Issue #695, opened 2021-07-30
Refactor `HTTPClient`


aboedo:
The current implementation of `HTTPClient` is mostly a direct port of its objective-c counterpart, but its logic is somewhat contrived, and it's not actually taking advantage of the two dispatchqueues it's managing. 

This entire class could be refactored so that: 
- it uses the corresponding queue to send stuff concurrently or serially
- its logic is simplified by leveraging the queue to, well, queue tasks, as oppposed to managing _another_ queue of tasks manually

PR #1427, opened 2022-03-29
Created `ErrorResponse` to abstract error deserialization

NachoSoto:
This is duplicated in most of the `NetworkOperation`s. This new shared implementation will be used by `HTTPClient`.

For #695.
See #1429 for how this is used.

### DIFF
--- a/Sources/Networking/HTTPResponse.swift
+++ b/Sources/Networking/HTTPResponse.swift
@@ -30,3 +30,70 @@ extension HTTPResponse: CustomStringConvertible {
     }
 
 }
+
+// MARK: -
+
+/// The response content of a failed request.
+struct ErrorResponse {
+
+    let code: BackendErrorCode
+    let message: String?
+    let attributeErrors: [String: String]
+
+}
+
+extension ErrorResponse {
+
+    /// Converts this `ErrorResponse` into an `ErrorCode` backed by the corresponding `BackendErrorCode`.
+    func asBackendError(with statusCode: HTTPStatusCode) -> Error {
+        var userInfo: [NSError.UserInfoKey: Any] = [
+            ErrorDetails.finishableKey: !statusCode.isServerError,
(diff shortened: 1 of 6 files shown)


Sources/Networking/Operations/Handling/UserInfoAttributeParser.swift
```diff
@@ -17,11 +17,8 @@ enum UserInfoAttributeParser {
 
     static func attributesUserInfoFromResponse(response: [String: Any], statusCode: HTTPStatusCode) -> [String: Any] {
```
NachoSoto, line 18: This whole class will go away soon, will be superseded by the new `ErrorResponse`.

Sources/Networking/HTTPResponse.swift
```diff
@@ -30,3 +30,70 @@ extension HTTPResponse: CustomStringConvertible {
     }
 
 }
+
+// MARK: -
+
+/// The response content of a failed request.
+struct ErrorResponse {
```
aboedo, line 37: nice!!! this will also help us get rid of some hacky stuff we have like the `finishableKey` and `RCSuccessfullySyncedKey`, which were just awkward ways to communicate information through the same existing dictionary back to other layers

Sources/Networking/HTTPStatusCode.swift
```diff
@@ -83,4 +83,8 @@ extension HTTPStatusCode {
         return 500...599 ~= self.rawValue
     }
 
+    var isSuccessfullySynced: Bool {
+        return !(self.isServerError || self == .notFoundError)
+    }
```
aboedo, line 88: ❤️

aboedo (approved):

Sources/Networking/HTTPResponse.swift
NachoSoto: 👍🏻

PR #1429, opened 2022-03-30
Replaced manual error response deserialization with `ErrorResponse`

NachoSoto:
Follow up to #1427.

The new way to create errors is like this:
```swift
ErrorResponse
    .from(response)
    .asBackendError(with: statusCode)
```

## Other changes:
- Removed `UserInfoAttributeParser`
- Simplified `JSONDecoder.decode` implementation by simply relying on `defaultJsonDecoder`
- Added support for "wrapped" error responses. This was previously handled by `UserInfoAttributeParser`.
- Removed all duplicated code parsing error responses
- The new `ErrorResponse` introduced in #1427 correctly decodes the format sent by the backend. Note that some tests had the wrong format, and `UserInfoAttributeParser` was parsing that wrong as well.

An upcoming PR will actually move these calls to `ErrorResponse` to `HTTPClient`.


Sources/FoundationExtensions/JSONDecoder+Extensions.swift
```diff
@@ -33,41 +33,63 @@ enum CodableError: Error, CustomStringConvertible {
...
         _ type: T.Type = T.self,
-        dictionary: [String: Any],
-        keyDecodingStrategy: KeyDecodingStrategy = .useDefaultKeys,
+        jsonData: Data,
         dateDecodingStrategy: DateDecodingStrategy = .deferredToDate,
-        dataDecodingStrategy: DataDecodingStrategy = .deferredToData
+        dataDecodingStrategy: DataDecodingStrategy = .deferredToData,
+        logErrors: Bool = true
     ) throws -> T {
-        self.keyDecodingStrategy = keyDecodingStrategy
```
NachoSoto, line 54: This `decode` method mutates the instance it's called on, which isn't a great practice. At least the `keyDecodingStrategy` is no longer changed and we rely on a single `defaultJsonDecoder` everywhere. I should look at the `date` and `data` decoding strategies, we might be able to just set them on `defaultJsonDecoder` too.

Sources/Networking/HTTPResponse.swift
```diff
@@ -90,3 +94,37 @@ extension ErrorResponse: Decodable {
     }
 
 }
+
+extension ErrorResponse {
+
+    /// For some endpoints the backend may return `ErrorResponse` inside of this wrapper.
+    private struct Wrapper: Decodable {
```
NachoSoto, line 101: This was previously handled by `UserInfoAttributeParser`.

Sources/FoundationExtensions/JSONDecoder+Extensions.swift
NachoSoto: Done: #1432

Sources/Networking/Operations/GetOfferingsOperation.swift
```diff
@@ -59,8 +59,9 @@ private extension GetOfferingsOperation {
                     return statusCode.isSuccessfulResponse
                     ? .success(response)
                     : .failure(
-                        ErrorUtils.backendError(withBackendCode: BackendErrorCode(code: response["code"]),
-                                                backendMessage: response["message"] as? String)
+                        ErrorResponse
```
taquitos, line 62: This is SO much better. I hacked on the errors a bunch just to get it working and surfacing what we wanted to surface, but this is a nice API 🧡

taquitos (approved):
YES. This is good.
yes-enthusiastically-yes

It's not Friday, yet, but I'm still dropping a gif.

PR #1425, opened 2022-03-29
`HTTPClient` now handles response deserialization + Introduced `NetworkError` and `BackendError`

NachoSoto:
Closes #695 and finishes [CF-195]. 
Depends on ~~#1431, #1432, #1433, #1437~~.

## Goals:

- [x] Handle all requests / response / deserialization errors in `HTTPClient`. Clients of `HTTPClient` will only have to handle the "happy" path: #1431
- [x] `HTTPClient` shouldn't return `Result.success` with failed responses, forcing clients to verify the response is actually a failure: https://github.com/RevenueCat/purchases-ios/pull/1431
- [x] Abstract error response deserialization / error creation: #1427
- [x] Abstract attribute error parsing: #1427 
- [x] Move response deserialization to `HTTPClient` based on a `ResponseType: Decodable` type, so the completion block will simply return a `Result<HTTPResponse<ResponseType>, Error>`
- [x] `ETagManager` should store response `Data` instead of a deserialized `[String: Any]`
- [x] Improved error types in `HTTPClient` to fix tests: #1437
- [x] Dealt with non-backwards compatible `ETagManager: #1438

## Changes:
- Replaced `HTTPResponse`'s body from `[String: Any]` to a generic `HTTPResponseBody`.
- Created `HTTPResponseBody` to abstract `Decodable` and provide some default implementations for types like `Data,` `[String: Any]` (for backwards compatibility to types that aren't `Decodable` yet), and `Decodable` itself.
- New `HTTPResponse.Result` typealias (`Result<HTTPResponse<HTTPResponseBody>, Error>`) used everywhere. This will allow replacing `Error` with a more specific `Error` so we can forward known typed errors, and make sure that we don't end up with the wrong error type, or with a very complex error hierarchy and the details buried in `underlyingError`.
- Each layer (only a few for now) has its own error type: `NetworkError`, `BackendError`, `OfferingsManager.Error`.
- `HTTPClient` for example has to produce `NetworkError`, operations produce `BackendError`
- The parent `BackendError` can have specific errors like `.missingAppUserID`, but also be simply a child error `case networkError(NetworkError)`
- All of these conform to a `ErrorCodeConvertible`, so there is a single point of code that converts from simple and readable errors (like `BackendError.emptySubscriberAttributes`, `.unexpectedBackendResponse(.loginResponseDecoding)`) into errors with all the context using `ErrorUtils`
- Converted `DNSError` into `NetworkError.dnsError`. Its functionality remains unchanged.
[...]

PR #1431, opened 2022-03-30
Moved response error and status checking to `HTTPClient`

NachoSoto:
Follow up to #1429. For #695.

This was duplicated in all operations.
`HTTPClient` now only forwards `Result.success` if the request actually succeeded,
leaving each `NetworkOperation` to just parse the body (which will become part of `HTTPClient` too).

Other changes:
- Simplified logic for when to retry requests in `HTTPClient` by using a new `Result.asOptionalResult`
- Removed several response handlers that don't need to do anything anymore
- Removed several tests that were basically duplicated:
  - The `finishable` checks are already tested as part of `ErrorResponse`.
  - All the checks for the error details are also part of `ErrorResponse`.
  - No longer need to have a test for each type of failed response since that's part of `HTTPClient` tests now.


Sources/Networking/HTTPClient.swift
```diff
@@ -219,55 +217,51 @@ private extension HTTPClient {
...
+                return .success(response)
             }
-        }()
-
-        return (
-            result: result
-                .flatMap {
-                    $0.map(Result.success)
-                        ?? .failure(ErrorUtils.unexpectedBackendResponseError())
-                },
```
NachoSoto, line 244: This is basically what `asOptionalResult` abstracts now.

Sources/Networking/HTTPClient.swift
```diff
@@ -219,55 +217,51 @@ private extension HTTPClient {
...
+            .asOptionalResult?
+            .mapError { ErrorUtils.networkError(withUnderlyingError: $0) }
+            .flatMap { response in
+                guard response.statusCode.isSuccessfulResponse else {
+                    return .failure(
+                        ErrorResponse
+                            .from(response.jsonObject)
+                            .asBackendError(with: statusCode)
+                    )
+                }
```
NachoSoto, line 237: These lines were in every single operation.

Sources/Networking/Operations/Handling/SubscriberAttributeHandler.swift
```diff
@@ -1,43 +0,0 @@
...
-//
-//      https://opensource.org/licenses/MIT
-//
-//  SubscriberAttributeHandler.swift
-//
-//  Created by Joshua Liebowitz on 11/18/21.
-
-import Foundation
-
-class SubscriberAttributeHandler {
```
NachoSoto, line 16: 👋🏻

Sources/Networking/Operations/Handling/NoContentResponseHandler.swift
```diff
@@ -1,39 +0,0 @@
...
-//
-//      https://opensource.org/licenses/MIT
-//
-//  NoContentResponseHandler.swift
-//
-//  Created by Joshua Liebowitz on 11/30/21.
-
-import Foundation
-
-class NoContentResponseHandler {
```
NachoSoto, line 16: 👋🏻

Sources/FoundationExtensions/Result+Extensions.swift
```diff
@@ -40,3 +40,21 @@ extension Result {
     }
 
 }
+
+extension Result where Success: OptionalType {
+
+    /// Converts a `Result<Success?, Error>` into `Result<Success, Error>?`
+    var asOptionalResult: Result<Success.Wrapped, Failure>? {
```
NachoSoto, line 47: I didn't write tests for this because the compiler is doing all the work checking the types. It's a simple type conversion with no added logic.

Sources/Networking/HTTPClient.swift
```diff
@@ -162,31 +162,29 @@ private extension HTTPClient {
         self.start(request: request)
     }
 
+    // - Returns: `nil` if the request must be retried
```
taquitos, line 165: This is cleaner 🧹

Sources/Networking/Operations/CreateAliasOperation.swift
```diff
@@ -16,15 +16,12 @@ import Foundation
 class CreateAliasOperation: CacheableNetworkOperation {
 
     private let aliasCallbackCache: CallbackCache<AliasCallback>
-    private let createAliasResponseHandler: NoContentResponseHandler
```
taquitos, line 19: 😻

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -18,51 +18,32 @@ class CustomerInfoResponseHandler {
...
-
-            let errorResponse = ErrorResponse.from(response)
-
-            var result: Result<CustomerInfo, Error> = {
-                // Only attempt to parse a response if we don't have an error status code from the backend.
-                if !isErrorStatusCode {
-                    return Result { try CustomerInfo.from(json: response) }
+        let errorResponse = ErrorResponse.from(response.value?.jsonObject ?? [:])
+
+        let result: Result<CustomerInfo, Error> = response
```
taquitos, line 24: Visually, this is more confusing to me. Is there a way to structure this so it's easier to parse? Maybe I'm the only one, but that was something I struggled with learning RxSwift/ReactiveCocoa. The many `.flatMap` after `.flatMap` with more nested function calls like `mapError` it's difficult for me to follow the flow of logic/transformations.

taquitos (commented):
This is a great improvement. I want to keep comments open for other folks to check it out. If people don't feel the same way I do about some of the confusing functional structure, I'm happy to approve 😄

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
NachoSoto: Yeah I brought up a similar comment [here](https://github.com/RevenueCat/purchases-ios/pull/1425#discussion_r839915462).
For this particular example this code is going away in that PR since the deserialization will be done by `HTTPClient`, so this is just temporary.

But the point is still valid, especially in that other PR.

NachoSoto:
This is the next PR that needs reviewing 🙏🏻

Sources/FoundationExtensions/Result+Extensions.swift
aboedo: I understand the point about the _current_ code being correct, but in practice, writing tests to check that the current code is correct is usually not that valuable in my experience. 
The real value in writing tests is future-proofing, so that if in the future another dev wants to modify or refactor this code, they can do so while resting assured that as long as tests pass, the code is still working correctly. 

Say that someone was debugging something, and in order to quickly test out what happens in the empty success case, they comment or remove the code in this implementation and replace it with a simple 

```swift
var asOptionalResult: Result<Success.Wrapped, Failure>? {
    return nil
}
```

Then they forget about it and never roll it back. Hopefully it would get caught in a PR, but if tests are passing, it might not be caught if it doesn't draw attention. 

If there are tests for this, it'll be easy to catch. Even if it doesn't make it to a full PR because other code is actually using the value and so the tests for that code fail, the tests will no longer be specific, and it might be hard to debug. 

So I'd always argue for making tests, even if the current code is guaranteed to be correct, in order to future-proof the code and ensure that other devs won't be able to accidentally break it.

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
aboedo: +1 on the functional code being tougher to understand. We can continue discussion on the other PR if this is temporary

aboedo:
love seeing so much code go away 🪄

Sources/FoundationExtensions/Result+Extensions.swift
NachoSoto: Always a great point that I keep forgetting. Added tests!

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
NachoSoto: Yeah this is gone later since all the parsing will be done by `HTTPClient`, but welcoming feedback on the other one.

Tests/UnitTests/FoundationExtensions/ResultExtensionsTests.swift
```diff
@@ -48,6 +48,25 @@ class ResultExtensionsTests: XCTestCase {
...
+    func testWithData() {
+        expect(Data.success("test").asOptionalResult) == OptionalData.some(.success("test"))
+    }
+
+    func testWithNoData() {
+        expect(Data.success(.none).asOptionalResult).to(beNil())
+    }
+
+    func testWithError() {
+        expect(Data.failure(.error1).asOptionalResult) == OptionalData.some(.failure(.error1))
```
NachoSoto, line 65: Testing with explicit types as well.

aboedo (approved):

PR #1432, opened 2022-03-30
Simplified `JSONDecoder.decode` decoding strategies

NachoSoto:
Follow up to #1429. The `decode` methods no longer mutate the underlying `JSONDecoder`, and instead the codebase uses a shared `defaultJSONDecoder` with the correct configuration.


taquitos (approved):
📅 🎈🐐

PR #1434, opened 2022-03-31
`NSData.asString()`: optimized string creation

NachoSoto:
I noticed that the existing implementation was creating many copies of the string as it grows, which isn't very efficient.
The new implementation allocates the required size initially and writes to the same buffer inside the loop.


taquitos (approved):
✂️ 🧵 🐐

PR #1436, opened 2022-03-31
`MockUserDefaults`: override `synchronize`

NachoSoto:
Right now the original `UserDefaults.synchronize` is being called which isn't great.


taquitos:
> Right now the original `UserDefaults.synchronize` is being called which isn't great.

It's not great, but it was for backwards compatibility < iOS 12
I think we can now remove this since we only target iOS 12 and above now.

NachoSoto:
This is the mock 😅 my point is that when mocking user defaults, calling `synchronize` was synchronizing the real user defaults (the parent class).

PR #1438, opened 2022-04-01
`ETagManager`: added test to verify that reading cached data in old format fails gracefully

NachoSoto:
Follow up to #1425.


taquitos (approved):
Fail

PR #1439, opened 2022-04-01
[wip] renaming confusing methods in the promoted purchase flow

aboedo:
This is just a wip for now, trying to illustrate how these methods could be named.


Sources/Purchasing/Purchases.swift
```diff
@@ -1784,15 +1784,19 @@ extension Purchases: PurchasesOrchestratorDelegate {
...
      * - Parameter product: ``StoreProduct`` the product that was selected from the app store.
+     * - Parameter startPromotedPurchaseBlock: Method that begins the purchase flow for the promoted purchase.
+     * If the app is ready to start the purchase flow when this delegate method is called, then this method
+     * should be called right away. Otherwise, the method should be stored as a property in memory, and then called
+     * once the app is ready to start the purchase flow. 
      */
     @objc
-    public func shouldPurchasePromoProduct(_ product: StoreProduct,
-                                           defermentBlock: @escaping DeferredPromotionalPurchaseBlock) {
+    func promotedPurchaseReadyToStart(for product: StoreProduct,
```
NachoSoto, line 1793: Yeah this didn't make sense to be exposed 👍🏻

Sources/Misc/Deprecations.swift
```diff
@@ -56,6 +56,14 @@ public extension Purchases {
         return await eligiblePromotionalOffers(forProduct: product)
     }
 
+//    @objc func shouldPurchasePromoProduct(_ product: StoreProduct,
```
NachoSoto, line 59: I'm actually gonna mark this as obsoleted, I don't think it made sense for users to call it anyway, it was more of a programmer error, so better to fail. Breaking compilation in this case I think is warranted.

PR #1441, opened 2022-04-01
[WIP] Created `SubscriberAttributesManagerIntegrationTests`

NachoSoto:
I made a lot of changes to how subscriber attributes are synced and the error handling, but we had no test coverage for it other than unit tests. None of the `SubscriberAttributesManager` have completion blocks so we never check if calling `setEmail` etc. actually works in production AFAIK.

This adds coverage for that.


Tests/BackendIntegrationTests/SubscriberAttributesManagerIntegrationTests.swift
```diff
@@ -0,0 +1,157 @@
...
+
+        let errors = await self.syncAttributes()
+        expect(errors).to(haveCount(1))
+
+        let error = try XCTUnwrap(errors.first ?? nil) as NSError
+        expect(error.domain) == RCPurchasesErrorCodeDomain
+        expect(error.code) == ErrorCode.invalidSubscriberAttributesError.rawValue
+        expect(error.subscriberAttributesErrors) == [
+            "$email": "Email address is not a valid email."
+        ]
```
NachoSoto, line 67: This is a failing test that should be fixed by #1429.

NachoSoto:
This is ready now 🎉 I added a bunch of tests to verify multiple scenarios, but we can extend it in the future.

Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
```diff
@@ -51,6 +51,12 @@ class BaseBackendIntegrationTests: XCTestCase {
         configurePurchases()
     }
 
+    override func tearDown() {
+        Purchases.clearSingleton()
```
NachoSoto, line 55: `Purchases.setDefaultInstance` does:
```swift
if isConfigured {
    Logger.info(Strings.configure.purchase_instance_already_set)
}
```
Which happens on every test. Now all integration tests clear the instance at the end.

Tests/BackendIntegrationTests/SubscriberAttributesManagerIntegrationTests.swift
```diff
@@ -0,0 +1,157 @@
...
+
+        let errors = await self.syncAttributes()
+        verifyAttributesSyncedWithNoErrors(errors, 2)
+    }
+
+}
+
+private extension SubscriberAttributesManagerIntegrationTests {
+
+    func syncAttributes() async -> [Error?] {
```
NachoSoto, line 124: `async` makes these tests so much better.

Tests/BackendIntegrationTests/SubscriberAttributesManagerIntegrationTests.swift
NachoSoto: Confirmed, this test will pass once `ErrorResponse` fixes attribute error parsing.

Tests/BackendIntegrationTests/SubscriberAttributesManagerIntegrationTests.swift
taquitos: Yeah, reading and making sense of the tests is so easy with out needing to build special block handling/waiting

taquitos (approved):
🕺

PR #1442, opened 2022-04-01
Changed `SystemInfo.isApplicationBackgrounded` to use `OperationDispatcher`

NachoSoto:
This solves a race condition that could cause operations to run out of order (discovered in #1441).

`Purchases.configure` triggers a `getCustomerInfo` call, which relies on `isApplicationBackgrounded`.
The previous implementation always had a runloop "jump", meaning that sometimes a subsequent request might happen out of order.
By using `OperationDispatcher.dispatchOnMainThread`, the jump is avoided if we're already in the main thread, meaning we run everything deterministically in order.

### DIFF
--- a/Sources/Misc/OperationDispatcher.swift
+++ b/Sources/Misc/OperationDispatcher.swift
@@ -20,6 +20,8 @@ class OperationDispatcher {
     private let workerQueue = DispatchQueue(label: "OperationDispatcherWorkerQueue")
     private let maxJitterInSeconds: Double = 5
 
+    static let `default`: OperationDispatcher = .init()
+
     func dispatchOnMainThread(_ block: @escaping () -> Void) {
         if Thread.isMainThread {
             block()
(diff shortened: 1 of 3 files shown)


Sources/Misc/OperationDispatcher.swift
```diff
@@ -20,6 +20,8 @@ class OperationDispatcher {
     private let workerQueue = DispatchQueue(label: "OperationDispatcherWorkerQueue")
     private let maxJitterInSeconds: Double = 5
 
+    static let main: OperationDispatcher = .init()
```
NachoSoto, line 23: The only reason the methods in this class aren't `static` is so we can mock it, but there's no reason why we can't share it.

Sources/Purchasing/Purchases.swift
```diff
@@ -245,14 +245,15 @@ public typealias DeferredPromotionalPurchaseBlock = (@escaping PurchaseCompleted
                      platformInfo: PlatformInfo? = Purchases.platformInfo,
                      useStoreKit2IfAvailable: Bool = false,
                      dangerousSettings: DangerousSettings? = nil) {
-        let operationDispatcher = OperationDispatcher()
+        let operationDispatcher = OperationDispatcher.main
```
aboedo, line 248: ❤️

aboedo (approved):

NachoSoto:
Changed to `default` because `main` might be confused with `DispatchQueue.main`.

taquitos (approved):
+1 for default 😄

PR #1444, opened 2022-04-01
Added `BackendErrorCode.subscriptionNotFoundForCustomer`

NachoSoto:
I was debugging this, so I guess it makes sense to at least add it to the `enum` even if we don't do anything special with it.


taquitos (approved):
🈴

PR #1446, opened 2022-04-05
Fix: local receipt parsing for StoreKitTest - part 1

aboedo:
This PR fixes local receipt parsing for StoreKitTest receipts. 

So far this only includes a single test, next PR is going to add more tests for different StoreKitTest receipts, and will also clean up the test code to make it a lot nicer. 

So this is part 1, but it does the main thing, which is to make it actually compatible. 

Main changes: 

### StoreKitTest receipts reported length value: 
These receipts use indefinite-length containers, so our calculation for the container length would break, because it would be understood to be zero. 
This first (naive) implementation assumes the indefinite-length containers to be as big as the entire payload, which is incorrect, but works in practice, so it's a decent temporary patch. 

### StoreKitTest receipts missing originalAppVersion
The [`originalAppVersion`](https://developer.apple.com/library/archive/releasenotes/General/ValidateAppStoreReceipt/Chapters/ReceiptFields.html) is non-optional for production and sandbox receipts, but the value is missing for StoreKitTest receipts, so this PR makes it optional. 

### StoreKitTest receipt structure
The data in StoreKItTest receipts is structured slightly differently to regular: 

Regular receipts: 
- Sequence
    - Main object (data object identifier)
    - Data
        - Sequence with single element
            - octet string
                - **Actual receipt data**

StoreKitTest receipts: 

- Sequence
    - Main object (data object identifier)
    - Data
        - Sequence with single element
            - octet string
                - **another octet string**
                    - **Actual receipt data**

I could of course not find any documentation for any of the changes above, this is just reverse-engineering. 

Follow-up PRs will: 
- Clean up tests code and make it a ton simpler (this is just a terrible and hacky copy-paste)
- Add more tests
- Fix the indefinite-length value calculation logic, to get the correct value instead of the naively-assumed value of "the entire payload"

### DIFF
--- a/Sources/LocalReceiptParsing/BasicTypes/AppleReceipt.swift
+++ b/Sources/LocalReceiptParsing/BasicTypes/AppleReceipt.swift
@@ -40,7 +40,7 @@ struct AppleReceipt: Equatable {
 
     let bundleId: String
     let applicationVersion: String
-    let originalApplicationVersion: String
+    let originalApplicationVersion: String?
     let opaqueValue: Data
     let sha1Hash: Data
     let creationDate: Date
@@ -58,7 +58,7 @@ struct AppleReceipt: Equatable {
         return [
             "bundleId": bundleId,
             "applicationVersion": applicationVersion,
-            "originalApplicationVersion": originalApplicationVersion,
+            "originalApplicationVersion": originalApplicationVersion ?? "<unknown>",
             "opaqueValue": opaqueValue,
             "sha1Hash": sha1Hash,
             "creationDate": creationDate,
--- a/Sources/LocalReceiptParsing/Builders/AppleReceiptBuilder.swift
+++ b/Sources/LocalReceiptParsing/Builders/AppleReceiptBuilder.swift
@@ -44,9 +44,18 @@ class AppleReceiptBuilder {
         guard let internalContainer = container.internalContainers.first else {
             throw ReceiptReadingError.receiptParsingError
         }
-        let receiptContainer = try containerBuilder.build(fromPayload: internalContainer.internalPayload)
-        let internalReceiptContainer = try containerBuilder.build(fromPayload: receiptContainer.internalPayload)
-        for receiptAttribute in internalReceiptContainer.internalContainers {
+        var receiptContainer = try containerBuilder.build(fromPayload: internalContainer.internalPayload)
+
+        // StoreKitTest receipts have their data embedded into 2 levels of octetString containers,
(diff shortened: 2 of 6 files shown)


Sources/LocalReceiptParsing/BasicTypes/AppleReceipt.swift
```diff
@@ -40,7 +40,7 @@ struct AppleReceipt: Equatable {
 
     let bundleId: String
     let applicationVersion: String
-    let originalApplicationVersion: String
+    let originalApplicationVersion: String?
```
aboedo, line 43: See PR description, `originalApplicationVersion` is apparently not reported in SKTest receipts

Sources/LocalReceiptParsing/Builders/AppleReceiptBuilder.swift
```diff
@@ -44,7 +44,17 @@ class AppleReceiptBuilder {
...
+        var receiptContainer = try containerBuilder.build(fromPayload: internalContainer.internalPayload)
+
+        // StoreKitTest receipts have their data embedded into 2 levels of octetString containers,
+        // Regular receipts have it in only one. At this point we've already unwrapped the upper level
+        // so we check whether we need to go one deeper.
+        let isStoreKitTestReceipt = receiptContainer.encodingType == .primitive
+                                    && receiptContainer.containerIdentifier == .octetString
+        if isStoreKitTestReceipt {
+            receiptContainer = try containerBuilder.build(fromPayload: receiptContainer.internalPayload)
+        }
```
aboedo, line 56: This is illustrated better in PR description

Tests/StoreKitUnitTests/LocalReceiptParserStoreKitTests.swift
```diff
@@ -0,0 +1,57 @@
...
+                completionCalled = true
+            } catch {
+                print("failed to parse. Error: \(error)")
+                completionCalled = true
+            }
+        }
+
+        expect(completionCalled).toEventually(beTrue())
+        let receipt = try XCTUnwrap(maybeReceipt)
+        expect(receipt.applicationVersion) == "1"
```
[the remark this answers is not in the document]
NachoSoto: 👍🏻

Sources/LocalReceiptParsing/BasicTypes/AppleReceipt.swift
aboedo: yeah exploring enforcing with the compiler is a good idea. I'm going to add that to the list of things to do in the follow-ups

PR #965, opened 2021-11-18
Detect ErrorCode.productAlreadyPurchasedError when SKError.unknown is actually cased by it

NachoSoto:
Fixes #392.

As per the conversation in that issue, this introspects and checks for the undocumented error domain and code.

PR #1450, opened 2022-04-05
Added workaround for StoreKit 1 incorrectly reporting purchase cancellations

### LINKED ISSUE #1445: `Purchases.shared.purchase(package: ) async` not being called when the user cancels
- [✅] I have updated Purchases SDK to the latest version
- [✅] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/CONTRIBUTING.md)
- [✅] I have searched the [Community](https://community.revenuecat.com)
- [✅] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [✅] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**

While doing a test purchase, I noticed that when cancelling the purchase, the Purchases.shared.purchase(package:) async does not return PurchaseResultData.userCancelled. It simply does nothing. 

Making the purchase works as expected, but not when the user cancels.

Link to the method being used.
[https://revenuecat-docs.netlify.app/documentation/revenuecat/purchases/purchase(package:)](url)

1. Environment
   1. Platform: Apple/iOS
   2. SDK version: 4.1.0
   3. OS version: 
   4. Xcode version: 13.3
   5. How widespread is the issue. Percentage of devices affected.

**Additional context**

Screen Shot 2022-04-05 at 2 14 08 AM

NachoSoto:
Fixes #1445.
Similar solution to #965, this attempts to introspect the unknown `SKError` to detect if it actually corresponds to a failed purchase.

### DIFF
--- a/Sources/Error Handling/SKError+Extensions.swift
+++ b/Sources/Error Handling/SKError+Extensions.swift
@@ -49,10 +49,14 @@ extension SKError {
             return ErrorUtils.invalidPromotionalOfferError(error: self)
         case .unknown:
             if let error = self.userInfo[NSUnderlyingErrorKey] as? NSError {
-                switch (error.domain, error.code) {
-                case ("ASDServerErrorDomain", 3532): // "You’re currently subscribed to this"
-                    // See https://github.com/RevenueCat/purchases-ios/issues/392
-                    return ErrorUtils.productAlreadyPurchasedError(error: self)
+                switch error.domain {
+                case ASDServerError.domain:
+                    switch ASDServerError.Code(rawValue: error.code) {
+                    case .currentlySubscribed:
+                        return ErrorUtils.productAlreadyPurchasedError(error: self)
+
+                    default: break
+                    }
 
                 default: break
                 }
@@ -61,13 +65,19 @@ extension SKError {
             return ErrorUtils.storeProblemError(error: self)
 
         @unknown default:
-            switch self.code.rawValue {
-            case 907: // "Unhandled exception"
+            switch SKError.UndocumentedCode(rawValue: self.code.rawValue) {
+            case .unhandledException:
                 if let error = self.userInfo[NSUnderlyingErrorKey] as? NSError {
-                    switch (error.domain, error.code) {
-                    case ("AMSErrorDomain", 6): // "Payment Sheet Failed"
-                        // See https://github.com/RevenueCat/purchases-ios/issues/1445
-                        return .purchaseCancelledError
+                    switch error.domain {
+                    case AMSError.domain:
+                        switch AMSError.Code(rawValue: error.code) {
+                            // See https://github.com/RevenueCat/purchases-ios/issues/1445
+                            // Cancellations sometimes show up as undocumented errors instead of regular cancellations
+                        case .paymentSheetFailed:
+                            return ErrorUtils.purchaseCancelledError(error: self)
+
+                        default: break
+                        }
 
                     default: break
                     }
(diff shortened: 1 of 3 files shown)


Tests/UnitTests/Purchasing/PurchasesTests.swift
```diff
@@ -2049,12 +2083,14 @@ class PurchasesTests: XCTestCase {
...
-        expect(receivedUnderlyingError?.domain).toEventually(equal(SKErrorDomain))
-        expect(receivedUnderlyingError?.code).toEventually(equal(SKError.Code.paymentCancelled.rawValue))
+        expect(receivedUserCancelled).toEventuallyNot(beNil())
+
+        expect(receivedUserCancelled) == true
+        expect(receivedError).toNot(beNil())
+        expect(receivedError?.domain) == RCPurchasesErrorCodeDomain
+        expect(receivedError?.code) == ErrorCode.purchaseCancelledError.rawValue
+        expect(receivedUnderlyingError?.domain) == SKErrorDomain
+        expect(receivedUnderlyingError?.code) == SKError.Code.paymentCancelled.rawValue
```
NachoSoto, line 2093: I've changed this to a single `toEventually`, that only checks that the value was actually set.
The rest aren't `toEventually` because if the test fails, it would be busy waiting for the condition to change, which it never would.

NachoSoto:
Test failure fixed by #1451.

Sources/Error Handling/SKError+Extensions.swift
```diff
@@ -61,6 +61,21 @@ extension SKError {
             return ErrorUtils.storeProblemError(error: self)
 
         @unknown default:
+            switch self.code.rawValue {
+            case 907: // "Unhandled exception"
```
taquitos, line 65: I hate we have to do this, but it's better us than our developers 😢

aboedo, line 65: Maybe we could create a constant for the magic number so the comment isn't needed? 

i.e.: 
```swift
switch self.code.rawValue {
    case .unhandledException: ...
```

taquitos (approved):
😄🐐

Sources/Error Handling/SKError+Extensions.swift
```diff
@@ -61,6 +61,21 @@ extension SKError {
...
+                if let error = self.userInfo[NSUnderlyingErrorKey] as? NSError {
+                    switch (error.domain, error.code) {
+                    case ("AMSErrorDomain", 6): // "Payment Sheet Failed"
```
aboedo, line 68: same thought here regarding coding in the constants

aboedo: maybe we can add a comment saying something like "cancellations sometimes show up as undocumented errors instead of regular cancellations"

aboedo (approved):
looks great! Thanks for taking care of this!

Sources/Error Handling/SKError+Extensions.swift
NachoSoto: Yup great idea

Sources/Error Handling/SKError+Extensions.swift
```diff
@@ -61,8 +65,66 @@ extension SKError {
...
+    static let domain = "AMSErrorDomain"
+
+    enum Code: Int {
+
+        // See https://github.com/RevenueCat/purchases-ios/issues/1445
+        case paymentSheetFailed = 6
+
+    }
+
+}
```
NachoSoto, line 130: I created these to better maintain these undocumented errors. Thanks for the suggestion @aboedo

PR #1452, opened 2022-04-05
Fix: ASN1 length calculation for indefinite-length

aboedo:
Follow-up to #1446. 

This one fixes the length calculation logic for indefinite-length receipts. 
In this type of receipt, the length bytes are set to a value of zero, and you find the actual length by finding a container with `objectIdentifier` = `endOfContent` (two bytes of 0x00 value). 

So the new logic for indefinite-length containers parses the inner containers until an endOfContent container is found, then sums up the length of all of them and uses that value. 

Definite-length receipts don't actually have `endOfContent` bytes, which makes the parsing logic slightly more complex, sadly.

### DIFF
--- a/Sources/LocalReceiptParsing/BasicTypes/ASN1Container.swift
+++ b/Sources/LocalReceiptParsing/BasicTypes/ASN1Container.swift
@@ -64,6 +64,14 @@ struct ASN1Length: Equatable {
 
     let value: Int
     let bytesUsedForLength: Int
+    let definition: LengthDefinition
+
+    enum LengthDefinition: Int {
+
+        case definite
+        case indefinite
+
+    }
 
 }
 
--- a/Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
+++ b/Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
@@ -24,8 +24,9 @@ class ASN1ContainerBuilder {
         let containerClass = try extractClass(byte: firstByte)
         let encodingType = try extractEncodingType(byte: firstByte)
         let containerIdentifier = try extractIdentifier(byte: firstByte)
+        let isConstructed = encodingType == .constructed
         let (length, internalContainers) = try extractLengthAndInternalContainers(data: payload.dropFirst(),
-                                                                                  isConstructed: encodingType == .constructed)
+                                                                                  isConstructed: isConstructed)
         let bytesUsedForIdentifier = 1
         let bytesUsedForMetadata = bytesUsedForIdentifier + length.bytesUsedForLength
 
--- a/Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
+++ b/Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
@@ -24,7 +24,7 @@ class ASN1ContainerBuilder {
         let containerClass = try extractClass(byte: firstByte)
         let encodingType = try extractEncodingType(byte: firstByte)
         let containerIdentifier = try extractIdentifier(byte: firstByte)
-        let length = try extractLength(data: payload.dropFirst())
+        let length = try extractLength(data: payload.dropFirst(), isConstructed: encodingType == .constructed)
         let bytesUsedForIdentifier = 1
         let bytesUsedForMetadata = bytesUsedForIdentifier + length.bytesUsedForLength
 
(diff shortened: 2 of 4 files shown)


Sources/LocalReceiptParsing/BasicTypes/ASN1Container.swift
```diff
@@ -64,6 +64,14 @@ struct ASN1Length: Equatable {
...
     let value: Int
     let bytesUsedForLength: Int
+    let definition: LengthDefinition
+
+    enum LengthDefinition: Int {
+
+        case definite
+        case indefinite
+
+    }
```
aboedo, line 74: this isn't really used, but it's useful when debugging

Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
```diff
@@ -55,6 +51,9 @@ private extension ASN1ContainerBuilder {
         while currentPayload.count > 0 {
             let internalContainer = try build(fromPayload: currentPayload)
             internalContainers.append(internalContainer)
+            if internalContainer.containerIdentifier == .endOfContent {
+                break
+            }
```
aboedo, line 56: this makes sense as a general thing, for all containers. But only indefinite-length containers seem to have this
¯\\_(ツ)_/¯

In any case, I don't think we need to check whether the length is indefinite, since it still makes sense to stop if you reach end of content

Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
```diff
@@ -84,7 +83,8 @@ private extension ASN1ContainerBuilder {
         return asn1Identifier
     }
 
-    func extractLength(data: ArraySlice<UInt8>) throws -> ASN1Length {
+    func extractLengthAndInternalContainers(data: ArraySlice<UInt8>,
+                                            isConstructed: Bool) throws -> (ASN1Length, [ASN1Container]) {
```
aboedo, line 87: this is awkward and I don't love it, but I couldn't separate the two steps, because: 
- for definite-length receipts, internal containers are determined by the receipt length
- for indefinite-length receipts, the actual length is determined by the internal containers' length
- I didn't want to generate internal containers twice

Alternatively I could have had the same old logic, but with a special case for indefinite receipts, where we initially set the length to zero, and then after building internal containers, create a new instance of ASN1Length with the correct value. 

That felt confusing and odd so I figured it's somewhat better to just determine both in a single step

Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
```diff
@@ -105,16 +105,28 @@ private extension ASN1ContainerBuilder {
...
+        } else if isConstructed {
+            let innerContainerData = data.dropFirst(bytesUsedForLength).prefix(lengthValue)
+            innerContainers = try buildInternalContainers(payload: innerContainerData)
         }
-        return ASN1Length(value: lengthValue, bytesUsedForLength: bytesUsedForLength)
+        let length = ASN1Length(value: lengthValue,
+                                bytesUsedForLength: bytesUsedForLength,
+                                definition: lengthDefinition)
+
+        return (length, innerContainers)
```
aboedo, line 129: tests for this coming in as a separate PR so that I can at least have code that I'm not uneasy about in `main`, this code does get the length calculation correctly

Sources/LocalReceiptParsing/BasicTypes/ASN1Container.swift
NachoSoto: I like it

Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
```diff
@@ -105,16 +106,28 @@ private extension ASN1ContainerBuilder {
...
         // This is called indefinite-length in ASN1 containers.
-        // When length == 0, the element's contents end when there are two subsequent 0x00 octets.
-        // This (naive) implementation just assumes that the end of content octets are at the end of the
-        // sequence, which is incorrect but works in practice.
-        // This code should be refactored to find the end of content octets, though. 
-        if lengthValue == 0 {
-            lengthValue = data.count - bytesUsedForLength
+        // When length == 0, the element's contents end when there's a container with .endOfContent identifier
+        // To get the length, we build the internal containers until we run into .endOfContent and sum up the bytes used
+        let lengthDefinition: ASN1Length.LengthDefinition = (isConstructed && lengthValue == 0)
```
NachoSoto, line 115: This is great 👍🏻

NachoSoto (approved):

PR #1454, opened 2022-04-05
length calculation tests

aboedo:
Follow-up to #1452, adds a couple of unit tests to ensure that we're getting the value correctly


aboedo:
This is the base PR that needs review 🙏

NachoSoto (approved):

PR #1456, opened 2022-04-05
More tests for receipt parsing with storekit test

aboedo:
Last item in the series, adds a couple of tests that ensure that we at least cover all of the fields in a receipt. 
There's still more than can be done in terms of messing around with purchases and see how the fields evolve, but I think this serves as a decent base, and we'll at least be able to use it to keep track of whether the parser is compatible with SKTest receipts.


NachoSoto (approved):
Awesome 🎉

PR #1460, opened 2022-04-06
Renamed `PurchasesDelegate.purchases(_:shouldPurchasePromoProduct:defermentBlock:)`

NachoSoto:
Finishing #1439.
Fixes [CF-464].

The goal of this change is to make the behavior and usage of this API more clear.

### DIFF
--- a/Sources/Misc/Obsoletions.swift
+++ b/Sources/Misc/Obsoletions.swift
@@ -17,7 +17,7 @@ import StoreKit
 // and therefore the `fatalError`s are unreachable.
 // See also: Development/Deprecations.md
 
-// swiftlint:disable file_length missing_docs
+// swiftlint:disable file_length missing_docs line_length
 
 public extension Purchases {
 
@@ -467,6 +467,16 @@ public extension Purchases {
                message: "Check eligibility for a discount using getPromotionalOffer:")
     func paymentDiscount(for discount: SKProductDiscount,
                          product: SKProduct) async throws -> SKPaymentDiscount {
+        fatalError()
+    }
+
+    @available(iOS, obsoleted: 1, message: "Use PurchasesDelegate.purchases(_:isReadyForPromotedProduct:purchase:)")
+    @available(tvOS, obsoleted: 1, message: "Use PurchasesDelegate.purchases(_:isReadyForPromotedProduct:purchase:)")
+    @available(watchOS, obsoleted: 1, message: "Use PurchasesDelegate.purchases(_:isReadyForPromotedProduct:purchase:)")
+    @available(macOS, obsoleted: 1, message: "Use PurchasesDelegate.purchases(_:isReadyForPromotedProduct:purchase:)")
+    @available(macCatalyst, obsoleted: 1, message: "Use PurchasesDelegate.purchases(_:isReadyForPromotedProduct:purchase:)")
+    @objc func shouldPurchasePromoProduct(_ product: StoreProduct,
+                                          defermentBlock: @escaping DeferredPromotionalPurchaseBlock) {
         fatalError()
     }
 
--- a/Sources/Purchasing/Purchases.swift
+++ b/Sources/Purchasing/Purchases.swift
@@ -32,7 +32,7 @@ public typealias PurchaseResultData = (transaction: StoreTransaction?,
 public typealias PurchaseCompletedBlock = (StoreTransaction?, CustomerInfo?, Error?, Bool) -> Void
 
 /**
- Deferred block for ``Purchases/shouldPurchasePromoProduct(_:defermentBlock:)``
+ Deferred block for ``PurchasesDelegate/purchases(_:isReadyForPromotedProduct:purchase:)``
  */
 public typealias DeferredPromotionalPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
 
@@ -1776,24 +1776,24 @@ extension Purchases: PurchasesOrchestratorDelegate {
     /**
      * Called when a user initiates a promotional in-app purchase from the App Store.
      *
-     * If your app is able to handle a purchase at the current time, run the deferment block in this method.
+     * If your app is able to handle a purchase at the current time, run the `startPurchase` block.
      *
-     * If the app is not in a state to make a purchase: cache the defermentBlock, then call the defermentBlock
+     * If the app is not in a state to make a purchase: cache the `startPurchase` block, then call it
      * when the app is ready to make the promotional purchase.
      *
-     * If the purchase should never be made, you don't need to ever call the defermentBlock and ``Purchases``
-     * will not proceed with promotional purchases.
+     * If the purchase should never be made, you don't need to ever call the `starrtPurchase` block
+     * and ``Purchases`` will not proceed with promotional purchases.
      *
      * - Parameter product: ``StoreProduct`` the product that was selected from the app store.
+     * - Parameter startPurchase: Method that begins the purchase flow for the promoted purchase.
+     * If the app is ready to start the purchase flow when this delegate method is called, then this method
+     * should be called right away. Otherwise, the method should be stored as a property in memory, and then called
+     * once the app is ready to start the purchase flow. 
      */
     @objc
-    public func shouldPurchasePromoProduct(_ product: StoreProduct,
-                                           defermentBlock: @escaping DeferredPromotionalPurchaseBlock) {
-        guard let delegate = delegate else {
-            return
-        }
-
-        delegate.purchases?(self, shouldPurchasePromoProduct: product, defermentBlock: defermentBlock)
+    internal func promotedPurchaseReadyToStart(for product: StoreProduct,
+                                               startPurchase: @escaping DeferredPromotionalPurchaseBlock) {
+        self.delegate?.purchases?(self, isReadyForPromotedProduct: product, purchase: startPurchase)
     }
 
 }
--- a/Sources/Purchasing/PurchasesDelegate.swift
+++ b/Sources/Purchasing/PurchasesDelegate.swift
@@ -47,13 +47,24 @@ import Foundation
     /**
      * Called when a user initiates a promotional in-app purchase from the App Store.
      * If your app is able to handle a purchase at the current time, run the deferment block in this method.
-     * If the app is not in a state to make a purchase: cache the defermentBlock,
-     * then call the defermentBlock when the app is ready to make the promotional purchase.
-     * If the purchase should never be made, you don't need to ever call the defermentBlock and
-     * ``Purchases`` will not proceed with promotional purchases.
-
+     * If the app is not in a state to make a purchase: cache the `startPurchase` block,
+     * then call the `startPurchase` block when the app is ready to make the promotional purchase.
+     *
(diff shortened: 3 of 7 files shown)


Sources/Misc/Obsoletions.swift
```diff
@@ -470,6 +470,16 @@ public extension Purchases {
         fatalError()
     }
 
+    @available(iOS, obsoleted: 1, message: "Use PurchasesDelegate.purchases(_:isReadyForPromotedProduct:purchase:)")
+    @available(tvOS, obsoleted: 1, message: "Use PurchasesDelegate.purchases(_:isReadyForPromotedProduct:purchase:)")
+    @available(watchOS, obsoleted: 1, message: "Use PurchasesDelegate.purchases(_:isReadyForPromotedProduct:purchase:)")
+    @available(macOS, obsoleted: 1, message: "Use PurchasesDelegate.purchases(_:isReadyForPromotedProduct:purchase:)")
+    @available(macCatalyst, obsoleted: 1, message: "Use PurchasesDelegate.purchases(_:isReadyForPromotedProduct:purchase:)")
+    @objc func shouldPurchasePromoProduct(_ product: StoreProduct,
```
NachoSoto, line 478: I marked this obsoleted instead of deprecated. I don't think it made sense for users to call it anyway, it was more of a programmer error, so better to fail. Breaking compilation in this case I think is warranted.

Sources/Purchasing/PurchasesDelegate.swift
```diff
@@ -47,13 +47,24 @@ import Foundation
...
+     * then call the `startPurchase` block when the app is ready to make the promotional purchase.
+     *
+     * If the purchase should never be made, you don't need to ever call the block and
+     * ``Purchases`` will not proceed with the promotional purchase.
+     *
      * - Parameter product: `StoreProduct` the product that was selected from the app store
      */
+    @objc optional func purchases(_ purchases: Purchases,
+                                  isReadyForPromotedProduct product: StoreProduct,
+                                  purchase startPurchase: @escaping DeferredPromotionalPurchaseBlock)
```
NachoSoto, line 60: The main change of this PR. I changed this slightly from @aboedo's suggestion, I think `purchases:isReadyForPromotedProduct:purchase:` reads nicely.

Sources/Purchasing/Purchases.swift
```diff
@@ -1776,24 +1776,24 @@ extension Purchases: PurchasesOrchestratorDelegate {
...
+     * If your app is able to handle a purchase at the current time, run the `startPurchase` block.
      *
-     * If the app is not in a state to make a purchase: cache the defermentBlock, then call the defermentBlock
+     * If the app is not in a state to make a purchase: cache the `startPurchase` block, then call it
      * when the app is ready to make the promotional purchase.
      *
-     * If the purchase should never be made, you don't need to ever call the defermentBlock and ``Purchases``
-     * will not proceed with promotional purchases.
+     * If the purchase should never be made, you don't need to ever call the `startPurchase` block
+     * and ``Purchases`` will not proceed with promotional purchases.
```
beylmk, line 1785: i think we should update all `promotionals` -> `promoted`

Sources/Purchasing/PurchasesDelegate.swift
beylmk: i was finding the is/has prefix a little confusing, considering there is no boolean involved here. is that a swifty and/or delegate thing? 

maybe something without is/has like `receievedPromotedProductPurchaseHook` or something. i don't like that necessarily just spitballing...

Sources/Purchasing/Purchases.swift
beylmk: i know we have an example coming in MagicWeather...but considering how insanely confusing this flow is, I wonder if we should have an example here too? or perhaps point people to MagicWeather?

NachoSoto: Oh yeah good idea. Do you mean the example link to test this?

beylmk: i just meant an example of how a customer would implement this. but yeah a link to apple's docs for testing would also be helpful! with the URL and all

NachoSoto: You mean with actual code? We'd have to chose between `Swift` and `Objective-C` :/

beylmk: yeah true. i thought we had other examples elsewhere as precedence but maybe that was pre-migration. anyways, i think the documentation update should help enough! maybe we can point to the MagicWeather call from our site docs on the flow.

NachoSoto: > yeah true. i thought we had other examples elsewhere as precedence but maybe that was pre-migration. 

Oh I think we have some of that for the `AsyncStream` which is Swift only anyway.

>  i think the documentation update should help enough! maybe we can point to the MagicWeather call from our site docs on the flow.

I already added some of your doc suggestions 👍🏻 
I'll add the example link for how to test it and [this link](https://developer.apple.com/documentation/storekit/original_api_for_in-app_purchase/testing_promoted_in-app_purchases) (cc @aboedo any chance you could make a short link for that? 🙏🏻)

Sources/Purchasing/PurchasesDelegate.swift
NachoSoto: > i was finding the is/has/should prefix a little confusing, considering there is no boolean involved here. is that a swifty and/or delegate thing?

Definitely more common for a Boolean, but I _think_ this is also a Cocoa-like name for a protocol method?

Sources/Misc/Obsoletions.swift
NachoSoto: > This was never meant to be `public`. Use `PurchasesDelegate.purchases(_:isReadyForPromotedProduct:purchase:)`
?

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -16,8 +16,8 @@ import StoreKit
 
 @objc protocol PurchasesOrchestratorDelegate {
 
-    func shouldPurchasePromoProduct(_ product: StoreProduct,
-                                    defermentBlock: @escaping DeferredPromotionalPurchaseBlock)
+    func promotedPurchaseReadyToStart(for product: StoreProduct,
+                                      startPurchase: @escaping DeferredPromotionalPurchaseBlock)
```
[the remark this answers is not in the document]
NachoSoto: Oh yeah. I was thinking users don't really reference the block type, but I guess they do when the implement the method. I'll rename it.

Sources/Purchasing/PurchasesDelegate.swift
```diff
@@ -47,13 +47,30 @@ import Foundation
...
+     * If the purchase should never be made, you don't need to ever call the block and
+     * ``Purchases`` will not proceed with the promotional purchase.
+     *
+     * This can be tested by opening a link like:
+     * itms-services://?action=purchaseIntent&bundleId=<BUNDLE_ID>&productIdentifier=<SKPRODUCT_ID>
+     *
      * - Parameter product: `StoreProduct` the product that was selected from the app store
+     *
+     * ### Related Articles:
+     * - [Apple Documentation](https://developer.apple.com/documentation/storekit/original_api_for_in-app_purchase/testing_promoted_in-app_purchases)
```
NachoSoto, line 62: @aboedo any chance you can create a short link for this? 🙏🏻

Sources/Misc/Obsoletions.swift
aboedo: LGTM

NachoSoto: Changed.

Sources/Purchasing/Purchases.swift
```diff
@@ -1771,26 +1771,27 @@ public extension Purchases {
...
      */
     @objc
-    public func shouldPurchasePromoProduct(_ product: StoreProduct,
-                                           defermentBlock: @escaping DeferredPromotionalPurchaseBlock) {
-        guard let delegate = delegate else {
-            return
-        }
-
-        delegate.purchases?(self, shouldPurchasePromoProduct: product, defermentBlock: defermentBlock)
+    internal func promotedPurchaseReadyToStart(for product: StoreProduct,
```
beylmk, line 1792: 👏  😻

Sources/Purchasing/PurchasesDelegate.swift
```diff
@@ -47,15 +47,32 @@ import Foundation
...
+     * This can be tested by opening a link like:
+     * itms-services://?action=purchaseIntent&bundleId=<BUNDLE_ID>&productIdentifier=<SKPRODUCT_ID>
+     *
      * - Parameter product: `StoreProduct` the product that was selected from the app store
+     *
+     * ### Related Articles:
+     * - [Apple Documentation](https://rev.cat/testing-promoted-in-app-purchases)
      */
+    @objc optional func purchases(_ purchases: Purchases,
+                                  readyForPromotedProduct product: StoreProduct,
```
beylmk, line 65: extra picky here, feel free to ignore...but maybe `readyToPurchasePromotedProduct`?

NachoSoto: Then the whole method would become `purchases(_: readyToPurchasePromotedProduct:purchase:)` instead of "ready for promoted product purchase".

beylmk (approved):
thanks for taking all my thoughts into account! 👏

NachoSoto:
Everyone else also good with this?

beylmk:
> Everyone else also good with this?

@aboedo if you get a chance to confirm this today that'd be great :) then i can rename the related flutter work

Sources/Purchasing/PurchasesDelegate.swift
```diff
@@ -47,13 +47,30 @@ import Foundation
...
+     * then call the `startPurchase` block when the app is ready to make the promotional purchase.
+     *
+     * If the purchase should never be made, you don't need to ever call the block and
+     * ``Purchases`` will not proceed with the promotional purchase.
+     *
+     * This can be tested by opening a link like:
+     * itms-services://?action=purchaseIntent&bundleId=<BUNDLE_ID>&productIdentifier=<SKPRODUCT_ID>
+     *
      * - Parameter product: `StoreProduct` the product that was selected from the app store
+     *
```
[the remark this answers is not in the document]
aboedo: bump

aboedo (approved):
naming looks good to me. I'm still a bit confused as to why the internal method isn't named the same as the external one, but since it's internal we can always have that discussion in a separate PR

aboedo:
Thanks for carrying this one to the finish line Nacho!

Sources/Purchasing/PurchasesDelegate.swift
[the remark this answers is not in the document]
NachoSoto: Oh sorry I missed this, will add.

NachoSoto:
> I'm still a bit confused as to why the internal method isn't named the same as the external one, but since it's internal we can always have that discussion in a separate PR

You mean `promotedPurchaseReadyToStart(for:startPurchase:)` vs `purchases(_:readyForPromotedProduct:purchase:)`?

Good point, I renamed it!

PR #1462, opened 2022-04-06
Extracted `CurrentUserProvider` protocol to simplify dependency injection

NachoSoto:
A lot of our classes depended on the mutable `IdentityManager`, when in reality they only needed a read-only `currentAppUserID`.
This PR changes the requirement, with the following advantages:

- Ensure all these classes (`AttributionPoster`, `PurchasesOrchestrator`, `TrialOrIntroPriceEligibiltyChecker`, `BeginRefundRequestHelper`, `ManageSubscriptionsHelper`) cannot logIn / logOut, only read the current user
- Simplify tests by only requiring a smaller type
- Simplifying instantiation of these types outside the normal `Purchases` initialization.

I've also created `MockCurrentUserProvider`, but changed the existing `MockIdentityManager` so that calls to `logIn`/`logOut` just fail. These aren't mocked, and they wouldn't have an impact on the mocked readable values, so calling them is most likely unexpected and should make the test fail.

### DIFF
--- a/Sources/Support/BeginRefundRequestHelper.swift
+++ b/Sources/Support/BeginRefundRequestHelper.swift
@@ -21,7 +21,7 @@ class BeginRefundRequestHelper {
 
     private let systemInfo: SystemInfo
     private let customerInfoManager: CustomerInfoManager
-    private let identityManager: IdentityManager
+    private let currentUserProvider: CurrentUserProvider
 
 #if os(iOS)
     @available(iOS 15.0, *)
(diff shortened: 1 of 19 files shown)


Sources/Support/BeginRefundRequestHelper.swift
```diff
@@ -93,7 +93,7 @@ private extension BeginRefundRequestHelper {
 
         do {
             customerInfo = try await self.customerInfoManager.customerInfo(
-                appUserID: self.identityManager.currentAppUserID
+                appUserID: self.currentUserProvider.currentAppUserID
```
aboedo, line 96: I ❤️ this. Ensuring misuse of an API, even by us, is just fantastic :chefskiss:

aboedo (approved):

PR #1464, opened 2022-04-06
`StoreKitIntegrationTests`: added more tests for `TrialOrIntroPriceEligibilityChecker`

NachoSoto:
Finishes [CF-497].
Depends on ~~#1462~~ and #1467.

This adds a lot more tests for `TrialOrIntroPriceEligibilityChecker` that run in both SK1 and SK2 thanks to #1467.

[CF-497]: https://revenuecats.atlassian.net/browse/CF-497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
@@ -11,7 +11,7 @@ import Nimble
 import StoreKitTest
 import XCTest
 
-// swiftlint:disable file_length
+// swiftlint:disable type_body_length file_length
 
 class StoreKit2IntegrationTests: StoreKit1IntegrationTests {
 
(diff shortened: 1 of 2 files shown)


Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -11,15 +11,9 @@ import RevenueCat
...
 import XCTest
 
-class StoreKit2IntegrationTests: StoreKit1IntegrationTests {
-
-    override class var sk2Enabled: Bool { return true }
-
-}
-
-class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
+class BaseStoreKitIntegrationTests: BaseBackendIntegrationTests {
```
NachoSoto, line 14: Extracting this base class with no tests so `BaseTrialOrIntroEligibilityCheckerIntegrationTests` can inherit from it.

Tests/BackendIntegrationTests/TrialOrIntroEligibilityCheckerIntegrationTests.swift
```diff
@@ -0,0 +1,104 @@
...
+                                     apiKey: Constants.apiKey,
+                                     attributionFetcher: .init(attributionFactory: .init(),
+                                                               systemInfo: systemInfo))
+        self.checker = .init(
+            receiptFetcher: .init(requestFetcher: .init(operationDispatcher: .default),
+                                  systemInfo: systemInfo),
+            introEligibilityCalculator: .init(productsManager: .init(systemInfo: systemInfo),
+                                              receiptParser: .init()),
+            backend: backend,
+            currentUserProvider: currentUserProvider,
```
NachoSoto, line 42: #1462 made this simpler so we don't have to create a whole `IdentityManager`.

Tests/BackendIntegrationTests/TrialOrIntroEligibilityCheckerIntegrationTests.swift
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    func checkEligibility(_ productIdentifiers: [String]) async throws -> [String: IntroEligibility] {
+        fatalError("This method must be overriden")
+    }
+
+    /// This test is similar to ``StoreKit1IntegrationTests/testEligibleForIntroBeforePurchaseAndIneligibleAfter``
+    /// except ran directly on `TrialOrIntroPriceEligibilityChecker`, which makes the test force
+    /// either the SK1 or SK2 implementation, so they can be tested independently.
+    final func runTestEligibleForIntroBeforePurchaseAndIneligibleAfter() async throws {
```
NachoSoto, line 55: Note that this function doesn't begin with `test`, so `XCTest` doesn't see it, and instead runs the tests in the subclasses (cc @taquitos)

Tests/BackendIntegrationTests/TrialOrIntroEligibilityCheckerIntegrationTests.swift
```diff
@@ -0,0 +1,104 @@
...
+
+    func testEligibleForIntroBeforePurchaseAndIneligibleAfter() async throws {
+        try await self.runTestEligibleForIntroBeforePurchaseAndIneligibleAfter()
+    }
+}
+
+class TrialOrIntroEligibilityCheckerStoreKit2IntegrationTests: BaseTrialOrIntroEligibilityCheckerIntegrationTests {
+
+    override func checkEligibility(_ productIdentifiers: [String]) async throws -> [String: IntroEligibility] {
+        guard #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) else {
```
NachoSoto, line 94: Not using `AvailabilityChecks` so that I can get away with not marking this method with `@available`, that way the shared test can simply run on all classes.

NachoSoto:
I'm still not super happy with having to instantiate a separate `TrialOrIntroPriceEligibilityChecker` instead of using `Purchases.shared`. Alternatives that I don't love either:
- Expose `Purchases.trialOrIntroPriceEligibilityChecker`
- Make `TrialOrIntroPriceEligibilityChecker` check `SystemInfo.useStoreKit2IfAvailable`
- Make `TrialOrIntroPriceEligibilityChecker` check `SystemInfo.useStoreKit2IfAvailable` to decide which implementation to use, *but only in DEBUG*, so in release it always biases for SK2

The last option would greatly simplify this test, and avoid having 2 separate `CurrentUserProvider`s in this test test (and every single dependency), which is a bad smell that might break in the future.

NachoSoto:
Gonna simplify this with #1467.

NachoSoto:
This is a little bit hard to review until #1467 is merged, so you can hold on until that's merged.

NachoSoto:
Okay this is ready to review again.

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -189,33 +191,85 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
...
+        try AvailabilityChecks.iOS13APIAvailableOrSkipTest()
+
+        try await self.purchaseMonthlyOffering()
+        try self.verifyEntitlementWentThrough()
+
+        let product2 = try await self.annualPackage.storeProduct
+
+        let eligibility = await Purchases.shared.checkTrialOrIntroDiscountEligibility(product: product2)
+
+        expect(eligibility) == .eligible
```
aboedo, line 230: the simplicity of these tests is just so nice to see

aboedo (approved):
✅🎉

PR #1465, opened 2022-04-06
fastlane bump: update master changelog

aboedo:
Fixes a missing step in `fastlane bump`: 
After updating changelog.latest.md, the changes from that file aren't pushed to the master changelog. 
This adds that step


NachoSoto (approved):
Makes sense.

PR #1467, opened 2022-04-06
Added `StoreKit2Setting` to replace `useStoreKit2IfAvailable`

NachoSoto:
This adds a third case `disabled` that allows tests to explicitly disable StoreKit 2.
Some types, like `TrialOrIntroPriceEligibilityChecker` and `TransactionsManager` always used SK2 because the implementation was better. But this meant that tests couldn't explicitly disable it (other than calling the internal methods directly).

This allows #1464 to be a lot simpler but being able to use the public `Purchases` methods, but forcing the implementation to use SK1.

Note that this is purely an implementation detail. `StoreKit2Setting` is `internal`, and the public API still only exposes the same boolean.

This basically solves [CF-497] too, since those tests will now force `TrialOrIntroPriceEligibilityChecker` to run the SK1 implementation.

[CF-497]: https://revenuecats.atlassian.net/browse/CF-497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Misc/StoreKit2Setting.swift
+++ b/Sources/Misc/StoreKit2Setting.swift
@@ -17,11 +17,12 @@ enum StoreKit2Setting {
     /// Never use SK2
     case disabled
 
-    /// Use SK2 if available for certain APIs that provide a better implementation
+    /// Use SK2 (if available in the current device) only for certain APIs that provide a better implementation
+    /// For example: intro eligibility, determining if a receipt has purchases, managing subscriptions.
     case enabledOnlyForOptimizations
 
-    /// Enable SK2 in all APIs if available
-    case enabledIfAvailable
+    /// Enable SK2 in all APIs if available in the current device
+    case enabledForCompatibleDevices
 
 }
 
@@ -29,10 +30,22 @@ extension StoreKit2Setting {
 
     init(useStoreKit2IfAvailable: Bool) {
         self = useStoreKit2IfAvailable
-            ? .enabledIfAvailable
+            ? .enabledForCompatibleDevices
             : .enabledOnlyForOptimizations
     }
 
     static let `default`: Self = .enabledOnlyForOptimizations
+
+}
+
+extension StoreKit2Setting: CustomDebugStringConvertible {
+
+    var debugDescription: String {
+        switch self {
+        case .disabled: return "disabled"
+        case .enabledOnlyForOptimizations: return "optimizations-only"
+        case .enabledForCompatibleDevices: return "enabled"
+        }
+    }
 
 }
--- a/Sources/Misc/StoreKit2Setting.swift
+++ b/Sources/Misc/StoreKit2Setting.swift
@@ -1,0 +1,38 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  StoreKit2Setting.swift
+//
+//  Created by Nacho Soto on 4/6/22.
+
+/// Defines when StoreKit 2 APIs may be used
+enum StoreKit2Setting {
+
+    /// Never use SK2
+    case disabled
+
+    /// Use SK2 if available for certain APIs that provide a better implementation
+    case enabledOnlyForOptimizations
+
+    /// Enable SK2 in all APIs if available
+    case enabledIfAvailable
+
+}
+
+extension StoreKit2Setting {
+
+    init(useStoreKit2IfAvailable: Bool) {
+        self = useStoreKit2IfAvailable
+            ? .enabledIfAvailable
+            : .enabledOnlyForOptimizations
+    }
+
+    static let `default`: Self = .enabledOnlyForOptimizations
+
+}
--- a/Sources/Networking/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient.swift
@@ -124,7 +124,7 @@ private extension HTTPClient {
             "X-Platform-Flavor": systemInfo.platformFlavor,
             "X-Client-Version": SystemInfo.appVersion,
             "X-Client-Build-Version": SystemInfo.buildVersion,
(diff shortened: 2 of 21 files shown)


Tests/UnitTests/Misc/SystemInfoTests.swift
```diff
@@ -66,20 +67,6 @@ class SystemInfoTests: XCTestCase {
         expect(try SystemInfo.withReceiptResult(.nilURL).isSandbox) == false
     }
 
-    func testUseStoreKit2IfAvailable() throws {
```
NachoSoto, line 69: This test was simply checking that the property was set in the constructor.
One could argue that can indeed break, but at some point we have to have some trust that the basic things are being done 😅 We can't have tests for every single constructor and property.

Sources/Networking/HTTPClient.swift
```diff
@@ -124,7 +124,7 @@ private extension HTTPClient {
             "X-Platform-Flavor": systemInfo.platformFlavor,
             "X-Client-Version": SystemInfo.appVersion,
             "X-Client-Build-Version": SystemInfo.buildVersion,
-            "X-StoreKit2-Enabled": "\(self.systemInfo.useStoreKit2IfAvailable)",
+            "X-StoreKit2-Enabled": "\(self.systemInfo.storeKit2Setting == .enabledIfAvailable)",
```
taquitos, line 127: Should this be `self.systemInfo.storeKit2Setting == .enabledIfAvailable || self.systemInfo.storeKit2Setting == . enabledOnlyForOptimizations ` ?

Sources/Misc/StoreKit2Setting.swift
```diff
@@ -0,0 +1,38 @@
...
+//  Created by Nacho Soto on 4/6/22.
+
+/// Defines when StoreKit 2 APIs may be used
+enum StoreKit2Setting {
+
+    /// Never use SK2
+    case disabled
+
+    /// Use SK2 if available for certain APIs that provide a better implementation
+    case enabledOnlyForOptimizations
```
taquitos, line 21: This isn't used, yet, I'm guessing a follow up PR uses it.

NachoSoto: `enabledOnlyForOptimizations` is the default value (see `default` below)
It's the current existing behavior when `useStoreKit2IfAvailable` is `false`. We were only using SK2 when it made sense for better implementations.

The new `case` is `disabled`, which allows us to not even use SK2 in those cases.

The fact that you were confused here and in the other comment is probably because this name is bad :/ but @aboedo and I couldn't come up with a better one.

Hopefully with this explanation this make a bit more sense, and maybe you have a suggestion for how to make this behavior more clear?

Sources/Networking/HTTPClient.swift
NachoSoto: See the other comment.

Sources/Misc/StoreKit2Setting.swift
aboedo: +1 to Nacho. FWIW the philosophy that this case applies is "use SK2 whenever its value is guaranteed to be more accurate than using SK1 because SK1 was abhorrent at doing this"

aboedo: which... we should probably elaborate on in this doc, including examples (which off the top of my head are: Intro eligibility, determining whether a receipt has any purchases on it, and manage subscriptions)

Sources/Networking/HTTPClient.swift
aboedo: this is a debug value for us as far as I know, so we don't have to even make it a bool. 
We could just send a new header that will be one of the 3 possible values

Sources/Misc/StoreKit2Setting.swift
```diff
@@ -0,0 +1,38 @@
...
+
+    /// Enable SK2 in all APIs if available
+    case enabledIfAvailable
```
aboedo, line 24: `enabledIfAvailable` reads a bit odd for me, how about `enabledForCompatibleDevices`?

Sources/Purchasing/TransactionsManager.swift
```diff
@@ -15,16 +15,18 @@ import StoreKit
...
+         receiptParser: ReceiptParser) {
+        self.storeKit2Setting = storeKit2Setting
         self.receiptParser = receiptParser
     }
 
     func customerHasTransactions(receiptData: Data, completion: @escaping (Bool) -> Void) {
-        // Note: this uses SK2 regardless of the value of `SystemInfo.useStoreKit2IfAvailable`
-        // because its implementation is more accurate.
-        if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) {
+        // Note: this uses SK2 (unless it's explicitly disabled) because its implementation is more accurate.
```
aboedo, line 28: ❤️

aboedo (commented):
Looking great! I left a couple of minor comments

Sources/Networking/HTTPClient.swift
NachoSoto: Oh great point!

Sources/Misc/StoreKit2Setting.swift
NachoSoto: Much better 👍🏻

NachoSoto (commented):
Updated from your comments 👍🏻

Sources/Networking/HTTPClient.swift
```diff
@@ -124,7 +124,7 @@ private extension HTTPClient {
             "X-Platform-Flavor": systemInfo.platformFlavor,
             "X-Client-Version": SystemInfo.appVersion,
             "X-Client-Build-Version": SystemInfo.buildVersion,
-            "X-StoreKit2-Enabled": "\(self.systemInfo.useStoreKit2IfAvailable)",
```
NachoSoto, line 127: I renamed the key as well.

taquitos (approved):
🏪  🐐

aboedo (approved):

PR #1468, opened 2022-04-06
Add SK2 configure parameter to DocC

codykerns:
For [CF-476](https://revenuecats.atlassian.net/browse/CF-496)

### Motivation
We support using SK2 (beta, experimentally), so we should include it in our DocC docs.

### Description
Adds the `configure` option with `useStoreKit2IfAvailable` to DocC docs


aboedo (approved):

NachoSoto (approved):
🎉

PR #1470, opened 2022-04-07
Push docs to separate repo

aboedo:
This separates the docs into a new repo, [`purchases-ios-docs`](https://github.com/revenuecat/purchases-ios-docs). Having it detached from the main repo means that we can easily have docs for multiple versions of the SDK live there without needing to worry about workflows (like for hotfixes). I already backfilled docs for 3.14.1, 4.0.0 and 4.1.0 (even 4.2.0-SNAPSHOT)

So this generates the docs as usual, but instead of adding them to the `docs` folder in this repo, it adds them to the other repo and pushes. After pushing they'll be automatically published. 

You can see this working in https://revenuecat.github.io/purchases-ios-docs. 

For different versions: 

- 4.2.0-SNAPSHOT: https://revenuecat.github.io/purchases-ios-docs/4.2.0-SNAPSHOT/documentation/revenuecat/
- 4.1.0: https://revenuecat.github.io/purchases-ios-docs/4.1.0/documentation/revenuecat/
- 4.0.0: https://revenuecat.github.io/purchases-ios-docs/4.0.0/documentation/revenuecat/
- 3.14.1: https://revenuecat.github.io/purchases-ios-docs/3.14.1

PR #1472, opened 2022-04-08
Delete old docs folders

aboedo:
Follow-up to #1470 

Removes the `/docs` and `LatestTagDocs` folders, since they're no longer needed


aboedo:
there's really nothing to review at the files level, this just removes the folders entirely

aboedo:
So if you have any questions we can discuss them on the PR conversation tab, the diff isn't particularly useful

NachoSoto (approved):
🔥

PR #1474, opened 2022-04-08
replaced old docs jobs with the new one

aboedo:
Removed the old jobs for generating a DocC archive and replaced with a new job that directly runs on tags.


NachoSoto (approved):
🎉

PR #1479, opened 2022-04-08
`SubscriberAttributesManager`: fixed potential race condition

NachoSoto:
Fixes https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/6125/workflows/d418661d-752d-403e-84d5-e7fdc32dad88/jobs/22409/tests

The tests introduced in #1441 exposed a potential race condition in the SDK.
On SDK initialization, `Purchases` calls `updateAllCaches`, which in turn makes `CustomerInfoManager` and `OfferingsManager` update its internal caches. Specifically, `CustomerInfoManager` makes a `getCustomerInfo` request ensuring that the backend knows about the current `appUserID`.

These 2 calls are made in the `OperationDispatcher.workerQueue`, so they're ensured proper ordering.

However, `SubscriberAttributesManager` enqueues the requests to post new attributes synchronously and outside this worker thread. If this happens _before_ `CustomerInfoManager` has had a chance to enqueue the `Backend.getCustomerInfo` request, the backend will fail with `SUBSCRIPTION_NOT_FOUND` errors.

### DIFF
--- a/Sources/Purchasing/Purchases.swift
+++ b/Sources/Purchasing/Purchases.swift
@@ -284,6 +284,7 @@ public typealias DeferredPromotionalPurchaseBlock = (@escaping PurchaseCompleted
         let attributionDataMigrator = AttributionDataMigrator()
         let subscriberAttributesManager = SubscriberAttributesManager(backend: backend,
                                                                       deviceCache: deviceCache,
+                                                                      operationDispatcher: operationDispatcher,
                                                                       attributionFetcher: attributionFetcher,
                                                                       attributionDataMigrator: attributionDataMigrator)
         let attributionPoster = AttributionPoster(deviceCache: deviceCache,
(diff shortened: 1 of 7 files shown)


aboedo (approved):
Awesome! 
it'd be nice to get a test in that reproduces the initial bug with 100% repro to ensure we don't re-introduce it in the future, but that's something we can do down the line. 

Also, this is worthy of doing a 4.2.1 hotfix IMO, even if it's not a new bug, just so that we ship the fix and not have the bug around any longer

NachoSoto:
> it'd be nice to get a test in that reproduces the initial bug with 100% repro to ensure we don't re-introduce it in the future, but that's something we can do down the line.

I thought about how to do it but couldn't think of a way other than simply using `MockOperationDispatcher` in `SubscriberAttributesManagerTests`, which wouldn't really be testing the whole picture.

> Also, this is worthy of doing a 4.2.1 hotfix IMO, even if it's not a new bug, just so that we ship the fix and not have the bug around any longer

I agree!

PR #1480, opened 2022-04-08
`NetworkError.errorResponse`: added status code to `userInfo`

NachoSoto:
This could help debug these errors.

Thanks to the error refactor this only had to be put in one place.

### DIFF
--- a/Sources/Error Handling/ErrorDetails.swift
+++ b/Sources/Error Handling/ErrorDetails.swift
@@ -14,7 +14,11 @@
 
 import Foundation
 
-class ErrorDetails: NSObject {
+enum ErrorDetails {
+
+    static let attributeErrorsKey: NSError.UserInfoKey = "attribute_errors"
+    static let attributeErrorsResponseKey: NSError.UserInfoKey = "attributes_error_response"
+    static let statusCodeErrorKey: NSError.UserInfoKey = "rc_response_status_code"
 
     static let readableErrorCodeKey: NSError.UserInfoKey = "readable_error_code"
     static let extraContextKey: NSError.UserInfoKey = "extra_context"
--- a/Sources/Error Handling/ErrorDetails.swift
+++ b/Sources/Error Handling/ErrorDetails.swift
@@ -14,15 +14,28 @@
 
 import Foundation
 
+extension NSError.UserInfoKey {
+
+    static let attributeErrors: NSError.UserInfoKey = "attribute_errors"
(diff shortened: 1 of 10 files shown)


Sources/Error Handling/ErrorDetails.swift
```diff
@@ -14,7 +14,11 @@
 
 import Foundation
 
-class ErrorDetails: NSObject {
+enum ErrorDetails {
```
NachoSoto, line 17: This didn't need to be an Objective-C class.

Sources/Error Handling/ErrorDetails.swift
```diff
@@ -14,7 +14,11 @@
...
+
+    static let attributeErrorsKey: NSError.UserInfoKey = "attribute_errors"
+    static let attributeErrorsResponseKey: NSError.UserInfoKey = "attributes_error_response"
+    static let statusCodeErrorKey: NSError.UserInfoKey = "rc_response_status_code"
```
NachoSoto, line 21: Now all these keys are in one place.

Tests/UnitTests/Networking/HTTPResponseTests.swift
```diff
@@ -31,7 +31,7 @@ class ErrorResponseTests: XCTestCase {
 
         expect(error.domain) == ErrorCode.errorDomain
         expect(error.code) == ErrorCode.invalidCredentialsError.rawValue
-        expect(error.userInfo[Backend.RCAttributeErrorsKey]).to(beNil())
+        expect(error.userInfo[ErrorDetails.attributeErrorsKey as String]).to(beNil())
```
aboedo, line 34: maybe we can have an extension so that we can do `ErrorDetails.attributeErrorsKey.string` directly instead of doing the cast everywhere?

aboedo (approved):

Tests/UnitTests/Networking/HTTPResponseTests.swift
NachoSoto: Yeah I thought about something like that.
How about:

```swift
extension NSError.UserInfoKey {
  static let attributeErrors = "xxx"
  // ... etc
}

enum ErrorDetails {
  static let attributeErrorsKey = NSError.UserInfoKey.attributeErrors as String
  // ... etc
}
```

So when we need the key we can just do `.attributeErrors`, but if not `ErrorDetails.attributeErrors` just works?

NachoSoto: Changed

PR #1481, opened 2022-04-08
`BackendErrorTests`: added explicit test for subscriber not found errors

NachoSoto:
The race condition fixed #1479 is luckily not a big deal because posting the attributes would have been retried.
This test ensures that behavior in a concise way.

### DIFF
--- a/Tests/UnitTests/Networking/BackendErrorTests.swift
+++ b/Tests/UnitTests/Networking/BackendErrorTests.swift
@@ -51,4 +51,20 @@ class BackendErrorTests: BaseErrorTests {
                              underlyingError: underlyingError as NSError)
     }
 
+    func testSubscriptionNotFoundErrorsArentSuccessfullySynced() {
+        // See https://github.com/RevenueCat/purchases-ios/pull/1479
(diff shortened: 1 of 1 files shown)


Tests/UnitTests/Networking/BackendErrorTests.swift
```diff
@@ -51,4 +51,20 @@ class BackendErrorTests: BaseErrorTests {
...
+        // This test ensures that if that race condition does happen
+        // at least the attributes won't be marked as synced
+
+        let response = ErrorResponse(
+            code: .subscriptionNotFoundForCustomer,
+            message: "Subscription not found for subscriber",
+            attributeErrors: [:]
+        )
+
+        let error: BackendError = .networkError(.errorResponse(response, .notFoundError))
```
NachoSoto, line 65: In reality this is equivalent to this (which is in `HTTPStatusCodeTests`:
```swift
expect(HTTPStatusCode.notFoundError.isSuccessfullySynced) == false
```
But that's an implementation detail. In the future `NetworkError.successfullySynced` might introspect the error information.

aboedo (approved):

PR #1484, opened 2022-04-08
update all jobs to use Xcode 13.3.0

aboedo:
Updates all CircleCI jobs to use Xcode 13.3.0. 

This also fixes the `docs_deploy` job, which needs 13.3.0 for static page generation with DocC.


NachoSoto (approved):
😍

PR #1486, opened 2022-04-11
test: fix flakiness in CI

aboedo:
`BackendIntegrationTests` are flaky right now in CI, but they're fine locally. We get failures from different tests on each run. 

I think this may be happening because `BackendIntegrationTests` started using SKTest, and SKTest needs time to initialize, but we didn't add that to `BackendIntegrationTests`. 

We do have an artificial delay for `StoreKitConfigTestCase`. 

So this ugly PR just makes `BaseBackendIntegrationTests` a subclass of `StoreKitConfigTestCase` so that it gets this artificial delay.

### DIFF
--- a/RevenueCat.xcodeproj/project.pbxproj
+++ b/RevenueCat.xcodeproj/project.pbxproj
@@ -118,6 +118,7 @@
 		2DE20B8626409EB8004C597D /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 2DE20B8526409EB8004C597D /* Assets.xcassets */; };
 		2DE20B8926409EB8004C597D /* Preview Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 2DE20B8826409EB8004C597D /* Preview Assets.xcassets */; };
 		2DE20B9226409ECF004C597D /* StoreKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 2DE20B9126409ECF004C597D /* StoreKit.framework */; };
+		2DE21DE928046EAA002297A4 /* StoreKitConfigTestCase.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2D90F8CB26FD2BA1009B9142 /* StoreKitConfigTestCase.swift */; };
 		2DE61A84264190830021CEA0 /* Constants.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2DE61A83264190830021CEA0 /* Constants.swift */; };
 		2DEAC2DD26EFE46E006914ED /* AppDelegate.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2DEAC2DC26EFE46E006914ED /* AppDelegate.swift */; };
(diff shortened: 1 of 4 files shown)


NachoSoto (commented):
This is helpful though 👍🏻 

One thing we need to disable in integration tests is the empty `Transaction` listener in `StoreKitConfigTestCase.setUp` though, I think that'll break the "production" code?
Also I think we need to add `CIRCLECI_STOREKIT_TESTS_DELAY_SECONDS` to the Integration tests, or are the environment variables shared across tests?

NachoSoto:
I've also pushed #1487 and #1488 to help debug these. This run failed with:
> ✖ testPurchaseMadeBeforeLogInWithExistingUserIsNotRetainedUnlessRestoreCalled, failed - expected to equal <1>, got <0>

Which is hard to debug.

aboedo:
> Also I think we need to add CIRCLECI_STOREKIT_TESTS_DELAY_SECONDS to the Integration tests, or are the environment variables shared across tests?

Env variables are repo-wide, so we should be good there

>One thing we need to disable in integration tests is the empty Transaction listener in StoreKitConfigTestCase.setUp though, I think that'll break the "production" code?

You should be able to have multiple observers, right? If not then we should dig deeper into that, since we'd have to warn customers to ensure that there's no observer set up anywhere else other than the SDK if using SK2

NachoSoto:
> You should be able to have multiple observers, right? If not then we should dig deeper into that, since we'd have to warn customers to ensure that there's no observer set up anywhere else other than the SDK if using SK2

Oh yeah I think you're right

RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -119,6 +119,8 @@
 		2DE20B8926409EB8004C597D /* Preview Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 2DE20B8826409EB8004C597D /* Preview Assets.xcassets */; };
 		2DE20B9226409ECF004C597D /* StoreKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 2DE20B9126409ECF004C597D /* StoreKit.framework */; };
 		2DE21DE928046EAA002297A4 /* StoreKitConfigTestCase.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2D90F8CB26FD2BA1009B9142 /* StoreKitConfigTestCase.swift */; };
+		2DE21DEC280470B0002297A4 /* UnitTestsConfiguration.storekit in Resources */ = {isa = PBXBuildFile; fileRef = 2D43017726EBFD7100BAB891 /* UnitTestsConfiguration.storekit */; };
```
NachoSoto, line 122: This is now in integration tests? I'm so confused, Integration tests use `RevenueCat_IntegrationPurchaseTesterConfiguration.storekit`.

NachoSoto:
FYI integration tests are no longer flaky in `main` 🎉

aboedo:
@NachoSoto I'm gonna close this out in that case, this was just a test to see if introducing the artificial delay would fix test flakiness in CI

PR #1488, opened 2022-04-11
Integration Tests: improved verification methods

NachoSoto:
Hoping this will make debugging flaky CI failures easier.

Example failure that's hard to debug:
>  ✖ testPurchaseMadeBeforeLogInWithExistingUserIsNotRetainedUnlessRestoreCalled, failed - expected to equal <1>, got <0>

- Added `file`/`line` information so the failure comes from the actual test line
- Added explicit descriptions to failures

### DIFF
--- a/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
@@ -323,27 +323,64 @@ private extension StoreKit1IntegrationTests {
     }
 
     @discardableResult
-    func verifyEntitlementWentThrough() throws -> EntitlementInfo {
+    func verifyEntitlementWentThrough(
+        file: FileString = #file, line: UInt = #line
+    ) throws -> EntitlementInfo {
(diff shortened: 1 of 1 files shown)


taquitos (approved):
🎈🐐

aboedo (approved):

NachoSoto:
I think the problem is the indirection we have in tests. Example from the test that failed:
```swift
try await self.purchaseMonthlyOffering()
try self.verifyEntitlementWentThrough()
```

The purchase goes through, but I wonder if there's a race condition here where `PurchasesDelegate` hasn't gotten an update yet?

I think we should change these to explicitly read the `CustomerInfo` from the purchase results.
And maybe have a test that does check that the delegate is _eventually_ updated.

What do you think @aboedo?

NachoSoto:
I think the majority of the test failures I've seen are related to this to be fair. I think the "unexpected backend errors" come _after_, which should be eliminated by #1487.
I think #1486 is still a good thing to have though, and I'll send a PR with my suggested change.

NachoSoto:
I'll merge this in the mean time!

NachoSoto:
Done in #1492.

PR #1491, opened 2022-04-11
Fix non-attribution integrations from trying to collect attribution data

codykerns:
For [CF-449](https://revenuecats.atlassian.net/browse/CF-449)

### Motivation
[OneSignal](https://docs.revenuecat.com/docs/onesignal), [Airship](https://docs.revenuecat.com/docs/airship), and [CleverTap](https://docs.revenuecat.com/docs/clevertap) are not attribution providers and the convenience setters should only be setting single subscriber attributes, and not additional device identifiers required for attribution providers.

### Description
Each of the integrations now only set their single specific subscriber attribute, and attempt to collect no other identifiers.

### DIFF
--- a/Tests/UnitTests/SubscriberAttributes/SubscriberAttributesManagerTests.swift
+++ b/Tests/UnitTests/SubscriberAttributes/SubscriberAttributesManagerTests.swift
@@ -849,7 +849,7 @@ class SubscriberAttributesManagerTests: XCTestCase {
     func testSetOnesignalID() {
         let onesignalID = "onesignalID"
         self.subscriberAttributesManager.setOnesignalID(onesignalID, appUserID: "kratos")
-        expect(self.mockDeviceCache.invokedStoreCount) == 4
+        expect(self.mockDeviceCache.invokedStoreCount) == 1
         guard let invokedParams = self.mockDeviceCache.invokedStoreParameters else {
             fatalError("no attributes received")
         }
@@ -865,7 +865,7 @@ class SubscriberAttributesManagerTests: XCTestCase {
 
         self.subscriberAttributesManager.setOnesignalID(nil, appUserID: "kratos")
 
-        expect(self.mockDeviceCache.invokedStoreCount) == 8
+        expect(self.mockDeviceCache.invokedStoreCount) == 2
         guard let invokedParams = self.mockDeviceCache.invokedStoreParameters else {
             fatalError("no attributes received")
         }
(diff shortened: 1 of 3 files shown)


aboedo (approved):
great work! 🚢

Tests/UnitTests/SubscriberAttributes/SubscriberAttributesManagerTests.swift
```diff
@@ -1600,4 +1610,17 @@ private extension SubscriberAttributesManagerTests {
         expect(ipReceived.isSynced) == false
     }
 
+    func checkDeviceIdentifiersAreNotSet() {
+        let invokedParams = self.mockDeviceCache.invokedStoreParametersList
+
+        let idfvNilAttribute = invokedParams.first(where: { $0.attribute.key == "$idfv" })
+        expect(idfvNilAttribute) == nil
```
NachoSoto, line 1617: `Nimble` doesn't support this, I think it's ignoring the expectation. For checking against `nil` you'd use `expect(x).to(beNil())`.

But I'd simply it since you're checking if it contains something, which would provide a much better failure message:
```swift
expect(invokedParams)
  .to(containElementSatisfying { $0.attribute.key == "$idfv" })
```

Tests/UnitTests/SubscriberAttributes/SubscriberAttributesManagerTests.swift
```diff
@@ -1600,4 +1610,17 @@ private extension SubscriberAttributesManagerTests {
...
+
+        let idfvaNilAttribute = invokedParams.first(where: { $0.attribute.key == "$idfa" })
+        expect(idfvaNilAttribute) == nil
```
NachoSoto, line 1620: Ditto:
```swift
expect(invokedParams)
  .to(containElementSatisfying { $0.attribute.key == "$idfa" })
```

Tests/UnitTests/SubscriberAttributes/SubscriberAttributesManagerTests.swift
```diff
@@ -1600,4 +1610,17 @@ private extension SubscriberAttributesManagerTests {
...
+
+        let ipNilAttribute = invokedParams.first(where: { $0.attribute.key == "$ip" })
+        expect(ipNilAttribute) == nil
```
NachoSoto, line 1623: Ditto.

Tests/UnitTests/SubscriberAttributes/SubscriberAttributesManagerTests.swift
codykerns: Gotcha, that makes sense- to clarify, since I'm checking if it doesn't contain the attribute, this is how I updated it: 
```
expect(invokedParams).toNot(containElementSatisfying({ $0.attribute.key == "$idfv" }))
```

Does that make sense?

NachoSoto: Yup!

NachoSoto (approved):

PR #1492, opened 2022-04-11
`StoreKitIntegrationTests`: removed incorrect `waitUntilCustomerInfoIsUpdated` and fixed race condition

NachoSoto:
This is probably a leftover from before integration tests got changed to async.

`Purchases` initialization can fetch an updated `CustomerInfo`, so assuming this will be 1 is a race condition waiting to happen, because there are potentially multiple customer info requests happening.

Also, SDK initialization begins with an initial request to offerings, which results in a get-create of the initial anonymous user. To avoid race conditions with when this request finishes and make all tests deterministic, integration tests now wait for the initial get offerings to finish

## Changes:
- Removed `waitUntilCustomerInfoIsUpdated`
- `verifyEntitlementWentThrough` now calls `verifyEntitlementWentThrough` to verify the `CustomerInfo` has the correct entitlement
- No longer checking the `PurchasesDelegate.customerInfo` in all tests, which could lead to race conditions
- Added a single test that verifies the delegate is notified. This might still be flaky, but at least this issue should be isolated to that new test now.
- Fixed race condition

### DIFF
--- a/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
@@ -67,8 +67,7 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
         let customerInfo = try await self.purchaseMonthlyOffering().customerInfo
         expect(customerInfo.entitlements.all.count) == 1
 
-        let entitlements = self.purchasesDelegate.customerInfo?.entitlements
-        expect(entitlements?[Self.entitlementIdentifier]?.isActive) == true
+        try verifyEntitlementWentThrough()
 
         let anonUserID = Purchases.shared.appUserID
         let identifiedUserID = "\(#function)_\(anonUserID)_".replacingOccurrences(of: "RCAnonymous", with: "")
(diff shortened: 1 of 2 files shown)


NachoSoto:
I'm currently debugging a separate failure that started happening after this was removed 🤔

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -391,26 +382,36 @@ private extension StoreKit1IntegrationTests {
...
     @discardableResult
-    func purchaseMonthlyOffering() async throws -> PurchaseResultData {
-        return try await Purchases.shared.purchase(package: self.monthlyPackage)
+    func purchaseMonthlyOffering(
+        file: FileString = #file,
+        line: UInt = #line
+    ) async throws -> PurchaseResultData {
+        let data = try await Purchases.shared.purchase(package: self.monthlyPackage)
+
+        try self.verifyEntitlementWentThrough(data.customerInfo,
```
NachoSoto, line 391: All purchases through this method now verify the entitlement for good measure (instead of most tests doing that repeatedly).

NachoSoto:
It's interesting, looking at the trend of failures in `main` it seems like #1484 is what made the failures more frequent.
Either way, I'm glad we managed to find (at least one of) the race condition.

aboedo (approved):
I was convinced I had approved already!! 😅

PR #1496, opened 2022-04-12
Converted `CustomerInfo` and related types to use `Codable`

NachoSoto:
### Depends on:
- #1537
- #1540
- #1541
- #1543
- #1546
- #1547
- #1550
- #1551,
- #1565

### Changes

- Removed all custom deserialization code thanks to #1537, #1541, and #1543.
- The format of the customer info response is now very concise and only in one place (`CustomerInfoResponse`) instead of being passed through dictionaries and magic strings everywhere.
- `CustomerInfo` was combining the dictionaries inside of `"subscriptions"` and `"non_subscriptions"`, despite them having different formats. This is now made explicit through type conversions between the two, (see `CustomerInfoResponse.allTransactionsByProductId`).
- Removed all custom `CustomerInfo` errors since deserialization is automatic, and the underlying `Error` information will be provided thanks to `ErrorUtils.logDecodingError`.
- `CustomerInfo` deserialization always wraps errors into `ErrorCode.customerInfoError`, which is now also covered in a new test.
- Added a new snapshot test that covers `CustomerInfo` serialization by storing the result in a JSON file.

### Other changes:
- `CustomerInfo` equality is now automatic through the `CustomerInfoResponse` `Equatable` conformance.
- Simplified `schemaVersion` handling
- All tests still use `CustomerInfo(data: [String: Any])`, but that method is only visible for tests now, and it uses the underlying `Decodable` implementation (see `CustomerInfo+TestExtensions.swift`).
- Added extra tests for `CustomerInfoResponse` that use a fixture `CustomerInfo.json`, which is easier to maintain than having JSON in code.
- `RawDataContainer` implementation is changed in #1565.

### TODO:
- [x] Fix all tests
- [x] Update `RawValueContainer`
- [x] Consider recovering `CustomerInfoError`
- [x] Look at all previously logged errors
- [x] Finish `CustomerInfoResponseTests`
- [x] Handle all TODOs
- [x] Make `RawDataContainer` actually store all the original data: #1565


NachoSoto:
Down to just 7 failing tests 🎉 almost there!

Sources/CodableExtensions/EntitlementData+Extensions.swift
```diff
@@ -1,30 +0,0 @@
...
-//
-//      https://opensource.org/licenses/MIT
-//
-//  EntitlementData+Extensions.swift
-//
-//  Created by Juanpe Catalán on 27/8/21.
-
-import Foundation
-
-extension EntitlementInfo.EntitlementData {
```
NachoSoto, line 16: This custom deserialization is no longer needed

Sources/CodableExtensions/EntitlementProductData+Extensions.swift
```diff
@@ -1,38 +0,0 @@
...
-//
-//      https://opensource.org/licenses/MIT
-//
-//  EntitlementProductData+Extensions.swift
-//
-//  Created by Juanpe Catalán on 27/8/21.
-
-import Foundation
-
-extension EntitlementInfo.ProductData {
```
NachoSoto, line 16: This was removed in #1540.

Sources/Identity/CustomerInfo.swift
```diff
@@ -87,15 +86,6 @@ import Foundation
...
     @objc public let originalApplicationVersion: String?
 
-    /**
-     * The underlying data for this `CustomerInfo`.
-     *
-     * - Note: the content and format of this data isn’t documented and is subject to change.
-     *         it's only meant for debugging purposes or for getting access to future data
-     *         without updating the SDK.
-     */
-    @objc public let rawData: [String: Any]
```
NachoSoto, line 97: This is now provided automatically through `RawDataContainer`.

Sources/Identity/CustomerInfo.swift
```diff
@@ -129,13 +119,12 @@ import Foundation
             return false
         }
 
-        return NSDictionary(dictionary: self.jsonObjectWithNoDate)
```
NachoSoto, line 132: Much simpler comparison! `CustomerInfoResponse` `Equatable` implementation ignores the date.

Sources/Identity/CustomerInfo.swift
```diff
@@ -164,263 +153,168 @@ import Foundation
...
-                json: customerJSON.stringRepresentation
-            )
-        }
+    /// Initializes a `CustomerInfo` creating a copy.
+    convenience init(customerInfo: CustomerInfo) {
+        self.init(data: customerInfo.data)
     }
 
-    static let currentSchemaVersion = "2"
+    fileprivate init(data: Data) {
```
NachoSoto, line 170: This method now stores each of the properties from the already decoded `Data`.

Sources/Identity/CustomerInfo.swift
```diff
@@ -164,263 +153,168 @@ import Foundation
...
-            let latestNonSubscriptionTransactionsByProductId = [String: [String: Any]](
-                uniqueKeysWithValues: nonSubscriptionsByProductId.map { productId, transactionsArray in
-                    (productId, transactionsArray.last ?? [:])
-                })
-
-            self.allTransactionsByProductId = latestNonSubscriptionTransactionsByProductId
-                .merging(subscriptionTransactionsByProductId, strategy: .keepOriginalValue)
-
-            self.allPurchases = latestNonSubscriptionTransactionsByProductId
-                .merging(subscriptionTransactionsByProductId) { (current, _) in current }
```
NachoSoto, line 338: This was super hacky and hard to follow. It was combining the dictionaries inside of "subscriptions" and "non_subscriptions", despite them having different formats. This is now made explicit through type conversions between the two.

Sources/Networking/Responses/CustomerInfoResponse.swift
```diff
@@ -0,0 +1,190 @@
...
+                     transactionIdentifier: nil,
+                     store: self.store,
+                     isSandbox: self.isSandbox)
+    }
+
+}
+
+extension CustomerInfoResponse.Subscriber {
+
+    var allTransactionsByProductId: [String: CustomerInfoResponse.Transaction] {
```
NachoSoto, line 177: `CustomerInfo` was combining the dictionaries inside of "subscriptions" and "non_subscriptions", despite them having different formats. This is now made explicit through type conversions between the two.

NachoSoto (commented):
This is ready for review now!

Sources/Identity/CustomerInfo.swift
joshdholtz: OMG, the new version is so ❤️ I was in here debugging something a while back and it was hard to follow. Great work!

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -120,45 +120,26 @@ class CustomerInfoManager {
...
-                    Logger.error("Error loading customer info from cache: \(error)")
-                }
-                return nil
-            }
-
-            if let schema = info.schemaVersion, schema == CustomerInfo.currentSchemaVersion {
+            if info.isInCurrentSchemaVersion {
                 return info
+            } else {
+                return nil
```
joshdholtz, line 128: Do we need to log anything here for this error? 🤔

Sources/Misc/RawDataContainer.swift
```diff
@@ -11,14 +11,33 @@
...
+    /// to future data while using an older version of the SDK.
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    public var rawData: [String: Any] {
+        do {
+            return try self.underlyingData.asDictionary()
+        } catch {
+            Logger.warn(Strings.codable.encoding_error(error))
+            return [:]
+        }
+    }
```
joshdholtz, line 41: Will this `rawData` not contain any data that was stripped away by the new lossy array and dictionary types? So like.. this won't show the raw response that this customer info was created from, right? It instead is just a dictionary representation of this object?

I'm not sure if that is a change from what this previously did but just wanted to understand and call that out just in case 😇

Sources/Networking/Responses/CustomerInfoResponse.swift
```diff
@@ -0,0 +1,190 @@
...
+//
+//      https://opensource.org/licenses/MIT
+//
+//  CustomerInfoResponse.swift
+//
+//  Created by Nacho Soto on 4/12/22.
+
+import Foundation
+
+struct CustomerInfoResponse {
```
joshdholtz, line 16: This whole file is soooooo 💯

Sources/Networking/Responses/CustomerInfoResponse.swift
```diff
@@ -0,0 +1,190 @@
...
+
+    var allTransactionsByProductId: [String: CustomerInfoResponse.Transaction] {
+        return self.allSubscriptionsByProductId.mapValues { $0.asTransaction }
+    }
+
+    var allSubscriptionsByProductId: [String: CustomerInfoResponse.Subscription] {
+        let subscriptions = self.subscriptions
+        let latestNonSubscriptionTransactionsByProductId = self.nonSubscriptions
+            .compactMapValues { $0.last }
+            .mapValues { $0.asSubscription }
```
joshdholtz, line 185: Why are non-subscriptions being included in `allSubscriptions` 🤔

joshdholtz (commented):
Leaving some early comments on my review while I continue reviewing 😊

Sources/Identity/CustomerInfoManager.swift
NachoSoto: We weren't logging anything before 🤷🏻‍♂️

Sources/Networking/Responses/CustomerInfoResponse.swift
NachoSoto: This was like this before:
```swift
self.allPurchases = latestNonSubscriptionTransactionsByProductId
    .merging(subscriptionTransactionsByProductId) { (current, _) in current }
```
Maybe I should change it to `allPurchasesByProductId` instead?

Sources/Misc/RawDataContainer.swift
NachoSoto: Oh shoot that's a great catch! Let me look into how we could support storing the original data.

Sources/Networking/Responses/CustomerInfoResponse.swift
joshdholtz: I _think_ that would make sense 🤷‍♂️

Sources/Identity/CustomerInfoManager.swift
joshdholtz: Probably fine then!

Sources/Networking/Responses/CustomerInfoResponse.swift
NachoSoto: DOne.

Sources/Identity/CustomerInfo.swift
```diff
@@ -164,263 +153,170 @@ import Foundation
...
 
+extension CustomerInfo: HTTPResponseBody {}
+
+// MARK: - Private
+
 private extension CustomerInfo {
 
-    static func createSubscriberDataError(_ error: Error, subscriberDictionary: [String: Any]) -> Error {
-        Logger.error(Strings.customerInfo.cant_instantiate_from_json_object(jsonObject: subscriberDictionary))
+    /// The actual contents of a `CustomerInfo`: the response with the associated version.
```
aboedo, line 255: `CustomerInfo` here should be in \`\`double backticks\`\` so you can click on it to navigate

aboedo, line 255: `CustomerInfo` here should be in `\`\double backticks \`\` so you can click on it to navigate

Sources/Identity/CustomerInfo.swift
```diff
@@ -164,263 +153,170 @@ import Foundation
...
+    struct Data {
```
aboedo, line 256: this confused me when reading the decoding code. I initially thought we'd added `response` and `schemaVersion` as extra properties in an extension of Foundation's Data. 
I think this approach makes sense, but perhaps we can call it something other than `Data` so there's no confusing it with a standard type?

Sources/Identity/CustomerInfo.swift
```diff
@@ -164,263 +153,170 @@ import Foundation
...
 }
 
-// Fixme: delete when `CustomerInfo` is `Decodable`
- extension CustomerInfo: HTTPResponseBody {
+private extension CustomerInfo {
+
+    func activeKeys(dates: [String: Date?]) -> Set<String> {
+        return Set(
+            dates
+                .lazy
```
aboedo, line 303: why is `lazy` needed here?

Sources/Misc/RawDataContainer.swift
```diff
@@ -11,14 +11,33 @@
...
 //  Created by Nacho Soto on 11/16/21.
 
-/// A type which exposes its underlying raw data, for debugging purposes or for getting access
+/// A type which exposes its underlying content for debugging purposes or for getting access
 /// to future data while using an older version of the SDK.
 public protocol RawDataContainer {
 
-    /// The type of the underlying raw data
-    associatedtype Content
+    /// The type of the `underlyingData` for this type.
```
aboedo, line 18: I think you get automatic linking to the symbol if you do \`\`RawDataContainer/underlyingData\`\` here

Sources/Misc/RawDataContainer.swift
```diff
@@ -11,14 +11,33 @@
...
-    /// The underlying data.
-    var rawData: Content { get }
+    /// The underlying content for debugging purposes or for getting access
+    /// to future data while using an older version of the SDK.
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    var underlyingData: Content { get }
+
+}
+
+/// Default implementation that encodes the `underlyingData` for public use.
```
aboedo, line 28: same thinking regarding auto linking to the symbol

Tests/UnitTests/Networking/Responses/CustomerInfoDecodingTests.swift
```diff
@@ -0,0 +1,96 @@
...
+        try super.setUpWithError()
+
+        self.customerInfo = try self.decodeFixture("CustomerInfo")
+    }
+
+    func testSchemaVersion() {
+        expect(self.customerInfo.schemaVersion) == "4"
+    }
+
+    func testResponseDataIsCorrect() throws {
```
aboedo, line 35: maybe we can spilt this up into multiple cases? the response is likely to grow over time, and the test case is already pretty big

aboedo, line 35: maybe we can spilt this up into multiple cases? the response is likely to grow over time, and the test case is already pretty big

aboedo (commented):
This looks amazing!! It's great to see the CustomerInfo decoding code go away, that was a pain to read. 

I left a few comments

Sources/Identity/CustomerInfo.swift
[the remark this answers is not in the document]
NachoSoto: Good point. The whole type is `CustomerInfo.Data`, maybe I can use that whole type for clarity?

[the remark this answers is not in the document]
aboedo: I think that works, although the reason I suggested renaming was so that in the future we don't use it by calling it `Data` and have this conversation again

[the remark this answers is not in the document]
NachoSoto: That's fair. Maybe `CustomerInfo.Contents`?

Sources/Identity/CustomerInfo.swift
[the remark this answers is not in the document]
NachoSoto: Avoids creating intermediate 3 intermediate collections, one for filtering, one for mapping, and then another copy for the set. This instead initializes the set going through each element only once.

Sources/Misc/RawDataContainer.swift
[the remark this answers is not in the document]
NachoSoto: Oh yeah thanks! Changed.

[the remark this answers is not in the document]
NachoSoto: Actually this is going away in #1565 (I put it separate because this was already big enough).

Sources/Purchasing/EntitlementInfos.swift
```diff
@@ -43,78 +43,55 @@ import Foundation
...
+                }
+
+                return (
+                    identifier,
+                    EntitlementInfo(identifier: identifier,
+                                    entitlement: entitlement,
+                                    subscription: subscription,
+                                    requestDate: requestDate)
+                )
+            }
```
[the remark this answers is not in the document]
NachoSoto: Yeah I guess it's personal preference, I'm happy to change the data manipulation method back.
What do you mean by "flat" though?

Tests/UnitTests/Networking/Responses/CustomerInfoDecodingTests.swift
NachoSoto: Good idea.

NachoSoto: Done. I changed the `setUp` to `class` so it only happens one for all tests.

NachoSoto: Done.

Sources/Misc/RawDataContainer.swift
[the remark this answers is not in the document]
NachoSoto: Never mind, the type is still there, so I updated the docstring 👍🏻

Sources/Identity/CustomerInfo.swift
aboedo: it seems like gh posted my comments once per time I edited them before posting the review? I find a different but in gh for every PR over 700 lines.

Sources/Identity/CustomerInfo.swift
NachoSoto: Changed!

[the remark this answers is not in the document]
NachoSoto: Changed!

Sources/Misc/RawDataContainer.swift
NachoSoto: Forgot to update this here, this is fixed by #1565.

Sources/Purchasing/EntitlementInfos.swift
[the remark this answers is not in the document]
NachoSoto: > well, not really flat, it's just that the new implementation does the filtering and matching between entitlementInfos and purchases as an operation within the constructor of the entitlementsByKey dictionary, which seems like a lot to do at once.

Yeah, one is imperative and this is functional. It's kind of a philosophical argument more than anything :P 

> It also took me a bit to realize that the entitlements in the signature and the one in declared in the method are different objects

Good point, I'll get rid of the unnecessary variable.

aboedo (approved):

Sources/Misc/RawDataContainer.swift
```diff
@@ -11,14 +11,33 @@
...
```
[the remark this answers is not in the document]
NachoSoto: I changed this in #1565 already :)

PR #1497, opened 2022-04-12
StoreKit 2 purchases: removed double request to `syncPurchases`

NachoSoto:
`StoreKit2TransactionListener` was triggering an asynchronous call to `syncPurchases` during transaction verification, and at the end of every purchase, another `syncPurchases` call was being made.
This refactor combines both calls into one by making `finish(transaction:)` wait for that call to finish.

### DIFF
--- a/Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
+++ b/Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
@@ -14,12 +14,18 @@
 import Foundation
 import StoreKit
 
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
 protocol StoreKit2TransactionListenerDelegate: AnyObject {
-    func transactionsUpdated()
+
+    func transactionsUpdated() async throws -> CustomerInfo
+
 }
 
 @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
 class StoreKit2TransactionListener {
+
+    /// Similar to ``PurchaseResultData`` but with an optional `CustomerInfo`
+    typealias ResultData = (userCancelled: Bool, customerInfo: CustomerInfo?, transaction: SK2Transaction?)
 
     private(set) var taskHandle: Task<Void, Never>?
     weak var delegate: StoreKit2TransactionListenerDelegate?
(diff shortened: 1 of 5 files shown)


Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
```diff
@@ -88,15 +96,16 @@ private extension StoreKit2TransactionListener {
...
+            return (verifiedTransaction, customerInfo)
         }
     }
 
-    func finish(transaction: StoreKit.Transaction) async {
+    func finish(transaction: StoreKit.Transaction) async throws -> CustomerInfo? {
         await transaction.finish()
-        delegate?.transactionsUpdated()
+
+        return try await delegate?.transactionsUpdated()
```
aboedo, line 108: `CustomerInfo` in the return here is nil only if the delegate is nil, right? Since the method in the delegate has a non-optional return value. 

Maybe we could make that more explicit with a `guard let delegate = self.delegate else { return nil }`

Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
```diff
@@ -48,20 +54,20 @@ class StoreKit2TransactionListener {
         self.taskHandle = nil
     }
 
-    /// - Returns: whether the user cancelled
+    /// - Returns: `nil` `CustomerInfo` if purchases were not synced
```
aboedo, line 57: this would only happen right now on a `userCancelled`, right?

NachoSoto: Or if the `delegate` is `nil`.

NachoSoto:
@aboedo sorry for the delay, just added your suggested change. Let me know if the `CustomerInfo` `nil` weirdness is still confusing.

Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
aboedo: right. Just to make sure I'm understanding the logic correctly, this is a delegate under our control, right? meaning that in practice it really shouldn't be `nil`, and only `userCancelled` would render a `nil`?

aboedo (approved):

Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
NachoSoto: Yeah that's right.

PR #1498, opened 2022-04-12
Unit Tests: localized test IAP to avoid warnings

NachoSoto:
Same as #1494 but for Unit Tests.
This makes the logs during tests a lot less noisy.


taquitos (approved):
👍

PR #1499, opened 2022-04-12
Pass storefront into ProductRequestData for receipt posting

taquitos:
[CF-434]

[CF-434]: https://revenuecats.atlassian.net/browse/CF-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/StoreKit1/StoreKitWrapper.swift
+++ b/Sources/Purchasing/StoreKit1/StoreKitWrapper.swift
@@ -34,6 +34,14 @@ class StoreKitWrapper: NSObject, SKPaymentTransactionObserver {
     @available(iOS 8.0, macOS 10.14, watchOS 6.2, macCatalyst 13.0, *)
     static var simulatesAskToBuyInSandbox = false
 
+    var currentStorefront: String? {
+        guard #available(iOS 13.0, tvOS 13.0, macOS 10.15, watchOS 6.2, *) else {
+            return nil
+        }
+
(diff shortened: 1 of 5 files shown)


Sources/Purchasing/StoreKit1/StoreKitWrapper.swift
```diff
@@ -34,6 +34,14 @@ class StoreKitWrapper: NSObject, SKPaymentTransactionObserver {
     @available(iOS 8.0, macOS 10.14, watchOS 6.2, macCatalyst 13.0, *)
     static var simulatesAskToBuyInSandbox = false
 
+    var currentStorefront: String? {
+        guard #available(iOS 13.0, tvOS 13.0, macOS 10.15, watchOS 6.2, *) else {
+            return nil
+        }
+
+        return self.paymentQueue.storefront?.countryCode
```
NachoSoto, line 42: Maybe we could create our own `Storefront` that abstracts `SKStorefront` and `StoreKit.Storefront` so we can pass that around instead of a String.

taquitos: Yeah, that's a good idea

NachoSoto: #1500

PR #1501, opened 2022-04-13
Added `XCTAsyncUnwrap` to simplify `XCTUnwrap` usage with `async`

NachoSoto:
Because Swift (still?) doesn't have support for "`reasync`", this currently required 2 lines.

PR #1500, opened 2022-04-13
Introduced `Storefront` to abstract SK1 `SKStorefront` and SK2 `StoreKit.Storefront`

NachoSoto:
To simplify #1499.

PR #1502, opened 2022-04-13
Added `Storefront.currentStorefront`

NachoSoto:
Follow up to #1500. Depends on #1501.

### DIFF
--- a/Sources/Purchasing/StoreKitAbstractions/Storefront.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/Storefront.swift
@@ -69,6 +69,32 @@ public class Storefront: NSObject, StorefrontType {
 
 }
 
+public extension Storefront {
+
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, macCatalyst 13.1, *)
+    private static var currentStorefrontType: StorefrontType? {
+        get async {
+            if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) {
+                let sk2Storefront = await StoreKit.Storefront.current
+                return sk2Storefront.map(SK2Storefront.init)
+            } else {
+                return SKPaymentQueue.default().storefront.map(SK1Storefront.init)
+            }
+        }
+    }
+
+    /// The current App Store storefront for the device.
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, macCatalyst 13.1, *)
+    static var currentStorefront: Storefront? {
+        get async {
+            return await self.currentStorefrontType.map(Storefront.from(storefront: ))
+        }
+    }
+
+}
+
+// MARK: -
+
 /// A type containing the location and unique identifier of an Apple App Store storefront.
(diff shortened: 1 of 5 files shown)


Sources/Purchasing/StoreKitAbstractions/Storefront.swift
```diff
@@ -69,6 +69,32 @@ public class Storefront: NSObject, StorefrontType {
...
+                return sk2Storefront.map(SK2Storefront.init)
+            } else {
+                return SKPaymentQueue.default().storefront.map(SK1Storefront.init)
+            }
+        }
+    }
+
+    /// The current App Store storefront for the device.
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, macCatalyst 13.1, *)
+    static var currentStorefront: Storefront? {
```
NachoSoto, line 88: I named it `currentStorefront` because `Storefront.current` would be ambiguous. Ideally we'd come up with a different name for our `Storefront` to avoid that but I couldn't think of anything.
We've done that in the past for other types by adding the word "store", like "StoreProduct", but that doesn't work in this case 😛

taquitos: It could be worse, but I'm ok with this.

taquitos (approved):
🛒 🐐

Tests/StoreKitUnitTests/StorefrontTests.swift
```diff
@@ -0,0 +1,47 @@
...
+
+        expect(currentStorefront.identifier) == expectedStorefront.id
+        expect(currentStorefront.countryCode) == expectedStorefront.countryCode
+    }
+
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, macCatalyst 13.1, *)
+    func testCurrentStorefrontSK1() async throws {
+        try AvailabilityChecks.iOS13APIAvailableOrSkipTest()
+
+        if #unavailable(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0) {
```
NachoSoto, line 36: Using the new `#unavailable` 😎

Tests/StoreKitUnitTests/StorefrontTests.swift
```diff
@@ -0,0 +1,47 @@
...
+            let expectedStorefront = try XCTUnwrap(SKPaymentQueue.default().storefront)
+            let currentStorefront = try await XCTAsyncUnwrap(await Storefront.currentStorefront)
+
+            expect(currentStorefront.identifier) == expectedStorefront.identifier
+            expect(currentStorefront.countryCode) == expectedStorefront.countryCode
+        } else {
+            throw XCTSkip("This test is for pre-iOS 15 APIs")
```
NachoSoto, line 43: To make this test run only BEFORE iOS 15 😄

Tests/StoreKitUnitTests/StorefrontTests.swift
taquitos: Already using Swift 5.6 features 👏

PR #1507, opened 2022-04-14
Added new `ErrorCode.offlineConnectionError`

NachoSoto:
This is a special enough error that it's worth distinguishing from other `ErrorCode.networkError`s.

I see a couple benefits:
- When opening an app that has the SDK, this used to print lots of errors that just said "Error performing request". Now it's more clear what's happening
- Users can respond to failed calls to the SDK more appropriately, whereas before it might not be clear why a request failed.

Also fixed `logErrorIfNeeded` not passing the file/function/line parameters to the log functions.

### DIFF
--- a/Sources/Error Handling/ErrorCode.swift
+++ b/Sources/Error Handling/ErrorCode.swift
@@ -57,6 +57,7 @@ import Foundation
     @objc(RCProductRequestTimedOut) case productRequestTimedOut = 32
     @objc(RCAPIEndpointBlocked) case apiEndpointBlockedError = 33
     @objc(RCInvalidPromotionalOfferError) case invalidPromotionalOfferError = 34
+    @objc(RCOfflineConnectionError) case offlineConnectionError = 35
 
     // swiftlint:enable missing_docs
 }
@@ -157,6 +158,9 @@ extension ErrorCode: DescribableError {
                    The information associated with this PromotionalOffer is not valid.
                    See https://rev.cat/ios-subscription-offers for more info.
                    """
+        case .offlineConnectionError:
+            return "Error performing request because the Internet Connection appears to be offline."
+
         @unknown default:
             return "Something went wrong."
         }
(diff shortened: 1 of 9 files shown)


Sources/Error Handling/ErrorCode.swift
```diff
@@ -157,6 +158,9 @@ extension ErrorCode: DescribableError {
                    The information associated with this PromotionalOffer is not valid.
                    See https://rev.cat/ios-subscription-offers for more info.
                    """
+        case .offlineConnectionError:
+            return "Error performing request because the Internet Connection appears to be offline."
```
taquitos, line 162: `Internet Connection` can be lowercase.

taquitos (approved):
Really big fan of this 🎉🐐

Sources/Error Handling/ErrorCode.swift
NachoSoto: Ok I'll change it.

Sources/Networking/HTTPClient.swift
```diff
@@ -187,13 +187,20 @@ private extension HTTPClient {
         if let networkError = networkError {
             return Result
                 .failure(networkError)
-                .mapError { error in
+                .mapError { error -> NetworkError in
                     if self.dnsChecker.isBlockedAPIError(networkError),
```
NachoSoto, line 191: I'm thinking it might be worth extracting this whole conversion out.

taquitos: I'm a big fan of that 😄

NachoSoto: Ugh I forgot I gave myself a comment cause I didn't get an e-mail for it lol
I'll do this in a separate PR.

taquitos: Hah, it seems separate PR worthy anyway

NachoSoto: #1518

PR #1503, opened 2022-04-13
CircleCI: added iOS 13 job

NachoSoto:
Will be useful to validate the tests after [CF-547].
Also considering that we support it, I think it's useful to run tests there too.

[CF-547]: https://revenuecats.atlassian.net/browse/CF-547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1510, opened 2022-04-14
`CircleCI`: install iOS 13 simulator to fix iOS 13 tests

NachoSoto:
Follow up to #1503.

Fixes the following error:
> [12:36:18]: No simulators found that are equal to the version of specifier (13.7) and greater than or equal to the version of deployment target (0)
> [12:36:18]: Ignoring 'iPhone 8 (13.7)', couldn’t find matching simulator
> [12:36:18]: Couldn't find any matching simulators for '["iPhone 8 (13.7)"]' - falling back to default simulator
> [12:36:18]: Found simulator "iPhone 8 (14.5)"

The tests weren't actually running on iOS 13.x.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -224,6 +224,11 @@ jobs:
     steps:
       - checkout
       - install-dependencies
+      - install-and-create-sim:
+          install-name: iOS 13.7 Simulator
+          sim-device-type: iPhone-8
+          sim-device-runtime: iOS-13-7
+          sim-name: iPhone 8 (13.7)
       - run:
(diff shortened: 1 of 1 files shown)


taquitos (approved):
At a certain point, we'll need to consider removing some sims so that PRs don't take a ton of time to run. We can always keep them on the merge->main. 
I feel like having two makes sense to me- like iOS12 and iOS15, but that's for another pr to discuss 😄

✅🐐

NachoSoto:
iOS 13 is **only** running on main as well!

PR #1512, opened 2022-04-14
Logging: removed duplicate log and cleared some warning noise

NachoSoto:
Clears a bunch of these errors:
> 2022-04-14 15:57:25.051397+0200 xctest[86968:5767146] [Purchases] - WARN: 🍎‼️ There is a problem with the SKPaymentTransaction missing a transaction date - this is an issue with the App Store.
> 2022-04-14 15:57:25.051530+0200 xctest[86968:5767146] [Purchases] - WARN: 🍎‼️ There is a problem with the SKPaymentTransaction missing a transaction identifier - this is an issue with the App Store.

### DIFF
--- a/Sources/Purchasing/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/PurchasesOrchestrator.swift
@@ -626,7 +626,6 @@ private extension PurchasesOrchestrator {
                         self.storeKitWrapper.finishTransaction(transaction)
                     }
                 } else {
-                    Logger.error(error)
                     completion?(storeTransaction, nil, purchasesError, false)
                 }
             }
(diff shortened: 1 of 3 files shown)


Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -626,7 +626,6 @@ private extension PurchasesOrchestrator {
                         self.storeKitWrapper.finishTransaction(transaction)
                     }
                 } else {
-                    Logger.error(error)
```
NachoSoto, line 629: This is already logged from:
```swift
let purchasesError = error.asPurchasesError
```

taquitos: I've been meaning to look into this- thank you!

taquitos (approved):
📝 📝 🐐

PR #1523, opened 2022-04-22
expose signed data in promo offers

aboedo:
This is what's causing the tests in https://github.com/RevenueCat/purchases-hybrid-common/pull/129 to fail - the type isn't exposed, so I'm temporarily using `@testable` just to compile locally, but we really can't have that in production. 

This PR exposes the type, so that in turn purchases-hybrid-common can send it to the hybrids. 

I'm not 100% sure if the hybrids actually ever use the type, but this is kind of the path of least resistance 😅 Since it means one less breaking change in the hybrids. I also don't really see an issue with exposing the type

### DIFF
--- a/Sources/Purchasing/StoreKitAbstractions/PromotionalOffer.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/PromotionalOffer.swift
@@ -31,7 +31,8 @@ public final class PromotionalOffer: NSObject {
 
     /// The ``StoreProductDiscount`` in this offer.
     @objc public let discount: StoreProductDiscount
-    let signedData: SignedData
+    /// The ``SignedData-swift.class`` provides information about the ``PromotionalOffer``'s signature.
+    @objc public let signedData: SignedData
 
     init(discount: StoreProductDiscountType, signedData: SignedData) {
         self.discount = StoreProductDiscount.from(discount: discount)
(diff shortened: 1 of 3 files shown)


taquitos (approved):
🎈🐐

🤙

Sources/Purchasing/StoreKitAbstractions/PromotionalOffer.swift
```diff
@@ -42,15 +43,30 @@ public final class PromotionalOffer: NSObject {
...
 
     /// Contains the details of a promotional offer discount that you want to apply to a payment.
-    internal struct SignedData {
-        var identifier: String
-        var keyIdentifier: String
-        var nonce: UUID
-        var signature: String
-        var timestamp: Int
+    @objc(RCPromotionalOfferSignedData)
+    class SignedData: NSObject {
```
NachoSoto, line 50: I had made this `internal` on purpose to show that it was just an implementation detail, but I guess if the hybrids need it...

Sources/Purchasing/StoreKitAbstractions/PromotionalOffer.swift
```diff
@@ -42,15 +43,30 @@ public final class PromotionalOffer: NSObject {
...
+        /// The subscription offer identifier.
+        @objc public private(set) var identifier: String
+        /// The key identifier of the subscription key.
+        @objc public private(set) var keyIdentifier: String
+        /// The nonce used in the signature.
+        @objc public private(set) var nonce: UUID
+        /// The cryptographic signature of the offer parameters, generated on RevenueCat's server.
+        @objc public private(set) var signature: String
+        /// The UNIX time, in milliseconds, when the signature was generated.
+        @objc public private(set) var timestamp: Int
```
NachoSoto, line 60: ⚠️ Can we make these immutable now that it's a `class` and we lost value semantics?

Sources/Purchasing/StoreKitAbstractions/PromotionalOffer.swift
```diff
@@ -42,15 +43,30 @@ public final class PromotionalOffer: NSObject {
...
+
+        init(identifier: String, keyIdentifier: String, nonce: UUID, signature: String, timestamp: Int) {
+            self.identifier = identifier
+            self.keyIdentifier = keyIdentifier
+            self.nonce = nonce
+            self.signature = signature
+            self.timestamp = timestamp
+            super.init()
```
NachoSoto, line 68: Nit: I think this needs to go first before setting the properties.

taquitos: > Nit: I think this needs to go first before setting the properties.

You only need to call `super.init()` if the subclass's initializer customizes some properties it is inheriting from (and it would go first). In this case, you can omit the call to `super.init()`

https://docs.swift.org/swift-book/LanguageGuide/Initialization.html# specifically search for `you can omit a call to super.init()` and it will bring you to the section that details this. I forget when they added this to swift, but I feel like it was maybe in 4.0.

PR #1525, opened 2022-04-22
expose setPushTokenString method

aboedo:
`purchases-hybrid-common` sets the push token as a string, since we can't send Data in the hybrids. 

It used to use a header that would expose the method, but that's not compatible with purchases-ios v4, since it was essentially exploiting how message-passing works for objc. 

This exposes the method, which I don't think is a bad thing, it seems kind of convenient if for whatever reason you already have it as a string.

### DIFF
--- a/Sources/Purchasing/Purchases.swift
+++ b/Sources/Purchasing/Purchases.swift
@@ -524,6 +524,18 @@ extension Purchases {
     }
 
     /**
+     * Subscriber attribute associated with the push token for the user.
+     *
+     * #### Related Articles
+     * -  [Subscriber attributes](https://docs.revenuecat.com/docs/subscriber-attributes)
+     *
+     * - Parameter pushToken: `nil` will delete the subscriber attribute.
+     */
+    @objc public func setPushTokenString(_ pushToken: String?) {
+        subscriberAttributesManager.setPushTokenString(pushToken, appUserID: appUserID)
+    }
+
+    /**
      * Subscriber attribute associated with the Adjust Id for the user.
      * Required for the RevenueCat Adjust integration.
      *
--- a/Tests/APITesters/ObjCAPITester/ObjCAPITester/RCPurchasesAPI.m
+++ b/Tests/APITesters/ObjCAPITester/ObjCAPITester/RCPurchasesAPI.m
@@ -113,6 +113,8 @@ BOOL isAnonymous;
     [p setDisplayName: @""];
     [p setPushToken: nil];
     [p setPushToken: [@"" dataUsingEncoding: NSUTF8StringEncoding]];
+    [p setPushTokenString: @""];
+    [p setPushTokenString: nil];
     [p setAdjustID: nil];
     [p setAdjustID: @""];
     [p setAppsflyerID: nil];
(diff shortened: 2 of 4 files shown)


Sources/Purchasing/Purchases.swift
```diff
@@ -523,6 +523,18 @@ extension Purchases {
...
 
+    /**
+     * Subscriber attribute associated with the push token for the user.
+     *
+     * #### Related Articles
+     * -  [Subscriber attributes](https://docs.revenuecat.com/docs/subscriber-attributes)
+     *
+     * - Parameter pushToken: `nil` will delete the subscriber attribute.
+     */
+    @objc public func setPushTokenString(_ pushToken: String?) {
```
NachoSoto, line 534: What do you think about overloading the name in Swift, and only changing it in Objective-C since it doesn't support it?
```suggestion
    @objc(setPushTokenString:)
    public func setPushToken(_ pushToken: String?) {
```

Sources/Purchasing/Purchases.swift
```diff
@@ -523,6 +523,18 @@ extension Purchases {
         subscriberAttributesManager.setPushToken(pushToken, appUserID: appUserID)
     }
 
+    /**
+     * Subscriber attribute associated with the push token for the user.
+     *
+     * #### Related Articles
+     * -  [Subscriber attributes](https://docs.revenuecat.com/docs/subscriber-attributes)
+     *
+     * - Parameter pushToken: `nil` will delete the subscriber attribute.
```
NachoSoto, line 532: Maybe reference the other method here, and this method in the other one?

Sources/Purchasing/Purchases.swift
aboedo: I should have added a comment on the PR to clarify that in advance

Sources/Purchasing/Purchases.swift
aboedo: great idea! adding

Sources/Purchasing/Purchases.swift
NachoSoto: Did you try `Purchases/setPushToken(_: String)` and `Purchases/setPushToken(_: Data)`?
I wonder if the "better" name is worth more than linking both methods with a DocC-string.

NachoSoto (approved):
Thanks for checking though! It's an ugly name (Swift has overloads for this purpose), but not the worst thing I suppose.

Sources/Purchasing/Purchases.swift
aboedo: image

I'd love to get that working, though, I just don't know a good way to do it

NachoSoto: We could just not add the related symbol? 🤷🏻‍♂️

aboedo: sure, but the other place where it happens is in Purchases.md([renders this site](https://revenuecat.github.io/purchases-ios-docs/4.2.1/documentation/revenuecat/purchases)) and in RevenueCat.md ([renders this site](https://revenuecat.github.io/purchases-ios-docs/4.2.1/documentation/revenuecat/)), which serve as indexes for public methods. 

If we add most methods in there but leave out any, they won't be rendered in the Revenuecat page at all, and they'll only show up at the bottom of the Purchases.md site

NachoSoto: Not worth wasting more time then, ship it!

PR #1531, opened 2022-04-25
`PackageType`: simplified `typesByDescription` and implemented `CustomDebugStringConvertible`

NachoSoto:


Sources/Purchasing/Package.swift
```diff
@@ -148,7 +158,7 @@ private extension PackageType {
      * - Parameter string: A string that maps to a enumeration value of type ``PackageType``
      * - Returns: a ``PackageType`` for the given string.
      */
-    class func packageType(from string: String) -> PackageType {
+    static func packageType(from string: String) -> PackageType {
```
NachoSoto, line 161: This doesn't need dynamic dispatch.

Sources/Purchasing/Package.swift
```diff
@@ -57,20 +74,13 @@ private extension PackageType {
...
 
-    static var typesByDescription: [String: PackageType] {
-        [
-            "$rc_lifetime": .lifetime,
-            "$rc_annual": .annual,
-            "$rc_six_month": .sixMonth,
-            "$rc_three_month": .threeMonth,
-            "$rc_two_month": .twoMonth,
-            "$rc_monthly": .monthly,
-            "$rc_weekly": .weekly
```
NachoSoto, line 71: These were all duplicated from `description` which was error prone.

aboedo (approved):

PR #1534, opened 2022-04-26
[CF-553] Post AdServices token

beylmk:
Posts the AdServices token to the backend.


Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostAdServicesTokenTests/iOS14-testPostAdServicesCallsHttpClient.6.json
```diff
@@ -0,0 +1,12 @@
+{
```
NachoSoto, line 1: Hmm something's weird here. The fact that there are 6 files means the operation is making 6 network requests.

beylmk: that's....not desirable... 

will take a look back here on monday. thanks for the info!

NachoSoto: it might not be your fault though. I saw the same issue in #1561, but I haven't been able to reproduce locally.
Since you did reproduce it it would be nice if you could debug why the request is happening multiple times 🙏🏻 or maybe it's something else entirely, but hopefully we can figure it out. I have 2 flights on Monday but might have some time to help out with this.

RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -1598,6 +1604,13 @@
 			path = Responses;
 			sourceTree = "<group>";
 		};
+		A55D5D56282B30EF00FA7623 /* Recovered References */ = {
```
beylmk, line 1607: unsure what i did to deserve this

aboedo: 😂 this is just Xcode telling you it likes you

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostAdServicesTokenTests/iOS14-testPostAdServicesCallsHttpClient.6.json
beylmk: @NachoSoto i confirmed the operation is not requesting multiple times, both when running this unit test and in the sample app... any other thoughts on why that could be happening?

RevenueCat.xcodeproj/project.pbxproj
beylmk: it happened after some merges, i wasn't the smartest with my branching on this 😿

Tests/UnitTests/Networking/Backend/BackendPostAttributionDataTests.swift
```diff
@@ -0,0 +1,39 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  BackendPostAttributionDataTests.swift
```
beylmk, line 10: another weird merge thing going on, this file should be deleted. looking into it now

Sources/Purchasing/Purchases.swift
```diff
@@ -1977,10 +1984,12 @@ private extension Purchases {
         updateAllCachesIfNeeded()
         dispatchSyncSubscriberAttributesIfNeeded()
 
+#if os(iOS) || os(macOS) || targetEnvironment(macCatalyst)
```
aboedo, line 1987: same thinking re: skipping catalyst

Sources/Purchasing/Purchases.swift
```diff
@@ -419,10 +422,12 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
         subscribeToAppStateNotifications()
         attributionPoster.postPostponedAttributionDataIfNeeded()
 
+#if os(iOS) || os(macOS) || targetEnvironment(macCatalyst)
```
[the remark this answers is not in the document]
aboedo: I'm not opposed to explicitly having it though

NachoSoto:
> We should figure out what's happening with the backend requests, I'm not sure why this would issue multiple ones.

I'm looking into it right now.

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostAdServicesTokenTests/iOS14-testPostAdServicesCallsHttpClient.6.json
NachoSoto: I couldn't reproduce locally though.

> Perhaps while taking snapshots, the app is becoming active multiple times quickly,

Can you explain what you mean by this? The duplicated snapshots come from the same test.

RevenueCat.xcodeproj/project.pbxproj
NachoSoto: You should be able to find this folder in Xcode and manually delete it.

NachoSoto (commented):
This looks good, but let's see if @aboedo's solution for the duplicate requests fixes it. You'll need to manually delete those duplicate snapshots.

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostAdServicesTokenTests/iOS14-testPostAdServicesCallsHttpClient.6.json
aboedo: I have almost no familiarity with how our snapshots are generated, I haven't had much time to play around with them. But my thinking was that if `appDidBecomeActive` gets called multiple times during the same test run, it would trigger multiple calls to this method. Same if we called configure multiple times in the same test run. 
Either of those would potentially start multiple calls to the method that fires the backend request

NachoSoto: Snapshots are generated every time `HTTPClient.perform(request)` is called (except in these tests, it's `MockHTTPClient`). So the presence of multiple files signals that the method is called multiple times throughout the lifetime of the test.

NachoSoto: I wonder if this was just a bad test run with a previous change. @beylmk I recommend removing the entire `__Snapshots__` directory and running the tests again. I'm doing that locally (even on iOS 12) and I don't reproduce the issue, it's only making the appropriate number of requests now.

beylmk: thanks @aboedo and @NachoSoto for the thoughts... i'll try deleting the directory and rerunning!

Sources/Purchasing/Purchases.swift
beylmk: FWIW we have ios + maccatalyst checks elsewhere -- `showPriceConsent` being one place..idk if we want to get those consistent?

beylmk:
@NachoSoto @aboedo this should be ready to re-review. deleting the snapshots and regenerating after andy's suggested change fixed things! thank you!

Sources/Purchasing/Purchases.swift
aboedo: it'd be great to establish clear rules for them, especially since they're mostly up to personal preference. Feel free to go with whichever you prefer for now

Sources/Attribution/AttributionPoster.swift
```diff
@@ -106,8 +109,28 @@ class AttributionPoster {
...
+        // set the cache in advance to avoid multiple post calls
+        let latestNetworkIdsAndAdvertisingIdsSentByNetwork =
+            self.deviceCache.latestNetworkAndAdvertisingIdsSent(appUserID: currentAppUserID)
+        var newDictToCache = latestNetworkIdsAndAdvertisingIdsSentByNetwork
+        newDictToCache[String(AttributionNetwork.adServices.rawValue)] = adServicesToken
+        self.deviceCache.set(latestNetworkAndAdvertisingIdsSent: newDictToCache, appUserID: currentAppUserID)
+
+         backend.post(adServicesToken: adServicesToken, appUserID: currentAppUserID) { error in
+             if let receivedNSError = error as NSError? {
+                 Logger.warn(Strings.attribution.adservices_token_post_failed(error: receivedNSError))
```
aboedo, line 127: should we just have this accept an `Error` so we don't need the cast? 
Also, is `receivedNSError` a `NSError?` at this point? if so, this would log even if `receivedNSError` is `nil`, right?

Sources/Attribution/AttributionPoster.swift
```diff
@@ -106,8 +109,28 @@ class AttributionPoster {
...
+
+                // if there's an error, reset the cache
+                newDictToCache[String(AttributionNetwork.adServices.rawValue)] = nil
+                self.deviceCache.set(latestNetworkAndAdvertisingIdsSent: newDictToCache, appUserID: currentAppUserID)
+             }
```
aboedo, line 132: should we have an `else` case here to log when it goes through correctly?

aboedo (commented):
looking great, left a question regarding logging an `NSError?`

NachoSoto:
Awesome! I can take a look later today

Sources/Attribution/AttributionPoster.swift
beylmk: it's a `BackendError`. i remember thinking there was a reason for the cast but after PTO i'm not remembering what that was.. switching to BackendError

Sources/Attribution/AttributionPoster.swift
```diff
@@ -106,8 +109,31 @@ class AttributionPoster {
             return
         }
 
-        Logger.debug("Logging attribution token for now to avoid lint warning: \(attributionToken)")
-        // post
+        self.post(adServicesToken: attributionToken)
+    }
+
+    func post(adServicesToken: String) {
```
NachoSoto, line 115: I think this can be `private`?

Tests/UnitTests/Mocks/MockDeviceCache.swift
```diff
@@ -221,4 +221,18 @@ class MockDeviceCache: DeviceCache {
...
+    var invokedSetLatestNetworkAndAdvertisingIdsSentCount = 0
+    var invokedSetLatestNetworkAndAdvertisingIdsSentParameters: (appUserID: String?, Void)?
+    var invokedSetLatestNetworkAndAdvertisingIdsSentParametersList = [(appUserID: String?, Void)]()
+
+    override func set(latestNetworkAndAdvertisingIdsSent: [String: String], appUserID: String) {
+        invokedSetLatestNetworkAndAdvertisingIdsSent = true
+        invokedSetLatestNetworkAndAdvertisingIdsSentCount += 1
+        invokedSetLatestNetworkAndAdvertisingIdsSentParameters = (appUserID, ())
+        invokedSetLatestNetworkAndAdvertisingIdsSentParametersList.append((appUserID, ()))
+        super.set(latestNetworkAndAdvertisingIdsSent: latestNetworkAndAdvertisingIdsSent, appUserID: appUserID)
```
NachoSoto, line 235: Does this need to call the real implementation? I don't think it's expected for the mock to update the underlying data.

Tests/UnitTests/Networking/Backend/BackendPostAdServicesTokenTests.swift
```diff
@@ -0,0 +1,42 @@
...
+
+@testable import RevenueCat
+
+class BackendPostAdServicesTokenTests: BaseBackendTests {
+
+    override func createClient() -> MockHTTPClient {
+        super.createClient(#file)
+    }
+
+    func testPostAdServicesCallsHttpClient() throws {
```
NachoSoto, line 27: 👍🏻

NachoSoto (commented):
Just a few minor comments!

Sources/Attribution/AttributionPoster.swift
```diff
@@ -106,8 +109,31 @@ class AttributionPoster {
...
+         backend.post(adServicesToken: adServicesToken, appUserID: currentAppUserID) { error in
+             guard let error = error else {
+                 Logger.debug(Strings.attribution.adservices_token_post_succeeded)
+                 return
+             }
+             
+             Logger.warn(Strings.attribution.adservices_token_post_failed(error: error))
+
+            // if there's an error, reset the cache
+            newDictToCache[String(AttributionNetwork.adServices.rawValue)] = nil
```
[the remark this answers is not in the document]
beylmk: yeah, it does always take an AttributionNetwork rawValue...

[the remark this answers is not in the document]
NachoSoto: Then I'd recommend changing the method, then you can use `dictionary.mapKeys { $0.rawValue }` inside :)

Tests/UnitTests/Mocks/MockBackend.swift
```diff
@@ -121,6 +121,24 @@ class MockBackend: Backend {
         completion(stubbedGetOfferingsCompletionResult!)
     }
 
+    var invokedPostAdServicesToken = false
+    var invokedPostAdServicesTokenCount = 0
+    var invokedPostAdServicesTokenParameters: (token: String, appUserID: String?)?
+    var invokedPostAdServicesTokenParametersList = [(token: String, appUserID: String?)]()
+    var stubbedPostAdServicesTokenCompletionResult: (BackendError?, Void)?
```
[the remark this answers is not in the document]
beylmk: was just copying the `stubbedPostSubscriberAttributesCompletionResult`, but agreed, will update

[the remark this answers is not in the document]
NachoSoto: Hmm yeah that seems wrong too. That's a leftover from Objective-C, we can use `Result` instead everywhere now :)

Sources/Attribution/AttributionPoster.swift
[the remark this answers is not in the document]
beylmk: on it!

Tests/UnitTests/Mocks/MockBackend.swift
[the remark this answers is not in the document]
beylmk: on it :)

Tests/UnitTests/Mocks/MockDeviceCache.swift
beylmk: yeah i did this to test that the postIfNeeded would skip if something was already set... suppose i can just use some booleans in MockDeviceCache though

beylmk:
@NachoSoto this should be ready for re-review! i believe i addressed all your comments..

Sources/Attribution/AttributionPoster.swift
```diff
@@ -134,22 +135,44 @@ class AttributionPoster {
         postponedAttributionData = postponedData
     }
 
+    private func post(adServicesToken: String) {
+        let currentAppUserID = self.currentUserProvider.currentAppUserID
+
+        // set the cache in advance to avoid multiple post calls
+        let latestNetworkIdsAndAdvertisingIdsSentByNetwork =
+            self.deviceCache.latestAdvertisingIdsByNetworkSent(appUserID: currentAppUserID)
+        var newDictToCache = latestNetworkIdsAndAdvertisingIdsSentByNetwork
```
NachoSoto, line 144: `latestNetworkIdsAndAdvertisingIdsSentByNetwork` seems used only to make a copy of it. You could define only one variable:
```suggestion
        var newDictToCache = self.deviceCache.latestAdvertisingIdsByNetworkSent(appUserID: currentAppUserID)
```

Sources/Caching/DeviceCache.swift
```diff
@@ -269,17 +269,34 @@ class DeviceCache {
...
 
-    func set(latestNetworkAndAdvertisingIdsSent: [String: String], appUserID: String) {
+    func set(latestAdvertisingIdsByNetworkSent: [AttributionNetwork: String], appUserID: String) {
         self.userDefaults.write {
-            $0.setValue(latestNetworkAndAdvertisingIdsSent,
+            // convert AttributionNetwork to Integer as String
+            let latestAdIdsByRawNetworkStringSent =
+                latestAdvertisingIdsByNetworkSent.reduce(into: [:]) { adIdsByRawNetworkString, adIdByNetwork in
+                    adIdsByRawNetworkString[String(adIdByNetwork.key.rawValue)] = adIdByNetwork.value
+                }
```
NachoSoto, line 298: Not for this PR, but this could be vastly simplified if we added `Dictionary.mapKeys`:
```swift
public func mapKeys<NewKey: Hashable>(_ transformer: (Key) -> NewKey) -> [NewKey: Value] {
	var result = [NewKey: Value](minimumCapacity: count)

	for (key, value) in self {
		result.updateValue(value, forKey: transformer(key))
	}

	return result
}
```
But I suppose we'd also need `compactMapKeys` to deal with `String` -> `AttributionNetwork` conversions.

Tests/UnitTests/Attribution/AttributionPosterTests.swift
```diff
@@ -76,43 +77,104 @@ class AttributionPosterTests: TestCase {
...
+        expect(self.backend.invokedPostAdServicesTokenCount) == 0
         expect(self.subscriberAttributesManager.invokedConvertAttributionDataAndSetCount) == 2
     }
 
+#if canImport(AdServices)
+    @available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *)
+    func testPostAdServicesTokenIfNeededSkipsIfAlreadySent() throws {
+        guard #available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *) else {
+            throw XCTSkip("Required API is not available for this test.")
+        }
```
NachoSoto, line 132: These 3 are duplicated in all of these tests.
Just a suggestion, feel free to ignore, but you could extract a `BaseAttributionPosterTests` that has the properties and `setUp` in this class that `AttributionPosterTests` and a new `AdServicesAttributionPosterTests` class can inherit.
Then you can put this inside of a shared `setUp` as well as making that whole class:
```swift
@available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *)
```

Tests/UnitTests/Mocks/MockAttributionFetcher.swift
```diff
@@ -16,8 +16,9 @@ class MockAttributionFetcher: AttributionFetcher {
     }
 
     var adServicesTokenCollectionCalled = false
+    var adServicesTokenToReturn: String? = "adServicesToken"
```
NachoSoto, line 19: I think the convention in other mocks is to call this `mockedAdServicesToken` or `stubbedAdServicesToken`.

Tests/UnitTests/Mocks/MockDeviceCache.swift
```diff
@@ -221,4 +221,25 @@ class MockDeviceCache: DeviceCache {
         invokedClearLatestNetworkAndAdvertisingIdsSentParameters = (appUserID, ())
         invokedClearLatestNetworkAndAdvertisingIdsSentParametersList.append((appUserID, ()))
     }
+
+    var invokedSetLatestNetworkAndAdvertisingIdsSent = false
+    var invokedSetLatestNetworkAndAdvertisingIdsSentCount = 0
+    var invokedSetLatestNetworkAndAdvertisingIdsSentParameters:
+        (adIdsByNetwork: [AttributionNetwork: String], appUserID: String?)?
+    var invokedSetLatestNetworkAndAdvertisingIdsSentParametersList =
+        [(adIdsByNetwork: [AttributionNetwork: String], appUserID: String?)?]()
```
NachoSoto, line 230: Does this need to be an array of optionals? I think we always append non-nil:
```suggestion
        [(adIdsByNetwork: [AttributionNetwork: String], appUserID: String?)]()
```

Tests/UnitTests/Purchasing/PurchasesTests.swift
```diff
@@ -230,6 +230,25 @@ class PurchasesTests: TestCase {
                 }
             )
         }
+
+        var invokedPostAdServicesToken = false
+        var invokedPostAdServicesTokenCount = 0
+        var invokedPostAdServicesTokenParameters: (token: String, appUserID: String?)?
+        var invokedPostAdServicesTokenParametersList = [(token: String, appUserID: String?)]()
+        var stubbedPostAdServicesTokenCompletionResult: (BackendError?, Void)?
```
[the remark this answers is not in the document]
NachoSoto: Bump

Tests/UnitTests/Attribution/AttributionPosterTests.swift
```diff
@@ -57,6 +57,7 @@ class AttributionPosterTests: TestCase {
                                               attributionFetcher: attributionFetcher,
                                               subscriberAttributesManager: subscriberAttributesManager)
         resetAttributionStaticProperties()
+        backend.stubbedPostAdServicesTokenCompletionResult = Result.success(())
```
NachoSoto, line 60: Nit: all these can be just `.success` or `.failure`, don't need `Result`.

Tests/UnitTests/Purchasing/PurchasesTests.swift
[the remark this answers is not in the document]
beylmk: i think this is a merge thing, will fix

Sources/Caching/DeviceCache.swift
beylmk: i can make a ticket!

beylmk: https://revenuecats.atlassian.net/browse/CSDK-105

Sources/Networking/Operations/PostAdServicesTokenOperation.swift
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        let request = HTTPRequest(method: .post(Body(aadAttributionToken: self.token)),
+                                  path: .postAdServicesToken(appUserID: appUserID))
+
+        self.httpClient.perform(
+            request,
+            authHeaders: self.authHeaders
+        ) { (response: HTTPResponse<HTTPEmptyResponseBody>.Result) in
+            completion()
```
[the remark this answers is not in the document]
beylmk: see andy's comment: https://github.com/RevenueCat/purchases-ios/pull/1534#discussion_r887136962

do we want both to be in the defer then?

beylmk:
@NachoSoto i know you already approved but just FYI i addressed comments (and replied to the one about `defer`)

Sources/Networking/Operations/PostAdServicesTokenOperation.swift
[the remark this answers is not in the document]
NachoSoto: Not sure I understand the comment, the completion here has to be called for both cases, and the response handler as well. My comment was regarding the fact that the operation shouldn't complete until we've called the completion handler.

Tests/UnitTests/Attribution/AttributionPosterTests.swift
```diff
@@ -57,6 +55,14 @@ class AttributionPosterTests: TestCase {
                                               attributionFetcher: attributionFetcher,
                                               subscriberAttributesManager: subscriberAttributesManager)
         resetAttributionStaticProperties()
+        backend.stubbedPostAdServicesTokenCompletionResult = .success(())
+    }
+
+    override func tearDown() {
+        super.tearDown()
```
NachoSoto, line 62: Nit: typically this is called at the end of the `tearDown` (Initialize the parent class first, and tear it down last).

Sources/Networking/Operations/PostAdServicesTokenOperation.swift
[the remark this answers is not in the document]
beylmk: @aboedo 👀 ^^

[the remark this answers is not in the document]
aboedo: I think both suggestions are just about the same - like Nacho said, we should call completion as the last step in the method, and what I suggested was to do it by having both this code path and the early return look similar, i.e.: 
```swift
guard let appUserID = try? self.configuration.appUserID.escapedOrError() else {
    defer { completion() } 
    self.responseHandler?(.missingAppUserID())
}

/// ...

{ (response: HTTPResponse<HTTPEmptyResponseBody>.Result) in
    defer { completion() } 
    self.responseHandler?(response.error.map(BackendError.networkError))
}
```

or 

```swift
guard let appUserID = try? self.configuration.appUserID.escapedOrError() else {
    self.responseHandler?(.missingAppUserID())
    completion()
}

/// ...

{ (response: HTTPResponse<HTTPEmptyResponseBody>.Result) in
    self.responseHandler?(response.error.map(BackendError.networkError))
    completion()
}
```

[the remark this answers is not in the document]
beylmk: ahhh i totally misunderstood your original comment, my bad. thank you for clarifying!

PR #1533, opened 2022-04-25
Prepare next version: 4.4.0-SNAPSHOT

revenuecat-ops:

PR #1536, opened 2022-04-26
`Fastfile`: fixed `replace_version_number` lane

NachoSoto:
See #1533.
The previous regular expression would incorrectly replace `55413C0025B778E200ECCA5A` with `554.4.0-SNAPSHOT025B778E200ECCA5A`, for example, because the `.`s in the previous version weren't escaped.


aboedo (commented):
great catch!

PR #1539, opened 2022-04-26
Fixed `ErrorUtils.logDecodingError`

NachoSoto:
## Changes:

- Improved output of decoding errors:

#### Before:
```
[Purchases] - ERROR: 😿‼️ Couldn't decode data from json. Error:
The operation couldn’t be completed. (RevenueCat.CodableError error 0.))
```

#### After:
```
[Purchases] - ERROR: 😿‼️ Couldn't decode data from json. Error:
Found unexpected value 'tienda' for type 'Store'
```

- `CodableStrings.decodingError` relied on `Error.localizedDescription`, so `CodableError` now conforms to `LocalizedError` to provide that as well
- Added incorrect `value` to the error to find out why it was invalid
- Removed extra `)` at the end of the error string
- Cleaned up `switch`es to use a single `let` syntax

### DIFF
--- a/Sources/CodableExtensions/PeriodType+Extensions.swift
+++ b/Sources/CodableExtensions/PeriodType+Extensions.swift
@@ -24,7 +24,7 @@ extension PeriodType: Decodable {
         }
 
         guard let type = Self.mapping[periodTypeString] else {
-            throw CodableError.unexpectedValue(PeriodType.self)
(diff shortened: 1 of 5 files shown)


NachoSoto:
Nothing like breaking decoding in the SDK to find out how good errors really are 😇

taquitos (approved):
Thousand Times Yes

PR #1543, opened 2022-04-26
Created `@LossyArray` and `@LossyDictionary`

NachoSoto:
Inspired by https://github.com/marksands/BetterCodable/blob/master/Sources/BetterCodable but supporting nested collections and with simplified code.

These property wrappers allows decoding `Array`s and `Dictionary`s and ignore elements that fail to decode.
For example, this will ignore any elements that aren't numbers, instead of failing to decode altogether.
```swift
struct Data: Decodable {
    @LossyArray var list: [Int]
    @LossyDictionary var map: [String: Int]
}
```
This does require that the values are the right type, but these wrappers can be composed with `@DefaultDecodable.EmptyArray` and `@DefaultDecodable.EmptyDictionary` introduced in #1537 to make it produce an empty array in case of any other type error:
```swift
struct Data: Decodable {
     @DefaultDecodable.EmptyArray @LossyArray var list: [Int]
     @DefaultDecodable.EmptyDictionary @LossyDictionary var map: [String: Int]
}
```

Because of limitations of the property wrappers in Swift, an extra type `LossyArrayDictionary` allows lossy decoding of types `[String: [Decodable]`.

This will be used to vastly simplify #1496.

### DIFF
--- a/Sources/FoundationExtensions/LossyCollections.swift
+++ b/Sources/FoundationExtensions/LossyCollections.swift
@@ -1,0 +1,221 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  LossyCollections.swift
+//
+//  Created by Nacho Soto on 4/26/22.
+
+import Foundation
+
+// Inspired by https://github.com/marksands/BetterCodable/blob/master/Sources/BetterCodable/
+
+// MARK: - LossyArray
+
+/// A property wrapper that allows decoding `Array`s and ignore elements that fail to decode.
+/// For example, this will ignore any elements that aren't numbers, instead of failing to decode altogether.
+/// ```
+/// struct Data: Decodable {
+///     @LossyArray var list: [Int]
+/// }
+/// ```
+/// This does require that the value is an `Array`,
+/// but this wrapper can be composed with `@DefaultDecodable.EmptyArray`
+/// to make it produce an empty array in case of any other type error.
+/// ```
+/// struct Data: Decodable {
+///      @DefaultDecodable.EmptyArray @LossyArray var list: [Int]
+/// }
+/// ```
+@propertyWrapper
+struct LossyArray<Value> {
+
+    var wrappedValue: [Value]
+
+    init(wrappedValue: [Value]) { self.wrappedValue = wrappedValue }
+
+}
+
+extension LossyArray: Decodable where Value: Decodable {
+
+    private struct AnyDecodableValue: Decodable {}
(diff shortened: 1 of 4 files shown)


Sources/FoundationExtensions/LossyCollections.swift
```diff
@@ -0,0 +1,221 @@
...
+
+}
+
+/// A property wrapper that allows decoding `Dictionary`s of `Array`s and ignore elements that fail to decode.
+/// This is a composition of `LossyArray` and `LossyDictionary`, that due to limitations of property wrappers,
+/// it necessitates its own type.
+@propertyWrapper
+struct LossyArrayDictionary<Value: Decodable> {
+
+    var wrappedValue: [String: [Value]]
```
joshdholtz, line 124: Is it safe to assume the key will always be a `String`? I mean, I can't see why it wouldn't be in any of our cases but not sure if that something we want to open up 🤷‍♂️

Tests/UnitTests/FoundationExtensions/DecoderExtensionTests.swift
```diff
@@ -89,14 +89,83 @@ class DecoderExtensionsIgnoreErrorsTests: XCTestCase {
...
+        let data = try Data.decode(json)
 
         expect(data.url).to(beNil())
     }
 
 }
 
+class DecoderExtensionsLossyCollectionTests: XCTestCase {
+
+    private struct Data: Codable, Equatable {
```
joshdholtz, line 102: I don't _think_ this needs to be changed but I was confused at how `Data` was being used thinking that all the `Data` objects in the tests were the `NSData` type of `Data` objects and then not realizing there is a custom `Data` struct here.

I would also name this something like `TestData` or something but that is just kind of a personal nit 🙃

joshdholtz (approved):
This looks awesome 😍 A few small questions but otherwise this looks 💯 and good to merge!

Sources/FoundationExtensions/LossyCollections.swift
NachoSoto: Good question. Swift's [`CodingKey`](https://developer.apple.com/documentation/swift/codingkey) allows keys to be `String` or `Int`, in a way that Integer keys can be implicitly converted to `String` when decoding JSON dictionaries. We don't use that, so for simplicity these property wrappers assume just `String`, but it would be _trivial_ to change these by adding a `CodingKey` `associatedtype` so that integer keys are supported as well.

Tests/UnitTests/FoundationExtensions/DecoderExtensionTests.swift
NachoSoto: Well this isn't just `Data` (as in the global namespace), it's `DecoderExtensionsLossyCollectionTests.Data` (it's namespaced inside the test class).

So in a way, it is like "Test Data". Does that make sense?

PR #1519, opened 2022-04-20
[CF-106] Fetch AdServices Token

beylmk:
AdServices token fetching and removal of old iAD code. Posting to the backend will be the next step.

* deprecates old automaticAppleSearchAdsAttributionCollection
* adds new automaticAdServicesAttributionTokenCollection
* fetches attribution token

There is a follow-up ticket to investigate linking, and we will probably switch to weak-linking then.

PR #1544, opened 2022-04-27
Move deprecations

beylmk:
As per discussion [here](https://github.com/RevenueCat/purchases-ios/pull/1519#discussion_r854523265), we should have a single place for deprecations.

I'm not sure if it's worth making purchasesOrchestrator and that one post method internal, though. If it is, perhaps we want to add a note to make them private again once the functions are obsoleted. Opening a quick PR for discussion.


NachoSoto:
> I'm not sure if it's worth making purchasesOrchestrator and that one post method internal, though.

Oh yeah definitely I don't think it's worth it. I'd keep the methods that require those for sure. Sorry I didn't realize that! I don't think the trade off is worth it.

Sources/Misc/Deprecations.swift
```diff
@@ -56,6 +56,64 @@ public extension Purchases {
         return await eligiblePromotionalOffers(forProduct: product)
     }
 
+    /**
+     * Enable debug logging. Useful for debugging issues with the lovely team @RevenueCat.
+     */
+    @available(*, deprecated, message: "use Purchases.logLevel instead")
+    @objc static var debugLogsEnabled: Bool {
```
aboedo, line 63: [not for this PR] I'm surprised the others in this file don't have `@objc`, did we miss that?

Sources/Misc/Deprecations.swift
```diff
@@ -56,6 +56,64 @@ public extension Purchases {
...
+        get { logLevel == .debug }
+        set { logLevel = newValue ? .debug : .info }
+    }
+
+    /**
+     * Deprecated
```
aboedo, line 69: ooff I guess we missed adding docs for this one (before this PR)

Could we add a link to https://docs.revenuecat.com/docs/user-ids and maybe a comment like 
```suggestion
     * Deprecated: Configure  behavior through the RevenueCat dashboard instead.
     * Determines whether or not users may reuse a subscription from an App or Play Store account that already has that subscription active.
     * ### Related articles: 
     * - https://docs.revenuecat.com/docs/user-ids
```

aboedo (approved):

Sources/Purchasing/Purchases.swift
```diff
@@ -226,7 +226,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
     private let productsManager: ProductsManager
     private let customerInfoManager: CustomerInfoManager
     private let trialOrIntroPriceEligibilityChecker: TrialOrIntroPriceEligibilityChecker
-    private let purchasesOrchestrator: PurchasesOrchestrator
+    internal let purchasesOrchestrator: PurchasesOrchestrator
```
NachoSoto, line 229: Like I said, I don't think this is worth the tradeoff of moving these out 🙏🏻

NachoSoto (changes requested):

Sources/Misc/Deprecations.swift
beylmk: made it for this PR 🤷 since the original work got undone anyway

Tests/UnitTests/Purchasing/PurchasesTests.swift
```diff
@@ -2182,42 +2182,6 @@ class PurchasesTests: XCTestCase {
         expect(invokedMethodParams.appUserID) == identityManager.currentAppUserID
     }
 
-    func testAttributionDataDontSendNetworkAppUserIdIfNotProvided() throws {
```
NachoSoto, line 2185: Do these need to be removed now?

beylmk: i thought i was on the adservices branch 🤦‍♀️ reverted

NachoSoto (approved):

Sources/Misc/Deprecations.swift
```diff
@@ -33,7 +33,7 @@ public extension Purchases {
     @available(watchOS, introduced: 6.2, deprecated: 1, renamed: "checkTrialOrIntroDiscountEligibility(productIdentifiers:)")
     @available(macOS, introduced: 10.15, deprecated: 1, renamed: "checkTrialOrIntroDiscountEligibility(productIdentifiers:)")
     @available(macCatalyst, introduced: 13.0, deprecated: 1, renamed: "checkTrialOrIntroDiscountEligibility(productIdentifiers:)")
-    func checkTrialOrIntroDiscountEligibility(_ productIdentifiers: [String]) async -> [String: IntroEligibility] {
+    @objc func checkTrialOrIntroDiscountEligibility(_ productIdentifiers: [String]) async -> [String: IntroEligibility] {
```
NachoSoto, line 36: I'm confused, why do these need `@objc`? it's an `async` version, the non-async is above, and that one doesn't have `@objc`?

PR #1545, opened 2022-04-27
`SK1StoreProductDiscount`: handle `SKProductDiscount.priceLocale` being `nil` and created `StoreKitWorkarounds`

### LINKED ISSUE #1521: App crashes during SKProductsRequest on iOS 13
On iOS 13 the app crashes during SKProductsRequest.

When SK1StoreProductDiscount is initialised it crashes on line 32: `self.currencyCode = sk1Discount.priceLocale.currencyCode` at `priceLocale`. 

1. Environment
   1. Platform: **iOS**
   2. SDK version: **4.2.1** _also happnes with 4.1.0_
   3. StoreKit 2 enabled (Y/N): **Yes**
   4. OS version: **13.6.1**
   5. Xcode version: **13.3**
   6. How widespread is the issue. Percentage of devices affected: **All devices with iOS 13 I guess**
   
Before the crash, there's just:
```
[Purchases] - DEBUG: 😻 SKProductsRequest did finish
[Purchases] - DEBUG: 😻 SKProductsRequest request received response
```

Just call `Purchases.configure(withAPIKey` and it happens every time.

```
Exception Type:  EXC_BREAKPOINT (SIGTRAP)
Exception Codes: 0x0000000000000001, 0x00000001d1360494
Termination Signal: Trace/BPT trap: 5
Termination Reason: Namespace SIGNAL, Code 0x5
Terminating Process: exc handler [6572]
Triggered by Thread:  0

Thread 0 name:
Thread 0 Crashed:
0   libswiftFoundation.dylib      	0x00000001d1360494 static Locale._unconditionallyBridgeFromObjectiveC(_:) + 332 (Locale.swift:473)
1   AppBlock                      	0x00000001051a7bc4 init + 84 (StoreProductDiscount.swift:169)
2   AppBlock                      	0x00000001051a7bc4 init + 84 (<compiler-generated>:85)
3   AppBlock                      	0x00000001051a7bc4 compactMap + 84 (SK1StoreProduct.swift:0)
4   AppBlock                      	0x00000001051a7bc4 SK1StoreProduct.discounts.getter + 1144
5   AppBlock                      	0x0000000105182010 discounts.get + 40 (StoreProduct.swift:96)
6   AppBlock                      	0x0000000105182010 extractDiscounts + 64 (ProductRequestData+Initialization.swift:71)
7   AppBlock                      	0x0000000105182010 specialized ProductRequestData.init(with:) + 1136 (ProductRequestData+Initialization.swift:30)
8   AppBlock                      	0x0000000105193fd4 init + 8 (PurchasesOrchestrator.swift:0)
9   AppBlock                      	0x0000000105193fd4 PurchasesOrchestrator.postReceipt(withTransaction:receiptData:products:) + 76
10  AppBlock                      	0x0000000105198b7c partial apply for closure #1 in PurchasesOrchestrator.fetchProductsAndPostReceipt(withTransaction:receiptData:) + 76 (PurchasesOrchestrator.swift:561)
11  AppBlock                      	0x000000010518fd20 closure #1 in closure #2 in PurchasesOrchestrator.products(withIdentifiers:completion:) + 256 (PurchasesOrchestrator.swift:117)
12  AppBlock                      	0x000000010518f2f0 thunk for @escaping @callee_guaranteed () -> () + 28 (<compiler-generated>:0)
13  libdispatch.dylib             	0x000000019b0959a8 _dispatch_call_block_and_release + 24 (init.c:1408)
14  libdispatch.dylib             	0x000000019b096524 _dispatch_client_callout + 16 (object.m:495)
15  libdispatch.dylib             	0x000000019b0485b4 _dispatch_main_queue_callback_4CF$VARIANT$mp + 904 (inline_internal.h:2484)
16  CoreFoundation                	0x000000019b34e6bc __CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__ + 12 (CFRunLoop.c:1749)
17  CoreFoundation                	0x000000019b349590 __CFRunLoopRun + 1724 (CFRunLoop.c:3069)
18  CoreFoundation                	0x000000019b348ba8 CFRunLoopRunSpecific + 424 (CFRunLoop.c:3192)
19  GraphicsServices              	0x00000001a54b8344 GSEventRunModal + 160 (GSEvent.c:2246)
20  UIKitCore                     	0x000000019f4843e4 UIApplicationMain + 1932 (UIApplication.m:4823)
21  libswiftUIKit.dylib           	0x00000001d15dd2b4 UIApplicationMain(_:_:_:_:) + 168 (UIKit.swift:528)
22  AppBlock                      	0x0000000104e6ff80 main + 80 (<compiler-generated>:18)
23  AppBlock                      	0x0000000104e6ff80 $main + 92 (LicensesViewController.swift:0)
24  AppBlock                      	0x0000000104e6ff80 main + 108
25  libdyld.dylib                 	0x000000019b1d08f0 start + 4
```

NachoSoto:
Fixes #1521.
Despite `SKProductDiscount.priceLocale` being non-optional, StoreKit might return `nil` `NSLocale`s.
This works around that to make sure the SDK doesn't crash when bridging to `Locale`.

### DIFF
--- a/RevenueCat.xcodeproj/project.pbxproj
+++ b/RevenueCat.xcodeproj/project.pbxproj
@@ -932,11 +932,11 @@
 				57C381DB27961547009E3940 /* SK2StoreProductDiscount.swift */,
 				57536A27278522B400E2AE7F /* SK2StoreTransaction.swift */,
 				57DE806C28074976008D6C6F /* Storefront.swift */,
+				573E7F082819B989007C9128 /* StoreKitWorkarounds.swift */,
 				FECF627761D375C8431EB866 /* StoreProduct.swift */,
 				B372EC53268FEDC60099171E /* StoreProductDiscount.swift */,
 				2D1015D9275959840086173F /* StoreTransaction.swift */,
 				2D1015DD275A57FC0086173F /* SubscriptionPeriod.swift */,
-				573E7F082819B989007C9128 /* StoreKitWorkarounds.swift */,
 			);
 			path = StoreKitAbstractions;
 			sourceTree = "<group>";
@@ -995,7 +995,6 @@
 		2D1DB44A26EA61AB00DDE736 /* StoreKit1 */ = {
 			isa = PBXGroup;
 			children = (
-				F5E4383626B852D700841CC8 /* StoreKitExtensions */,
 				EFB3CBAA73855779FE828CE2 /* ProductsFetcherSK1.swift */,
 				2D991AC9268BA56900085481 /* StoreKitRequestFetcher.swift */,
 				2D4E926426990AB1000E10B0 /* StoreKitWrapper.swift */,


Sources/Misc/PriceFormatterProvider.swift
```diff
@@ -20,26 +20,24 @@ class PriceFormatterProvider {
...
         func makePriceFormatterForSK1(with locale: Locale) -> NumberFormatter {
             let formatter = NumberFormatter()
             formatter.numberStyle = .currency
             formatter.locale = locale
             return formatter
         }
 
-        if cachedPriceFormatterForSK1 == nil || cachedPriceFormatterForSK1?.locale != locale {
-            cachedPriceFormatterForSK1 = makePriceFormatterForSK1(with: locale)
+        if self.cachedPriceFormatterForSK1 == nil || self.cachedPriceFormatterForSK1?.locale != locale {
```
NachoSoto, line 30: Just some cleanup in this file to make it consistent with the style guide.

Tests/UnitTests/Mocks/MockSK1Product.swift
```diff
@@ -51,7 +51,7 @@ class MockSK1Product: SK1Product {
 
     @available(iOS 12.2, macCatalyst 13.0, tvOS 12.2, macOS 10.13.2, *)
     override var discounts: [SKProductDiscount] {
-        return (mockDiscount != nil) ? [mockDiscount!] : []
+        return self.mockDiscount.map { [$0] } ?? []
```
NachoSoto, line 54: No force unwrap.

Tests/UnitTests/Mocks/MockSKProductDiscount.swift
```diff
@@ -7,17 +7,24 @@ import Foundation
...
+    var mockLocale: Locale? = Locale(identifier: "USD")
+    override var priceLocale: Locale {
+        return self.mockLocale ?? super.priceLocale
+    }
+
+}
+
+// See https://github.com/RevenueCat/purchases-ios/issues/1521
+@available(iOS 11.2, macCatalyst 13.0, tvOS 11.2, macOS 10.13.2, *)
+class MockSKProductDiscountWithNoPriceLocale: SKProductDiscount {
```
NachoSoto, line 21: I had to make a separate `class` to be able to simulate this. Overriding the property was making it crash here without letting the extension detect the `nil` `priceLocale`.

Sources/Purchasing/StoreKitAbstractions/SK1StoreProductDiscount.swift
```diff
@@ -77,3 +79,20 @@ private extension StoreProductDiscount.PaymentMode {
...
 
 }
+
+@available(iOS 11.2, macOS 10.13.2, tvOS 11.2, watchOS 6.2, *)
+private extension SK1ProductDiscount {
+
+    // See https://github.com/RevenueCat/purchases-ios/issues/1521
+    // Despite `SKProductDiscount.priceLocale` being non-optional, StoreKit might return `nil` `NSLocale`s.
+    // This works around that to make sure the SDK doesn't crash when bridging to `Locale`.
+    var optionalLocale: Locale? {
```
aboedo, line 89: the sadness. 

We've seen [these a few times before](https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/StoreKit1/StoreKitExtensions/SKProduct%2BExtensions.swift#L19), I wonder if we should start having like a file `StoreKit+NullabilityFixes`

aboedo (approved):
great work! another storekit bug bites the dust

Sources/Purchasing/StoreKitAbstractions/SK1StoreProductDiscount.swift
NachoSoto: Actually I love that idea. `StoreKitWorkarounds`? To isolate the "real" code versus special cases for Apple bugs, like a certain other PDF company used to do with UIKit. That way we can keep track of these and hopefully deprecate them over time.

aboedo: ❤️ it

NachoSoto: Done.

Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+}
+
+extension SKPaymentTransaction {
+
+    /// Considering issue https://github.com/RevenueCat/purchases-ios/issues/279, sometimes `payment`
+    /// and `productIdentifier` can be `nil`, in this case, they must be treated as nullable.
+    /// Due to that an optional reference is created so that the compiler would allow us to check for nullability.
+    var paymentIfPresent: SKPayment? {
```
aboedo, line 38: LOL I'd forgotten about this one, I was thinking about the one in `SKProduct+Extensions`. I then looked up the words "although" and "even though" in our codebase and they're all tied to storekit bugs 😂

aboedo: we don't have to unify them all in this PR, though, we can keep consolidating in the future

aboedo (approved):

Sources/Logging/Strings/StoreKitStrings.swift
```diff
@@ -71,6 +73,10 @@ extension StoreKitStrings: CustomStringConvertible {
         case .sk1_no_known_product_type:
             return "This StoreProduct represents an SK1 product, the type of product cannot be determined, " +
             "the value will be undefined. Use `StoreProduct.productCategory` instead."
+
+        case .sk1_discount_missing_locale:
+            return "There is a problem with this SKProductDiscount missing a Locale - " +
+            "this is an issue with the App Store."
```
taquitos, line 79: "This is an issue with the App Store" is there anything we can communicate around what the developer can do in this case? cc @codykerns  for visibility

NachoSoto: I'm waiting to [hear back from the user regarding why they think it might be `nil`](https://github.com/RevenueCat/purchases-ios/issues/1521#issuecomment-1111186139). Once we get more context maybe we can update this.

Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
NachoSoto: Haha that's funny I forgot about that too, so this is 3 of them. I just moved it too.

Sources/Logging/Strings/StoreKitStrings.swift
NachoSoto: The user sent us some screenshots and I can't think of anything wrong in his set up. My only guess is a StoreKit bug related to the CZK App Store leading to a `nil` `NSLocale`. So @taquitos to your point, I can't think of anything the developer can do in this case, so we might just be able to communicate "hey this isn't our bug, sorry".
Maybe I'll change this to say "We will default to the device's current Locale".
What do you think /cc @aboedo?

aboedo: Historically we've gone with stuff like "this is an App Store quirk". I don't think there's much that we can do - defaulting to another Locale could be dangerous, like, we could end up showing the wrong prices for a product, and that's kind of worse than just erroring out or even crashing. 
In this particular case, Apple is breaking the nullability contract, so I don't imagine it's tied to any specific setup on the dev side. 

Hopefully SK2 will reduce the number of times we run into stuff like this given it's built in Swift

NachoSoto: > defaulting to another Locale could be dangerous, like, we could end up showing the wrong prices for a product, and that's kind of worse than just erroring out or even crashing.

That's a good point. That's what this PR is doing though. We could make it ignore the discount, but then this user's discount won't work at all :/

Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
```diff
@@ -7,13 +7,47 @@
...
 //
-//  Created by Nacho Soto on 12/13/21.
+//  Created by Nacho Soto on 4/27/22.
 
 import StoreKit
 
+@available(iOS 11.2, macOS 10.13.2, tvOS 11.2, watchOS 6.2, *)
+extension SK1ProductDiscount {
+
+    // See https://github.com/RevenueCat/purchases-ios/issues/1521
```
joshdholtz, line 19: I love seeing backlinks to issues 😍

joshdholtz (approved):
🔥 🔥 🔥 LGTM

Sources/Logging/Strings/StoreKitStrings.swift
NachoSoto: @joshdholtz @aboedo what do you guys think? We have no more info about what could cause this so I don't know what else we could add.

joshdholtz: @NachoSoto I'm out of ideas what could cause this issue. I really don't think its anything we or the developer could do here. I think this is the best way we can handle for now 😇

taquitos (approved):
I like this.

Double check the alphabetization of the `StoreKitAbstractions` group 😄

NachoSoto:
> Double check the alphabetization of the StoreKitAbstractions group 😄

Fixed! I wonder if we should automate this though, otherwise it feels like busy work :/

PR #1547, opened 2022-04-27
Improved `EntitlementInfosTests`

NachoSoto:
I broke these tests in a refactor and these improvements are making it much easier to figure out what's failing. Right now the failures come from the `verify` methods so it's hard to find the root issue.

## Changes:
- Use `#file` and `#line` to make failures show up in the right line
- Use `ISO8601DateFormatter`
- Moved verification methods to an extension separate from the tests


NachoSoto:
Yeah, though I wish `@_transparent` was public to make this simpler :P

PR #1551, opened 2022-04-28
`PeriodType`/`PurchaseOwnershipType`/`Store`: conform to `Encodable`

NachoSoto:
These will be used in an upcoming PR.


NachoSoto:
I should add tests for these.

NachoSoto:
Adding tests in a future PR because I want to use a helper method added in #1537.

NachoSoto:
Added tests: https://github.com/RevenueCat/purchases-ios/pull/1558

PR #1556, opened 2022-05-07
Create Configuration object and API to configure Purchases

taquitos:
Resolves [CF-443]

Encapsulate all configuration options into a single `Configuration` object that is used to configure the `Purchases` object. This object has two new properties- `networkTimeout: TimeInterval` and `storeKit1Timeout: TimeInterval`

To create a `Configuration` object you use the `Configuration.Builder` object.

New API on `Purchases`:
```
@objc(configureWithConfiguration:)
@discardableResult static func configure(with configuration: Configuration) -> Purchases
```

[CF-443]: https://revenuecats.atlassian.net/browse/CF-443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sources/Networking/HTTPClient.swift
```diff
@@ -18,23 +18,29 @@ class HTTPClient {
...
     private let state: Atomic<State> = .init(.initial)
     private let eTagManager: ETagManager
     private let dnsChecker: DNSCheckerType.Type
 
     init(
         systemInfo: SystemInfo,
         eTagManager: ETagManager,
-        dnsChecker: DNSCheckerType.Type = DNSChecker.self
+        dnsChecker: DNSCheckerType.Type = DNSChecker.self,
+        timeoutSeconds: Int = Configuration.networkTimeoutSecondsDefault
```
NachoSoto, line 33: Should we use `TimeInterval` here too?

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,150 @@
...
+ * To configure your `Purchases` instance using this object, follow these steps.
+ *
+ * **Steps:**
+ * 1. Call ``Configuration/configurationBuilder(withAPIKey:)`` To obtain a ``ConfigurationBuilder`` object.
+ * 2. Set this builder's properties using the "`with(`" functions.
+ * 3. Call ``ConfigurationBuilder/build()`` to obtain the `Configuration` object.
+ * 4. Pass the `Configuration` object into ``Purchases/configure(withConfiguration:)``.
+ *
+ */
+@objc(RCConfiguration) public class Configuration: NSObject {
```
NachoSoto, line 29: Yesssss this is a great refactor

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,150 @@
...
+    let apiKey: String
+    let appUserID: String?
+    let observerMode: Bool
+    let userDefaults: UserDefaults?
+    let storeKit2Setting: StoreKit2Setting
+    let dangerousSettings: DangerousSettings?
+    let networkTimeoutSeconds: Int
+    let storeKit1TimeoutSeconds: Int
+
+    override public init() {
```
NachoSoto, line 44: Is this needed? If we don't implement it it wouldn't be callable?
Or we can also mark it `unavailable`

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,150 @@
...
+
+    private init(withConfigurationBuilder configurationBuilder: ConfigurationBuilder) {
+        self.apiKey = configurationBuilder.apiKey
+        self.appUserID = configurationBuilder.appUserID
+        self.observerMode = configurationBuilder.observerMode
+        self.userDefaults = configurationBuilder.userDefaults
+        self.storeKit2Setting = configurationBuilder.storeKit2Setting
+        self.dangerousSettings = configurationBuilder.dangerousSettings
+
+        if configurationBuilder.storeKit1TimeoutSeconds < 5 {
```
NachoSoto, line 56: This can be `self.timeout = max(5, timeout)`

Sources/Purchasing/Purchases.swift
```diff
@@ -1828,13 +1862,17 @@ public extension Purchases {
                                                       observerMode: Bool,
                                                       userDefaults: UserDefaults?,
                                                       storeKit2Setting: StoreKit2Setting,
+                                                      storeKitTimeoutSeconds: Int,
+                                                      networkTimeoutSeconds: Int,
                                                       dangerousSettings: DangerousSettings?) -> Purchases {
         let purchases = Purchases(apiKey: apiKey,
```
NachoSoto, line 1868: Should we mark the other configure methods as deprecated?

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
```diff
@@ -173,9 +173,7 @@ extension ProductsFetcherSK1: SKProductsRequestDelegate {
                     completion(.failure(error))
                 }
             } else {
-                let delayInSeconds = Int((self.requestTimeout.seconds / 10).rounded())
-
-                queue.asyncAfter(deadline: .now() + .seconds(delayInSeconds)) { [self] in
+                queue.asyncAfter(deadline: .now() + .seconds(3)) { [self] in
```
NachoSoto, line 176: Why 3?

Tests/APITesters/SwiftAPITester/SwiftAPITester/ConfigurationAPI.swift
```diff
@@ -0,0 +1,24 @@
...
+//  SwiftAPITester
+//
+//  Created by Joshua Liebowitz on 5/6/22.
+//
+
+import Foundation
+import RevenueCat
+
+func checkConfigurationAPI() {
+    let configuration = Configuration
```
NachoSoto, line 12: Love it

NachoSoto (commented):
I love this refactor! I can take a closer look on Monday but just leaving a couple of comments for now :)

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,150 @@
...
+ * ``Configuration`` can be used when configuring the ``Purchases`` instance. It is not required to be used, but
+ * highly recommended. This class follows a builder pattern.
+ *
+ * To configure your `Purchases` instance using this object, follow these steps.
+ *
+ * **Steps:**
+ * 1. Call ``Configuration/configurationBuilder(withAPIKey:)`` To obtain a ``ConfigurationBuilder`` object.
+ * 2. Set this builder's properties using the "`with(`" functions.
+ * 3. Call ``ConfigurationBuilder/build()`` to obtain the `Configuration` object.
+ * 4. Pass the `Configuration` object into ``Purchases/configure(withConfiguration:)``.
```
aboedo, line 26: We could even have a swift example here by doing something like 
```
* ```swift
* let configuration = Configuration.configurationBuilder(...
* ...
* ```
*
```

Sources/Purchasing/Configuration.swift
aboedo: ❤️ it

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,150 @@
...
+    let storeKit2Setting: StoreKit2Setting
+    let dangerousSettings: DangerousSettings?
+    let networkTimeoutSeconds: Int
+    let storeKit1TimeoutSeconds: Int
+
+    override public init() {
+        fatalError("Use static function configurationBuilder(withAPIKey:) to configure.")
+    }
+
+    private init(withConfigurationBuilder configurationBuilder: ConfigurationBuilder) {
```
aboedo, line 48: I think the swifty way here would be 
```suggestion
    private init(with configurationBuilder: ConfigurationBuilder) {
```

Sources/Purchasing/Configuration.swift
aboedo: feels like we're hardcoding the magical `5` which would be a `minTimeoutInSeconds`, maybe we can make that a property for readability? 
Also, if you do pass in a timeout that's lower than the min, and we're overriding that value, it feels like we should log a warning so you know that your timeout isn't being used

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,150 @@
...
+        private(set) var observerMode: Bool = false
+        private(set) var userDefaults: UserDefaults?
+        private(set) var storeKit2Setting: StoreKit2Setting = .init(useStoreKit2IfAvailable: false)
+        private(set) var dangerousSettings: DangerousSettings?
+        private(set) var networkTimeoutSeconds: Int = Configuration.networkTimeoutSecondsDefault
+        private(set) var storeKit1TimeoutSeconds: Int = Configuration.storeKitTimeoutSecondsDefault
+
+        override public init() {
+            fatalError("Use init(withAPIKey:).")
+        }
```
aboedo, line 88: these aren't needed in swift

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,150 @@
...
+
+        /// Set or update an `appUserID`.
+        @objc public func with(appUserID: String) -> ConfigurationBuilder {
+            self.appUserID = appUserID
+            return self
+        }
+
+        /// Set or update `observerMode`.
+        @objc public func with(observerMode: Bool) -> ConfigurationBuilder {
+            self.observerMode = observerMode
```
aboedo, line 109: The `or update` part here seems like it might need more clarification - you really shouldn't be updating this value after the SDK has been set. 
Since we're storing these properties in Purchases.swift, the values will be stored so if you try to update the configuration after configuration it'll be ignored. 
Maybe we could add that to the docstring?

Sources/Purchasing/Purchases.swift
```diff
@@ -1631,6 +1636,33 @@ public extension Purchases {
...
 
+    /**
+     * Configures an instance of the Purchases SDK with a specified ``Configuration``.
+     *
+     * The instance will be set as a singleton.
+     * You should access the singleton instance using ``Purchases/shared``
+     *
+     * - Note: Use this initializer if you wish to adjust network request timeouts.
+     *
+     * - Parameter configuration: The ``Configuration`` object you wish to use to configure `Purchases`
```
aboedo, line 1647: I think we're missing a double-backtick at the end for \`\`Purchases\`\`

Sources/Purchasing/Purchases.swift
```diff
@@ -1958,6 +1996,16 @@ internal extension Purchases {
...
+    // Used for testing
+    var networkTimeoutSeconds: Int {
+        return self.backend.networkTimeoutSeconds
+    }
+
+    // Used for testing
+    var storeKitTimeoutSeconds: DispatchTimeInterval {
+        return self.productsManager.requestTimeout
+    }
+
```
aboedo, line 2008: Maybe we should have these in the tests instead?

aboedo: or as a separate file?

Tests/UnitTests/SubscriberAttributes/BackendSubscriberAttributesTests.swift
```diff
@@ -329,7 +329,10 @@ class BackendSubscriberAttributesTests: XCTestCase {
         let eTagManager = MockETagManager(userDefaults: MockUserDefaults())
         self.mockETagManager = eTagManager
 
-        return MockHTTPClient(systemInfo: self.systemInfo, eTagManager: eTagManager, sourceTestFile: file)
+        return MockHTTPClient(systemInfo: self.systemInfo,
+                              eTagManager: eTagManager,
+                              timeoutSeconds: 7,
```
aboedo, line 334: why do some use `7` and some `3`? Are they fully arbitrary?

aboedo (commented):
I absolutely love this 👏

Tests/APITesters/SwiftAPITester/SwiftAPITester/ConfigurationAPI.swift
```diff
@@ -0,0 +1,24 @@
...
+        .configurationBuilder(withAPIKey: "")
+        .with(apiKey: "")
+        .with(appUserID: "")
+        .with(observerMode: false)
+        .with(userDefaults: UserDefaults.standard)
+        .with(dangerousSettings: DangerousSettings())
+        .with(networkTimeoutSeconds: 1)
+        .with(storeKit1TimeoutSeconds: 1)
+        .with(usesStoreKit2IfAvailable: false)
+        .build()
```
aboedo, line 22: The usage looks awesome :chefskiss:

It does somewhat make me wonder whether we should name the `configurationBuilder` method something that looks more RevenueCat-specific. Otherwise, in Swift, that line will look generic enough that it could totally have belonged to any SDK, and it might be confusing to devs maintaining apps that use RC

Sources/Purchasing/Configuration.swift
NachoSoto: And I'd rename it `Configuration.Builder`

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
taquitos: Human compiler optimization
`self.requestTimeout.seconds / 10`  -> `self.requestTimeout.seconds = 30` -> `3`

Sources/Purchasing/Configuration.swift
taquitos: Old habits die hard.

Sources/Networking/HTTPClient.swift
taquitos: I was thinking about it, I wanted to get folks' opinions since I don't have a strong one. I lean towards Int only because it's more comfortable as an API for me.

Sources/Purchasing/Configuration.swift
taquitos: I miss ObjC, OK? 😆

Sources/Purchasing/Configuration.swift
taquitos: Ohhh, right, I didn't realize that. I'll update.

Tests/UnitTests/SubscriberAttributes/BackendSubscriberAttributesTests.swift
taquitos: I like to keep it arbitrary- I can test that I modified it from the default.

Tests/APITesters/SwiftAPITester/SwiftAPITester/ConfigurationAPI.swift
taquitos: True, I'll think a bit about this.

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,150 @@
...
+    static let networkTimeoutSecondsDefault: Int = 60
+
+    let apiKey: String
+    let appUserID: String?
+    let observerMode: Bool
+    let userDefaults: UserDefaults?
+    let storeKit2Setting: StoreKit2Setting
+    let dangerousSettings: DangerousSettings?
+    let networkTimeoutSeconds: Int
+    let storeKit1TimeoutSeconds: Int
```
NachoSoto, line 42: By using `TimeInterval` we could name these just `networkTimeout` and `storeKitTimeout`?

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,150 @@
...
+
+        if configurationBuilder.networkTimeoutSeconds < 5 {
+            self.networkTimeoutSeconds = 5
+        } else {
+            self.networkTimeoutSeconds = configurationBuilder.networkTimeoutSeconds
+        }
+    }
+
+    /// Factory method for the ``ConfigurationBuilder`` object that is required to create a `Configuration`
+    @objc public static func configurationBuilder(withAPIKey apiKey: String) -> ConfigurationBuilder {
```
NachoSoto, line 70:
```suggestion
    @objc public static func builder(withAPIKey apiKey: String) -> Builder {
```

So this becomes `let builder: Configuration.Builder = Configuration.builder(withAPIKey:)` instead of `let builder: Configuration.ConfigurationBuilder = Configuration.configurationBuilder(withAPIKey:)`

Sources/Purchasing/Purchases.swift
```diff
@@ -1631,6 +1636,33 @@ public extension Purchases {
...
+     *
+     * - Parameter configuration: The ``Configuration`` object you wish to use to configure `Purchases`
+     *
+     * - Returns: An instantiated ``Purchases`` object that has been set as a singleton.
+     *
+     * - Important: See ``Configuration/ConfigurationBuilder`` for more information about configurable properties.
+     * 
+     */
+    @objc(configureWithConfiguration:)
+    @discardableResult static func configure(withConfiguration configuration: Configuration) -> Purchases {
```
NachoSoto, line 1655: Ditto `configure(with:)`

Sources/Purchasing/Purchases.swift
NachoSoto: Then you could move them to `Deprecations.swift` to clean up this file of all those overloads :D

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
NachoSoto: But that assumes that the timeout is 30? 🤨

Tests/APITesters/ObjCAPITester/ObjCAPITester/RCPurchasesAPI.m
```diff
@@ -31,6 +31,7 @@ @implementation RCPurchasesAPI
 
 + (void)checkAPI {
     RCPurchases *p = [RCPurchases configureWithAPIKey:@""];
+    [RCPurchases configureWithConfiguration:[[RCConfiguration alloc] init]];
```
NachoSoto, line 34: I wonder if the Obj-C method should also be `[RCPurchases configureWith:]`?

Tests/UnitTests/Purchasing/PurchasesTests.swift
```diff
@@ -375,6 +377,17 @@ class PurchasesTests: XCTestCase {
         expect(Purchases.isConfigured) == true
     }
 
+    func testConfigurationPassedThroughTimeouts() {
```
NachoSoto, line 380: 👍🏻

Tests/UnitTests/SubscriberAttributes/BackendSubscriberAttributesTests.swift
NachoSoto: Do you mind checking that the total test time didn't go up? I myself introduced some very slow tests a while back that I then fixed by making sure we didn't wait this long in some tests so it would be nice to make sure that didn't regress 🙏🏻

NachoSoto (commented):
Just minor comments, but this looks great 😻

Sources/Networking/HTTPClient.swift
taquitos: ✅

Sources/Purchasing/Configuration.swift
taquitos: ✅

Sources/Purchasing/Configuration.swift
taquitos: ✅

Tests/UnitTests/SubscriberAttributes/BackendSubscriberAttributesTests.swift
taquitos: This would reduce the time, the default `NSURLSession` timeout is 60 seconds, this sets the full response and request timeout to 7 seconds instead 😄

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
taquitos: Yup, check it out- https://github.com/RevenueCat/purchases-ios/pull/1556/files#diff-b080ecaed1fab9b6af95a6a40653aa15720509e51c3faa1aa71263f99a4ff1cbL30

Tests/APITesters/SwiftAPITester/SwiftAPITester/ConfigurationAPI.swift
taquitos: 🕺

Sources/Purchasing/Configuration.swift
taquitos: ✅

Sources/Purchasing/Configuration.swift
taquitos: ✅

Sources/Purchasing/Configuration.swift
taquitos: Ok, `Set or update` is meant to be `Set or update default <property value> from default`
The builder is only used to initialize the SDK but some values might already have defaults, like the `userDefaults` so if you set it, you're actually updating the config. This is super pedantic and I'll just update them all to be `set`.

Sources/Purchasing/Purchases.swift
taquitos: I kinda liked it as single, since we're already in the file, but you're probably right- for consistency double back ticks makes sense.

Sources/Purchasing/Purchases.swift
taquitos: Oh I remember why I didn't do that- I'm reaching into a private variable that isn't exposed to the test targets.

Sources/Purchasing/Configuration.swift
taquitos: ✅

Sources/Purchasing/Purchases.swift
taquitos: ✅

Tests/APITesters/ObjCAPITester/ObjCAPITester/RCPurchasesAPI.m
taquitos: I don't think so. Usually the method names are extra descriptive of the first parameter.

Sources/Purchasing/Configuration.swift
taquitos: ✅

Sources/Purchasing/Configuration.swift
taquitos: ✅

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,147 @@
...
+
+    /// Factory method for the ``Configuration/Builder`` object that is required to create a `Configuration`
+    @objc public static func builder(withAPIKey apiKey: String) -> Builder {
+        return Builder(withAPIKey: apiKey)
+    }
+
+    /// The Builder for ```Configuration```.
+    @objc(RCConfigurationBuilder) public class Builder: NSObject {
+
+        static let minimumTimeout = TimeInterval(5)
```
NachoSoto, line 62: This can be `private`.

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
NachoSoto: I'm confused. This class has a `requestTimeout` parameter` that can take any value. Before this delay was 1/10th of that value, but now it's hardcoded as 3 seconds, regardless of the value of that parameter. But what if we have a different `requestTimeout`?

Tests/UnitTests/Mocks/MockBackend.swift
```diff
@@ -34,7 +34,7 @@ class MockBackend: Backend {
         let systemInfo = try! MockSystemInfo(platformInfo: nil, finishTransactions: false, dangerousSettings: nil)
         let attributionFetcher = AttributionFetcher(attributionFactory: MockAttributionTypeFactory(),
                                                     systemInfo: systemInfo)
-        self.init(httpClient: MockHTTPClient(systemInfo: systemInfo, eTagManager: MockETagManager()),
+        self.init(httpClient: MockHTTPClient(systemInfo: systemInfo, eTagManager: MockETagManager(), timeoutSeconds: 7),
```
NachoSoto, line 37: Looks like all `MockHTTPClient` are using the same `7`. Consider that the value likely doesn't matter since it doesn't do real requests, maybe move that value to the constructor so we don't need this magic number in all the tests that use it?

NachoSoto (commented):
Just a couple more minor comments. The main things left I think are
- What do you think about making the `Configuration` `Purchases.configure` method the main one, and deprecate the rest?
- Magic 3 second delay thing.

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,147 @@
...
+            self.storeKit1Timeout = valueOrMinimum(timeout: storeKit1TimeoutSeconds)
+            return self
+        }
+
+        /// Generate a ``Configuration`` object given the values configured by this builder.
+        @objc public func build() -> Configuration {
+            return Configuration(with: self)
+        }
+
+        private func valueOrMinimum(timeout: TimeInterval) -> TimeInterval {
```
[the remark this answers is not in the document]
taquitos: That's a much better name. ✅

Sources/Purchasing/Configuration.swift
taquitos: ✅

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,147 @@
...
+ * 1. Call ``Configuration/builder(withAPIKey:)`` To obtain a ``Builder`` object.
+ * 2. Set this builder's properties using the "`with(`" functions.
+ * 3. Call ``Builder/build()`` to obtain the `Configuration` object.
+ * 4. Pass the `Configuration` object into ``Purchases/configure(with:)``.
+ *
+ */
+@objc(RCConfiguration) public class Configuration: NSObject {
+
+    static let storeKitRequestTimeoutDefault = TimeInterval(30)
+    static let networkTimeoutDefault = TimeInterval(60)
```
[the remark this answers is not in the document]
taquitos: Oh yeah, duh. This is nicer 😄
✅

Sources/Purchasing/Configuration.swift
```diff
@@ -0,0 +1,147 @@
...
+    let apiKey: String
+    let appUserID: String?
+    let observerMode: Bool
+    let userDefaults: UserDefaults?
+    let storeKit2Setting: StoreKit2Setting
+    let dangerousSettings: DangerousSettings?
+    let networkTimeoutSeconds: TimeInterval
+    let storeKit1TimeoutSeconds: TimeInterval
+
+    private init(with configurationBuilder: Builder) {
```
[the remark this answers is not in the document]
taquitos: I knew this API looked off. ✅

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
taquitos: IMO, it would be better to have exponential backoff on the retry instead of just a percentage of the original value. I chose `3` purely because the default math works out to `3`. Stepping back for a second- I don't think there's a benefit to making the retry time be based on the request timeout value. It seems just as arbitrary as `3`.  Like, why does it matter that the retry is calculated based on a timeout? I could see retry based on network throughput/latency, but I don't think timeout value is a good proxy for that. 
To be clear, I'm not trying to be argumentative, I just think we could hardcode any value < 10 seconds here and still be fine 😄
Thoughts?

Tests/UnitTests/Purchasing/ProductsFetcherSK1Tests.swift
```diff
@@ -8,7 +8,8 @@ class ProductsFetcherSK1Tests: TestCase {
     var productsRequestFactory: MockProductsRequestFactory!
     var productsFetcherSK1: ProductsFetcherSK1!
 
-    private static let defaultTimeout: DispatchTimeInterval = .seconds(5)
+    private static let defaultTimeout = TimeInterval(2)
+    private static let defaultTimeoutInterval = DispatchTimeInterval.seconds(Int(defaultTimeout))
```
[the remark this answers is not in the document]
taquitos: That's a really great point. ✅

Sources/Purchasing/Purchases.swift
taquitos: ✅

Sources/Purchasing/Configuration.swift
taquitos: ✅

Sources/Purchasing/Configuration.swift
taquitos: dancing goats

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
NachoSoto: > To be clear, I'm not trying to be argumentative,

Me neither, sorry! 😅 I was just confused cause I didn't understand how this is equivalent.

> Stepping back for a second- I don't think there's a benefit to making the retry time be based on the request timeout value

Well the main benefit is so that tests don't take 3 seconds when we try out the retry mechanism. That's why I changed this to just a few milliseconds.

> Like, why does it matter that the retry is calculated based on a timeout?

That's a good point, maybe we can have a separate parameter? One for timeout, another one for retry delays. Then the tests don't assume what this value is, and it can be very small.

However, even for production 3 seconds might be a lot. This could mean that a purchase adds 3 extra seconds?

Tests/APITesters/ObjCAPITester/ObjCAPITester/RCPurchasesAPI.m
NachoSoto: I have your opposite problem haha, I only think in Swift now 😅

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
taquitos: OOooooo ok ok, that makes sense. I'll add another param for retry delay, I like that.

Tests/UnitTests/Mocks/MockBackend.swift
taquitos: ✅

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
taquitos: Actually... Imma leave this out and return it back to `let delayInSeconds = Int((self.requestTimeout.seconds / 10).rounded())`
Adding another parameter that is timeout/retry related really calls for a new enum in the configuration object and I don't think we need that yet.

Sources/Networking/HTTPClient.swift
```diff
@@ -18,23 +18,29 @@ class HTTPClient {
...
         systemInfo: SystemInfo,
         eTagManager: ETagManager,
-        dnsChecker: DNSCheckerType.Type = DNSChecker.self
+        dnsChecker: DNSCheckerType.Type = DNSChecker.self,
+        timeoutSeconds: TimeInterval = Configuration.networkTimeoutDefault
     ) {
         let config = URLSessionConfiguration.ephemeral
         config.httpMaximumConnectionsPerHost = 1
+        config.timeoutIntervalForRequest = timeoutSeconds
+        config.timeoutIntervalForResource = timeoutSeconds
```
NachoSoto, line 38: 👍🏻 I can't believe we didn't have these.

Sources/Networking/HTTPRequest.swift
```diff
@@ -16,6 +16,7 @@ struct HTTPRequest {
 
     let method: Method
     let path: Path
+    let timeoutSeconds = Configuration.networkTimeoutDefault
```
NachoSoto, line 19: Is this needed here? I'm confused why this wouldn't use `HTTPClient.networkTimeout`

Sources/Networking/HTTPClient.swift
```diff
@@ -18,23 +18,29 @@ class HTTPClient {
...
     private let state: Atomic<State> = .init(.initial)
     private let eTagManager: ETagManager
     private let dnsChecker: DNSCheckerType.Type
 
     init(
         systemInfo: SystemInfo,
         eTagManager: ETagManager,
-        dnsChecker: DNSCheckerType.Type = DNSChecker.self
+        dnsChecker: DNSCheckerType.Type = DNSChecker.self,
+        timeoutSeconds: TimeInterval = Configuration.networkTimeoutDefault
```
NachoSoto, line 33: Nit: name this `requestTimeout` instead for consistency with the rest?

Sources/Purchasing/Purchases.swift
```diff
@@ -1632,194 +1637,49 @@ public extension Purchases {
...
-    /**
-     * Configures an instance of the Purchases SDK with a specified API key and app user ID.
-     *
-     * The instance will be set as a singleton.
-     * You should access the singleton instance using ``Purchases/shared``
-     *
-     * - Note: Best practice is to use a salted hash of your unique app user ids.
-     *
-     * - Warning: Use this initializer if you have your own user identifiers that you manage.
+     * - Note: Use this initializer if you wish to adjust network request timeouts.
```
NachoSoto, line 1645: I think this isn't relevant anymore now that it's the main configure method?

Sources/Purchasing/Purchases.swift
```diff
@@ -1632,194 +1637,49 @@ public extension Purchases {
...
+     * - Important: See ``Configuration/Builder`` for more information about configurable properties.
+     * 
      */
-    @objc(configureWithAPIKey:appUserID:observerMode:)
-    @discardableResult static func configure(withAPIKey apiKey: String,
-                                             appUserID: String?,
-                                             observerMode: Bool) -> Purchases {
-        configure(withAPIKey: apiKey, appUserID: appUserID, observerMode: observerMode, userDefaults: nil)
+    @objc(configureWithConfiguration:)
+    @discardableResult static func configure(with configuration: Configuration) -> Purchases {
```
NachoSoto, line 1655: Yess so nice to have only one now 🕺🏻 

We'll we need to update the docs to reflect this?

Sources/Purchasing/Purchases.swift
```diff
@@ -1632,194 +1637,49 @@ public extension Purchases {
...
-    }
-
-    /**
-     * Configures an instance of the Purchases SDK with a custom userDefaults.
-     *
-     * Use this constructor if you want to sync status across a shared container,
-     * such as between a host app and an extension.
-     * The instance of the `Purchases` SDK will be set as a singleton.
-     * You should access the singleton instance using ``Purchases/shared``
+     * - Note: Use this initializer if your app does not have an account system.
```
NachoSoto, line 1672: Not relevant either? Depends on how the `Configuration` is defined?

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
taquitos: ✅

Sources/Purchasing/Purchases.swift
taquitos: ✅

Tests/StoreKitUnitTests/StoreKitConfigTestCase.swift
```diff
@@ -20,7 +20,8 @@ import XCTest
 @available(iOS 14.0, tvOS 14.0, macOS 11.0, watchOS 6.2, *)
 class StoreKitConfigTestCase: TestCase {
 
-    static var requestTimeout: DispatchTimeInterval = .seconds(60)
+    static var requestTimeout: TimeInterval = 60
+    static var requestDispatchTimeout = DispatchTimeInterval.seconds(Int(requestTimeout))
```
NachoSoto, line 24: I think this should be a derived property, so that it responds to changes to `requestTimeout`, right?
```suggestion
    static var requestDispatchTimeout: DispatchTimeInterval {
      return .seconds(Int(Self.requestTimeout))
    }
```

NachoSoto (approved):

Sources/Networking/HTTPRequest.swift
taquitos: AYyyyyyy oops.

taquitos: ✅

Sources/Networking/HTTPClient.swift
taquitos: Hah, same

Sources/Networking/HTTPClient.swift
taquitos: yeah, updated! ✅

Sources/Purchasing/Purchases.swift
taquitos: CC'ing @codykerns

Sources/Purchasing/Purchases.swift
taquitos: Oh yeah, good point.
✅

Sources/Purchasing/Purchases.swift
taquitos: I kinda want to keep this one around since I think it's the most common use case we have.

Tests/StoreKitUnitTests/StoreKitConfigTestCase.swift
taquitos: Oh, good call. ✅

Sources/Purchasing/Purchases.swift
NachoSoto: I'm confused, this is the only method, so you'd use it whether you have an account system or not?

Sources/Networking/Operations/GetOfferingsOperation.swift
```diff
@@ -44,7 +44,7 @@ private extension GetOfferingsOperation {
             return
         }
 
-        let request = HTTPRequest(method: .get, path: .getOfferings(appUserID: appUserID))
+        let request = HTTPRequest(method: .get, path: .getOfferings(appUserID: appUserID), timeout: httpClient.timeout)
```
NachoSoto, line 47: This feels backwards. `HTTPClient` is already initialized with a timeout, so why pass the very same timeout to every request?

NachoSoto (changes requested):

Sources/Purchasing/Purchases.swift
taquitos: We have both:
`@discardableResult static func configure(with configuration: Configuration) -> Purchases {`

and 
`@discardableResult static func configure(withAPIKey apiKey: String) -> Purchases {`

The first is the preferred method for folks who want to do more than just set an API key.
The second is for people how have the most simple setup and don't require customization. Like if they don't have an account system you wouldn't set the `appUserID` during init. But if you did, you could use the `Configuration.Builder` to set that property.

NachoSoto: Oooh sorry, that makes sense 👍🏻

Sources/Networking/Operations/GetOfferingsOperation.swift
taquitos: You're right, I was being heavy handed with a replacement

taquitos: ✅

Sources/Networking/HTTPRequest.swift
```diff
@@ -16,7 +16,6 @@ struct HTTPRequest {
 
     let method: Method
     let path: Path
-    let timeout: TimeInterval
```
taquitos: I had this in here and it shouldn't have been @NachoSoto

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -414,7 +415,7 @@ class HTTPClientTests: TestCase {
 
     #if os(macOS) || targetEnvironment(macCatalyst)
     func testAlwaysPassesAppleDeviceIdentifierWhenIsSandbox() {
-        let request = HTTPRequest(method: .get, path: .mockPath)
+        let request = HTTPRequest(method: .get, path: .mockPath, timeoutSeconds: self.client.timeout)
```
NachoSoto, line 418: I think this is left over too?

NachoSoto (approved):
Ship it!

Tests/UnitTests/Networking/HTTPClientTests.swift
taquitos: Hah, yup.

taquitos: ✅

Sources/Purchasing/Purchases.swift
taquitos: 🎉

danibachar:
Whoop whoop. Thank you guys for taking care of it

Issue #1382, opened 2022-03-11
Respond to Storefront updates


aboedo:
When the Storefront is changed while the app is live, we have some stuff in cache that needs to be updated. 

This could lead to some issues, [outlined here](https://github.com/RevenueCat/purchases-ios/pull/1374#discussion_r824724896). 

We should: 

- Listen to Storefront updates ([SK2 method here](https://developer.apple.com/documentation/storekit/storefront/storefronts/asynciterator#), [SK1 method here](https://developer.apple.com/documentation/storekit/skpaymenttransactionobserver/3242936-paymentqueuedidchangestorefront#))
- When the storefront is updated, clear caches for offerings + ProductsFetcherSK1 and ProductsFetcherSK2
- Bonus points: eagerly re-fetch products that had beeb previously fetched in productsFetchers + re-fetch offerings
- Add a new method in purchasesDelegate that ports that notification over to the developer. 

The case where the storefront is updated while the app is open is fairly rare, so this won't affect many apps in real life. But long-running apps could be affected, and it isn't _impossible_ to run into this. 

At the very least, if we're eagerly refreshing offerings and products when the storefront is updated, apps will get fresh products when calling `getOfferings` again, which might already be enough for most developers.

Juanpe:
I could do it if it's still unassigned 🙂

aboedo:
That makes total sense to me, and it sounds great!!

Juanpe:
Hi guys 👋🏼! To finish this issue, I'd like to do the last task, create a new `PurchasesDelegate` method, but I have a doubt. When should be the best time to call this method?

1. When the Storefront changes WITHOUT waiting to clear and re-fetch the cache stuff.
2. When the Storefront changes and we clear the cache WITHOUT waiting to re-fetch the cache stuff.
3. When the Storefront changes, we clear the cache and re-fetch the cache stuff again.

To me, option 3 would make more sense; maybe the developer needs to refresh the UI. I would even say that we could add two methods when the Storefront changes and the other when re-fetching the cache stuff, although I don't know if we should add so many methods to the delegate protocol.

WDYT? 🤔

NachoSoto:
I'm not sure if we need to add anything. Users can already use `StoreKit.Storefront.updates` to be notified of updates, and manually refetch products if they need to (which thanks to your updates would be up-to-date). Does that make sense?

PR #1557, opened 2022-05-09
Prepare the codebase to listen to the Storefront changes (1/4)

Juanpe:
Subtask of #1382 

### Description

This PR is the first step to preparing the codebase to respond to **Storefront** changes to update some stuff that is in the cache. 

Changes:
- Listen to Storefront changes for SK1 via `SKPaymentTransactionObserver` protocol using the `paymentQueueDidChangeStorefront` method.
- Listen to Storefront changes for SK2, creating a new listener (`StoreKit2TransactionListener `). 
- Update `PurchasesOrchestrator`, adding the new delegate methods.
- Fix failing tests related to these changes and add new ones.


Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -67,7 +67,10 @@ class PurchasesOrchestrator {
     private let lock = NSRecursiveLock()
 
     @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
-    lazy var storeKit2Listener = StoreKit2TransactionListener(delegate: self)
+    lazy var storeKit2TransactionListener = StoreKit2TransactionListener(delegate: self)
```
Juanpe, line 70: In the future, we could create a single listener that would be composed of other "sublisteners"

Tests/StoreKitUnitTests/StoreKit2StorefrontListenerTests.swift
```diff
@@ -0,0 +1,46 @@
...
+//  StoreKit2StorefrontListenerTests.swift
+//
+//  Created by Juanpe Catalán on 9/5/22.
+
+import Nimble
+@testable import RevenueCat
+import XCTest
+
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+class StoreKit2StorefrontListenerTests: XCTestCase {
```
Juanpe, line 19: I tried to add a test case to cover when the Storefront changes, but I couldn’t. Is there is a way? 🤔

Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
```diff
@@ -0,0 +1,46 @@
...
+import StoreKit
+
+protocol StoreKit2StorefrontListenerDelegate: AnyObject {
+    func storefrontDidUpdate()
+}
+
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+class StoreKit2StorefrontListener {
+
+    private(set) var taskHandle: Task<Void, Never>?
```
NachoSoto, line 24: I would implement `didSet` here like so:
```swift
didSet {
  if self.taskHandle !== oldValue {
    oldValue?.cancel()
  }
}
```

Then you only need to set it to `nil` or the new value.

Juanpe: much better 👍🏼 Thanks :)

Juanpe: Just a question, why do you have to check if they are identical, checking if they are not `equal` wouldn't be enough? 🤔

Juanpe: Done! ✅  [cee546e](https://github.com/RevenueCat/purchases-ios/pull/1557/commits/cee546efcbb20d7163454f5a65a050db76ece160)

Sources/Purchasing/PurchasesOrchestrator.swift
NachoSoto: Can this be `private(set)`?

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -67,7 +67,10 @@ class PurchasesOrchestrator {
...
+
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+    lazy var storeKit2StorefrontListener = StoreKit2StorefrontListener(delegate: self)
```
NachoSoto, line 73: Ditto `private(set)`

Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
```diff
@@ -0,0 +1,52 @@
...
+            for await _ in StoreKit.Storefront.updates {
+                guard let self = self else { break }
+                self.delegate?.storefrontDidUpdate()
+            }
+        }
+    }
+
+    deinit {
+        taskHandle?.cancel()
+        taskHandle = nil
```
NachoSoto, line 49: This is duplicated, you can just do `self.taskHandle = nil` now :)
The advantage of what I suggested is that now it's impossible to have a reference to a cancelled task :)

Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    weak var delegate: StoreKit2StorefrontListenerDelegate?
+
+    init(delegate: StoreKit2StorefrontListenerDelegate?) {
+        self.delegate = delegate
+    }
+
+    func listenForStorefrontChanges() {
+        taskHandle = Task { [weak self] in
```
NachoSoto, line 39: Nit: our style guide requires explicit `self`:
```suggestion
        self.taskHandle = Task { [weak self] in
```

Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
```diff
@@ -0,0 +1,52 @@
...
+            for await _ in StoreKit.Storefront.updates {
+                guard let self = self else { break }
+                self.delegate?.storefrontDidUpdate()
```
NachoSoto, line 42: Maybe put this inside of `MainActor.run {}`?

Tests/StoreKitUnitTests/StoreKit2StorefrontListenerTests.swift
NachoSoto: Yeah I don't think we can change the storefront mid-way through a test run unfortunately.

Tests/StoreKitUnitTests/StoreKit2StorefrontListenerTests.swift
```diff
@@ -0,0 +1,46 @@
...
+
+    private var listener: StoreKit2StorefrontListener! = nil
+
+    override func setUp() {
+        super.setUp()
+
+        self.listener = .init(delegate: nil)
+    }
+
+    func testStopsListeningToChanges() throws {
```
NachoSoto, line 29: I think you also need an `@available` annotation in the test too?

Tests/StoreKitUnitTests/StoreKit2StorefrontListenerTests.swift
```diff
@@ -0,0 +1,46 @@
...
+        try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
+
+        var handle: Task<Void, Never>?
+
+        expect(self.listener!.taskHandle).to(beNil())
```
NachoSoto, line 34: You don't need the `!`. It's better if tests assume it's non-nil :)
```suggestion
        expect(self.listener.taskHandle).to(beNil())
```

Tests/StoreKitUnitTests/StoreKit2StorefrontListenerTests.swift
```diff
@@ -0,0 +1,46 @@
...
+
+        self.listener!.listenForStorefrontChanges()
+        handle = self.listener!.taskHandle
+
+        expect(handle).toNot(beNil())
+        expect(handle?.isCancelled) == false
+
+        self.listener = nil
+        expect(handle?.isCancelled) == true
```
NachoSoto, line 43: This is nice.

NachoSoto (approved):
Awesome!

Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
```diff
@@ -0,0 +1,52 @@
...
+//  StoreKit2StorefrontListener.swift
+//
+//  Created by Juanpe Catalán on 5/5/22.
+
+import Foundation
+import StoreKit
+
+protocol StoreKit2StorefrontListenerDelegate: AnyObject {
+    func storefrontDidUpdate()
+}
```
NachoSoto, line 19: Nit spaces:
```suggestion
protocol StoreKit2StorefrontListenerDelegate: AnyObject {

    func storefrontDidUpdate()
    
}
```

Tests/StoreKitUnitTests/StoreKit2StorefrontListenerTests.swift
Juanpe: 🤔 I don't think so. [I've applied the `@available` attribute to the whole class](https://github.com/RevenueCat/purchases-ios/blob/cee546efcbb20d7163454f5a65a050db76ece160/Tests/StoreKitUnitTests/StoreKit2StorefrontListenerTests.swift#L18) so I think it’s not needed, is it?

Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
Juanpe: Yeah! good catch! I forgot to update this line 🤦🏼‍♂️

Sources/Purchasing/PurchasesOrchestrator.swift
Juanpe: same here 🙂

Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
Juanpe: Done! ✅  ([171db44](https://github.com/RevenueCat/purchases-ios/pull/1557/commits/171db446d23bf8e2adfbf138cb65419f16333c07))

Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
Juanpe: Good point 👍🏼

Juanpe: Done! ✅ ([ad5e799](https://github.com/RevenueCat/purchases-ios/pull/1557/commits/ad5e799ac4cf64a1e43076ab3b98b570f58c2e16))

Sources/Purchasing/PurchasesOrchestrator.swift
NachoSoto: That's fine then!

NachoSoto: That's fine then!

Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
Juanpe: I've reverted the change because when the listener is released, the `didSet` observer is not called 😅 so the task is not cancelled. [I've cancelled the task when the listener is released](https://github.com/RevenueCat/purchases-ios/blob/ad5e799ac4cf64a1e43076ab3b98b570f58c2e16/Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift#L52) and [modified the test cases](https://github.com/RevenueCat/purchases-ios/blob/ad5e799ac4cf64a1e43076ab3b98b570f58c2e16/Tests/StoreKitUnitTests/StoreKit2StorefrontListenerTests.swift#L29) to cover this scenario.

NachoSoto: Oooh that's a great catch I forgot!!

NachoSoto:
I'll wait to see if @aboedo has any feedback before merging this.

Juanpe:
Hi guys! 👋🏼  Did you have a chance to take a look at this PR? 🙂 The second PR is ready to be opened but I can't because I'm working on a fork and I can't select this branch as the base of the other PR. So, I have to wait for this PR to be merged to open the other one, otherwise, you would see the same changes 😅

taquitos (approved):
🎈🐐

PR #1559, opened 2022-05-10
Created `TestCase` superclass to allow re-generating snapshots

NachoSoto:
## Motivation:

There is no way to run code statically like in Obj-C using `+ [NSObject initialize]`, so this ensures (through a new `SwiftLint` rule) that all test classes inherit this common test class instead of `XCTestCase` directly.
A future PR will make use of this new `CIRCLECI_TESTS_GENERATE_SNAPSHOTS` environment variable to allow regenerating snapshots in CI.

I've also moved the duplicated `CurrentTestCaseTracker` usages to a single point.

### DIFF
--- a/Tests/UnitTests/Networking/Backend/BaseBackendTest.swift
+++ b/Tests/UnitTests/Networking/Backend/BaseBackendTest.swift
@@ -18,7 +18,7 @@ import XCTest
 
 @testable import RevenueCat
 
-class BaseBackendTests: XCTestCase {
+class BaseBackendTests: TestCase {
 
     private(set) var systemInfo: SystemInfo!
     private(set) var httpClient: MockHTTPClient!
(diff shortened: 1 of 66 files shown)


Tests/UnitTests/Networking/Backend/BaseBackendTest.swift
```diff
@@ -40,14 +40,6 @@ class BaseBackendTests: XCTestCase {
...
     }
 
-    override class func setUp() {
-        XCTestObservationCenter.shared.addTestObserver(CurrentTestCaseTracker.shared)
-    }
-
-    override class func tearDown() {
-        XCTestObservationCenter.shared.removeTestObserver(CurrentTestCaseTracker.shared)
-    }
-
```
NachoSoto, line 50: This is in `TestCase` now.

Tests/UnitTests/SubscriberAttributes/BackendSubscriberAttributesTests.swift
```diff
@@ -71,14 +71,6 @@ class BackendSubscriberAttributesTests: XCTestCase {
...
     }
 
-    override class func setUp() {
-        XCTestObservationCenter.shared.addTestObserver(CurrentTestCaseTracker.shared)
-    }
-
-    override class func tearDown() {
-        XCTestObservationCenter.shared.removeTestObserver(CurrentTestCaseTracker.shared)
-    }
-
```
NachoSoto, line 81: Ditto

joshdholtz (approved):
💯 LGTM

PR #1504, opened 2022-04-13
Backend snapshot tests: separate snapshot for each major version

NachoSoto:
Fixes [CF-547].

I recommend reviewing this ignoring JSON files:https://github.com/RevenueCat/purchases-ios/pull/1504/files?file-filters%5B%5D=.pbxproj&file-filters%5B%5D=.swift&show-deleted-files=true&show-viewed-files=true

This also combines the duplicate copies of some tests for each version, as only 1 is needed now. `assertSnapshot` encodes the version in the filename.
Combined:
- `BackendPostRecepitDataTests`
- `BackendSubscriberAttributesTests`

No more unnecessary skipped tests.
- Before: `Number of tests skipped | 112`
- Now: `Number of tests skipped | 58`

### Example:
Screen Shot 2022-04-13 at 16 07 15

[CF-547]: https://revenuecats.atlassian.net/browse/CF-547?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### ⚠️ Note:

Because iOS 12.x simulators currently crash on M1 Macs, I haven't generated those snapshots yet.

PR #1366, opened 2022-03-08
`MockHTTPClient`: encode the entire `HTTPRequest` into snapshots

NachoSoto:
This builds up on top of all the prior PRs that allow for easily encoding requests into simple `struct`s.
Prior to this, tests had to manually check the different parameters of requests (path, headers, body, method, etc.). Now it's all automatic and encoded together in the snapshot files.

Example snapshot:
```json
{
  "headers" : {
    "Authorization" : "Bearer asharedsecret"
  },
  "request" : {
    "body" : {
      "fetch_token" : "AQID",
      "product_identifiers" : [
        "producta",
        "productb",
        "productc",
        "productd"
      ]
    },
    "method" : "POST",
    "url" : "https://api.revenuecat.com/v1/subscribers/user/intro_eligibility"
  }
}
```

PR #1560, opened 2022-05-10
`Snapshotting.formattedData`: use `JSONEncoder.OutputFormatting.withoutEscapingSlashes` if available

NachoSoto:
This was removed after a [PR review comment](https://github.com/RevenueCat/purchases-ios/pull/1366#discussion_r823106740) because it would have lead to different snapshots in iOS 12.

However, since #1504 we have separate snapshots for each iOS version, so this cleans up the output when possible.

### Example:
```diff
- "url" : "https:\/\/api.revenuecat.com\/v1\/subscribers\/user\/alias"
+ "url" : "https://api.revenuecat.com/v1/subscribers/user/alias"
```


NachoSoto:
I can manually update the snapshots, but it's a perfect test case for #1561 yeah.

aboedo (approved):

PR #1564, opened 2022-05-11
Generating new test snapshots

revenuecat-ops:


NachoSoto:
@joshdholtz could you change the Xcode output to raw so we can look at the logs? I can't run iOS 12 tests locally so I can't debug why for some reason some of the tests are doing way too much requests (that's why there's up to 12 json files for some of the tests).

PR #1565, opened 2022-05-11
Added tests and fix to ensure `RawDataContainer` includes all data

NachoSoto:
Follow up to #1496. Depends on ~~#1567,  #1568, #1604~~.

The solution implemented in #1496 was very simple, but [as pointed out by @joshdholtz](https://github.com/RevenueCat/purchases-ios/pull/1496#discussion_r869463028), it was only encoding the data that was formally decoded by the `struct`s.
We had no test coverage for this, so it went unnoticed.

The new solution adds a new field:
```swift
@IgnoreEncodable @IgnoreHashable
var rawData: [String: Any]
```

Which, together with a new helper method `decodeRawData()`, allows storing all the data to be used by `CustomerInfo`'s `RawDataContainer` implementation.

Unfortunately, despite many attempts, I realized that this can't be abstracted into a property wrapper, because a `KeyedDecodingContainer` can't access the parent `Decoder`.
That's why the `CustomerInfoResponse` and `CustomerInfoResponse.Entitlement` must manually call this method.

### DIFF
--- a/Sources/Misc/RawDataContainer.swift
+++ b/Sources/Misc/RawDataContainer.swift
@@ -15,27 +15,32 @@
 /// to future data while using an older version of the SDK.
 public protocol RawDataContainer {
 
-    /// The type of the ``RawDataContainer/underlyingData`` for this type.
-    associatedtype Content: Encodable
+    /// The type of the ``RawDataContainer/rawData`` for this type.
+    associatedtype Content
 
     /// The underlying content for debugging purposes or for getting access
     /// to future data while using an older version of the SDK.
-    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
-    var underlyingData: Content { get }
+    var rawData: Content { get }
 
 }
 
-/// Default implementation that encodes the `underlyingData` for public use.
-extension RawDataContainer {
+extension Decoder {
 
-    /// The underlying data for this type, for debugging purposes or for getting access
-    /// to future data while using an older version of the SDK.
-    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
-    public var rawData: [String: Any] {
+    /// Decodes the entire content of this `Decoder` into `[String: Any]` to be used for a `RawDataContainer` type.
+    func decodeRawData() -> [String: Any] {
         do {
-            return try self.underlyingData.asDictionary()
+            let value = try self.singleValueContainer()
+                .decode(AnyDecodable.self)
+                .asAny
+
+            guard let dictionary = value as? [String: Any] else {
+                Logger.warn(Strings.codable.unexpectedValueError(type: type(of: value), value: value))
+                return [:]
+            }
+
+            return dictionary
         } catch {
-            Logger.warn(Strings.codable.encoding_error(error))
+            Logger.warn(Strings.codable.decoding_error(error))
             return [:]
         }
     }
--- a/Sources/Networking/Responses/CustomerInfoResponse.swift
+++ b/Sources/Networking/Responses/CustomerInfoResponse.swift
@@ -15,8 +15,12 @@ import Foundation
 
 struct CustomerInfoResponse {
 
-    var requestDate: Date
     var subscriber: Subscriber
+
+    @IgnoreHashable
+    var requestDate: Date
+    @IgnoreEncodable @IgnoreHashable
+    var rawData: [String: Any]
 
 }
 
@@ -74,6 +78,9 @@ extension CustomerInfoResponse {
         var productIdentifier: String
         var purchaseDate: Date?
 
+        @IgnoreEncodable @IgnoreHashable
+        var rawData: [String: Any]
+
     }
 
 }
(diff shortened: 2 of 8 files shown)


Sources/FoundationExtensions/Decoder+Extensions.swift
```diff
@@ -77,9 +77,48 @@ extension DefaultValue: Encodable where Value: Encodable {
...
+
+extension IgnoreEncodable: Decodable where Value: Decodable {
+
+    init(from decoder: Decoder) throws {
+        self.wrappedValue = try decoder.singleValueContainer().decode(Value.self)
+    }
+
+}
+
+extension IgnoreEncodable: Encodable {
```
NachoSoto, line 104: A transparent benefit of this is that `@IgnoreEncodable` conforms to `Encodable` without relying on `Value` (because it doesn't actually do anything). Which means that it can be used for types that aren't `Encodable` (like `[String: Any]`):
```swift
struct Data: Encodable {
  @IgnoreEncodable var rawData: [String: Any] // this compiles!
}
```

Sources/Misc/Obsoletions.swift
```diff
@@ -719,3 +719,16 @@ public extension Purchases {
...
+
+extension EntitlementInfo {
+
+    @available(iOS, obsoleted: 1, message: "Use `CustomerInfo.rawData`")
+    @available(tvOS, obsoleted: 1, message: "Use `CustomerInfo.rawData`")
+    @available(watchOS, obsoleted: 1, message: "Use `CustomerInfo.rawData`")
+    @available(macOS, obsoleted: 1, message: "Use `CustomerInfo.rawData`")
+    @available(macCatalyst, obsoleted: 1, message: "Use `CustomerInfo.rawData`")
+    public var rawData: [String: Any] {
+        fatalError()
```
aboedo, line 731: this honestly doesn't feel great - we _just_ added the feature and we're already obsoleting it

aboedo: also, this is an obsoletion with a fatal error, this would mean a major version

aboedo: naive question: why can't this be loaded into the EntitlementInfo as `rawData["subscriber"]["entitlements"]`?

NachoSoto: Yeah looking back this definitely doesn't feel like a good tradeoff, hopefully we can find an easy way to do this.
I think I just thought of a way, standby! I'm holding on merging #1496 because this is kind of a blocker.

NachoSoto: Okay no more deprecation :) And I added some extra data to the test fixture, and verified that it gets decoded as part of the raw data :)

Sources/Misc/RawDataContainer.swift
```diff
@@ -15,27 +15,32 @@
...
+    func decodeRawData() -> [String: Any] {
         do {
-            return try self.underlyingData.asDictionary()
+            let value = try self.singleValueContainer()
+                .decode(AnyDecodable.self)
+                .asAny
+
+            guard let dictionary = value as? [String: Any] else {
+                Logger.warn(Strings.codable.unexpectedValueError(type: type(of: value), value: value))
+                return [:]
```
NachoSoto, line 38: This logs an error and simply returns empty data. We wouldn't want to make the entire structure fail to decode because of this.

Sources/Networking/Responses/CustomerInfoResponse.swift
```diff
@@ -74,16 +78,34 @@ extension CustomerInfoResponse {
         var productIdentifier: String
         var purchaseDate: Date?
 
+        @IgnoreEncodable @IgnoreHashable
+        var rawData: [String: Any]
```
NachoSoto, line 82: I think this is pretty elegant 🤓

Sources/Networking/Responses/CustomerInfoResponse.swift
```diff
@@ -74,16 +78,34 @@ extension CustomerInfoResponse {
...
+extension CustomerInfoResponse.Entitlement: Decodable {
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        self.expiresDate = try container.decodeIfPresent(Date.self, forKey: .expiresDate)
+        self.productIdentifier = try container.decode(String.self, forKey: .productIdentifier)
+        self.purchaseDate = try container.decodeIfPresent(Date.self, forKey: .purchaseDate)
+
+        self.rawData = decoder.decodeRawData()
```
NachoSoto, line 104: This is the only downside, but it's not too bad. We must manually implement this method. But reusing `decodeRawData()` it's pretty clean.

Sources/Networking/Responses/CustomerInfoResponse.swift
```diff
@@ -15,9 +15,13 @@ import Foundation
 
 struct CustomerInfoResponse {
 
-    var requestDate: Date
     var subscriber: Subscriber
 
+    @IgnoreHashable
+    var requestDate: Date
+    @IgnoreEncodable @IgnoreHashable
+    var rawData: [String: Any]
```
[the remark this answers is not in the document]
aboedo: nice!! Perhaps we could add that as a comment here?

Sources/Networking/Responses/CustomerInfoResponse.swift
aboedo: yeah, the syntax for this is :chefskiss:

Sources/Networking/Responses/CustomerInfoResponse.swift
```diff
@@ -74,16 +78,34 @@ extension CustomerInfoResponse {
...
 
 extension CustomerInfoResponse.Subscriber: Codable, Hashable {}
-extension CustomerInfoResponse.Entitlement: Codable, Hashable {}
 extension CustomerInfoResponse.Subscription: Codable, Hashable {}
 
+extension CustomerInfoResponse.Entitlement: Hashable {}
+extension CustomerInfoResponse.Entitlement: Encodable {}
+extension CustomerInfoResponse.Entitlement: Decodable {
+
+    init(from decoder: Decoder) throws {
```
aboedo, line 97: maybe we could add a comment here explaining why we need to implement this method manually?

aboedo (approved):

Sources/Networking/Responses/CustomerInfoResponse.swift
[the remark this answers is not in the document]
NachoSoto: Oh yeah, I removed the comment from the previous implementation but that would be good here.

Sources/Networking/Responses/CustomerInfoResponse.swift
NachoSoto: Done.

PR #1566, opened 2022-05-11
Added `String.asData` to avoid dealing with `Data?`

NachoSoto:
Turns out `Data(string.utf8)` is a much simpler way of converting `String` to `Data` that doesn't lead to an `Optional`.
UTF-8 can represent all valid Unicode code points, so this was always safe, but now we know at compile time.

The reason the old variant returned an `Optional` is because it accepts other encodings like `.ascii` which might not be able to represent everything in the `String`.

This simplifies several parts of the code that were dealing with an error that was impossible.

### DIFF
--- a/Sources/Purchasing/StoreKitAbstractions/PromotionalOffer.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/PromotionalOffer.swift
@@ -81,41 +81,15 @@ extension PromotionalOffer.SignedData {
                                  timestamp: self.timestamp as NSNumber)
     }
 
-    /// - Throws: ``ErrorCode/unexpectedBackendResponseError`` if the signature cannot be encoded in UTF8
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
     var sk2PurchaseOption: Product.PurchaseOption {
-        get throws {
-            guard let signatureData = self.signature.data(using: .utf8) else {
-                throw ErrorUtils.unexpectedBackendResponse(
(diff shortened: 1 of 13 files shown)


Sources/Purchasing/StoreKitAbstractions/PromotionalOffer.swift
```diff
@@ -81,41 +81,15 @@ extension PromotionalOffer.SignedData {
...
 
-    /// - Throws: ``ErrorCode/unexpectedBackendResponseError`` if the signature cannot be encoded in UTF8
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
     var sk2PurchaseOption: Product.PurchaseOption {
-        get throws {
-            guard let signatureData = self.signature.data(using: .utf8) else {
-                throw ErrorUtils.unexpectedBackendResponse(
-                    withSubError: PromotionalOffer.Error.invalidSignature(self.signature)
-                )
-            }
```
NachoSoto, line 92: cc @aboedo since we talked about this error. No longer needed :)

NachoSoto, line 92: cc @aboedo since we talked about this error. No longer needed :)

aboedo: awesome!!

aboedo: there's nothing better than killing error handling because the error can't possibly happen anymore 👏

aboedo (approved):

PR #1567, opened 2022-05-11
Added `@IgnoreEncodable` property wrapper

NachoSoto:
This allows having properties in an `Encodable` type that won't be encoded.
This will be used to storing the `rawData` when decoding our types (see #1496).

### DIFF
--- a/Sources/FoundationExtensions/Decoder+Extensions.swift
+++ b/Sources/FoundationExtensions/Decoder+Extensions.swift
@@ -77,9 +77,48 @@ extension DefaultValue: Encodable where Value: Encodable {
 
 }
 
+// MARK: - IgnoreEncodable
+
+/// A property wrapper that allows not encoding a value.
+/// - Example:
+/// ```
+/// struct Data {
+///     @IgnoreEncodable var data: String // this value won't be encoded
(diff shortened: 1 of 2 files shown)


Sources/FoundationExtensions/Decoder+Extensions.swift
```diff
@@ -77,9 +77,48 @@ extension DefaultValue: Encodable where Value: Encodable {
...
+
+extension IgnoreEncodable: Decodable where Value: Decodable {
+
+    init(from decoder: Decoder) throws {
+        self.wrappedValue = try decoder.singleValueContainer().decode(Value.self)
+    }
+
+}
+
+extension IgnoreEncodable: Encodable {
```
NachoSoto, line 104: A transparent benefit of this is that `@IgnoreEncodable` conforms to `Encodable` without relying on `Value` (because it doesn't actually do anything). Which means that it can be used for types that aren't `Encodable` (like `[String: Any]`):
```swift
struct Data: Encodable {
  @IgnoreEncodable var rawData: [String: Any] // this compiles!
}
```

aboedo (approved):

PR #1568, opened 2022-05-11
Created `AnyDecodable`

NachoSoto:
This will be used to decode the entire data in JSON for the new implementation of `RawDataContainer` in #1496.

### DIFF
--- a/Sources/Misc/AnyDecodable.swift
+++ b/Sources/Misc/AnyDecodable.swift
@@ -1,0 +1,143 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  AnyDecodable.swift
+//
+//  Created by Nacho Soto on 5/11/22.
+
+import Foundation
+
+/// Type erased `Any` that conforms to `Decodable`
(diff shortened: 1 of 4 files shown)


Sources/Misc/AnyDecodable.swift
```diff
@@ -0,0 +1,143 @@
...
+                )
+            )
+        }
+    }
+
+}
+
+extension AnyDecodable {
+
+    var asAny: Any {
```
NachoSoto, line 63: This allows decoding all the data in JSON into `[String: Any]`, which normally wouldn't be `Decodable`.

Sources/Misc/AnyDecodable.swift
```diff
@@ -0,0 +1,143 @@
...
+extension AnyDecodable: Decodable {
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+
+        if container.decodeNil() {
+            self = .null
+        } else if let value = try? container.decode(String.self) {
+            self = .string(value)
+        } else if let value = try? container.decode(Int.self) {
```
taquitos, line 38: I shouldn't be surprised, yet I am- if you pass in `2.5` it won't try to decode as an `Int`, it will work as a `Double`. That's neat.

taquitos (approved):
🎈🐐

Sources/Misc/AnyDecodable.swift
NachoSoto: Types are nice

PR #1572, opened 2022-05-12
Added `OfferingsResponse` decoding tests

NachoSoto:


Sources/Purchasing/Package.swift
```diff
@@ -76,7 +76,7 @@ private extension PackageType {
         }
     }
 
-    static let typesByDescription: [String: PackageType] = PackageType
+    fileprivate static let typesByDescription: [String: PackageType] = PackageType
         .allCases
         .filter { $0.description != nil }
         .dictionaryWithKeys { $0.description! }
```
aboedo, line 82: really feels like we should move PackageType to its own file, it's grown quite a bit over time

aboedo (approved):

PR #1575, opened 2022-05-12
`BackendGetIntroEligibilityTests`: fixed test that was passing before anything ran

NachoSoto:
Operations run asynchronously, but this test was not waiting, and therefore the only assertion always passed no matter what.
Now this actually ensures that when the operation ends, there are still no requests being made.


taquitos (approved):
🤦‍♂️🐐

PR #1576, opened 2022-05-12
`GetIntroEligibilityOperation`: replaced response parsing with `Decodable`

NachoSoto:
The new response is now represented by `GetIntroEligibilityResponse`, and has dedicated tests (`GetIntroEligibilityDecodingTests`) using a JSON fixture.
The operation becomes much simpler now, plus I've done a lot of code removing that was redundant.

The actual response handling is now 3 lines:
```swift
let result: [String: IntroEligibility] = Set(productIdentifiers)
  .dictionaryWithValues { eligibilities?[$0] ?? .unknown }
  .mapValues(IntroEligibility.init)
```

All the tests in `BackendGetIntroEligibilityTests` pass with no changes.


Sources/Networking/Responses/GetIntroEligibilityResponse.swift
```diff
@@ -0,0 +1,39 @@
...
+
+import Foundation
+
+struct GetIntroEligibilityResponse {
+
+    var eligibilityByProductIdentifier: [String: IntroEligibilityStatus]
+
+}
+
+extension GetIntroEligibilityResponse: HTTPResponseBody {
```
NachoSoto, line 22: Because we use a dedicated protocol `HTTPResponseBody` we can implement a custom deserialization in this case.

Sources/Networking/Responses/GetIntroEligibilityResponse.swift
```diff
@@ -0,0 +1,39 @@
...
+
+    static func create(with data: Data) throws -> Self {
+        let response = try [String: Bool?].create(with: data)
```
NachoSoto, line 25: But still rely on automatic deserialization from the root type `[String: Bool?]`

Sources/Networking/Operations/GetIntroEligibilityOperation.swift
```diff
@@ -86,40 +74,28 @@ private extension GetIntroEligibilityOperation {
...
+    ) {
+        let eligibilities = result.value?.body.eligibilityByProductIdentifier
 
-            if response.result.value?.statusCode.isSuccessfulResponse != true {
-                eligibilitiesByProductIdentifier = [:]
-            }
-
-            guard let eligibilitiesByProductIdentifier = eligibilitiesByProductIdentifier else {
-                return response.unknownEligibilityClosure()
-            }
```
NachoSoto, line 109: All these extra checks were unnecessary

Sources/Networking/Operations/GetIntroEligibilityOperation.swift
```diff
@@ -48,35 +48,23 @@ private extension GetIntroEligibilityOperation {
...
-            self.responseHandler(eligibilities, nil)
+            self.responseHandler(unknownEligibilities, nil)
             completion()
 
             return
         }
 
-        // Closure we can use for both missing appUserID as well as server error where we have an unknown
-        // eligibility status.
-        let unknownEligibilityClosure: () -> [String: IntroEligibility] = {
```
NachoSoto, line 70: This whole thing was the same as
```swift
Set(self.productIdentifiers)
  .dictionaryWithValues { _ in IntroEligibility(eligibilityStatus: .unknown) }
```
Which was already done above.

Sources/Networking/Operations/GetIntroEligibilityOperation.swift
```diff
@@ -86,40 +74,28 @@ private extension GetIntroEligibilityOperation {
...
 
-        httpClient.perform(request, authHeaders: self.authHeaders) { (response: HTTPResponse<[String: Any]>.Result) in
-            let eligibilityResponse = IntroEligibilityResponse(result: response.mapError { $0.asPurchasesError },
-                                                               productIdentifiers: self.productIdentifiers,
-                                                               unknownEligibilityClosure: unknownEligibilityClosure,
-                                                               completion: self.responseHandler)
-            self.handleIntroEligibility(response: eligibilityResponse)
+        httpClient.perform(
+            request, authHeaders: self.authHeaders
+        ) { (response: HTTPResponse<GetIntroEligibilityResponse>.Result) in
```
NachoSoto, line 79: Replaced `HTTPResponse<[String: Any]>` with `HTTPResponse<GetIntroEligibilityResponse>` to get automatic decoding 🎉

Sources/Networking/Operations/GetIntroEligibilityOperation.swift
```diff
@@ -86,40 +74,28 @@ private extension GetIntroEligibilityOperation {
...
 
-            if response.result.value?.statusCode.isSuccessfulResponse != true {
-                eligibilitiesByProductIdentifier = [:]
-            }
-
-            guard let eligibilitiesByProductIdentifier = eligibilitiesByProductIdentifier else {
-                return response.unknownEligibilityClosure()
-            }
+        let result: [String: IntroEligibility] = Set(productIdentifiers)
+            .dictionaryWithValues { eligibilities?[$0] ?? .unknown }
```
aboedo, line 95: total nitpick, feel free to ignore: 
at first glance I thought this was `eligibilities[0]` and thought that was weird, then I realized I had missed the `$`. 
I think for places where we're reading using `[ ]` we could name the variable for the sake of readability
```suggestion
            .dictionaryWithValues { productID in eligibilities?[productID] ?? .unknown }
```

aboedo (approved):

Sources/Networking/Operations/GetIntroEligibilityOperation.swift
NachoSoto: Yup that's a good point.

PR #1577, opened 2022-05-12
`[String: Any]` no longer needs to conform to `HTTPResponseBody`

NachoSoto:
There are no leftover usages of `HTTPClient` that need `[String: Any]` deserialization! 🎉


taquitos (approved):
👍 🐐

PR #1583, opened 2022-05-17
Clear cache after changing storefront

Juanpe:
Follow up to #1557. Subtask of #1382 

### Description

This PR aims to clear the stuff in the cache when the Storefront changes.

Changes:
- Add `clearCachedOfferings` method to `DeviceCache`.
- Add `clearCache` method to `ProductFetcherSK1`.
- Make `internal` the `clearCache` method in `ProductsFetcherSK2`.
- Handle the **Storefront** changes in the `PurchasesOrchestator`.
- Add tests to cover some scenarios.


Sources/Purchasing/ProductsManager.swift
```diff
@@ -71,6 +71,17 @@ class ProductsManager: NSObject {
         productsFetcherSK1.cacheProduct(product)
     }
 
+    func clearCachedProducts() {
+        if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *),
+           self.systemInfo.storeKit2Setting == .enabledForCompatibleDevices {
+            Task {
```
NachoSoto, line 77: See `SwiftStyleGuide`, this should be `Task<Void, Never>` to make sure we don't ignore errors.

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -767,6 +767,11 @@ private extension PurchasesOrchestrator {
                  completion: completion)
     }
 
+    func handleChangeStorefront() {
```
NachoSoto, line 770: This sounds weird. I'd rename it either `handleChangedStorefront` or `handleStorefrontChange`.

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
```diff
@@ -188,6 +188,12 @@ extension ProductsFetcherSK1: SKProductsRequestDelegate {
         }
     }
 
+    func clearCache() {
+        queue.async {
```
NachoSoto, line 192: Explicit `self` per the style guide:
```suggestion
        self.queue.async {
```

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -379,6 +379,23 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
         expect(self.backend.invokedPostOfferParameters?.offerIdentifier) == storeProductDiscount.offerIdentifier
     }
 
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func testClearCachedProductsAndOfferingsAfterStorefrontChangesWithSK2() async throws {
```
NachoSoto, line 383: Awesome

Tests/UnitTests/Purchasing/ProductsFetcherSK1Tests.swift
```diff
@@ -186,4 +186,21 @@ class ProductsFetcherSK1Tests: TestCase {
         expect(request.cancelCalled) == false
     }
 
+    func testMakesNewRequestAfterClearingCachedProductCorrectly() {
+        let productIdentifiers = Set(["1", "2", "3"])
+        let mockProducts: Set<SK1Product> = Set(productIdentifiers.map {
+            MockSK1Product(mockProductIdentifier: $0)
+        })
+
+        mockProducts.forEach { productsFetcherSK1.cacheProduct($0) }
```
NachoSoto, line 195: This can be
```suggestion
        mockProducts.forEach(self.productsFetcherSK1.cacheProduct)
```

Tests/UnitTests/Purchasing/ProductsFetcherSK1Tests.swift
```diff
@@ -186,4 +186,21 @@ class ProductsFetcherSK1Tests: TestCase {
         expect(request.cancelCalled) == false
     }
 
+    func testMakesNewRequestAfterClearingCachedProductCorrectly() {
```
NachoSoto, line 189: Would it be possible to add a similar test for SK2's fetcher?

Sources/Purchasing/ProductsManager.swift
Juanpe: 👍🏼 I've to check again the style guide to be up to date.

Sources/Purchasing/PurchasesOrchestrator.swift
Juanpe: Yep. Totally agree. I think `handleStorefrontChange` is more appropriated 👍🏼 Thanks

Tests/UnitTests/Purchasing/ProductsFetcherSK1Tests.swift
Juanpe: much better 👍🏼

Tests/UnitTests/Purchasing/ProductsFetcherSK1Tests.swift
Juanpe: Yeah, it could be great. TBH, I didn't do it because I don't know exactly to validate we fetch the products after clearing the cache. 

I think a possible way is:

```
1. Cache a product
2. Clear the cache
3. Fetch the products.
4. Validate that the obtained products are not the same as I previously cached.5. 
```

The problem is [the method to cache products is `private`](https://github.com/RevenueCat/purchases-ios/blob/588d00bdd8493003705057160b47df326bfc6ebc/Sources/Purchasing/StoreKit2/ProductsFetcherSK2.swift#L69), and it's not available for testing. We could make it internal just for testing purposes like [we did for `ProductsFetcherSK1`](https://github.com/RevenueCat/purchases-ios/blob/588d00bdd8493003705057160b47df326bfc6ebc/Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift#L185)...WDYT? 🤔

Sources/Purchasing/ProductsManager.swift
Juanpe: Done ✅ [a4f3e18](https://github.com/RevenueCat/purchases-ios/pull/1583/commits/a4f3e18e62042f9eaf98cdffac66a43e3aa2fbb7)

Sources/Purchasing/PurchasesOrchestrator.swift
Juanpe: Done ✅ [eec2fe8](https://github.com/RevenueCat/purchases-ios/pull/1583/commits/eec2fe86fac99cd5b5298f03f07a90a2916a7315)

Sources/Purchasing/ProductsManager.swift
```diff
@@ -71,6 +71,17 @@ class ProductsManager: NSObject {
         productsFetcherSK1.cacheProduct(product)
     }
 
+    func clearCachedProducts() {
+        if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *),
+           self.systemInfo.storeKit2Setting == .enabledForCompatibleDevices {
+            Task<Void, Never> {
+                await productsFetcherSK2.clearCache()
```
NachoSoto, line 78: I just realized that `ProductsFetcherSK2` was already keeping track of the `Storefront` and clearing the cache itself, so this now feels like duplicated logic. Perhaps the best solution is to remove that custom logic from within `ProductsFetcherSK2`?

Juanpe: Yep, you're right, the worst thing is we keep the outdated products in the cache, but I think it's not a real problem since we're going to re-fetch the product when the Storefront changes in the next step 👍🏼 

Let me change this code 👍🏼

Juanpe: Done ✅ ([38ca1e7](https://github.com/RevenueCat/purchases-ios/pull/1583/commits/38ca1e79af91ee07033aac18ece05158018afe28))

Sources/Purchasing/ProductsManager.swift
```diff
@@ -74,9 +74,7 @@ class ProductsManager: NSObject {
     func clearCachedProducts() {
         if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *),
            self.systemInfo.storeKit2Setting == .enabledForCompatibleDevices {
-            Task<Void, Never> {
-                await productsFetcherSK2.clearCache()
-            }
+            // We have to re-fetch the products
```
Juanpe, line 77: I wanted to use `#unavailable` here, but I decided to leave this logic because we're going to re-fetch the products in the next step

Tests/UnitTests/Purchasing/ProductsFetcherSK1Tests.swift
NachoSoto: Hmm yeah I don't see an easy way without exposing implementation details, so never mind.

Sources/Purchasing/ProductsManager.swift
NachoSoto: Oh looks like you opted for leaving only the internal cache-clearing implementation?
What do you think of doing it the other way around for consistency with SK1?

Juanpe: mm...sorry, I don't know exactly what you mean 🤔

NachoSoto: I mentioned this earlier:
> `ProductsFetcherSK2` was already keeping track of the Storefront and clearing the cache itself, so this now feels like duplicated logic. 

So you removed the logic from `ProductsManager`. I was thinking that maybe instead it made more sense to make `ProductsManager` clear the cache like you had before, and instead remove the current `Storefront` keeping logic from `ProductsFetcherSK2`, so it's consistent with how `SK1`'s fetcher works.

Juanpe: 👋🏼  I was trying to do the change but I found out that you put [a note in the code to warn about using `Storefront.updates`](https://github.com/RevenueCat/purchases-ios/blob/395322ac7ef437f6b2f880b2057559b11954bdc4/Sources/Purchasing/StoreKit2/ProductsFetcherSK2.swift#L94) could cause race conditions ([original commit](https://github.com/RevenueCat/purchases-ios/commit/e9a8dcf8076a7b2463dfdd789c8b635f5ddbc934)) 🤔 

Do you think we could have some problems using `Storefront` updates in production or it's something related to the tests?

Tests/UnitTests/Purchasing/ProductsFetcherSK1Tests.swift
aboedo: I suppose our choices at this point are: 
1. Exposing the fact that `ProductsFetcherSK2` caches objects at an `internal` level, solely for the purpose of testing
2. Not testing the caching that does happen in `ProductsFetcherSK2`
3. A larger refactor to introduce DI, but I can't think of a way of doing that that doesn't conceptually do the same as 1.

I'd opt for 1., I'd rather have a bit more test coverage than be overly purist with the internal API here

Sources/Purchasing/ProductsManager.swift
aboedo: is the implementation of this coming in a separate PR?

Juanpe: yep, the logic was to be included in step 3 but finally we decided to modify the strategy. It's related to `[this thread](https://github.com/RevenueCat/purchases-ios/pull/1583#discussion_r878370125)`

Sources/Purchasing/ProductsManager.swift
Juanpe: Ey guys, Did you guys finally get a look at this? 🤔 Thanks!

Tests/UnitTests/Purchasing/ProductsFetcherSK1Tests.swift
Juanpe: Yep, we could do it the option 1. In fact, we are doing the same for the `ProductsFetcherSK1` 🙂

> in tests, would we be able to do an extension ProductsFetcherSK2 that introduces a variable that exposes the cached products? It's still conceptually the same as 1., just doing it in the Unit tests code instead of the main implementation.

As far as I know, it's not possible. We only can use the `public` code for testing, and`internal` code is possible if we include the framework as `@testable`, but `private` it's not possible.

Sources/Purchasing/ProductsManager.swift
aboedo: could we point this PR to an integration brach? So that we have `main` always in a consistent state - it might otherwise be daunting if we have a bug report and when investigating we see that `clearCachedProducts` has been called, we might reasonably assume that cache has been cleared unless we see that the method is a no-op for sk2

aboedo (commented):
great work on this! 💪

aboedo:
created the branch `observe_storefront_changes`

aboedo:
I updated the base branch, it's based off of the current `main`

Juanpe:
Hi 👋🏼 !  I made the suggested changes:

- Clear the cached products for SK2.
- Remove the logic that invalidates the cached products when the Storefront identifier changes, in order to use the same strategy for both SK1 and SK2. ([discussion](https://github.com/RevenueCat/purchases-ios/pull/1583#discussion_r878370125)).
- Create a test case for SK2 to validate that cached is cleared correctly. I included the test in the **StoreKitUnitTests** instead of **UnitTests** because I had to use the StoreKit Configuration file. I think it's ok since the tests for ProductManaget are also included in the **StoreKitUnitTests**. WDYT?


> also, I think we might still want to call this for the .enabledOnlyForOptimizations value, since if you have that value set, sk2 might be in use for some things. I think we might wanna make the rule something like:

Finally, I didn't add the logic that clears the cache for both SK1 and SK2 when StoreKit 2 is enabled for optimizations. Why?  I've checked and although SK2 is used for some things, it's not used to fetch the products, I mean, [the products are only fetched when the StoreKit 2 is enabled for compatible devices](https://github.com/RevenueCat/purchases-ios/blob/d80eebbb3ec8403e426052dfdee2a2749964b568/Sources/Purchasing/ProductsManager.swift#L43). Please, let me know if I missed something or you think it would be safer to clear the cached products for SK1.

Sources/Purchasing/StoreKit2/ProductsFetcherSK2.swift
```diff
@@ -43,20 +42,25 @@ actor ProductsFetcherSK2 {
...
+    }
+
 }
 
 @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
-private extension ProductsFetcherSK2 {
+extension ProductsFetcherSK2 {
+    /// Read-only variable declared for testing only.
+    var cachedProducts: [SK2StoreProduct] {
+        Array(self.cachedProductsByIdentifier.values)
```
NachoSoto, line 55: Why not make `cachedProductsByIdentifier)` as `private(set)`?

Juanpe: Well, this was [my first approach](https://github.com/RevenueCat/purchases-ios/commit/f3c4c15ae3a6a13f039bda9f1577d441bb19805e).  However, I thought it might be better to have a read-only property for testing purposes, but yeah, I don't have a strong preference for that, I could make the getter function internal

WDYT?

Juanpe: Totally agree. KIS principle! 😬

Juanpe: Done ✅ ([d302f8f](https://github.com/RevenueCat/purchases-ios/pull/1583/commits/d302f8ffd66cc3d5feeed9bc694e5a90b80708bb))

aboedo:
updated the base branch so the diff doesn’t go nuts and re-requested reviews

taquitos:
I'll review this Monday 😄

Sources/Purchasing/StoreKit2/ProductsFetcherSK2.swift
```diff
@@ -67,36 +63,11 @@ private extension ProductsFetcherSK2 {
...
-
-        self.cachedProductsStorefrontIdentifier = storeFrontIdentifier
         self.cachedProductsByIdentifier += products.dictionaryWithKeys {
             $0.productIdentifier
         }
     }
 
-    /// - Returns: current `Storefront` identifier
-    func invalidateOutdatedProductsIfNeeded() async -> String? {
-        let storeFrontIdentifier = await self.currentStorefrontIdentifier
```
Juanpe, line 80: We don't have to do this validation because now, we manage the Storefront changes and clear the caches.

Sources/Purchasing/ProductsManager.swift
taquitos: @NachoSoto I think this got lost in your notifications 😄

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
```diff
@@ -188,6 +188,12 @@ extension ProductsFetcherSK1: SKProductsRequestDelegate {
         }
     }
 
+    func clearCache() {
+        self.queue.async {
```
taquitos, line 192: I feel like using dispatch barrier might be a good move here, so all previous requests can complete before we clear. 
`.async(flags: .barrier)`

taquitos (commented):
This looks awesome to me, I'd love your thoughts on the `.async(flags: .barrier)` comment.

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
Juanpe: Mmm...it's a [serial queue](https://github.com/RevenueCat/purchases-ios/blob/d302f8ffd66cc3d5feeed9bc694e5a90b80708bb/Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift#L27) and as far as I know the barrier flag only affects to concurrent queues, doesn't it? Let me know if I'm wrong.

taquitos: No no, you're right 😄, I'll wait for Nacho to do one more pass, but this looks great.

Juanpe: Ah ok ok 🙂 thanks for the review 🤙🏼

Sources/Purchasing/StoreKit2/ProductsFetcherSK2.swift
NachoSoto: 👍🏻

Tests/StoreKitUnitTests/ProductsFetcherSK2Tests.swift
```diff
@@ -0,0 +1,38 @@
...
+        super.setUp()
+        self.productsFetcherSK2 = ProductsFetcherSK2()
+    }
+
+    func testCachedProductsAreEmptyAfterClearingCachedProductCorrectly() async throws {
+        _ = try await productsFetcherSK2.product(withIdentifier: Self.productID)
+
+        await productsFetcherSK2.clearCache()
+
+        let cachedProducts = await productsFetcherSK2.cachedProductsByIdentifier
```
NachoSoto, line 34: I would check that this isn't empty before clearing the cache.

Tests/StoreKitUnitTests/ProductsFetcherSK2Tests.swift
```diff
@@ -0,0 +1,38 @@
...
+//
+//  Created by Juanpe Catalán on 1/6/22.
+
+import Nimble
+import XCTest
+
+@testable import RevenueCat
+
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+class ProductsFetcherSK2Tests: StoreKitConfigTestCase {
```
NachoSoto, line 20: Whoa we didn't have any tests before? 😕

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -822,6 +822,11 @@ private extension PurchasesOrchestrator {
                  completion: completion)
     }
 
+    func handleStorefrontChange() {
+        productsManager.clearCachedProducts()
+        deviceCache.clearCachedOfferings()
```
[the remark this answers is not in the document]
Juanpe: Much better 🙂

NachoSoto (approved):
I think this is ready! Looks a lot better with the logic taken out now of both fetchers :)

Tests/StoreKitUnitTests/ProductsFetcherSK2Tests.swift
```diff
@@ -0,0 +1,38 @@
...
+        self.productsFetcherSK2 = ProductsFetcherSK2()
+    }
+
+    func testCachedProductsAreEmptyAfterClearingCachedProductCorrectly() async throws {
+        _ = try await productsFetcherSK2.product(withIdentifier: Self.productID)
+
+        await productsFetcherSK2.clearCache()
+
+        let cachedProducts = await productsFetcherSK2.cachedProductsByIdentifier
+        expect(cachedProducts.count) == 0
```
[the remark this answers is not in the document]
Juanpe: Much better 👍🏼

Tests/StoreKitUnitTests/ProductsFetcherSK2Tests.swift
Juanpe: Good point 👍🏼

Tests/StoreKitUnitTests/ProductsFetcherSK2Tests.swift
```diff
@@ -0,0 +1,38 @@
...
+class ProductsFetcherSK2Tests: StoreKitConfigTestCase {
+
+    var productsFetcherSK2: ProductsFetcherSK2!
+
+    override func setUp() {
+        super.setUp()
+        self.productsFetcherSK2 = ProductsFetcherSK2()
+    }
+
+    func testCachedProductsAreEmptyAfterClearingCachedProductCorrectly() async throws {
```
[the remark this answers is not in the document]
Juanpe: Good catch :)

Tests/StoreKitUnitTests/ProductsFetcherSK2Tests.swift
Juanpe: Done ✅ ([9d556b3](https://github.com/RevenueCat/purchases-ios/pull/1583/commits/9d556b3a6fe130aeee3d363ec5820208c445ba0e))

Juanpe:
Hi guys, I made the suggested changes but I didn't re-request the review. Should I do it? Please, let me know, ok? 🙂 Thanks!

NachoSoto:
I'm about to run tests on here and merge this 🎉

NachoSoto:
Looks `StoreProductTests.testSk2PriceFormatterReactsToStorefrontChanges` is failing:
Screen Shot 2022-06-16 at 09 27 37

Tests/StoreKitUnitTests/StoreProductTests.swift
```diff
@@ -290,6 +290,11 @@ class StoreProductTests: StoreKitConfigTestCase {
         testSession.locale = Locale(identifier: "en_EN")
         await changeStorefront("USA")
 
+        // Note: this test passes only because the fetcher is recreated
+        // therefore clearing the cache. `ProductsFetcherSK2` does not
+        // detect Storefront changes to invalidate the cache.
```
NachoSoto, line 295: Oh makes sense.
Can you expand this comment to explain that this is done by `StoreKit2StorefrontListenerDelegate`?

NachoSoto: Also maybe it makes more sense to call `clearCache()` on the fetcher to mimic the real-world behavior?

Juanpe: Yep, it's a good point, let me expand it 👍🏼

Juanpe: yep, I thought the same, but I wanted to use the same logic for both fetchers. 🤔 BTW, I agree with you that it would mimic the real-world behaviour. Let me change it

NachoSoto (approved):
All tests passing!

Tests/StoreKitUnitTests/StoreProductTests.swift
Juanpe: Done ✅ ([645cf5d](https://github.com/RevenueCat/purchases-ios/pull/1583/commits/645cf5d101a7ecba345dd18749b5734b59a02187)). I've updated the test cases. Now, we clear manually the cache to mimic the real-world behaviour. Moreover, I've updated the comment to indicate that now, the changes of the Storefront are managed by the SK2 listener delegate. 

I think, now it looks better 👍🏼 Thanks!

NachoSoto:
Thanks again for the hard work with this!

PR #1584, opened 2022-05-18
Fix description of `StoreTransaction`

aboedo:
Just noticed this copy/pasta description in `StoreTransaction`


NachoSoto (approved):
Good catch

PR #1585, opened 2022-05-18
CF-527 - Auto generate release changelog

joshdholtz:
### Motivation
[CF-527](https://revenuecats.atlassian.net/browse/CF-527)

Generating the changelog manually (getting a list of everything merged in since last release) was a lot of work.

### Description
This will now automate that by getting fetching all of the commits / PRs and generate the changelog with the title and mentioning the GitHub username.

The logic was pulled from the recent change I made to _fastlane_ for doing this exact same thing.

The current process will still allow the developer calling this lane to edit the changelog in their specified editor (if things need to be cleaned up) but it shouuld save time by not needing to manually enter all of this.

This could also eventually be moved to a custom action/plugin that we use for other SDKs.


fastlane/Fastfile
```diff
@@ -245,6 +246,58 @@ platform :ios do
     )
   end
 
+  lane :generate_changelog do
+
+  end
```
aboedo, line 251: looks like this one wasn't implemented / used?

fastlane/Fastfile
```diff
@@ -245,6 +246,58 @@ platform :ios do
...
+
+  # Inspired by fastlane's changelog
```
aboedo, line 253: 🚀

fastlane/Fastfile
```diff
@@ -245,6 +246,58 @@ platform :ios do
...
+  # https://github.com/fastlane/fastlane/blob/eb4584e0394c1fecccd70d4313260f4707f2af80/fastlane/Fastfile#L346-L380
+  desc "Generate changelog from GitHub compare and PR data for mentioning GitHub usernames in release notes"
+  lane :github_changelog do |options|
+    last_version = sh("git describe --tags --abbrev=0").strip
```
aboedo, line 257: 🤔 would this work when hotfixing an older major?

aboedo: we have a lane around for getting the version number from the file called VERSION, we could use that instead

fastlane/Fastfile
```diff
@@ -273,15 +326,22 @@ platform :ios do
...
+
     # Open CHANGELOG.latest.md in editor
     changelog_filename = "CHANGELOG.latest.md"
     changelog_path = File.absolute_path("../#{changelog_filename}")
     editor = ENV['FASTLANE_EDITOR'] || ENV['EDITOR'] || 'vim'
 
+    if generated_contents.size > 0
+      UI.message("Using auto generated contents:\n#{generated_contents}")
+      File.write(changelog_path, generated_contents)
+    end
```
aboedo, line 339: if size == 0, should we error out?

aboedo (commented):
looking great! left a few questions / comments

fastlane/Fastfile
joshdholtz: Ah, whoops! Forgot I did that 😛

fastlane/Fastfile
joshdholtz: @aboedo That didn't work because at the time its the `<next-verison>-SNAPSHOT` 😬 

The latest tag that was made was `3.X.X` something but that tag is on different branches (and not main) so this command actually found the correct one which was `4.3.0` 🙃

I was hoping to use `.version` but it was already bumped 🤷‍♂️

fastlane/Fastfile
joshdholtz: Ummm, that is probably a good idea! I'll add that 💪

fastlane/Fastfile
aboedo: oohh right that makes sense! 👍

aboedo (approved):

PR #1586, opened 2022-05-18
CF-653: Add `AmazonStore` to `Store` enum

fire-at-will:
This PR adds `AmazonStore` to the `Store` enum.

### Checklist
- TODO: Create corresponding PRs for hybrid SDKs

### Motivation
Resolves [CF-653](https://revenuecats.atlassian.net/browse/CF-653).

### Description
- Added the `AmazonStore` value to the `Store` enum with a value of `6` to ensure that any implementations relying on `Unknown` being `5` don't break.
- Ensured that the `name` value of the case matches the value used by the backend.
- Updated relevant `switch` statements to include the new case.
- Fixed a few Swiftlint warnings to get the CI to pass

### DIFF
--- a/Sources/CodableExtensions/Store+Extensions.swift
+++ b/Sources/CodableExtensions/Store+Extensions.swift
@@ -55,6 +55,7 @@ private extension Store {
         case .playStore: return "play_store"
         case .stripe: return "stripe"
         case .promotional: return "promotional"
+        case .amazonStore: return "amazon_store"
         case .unknownStore: return nil
         }
     }
--- a/Sources/CodableExtensions/Store+Extensions.swift
+++ b/Sources/CodableExtensions/Store+Extensions.swift
@@ -55,7 +55,7 @@ private extension Store {
         case .playStore: return "play_store"
         case .stripe: return "stripe"
         case .promotional: return "promotional"
-        case .amazonStore: return "amazon_store"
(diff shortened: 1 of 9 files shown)


fire-at-will:
Thanks @aboedo! 

I've updated the [testParseStoreFromSubscription](https://github.com/RevenueCat/purchases-ios/blob/b184e50bfb946e4460a57f098ddc4d616c299eca/Tests/UnitTests/Purchasing/EntitlementInfosTests.swift#L556) and [testParseStoreFromNonSubscription](https://github.com/RevenueCat/purchases-ios/blob/b184e50bfb946e4460a57f098ddc4d616c299eca/Tests/UnitTests/Purchasing/EntitlementInfosTests.swift#L691) tests - I figured there had to have been a test for it somewhere but couldn't find them! 

I'd love to look into modularizing this test file and some of its tests to make it a little more manageable, but that's work for another PR

NachoSoto (approved):

aboedo (approved):
> I'd love to look into modularizing this test file and some of its tests to make it a little more manageable, but that's work for another PR

Totally agree, that one's gotten out of hand over time. We can work on it as a separate thing 👍 

In the meantime, this looks great!! Feel free to merge, and it will get shipped as part of the next update! 🚢

PR #1588, opened 2022-05-19
`StoreKitIntegrationTests`: added test to check applying a promotional offer during subscription

NachoSoto:
See [iOS 15.5 release notes](https://developer.apple.com/documentation/ios-ipados-release-notes/ios-ipados-15_5-release-notes):

> Fixed an issue so developers using StoreKit 2 can now initiate a subscription offer purchase for a product that the customer currently subscribes to. (89152302).

This adds a test to verify this behavior.

Blocked for #2020.

## TODO:

- [x] Update CircleCI: https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions

### DIFF
--- a/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
@@ -442,7 +442,9 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
         entitlement = try await self.verifyEntitlementWentThrough(customerInfo)
         let transaction = try await Self.findTransaction(for: product.productIdentifier)
 
-        self.verifyPromotionalOffer(offer, entitlement, transaction)
+        expect(entitlement.latestPurchaseDate) != entitlement.originalPurchaseDate
+        expect(transaction.offerID) == offer.discount.offerIdentifier
+        expect(transaction.offerType) == .promotional
     }
 
 }
@@ -591,28 +593,6 @@ private extension StoreKit1IntegrationTests {
             beEmpty(),
             description: "Expected no entitlements. Got: \(customerInfo.entitlements.all)"
         )
-    }
-
-    @available(iOS 15.2, tvOS 15.2, macOS 12.1, watchOS 8.3, *)
-    func verifyPromotionalOffer(
-        _ offer: PromotionalOffer,
-        _ entitlement: EntitlementInfo,
-        _ transaction: Transaction,
-        file: FileString = #file,
-        line: UInt = #line
-    ) {
-        expect(
-            file: file, line: line,
-            entitlement.latestPurchaseDate
-        ) != entitlement.originalPurchaseDate
-        expect(
-            file: file, line: line,
-            transaction.offerID
-        ) == offer.discount.offerIdentifier
-        expect(
-            file: file, line: line,
-            transaction.offerType
-        ) == .promotional
     }
 
     func expireSubscription(_ entitlement: EntitlementInfo) async throws {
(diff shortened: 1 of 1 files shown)


Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -292,11 +292,64 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
...
+        expect(transaction.offerType) == .promotional
+    }
+
     @available(iOS 15.2, tvOS 15.2, macOS 12.1, watchOS 8.3, *)
     func testPurchaseWithPromotionalOffer() async throws {
         try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
-        try XCTSkipIf(Self.storeKit2Setting == .enabledForCompatibleDevices,
-                      "This test is not currently passing on SK2")
+//        try XCTSkipIf(Self.storeKit2Setting == .enabledForCompatibleDevices,
+//                      "This test is not currently passing on SK2")
```
NachoSoto, line 352: Disabled this for now cause I want to see if it's passing now too on iOS 15.5.

NachoSoto:
As expected this isn't currently passing.
I'll try with Xcode 13.4 / iOS 15.5 [when it becomes available](https://circleci.com/docs/2.0/testing-ios/#supported-xcode-versions).

taquitos:
Note: I rebased this just now 😄

.circleci/config.yml
```diff
@@ -179,6 +179,11 @@ jobs:
     steps:
       - checkout
       - install-dependencies
+      - install-and-create-sim:
+          install-name: iOS 15.4 Simulator
+          sim-device-type: iPhone-13
+          sim-device-runtime: iOS-15-4
+          sim-name: iPhone 13 (15.4)
```
NachoSoto, line 186: We still need to wait for 15.5 though.

taquitos: Yup, without digging into it, I thought we might have it available because it's on my machine with the same version of Xcode as CircleCI.

NachoSoto:
We can use [13.4.1](https://developer.apple.com/documentation/xcode-release-notes/xcode-13_4_1-release-notes).

NachoSoto:
This still doesn't seem to be working:
> Purchase did not return a transaction: Error Domain=ASDServerErrorDomain Code=3903 "The server encountered an error" UserInfo={NSLocalizedDescription=The server encountered an error}

taquitos:
Sorry, I was cleaning up some old branches and this one got caught in the cross-fire

NachoSoto:
> Oh, that being said, I think you can drop the config.yml changes and this should work now.

What makes you say that? I don't think they've fixed this yet, but let me try it.

taquitos:
> > Oh, that being said, I think you can drop the config.yml changes and this should work now.
> 
> What makes you say that? I don't think they've fixed this yet, but let me try it.

I thought "retest with 15.5" was "will work with 15.5", I misread it.

NachoSoto:
> I thought "retest with 15.5" was "will work with 15.5", I misread it.

It should but it still doesn't hehe. I verified that locally when that came out :(

NachoSoto:
Still not passing on iOS 16 beta 5.

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
NachoSoto: #2020

NachoSoto:
I just realized this doesn't work probably for the same reason #2020 doesn't.

NachoSoto:
This is passing now 🎉

PR #1591, opened 2022-05-20
Ensure SubscriptionPeriods are represented as 1week instead of 7days

fire-at-will:
### Checklist
- [x] If applicable, create follow-up issues for `purchases-android` and hybrids (N/A)

### Motivation
This PR resolves [CF-659](https://revenuecats.atlassian.net/browse/CF-659) & `react-native-purchases` issue [#348](https://github.com/RevenueCat/react-native-purchases/issues/348).

Occasionally, StoreKit will state that a period is 7 days instead of 1 week. This unexpected behavior can cause unintended issues for customers on their paywalls and generally decreases developer confidence in the Offerings/Products provided by the SDK.

### Description
- Implemented a `normalization` function to be run on the input parameters of the `SubscriptionPeriod` class that, if possible to do so evenly, converts days into weeks and months into years.
- This function is static since it needs to be executed before the `SubscriptionPeriod` class has completed instantiation.
- Conversion of weeks to months and weeks to years was intentionally left out since those conversion rates don't use whole numbers (variable number of weeks in a month and 52.14 weeks per year)

### DIFF
--- a/Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
@@ -75,3 +75,34 @@ extension SK1Product {
     }
 
 }
+
+extension SubscriptionPeriod {
+
+    /// This function simplifies large numbers of days into months and large numbers
+    /// of months into years if there are no leftover units after the conversion.
+    ///
+    /// Occassionally, StoreKit seems to send back a value 7 days for a 7day trial
+    /// instesad of a value of 1 week for a trial of 7 days in length.
+    /// Source: https://github.com/RevenueCat/react-native-purchases/issues/348
+    internal static func normalizeValueAndUnits(
+        value: Int,
+        unit: Unit
+    ) -> (value: Int, unit: Unit) {
+        switch unit {
+        case .day:
+            if value % 7 == 0 {
+                let numberOfWeeks = value / 7
+                return (value: numberOfWeeks, unit: .week)
+            }
+        case .month:
+            if value % 12 == 0 {
+                let numberOfYears = value / 12
+                return (value: numberOfYears, unit: .year)
+            }
+        case .week, .year:
+            break
+        }
+
+        return (value: value, unit: unit)
+    }
+}
--- a/Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
@@ -82,7 +82,7 @@ extension SubscriptionPeriod {
     /// of months into years if there are no leftover units after the conversion.
     ///
     /// Occassionally, StoreKit seems to send back a value 7 days for a 7day trial
-    /// instesad of a value of 1 week for a trial of 7 days in length.
+    /// instead of a value of 1 week for a trial of 7 days in length.
     /// Source: https://github.com/RevenueCat/react-native-purchases/issues/348
     internal static func normalizeValueAndUnits(
         value: Int,
--- a/Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
@@ -89,17 +89,17 @@ extension SubscriptionPeriod {
         case .day:
             if value.isMultiple(of: 7) {
                 let numberOfWeeks = value / 7
-                return SubscriptionPeriod(value: numberOfWeeks, unit: .week)
+                return .init(value: numberOfWeeks, unit: .week)
             }
         case .month:
             if value.isMultiple(of: 12) {
                 let numberOfYears = value / 12
-                return SubscriptionPeriod(value: numberOfYears, unit: .year)
+                return .init(value: numberOfYears, unit: .year)
             }
         case .week, .year:
             break
         }
 
-        return SubscriptionPeriod(value: value, unit: unit)
+        return self
     }
 }
--- a/Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
@@ -84,25 +84,22 @@ extension SubscriptionPeriod {
     /// Occassionally, StoreKit seems to send back a value 7 days for a 7day trial
     /// instead of a value of 1 week for a trial of 7 days in length.
     /// Source: https://github.com/RevenueCat/react-native-purchases/issues/348
-    internal static func normalizeValueAndUnits(
-        value: Int,
-        unit: Unit
-    ) -> (value: Int, unit: Unit) {
+    internal func normalized() -> SubscriptionPeriod {
         switch unit {
         case .day:
-            if value % 7 == 0 {
+            if value.isMultiple(of: 7) {
                 let numberOfWeeks = value / 7
-                return (value: numberOfWeeks, unit: .week)
+                return SubscriptionPeriod(value: numberOfWeeks, unit: .week)
             }
         case .month:
-            if value % 12 == 0 {
+            if value.isMultiple(of: 12) {
                 let numberOfYears = value / 12
-                return (value: numberOfYears, unit: .year)
+                return SubscriptionPeriod(value: numberOfYears, unit: .year)
             }
         case .week, .year:
             break
         }
 
-        return (value: value, unit: unit)
+        return SubscriptionPeriod(value: value, unit: unit)
     }
 }
--- a/Sources/Purchasing/StoreKitAbstractions/SubscriptionPeriod.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/SubscriptionPeriod.swift
@@ -29,7 +29,8 @@ public class SubscriptionPeriod: NSObject {
     public init(value: Int, unit: Unit) {
         assert(value > 0, "Invalid value: \(value)")
 
-        (self.value, self.unit) = Self.normalizeValueAndUnits(value: value, unit: unit)
+        self.value = value
+        self.unit = unit
     }
 
     /// Units of time used to describe subscription periods.
@@ -54,6 +55,7 @@ public class SubscriptionPeriod: NSObject {
         }
 
         return .init(value: sk1SubscriptionPeriod.numberOfUnits, unit: unit)
+            .normalized()
     }
 
     @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8, *)
@@ -63,6 +65,7 @@ public class SubscriptionPeriod: NSObject {
         }
 
         return .init(value: sk2SubscriptionPeriod.value, unit: unit)
+            .normalized()
     }
 
     public override func isEqual(_ object: Any?) -> Bool {
--- a/Sources/Purchasing/StoreKitAbstractions/SubscriptionPeriod.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/SubscriptionPeriod.swift
@@ -29,8 +29,7 @@ public class SubscriptionPeriod: NSObject {
     public init(value: Int, unit: Unit) {
         assert(value > 0, "Invalid value: \(value)")
 
-        self.value = value
-        self.unit = unit
+        (self.value, self.unit) = Self.normalizeValueAndUnits(value: value, unit: unit)
     }
 
     /// Units of time used to describe subscription periods.
--- a/Tests/UnitTests/Purchasing/StoreKitAbstractions/StoreKitWorkaroundsTests.swift
+++ b/Tests/UnitTests/Purchasing/StoreKitAbstractions/StoreKitWorkaroundsTests.swift
@@ -34,7 +34,7 @@ class StoreKitWorkaroundsTests: TestCase {
             (112, .day, 16, .week),
             (356, .day, 356, .day),
             (712, .day, 712, .day),
-            // Test week simplicfication
+            // Test week simplification
             (1, .week, 1, .week),
             (4, .week, 4, .week),
             (24, .week, 24, .week),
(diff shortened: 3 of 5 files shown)


Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
```diff
@@ -75,3 +75,34 @@ extension SK1Product {
...
 
 }
+
+extension SubscriptionPeriod {
+
+    /// This function simplifies large numbers of days into months and large numbers
+    /// of months into years if there are no leftover units after the conversion.
+    ///
+    /// Occassionally, StoreKit seems to send back a value 7 days for a 7day trial
+    /// instesad of a value of 1 week for a trial of 7 days in length.
```
vegaro, line 85: Typo instead

Sources/Purchasing/StoreKitAbstractions/SubscriptionPeriod.swift
```diff
@@ -29,8 +29,7 @@ public class SubscriptionPeriod: NSObject {
     public init(value: Int, unit: Unit) {
         assert(value > 0, "Invalid value: \(value)")
 
-        self.value = value
-        self.unit = unit
+        (self.value, self.unit) = Self.normalizeValueAndUnits(value: value, unit: unit)
```
vegaro, line 32: Love that destructuring :chefskiss:

vegaro (approved):
Looks good! Did you manage to reproduce it by the way? We should probably file a radar if you haven't done yet

Tests/UnitTests/Purchasing/StoreKitAbstractions/StoreKitWorkaroundsTests.swift
```diff
@@ -0,0 +1,62 @@
...
+            (7, .day, 1, .week),
+            (14, .day, 2, .week),
+            (21, .day, 3, .week),
+            (28, .day, 4, .week),
+            (30, .day, 30, .day),
+            (56, .day, 8, .week),
+            (112, .day, 16, .week),
+            (356, .day, 356, .day),
+            (712, .day, 712, .day),
+            // Test week simplicfication
```
NachoSoto, line 37:
```suggestion
            // Test week simplification
```

NachoSoto (changes requested):
Looks good in general, but I wouldn't normalize all `SubscriptionPeriod` creation, that feels like spooky action at a distance.
And because we don't have tests for `SubscriptionPeriod` initialization (since it was just a basic data type), this is accidentally mutating the data that goes in.

Instead of `normalize` though, since we want to keep the type immutable, I would change it to `normalized`:
```swift
extension SubscriptionPeriod { 
  func normalized() -> Self {
    return .init(...)
  }
}
```

Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
```diff
@@ -75,3 +75,34 @@ extension SK1Product {
...
+    /// Source: https://github.com/RevenueCat/react-native-purchases/issues/348
+    internal static func normalizeValueAndUnits(
+        value: Int,
+        unit: Unit
+    ) -> (value: Int, unit: Unit) {
+        switch unit {
+        case .day:
+            if value % 7 == 0 {
```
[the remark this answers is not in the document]
fire-at-will: Whaaaaat I didn't know Integer had this function!! 🤯🤯

Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
```diff
@@ -75,3 +75,34 @@ extension SK1Product {
...
+    /// This function simplifies large numbers of days into months and large numbers
+    /// of months into years if there are no leftover units after the conversion.
+    ///
+    /// Occassionally, StoreKit seems to send back a value 7 days for a 7day trial
+    /// instesad of a value of 1 week for a trial of 7 days in length.
+    /// Source: https://github.com/RevenueCat/react-native-purchases/issues/348
+    internal static func normalizeValueAndUnits(
+        value: Int,
+        unit: Unit
+    ) -> (value: Int, unit: Unit) {
```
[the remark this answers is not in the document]
fire-at-will: I don't think making this a mutating function on `SubscriptionPeriod` is possible since both `value` and `unit` are `let` constants. Additionally, I'm not sure if making this function's signature
```swift
internal static func normalize(_ subscriptionPeriod: SubscriptionPeriod) -> SubscriptionPeriod
```
is feasible either since the `normalize` function is called from within `SubscriptionPeriod`'s initializer, and such a call would lead to an endless call loop.

I'm not a fan of the function being static, but I think the only alternative is having the developer call `normalize` every time a `SubscriptionPeriod` is instantiated, which doesn't seem like a good design.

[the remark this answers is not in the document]
fire-at-will: Oh, I just saw [this comment](https://github.com/RevenueCat/purchases-ios/pull/1591#discussion_r878358928) 😅

I think it makes a lot of sense! I'll adopt that approach since this is just meant to be a StoreKit fix.

Tests/UnitTests/Purchasing/StoreKitAbstractions/StoreKitWorkaroundsTests.swift
```diff
@@ -0,0 +1,62 @@
...
+import Nimble
+@testable import RevenueCat
+import XCTest
+
+class StoreKitWorkaroundsTests: TestCase {
+
+    func testSimplifyValueAndUnits() {
+        let expectations: [(
+            inputValue: Int, inputUnit: SubscriptionPeriod.Unit,
+            expectedValue: Int, expectedUnit: SubscriptionPeriod.Unit
```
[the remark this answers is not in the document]
fire-at-will: Great feedback! It's so convenient that Nimble has the `description` parameter 🙂

Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
```diff
@@ -75,3 +75,31 @@ extension SK1Product {
...
+    ///
+    /// Occassionally, StoreKit seems to send back a value 7 days for a 7day trial
+    /// instead of a value of 1 week for a trial of 7 days in length.
+    /// Source: https://github.com/RevenueCat/react-native-purchases/issues/348
+    internal func normalized() -> SubscriptionPeriod {
+        switch unit {
+        case .day:
+            if value.isMultiple(of: 7) {
+                let numberOfWeeks = value / 7
+                return SubscriptionPeriod(value: numberOfWeeks, unit: .week)
```
NachoSoto, line 92: This could also be:
```suggestion
                return .init(value: numberOfWeeks, unit: .week)
```

Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
```diff
@@ -75,3 +75,31 @@ extension SK1Product {
...
+        case .month:
+            if value.isMultiple(of: 12) {
+                let numberOfYears = value / 12
+                return SubscriptionPeriod(value: numberOfYears, unit: .year)
+            }
+        case .week, .year:
+            break
+        }
+
+        return SubscriptionPeriod(value: value, unit: unit)
```
NachoSoto, line 103: Now you can just do this 😄 
```suggestion
        return self
```

Tests/UnitTests/Purchasing/StoreKitAbstractions/StoreKitWorkaroundsTests.swift
```diff
@@ -0,0 +1,69 @@
...
+            // Ensure year inputs return the same value
+            (1, .year, 1, .year),
+            (2, .year, 2, .year)
+        ]
+
+        for expectation in expectations {
+            let outputPeriod = SubscriptionPeriod(
+                value: expectation.inputValue,
+                unit: expectation.inputUnit
+            ).normalized()
```
NachoSoto, line 56: Maybe it makes sense to compare the whole period now?
```swift
let period = SubscriptionPeriod(
  value: expectation.inputValue,
  unit: expectation.inputUnit
)
let normalized = period.normalized()
let expected = SubscriptionPeriod(
  value: expectation.expectedValue,
  unit: expectation.expectedUnit
)

expect(period).to(
  equal(expected),
  description: "..."
)
```

Sources/Purchasing/StoreKitAbstractions/SubscriptionPeriod.swift
```diff
@@ -64,6 +65,7 @@ public class SubscriptionPeriod: NSObject {
         }
 
         return .init(value: sk2SubscriptionPeriod.value, unit: unit)
+            .normalized()
```
NachoSoto, line 68: Awesome.

fire-at-will:
@NachoSoto @vegaro Thanks for the feedback! 😄 

I made the following changes in [this commit](https://github.com/RevenueCat/purchases-ios/pull/1591/commits/3d2af2fb7b3147f1f49fe373ee73422825bc3bdc):
- Change the `normalizeValueAndUnit` function signature to `internal func normalized() -> SubscriptionPeriod`. Note that this is no longer `static` 🥳
- Move `normalized` call site from `SubscriptionPeriod.init` to `SubscriptionPeriod.from(sk1SubscriptionPeriod: SKProductSubscriptionPeriod)` and `SubscriptionPeriod.from(sk2SubscriptionPeriod: StoreKit.Product.SubscriptionPeriod)` so the fix is only used when the `SubscriptionPeriod` is created from StoreKit.
- Added descriptions to the assertions in the unit test.
- Change normalization logic to use `Int.isMultipleOf` instead of the modulo operator.
- Fixed typos 😎

Sources/Purchasing/StoreKitAbstractions/SubscriptionPeriod.swift
```diff
@@ -55,6 +55,7 @@ public class SubscriptionPeriod: NSObject {
         }
 
         return .init(value: sk1SubscriptionPeriod.numberOfUnits, unit: unit)
+            .normalized()
```
NachoSoto, line 58: We can't test SK2's because it has no initializers, but maybe add a test in `SubscriptionPeriodTests` that ensures this is called here? It doesn't have to be comprehensive at all since you have all those examples in `StoreKitWorkaroundsTests`, maybe just one.

NachoSoto (approved):
🎉

Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
fire-at-will: You're right! 🤯 
The fewer objects the SDK creates in RAM, the better 😃

NachoSoto:
I'm excited for this change.
I just realized I had this in my app:
```swift
var periodTitle: String {
	assert(self.unit == .week, "Only week trials supported: \(self.unit)")
	
	return Plurals.Pro.Trial.intervalWeeks(self.value)
}
```
The `assert` won't run in production, but I'd probably be displaying "7 weeks" if users ran into this issue 😵‍💫

Sources/Purchasing/StoreKitAbstractions/SubscriptionPeriod.swift
fire-at-will: Oooh, added in one with a few different scenarios to test. Great idea!

Tests/UnitTests/Purchasing/StoreKitAbstractions/StoreKitWorkaroundsTests.swift
fire-at-will: That's a great way to clean this test up, made the changes

Tests/UnitTests/Purchasing/StoreKitAbstractions/SubscriptionPeriodTests.swift
```diff
@@ -73,4 +73,63 @@ class SubscriptionPeriodTests: TestCase {
...
+                    Expected \(description(for: sk1SubscriptionPeriod)) to become \(expected.debugDescription).
+                    """
+            )
+        }
+    }
+
+    /// Necessary since SKProductSubscriptionPeriod.debugDescription & SKProductSubscriptionPeriod.description
+    /// return the object's address in memory.
+    @available(iOS 11.2, *)
+    private func description(for skProductSubscriptionPeriod: SKProductSubscriptionPeriod) -> String {
```
NachoSoto, line 116: Instead of doing this, let's implement `CustomStringConvertible ` in the tests:
```swift
extension SubscriptionPeriod: CustomStringConvertible {

  var description: String {}

}
```

Then you shouldn't need to manually call the method.

Tests/UnitTests/Purchasing/StoreKitAbstractions/StoreKitWorkaroundsTests.swift
```diff
@@ -0,0 +1,70 @@
...
+            )
+            let normalized = period.normalized()
+            let expected = SubscriptionPeriod(
+                value: expectation.expectedValue,
+                unit: expectation.expectedUnit
+            )
+
+            expect(normalized).to(
+                equal(expected),
+                description: "Expected \(period.debugDescription) to become \(expected.debugDescription)."
```
NachoSoto, line 65: Then you can use the same `"Expected \(normalized) to become \(expected)"` here too.

vegaro (approved):

Tests/UnitTests/Purchasing/StoreKitAbstractions/SubscriptionPeriodTests.swift
NachoSoto: I see this is where I got it wrong, sorry! I should have realized you had `SKProductSubscriptionPeriod` as the parameter not `SubscriptionPeriod`.

Issue #1528, opened 2022-04-25
EXC_BAD_ACCESS at StoreKit2TransactionListener.swift line 32


mrcperello:
- [x] I have updated Purchases SDK to the latest version
- [ ] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/CONTRIBUTING.md) -> Link doesn't work.
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Description**
Fatal crash when calling `Purchases.configure(withAPIKey:)`

1. Environment
   1. Platform: iPad
   2. SDK version: 4.1.0
   3. StoreKit 2 enabled (Y/N): Y
   4. OS version: iPadOS 15.0.0
   5. Xcode version: -
   6. How widespread is the issue. Percentage of devices affected: *1 user two instances*
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issuelogs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior: *On app launch, we configure the Purchases SDK, and while configuring it crashes.*
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

```
Crashed: com.apple.main-thread
0  libswiftCore.dylib             0x3a3acc swift::ResolveAsSymbolicReference::operator()(swift::Demangle::__runtime::SymbolicReferenceKind, swift::Demangle::__runtime::Directness, int, void const*) + 124
1  libswiftCore.dylib             0x3c9d58 swift::Demangle::__runtime::Demangler::demangleSymbolicReference(unsigned char) + 252
2  libswiftCore.dylib             0x3c65f8 swift::Demangle::__runtime::Demangler::demangleType(__swift::__runtime::llvm::StringRef, std::__1::function<swift::Demangle::__runtime::Node* (swift::Demangle::__runtime::SymbolicReferenceKind, swift::Demangle::__runtime::Directness, int, void const*)>) + 324
[...]

PR #1593, opened 2022-05-20
`PurchasesOrchestrator`: don't listen for StoreKit 2 transactions if it's disabled

NachoSoto:
I tried adding tests for this, but because the listener is created internally, there is no time to override it with the mock before verifying if `listenForTransactions` is called.

Will at least prevent #1528 if SK2 is disabled.

### DIFF
--- a/Sources/Purchasing/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/PurchasesOrchestrator.swift
@@ -99,7 +99,8 @@ class PurchasesOrchestrator {
         self.manageSubscriptionsHelper = manageSubscriptionsHelper
         self.beginRefundRequestHelper = beginRefundRequestHelper
 
-        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *) {
+        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *),
+           systemInfo.storeKit2Setting == .enabledForCompatibleDevices {
             storeKit2TransactionListener.listenForTransactions()
             storeKit2StorefrontListener.listenForStorefrontChanges()
         }


Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -99,7 +99,8 @@ class PurchasesOrchestrator {
         self.manageSubscriptionsHelper = manageSubscriptionsHelper
         self.beginRefundRequestHelper = beginRefundRequestHelper
 
-        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *) {
+        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *),
+           systemInfo.storeKit2Setting == .enabledForCompatibleDevices {
```
Juanpe, line 103: 👌🏼

aboedo: feels like we should have an accompanying test for this, so we don't inadvertently break it in the future. It can be done as a separate PR

aboedo (approved):
approved, but could you add a test for it?

NachoSoto:
See the description:

> I tried adding tests for this, but because the listener is created internally, there is no time to override it with the mock before verifying if listenForTransactions is called.

I couldn't think of a way to add a test for it.

aboedo:
@NachoSoto I see, good catch. Good to go then! 
Once we have https://github.com/RevenueCat/purchases-ios/pull/1596/ merged we should be able to inject the property and then make tests for this, right?

NachoSoto:
> Once we have https://github.com/RevenueCat/purchases-ios/pull/1596 merged we should be able to inject the property and then make tests for this,

The rabbit hole goes deep and deep:
- We need 2 constructors of `PurchasesOrchestrator`, one with an `@available` annotation
- `Purchases` needs to do `if #available` to decide which constructor to call

Here's a proposed solution: #1618

It adds a fair amount of complexity, but I guess it's important to test this.

NachoSoto:
I'll merge this for now, maybe we can have it out in the release.

PR #1596, opened 2022-05-20
`PurchasesOrchestrator`: removed `lazy` hack for properties with `@available`

NachoSoto:
Fixes [CF-682], and probably #1528.
Looks like possibly something was broken demangling `lazy` properties with `@available` in iOS 15.0 (per the 2 bug reports).

This changes the approach to simply having `Any?` as the value, and making the exposed `@available` property a derived one.

[CF-682]: https://revenuecats.atlassian.net/browse/CF-682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/PurchasesOrchestrator.swift
@@ -66,11 +66,23 @@ class PurchasesOrchestrator {
     private let beginRefundRequestHelper: BeginRefundRequestHelper
     private let lock = NSRecursiveLock()
 
-    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
-    lazy var storeKit2TransactionListener = StoreKit2TransactionListener(delegate: self)
+    // Can't have these properties with `@available`.
+    // swiftlint:disable identifier_name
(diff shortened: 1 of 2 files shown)


Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -66,11 +66,23 @@ class PurchasesOrchestrator {
     private let beginRefundRequestHelper: BeginRefundRequestHelper
     private let lock = NSRecursiveLock()
```
Juanpe, line 68: I think it's a great workaround, but another possible solution could be to create an object that encapsulates all this logic, something like `StoreKit2Listeners` without `@available`, and it's only initialised if StoreKit2 is available

NachoSoto: `StoreKit2TransactionListener` is already a wrapper, so that feels like creating a wrapper for a wrapper :P

aboedo (approved):

PR #1602, opened 2022-05-24
`PurchasesOrchestrator.purchase(sk2Product:promotionalOffer:)`: simplified implementation with new operator

NachoSoto:
Both `return` statements where the same, the `else` was just providing a default value to `customerInfoIfSynced`.
Because `??` doesn't support an `async` function, we couldn't use that. By using this new operator we can simplify the implementation to a single `return` statement.

### DIFF
--- a/Sources/FoundationExtensions/Operators+Extensions.swift
+++ b/Sources/FoundationExtensions/Operators+Extensions.swift
@@ -14,6 +14,7 @@
 infix operator ???
 
 /// Equivalent to `??` but allows an `async` default value.
+/// See https://github.com/apple/swift-evolution/blob/main/proposals/0296-async-await.md#future-directions
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
 func ??? <T>(value: T?, defaultValue: @autoclosure () async throws -> T) async rethrows -> T {
     if let value = value {
--- a/Sources/FoundationExtensions/Operators+Extensions.swift
+++ b/Sources/FoundationExtensions/Operators+Extensions.swift
@@ -1,0 +1,24 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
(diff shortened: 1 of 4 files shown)


Sources/FoundationExtensions/Operators+Extensions.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+//  Operators+Extensions.swift
+//
+//  Created by Nacho Soto on 5/23/22.
+
+infix operator ???
+
+/// Equivalent to `??` but allows an `async` default value.
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+func ??? <T>(value: T?, defaultValue: @autoclosure () async throws -> T) async rethrows -> T {
```
NachoSoto, line 18: This won't be necessary [when Swift gets `reasync`](https://twitter.com/slava_pestov/status/1528926247828525057?s=21&t=armm7A8ik5cATfXmPQXOjg).

aboedo: maybe we can add a link here pointing to https://github.com/apple/swift-evolution/blob/main/proposals/0296-async-await.md#future-directions? That way if a dev runs into this in the future after `reasync` gets implemented, they'll know that it serves a similar purpose and can be replaced?

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -356,13 +356,10 @@ class PurchasesOrchestrator {
             .handle(purchaseResult: result)
         let transaction = sk2Transaction.map(StoreTransaction.init(sk2Transaction:))
 
-        if let customerInfo = customerInfoIfSynced {
-            return (transaction, customerInfo, userCancelled)
-        } else {
-            let customerInfo = try await self.syncPurchases(receiptRefreshPolicy: .always, isRestore: false)
+        let customerInfo = try await customerInfoIfSynced
+        ??? (await self.syncPurchases(receiptRefreshPolicy: .always, isRestore: false))
```
aboedo, line 360: this is cool

aboedo (approved):

Sources/FoundationExtensions/Operators+Extensions.swift
NachoSoto: Done.

PR #1604, opened 2022-05-24
Added `@IgnoreHashable` property wrapper

NachoSoto:
This will be used for the new implementation of `CustomerInfoResponse`, and also for `var rawData: [String: Any]` properties that can't conform to `Hashable`, and that we wouldn't want to use for comparison since they contain duplicated data.


aboedo (approved):
cool idea!

Issue #1594, opened 2022-05-20
`showManageSubscriptions()` never uses native AppStore method


philmitchell:
- [ x] I have updated Purchases SDK to the latest version
- [ x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Development/CONTRIBUTING.md)
- [ x] I have searched the [Community](https://community.revenuecat.com)
- [x ] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x ] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**

Even on a device running iOS15, `Purchases.showManageSubscriptions()` loads the subscriptions UI via Safari rather than a native action sheet. 

1. Environment
   1. Platform: iOS
   2. SDK version: 4.3
   3. StoreKit 2 enabled (Y/N): Have tried it both ways
   4. OS version: 15.4.1
   5. Xcode version: 13.3
   6. How widespread is the issue. Percentage of devices affected. 100%
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior
Where supported, it should open the native action sheet, which provides a superior UX.

5. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

This seems to be because we have:

  `SystemInfo.appleSubscriptionsURL = URL(string: "https://rev.cat/manage-apple-subscription")`

This url never passes the `isAppleSubscription()` test, therefore we always fallback to the `openURL()` method.

**Additional context**

aboedo:
Hi @philmitchell 👋 
Great catch!! That's a refactor gone bad. We'll get it fixed 💪 
Thanks for reporting!

NachoSoto:
Submitted a fix for this. We'll probably have it out in a release later today :)

NachoSoto:
The fix for this is out in 4.4.0

PR #1607, opened 2022-05-24
`ManageSubscriptionsHelper`: fixed discrepancy between `SystemInfo.isAppleSubscription(managementURL:)` and `SystemInfo.appleSubscriptionsURL`

NachoSoto:
Fixes [CF-684] and #1594.

`SystemInfo.appleSubscriptionsURL` was a short URL that didn't pass the `SystemInfo.isAppleSubscription(managementURL:)` test. This is fixed and covered by a test now.

### Other changes:
- Made `SystemInfo.isAppleSubscription(managementURL:)` and `SystemInfo.appleSubscriptionsURL` static since they don't need any properties in `SystemInfo`.
- Forced-unwrapped `SystemInfo.appleSubscriptionsURL` to remove unnecessary error handling. This is safe because the URL is provided statically at compile time. Per the docs "This initializer returns nil if the string doesn’t represent a valid URL. For example, an empty string or one containing characters that are illegal in a URL produces nil.". This is known at compile time, and covered by tests, so it could never return `nil`.
- Combined redundant tests in `ManageSubscriptionsHelper`. They were all doing exactly the same for both platforms.

[CF-684]: https://revenuecats.atlassian.net/browse/CF-684?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Tests/UnitTests/Misc/SystemInfoTests.swift
+++ b/Tests/UnitTests/Misc/SystemInfoTests.swift
@@ -5,7 +5,7 @@ import XCTest
 
 class SystemInfoTests: TestCase {
 
-    func testproxyURL() {
+    func testProxyURL() {
         let defaultURL = URL(string: "https://api.revenuecat.com")
         expect(SystemInfo.serverHostURL) == defaultURL
         expect(SystemInfo.proxyURL).to(beNil())
@@ -22,7 +22,7 @@ class SystemInfoTests: TestCase {
         expect(SystemInfo.serverHostURL) == defaultURL
     }
 
-    func testsystemVersion() {
+    func testSystemVersion() {
         expect(SystemInfo.systemVersion) == ProcessInfo().operatingSystemVersionString
     }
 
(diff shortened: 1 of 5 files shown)


Tests/UnitTests/Misc/SystemInfoTests.swift
```diff
@@ -67,6 +67,26 @@ class SystemInfoTests: TestCase {
...
+        expect(SystemInfo.isAppleSubscription(managementURL: URL(string: "www.google.com")!)) == false
+    }
+
+    func testIsAppleSubscriptionURLWithStandardURL() {
+        expect(SystemInfo.isAppleSubscription(managementURL: SystemInfo.appleSubscriptionsURL)) == true
+    }
+
+    func testIsAppleSubscriptionURLWithOtherStandardURL() {
+        expect(SystemInfo.isAppleSubscription(
+            managementURL: URL(string: "https://apple.com/manage_subscription")!
```
NachoSoto, line 80: This is what we have in `CustomerInfo` test data, and I believe what the backend returns.

aboedo (approved):

Tests/UnitTests/Misc/SystemInfoTests.swift
NachoSoto: Oh gotcha I'll remove this test then.

PR #1608, opened 2022-05-24
`Purchases.customerInfo()`: added overload with a new `CacheFetchPolicy`

NachoSoto:
Fixes [CF-19].
This allows specifying `CustomerInfoManager` fetching behavior, both as a public API for users, and also internally.
Some future uses of `CustomerInfoManager` might want to ensure that we don't rely on cached data, for example.

Thanks to the `APITester` we know that the existing methods still work, this is purely an additive change.

This is the core of the change:
```swift
/// Specifies the behavior for a caching API.
@objc(RCCacheFetchPolicy)
public enum CacheFetchPolicy: Int {
    /// Returns values from the cache, or throws an error if not available.
    case fromCacheOnly

    /// Always fetch the most up-to-date data.
    case fetchCurrent

    /// Returns the cached data if available and not stale, or fetches up-to-date data.
    case notStaleCachedOrFetched

    /// Default behavior: returns the cached data if available (even if stale), or fetches up-to-date data.
    case cachedOrFetched

    /// Default ``CacheFetchPolicy`` behavior.
    public static let `default`: Self = .cachedOrFetched
}
```

The new `CustomerInfoManagerGetCustomerInfoTests` check all the different combinations of fetching behavior.
I've taken some of the existing tests and refactored them to be `async`, as well as added a bunch more tests.

> **Note**: 
There is also one big hidden bug fix:

`Purchases.getCustomerInfo` now calls `BackendError.asPurchasesError`, so we don't end up returning an internal error type.
This was missed because we always return generic `Error` types, so we have no way to know at compile time that we only return `ErrorCode`s.

[CF-19]: https://revenuecats.atlassian.net/browse/CF-19?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Error Handling/BackendError.swift
+++ b/Sources/Error Handling/BackendError.swift
@@ -23,6 +23,7 @@ enum BackendError: Error, Equatable {
     case emptySubscriberAttributes(Source)
     case missingReceiptFile(Source)
     case missingTransactionProductIdentifier(Source)
+    case missingCachedCustomerInfo(Source)
     case unexpectedBackendResponse(UnexpectedBackendResponseError, extraContext: String?, Source)
 
 }
@@ -51,6 +52,12 @@ extension BackendError {
         file: String = #fileID, function: String = #function, line: UInt = #line
     ) -> Self {
         return .missingReceiptFile(.init(file: file, function: function, line: line))
+    }
+
+    static func missingCachedCustomerInfo(
+        file: String = #fileID, function: String = #function, line: UInt = #line
+    ) -> Self {
+        return .missingCachedCustomerInfo(.init(file: file, function: function, line: line))
     }
 
     static func unexpectedBackendResponse(
@@ -93,6 +100,12 @@ extension BackendError: ErrorCodeConvertible {
                 line: source.line
             )
 
+        case let .missingCachedCustomerInfo(source):
+            return ErrorUtils.customerInfoError(withMessage: Strings.purchase.missing_cached_customer_info.description,
+                                                fileName: source.file,
+                                                functionName: source.function,
+                                                line: source.line)
+
         case let .unexpectedBackendResponse(error, extraContext, source):
             return ErrorUtils.unexpectedBackendResponse(withSubError: error,
                                                         extraContext: extraContext,
@@ -128,6 +141,7 @@ extension BackendError {
              .emptySubscriberAttributes,
              .missingReceiptFile,
              .missingTransactionProductIdentifier,
+             .missingCachedCustomerInfo,
              .unexpectedBackendResponse:
             return nil
         }
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -61,22 +61,22 @@ class CustomerInfoManager {
     func fetchAndCacheCustomerInfoIfStale(appUserID: String,
                                           isAppBackgrounded: Bool,
                                           completion: ((Result<CustomerInfo, BackendError>) -> Void)?) {
-        let cachedCustomerInfo = cachedCustomerInfo(appUserID: appUserID)
-        let isCacheStale = deviceCache.isCustomerInfoCacheStale(appUserID: appUserID,
-                                                                isAppBackgrounded: isAppBackgrounded)
+        let cachedCustomerInfo = self.cachedCustomerInfo(appUserID: appUserID)
+        let isCacheStale = self.deviceCache.isCustomerInfoCacheStale(appUserID: appUserID,
+                                                                     isAppBackgrounded: isAppBackgrounded)
 
         guard !isCacheStale, let customerInfo = cachedCustomerInfo else {
             Logger.debug(isAppBackgrounded
                             ? Strings.customerInfo.customerinfo_stale_updating_in_background
                             : Strings.customerInfo.customerinfo_stale_updating_in_foreground)
-            fetchAndCacheCustomerInfo(appUserID: appUserID,
-                                      isAppBackgrounded: isAppBackgrounded,
-                                      completion: completion)
+            self.fetchAndCacheCustomerInfo(appUserID: appUserID,
+                                           isAppBackgrounded: isAppBackgrounded,
+                                           completion: completion)
             return
         }
 
         if let completion = completion {
-            operationDispatcher.dispatchOnMainThread {
+            self.operationDispatcher.dispatchOnMainThread {
                 completion(.success(customerInfo))
             }
         }
@@ -90,27 +90,46 @@ class CustomerInfoManager {
         sendUpdateIfChanged(customerInfo: info)
     }
 
-    func customerInfo(appUserID: String, completion: ((Result<CustomerInfo, BackendError>) -> Void)?) {
-        let infoFromCache = cachedCustomerInfo(appUserID: appUserID)
-        var completionCalled = false
+    func customerInfo(
+        appUserID: String,
+        fetchPolicy: CacheFetchPolicy,
+        completion: ((Result<CustomerInfo, BackendError>) -> Void)?
+    ) {
+        switch fetchPolicy {
+        case .fromCacheOnly:
+            completion?(
+                Result(self.cachedCustomerInfo(appUserID: appUserID), .missingCachedCustomerInfo())
+            )
 
-        if let infoFromCache = infoFromCache {
-            Logger.debug(Strings.customerInfo.vending_cache)
-            if let completion = completion {
-                completionCalled = true
-                operationDispatcher.dispatchOnMainThread {
-                    completion(.success(infoFromCache))
+        case .fetchCurrent:
+            self.systemInfo.isApplicationBackgrounded { isAppBackgrounded in
+                self.fetchAndCacheCustomerInfo(appUserID: appUserID,
+                                               isAppBackgrounded: isAppBackgrounded,
+                                               completion: completion)
+            }
+
+        case .cachedOrFetched:
+            let infoFromCache = self.cachedCustomerInfo(appUserID: appUserID)
+            var completionCalled = false
+
+            if let infoFromCache = infoFromCache {
+                Logger.debug(Strings.customerInfo.vending_cache)
+                if let completion = completion {
+                    completionCalled = true
+                    self.operationDispatcher.dispatchOnMainThread {
+                        completion(.success(infoFromCache))
+                    }
                 }
             }
-        }
 
-        // Prevent calling completion twice.
-        let completionIfNotCalledAlready = completionCalled ? nil : completion
+            // Prevent calling completion twice.
+            let completionIfNotCalledAlready = completionCalled ? nil : completion
 
-        systemInfo.isApplicationBackgrounded { isAppBackgrounded in
-            self.fetchAndCacheCustomerInfoIfStale(appUserID: appUserID,
-                                                  isAppBackgrounded: isAppBackgrounded,
-                                                  completion: completionIfNotCalledAlready)
+            self.systemInfo.isApplicationBackgrounded { isAppBackgrounded in
+                self.fetchAndCacheCustomerInfoIfStale(appUserID: appUserID,
+                                                      isAppBackgrounded: isAppBackgrounded,
+                                                      completion: completionIfNotCalledAlready)
+            }
         }
     }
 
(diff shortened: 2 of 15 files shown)


Sources/Purchasing/CacheFetchPolicy.swift
```diff
@@ -0,0 +1,30 @@
...
+public enum CacheFetchPolicy: Int {
+
+    /// Returns values from the cache, or throws an error if not available.
+    case fromCacheOnly
+
+    /// Always fetch the most up-to-date data.
+    case fetchCurrent
+
+    /// Default behavior: retruns the cache data if available and not stale, or fetches up-to-date data.
+    case cachedOrFetched
```
NachoSoto, line 25: Feedback welcome on these names.

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
```diff
@@ -415,3 +378,129 @@ class CustomerInfoManagerTests: TestCase {
     }
 
 }
+
+// iOS 13.0+ only because these tests are async
```
NachoSoto, line 382: Not ideal that we don't run these on iOS 12.x, but a few weeks ago we weren't even running any tests on iOS 12. So I think it's a good compromise, because the tests are _a lot_ simpler.

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
```diff
@@ -415,3 +378,129 @@ class CustomerInfoManagerTests: TestCase {
...
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+class CustomerInfoManagerGetCustomerInfoTests: BaseCustomerInfoManagerTests {
```
NachoSoto, line 384: I think I covered all scenarios, but let me know if I missed any or if it'd be worth adding any more.

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -214,9 +233,14 @@ class CustomerInfoManager {
...
     @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
-    func customerInfo(appUserID: String) async throws -> CustomerInfo {
+    func customerInfo(
+        appUserID: String,
+        fetchPolicy: CacheFetchPolicy
+    ) async throws -> CustomerInfo {
         return try await withCheckedThrowingContinuation { continuation in
-            return self.customerInfo(appUserID: appUserID, completion: continuation.resume)
+            return self.customerInfo(appUserID: appUserID,
+                                     fetchPolicy: fetchPolicy,
```
NachoSoto, line 242: I changed all these calls to explicitly pass the previous behavior. But now that we have this feature, it's possible that we might want to customize it and use a different one.

Sources/Error Handling/BackendError.swift
```diff
@@ -93,6 +100,12 @@ extension BackendError: ErrorCodeConvertible {
                 line: source.line
             )
 
+        case let .missingCachedCustomerInfo(source):
+            return ErrorUtils.customerInfoError(withMessage: Strings.purchase.missing_cached_customer_info.description,
+                                                fileName: source.file,
+                                                functionName: source.function,
+                                                line: source.line)
+
```
NachoSoto, line 108: The new error infrastructure makes this so easy 🎉

Sources/Purchasing/CacheFetchPolicy.swift
```diff
@@ -0,0 +1,30 @@
...
+public enum CacheFetchPolicy: Int {
+
+    /// Returns values from the cache, or throws an error if not available.
+    case fromCacheOnly
+
+    /// Always fetch the most up-to-date data.
+    case fetchCurrent
+
+    /// Default behavior: returns the cache data if available and not stale, or fetches up-to-date data.
+    case cachedOrFetched
```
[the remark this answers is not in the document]
NachoSoto: Great catch! This is what I arrived at, let me know what you think:
```swift
public enum CacheFetchPolicy: Int {
    /// Returns values from the cache, or throws an error if not available.
    case fromCacheOnly

    /// Always fetch the most up-to-date data.
    case fetchCurrent

    /// Returns the cached data if available and not stale, or fetches up-to-date data.
    case notStaleCachedOrFetched

    /// Default behavior: returns the cached data if available (even if stale), or fetches up-to-date data.
    case cachedOrFetched

    /// Default ``CacheFetchPolicy`` behavior.
    public static let `default`: Self = .cachedOrFetched
}
```

You're right the docs for `cachedOrFetched` were wrong, but the implementation and tests were correct.

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -90,27 +90,67 @@ class CustomerInfoManager {
...
+                    if let completion = completion {
+                        self.operationDispatcher.dispatchOnMainThread {
+                            completion(.success(infoFromCache))
+                        }
+                    }
+                } else {
+                    self.fetchAndCacheCustomerInfo(appUserID: appUserID,
+                                                   isAppBackgrounded: isAppBackgrounded,
+                                                   completion: completion)
+                }
```
NachoSoto, line 152: This behavior makes a lot more sense. I wonder if we should change some calls to use this instead now.

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
```diff
@@ -217,61 +224,58 @@ class CustomerInfoManagerTests: TestCase {
...
-        let appUserID = "myUser"
-        mockDeviceCache.cachedCustomerInfo[appUserID] = object
+        mockDeviceCache.cachedCustomerInfo[Self.appUserID] = object
 
-        customerInfoManager.sendCachedCustomerInfoIfAvailable(appUserID: appUserID)
+        customerInfoManager.sendCachedCustomerInfoIfAvailable(appUserID: Self.appUserID)
         expect(self.mockOperationDispatcher.invokedDispatchOnMainThreadCount) == 1
     }
 
     func testCustomerInfoReturnsFromCacheIfAvailable() {
```
NachoSoto, line 272: All the existing tests are left unchanged, meaning the existing default behavior is still the same.

Sources/Identity/CustomerInfoManager.swift
aboedo: I agree. Let's run it by the support team, they might have thoughts based on customer feedback.

I think we can merge this PR as is (with the current behavior), then ask the support team (and broader team) for feedback regarding the proposed change, then if folks agree make a release that makes this the default. 
We'd have to also make a purchases-android equivalent release so that we don't have inconsistent behavior in the hybrids.

Sources/Purchasing/Purchases.swift
```diff
@@ -966,21 +967,33 @@ public extension Purchases {
...
      * Get latest available customer  info.
-     * Returns a value immediately if ``CustomerInfo`` is cached.
+     *
+     * - Parameter fetchPolicy: The behavior for what to do regarding caching.
+     * - Parameter completion: A completion block called when customer info is available and not stale.
+     */
+    @objc func getCustomerInfo(fetchPolicy: CacheFetchPolicy, completion: @escaping (CustomerInfo?, Error?) -> Void) {
+        self.customerInfoManager.customerInfo(appUserID: self.appUserID,
+                                              fetchPolicy: fetchPolicy) { result in
+            completion(result.value, result.error?.asPurchasesError)
```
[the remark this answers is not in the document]
aboedo: good catch

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
```diff
@@ -415,3 +418,176 @@ class CustomerInfoManagerTests: TestCase {
...
+        self.customerInfoManager.cache(customerInfo: self.mockCustomerInfo, appUserID: Self.appUserID)
+
+        let result = try await self.customerInfoManager.customerInfo(appUserID: Self.appUserID,
+                                                                     fetchPolicy: .cachedOrFetched)
+
+        expect(result) == self.mockCustomerInfo
+        expect(self.mockBackend.invokedGetSubscriberDataCount) == 1
+    }
+
+    func testCustomerInfoCachedOrFetchedFetchesIfNoCache() async throws {
```
aboedo, line 508: the only thing I'd add is to have a case for "returns error if cache was stale and new cache can't be fetched" in case of network errors

aboedo (approved):
great work on this!

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
NachoSoto: Good idea, added for both `.cachedOrFetched` and `.notStaleCachedOrFetched`.

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -90,27 +90,67 @@ class CustomerInfoManager {
...
-    func customerInfo(appUserID: String, completion: ((Result<CustomerInfo, BackendError>) -> Void)?) {
-        let infoFromCache = cachedCustomerInfo(appUserID: appUserID)
-        var completionCalled = false
+    func customerInfo(
+        appUserID: String,
+        fetchPolicy: CacheFetchPolicy,
+        completion: ((Result<CustomerInfo, BackendError>) -> Void)?
+    ) {
+        switch fetchPolicy {
+        case .fromCacheOnly:
```
tonidero, line 99: I'm working on the android side of this, and one thing I'm wondering is, do we want to fetch new data even with this fetch policy (if stale)? I think that might be better so we update as early as possible, but I could also understand devs thinking that the cache will not be updated in this case... Though if that's the case we should make sure we handle all other possible places where the cache can be updated. cc @NachoSoto @aboedo

aboedo: The [docs for CachePolicy](https://revenuecat.github.io/purchases-ios-docs/4.6.1/documentation/revenuecat/cachefetchpolicy) might help illustrate the desired behavior here. 
For this specific case, if there's nothing in cache, we'd return an error right away no fetching. If there's something in cache, it gets returned even if stale. FWIW I think most apps will not want to use this behavior, and will opt for any of the others

aboedo: Happy to clarify further if needed!

tonidero: I agree that this won't be used much by devs so it's not that big of a deal. The docs don't specify whether this would initiate a fetch, and it doesn't either for some other policies, like the `cacheOrFetched` policy, it says: `Default behavior: returns the cached data if available (even if stale), or fetches up-to-date data.`, it says `or` but it can be an `and` if the cache is stale (I'm probably nitpicking here).

For now I will follow iOS behavior on this, but I'm more concerned about this other question: https://github.com/RevenueCat/purchases-ios/pull/1608#discussion_r895928743 @aboedo

aboedo: That is certainly valid feedback. We should update the docs to make it 100% unambiguous, i.e.: specify for each one: 
- If cache is available and valid, ...
- If cache is available and stale, ... 
- If cache is not available, ...
Where the second part specifies whether an error or value is returned, whether it'll trigger a fetch, and whether it'll wait for the fetch before returning or it'll trigger something in the delegate instead.

Maybe we could specify that in the Android implementation and copy the text back to this one?

Sources/Identity/CustomerInfoManager.swift
tonidero: Well, I can see your reasoning... I think this is ok as long as this option is not the default... But if we wanted to make this the default as was mentioned on this comment, we probably want to be more lenient and allow to use the stale data... I'm sure there are a ton of people that would use the apps while offline that would cause this to error out even if we had a cached value that the customer can potentially use.

NachoSoto: > Well, I can see your reasoning... I think this is ok as long as this option is not the default... But if we wanted to make this the default as was mentioned on this comment, we probably want to be more lenient and allow to use the stale data...

Totally agreed.

So we want this to be `notStaleCachedOrFetchedUnlessThatFailsInWhichCaseStaleCache`. I think this is pointing to the fact that really there's a lot of different combinations, and we can't offer them all. So maybe the right answer is to make sure this is documented correctly (could add a note to this policy).
To create that behavior users could do it themselves. I think it's best if we provide the primitives that users can compose around:
```swift
let customerInfo: CustomerInfo?

do {
  customerInfo = purchases.customerInfo(fetchPolicy: .notStaleCachedOrFetched)
} catch {
  customerInfo = purchases.customerInfo(fetchPolicy: .fromCacheOnly)
}
```

What do you guys think?

tonidero: Yeah, I think with the current options, users have a lot of flexibility, so as long as it's well documented, customers should be able to do what they need. However, as I mentioned, I wouldn't change the default to `notStaleCachedOrFetched` since I think that can potentially cause a bunch of issues for users.

We can talk about what the default behavior should be if we think the current behavior (`cachedOrFetched`) is causing issues

PR #1609, opened 2022-05-25
Split `PurchasesTests` step 1: extracted `BasePurchasesTests`

NachoSoto:
`PurchasesTests.swift` is nearly 3000 lines. This is the first step towards making separate smaller files like I did with `BackendTests.swift` in #1365.

PR #1610, opened 2022-05-25
Split `PurchasesTests` step 2: extracted `PurchasesConfiguringTests` and `PurchasesGetCustomerInfoTests`

NachoSoto:
Follow up to #1609.


Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
```diff
@@ -210,22 +212,19 @@ extension BasePurchasesTests {
...
             try! CustomerInfo(data: BasePurchasesTests.emptyCustomerInfoData)
         )
 
         override func getCustomerInfo(appUserID: String, completion: @escaping Backend.CustomerInfoResponseHandler) {
-            getSubscriberCallCount += 1
-            userID = appUserID
+            self.getSubscriberCallCount += 1
+            self.userID = appUserID
 
-            if !timeout {
```
NachoSoto, line 224: This property made it so that this method wouldn't invoke the callback, which made the tests not actually do anything.
Instead, I've updated the tests to ensure that when the callback is called, the data is correct.

aboedo (approved):

PR #1616, opened 2022-05-26
Fix fastlane changelog_path error and use 15.4 sim on CircleCI

joshdholtz:
### Motivation/Description

Another issue when trying to generate changelog 🙈

Also, the 15.2 simulator install was really slow so using the 15.4 simulator which is already installed on CircleCI.


fire-at-will (approved):
Super hyped for these improved CI times 👏 👏

PR #1624, opened 2022-05-27
[CF-601] Fix docs generation failure

aboedo:
Fixed a bug where we'd get an error after updating docs. 
The issue was that we're creating a temporary directory that gets automatically cleaned up, but we actually moved that directory before the cleanup code would happen.

So the cleanup code raised an error because there was nothing to clean up.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ GEM
   specs:
     CFPropertyList (3.0.5)
       rexml
-    activesupport (6.1.5)
+    activesupport (6.1.6)
       concurrent-ruby (~> 1.0, >= 1.0.2)
       i18n (>= 1.6, < 2)
       minitest (>= 5.1)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -3,7 +3,7 @@ GEM
   specs:
     CFPropertyList (3.0.5)
       rexml
-    activesupport (6.1.5)
```
aboedo, line 6: also updated gems because why not ¯\_(ツ)_/¯

fastlane/Fastfile
```diff
@@ -518,7 +518,7 @@ platform :ios do
               # and push the changes
               docs_destination_folder = "docs/#{version_number}"
               index_destination_path = "docs/index.html"
-              FileUtils.mv docs_generation_folder, docs_destination_folder
+              FileUtils.cp_r docs_generation_folder, docs_destination_folder
```
aboedo, line 521: this is the fix - the folder would get _moved_, so then when ruby goes to remove it from the tmp directory, it's not there anymore. 
Fixed by just doing a copy instead of a move. The copy will also get automatically cleaned up afterwards, since both the original and the copy are stored in temporary directories.

aboedo:
not sure what's going on with tests now, but it's unrelated to the contents of this PR

NachoSoto (approved):

PR #1625, opened 2022-05-30
Fix for not being able to read receipts on watchOS

patrickbusch:
### Checklist

### DIFF
--- a/Sources/Purchasing/ReceiptFetcher.swift
+++ b/Sources/Purchasing/ReceiptFetcher.swift
@@ -71,7 +71,7 @@ private extension ReceiptFetcher {
         let firstOSVersionWithoutBug: OperatingSystemVersion = OperatingSystemVersion(majorVersion: 7,
                                                                                       minorVersion: 0,
                                                                                       patchVersion: 0)
-        let isBelowFirstOSVersionWithoutBug = ProcessInfo.processInfo.isOperatingSystemAtLeast(firstOSVersionWithoutBug)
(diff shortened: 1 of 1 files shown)


Sources/Purchasing/ReceiptFetcher.swift
```diff
@@ -71,7 +71,7 @@ private extension ReceiptFetcher {
         let firstOSVersionWithoutBug: OperatingSystemVersion = OperatingSystemVersion(majorVersion: 7,
                                                                                       minorVersion: 0,
                                                                                       patchVersion: 0)
-        let isBelowFirstOSVersionWithoutBug = ProcessInfo.processInfo.isOperatingSystemAtLeast(firstOSVersionWithoutBug)
+        let isBelowFirstOSVersionWithoutBug = !ProcessInfo.processInfo.isOperatingSystemAtLeast(firstOSVersionWithoutBug)
```
aboedo, line 74: Great catch!! This was introduced during the Swift migration, it [is correct in v3](https://github.com/RevenueCat/purchases-ios/blob/3.14.2/Purchases/Purchasing/RCReceiptFetcher.m#L27). 

Thanks so much for reporting and fixing!! 🎉 👏 🏆 

Merging and we'll get it shipped soon!

aboedo (approved):

aboedo:
There's a lint issue now because the line goes over 120 chars, but we'll fix it as a separate thing

aboedo:
@patrickbusch thanks once more for your help on this one! the fix has been shipped as a part of 4.5.0 🎉

PR #1631, opened 2022-05-31
Removed all deprecation warnings in tests / SwiftAPITester

NachoSoto:
The project now builds with no warnings 🎉

PR #1635, opened 2022-05-31
Treat warnings as errors

NachoSoto:
Now that #1631 fixed all warnings (plus a few more fixed here that I missed), we can enforce this to make sure we don't add any more in the future.


taquitos:
I cancelled the iOS 15 tests, it was taking over 2 hours 🙀

NachoSoto:
They're passing now :)

taquitos (approved):
🌶️ 🐐

PR #1642, opened 2022-06-01
Add Swift Package Index badges for Swift version and platforms

joshdholtz:
### Motivation

I noticed while browsing Swift Package Index that there are badges for Swift Package Index that we can use for supported Swift versions and platforms.

These badges will automatically update with what SPI detects so we don't need to maintain these.

### Description

Adds two badgest into README

Totally fine if we don't want to merge this PR (if its too many badges) but figured I'd quickly make this PR to see if any interest 🤷‍♂️

Screen Shot 2022-05-31 at 7 38 46 PM


NachoSoto (approved):
We need more badges!

image

NachoSoto:
@joshdholtz wanna merge this? :)

PR #1645, opened 2022-06-02
Extracted `SandboxEnvironmentDetector`

NachoSoto:
### Motivation:

For [CSDK-53], `EntitlementInfos` will need access to the environment. This provides a lighter interface for that, as well as a simple implementation to inject that doesn't require the entire `SystemInfo`

[CSDK-53]: https://revenuecats.atlassian.net/browse/CSDK-53?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1647, opened 2022-06-02
`EntitlementInfos`: added `activeInAnyEnvironment` and `activeInCurrentEnvironment`

NachoSoto:
Fixes [CSDK-53].
Depends on #1645.

This is backwards compatible. `EntitlementInfos.active` maintains the previous behavior of returning `activeInAnyEnvironment`

[CSDK-53]: https://revenuecats.atlassian.net/browse/CSDK-53?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/EntitlementInfos.swift
+++ b/Sources/Purchasing/EntitlementInfos.swift
@@ -81,7 +81,8 @@ public extension EntitlementInfos {
     }
 
     /// Dictionary of active ``EntitlementInfo`` objects keyed by their identifiers.
-    /// - Note: these can be active on any environment.
+    /// - Note: When queried from the sandbox environment, it only returns entitlements active in sandbox.
+    /// When queried from production, this only returns entitlements active in production.
     /// - Seealso: ``activeInAnyEnvironment``
     @objc var activeInCurrentEnvironment: [String: EntitlementInfo] {
         return self.activeInAnyEnvironment.filter { [isSandbox = self.sandboxEnvironmentDetector.isSandbox] in
@@ -90,8 +91,7 @@ public extension EntitlementInfos {
     }
 
     /// Dictionary of active ``EntitlementInfo`` objects keyed by their identifiers.
-    /// - Note: this only returns entitlements active in sandbox when queried from the sandbox
-    /// or active in production when queried from production.
+    /// - Note: these can be active on any environment.
     /// - Seealso: ``activeInCurrentEnvironment``
     @objc var activeInAnyEnvironment: [String: EntitlementInfo] {
         return self.all.filter { $0.value.isActive }
--- a/Sources/Purchasing/EntitlementInfos.swift
+++ b/Sources/Purchasing/EntitlementInfos.swift
@@ -25,13 +25,6 @@ import Foundation
      */
     @objc public let all: [String: EntitlementInfo]
 
-    /**
-     Dictionary of active ``EntitlementInfo`` (`RCEntitlementInfo`) objects keyed by entitlement identifier.
-     */
-    @objc public var active: [String: EntitlementInfo] {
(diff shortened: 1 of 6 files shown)


Tests/UnitTests/Purchasing/EntitlementInfosTests.swift
```diff
@@ -1108,6 +1181,50 @@ private extension EntitlementInfosTests {
...
+    ) throws {
+        let entitlements = try self.extractEntitlements(identifier: identifier, inSandbox: sandbox)
+        let entitlement = try XCTUnwrap(entitlements[identifier])
+
+        expect(file: file, line: line, entitlement.identifier) == identifier
+        expect(file: file, line: line, entitlements.all.keys).to(contain([identifier]))
+        expect(file: file, line: line, entitlement.isActive) == true
+
+        expect(file: file, line: line, Set(entitlements.activeInCurrentEnvironment.keys))
+        == (expectedEntitlementActive ? [identifier] : [])
```
NachoSoto, line 1225: I've made sure these tests only pass if the implementation is
```swift
return self.activeInAnyEnvironment.filter { [isSandbox = self.sandboxEnvironmentDetector.isSandbox] in
  $0.value.isSandbox == isSandbox
}
```
And not with
```swift
return self.all.filter { [isSandbox = self.sandboxEnvironmentDetector.isSandbox] in
  $0.value.isSandbox == isSandbox
}
```

Sources/Purchasing/EntitlementInfos.swift
```diff
@@ -66,31 +65,67 @@ import Foundation
...
+    // MARK: -
+
+    private let sandboxEnvironmentDetector: SandboxEnvironmentDetector
+
+}
+
+public extension EntitlementInfos {
+
+    /// Dictionary of active ``EntitlementInfo`` objects keyed by their identifiers.
+    /// - Warning: this is equivalent to ``activeInAnyEnvironment``
```
NachoSoto, line 77: ⚠️ I've made this a warning so it's clear.

Sources/Purchasing/EntitlementInfos.swift
```diff
@@ -66,31 +65,67 @@ import Foundation
...
+    /// - Seealso: ``activeInAnyEnvironment``
+    @objc var activeInCurrentEnvironment: [String: EntitlementInfo] {
+        return self.activeInAnyEnvironment.filter { [isSandbox = self.sandboxEnvironmentDetector.isSandbox] in
+            $0.value.isSandbox == isSandbox
+        }
+    }
+
+    /// Dictionary of active ``EntitlementInfo`` objects keyed by their identifiers.
+    /// - Note: this only returns entitlements active in sandbox when queried from the sandbox
+    /// or active in production when queried from production.
```
taquitos, line 94: I was confused about how this relates to non-sandbox, maybe an update to the verbiage:
```
    /// - Note: When queried from the sandbox environment, it only returns entitlements active in sandbox.
    /// When queried from production, this will include both production and sandbox.
    
```

NachoSoto: Much better thanks.

NachoSoto: Actually yeah the note was completely wrong.

NachoSoto: Thanks for catching this, the docstrings were backwards!

NachoSoto: Updated.

taquitos (approved):
🎈🐐

PR #1648, opened 2022-06-02
`SubscriberAttribute`: converted into `struct`

NachoSoto:
### Motivation:

Initially I wanted to make `SubscriberAttribute.asDictionary` work using `Codable`. However, this isn't the right refactor because we need the actual raw types to be encoded inside `UserDefaults`.
At least this refactor provides several improvements:

### Changes:

- `SubscriberAttribute` now has value semantics
- ⚠️  `SubscriberAttribute.isSynced` used to be mutable, with reference semantics, which was dangerous
- `SubscriberAttribute: Equatable` implementation is now automatic
- Cleaned up `SubscriberAttribute.asDictionary` implementation, extracting keys into an `enum`
- Moved `Dictionary` -> `SubscriberAttribute` creation to the type itself
- ⚠️ `SubscriberAttribute.init(dictionary:)` returns an `Optional` instead of using force-unwrap, avoiding potential crashes
- Added more serialization / deserialization tests
- Removed `SubscriberAttributesMarshaller` and moved method into `SubscriberAttribute` extension
- Moved `SubscriberAttributeDict` into `SubscriberAttribute.Dictionary`


Sources/SubscriberAttributes/SubscriberAttribute.swift
```diff
@@ -14,70 +14,47 @@
 
 import Foundation
 
-class SubscriberAttribute {
+struct SubscriberAttribute {
```
NachoSoto, line 17: 👋🏻  Objective-C

Sources/SubscriberAttributes/SubscriberAttribute.swift
```diff
@@ -14,70 +14,47 @@
...
-
-    let setTime: Date
-    let key: String
-    let value: String
+    var setTime: Date
+    var key: String
+    @DefaultDecodable.EmptyString
+    var value: String
+    @DefaultDecodable.False
     var isSynced: Bool
```
NachoSoto, line 24: This had mutable reference semantics 🤮

Sources/Caching/DeviceCache.swift
```diff
@@ -303,17 +302,6 @@ class DeviceCache {
 
     // MARK: - Helper functions
 
-    static func newAttribute(dictionary: [String: Any]) -> SubscriberAttribute {
```
NachoSoto, line 306: This is `SubscriberAttribute.init?(dictionary:)` now.

Sources/FoundationExtensions/Date+Extensions.swift
```diff
@@ -17,7 +17,7 @@ import Foundation
 extension NSDate {
 
     func millisecondsSince1970AsUInt64() -> UInt64 {
-        return UInt64(self.timeIntervalSince1970 * 1000.0)
+        return (self as Date).millisecondsSince1970AsUInt64()
```
NachoSoto, line 20: This was duplicated.

Sources/Networking/Operations/Handling/SubscriberAttributesMarshaller.swift
```diff
@@ -1,27 +0,0 @@
...
-//
-//  SubscriberAttributesMarshaller.swift
-//
-//  Created by Joshua Liebowitz on 11/18/21.
-
-import Foundation
-
-enum SubscriberAttributesMarshaller {
-
-    // fixme: make `SubscriberAttributeDict` `Encodable` instead
```
NachoSoto, line 18: This was the goal of this PR, but as I explained in the description it didn't make sense. At least the existing implementation is cleaner now.

taquitos (approved):
Great change, only one update, then 🚢🐐

Sources/SubscriberAttributes/SubscriberAttribute.swift
```diff
@@ -14,75 +14,95 @@
...
-
-    static let keyKey = "key"
-    static let valueKey = "value"
-    static let setTimeKey = "setTime"
-    static let isSyncedKey = "isSynced"
-
-    let setTime: Date
-    let key: String
-    let value: String
+    var setTime: Date
```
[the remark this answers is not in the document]
NachoSoto: It doesn't matter though since this has value semantistas now :) so the properties are only mutable if the container is.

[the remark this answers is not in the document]
NachoSoto: Sure I can change it

PR #1651, opened 2022-06-03
Fix decoding issue caused by LossyDictionary for subscriptions and active entitlements

joshdholtz:
### Motivation

Fixes https://revenuecats.atlassian.net/browse/SDKONCALL-20

### Description

`LossyDictionary` was trying to convert the keys of product identifiers with a `JSONEncoder` and `JSONDecoder` encoding strategy (snake and camel case) when writing and reading from the cache.

These keys should not be using any encoding strategies (since the product identifiers are not getting mapped to any properties). Removing the `LossyDictionary` allows the product identifier keys to remain as they case they were intended to be.

This also fixes the test that would have caught this. The test for `CustomerInfo` had two subscriptions that expired in `2018` to `activeSubscriptions` was returning empty array for pre-cache and post-cache. There also a few more product identifier case formats added to this to try and catch any other errors.

### DIFF
--- a/Sources/Networking/Responses/CustomerInfoResponse.swift
+++ b/Sources/Networking/Responses/CustomerInfoResponse.swift
@@ -36,11 +36,11 @@ extension CustomerInfoResponse {
         var originalApplicationVersion: String?
         var originalPurchaseDate: Date?
         var firstSeen: Date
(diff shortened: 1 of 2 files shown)


aboedo (approved):

aboedo:
tested on purchaseTester as well as with the unit test changes from the PR

Issue #1650, opened 2022-06-03
On app launch entitlement identifier is lowercased string and not as defined in the web portal


cherpake:
When launching the app - entitlement identifier appears to be lowered case string "pro" and not as it's defined in the web portal "Pro". Then after fetching current customer info from the web - it does appear as expected "Pro".

The problem is that customer info (including entitlements) is encoded to json with snake case - which turns "Pro" into "pro".

```
(lldb) po customerInfo
<CustomerInfo:
originalApplicationVersion=35,
latestExpirationDate=Optional(2022-06-08 17:58:04 +0000),
activeEntitlements=["Pro": "<EntitlementInfo: \"\nidentifier=Pro,\nisActive=true,\nwillRenew=false,\nperiodType=PeriodType(rawValue: 0),\nlatestPurchaseDate=Optional(2018-02-23 09:40:18 +0000),\noriginalPurchaseDate=Optional(2018-02-23 09:40:18 +0000),\nexpirationDate=nil,\nstore=Store(rawValue: 0),\nproductIdentifier=com.cherpake.macrc.all,\nisSandbox=false,\nunsubscribeDetectedAt=nil,\nbillingIssueDetectedAt=nil,\nownershipType=PurchaseOwnershipType(rawValue: 0)\n>"],
activeSubscriptions=["com.cherpake.control.1msub": "expiresDate: Optional(2022-06-08 17:58:04 +0000)"],
nonSubscriptionTransactions=[<RCStoreTransaction: 0x281384ea0>, <RCStoreTransaction: 0x281385440>, <RCStoreTransaction: 0x2813853b0>],
requestDate=2022-06-02 18:48:28 +0000,
firstSeen=2022-05-08 17:57:30 +0000,
originalAppUserId=$RCAnonymousID:47027854016b41a9a201dcaa0d643615,
[...]

PR #1655, opened 2022-06-03
Added more failing tests for `CustomerInfo` encoding regression

NachoSoto:
See #1650.

I've added a failing subscription identifier in the fixture `CustomerInfo.json` so that the `CustomerInfoDecodingTests` would catch this.
Also updated the test introduced in #1651 to include one more case and improved the assertions.


taquitos (approved):
🎈🐐
image

PR #1657, opened 2022-06-06
Removed `LossyCollections`

NachoSoto:
See #1651.

Since that regression fix, we're no longer using it because it's not possible to make it decode dictionary keys without changing their case.
Therefore we can't ensure data isn't changed in the process.


Tests/UnitTests/Purchasing/TransactionsFactoryTests.swift
```diff
@@ -40,42 +40,6 @@ class TransactionsFactoryTests: TestCase {
         expect(list).to(beEmpty())
     }
 
-    func testBuildsCorrectlyEvenIfSomeTransactionsCantBeBuilt() throws {
```
NachoSoto, line 43: ⚠️ note that this test would no longer pass (same when decoding `CustomerInfo`.

taquitos (approved):
🎈🐐
image

PR #1658, opened 2022-06-06
`DefaultDecodable.EmptyDictionary`: added new test to verify partial decoding failure behavior

NachoSoto:
The new test checks the behavior that even if only part of the dictionary fails to decode, the whole dictionary becomes empty.

### DIFF
--- a/Tests/UnitTests/FoundationExtensions/DecoderExtensionTests.swift
+++ b/Tests/UnitTests/FoundationExtensions/DecoderExtensionTests.swift
@@ -260,14 +260,36 @@ class DecoderExtensionsDefaultDecodableTests: TestCase {
     }
 
     func testDecodesEmptyArrayForIncorrectType() throws {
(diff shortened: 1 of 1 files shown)


taquitos (approved):
🚢🐐 🥂

Tests/UnitTests/FoundationExtensions/DecoderExtensionTests.swift
```diff
@@ -260,14 +260,36 @@ class DecoderExtensionsDefaultDecodableTests: TestCase {
     }
 
     func testDecodesEmptyArrayForIncorrectType() throws {
-        let json = "{\"list\": \"this is not an array\"}"
+        let json = "{\"array\": \"this is not an array\"}"
```
NachoSoto, line 263: These tests were wrong cause this was the wrong key, so of course the test failed, since there was no data.

NachoSoto:
Failing tests are unrelated.

PR #1659, opened 2022-06-06
`CustomerInfoDecodingTests`: added test to check decoding invalid data behavior

NachoSoto:
See also #1658.
This updates the behavior of `CustomerInfoResponse` decoding so that having invalid content inside of one of the structures fails more loudly instead of getting ignored.

For that, `@DefaultValue` no longer ignores errors, and instead `@IgnoreDecodeErrors` must be used.
For some of the values in `CustomerInfoResponse` this makes sense (like `Store`), but not for things like subscriptions where we would be losing data.


NachoSoto:
I'm going to change this behavior since it's not what we want, so that the whole `CustomerInfo` fails to decode in this case.

taquitos (approved):
🎈🐐
image

NachoSoto:
This depends on #1657

PR #1561, opened 2022-05-11
Allow CircleCI to generate new test snapshots

joshdholtz:
Depends on https://github.com/RevenueCat/purchases-ios/pull/1589

### Motivation
Our snapshot testing sometimes needs regeneration and it requires regeneration of iOS 12. iOS 12 simulators only run on Intel machines and not all of us have them.

This PR will allow a CircleCI job to be manually triggered to regenerate snapshot tests

#### Step 1: Trigger CircleCI workflow
Run `bundle exec fastlane generate_snapshots`

#### Step 2: Run CircleCI workflow with generate snapshots
This will start a series of jobs on CircleCI that will have the `CIRCLECI_TESTS_GENERATE_SNAPSHOTS` set to `true` that will regenerate new/updated snapshot test assets.

Screen Shot 2022-05-12 at 9 02 43 PM

#### Step 3: Create pull requests

If there are any files changed, a new PR will be created against the base branch (set earlier in the `generate_snapshots` lane)

Example PR 👉  https://github.com/RevenueCat/purchases-ios/pull/1578

Screen Shot 2022-05-12 at 8 57 46 PM

PR #1662, opened 2022-06-06
`BackendPostReceiptDataTests`: added new test with `ProductRequestData`

NachoSoto:
Looking into [SDKONCALL-19], realized that there's no test that explicitly checks the `ProductRequestData` is sent along with the receipt data.

## TODO:
- [x] Add new snapshots for other iOS versions (cough #1561).

[SDKONCALL-19]: https://revenuecats.atlassian.net/browse/SDKONCALL-19?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Tests/UnitTests/Networking/Backend/BackendPostReceiptDataTests.swift
+++ b/Tests/UnitTests/Networking/Backend/BackendPostReceiptDataTests.swift
@@ -51,6 +51,35 @@ class BackendPostReceiptDataTests: BaseBackendTests {
         expect(completionCalled).toEventually(beTrue())
     }
 
+    func testPostsReceiptDataWithProductDataCorrectly() throws {
+        let path: HTTPRequest.Path = .postReceiptData
+
+        httpClient.mock(
+            requestPath: path,
+            response: .init(statusCode: .success, response: Self.validCustomerResponse)
+        )
+
+        var completionCalled = false
+
+        let isRestore = false
+        let observerMode = true
+        let productData: ProductRequestData = .createMockProductData(currencyCode: "USD")
+
+        backend.post(receiptData: Self.receiptData,
+                     appUserID: Self.userID,
+                     isRestore: isRestore,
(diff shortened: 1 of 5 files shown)


Tests/UnitTests/Networking/Backend/BackendPostReceiptDataTests.swift
```diff
@@ -51,6 +51,35 @@ class BackendPostReceiptDataTests: BaseBackendTests {
...
+                     productData: productData,
+                     presentedOfferingIdentifier: nil,
+                     observerMode: observerMode,
+                     subscriberAttributes: nil,
+                     completion: { _ in
+            completionCalled = true
+        })
+
+        expect(self.httpClient.calls).toEventually(haveCount(1))
+        expect(completionCalled).toEventually(beTrue())
```
taquitos, line 80: I think this can be `expect(completionCalled) == true` since it's waiting for `expect(self.httpClient.calls).toEventually(haveCount(1))` anyway

taquitos (approved):
🏎🐐
download-48

Tests/UnitTests/Networking/Backend/BackendPostReceiptDataTests.swift
NachoSoto: Yup 👍🏻

NachoSoto: I think this order makes more sense too:
```swift
expect(completionCalled).toEventually(beTrue())
expect(self.httpClient.calls).to(haveCount(1))
```

taquitos: Oh, yeah, good point

PR #983, opened 2021-11-22
Refactored `DeviceCache` to improve thread safety

NachoSoto:
Fixes #739.

### Changes:
- Created `Atomic`, based on `Lock`, to abstract atomic access and modifications to values.
- Simplified `InMemoryCachedObject` using `Atomic` (tests still pass!).
- Created `SynchronizedUserDefaults` using `Atomic`, to enforce atomic access to `UserDefaults`.
- Refactored `DeviceCache` using `Atomic` to enforce all access to `UserDefaults` are synchronized:
  - `DeviceCache` no longer stores a `UserDefaults` instance (only `SynchronizedUserDefaults`), so we guarantee we don't do unsynchronized modifications.
  - The old `threadUnsafe_` methods are now `static`, which ensures they don't access any local state. Instead, `UserDefaults` is injected, and `DeviceCache` calls these methods using `SynchronizedUserDefaults.perform`
  - The combination of both of these things eliminates potential for programmer error within the class.
  - There's still potential for what's described in #739, that a _user_ of `DeviceCache` reads data, and THEN writes it again, without synchronizing both operations. This PR doesn't address that, but I haven't found any examples of that being an issue, and if we do, we should move both operations into a single method in `DeviceCache` that can enforce the semantics.

PR #1664, opened 2022-06-07
`PurchasesOrchestrator`: replaced manual `Lock` with `Atomic`

NachoSoto:
See #983. This is less error-prone as it's impossible to have deadlocks, or to access the data without locking (like we were doing in some cases).
Also, this approach has 2 separate locks for each property, so unrelated accesses won't lock each other now.


Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -509,12 +509,10 @@ extension PurchasesOrchestrator: StoreKitWrapperDelegate {
         guard let delegate = delegate else { return false }
 
         let storeProduct = StoreProduct(sk1Product: product)
-        lock.lock()
         delegate.readyForPromotedProduct(storeProduct) { completion in
-            self.purchaseCompleteCallbacksByProductID[product.productIdentifier] = completion
+            self.purchaseCompleteCallbacksByProductID.modify { $0[product.productIdentifier] = completion }
             storeKitWrapper.add(payment)
```
taquitos, line 514: I kinda think it makes sense to add this in to the `.modify` block since that guarantees the same behavior we had with the lock.

taquitos (approved):
I *THINK* there's a change here you should make, but let me know what you think.

Sources/Purchasing/PurchasesOrchestrator.swift
NachoSoto: The thing is this is an `@escaping` block, so the outside `lock` was likely getting unlocked before this was even modified. Does that make sense?

So the access to `purchaseCompleteCallbacksByProductID` wasn't protected. I do see that maybe the intent is that the whole thing was, and therefore maybe `storeKitWrapper.add(payment)` could be inside now. But considering that was missed, and therefore this is the same behavior, maybe this makes more sense?

taquitos: Ahhh, yeah. Makes sense.

PR #1665, opened 2022-06-07
`PurchasesOrchestrator.handleDeferredTransaction`: check `NSError.domain` too

NachoSoto:
Comparing only error codes can lead to collisions with a completely different error.
Also refactored this check into an existing extension.


taquitos (approved):
🎈🐐

PR #1666, opened 2022-06-07
`PurchasesOrchestrator`: replaced calls to `syncPurchases` with posting receipt for an individual product during SK2 purchases

NachoSoto:
Fixes [SDKONCALL-19].
SK2 purchases were always calling `syncPurchases`, which called `Backend.post(receiptData:...)` with no `productData`, which meant that after every SK2 purchase we never posted the price.

This refactor makes a lot of methods in `PurchasesOrchestrator` use `StoreTransaction` so they can be used during SK2 purchases as well. Specifically, `handlePurchasedTransaction` is now called after both SK1 and SK2 purchases.

### Other changes:

- `StoreKit2TransactionListenerDelegate` no longer needs to return `CustomerInfo`
- `StoreKit2TransactionListenerDelegate.handle(purchaseResult:)` no longer notifies the delegate (only done by `StoreKit.Transaction.updates`), which was leading to duplicate `syncPurchases` calls.
- If the purchase is cancelled, `syncPurchases` isn't called. Instead, the `CustomerInfo` is fetched from `CustomerInfoManager`

[SDKONCALL-19]: https://revenuecats.atlassian.net/browse/SDKONCALL-19?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/PurchasesOrchestrator.swift
@@ -417,7 +417,8 @@ class PurchasesOrchestrator {
         let customerInfo: CustomerInfo
 
         if let transaction = transaction {
-            customerInfo = try await self.handlePurchasedTransaction(transaction)
+            customerInfo = try await self.handlePurchasedTransaction(transaction,
+                                                                     refreshPolicy: .always)
         } else {
             customerInfo = try await self.customerInfoManager.customerInfo(appUserID: self.appUserID,
                                                                            fetchPolicy: .cachedOrFetched)
@@ -550,8 +551,9 @@ extension PurchasesOrchestrator: StoreKitWrapperDelegate {
 private extension PurchasesOrchestrator {
 
     func handlePurchasedTransaction(_ transaction: StoreTransaction,
+                                    refreshPolicy: ReceiptRefreshPolicy = .onlyIfEmpty,
                                     storefront: StorefrontType?) {
-        self.receiptFetcher.receiptData(refreshPolicy: .onlyIfEmpty) { receiptData in
+        self.receiptFetcher.receiptData(refreshPolicy: refreshPolicy) { receiptData in
             if let receiptData = receiptData,
                !receiptData.isEmpty {
                 self.fetchProductsAndPostReceipt(withTransaction: transaction,
@@ -887,7 +889,10 @@ private extension Error {
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
 extension PurchasesOrchestrator {
 
-    private func handlePurchasedTransaction(_ transaction: StoreTransaction) async throws -> CustomerInfo {
+    private func handlePurchasedTransaction(
+        _ transaction: StoreTransaction,
+        refreshPolicy: ReceiptRefreshPolicy
+    ) async throws -> CustomerInfo {
         let storefront = await Storefront.currentStorefront
 
         return try await withCheckedThrowingContinuation { continuation in
@@ -898,7 +903,9 @@ extension PurchasesOrchestrator {
                 }
             )
 
-            self.handlePurchasedTransaction(transaction, storefront: storefront)
+            self.handlePurchasedTransaction(transaction,
+                                            refreshPolicy: refreshPolicy,
+                                            storefront: storefront)
         }
     }
 
--- a/Sources/Purchasing/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/PurchasesOrchestrator.swift
@@ -405,6 +405,8 @@ class PurchasesOrchestrator {
             throw ErrorUtils.purchasesError(withStoreKitError: error)
         }
 
+        // `userCancelled` above comes from `StoreKitError.userCancelled`.
+        // This detects if `Product.PurchaseResult.userCancelled` is true.
         let (userCancelled, sk2Transaction) = try await self.storeKit2TransactionListener
             .handle(purchaseResult: result)
         let transaction = sk2Transaction.map(StoreTransaction.init(sk2Transaction:))
@@ -414,6 +416,8 @@ class PurchasesOrchestrator {
             customerInfo = try await self.handlePurchasedTransaction(transaction,
                                                                      refreshPolicy: .always)
         } else {
+            // `transaction` would be `nil` for `Product.PurchaseResult.pending` and
+            // `Product.PurchaseResult.userCancelled`.
             customerInfo = try await self.customerInfoManager.customerInfo(appUserID: self.appUserID,
                                                                            fetchPolicy: .cachedOrFetched)
         }
@@ -704,7 +708,7 @@ private extension PurchasesOrchestrator {
     func handleReceiptPost(withTransaction transaction: StoreTransaction,
                            result: Result<CustomerInfo, BackendError>,
                            subscriberAttributes: SubscriberAttributeDict?) {
-        func finishTransaction() {
+        func finishTransactionIfNeeded() {
             if self.finishTransactions, let sk1Transaction = transaction.sk1Transaction {
                 self.storeKitWrapper.finishTransaction(sk1Transaction)
             }
@@ -725,7 +729,7 @@ private extension PurchasesOrchestrator {
                 self.customerInfoManager.cache(customerInfo: customerInfo, appUserID: appUserID)
                 completion?(transaction, customerInfo, nil, false)
 
-                finishTransaction()
+                finishTransactionIfNeeded()
 
             case let .failure(error):
                 let purchasesError = error.asPurchasesError
@@ -733,7 +737,7 @@ private extension PurchasesOrchestrator {
                 completion?(transaction, nil, purchasesError, false)
 
                 if finishable {
-                    finishTransaction()
+                    finishTransactionIfNeeded()
                 }
             }
         }
--- a/Sources/Purchasing/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/PurchasesOrchestrator.swift
@@ -493,7 +493,7 @@ extension PurchasesOrchestrator: StoreKitWrapperDelegate {
         switch transaction.transactionState {
         case .restored, // for observer mode
              .purchased:
-            self.handlePurchasedTransaction(storeTransaction)
+            self.handlePurchasedTransaction(storeTransaction, storefront: storeKitWrapper.currentStorefront)
         case .purchasing:
             break
         case .failed:
@@ -549,11 +549,14 @@ extension PurchasesOrchestrator: StoreKitWrapperDelegate {
 // MARK: Transaction state updates.
 private extension PurchasesOrchestrator {
 
-    func handlePurchasedTransaction(_ transaction: StoreTransaction) {
+    func handlePurchasedTransaction(_ transaction: StoreTransaction,
+                                    storefront: StorefrontType?) {
         self.receiptFetcher.receiptData(refreshPolicy: .onlyIfEmpty) { receiptData in
             if let receiptData = receiptData,
                !receiptData.isEmpty {
-                self.fetchProductsAndPostReceipt(withTransaction: transaction, receiptData: receiptData)
+                self.fetchProductsAndPostReceipt(withTransaction: transaction,
+                                                 receiptData: receiptData,
+                                                 storefront: storefront)
             } else {
                 self.handleReceiptPost(withTransaction: transaction,
                                        result: .failure(.missingReceiptFile()),
@@ -647,10 +650,13 @@ private extension PurchasesOrchestrator {
         }
     }
 
-    func fetchProductsAndPostReceipt(withTransaction transaction: StoreTransaction, receiptData: Data) {
+    func fetchProductsAndPostReceipt(
+        withTransaction transaction: StoreTransaction,
+        receiptData: Data,
+        storefront: StorefrontType?
+    ) {
         if let productIdentifier = transaction.productIdentifier.notEmpty {
             self.products(withIdentifiers: [productIdentifier]) { products in
-                let storefront = self.storeKitWrapper.currentStorefront
                 self.postReceipt(withTransaction: transaction,
                                  receiptData: receiptData,
                                  products: Set(products),
@@ -882,6 +888,8 @@ private extension Error {
 extension PurchasesOrchestrator {
 
     private func handlePurchasedTransaction(_ transaction: StoreTransaction) async throws -> CustomerInfo {
+        let storefront = await Storefront.currentStorefront
+
         return try await withCheckedThrowingContinuation { continuation in
             self.addPurchaseCompletedCallback(
                 productIdentifier: transaction.productIdentifier,
@@ -890,7 +898,7 @@ extension PurchasesOrchestrator {
                 }
             )
 
-            self.handlePurchasedTransaction(transaction)
+            self.handlePurchasedTransaction(transaction, storefront: storefront)
         }
     }
 
--- a/Sources/Purchasing/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/PurchasesOrchestrator.swift
@@ -320,7 +320,7 @@ class PurchasesOrchestrator {
 
         }
 
-        productsManager.cacheProduct(sk1Product)
+        self.productsManager.cacheProduct(sk1Product)
 
         let addPayment: Bool = self.purchaseCompleteCallbacksByProductID.modify { callbacks in
             guard callbacks[productIdentifier] == nil else {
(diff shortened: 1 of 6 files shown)


NachoSoto:
Unit tests pass, but looking into integration tests now (glad to have those since this is a critical change).

NachoSoto:
This simple test fails 🎉 looking into why

```swift
func testCanMakePurchase() async throws {
  try await self.purchaseMonthlyOffering()
}
```

NachoSoto:
Integration tests fixed :) This is ready

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -855,9 +884,31 @@ private extension Error {
 
 }
 
+// MARK: - Async extensions
+
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
 extension PurchasesOrchestrator {
 
-    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    private func handlePurchasedTransaction(
```
NachoSoto, line 892: This is an `async` wrapper on top of `self.handlePurchasedTransaction`.

Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
```diff
@@ -96,19 +100,18 @@ private extension StoreKit2TransactionListener {
             )
 
         case .verified(let verifiedTransaction):
-            let customerInfo = try await self.finish(transaction: verifiedTransaction)
+            await self.finish(transaction: verifiedTransaction)
 
-            return (verifiedTransaction, customerInfo)
+            if notifyDelegate, let delegate = self.delegate {
+                _ = try await delegate.transactionsUpdated()
+            }
```
NachoSoto, line 107: This avoids calling the delegate during purchases (leading to a duplicate `syncPurchases` call). Now this is only called for transactions that come from `StoreKit.Transaction.updates`

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -332,24 +335,40 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
     }
 
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
-    func testPurchaseSK2PackageReturnsMissingReceiptErrorIfSendReceiptFailed() async throws {
+    func testPurchaseSK2PackageRetrunsCustomerInfoForFailedTransaction() async throws {
```
NachoSoto, line 338: This checks that `CustomerInfo` is returned even when the purchase failed.

Tests/UnitTests/Purchasing/Purchases/PurchasesPurchasingTests.swift
```diff
@@ -574,12 +574,12 @@ class PurchasesPurchasingTests: BasePurchasesTests {
...
         expect(receivedError).toEventuallyNot(beNil())
-        expect(receivedError?.domain).toEventually(equal(RCPurchasesErrorCodeDomain))
-        expect(receivedError?.code).toEventually(equal(ErrorCode.operationAlreadyInProgressForProductError.rawValue))
-        expect(self.storeKitWrapper.addPaymentCallCount).to(equal(1))
-        expect(receivedUserCancelled).toEventually(beFalse())
+        expect(receivedInfo).to(beNil())
+        expect(receivedError?.domain) == RCPurchasesErrorCodeDomain
+        expect(receivedError?.code) == ErrorCode.operationAlreadyInProgressForProductError.rawValue
+        expect(self.storeKitWrapper.addPaymentCallCount) == 1
+        expect(receivedUserCancelled) == false
```
NachoSoto, line 582: NFC, just improving expectations to a single "eventually".

taquitos (approved):
Nice fix, it's slightly confusing to follow but I think that's to be expected given we're dealing with SK1/SK2, and whether or not we're calling the delegate/syncing/posting receipts. I guess this is why people use RevenueCat 😄

NachoSoto:
Heads up this depends on #1664 and #1665.

aboedo:
just to clarify, in the PR description it says 

> SK2 purchases were always calling syncPurchases, which called Backend.post(receiptData:...) with no productData, which meant that after every SK2 purchase we never posted the price.

But my understanding is that we were sending two posts, right? one with and one without the product details?

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -403,19 +403,26 @@ class PurchasesOrchestrator {
...
+                customerInfo: try await self.customerInfoManager.customerInfo(appUserID: self.appUserID,
+                                                                              fetchPolicy: .cachedOrFetched),
                 userCancelled: true
             )
         } catch {
             throw ErrorUtils.purchasesError(withStoreKitError: error)
         }
 
-        let (userCancelled, customerInfoIfSynced, sk2Transaction) = try await self.storeKit2TransactionListener
+        let (userCancelled, sk2Transaction) = try await self.storeKit2TransactionListener
```
aboedo, line 414: don't we already know whether this is a cancellation from the lines before this?

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -541,11 +550,15 @@ extension PurchasesOrchestrator: StoreKitWrapperDelegate {
 // MARK: Transaction state updates.
 private extension PurchasesOrchestrator {
 
-    func handlePurchasedTransaction(_ transaction: SKPaymentTransaction) {
-        receiptFetcher.receiptData(refreshPolicy: .onlyIfEmpty) { receiptData in
+    func handlePurchasedTransaction(_ transaction: StoreTransaction,
+                                    refreshPolicy: ReceiptRefreshPolicy = .onlyIfEmpty,
```
aboedo, line 554: 🤔 why do we have `onlyIfEmpty` in this case? If we're calling this method, we _know_ that there was a purchase, right? I feel like we should always refresh in that case, there's a good chance that the receipt isn't up to date otherwise

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -555,11 +568,13 @@ private extension PurchasesOrchestrator {
     }
 
     func handleFailedTransaction(_ transaction: SKPaymentTransaction) {
+        let storeTransaction = StoreTransaction(sk1Transaction: transaction)
+
         if let error = transaction.error,
-           let completion = getAndRemovePurchaseCompletedCallback(forTransaction: transaction) {
+           let completion = self.getAndRemovePurchaseCompletedCallback(forTransaction: storeTransaction) {
```
aboedo, line 574: on a separate note, this brings up an issue that @vegaro brought up to me yesterday (some of his thoughts are captured [here](https://revenuecats.atlassian.net/browse/CSDK-99)): 
StoreTransaction currently represents Storekit transactions + backend parsed transactions. 
But... they're really different. 

For starters, a `BackendParsedTransaction` would absolutely not make sense here, meaning that we're violating Liskov principle, right? 

We can discuss it further on the ticket, but I did want to bring up that it feels like our abstraction there might get problematic over time (it already has some inconsistencies in hybrids)

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -610,25 +627,38 @@ extension PurchasesOrchestrator: StoreKit2StorefrontListenerDelegate {
...
-    func fetchProductsAndPostReceipt(withTransaction transaction: SKPaymentTransaction, receiptData: Data) {
-        if let productIdentifier = transaction.productIdentifier {
+    func fetchProductsAndPostReceipt(
+        withTransaction transaction: StoreTransaction,
+        receiptData: Data,
+        storefront: StorefrontType?
+    ) {
+        if let productIdentifier = transaction.productIdentifier.notEmpty {
             self.products(withIdentifiers: [productIdentifier]) { products in
-                let storefront = self.storeKitWrapper.currentStorefront
```
aboedo, line 631: If the `productIdentifier` is empty, it means we got a bad `SKTransaction`. 
Current behavior is to just not send it to the backend. 

But... 
At this point, the user got charged. They won't have access, and they'll get an error. That's a terrible experience. Sure, it's because of a bug in StoreKit, but it's a bug in the price, and the only value of fetching the product here is to send the prices to our backend. So the user gets a bad experience but we're not preventing any issues on their side, this is just for accounting purposes. 

The user should come first. If they made a purchase, and it went through, we should prioritize granting them access. Getting the products comes second. 

This is illustrated in the current flow by the fact that we don't check whether we actually could fetch the product - there's no guarantee that we will (the request could fail) but we don't care, we just send the receipt. 

We should do the same for the case where the SKTransaction is broken. The receipt should still be fine, so even if the transaction is broken, we can send the transaction to the backend, and get the customerInfo, and preserve user experience, in exchange for slightly worse analytics when `SKTransaction`s are broken. 

It's not like the customer would be able to fix this by retrying either, they'll already have purchased the product

aboedo: So I'd suggest to: 
- Log when the transaction is bad, that'll help us debug issues with prices
- Always refresh and send the receipt, even if the transaction was bad

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -660,23 +690,29 @@ private extension PurchasesOrchestrator {
...
     }
 
-    func handleReceiptPost(withTransaction transaction: SKPaymentTransaction,
+    func handleReceiptPost(withTransaction transaction: StoreTransaction,
                            result: Result<CustomerInfo, BackendError>,
                            subscriberAttributes: SubscriberAttributeDict?) {
-        operationDispatcher.dispatchOnMainThread {
+        func finishTransaction() {
+            if self.finishTransactions, let sk1Transaction = transaction.sk1Transaction {
+                self.storeKitWrapper.finishTransaction(sk1Transaction)
```
aboedo, line 711: the RTTI here feels a little weird... why are we special casing SK1? [SK2 transactions have a finish](https://developer.apple.com/documentation/storekit/transaction) method too

aboedo (changes requested):
still reviewing but I left a few comments

NachoSoto:
Thanks for the amazingly thorough review @aboedo! This is a change in a very critical part of the code, so I really appreciate the close look.

NachoSoto:
> But my understanding is that we were sending two posts, right? one with and one without the product details?

@aboedo correct, we were posting receipts twice. But no, none of them contained the product details, because in both cases they came from calls to `syncPurchases`, which [never passes `productData`](https://github.com/RevenueCat/purchases-ios/blob/4.5.2/Sources/Purchasing/PurchasesOrchestrator.swift#L772).

Sources/Purchasing/PurchasesOrchestrator.swift
NachoSoto: The line before catches `StoreKitError.userCancelled`. This comes from `Product.PurchaseResult.userCancelled`. It's unclear when StoreKit would throw one versus returning the other, but we need to handle both.

Sources/Purchasing/PurchasesOrchestrator.swift
NachoSoto: I added that default value to keep the existing behavior to the previous callers.
Looks like there is only one place: `storeKitWrapper(:updatedTransaction:)`. I suppose it makes sense for that one to use `.always` as well?

Sources/Purchasing/PurchasesOrchestrator.swift
NachoSoto: That one is definitely my fault, I didn't realize that those represented 2 different things at the time, and because they had basically the same interface I assumed they could be represented by the same type :/

Sources/Purchasing/PurchasesOrchestrator.swift
NachoSoto: For context, this PR doesn't change the existing behavior. The reason I changed this to `notEmpty` is because of [this](https://github.com/RevenueCat/purchases-ios/blob/4.5.2/Sources/Purchasing/StoreKitAbstractions/SK1StoreTransaction.swift#L21) (`StoreTransactionType.productIdentifier` is not `Optional`) whereas previously the `nil` check did that.

> Log when the transaction is bad, that'll help us debug issues with prices

That's always done by [`SKPaymentTransaction.productIdentifier`](https://github.com/RevenueCat/purchases-ios/blob/4.5.2/Sources/Purchasing/StoreKitAbstractions/SK1StoreTransaction.swift#L43).

> Always refresh and send the receipt, even if the transaction was bad

I think that makes sense. So to make sure I understand, basically change the `else` to:
```swift
self.postReceipt(withTransaction: transaction,
                 receiptData: receiptData,
                 products: [], // No product data, but still post the receipt
                 storefront: storefront)
```

I can add unit tests for this too.

Sources/Purchasing/PurchasesOrchestrator.swift
NachoSoto: Yeah I agree this is messy. The reason is SK2 transactions are finished by [`StoreKit2TransactionListener`](https://github.com/RevenueCat/purchases-ios/blob/4.5.2/Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift#L107), so I wanted to keep the existing behavior here. Even though this method uses any `StoreTransaction`, I didn't want to call that twice on SK2 transactions :/

But looking at this now, maybe we can remove the `finish` call in `StoreKit2TransactionListener` and call it here for consistency?

Sources/Purchasing/PurchasesOrchestrator.swift
aboedo: I think so - if anything is coming in from StoreKit it should be safe to assume that there's an apple ID logged into the device

Sources/Purchasing/PurchasesOrchestrator.swift
NachoSoto: Kinda scary that this doesn't make any test fail:
Screen Shot 2022-06-09 at 11 30 47

NachoSoto: Huh I see, there's just no test coverage for it.
It is an edge case after all, but I'll add a test.

NachoSoto: Turns out this breaks during tests, because we rely on product identifiers for the callbacks keys...

NachoSoto: I'm gonna leave this for a separate PR to deal with this problem.

NachoSoto: Filed [CSDK-119]

Sources/Purchasing/PurchasesOrchestrator.swift
NachoSoto: Filed [CSDK-120]

Sources/Purchasing/PurchasesOrchestrator.swift
NachoSoto: I also added a failing test: https://github.com/RevenueCat/purchases-ios/pull/1680

Sources/Purchasing/PurchasesOrchestrator.swift
aboedo: I think that's probably a good call, although it doesn't need to be a part of this PR

NachoSoto: Okay let me make a Jira for it then.

NachoSoto: https://revenuecats.atlassian.net/browse/CSDK-122

This is ready for a final look then :)

aboedo (approved):
looks great! I left a couple of minor comments

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -403,19 +403,26 @@ class PurchasesOrchestrator {
...
             .handle(purchaseResult: result)
         let transaction = sk2Transaction.map(StoreTransaction.init(sk2Transaction:))
+        let customerInfo: CustomerInfo
 
-        let customerInfo = try await customerInfoIfSynced
-        ??? (await self.syncPurchases(receiptRefreshPolicy: .always, isRestore: false))
+        if let transaction = transaction {
+            customerInfo = try await self.handlePurchasedTransaction(transaction,
+                                                                     refreshPolicy: .always)
+        } else {
```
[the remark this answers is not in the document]
NachoSoto: Yup, added.

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -660,23 +688,29 @@ private extension PurchasesOrchestrator {
...
                                    subscriberAttributes: unsyncedAttributes)
         }
     }
 
-    func handleReceiptPost(withTransaction transaction: SKPaymentTransaction,
+    func handleReceiptPost(withTransaction transaction: StoreTransaction,
                            result: Result<CustomerInfo, BackendError>,
                            subscriberAttributes: SubscriberAttributeDict?) {
-        operationDispatcher.dispatchOnMainThread {
+        func finishTransaction() {
```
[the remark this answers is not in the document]
NachoSoto: Good point, changed.

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -295,6 +297,7 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
         _ = try await orchestrator.purchase(sk2Product: product, promotionalOffer: nil)
 
         expect(self.backend.invokedPostReceiptDataCount) == 1
+        expect(self.backend.invokedPostReceiptDataParameters?.productData).toNot(beNil())
```
NachoSoto, line 300: I added these checks in several of the tests too.

Sources/Purchasing/PurchasesOrchestrator.swift
NachoSoto: Done: https://github.com/RevenueCat/purchases-ios/pull/1703

PR #1668, opened 2022-06-07
`CodableStrings.decoding_error`: added underlying error information

NachoSoto:
While debugging `CustomerInfo` decoding errors for example, this is the existing error logged in the console:
```
ERROR: 😿‼️ Couldn't decode data from json. Error:
There was a problem related to the customer info.
```

With this change:
```
ERROR: 😿‼️ Couldn't decode data from json.
Error: There was a problem related to the customer info..
Underlying error: Swift.DecodingError.typeMismatch(Swift.Dictionary<Swift.String, RevenueCat.CustomerInfoResponse.Subscription>, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "subscriber", intValue: nil), CodingKeys(stringValue: "subscriptions", intValue: nil)], debugDescription: "Expected to decode Dictionary<String, Subscription> but found an array instead.", underlyingError: nil))
```

### DIFF
--- a/Sources/Logging/Strings/CodableStrings.swift
+++ b/Sources/Logging/Strings/CodableStrings.swift
@@ -44,7 +44,14 @@ extension CodableStrings: CustomStringConvertible {
         case let .encoding_error(error):
             return "Couldn't encode data into json. Error:\n\(error.localizedDescription)"
         case let .decoding_error(error):
-            return "Couldn't decode data from json. Error:\n\(error.localizedDescription)"
+            let underlyingErrorMessage: String
+            if let underlyingError = (error as NSError).userInfo[NSUnderlyingErrorKey] as? NSError {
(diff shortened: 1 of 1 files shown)


taquitos (approved):
🎈🐐
image

NachoSoto:
Flaky tests.
Something I was seeing today, sometimes tests find a pre-existing receipt in `Bundle.appStoreReceiptURL`. I wonder if we should delete that if it exists at the beginning of every test?

taquitos:
> Flaky tests. Something I was seeing today, sometimes tests find a pre-existing receipt in `Bundle.appStoreReceiptURL`. I wonder if we should delete that if it exists at the beginning of every test?

I think cleaning up any data that could persist between launches makes sense. It's a bummer this doesn't happen all the time, but I'm not shocked.

NachoSoto:
#1671

PR #1672, opened 2022-06-07
`Result.init(value:error:)`: avoid creating error if value is provided

NachoSoto:
Our implementation of `ErrorUtils.error(with:message:underlyingError:extraUserInfo:)` always calls `ErrorUtils.logErrorIfNeeded`, this means that even though the error wasn't actually used, it was being logged as if it had occurred.
By using `@autoclosure` we avoid this and instead create it lazily.

This incorrect logging lead to [CSDK-112], which won't be confusing anymore.

[CSDK-112]: https://revenuecats.atlassian.net/browse/CSDK-112?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/FoundationExtensions/Result+Extensions.swift
+++ b/Sources/FoundationExtensions/Result+Extensions.swift
@@ -15,10 +15,10 @@ extension Result {
 
     /// Creates a `Result` from either a value or an error.
     /// This is useful for converting from old Objective-C closures to new APIs.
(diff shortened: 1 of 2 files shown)


NachoSoto:
Thanks @vegaro for finding this "fun" bug 🎉

taquitos (approved):
🎈🐐
download-39

PR #1680, opened 2022-06-09
[WIP] Fix behavior dealing with `nil` `SKPaymentTransaction.productIdentifier` during purchase

NachoSoto:
See [CSDK-119]


Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -227,6 +227,33 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
         expect(self.backend.invokedPostReceiptDataCount) == 1
     }
 
+    func testPurchaseSK1PackageWithNoProductIdentifierPostsReceipt() async throws {
```
NachoSoto, line 230: This currently fails.

Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -302,9 +302,8 @@ class PurchasesOrchestrator {
                   package: Package?,
                   completion: @escaping PurchaseCompletedBlock) {
         guard let productIdentifier = sk1Product.extractProductIdentifier(withPayment: payment) else {
-            Logger.error(Strings.purchase.could_not_purchase_product_id_not_found)
```
NachoSoto, line 305: Creating the error already logs it, so this was duplicated. The new code relies on the `String` without hardcoding another error message.

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -237,6 +237,40 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
         expect(self.backend.invokedPostReceiptDataParameters?.productData).toNot(beNil())
     }
 
+    func testPurchaseSK1PackageWithNoProductIdentifierDoesNotPostReceipt() async throws {
```
NachoSoto, line 240: ⚠️ We didn't have any test coverage for this scenario, so we didn't know that the method in `StoreKitWorkarounds` actually made the whole purchase hang.

NachoSoto:
This is ready for review.

Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
```diff
@@ -46,32 +46,24 @@ extension SKPaymentTransaction {
...
+extension SKPayment {
 
     /// Attempts to find a non-nil `productIdentifier`.
     ///
-    /// Although both `SK1Product.productIdentifier` and `SKPayment.productIdentifier`
-    /// are supposed to be non-nil, we've seen instances where this is not true.
-    /// so we cast into optionals in order to check nullability, and try to fall back if possible.
-    func extractProductIdentifier(withPayment payment: SKPayment,
-                                  fileName: String = #fileID,
+    /// Although `SKPayment.productIdentifier` is supposed to be non-nil,
```
taquitos, line 53: Suggested edit:
```
    /// Although `SKPayment.productIdentifier` is supposed to be non-nil, we've seen instances where this is not true.
    /// To handle this case, we cast `productIdentifier` to `Optional` in order to check nullability.
    
```

taquitos (approved):
Slight wording suggestion, but 👍

Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
NachoSoto: Thanks much better.

aboedo (changes requested):
I left a question, lmk if I'm reading it wrong

NachoSoto:
What's the question? 🤓

aboedo:
LOL not sure if it didn't get posted or I'm going nuts, but I'll re-post the question, just a second

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -301,10 +301,9 @@ class PurchasesOrchestrator {
                   payment: SKMutablePayment,
                   package: Package?,
                   completion: @escaping PurchaseCompletedBlock) {
-        guard let productIdentifier = sk1Product.extractProductIdentifier(withPayment: payment) else {
-            Logger.error(Strings.purchase.could_not_purchase_product_id_not_found)
-            let errorMessage = "There was a problem purchasing the product: productIdentifier was nil"
-            completion(nil, nil, ErrorUtils.unknownError(message: errorMessage), false)
+        guard let productIdentifier = payment.extractProductIdentifier() else {
+            let message = Strings.purchase.could_not_purchase_product_id_not_found.description
+            completion(nil, nil, ErrorUtils.storeProblemError(withMessage: message), false)
```
aboedo, line 306: why are we ignoring the product identifier from the SKProduct now? 
The order should be: 
try getting the product id from the product, _then_ from the payment, and then if we still have nothing, raise an error

aboedo:
re-added the comment here https://github.com/RevenueCat/purchases-ios/pull/1680/files#r901905200

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: That's explained in the PR description 😇

NachoSoto: Basically that was nice, the problem is when the transaction comes back, we only have that, and if the identifier there is `nil` we never called the callback. So the test I wrote (for example) hung forever.

So because we didn't have test coverage for that workaround we didn't realize that it was enough to begin the purchase, but then we can't complete it.

aboedo: I read the description but my brain didn't process it 🤦‍♂️ my apologies. 
That makes total sense. I keep thinking that there has to be something that we can do to improve this case, but ultimately I feel like the solution will be to start moving people to SK2 as it matures. 
In the meantime, I do think we should have an extensive comment here, since it won't be immediately obvious to folks maintaining this

aboedo (approved):

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: > but ultimately I feel like the solution will be to start moving people to SK2 as it matures.

Agreed.

> In the meantime, I do think we should have an extensive comment here, since it won't be immediately obvious to folks maintaining this

Good call, adding.

PR #1683, opened 2022-06-09
Introduce Attribution class to encapsulate all public Attribution API.

taquitos:
Extract all the attribution-related functions from `Purchases` into their own extension.
Move implementation of these functions into a new class `Attribution`
Proxy the calls to `Attribution`

Goal: Remove hard-coded attribution functions from `Purchases` object in to future. Reduce `Purchases` class API by grouping related API under more appropriate locations.


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -215,6 +215,8 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
         set { systemInfo.finishTransactions = newValue }
     }
 
+    @objc let attribution: Attribution
```
NachoSoto, line 218: ❤️

NachoSoto (commented):
I absolutely love this. It's exactly the type of refactor that huge `Purchases` class needs.

I guess the only thing left here would be to add deprecations / obsoletions in an extension to point to the new methods.

We could start with simply `deprecation` which would be a warning, and I think we could ship that without a major version?

taquitos:
> I absolutely love this. It's exactly the type of refactor that huge `Purchases` class needs.
> 
> I guess the only thing left here would be to add deprecations / obsoletions in an extension to point to the new methods.
> 
> We could start with simply `deprecation` which would be a warning, and I think we could ship that without a major version?

That's exactly right, I didn't want to get too deep into the deprecation stuff until I had more signal that it was the right direction. Now that I've heard back from a few folks, I'll go ahead and close this and do the needful.

Issue #1597, opened 2022-05-21
[SANDBOX] Testing with low subscription interval latestPurchaseDate is often equal to expirationDate rendering entitlement inactive


ZeeWanderer:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Development/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
When testing purchases in sandbox with 30 second subscription intervals `latestPurchaseDate` is often equal to `expirationDate` making entitlement inactive on purchase.  Usually, this problem corrects itself on the next subscription tick. Since this does not happen in production or on higher subscription intervals (thou it does sometimes happen on 5-minute intervals but significantly less often) i was content to leave it be, but Apple rejected my app due to this bug. I am not completely sure it was exactly this but it seems like they are testing with low subscription intervals and do not wait 30-60 seconds for the next tick. I worked around this by detecting specific circumstances and treating entitlement as false negative but in my opinion, this is a bug.

Example of such entitlement:
```
Logica::Store::update_state: entitlements: ["base_content_access": <EntitlementInfo: "
identifier=base_content_access,
isActive=false,
willRenew=true,
periodType=PeriodType(rawValue: 1),
latestPurchaseDate=Optional(2022-05-21 13:58:39 +0000),
originalPurchaseDate=Optional(2022-05-21 13:58:39 +0000),
expirationDate=Optional(2022-05-21 13:58:39 +0000),
store=Store(rawValue: 0),
productIdentifier=fivesysdev.logica.renewable_subscription,
isSandbox=true,
unsubscribeDetectedAt=nil,
billingIssueDetectedAt=nil,
ownershipType=PurchaseOwnershipType(rawValue: 0)
>]
```

1. Environment
   1. Platform: iOS
   2. SDK version: 4.3.0
   3. StoreKit 2 enabled (Y/N): N
   4. OS version: iOS 15.5
   5. Xcode version: 13.4 (13F17a)
[...]

PR #1684, opened 2022-06-09
EntitlementInfo.isActive returns true if requestDate == expirationDate

beylmk:
Giving the end users back a millisecond of entitlement access, and hopefully solving [a Sandbox issue](https://github.com/RevenueCat/purchases-ios/issues/1597) wherein entitlements are returned as inactive because expirationDate == latestPurchaseDate for 30 second subscription renewals.

Did I miss any repercussions around making this change?

### DIFF
--- a/Tests/UnitTests/Purchasing/EntitlementInfosTests.swift
+++ b/Tests/UnitTests/Purchasing/EntitlementInfosTests.swift
@@ -148,6 +148,35 @@ class EntitlementInfosTests: TestCase {
         try verifyEntitlementActive()
     }
 
+    func testSubscriptionActiveIfExpiresDateEqualsRequestDate() throws {
+        let expirationAndRequestDate = "2019-08-16T10:30:42Z"
+        stubResponse(
+                entitlements: [
+                    "pro_cat": [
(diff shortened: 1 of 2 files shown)


NachoSoto:
I think this is safe.
I would add a test case for it though.

Tests/UnitTests/Purchasing/EntitlementInfosTests.swift
```diff
@@ -148,6 +148,35 @@ class EntitlementInfosTests: TestCase {
         try verifyEntitlementActive()
     }
 
+    func testSubscriptionActiveIfExpiresDateEqualsRequestDate() throws {
+        let expirationAndRequestDate = "2019-08-16T10:30:42Z"
```
NachoSoto, line 152: Beautiful

NachoSoto (approved):

NachoSoto:
@aboedo this looks good to you too?

aboedo:
looks good, although you're stealing a second's worth of subscription away from folks 😅 (jk, a second in a subscription in production is nothing)

PR #1687, opened 2022-06-10
Fixed thread unsafe usage of `toEventually`

NachoSoto:
Thanks @taquitos for finding this.

Our existing use as illustrated in the diff could lead to race conditions, because the `completionCalled.toEventually(beTrue())` would potentially evaluate before the next line, leading to flaky tests.


taquitos (approved):
Awesome! fun. Threading is

PR #1695, opened 2022-06-13
Remove create alias

taquitos:
Resolves [CSDK-152]

Deprecated in RCv3, removed in RCv4 🎉

[CSDK-152]: https://revenuecats.atlassian.net/browse/CSDK-152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1696, opened 2022-06-13
Extract Backend dependencies into a BackendConfiguration object

taquitos:
Extract various dependencies that are required to instantiate the `Backend` object.

Refactor to assist with the next few PRs that extract out other pieces
See [CSDK-82]
Resolves [CSDK-155]
Depends on [CSDK-152]

Blocked by https://github.com/RevenueCat/purchases-ios/pull/1695

[CSDK-82]: https://revenuecats.atlassian.net/browse/CSDK-82?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CSDK-155]: https://revenuecats.atlassian.net/browse/CSDK-155?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CSDK-152]: https://revenuecats.atlassian.net/browse/CSDK-152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Networking/Backend.swift
+++ b/Sources/Networking/Backend.swift
@@ -35,12 +35,11 @@ class Backend {
                      eTagManager: ETagManager,
                      attributionFetcher: AttributionFetcher,
                      dateProvider: DateProvider = DateProvider()) {
-        let httpClient = HTTPClient(systemInfo: systemInfo,
+        let httpClient = HTTPClient(apiKey: apiKey,
+                                    systemInfo: systemInfo,
                                     eTagManager: eTagManager,
                                     requestTimeout: httpClientTimeout)
-        let config = BackendConfiguration(apiKey: apiKey,
-                                          authHeaders: HTTPClient.authorizationHeader(withAPIKey: apiKey),
-                                          httpClient: httpClient,
+        let config = BackendConfiguration(httpClient: httpClient,
                                           operationQueue: QueueProvider.queue,
                                           dateProvider: dateProvider)
         self.init(backendConfig: config, attributionFetcher: attributionFetcher)
@@ -101,7 +100,6 @@ class Backend {
               appUserID: String,
               completion: @escaping OfferSigningResponseHandler) {
         let config = NetworkOperation.UserSpecificConfiguration(httpClient: self.config.httpClient,
-                                                                authHeaders: self.config.authHeaders,
                                                                 appUserID: appUserID)
 
         let postOfferData = PostOfferForSigningOperation.PostOfferForSigningData(offerIdentifier: offerIdentifier,
@@ -119,7 +117,6 @@ class Backend {
               appUserID: String,
               completion: SimpleResponseHandler?) {
         let config = NetworkOperation.UserSpecificConfiguration(httpClient: self.config.httpClient,
-                                                                authHeaders: self.config.authHeaders,
                                                                 appUserID: appUserID)
         let postAttributionDataOperation = PostAttributionDataOperation(configuration: config,
                                                                         attributionData: attributionData,
@@ -132,7 +129,6 @@ class Backend {
                newAppUserID: String,
                completion: @escaping LogInResponseHandler) {
         let config = NetworkOperation.UserSpecificConfiguration(httpClient: self.config.httpClient,
-                                                                authHeaders: self.config.authHeaders,
                                                                 appUserID: currentAppUserID)
         let loginOperation = LogInOperation(configuration: config,
                                             newAppUserID: newAppUserID,
@@ -146,7 +142,6 @@ class Backend {
 
     func getOfferings(appUserID: String, completion: @escaping OfferingsResponseHandler) {
         let config = NetworkOperation.UserSpecificConfiguration(httpClient: self.config.httpClient,
-                                                                authHeaders: self.config.authHeaders,
                                                                 appUserID: appUserID)
         let getOfferingsOperation = GetOfferingsOperation(configuration: config,
                                                           offeringsCallbackCache: self.offeringsCallbacksCache)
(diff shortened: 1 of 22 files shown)


Sources/Networking/Backend.swift
```diff
@@ -155,34 +141,49 @@ class Backend {
...
+        self.config.operationQueue.addOperation(getIntroEligibilityOperation)
+    }
+
+}
+
+extension Backend {
+
+    enum QueueProvider {
+
+        static var queue: OperationQueue {
```
NachoSoto, line 180: I would turn this into a `func` to make it more clear that it's creating a new `OperationQueue` every time it's called.
Maybe also rename it?

```suggestion
        static func createQueue() -> OperationQueue {
```

Sources/Networking/SubscribersAPI.swift
```diff
@@ -15,50 +15,41 @@ import Foundation
...
-    private let httpClient: HTTPClient
-    private let operationQueue: OperationQueue
-    private let authHeaders: [String: String]
+    private let backendConfig: BackendConfiguration
     private let customerInfoCallbackCache: CallbackCache<CustomerInfoCallback>
     private let attributionFetcher: AttributionFetcher
-    private let dateProvider: DateProvider
 
-    init(httpClient: HTTPClient,
+    init(backendConfig: BackendConfiguration,
```
NachoSoto, line 22: So much better.

Sources/Networking/BackendConfiguration.swift
```diff
@@ -0,0 +1,44 @@
...
+    let dateProvider: DateProvider
+
+    init(apiKey: String,
+         authHeaders: [String: String],
+         httpClient: HTTPClient,
+         operationQueue: OperationQueue,
+         callbackQueue: DispatchQueue = DispatchQueue(label: "Backend callbackQueue"),
+         dateProvider: DateProvider = DateProvider()) {
+        self.apiKey = apiKey
+        self.authHeaders = authHeaders
```
NachoSoto, line 33: Are these 2 needed, since we're already passing `HTTPClient`?

Tests/UnitTests/Networking/Backend/BaseBackendTest.swift
```diff
@@ -34,10 +34,13 @@ class BaseBackendTests: TestCase {
         self.httpClient = self.createClient()
         let attributionFetcher = AttributionFetcher(attributionFactory: MockAttributionTypeFactory(),
                                                     systemInfo: self.systemInfo)
-        self.backend = Backend(httpClient: self.httpClient,
-                               apiKey: Self.apiKey,
-                               attributionFetcher: attributionFetcher,
-                               dateProvider: MockDateProvider(stubbedNow: MockBackend.referenceDate))
+        let authHeaders = type(of: self.httpClient).authorizationHeader(withAPIKey: Self.apiKey)
```
NachoSoto, line 37: I'm confused about the usage of `type(of:)` here. Did you mean to use `HTTPClient.authorizationHeader`?
The latter is known at compile time, this is relying on the type at runtime. And since most of our classes are pretty much `final`, it should be the same.

Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
```diff
@@ -42,10 +42,15 @@ class BasePurchasesTests: TestCase {
...
         self.attributionFetcher = MockAttributionFetcher(attributionFactory: MockAttributionTypeFactory(),
                                                          systemInfo: systemInfoAttribution)
-        self.backend = MockBackend(httpClient: MockHTTPClient(systemInfo: self.systemInfo,
-                                                              eTagManager: MockETagManager()),
-                                   apiKey: "mockAPIKey",
-                                   attributionFetcher: self.attributionFetcher)
+
+        let apiKey = "mockAPIKey"
+        let httpClient = MockHTTPClient(systemInfo: self.systemInfo, eTagManager: MockETagManager())
+        let config = BackendConfiguration(apiKey: apiKey,
```
NachoSoto, line 48: This is why I made the comment above regarding the api key. It seems weird that we need to pass it even though we're creating the `HTTPClient` already.

NachoSoto (commented):
Great refactor.

My only concern is the new need to also pass api key and auth headers.

Sources/Networking/BackendConfiguration.swift
NachoSoto: 👍🏻

Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
taquitos: It's weird because the HTTPClient doesn't get instantiated with an APIKey, so if you want auth-headers, you need to pass in an apikey that we can use with the `HTTPClient.authorizationHeader(withAPIKey apiKey: String) ` method. I'll refactor this to make more sense.

Sources/Networking/BackendConfiguration.swift
```diff
@@ -15,22 +15,16 @@ import Foundation
...
     let httpClient: HTTPClient
 
     let callbackQueue: DispatchQueue
     let operationQueue: OperationQueue
     let dateProvider: DateProvider
 
-    init(apiKey: String,
-         authHeaders: [String: String],
-         httpClient: HTTPClient,
+    init(httpClient: HTTPClient,
```
taquitos, line 24: @NachoSoto I blame you for this commit.... the ripple effect was pretty big, but I think this refactor moves us in the right direction 😄

PR #1699, opened 2022-06-14
iOS 16: added new snapshots

NachoSoto:
Also added this version to iOS `OSVersionEquivalent`

I've created these by simply copying the iOS 15 version, which allows us to verify that the existing snapshots still make the iOS 16 tests pass with no changes.


Tests/UnitTests/Networking/Backend/__Snapshots__/BackendCreateAliasTests/iOS16-testAliasCallsBackendProperly.1.json
```diff
@@ -0,0 +1,12 @@
+{
```
NachoSoto, line 1: I'll wait to remove these alias ones when #1695 is merged.

NachoSoto:
Actually hold on. I'm going to create these snapshots by duplicating the iOS 15 ones, to make sure tests pass with no changes, instead of creating brand new files that are potentially different.

That way we can verify nothing changed.

taquitos (approved):
🤳🐐
image

PR #1671, opened 2022-06-07
`Integration Tests`: always clear `Bundle.appStoreReceiptURL` before every test

NachoSoto:
I've noticed sometimes tests start with some receipts that are pending to submit, which could lead to flaky tests.
Hopefully this makes them less flaky.

PR #1700, opened 2022-06-14
`StoreKitConfigTestCase`: delete `Bundle.appStoreReceiptURL` after every test

NachoSoto:
Equivalent to #1671 but for unit tests.

In iOS 16 (as of beta 1), the receipt persists across tests, despite the use of `SKTestSession.resetToDefaultState`.
This might also help with flaky tests pre-iOS 16 too.

Also note that I fixed this method in `BaseBackendIntegrationTests`. Using `URL.absoluteString` was incorrect there, so the file was probably never deleted.

I filed a feedback for this:
Screen Shot 2022-06-14 at 11 27 13


NachoSoto:
Tests failures are unrelated to this.

taquitos (approved):
🤞 🐐 
image

PR #1702, opened 2022-06-14
`ROT13`: optimized initialization and removed magic numbers

NachoSoto:
### Changes:
- Using `Dictionary.init(minimumCapacity:)` to allocate the necessary size initially instead of growing when adding data to it
- Defined `uppercase` in terms of `lowercase`
- Removed magic `13` and `26` numbers


taquitos (approved):
🚢🐐

image

PR #1703, opened 2022-06-14
`PurchasesOrchestrator.handlePurchasedTransaction`: always refresh receipt data

NachoSoto:
Fixes [CSDK-120].
See conversation: https://github.com/RevenueCat/purchases-ios/pull/1666#discussion_r892423490

> why do we have `onlyIfEmpty` in this case? If we're calling this method, we know that there was a purchase, right? I feel like we should always refresh in that case, there's a good chance that the receipt isn't up to date otherwise

[CSDK-120]: https://revenuecats.atlassian.net/browse/CSDK-120?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -413,8 +413,7 @@ class PurchasesOrchestrator {
         let customerInfo: CustomerInfo
 
         if let transaction = transaction {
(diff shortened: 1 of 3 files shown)


taquitos (approved):
🥤🐐

PR #1704, opened 2022-06-14
`PurchasesOrchestrator`: unify finish transactions between SK1 and SK2

NachoSoto:
Fixes [CSDK-122].

Before this change, transactions were finished separately for SK1 and SK2. Now this is combined through a common method in `StoreTransactionType`.

[CSDK-122]: https://revenuecats.atlassian.net/browse/CSDK-122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/TransactionsFactory.swift
+++ b/Sources/Purchasing/TransactionsFactory.swift
@@ -54,4 +54,8 @@ private struct BackendParsedTransaction: StoreTransactionType {
         self.quantity = 1
     }
 
+    func finish(_ wrapper: StoreKitWrapper) {
+        // Nothing to do
+    }
+
 }
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -588,9 +588,7 @@ private extension PurchasesOrchestrator {
             }
         }
 
-        if finishTransactions {
-            storeKitWrapper.finishTransaction(transaction)
-        }
+        self.finishTransactionIfNeeded(storeTransaction)
     }
 
     func handleDeferredTransaction(_ transaction: SKPaymentTransaction) {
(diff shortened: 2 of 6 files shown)


Sources/Purchasing/PurchasesOrchestrator.swift
```diff
@@ -843,6 +843,17 @@ private extension PurchasesOrchestrator {
...
+    func finish(_ transaction: StoreTransaction) async {
+        if let sk1Transaction = transaction.sk1Transaction {
+            self.storeKitWrapper.finishTransaction(sk1Transaction)
+        } else if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *),
+                  let sk2Transaction = transaction.sk2Transaction {
+            await sk2Transaction.finish()
+        } else {
+            // TODO: warning
+        }
+    }
```
NachoSoto, line 855: cc @aboedo since this came up in [our conversation](https://github.com/RevenueCat/purchases-ios/pull/1666#discussion_r892450066).

This looks nice in principle, but because `StoreKit.Transaction.finish` is `async`, we'd have to make a lot of `PurchasesOrchestrator` `async` just to support this.

I wanted to open this PR just in case what you had in mind was different. But I think what's best is what we have right now: finishing transactions in 2 different places for SK1 and SK2.

aboedo: we don't really need to wait for the result, though, right? Since we're not doing anything with it? 
i.e.: wouldn't we be able to have this function _not_ be async, and then just have it start a task that calls `finish` for sk2?

NachoSoto: I thought about it. There could be race conditions, but that's unclear not knowing exactly what finish does. The docs don't make it clear,, but I suspect you're probably right!

aboedo (approved):

Sources/Purchasing/TransactionsFactory.swift
```diff
@@ -54,4 +54,8 @@ private struct BackendParsedTransaction: StoreTransactionType {
         self.quantity = 1
     }
 
+    func finish(_ wrapper: StoreKitWrapper) {
+        // Nothing to do
+    }
```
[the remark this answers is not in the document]
aboedo: I don't recall a jira task for it, feel free to create one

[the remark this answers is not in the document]
NachoSoto: [CF-781]

PR #1706, opened 2022-06-16
Extract Offerings-related things into OfferingsAPI

taquitos:
Resolves [CSDK-154]
Extracts all the remaining logic from `Backend.swift` leaving only proxied methods.

[CSDK-154]: https://revenuecats.atlassian.net/browse/CSDK-154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


taquitos:
Note: I also reorganized the `Networking` folder so that all the HTTPClient-related things went into a HTTPClient group, leaving only the `Backend`, `BackendConfiguration`, and API classes.

Sources/Networking/Backend.swift
```diff
@@ -181,3 +149,7 @@ extension Backend {
     }
 
 }
+
+extension Backend {
+
+}
```
NachoSoto, line 155: Looks like this is a leftover?

NachoSoto (approved):
😻

Sources/Networking/Backend.swift
taquitos: ✅

PR #1711, opened 2022-06-16
Create and plumb MockIdentityAPI

taquitos:
resolves [CSDK-251]

[CSDK-251]: https://revenuecats.atlassian.net/browse/CSDK-251?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Tests/UnitTests/Identity/IdentityManagerTests.swift
+++ b/Tests/UnitTests/Identity/IdentityManagerTests.swift
@@ -26,7 +26,7 @@ class IdentityManagerTests: TestCase {
 
     override func setUpWithError() throws {
         try super.setUpWithError()
-        self.mockIdentityAPI = mockBackend.identity as? MockIdentityAPI
+        self.mockIdentityAPI = try XCTUnwrap(mockBackend.identity as? MockIdentityAPI)
         self.mockCustomerInfo = try CustomerInfo(data: [
             "request_date": "2019-08-16T10:30:42Z",
             "subscriber": [
--- a/Tests/UnitTests/Identity/IdentityManagerTests.swift
+++ b/Tests/UnitTests/Identity/IdentityManagerTests.swift
@@ -14,6 +14,7 @@ class IdentityManagerTests: TestCase {
     private let mockBackend = MockBackend()
     private var mockCustomerInfoManager: MockCustomerInfoManager!
 
+    private var mockIdentityAPI: MockIdentityAPI!
     private var mockCustomerInfo: CustomerInfo!
 
     private func create(appUserID: String?) -> IdentityManager {
@@ -25,7 +26,7 @@ class IdentityManagerTests: TestCase {
 
     override func setUpWithError() throws {
         try super.setUpWithError()
-
+        self.mockIdentityAPI = mockBackend.identity as? MockIdentityAPI
         self.mockCustomerInfo = try CustomerInfo(data: [
             "request_date": "2019-08-16T10:30:42Z",
             "subscriber": [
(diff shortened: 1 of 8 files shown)


Tests/UnitTests/Identity/IdentityManagerTests.swift
```diff
@@ -25,7 +26,7 @@ class IdentityManagerTests: TestCase {
 
     override func setUpWithError() throws {
         try super.setUpWithError()
-
+        self.mockIdentityAPI = mockBackend.identity as? MockIdentityAPI
```
NachoSoto, line 29: To avoid a potential crash in the test later, I'd change this to:
```suggestion
        self.mockIdentityAPI = try XCTUnwrap(mockBackend.identity as? MockIdentityAPI)
```
So this failure is caught early.

Tests/UnitTests/Mocks/MockIdentityAPI.swift
```diff
@@ -0,0 +1,51 @@
...
+
+class MockIdentityAPI: IdentityAPI {
+
+    var invokedLogIn = false
+    var invokedLogInCount = 0
+    var invokedLogInParameters: (currentAppUserID: String, newAppUserID: String)?
+    var invokedLogInParametersList = [(currentAppUserID: String, newAppUserID: String)]()
+    var stubbedLogInCompletionResult: Result<(info: CustomerInfo, created: Bool), BackendError>?
+
+    // swiftlint:disable force_try
```
NachoSoto, line 25: Maybe move this on top of the `MockSystemInfo` and change it to `disable:next`?

Tests/UnitTests/Mocks/MockIdentityAPI.swift
```diff
@@ -0,0 +1,51 @@
...
+import Foundation
+@testable import RevenueCat
+
+class MockIdentityAPI: IdentityAPI {
+
+    var invokedLogIn = false
+    var invokedLogInCount = 0
+    var invokedLogInParameters: (currentAppUserID: String, newAppUserID: String)?
+    var invokedLogInParametersList = [(currentAppUserID: String, newAppUserID: String)]()
+    var stubbedLogInCompletionResult: Result<(info: CustomerInfo, created: Bool), BackendError>?
```
NachoSoto, line 23: Maybe move these on top of the `logIn` method? I think other mocks are formatted like that so every method is next to these values.

Tests/UnitTests/Networking/Backend/BaseBackendTest.swift
```diff
@@ -37,7 +37,14 @@ class BaseBackendTests: TestCase {
...
                                                  dateProvider: MockDateProvider(stubbedNow: MockBackend.referenceDate))
-        self.backend = Backend(backendConfig: backendConfig, attributionFetcher: attributionFetcher)
+
+        let identity = IdentityAPI(backendConfig: backendConfig)
+        let offerings = OfferingsAPI(backendConfig: backendConfig)
+        let customer = CustomerAPI(backendConfig: backendConfig, attributionFetcher: attributionFetcher)
+        self.backend = Backend(backendConfig: backendConfig,
+                               customerAPI: customer,
+                               identityAPI: identity,
+                               offeringsAPI: offerings)
```
NachoSoto, line 47: I think this could use the other constructor so it doesn't manually create all the inputs?

NachoSoto (approved):
Another nice refactor👌🏻

Tests/UnitTests/Networking/Backend/BaseBackendTest.swift
taquitos: I was going to use the mocks here, but it turns out that isn't the right move and I forgot to switch this back to using the more friendly constructor! Thanks for catching that.

Tests/UnitTests/Mocks/MockIdentityAPI.swift
taquitos: Yup, you're totally right.

Tests/UnitTests/Identity/IdentityManagerTests.swift
taquitos: 💯

PR #1714, opened 2022-06-19
`SubscriberAttributesManagerIntegrationTests`: verify individual synced attributes

NachoSoto:
In order to fix [CSDK-36] we first need to be able to verify which user is getting which attributes.
This sets up a way to do that in integration tests.


Tests/BackendIntegrationTests/SubscriberAttributesManagerIntegrationTests.swift
```diff
@@ -76,59 +101,111 @@ class SubscriberAttributesManagerIntegrationTests: BaseBackendIntegrationTests {
...
-        self.attribution.setDisplayName("User 2")
+        _ = try await Purchases.shared.logIn(user2)
+        self.attribution.setDisplayName(name2)
 
         let errors = await self.syncAttributes()
-        verifyAttributesSyncedWithNoErrors(errors, 2)
+        self.verifyAttributesSyncedWithNoErrors(errors, 2)
+
+        // Note: this is being associated to the second user
+        self.verifySyncedAttribute(user2, [reserved(.displayName): name1])
```
NachoSoto, line 179: ⚠️ This already illustrates part of the bug. Notice that this is `user2`.

PR #1716, opened 2022-06-19
`logIn`/`logOut`: fix: sync attributes before aliasing

NachoSoto:
Fixes [CSDK-36].

Depends on #1715.


Sources/Identity/IdentityManager.swift
```diff
@@ -100,10 +129,10 @@ class IdentityManager: CurrentUserProvider {
         }
     }
 
-    func logOut(completion: (Error?) -> Void) {
+    func performLogOut(completion: (Error?) -> Void) {
```
NachoSoto, line 132: This could use a better name too.

Sources/Identity/IdentityManager.swift
```diff
@@ -20,23 +20,32 @@ protocol CurrentUserProvider {
 
 }
 
+protocol AttributeSyncing {
```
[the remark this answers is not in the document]
taquitos: I would have made it even worse with `AttributeSyncable` so yeah, I've got nothing.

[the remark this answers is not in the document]
taquitos: It's an old habit but with interfaces/protocols when it adds some ability, I always add `able` to the end.

Sources/SubscriberAttributes/SubscriberAttributesManager.swift
```diff
@@ -232,6 +232,17 @@ class SubscriberAttributesManager {
             network: network.rawValue)
         setAttributes(convertedAttribution, appUserID: appUserID)
     }
+
+}
+
+extension SubscriberAttributesManager: AttributeSyncing {
+
+    func syncSubscriberAttributesIfNeeded(currentAppUserID: String, completion: @escaping (() -> Void)) {
```
taquitos, line 240: I kinda think it should be `syncSubscriberAttributes` since the `ifNeeded` implies that it might not sync. The current implementation does seem to always sync each time

taquitos (approved):
Naming question, but otherwise looks ✅

Sources/Identity/IdentityManager.swift
NachoSoto: There are 2 `logOut` methods now. The `internal` one performs the sync first.

Sources/SubscriberAttributes/SubscriberAttributesManager.swift
NachoSoto: This is the same name as `Attribution.syncSubscriberAttributesIfNeeded`, it comes from the fact that there might not be attributes to sync for the given user, but maybe there's a better way to phrase that?
Or perhaps, that's obvious (if there's nothing to sync, welp, then do do anything).

I'll remove that suffix 👍🏻

Sources/Identity/IdentityManager.swift
NachoSoto: I'm not a big fan of that cause that's exposing an implementation detail in the name :/

taquitos: Yeah, I'm on the fence, though. If it's internal, the naming actually serves to remind us why we have it. If it were public API I'd definitely say you're 100% correct.

taquitos: Yeah, ultimately, I don't have a strong preference for non-public api naming here. What you have `perform` is totally workable IMO

Issue #1717, opened 2022-06-20
Configuration.Builder .with(appUserID: )  not accepting nil as a parameter


unxavi:
- [x ] I have updated Purchases SDK to the latest version
- [ x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Development/CONTRIBUTING.md)
- [ x] I have searched the [Community](https://community.revenuecat.com)
- [ x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
This might be a bug on the SDK or on the documentation. The documentation of the SDK stays:

```
 /**
         * Set an `appUserID`.
         * - Parameter appUserID: The unique app user id for this user. This user id will allow users to share their
         * purchases and subscriptions across devices. Pass `nil` or an empty string if you want ``Purchases``
         * to generate this for you.
         *
         * - Note: Best practice is to use a salted hash of your unique app user ids.
         *
         * - Important: Set this property if you have your own user identifiers that you manage.
         */
        @objc public func with(appUserID: String) -> RevenueCat.Configuration.Builder
```

But the param for `appUserID`is `String` and not `String?`

1. Environment
   1. Platform: iOS
   2. SDK version: 3.7.0
   3. StoreKit 2 enabled (Y/N): N
   4. OS version: All
   5. Xcode version: 13.4.1
   6. How widespread is the issue. Percentage of devices affected.

**Additional context**
None

aboedo:
Hey @unxavi! Thanks for opening!! 

Perhaps we could do a better job at documenting the intended usage here. 

From our [SDK reference docs](https://revenuecat.github.io/purchases-ios-docs/4.6.1/documentation/revenuecat/purchases/configure(with:)): 

```swift
 Purchases.configure(
     with: Configuration.Builder(withAPIKey: Constants.apiKey)
              .with(usesStoreKit2IfAvailable: true)
              .with(observerMode: false)
              .with(appUserID: "<app_user_id>")
              .build()
     )
```

So... basically you don't need to call `.with` unless you want to specify the `appUserID`. If you don't call it at all, it's the same as passing `nil`. 

I.e.: in order to pass in `nil`, just call
```swift
 Purchases.configure(with: "<api_key>")
```

I hope that makes sense!

unxavi:
[...]

PR #1720, opened 2022-06-21
`Purchases.configure`: added overload taking a `Configuration.Builder`

NachoSoto:
Fixes [CF-780].

See https://github.com/RevenueCat/purchases-ios/issues/1717#issuecomment-1160718519 for an example of a user trying to do this.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1442,6 +1442,35 @@ public extension Purchases {
     }
 
     /**
+     * Configures an instance of the Purchases SDK with a specified ``Configuration/Builder``.
+     *
+     * The instance will be set as a singleton.
+     * You should access the singleton instance using ``Purchases/shared``
+     *
+     * - Parameter builder: The ``Configuration/Builder`` object you wish to use to configure ``Purchases``
+     *
+     * - Returns: An instantiated ``Purchases`` object that has been set as a singleton.
+     *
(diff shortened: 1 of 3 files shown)


Sources/DocCDocumentation/DocCDocumentation.docc/RevenueCat.md
```diff
@@ -55,7 +55,7 @@ Or browse our iOS sample apps:
 
 ### Configuring the SDK
 - ``Purchases/configure(withAPIKey:)``
-- ``Purchases/configure(with:)``
+- ``Purchases/configure(with:)-6oipy``
```
NachoSoto, line 58: This helps DocC resolve the overload.
I decided not to add the new one here since it doesn't really add anything new, it's just a shortcut.

taquitos: So it doesn't show up like that in the docs? Cool.

taquitos (approved):
🐐👷‍♂️

download-77

Sources/DocCDocumentation/DocCDocumentation.docc/RevenueCat.md
NachoSoto: No hehe, it'll just show the right overload type:

taquitos: That's what I thought, just had to double check cause I'm not familiar with how it works 😄

PR #1721, opened 2022-06-21
`Purchases.customerInfo(fetchPolicy:)`: actually use `fetchPolicy` parameter

NachoSoto:
This was being ignored in the `async` method


taquitos (approved):
🤦‍♂️

PR #1722, opened 2022-06-21
New SwiftUI Purchase Tester example

joshdholtz:
### Motivation / Description

New app that tests almost all iOS SDK functionality.

This is not an example like MagicWeather that shows how to use RevenueCat in an app but will instead be an example for testing all the SDK functions on _any_ RevenueCat app.

### DIFF
--- a/Examples/PurchaseTester/Constants.swift
+++ b/Examples/PurchaseTester/Constants.swift
@@ -1,0 +1,21 @@
+//
+//  Constants.swift
+//  PurchaseTester
+//
+//  Created by Josh Holtz on 6/21/22.
+//
+
+import Foundation
+
+/*
+ Configuration file for your app's RevenueCat settings.
+ */
+
+struct Constants {
+
+    /*
+     The API key for your app from the RevenueCat dashboard: https://app.revenuecat.com
+     */
+    #error("Modify this property to reflect your app's API key, then comment this line out.")
+    static let apiKey = "api_key"
+}
--- a/Examples/PurchaseTester/Shared/Views/Customer/SubscriberAttributesView.swift
+++ b/Examples/PurchaseTester/Shared/Views/Customer/SubscriberAttributesView.swift
@@ -1,0 +1,153 @@
+//
+//  SubscriberAttributesView.swift
+//  PurchaseTester
+//
+//  Created by Josh Holtz on 2/3/22.
+//
+
+import Foundation
+import SwiftUI
+import RevenueCat
+
+struct SubscriberAttributesView: View {
+   
+
+    enum RevenueCatKind: String, CaseIterable, Identifiable {
+        var id: String {
+            return self.rawValue
+        }
+        
+        case setAd
+        case setEmail
+        case setDisplayName
+        case setKeyword
+        case setCampaign
+        case setCreative
+        case setAdGroup
+        case setPushToken
+        case setMediaSource
+        case setPhoneNumber
+    }
+    
+    enum ThirdPartyKind: String, CaseIterable, Identifiable {
+        var id: String {
+            return self.rawValue
+        }
+        
+        case setAdjustID
+        case setAppsflyerID
+        case setAirshipChannelID
+        case setCleverTapID
+        case setMparticleID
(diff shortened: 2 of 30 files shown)


aboedo:
Look at that thing 😻 This is awesome

Examples/PurchaseTester/Constants.swift
```diff
@@ -0,0 +1,21 @@
+//
```
aboedo, line 1: Could we have this live in `Tests/TestingApps` instead since it's not really an example?

Examples/PurchaseTester/Shared/Views/Customer/SubscriberAttributesView.swift
```diff
@@ -0,0 +1,153 @@
...
+        .textFieldAlert(isShowing: self.$showAttributesAlert, title: "Attribute", fields: [
+            ("Key", "Key name", self.$attributeKey),
+            ("Value", "Value", self.$attributeValue)
+        ]) {
+            Purchases.shared.setAttributes([self.attributeKey:  self.attributeValue])
+        }
+        .textFieldAlert(isShowing: self.$showOtherAlert, title: self.revenueCatKind?.rawValue ?? self.thirdPartyKind?.rawValue ?? "<ERROR>", fields: [
+            ("Value", "Value", self.$otherValue),
+        ]) {
+            if let kind = self.revenueCatKind {
```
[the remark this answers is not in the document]
joshdholtz: I might be missing a few here! This is also using the deprecated version of these so those should be updated with the new attribution namespace thing at some point 🤷‍♂️

Examples/PurchaseTester/Constants.swift
joshdholtz: I think this makes sense! I can move it 💪

aboedo (approved):
🚀🚀🚀

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/Offerings/PromoOfferDetailsView.swift
```diff
@@ -0,0 +1,121 @@
...
+//
+
+import Foundation
+import SwiftUI
+import RevenueCat
+
+struct PromoOfferDetailsView: View {
+    let package: RevenueCat.Package
+    
+    @State var eligibility: [RevenueCat.StoreProductDiscount: RevenueCat.PromotionalOffer?]? = nil
```
NachoSoto, line 15: Do you need all these `RevenueCat.`s everywhere?

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/Offerings/PromoOfferDetailsView.swift
```diff
@@ -0,0 +1,121 @@
...
+                if let promotionalOffer = self.eligibility?[discount] as? PromotionalOffer {
+                    Text("Accept promo offer")
+                        .foregroundColor(.blue)
+                        .padding(.vertical, 10)
+                        .onTapGesture {
+                            purchasePromo(promotionalOffer: promotionalOffer)
+                        }
+                }
+            }
+        }.onAppear {
```
NachoSoto, line 106: Ditto: use `.task`

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/Offerings/PromoOfferDetailsView.swift
NachoSoto, line 15: Also this `@State` property needs to be `private`.

joshdholtz: Oh! This is because I also had a V3 version and so I was using `RevenueCat.Package` and `Purchases.Package` 🙃 I will remove!

joshdholtz:
Updated with a bunch of suggestions from @NachoSoto 💪

PR #1726, opened 2022-06-21
`TrialOrIntroPriceEligibilityChecker`: fixed integration test to ensure receipt is always loaded

NachoSoto:
The reason the receipt data is required, is because if it's `nil`, it ends up becoming empty because of `data ?? Data()`, but `GetIntroEligibilityOperation` never actually runs because that's empty.

To make sure the test doesn't fail, we begin by ensuring the receipt data is loaded first.

Also documented the reason for `ReceiptRefreshPolicy.never`.

### DIFF
--- a/Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
+++ b/Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
@@ -82,7 +82,9 @@ class TrialOrIntroPriceEligibilityChecker {
 
     func sk1CheckEligibility(_ productIdentifiers: [String],
                              completion: @escaping ReceiveIntroEligibilityBlock) {
-        receiptFetcher.receiptData(refreshPolicy: .never) { data in
+        // We don't want to refresh receipts because it will likely prompt the user for their credentials,
+        // and intro eligibility is triggered programmatically.
+        self.receiptFetcher.receiptData(refreshPolicy: .never) { data in
             if #available(iOS 12.0, macOS 10.14, tvOS 12.0, watchOS 6.2, *),
                let data = data {
                 self.sk1CheckEligibility(with: data,
--- a/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
@@ -208,6 +208,12 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
     func testEligibleForIntroBeforePurchase() async throws {
         try AvailabilityChecks.iOS13APIAvailableOrSkipTest()
 
+        if Self.storeKit2Setting == .disabled {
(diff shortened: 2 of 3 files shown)


taquitos (approved):
🤯 🐐 

image

Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
```diff
@@ -82,7 +82,7 @@ class TrialOrIntroPriceEligibilityChecker {
 
     func sk1CheckEligibility(_ productIdentifiers: [String],
                              completion: @escaping ReceiveIntroEligibilityBlock) {
-        receiptFetcher.receiptData(refreshPolicy: .never) { data in
+        receiptFetcher.receiptData(refreshPolicy: .onlyIfEmpty) { data in
```
NachoSoto, line 85: cc @aboedo any reason why this should be `.never`?

aboedo: yeah, sadly. 
refreshing here will likely prompt the user for their credentials, and intro eligibility should be triggered programmatically. 

In theory, this should only be a problem in sandbox, but I have a vague recollection of having moved to `never` after this caused a lot of customer confusion and headaches

aboedo: that makes total sense. 

Also, would you mind adding what I explained as a comment? Your question about why this should be `.never` is very valid and there's no reason maintainers should know the explanation before seeing this code, I should have added a comment myself when adding the `.never` 🤦

NachoSoto:
@aboedo I've updated this:
- Change implementation to only sandbox
- Added docs
- Added a unit test to cover this

Because Integration Tests aren't detecting "sandbox" I still need those tests separately.

Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
```diff
@@ -82,7 +82,15 @@ class TrialOrIntroPriceEligibilityChecker {
 
     func sk1CheckEligibility(_ productIdentifiers: [String],
                              completion: @escaping ReceiveIntroEligibilityBlock) {
-        receiptFetcher.receiptData(refreshPolicy: .never) { data in
+        // In production there’s always a receipt.
+        // But in sandbox, the receipt might not be there for an app on a fresh install (or in tests).
+        // We don't want to always refresh receipts in production because it will
+        // likely prompt the user for their credentials, and intro eligibility should be triggered programmatically.
+        let refreshPolicy: ReceiptRefreshPolicy = self.systemInfo.isSandbox
+            ? .onlyIfEmpty
```
NachoSoto, line 90: Just rebased. Integration tests will use this now :)

NachoSoto:
All tests passing 🎉

PR #1727, opened 2022-06-22
AdServices

joshdholtz:
### Motivation

@beylmk did all of the hard work on this so she gets credit for this PR 😊 

#### New API

```swift
Purchases.shared.attribution.enableAdServicesAttributionTokenCollection()
```

⚠️ This was originally planned as `Purchases.automaticAdServicesAttributionTokenCollection = true` but changed when #1693 was introduced

💭 Any thoughts on this public API would be appreciated! Not sure if this feels too hidden or too long? 🤷‍♂️ 

#### Behavior

- Will attempt to collect and post the AdServices token when `enableAdServicesAttributionTokenCollection` is called
- Will attempt to send when app enters foreground
- Will only send the token once


RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -340,13 +338,17 @@
...
+		A55D5D66282ECCC100FA7623 /* PostAdServicesTokenOperation.swift in Sources */ = {isa = PBXBuildFile; fileRef = A55D5D65282ECCC100FA7623 /* PostAdServicesTokenOperation.swift */; };
 		A563F586271E072B00246E0C /* MockBeginRefundRequestHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = A563F585271E072B00246E0C /* MockBeginRefundRequestHelper.swift */; };
 		A563F589271E1DAD00246E0C /* MockBeginRefundRequestHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = A563F585271E072B00246E0C /* MockBeginRefundRequestHelper.swift */; };
+		A56C2E012819C33500995421 /* BackendPostAdServicesTokenTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = A56C2E002819C33500995421 /* BackendPostAdServicesTokenTests.swift */; };
 		A56F9AB126990E9200AFC48F /* CustomerInfo.swift in Sources */ = {isa = PBXBuildFile; fileRef = A56F9AB026990E9200AFC48F /* CustomerInfo.swift */; };
+		A5B6CDD8280F3843007629D5 /* AdServices.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = A5B6CDD5280F3843007629D5 /* AdServices.framework */; platformFilters = (ios, maccatalyst, macos, ); settings = {ATTRIBUTES = (Weak, ); }; };
```
NachoSoto, line 351: 👍🏻

Sources/Attribution/AttributionNetwork.swift
```diff
@@ -22,37 +22,43 @@ import Foundation
     /**
      Apple's search ads
      */
-    case appleSearchAds
+    @available(*, deprecated, message: "use adServices")
```
NachoSoto, line 25: Maybe this could use some more explanation.

NachoSoto: Or replace with `renamed: "adServices"`

Sources/Logging/Strings/AttributionStrings.swift
```diff
@@ -115,8 +119,21 @@ extension AttributionStrings: CustomStringConvertible {
         case .missing_advertiser_identifiers:
             return "Attribution error: identifierForAdvertisers is missing"
 
-        case .unknown_sk2_product_discount_type(let rawValue):
```
NachoSoto, line 118: Good call

Sources/Networking/Backend.swift
```diff
@@ -96,7 +94,6 @@ class Backend {
               completion: CustomerAPI.SimpleResponseHandler?) {
         self.customer.post(subscriberAttributes: subscriberAttributes, appUserID: appUserID, completion: completion)
     }
-
```
NachoSoto, line 99: Style guide. This is still needed.

Sources/Purchasing/Purchases/Attribution.swift
```diff
@@ -40,6 +47,34 @@ import Foundation
 
 }
 
+public extension Attribution {
+
+    /**
+     * Enable automatic collection of AdServices attribution token.
+     *
+     * Should match OS availability in https://developer.apple.com/documentation/ad_services
```
NachoSoto, line 55: ⚠️  think this shouldn't be a `public` comment

Sources/Purchasing/Purchases/Attribution.swift
```diff
@@ -40,6 +47,34 @@ import Foundation
...
+     */
+    @available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *)
+    @available(tvOS, unavailable)
+    @available(watchOS, unavailable)
+    @objc func enableAdServicesAttributionTokenCollection() {
+        self.automaticAdServicesAttributionTokenCollection = true
+        self.postAdServicesTokenIfNeeded()
```
NachoSoto, line 62: This is only called after setting it to true, so it doesn't need the "if needed" and guard.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -505,13 +500,6 @@ extension Purchases {
         attributionPoster.post(attributionData: data, fromNetwork: network, networkUserId: networkUserId)
     }
 
-    private func postAppleSearchAddsAttributionCollectionIfNeeded() {
-        guard Self.automaticAppleSearchAdsAttributionCollection else {
```
NachoSoto, line 509: This is no longer used, but it's only deprecated. I think that might confuse users because it's not just deprecated; but it will no longer work.

The alternative is to obsolete it, but that would require a new major version. Either way it's breaking behavior though. Should we make the old property "secretly" enable the new one instead?

Tests/APITesters/SwiftAPITester/SwiftAPITester/PurchasesAPI.swift
```diff
@@ -263,6 +263,7 @@ private func checkDeprecatedMethods(_ purchases: Purchases) {
 
     Purchases.addAttributionData([String: Any](), from: AttributionNetwork.adjust, forNetworkUserId: "")
     Purchases.addAttributionData([String: Any](), from: AttributionNetwork.adjust, forNetworkUserId: nil)
+    let _: Bool = Purchases.automaticAppleSearchAdsAttributionCollection
```
NachoSoto, line 266: 👍🏻

NachoSoto: I just realized this is only checking the getter and not the setter?

Tests/UnitTests/Purchasing/Purchases/PurchasesAttributionDataTests.swift
```diff
@@ -73,74 +56,4 @@ class PurchasesAttributionDataTests: BasePurchasesTests {
         expect(invokedParameters.appUserID) == self.purchases?.appUserID
     }
 
-    func testAttributionDataSendsNetworkAppUserId() throws {
```
NachoSoto, line 76: Should these have the equivalent tests now?

Tests/UnitTests/SubscriberAttributes/SubscriberAttributesManagerTests.swift
```diff
@@ -525,7 +525,7 @@ class SubscriberAttributesManagerTests: TestCase {
         mockDeviceCache.stubbedUnsyncedAttributesForAllUsersResult = allAttributes
 
         let mockError: BackendError = .missingAppUserID()
-        mockBackend.stubbedPostSubscriberAttributesCompletionResult = (mockError, ())
+        mockBackend.stubbedPostSubscriberAttributesCompletionResult = Result.failure(mockError)
```
NachoSoto, line 528: This can be
```suggestion
        mockBackend.stubbedPostSubscriberAttributesCompletionResult = .failure(mockError)
```

NachoSoto (changes requested):
The main issue is the old API still being available but "broken"

Sources/Purchasing/Purchases/Attribution.swift
beylmk: oh also -- the "ifNeeded" kinda covers checking whether it's already been posted, i.e. if a token exists in UserDefaults

NachoSoto: Oh I missed that! Never mind then.

Tests/APITesters/SwiftAPITester/SwiftAPITester/PurchasesAPI.swift
NachoSoto: The property is still there, it's just deprecated.

Sources/Purchasing/Purchases/Purchases.swift
beylmk: so... we could make `automaticAppleSearchAdsAttributionCollection` enable `attribution.automaticAdServicesAttributionTokenCollection`, but that wouldn't work for devices below 14.3... 

if we want devices <14.3 to work, we would need to add back all of the appleSearchAds stuff.

beylmk: there was a slack discussion that we are okay with cutting support for <14.3. but i understand this is a semver issue...

NachoSoto: The main thing is how can we do this without making it seem like everything is working for users setting the old property, but then realizing it's not doing anything.

beylmk: @aboedo thoughts on a 5.0.0 obsoleting iad?

Tests/APITesters/ObjCAPITester/ObjCAPITester/RCPurchasesAPI.m
```diff
@@ -16,6 +16,7 @@ @implementation RCPurchasesAPI
 NSString *version;
 
 BOOL automaticAppleSearchAdsAttributionCollection;
+BOOL automaticAdServicesAttributionTokenCollection;
```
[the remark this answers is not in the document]
joshdholtz: Oh, good call! I deleted the usage but forget about this 🤦‍♂️

Sources/Purchasing/Purchases/Purchases.swift
beylmk: seeing info coming through for a pre-14.3 user :/ will add all the iad code back.

Sources/Misc/Deprecations.swift
```diff
@@ -143,6 +143,12 @@ public extension Purchases {
         )
     }
 
+    /**
+     * Enable automatic collection of Apple Search Ads attribution. Defaults to `false`.
+     */
+    @available(*, deprecated, message: "Use Purchases.automaticAdServicesAttributionTokenCollection instead")
```
NachoSoto, line 149: Instead of `message` this would be better with `renamed` so it provides a fix-it (it's also missing the new `attribution`):
```suggestion
    @available(*, deprecated, renamed: "attribution.automaticAdServicesAttributionTokenCollection")
```
I think that should be correct but you can check it in the API testers.

Sources/Caching/DeviceCache.swift
```diff
@@ -274,17 +274,34 @@ class DeviceCache {
...
+            // convert keys from UserDefault from Integer String to AttributionNetwork
+            let latestSent: [AttributionNetwork: String] =
+                latestAdvertisingIdsByRawNetworkSent.reduce(into: [:]) { adIdsByNetwork, adIdByRawNetworkString in
+                    if let networkRawValue = Int(adIdByRawNetworkString.key),
+                       let attributionNetwork = AttributionNetwork(rawValue: networkRawValue) {
+                    adIdsByNetwork[attributionNetwork] = adIdByRawNetworkString.value
+                } else {
+                    Logger.error(Strings.attribution.latest_attribution_sent_user_defaults_invalid)
+                }
+            }
```
NachoSoto, line 291: #1734 is in `main` now so this can be simplified:
```suggestion
           let latestSent: [AttributionNetwork: String] =
                latestAdvertisingIdsByRawNetworkSent.compactMapKeys { 
                    guard let networkRawValue = Int(adIdByRawNetworkString.key),
                       let attributionNetwork = AttributionNetwork(rawValue: networkRawValue) else {
                            Logger.error(Strings.attribution.latest_attribution_sent_user_defaults_invalid)    
                            return nil
                       }

                       return attributionNetwork
                   }
```

Sources/Purchasing/Purchases/Attribution.swift
```diff
@@ -40,6 +47,32 @@ import Foundation
...
 }
 
+public extension Attribution {
+
+    /**
+     * Enable automatic collection of AdServices attribution token.
+     */
+    @available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *)
+    @available(tvOS, unavailable)
+    @available(watchOS, unavailable)
```
NachoSoto, line 57: I would put these in the `extension` to avoid repeating them in both methods.

Tests/UnitTests/Attribution/AttributionPosterTests.swift
```diff
@@ -57,245 +55,118 @@ class AttributionPosterTests: TestCase {
...
-    override func tearDown() {
-        super.tearDown()
-        UserDefaults.standard.removePersistentDomain(forName: userDefaultsSuiteName)
-        UserDefaults.standard.synchronize()
-        resetAttributionStaticProperties()
+}
+
+#if canImport(AdServices)
+@available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *)
+class AdServicesAttributionPosterTests: BaseAttributionPosterTests {
```
NachoSoto, line 80: Ditto this needs an availability check in `setUpWithError`.

Tests/UnitTests/Purchasing/Purchases/PurchasesTests.swift
```diff
@@ -0,0 +1,855 @@
...
+//  Created by RevenueCat.
+//
+
+import Nimble
+import StoreKit
+import XCTest
+
+@testable import RevenueCat
+
+class PurchasesTests: BasePurchasesTests {
```
NachoSoto, line 19: ⚠️ this file was split starting from #1609, but looks like it got re-added here!

NachoSoto (changes requested):
I just noticed this re-added `PurchasesTests`. If there are new tests there, you can put them in one of the subclasses or create another `BasePurchasesTest` subclass.

Sources/Attribution/AfficheClientProxy.swift
```diff
@@ -0,0 +1,62 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  AfficheClientProxy.swift
+//
+//  Created by Juanpe Catalán on 14/7/21.
+//
+
+import Foundation
+
+typealias AttributionDetailsBlock = ([String: NSObject]?, Error?) -> Void
```
NachoSoto: I would move this inside `AfficheClientProxy` to namespace it instead of putting it in the global namespace.
Also could this be `Result<[String: NSObject], Error>`?

Sources/Attribution/AfficheClientProxy.swift
```diff
@@ -0,0 +1,62 @@
...
+
+// We need this class to avoid Kid apps being rejected for getting idfa. It seems like App
+// Review uses some grep to find the class names, so we ended up creating a fake class that
+// exposes the same methods we're looking for in ADClient to call the same methods and mangling
+// the class names. So that Apple can't find them during the review, but we can still access them on runtime.
+// You can see the class here: https://rev.cat/fake-affiche-client
+class FakeAfficheClient: NSObject {
+
+    // We need this method to be available as an optional implicitly unwrapped method for `AnyClass`.
```
NachoSoto: I'm not sure what this means. This isn't an optional?

Sources/Attribution/AfficheClientProxy.swift
```diff
@@ -0,0 +1,62 @@
...
+    @objc static func sharedClient() -> FakeAfficheClient {
+        FakeAfficheClient()
+    }
+
+    // We need this method to be available as an optional implicitly unwrapped method for `AnyClass`.
```
NachoSoto: Ditto

Sources/Attribution/AttributionFetcher.swift
```diff
@@ -73,6 +73,21 @@ class AttributionFetcher {
         return nil
     }
 
+    func afficheClientAttributionDetails(completion: @escaping ([String: NSObject]?, Error?) -> Void) {
+        // Should match available platforms in
+        // https://developer.apple.com/documentation/iad/adclient?language=swift
```
NachoSoto: There's no longer an availability here though?

Sources/Attribution/AttributionFetcher.swift
```diff
@@ -73,6 +73,21 @@ class AttributionFetcher {
         return nil
     }
 
+    func afficheClientAttributionDetails(completion: @escaping ([String: NSObject]?, Error?) -> Void) {
```
NachoSoto: Should this be `AttributionDetailsBlock`?

Sources/Attribution/AttributionPoster.swift
```diff
@@ -87,10 +81,45 @@ class AttributionPoster {
...
+
+        guard self.latestNetworkIdAndAdvertisingIdentifierSent(network: .appleSearchAds) == nil else {
+            return
+        }
+
+        attributionFetcher.afficheClientAttributionDetails { attributionDetails, error in
+            guard let attributionDetails = attributionDetails,
+                  error == nil else {
+                return
+            }
```
NachoSoto, line 112: If you use `Result` then this wouldn't be needed.

Sources/Attribution/AttributionPoster.swift
```diff
@@ -53,19 +53,13 @@ class AttributionPoster {
...
 
         let currentAppUserID = self.currentUserProvider.currentAppUserID
-        let latestNetworkIdsAndAdvertisingIdsSentByNetwork =
-            deviceCache.latestAdvertisingIdsByNetworkSent(appUserID: currentAppUserID)
-        let latestSentToNetwork = latestNetworkIdsAndAdvertisingIdsSentByNetwork[network]
-
-        let newValueForNetwork = "\(identifierForAdvertisers ?? "(null)")_\(networkUserId ?? "(null)")"
-        guard latestSentToNetwork != newValueForNetwork else {
-            Logger.debug(Strings.attribution.skip_same_attributes)
+        guard let newDictToCache = getNewDictToCache(currentAppUserID: currentAppUserID,
```
NachoSoto, line 56: Nit: explicit `self` per the style guide.

Tests/UnitTests/Mocks/MockAttributionTypeFactory.swift
```diff
@@ -16,6 +16,25 @@ import Foundation
...
 @testable import RevenueCat
 
+class MockAdClientProxy: AfficheClientProxy {
+
+    static var mockAttributionDetails: [String: NSObject] = [
+        "Version3.1":
+            [
+                "iad-campaign-id": 15292426,
+                "iad-attribution": true
+            ] as NSObject
```
NachoSoto, line 26: I think this `as NSObject` is probably not needed.

Tests/UnitTests/Mocks/MockAttributionTypeFactory.swift
```diff
@@ -16,6 +16,25 @@ import Foundation
...
+    ]
+    static var mockError: Error?
+    static var requestAttributionDetailsCallCount = 0
+
+    override func requestAttributionDetails(_ completionHandler: @escaping AttributionDetailsBlock) {
+        Self.requestAttributionDetailsCallCount += 1
+        completionHandler(Self.mockAttributionDetails, Self.mockError)
```
NachoSoto, line 33: Ditto here. Using `Result` avoids potentially setting a value and an error at the same time (or neither).

Sources/Misc/Deprecations.swift
```diff
@@ -146,7 +146,7 @@ public extension Purchases {
     /**
      * Enable automatic collection of Apple Search Ads attribution. Defaults to `false`.
      */
-    @available(*, deprecated, message: "Use Purchases.automaticAdServicesAttributionTokenCollection instead")
+    @available(*, deprecated, message: "Use Purchases.attribution.enableAdServicesAttributionTokenCollection() instead")
```
NachoSoto, line 149: I'm confused, this is no longer in the PR?

NachoSoto:
Sorry I think I was reviewing the wrong diff 😕

Sources/Caching/DeviceCache.swift
```diff
@@ -274,17 +274,31 @@ class DeviceCache {
...
-            let latestNetworkAndAdvertisingIdsSent = $0.object(forKey: key) as? [String: String] ?? [:]
-            return latestNetworkAndAdvertisingIdsSent
+            let latestAdvertisingIdsByRawNetworkSent = $0.object(forKey: key) as? [String: String] ?? [:]
+
+            // convert keys from UserDefault from Integer String to AttributionNetwork
+            let latestSent: [AttributionNetwork: String] =
+                 latestAdvertisingIdsByRawNetworkSent.compactMapKeys { network in
+                     guard let networkRawValue = Int(network),
+                        let attributionNetwork = AttributionNetwork(rawValue: networkRawValue) else {
+                             Logger.error(Strings.attribution.latest_attribution_sent_user_defaults_invalid)
```
NachoSoto, line 287: I would include the `network` string in the error, so if we get a report with this message we can figure out why.

Sources/Caching/DeviceCache.swift
```diff
@@ -274,17 +274,31 @@ class DeviceCache {
...
+
+            return latestSent
         }
     }
 
-    func set(latestNetworkAndAdvertisingIdsSent: [String: String], appUserID: String) {
+    func set(latestAdvertisingIdsByNetworkSent: [AttributionNetwork: String], appUserID: String) {
         self.userDefaults.write {
-            $0.setValue(latestNetworkAndAdvertisingIdsSent,
+            // convert AttributionNetwork to Integer as String
```
NachoSoto, line 299: I think this comment doesn't add anything, that's implied in the types.

Tests/UnitTests/Mocks/MockDeviceCache.swift
```diff
@@ -227,4 +227,23 @@ class MockDeviceCache: DeviceCache {
...
+        invokedSetLatestNetworkAndAdvertisingIdsSentCount += 1
+        invokedSetLatestNetworkAndAdvertisingIdsSentParameters = (latestAdvertisingIdsByNetworkSent, appUserID)
+        invokedSetLatestNetworkAndAdvertisingIdsSentParametersList.append(
+            (latestAdvertisingIdsByNetworkSent, appUserID)
+        )
+    }
+
+    override func latestAdvertisingIdsByNetworkSent(appUserID: String) -> [AttributionNetwork: String] {
+        return invokedSetLatestNetworkAndAdvertisingIdsSentParameters?.adIdsByNetwork ?? [:]
+    }
```
NachoSoto, line 248: Nit: new line after this.

NachoSoto (commented):
I think this is looking good!
The main thing left is the `PurchasesTests` file being re-added.

NachoSoto:
I see what happened, I reviewed a bunch of (old) code that was simply re-added. You can ignore all those comments since that's existing code!

beylmk:
> I see what happened, I reviewed a bunch of (old) code that was simply re-added. You can ignore all those comments since that's existing code!

thanks for enduring all the confusion on this one 😬

Tests/UnitTests/Attribution/AttributionPosterTests.swift
beylmk: doing the availability check in `setUpWithError` leads to the "redundant availability check" error... and it seems like we've avoided making more specific checks in the `AvailabilityChecks` class?

beylmk:
@NachoSoto i believe all updates are done/replied to... we won't actually merge this btw, we are going to use this branch to release a beta.

Sources/Attribution/AttributionPoster.swift
```diff
@@ -130,6 +123,23 @@ class AttributionPoster {
...
 
+    // should match OS availability in https://developer.apple.com/documentation/ad_services
+    @available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *)
+    @available(tvOS, unavailable)
+    @available(watchOS, unavailable)
+    func postAdServicesTokenIfNeeded() {
+        let latestTokenSent = latestNetworkIdAndAdvertisingIdentifierSent(network: .adServices)
+        guard latestTokenSent == nil else {
+            return
+        }
```
NachoSoto, line 134: Since `latestTokenSent` isn't used in the method we don't need a variable for it (it's kind of confusing because it's always going to be `nil`. I'd change this to:
```suggestion
        guard self.latestNetworkIdAndAdvertisingIdentifierSent(network: .adServices) == nil else {
            return
        }
```

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostAttributionDataTests/iOS12-testPostAttributesPutsDataInDataKey.1.json
```diff
@@ -1,16 +0,0 @@
-{
```
NachoSoto, line 1: I suspect this file is still needed to make iOS 12 tests pass.

NachoSoto (approved):
Awesome!

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostAttributionDataTests/iOS12-testPostAttributesPutsDataInDataKey.1.json
beylmk: ah yes, my bad. updating

Sources/Attribution/AttributionNetwork.swift
beylmk: idk i hesitate since it's not really a rename, it results in different data being collected. but at the same time, the auto-correction is nice.

NachoSoto: Oh yeah that's a great point

Sources/Attribution/AttributionPoster.swift
NachoSoto: Bump

Tests/UnitTests/Attribution/AttributionPosterTests.swift
NachoSoto: I think without it these tests will fail on iOS 13 though (which won't be caught until this is merged in `main`.

Tests/UnitTests/Purchasing/Purchases/PurchasesAttributionDataTests.swift
NachoSoto: Bump?

NachoSoto (commented):
I think this is ready except for the iOS 12/13 issue with availability checks. We can add a new method in `AvailabilityChecks` to deal with that.

PR #1729, opened 2022-06-22
update docs for `showManageSubscriptions`

### LINKED ISSUE #1728: Confusing documentation for showManageSubscriptions:
/**
     * Use this function to open the manage subscriptions page.

     * - Parameter completion: A completion block that is called when the modal is closed.
     * If it was not successful, there will be an `Error`.

     * If the manage subscriptions page can't be opened, the ``CustomerInfo/managementURL`` in
     * the ``CustomerInfo`` will be opened. If ``CustomerInfo/managementURL`` is not available,
     * the App Store's subscription management section will be opened.
     *
     * The `completion` block will be called when the modal is opened, not when it's actually closed.
     * This is because of an undocumented change in StoreKit's behavior between iOS 15.0 and 15.2,
     * where 15.0 would return when the modal was closed,
     * and 15.2 returns when the modal is opened.
     */

- The `completion` block will be called when the modal is opened, not when it's actually closed.
- Parameter completion: A completion block that is called when the modal is closed.

aboedo:
Fixes #1728

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1297,10 +1297,12 @@ public extension Purchases {
 
     /**
      * Use this function to open the manage subscriptions page.
-
-     * - Parameter completion: A completion block that is called when the modal is closed.
-     * If it was not successful, there will be an `Error`.
-
+     *
+     * - Parameter completion: A completion block that will be called when the modal is opened,
+     * not when it's actually closed. ß∫This is because of an undocumented change in StoreKit's behavior
+     * between iOS 15.0 and 15.2, where 15.0 would return when the modal was closed, and 15.2 returns
+     * when the modal is opened.
+     *
      * If the manage subscriptions page can't be opened, the ``CustomerInfo/managementURL`` in
      * the ``CustomerInfo`` will be opened. If ``CustomerInfo/managementURL`` is not available,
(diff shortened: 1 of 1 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1297,10 +1297,12 @@ public extension Purchases {
 
     /**
      * Use this function to open the manage subscriptions page.
-
-     * - Parameter completion: A completion block that is called when the modal is closed.
-     * If it was not successful, there will be an `Error`.
-
+     *
+     * - Parameter completion: A completion block that will be called when the modal is opened,
+     * not when it's actually closed. ß∫This is because of an undocumented change in StoreKit's behavior
```
NachoSoto, line 1302:
```suggestion
     * not when it's actually closed. This is because of an undocumented change in StoreKit's behavior
```

NachoSoto (approved):

NachoSoto:
I like the math notation though :P

aboedo:
@NachoSoto you're right, I'm gonna keep the math annotation. It makes it look cooler

Sources/Purchasing/Purchases/Purchases.swift
aboedo: great catch 😅 I'm not even sure how I managed to add that in there

PR #1730, opened 2022-06-22
Integration Tests: ensure that environment is sandbox

NachoSoto:
For some reason the url is `[app_bundle]/StoreKit/receipt` in integration tests.
This is not correct, so I've added a couple of tests: one for entitlements, and another one to check the raw output value of that property.

This is necessary to make sure that receipts are loaded correctly after #1726.

### DIFF
--- a/Sources/Misc/SandboxEnvironmentDetector.swift
+++ b/Sources/Misc/SandboxEnvironmentDetector.swift
@@ -21,7 +21,7 @@ protocol SandboxEnvironmentDetector {
 }
 
 /// ``SandboxEnvironmentDetector`` that uses a `Bundle` to detect the environment
-final class DefaultSandboxEnvironmentDetector: SandboxEnvironmentDetector {
+final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
 
     private let bundle: Bundle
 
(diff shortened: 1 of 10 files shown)


Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -37,4 +37,11 @@ final class DefaultSandboxEnvironmentDetector: SandboxEnvironmentDetector {
         return url.path.contains("sandboxReceipt")
     }
 
+    #if DEBUG
+    // Mutable in tests so it can be overriden
+    static var `default`: SandboxEnvironmentDetector = BundleSandboxEnvironmentDetector()
+    #else
+    static let `default`: SandboxEnvironmentDetector = BundleSandboxEnvironmentDetector()
```
aboedo, line 44: `var` vs `let` here is clever! FWIW just having it be internal would likely have been enough for testing purposes, but this adds extra safety

aboedo (approved):

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -49,6 +49,10 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
         return .disabled
     }
 
+    func testIsSandbox() {
```
NachoSoto, line 52: I moved this here so it runs in SK1 and SK2 separately, in case that logic ever diverges (like in iOS 16 potentially?)

PR #1733, opened 2022-06-23
Replaced `CustomerInfo.nonSubscriptionTransactions` with a new non-`StoreTransaction` type

NachoSoto:
Fixes [CF-781]

Making `CustomerInfo.nonSubscriptionTransactions` return type a `StoreTransaction` was my mistake from an earlier refactor. I saw the two types were similar, so I made them conform to the same interface.
However, they represent different concepts.

### Changes:

- Added new `NonSubscriptionTransaction` type
- Deprecated `CustomerInfo.nonSubscriptionTransactions` with a `renamed` for a new `nonSubscriptions`
- Made the old `BackendParsedTransaction` initialization depend on the new type to avoid duplicating code
- Add new test to cover the new method, and left the existing one
- Added more checks in `CustomerInfoTest` for missing properties
- Updated all APITesters

[CF-781]: https://revenuecats.atlassian.net/browse/CF-781?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/TransactionsFactory.swift
+++ b/Sources/Purchasing/TransactionsFactory.swift
@@ -18,44 +18,14 @@ enum TransactionsFactory {
 
     static func nonSubscriptionTransactions(
         withSubscriptionsData subscriptionsData: [String: [CustomerInfoResponse.Transaction]]
-    ) -> [StoreTransaction] {
+    ) -> [NonSubscriptionTransaction] {
         subscriptionsData
-            .flatMap { (productID, transactions) -> [StoreTransaction] in
+            .flatMap { (productID, transactions) -> [NonSubscriptionTransaction] in
                 transactions
                     .lazy
-                    .compactMap { BackendParsedTransaction(with: $0, productID: productID) }
-                    .map { StoreTransaction($0) }
+                    .compactMap { NonSubscriptionTransaction(with: $0, productID: productID) }
             }
             .sorted { $0.purchaseDate < $1.purchaseDate }
-    }
-
-}
-
-/// `StoreTransactionType` backed by data parsed from the server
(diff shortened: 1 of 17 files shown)


NachoSoto:
Bump? 👀

aboedo:
adding @vegaro since he brought this up originally

aboedo:
why did that close it?? I think GH's having some issues right now, sorry, I'll reopen

aboedo (commented):
Looks great to me, a clean fix ✨ 
Let's wait for Cesar to take a look too before merging in case I'm missing something

Sources/Purchasing/TransactionsFactory.swift
```diff
@@ -18,44 +18,14 @@ enum TransactionsFactory {
...
 
 }
-
-/// `StoreTransactionType` backed by data parsed from the server
-private struct BackendParsedTransaction: StoreTransactionType {
-
-    let productIdentifier: String
-    let purchaseDate: Date
-    let transactionIdentifier: String
-    let quantity: Int
```
vegaro, line 40: Would it make sense to add this property to the new model? I see it's been set to 1, so I'm not sure it's coming from the backend

NachoSoto: It wasn't. We only had it so we could expose the SK2 version, but since it was a dummy value I thought not to until the backend sends it.

vegaro (approved):
Awesome, we should rename the Android model too (it's called `Transaction`). I created https://revenuecats.atlassian.net/browse/CSDK-308 to track that.

PR #1735, opened 2022-06-23
`DeviceCache`: added default `sandboxEnvironmentDetector` value

NachoSoto:
Turns out that `purchases-hybrid-common` relied on the previous constructor.
For backwards compatibility, we can provide this default value.


NachoSoto:
Oh I got this wrong never mind.

PR #1736, opened 2022-06-23
`FatalErrorUtil`: don't override `fatalError` on release builds

NachoSoto:
This hack is useful for tests, but since it's not needed in production, better to avoid compiling it and overriding Swift's `fatalError`.


aboedo (approved):
good call

PR #1742, opened 2022-06-28
`PurchasesOrchestrator`: don't log attribute errors if there are none

NachoSoto:
For an error like `ErrorCode.invalidReceiptError` there won't be any, so this results in an empty confusing log message.


aboedo (approved):
👌

PR #1743, opened 2022-06-28
Re-fetch cached offerings and products after Storefront changes (3/4) 

Juanpe:
Follow up to https://github.com/RevenueCat/purchases-ios/pull/1583. Subtask of https://github.com/RevenueCat/purchases-ios/issues/1382

### Description

This PR aims to invalidate and re-fetch if appropriate the stuff in the cache when the Storefront changes. I've used both `OfferingsManager` and `ProductsManager` to manage the logic of invalidate and re-fetch the data.

Changes:
- Add `invalidateAndReFetchCachedOfferingsIfAppropiate` method to `OfferingsManager` class.
- Rename `clearCachedProducts` method to `invalidateAndReFetchCachedProductsIfAppropiate` in `ProductsManager` class.
- Add `OfferingsManager` as a dependency of `PurchasesOrchestrator`.
- Clear cache methods returns the identifiers of the elements that have been removed or an empty list if there was no stuff in the cache.
- Add tests to cover some scenarios.


Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
```diff
@@ -186,9 +186,15 @@ extension ProductsFetcherSK1: SKProductsRequestDelegate {
         }
     }
 
-    func clearCache() {
+    /// - Returns (via callback): The product identifiers that were removed, or empty if there were not
+    ///   cached products.
+    func clearCache(completion: @escaping ((Set<String>) -> Void)) {
```
NachoSoto, line 191: Maybe provide a default value for the callback?
```suggestion
    func clearCache(completion: @escaping ((Set<String>) -> Void)? = nil) {
```

Sources/Purchasing/StoreKit2/ProductsFetcherSK2.swift
```diff
@@ -43,8 +43,14 @@ actor ProductsFetcherSK2 {
         }
     }
 
-    func clearCache() {
-        self.cachedProductsByIdentifier.removeAll(keepingCapacity: false)
+    /// - Returns: The product identifiers that were removed, or empty if there were not
+    ///   cached products.
+    func clearCache() -> Set<String> {
```
NachoSoto, line 48: Ditto:
```suggestion
    @discardableResult
    func clearCache() -> Set<String> {
```

Sources/Purchasing/StoreKit2/ProductsFetcherSK2.swift
```diff
@@ -43,8 +43,14 @@ actor ProductsFetcherSK2 {
...
+        let cachedProductIdentifiers = self.cachedProductsByIdentifier.keys
+        if !cachedProductIdentifiers.isEmpty {
+            self.cachedProductsByIdentifier.removeAll(keepingCapacity: false)
+        }
+        return Set(cachedProductIdentifiers)
```
NachoSoto, line 53: I was like "this is a race condition!". Then realized this is an `actor` 😄

Sources/Purchasing/ProductsManager.swift
```diff
@@ -82,14 +82,20 @@ class ProductsManager: NSObject {
...
 
-    func clearCachedProducts() {
+    func invalidateAndReFetchCachedProductsIfAppropiate() {
         if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *),
            self.systemInfo.storeKit2Setting == .enabledForCompatibleDevices {
             Task<Void, Never> {
-                await productsFetcherSK2.clearCache()
+                let removedProductIdentifiers = await productsFetcherSK2.clearCache()
+                if !removedProductIdentifiers.isEmpty {
+                    _ = try? await self.productsFetcherSK2.products(identifiers: removedProductIdentifiers)
```
NachoSoto, line 91: I think this makes sense 👍🏻 
Maybe we can add a debug log here? :)

Sources/Purchasing/ProductsManager.swift
```diff
@@ -82,14 +82,20 @@ class ProductsManager: NSObject {
...
+                }
             }
         } else {
-            productsFetcherSK1.clearCache()
+            productsFetcherSK1.clearCache { [weak self] removedProductIdentifiers in
+                guard !removedProductIdentifiers.isEmpty else { return }
+                self?.productsFetcherSK1.products(withIdentifiers: removedProductIdentifiers, completion: { _ in })
+            }
```
NachoSoto, line 98: I would just capture `productsFetcherSK1`. This completion block is `@escaping`, but it's short lived so it won't be a reference cycle:
```suggestion
            productsFetcherSK1.clearCache { [productsFetcherSK1] removedProductIdentifiers in
                guard !removedProductIdentifiers.isEmpty else { return }
                productsFetcherSK1.products(withIdentifiers: removedProductIdentifiers, completion: { _ in })
            }
```

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -94,6 +94,15 @@ class OfferingsManager {
         return productIDsFromBackend.subtracting(productIDsFromStore)
     }
 
+    func invalidateAndReFetchCachedOfferingsIfAppropiate(appUserID: String) {
+        let cachedOfferings = self.deviceCache.cachedOfferings
+        self.deviceCache.clearCachedOfferings()
+
+        if cachedOfferings != nil {
+            offerings(appUserID: appUserID, completion: { _ in })
```
NachoSoto, line 102:
```suggestion
            self.offerings(appUserID: appUserID, completion: { _ in })
```

NachoSoto (commented):
This looks great! Just a few suggestions.

Sources/Purchasing/OfferingsManager.swift
Juanpe: 👍🏼

Sources/Purchasing/ProductsManager.swift
Juanpe: much better 👍🏼

Sources/Purchasing/ProductsManager.swift
Juanpe: Do you mean only when the request fails or just to inform that the products have been re-fetched? 🤔

Juanpe:
> This looks great! Just a few suggestions.

Thanks for the feedback 👌🏼 I made the suggested changes. LMK if you find anything else that could be improved 👍🏼

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
```diff
@@ -186,9 +186,16 @@ extension ProductsFetcherSK1: SKProductsRequestDelegate {
...
-    func clearCache() {
+    /// - Returns (via callback): The product identifiers that were removed, or empty if there were not
+    ///   cached products.
+    func clearCache(completion: ((Set<String>) -> Void)? = nil) {
         self.queue.async {
-            self.cachedProductsByIdentifier.removeAll()
+            let cachedProductIdentifiers = self.cachedProductsByIdentifier.keys
+            if !cachedProductIdentifiers.isEmpty {
+                Logger.debug(Strings.offering.product_cache_invalid_for_storefront_change)
+                self.cachedProductsByIdentifier.removeAll()
```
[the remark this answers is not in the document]
Juanpe: Just to be sure, is there a possibility that there may not be the same number of products in different Storefronts?

NachoSoto:
There's a couple lint issues in `MockOfferingsManager`. Maybe add `switlint:ignore identifier_name`?

Juanpe:
> There's a couple lint issues in `MockOfferingsManager`. Maybe add `switlint:ignore identifier_name`?

Oups, you're right. I've disable this rule for these properties. ([9988fb5](https://github.com/RevenueCat/purchases-ios/pull/1743/commits/9988fb5e861f4b294345919df4ebea538373afd7))

BTW, my local copy of the project has 35 warnings related to the linter. Is it ok? I checked and I think I have the config files up to date 🤔 
Screenshot 2022-07-05 at 14 31 02

Tests/UnitTests/Mocks/MockOfferingsManager.swift
```diff
@@ -63,4 +63,18 @@ class MockOfferingsManager: OfferingsManager {
         completion?(stubbedUpdateOfferingsCompletionResult!)
     }
 
+    var invokedInvalidateAndReFetchCachedOfferingsIfAppropiate = false
+    var invokedInvalidateAndReFetchCachedOfferingsIfAppropiateCount = 0
+
+    // swiftlint:disable identifier_name
```
NachoSoto, line 69: This is just my technique, if you're disabling a rule next to the code that breaks it I'd either re-enable it afterwards or do `disable:next`. Otherwise I would move this to the top of the file.

NachoSoto (approved):
Weird. What version of SwiftLint do you have? I wonder if there a new rule that got added?

Tests/UnitTests/Mocks/MockOfferingsManager.swift
Juanpe: Makes sense.  I used this technique because this code is at the end of this file and as far as I know this disable only affects the code below.  However, I agree with you that we need to re-enable this rule after code because we may add more code in the future that would be affected by this disable.  Let me change it 👍🏼

Juanpe:
I made the change. Finally, I disable the `identifier_name` for the entire file. It's a mock and we did the same for other classes, like `MockTrialOrIntroPriceEligibilityChecker` or `MockSubscriberAttributesManager`. 

BTW, in order to make all the necessary changes...could you please review [the comment](https://github.com/RevenueCat/purchases-ios/pull/1743#discussion_r913146139) related to keep the capacity after cleaning the cache?

Thanks for the review and the feedback! 🤙

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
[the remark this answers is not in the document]
Juanpe: yep, well, `keepingCapacity: false` was already in SK2 and `false` is the default value, but I think it might be a good idea to explicitly state that we are completely clearing the storage 👍🏼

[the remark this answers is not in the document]
Juanpe: Done ✅ ([0f59e57](https://github.com/RevenueCat/purchases-ios/pull/1743/commits/0f59e57cd4d305e2ebf3fc967dc33d0825c3c55c))

PR #1744, opened 2022-06-30
`RELEASING.md`: fixed `bump` lane name

NachoSoto:
Looks like this was changed in #1719 but not updated.


tonidero (approved):
Ahh thanks for fixing that!

PR #1747, opened 2022-07-01
`fastlane release_checks`: removed call to `check_autogenerated_header_up_to_date`

NachoSoto:
This was added in #1719, but unfortunately [it fails when calling `ensure_git_status_clean`](https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/7379/workflows/f8feaf5d-42f7-4154-8782-9c2f1d36aa6b/jobs/30726).
So since the header is already automatically generated, we can just go back to the previous behavior and not call this.


tonidero (approved):
Hmm weird... Any idea what files are modified? I will investigate about it next week.

NachoSoto:
I'm not sure, it could be fastlane changing the README?

PR #1752, opened 2022-07-06
`bug_report.md`: clarify SK2 support

NachoSoto:
We've seen several reports where users say "yes" maybe because they assume that's the default.
This makes it more clear that it's only enabled if they're passing the flag.


tonidero (approved):
Makes sense!

PR #1754, opened 2022-07-08
`logErrorIfNeeded`: also log message if present

NachoSoto:
### Before:

For errors with a custom `message` it turns out that we were never logging it.
This was extra confusing for `OfferingsManager.Error.configurationError`, which ended up displaying this:

> 😿‼️ There is an issue with your configuration. Check the underlying error for more details.

With no underlying details.

### After:

> 😿‼️ There is an issue with your configuration. Check the underlying error for more details.
> There's a problem with your configuration. None of the products registered in the RevenueCat dashboard could be fetched from App Store Connect (or the StoreKit Configuration file if one is being used).
> More information: https://rev.cat/why-are-offerings-empty

### DIFF
--- a/Sources/Error Handling/ErrorUtils.swift
+++ b/Sources/Error Handling/ErrorUtils.swift
@@ -518,7 +518,7 @@ private extension ErrorUtils {
                                          fileName: String = #fileID,
                                          functionName: String = #function,
                                          line: UInt = #line) {
-        let message = code.description + (message.map { " " + $0 } ?? "")
+        let formattedMessage = code.description + (message.map { " " + $0 } ?? "")
 
         switch code {
         case .networkError,
@@ -544,7 +544,7 @@ private extension ErrorUtils {
                 .invalidPromotionalOfferError,
                 .offlineConnectionError:
                 Logger.error(
-                    message,
+                    formattedMessage,
                     fileName: fileName,
                     functionName: functionName,
                     line: line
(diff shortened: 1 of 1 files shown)


Sources/Error Handling/ErrorUtils.swift
```diff
@@ -513,9 +514,12 @@ private extension ErrorUtils {
 
     // swiftlint:disable:next function_body_length
     private static func logErrorIfNeeded(_ code: ErrorCode,
+                                         message: String?,
                                          fileName: String = #fileID,
                                          functionName: String = #function,
                                          line: UInt = #line) {
+        let message = code.description + (message.map { " " + $0 } ?? "")
```
tonidero, line 521: Hmm maybe rename this to `formattedMessage` or something like that? so we don't shadow the `message` parameter.

Sources/Error Handling/ErrorUtils.swift
```diff
@@ -471,6 +471,7 @@ private extension ErrorUtils {
 
         Self.logErrorIfNeeded(
             code,
+            message: message,
```
tonidero, line 474: Does the message include all the useful information or should we also include the `underlyingError` or any other info?

tonidero (approved):
Just a couple small comments, but looking good!

Sources/Error Handling/ErrorUtils.swift
NachoSoto: Good question. The underlying error is sometimes printed separately so it could be duplicated.

PR #1755, opened 2022-07-11
`EntitlementInfo`: added `isActiveInCurrentEnvironment` and `isActiveInAnyEnvironment`

NachoSoto:
A while ago, #1647 added these properties to `EntitlementInfos`: `activeInCurrentEnvironment` and `activeInAnyEnvironment`.
However, the only way to use them was by filtering the `EntitlementInfos`.

A common pattern is to use:
```swift
customerInfo.entitlements.all[entitlementIdentifier].isActive
```

So this adds a couple more properties to allow filtering using the `all` dictionary, directly from the `EntitlementInfo` itself.

For that, I moved the logic down one level, and made the existing methods rely on those new ones instead.
To test the new properties I simply added 2 new expectations to the shared methods used [in the existing tests](https://github.com/RevenueCat/purchases-ios/blob/829b69041da4bd53c5de3b1beb3034d8e8f2532a/Tests/UnitTests/Purchasing/EntitlementInfosTests.swift#L1116-LL1154).

### DIFF
--- a/Sources/Purchasing/EntitlementInfo.swift
+++ b/Sources/Purchasing/EntitlementInfo.swift
@@ -85,6 +85,10 @@ extension PeriodType: DefaultValueProvider {
 
     /**
      True if the user has access to this entitlement
+     - Warning: this is equivalent to ``isActiveInAnyEnvironment``
+
+     #### Related Symbols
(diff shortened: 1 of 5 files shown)


beylmk (approved):
🔥 love it

PR #1757, opened 2022-07-12
Plumb platformInfo in Configuration for PHC use

taquitos:
We aren't passing in flavory things to the Purchases init. This updates the Configuration builder to take in PlatformInfo from the hybrids, but keeps it as an internal detail that developers don't need to worry about.

resolves [CSDK-315]

[CSDK-315]: https://revenuecats.atlassian.net/browse/CSDK-315?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sources/Purchasing/Configuration.swift
```diff
@@ -164,6 +167,12 @@ import Foundation
             return self
         }
 
+        /// Set `platformInfo`. Internal details, no need to expose.
+        func with(platformInfo: Purchases.PlatformInfo) -> Builder {
```
NachoSoto, line 171: Maybe make this explicitly `internal` instead of putting that in the comment?

taquitos: Fair

PR #1760, opened 2022-07-12
Update Configuration.swift

taquitos:
internal protection level is not accessible to PurchasesHybridCommon through our Purchases extension.
Need to expose it publicly.

Double-checked locally with PurchasesHybridCommon pointing to local dev.

🤦‍♂️

### DIFF
--- a/Sources/Purchasing/Configuration.swift
+++ b/Sources/Purchasing/Configuration.swift
@@ -168,7 +168,7 @@ import Foundation
         }
 
         /// Set `platformInfo`.
-        internal func with(platformInfo: Purchases.PlatformInfo) -> Builder {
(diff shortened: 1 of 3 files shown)


NachoSoto:
We use other `internal` stuff, you could `@testable import RevenueCat`.

taquitos:
Great point about using `@testable import RevenueCat`

taquitos:
It was a nice idea, but it doesn't work without setting RevenueCat to compile for testability, too

taquitos:
After this merges, I can do a patch release for `4.8.1` unless we think flipping internal -> public on the Configuration change here warrants a minor.

NachoSoto (approved):
You'll need to add this to the APITesters

PR #1761, opened 2022-07-13
Configure Danger, enforce labels

vegaro:
- Adds Danger
- Adds a check for PR labels

This is how the check looks like:

Screen Shot 2022-08-05 at 9 01 39 AM

This PR initially added Jira link checks, but the lack of customization on the danger-jira plugin made me decide to drop the danger check. We should probably write our own/fork that plugin and add more options.

[CSDK-127]

[CSDK-127]: https://revenuecats.atlassian.net/browse/CSDK-127?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vegaro:
Screen Shot 2022-07-19 at 6 51 42 PM

For reference this is how it looks when Jira check fails

NachoSoto:
I'm not sure Jira links in commit titles is a good idea IMO, for several reasons:
- Open source repo, but external contributors don't have access to Jira.
- If in the future RC moves to another issue tracking system, that becomes outdated
- Commit messages are meant to help finding a change you might be looking for. The JIRA-123 part doesn't help with that.

I'm all however for including that in the commit message though for extra context, which can also be searched in git history. Just don't think that makes sense in a title.

vegaro:
That's great feedback @NachoSoto and I agree with you. 

I think the Danger plugin also checks if the description of the PR has the jira ticket, let me confirm. Do you think that would be better?

vegaro:
ok so I retested it @NachoSoto and it looks like it also looks for the JIRA issue in the PR body and the PR title. I think having it in the body is good right? If we have external contributors we can edit the PR body or the PR title ourselves too.

What do you think?

NachoSoto:
> ok so I retested it @NachoSoto and it looks like it also looks for the JIRA issue in the PR body and the PR title. I think having it in the body is good right? If we have external contributors we can edit the PR body or the PR title ourselves too.
> 
> 
> 
> What do you think?

That would be great!

vegaro:
I removed the changelog check stuff because we want to automate it in a better way :)

Dangerfile
```diff
@@ -0,0 +1,9 @@
+jira.check(
+  key: ["CSDK", "CF", "SDKONCALL"],
+  url: "https://revenuecats.atlassian.net/browse",
+  search_title: true,
+  search_commits: false,
+  fail_on_warning: false,
+  report_missing: true,
+  skippable: true # skippable by adding [no-jira] to PR title or body
```
taquitos, line 8: 🧡 like the skippable, but maybe also enforce a skip reason? Like if you use the [no-jira] we should have a description that makes it obvious why we skipped? like require `skip jira reason:`
Thoughts?

vegaro: I like it. The thing is that this is a plugin and it doesn't have the option. I would also like to change the warning message. Maybe I fork it?

taquitos: Yeah, it looks abandoned 😿 I think your best bet is a fork. I looked at the other forks, and nothing sticks out as one we should use.

vegaro: Yes... the goal of this sprint is to work on release trains so I might merge it like this and work on the fork in the future. We are going to still use Danger for the release trains (to check if PRs are label correctly) so this PR is still useful

vegaro:
I removed the danger-jira stuff since we would like to do some stuff the plugin doesn't support:
- Make it optional
- Change the error message

I think it's better to leave it for another sprint and focus on releases this sprint

vegaro:
I created a new bot [RevenueCat-Danger-Bot](https://github.com/RevenueCat-Danger-Bot) because it is recommended in https://danger.systems/guides/getting_started.html#oss-projects that the bots don't have access to the repositories and RCGitBot has access. 

I am not sure what the security concern is. I guess the problem is that the token could be leaked in CI. But in that case the RCGitBot token we have right now can also be leaked right?

.circleci/config.yml
```diff
@@ -534,3 +554,6 @@ workflows:
       - docs-deploy:
           xcode_version: '14.0.0'
           <<: *release-tags
+  danger:
+    jobs:
+      - danger
```
tonidero, line 559: Do we need to specify in what branches we need to run this? I guess we want every branch that has a PR associated? (Though not sure how to do that in circleci)

tonidero (commented):
Mostly a question and a comment but looking good.

.circleci/config.yml
vegaro: With the settings we have in CircleCI for purchases-ios it will work like that and this will run in every branch with a PR

Screen Shot 2022-08-05 at 2 02 35 PM

tonidero: Oh interesting... I don't think that's the case for all our repos, which might lead to inconsistencies. But it should be ok in this repo then 👍

tonidero (approved):

PR #1769, opened 2022-07-22
Split changelog in sections depending on PR label 

vegaro:
Modifies the github_changelog lane to look into PR labels so changelog can be automatically split in different sections. These labels can also be used to determine the next version number (patch, minor, major).


fastlane/CommonFastfile
```diff
@@ -121,7 +121,23 @@ lane :github_changelog do |options|
   # Temporary workaround to get around GitHub rate limit
   rate_limit_sleep = ENV["GITHUB_RATE_LIMIT_SLEEP"].to_i
 
-  formatted = commits.map do |commit|
+  supported_types = [:breaking, :build, :ci, :docs, :feat, :fix, :perf, :refactor, :style, :test]
```
vegaro, line 124: this is one of my first adventures with Ruby, so beware and please be critical

fastlane/CommonFastfile
```diff
@@ -121,7 +121,11 @@ lane :github_changelog do |options|
   # Temporary workaround to get around GitHub rate limit
   rate_limit_sleep = ENV["GITHUB_RATE_LIMIT_SLEEP"].to_i
 
-  formatted = commits.map do |commit|
+  supported_types = ["breaking", "build", "ci", "docs", "feat", "fix", "perf", "refactor", "style", "test"]
```
tonidero, line 124: It will be good to move this to the fastlane plugin once we start using it. 🎉 for shared code and tests!

fastlane/CommonFastfile
```diff
@@ -142,11 +146,43 @@ lane :github_changelog do |options|
       item = items.first
       message = "#{item['title']} (##{item['number']})"
       username = item["user"]["login"]
+      change_types = item["labels"].map do |label|
+        label_name = label["name"]
+        supported_types.include? label_name || "other"
```
tonidero, line 151: This might be wip, and I might be wrong, but I think this would map the labels to booleans? Since the last line within the `map` returns a boolean.

Another thing is that I prefer to use parentheses for most of these, and I think it's recommended according to rubocop.

vegaro: Oh I think this was working and I broke it in one of the last commits 🙈

fastlane/CommonFastfile
vegaro: yes! that's the idea :)

fastlane/CommonFastfile
vegaro: fixed

fastlane/CommonFastfile
```diff
@@ -142,11 +146,43 @@ lane :github_changelog do |options|
       item = items.first
       message = "#{item['title']} (##{item['number']})"
       username = item["user"]["login"]
+      change_types = item["labels"].map do |label|
+        label_name = label["name"]
+        (supported_types.include? label_name) ? label_name : "other"
+      end.uniq
```
NachoSoto, line 152: I'm no the biggest expert on Ruby. To make sure I follow this, looks like it will only include PRs with the labels in `supported_types`, but categorize the ones that aren't in `changelog_sections` as "other"?

fastlane/CommonFastfile
```diff
@@ -142,11 +146,43 @@ lane :github_changelog do |options|
       item = items.first
       message = "#{item['title']} (##{item['number']})"
       username = item["user"]["login"]
+      change_types = item["labels"].map do |label|
+        label_name = label["name"]
```
NachoSoto, line 150: If I'm following this correctly, does this mean that if a PR has multiple matching labels, it'll only apply it to the last label found?

NachoSoto (commented):
I think this looks good, just wondering what the behavior is with multiple matching labels.

fastlane/CommonFastfile
vegaro: This is more or less the logic:
- Get all supported labels. If not supported, the label will be equal to "other
- If there are multiple labels:
	- And one is "other". This can happen if it's labeled with a non-supported type, like for example "HOLD", "fix" and "breaking"
		- Remove "other" and leave the other supported types
	- If tagged with "breaking", it will go to the breaking changes section directly since it has preference. For example if "breaking" and "fix", it will go to the breaking changes section
	- If tagged with "feat" and "fix" (this is wrong tagging, it should be either), it will go to the new features section
- If tagged with something that's not "breaking", "fix" or "feat" it just goes directly to "Other changes" since I think having too many sections in the changelog can make it hard to read and most of the types (like "test", "docs", "build") can just be categorized as "Other changes"

What do you think? Some things that I am not sure about are: 
- Should we support multiple labels? I don't think the angular convention supports this (besides breaking + something else) and I think it might be overkill
- Do we enforce not tagging with multiple labels? I think we can do this Danger. Can be an improvement and not necessary for this PR
- Do we add more sections to the changelog?
- Should breaking changes be a section? Or something in the changelog lines, something like:
```
## New Features
- `*BREAKING* Removes StoreKit (#1771) via vegaro (@vegaro)` 
- `Adds StoreKit 3 option (#1772) via vegaro (@vegaro)` 
```
Instead of:

```
## Breaking Changes
- Removes StoreKit (#1771) via vegaro (@vegaro)` 
## New Features
- `Adds StoreKit 3 option (#1772) via vegaro (@vegaro)` 
```

@RevenueCat/core-sdk  I would love your input on this

tonidero (commented):
Just a couple of things but looking good

fastlane/CommonFastfile
```diff
@@ -142,11 +146,47 @@ lane :github_changelog do |options|
       item = items.first
       message = "#{item['title']} (##{item['number']})"
       username = item["user"]["login"]
+      change_types = item["labels"].map do |label|
```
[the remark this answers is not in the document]
vegaro: oh yes, I like that, much easier

fastlane/CommonFastfile
```diff
@@ -142,11 +146,45 @@ lane :github_changelog do |options|
       item = items.first
       message = "#{item['title']} (##{item['number']})"
       username = item["user"]["login"]
+
+      change_types = item["labels"].map{ |label_info| label_info["name"] }.select { |label| supported_types.include? label }.uniq
+
+      if change_types.size > 1
+        # If there are more than one label, and one is "other", remove "other" so the other one is used
+        change_types.delete_if { |n| n == "other" }
+      end
```
tonidero, line 155: This isn't needed anymore I think?

vegaro: good catch! I forgot to remove it :)

fastlane/CommonFastfile
NachoSoto: I think what you did makes a lot of sense 👍🏻  especially for now. If we run into other edge cases we can either fix them later, or just move a commit manually as well. This covers the majority of what we need I think :)

vegaro:
I am going to move the changes into the fastlane plugin with the common code

vegaro:
Closing since https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/10 adds this to our Fastlane plugin

PR #1770, opened 2022-07-25
Integration Tests / StoreKit tests: limit execution to 3 minutes

NachoSoto:
Reduced from the default of 10 minutes. If a test hasn't passed by then, it's probably stuck.


taquitos (approved):
I was thinking about this last week, totally agree!

NachoSoto:
Actually let me do the same for regular tests too.

NachoSoto:
Changed. Mostly important for `StoreKitUnitTests` which can get stuck due to `SKTestSession` bugs.

taquitos (approved):

PR #1771, opened 2022-07-25
Integration Tests: store result bundle

NachoSoto:
Equivalent to https://github.com/RevenueCat/purchases-hybrid-common/pull/181

This allows us to get more details on the failures. I'm planning to do the same for regular iOS tests as well, which will allow us to look at snapshot test failures, etc.

PR #1773, opened 2022-07-27
CircleCI: always store compressed result bundle

NachoSoto:
Follow up to #1771 for all test jobs. This allows us download the `.xcresult` file so we can look at all the test results in detail, including snapshot failures.

image


taquitos (approved):
WinZip

PR #1776, opened 2022-07-27
Error logging: `logErrorIfNeeded` no longer prints message if it's the same as the error description

NachoSoto:
Working on [SDKONCALL-62] I realized that there are many instances where we end up logging the same message twice.

This PR not only improves the logic, but adds much needed test coverage to `ErrorUtils.logErrorIfNeeded` and error logging in general.

[SDKONCALL-62]: https://revenuecats.atlassian.net/browse/SDKONCALL-62?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Tests/UnitTests/Purchasing/ErrorCodeTests.swift
```diff
@@ -181,3 +181,100 @@ class ErrorCodeTests: TestCase {
...
+
+    override func tearDown() {
+        Logger.logHandler = self.originalLogHandler
+
+        super.tearDown()
+    }
+
+    func testPurchaseErrorsAreLoggedAsApppleErrors() {
+        let underlyingError = NSError(domain: SKErrorDomain, code: SKError.Code.paymentInvalid.rawValue)
+        let error = ErrorUtils.purchaseNotAllowedError(error: underlyingError as Error)
```
NachoSoto, line 209:
```suggestion
        let error = ErrorUtils.purchaseNotAllowedError(error: underlyingError)
```

Tests/UnitTests/Purchasing/ErrorCodeTests.swift
```diff
@@ -181,3 +181,100 @@ class ErrorCodeTests: TestCase {
     }
 
 }
+
+import StoreKit
+
+class ErrorUtilsTests: TestCase {
```
tonidero, line 187: Hmm would it be better moving these to their own file?

Tests/UnitTests/Purchasing/ErrorCodeTests.swift
```diff
@@ -181,3 +181,100 @@ class ErrorCodeTests: TestCase {
...
+
+    private var originalLogHandler: VerboseLogHandler!
+    private var loggedMessages: [(level: LogLevel, message: String)] = []
+
+    override func setUp() {
+        super.setUp()
+
+        self.originalLogHandler = Logger.logHandler
+        Logger.logHandler = { level, message, _, _, _ in
+            self.loggedMessages.append((level, message))
```
tonidero, line 197: I think it might be ok for the tests, but should we use a `weak self` here?  (Been a while since I did some iOS development)

Tests/UnitTests/Purchasing/ErrorCodeTests.swift
```diff
@@ -181,3 +181,100 @@ class ErrorCodeTests: TestCase {
...
+    ) {
+        expect(
+            file: file,
+            line: line,
+            self.loggedMessages
+        ).to(
+            containElementSatisfying { level, message in
+                (level == .error &&
+                 message.contains(intent.prefix) &&
+                 message.contains(error.localizedDescription))
```
tonidero, line 274: Hmm NABD but I wonder if this logic is too complex to have in the tests... I think we should just match the test error messages exactly as they are if possible, even if the test fails when the error message changes.

tonidero (approved):
A few questions but LGTM!

Tests/UnitTests/Purchasing/ErrorCodeTests.swift
NachoSoto: Oh yeah let me do that.

Tests/UnitTests/Purchasing/ErrorCodeTests.swift
NachoSoto: The log handler is restored, so the cycle is broken, but it doesn't hurt 👍🏻

Tests/UnitTests/Purchasing/ErrorCodeTests.swift
NachoSoto: I did that in the last test I wrote because it was better, so I agree. Let me do that here too.

Tests/UnitTests/Purchasing/ErrorCodeTests.swift
NachoSoto: Well there's no cycle, but we probably don't want to make the log handler have a strong reference to the test instance.

Tests/UnitTests/Purchasing/ErrorCodeTests.swift
NachoSoto: Done

Tests/UnitTests/Purchasing/ErrorCodeTests.swift
NachoSoto: Done

Tests/UnitTests/Purchasing/ErrorCodeTests.swift
NachoSoto: Done

PR #1777, opened 2022-07-27
fix another broken link in docC docs

aboedo:
Fix another broken link. Again, not sure why we're not getting a warning


NachoSoto (approved):
Does this not work only when CMD+clicking in Xcode or is it broken in the generated docs too?

aboedo:
it looks like it works correctly in the generated docs, just not on cmd click

aboedo:
which does explain why we're not getting a warning

PR #1387, opened 2022-03-11
`HTTPClient.perform`: changed response to `Result<[String: Any], Error>`

NachoSoto:
Currently pretty much all closures in the codebase use `(Value?, Error?)`, which is a leftover from Objective-C days. These can represent 4 different cases:
- **Value / no-error**
- **No-value / error**
- _Value / error_
- _No-value / no-error_

Out of those 4, only 2 are really valid, which meant that a lot of code was spent handling the other 2, either creating errors (like `UnexpectedBackendResponseSubErrorCode.loginMissingResponse`) that weren't possible, or hard to follow code because both a value and an error were possible.

In tests, too, the mocks were initially set up to produce no value and no error, which were in many cases triggering code paths that weren't legal with real code.

This change removes this possibility all together by ensuring **at compile time**, that one of the two is always present.

This is step 1 in a series of refactors.
- Step 2 will move _a lot of duplicated code_ that handles status codes != 20x, error handling, parsing error responses, etc. up to `HTTPClient` and out of the operations, making the operations a lot simpler.
- Step 3 will change `HTTPClient` to return a `Response: Decodable` instead of `[String: Any]`, moving more of the response serialization logic to `HTTPClient`, and leaving the operations to only focus on the "happy path".

PR #1778, opened 2022-07-27
`CustomerInfoResponseHandler`: return `CustomerInfo` instead of error if the response was successful

NachoSoto:
Fixes [SDKONCALL-62].
The behavior in version 3.x was that we would return a `CustomerInfo` AND the attribute errors. [The test in that version](https://github.com/RevenueCat/purchases-ios/blob/3.14.2/PurchasesTests/SubscriberAttributes/BackendSubscriberAttributesTests.swift#L370-L404) however only verified that the error was returned.

When we converted `(Value, Error)` callbacks to using `Result` (#1387), I assumed this was an error. Because of that incomplete test, we didn't notice this wasn't the desired behavior.

If the backend returns a 200, it's assumed that the errors aren't more important than the correct `CustomerInfo`.

The new implementation simply logs the attribute errors (which is now tested), and returns the `CustomerInfo`:
```
😿‼️ One or more of the attributes sent could not be saved. ["$email": "email is not in valid format"]
```

[SDKONCALL-62]: https://revenuecats.atlassian.net/browse/SDKONCALL-62?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Tests/UnitTests/SubscriberAttributes/BackendSubscriberAttributesTests.swift
```diff
@@ -142,49 +142,104 @@ class BackendSubscriberAttributesTests: TestCase {
...
 
         let subscriberAttributesByKey: [String: SubscriberAttribute] = [
             subscriberAttribute1.key: subscriberAttribute1,
             subscriberAttribute2.key: subscriberAttribute2
         ]
-        var receivedError: BackendError?
+
+        var loggedMessages = [String]()
+        let originalLogHandler = Logger.logHandler
+        defer { Logger.logHandler = originalLogHandler }
```
tonidero, line 174: 😮 Didn't know about this. Interesting!

Sources/Networking/HTTPClient/HTTPResponse.swift
```diff
@@ -86,7 +86,7 @@ extension ErrorResponse {
 
         return ErrorUtils.backendError(
             withBackendCode: self.code,
-            backendMessage: self.message,
+            message: self.attributeErrors.description,
```
tonidero, line 89: Hmm are we still using the `backendMessage` parameter anywhere? I couldn't find a usage of it but I might be wrong.

Tests/UnitTests/SubscriberAttributes/__Snapshots__/BackendSubscriberAttributesTests/iOS14-testPostReceiptWithSubscriberAttributesPassesErrorIfStatusCodeIsNotSuccess.1.json
```diff
@@ -0,0 +1,29 @@
+{
+  "headers" : {
```
tonidero, line 2: Just to make sure, these files are snapshots of the requests done during the unit tests. Then the tests check that the request remains the same, correct? This is nice!

tonidero (approved):
LGTM!

Tests/UnitTests/SubscriberAttributes/__Snapshots__/BackendSubscriberAttributesTests/iOS14-testPostReceiptWithSubscriberAttributesPassesErrorIfStatusCodeIsNotSuccess.1.json
NachoSoto: Yup exactly :) without needing to check each thing individually.

Sources/Networking/HTTPClient/HTTPResponse.swift
NachoSoto: Oh good catch! I changed this before I fixed #1776 and forgot to add it back. Thanks for the thorough review 👏🏻

taquitos (commented):
Ugh, this review was sitting in draft.

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -20,12 +20,15 @@ class CustomerInfoResponseHandler {
...
-                if $0.body.errorResponse.attributeErrors.isEmpty {
-                    return .success($0.body.customerInfo)
-                } else {
-                    return .failure(.errorResponse($0.body.errorResponse, $0.statusCode))
+            .map {
+                // If the response was successful we always want to return the `CustomerInfo`.
+                if !$0.body.errorResponse.attributeErrors.isEmpty {
+                    // If there are any, log attribute errors.
+                    // Creating the error implicitly logs it.
+                    _ = $0.body.errorResponse.asBackendError(with: $0.statusCode)
```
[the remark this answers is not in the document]
NachoSoto: I agree. I didn't feel _too_ bad about this instance because it's covered by a test, so removing it would make it fail.
I don't love the implicit logging, but I think explicit logging of errors everywhere would be even worse :/ so I'm not sure how to improve this at the moment.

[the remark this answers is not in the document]
taquitos: Yeah, that's a great point, as long as it's covered by a test, this is much more palatable 😄

Issue #1765, opened 2022-07-18
[SANDBOX] StoreKit config not working when using Xcode 14/iOS 16 beta 3


ZeeWanderer:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Development/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
StoreKit configuration file stopped working on XCode 14 beta 3 + iOS 16 beta 3, the same configuration works on XCode 13.4.1 + IOS 15.5. I tried resaving public testing certificate with beta xcode but it produces the same file. I also tried recreating Config file with Xcode 14. Logs attached, thou logs for ios15.5+Xcode13.4.1 are from a simulator as rn i don't have a real device with 15.X to produce the log.

1. Environment
   1. Platform: iOS
   2. SDK version: 4.9.0
   3. StoreKit 2 (_enabled with `useStoreKit2IfEnabled`_) (Y/N): N
   4. OS version: 16 beta 3
   5. Xcode version: 14 beta 3
   6. How widespread is the issue. Percentage of devices affected.
2. [RC_ios15.5+Xcode13.4.1.log](https://github.com/RevenueCat/purchases-ios/files/9132176/RC_ios15.5%2BXcode13.4.1.log)  [RC_ios16.b3+Xcode14.b3.log](https://github.com/RevenueCat/purchases-ios/files/9132335/RC_ios16.b3%2BXcode14.b3.log)
3. 
4. 

**Additional context**

**EDIT:**
If StoreKit 2 is enabled the config works fine
log: [RC_ios16.b3+Xcode14.b3+SK2.log](https://github.com/RevenueCat/purchases-ios/files/9132398/RC_ios16.b3%2BXcode14.b3%2BSK2.log)

beylmk:
Hey @ZeeWanderer, thanks for the detailed issue. It's definitely possible this was broken in the latest xcode beta release. I'll try it out myself, but need to download beta 3 first 😅 will report back here once done!

beylmk:
[...]

PR #1780, opened 2022-07-28
Add missing logs to ProductsFetcherSK2

beylmk:
While investigating https://github.com/RevenueCat/purchases-ios/issues/1765, I noticed logs between the `ProductsFetcherSK1` and `ProductsFetcherSK2` don't match, causing confusion. This PR adds some missing logs to `ProductsFetcherSK2`


Sources/Purchasing/StoreKit2/ProductsFetcherSK2.swift
```diff
@@ -29,10 +29,17 @@ actor ProductsFetcherSK2 {
     func products(identifiers: Set<String>) async throws -> Set<SK2StoreProduct> {
         do {
             if let cachedProducts = await self.cachedProducts(withIdentifiers: identifiers) {
+                let cachedIdentifiers = Set(cachedProducts.map { $0.productIdentifier })
+                Logger.debug(Strings.offering.products_already_cached(identifiers: cachedIdentifiers))
                 return cachedProducts
             }
+            
+            Logger.debug(
+                Strings.offering.no_cached_requests_and_products_starting_skproduct_request(identifiers: identifiers)
```
beylmk, line 38: this one doesn't perfectly translate, since we don't cache requests for sk2....maybe we should?

NachoSoto: > maybe we should?

It's not trivial with `async`, I think that's why we don't do it.

The "no cached requests" part of the message isn't a big deal, but the latter part of the message would be weird here:
> starting `SKProducts` request for

Since that's `SK1`. So maybe it makes sense to add a new string?

Sources/Purchasing/StoreKit2/ProductsFetcherSK2.swift
```diff
@@ -29,10 +29,17 @@ actor ProductsFetcherSK2 {
...
+                Logger.debug(Strings.offering.products_already_cached(identifiers: cachedIdentifiers))
                 return cachedProducts
             }
 
+            Logger.debug(
+                Strings.offering.no_cached_requests_and_products_starting_skproduct_request(identifiers: identifiers)
+            )
+
             let storeKitProducts = try await StoreKit.Product.products(for: identifiers)
+            Logger.rcSuccess(Strings.storeKit.skproductsrequest_received_response)
```
NachoSoto, line 42: Ditto here. Alternatively we could change these strings to be agnostic of the types? So something like "product request received response", etc.

NachoSoto (commented):
Thanks for doing this!

Sources/Purchasing/StoreKit2/ProductsFetcherSK2.swift
beylmk: well we do have a completion handler up the call chain in `ProductsManager.sk2Products`:

` 
func sk2Products(withIdentifiers identifiers: Set<String>,
                     completion: @escaping (Result<Set<SK2StoreProduct>, Error>) -> Void) 
`

...so we could theoretically hold a map of identifiers -> handlers like `ProductsFetcherSK1` does 🤷

Sources/Purchasing/ProductsManager.swift
```diff
@@ -119,9 +121,11 @@ private extension ProductsManager {
             productsFetcherSK1.products(withIdentifiers: removedProductIdentifiers, completion: { result in
                 switch result {
                 case.success:
-                    Logger.debug(Strings.storeKit.skproductsrequest_finished)
+                    // this one is duplicate, though other in ProductsFetcherSK1 is rcSuccess
```
beylmk, line 124: options i see are:
*. remove duplication -- need to look into hierarchy of logging, when would rcSuccess happen vs debug?
* ensure sk2 has the same duplication

beylmk:
@NachoSoto comments should be addressed, except for a couple open conversations... curious what you think about:
* caching the callbacks for sk2 products requests
* those log duplications...left [comments](https://github.com/RevenueCat/purchases-ios/pull/1780#discussion_r932499046) for discussion

Sources/Logging/Strings/StoreKitStrings.swift
```diff
@@ -46,20 +48,20 @@ extension StoreKitStrings: CustomStringConvertible {
...
 
         case .skproductsrequest_timed_out(let afterTimeInSeconds):
             return "SKProductsRequest took longer than \(afterTimeInSeconds) seconds, " +
             "cancelling request and returning an empty set. This seems to be an App Store quirk. " +
             "If this is happening to you consistently, you might want to try using a new Sandbox account. " +
             "More information: https://rev.cat/skproductsrequest-hangs"
 
-        case .skproductsrequest_finished:
+        case .store_product_request_did_finish:
             return "SKProductsRequest did finish"
```
NachoSoto, line 61: This can change as well:
```suggestion
            return "Store product request did finish"
```

Sources/Logging/Strings/StoreKitStrings.swift
```diff
@@ -77,6 +79,10 @@ extension StoreKitStrings: CustomStringConvertible {
         case .sk1_discount_missing_locale:
             return "There is an issue with the App Store, this SKProductDiscount is missing a Locale - " +
             "The current device Locale will be used instead."
+
+        case .no_cached_requests_and_products_starting_skproduct_request(let identifiers):
+            return "No existing requests and " +
+                "products not cached, starting SKProducts request for: \(identifiers)"
```
NachoSoto, line 85: Ditto, you're still using this in SK2, maybe change it to `Store product request`?

Sources/Purchasing/ProductsManager.swift
NachoSoto: I don't think we've looked into these logs very much in detail, so this refactor is great 👍🏻 

`rcSuccess` is just a success `LogIntent`, but still `LogLevel.debug`. I think that's good for these success cases.

Sources/Purchasing/ProductsManager.swift
```diff
@@ -119,9 +121,11 @@ private extension ProductsManager {
...
+                    Logger.debug(Strings.storeKit.store_product_request_did_finish)
                 case .failure(let error):
-                    Logger.debug(Strings.storeKit.skproductsrequest_failed(error: error))
+                    // this one is duplicate, though other in ProductsFetcherSK1 is appleError
```
NachoSoto, line 127: Yeah I'd say this one should be `appleError` too.

Sources/Purchasing/ProductsManager.swift
```diff
@@ -119,9 +121,11 @@ private extension ProductsManager {
             productsFetcherSK1.products(withIdentifiers: removedProductIdentifiers, completion: { result in
                 switch result {
                 case.success:
```
NachoSoto, line 123: Hun looks like this is missing a space?
```suggestion
                case .success:
```

NachoSoto (commented):
I think this is good, looks like the only 2 things left:
- Don't mention `SKProduct` in SK2
- Consistent debug / success / error calls.

Sources/Logging/Strings/StoreKitStrings.swift
beylmk: was leaving this one in case we decided to cache requests... might be worth a different string until we decide to cache

NachoSoto: We cache products, just not concurrent requests, which might be good enough?

Sources/Logging/Strings/StoreKitStrings.swift
```diff
@@ -46,20 +48,20 @@ extension StoreKitStrings: CustomStringConvertible {
         case .skrequest_failed(let error):
             return "SKRequest failed: \(error.localizedDescription)"
 
-        case .skproductsrequest_failed(let error):
-            return "SKProductsRequest failed! error: \(error.localizedDescription)"
+        case .store_products_request_failed(let error):
+            return "Store products request failed! error: \(error.localizedDescription)"
```
NachoSoto, line 52: Nit: 
```suggestion
            return "Store products request failed! Error: \(error.localizedDescription)"
```

Sources/Purchasing/ProductsManager.swift
```diff
@@ -134,9 +129,10 @@ private extension ProductsManager {
             if !removedProductIdentifiers.isEmpty {
                 do {
                     _ = try await self.productsFetcherSK2.products(identifiers: removedProductIdentifiers)
-                    Logger.debug(Strings.storeKit.skproductsrequest_finished)
+
+                    Logger.debug(Strings.storeKit.store_product_request_did_finish)
                 } catch {
-                    Logger.debug(Strings.storeKit.skproductsrequest_failed(error: error))
+                    Logger.debug(Strings.storeKit.store_products_request_failed(error: error))
```
NachoSoto, line 135: Nit: For consistency I'd keep the "finished" wording, or change this to `did_fail` as well.

NachoSoto (approved):

Sources/Purchasing/ProductsManager.swift
beylmk: sorry, i'm lost, which finished wording?

NachoSoto: `store_products_request_failed` vs `store_product_request_did_finish` vs `store_product_request_received_response`.
I would use "did {fail/finish/receive}" for all, or the past tense, for consistency.

Just mentioning that since you changed `skproductsrequest_finished` to `store_product_request_did_finish`

beylmk: oh got it, i was looking for "finished" in that string you commented on. monday brain, thanks

PR #1781, opened 2022-07-28
`Logger`: wrap `message` in `@autoclosure` to avoid creating when `LogLevel` is disabled

NachoSoto:
I realized in #1780 that we might be doing unnecessary work creating messages that are never logged. This is a small optimization to prevent that.


aboedo (approved):
nice

PR #1784, opened 2022-07-29
`OSVersionEquivalent.current`: evaluate only once

NachoSoto:
This was a getter which meant that it was being evaluated unnecessarily for every mock network request. Now it's only set once.


taquitos (approved):
🐙🐐
download-61

PR #1800, opened 2022-08-08
Release/4.10.0

joshdholtz:
## New Features
* New AdServices Integration (#1727) via Josh Holtz (@joshdholtz)
## Bugfixes
* `OfferingsManager`: expose underlying error when `ProductsManager` returns an error (#1792) via NachoSoto (@NachoSoto)
* Add missing logs to ProductsFetcherSK2 (#1780) via beylmk (@beylmk)
## Other Changes
* AdServices: Fix failing tests on main in iOS 12 and 13 - IOSAttributionPosterTests (#1797) via Josh Holtz (@joshdholtz)
* Invalidates gem caches and separates danger and macOS caches (#1798) via Cesar de la Vega (@vegaro)
* Pass CircleCI branch to prepare_next_version job (#1796) via Toni Rico (@tonidero)
* Configure Danger, enforce labels (#1761) via Cesar de la Vega (@vegaro)
* Support for new fastlane internal plugin for automation (#1779) via Toni Rico (@tonidero)

PR #1803, opened 2022-08-09
Replace usages of replace_in with replace_text_in_files action

tonidero:
### Description
This was fixed by @joshdholtz in #1800. However, we created an action to replace text in a list of files so this changes that to use that action.


taquitos (approved):
📑 🐐

PR #1804, opened 2022-08-09
`Atomic: ExpressibleByNilLiteral`

NachoSoto:
Small simplification to make initializing `Atomic<Optional>` easier to create.

### DIFF
--- a/Sources/Misc/Atomic.swift
+++ b/Sources/Misc/Atomic.swift
@@ -74,6 +74,8 @@ internal final class Atomic<T> {
 
 }
 
+// Syntactic sugar that allows initializing an `Atomic` optional value by directly assigning `nil`,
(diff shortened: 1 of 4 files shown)


Sources/Misc/Atomic.swift
```diff
@@ -73,3 +73,11 @@ internal final class Atomic<T> {
     }
 
 }
+
+extension Atomic: ExpressibleByNilLiteral where T: OptionalType {
```
aboedo, line 77: this may be somewhat redundant, but I think it'd be nice to add a comment like: 
```swift
// this is syntactic sugar that allows us to initialize an optional Atomic value to by directly assigning nil, 
// i.e.: let foo : Atomic<Foo?> = nil instead of the more indirect let foo : Atomic<Foo?> = .init(nil)
```

aboedo (approved):

Sources/Misc/Atomic.swift
NachoSoto: Great idea, added!

PR #1806, opened 2022-08-09
`PriceFormatterProvider: Sendable` conformance and fixed thread-safety

NachoSoto:
For [CSDK-379].

This class was used all over the place **but it wasn't thread-safe**. The Swift compiler enforces this now that it conforms to Sendable (note it's not unchecked!).
Because `Atomic` guarantees correctness, this thread-safety can be enforced now by the compiler.

[CSDK-379]: https://revenuecats.atlassian.net/browse/CSDK-379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Misc/Lock.swift
+++ b/Sources/Misc/Lock.swift
@@ -13,7 +13,7 @@
 
 import Foundation
 
-internal final class Lock {
+internal final class Lock: Sendable {
 
     private let recursiveLock: NSRecursiveLock = {
         let lock = NSRecursiveLock()
--- a/Sources/Misc/Lock.swift
+++ b/Sources/Misc/Lock.swift
@@ -13,7 +13,7 @@
 
 import Foundation
 
-internal final class Lock: Sendable {
+internal final class Lock {
 
     private let recursiveLock: NSRecursiveLock = {
         let lock = NSRecursiveLock()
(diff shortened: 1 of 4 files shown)


Sources/Misc/Lock.swift
```diff
@@ -13,7 +13,7 @@
 
 import Foundation
 
-internal final class Lock {
+internal final class Lock: Sendable {
 
     private let recursiveLock: NSRecursiveLock = {
```
NachoSoto, line 18: `NSRecursiveLock` isn't `Sendable` before iOS 16. I'll fix that.

Sources/Misc/Lock.swift
```diff
@@ -31,3 +31,10 @@ internal final class Lock {
     }
 
 }
+
+#if swift(>=5.7)
+extension Lock: Sendable {}
+#else
+// `NSRecursiveLock` isn't `Sendable` until iOS 16.0 / Swift 5.7
```
aboedo, line 38: 👍

aboedo (approved):

PR #1808, opened 2022-08-09
Moved `AttributionFetcherError` inside `AttributionFetcher`

NachoSoto:
This removes this unnecessary type in the global namespace.
Also removed 2 `cases` introduced in #1727 that aren't actually used.

### DIFF
--- a/Sources/Attribution/AttributionFetcher.swift
+++ b/Sources/Attribution/AttributionFetcher.swift
@@ -23,15 +23,6 @@ import WatchKit
 import AdServices
 #endif
 
-enum AttributionFetcherError: Error {
-
-    case identifierForAdvertiserUnavailableForPlatform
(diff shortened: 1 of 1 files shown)


Sources/Attribution/AttributionFetcher.swift
```diff
@@ -23,15 +23,6 @@ import WatchKit
 import AdServices
 #endif
 
-enum AttributionFetcherError: Error {
-
-    case identifierForAdvertiserUnavailableForPlatform
-    case identifierForAdvertiserFrameworksUnavailable
-    case adServicesNotAvailable
-    case adServicesTokenFetchError
```
NachoSoto, line 31: I removed these 2 cause they aren't used, but it was harder to see cause this was an `internal` type

aboedo (approved):

PR #1810, opened 2022-08-09
`Dangerfile`: added "next_release" as supported label

NachoSoto:
See also https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/17

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,5 +1,5 @@
 def fail_if_no_supported_label_found
-  supported_types = ["breaking", "build", "ci", "docs", "feat", "fix", "perf", "refactor", "style", "test"]
+  supported_types = ["breaking", "build", "ci", "docs", "feat", "fix", "perf", "refactor", "style", "test", "next_release"]
 
   supported_labels_in_pr = supported_types & github.pr_labels
   no_supported_label = supported_labels_in_pr.empty?


Dangerfile
```diff
@@ -1,5 +1,5 @@
 def fail_if_no_supported_label_found
-  supported_types = ["breaking", "build", "ci", "docs", "feat", "fix", "perf", "refactor", "style", "test"]
+  supported_types = ["breaking", "build", "ci", "docs", "feat", "fix", "perf", "refactor", "style", "test", "next_release"]
```
tonidero, line 2: Hmm I wonder if we should rename to `release`... I'm guessing there will be 2 PRs usually tagged with this, the "bump" PR which is the actual release, and the "prepare next version" PR, bumping to the next snapshot.

tonidero (approved):

Dangerfile
NachoSoto: I don't have a strong preference, I went with `next_release` because that's a label that already exists (and I think it's default in GitHub?)

tonidero: Yeah me neither... I just didn't know if the "next" was necessary... But don't have a strong opinion so we can leave as is

vegaro (approved):

PR #1814, opened 2022-08-10
PurchaseTester- Update Podfile.lock

taquitos:


aboedo:
we should probably remove the file from versioning altogether, since it's kinda always lagging behind and we don't ship PurchaseTester to a store or anything

aboedo (approved):

aboedo:
not needed for this PR though

PR #1815, opened 2022-08-12
Adds an approval job that will tag the release

vegaro:
Creates an approval job in CircleCi on release branches. After the job is approved, a tag will be created in the release branch and the release process will begin.

There's a change of process compared to the current process. The tag has to be performed in the release branch and not in main. This is to prevent shipping extra commits that might have been merged to `main` after the release branch was created.

[CSDK-136]

[CSDK-136]: https://revenuecats.atlassian.net/browse/CSDK-136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


fastlane/Fastfile
```diff
@@ -223,6 +223,17 @@ platform :ios do
     github_release(version: version_number)
   end
 
+  desc "Tag current branch with current version number"
+  lane :tag_current_branch do |options|
+    version_number = current_version_number
+    add_git_tag(
```
tonidero, line 229: Should we maybe check if the tag already exists not do anything? We could delete the tag and retag, but it's dangerous since we might be releasing the same version multiple times, so I suggest the first option.

.circleci/config.yml
```diff
@@ -546,6 +572,13 @@ workflows:
           <<: *release-tags-and-branches
   deploy:
     jobs:
+      - hold:
+          type: approval
```
tonidero, line 576: One question, if there are multiple commits in the release branch, will it launch a new job for each one and cancel the previous one?

tonidero (approved):
LGTM!

PR #1817, opened 2022-08-12
Update documentation for presentCodeRedemptionSheet

taquitos:
Clarifies why it's not available for macCatalyst
improves https://github.com/RevenueCat/react-native-purchases/issues/401

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1228,7 +1228,7 @@ public extension Purchases {
 
 #if os(iOS)
 
-     /**
+    /**
      * Displays a sheet that enables users to redeem subscription offer codes that you generated in App Store Connect.
      *
      * - Important: Even though the docs in `SKPaymentQueue.presentCodeRedemptionSheet`
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1229,6 +1229,13 @@ public extension Purchases {
 #if os(iOS)
     /**
      * Displays a sheet that enables users to redeem subscription offer codes that you generated in App Store Connect.
+     *
+     * - Note: Even though the docs in `SKPaymentQueue.presentCodeRedemptionSheet`
+     * say that it's available on Catalyst 14.0, there is a note:
+     *
+     * `This function doesn’t affect Mac apps built with Mac Catalyst.`
+     *
+     * when, in fact, it crashes when called both from Catalyst and also when running as "Designed for iPad".
      */
     @available(iOS 14.0, *)
     @available(watchOS, unavailable)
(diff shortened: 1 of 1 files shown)


NachoSoto (approved):

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1229,6 +1229,13 @@ public extension Purchases {
 #if os(iOS)
     /**
      * Displays a sheet that enables users to redeem subscription offer codes that you generated in App Store Connect.
+     *
+     * - Note: Even though the docs in `SKPaymentQueue.presentCodeRedemptionSheet`
```
aboedo, line 1233: There's syntax to render this as a warning, can't recall if it's `- Warn:` or `- Warning:` or what it is, but it's more appropriate for this case since it's potentially app-breaking

aboedo: I sent some notes over slack about fixing the react-native issue

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1229,6 +1229,13 @@ public extension Purchases {
 #if os(iOS)
     /**
      * Displays a sheet that enables users to redeem subscription offer codes that you generated in App Store Connect.
+     *
+     * - Warning: Even though the docs in `SKPaymentQueue.presentCodeRedemptionSheet`
+     * say that it's available on Catalyst 14.0, there is a note:
+     *
+     * `This function doesn’t affect Mac apps built with Mac Catalyst.`
+     *
+     * when, in fact, it crashes when called both from Catalyst and also when running as "Designed for iPad".
```
aboedo, line 1238: This is broken when rendering:
image

A couple of suggestions: 
- Using `important` instead of warning since the app won't build in catalyst because we're explicitly making it unavailable to protect from this runtime crash
- packing lines together so they become a part of the same block
- Adding quotes so it looks a little clearer

```suggestion
     /**
 * Displays a sheet that enables users to redeem subscription offer codes that you generated in App Store Connect.
 *
 * - Important: Even though the docs in `SKPaymentQueue.presentCodeRedemptionSheet`
 * say that it's available on Catalyst 14.0, there is a note:
 * "`This function doesn’t affect Mac apps built with Mac Catalyst.`"
 * when, in fact, it crashes when called both from Catalyst and also when running as "Designed for iPad".
 * This is why RevenueCat's SDK makes it unavailable in mac catalyst.
 */
```

aboedo: The changes would look like this

image

aboedo: You can test this yourself by building the framework, then going to the generated docs in xcode with cmd + shift + 0

aboedo: alternatively `bundle exec fastlane preview_docs`

aboedo (changes requested):

Sources/Purchasing/Purchases/Purchases.swift
taquitos: This is what happens when I try to put the note within the note on a separate line to distinguish it from the rest of the original text.

aboedo (approved):

PR #1819, opened 2022-08-13
`Lock`: changed default implementation to use `NSLock`

NachoSoto:
`Atomic` can't be used recursively (due to Swift's `inout` limitation), so we pay an unnecessary price in performance using `NSRecursiveLock` that we don't need.
See https://swiftrocks.com/thread-safety-in-swift for a benchmark.

The implementation still allows using `NSRecursiveLock` when necessary.

### DIFF
--- a/Sources/Misc/Lock.swift
+++ b/Sources/Misc/Lock.swift
@@ -37,7 +37,7 @@ internal final class Lock {
     private init(_ lock: UnderlyingType) { self.lock = lock }
 
     /// Creates an instance based on `LockType`
-    convenience init(_ type: LockType) {
+    convenience init(_ type: LockType = .nonRecursive) {
         self.init(type.create())
     }
 
--- a/Sources/Misc/Lock.swift
+++ b/Sources/Misc/Lock.swift
@@ -13,19 +13,39 @@
 
 import Foundation
 
+/// A lock abstraction over an instance of `NSLocking`
 internal final class Lock {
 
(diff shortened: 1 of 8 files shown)


NachoSoto:
Bump @RevenueCat/core-sdk

Sources/Misc/Lock.swift
```diff
@@ -13,19 +13,39 @@
...
+    #else
+    // `NSRecursiveLock` and `NSLock` aren't `Sendable` until iOS 16.0 / Swift 5.7
+    private typealias UnderlyingType = NSLocking
+    #endif
+
+    private let lock: UnderlyingType
+    private init(_ lock: UnderlyingType) { self.lock = lock }
+
+    /// Initializes an instance backed by a non-recursive `NSLock`
+    convenience init() {
```
aboedo, line 30: kinda wonder if it now makes sense to have something like `createNonRecursive` so it's very clear when you're using them and you don't miss the fact that there's a choice. 
or even `lock = Lock(.recursive)` and `lock = Lock(.nonRecursive)`

aboedo: what do you think?

aboedo (approved):

Sources/Misc/Lock.swift
NachoSoto: Oh yeah totally 👍🏻 I'll change this.

NachoSoto: Done!

PR #1821, opened 2022-08-13
`SwiftLint`: cleaned up output

NachoSoto:
The output of `SwiftLint` is pretty noisy, which sometimes makes it hard to see other build warnings or errors in Xcode:

Screenshot 2022-08-12 at 12 23 31

This removes the "Linting... " lines when ran from Xcode.

### DIFF
--- a/scripts/swiftlint.sh
+++ b/scripts/swiftlint.sh
@@ -22,7 +22,8 @@ if which swiftlint >/dev/null; then
   echo "linter command: ${lint_command}"
   
   pushd "${source_path}"
-  $lint_command
+  # Run swiftlint but filter out "Linting ..." to clean up output
+  $lint_command 2>&1 | grep -v 'Linting '
   popd
 else
   echo "Warning: SwiftLint not installed in ${HOMEBREW_BINARY_DESTINATION}, download from https://github.com/realm/SwiftLint"


tonidero (approved):

RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -2234,7 +2234,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "scripts/swiftlint.sh\n";
+			shellScript = "# Run swiftlint but filter out \"Linting ...\" to clean up output\nscripts/swiftlint.sh 2>&1 | grep -v 'Linting '\n";
```
aboedo, line 2237: looks good! although maybe we can do it at the swiftlint.sh level so it's easier to find an edit in the future?

aboedo (approved):
❤️

RevenueCat.xcodeproj/project.pbxproj
NachoSoto: I thought that maybe we wouldn't want to hide what that's doing when ran directly, only to clean up the output in Xcode. What do you think?

NachoSoto: That's why I put it here, but let me know if you think it's worth moving it there.

aboedo: Yeah, that's a good point. We could make it an optional parameter too and pass it in here.

I may be over complicating things, no need to listen to me. 

When running locally, though, I just run `swiftlint`

NachoSoto: That's a good point, this doesn't really add much on top of `SwiftLint`, so no reason not to put this code in there instead. I'll change it.

PR #1827, opened 2022-08-13
`Backend`: fixed potential race conditions introduced by `OperationDispatcher.dispatchOnWorkerThread(withRandomDelay:)`

NachoSoto:
Fixes [CSDK-391]. As described in that ticket, calling a method that introduces a random delay before performing an API request could lead to a race condition. Example:

> Request 1: delay 0 seconds. Start now.
> Request 2: delay 2 seconds. Waiting…
> Request 1 finishes.
> Request 2: 2 seconds past. There is no in progress request, so we start it again

This also solves [CSDK-394]. If any part of our code (or users) call a method that would enqueue a request multiple times, we need to insert it into the callback queue right away.
That's what this does: it moves the random delay to *after* the callback has been added to the `CallbackCache`.

[CSDK-391]: https://revenuecats.atlassian.net/browse/CSDK-391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CSDK-394]: https://revenuecats.atlassian.net/browse/CSDK-394?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -35,26 +35,26 @@ class CustomerInfoManager {
     func fetchAndCacheCustomerInfo(appUserID: String,
                                    isAppBackgrounded: Bool,
                                    completion: ((Result<CustomerInfo, BackendError>) -> Void)?) {
-        deviceCache.setCacheTimestampToNowToPreventConcurrentCustomerInfoUpdates(appUserID: appUserID)
-        operationDispatcher.dispatchOnWorkerThread(withRandomDelay: isAppBackgrounded) {
-            self.backend.getCustomerInfo(appUserID: appUserID) { result in
-                switch result {
-                case let .failure(error):
-                    self.deviceCache.clearCustomerInfoCacheTimestamp(appUserID: appUserID)
-                    Logger.warn(Strings.customerInfo.customerinfo_updated_from_network_error(error: error))
+        self.deviceCache.setCacheTimestampToNowToPreventConcurrentCustomerInfoUpdates(appUserID: appUserID)
 
-                case let .success(info):
-                    self.cache(customerInfo: info, appUserID: appUserID)
-                    Logger.rcSuccess(Strings.customerInfo.customerinfo_updated_from_network)
-                }
+        self.backend.getCustomerInfo(appUserID: appUserID,
+                                     withRandomDelay: isAppBackgrounded) { result in
+            switch result {
+            case let .failure(error):
+                self.deviceCache.clearCustomerInfoCacheTimestamp(appUserID: appUserID)
+                Logger.warn(Strings.customerInfo.customerinfo_updated_from_network_error(error: error))
 
-                if let completion = completion {
-                    self.operationDispatcher.dispatchOnMainThread {
-                        completion(result)
-                    }
-                }
-
+            case let .success(info):
+                self.cache(customerInfo: info, appUserID: appUserID)
+                Logger.rcSuccess(Strings.customerInfo.customerinfo_updated_from_network)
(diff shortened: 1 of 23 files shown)


Sources/Identity/CustomerInfoManager.swift
```diff
@@ -35,26 +35,26 @@ class CustomerInfoManager {
     func fetchAndCacheCustomerInfo(appUserID: String,
                                    isAppBackgrounded: Bool,
                                    completion: ((Result<CustomerInfo, BackendError>) -> Void)?) {
-        deviceCache.setCacheTimestampToNowToPreventConcurrentCustomerInfoUpdates(appUserID: appUserID)
-        operationDispatcher.dispatchOnWorkerThread(withRandomDelay: isAppBackgrounded) {
```
NachoSoto, line 39: The only change here is that this is removed.

Sources/Networking/OfferingsAPI.swift
```diff
@@ -36,7 +38,9 @@ class OfferingsAPI {
         let offeringsCallback = OfferingsCallback(cacheKey: getOfferingsOperation.cacheKey, completion: completion)
         let cacheStatus = self.offeringsCallbacksCache.add(callback: offeringsCallback)
 
-        self.backendConfig.operationQueue.addCacheableOperation(getOfferingsOperation, cacheStatus: cacheStatus)
+        self.backendConfig.addCacheableOperation(getOfferingsOperation,
```
NachoSoto, line 41: Instead of adding it directly to the queue, a new method allows to introduce a random delay, but only *after* the callback is in the cache.

taquitos: Ahhhh I see, yeah, that's a much better way of doing it.

taquitos (approved):
💵💵🐐

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -38,7 +38,7 @@ class OfferingsManager {
     }
 
     func offerings(appUserID: String, completion: ((Result<Offerings, Error>) -> Void)?) {
-        guard let cachedOfferings = deviceCache.cachedOfferings else {
+        guard let cachedOfferings = self.deviceCache.cachedOfferings else {
```
aboedo, line 41: [not needed for this PR]
Here's the other bug that I was talking about in slack: we assume that not having cached offerings means that we need to request them, but a request might already be in flight. 
So here the check should have been something like 
`if cache is not stale then append to callbacks`, to account for a possible request in flight (this is why we set the cache timestamp to now before we actually have the cache returned from the backend). 

Then again this concurrency control predates the request caching that we have now, which automatically takes care of this bit. It'd be nice to move the bits that do `setOfferingsCacheTimestampToNow` to only do that after we actually have some data from the backend, but that's not needed for now. 

Same thinking for CustomerInfo, they both had the same system in place

aboedo (approved):
great work on this!

Sources/Purchasing/OfferingsManager.swift
NachoSoto: I sent you a message about this, because I _think_ this solves that as well.
Merging this PR for now but let me know if I misunderstood and happy to write a failing test for that scenario as well.

NachoSoto: See #1839.

PR #1834, opened 2022-08-15
`CallbackCache`: simplified implementation using `Atomic`

NachoSoto:
We don't actually need to share a single queue to synchronize all the callback caches.
This new implementation also avoids using the risky `DispatchQueue.sync`.

This also fixes potential issues because `cachedCallbacksByKey` was exposed, and it could have been read without synchronizing access with the queue. This is now impossible.


taquitos (approved):
surprised-pikachu-s

PR #1835, opened 2022-08-15
`CallbackCache: Sendable` conformance

NachoSoto:
The change in #1834 made this possible because the type is now immutable.


taquitos (approved):
🎈🐐

PR #1839, opened 2022-08-16
`DeviceCache`: no longer set cache timestamp before beginning request

NachoSoto:
Follow up to [CSDK-391]. As pointed out by @aboedo, we have now 2 caching mechanisms: `CallbackCache` and `DeviceCache` timestamps.
These 2 calls to set the timestamp (through `InMemoryCachedObject.updateCacheTimestamp(date:)` _before_ starting requests was there before `CallbackCache` was a thing.

However, with that solution, especially after the fix in #1827, the timestamp change is unnecessary. Its purpose was to make sure that a subsequent request didn't trigger an update if it was already in progress, but again, that's already handled by `CallbackCache` now.

The benefit of removing this is also that if a request fails, the timestamp cache wouldn't be an invalid `Date`, since nothing was actually fetched. In practice this isn't an issue because we clear the date on failure (See `OfferingsManager.handleOfferingsUpdateError` for example), but now we wouldn't need to rely on that.

### DIFF
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -35,8 +35,6 @@ class CustomerInfoManager {
     func fetchAndCacheCustomerInfo(appUserID: String,
                                    isAppBackgrounded: Bool,
                                    completion: ((Result<CustomerInfo, BackendError>) -> Void)?) {
-        self.deviceCache.setCacheTimestampToNowToPreventConcurrentCustomerInfoUpdates(appUserID: appUserID)
-
         self.backend.getCustomerInfo(appUserID: appUserID,
                                      withRandomDelay: isAppBackgrounded) { result in
             switch result {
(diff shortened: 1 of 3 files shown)


taquitos (approved):
☠️🐐

NachoSoto:
@aboedo this looks good to you?

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -35,8 +35,6 @@ class CustomerInfoManager {
     func fetchAndCacheCustomerInfo(appUserID: String,
                                    isAppBackgrounded: Bool,
                                    completion: ((Result<CustomerInfo, BackendError>) -> Void)?) {
-        self.deviceCache.setCacheTimestampToNowToPreventConcurrentCustomerInfoUpdates(appUserID: appUserID)
-
```
aboedo, line 39: just to clarify, it _still_ gets set to `now` after the backend call is successful and the cache is set, right?

aboedo (approved):

Sources/Identity/CustomerInfoManager.swift
NachoSoto: Yeah and there are tests for that too.

PR #1841, opened 2022-08-16
Purchase cancellations: unify behavior between SK1 and SK2

NachoSoto:
Fixes [CSDK-113] and https://github.com/RevenueCat/purchases-flutter/issues/403

### Changes:

- SK2 purchase now forwards an error (`ErrorCode.purchaseCancelledError`) if the purchase is cancelled. This matches the behavior in SK1.
- In order to retain the behavior on `async` APIs, this error is ignored if `userCancelled` is `true`, to ensure that `PurchaseResultData` is returned instead of an error being thrown.

Unfortunately, as explained in [CSDK-113], we can't test this because there's no way to fake cancellations (_FB10300403_).

[CSDK-113]: https://revenuecats.atlassian.net/browse/CSDK-113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[CSDK-113]: https://revenuecats.atlassian.net/browse/CSDK-113?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Misc/Purchases+async.swift
+++ b/Sources/Misc/Purchases+async.swift
@@ -66,7 +66,7 @@ extension Purchases {
     func purchaseAsync(product: StoreProduct) async throws -> PurchaseResultData {
         return try await withCheckedThrowingContinuation { continuation in
             purchase(product: product) { transaction, customerInfo, error, userCancelled in
-                continuation.resume(with: Result(customerInfo, error)
+                continuation.resume(with: Result(customerInfo, error?.ignoreIfPurchaseCancelled(userCancelled))
                                         .map { PurchaseResultData(transaction, $0, userCancelled) })
             }
         }
@@ -76,7 +76,7 @@ extension Purchases {
     func purchaseAsync(package: Package) async throws -> PurchaseResultData {
         return try await withCheckedThrowingContinuation { continuation in
             purchase(package: package) { transaction, customerInfo, error, userCancelled in
-                continuation.resume(with: Result(customerInfo, error)
+                continuation.resume(with: Result(customerInfo, error?.ignoreIfPurchaseCancelled(userCancelled))
                                         .map { PurchaseResultData(transaction, $0, userCancelled) })
             }
         }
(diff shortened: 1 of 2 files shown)


Sources/Misc/Purchases+async.swift
```diff
@@ -87,7 +87,7 @@ extension Purchases {
         return try await withCheckedThrowingContinuation { continuation in
             purchase(product: product,
                      promotionalOffer: promotionalOffer) { transaction, customerInfo, error, userCancelled in
-                continuation.resume(with: Result(customerInfo, error)
+                continuation.resume(with: Result(customerInfo, error?.ignoreIfPurchaseCancelled(userCancelled))
```
aboedo, line 90: feels like we're duplicating this code a lot, should we be moving this up the stack?

aboedo (approved):
looks good! sucks to not be able to test this programmatically. 
I assume it works fine locally right? 
Have you tried pointing a hybrid sdk to this branch to verify that they don't crash now? lmk if you need hand setting that up

Sources/Misc/Purchases+async.swift
NachoSoto: We can't, we do want to send the error in the completion block based APIs (that's what broke PHC, that SK2 wasn't doing that), but we don't want to throw an error in the `async` APIs. Does that make sense?

NachoSoto: So to summarize:
- Completion block APIs need both the error and the cancellation parameter. This is consistent with SK1 and SK2, users can choose how to read cancellations
- `async` APIs can't have an error and a value, so throwing in case of cancellation seems like a bad API, especially because the returned value has a `userCancelled` boolean.

So we can't push this up because we only want to remove the error in `async` APIs. Does that make sense?

NachoSoto:
> I assume it works fine locally right?
> Have you tried pointing a hybrid sdk to this branch to verify that they don't crash now? lmk if you need hand setting that up

Yup, I tried this in Flutter, but I just tried it on iOS through the `async` APIs and that's working as well.

aboedo (approved):

PR #1842, opened 2022-08-17
Release train

vegaro:
Creates a new workflow that will run weekly and that will perform the following:
- Determine next version to release using PR labels
- If next version to release is not a major release
- Open a PR preparing the release

Depends on https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/22

[CSDK-135]

[CSDK-135]: https://revenuecats.atlassian.net/browse/CSDK-135?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


NachoSoto (approved):
This looks amazing 👏🏻

aboedo (approved):

fastlane/Fastfile
```diff
@@ -47,11 +47,23 @@ platform :ios do
...
   end
 
+  desc "Automatically bumps version, update swift header, edit changelog, and create pull request"
+  lane :automatic_bump do |options|
+    next_version, type_of_bump = determine_next_version_using_labels(
+      repo_name: repo_name,
+      github_rate_limit: options[:github_rate_limit]
+    )
+    options[:next_version] = next_version
+    UI.user_error!('Skipping automatic bump since the next version is a major release') if type_of_bump == :major
```
tonidero, line 63: I wonder if this should have been a normal message instead of an error, since this is expected... I guess it's NABD since it shouldn't block anything though.

NachoSoto: Maybe Danger could show a warning if a PR has `breaking` as a label:

> Warning: this change will require a new major version

tonidero: Well I think this action would be executed from `main` most of the time, not from a PR, so I don't think Danger can warn us easily...

NachoSoto: I didn't mean in this action, I meant in PRs that have that. So before this action ends up failing, the PR that introduces that breaking change warns us about this happening once merged in `main`. Does that make sense?

tonidero: Ahh yeah, that could be useful. It can be a good reminder 👍

PR #1843, opened 2022-08-17
`MagicWeatherSwiftUI`: updated to use `async` APIs

NachoSoto:
I was testing some of these with a recent change, and I realized that it's probably best to have this example use the "modern" APIs anyway.
Not the best implementation, since we should move these methods to the view model, but this was the simplest way to do it for now without a huge refactor.

It's also not ideal that we were and are ignoring errors, but luckily the SDK logs them.


taquitos (approved):
🧙‍♂️🐐

PR #1844, opened 2022-08-17
`ErrorResponse`: don't add attribute errors to message if empty

NachoSoto:
Follow up to #1776. That PR added a lot of tests to cover error logging, and #1778 introduced a small error that wasn't covered: if attribute errors are empty, the `localizedDescription` could end up becoming `"[:]"`. This fixes that.

screenshot_2022-08-17_at_09 19 37_720

## Examples:

This was the underlying error returned (the output of `error.localizedDescription`):

And this is what was logged (notice that I also fixed `ErrorCode.invalidCredentialsError` to not log as an "Apple Error"):


taquitos (approved):
🎈🐐

PR #1849, opened 2022-08-17
`watchOS`: fixed crash on single-target apps

NachoSoto:
Fixes [CSDK-416].

Starting on `watchOS 9.0` (Xcode 14.0), `watchOS` apps can be configured as a single target.
Apps that run in this single-target crash when calling `WKExtension.shared`:
> -[WKExtension init]:62: Critical failure. Simulating crash: Condition failed:"NO". WKExtension is part of WatchKit.framework and can only be instanciated from a WatchKit Extension

To prevent this, we can use `WKApplication` if it's available.

[CSDK-416]: https://revenuecats.atlassian.net/browse/CSDK-416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Misc/SystemInfo.swift
+++ b/Sources/Misc/SystemInfo.swift
@@ -226,11 +226,16 @@ private extension SystemInfo {
         // Calling `WKExtension.shared` on these single-target apps crashes.
         //
         // `WKApplication` provides a more accurate value if it's available, and it avoids that crash.
+        #if swift(>=5.7)
         if #available(watchOS 7.0, *) {
             return WKApplication.shared().applicationState == .background
         } else {
             return WKExtension.shared().applicationState == .background
         }
+        #else
+        // Before Xcode 14, single-target extensions aren't supported (and WKApplication isn't available)
+        return WKExtension.shared().applicationState == .background
+        #endif
     }
 
     #endif
--- a/Sources/Misc/SystemInfo.swift
+++ b/Sources/Misc/SystemInfo.swift
@@ -197,11 +197,12 @@ private extension SystemInfo {
     #elseif os(macOS)
         return false
     #elseif os(watchOS)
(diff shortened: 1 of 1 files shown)


tonidero:
Looks like tests failed. Looks good otherwise.

NachoSoto:
Yeah it compiled locally with Xcode 14 and 13, I'll need to look into it.

NachoSoto:
Fixed. Turns out `WKApplication` is available on watchOS 7.0 and after, but it's not public API until Xcode 14.

Sources/Misc/SystemInfo.swift
```diff
@@ -211,7 +212,30 @@ private extension SystemInfo {
...
+        #if swift(>=5.7)
+        if #available(watchOS 7.0, *) {
+            return WKApplication.shared().applicationState == .background
+        } else {
+            return WKExtension.shared().applicationState == .background
+        }
+        #else
+        // Before Xcode 14, single-target extensions aren't supported (and WKApplication isn't available)
+        return WKExtension.shared().applicationState == .background
+        #endif
```
tonidero, line 238: Well this becomes slightly more complex, having to check for 2 things... But well, I don't have a better way to do it and it's not too bad.

tonidero (approved):

aboedo (approved):

Sources/Misc/SystemInfo.swift
NachoSoto: Yeah, but moving forward we'll only have one. Once Xcode 14 is mandatory, and when we drop anything before `watchOS 7.0`, there's really only one implementation :)

PR #1852, opened 2022-08-18
Build fix- Update package requirements for MagicWeather

taquitos:
pull in up to next major, fix build issue

### DIFF
--- a/Examples/MagicWeather/MagicWeather.xcodeproj/project.pbxproj
+++ b/Examples/MagicWeather/MagicWeather.xcodeproj/project.pbxproj
@@ -459,8 +459,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/RevenueCat/purchases-ios.git";
 			requirement = {
-				branch = main;
-				kind = branch;
+				kind = upToNextMajorVersion;
+				minimumVersion = 4.0.0;
 			};
 		};
 /* End XCRemoteSwiftPackageReference section */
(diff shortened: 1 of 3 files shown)


NachoSoto (approved):

Examples/MagicWeather/MagicWeather.xcodeproj/project.pbxproj
```diff
@@ -459,8 +459,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/RevenueCat/purchases-ios.git";
 			requirement = {
-				branch = main;
-				kind = branch;
```
tonidero, line 463: Hmm honestly I don't know, but will this cause the MagicWeather app to only work with deployed versions of `purchases-ios`? I think it's nice to test with the local version of `purchases-ios`, but then again I don't think it's a big deal, so I'm ok with it.

taquitos: Oh, that's true. It's already pinned a revision, I was thinking it make sense just to keep it at what our customers would use. PurchaseTester feels like it should be pointed to main, but maybe these sample apps make sense to keep pinned to releases?

tonidero: Yeah, I think it's good to keep at least a sample app pointing to the latest release, but having one where we can easily test is also good. Another thing is that I'm slightly concerned about forgetting to update this when there is a new major... Maybe we can try to automate this but it would be tricky... Honestly I'm a bit torn about this so I think it would be good to discuss it on Slack (I'm out right now but can initiate the discussion on monday)

PR #1855, opened 2022-08-19
`IntegrationTests`: actually fail test if tests aren't configured

NachoSoto:
Looks like calling `XCTFail` followed by throwing an error was making the test continue for some reason, which was then making it crash instead because `Purchases.shared` isn't set.
By removing the `XCTFail`, `XCTest` now correctly sees this as a test failure and doesn't run the tests.

See also https://github.com/RevenueCat/purchases-hybrid-common/pull/210


aboedo (approved):
super weird

NachoSoto:
Might be a regression in iOS 16, but this is simpler anyway.

PR #1861, opened 2022-08-23
Added `beginRefundRequest` overload with completion block

NachoSoto:
Fixes [CSDK-218].

The underlying implementation provided by Apple is `async` only, which is still available to Objective-C. However, to be able to use it from Swift, it must be used with `await`.
This exposes an overload that allows using it with a completion-block based API. Because we don't need these in Objective-C (those are already available), these blocks can use the Swift-only `Result`.

[CSDK-218]: https://revenuecats.atlassian.net/browse/CSDK-218?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Misc/Purchases+nonasync.swift
+++ b/Sources/Misc/Purchases+nonasync.swift
@@ -1,0 +1,109 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  Purchases+nonasync.swift
+//
+//  Created by Nacho Soto on 8/22/22.
+
+import Foundation
+
+/// This extension holds the biolerplate logic to convert async methods to completion blocks APIs.
+/// Because `async` APIs are implicitly available in Objective-C, these can be Swift only.
+public extension Purchases {
+
+    #if os(iOS)
+    /**
+     * Presents a refund request sheet in the current window scene for
+     * the latest transaction associated with the `productID`
+     *
+     * - Parameter productID: The `productID` to begin a refund request for.
+     * - Parameter completion: A completion block that is called when the ``RefundRequestStatus`` is returned.
+     * Keep in mind the status could be ``RefundRequestStatus/userCancelled``
+     * If the request was unsuccessful, no active entitlements could be found for the user,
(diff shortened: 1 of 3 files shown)


Tests/APITesters/SwiftAPITester/SwiftAPITester/PurchasesAPI.swift
```diff
@@ -232,12 +234,19 @@ private func checkAsyncMethods(purchases: Purchases) async {
         let _: RefundRequestStatus = try await purchases.beginRefundRequest(forEntitlement: "")
         let _: RefundRequestStatus = try await purchases.beginRefundRequestForActiveEntitlement()
 
-        // Deprecated
```
NachoSoto, line 235: This was a leftover from an old method that's been obsoleted.

Sources/Misc/Purchases+nonasync.swift
```diff
@@ -0,0 +1,109 @@
...
+     * If a user could have more than one entitlement at a time, use ``beginRefundRequest(forEntitlement:)`` instead.
+     */
+    @available(iOS 15.0, *)
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(tvOS, unavailable)
+    func beginRefundRequestForActiveEntitlement(
+        completion: @escaping (Result<RefundRequestStatus, Error>) -> Void
+    ) {
+        call(completion) {
```
tonidero, line 87: It feels a bit weird to `call` the completion before the actual block... Maybe we could rename to `awaitWithCompletion` or something like that?

tonidero (approved):
Just a rename suggestion, looks good otherwise

Sources/Misc/Purchases+nonasync.swift
NachoSoto: Yeah this was a bad name, since it was a private function I didn't think too hard about it hehe, but yeah that's a better name I'll change it,.

NachoSoto: Yeah this was a bad name, since it was a private function I didn't think too hard about it hehe, but yeah that's a better name I'll change it.

NachoSoto: Actually the Swifty version would be `await(with completion:)`, but we can't make a function named `await`. I'll rename this to `call(with completion:)`

PR #1871, opened 2022-08-30
`ErrorUtils`: added test to verify that returned errors can be converted to `ErrorCode`

NachoSoto:
This is a key part of how we document error handling in [`V4_API_Migration_guide.md`](https://github.com/RevenueCat/purchases-ios/blob/4.10.3/Sources/DocCDocumentation/DocCDocumentation.docc/V4_API_Migration_guide.md#error-handling).

I'm working on an error type-safety refactor that broke this invariant, which wasn't covered in any test.

### DIFF
--- a/Tests/UnitTests/Purchasing/ErrorUtilsTests.swift
+++ b/Tests/UnitTests/Purchasing/ErrorUtilsTests.swift
@@ -38,6 +38,13 @@ class ErrorUtilsTests: TestCase {
         super.tearDown()
     }
 
+    func testErrorsCanBeConvertedToErrorCode() throws {
+        let error = ErrorUtils.customerInfoError()
+        let errorCode = try XCTUnwrap(error as? ErrorCode, "Error couldn't be converted to ErrorCode")
+
+        expect(errorCode).to(matchError(error))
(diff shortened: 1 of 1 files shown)


Tests/UnitTests/Purchasing/ErrorUtilsTests.swift
```diff
@@ -38,6 +38,13 @@ class ErrorUtilsTests: TestCase {
         super.tearDown()
     }
 
+    func testErrorsCanBeConvertedToErrorCode() throws {
+        let error = ErrorUtils.customerInfoError()
```
tonidero, line 42: I think this should be enough, but should we do this for all error types? It might be a lot of tests though 😅

tonidero (approved):

Tests/UnitTests/Purchasing/ErrorUtilsTests.swift
NachoSoto: Yeah I think that's not necessary. This is indirectly testing `ErrorUtils.error(with:message:...)`, but that's `private`.

PR #1437, opened 2022-03-31
Introduced `NetworkError` and `BackendError`

NachoSoto:
## Changes:

- Each layer (only a few for now) has its own error type: `NetworkError`, `BackendError`, `OfferingsManager.Error`.
- `HTTPClient` for example has to produce `NetworkError`, operations produce `BackendError`

```diff
struct HTTPResponse<Body: HTTPResponseBody> {
-   typealias Result = Swift.Result<Self, Error>
+   typealias Result = Swift.Result<Self, NetworkError>
}
```

- The parent `BackendError` can have specific errors like `.missingAppUserID`, but also be simply a child error `case networkError(NetworkError)`
- All of these conform to a `ErrorCodeConvertible`, so there is a single point of code that converts from simple and readable errors (like `BackendError.emptySubscriberAttributes`, `.unexpectedBackendResponse(.loginResponseDecoding)`) into errors with all the context using `ErrorUtils`
- Tests also become simpler:

```diff
-        expect(receivedError?.domain).toEventually(equal(RCPurchasesErrorCodeDomain))
-        expect(receivedError?.code).toEventually(
-            equal(ErrorCode.unexpectedBackendResponseError.rawValue))
-        expect(receivedUnderlyingError?.code).toEventually(
-            equal(UnexpectedBackendResponseSubErrorCode.postOfferIdMissingOffersInResponse.rawValue))
+
+        expect(receivedError) == .unexpectedBackendResponse(.postOfferIdMissingOffersInResponse)
```
- Converted `DNSError` into `NetworkError.dnsError`. Its functionality remains unchanged.
- Removed `Backend.RCSuccessfullySyncedKey` and `ErrorDetails.finishableKey` in favor of tested properties on `NetworkError`

PR #1879, opened 2022-08-31
Introduced type-safe `PurchasesError` and fixed some incorrect returned error types

NachoSoto:
This is the last change from what was started with #1437. That PR introduced a few private error types, but for the errors returned in the public `APIs` we were still passing around `Error`s everywhere, which didn't provide any compile-time guarantees.

### Goals:
- Ensure that public APIs *only* return `ErrorCode`s wrapped with the additional `userInfo`, and not accidentally return other types like `SKError`, `BackendError`, `ErrorCode` directly, etc.
- Guarantee type-safety when passing errors around within our internal implementations instead of assuming certain thrown errors are of a particular type.

### Changes:
- Created `PurchasesError` for wrapping an `ErrorCode` with the additional `userInfo`.
- Changed returned errors from `Error` to `NSError`. This is a change in the API, but the new type is a more specific `Error` so it shouldn't lead to any changes in client apps.
- Because `PurchasesError` isn't implicitly convertible to `NSError`, this ensures that we don't return those values directly, and instead use `PurchasesError.asPublicError`.
- Changed `ErrorCodeConvertible` to a now more precise `PurchasesErrorConvertible`.

Thanks to this new type-safety this PR fixes at least 3 bugs where we were returning private errors instead of `ErrorCode`s.

### Testing:

Thanks to #1871 we know that the returned errors are still convertible to `ErrorCode` so users can `switch` over them.

### DIFF
--- a/Sources/Error Handling/ErrorUtils.swift
+++ b/Sources/Error Handling/ErrorUtils.swift
@@ -33,7 +33,7 @@ enum ErrorUtils {
         withUnderlyingError underlyingError: Error? = nil,
         extraUserInfo: [NSError.UserInfoKey: Any]? = nil,
         fileName: String = #fileID, functionName: String = #function, line: UInt = #line
-    ) -> Error {
+    ) -> PurchasesError {
 
         let errorCode: ErrorCode
         if case NetworkError.dnsError(_, _, _)? = underlyingError {
@@ -54,7 +54,7 @@ enum ErrorUtils {
      */
     static func offlineConnectionError(
         fileName: String = #fileID, functionName: String = #function, line: UInt = #line
-    ) -> Error {
+    ) -> PurchasesError {
         return error(with: .offlineConnectionError,
                      fileName: fileName, functionName: functionName, line: line)
     }
@@ -73,7 +73,7 @@ enum ErrorUtils {
     static func backendError(
         withBackendCode backendCode: BackendErrorCode, backendMessage: String?,
         fileName: String = #fileID, functionName: String = #function, line: UInt = #line
-    ) -> Error {
+    ) -> PurchasesError {
         return backendError(withBackendCode: backendCode, backendMessage: backendMessage, extraUserInfo: nil,
                             fileName: fileName, functionName: functionName, line: line)
     }
@@ -86,7 +86,7 @@ enum ErrorUtils {
     static func unexpectedBackendResponseError(
         extraUserInfo: [NSError.UserInfoKey: Any]? = nil,
         fileName: String = #fileID, functionName: String = #function, line: UInt = #line
-    ) -> Error {
+    ) -> PurchasesError {
         return error(with: ErrorCode.unexpectedBackendResponseError, extraUserInfo: extraUserInfo,
                      fileName: fileName, functionName: functionName, line: line)
     }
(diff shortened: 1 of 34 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -559,10 +559,10 @@ public extension Purchases {
      * - ``Purchases/appUserID``
      */
     @objc(logIn:completion:)
-    func logIn(_ appUserID: String, completion: @escaping (CustomerInfo?, Bool, Error?) -> Void) {
+    func logIn(_ appUserID: String, completion: @escaping (CustomerInfo?, Bool, PublicError?) -> Void) {
         self.identityManager.logIn(appUserID: appUserID) { result in
             self.operationDispatcher.dispatchOnMainThread {
-                completion(result.value?.info, result.value?.created ?? false, result.error)
+                completion(result.value?.info, result.value?.created ?? false, result.error?.asPurchasesError)
```
NachoSoto, line 565: This was a bug, we were returning `BackendError`s! Caught by the compiler now.

NachoSoto:
Yess I got this all working with type-safe public errors 🤪

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -391,7 +391,7 @@ final class PurchasesOrchestrator {
                 ))
 
                 DispatchQueue.main.async {
-                    completion(nil, nil, error, false)
+                    completion(nil, nil, ErrorUtils.purchasesError(withUntypedError: error).asPublicError, false)
```
NachoSoto, line 394: This is needed because of untyped `throws` in Swift, which doesn't allow us to guarantee type safety here.
However this is pretty well covered in tests.

Sources/Error Handling/ErrorUtils.swift
```diff
@@ -220,10 +230,10 @@ enum ErrorUtils {
         message: String? = nil,
         underlyingError: Error? = nil,
         fileName: String = #fileID, functionName: String = #function, line: UInt = #line
-    ) -> Error {
-        return error(with: ErrorCode.configurationError,
-                     message: message, underlyingError: underlyingError,
-                     fileName: fileName, functionName: functionName, line: line)
+    ) -> PurchasesError {
+        return self.error(with: ErrorCode.configurationError,
```
tonidero, line 234: Hmm there are a ton of these that call `error` without the `self`. Also, this is a static function so this shouldn't be needed I think? (I might be wrong)

Sources/Error Handling/PurchasesError.swift
```diff
@@ -0,0 +1,67 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  PurchasesError.swift
+//
+//  Created by Nacho Soto on 8/31/22.
+
+import Foundation
+
+/// An error returned by a `RevenueCat` public API.
+public typealias PublicError = NSError
```
tonidero, line 17: Hmm I wonder if we should rename this... `PublicError` from the point of view of SDK users doesn't make that much sense... Maybe `RCError`? `PurchasesError` is already used... Can't think of other names at the moment 😅

tonidero (commented):
It looks good, like the extra security this provides us 😄 . Just a couple of questions.

Sources/Error Handling/PurchasesError.swift
tonidero: Hmm I guess I don't love any of the current options but I have a slight preference for `ExposedError`. Also, not sure if we should namespace the errors to RC? As in `RCExposedError`?

tonidero (approved):
There is the question of that rename but the PR looks great!

Sources/Error Handling/PurchasesError.swift
NachoSoto: We don't need to namespace in Swift (since it's already namespaces by the module name `RevenueCat`. And this `typealias` is transparent in Objective-C, meaning APIs see `NSError` instead.

PR #1880, opened 2022-08-31
`Integration Tests`: actually verify that entitlement is active

NachoSoto:
`verifyEntitlementWentThrough` was only looking for an entitlement inside of `EntitlementInfos.all`, but not verifying that it's active.


vegaro (approved):

NachoSoto:
This is failing, I wonder if it's the same issue we've been discussing.

NachoSoto:
This makes [TRIAGE-82] even more reproducible.

aboedo (approved):

PR #1881, opened 2022-09-01
Added `StoreKit2Setting.shouldOnlyUseStoreKit2`

NachoSoto:
A couple of helpers that can be useful, for #1882 for example.

### DIFF
--- a/Sources/Misc/StoreKit2Setting.swift
+++ b/Sources/Misc/StoreKit2Setting.swift
@@ -38,6 +38,27 @@ extension StoreKit2Setting {
 
 }
 
+extension StoreKit2Setting {
+
+    /// - Returns: `true` if SK2 is available in this device.
+    static var isStoreKit2Available: Bool {
+        if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
+            return true
+        } else {
+            return false
+        }
+    }
+
+    /// - Returns: `true` iff SK2 is enabled and it's available.
+    var shouldOnlyUseStoreKit2: Bool {
+        switch self {
+        case .disabled, .enabledOnlyForOptimizations: return false
+        case .enabledForCompatibleDevices: return Self.isStoreKit2Available
+        }
+    }
+
+}
+
 extension StoreKit2Setting: CustomDebugStringConvertible {
 
     var debugDescription: String {
--- a/Tests/UnitTests/Misc/StoreKit2SettingTests.swift
+++ b/Tests/UnitTests/Misc/StoreKit2SettingTests.swift
@@ -1,0 +1,61 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  StoreKit2SettingTests.swift
+//
+//  Created by Nacho Soto on 9/1/22.
+
+import Nimble
+import XCTest
+
+@testable import RevenueCat
+
+class StoreKit2SettingTests: TestCase {
+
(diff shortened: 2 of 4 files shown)


Sources/Misc/StoreKit2Setting.swift
```diff
@@ -38,6 +38,27 @@ extension StoreKit2Setting {
...
+    /// - Returns: `true` if SK2 is available in this device.
+    static var isStoreKit2Available: Bool {
+        if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
+            return true
+        } else {
+            return false
+        }
+    }
+
+    /// - Returns: `true` iff SK2 is enabled and it's available.
```
tonidero, line 52: Typo 
```suggestion
    /// - Returns: `true` if SK2 is enabled and it's available.
```

Tests/UnitTests/Misc/StoreKit2SettingTests.swift
```diff
@@ -0,0 +1,61 @@
...
+    func testShouldOnlyUseStoreKitFalseIfNotAvailable() throws {
+        if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
+            throw XCTSkip("Test only for older devices")
+        }
+
+        expect(StoreKit2Setting.enabledForCompatibleDevices.shouldOnlyUseStoreKit2) == false
+    }
+
+    func testShouldOnlyUseStoreKitTrueIfAvailable() throws {
+        guard #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) else {
```
tonidero, line 38: Since this condition is kind of repeated in a few places, I wonder if we could extract it somewhere... Though I guess that's exactly what `isStoreKit2Available` does, so this little repetition might be ok :P

Sources/Misc/StoreKit2Setting.swift
```diff
@@ -38,6 +38,27 @@ extension StoreKit2Setting {
...
 
+extension StoreKit2Setting {
+
+    /// - Returns: `true` if SK2 is available in this device.
+    static var isStoreKit2Available: Bool {
+        if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
+            return true
+        } else {
+            return false
+        }
```
tonidero, line 49: Could this work?
```suggestion
        return #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
```

tonidero (approved):
LGTM!

Sources/Misc/StoreKit2Setting.swift
NachoSoto: `iff` means "if and only if" 😅

Tests/UnitTests/Misc/StoreKit2SettingTests.swift
NachoSoto: No but you're right, let me move it to `AvailabilityChecks` :)

Sources/Misc/StoreKit2Setting.swift
NachoSoto: I'm pretty sure it doesn't unfortunately but let me try it

Sources/Misc/StoreKit2Setting.swift
tonidero: Ahhh haha just learned something 😅. It's good then!

Sources/Misc/StoreKit2Setting.swift
NachoSoto: Confirmed you can't. I hope Swift makes all this stuff expressions in the future.

Tests/UnitTests/Misc/StoreKit2SettingTests.swift
NachoSoto: Done.

PR #1882, opened 2022-09-01
Disable SK1's `StoreKitWrapper` if SK2 is enabled and available

NachoSoto:
When SK2 is enabled, `StoreKit2TransactionListener` is enough to handle transactions, so it's no longer necessary to have that and `StoreKitWrapper` both listening to them.

Depends on #1881 and #1879.

### TODO:

- [x] Improve handling of attempting to purchase SK1 products when it's implicitly disabled. This shouldn't happen (covered well in tests), but we need to handle that more gracefully.
- [x] What about `presentCodeRedemptionSheet` and `showPriceConsentIfNeeded`? We could have a dedicated "`SK1Wrapper`" with support for only those 2 functions, since they don't have an SK2 equivalent.
- [x] Add tests to verify this behavior.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -249,6 +249,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
     private let receiptFetcher: ReceiptFetcher
     private let requestFetcher: StoreKitRequestFetcher
     private let storeKitWrapper: StoreKitWrapper?
+    private let paymentQueueWrapper: PaymentQueueWrapper
     private let systemInfo: SystemInfo
     private var customerInfoObservationDisposable: (() -> Void)?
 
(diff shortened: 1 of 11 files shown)


NachoSoto:
I believe all tests should pass.

aboedo:
I think having a separate wrapper for the ones that are SK1-only should work. The only reason we have a wrapper for them in the first place is to make mocking easier, and that function can still be served by a separate wrapper

NachoSoto:
This is ready now.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1678,6 +1687,11 @@ internal extension Purchases {
         return self.systemInfo.isSandbox
     }
 
+    /// For testing purposes
+    var isStoreKit1Configured: Bool {
+        return self.storeKitWrapper != nil
```
aboedo, line 1692: starting to think we should rename this to `storeKit1Wrapper`

Sources/Purchasing/StoreKit1/PaymentQueueWrapper.swift
```diff
@@ -0,0 +1,56 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  PaymentQueueWrapper.swift
+//
+//  Created by Nacho Soto on 9/4/22.
+
+import Foundation
+import StoreKit
+
+final class PaymentQueueWrapper {
```
aboedo, line 17: :chefskiss:

aboedo (approved):

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: Yeah that's a good idea, let me do that.

NachoSoto: Actually let me send a separate PR for that so we can keep the history of this particular change easier to follow in the future.

PR #1884, opened 2022-09-05
`Offering`: improved confusing log for `PackageType.custom`

NachoSoto:
Fixes [CSDK-450].

We were logging a message for packages with custom duration that says "Unknown subscription length for package, ignoring."
This is very confusing for customers, as it looks like something is wrong. However it’s expected and normal behavior for packages with custom duration.

[CSDK-450]: https://revenuecats.atlassian.net/browse/CSDK-450?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Logging/Strings/OfferingStrings.swift
+++ b/Sources/Logging/Strings/OfferingStrings.swift
@@ -38,6 +38,8 @@ enum OfferingStrings {
     case configuration_error_no_products_for_offering
     case offering_empty(offeringIdentifier: String)
     case product_details_empty_title(productIdentifier: String)
+    case unknown_package_type(Package)
+    case custom_package_type(Package)
 
 }
 
@@ -126,6 +128,11 @@ extension OfferingStrings: CustomStringConvertible {
         case let .product_details_empty_title(identifier):
             return "Empty Product titles are not supported. Found in product with identifier: \(identifier)"
 
+        case let .unknown_package_type(package):
+            return "Unknown subscription length for package '\(package.offeringIdentifier)'. Ignoring."
+
+        case let .custom_package_type(package):
+            return "Package '\(package.offeringIdentifier)' has a custom duration. Ignoring."
         }
     }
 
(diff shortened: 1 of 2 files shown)


Sources/Purchasing/Offering.swift
```diff
@@ -138,11 +139,13 @@ import Foundation
             case .custom where package.storeProduct.productCategory == .nonSubscription:
                 // Non-subscription product, ignoring
```
NachoSoto, line 140: I still left this extra `case` because it doesn't even warrant a log.

Sources/Purchasing/Offering.swift
```diff
@@ -109,6 +109,7 @@ import Foundation
         return package(identifier: key)
     }
 
+    // swiftlint:disable:next cyclomatic_complexity
```
tonidero, line 112: Hmm might be good to try to refactor this to avoid this issue... But NABD.

Sources/Logging/Strings/OfferingStrings.swift
```diff
@@ -38,6 +38,8 @@ enum OfferingStrings {
     case configuration_error_no_products_for_offering
     case offering_empty(offeringIdentifier: String)
     case product_details_empty_title(productIdentifier: String)
+    case unknown_package_type(Package)
+    case custom_package_type(Package)
```
tonidero, line 42: I see the cases above have a parameter name in addition to the type. Is that ok? And if so, it might be a good idea to consolidate them.

tonidero (approved):
Looks good!

Sources/Logging/Strings/OfferingStrings.swift
NachoSoto: It's useful for parameters like `String`s because it's not clear what they contain. But a package is always going to be a package 😅

Sources/Purchasing/Offering.swift
NachoSoto: Yeah I thought about it but this needs to be in `init` in order to make the type immutable so I couldn't think of how to extract it out. Ultimately it's just a factor of the number of `case`s and putting a limit to that is kind of arbitrary.

PR #1885, opened 2022-09-05
`PurchasesOrchestrator`: don't log warning if `allowSharingAppStoreAccount` setting was never explicitly set

NachoSoto:
Fixes [CSDK-454].

This should reduce a lot of the noise during initialization. The internal backing property was already `Bool?`, so we can check if it was ever set, and if not, not log anything.
Using `TestLogHandler` introduced in #1858 to verify this behavior.

[CSDK-454]: https://revenuecats.atlassian.net/browse/CSDK-454?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aboedo (approved):
awesome! cc @codykerns who was curious about implementation approaches for this one

PR #1886, opened 2022-09-06
Renamed `StoreKitWrapper` to `StoreKit1Wrapper`

NachoSoto:
Follow up to #1882. Now that `StoreKit1Wrapper` is not initialized for SK2, it makes sense to clarify that this class is only for SK1.
This is just a simple find-replace.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -137,15 +137,15 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
      */
     @available(iOS 8.0, macOS 10.14, watchOS 6.2, macCatalyst 13.0, *)
     @objc public static var simulatesAskToBuyInSandbox: Bool {
-        get { StoreKitWrapper.simulatesAskToBuyInSandbox }
-        set { StoreKitWrapper.simulatesAskToBuyInSandbox = newValue }
+        get { StoreKit1Wrapper.simulatesAskToBuyInSandbox }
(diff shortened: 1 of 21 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -137,15 +137,15 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
      */
     @available(iOS 8.0, macOS 10.14, watchOS 6.2, macCatalyst 13.0, *)
     @objc public static var simulatesAskToBuyInSandbox: Bool {
```
tonidero, line 139: Not needed for this PR but should the name here and `canMakePayments` also reference that this is a SK1 only property?

tonidero (approved):

Sources/Purchasing/Purchases/Purchases.swift
aboedo: I think we should keep those as they are: 
- They're public variables, so a rename is a breaking change
- There aren't any SK2 equivalents, you're always using SK1 for either of them
- It doesn't really make a difference for the customer for these two

aboedo (approved):

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: > this is a SK1 only property?

Yeah the fact that it uses an SK1 method is an implementation detail. Hopefully SK2 will add another method and we'll migrate to it in the future.

PR #1887, opened 2022-09-06
Enabled `DEAD_CODE_STRIPPING`

NachoSoto:
I did some tests and this didn't seem to reduce the binary size of the framework by much, but it's recommended by Xcode during "update to recommended settings".
This is also the default on new projects.


tonidero (approved):
From what I've read about this, I think it should be ok

PR #1888, opened 2022-09-06
`CustomerInfoManager`/`OfferingsManager`: improved display of underlying errors

NachoSoto:
Improved these 2 logs:

> 🍎‼️ Error fetching offerings - The operation couldn’t be completed. (RevenueCat.OfferingsManager.Error error 0.)
Underlying error: The operation couldn’t be completed. (RevenueCat.BackendError error 0.)

> ⚠️ Attempt to update CustomerInfo from network failed.
The operation couldn’t be completed. (RevenueCat.BackendError error 0.)

These 2 now will display the underlying `NetworkError` instead of the opaque "(RevenueCat.BackendError error 0.)".

### DIFF
--- a/Sources/Purchasing/OfferingsManager.swift
+++ b/Sources/Purchasing/OfferingsManager.swift
@@ -240,6 +240,7 @@ extension OfferingsManager.Error: CustomNSError {
 
     fileprivate var underlyingError: Error? {
         switch self {
+        case let .backendError(.networkError(error)): return error
         case let .backendError(error): return error
         case let .configurationError(_, error, _): return error
         case .noOfferingsFound: return nil
--- a/Sources/Error Handling/BackendError.swift
+++ b/Sources/Error Handling/BackendError.swift
@@ -153,6 +153,27 @@ extension BackendError {
 
 extension BackendError {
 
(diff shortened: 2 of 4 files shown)


Sources/Purchasing/OfferingsManager.swift
```diff
@@ -240,6 +240,7 @@ extension OfferingsManager.Error: CustomNSError {
 
     fileprivate var underlyingError: Error? {
         switch self {
+        case let .backendError(.networkError(error)): return error
```
tonidero, line 243: Should we use the `underlyingError` of the `BackendError` here?

tonidero (approved):
LGTM! Just a question.

Sources/Purchasing/OfferingsManager.swift
NachoSoto, line 243: This is why typed errors are amazing.

NachoSoto: That's not defined, but it could be another way to handle this yeah.
I'd say for now we don't need to expose that for anything else so this is a bit more direct, but we could change it in the future.

PR #1898, opened 2022-09-09
`Integration Tests`: added test to verify re-subscription behavior

NachoSoto:


Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -293,6 +293,26 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
...
+
+        // 1. Subscribe
+        let customerInfo = try await subscribe()
+        let entitlement = try XCTUnwrap(customerInfo.entitlements[Self.entitlementIdentifier])
+
+        // 2. Expire subscription
+        try await self.expireSubscription(entitlement)
+
+        // 3. Resubscribe
+        try await subscribe()
```
tonidero, line 313: Hmm not that familiar with this so I might be missing something but looks like we are not doing any expectations or assertions, so I guess this would just verify that it doesn't crash?

NachoSoto: Yeah implicitly asserting that there are no errors. Also the test's purchase method verified that the entitlement is active after purchasing.

tonidero (approved):

PR #1901, opened 2022-09-10
`PaymentQueueWrapper`: handle promotional purchase requests from App Store when SK1 is disabled

NachoSoto:
Follow up to #1882. After that PR, `StoreKit1Wrapper` was no longer listening to `SKPaymentQueueDelegate.paymentQueue(:shouldAddStorePayment:for:)`, which means that purchases initiated from the App Store were being ignored by the SDK if SK2 is enabled.

To fix this, `PaymentQueueWrapper` now becomes the `SKPaymentQueueDelegate` only when `StoreKit1Wrapper` is not initialized, and `PurchasesOrchestrator` initiates an SK2 purchase instead.

## Other changes:
- `PurchasesOrchestrator` purchase methods no longer need the whole `PromotionalOffer`, only the `SignedData`. This allows the new implementation to create that data from only the `SKPaymentDiscount` provided from the `SKPayment`.
- Added tests for `PromotionalOffer.SignedData`

## TODO:
- [x] Log warning if fetching product failed.
- [x] Handle discounts 
- [x] Test that `StoreKit1Wrapper` is the delegate if SK1 is enabled
- [x] Test that `PaymentQueueWrapper` is the delegate otherwise
- [x] Test `PromotionalOffer.SignedData` initialization from SK1


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -603,6 +603,44 @@ extension PurchasesOrchestrator: StoreKit1WrapperDelegate {
...
+    func paymentQueueWrapper(
+        _ wrapper: PaymentQueueWrapper,
+        shouldAddStorePayment payment: SKPayment,
+        for product: SK1Product
+    ) -> Bool {
+        // `PurchasesOrchestrator` becomes `PaymentQueueWrapperDelegate` only
+        // when `StoreKit1Wrapper` is not initialized, which means that promoted purchases
+        // need to be handled as a SK2 purchase.
+        // This method converts the `SKPayment` into an SK2 purchase by fetching the product again.
+        assert(self.storeKit1Wrapper == nil, "This method should not be invoked if SK1 is enabled")
```
NachoSoto, line 617: This is not a `precondition` so it will be ignored in release builds. Still useful to catch potential misconfigurations in tests.

NachoSoto:
I'll finish this today.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
tonidero: Should we also log an error if this gets called for any reason with SK1 in release builds?

Sources/Purchasing/StoreKit1/PaymentQueueWrapper.swift
```diff
@@ -14,12 +14,33 @@
 import Foundation
 import StoreKit
 
-final class PaymentQueueWrapper {
+protocol PaymentQueueWrapperDelegate: AnyObject, Sendable {
```
tonidero, line 17: It's a bit confusing that this is defined in the `StoreKit1` folder but we will be using it in SK2. Maybe it should be moved out of that folder?

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -603,6 +603,57 @@ extension PurchasesOrchestrator: StoreKit1WrapperDelegate {
...
+                                  package: nil) { transaction, customerInfo, error, cancelled in
+                        completion(transaction, customerInfo, error, cancelled)
+                    }
+                }
+            }
+
+            delegate.readyForPromotedProduct(product, purchase: startPurchase)
+        }
+
+        return false
```
tonidero, line 652: Looks like we always return false in this function. Also, I'm not sure what the returned boolean is supposed to mean... Maybe store in a variable with a meaningful name so it's easier to understand the returned value?

Tests/UnitTests/Purchasing/StoreKitAbstractions/PromotionalOfferTests.swift
```diff
@@ -0,0 +1,68 @@
...
+
+    static let randomOffer: PromotionalOffer.SignedData = .init(
+        identifier: "identifier \(Int.random(in: 0..<1000))",
+        keyIdentifier: "key identifier \(Int.random(in: 0..<1000))",
+        nonce: .init(),
+        signature: "signature \(Int.random(in: 0..<1000))",
+        timestamp: Int.random(in: 0..<1000)
+    )
+
+    func setUpPurchases(enableStoreKit2: Bool) {
```
tonidero, line 61: Hmm I might be missing something but this seems unused?

tonidero (commented):
LGTM but will leave to people more familiar with this to do another review.

Sources/Purchasing/StoreKit1/PaymentQueueWrapper.swift
NachoSoto: It does wrap in `SK1` type. When we get rid of SK1 support it'll help that it's there :)

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: Yeah that confused me to in the old implementation. Let me dig into why and document it 👍🏻

NachoSoto: Done.
I added this note:
> // See `SKPaymentTransactionObserver.paymentQueue(_:shouldAddStorePayment:for:)`
// Returns `false` to indicate that the app will defer the purchase and be handled
// when the user calls the purchase callback.

Tests/UnitTests/Purchasing/StoreKitAbstractions/PromotionalOfferTests.swift
NachoSoto: Good catch, I didn't need this after all and forgot to remove it.

aboedo (approved):

PR #1902, opened 2022-09-13
`Purchases`: only expose testing data on `DEBUG`

NachoSoto:
These are only used for tests, so we don't need to compile them in the final binary.


tonidero (approved):
LGTM!

PR #1860, opened 2022-08-23
`CircleCI`: added job for building SDK with `SPM`

NachoSoto:
Fixes [CSDK-221].

SPM requires explicit `import Foundation` statements. Missing these could go unnoticed, so this new step will ensure that we don't break compilation with SPM.

- Example of [successful build](https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/7913/workflows/129b5e08-cdbe-4b3e-8606-8709655df205/jobs/35323/parallel-runs/0/steps/0-107).
- Example of [failed build from a missing `import Foundation`](https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/7914/workflows/4296ba99-f966-40d0-8fac-f8ece4db39aa/jobs/35329/parallel-runs/0/steps/0-102
).

[CSDK-221]: https://revenuecats.atlassian.net/browse/CSDK-221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1903, opened 2022-09-13
`spm build` CI job: changed to release build

NachoSoto:
See #1860.

We want to ensure that the framework can be correctly compiled with SPM, so I think it makes sense to ensure that the _RELEASE_ version of it can be compiled specifically.
We have a lot of coverage of `DEBUG` builds basically by most other CI jobs, so this provides coverage of the release build as well.

See #1902 for example for the type of change that could break release builds if not covered in CI.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -221,7 +221,7 @@ jobs:
     steps:
       - checkout
       - run:
-          name: SPM Build
+          name: SPM Release Build
           command: swift build -c release
           no_output_timeout: 30m
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -222,7 +222,7 @@ jobs:
       - checkout
       - run:
           name: SPM Build
-          command: swift build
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -222,7 +222,7 @@ jobs:
       - checkout
       - run:
           name: SPM Build
-          command: swift build
+          command: swift build -c release
```
tonidero, line 225: Should we just have both? A `spm-build-debug` and a `spm-build-release` jobs. But well, if this is just building the SDK, as you said, that might be covered there so I'm ok keeping only the release version, but maybe renaming the job is still worth it.

NachoSoto: Yeah I'll rename it.
Like I said in the description I think we cover debug builds a lot in every other job.

NachoSoto: Done.

tonidero (approved):

PR #1904, opened 2022-09-13
`StoreKitUnitTests`:  compile on iOS 11.0+

NachoSoto:
The entire target had a deployment target of iOS 14.0 because `SKTestSession` requires it. However, not all tests in that target use `StoreKitConfigTestCase`, and those weren't running before iOS 14.0.

To fix this, I've removed the deployment target overrides from all targets (expect for integration tests, those have a higher one for simplicity), and instead defined them only once at the *project level*, so all targets inherit that automatically.

`StoreKitConfigTestCase` will skip the test automatically if the iOS is below 14.0, but the other tests will now run on iOS 12 and 13.


RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -3218,15 +3184,19 @@
...
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = iphoneos;
 				SUPPORTS_MACCATALYST = YES;
 				SWIFT_OPTIMIZATION_LEVEL = "-Onone";
 				SWIFT_STRICT_CONCURRENCY = targeted;
 				SWIFT_TREAT_WARNINGS_AS_ERRORS = YES;
+				TVOS_DEPLOYMENT_TARGET = 11.0;
 				VERSIONING_SYSTEM = "apple-generic";
 				VERSION_INFO_PREFIX = "";
+				WATCHOS_DEPLOYMENT_TARGET = 6.2;
```
NachoSoto, line 3199: ✅  These are now at the project level instead of duplicated in each target.

Tests/StoreKitUnitTests/TestHelpers/StoreKitTestHelpers.swift
```diff
@@ -69,44 +70,44 @@ extension StoreKitConfigTestCase {
     ) async {
         self.testSession.storefront = new
 
-        if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
```
NachoSoto, line 72: I *think* this isn't necessary, but let's see if tests pass without it.

Tests/UnitTestsHostApp/SceneDelegate.swift
```diff
@@ -1,60 +0,0 @@
...
-//
-//      https://opensource.org/licenses/MIT
-//
-//  SceneDelegate.swift
-//
-//  Created by Andrés Boedo on 9/24/21.
-
-import UIKit
-
-class SceneDelegate: UIResponder, UIWindowSceneDelegate {
```
NachoSoto, line 16: This wasn't needed in the host app (and it's iOS 14.0+ only, so I just removed it).

Tests/StoreKitUnitTests/StoreKitConfigTestCase.swift
```diff
@@ -36,16 +37,20 @@ class StoreKitConfigTestCase: TestCase {
     var userDefaults: UserDefaults!
 
     override func setUpWithError() throws {
-        testSession = try SKTestSession(configurationFileNamed: "UnitTestsConfiguration")
-        testSession.resetToDefaultState()
-        testSession.disableDialogs = true
-        testSession.clearTransactions()
+        try super.setUpWithError()
+
+        try AvailabilityChecks.iOS14APIAvailableOrSkipTest()
```
tonidero, line 42: Just a question, can this still be called before iOS 14 with the `@available` annotation?

tonidero (approved):
It makes sense to me

Tests/StoreKitUnitTests/StoreKitConfigTestCase.swift
NachoSoto: It can. The annotation is only for compilation, but `XCTest` is not very smart and runs the tests anyway. That's why we have `AvailabilityChecks`.

Tests/StoreKitUnitTests/TrialOrIntroPriceEligibilityCheckerSK1Tests.swift
```diff
@@ -16,7 +16,9 @@ import Nimble
 import StoreKitTest
 import XCTest
 
-// swiftlint:disable:next type_name
+// swiftlint:disable type_name
```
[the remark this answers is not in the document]
NachoSoto: This file only has one class (like most test classes) so I think that's unnecessary. Extensions are common to be added at the bottom, but they don't have "type names" so this wouldn't apply.

.circleci/config.yml
```diff
@@ -574,10 +574,8 @@ workflows:
           xcode_version: '13.4.1'
       - run-test-ios-13:
           xcode_version: '13.4.1'
-          <<: *release-branches-and-main
```
NachoSoto, line 577: Oops I forgot to remove this. Re-added in #1909

PR #1907, opened 2022-09-15
Created `CachingProductsManager` to provide consistent caching logic when fetching products

NachoSoto:
Fixes [SDKONCALL-129].

### Goal

The main feature that this change introduces is providing a _request_ cache to `ProductsFetcherSK2` (like `ProductsFetcherSK1` has).
It already implements a _product_ cache, but concurrent requests for the same products result in fetching those multiple times. It does not cache products until they're done being fetched.

This was observed in the logs when doing something simple like purchasing a product. This is running `testCanPurchaseProduct` integration test:
> DEBUG: ℹ️ No existing products cached, starting store products request for: ["com.revenuecat.monthly_4.99.1_week_intro", "com.revenuecat.annual_39.99.2_week_intro", "com.revenuecat.weekly_1.99.3_day_intro"]
DEBUG: ℹ️ No existing products cached, starting store products request for: ["com.revenuecat.monthly_4.99.1_week_intro", "com.revenuecat.annual_39.99.2_week_intro", "com.revenuecat.weekly_1.99.3_day_intro"]
DEBUG: 😻 Store products request request received response
DEBUG: ℹ️ Store products request finished
DEBUG: 😻 Store products request request received response
DEBUG: ℹ️ Store products request finished

With this PR, we avoid the double request due to that race condition:

> DEBUG: ℹ️ No existing products cached, starting store products request for: ["com.revenuecat.annual_39.99.2_week_intro", "com.revenuecat.weekly_1.99.3_day_intro", "com.revenuecat.monthly_4.99.1_week_intro"]
DEBUG: 😻 Store products request request received response
DEBUG: ℹ️ Store products request finished

### How

Instead of implementing that logic _only_ on `ProductsFetcherSK2` or in `ProductsManager`, this aims to keep the implementation simple in `ProductsManager` and each of the fetchers, and instead uses the [decorator pattern](https://en.wikipedia.org/wiki/Decorator_pattern) to provide that new functionality in a way that's simple to test and maintain, independent of the logic for actually fetching products.

To accomplish that, this adds a new `protocol`: `ProductsManagerType`, with 2 main methods (more about why the need for an SK2 special method below):
```swift
/// Fetches the ``StoreProduct``s with the given identifiers
/// The returned products will be SK1 or SK2 backed depending on the implementation and configuration.
func products(withIdentifiers identifiers: Set<String>, completion: @escaping Completion)

/// Fetches the `SK2StoreProduct`s with the given identifiers.
func sk2Products(withIdentifiers identifiers: Set<String>, completion: @escaping SK2Completion)
```

Those are only completion-block APIs. To avoid repetition, an `extension` on the `protocol` itself provides an `async` API, wrapping on top of these 2 methods, regardless of which `ProductsManagerType` they're being used on.

The hierarchy is now as follows:
- `CachingProductsManager`
    - `ProductsManager`
        - `ProductsFetcherSK1`
        - `ProductsFetcherSK2`

#1908 gets rid of the now duplicate caching logic in `ProductsManager` and the fetchers.

### Other notes

What makes the implementation a bit more convoluted is the need for a special SK2 method.
This is necessary specifically for `TrialOrIntroPriceEligibilityChecker` which requires fetching only SK2 products regardless of `StoreKit2Setting`. In order to guarantee this invariant at compile time through types, a separate cache is maintained with `SK2StoreProduct`s. However, the main caching logic is *not* duplicated and instead shared for both.

The main tests for `CachingProductsManager` are simple unit tests that don't make use of `StoreKitConfigTestCase`.
Because creating SK2 products does require that, I added a separate test class for those.

### Future changes

The caching logic is very simple, just like what we had for the fetchers. One missing piece is that if products A and B are requested, and only A is in the cache, this is treated as a complete cache miss.

The fetchers had the same problem, but now that this caching logic is isolated and much easier to test, we can evolve and solve that feature in a more maintainable way.

[SDKONCALL-129]: https://revenuecats.atlassian.net/browse/SDKONCALL-129?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/CachingProductsManager.swift
+++ b/Sources/Purchasing/CachingProductsManager.swift
@@ -1,0 +1,188 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  CachingProductsManager.swift
+//
+//  Created by Nacho Soto on 9/14/22.
+
+import Foundation
+
+/// `ProductsManagerType` decorator that adds caching behavior on each request.
+/// The product results are cached, and it avoids performing concurrent duplicate requests for the same products.
+final class CachingProductsManager {
+
+    private let manager: ProductsManagerType
+
+    private let productCache: Atomic<[String: StoreProduct]> = .init([:])
+    private let requestCache: Atomic<[Set<String>: [Completion]]> = .init([:])
+
+    #if swift(>=5.7)
+    private let _sk2ProductCache: (any Sendable)?
+    private let _sk2RequestCache: (any Sendable)?
+    #else
+    private let _sk2ProductCache: any
+    private let _sk2RequestCache: any
+    #endif
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    private var sk2ProductCache: Atomic<[String: SK2StoreProduct]> {
+        // swiftlint:disable:next force_cast
+        return self._sk2ProductCache as! Atomic<[String: SK2StoreProduct]>
+    }
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    private var sk2RequestCache: Atomic<[Set<String>: [SK2Completion]]> {
+        // swiftlint:disable:next force_cast
+        return self._sk2RequestCache as! Atomic<[Set<String>: [SK2Completion]]>
+    }
+
+    init(manager: ProductsManagerType) {
+        assert(!(manager is CachingProductsManager), "Decorating CachingProductsManager with itself")
+
+        self.manager = manager
+
+        if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) {
+            self._sk2ProductCache = Atomic<[String: SK2StoreProduct]>([:])
+            self._sk2RequestCache = Atomic<[Set<String>: [SK2Completion]]>([:])
+        } else {
+            self._sk2ProductCache = nil
+            self._sk2RequestCache = nil
+        }
+    }
+
+    func clearCache() {
+        self.productCache.value.removeAll(keepingCapacity: true)
+        if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) {
+            self.sk2ProductCache.value.removeAll(keepingCapacity: true)
+        }
+
+        self.manager.clearCache()
+    }
+
+}
+
+extension CachingProductsManager: ProductsManagerType {
+
+    func products(withIdentifiers identifiers: Set<String>, completion: @escaping Completion) {
+        Self.products(with: identifiers,
+                      completion: completion,
+                      productCache: self.productCache,
+                      requestCache: self.requestCache) { identifiers, completion in
+            self.manager.products(withIdentifiers: identifiers, completion: completion)
+        }
+    }
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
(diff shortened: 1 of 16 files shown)


NachoSoto:
@RevenueCat/core-sdk bump

Tests/UnitTests/Purchasing/CachingProductsManagerTests.swift
```diff
@@ -0,0 +1,160 @@
...
+//
+//  Created by Nacho Soto on 9/14/22.
+
+import Nimble
+@testable import RevenueCat
+import XCTest
+
+@available(iOS 13.0, tvOS 13.0, watchOS 6.2, macOS 10.15, *)
+@MainActor
+class CachingProductsManagerTests: TestCase {
```
tonidero, line 20: To make this more explicit, should we rename this to `StoreKit1CachingProductsManagerTests`?

Tests/UnitTests/Purchasing/Purchases/PurchasesDeferredPurchasesTests.swift
```diff
@@ -136,7 +136,8 @@ class PurchaseDeferredPurchasesTests: BasePurchasesTests {
         self.storeKit1WrapperDelegate.storeKit1Wrapper(self.storeKit1Wrapper, updatedTransaction: transaction)
 
         expect(self.mockProductsManager.invokedCacheProduct) == true
-        expect(self.mockProductsManager.invokedCacheProductParameter) == self.product
+        expect(self.mockProductsManager.invokedCacheProductParameter.map(StoreProduct.from(product:)))
+        == StoreProduct(sk1Product: self.product)
```
tonidero, line 140: Hmm I'm wondering, why do we need to wrap the. products with `StoreProduct`?

tonidero (approved):
Just a few questions but looking great!

Sources/Purchasing/CachingProductsManager.swift
```diff
@@ -0,0 +1,188 @@
...
+        if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) {
+            self._sk2ProductCache = Atomic<[String: SK2StoreProduct]>([:])
+            self._sk2RequestCache = Atomic<[Set<String>: [SK2Completion]]>([:])
+        } else {
+            self._sk2ProductCache = nil
+            self._sk2RequestCache = nil
+        }
+    }
+
+    func clearCache() {
```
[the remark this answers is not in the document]
NachoSoto: That would break requests that are currently in progress.

Sources/Purchasing/CachingProductsManager.swift
```diff
@@ -0,0 +1,188 @@
...
+            Logger.debug(
+                Strings.offering.products_already_cached(
+                    identifiers: Set(cachedProducts.map { $0.productIdentifier})
+                )
+            )
+
+            completion(.success(cachedProducts))
+        } else {
+            let requestInProgress = Self.save(completion, for: identifiers, requestCache: requestCache)
+            guard !requestInProgress else { return }
```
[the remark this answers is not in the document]
NachoSoto: Yeah that's a good idea.

Tests/UnitTests/Purchasing/CachingProductsManagerTests.swift
NachoSoto: This test suite is testing `CachingProductsManager` independently of what the underlying `ProductsManager` is, so it would be confusing to call it SK1 since it works just the same for SK2.

Tests/UnitTests/Purchasing/Purchases/PurchasesDeferredPurchasesTests.swift
NachoSoto: This type is `StoreProductType` which isn't `Equatable`.

PR #1908, opened 2022-09-15
`ProductsFetcherSK2`: removed now redundant caching logic

NachoSoto:
Follow up to #1907.

This logic is now implemented and shared in `CachingProductsManager`.

I've decided to keep the logic in `ProductsFetcherSK1` because it's intertwined with the retry mechanism, and it wasn't trivial to remove caching while leaving that logic.

### DIFF
--- a/Sources/Purchasing/ProductsManager.swift
+++ b/Sources/Purchasing/ProductsManager.swift
@@ -108,31 +108,11 @@ class ProductsManager: NSObject, ProductsManagerType {
         }
     }
 
-    func cache(_ product: StoreProductType) {
-        switch product {
-        case let sk1Product as SK1Product:
-            self.productsFetcherSK1.cacheProduct(sk1Product)
-
-        case let product as StoreProduct:
-            if let sk1Product = product.sk1Product {
-                self.productsFetcherSK1.cacheProduct(sk1Product)
-            }
-
-        default:
-            // Ignoring otherwise.
-            // This is temporary, as the caching logic is going away from here
-            // in favor of the new `CachingProductsManager`
-            break
-        }
-    }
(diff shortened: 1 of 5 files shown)


Sources/Purchasing/ProductsManager.swift
```diff
@@ -108,31 +108,11 @@ class ProductsManager: NSObject, ProductsManagerType {
...
+    // See `CachingProductsManager`.
+    func cache(_ product: StoreProductType) {}
     func clearCache() {
-        if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *),
-           self.systemInfo.storeKit2Setting == .enabledForCompatibleDevices {
-            self.invalidateAndReFetchCachedSK2Products()
-        } else {
-            self.invalidateAndReFetchCachedSK1Products()
-        }
+        self.productsFetcherSK1.clearCache()
```
tonidero, line 115: I guess we want to remove the caching from the `ProductsFetcherSK1` in the future, since we are now caching it in the `CachingProductsManager`. Did we create a Jira task for it?

tonidero (approved):
Looks good!

Sources/Purchasing/ProductsManager.swift
NachoSoto: Yeah that's why I wrote this:
> I've decided to keep the logic in ProductsFetcherSK1 because it's intertwined with the retry mechanism, and it wasn't trivial to remove caching while leaving that logic.

We could file a Jira to do it in the future, but I wonder if it's worth spending the time on that if SK1 is going away eventually /cc @aboedo thoughts?

PR #1909, opened 2022-09-15
`CircleCI`: changed all jobs to use Xcode 14

NachoSoto:
Now that [Xcode 14 RC](https://discuss.circleci.com/t/xcode-14-rc-released/45334) is out, we should make it the default for all the jobs moving forward.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -579,6 +579,8 @@ workflows:
           xcode_version: '14.0.0'
           <<: *release-branches-and-main
       - build-tv-watch-and-macos:
+          # This is the only job not using Xcode 14 yet
+          # because it does not contain the new macOS SDK until Ventura is release.
           xcode_version: '13.4.1'
       - release-checks:
           xcode_version: '14.0.0'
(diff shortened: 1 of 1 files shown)


NachoSoto:
Let's see if they all pass though 🍿

NachoSoto:
Ugh. Mac building fails because Xcode 14 doesn't include the new SDK, and so checking for swift's version is the wrong check... I'll downgrade that job for now.

.circleci/config.yml
```diff
@@ -561,43 +561,45 @@ workflows:
...
       - run-test-ios-13:
-          xcode_version: '13.4.1'
+          xcode_version: '14.0.0'
+          <<: *release-branches-and-main
       - run-test-ios-12:
-          xcode_version: '13.4.1'
+          xcode_version: '14.0.0'
+          <<: *release-branches-and-main
       - build-tv-watch-and-macos:
           xcode_version: '13.4.1'
```
tonidero, line 582: Should we add a comment here to revisit it in the future?

tonidero (approved):
LGTM!

.circleci/config.yml
NachoSoto: Good idea, added.

PR #1869, opened 2022-08-29
Fixed crash on `async` SK1 cancelled purchase

NachoSoto:
Fixes #1868.
See https://github.com/RevenueCat/purchases-ios/issues/1868#issuecomment-1229614774 for a summary of the changes in #1841.

### Before #1841:
- SK1 / completion-block: no `CustomerInfo` + cancelled + `ErrorCode.purchaseCancelledError`
- SK1 / `async`: thrown `ErrorCode.purchaseCancelledError`
- SK2 / completion-block: `CustomerInfo` + cancelled + no error
- SK2 / `async`: `CustomerInfo` + cancelled (not error)

### After #1841:
- SK1 / completion-block:  no `CustomerInfo` + cancelled + `ErrorCode.purchaseCancelledError`
- SK1 / `async`: no `CustomerInfo` + ignored error -> crash! :rotating_light:
- SK2 / completion-block: `CustomerInfo` + cancelled + `ErrorCode.purchaseCancelledError`
- SK2 / `async`: `CustomerInfo` + cancelled (not error)

### After this PR:
- SK1 / completion-block:  no `CustomerInfo` + cancelled + `ErrorCode.purchaseCancelledError`
- **SK1 / `async`: thrown `ErrorCode.purchaseCancelledError`** :warning:
- SK2 / completion-block: `CustomerInfo` + cancelled + `ErrorCode.purchaseCancelledError`
- SK2 / `async`: `CustomerInfo` + cancelled (not error)

The change in #1841 was slightly incorrect. The `ignoreIfPurchaseCancelled` didn't make a lot of sense because `Result(value:error:)` already ignored the error if the value wasn't `nil`.

This change still doesn't get us to a fully consistent behavior across SK1 and SK2 but it adds coverage for this crash and temporarily fixes it.

PR #1910, opened 2022-09-16
`StoreKit 1`: changed result of cancelled purchases to be consistent with `StoreKit 2`

NachoSoto:
Fixes [CSDK-464]. Follow up to #1869.

SK1 and SK2 had inconsistent behavior during cancelled purchases.

### Before this PR:
- SK1 / completion-block:  no `CustomerInfo` + cancelled + `ErrorCode.purchaseCancelledError`
- **SK1 / `async`: thrown `ErrorCode.purchaseCancelledError`** :warning:
- SK2 / completion-block: `CustomerInfo` + cancelled + `ErrorCode.purchaseCancelledError`
- SK2 / `async`: `CustomerInfo` + cancelled (not error)

### After this PR:

The output is now consistent across the board:

- SK1 / completion-block:  `CustomerInfo` + cancelled + `ErrorCode.purchaseCancelledError`
- SK1 / `async`: `CustomerInfo` + cancelled (not error)
- SK2 / completion-block: `CustomerInfo` + cancelled + `ErrorCode.purchaseCancelledError`
- SK2 / `async`: `CustomerInfo` + cancelled (not error)

## Notes

This is of course a somewhat backwards incompatible change, but this output is now what makes the most sense based on the types. **This would only break existing code if users are relying on the existence of `CustomerInfo` alone without checking `error` or `userCancelled`.**

- Completion-block API:

There's 2 ways of detecting cancellation, both valid and reasonable. Either checking that `cancelled` is `true`, or comparing `error` with `ErrorCode.purchaseCancelledError`

- `async` API:

A cancelled purchase no longer throws. This didn't make sense in SK1, because it meant that `userCancelled` was always `false` in the case that the purchase method returned. This also matches `StoreKit`'s  behavior in `Product.purchase()`

[CSDK-464]: https://revenuecats.atlassian.net/browse/CSDK-464?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -640,10 +640,10 @@ private extension PurchasesOrchestrator {
 
             if isCancelled {
                 self.customerInfoManager.customerInfo(appUserID: self.appUserID,
-                                                      fetchPolicy: .cachedOrFetched) { result in
+                                                      fetchPolicy: .cachedOrFetched) { customerInfo in
                     self.operationDispatcher.dispatchOnMainActor {
                         completion(storeTransaction,
-                                   result.value,
+                                   customerInfo.value,
                                    purchasesError.asPublicError,
                                    true)
                     }
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -636,11 +636,25 @@ private extension PurchasesOrchestrator {
            let completion = self.getAndRemovePurchaseCompletedCallback(forTransaction: storeTransaction) {
             let purchasesError = ErrorUtils.purchasesError(withSKError: error)
 
-            self.operationDispatcher.dispatchOnMainActor {
-                completion(storeTransaction,
-                           nil,
-                           purchasesError.asPublicError,
-                           purchasesError.isCancelledError)
+            let isCancelled = purchasesError.isCancelledError
+
+            if isCancelled {
+                self.customerInfoManager.customerInfo(appUserID: self.appUserID,
+                                                      fetchPolicy: .cachedOrFetched) { result in
+                    self.operationDispatcher.dispatchOnMainActor {
+                        completion(storeTransaction,
+                                   result.value,
+                                   purchasesError.asPublicError,
+                                   true)
+                    }
+                }
+            } else {
+                self.operationDispatcher.dispatchOnMainActor {
+                    completion(storeTransaction,
+                               nil,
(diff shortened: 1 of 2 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -636,11 +636,25 @@ private extension PurchasesOrchestrator {
...
-                           purchasesError.asPublicError,
-                           purchasesError.isCancelledError)
+            let isCancelled = purchasesError.isCancelledError
+
+            if isCancelled {
+                self.customerInfoManager.customerInfo(appUserID: self.appUserID,
+                                                      fetchPolicy: .cachedOrFetched) { result in
+                    self.operationDispatcher.dispatchOnMainActor {
+                        completion(storeTransaction,
+                                   result.value,
```
tonidero, line 646: Maybe rename `result` to `customerInfo`?
```suggestion
                                                      fetchPolicy: .cachedOrFetched) { customerInfo in
                    self.operationDispatcher.dispatchOnMainActor {
                        completion(storeTransaction,
                                   customerInfo.value,
```

Tests/UnitTests/Purchasing/Purchases/PurchasesPurchasingTests.swift
```diff
@@ -334,13 +334,13 @@ class PurchasesPurchasingTests: BasePurchasesTests {
...
-        _ = Task {
+        // Need to do this async so the code below can invoke the `updatedTransaction` delegate method.
+        _ = Task<Void, Never> {
             do {
-                _ = try await self.purchases.purchase(
-                    product: product
-                )
+                result = try await self.purchases.purchase(product: product)
             } catch {
                 receivedError = error as NSError
```
tonidero, line 345: Could we replace this with a `fail`?

tonidero (approved):
I can see how this changes the behavior and users could potentially be relying in the old behavior... While I wouldn't consider it a breaking change, we should mention this prominently in the Changelog on the next release.

Tests/UnitTests/Purchasing/Purchases/PurchasesPurchasingTests.swift
NachoSoto: No because it's inside of a `Task` so it's happening asynchronously.
There's a comment above explaining why it has to be this way:
> // Need to do this async so the code below can invoke the `updatedTransaction` delegate method.

NachoSoto:
> we should mention this prominently in the Changelog on the next release.

💯

PR #1292, opened 2022-02-13
`GetCustomerInfoOperation`: improve logic for caching responses

NachoSoto:
Fixes [sc-12802].

## Background

Before this change, the sequence illustrated by the new test `testGetsUpdatedSubscriberInfoAfterPost` (started in #1249, with a few modifications here) would fail:
- GET /subscribers/
- POST /receipts
- GET /subscribers/

#1261 fixed the race condition in the test, making it so that operations are actually run serially.
However, when the first GET request finished, the second one shared the cache key, so it received the (soon to be) outdated `CustomerInfo`.

## The fix

Unfortunately this fix is isolated for this specific issue, and not a generalized one. A more comprehensive rewrite of `HTTPClient` is needed ([sc-9521]).
I came up with several other alternatives, but what I ended up going with here involves the least amount of moving pieces.

If a `GetCustomerInfoOperation` request is enqueued with `PostReceiptDataOperation` in the queue, then a new unique cache key is created to avoid past `GetCustomerInfoOperation` to share outdated data.

If we simply added a random number to the `GetCustomerInfoOperation` cache key, the new test would have also passed. No other test verified the caching behavior, which is what the new `testGetSubscriberInfoDoesNotMakeTwoRequests` checks.

PR #1911, opened 2022-09-16
`GetCustomerInfoAPI`: avoid making a request if there's any `PostReceiptDataOperation` in progress

NachoSoto:
Fixes [CSDK-419].

This is an improvement over the fix in #1292. In there, we avoided appending on an existing `CustomerInfoOperation` if there were any `PostReceiptDataOperation`s in progress, because those might come back with outdated entitlements.
This fix removes that cache key hack, and instead *reuses* the entire response to that `PostReceiptDataOperation` by "stealing" that request's cache key.

This is perfectly captured by the existing test `BackendPostReceiptDataTests.testGetsUpdatedSubscriberInfoAfterPost`, added in #1292. Amazingly enough, that test still passes with one minor change: the entire process requires one fewer API call :tada: 🐼

[CSDK-419]: https://revenuecats.atlassian.net/browse/CSDK-419?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Networking/Caching/CustomerInfoCallback.swift
+++ b/Sources/Networking/Caching/CustomerInfoCallback.swift
@@ -17,14 +17,52 @@ struct CustomerInfoCallback: CacheKeyProviding {
 
     typealias Completion = (Result<CustomerInfo, BackendError>) -> Void
 
-    let cacheKey: String
-    let source: NetworkOperation.Type
-    let completion: Completion
+    var cacheKey: String
+    var source: NetworkOperation.Type
+    var completion: Completion
 
     init(operation: CacheableNetworkOperation, completion: @escaping Completion) {
         self.cacheKey = operation.cacheKey
         self.source = type(of: operation)
         self.completion = completion
+    }
+
+}
+
+extension CustomerInfoCallback {
+
+    func withNewCacheKey(_ newKey: String) -> Self {
+        var copy = self
+        copy.cacheKey = newKey
+
+        return copy
+    }
+
+}
+
+// MARK: - CallbackCache helpers
+
+extension CallbackCache where T == CustomerInfoCallback {
+
+    func addOrAppendToPostReceiptDataOperation(callback: CustomerInfoCallback) -> CallbackCacheStatus {
+        if let existing = self.callbacks(ofType: PostReceiptDataOperation.self).last {
+            return self.add(callback: callback.withNewCacheKey(existing.cacheKey))
+        } else {
+            return self.add(callback: callback)
+        }
+    }
+
+    private func callbacks(ofType type: NetworkOperation.Type) -> [T] {
+        return self
+            .cachedCallbacksByKey
+            .value
+            .lazy
+            .flatMap(\.value)
+            .filter { $0.source == type }
+    }
+
+    private func callbackCount(ofType type: NetworkOperation.Type) -> Int {
+        return self.callbacks(ofType: type).count
     }
 
 }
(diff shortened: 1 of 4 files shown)


Sources/Networking/Caching/CustomerInfoCallback.swift
```diff
@@ -28,3 +28,41 @@ struct CustomerInfoCallback: CacheKeyProviding {
...
+    private func callbacks(ofType type: NetworkOperation.Type) -> [T] {
+        return self
+            .cachedCallbacksByKey
+            .value
+            .lazy
+            .flatMap(\.value)
+            .filter { $0.source == type }
+    }
+
+    private func callbackCount(ofType type: NetworkOperation.Type) -> Int {
```
tonidero, line 64: This seems unused?

Sources/Networking/Caching/CustomerInfoCallback.swift
```diff
@@ -17,9 +17,9 @@ struct CustomerInfoCallback: CacheKeyProviding {
 
     typealias Completion = (Result<CustomerInfo, BackendError>) -> Void
 
-    let cacheKey: String
-    let source: NetworkOperation.Type
-    let completion: Completion
+    var cacheKey: String
+    var source: NetworkOperation.Type
+    var completion: Completion
```
tonidero, line 22: Could we add another init instead of making this mutable? Seems safer to me but lmk if you think there are problems with that.

Sources/Networking/Caching/CustomerInfoCallback.swift
```diff
@@ -28,3 +28,41 @@ struct CustomerInfoCallback: CacheKeyProviding {
     }
 
 }
+
+extension CustomerInfoCallback {
+
+    func withNewCacheKey(_ newKey: String) -> Self {
```
tonidero, line 34: Should this (or the extension) be `private`?

tonidero (approved):
A couple of things but looking good!

Sources/Networking/Caching/CustomerInfoCallback.swift
NachoSoto: I'd recommend reading on Swift's value types. This isn't like having mutable data in other languages, it's only mutable for mutables copies, and even then, because it doesn't have reference semantics, you can't have spooky action at a distance by changing one "instance" and mutating something else. So it's perfectly safe.

Sources/Networking/Caching/CustomerInfoCallback.swift
NachoSoto: Oh yeah that was for the old implementation. Thanks.

Sources/Networking/Caching/CustomerInfoCallback.swift
NachoSoto: Good call, dine,

NachoSoto:
@aboedo I'll wait for you since this is a pretty critical change and I want to make sure you're ok with it too.

Tests/UnitTests/Networking/Backend/BackendPostReceiptDataTests.swift
```diff
@@ -432,8 +432,7 @@ class BackendPostReceiptDataTests: BaseBackendTests {
 
         expect(self.httpClient.calls.map { $0.request.path }) == [
             getCustomerInfoPath,
-            .postReceiptData,
-            getCustomerInfoPath
```
NachoSoto, line 436: 🐼

PR #1913, opened 2022-09-20
`Purchases.beginRefundRequest`: ensured errors are `PublicError`

NachoSoto:
This was missed in #1879. Without this, we might have been forwarding some private errors instead of ensuring that we only sent `CodeError`s

### DIFF
--- a/Sources/Error Handling/ErrorUtils.swift
+++ b/Sources/Error Handling/ErrorUtils.swift
@@ -286,7 +286,6 @@ enum ErrorUtils {
      * If the error is already a `PurchasesError`, this simply returns the same value,
      * otherwise it wraps it into a ``ErrorCode/unknownError``.
      */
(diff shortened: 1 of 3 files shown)


Sources/Error Handling/ErrorUtils.swift
```diff
@@ -286,7 +286,6 @@ enum ErrorUtils {
      * If the error is already a `PurchasesError`, this simply returns the same value,
      * otherwise it wraps it into a ``ErrorCode/unknownError``.
      */
-    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
```
tonidero, line 289: Leftover from before?

tonidero (approved):
LGTM!

Sources/Error Handling/ErrorUtils.swift
NachoSoto: Yeah it was only used from another method with the same availability so I didn't notice it didn't need it.

PR #1916, opened 2022-09-20
`PurchaseTesterSwiftUI`: removed unnecessary `UIApplicationDelegate`

NachoSoto:
Using `init` in the `@main` app is the `SwiftUI` way of setting up `Purchases`

### DIFF
--- a/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/PurchaseTesterApp.swift
+++ b/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/PurchaseTesterApp.swift
@@ -11,7 +11,18 @@ import RevenueCat
 
 @main
 struct PurchaseTesterApp: App {
-    @UIApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
+
+    private let delegate = Delegate()
+
+    init() {
+        Purchases.logLevel = .debug
+        Purchases.configure(with: Configuration
+            .builder(withAPIKey: Constants.apiKey)
+            .with(usesStoreKit2IfAvailable: true)
+            .build())
+        Purchases.shared.delegate = self.delegate
+    }
+
     @State private var revenueCatCustomerData = RevenueCatCustomerData()
     
     var body: some Scene {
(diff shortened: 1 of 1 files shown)


Tests/TestingApps/PurchaseTesterSwiftUI/Shared/PurchaseTesterApp.swift
```diff
@@ -11,7 +11,18 @@ import RevenueCat
 
 @main
 struct PurchaseTesterApp: App {
-    @UIApplicationDelegateAdaptor(AppDelegate.self) var appDelegate
+
+    private let delegate = Delegate()
```
tonidero, line 15: Hmm I wonder, could we make this a local variable? Currently it's only used inside the `init` so we could just create it there I think.

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/PurchaseTesterApp.swift
```diff
@@ -11,7 +11,18 @@ import RevenueCat
...
+
+    init() {
+        Purchases.logLevel = .debug
+        Purchases.configure(with: Configuration
+            .builder(withAPIKey: Constants.apiKey)
+            .with(usesStoreKit2IfAvailable: true)
+            .build())
```
tonidero, line 22: Nitpick. I know this is how it was before but it was a bit hard to read at the beginning, should we store the configuration in a variable and pass that to the `configure` method? Would improve readability a bit I think.

tonidero (approved):

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/PurchaseTesterApp.swift
NachoSoto: That wouldn't work. The delegate is `weak` so we do need to keep a reference around.

tonidero: Ahh you're totally right 👍

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/PurchaseTesterApp.swift
NachoSoto: Done.

PR #1917, opened 2022-09-20
`PurchaseTester`: allow HTTP requests and enable setting `ProxyURL`

NachoSoto:
Fixes [CSDK-457].

[CSDK-457]: https://revenuecats.atlassian.net/browse/CSDK-457?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Tests/TestingApps/PurchaseTesterSwiftUI/Shared/PurchaseTesterApp.swift
```diff
@@ -16,6 +16,11 @@ struct PurchaseTesterApp: App {
 
     init() {
         Purchases.logLevel = .debug
+
+        if let proxyURL = Constants.proxyURL {
+            Purchases.proxyURL = URL(string: proxyURL)!
+        }
```
tonidero, line 22: Should we also add this and the http support changes to the MagicWeather apps? Not sure of the use cases for this, so it might not be needed.

tonidero (approved):
Just a question but LGTM!

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/PurchaseTesterApp.swift
NachoSoto: That's more of an example app. I think the Jira was specific for these because these apps are more geared towards testing internal functionality. @aboedo might have more insight into the motivation for this change though, but I'd say we don't need to do that there.

aboedo: Yeah, exactly, these changes are meant for the purchase testing apps, which are internal.
The other ones are meant for customers to have an idea of what an app with RC installed looks like and have code that they can copy and paste

PR #1918, opened 2022-09-20
`CircleCI`: changed iOS 12/13 to use Xcode 13

NachoSoto:
Looks like this broke in #1909. Not sure why it's not finding those simulators, but we can just run those old versions with an older Xcode for now.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -573,10 +573,10 @@ workflows:
       - run-test-ios-14:
           xcode_version: '14.0.0'
       - run-test-ios-13:
-          xcode_version: '14.0.0'
+          xcode_version: '13.4.1'
           <<: *release-branches-and-main
       - run-test-ios-12:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -573,10 +573,10 @@ workflows:
       - run-test-ios-14:
           xcode_version: '14.0.0'
       - run-test-ios-13:
-          xcode_version: '14.0.0'
+          xcode_version: '13.4.1'
           <<: *release-branches-and-main
       - run-test-ios-12:
-          xcode_version: '14.0.0'
+          xcode_version: '13.4.1'
```
tonidero, line 579: Maybe add a comment explaining why we are not using xcode 14?

tonidero (approved):

.circleci/config.yml
NachoSoto: Good poitm

PR #1921, opened 2022-09-21
Documentation: removed references to deprecated `Purchases.configure` methods

NachoSoto:


aboedo (commented):
I am now realizing that I never documented this, but the reason I had these deprecated sections is because if you don't, public methods will still get a link in the docs in the Purchases type, but with nothing telling you that they're deprecated (at least not until you go into each of them)

| Without deprecated section | With deprecated section |
| :-: | :-: |
| Screen Shot 2022-09-21 at 4 15 58 PM | Screen Shot 2022-09-21 at 4 16 32 PM |

NachoSoto:
Ooooh makes sense. Closing this then.

PR #1922, opened 2022-09-21
`StoreKit 2`: enabled by default

NachoSoto:
The main change is `StoreKit2Setting.default` is now `enabledForCompatibleDevices`, but I've also made sure that all places in the codebase are derived from this value.


tonidero (approved):
Looks great to me!

PR #1924, opened 2022-09-21
`ErrorUtils`: added test to verify that `PublicError`s can be `catch`'d as `ErrorCode`

NachoSoto:
This is a key part of how we document error handling in the SDK, so it's important that it's covered in a test.
The second test also verifies that this same error contains all the metadata.

This is a follow up to #1879.


NachoSoto:
I was about to rely on this for the implementation of [CSDK-451], and realized I wasn't quite 100% sure this worked, even though it should.

Now we know we won't break this.

tonidero (approved):

PR #1925, opened 2022-09-21
`Integration Tests`: relaunch tests when retrying failures

NachoSoto:
I've observed that a lot of the time flaky tests still don't pass after a retry.
This change can probably decrease the likelyhood that a retry will still fail, by ensuring that we create a brand new `SKTestSession` in a new simulator session.


NachoSoto:
Welp flaky tests are still flaky

tonidero (approved):
Since test keeps failing, not sure if we want to merge it but looks good and shouldn't hurt. Also, would this slow down tests too much? I'm guessing spinning new processes would be slower.

NachoSoto:
This failed in 5 minutes vs another sample failure in `main` that took 9 minutes, so I'd say no.

PR #1926, opened 2022-09-21
`OperationQueue`: log debug message when requests are found in cache and skipped

NachoSoto:
Right now there is no way to know when a request is being reused. If this were to lead to a hard to debug issue, at lease we'll now have evidence for this in the logs.


aboedo (approved):

NachoSoto:
Huh the new test fails in `tvOS` 🧐

PR #1928, opened 2022-09-21
Fixed `tvOS` tests

NachoSoto:
Turns out we hadn't been running tests on `tvOS` for a while:
> [19:08:22]: Couldn't find any matching simulators for '["Apple TV (15.2)"]' - falling back to default simulator
> [19:08:22]: Found simulator "iPhone 13 Pro (15.5)"

I found this went I went to dig into [this failure](https://github.com/RevenueCat/purchases-ios/pull/1926#issuecomment-1254132253). Turns out that wasn't running on `tvOS` anyway.

This turns them back on, fixes the compilation issues on `tvOS` tests, and ensures that we don't run into this issue again.

### Changes:
- Renamed device in lane `test_tvos`
- Enabled `ensure_devices_found` on `iOS` and `tvOS` lanes. This will also ensure that we don't stop running tests on iOS XX because it doesn't find that simulator.
- Removed `UnitTestHostApp` storyboards and scene, since those aren't supported in `tvOS`, and aren't needed because there's no UI.
- Made `StoreKitUnitTests` and `UnitTestHostApp` targets universal
- No longer linking `AdServices` on `tvOS` in test targets
- Disabled attribution tests on `tvOS`. Those methods weren't being compiled, but we were testing that a message was being sent, even though it no longer was.
- Changed iOS 14 tests to use Xcode 13 due to https://github.com/xcpretty/xcode-install/issues/466 (same as #1918).


fastlane/Fastfile
```diff
@@ -121,6 +121,7 @@ platform :ios do
     scan(
       step_name: "scan - iPhone",
       device: ENV['SCAN_DEVICE'] || "iPhone 12 (15.2)",
+      ensure_devices_found: true,
```
NachoSoto, line 124: Thanks for the knowledge @joshdholtz!

.circleci/config.yml
```diff
@@ -288,6 +288,11 @@ jobs:
     steps:
       - checkout
       - install-dependencies
+      - install-and-create-sim:
+          install-name: iOS 14.5 Simulator
+          sim-device-type: iPhone-8
+          sim-device-runtime: iOS-14-5
+          sim-name: iPhone 8 (14.5)
```
NachoSoto, line 295: This was missing now in Xcode 14.0

.circleci/config.yml
```diff
@@ -235,7 +235,7 @@ jobs:
           command: bundle exec fastlane test_ios
           no_output_timeout: 30m
           environment:
-            SCAN_DEVICE: iPhone 13 (16.0)
+            SCAN_DEVICE: iPhone 14 (16.0)
```
NachoSoto, line 238: This was wrong too.

Tests/StoreKitUnitTests/BeginRefundRequestHelperTests.swift
```diff
@@ -60,37 +62,6 @@ class BeginRefundRequestHelperTests: TestCase {
...
-        #if os(watchOS) || os(macOS) || os(tvOS)
-        expectFatalError(expectedMessage: Strings.purchase.begin_refund_request_unsupported.description) {
-            _ = try await helper.beginRefundRequestforActiveEntitlement()
-        }
-        #else
-        throw XCTSkip("beginRefundRequestForEntitlement is not available for this test.")
-        #endif
-    }
-
-#if os(iOS)
```
joshdholtz, line 93: Why are these being deleted? 🤔

joshdholtz (approved):
Just one question on deleted tests but approving because I assume there is a good reason 😊

Tests/StoreKitUnitTests/BeginRefundRequestHelperTests.swift
NachoSoto: Oh I forgot to explain. The expectation wasn't being met (maybe it was before?). The methods aren't even being compiled anymore.

NachoSoto: Literally no code used that message string

PR #1929, opened 2022-09-21
`StoreKitIntegrationTests`: print `AppleReceipt` data whenever `verifyEntitlementWentThrough` fails

NachoSoto:
This will help us debug flaky tests.

Example:
> INFO: 😻💰 Purchased product - 'com.revenuecat.monthly_4.99.1_week_intro'
DEBUG: ℹ️ Force refreshing the receipt to get latest transactions from Apple.
DEBUG: ℹ️ Loaded receipt from url file:///Users/nachosoto/Library/Developer/CoreSimulator/Devices/A3576DC2-355E-45BA-B32C-D2C0A3811BB4/data/Containers/Data/Application/EDF3B7C8-A4BD-4629-813E-0934528AF02E/StoreKit/receipt
INFO: ℹ️ Receipt parsed successfully
INFO: ℹ️ Receipt content:
["inAppPurchases": [], "creationDate": 2022-09-21 21:00:07 +0000, "opaqueValue": 8 bytes, "applicationVersion": "1", "originalApplicationVersion": "<unknown>", "bundleId": "com.revenuecat.StoreKitTestApp", "expirationDate": 4001-01-01 00:00:00 +0000, "sha1Hash": 20 bytes]
/Users/nachosoto/dev/purchases-ios/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift:64: error: -[BackendIntegrationTests.StoreKit2IntegrationTests testCanPurchaseProduct] : failed - Expected Entitlement. Got: [:]
expected to have Dictionary<String, EntitlementInfo> with count 1, got 0
Actual Value: [:]

I've also added the receipt as a test attachment so we can download it from `CircleCI`:

Screenshot 2022-09-21 at 14 33 01

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1051,6 +1051,20 @@ internal extension Purchases {
         return self.storeKit1Wrapper != nil
     }
 
+    #if DEBUG
+
+    /// Returns the parsed `AppleReceipt`
+    ///
+    /// - Warning: this is only meant for integration tests, as a way to debug purchase failures.
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    func fetchReceipt(_ policy: ReceiptRefreshPolicy) async throws -> AppleReceipt? {
+        let receipt = await self.receiptFetcher.receiptData(refreshPolicy: policy)
+
+        return try receipt.map { try ReceiptParser().parse(from: $0) }
+    }
+
+    #endif
+
     /// - Parameter syncedAttribute: will be called for every attribute that is updated
     /// - Parameter completion: will be called once all attributes have completed syncing
     /// - Returns: the number of attributes that will be synced
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1053,7 +1053,7 @@ internal extension Purchases {
 
     #if DEBUG
 
-    /// Returns the parsed `AppleReceipt`
+    /// - Returns: the parsed `AppleReceipt`
     ///
     /// - Warning: this is only meant for integration tests, as a way to debug purchase failures.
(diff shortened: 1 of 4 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1051,6 +1051,20 @@ internal extension Purchases {
         return self.storeKit1Wrapper != nil
     }
 
+    #if DEBUG
+
+    /// - Returns: the parsed `AppleReceipt`
+    ///
+    /// - Warning: this is only meant for integration tests, as a way to debug purchase failures.
```
aboedo, line 1058: ❤️ 

I think this would be a useful thing to have in general with `#if debug` and just post it to debug logs. 
I bet it would save us a ton of work debugging customer issues

aboedo: we could have it print whenever a transaction goes through

aboedo (approved):

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: I thought about that. Are you not worried about performance implications of forcing a receipt refresh and parsing every time?

aboedo: Not too worried about it for debug builds. 
The refresh part we wouldn't really need to do, we'd just parse whatever we're about to send the backend. 
And the parsing is pretty fast in my experience. Haven't tested it again in a while, but the parser goes through the receipt only once, and leverages array slices to ensure we're not copying chunks of the receipt to different places to memory. 

The loading of the file from disk is probably the heaviest bit, and that's already done once, we could reuse it.

NachoSoto: [CSDK-478]

NachoSoto: Done in #1940. Instead of `#if DEBUG` I've checked `if Logger.logLevel == .debug` so that it can always be enabled if needed.

PR #1931, opened 2022-09-22
`SnapshotTesting`: require version 1.9.0 to keep supporting iOS 12/13 tests

NachoSoto:
[The new 1.10.0 version](https://github.com/pointfreeco/swift-snapshot-testing/releases/tag/1.10.0) requires iOS 13+, which is making CI builds all fail.

We'll be able to upgrade when we drop support for those version, but for now we're stuck with `1.9.0`


joshdholtz (approved):
LGTM 🚀

PR #1932, opened 2022-09-22
`CacheableNetworkOperation`: fixed race condition in new test

NachoSoto:
The test introduced by #1926 passed locally, but sometimes failed in CI due to a race condition.

What we're testing is that the second request reuses the callback of the first one, **if called while the first one is still in progress**. However, nothing in the test ensures that's the case.
This introduces an optional delay in `MockHTTPClient` so we can ensure that.

Another reason that test was failing randomly is because we had another test modifying global state, which interferes with every other test. We run tests in parallel, so it was making the new test that requires `debug` logs to fail randomly, if this test happened to run before.

_Note: because of the nature of these 2 tests, they don't actually wait for the whole response to finish, so they still pass in ~0.006s_

### DIFF
--- a/Tests/UnitTests/Purchasing/Purchases/PurchasesConfiguringTests.swift
+++ b/Tests/UnitTests/Purchasing/Purchases/PurchasesConfiguringTests.swift
@@ -209,17 +209,6 @@ class PurchasesConfiguringTests: BasePurchasesTests {
         expect(SystemInfo.serverHostURL) == defaultHostURL
     }
 
-    @available(*, deprecated) // Ignore deprecation warnings
-    func testSetDebugLogsEnabledSetsTheCorrectValue() {
-        Logger.logLevel = .warn
-
-        Purchases.debugLogsEnabled = true
-        expect(Logger.logLevel) == .debug
-
-        Purchases.debugLogsEnabled = false
-        expect(Logger.logLevel) == .info
-    }
-
     func testIsAnonymous() {
         setupAnonPurchases()
         expect(self.purchases.isAnonymous).to(beTrue())
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -128,6 +128,7 @@ platform :ios do
       number_of_retries: 5,
       result_bundle: true,
       testplan: "CI-AllTests",
+      configuration: 'Debug',
       output_directory: "fastlane/test_output/xctest/ios"
     )
   end
@@ -144,6 +145,7 @@ platform :ios do
       number_of_retries: 5,
       result_bundle: true,
       testplan: "CI-AllTests",
+      configuration: 'Debug',
       output_directory: "fastlane/test_output/xctest/tvos"
     )
   end
@@ -342,6 +344,7 @@ platform :ios do
       number_of_retries: 3,
       result_bundle: true,
       testplan: "CI-BackendIntegration",
+      configuration: 'Debug',
       output_directory: "fastlane/test_output/xctest/ios"
     )
   end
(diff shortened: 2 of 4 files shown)


NachoSoto:
Huh this is still failing in iOS 14 😶

NachoSoto:
For some reason debug logs aren't enabled when running the iOS 15 tests, but can't figure out why. We're compiling on debug, and `Logger` sets the default level to `.debug`... nothing else changes it.

NachoSoto:
Tests are still failing. I need to figure out why on iOS 15 the `Logger.logLevel` is being set to Release in CI.

NachoSoto:
In the mean time it would be nice to merge https://github.com/RevenueCat/purchases-ios/pull/1928 so we're running tests in the device we actually want them too.

joshdholtz (approved):
LGTM 🚀

Tests/UnitTests/Purchasing/Purchases/PurchasesConfiguringTests.swift
```diff
@@ -209,17 +209,6 @@ class PurchasesConfiguringTests: BasePurchasesTests {
...
-    func testSetDebugLogsEnabledSetsTheCorrectValue() {
-        Logger.logLevel = .warn
-
-        Purchases.debugLogsEnabled = true
-        expect(Logger.logLevel) == .debug
-
-        Purchases.debugLogsEnabled = false
-        expect(Logger.logLevel) == .info
-    }
-
```
NachoSoto, line 222: ⚠️  I'm removing this test to fix tests in CI for 2 reasons:
- It's testing a deprecated method
- It's modifying global state, which interferes with every other test. We run tests in parallel, so this is making the new test that requires debug logs to fail randomly, if this test happened to run before (it was changing the  level back to ` .info`).

fastlane/Fastfile
```diff
@@ -128,6 +128,7 @@ platform :ios do
       number_of_retries: 5,
       result_bundle: true,
       testplan: "CI-AllTests",
+      configuration: 'Debug',
```
NachoSoto, line 131: ⚠️ Explicitly setting these to Debug.

`Fastlane`'s `scan` documents that `Release` is the default. However, if we don't pass the value, `scan` wasn't passing any configuration to `xcodebuild`, so it was relying on the `scheme`'s default, which is `Debug`.
Instead of relying on that, which might change in the future, this makes the configuration explicit.

NachoSoto:
Can I get another review of this?

Tests/UnitTests/Purchasing/Purchases/PurchasesConfiguringTests.swift
aboedo: is there any way to
- back up the previous value
- run this atomically
- return the previous value?

aboedo (approved):

Tests/UnitTests/Purchasing/Purchases/PurchasesConfiguringTests.swift
aboedo: like, deprecated or not, if we break this, it's still a problem

NachoSoto: > run this atomically

We would have to block the entire `Logger` from logging anything, and / or block all other tests from running. I can't think of a way to do that.

NachoSoto:
Remaining failing tests are fixed by #1935.

Tests/UnitTests/Purchasing/Purchases/PurchasesConfiguringTests.swift
NachoSoto: > like, deprecated or not, if we break this, it's still a problem

For sure. I just can't think of a way to test this global state without messing with every other test.

NachoSoto: Do you have any ideas? :/

aboedo: I do not 😞

PR #1933, opened 2022-09-22
`AppleReceipt`: refactored declarations into nested types

NachoSoto:
Before this change we had a lot of global types. Now they're all nested inside of `AppleReceipt`, `InAppPurchase`, and `InAppPurchaseBuilder`.


tonidero (approved):
Nice cleanup!

PR #1934, opened 2022-09-22
`Logger`: refactored default `LogLevel` definition

NachoSoto:
Defining the property "twice" was a big weird and hard to read. Now `logLevel` is only defined once, with the default value extracted into a separate method.

### DIFF
--- a/Sources/Logging/Logger.swift
+++ b/Sources/Logging/Logger.swift
@@ -48,7 +48,7 @@ public typealias LogHandler = (_ level: LogLevel,
 
 enum Logger {
 
-    static var logLevel: LogLevel = Self.debugLogLevel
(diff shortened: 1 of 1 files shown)


Sources/Logging/Logger.swift
```diff
@@ -72,6 +69,14 @@ enum Logger {
 
     static var verbose: Bool = false
 
+    private static let debugLogLevel: LogLevel = {
```
tonidero, line 72: Should this be `defaultLogLevel`?

tonidero (approved):

Sources/Logging/Logger.swift
NachoSoto: OMG yes 😇

PR #1935, opened 2022-09-23
`BasePurchasesTests`: changed default back to SK1

NachoSoto:
Follow up to #1922. Most of the tests from the subclasses assume SK2 is disabled. This changes the default for those back to SK1.
The tests that do require SK2, define so explicitly.

I also replaced a bunch of force-unwraps with `XCTUnwrap` so tests don't crash (which is what was happening when running these tests with SK2 enabled).

### DIFF
--- a/Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
+++ b/Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
@@ -229,7 +229,7 @@ class BasePurchasesTests: TestCase {
 
     var storeKit2Setting: StoreKit2Setting {
         // Even though the new default is StoreKit 2, most of the tests from this parent class
-        // we written for SK1. Therefore we want to default to it being disabled.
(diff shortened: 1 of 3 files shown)


Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
```diff
@@ -225,6 +227,12 @@ class BasePurchasesTests: TestCase {
         self.storeKit1Wrapper.delegate?.storeKit1Wrapper(self.storeKit1Wrapper, updatedTransaction: transaction)
     }
 
+    var storeKit2Setting: StoreKit2Setting {
+        // Even though the new default is StoreKit 2, most of the tests from this parent class
+        // we written for SK1. Therefore we want to default to it being disabled.
```
aboedo, line 232:
```suggestion
        // were written for SK1. Therefore we want to default to it being disabled.
```

aboedo (approved):

PR #1936, opened 2022-09-23
Fixed iOS 12 tests

NachoSoto:
I blindly copy-pasted this snapshot and forgot that iOS 12 has different formatting. This was wrong in #1926.

It wasn't caught originally because we hadn't been running tests on iOS 12 until #1928.


.circleci/config.yml
```diff
@@ -582,7 +582,6 @@ workflows:
       - run-test-ios-12:
           # Simulator fails to install on Xcode 14
           xcode_version: '13.4.1'
-          <<: *release-branches-and-main
```
NachoSoto, line 585: ⚠️  REMOVE BEFORE MERGING

Just making sure this actually fixes the tests.

aboedo (approved):

PR #1942, opened 2022-09-27
Moved `InAppPurchase` into `AppleReceipt`

NachoSoto:
The name `InAppPurchase` is too general. This type in particular is specific to the content of `AppleReceipt`, so nesting it makes that clear.
Follow up to #1933.


tonidero (approved):
Looks good!

PR #1943, opened 2022-09-27
Abstracted conversion from `async` to completion-block APIs

NachoSoto:
Created `Async` with some shared methods that make this conversion simpler in a type-safe way.

Helps turn this:
```swift
_ = Task<Void, Never> {
    do {
        completion(.success(try await f()))
    } catch {
        completion(.failure(ErrorUtils.error())
    }
}
```

Into this:
```swift
Async.call(with: completion) {
    try await f()
}
```

### DIFF
--- a/Sources/FoundationExtensions/AsyncExtensions.swift
+++ b/Sources/FoundationExtensions/AsyncExtensions.swift
@@ -32,7 +32,7 @@ internal enum Async {
     }
 
     /// Invokes an `async throws` method and calls `completion` with the result.
-    /// /// Ensures that the returned error is `PurchasesError`.
+    /// Ensures that the returned error is `PurchasesError`.
     static func call<T>(
         with completion: @escaping (Result<T, PurchasesError>) -> Void,
         asyncMethod method: @escaping () async throws -> T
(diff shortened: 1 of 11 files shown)


Sources/FoundationExtensions/AsyncExtensions.swift
```diff
@@ -0,0 +1,59 @@
...
+            do {
+                completion(.success(try await method()))
+            } catch {
+                completion(.failure(ErrorUtils.purchasesError(withUntypedError: error).asPublicError))
+            }
+        }
+    }
+
+    /// Invokes an `async throws` method and calls `completion` with the result.
+    /// /// Ensures that the returned error is `PurchasesError`.
```
tonidero, line 35:
```suggestion
    /// Ensures that the returned error is `PurchasesError`.
```

tonidero (approved):
Nice cleanup!

Sources/FoundationExtensions/AsyncExtensions.swift
NachoSoto: Oops good catch!

PR #1945, opened 2022-09-27
`ReceiptFetcher`: added retry mechanism

NachoSoto:
Fixes [CSDK-478]

~~Depends on #1941, #1942, and #1943.~~

### Example:

Log of a successful purchase after retrying 🎉 

> INFO: 💰 Purchasing Product 'com.revenuecat.monthly_4.99.1_week_intro'
DEBUG: ℹ️ Loaded receipt from url file:///Users/nachosoto/Library/Developer/CoreSimulator/Devices/A3576DC2-355E-45BA-B32C-D2C0A3811BB4/data/Containers/Data/Application/7ACD12DA-2A12-4CF1-8A76-8295E40B64EF/StoreKit/receipt
INFO: ℹ️ Receipt parsed successfully
WARN: 🍎‼️ Local receipt is still missing purchase for 'com.revenuecat.monthly_4.99.1_week_intro': 
{
  "opaque_value" : "79v2XwwAAAA=",
  "sha1_hash" : "GhOmfuioo2Z4R3zX8K60l4b0qgs=",
  "bundle_id" : "com.revenuecat.StoreKitTestApp",
  "in_app_purchases" : [
    {
      "quantity" : 1,
      "product_id" : "com.revenuecat.monthly_4.99.1_week_intro",
      "purchase_date" : "2022-09-28T01:14:24Z",
      "transaction_id" : "0",
      "is_in_intro_offer_period" : true,
      "expires_date" : "2022-09-28T01:14:34Z"
    }
  ],
  "application_version" : "1",
  "creation_date" : "2022-09-28T01:14:24Z",
  "expiration_date" : "4001-01-01T00:00:00Z"
}
DEBUG: ℹ️ Retrying Receipt fetch after  5 seconds
DEBUG: ℹ️ API request completed: POST /v1/receipts 200
DEBUG: ℹ️ PostReceiptDataOperation: Finished
DEBUG: ℹ️ Serial request done: POST receipts, 0 requests left in the queue
DEBUG: ℹ️ Sending updated CustomerInfo to delegate.
DEBUG: ℹ️ Found 0 unsynced attributes for App User ID: $RCAnonymousID:ec326c1613c14b489980eace90dceabb
DEBUG: ℹ️ Force refreshing the receipt to get latest transactions from Apple.
[...]

PR #1952, opened 2022-09-28
`Integration Tests`: simplified `testIneligibleForIntroAfterPurchaseExpires` to fix flakiness

NachoSoto:
Fixes [CSDK-479].

### Changes:
- Removed extra `verifyEntitlementWentThrough`. #1880 introduced a change so that weekly subscriptions aren't verified, because they expire within a second. This make the test flaky due to that race condition.
- Removed `assertNoActiveSubscription`: I tried calling `SKTestSession.disableAutoRenewForTransaction`, but sometimes the server still thinks the subscription has auto-renewed. This was making the test flaky. Turns out that even if it's not active, the eligibility test still passes as expected.
- Removed call to `syncPurchases`. It doesn't matter what state the server is in, as long as locally `SKTestSession` has an expired subscription.

This, together with #1945, should fix the last of the issues causing flaky integration tests 🤞🏻

[CSDK-479]: https://revenuecats.atlassian.net/browse/CSDK-479?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
@@ -252,15 +252,17 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
     func testIneligibleForIntroAfterPurchaseExpires() async throws {
         let product = try await self.weeklyPackage.storeProduct
 
+        // 1. Purchase weekly offering
         let customerInfo = try await self.purchaseWeeklyOffering().customerInfo
-        let entitlement = try await self.verifyEntitlementWentThrough(customerInfo)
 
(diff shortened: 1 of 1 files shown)


vegaro (approved):
> I tried calling SKTestSession.disableAutoRenewForTransaction, but sometimes the server still thinks the subscription has auto-renewed.

Is this an issue we should fix?

NachoSoto:
I talked about it with @aboedo. The backend has no way of knowing if an `SKTestSession` subscription is going to renew or not.

PR #1955, opened 2022-09-29
Remind about updating docs and parity spreadsheet on minor releases

vegaro:
Adds two approval jobs and a Slack notification posted to the `#sdk-prs` channel. After this is merged, it can be moved to the orb so it can be shared across all repos

Screen Shot 2022-09-29 at 1 40 59 PM

[CSDK-420]

[CSDK-420]: https://revenuecats.atlassian.net/browse/CSDK-420?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -562,6 +562,57 @@ jobs:
           name: Create automatic PR
           command: bundle exec fastlane automatic_bump github_rate_limit:10
 
+  notify-on-non-patch-release-branches:
+    docker:
+      - image: 'cimg/base:stable'
+    steps:
+      - slack/notify:
+          custom: |
+              {
+                "text": "Public facing changes detected",
+                "blocks": [
+                  {
+                    "type": "header",
+                    "text": {
+                      "type": "plain_text",
+                      "text": "REMINDER :raised_hand:",
+                      "emoji": true
+                    }
+                  },
+                  {
+                    "type": "section",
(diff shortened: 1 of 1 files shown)


tonidero (commented):
Hmm so these approvals won't actually perform any actions right? It's just a reminder of actions we need to take on minors right now. I'm wondering if we are over-engineering here. Maybe just slacking a reminder would be enough? While this adds an active step to make sure someone has updated docs/parity sheet, I'm not sure if that's necessary. Happy to chat about it more though, if you think it's useful though

vegaro:
Let me cc @aboedo for that conversation @tonidero 

I think the fact that the approval step can be completely ignored and nothing will happen because the release has already been completed makes it a little bit useless and the important part here is the slack notification. 

Maybe we could move the approval earlier in the process, but I don't want to add a lot of friction to the release process.

aboedo:
Happy to go with whatever you prefer here, we can always change it later on if it's not woring

vegaro:
ok, I think I prefer just the slack notification, so I am going to go with that

tonidero (approved):
Looks like there is an error with the circleci file. Once that's fixed :shipit:

PR #1957, opened 2022-09-29
`PostReceiptDataOperation`: added `initiationSource` parameter

NachoSoto:
For [TRIAGE-82].

[TRIAGE-82]: https://revenuecats.atlassian.net/browse/TRIAGE-82?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Networking/Operations/PostReceiptDataOperation.swift
+++ b/Sources/Networking/Operations/PostReceiptDataOperation.swift
@@ -23,6 +23,7 @@ class PostReceiptDataOperation: CacheableNetworkOperation {
         let productData: ProductRequestData?
         let presentedOfferingIdentifier: String?
         let observerMode: Bool
+        let initiationSource: ProductRequestData.InitiationSource
         let subscriberAttributesByKey: SubscriberAttribute.Dictionary?
 
     }
@@ -101,6 +102,7 @@ extension PostReceiptDataOperation.PostData: Encodable {
         case appUserID
         case isRestore
         case observerMode
+        case initiationSource
         case attributes
         case presentedOfferingIdentifier
 
@@ -113,6 +115,7 @@ extension PostReceiptDataOperation.PostData: Encodable {
         try container.encode(self.appUserID, forKey: .appUserID)
         try container.encode(self.isRestore, forKey: .isRestore)
         try container.encode(self.observerMode, forKey: .observerMode)
+        try container.encode(self.initiationSource, forKey: .initiationSource)
 
         if let productData = self.productData {
             try productData.encode(to: encoder)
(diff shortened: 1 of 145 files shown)


NachoSoto:
This is ready (just need to fix the snapshots).
Integration tests will fail until backend is deployed (cc @antoniobg no rush though)

Sources/Purchasing/ProductRequestData.swift
```diff
@@ -20,6 +20,20 @@ import StoreKit
 /// - SeeAlso: `Backend/post(receiptData:appUserID:isRestore:productData:...`
 struct ProductRequestData {
 
+    /// Determins what triggered a receipt to be posted
+    enum InitiationSource: CaseIterable {
```
tonidero, line 24: Hmm what do you think about `ReceiptPostSource`? Feels like a better name to me but looks like renaming would require a backend change and I don't think it's a big deal.

Sources/Purchasing/ProductRequestData.swift
```diff
@@ -20,6 +20,20 @@ import StoreKit
 /// - SeeAlso: `Backend/post(receiptData:appUserID:isRestore:productData:...`
 struct ProductRequestData {
 
+    /// Determins what triggered a receipt to be posted
```
tonidero, line 23:
```suggestion
    /// Determines what triggered a receipt to be posted
```

tonidero (approved):
LGTM, just a naming suggestion but NABD

Sources/Networking/Operations/PostReceiptDataOperation.swift
```diff
@@ -130,3 +133,27 @@ extension PostReceiptDataOperation.PostData: Encodable {
     }
 
 }
+
+// MARK: - InitiationSource
+
+extension ProductRequestData.InitiationSource: Encodable, RawRepresentable {
```
tonidero, line 139: Hmm I do wonder if this would be better placed in the `ProductRequestData.swift` file... If it's only needed from here, this would be ok though

Sources/Purchasing/ProductRequestData.swift
NachoSoto: Thanks

Sources/Purchasing/ProductRequestData.swift
NachoSoto: We don't have to match the backend, but I can see either way.
If we don't, we can have something more "Swifty", but I think the advantage of having this match the backend is familiarity, so this term is the same across different repos.
What do you think?

NachoSoto: Maybe another thing to realize is that this isn't just `InitiationSource`, it's namespaced as `ProductRequestData.InitiationSource`.

Sources/Networking/Operations/PostReceiptDataOperation.swift
NachoSoto: My goal with this is to keep the type definition simpler. This is more of an implementation detail of how the source is posted to the backend.

PR #1967, opened 2022-10-07
Moved `SKPaymentQueue.presentCodeRedemptionSheet` to `StoreKitWorkarounds`

NachoSoto:
NFC.

I'll be using this method from 2 different places, and it belongs here with the other workarounds.

### DIFF
--- a/Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
@@ -119,3 +119,25 @@ extension ReceiptFetcher {
     }
 
 }
+
+extension SKPaymentQueue {
+
+    @available(iOS 14.0, *)
+    @available(macOS, unavailable)
+    @available(tvOS, unavailable)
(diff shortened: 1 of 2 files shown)


Sources/Purchasing/StoreKitAbstractions/StoreKitWorkarounds.swift
```diff
@@ -119,3 +119,25 @@ extension ReceiptFetcher {
...
 }
+
+extension SKPaymentQueue {
+
+    @available(iOS 14.0, *)
+    @available(macOS, unavailable)
+    @available(tvOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(macCatalyst, unavailable)
+    func presentCodeRedemptionSheetIfAvailable() {
```
tonidero, line 130: Should we add a couple tests for this?

NachoSoto: We would have to test `MockPaymentQueue` as a way to test that the method is called, but it feels silly to test a mock :/

PR #1968, opened 2022-10-07
`PaymentQueueWrapper`: improved abstraction for active `SKPaymentQueue` wrapper

### LINKED ISSUE #1964: Unable to buy the same Product twice in the same app session
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Development/CONTRIBUTING.md)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

I'm unable to purchase the same consumable twice in the same app session. Due to our app having various items for purchase, we've set up 3 Products for purchase and the items for sale in our app are linked to one of the 3 Products, which sets the item's price. The issue we're having is that the SDK doesn't attempt to make a purchase if the same Product was bought in the current app session. We're calling .purchase(Product) and on the second purchase, we automatically get a successful callback and the transactions made for the previous Product is returned, instead of making a new purchase. Closing and reopening the app apparently clears the cache and the user is able to purchase the Product then, but again only the first one.

1. Environment
   1. Platform: iOS
   2. SDK version: 4.13.1
   3. StoreKit 2 (_enabled with `useStoreKit2IfEnabled`_) (Y/N): Y
   4. OS version: 16.0.2
   5. Xcode version: 14.0.1
   6. How widespread is the issue. Percentage of devices affected. 100%

**Additional context**

I'm using the sandbox environment

NachoSoto:
This is a follow up to #1901 and #1882. Depends on #1967.
Essentially before this change, `Purchases` had both `StoreKit1Wrapper?` and a `PaymentQueueWrapper`. The later wasn't actually needed when the first was not-nil, so it was essentially storing `(a?, b?)`. What we really wanted to express is that it could have one OR the other.

This introduces `Either` to express this requirement in the type system, as well as `PaymentQueueWrapperType` that will be used to fix #1964.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -227,8 +227,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
     private let purchasesOrchestrator: PurchasesOrchestrator
     private let receiptFetcher: ReceiptFetcher
     private let requestFetcher: StoreKitRequestFetcher
-    private let storeKit1Wrapper: StoreKit1Wrapper?
-    private let paymentQueueWrapper: PaymentQueueWrapper
+    private let paymentQueueWrapper: EitherPaymentQueueWrapper
     private let systemInfo: SystemInfo
     private var customerInfoObservationDisposable: (() -> Void)?
 
@@ -267,10 +266,10 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
                               eTagManager: eTagManager,
                               operationDispatcher: operationDispatcher,
                               attributionFetcher: attributionFetcher)
-        let storeKit1Wrapper: StoreKit1Wrapper? = systemInfo.storeKit2Setting.shouldOnlyUseStoreKit2
-        ? nil
-        : StoreKit1Wrapper()
-        let paymentQueueWrapper = storeKit1Wrapper?.createPaymentQueueWrapper() ?? .init()
+
+        let paymentQueueWrapper: EitherPaymentQueueWrapper = systemInfo.storeKit2Setting.shouldOnlyUseStoreKit2
+            ? .right(.init())
+            : .left(.init())
 
         let offeringsFactory = OfferingsFactory()
         let userDefaults = userDefaults ?? UserDefaults.standard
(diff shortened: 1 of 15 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -455,16 +451,14 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
...
         } else {
             Logger.warn(Strings.configure.autoSyncPurchasesDisabled)
         }
 
         /// If SK1 is not enabled, `PaymentQueueWrapper` needs to handle transactions
         /// for promotional offers to work.
-        if !self.isStoreKit1Configured {
-            self.paymentQueueWrapper.delegate = purchasesOrchestrator
-        }
+        self.paymentQueueWrapper.right?.delegate = purchasesOrchestrator
```
tonidero, line 461: Hmm I wonder if we could add another extension to `EitherPaymentQueueWrapper` to access the `PaymentQueueWrapper`. Something like `.paymentQueueWrapper` instead of `.right` to make clearer what it is. So this would become: `self.paymentQueueWrapper.paymentQueueWrapper?.delegate = purchasesOrchestrator`

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -479,8 +473,8 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
 
     deinit {
         self.notificationCenter.removeObserver(self)
-        self.storeKit1Wrapper?.delegate = nil
-        self.paymentQueueWrapper.delegate = nil
+        self.paymentQueueWrapper.left?.delegate = nil
+        self.paymentQueueWrapper.right?.delegate = nil
```
tonidero, line 477: These could potentially use the `storeKit1Wrapper` (and `paymentQueueWrapper` if we add it) extension functions instead of `left` and `right`.

tonidero (approved):
Looks good, just a comment about a possible change but NABD.

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: Yeah that's a good suggestion, let me do that.

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: Changed

PR #1969, opened 2022-10-07
`Async.call` method to convert completion-block call to `async`

NachoSoto:
This simplifies calling completion-block APIs as `async`. The opposite of what #1943 introduced.

### DIFF
--- a/Sources/FoundationExtensions/AsyncExtensions.swift
+++ b/Sources/FoundationExtensions/AsyncExtensions.swift
@@ -56,4 +56,17 @@ internal enum Async {
         }
     }
 
+    static func call<Value, Error: Swift.Error>(
+        method: (@escaping @Sendable (Result<Value, Error>) -> Void) -> Void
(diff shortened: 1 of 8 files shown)


Sources/FoundationExtensions/AsyncExtensions.swift
```diff
@@ -56,4 +56,17 @@ internal enum Async {
...
     }
 
+    static func call<Value, Error: Swift.Error>(
+        method: (@escaping @Sendable (Result<Value, Error>) -> Void) -> Void
+    ) async throws -> Value {
+        return try await withCheckedThrowingContinuation { continuation in
+            @Sendable
+            func complete(_ result: Result<Value, Error>) {
+                continuation.resume(with: result)
+            }
```
NachoSoto, line 66: Without this intermediate `func` it doesn't compile before Xcode 14.1.

tonidero (approved):

PR #1975, opened 2022-10-10
`OfferingsManager`: added missing test for ignoring missing products

NachoSoto:
We had no test coverage for [`OfferingsManager`'s ability to ignore products that aren't found](https://github.com/RevenueCat/purchases-ios/blob/4.13.1/Sources/Purchasing/OfferingsManager.swift#L134-L140).

Added this before expanding this behavior in #1976.

### DIFF
--- a/Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
+++ b/Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
@@ -64,6 +64,7 @@ extension OfferingsManagerTests {
         expect(result).to(beSuccess())
 
         let unwrappedOfferings = try XCTUnwrap(result?.value)
+        expect(unwrappedOfferings.all).to(beEmpty())
         expect(unwrappedOfferings["base"]).to(beNil())
     }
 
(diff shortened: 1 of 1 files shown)


Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
```diff
@@ -64,6 +64,7 @@ extension OfferingsManagerTests {
         expect(result).to(beSuccess())
 
         let unwrappedOfferings = try XCTUnwrap(result?.value)
+        expect(unwrappedOfferings.all).to(beEmpty())
```
tonidero, line 67: If we add this, would the next check still be necessary?

tonidero (approved):

Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
NachoSoto: Nope, good point. I think it's still good to leave it to make it explicit that we're also making sure the expected entitlement isn't there.

PR #1976, opened 2022-10-10
`OfferingsManager`: added ability to fail if any product is not found

NachoSoto:
The default behavior is unchanged: if any products aren't found, those are logged, but an `Offerings` value is returned with the found products.

This introduces a new `OfferingsManager.FetchPolicy` that allows receiving an error if any product is not found. For the time being, this is `internal` only, and will be used for [CSDK-451], since we don't want to ignore those errors when using this new SDK tester.

[CSDK-451]: https://revenuecats.atlassian.net/browse/CSDK-451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Logging/Strings/OfferingStrings.swift
+++ b/Sources/Logging/Strings/OfferingStrings.swift
@@ -34,7 +34,7 @@ enum OfferingStrings {
     case fetching_products_finished
     case fetching_products(identifiers: Set<String>)
     case completion_handlers_waiting_on_products(handlersCount: Int)
-    case configuration_error_skproducts_not_found
+    case configuration_error_products_not_found
     case configuration_error_no_products_for_offering
     case offering_empty(offeringIdentifier: String)
     case product_details_empty_title(productIdentifier: String)
@@ -48,7 +48,7 @@ extension OfferingStrings: CustomStringConvertible {
     var description: String {
         switch self {
         case .cannot_find_product_configuration_error(let identifiers):
-            return "Could not find SKProduct for \(identifiers) " +
+            return "Could not find products with identifiers: \(identifiers)." +
                 "\nThere is a problem with your configuration in App Store Connect. " +
                 "\nMore info here: https://errors.rev.cat/configuring-products"
 
@@ -107,7 +107,7 @@ extension OfferingStrings: CustomStringConvertible {
         case .completion_handlers_waiting_on_products(let handlersCount):
             return "\(handlersCount) completion handlers waiting on products"
 
-        case .configuration_error_skproducts_not_found:
+        case .configuration_error_products_not_found:
             return "There's a problem with your configuration. None of the products registered in the RevenueCat " +
             "dashboard could be fetched from App Store Connect (or the StoreKit Configuration file " +
             "if one is being used). \nMore information: https://rev.cat/why-are-offerings-empty"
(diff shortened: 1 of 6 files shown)


Sources/Logging/Strings/OfferingStrings.swift
```diff
@@ -48,7 +48,7 @@ extension OfferingStrings: CustomStringConvertible {
     var description: String {
         switch self {
         case .cannot_find_product_configuration_error(let identifiers):
-            return "Could not find SKProduct for \(identifiers) " +
```
NachoSoto, line 51: ⚠️ This is improved, no longer referencing `SKProduct` 😅

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -134,9 +144,19 @@ private extension OfferingsManager {
             let missingProductIDs = self.getMissingProductIDs(productIDsFromStore: Set(productsByID.keys),
                                                               productIDsFromBackend: productIdentifiers)
             if !missingProductIDs.isEmpty {
-                Logger.appleWarning(
-                    Strings.offering.cannot_find_product_configuration_error(identifiers: missingProductIDs)
-                )
+                switch fetchPolicy {
+                case .ignoreNotFoundProducts:
+                    Logger.appleWarning(
```
tonidero, line 149: Do we need to `return` after this?

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -171,6 +191,20 @@ private extension OfferingsManager {
 
 }
 
+extension OfferingsManager {
+
+    /// Determines the behavior when products in an `Offering` are not found
+    internal enum FetchPolicy {
+
+        case ignoreNotFoundProducts
+        case failIfProductsAreMissing
```
tonidero, line 200: Hmm do we have plans to add other policies here? Just wondering whether it would be better to just make this a boolean instead of an enum.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -576,15 +576,28 @@ public extension Purchases {
     func logOut() async throws -> CustomerInfo {
         return try await logOutAsync()
     }
+
     @objc func getOfferings(completion: @escaping (Offerings?, PublicError?) -> Void) {
-        self.offeringsManager.offerings(appUserID: appUserID) { result in
+        self.getOfferings(fetchPolicy: .default, completion: completion)
+    }
+
+    internal func getOfferings(
```
tonidero, line 584: Looks like we are not testing these new methods directly. This is called by the existing `getOfferings` method, so it's tested with a `default` fetch policy but since it's a public method, even if it's internal, I think we should add a quick test for this just in case.

tonidero (approved):
Looks good! Just a couple of comments

Sources/Purchasing/OfferingsManager.swift
NachoSoto: No, not in this case. This is the existing behavior, which was to simply log and keep going with the products that are found.

Sources/Purchasing/OfferingsManager.swift
NachoSoto: Even if we don't, this allows us to document the behavior in the name. As for performance, this has essentially the same representation as a boolean.
The advantage over a `Bool` is that we don't have to wonder in every place what `true` or `false` means, and that's much more error prone.

PR #1971, opened 2022-10-07
Created `InternalAPI` for "health" request

NachoSoto:
For [CSDK-451].
Depends on #1969 and #1970.

## TODO:
- [x] Add endpoint to API: https://github.com/RevenueCat/khepri/pull/4615
- [x] Updated snapshots

### Other changes:
- Allow `HTTPRequest.Path`s to be non-authenticated

[CSDK-451]: https://revenuecats.atlassian.net/browse/CSDK-451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1977, opened 2022-10-10
Introduced `PurchasesDiagnostics` to help diagnose SDK configuration errors

NachoSoto:
Finishes [CSDK-451].
This new small public API allows users to quickly figure out if everything in the SDK is correctly configured in a simple way:
```swift
let diagnostics = PurchasesDiagnostics.default
do {
    try await diagnostics.testSDKHealth()
} catch {
    print(error)
}
```

The specific underlying errors will provide information about what failed.

We can continue growing this to check for more specific things, but for now it does 4 things:
- Verify API connectivity: networking issues, firewalling, etc.
- Verify API key is correct
- Verify `Offerings` are configured correctly
- Verify that all products in `Offerings` are configured correctly and found in `StoreKit`

This new API is covered by:
- API testers
- Unit tests
- Integration tests (both on `SK1` and `SK2`)

_I've taken the "shortcut" of making this `async` only (while still compatible with Objective-C), which means it's not compatible with iOS 12.x. But that made the implementation a lot simpler, which I think is a useful tradeoff._

---

This is a new API so I made sure it's documented with an example:

Screenshot 2022-10-10 at 14 38 28

### Depends on:
- #1970
- #1971
- #1973
- #1974
- #1975
- #1976

[CSDK-451]: https://revenuecats.atlassian.net/browse/CSDK-451?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sources/Support/SDKTester.swift
```diff
@@ -0,0 +1,170 @@
...
+}
+
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+extension SDKTester {
+
+    /// An error that represents a failing step in ``SDKTester``
+    public enum Error: Swift.Error {
+
+        /// Connection to the API failed
+        case failedConectingToAPI(Swift.Error)
```
vegaro, line 53: There's a small typo here

```suggestion
        case failedConnectingToAPI(Swift.Error)
```

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -941,3 +941,18 @@ public protocol PurchasesSwiftType: AnyObject {
...
+/// Interface for ``Purchases``'s internal-only methods.
+internal protocol InternalPurchasesType: AnyObject {
+
+    /// Performs an unauthenticated request to the API to verify connectivity.
+    /// - Throws: `PublicError` if request failed.
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    func healthRequest() async throws
+
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    func offerings(fetchPolicy: OfferingsManager.FetchPolicy) async throws -> Offerings
```
vegaro, line 956: Why is `offerings` considered internal?

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1054,14 +1054,28 @@ extension Purchases: @unchecked Sendable {}
...
+/// Necessary because `ErrorUtils` inside of `Purchases` finds the obsoleted type.
+private typealias NewErrorUtils = ErrorUtils
+
 internal extension Purchases {
 
     var isStoreKit1Configured: Bool {
         return self.paymentQueueWrapper.sk1Wrapper != nil
     }
 
-    #if DEBUG
```
vegaro, line 1063: Just checking about this change

Tests/UnitTests/Misc/SDKTesterTests.swift
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    func testInvalidAPIKey() {
+        let error = SDKTester.Error.invalidAPIKey
+
+        expect(error.errorUserInfo[NSUnderlyingErrorKey] as? NSNull).toNot(beNil())
+        expect(error.localizedDescription) == "API key is not valid"
+    }
+
+    func testFailedConnectingToAPI() {
```
vegaro, line 117: I think this test should be named `testFetchingOfferingsError` , and we would be missing one that returns `failedConectingToAPI`

Tests/UnitTests/Misc/SDKTesterTests.swift
```diff
@@ -0,0 +1,133 @@
...
+        let underlyingError = OfferingsManager.Error.missingProducts(identifiers: ["a"]).asPublicError
+        let error = SDKTester.Error.failedFetchingOfferings(underlyingError)
+
+        expect(error.errorUserInfo[NSUnderlyingErrorKey] as? NSError).to(matchError(underlyingError))
+        expect(error.localizedDescription) == "Failed fetching offerings: \(underlyingError.localizedDescription)"
+    }
+
+    func testFetchingOfferingsError() {
```
vegaro, line 125: And I think this test is also wrongly named

vegaro (changes requested):

Sources/Support/SDKTester.swift
NachoSoto: Thanks.

Sources/Purchasing/Purchases/PurchasesType.swift
NachoSoto: This is just an overload that takes `OfferingsManager.FetchPolicy`, which is internal. It exposes the feature added in #1976. The method without parameters is still part of the `public PurchasesType protocol`.

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: Oh good catch, I think this was an old change I didn't need anymore, I can add it again.

Tests/UnitTests/Misc/SDKTesterTests.swift
NachoSoto: Oops thanks!

Tests/UnitTests/Misc/SDKTesterTests.swift
NachoSoto: Good catch, I messed up all these tests.

vegaro (approved):

aboedo:
We should be very, very explicit with our docs for this one. Perhaps @codykerns and / or the Support team can help us figure out a nice copy? 
Similar thinking regarding the API: maybe we can call it something like `testRevenueCatIntegration` or something? Like, if it just says "test" it might IMO feel like something that's internal to the SDK, just meant for us to test it

NachoSoto:
Similar thinking regarding the API: maybe we can call it something like `testRevenueCatIntegration` or something? 

Oh yeah much better 👍🏻 
I started with `test` as I was exploring and prototyping this and never went back to change the name 🤦🏻‍♂️

@codykerns any thoughts on the rest of the documentation in here?

Sources/Support/SDKTester.swift
```diff
@@ -0,0 +1,170 @@
...
+/// let tester = SDKTester.default
+/// do {
+///     try await tester.testRevenueCatIntegration()
+/// } catch {
+///     print("SDKTester failed: \(error.localizedDescription)")
+/// }
+/// ```
+@objc(RCSDKTester)
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+public final class SDKTester: NSObject {
```
codykerns, line 30: I kind of think that naming this class more like a support or diagnostics tool, like `RevenueCatDiagnostics` or `PurchasesDiagnostics` as it more indicates symptoms of a potentially broken implementation. I think it would make it less ambiguous too, as the `test` terminology might be confusing/ pointing toward it being an internal thing rather than publicly accessible

Sources/Support/SDKTester.swift
```diff
@@ -0,0 +1,170 @@
...
+
+// MARK: - Implementation
+
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+extension SDKTester {
+
+    /// Perform tests to ensure SDK is configured correctly.
+    /// - `Throws`: ``SDKTester/Error`` if any step fails
+    @objc(testRevenueCatIntegrationWithCompletion:)
+    public func testRevenueCatIntegration() async throws {
```
codykerns, line 76:
```suggestion
    public func testBasicIntegrationHealth() async throws {
```

If we rename the class, I think it's implied that it's for RevenueCat. Let's call it something like this to give room for other potential 'tests' in the future

Sources/Support/SDKTester.swift
NachoSoto: I like `PurchasesDiagnostics` 👍🏻

Sources/Support/SDKTester.swift
NachoSoto: I like `PurchasesDiagnostics.testSDKHealth` 👍🏻

NachoSoto:
Updated to `PurchasesDiagnostics.testSDKHealth` which is much better, thanks @codykerns!

NachoSoto:
Idea from @aboedo: we can expand the [`error` descriptions](https://github.com/RevenueCat/purchases-ios/blob/62e324be4d2967634f2355377314d638e1dd388a/Sources/Support/PurchasesDiagnostics.swift#L152) with more details to help developers fix each one.
For example:
> API key is not valid. You can obtain the API key for your app at https://whatever.com/. If you’re getting this message while using a key from https://whatever.com/, make sure that you’re using the key specific to this platform (i.e.: you can’t use the Android key on Apple platforms). A correct API key for this platform will look something like appl_ligarsgiajsgeasg or oiajwgeoieagoesgea if you’re using a legacy API key.

codykerns:
@NachoSoto do you want me to write up the potential errors messages, etc? I think that's a great idea

NachoSoto:
We can leave that for a future PR if everyone is okay with merging this for now?

PR #1982, opened 2022-10-14
Fixed `ErrorUtils.purchasesError(withUntypedError:)` on iOS 12

NachoSoto:
Fixes [CSDK-491].
#1973 introduced an iOS-12 only regression:

image

For some reason the Swift runtime on iOS 12 fails to get the value from the `Error` existential, and gets a null pointer instead.

This workaround works, and it's equivalent on later iOS versions, just a bit redundant.

[CSDK-491]: https://revenuecats.atlassian.net/browse/CSDK-491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Error Handling/ErrorUtils.swift
+++ b/Sources/Error Handling/ErrorUtils.swift
@@ -290,16 +290,36 @@ enum ErrorUtils {
         withUntypedError error: Error,
         fileName: String = #fileID, functionName: String = #function, line: UInt = #line
     ) -> PurchasesError {
-        switch error {
-        case let purchasesError as PurchasesError:
-            return purchasesError
-        case let convertible as PurchasesErrorConvertible:
-            return convertible.asPurchasesError
-        default:
-            return ErrorUtils.unknownError(
-                error: error,
-                fileName: fileName, functionName: functionName, line: line
-            )
+        if #available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *) {
+            switch error {
+            case let purchasesError as PurchasesError:
+                return purchasesError
+            // This line crashes on iOS 12.x only (see https://github.com/RevenueCat/purchases-ios/pull/1982).
+            case let convertible as PurchasesErrorConvertible:
+                return convertible.asPurchasesError
+            default:
+                return ErrorUtils.unknownError(
+                    error: error,
+                    fileName: fileName, functionName: functionName, line: line
+                )
+            }
+        } else {
+            switch error {
+            case let purchasesError as PurchasesError:
+                return purchasesError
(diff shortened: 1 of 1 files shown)


Sources/Error Handling/ErrorUtils.swift
```diff
@@ -293,6 +293,14 @@ enum ErrorUtils {
         switch error {
         case let purchasesError as PurchasesError:
             return purchasesError
+
+        // `as PurchasesErrorConvertible` below covers the following types
+        // but that crashes on iOS 12.x, so these explicit casts are a workaround.
```
aboedo, line 298: we don't have a radar or something we can link to here right?

aboedo (approved):

Sources/Error Handling/ErrorUtils.swift
aboedo: this'd break again in the future if we add another type of error but don't add it here before `PurchasesErrorConvertible`, right?

aboedo: are there any cases that _aren't_ covered by `BackendError` + `NetworkError`? if not maybe we can skip the `PurchasesErrorConvertible` line, so that if we add a new error type in the future it gets handled as default without crashing on iOS 12

NachoSoto: > we don't have a radar or something we can link to here right?

Whatever it is, it's already fixed in future iOS versions, so they're not gonna bother. I also tried updating to the latest iOS 12.x to see if the fix is there, but 12.4 is the latest simulator we have available.

NachoSoto: > are there any cases that aren't covered by BackendError + NetworkError?

In practice right now, no. But there _are_ other types that conform to `PurchasesErrorConvertible`, just wouldn't be used in the code paths this is called now. But we can't guarantee that for the future unfortunately.

> so that if we add a new error type in the future it gets handled as default without crashing on iOS 12

That would break the behavior in later iOS versions though. It's unfortunate we even need this in the first place, but we do because `throws` is not typed, and we're using it more and more with `async throws` instead of typed `Result`s in completion-block APIs. Using this is what ensures that we don't end up throwing a `StoreKitError` instead of our own `ErrorCode`.

If we miss a type here, we would send "unknown error" instead, even though it's not really unknown.

But maybe that _is_ the best approach for iOS 12. Let me change this to a runtime check so at least we avoid the crash! Stand by.

NachoSoto: Updated.

aboedo (approved):

PR #1990, opened 2022-10-24
CI: added job to test compilation with `Xcode 13.2.1` / `Swift 5.5`

NachoSoto:
To cover [SDKONCALL-156].

Screenshot 2022-10-24 at 09 17 12

[SDKONCALL-156]: https://revenuecats.atlassian.net/browse/SDKONCALL-156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1992, opened 2022-10-24
CI: use `Xcode 14.1`

NachoSoto:
It contains the updated `macOS` SDK so we should be able to use it for most jobs.
#1990 will ensure that we maintain backwards compatibility with the oldest Xcode version we support.

I disabled `watchOS` from `check_pods` until https://github.com/CocoaPods/CocoaPods/issues/11558 is fixed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -252,7 +252,7 @@ jobs:
           command: bundle exec fastlane test_ios
           no_output_timeout: 30m
           environment:
-            SCAN_DEVICE: iPhone 14 (16.0)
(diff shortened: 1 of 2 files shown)


NachoSoto:
`build-tv-watch-and-macos` fails for some reason, need to look into why.

NachoSoto:
This is ready. Once https://github.com/CocoaPods/CocoaPods/issues/11558 is fixed we can re-add `watchOS` to `pod lib lint`

NachoSoto:
This should be ready.

tonidero (approved):

PR #1997, opened 2022-10-25
Documentation: fixed missing docs from inherited symbols

NachoSoto:
Fixes [CSDK-508].

#1912 moved all the docs to the `protocol`. The generated docs locally were correct, so I didn't notice that [the online docs](https://revenuecat.github.io/purchases-ios-docs/4.13.2/documentation/revenuecat/purchases/offerings()) were wrong.

I checked and Xcode was generating them with an extra parameter:
Screenshot 2022-10-25 at 12 04 38

This is also [documented](https://www.swift.org/documentation/docc/documenting-a-swift-framework-or-package), and it works like magic.

`--enable-inherited-docs` doesn't seem to be needed to fix this, but it doesn't hurt to be explicit.

[CSDK-508]: https://revenuecats.atlassian.net/browse/CSDK-508?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -722,8 +722,7 @@ workflows:
           xcode_version: '13.4.1'
           <<: *release-tags
       - docs-deploy:
-          # Not using Xcode 14 yet because it does not contain the new macOS SDK until Ventura is released.
-          xcode_version: '13.4.1'
+          xcode_version: '14.1.0'
           <<: *release-tags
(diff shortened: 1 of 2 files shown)


NachoSoto:
Yup I'll do that.

PR #1999, opened 2022-10-25
`PurchaseTesterSwiftUI`: allow configuring API key at runtime

NachoSoto:
Fixes [CSDK-456].

### Configuration screen:
Simulator Screen Shot - iPhone 13 Pro Max - 2022-10-25 at 15 02 26

### Displaying errors:
_Note that if you enter an invalid API key and there's a credentials issue, you can tap on "reconfigure" to start over:

Simulator Screen Shot - iPhone 13 Pro - 2022-10-26 at 13 35 42

[CSDK-456]: https://revenuecats.atlassian.net/browse/CSDK-456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/ConfiguredPurchases.swift
+++ b/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/ConfiguredPurchases.swift
@@ -1,0 +1,52 @@
+//
+//  ConfiguredPurchases.swift
+//  PurchaseTester
+//
+//  Created by Nacho Soto on 10/25/22.
+//
+
+import RevenueCat
+
+final class ConfiguredPurchases {
+
+    let purchases: Purchases
+    private let delegate: Delegate
+
+    init(purchases: Purchases) {
+        self.purchases = purchases
+        self.delegate = Delegate()
+
+        self.purchases.delegate = self.delegate
+    }
+
+    convenience init(
+        apiKey: String,
+        proxyURL: String?,
+        useStoreKit2: Bool
+    ) {
+        Purchases.logLevel = .debug
+
+        if let proxyURL {
+            Purchases.proxyURL = URL(string: proxyURL)!
+        }
+
+        let purchases = Purchases.configure(
+            with: .builder(withAPIKey: apiKey)
+                .with(usesStoreKit2IfAvailable: useStoreKit2)
+                .build()
+        )
+
+        self.init(purchases: purchases)
+    }
+
+}
+
+private final class Delegate: NSObject, PurchasesDelegate {
+
+    func purchases(_ purchases: Purchases, readyForPromotedProduct product: StoreProduct, purchase makeDeferredPurchase: @escaping StartPurchaseBlock) {
+        makeDeferredPurchase { (transaction, customerInfo, error, success) in
+            print("Yay")
+        }
+    }
+
+}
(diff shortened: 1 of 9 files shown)


NachoSoto:
Actually I'm gonna save it in user defaults so it remembers the last used one.

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/ConfiguredPurchases.swift
```diff
@@ -0,0 +1,52 @@
...
+        self.init(purchases: purchases)
+    }
+
+}
+
+private final class Delegate: NSObject, PurchasesDelegate {
+
+    func purchases(_ purchases: Purchases, readyForPromotedProduct product: StoreProduct, purchase makeDeferredPurchase: @escaping StartPurchaseBlock) {
+        makeDeferredPurchase { (transaction, customerInfo, error, success) in
+            print("Yay")
```
tonidero, line 48: Should we print something more meaningful? I guess this will not be called usually but something more meaningful would be better :P

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/PurchaseTesterApp.swift
```diff
@@ -12,69 +12,41 @@ import RevenueCat
...
-                    for await customerInfo in Purchases.shared.customerInfoStream {
-                        self.revenueCatCustomerData.customerInfo = customerInfo
-                        self.revenueCatCustomerData.appUserID = Purchases.shared.appUserID
+            NavigationView {
+                if let configuration {
+                    ContentView(configuration: configuration)
+                        .toolbar {
+                            ToolbarItem(placement: .navigationBarLeading) {
+                                Button {
+                                    self.configuration = nil
```
tonidero, line 26: Does the iOS SDK keep some sort of shared instance of `Purchases`? Just wondering if this is enough to reset it.

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/ConfigurationView.swift
```diff
@@ -0,0 +1,111 @@
...
+                    self.saveData()
+                    self.onContinue(self.data)
+                } label: {
+                    Text("Continue")
+                }
+                .disabled(!self.contentIsValid)
+            }
+        }
+        .onChange(of: self.data) { _ in
+            self.saveData()
```
tonidero, line 69: A UX question, should we only save data when pressing the "Continue" button? I think it shouldn't matter much, but seems more intuitive for me... For example, if a user wants to reconfigure the sdk, starts changing the data but then restarts the app, I would expect to still have the old information... NABD though.

tonidero (approved):
Just some questions but looking good!

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/ConfiguredPurchases.swift
NachoSoto: I agree, I just copied the existing implementation. There's tons we could improve but this PR focuses on just this feature.

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/PurchaseTesterApp.swift
NachoSoto: Technically not, but it will be re-created when it gets re-configured.

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/ConfigurationView.swift
NachoSoto: Yeah, I could see either way. I don't think this is that "wrong", so if you don't mind I'm going to leave it as is so I don't have to re-test this 9-day old PR.

PR #2003, opened 2022-10-26
Deploy `PurchaseTesterSwiftUI` to TestFlight

NachoSoto:
Fixes [CSDK-510].

IMG_495339BA4A2F-1

Depends on #1999.

Also set up CI as per [the docs](https://docs.fastlane.tools/best-practices/continuous-integration/#storing-a-manually-verified-session-using-spaceauth):
Added these 2 environment variables:
- `SPACESHIP_ONLY_ALLOW_INTERACTIVE_2FA`
- `FASTLANE_SESSION`

[CSDK-510]: https://revenuecats.atlassian.net/browse/CSDK-510?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


NachoSoto:
This is ready 🎉

NachoSoto:
I'm going to submit a catalyst version too

fastlane/Fastfile
```diff
@@ -465,6 +465,28 @@ platform :ios do
     end
   end
 
+  desc "Build and deploy PurchaseTesterSwiftUI"
+  lane :deploy_purchase_tester do
+    match(verbose: true, readonly: true)
+    increment_build_number(
+      build_number: latest_testflight_build_number + 1,
```
tonidero, line 472: I'm using the CircleCI build number here, but then, this is probably fine as well 👍

Tests/TestingApps/PurchaseTesterSwiftUI/fastlane/Fastfile
```diff
@@ -0,0 +1 @@
+import("../../../../fastlane/Fastfile")
```
tonidero, line 1: With this I guess the `PurchaseTesterSwiftUI` project will have access to all the project fastfile's functions... There are probably many that shouldn't be called from here, but then again, I think it should be ok.

tonidero (approved):
Looks great!

Tests/TestingApps/PurchaseTesterSwiftUI/fastlane/Fastfile
NachoSoto: Yeah it's fine. It's the same approach we use in the installation tests.

Tests/TestingApps/PurchaseTesterSwiftUI/AppIcon.xcassets/Contents.json
```diff
@@ -0,0 +1,6 @@
+{
```
aboedo, line 1: the icon is glorious 🎨

fastlane/Fastfile
aboedo: I think the latest Xcode versions were supposed to auto-bump build numbers? 
But then again, that might be only if you use Xcode, not if you go through command line

aboedo (approved):

fastlane/Fastfile
NachoSoto: Yeah and I use that for my own app now, but it wasn't working without bumping the build number manually like this. So they probably built that feature into the UI only (like when Swift shipped 😂)

PR #2006, opened 2022-10-27
`Async.call`: added non-throwing overload

NachoSoto:
This will be used by an upcoming PR.

### DIFF
--- a/Sources/FoundationExtensions/AsyncExtensions.swift
+++ b/Sources/FoundationExtensions/AsyncExtensions.swift
@@ -18,6 +18,13 @@ internal enum Async {
 
     /// Invokes an `async throws` method and calls `completion` with the result.
     /// Ensures that the returned error is `PublicError`.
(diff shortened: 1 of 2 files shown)


Sources/FoundationExtensions/AsyncExtensions.swift
```diff
@@ -69,4 +70,18 @@ internal enum Async {
         }
     }
 
+    /// Invokes a completion-block based API and returns the method `async`hronoiusly.
```
aboedo, line 73: I'm kinda lost as to how this'd be used, maybe we can add an example here to the docstring?

NachoSoto: I added examples for all methods in the file.

NachoSoto:
@RevenueCat/coresdk bump.

aboedo (approved):

Issue #1893, opened 2022-09-08
Bug: CheckTrialOrIntroDiscountEligibility too slow


jesus-mg-ios:
- [X] I have updated Purchases SDK to the latest version
- [-] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Development/CONTRIBUTING.md)
- [X ] I have searched the [Community](https://community.revenuecat.com)
- [ X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ X] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**

We are using checkTrialOrIntroDiscountEligibility in iOS with swift  after get the product, and we notice that this call is too slow (even 45s - 1 min). 
```
Purchases.shared.checkTrialOrIntroDiscountEligibility(productIdentifiers: ["productIdentifier"]) { result in
....
}
```

1. Environment
   1. Platform: iOS
   2. SDK version: 4.10
   3. StoreKit 2 (_enabled with `useStoreKit2IfEnabled`_) (Y/N): N
   4. OS version: 15.5
   5. Xcode version: 13.4.1
   6. How widespread is the issue. Percentage of devices affected: Unknown/Random
**Additional context**

Environment: Real device & simulator. iOS 15.5
SDK Version: "4.10.0"
Code that is too slow
```
Purchases.shared.checkTrialOrIntroDiscountEligibility(productIdentifiers: ["productIdentifier"]) { result in
....
}
```

aboedo:
@jesus-mg-ios would you mind trying out the branch from this PR to see if you get better results? https://github.com/RevenueCat/purchases-ios/pull/1894

It just disables SK2 for this particular operation.

jesus-mg-ios:
Thanks, I'm gonna try it!

jesus-mg-ios:
Hi @aboedo, the change was amazingly faster.

We're gonna share with you the logs before and 
after the change.

[RevenueCatLogsBeforeFixBranch_4_11_0 copy.txt](https://github.com/RevenueCat/purchases-ios/files/9610523/RevenueCatLogsBeforeFixBranch_4_11_0.copy.txt)

[RevenueCatLogsAfterFixBranch_4_12_0 copy.txt](https://github.com/RevenueCat/purchases-ios/files/9610508/RevenueCatLogsAfterFixBranch_4_12_0.copy.txt)

[...]

PR #2007, opened 2022-10-27
Created `CachingTrialOrIntroPriceEligibilityChecker`

NachoSoto:
Fixes [CSDK-493].

This will help a little bit with the SK2 performance issues computing eligibility (#1893).

[CSDK-493]: https://revenuecats.atlassian.net/browse/CSDK-493?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/CachingTrialOrIntroPriceEligibilityChecker.swift
+++ b/Sources/Purchasing/CachingTrialOrIntroPriceEligibilityChecker.swift
@@ -51,10 +51,23 @@ extension CachingTrialOrIntroPriceEligibilityChecker {
             completion(cached)
         } else {
             self.checker.checkEligibility(productIdentifiers: Array(missingProducts)) { result in
-                self.cache.value += result
+                self.cache.value += result.filter { $0.value.shouldCache }
 
                 completion(cached + result)
             }
+        }
+    }
+
+}
+
+// MARK: - Private
+
+private extension IntroEligibility {
+
+    var shouldCache: Bool {
+        switch self.status {
+        case .noIntroOfferExists, .ineligible, .eligible: return true
+        case .unknown: return false
         }
     }
 
--- a/Sources/Purchasing/CachingTrialOrIntroPriceEligibilityChecker.swift
+++ b/Sources/Purchasing/CachingTrialOrIntroPriceEligibilityChecker.swift
@@ -45,13 +45,20 @@ extension CachingTrialOrIntroPriceEligibilityChecker {
         // It's a fine compromise to keep this implementaton simpler.
         let cached = self.cache.value.filter { uniqueProductIdentifiers.contains($0.key) }
 
+        if !cached.isEmpty {
+            Logger.debug(Strings.eligibility.found_cached_eligibility_for_products(Set(cached.keys)))
+        }
+
         let missingProducts = uniqueProductIdentifiers.subtracting(cached.keys)
 
         if missingProducts.isEmpty {
             completion(cached)
         } else {
             self.checker.checkEligibility(productIdentifiers: Array(missingProducts)) { result in
-                self.cache.value += result.filter { $0.value.shouldCache }
+                let productsToCache = result.filter { $0.value.shouldCache }
+
+                Logger.debug(Strings.eligibility.caching_intro_eligibility_for_products(Set(productsToCache.keys)))
+                self.cache.value += productsToCache
 
                 completion(cached + result)
             }
--- a/Sources/Purchasing/CachingTrialOrIntroPriceEligibilityChecker.swift
+++ b/Sources/Purchasing/CachingTrialOrIntroPriceEligibilityChecker.swift
@@ -1,0 +1,61 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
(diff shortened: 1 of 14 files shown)


Sources/FoundationExtensions/AsyncExtensions.swift
```diff
@@ -69,4 +70,18 @@ internal enum Async {
         }
     }
 
+    /// Invokes a completion-block based API and returns the method `async`hronoiusly.
```
tonidero, line 73:
```suggestion
    /// Invokes a completion-block based API and returns the method `async`hronously.
```

Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
```diff
@@ -135,6 +135,20 @@ extension TrialOrIntroPriceEligibilityCheckerType {
 
 }
 
+/// Default overload implementation that exposes an `async` API.
+@available(iOS 13.0, tvOS 13.0, watchOS 6.2, macOS 10.15, *)
+extension TrialOrIntroPriceEligibilityCheckerType {
+
+    func checkEligibility(productIdentifiers: [String]) async -> [String: IntroEligibility] {
```
tonidero, line 142: Hmm, this seems unused other than in tests?

tonidero (approved):
LGTM!

Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
NachoSoto: it's actually [public API](https://github.com/RevenueCat/purchases-ios/blob/4.14.0/Sources/Purchasing/Purchases/Purchases.swift#L731).

tonidero: But that's the completion version, not this one with async right?

NachoSoto: Oooh good catch! I moved this to just the test file.

NachoSoto:
Oh wait we need to invalidate this cache when we make purchases or they expire of course 🤦🏻‍♂️
Caching and off-by-one errors, the 3 hardest problems in computer science.

NachoSoto:
This is ready for review again @RevenueCat/coresdk, I'm now clearing the cache when subscriptions change, which should fix integration tests.

Sources/Purchasing/CachingTrialOrIntroPriceEligibilityChecker.swift
```diff
@@ -0,0 +1,107 @@
...
+
+/// `TrialOrIntroPriceEligibilityCheckerType` decorator that adds caching behavior on each request.
+final class CachingTrialOrIntroPriceEligibilityChecker: TrialOrIntroPriceEligibilityCheckerType {
+
+    private let checker: TrialOrIntroPriceEligibilityCheckerType
+    private let notificationCenter: NotificationCenter
+
+    private let cache: Atomic<[String: IntroEligibility]> = .init([:])
+
+    private var observation: NSObjectProtocol!
```
tonidero, line 26: Hmm I'm probably missing something but should this be `private let observation: NSObjectProtocol`?

tonidero (approved):

Sources/Purchasing/CachingTrialOrIntroPriceEligibilityChecker.swift
NachoSoto: The observation requires escaping `self` in the closure, and that's not fully initialized without setting this value. This is solved in Kotlin with `lateinit` but Swift doesn't have that.

tonidero: Ahh ok, thanks for explaining!

aboedo: this might not be obvious to a maintainer, and they might be inclined to try to spend some time getting rid of the IUO. Maybe we could add a comment explaining why this is needed?

aboedo (approved):

Sources/Purchasing/CachingTrialOrIntroPriceEligibilityChecker.swift
NachoSoto: It's a normal pattern but yeah I can add it.

NachoSoto: Done.

NachoSoto:
Updated this to use the new notification.

tonidero (approved):

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1235,6 +1238,11 @@ internal extension Purchases {
 
 private extension Purchases {
 
+    func handleCustomerInfoChanged(_ customerInfo: CustomerInfo) {
+        self.trialOrIntroPriceEligibilityChecker.clearCache()
```
NachoSoto, line 1242: @aboedo per our discussion, this ordering now ensures there are no race conditions.

NachoSoto:
Updated this to clear caches using the existing `CustomerInfoManager` observation mechanism.

aboedo (approved):

PR #2015, opened 2022-10-31
Avoid posting empty receipts by making`TransactionsManager` always use `SK1` implementation

NachoSoto:
Fixes [CSDK-513].

The SDK still uses `SK1` when it comes to fetching receipts to post to the backend.
The `SK1` implementation checks whether there are transactions inside of the receipt.

However, this could lead to issues if `SK2` has transactions but SK1 doesn't. In those cases, we would end up posting an empty receipt. This PR fixes that.

[CSDK-513]: https://revenuecats.atlassian.net/browse/CSDK-513?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sources/Purchasing/TransactionsManager.swift
```diff
@@ -15,33 +15,14 @@ import StoreKit
...
-
-    func sk1CheckCustomerHasTransactions(receiptData: Data) -> Bool {
-        receiptParser.receiptHasTransactions(receiptData: receiptData)
-    }
-
-    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
-    func sk2CheckCustomerHasTransactions() async -> Bool {
-        await StoreKit.Transaction.all.contains { _ in true }
+    func customerHasTransactions(receiptData: Data) -> Bool {
+        return self.receiptParser.receiptHasTransactions(receiptData: receiptData)
```
aboedo, line 25: let's add a note here explaining that we're always using SK1, even though SK2 might potentially be more accurate, because we need to reflect whatever's on the receipt if we're posting SK1 receipts

aboedo (approved):
One small note, ready to go otherwise

Sources/Purchasing/TransactionsManager.swift
NachoSoto: Done.

PR #2017, opened 2022-11-01
Makes hold job wait for installation tests to pass

vegaro:
Equivalent of https://github.com/RevenueCat/purchases-android/pull/644/files

Fixes two things:
- Makes sure installation tests pass before deploying anything
- Prevents installation tests from running also when the branch is tagged (after the hold job is approved), which was unnecessary and caused unnecessary waiting time


tonidero (approved):

NachoSoto (commented):
Awesome!

PR #2020, opened 2022-11-02
Fixed purchasing with `PromotionalOffer`s using `StoreKit 2`

NachoSoto:
Fixes [SDKONCALL-160], [SDKONCALL-182], #2114, and https://github.com/RevenueCat/react-native-purchases/issues/455.
Depends on #2118 and https://github.com/RevenueCat/khepri/pull/4843 / https://github.com/RevenueCat/khepri/pull/4852.

## Changes:
- Base64decode `signature`
- Fixed signature on backend for SK2: https://github.com/RevenueCat/khepri/pull/4843 / https://github.com/RevenueCat/khepri/pull/4852
- Added tests
- Added specific error if signature can't be decoded:
> [Purchases] - ERROR: 😿‼️ The information associated with this PromotionalOffer is not valid.
See https://rev.cat/ios-subscription-offers for more info. The signature generated by RevenueCat could not be decoded: signature 914

[SDKONCALL-160]: https://revenuecats.atlassian.net/browse/SDKONCALL-160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SDKONCALL-182]: https://revenuecats.atlassian.net/browse/SDKONCALL-182?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sources/Purchasing/StoreKitAbstractions/PromotionalOffer.swift
```diff
@@ -122,11 +122,20 @@ extension PromotionalOffer.SignedData {
 
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
     var sk2PurchaseOption: Product.PurchaseOption {
+        let signature: Data
+
+        if let decoded = Data(base64Encoded: self.signature) {
```
NachoSoto, line 127: I was hoping this would fix it, but the test is still failing :/

NachoSoto: I'm out of ideas, I'll probably have to file a Radar and pray? 🙏🏻

NachoSoto: ⚠️ 

Looks like one of our customers tried to do this without RC, using `StoreKit` directly, and he gets the same result, so that confirms the fact that this is an Apple bug.
He's opened a TSI with Apple, so hopefully they can address this soon.

In the mean time, I'll leave this PR open. I believe the fix here will still be required (doing the `base64` decode). But once Apple fixes it on their end we can use this PR to confirm, as well as the now enabled test.

Sources/Purchasing/StoreKitAbstractions/PromotionalOffer.swift
```diff
@@ -122,11 +122,20 @@ extension PromotionalOffer.SignedData {
...
+            signature = decoded
+        } else {
+            // TODO: log warning
```
aboedo, line 130: Yeah, definitely worth logging something loud here. 
Stuff logged from here would also show up in the diagnostics API call, right? since we get offerings and parse products?

NachoSoto: This property is only used when purchasing though.

Sources/Purchasing/StoreKitAbstractions/PromotionalOffer.swift
wendland99: Do you have the ticket # with Apple?

mamouneyya: Not sure about the original, but we reported it today as well under: FB11764024.

NachoSoto:
I just realized this _cursed_ PR is # 2020 🙈

aboedo:
> I just realized this cursed PR is # 2020 🙈
#destiny

mamouneyya:
**Update:** We noticed today that this started to work actually. Also, I can confirm that the decoding logic in this PR is important. When not decoding the signature, I was getting the `invalidOfferSignature` error code every time.

Something I'm a bit confused about is the fact that this is now working even when using a simulator and a StoreKit testing store file. I thought that in that scenario, Xcode won't make any remote request, so I cannot understand how this was fixed from the App Store server side. Glad it is though.

NachoSoto:
Interesting, our integration test is still failing. I'll dig more next week though.

NachoSoto:
> Update: We noticed today that this started to work actually. Also, I can confirm that the decoding logic in this PR is important. When not decoding the signature, I was getting the invalidOfferSignature error code every time.

Weird, I can't get this to work even with this fix. Are you doing something else differently? Is it running on iOS 16.1.2 only or something?

NachoSoto:
Thanks, I just ran it on iOS 15.x and it's still failing. I verified in our backend that the username is never empty.

mamouneyya:
> Thanks, I just ran it on iOS 15.x and it's still failing. I verified in our backend that the username is never empty.

By never empty, do you mean that you're actually using a value when generating the signature? If that's the case, then you'll have to pass an [`appAccountToken(_:)`](https://developer.apple.com/documentation/storekit/product/purchaseoption/3749440-appaccounttoken) purchase option with the exact same value.

NachoSoto:
> By never empty, do you mean that you're actually using a value when generating the signature? If that's the case, then you'll have to pass an [.appAccountToken(_:)](https://developer.apple.com/documentation/storekit/product/purchaseoption/3749440-appaccounttoken) purchase option with the exact same value.

Whoa we're definitely not doing that.

This is still not working:
```swift
options.insert(signedData.sk2PurchaseOption)
options.insert(.appAccountToken(UUID(uuidString: self.appUserID)!))
```

But I feel like we're on the right track. I missed this:

> If you're providing an `appAccountToken` in the purchase options, you must include that value when you generate the signature. Use lowercase for the UUID string representations of the app account token and the nonce in the siganture.

NachoSoto:
(Testing now calling `lowercased()`)

NachoSoto:
> `applicationUsername` or `appAccountToken`
An optional string value that you define; may be an empty string. If your app uses `applicationUsername`, provide applicationUsername. If your app uses `appAccountToken`, provide `appAccountToken`. Use lowercase for the string representation of `appAccountToken`.

Our backend isn't lowercasing the user ID for SK2, so that's likely the issue.

NachoSoto:
CONFIRMED!
Using a lowercase user ID (and passing the missing `.appAccountToken`) makes the test pass.

Thanks a lot for that pointer @mamouneyya! We'll get this fixed.

mamouneyya:
> CONFIRMED! Using a lowercase user ID (and passing the missing `.appAccountToken`) makes the test pass.
> 
> Thanks a lot for that pointer @mamouneyya! We'll get this fixed.

Wohoo! Glad you got it working! 🎉

Sources/Purchasing/StoreKitAbstractions/PromotionalOffer.swift
```diff
@@ -122,13 +129,36 @@ extension PromotionalOffer.SignedData {
...
-            signature: self.signature.asData,
-            timestamp: self.timestamp
-        )
+        get throws {
+            let signature: Data
+
+            if let decoded = Data(base64Encoded: self.signature) {
+                signature = decoded
+            } else {
+                throw Error.failedToDecodeSignature(self.signature)
```
NachoSoto, line 138: I improved error handling here (the last TODO) and added a few tests.

aboedo:
thinking more about this: 
if, for affected versions, the backend sent the signature as utf 8 instead of base64 encoded, wouldn't that fix them?

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -467,6 +467,33 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
...
+        self.customerInfoManager.stubbedCachedCustomerInfoResult = self.mockCustomerInfo
+        self.backend.stubbedPostReceiptResult = .success(self.mockCustomerInfo)
+
+        let product = try await self.fetchSk2Product()
+        let offer = PromotionalOffer.SignedData(
+            identifier: "identifier \(Int.random(in: 0..<1000))",
+            keyIdentifier: "key identifier \(Int.random(in: 0..<1000))",
+            nonce: .init(),
+            // This should be base64
+            signature: "signature \(Int.random(in: 0..<1000))",
```
aboedo, line 483: why not base64 the result of this so it looks realistic?

aboedo (approved):

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
NachoSoto: This is 
> testPurchaseSK2PackageWithInvalidPromotionalOfferSignatureThrowsError

So I'm testing that if it's _not_ base64 (so it's just a random string), it produces the correct error instead of crashing or something.

NachoSoto:
Merging 🎉

KrisJulioRSVP:
This helps my problem, im just passing empty string in `appAccountToken`. Thank you guys! Apple documentation is pain in the a$$

PR #2024, opened 2022-11-04
`ReceiptRefreshPolicy.retryUntilProductIsFound`: default to returning "invalid" receipt

NachoSoto:
See #1945.

To make this retry mechanism less risky when eventually enabled, this change makes the "worst case scenario" (where retrying didn't help, or `AppleReceipt.containsActivePurchase` has issues) equivalent to `ReceiptFetcher.onlyIfEmpty`.

_Note: this is currently not enabled in production, only in Integration Tests._

### DIFF
--- a/Tests/UnitTests/Mocks/MockReceiptParser.swift
+++ b/Tests/UnitTests/Mocks/MockReceiptParser.swift
@@ -54,7 +54,7 @@ class MockReceiptParser: ReceiptParser {
             // This is used to mock changing receipts over time.
             return try self.stubbedParseResults[self.invokedParseCount - 1].get()
         } else {
-            return try XCTUnwrap(self.stubbedParseResults.first?.get())
+            return try XCTUnwrap(self.stubbedParseResults.first).get()
         }
     }
 
--- a/Tests/UnitTests/Purchasing/ReceiptFetcherTests.swift
+++ b/Tests/UnitTests/Purchasing/ReceiptFetcherTests.swift
@@ -240,6 +240,19 @@ final class RetryingReceiptFetcherTests: BaseReceiptFetcherTests {
         ]
     }
 
+    func testStopsRetryingEvenIfParsingReceiptKeepsThrowingError() async {
+        let invalidData = self.mockReceiptWithInvalidData()
+
+        let data = await self.fetch(productIdentifier: Self.productID, retries: 1)
+        expect(data) == invalidData
+
+        expect(self.mockRequestFetcher.refreshReceiptCalledCount) == 2
+        expect(self.mockReceiptParser.invokedParseParametersList) == [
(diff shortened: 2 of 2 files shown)


Sources/Purchasing/ReceiptFetcher.swift
```diff
@@ -151,10 +151,11 @@ private extension ReceiptFetcher {
         sleepDuration: DispatchTimeInterval
     ) async -> Data {
         var retries = 0
+        var data: Data = .init()
 
         repeat {
             retries += 1
-            let data = await self.refreshReceipt()
+            data = await self.refreshReceipt()
```
tonidero, line 158: When the receipt can't be parsed on the last retry, we would return that receipt. Is that correct?

tonidero (approved):
Just a question, if that's ok, this looks good.

Sources/Purchasing/ReceiptFetcher.swift
tonidero: We could add a couple of tests for these returning an unparseable receipt and an empty receipt if that's the case.

NachoSoto: > When the receipt can't be parsed on the last retry, we would return that receipt. Is that correct?

Correct, as explained in the description:
> To make this retry mechanism less risky when eventually enabled, this change makes the "worst case scenario" [...] equivalent to `ReceiptFetcher.onlyIfEmpty`.

tonidero: > The current tests already cover that though?

Well I think we have tests for receipts without a purchase but not for unparseable receipts? (I might be missing something since I'm not that familiar with the parsing logic)

NachoSoto: Is `testStopsRetryingEvenIfDataIsInvalid` what you're describing?

tonidero:
```
    func testStopsRetryingEvenIfDataIsInvalid() async {
        self.mock(receipt: Self.receiptWithoutPurchases)
        let invalidData = Self.receiptWithoutPurchases.asData

        let data = await self.fetch(productIdentifier: Self.productID, retries: 2)
        expect(data) == Data()
        expect(data) == invalidData

        expect(self.mockReceiptParser.invokedParseParametersList) == [
            invalidData,
            invalidData,
            invalidData
        ]
        expect(self.mockRequestFetcher.refreshReceiptCalledCount) == 3
    }
```

That seems to be using a `receiptWithoutPurchases`. I was thinking more along the lines of a receipt that is missing a critical field or something like that. So basically a test where we make `self.receiptParser.parse` throw an exception.

PR #2030, opened 2022-11-07
CI: fixed `iOS 14` tests Xcode version

NachoSoto:
See https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/8749/workflows/f4994fe6-6ea3-4165-aaf0-4c5ee481260c/jobs/43015

We can't run these tests on Xcode 14 without first installing the runtime (see https://circle-macos-docs.s3.amazonaws.com/image-manifest/v8973/index.html).
Not sure if there's an API for that, and even if there was, it would be way too slow to bother instead of using an image that has the runtime already installed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -326,6 +326,11 @@ jobs:
     steps:
       - checkout
       - install-dependencies
+      - install-and-create-sim:
+          install-name: iOS 14.5 Simulator
+          sim-device-type: iPhone-8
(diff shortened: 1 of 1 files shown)


NachoSoto:
We missed this because CircleCI / GitHub were hiding the job instead of marking it as failed 🤨

tonidero (approved):

PR #2032, opened 2022-11-07
`ErrorCode.unknownBackendError`: include original error code

NachoSoto:
Currently we only include the message and not the error code. So when these unknown errors are returned, unless the message is clear, we might not known which code the backend actually sent.
Example:
> There was an unknown backend error. The purchased product was missing in the receipt. This is typically due to a bug in StoreKit.

In this case it's clear, but nowhere in the logs we can find the error code.

This adds the error code to those messages (only when the code is unknown), as well as to `userInfo`.


vegaro (approved):
This is going to be very useful

PR #2033, opened 2022-11-07
Added `BackendErrorCode.purchasedProductMissingInAppleReceipt`

NachoSoto:
See https://github.com/RevenueCat/khepri/pull/4716
This makes that error `ErrorCode.invalidReceiptError` instead of `.unknownBackendError`.

I've also added #2032 to improve error messages when these errors are unknown in the future.

### DIFF
--- a/Sources/Error Handling/BackendErrorCode.swift
+++ b/Sources/Error Handling/BackendErrorCode.swift
@@ -90,7 +90,8 @@ extension BackendErrorCode {
             return .storeProblemError
         case .cannotTransferPurchase:
             return .receiptAlreadyInUseError
-        case .invalidReceiptToken, .purchasedProductMissingInAppleReceipt:
+        case .invalidReceiptToken,
+                .purchasedProductMissingInAppleReceipt:
             return .invalidReceiptError
         case .invalidAppStoreSharedSecret,
              .invalidAuthToken,
--- a/Sources/Error Handling/BackendErrorCode.swift
+++ b/Sources/Error Handling/BackendErrorCode.swift
@@ -42,6 +42,7 @@ enum BackendErrorCode: Int, Error {
     case subscriptionNotFoundForCustomer = 7259
     case invalidSubscriberAttributes = 7263
     case invalidSubscriberAttributesBody = 7264
+    case purchasedProductMissingInAppleReceipt = 7712
 
     /**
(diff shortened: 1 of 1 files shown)


Sources/Error Handling/BackendErrorCode.swift
```diff
@@ -89,7 +90,7 @@ extension BackendErrorCode {
             return .storeProblemError
         case .cannotTransferPurchase:
             return .receiptAlreadyInUseError
-        case .invalidReceiptToken:
+        case .invalidReceiptToken, .purchasedProductMissingInAppleReceipt:
```
tonidero, line 93: Nitpick, these multi-cases seem to be set in multiple lines here. Just to be consistent:
```suggestion
        case .invalidReceiptToken, 
             .purchasedProductMissingInAppleReceipt:
```

tonidero: Also, not sure if we have tests for these... If we do, it might be good to add one, if not, I guess something to think about.

tonidero (approved):

Sources/Error Handling/BackendErrorCode.swift
NachoSoto: What type of test would you write, other than simply duplicating this logic?

tonidero: Yeah I think it would be good to have a test that checks the mapping from backend code to purchases code for each of these values. As you said, it's basically duplicating the logic, so not a huge benefit, but might catch some issues if someone changes the mapping at some point and moves one of the error codes accidentally. That said, I don't think this is needed for this PR

NachoSoto: Like you said yourself, that would just be duplicating the logic. This is purely data-driven, so the data is the test as well.

> for each of these values.

In order to do this cleanly, we would need to have a mapping for each of the errors, instead of duplicating each test. Then iterate the mapping and assert it's correct... but wait, that mapping _is_ the data we're testing, so there would be no valid in that duplication.

tonidero: Well I agree that it's redundant, but there is some security in redundancy. In any case, I agree the chances of this being changed mistakenly are minimal and if we did this, we would probably have to duplicate logic for tests in a bunch of places so I'm ok with leaving this.

PR #2035, opened 2022-11-07
`Integration Tests`: run on iOS 16

NachoSoto:
iOS 16 is here to stay, so we should make sure our tests run there too.

One could make the case for running integration tests in all versions of iOS, but I think for now running on the latest is a good default.


NachoSoto:
FYI integration tests are failing now because of https://github.com/RevenueCat/khepri/pull/4716

tonidero (approved):

PR #2036, opened 2022-11-07
`CI`: run tests on iOS 11

NachoSoto:
We support iOS 11, so we should make sure tests pass there too.


NachoSoto:
I'm testing with Xcode 13.0.0 (which technically we don't support), and we still can't install an iOS 11 simulator :(

PR #2038, opened 2022-11-08
`PostReceiptDataOperation`: log Apple error when purchase equals expiration date

NachoSoto:
A large amount of flaky integration tests are due to this bug, so detecting it early in the logs will help debug this.
Radar: `FB11767559`.

`productType` is missing in `StoreKitTest` purchases, so I changed `isSubscription` to default to checking if there is an expiration when the type is `.unknown`.
To simplify the code, I changed `ProductType?` to be non-optional, since we already have `.unknown` as a case.

### Example:
```
[00:27:56]: ▸ 2022-11-09 00:27:56.000533+0000 BackendIntegrationTestsHostApp[3942:17000] [Purchases] - DEBUG: ℹ️ PostReceiptDataOperation: Posting receipt: {
[00:27:56]: ▸   "opaque_value" : "1n+XXw8AAAA=",
[00:27:56]: ▸   "sha1_hash" : "unR7vAzPaUlB0TtZ09z5jMyqqAs=",
[00:27:56]: ▸   "bundle_id" : "com.revenuecat.StoreKitTestApp",
[00:27:56]: ▸   "in_app_purchases" : [
[00:27:56]: ▸     {
[00:27:56]: ▸       "quantity" : 1,
[00:27:56]: ▸       "product_id" : "com.revenuecat.weekly_1.99.3_day_intro",
[00:27:56]: ▸       "purchase_date" : "2022-11-09T00:27:50Z",
[00:27:56]: ▸       "transaction_id" : "0",
[00:27:56]: ▸       "is_in_intro_offer_period" : true,
[00:27:56]: ▸       "expires_date" : "2022-11-09T00:27:50Z"
[00:27:56]: ▸     },
[00:27:56]: ▸     {
[00:27:56]: ▸       "product_id" : "com.revenuecat.weekly_1.99.3_day_intro",
[00:27:56]: ▸       "quantity" : 1,
[00:27:56]: ▸       "transaction_id" : "1",
[00:27:56]: ▸       "is_in_intro_offer_period" : false,
[00:27:56]: ▸       "expires_date" : "2022-11-09T00:28:00Z",
[00:27:56]: ▸       "original_purchase_date" : "2022-11-09T00:27:50Z",
[00:27:56]: ▸       "original_transaction_id" : "0",
[00:27:56]: ▸       "purchase_date" : "2022-11-09T00:27:50Z"
[00:27:56]: ▸     }
[00:27:56]: ▸   ],
[00:27:56]: ▸   "application_version" : "1",
[...]

PR #2039, opened 2022-11-09
`Integration Tests`: changed weekly to monthly subscriptions to work around 0-second subscriptions

NachoSoto:
As exposed by #2038, weekly subscriptions were leading to false negatives because they were always expired by the time the purchase was made (using `TimeRate.monthlyRenewalEveryThirtySeconds`), instead of lasting 10 seconds as documented. This is reported as `FB11767559`.

The advantage of using weekly subscriptions was to also work around `FB11767679`, which means that we have to manually _wait_ for the subscription to expire, as `SKTestSession.expireSubscription(productIdentifier:)` doesn't actually change the receipt.
Because we had to wait, using weekly subscriptions was faster, since we only needed to wait 10 seconds.

This does mean that expirations in these tests now have to wait 30 seconds instead of 10, but now they're no longer flaky :tada:
Looks like tests right now pass in CI in less than 7 minutes, which I believe is totally acceptable, especially for the benefit of no longer having these flaky tests.

### DIFF
--- a/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
@@ -268,9 +268,7 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
     }
 
     func testNoIntroOfferIfProductHasNoIntro() async throws {
-        let product = try await XCTAsyncUnwrap(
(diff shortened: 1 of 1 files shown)


vegaro (approved):

aboedo (approved):
> Looks like tests right now pass in CI in less than 7 minutes, which I believe it totally acceptable, especially for the benefit of no longer having these flaky tests.

💯

NachoSoto:
Oops I merged this into the other branch, let me re-open the PR..

Issue #1527, opened 2022-04-25
crash: EXC_BAD_ACCESS in 4.2.1, DeviceCache.swift - Line 123


quantamrhino:
- [* ] I have updated Purchases SDK to the latest version
- [*] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/CONTRIBUTING.md)
- [*] I have searched the [Community](https://community.revenuecat.com)
- [* ] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [*] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
 Crash in DeviceCache.cache(customerInfo:appUserID:) + 123 
Crashed: Backend callbackQueue
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000b2005a2e0

see stack trace below

1. Environment
   1. Platform: iOS
   2. SDK version: 4.2.1
   3. StoreKit 2 enabled (Y/N): not sure.
   4. OS version: 15.5.0
   5. Xcode version: 13.3.1
   6. How widespread is the issue. Percentage of devices affected. unknown as of now
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
  no logs available for this crash
4. Steps to reproduce, with a description of expected vs. actual behavior
No steps as such - happened randomly
6. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

```
Crashed: Backend callbackQueue
EXC_BAD_ACCESS KERN_INVALID_ADDRESS 0x0000000b2005a2e0
0
libobjc.A.dylib
objc_release + 16
1
Foundation
-[NSObject(NSKeyValueObservingPrivate) _changeValueForKeys:count:maybeOldValuesDict:maybeNewValuesDict:usingBlock:] + 676
2
Foundation
-[NSObject(NSKeyValueObservingPrivate) _notifyObserversOfChangeFromValuesForKeys:toValuesForKeys:] + 816
3
CoreFoundation
-[CFPrefsSource forEachObserver:] + 332
4
CoreFoundation
-[CFPrefsSource _notifyObserversOfChangeFromValuesForKeys:toValuesForKeys:] + 112
5
CoreFoundation
___CFPrefsDeliverPendingKVONotificationsGuts_block_invoke + 440
6
CoreFoundation
[...]

PR #2046, opened 2022-11-09
Changed default `UserDefaults` from `.standard` to our own Suite

NachoSoto:
Fixes [CSDK-16].
See https://docs.google.com/document/d/1DDyed98RFpIvUwUPJcetc7X4fBhDzf-r1trC4g3zf4c/edit# for reasoning and documentation.

### Motivation

The RevenueCat SDK has always used `UserDefaults.standard` as the default value. The advantage of this was simplicity, however, this brings a few challenges.

### Issues

- While `UserDefaults` is itself thread-safe, our users might be observing `UserDefaults.defaults` changes via `KVO`, `NotificationCenter`, or `@AppStorage` (see https://github.com/RevenueCat/purchases-ios/issues/1527). We update `UserDefaults` from background threads, and users might be getting updates on those threads without handling thread-safety correctly.
- Users can accidentally delete data from the shared `UserDefaults` (see https://rev.cat/userdefaults-crash).

Using a custom `UserDefaults` solves all these issues.

### Change

The new `UserDefaults.revenueCatSuite` is used when:
- Only if users aren’t overriding the `UserDefaults` parameter
- Only if `UserDefaults.standard` has no value for `com.revenuecat.userdefaults.appUserID.new` (the user hasn't used the SDK in that app before).

#### Logged during app start:

> PurchaseTester[35110:6323140] DEBUG: ℹ️ Configuring SDK using RevenueCat's UserDefaults suite.

### Alternatives Considered
See https://docs.google.com/document/d/1c2B4MNAFKs_DcFJ3OE_KnmryHCArBkjlZbAD9_kSXu0/edit#

We considered also migrating data for everyone into the new `UserDefaults`.
This new approach is much simpler, less error prone, easier to implement, and maintain.
The only downside is that existing users won’t benefit from this change, but @aboedo correctly pointed out: if the app is working for them, no reason to change anything.

So in this way, this becomes a purely forward-looking improvement.

[CSDK-16]: https://revenuecats.atlassian.net/browse/CSDK-16?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/FoundationExtensions/UserDefaults+Extensions.swift
+++ b/Sources/FoundationExtensions/UserDefaults+Extensions.swift
@@ -15,20 +15,6 @@ import Foundation
 
 extension UserDefaults {
 
-    #if DEBUG
-    /// The "default" `UserDefaults` to use for the SDK.
-    ///
-    /// Moving foward, this default is `UserDefaults.revenueCatSuite`,
-    /// but existing users will continue using `UserDefaults.standard` for compatibility.
-    /// This is determined by the presence of an app user ID in `UserDefaults.standard`.
-    ///
-    /// Computed property only in `DEBUG` to be able to test it under different conditions.
-    /// In release mode, it gets cached.
-    static var `default`: UserDefaults { .computeDefault() }
-    #else
-    static let `default`: UserDefaults = .computeDefault()
-    #endif
-
     // These are the only 2 documented reasons why `.init(suiteName:)` might return `nil`:
     // - "Because a suite manages the defaults of a specified app group, a suite name
     // must be distinct from your app’s main bundle identifier.
@@ -37,12 +23,17 @@ extension UserDefaults {
     // Because we know at compile time that it's neither of those, this is a safe force-unwrap.
     static let revenueCatSuite: UserDefaults = .init(suiteName: UserDefaults.revenueCatSuiteName)!
 
+    private static let revenueCatSuiteName = "com.revenuecat.user_defaults"
+
 }
 
-private extension UserDefaults {
+extension UserDefaults {
 
-    static let revenueCatSuiteName = "com.revenuecat.user_defaults"
-
+    /// Determines the "default" `UserDefaults` to use for the SDK.
+    ///
+    /// Moving foward, this default is `UserDefaults.revenueCatSuite`,
+    /// but existing users will continue using `UserDefaults.standard` for compatibility.
+    /// This is determined by the presence of an app user ID in `UserDefaults.standard`.
     static func computeDefault() -> UserDefaults {
         let standard: UserDefaults = .standard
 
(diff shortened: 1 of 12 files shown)


vegaro (approved):
LGTM!

NachoSoto:
Thanks!
I'll let this up for a few more days since it's an important change, also in case @aboedo wants to take a look at it.

NachoSoto:
I just did some extra testing with PurchaseTester and this is working as expected.

Sources/FoundationExtensions/UserDefaults+Extensions.swift
```diff
@@ -0,0 +1,58 @@
...
+    /// Moving foward, this default is `UserDefaults.revenueCatSuite`,
+    /// but existing users will continue using `UserDefaults.standard` for compatibility.
+    /// This is determined by the presence of an app user ID in `UserDefaults.standard`.
+    ///
+    /// Computed property only in `DEBUG` to be able to test it under different conditions.
+    /// In release mode, it gets cached.
+    static var `default`: UserDefaults { .computeDefault() }
+    #else
+    static let `default`: UserDefaults = .computeDefault()
+    #endif
```
aboedo, line 30: This feels like an odd workaround to solve a problem that the rest of the project solves with DI - we even pass in the userDefaults when initializing purchases. 

Not saying that doing all the DI in one place is necessarily better, but it does have the advantage of looking exactly the same in all our SDKs, whereas this would be probably impossible to achieve in most languages.

Sources/FoundationExtensions/UserDefaults+Extensions.swift
```diff
@@ -0,0 +1,58 @@
...
+
+private extension UserDefaults {
+
+    static let revenueCatSuiteName = "com.revenuecat.user_defaults"
+
+    static func computeDefault() -> UserDefaults {
+        let standard: UserDefaults = .standard
+
+        if standard.value(forKey: DeviceCache.CacheKeys.appUserDefaults.rawValue) != nil {
+            Logger.debug(Strings.configure.using_user_defaults_standard)
```
aboedo, line 50: it'd be nice to also provide the _reason_ here, i.e.: "found existing data in standard user defaults, so we will use userdefaults.standard"

aboedo (approved):
Looks awesome! Left a question re: DI

Sources/FoundationExtensions/UserDefaults+Extensions.swift
NachoSoto: That's a good point, we don't need to cache this because we only call it once!

NachoSoto: Removed!

Sources/FoundationExtensions/UserDefaults+Extensions.swift
NachoSoto: Done

PR #2048, opened 2022-11-11
`HTTPClient`: added log for failed requests

NachoSoto:
I'm looking into [CSDK-517], and it's hard to tell why we ended up with concurrent requests. It appears that a second request begins before the last one started.
That's seemingly impossible, and the only explanation for the logs missing "request completed" is that the first request failed.

This extra log will help us understand what's happening.

[CSDK-517]: https://revenuecats.atlassian.net/browse/CSDK-517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -214,9 +214,6 @@ private extension HTTPClient {
 
         let statusCode = HTTPStatusCode(rawValue: httpURLResponse.statusCode)
 
-        Logger.debug(Strings.network.api_request_completed(request.httpRequest,
-                                                           httpCode: statusCode))
-
         return Result
             .success(dataIfAvailable(statusCode))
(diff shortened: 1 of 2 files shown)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -244,6 +241,13 @@ private extension HTTPClient {
         )
 
         if let response = response {
+            switch response {
+            case let .success(response):
+                Logger.debug(Strings.network.api_request_completed(request.httpRequest, httpCode: response.statusCode))
+            case let .failure(error):
+                Logger.debug(Strings.network.api_request_failed(request.httpRequest, error: error))
```
NachoSoto, line 248: `debug` only so we don't end up logging this error twice.

PR #2057, opened 2022-11-15
`CustomerInfoManager`: post notification when `CustomerInfo` changes

NachoSoto:
This will be used for #2007.

### DIFF
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -22,16 +22,20 @@ class CustomerInfoManager {
     private let deviceCache: DeviceCache
     private let backend: Backend
     private let systemInfo: SystemInfo
+    private let notificationCenter: NotificationCenter
+
     private let customerInfoCacheLock = Lock()
 
     init(operationDispatcher: OperationDispatcher,
          deviceCache: DeviceCache,
          backend: Backend,
-         systemInfo: SystemInfo) {
+         systemInfo: SystemInfo,
+         notificationCenter: NotificationCenter) {
         self.operationDispatcher = operationDispatcher
         self.deviceCache = deviceCache
         self.backend = backend
         self.systemInfo = systemInfo
+        self.notificationCenter = notificationCenter
     }
 
     func fetchAndCacheCustomerInfo(appUserID: String,
@@ -83,11 +87,11 @@ class CustomerInfoManager {
     }
 
     func sendCachedCustomerInfoIfAvailable(appUserID: String) {
-        guard let info = cachedCustomerInfo(appUserID: appUserID) else {
+        guard let info = self.cachedCustomerInfo(appUserID: appUserID) else {
             return
         }
 
-        sendUpdateIfChanged(customerInfo: info)
+        self.sendUpdateIfChanged(customerInfo: info)
     }
 
     func customerInfo(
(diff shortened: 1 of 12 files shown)


Sources/Identity/CustomerInfoManager.swift
```diff
@@ -231,21 +235,27 @@ class CustomerInfoManager {
...
                 Logger.debug(Strings.customerInfo.sending_updated_customerinfo_to_delegate)
             } else {
                 Logger.debug(Strings.customerInfo.sending_latest_customerinfo_to_delegate)
             }
 
+            if let previousSubscriptions = self.lastSentCustomerInfo?.activeSubscriptions,
+               previousSubscriptions != customerInfo.activeSubscriptions {
+                self.notificationCenter.post(name: Self.customerInfoChangedActiveSubscriptionsNotification,
+                                             object: customerInfo)
+            }
```
aboedo, line 250: I think we should just play it on the safe side here and invalidate cache on _any_ change to the customerInfo. 
What if for some reason the customerInfo just backfilled a bunch of expired subscriptions that would actually change the result of eligibility? 
For example: 

1. Log in as user A for the first time. Eligible for all things. 
2. Restore purchases, there were a bunch of old purchases that this person made as an anonymous user last year. 
3. User is no longer eligible, but there are no active subscriptions, so we're not refreshing cache

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -231,21 +235,27 @@ class CustomerInfoManager {
...
+
             self.lastSentCustomerInfo = customerInfo
-            self.operationDispatcher.dispatchOnMainThread { [observers = self.customerInfoObserversByIdentifier] in
-                for closure in observers.values {
-                    closure(customerInfo)
+
+            if !self.customerInfoObserversByIdentifier.isEmpty {
+                self.operationDispatcher.dispatchOnMainThread { [observers = self.customerInfoObserversByIdentifier] in
+                    for closure in observers.values {
+                        closure(customerInfo)
```
aboedo, line 257: Curious about this, why don't we reuse this mechanism for getting notified about changes, instead of adding the notificationCenter notification?

aboedo (changes requested):
I think there could be cases where we're not invalidating correctly

Sources/Identity/CustomerInfoManager.swift
NachoSoto: > I think we should just play it on the safe side here and invalidate cache on any change to the customerInfo.

Makes sense 👍🏻  great feedack.

Sources/Identity/CustomerInfoManager.swift
NachoSoto: For one because this was checking only when active subscriptions changed.
Now that I'm changing the cache invalidation to checking `CustomerInfo` changes we could.

However, this is used by #2007. Using a notification makes it a lot simpler so that we don't have to manually couple `TrialOrIntroPriceEligibilityChecker` to `CustomerInfoManager`. What do you think?

NachoSoto:
Updated this to post a notification any time `CustomerInfo` changes.

NachoSoto:
Bump @aboedo

tonidero (approved):
LGTM!

Sources/Identity/CustomerInfoManager.swift
aboedo: the part that I find a bit concerning is that we have 2 things updating from a change in customerInfo, but we don't have much control over the ordering. 
if we're using this for cache invalidation, doesn't it mean that it's possible that you have a closure coming in from here that uses an outdated cache of intro eligibility values?

NachoSoto: The 2 sources is a valid concern. I think they serve different purposes though.
The notification pattern is useful for 1-N notifications without strongly coupling types, and it simplifies the hierarchy. The current dependency injection approach is prone to retain cycles and leaks as exposed in the other PRs.

As for ordering, I'm not sure that's an issue in this case. Not sure what you mean exactly. #2007 won't recreate the eligibility, simply invalidate the cache for the next time it's requested.

NachoSoto: Bump @aboedo.

NachoSoto:
Closing this, will replace #2007 with using the existing observation methods.

PR #2059, opened 2022-11-15
Created `TimingUtil` to measure and log methods that are too slow

NachoSoto:
For [CSDK-512].

This will print messages like:
> 🍎⚠️ Computation took too long (2.13 seconds)

#### `async` methods:

```swift
let (result, duration) = try await TimingUtil.measure {
   return try await asyncMethod()
}
```

```swift
let result = try await TimingUtil.measureAndLogIfTooSlow(
    threshold: 2,
    message: "Computation too slow",
    level: .warn,
    intent: .appleWarning
) {
   try await asyncMethod()
}
```

#### Completion-block methods:

```swift
TimingUtil.measure { completion in
    work { completion($0) }
} result: { result, duration in
    print("Result: \(result) calculated in \(duration) seconds")
}
```

```swift
TimingUtil.measureAndLogIfTooSlow(
    threshold: 2,
    message: "Computation too slow",
    level: .warn,
    intent: .appleWarning
) { completion in
    work { completion($0) }
} result: { result in
    print("Finished computing: \(result)")
}
```

[CSDK-512]: https://revenuecats.atlassian.net/browse/CSDK-512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Misc/TimingUtil.swift
+++ b/Sources/Misc/TimingUtil.swift
@@ -1,0 +1,168 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  TimingUtil.swift
+//
+//  Created by Nacho Soto on 11/15/22.
+
+import Foundation
+
+internal enum TimingUtil {
+
+    typealias Duration = TimeInterval
+
+}
+
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+extension TimingUtil {
+
+    /// Measures the time to execute `work` and returns the result and the duration.
+    /// Example:
+    /// ```swift
+    /// let (result, duration) = try await TimingUtil.measure {
+    ///    return try await asyncMethod()
+    /// }
+    /// ```
+    static func measure<Value>(
+        _ work: () async throws -> Value
+    ) async rethrows  -> (result: Value, duration: Duration) {
+        let start: DispatchTime = .now()
+        let result = try await work()
+
+        return (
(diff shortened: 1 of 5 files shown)


Sources/Misc/TimingUtil.swift
```diff
@@ -0,0 +1,168 @@
...
+extension TimingUtil {
+
+    /// Measures the time to execute `work` and returns the result and the duration.
+    /// Example:
+    /// ```swift
+    /// let (result, duration) = try await TimingUtil.measure {
+    ///    return try await asyncMethod()
+    /// }
+    /// ```
+    static func measure<Value>(
```
aboedo, line 32: I'm not a huge fan of static methods for stuff like this because they complicate DI, i.e: if we wanted to test that some consumer of this method is indeed calling it, static funcs will make that a bit more awkward

aboedo (approved):

Sources/Misc/TimingUtil.swift
NachoSoto: I'm on the same page, using `static` stuff all over the place is a big smell. 

I wish we had a proper DI framework though, cause this would require wiring this down 3 or 4 layers down.
I'm torn, I'd say YAGNI, but we could inject it later if we need to?

NachoSoto: > if we wanted to test that some consumer of this method is indeed calling it, static funcs will make that a bit more awkward

I'd say this might not be that critical of a thing to test. If we stopped calling it all we'd lose is a log.

We can also test that we're calling this using `TestLogger`.

NachoSoto: I'll go ahead and merge this since we can easily change this to be non-static to improve testing (or use `TestLogger`).

PR #2061, opened 2022-11-15
`StoreKit`: added logs when purchasing and product requests are too slow

NachoSoto:
Using #2059 in a few places for now.
We can add this to more places, and tweak the thresholds.

Fixes [CSDK-512].

[CSDK-512]: https://revenuecats.atlassian.net/browse/CSDK-512?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
+++ b/Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
@@ -86,12 +86,16 @@ final class ProductsFetcherSK1: NSObject {
 
     func products(withIdentifiers identifiers: Set<String>,
                   completion: @escaping (Result<Set<SK1StoreProduct>, PurchasesError>) -> Void) {
-        self.sk1Products(withIdentifiers: identifiers) { skProducts in
-            let result = skProducts
-                .map { Set($0.map(SK1StoreProduct.init)) }
-
-            completion(result)
-        }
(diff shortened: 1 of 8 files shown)


Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
```diff
@@ -86,12 +86,16 @@ final class ProductsFetcherSK1: NSObject {
 
     func products(withIdentifiers identifiers: Set<String>,
                   completion: @escaping (Result<Set<SK1StoreProduct>, PurchasesError>) -> Void) {
-        self.sk1Products(withIdentifiers: identifiers) { skProducts in
-            let result = skProducts
-                .map { Set($0.map(SK1StoreProduct.init)) }
-
-            completion(result)
-        }
+        TimingUtil.measureAndLogIfTooSlow(
```
aboedo, line 89: I'd be great to actually test that this class actually logs if this is too slow, to future-proof so that a refactor can't accidentally get rid of the logging

aboedo (approved):
the `TimingUtil` is really cool 😎
And the API for it is :chefskiss:

We should try to find a way to test that classes are actually using it, though, but that doesn't have to block this PR

aboedo (approved):

PR #2062, opened 2022-11-16
[WIP] Separate `ReceiptParser` into SPM

NachoSoto:
Fixes [CSDK-17]

Screenshot 2022-11-16 at 10 30 01

### TODO:
~~- [ ] Create separate SPM for shared types (`Logger`?)~~ This would be a pain to do with CocoaPods
- [ ] Document public methods
- [ ] Add new API to `APITester`s
- [x] Fix CocoaPods (https://stackoverflow.com/questions/67894280/how-to-import-a-cocoapods-subspec-module-instead-of-the-entire-pod-in-swift)
- [ ] Verify `Carthage` integration still works
- [ ] `README`
- [x] Move `ReceiptParser` tests to SPM (include #2043)
- [ ] Fix tests
- [ ] Move calls to `Logger` outside module
- [ ] Verify `release-checks` all pass


NachoSoto:
Turns out this approach might not be supported: https://forums.swift.org/t/unable-to-integrate-a-remote-package-that-has-local-packages/53146/5

Might have to keep the folder structure the same, and instead expose a separate target in the main `Package.swift`.

We get the same result, but I was excited about this approach because it would allow creating many more smaller SPMs that would help us keep different parts of the SDK isolated like we do in Android.

NachoSoto:
I filed a Feedback for this:
Screenshot 2022-11-16 at 10 44 01

NachoSoto:
I'm gonna need a 3rd workaround for my 2nd workaround... Re-opening the project breaks it:
Screenshot 2022-11-16 at 12 39 07

Package.swift
```diff
@@ -34,8 +29,22 @@ let package = Package(
...
+                 targets: ["RevenueCat"]),
+        .library(name: "ReceiptParser",
+                 type: .static,
+                 targets: ["ReceiptParser"])
     ],
     dependencies: dependencies,
-    targets: resolveTargets()
+    targets: [
+        .target(name: "RevenueCat",
+                dependencies: [.byName(name: "ReceiptParser")],
```
aboedo, line 40: In my experience, this will complicate things quite a bit. 
It's a total headache to manage imports from a local package, we had something like this back in the objc days with PurchasesCoreSwift and it broke hybrids in all sorts of ways. 

I think we should just have these two be independent packages, which just happen to have the same files. 
If you use RC, you don't need the ReceiptParser package. If you use the ReceiptParser package, and you want to move to the full RevenueCat, you need to remove ReceiptParser and replace it with RevenueCat.

aboedo: One bummer with this is that SPM from Xcode might prompt you to install both by default, right?

NachoSoto: > I think we should just have these two be independent packages, which just happen to have the same files.

That's a great idea, which will simplify this tremendously.
I wanted to have a call with you to discuss the approach, but thanks for providing the best path forward already :D

aboedo: happy to chat about it whenever!

NachoSoto:
New approach: #2096.

PR #2063, opened 2022-11-17
use `transactions.all` to observe transactions in observer mode with SK2

aboedo:
While it's not documented, `Transaction.updates` will only send updates for purchases being made on other devices or renewed. Purchases made on the same device will not receive an update through `.updates`. Instead, those get them through `.all`. 

This draft PR updates the logic so that in observer mode we don't miss transactions made on the same device.

### DIFF
--- a/Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
+++ b/Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
@@ -37,18 +37,28 @@ class StoreKit2TransactionListener {
         self.delegate = delegate
     }
 
-    func listenForTransactions() {
+    func listenForTransactions(observerMode: Bool) {
         self.taskHandle?.cancel()
         self.taskHandle = Task { [weak self] in
-            for await result in StoreKit.Transaction.updates {
-                guard let self = self else { break }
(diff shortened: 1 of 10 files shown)


Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
```diff
@@ -37,7 +37,12 @@ class StoreKit2TransactionListener {
     func listenForTransactions() {
         self.taskHandle?.cancel()
         self.taskHandle = Task { [weak self] in
-            for await result in StoreKit.Transaction.updates {
+            // Transaction.updates only sends updates from transactions made on different devices,
+            // so on observer mode we need to get `all` instead.
+
+            // TODO: ability to access observer mode setting from here
+            let transactionsStream = observerModeEnabled ? StoreKit.Transaction.all : StoreKit.Transaction.updates
```
aboedo, line 44: @NachoSoto I can wrap this up tonight if you don't have the time, basically we need to pass in system info or the setting here

NachoSoto (changes requested):
See #2069 for why this implementation would differ from SK1.

aboedo:
closing since this isn't needed

PR #2066, opened 2022-11-17
`SKTestSession`: finish all unfinished transactions before starting each test

NachoSoto:
Filed this as `FB11799675`. I noticed some tests began with prior transactions (despite using `SKTestSession.clearTransactions`), so this workaround ensures each test begins in a clean state.

### DIFF
--- a/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
@@ -53,6 +53,9 @@ class BaseBackendIntegrationTests: XCTestCase {
         }
 
         if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
+            // Despite calling `SKTestSession.clearTransactions` tests sometimes
+            // begin with leftover transactions. This ensures that we remove them
+            // to always start with a clean state.
             await self.finishAllUnfinishedTransactions()
         }
 
--- a/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
@@ -52,6 +52,10 @@ class BaseBackendIntegrationTests: XCTestCase {
             throw ErrorUtils.configurationError(message: "Must set configuration in `Constants.swift`")
         }
 
+        if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
+            await self.finishAllUnfinishedTransactions()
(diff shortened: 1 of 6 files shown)


Tests/StoreKitUnitTests/TestHelpers/StoreKitConfigTestCase+Extensions.swift
```diff
@@ -0,0 +1,151 @@
...
+//
+//  Created by Nacho Soto on 11/17/22.
+
+import Nimble
+@testable import RevenueCat
+@preconcurrency import StoreKit
+import XCTest
+
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+extension StoreKitConfigTestCase {
```
NachoSoto, line 20: I moved this from `StoreKitTestHelpers.swift` into this file since this is only for `StoreKitConfigTestCase` and not relevant for integration tests.

NachoSoto:
I think this is exposing [CSDK-517] even more frequently.
(See #2055).

NachoSoto:
Actually, I wonder if this will make that issue better (see https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/8929/workflows/e990dfcf-effb-4923-b688-1e3655aee62a/jobs/44610).
I bet there are pending transactions, and that's why even retrying triggers the same issue.

Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
```diff
@@ -52,6 +52,10 @@ class BaseBackendIntegrationTests: XCTestCase {
             throw ErrorUtils.configurationError(message: "Must set configuration in `Constants.swift`")
         }
 
+        if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
+            await self.finishAllUnfinishedTransactions()
+        }
```
aboedo, line 57: let's add a comment explaining why this is needed, it'd be reasonable for a maintainer to assume this is a bug

aboedo (approved):

Tests/StoreKitUnitTests/TestHelpers/StoreKitConfigTestCase+Extensions.swift
NachoSoto: Done.

Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
NachoSoto: Done

PR #2068, opened 2022-11-17
`Integration Tests`: always run them in random order locally

NachoSoto:
This is already set to random in `CI-BackendIntegration.xctestplan`, but it's useful to default to random when running them locally too, to potentially catch any state issues.


taquitos (approved):
🐐🎈

tonidero (approved):

PR #2069, opened 2022-11-17
Added integration and unit tests to verify observer mode behavior

NachoSoto:
See [TRIAGE-179].

I wrote all these to cover the change in #2063. I wrote a failing test that only passed with the change in #2063. However, when running that in SK1, that would fail.
With the change as suggested in #2063, the SK2 behavior would differ from SK1.

From the `Transaction.all` docs:
> This sequence returns the user’s transaction history current to the moment you access the sequence. The sequence emits a finite number of transactions. If the App Store processes new transactions for the user while you’re accessing this sequence, the new transactions appear in the transaction listener, updates.

These new tests verify the behavior of the SDK when encountering transactions created prior to initializing the SDK, and some made after initialization.

Integration tests now are as follows:
- `BaseStoreKitIntegrationTests`
  - `StoreKit1IntegrationTests`
  - `StoreKit2IntegrationTests`
  - `BaseStoreKitObserverModeIntegrationTests`:
    - `StoreKit1ObserverModeIntegrationTests`
    - `StoreKit2ObserverModeIntegrationTests`
    - `StoreKit1ObserverModeWithExistingPurchasesTests`
    - `StoreKit2ObserverModeWithExistingPurchasesTests`

Depends on #2064, #2066, and #2134.

[TRIAGE-179]: https://revenuecats.atlassian.net/browse/TRIAGE-179?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
@@ -481,14 +481,28 @@ class StoreKit2ObserverModeIntegrationTests: StoreKit1ObserverModeIntegrationTes
 
     override class var storeKit2Setting: StoreKit2Setting { return .enabledForCompatibleDevices }
 
+    override func setUp() async throws {
+        try await super.setUp()
+
+        try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
+    }
+
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
     func testPurchaseInDevicePostsReceipt() async throws {
-        try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
-
         try await self.purchaseProductFromStoreKit()
 
-        let info = try await Purchases.shared.restorePurchases()
-        try await self.verifyEntitlementWentThrough(info)
+        await asyncWait(
+            until: {
+                let entitlement = try? await Purchases.shared
+                    .customerInfo(fetchPolicy: .fetchCurrent)
+                    .entitlements[Self.entitlementIdentifier]
+
+                return entitlement?.isActive == true
+            },
+            timeout: .seconds(30),
+            pollInterval: .seconds(1),
+            description: "Entitlement didn't become active"
+        )
     }
 
 }
@@ -775,6 +789,7 @@ private extension BaseStoreKitIntegrationTests {
         }
     }
 
+    /// Purchases a product directly with StoreKit.
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
     @discardableResult
     func purchaseProductFromStoreKit() async throws -> Product.PurchaseResult {
--- a/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
@@ -423,7 +423,7 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
 
         // 4. Verify purchase went through
 
-        let entitlement = try await self.verifyEntitlementWentThrough(customerInfo)
+        try await self.verifyEntitlementWentThrough(customerInfo)
     }
 
     @available(iOS 15.2, tvOS 15.2, macOS 12.1, watchOS 8.3, *)
--- a/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
@@ -8,31 +8,21 @@
 
 import Nimble
 @testable import RevenueCat
+@preconcurrency import StoreKit // `PurchaseResult` is not `Sendable`
 import StoreKitTest
 import UniformTypeIdentifiers
 import XCTest
 
-// swiftlint:disable file_length type_body_length
+// swiftlint:disable file_length type_body_length type_name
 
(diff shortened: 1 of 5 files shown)


NachoSoto:
This is ready for review.

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -433,10 +453,97 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
...
+        try self.testSession.buyProduct(productIdentifier: Self.monthlyNoIntroProductID)
+
+        let info = try await Purchases.shared.restorePurchases()
+        try await self.verifyEntitlementWentThrough(info)
+    }
+
+}
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+class StoreKit2ObserverModeWithExistingPurchasesTests: StoreKit1ObserverModeWithExistingPurchasesTests {
```
tonidero, line 492: Hmm we don't have any tests here, I guess the plan is to add some later? If so, should we add a comment or something? We can also delete it for now.

tonidero (approved):

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -433,10 +453,97 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
...
+
+    override class var storeKit2Setting: StoreKit2Setting { return .enabledForCompatibleDevices }
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func testPurchaseInDevicePostsReceipt() async throws {
+        try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
+
+        try await self.purchaseProductFromStoreKit()
+
+        let info = try await Purchases.shared.restorePurchases()
```
aboedo, line 474: I'm confused about this - when we do restore we're forcing a receipt post, so this effectively is just testing that if the purchase there it gets posted, right? 
like, shouldn't this test just be "purchase from storekit, then verify entitlement went through"? Same question for the other one

aboedo: To be clear, my concern is that the minute that we call `restorePurchases`, we're not really testing observer mode behavior

NachoSoto: Okay I see what you're saying. I guess it's a very basic test. What else would you do if we don't restore purchases to verify what the SDK does in this case of purchasing from "outside" the SDK?

aboedo: after a purchase was made from "outside", the listeners should get notified and we should: 
- get a notification from the customerInfoChanged delegate
- if you call getCustomerInfo, it should have the entitlements
If those are not happening, then observer mode isn't working. 
It could be that it just doesn't work well with StoreKitTest or something, but if we call restorePurchases we do a force sync to backend, so we're not really doing what someone in observer mode would be doing

NachoSoto: Oh cool, let me try changing the test to that then.

NachoSoto: I updated this using #2134.

The good news: this works great! It waits for the transaction to be detected by the listener, which posts it to the backend, and then we see the entitlement.
The bad news: for some reason the transaction listener doesn't get the update until like 20-30 seconds later, which makes this test extremely slow (therefore why I added such a high `timeout`).

I think it's a very super useful test, but it being so slow is annoying. What do you think?

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -449,10 +469,111 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
...
+        try await super.setUp()
+
+        try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
+    }
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func testPurchaseInDevicePostsReceipt() async throws {
+        try await self.purchaseProductFromStoreKit()
+
+        await asyncWait(
```
NachoSoto, line 494: @aboedo this is the new test.

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -449,10 +469,111 @@ class StoreKit1IntegrationTests: BaseBackendIntegrationTests {
...
+        // 3. Configure SDK
+        try await super.setUp()
+    }
+
+    func testDoesNotSyncExistingPurchase() async throws {
+        try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
+
+        let info = try await Purchases.shared.customerInfo(fetchPolicy: .fetchCurrent)
+        self.assertNoPurchases(info)
+    }
```
aboedo, line 566: this test case is interesting. ideally purchases don't go through before the SDK gets configured, otherwise, this is what happens - we miss the purchase. 

in real life the most common case is a renewal which happens with the SDK being closed, and in that case, it will still show up through the queue. Same for purchases on other devices.

aboedo (approved):

PR #2070, opened 2022-11-17
`StoreKit2` listeners: set `Task` `priority` to `.utility`

NachoSoto:
A small performance improvement, since technically these don't need higher priorities.

### DIFF
--- a/Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
+++ b/Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
@@ -38,7 +38,7 @@ class StoreKit2StorefrontListener {
     }
 
     func listenForStorefrontChanges() {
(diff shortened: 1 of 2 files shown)


aboedo:
🤔 neither of them should happen often so it shouldn't make a difference in practice, but I do feel like that might be more of an argument for updating as quickly as possible if they do happen, particularly for transactions. No strong feelings about this one, though

NachoSoto:
Yeah I can see either way. The way I was thinking about it is that this is something that the SDK is running in the background and shouldn't have a very high priority that could potentially make the UI of an app slow down or drop frames.

But maybe "utility" would make more sense?
From the docs:
> Utility priority is intended for user-visible tasks such as fetching data from the network, copying files, or importing data.

aboedo:
that seems pretty applicable

NachoSoto:
Changed!

aboedo (approved):

PR #2071, opened 2022-11-18
`Tests`: replaced `toEventually` with new `waitUntilValue` to simplify tests

NachoSoto:
Example:
```diff
-        var result: Result<Offerings, OfferingsManager.Error>?
-        offeringsManager.offerings(appUserID: MockData.anyAppUserID) {
-            result = $0
+        let result = waitUntilValue { completed in
+            self.offeringsManager.offerings(appUserID: MockData.anyAppUserID) {
+                completed($0)
+            }
         }
-        expect(result).toEventuallyNot(beNil())
```

This removes the need to _poll_ for when the value is returned, and instead it gets detected immediately, potentially providing a significant boost in test execution speed.

I've also used `Nimble`'s own `waitUntil` for cases where we don't need to return a value.

 #### `toEventually` usage:
- Before this PR: 353
- After: 229

There's still a lot of legitimate usage of `toEventually`, but I'm sure I missed other places where this can also simplify the code. This is a good start for now.

### DIFF
--- a/Tests/StoreKitUnitTests/ProductsManagerTests.swift
+++ b/Tests/StoreKitUnitTests/ProductsManagerTests.swift
@@ -23,13 +23,10 @@ class ProductsManagerTests: StoreKitConfigTestCase {
         let manager = try createManager(storeKit2Setting: .disabled)
 
         let identifier = "com.revenuecat.monthly_4.99.1_week_intro"
-        var receivedProducts: Result<Set<StoreProduct>, PurchasesError>?
-
-        manager.products(withIdentifiers: Set([identifier])) { products in
-            receivedProducts = products
+        let receivedProducts = waitUntilValue(timeout: Self.requestDispatchTimeout) { completed in
+            manager.products(withIdentifiers: Set([identifier]), completion: completed)
         }
 
-        expect(receivedProducts).toEventuallyNot(beNil(), timeout: Self.requestDispatchTimeout)
         let unwrappedProducts = try XCTUnwrap(receivedProducts?.get())
 
         let product = try XCTUnwrap(unwrappedProducts.onlyElement).product
@@ -46,13 +43,10 @@ class ProductsManagerTests: StoreKitConfigTestCase {
         let manager = try createManager(storeKit2Setting: .enabledForCompatibleDevices)
 
         let identifier = "com.revenuecat.monthly_4.99.1_week_intro"
-        var receivedProducts: Result<Set<StoreProduct>, PurchasesError>?
-
-        manager.products(withIdentifiers: Set([identifier])) { products in
-            receivedProducts = products
+        let receivedProducts = waitUntilValue(timeout: Self.requestDispatchTimeout) { completed in
+            manager.products(withIdentifiers: Set([identifier]), completion: completed)
         }
 
-        expect(receivedProducts).toEventuallyNot(beNil(), timeout: Self.requestDispatchTimeout)
         let unwrappedProducts = try XCTUnwrap(receivedProducts?.get())
 
         let product = try XCTUnwrap(unwrappedProducts.onlyElement).product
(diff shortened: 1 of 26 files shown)


Tests/StoreKitUnitTests/ProductsManagerTests.swift
```diff
@@ -46,13 +43,10 @@ class ProductsManagerTests: StoreKitConfigTestCase {
         let manager = try createManager(storeKit2Setting: .enabledForCompatibleDevices)
 
         let identifier = "com.revenuecat.monthly_4.99.1_week_intro"
-        var receivedProducts: Result<Set<StoreProduct>, PurchasesError>?
```
NachoSoto, line 49: No longer need to specify types every time, now they're inferred by the method we call.

NachoSoto: Never mind, that only works in Xcode 14.

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -646,12 +646,12 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
         let message = "Failed to get managementURL from CustomerInfo. Details: customerInfo is nil."
         mockManageSubsHelper.mockError = ErrorUtils.customerInfoError(withMessage: message)
 
-        var receivedError: Error?
-        orchestrator.showManageSubscription { error in
-            receivedError = error
+        let receivedError: Error? = waitUntilValue { completed in
+            self.orchestrator.showManageSubscription { error in
+                completed(error)
+            }
```
vegaro, line 652: This can be:

```
let receivedError: Error? = waitUntilValue { completed in
	self.orchestrator.showManageSubscription(completion: completed)
}
```
Correct? Not sure if that's more legible though

Tests/UnitTests/Purchasing/Purchases/PurchasesRestoreTests.swift
```diff
@@ -27,23 +27,26 @@ class PurchasesRestoreTests: BasePurchasesTests {
...
-            receivedError = error as NSError?
+        let receivedError = waitUntilValue { completed in
+            self.purchases.restorePurchases { (_, error) in
+                completed(error as NSError?)
+            }
         }
 
-        expect(self.mockOperationDispatcher.invokedDispatchOnMainThreadCount) == 1
-        expect(receivedError?.code).toEventually(equal(ErrorCode.missingReceiptFileError.rawValue))
+        expect(self.mockOperationDispatcher.invokedDispatchOnMainThreadCount) == 3
```
vegaro, line 48: why has this been changed to 3?

vegaro (commented):
still have some files to check but looking good so far

Tests/UnitTests/Purchasing/Purchases/PurchasesRestoreTests.swift
NachoSoto: The 1 was passing because it is 1 at some point. But in practice once the operation changes this became 3. I suppose I should change this to `>= 1` to match the same semantics as before, what do you think?

vegaro: Could `>= 1` give a false positive? Right now, there are 3 invokes on the main thread, one of them being the restoring one, which is the one we are testing for. If that one stops being called on the main thread (total of 2 invokes on main thread), the test will still pass

vegaro (approved):
Just that question about the `>= 1`

Tests/UnitTests/Purchasing/Purchases/PurchasesRestoreTests.swift
NachoSoto: Probably, but that was an issue with the existing test as well.

NachoSoto: I looked into this more. This just looks like a bad test. The same operation dispatcher is used for all of `Purchases`, so other background tasks are calling `dispatchOnMainThread` as well. The test has always passed because it was effectively checking `>= 1` with that `toEventually`. Eventually it was 1, *even before `restorePurchases` was called*, but that's not how the test was written.

~~Unfortunately can't think of a way to improve the test since we have no way to identify which operation was invoked.~~

Scratch that, of course there's a simple way haha

NachoSoto: Since the test is `testRestorePurchasesCallsCompletionOnMainThreadWhenMissingReceipts` I added this in the completion block:
```swift
expect(Thread.isMainThread) == true
``

PR #2076, opened 2022-11-21
Fixed lint issues with `SwiftLint 0.5.0`

NachoSoto:
[The new release](https://github.com/realm/SwiftLint/releases/tag/0.50.0) is much faster as [it uses](https://github.com/realm/SwiftLint/pull/4216) the [new SwiftParser](https://github.com/apple/swift-syntax).

### Changes:
- Disabled `orphaned_doc_comment`: it provides warnings for non-public declarations (see https://github.com/realm/SwiftLint/issues/4573#issuecomment-1322502917), but those are still useful for us.
- `function_body_length` has a new logic
- `large_tuple` seems to have new default thresholds, so I've specified them in `.swiftlint` to relax them. Tuples will most likely even become `Equatable` soon in Swift 6.0 so they're here to stay.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -16,6 +16,9 @@ opt_in_rules:
   - vertical_parameter_alignment
   - vertical_parameter_alignment_on_call
 
+disabled_rules:
+  - orphaned_doc_comment
+
 custom_rules:
   xctestcase_superclass:
     included: ".*\\.swift"
@@ -28,6 +31,10 @@ identifier_name:
   max_length: 
     warning: 60 
     error: 80
+
+large_tuple:
+  warning: 4
+  error: 5
 
 missing_docs:
   excludes_inherited_types: false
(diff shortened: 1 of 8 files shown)


.swiftlint.yml
```diff
@@ -16,6 +16,9 @@ opt_in_rules:
   - vertical_parameter_alignment
   - vertical_parameter_alignment_on_call
 
+disabled_rules:
+  - orphaned_doc_comment
```
aboedo, line 20: do we get false positives often enough that it's worth disabling at the project level rather than on a case-by-case basis?

aboedo (approved):

.swiftlint.yml
NachoSoto: We get a fair amount of `internal` declarations (though that [should get fixed](https://github.com/realm/SwiftLint/issues/4573#issuecomment-1322627749), but we also have some blocks of text like:
```swift
/**
Some explanation about a method
*/

let some = code()
```

I was disabling them individually, but after a while I realized that more docs are always better so not sure we get any benefit from the rule.

NachoSoto: I'll get this merged and we can re-enable the rule when they fix that regression.

aboedo: 👍

Issue #1906, opened 2022-09-15
Has any one meet `Cached appUserID has been deleted from user defaults.` crash on iOS16?


alfredcc:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Development/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

1. Environment
   1. Platform: iOS
   2. SDK version: 4
   3. StoreKit 2 (_enabled with `useStoreKit2IfEnabled`_) (Y/N): Y
   4. OS version: iOS16
   5. Xcode version: Xcode14
  
**Additional context**
```
RevenueCat/DeviceCache.swift:70: Fatal error: [Purchases] - Cached appUserID has been deleted from user defaults. This leaves the SDK in an undetermined state. Please make sure that RevenueCat entries in user defaults don't get deleted by anything other than the SDK. More info: https://rev.cat/userdefaults-crash
```
```
0x3820c _assertionFailure(_:_:file:line:flags:) + 312
1  Grow                           0x95260c DeviceCache.handleUserDefaultsChanged(notification:) + 4388398604 (<compiler-generated>:4388398604)
2  Grow                           0x95267c @objc DeviceCache.handleUserDefaultsChanged(notification:) + 4388398716 (<compiler-generated>:4388398716)
3  CoreFoundation                 0x377c4 __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 148
4  CoreFoundation                 0xde75c ___CFXRegistrationPost_block_invoke + 88
5  CoreFoundation                 0xc1a34 _CFXRegistrationPost + 440
6  CoreFoundation                 0x4bdc8 _CFXNotificationPost + 704
7  Foundation                     0x5c4dc -[NSNotificationCenter postNotificationName:object:userInfo:] + 92
[...]

PR #2079, opened 2022-11-22
`DeviceCache`: removed `fatalError` for users not overriding `UserDefaults`

NachoSoto:
We can do this now thanks to #2046. Solves #1906.
We no longer need to do this check if we're using our own `UserDefaults`, which simplifies the code, avoids a dangerous `fatalError`, and makes [CSDK-519] less likely.

[CSDK-519]: https://revenuecats.atlassian.net/browse/CSDK-519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Caching/DeviceCache.swift
+++ b/Sources/Caching/DeviceCache.swift
@@ -36,7 +36,7 @@ class DeviceCache {
     /// cleared from under the SDK
     private let appUserIDHasBeenSet: Atomic<Bool> = false
 
-    private var userDefaultsObserver: NSObjectProtocol!
+    private var userDefaultsObserver: NSObjectProtocol?
 
     convenience init(sandboxEnvironmentDetector: SandboxEnvironmentDetector,
                      userDefaults: UserDefaults) {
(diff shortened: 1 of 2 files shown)


Sources/Caching/DeviceCache.swift
```diff
@@ -57,14 +57,21 @@ class DeviceCache {
...
-            queue: nil, // Run synchronously on the posting thread
-            using: { [weak self] notification in
-                self?.handleUserDefaultsChanged(notification: notification)
-            }
-        )
+        // Observe `UserDefaults` changes through `handleUserDefaultsChanged`
+        // to ensure that users don't remove the data from the SDK, which would
+        // leave it in an undetermined state. See https://rev.cat/userdefaults-crash
+        // If the user is not using a custom `UserDefaults`, we don't need to
+        // because they have no access to it.
```
aboedo, line 64: the good news is that folks who _are_ setting their custom user defaults for revenuecat usually know not to delete its contents, so the code will likely very rarely run

aboedo (approved):

PR #2080, opened 2022-11-23
Added `LogLevel.verbose`

NachoSoto:
I want to add some extra logs to figure out [CSDK-517] which would be too verbose (no pun intended) for `debug`, so this will allow us to have some additional logging that wouldn't be enabled by default.

[CSDK-517]: https://revenuecats.atlassian.net/browse/CSDK-517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Logging/LogIntent.swift
+++ b/Sources/Logging/LogIntent.swift
@@ -16,10 +16,11 @@ import Foundation
 
 enum LogIntent {
 
-    case appleError
-    case appleWarning
+    case verbose
     case info
     case purchase
+    case appleWarning
+    case appleError
     case rcError
     case rcPurchaseSuccess
     case rcSuccess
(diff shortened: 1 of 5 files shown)


vegaro (commented):
This is breaking our public API correct?

NachoSoto:
By breaking do you mean because we added a new `case`?
Technically yes, so we could tag this as `feat` if we want to be 100% correct, but I'd be surprised anyone is `switch`ing over our `LogLevel`?

Well I suppose users implementing their own `LogHandler` might.

aboedo (approved):

aboedo:
@NachoSoto this does technically fall under feat since it's a new enum case. 
But we do modify the values, so that's the breaking part, since it's an int enum and we're effectively moving them all by 1. Maybe we can just keep the existing numbering by setting it explicitly, that way we're being extra safe and not modifying an existing public value

NachoSoto:
Oh yeah that's a great point!!

NachoSoto:
Changed to stable IDs and made this `feat` to be 100% safe.

vegaro:
Thanks for explaining @aboedo , that's exactly what I meant. I should've been more clear about my concern :)

PR #2088, opened 2022-11-28
`TimingUtil`: fixed Xcode 13.2 compilation

NachoSoto:
This compiles fine in Xcode 14 and Xcode 13.4, but produces the following error in Xcode 13.2:
```
static method 'measureAndLogIfTooSlow(threshold:message:level:intent:_:)' requires that 'PurchaseStrings' conform to 'Sendable'
```

To fix that, we can pass the `String` directly instead of a `CustomStringConvertible`.

See https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/9028/workflows/fdd6cb34-a28f-4f1c-9e56-
abecc4eab8c0/jobs/45630

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -425,7 +425,7 @@ final class PurchasesOrchestrator {
         do {
             result = try await TimingUtil.measureAndLogIfTooSlow(
                 threshold: .purchase,
-                message: Strings.purchase.sk2_purchase_too_slow) {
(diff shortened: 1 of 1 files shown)


NachoSoto:
I considered always running the `Xcode 13.x` job in every PR, but I think catching these things in `main` is fine. Thoughts?

NachoSoto:
Oh wait, we do compile on Xcode 13.4 in every PR (the `run-test-ios-14` job), but only do `Xcode 13.2` on `main`, which is the only thing that was broken 🤦🏻

NachoSoto:
This was super subtle, good thing #1990 added a CI job so we don't break it again.

vegaro (approved):

PR #2094, opened 2022-11-28
Fixed `SwiftLint` installation

NachoSoto:
We need an older version compatible with `macOS 11`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ commands:
             - vendor/bundle
       - restore_cache:
           keys:
-            - homebrew-tap-cache
+            - homebrew-tap-cache-0.46.5
       - run:
           name: Install swiftlint 0.46.5
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -40,9 +40,21 @@ commands:
           key: v1-gem-cache-{{ checksum "Gemfile.lock" }}
           paths:
             - vendor/bundle
+      - restore_cache:
+          keys:
+            - homebrew-tap-cache
```
vegaro, line 45: This will never change and will always be a cache hit (note how in line 40 for example we are adding a checksum of Gemfile.lock to the key). I think it's fine since it's an older version of swiftlint what we are installing, but wanted to make sure

NachoSoto: Right, there's no "lock" file in this case so I guess the only solution is to change the key if we ever change the version.
But you just answered my own question 🙈 https://revenuecat.slack.com/archives/C03HPN9UE80/p1669652256396329

vegaro (approved):

.circleci/config.yml
NachoSoto: Oh great idea 👍🏻

PR #2096, opened 2022-11-28
[WIP] Separate `ReceiptParser` SPM Package

NachoSoto:
Fixes [CSDK-17].

This has some drawbacks and it's not as nice as having an isolated package, but it's A LOT simpler than the approach in #2062.

### What's new:
- New package can be built separately: `swift build --target ReceiptParser`
- The package can be used separately when integrating in Xcode:
Screenshot 2022-11-29 at 09 29 10
- New `target` and `scheme` allows building directly from Xcode:
Screenshot 2022-11-28 at 14 38 19
Screenshot 2022-11-28 at 14 35 56
- New CI job to ensure package aways builds

### TODO:
- [x] Figure out overlapping sources (_fixed it by creating a `symlink`_)
- [x] Add scheme to be able to build new package from Xcode
- [x] Extract `LoggerType`: #2098
- [x] Make types `public`
- [x] Document public methods
- [ ] Document new `ReceiptParser` Package exists
- [x] Add new API to `APITester`s
~~- [ ] Expose to Obj-C~~
- [x] Add Installation test
- [ ] Change installation test reference back to `main`
- [x] Verify `Carthage` integration still works
- [x] Verify `release-checks` all pass


LocalReceiptParsing
```diff
@@ -0,0 +1 @@
+Sources/LocalReceiptParsing
```
NachoSoto, line 1: Symlink hack FTW.

aboedo: Do docC docs continue to render well after this change? Asking because I ran into some issues with folder structure before and that's why I ended up having a `Sources` folder in the first place (although it also happened to be a good thing)

aboedo:
nice work on this! Perhaps for the sake of shipping and iterating, we could split this into one PR for a public API to get the parsed receipt + one for the SPM and target changes?

NachoSoto:
> we could split this into one PR for a public API to get the parsed receipt + one for the SPM and target changes?

Yeah that's my plan 👍🏻

LocalReceiptParsing
NachoSoto: Docs are WIP, but yeah:
Screenshot 2022-12-14 at 11 10 04

At least these docs are generated correctly for the `Purchases` package. Not sure about the `ReceiptParser` package but if this doesn't work I'd be out of ideas :/

NachoSoto:
Reopened this as #2154 and #2155.

PR #2099, opened 2022-11-29
`ReceiptParser.parse` always throws `ReceiptParser.Error`

NachoSoto:
This will be used for [CSDK-17]. 

Since this will be a public API, we want to make sure we provide consistent and documented errors.
This changes a few thrown `BitShiftError`s into `ReceiptParser.Error`.

### DIFF
--- a/Sources/LocalReceiptParsing/DataConverters/UInt8+Extensions.swift
+++ b/Sources/LocalReceiptParsing/DataConverters/UInt8+Extensions.swift
@@ -43,19 +43,21 @@ extension BitShiftError: CustomStringConvertible {
 
 extension UInt8 {
 
+    /// - Throws: `BitShiftError`
     func bitAtIndex(_ index: UInt8) throws -> UInt8 {
         guard index <= 7 else { throw BitShiftError.invalidIndex(index) }
         let shifted = self >> (7 - index)
         return shifted & 0b1
     }
 
+    /// - Throws: `BitShiftError`
     func valueInRange(from: UInt8, to: UInt8) throws -> UInt8 {
         guard to <= 7 else { throw BitShiftError.invalidIndex(to) }
         guard from <= to else { throw BitShiftError.rangeFlipped(from: from, to: to) }
 
         let range: UInt8 = to - from + 1
         let shifted = self >> (7 - to)
-        let mask = try maskForRange(range)
+        let mask = try self.maskForRange(range)
         return shifted & mask
     }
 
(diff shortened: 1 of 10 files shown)


Sources/LocalReceiptParsing/ReceiptParsingError.swift
```diff
@@ -15,28 +15,28 @@
 
 import Foundation
 
-enum ReceiptReadingError: Error, Equatable {
+extension ReceiptParser {
 
-    case missingReceipt,
```
NachoSoto, line 20: This one was never used.

NachoSoto:
Hold on, the protocol needs to get rid of the optional parameters... https://twitter.com/nachosoto/status/1597683789005611008?s=61&t=U9GVG3DNVeKOdZfhsuJEKw

vegaro (approved):

Sources/LocalReceiptParsing/DataConverters/UInt8+Extensions.swift
```diff
@@ -43,26 +43,29 @@ extension BitShiftError: CustomStringConvertible {
 
 extension UInt8 {
 
+    /// - Throws: `BitShiftError`
     func bitAtIndex(_ index: UInt8) throws -> UInt8 {
         guard index <= 7 else { throw BitShiftError.invalidIndex(index) }
```
aboedo, line 48: should we have this one also throw a specific receiptParser.error?

Sources/LocalReceiptParsing/ReceiptParsingError.swift
aboedo: I suppose that's because the parser takes in a data object, so it doesn't allow you to call it with no receipt. 
For the public API I think that's something that we could abstract away, right? 
Meaning the fetching and / or refreshing of the receipt, so that the API allows you to either send it receipt data or just get the default receipt from device. Or maybe just the latter

aboedo (approved):
Looks good! left a question about potentially also including bitShiftError as one of the receipt parser errors

Sources/LocalReceiptParsing/ReceiptParsingError.swift
NachoSoto: Not sure what you mean.
As for why this was not used, it's because it's detected earlier and we create an `ErrorCode` instead.

Sources/LocalReceiptParsing/DataConverters/UInt8+Extensions.swift
NachoSoto: You mean instead of having this sub-error that gets wrapped later?
These are `UInt8` extensions, so they kind of need a specific error, `ReceiptParser.Error` wouldn't make sense.

PR #2100, opened 2022-11-29
`Tests`: ensure `Purchases` is not configured multiple times

NachoSoto:
Continuing to look into [CSDK-517], I found this several times in the logs:
> Purchases instance already set. Did you mean to configure two Purchases objects?

In order to ensure that tests don't leave around unwanted state, I've turned this error into an assertion _only when running tests_, as well as fixed those cases.


Sources/FoundationExtensions/UIApplication+RCExtensions.swift
```diff
@@ -40,14 +40,4 @@ extension UIApplication {
 
 }
 
-#if DEBUG
-
-private extension UIApplication {
```
vegaro, line 45: Why was this an extension on UIApplication before?

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -500,6 +500,11 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
     static func setDefaultInstance(_ purchases: Purchases) {
         self.purchases.modify { currentInstance in
             if currentInstance != nil {
+                #if DEBUG
+                if ProcessInfo.isRunningUnitTests {
+                    preconditionFailure(Strings.configure.purchase_instance_already_set.description)
+                }
+                #endif
```
aboedo, line 507: does that mean that any apps that have tests using our SDK will also crash here? If so we should find a way to limit this to just _our_ testing, a customer should be able to do whatever they want in their tests

NachoSoto: Oh that's a fair point. But considering this is undesired usage of the SDK, and crashing during tests has no user impact, isn't that a good thing?
Running the SDK concurrently can lead to the concurrent post receipt operations issue.

Sources/FoundationExtensions/UIApplication+RCExtensions.swift
NachoSoto: For no particular reason :P

PR #2102, opened 2022-11-29
Fixed `LogLevel` ordering and added tests

NachoSoto:
A last minute change in #2080 made `LogLevel.verbose` have a higher `rawValue` to maintain backwards compatibility.
Unfortunately, we we re relying on these to implement ordering.

This PR makes that explicit by making `LogLevel` conform to `Comparable`, and added tests to verify this behavior in `Logger`.


NachoSoto:
Didn't add a `fix` label to this because technically we never shipped this wrong on any release.

Tests/UnitTests/TestHelpers/TestLogHandler.swift
```diff
@@ -139,7 +139,7 @@ extension TestLogHandler: LogMessageObserver {
         }
     }
 
-    private static let messageLimit = 100
+    private static let messageLimit = 200
```
NachoSoto, line 142: With `LogLevel.verbose` logs now enabled we need this higher.

NachoSoto: Actually maybe not
```suggestion
    private static let messageLimit = 100
```

PR #1057, opened 2021-12-10
Automate generating `RevenueCat-Swift.h` for SPM

NachoSoto:
- Fixes [sc-11746]
- See https://github.com/RevenueCat/purchases-ios/blob/362f023bf06f11efdf5b45db9d8f133c47e499d6/docs/V4_API_Updates.md#objc--spm
- Bug report: https://bugs.swift.org/browse/SR-15154

PR #2103, opened 2022-11-29
`CircleCI`: cache Homebrew installation

NachoSoto:
We currently spend almost 3 minutes on every build installing `swiftlint` and `xcbeautify`, so this is worth optimizing.

As brought up in #2094, the cache key should have some sort of hash of what's installed.
However, we don't have a way to list the dependencies. For the time being, this should be enough. We can bump the "version" number to invalidate the cache whenever we need it to update.

This makes installing `SwiftLint` go from almost 3 minutes to just a few seconds (around 5 between restoring cache and `brew link`).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -106,12 +106,35 @@ commands:
     steps:
       - install-bundle-dependencies:
           directory: << parameters.directory >>
+      - restore_cache:
+          keys:
+            - homebrew-cache-v4
+      - install-brew-dependency:
+          dependency_name: 'xcbeautify'
+      - install-brew-dependency:
+          dependency_name: 'swiftlint'
+      - save_cache:
+          key: homebrew-cache-v4
+          paths:
+            - /usr/local/Cellar/swiftlint/
+            - /usr/local/Cellar/xcbeautify/
+            - /Users/$USER/Library/Caches/Homebrew/
+  
+  install-brew-dependency:
+    parameters:
+      dependency_name:
+        type: string
+    steps:
       - run:
-          name: Install swiftlint
-          command: brew install swiftlint
-      - run:
-          name: Install xcbeautify
-          command: brew install xcbeautify
+          name: Install << parameters.dependency_name >>
+          command: |
+            if which << parameters.dependency_name >> > /dev/null 2>&1; then
+                echo "Skipping installation, already found."
+                exit 0;
+            fi
+            brew install << parameters.dependency_name >>
+          environment:
+            HOMEBREW_NO_INSTALL_CLEANUP: 1
 
   install-rubydocker-dependencies:
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -128,6 +128,9 @@ commands:
       - run:
           name: Install << parameters.dependency_name >>
           command: |
+            # Link dependency in case it was found in the cache
+            brew link << parameters.dependency_name >> || true
+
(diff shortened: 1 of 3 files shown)


NachoSoto:
Ugh what's slow is the `brew` autoupdate (`update`) behavior.
Getting rid of that leads to an old version of SwiftLint unfortunately, so not much we can do.

.circleci/config.yml
```diff
@@ -106,12 +106,38 @@ commands:
...
+            # Link dependency in case it was found in the cache
+            brew link << parameters.dependency_name >> || true
+
+            if which << parameters.dependency_name >> > /dev/null 2>&1; then
+                echo "Skipping installation, already found."
+                exit 0;
+            fi
+            brew install << parameters.dependency_name >>
+          environment:
+            HOMEBREW_NO_INSTALL_CLEANUP: 1
```
NachoSoto, line 140: This is unnecessary in CI and it also saves some time.

.circleci/config.yml
```diff
@@ -106,12 +106,38 @@ commands:
...
-          command: brew install swiftlint
-      - run:
-          name: Install xcbeautify
-          command: brew install xcbeautify
+          name: Install << parameters.dependency_name >>
+          command: |
+            # Link dependency in case it was found in the cache
+            brew link << parameters.dependency_name >> || true
+
+            if which << parameters.dependency_name >> > /dev/null 2>&1; then
```
NachoSoto, line 134: Avoid calling `brew install` which will trigger a `brew update`.
Makes this go from 2 and a half minutes to 1 second.

.circleci/config.yml
```diff
@@ -106,12 +106,38 @@ commands:
...
+                echo "Skipping installation, already found."
+                exit 0;
+            fi
```
joshdholtz, line 137: What happens here if a dependency needs to be updated? It seems like if it's found it will just skip and not potentially update? Do we have to change the cache keys?

NachoSoto: Yeah, we'll just bump the cache key :)

joshdholtz: Could I suggest we try using a Homebrew Bundle? 🤷‍♂️ 

👉  https://github.com/Homebrew/homebrew-bundle

It creates a lock file that we could use the hash of the lock file as the cache key? And then we can update the lock file locally for new versions

NachoSoto: That doesn't allow specifying versions though?
> Homebrew does not support installing specific versions of a library, only the most recent one, so there is no good mechanism for storing installed versions in a .lock file.

joshdholtz: This is what _fastlane_'s lock file looks like - https://github.com/fastlane/fastlane/blob/master/fastlane/swift/formatting/Brewfile.lock.json

That quote is very confusing to me now 🙈

NachoSoto: Oh ok let me try. We can keep the rest of this PR the same, just use that as a hash 👍🏻

NachoSoto: Actually I just found [this](https://github.com/Homebrew/homebrew-bundle/issues/1057):
> Should we commit Brewfile.lock.json to git? #1057
> No, not really, unfortunately.

NachoSoto: Meh ignore that, I think having it will help a lot of our purpose here 👍🏻

NachoSoto: @joshdholtz done.

aboedo (approved):

Issue #2107, opened 2022-11-30
Processing under the main thread hangs the UI


jesus-mg-ios:
- [X] I have updated Purchases SDK to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Development/CONTRIBUTING.md)
- [X] I have searched the [Community](https://community.revenuecat.com)
- [X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [X] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
I'm investigating another bug, but trying some recommendations from yours, like using sk1 and enabling log level to debug, I find that the application crashes by memory or hangs the UI.

1. Environment
   1. Platform: iOS
   2. SDK version: 4.14.3
   3. StoreKit 2 (_disabled with `useStoreKit2IfEnabled(false)`_) (Y/N): Which SDK uses by default (I think is SK1 right now)
   4. OS version:  x.x.x
   5. Xcode version: 14.1.0
   6. How widespread is the issue. Percentage of devices affected. We don't know because depends on transactions.
2. Debug logs:
[logsRevenuecat.zip](https://github.com/RevenueCat/purchases-ios/files/10120957/logsRevenuecat.zip)

3. Steps to reproduce:

The bug is reproduced when the SDK starts to process transactions. I think that part of the process is done inside of the Main actor, using the main thread for that. When It processes only a few transactions, all is working as expected, but when the number of transactions is high, the UI hangs, and memory increases exponentially.

4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

<img width="830" alt="image" src="https://user-images.githubusercontent.com/85997060/204742016-c84b49be-2b57-4a46-8e96-ee3264e0afdf.png">

I don't know if it makes sense, but maybe there's a way to use a background thread to process them or maybe create a global actor that uses other queues (custom if it's possible). Also, processing a great number of transactions could take time, maybe there's a way to process a prefix or suffix of 50, ordering them by date && success result and don't communicate the rest of them directly to back doing batches.

**Additional context**

RCGitBot:
👀 [SDKONCALL-180] We've just linked this issue to our internal tracker and notified the team. Thank you for reporting, we're checking this out!

NachoSoto:
[...]

PR #2110, opened 2022-11-30
`Configuration.with(appUserID:)`: allow passing `nil` and added new tests

NachoSoto:
The docs had this:
> Pass `nil` or an empty string if you want ``Purchases`` to generate this for you.

But `nil` wasn't a valid input.
I noticed [this reported issue](https://github.com/RevenueCat/purchases-ios/issues/2107#issuecomment-1332007011) where the user was doing this:
```swift
.with(appUserID: userUID ?? "")
```

Which technically would have the same semantics, since we allow empty and convert it to anonymous, but now this is more clear and it matches the docstring.

### DIFF
--- a/Sources/Identity/IdentityManager.swift
+++ b/Sources/Identity/IdentityManager.swift
@@ -70,11 +70,8 @@ class IdentityManager: CurrentUserProvider {
     }
 
     var currentUserIsAnonymous: Bool {
-
-        let anonymousFoundRange = currentAppUserID.range(of: IdentityManager.anonymousRegex,
-                                                         options: .regularExpression)
-        let currentAppUserIDLooksAnonymous = anonymousFoundRange != nil
-        let isLegacyAnonymousAppUserID = currentAppUserID == deviceCache.cachedLegacyAppUserID
+        let currentAppUserIDLooksAnonymous = Self.userIsAnonymous(self.currentAppUserID)
+        let isLegacyAnonymousAppUserID = self.currentAppUserID == deviceCache.cachedLegacyAppUserID
 
         return currentAppUserIDLooksAnonymous || isLegacyAnonymousAppUserID
     }
@@ -93,6 +90,16 @@ class IdentityManager: CurrentUserProvider {
 
     static func generateRandomID() -> String {
         "$RCAnonymousID:\(UUID().uuidString.replacingOccurrences(of: "-", with: "").lowercased())"
+    }
+
+}
+
+extension IdentityManager {
+
+    static func userIsAnonymous(_ appUserId: String) -> Bool {
+        let anonymousFoundRange = appUserId.range(of: IdentityManager.anonymousRegex,
+                                                  options: .regularExpression)
+        return anonymousFoundRange != nil
     }
 
 }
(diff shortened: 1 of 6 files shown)


Tests/UnitTests/Purchasing/Purchases/PurchasesConfiguringTests.swift
```diff
@@ -92,6 +92,34 @@ class PurchasesConfiguringTests: BasePurchasesTests {
...
+        expect(purchases.appUserID) == Self.appUserID
+    }
+
+    func testUserIdIsSetToAnonymousWhenConfiguringWithEmptyUserID() {
+        let purchases = Purchases.configure(
+            with: .init(withAPIKey: "")
+                .with(appUserID: "")
+        )
+        expect(purchases.appUserID).toNot(beEmpty())
+        expect(IdentityManager.userIsAnonymous(purchases.appUserID))
```
NachoSoto, line 109: Weirdly this fails because `DeviceCache` has a previously set `user ID`.
Is that expected? Shouldn’t the SDK “log out” when initializing with nil even if previously there was a user?

unxavi:
@NachoSoto https://github.com/RevenueCat/purchases-ios/issues/1717 this has been reported before, but the logout issue that you said sounds interesting

Tests/UnitTests/Purchasing/Purchases/PurchasesConfiguringTests.swift
```diff
@@ -92,6 +92,68 @@ class PurchasesConfiguringTests: BasePurchasesTests {
...
+        let purchases = Purchases.configure(
+            with: .init(withAPIKey: "")
+                .with(userDefaults: userDefaults)
+                .with(appUserID: newUserID)
+        )
+
+        expect(purchases.appUserID) == newUserID
+    }
+
+    func testNilUserIdIsIgnoredIfPreviousUserExists() {
```
NachoSoto, line 138: @vegaro @aboedo this is what I was talking about. To me this seems like a bug, but I wanted to at least document the current behavior.

Sources/Identity/IdentityManager.swift
```diff
@@ -32,7 +32,7 @@ class IdentityManager: CurrentUserProvider {
     private let customerInfoManager: CustomerInfoManager
     private let attributeSyncing: AttributeSyncing
 
-    internal static let anonymousRegex = #"\$RCAnonymousID:([a-z0-9]{32})$"#
+    private static let anonymousRegex = #"\$RCAnonymousID:([a-z0-9]{32})$"#
```
NachoSoto, line 35: This is `private` now 🎉

Tests/UnitTests/Purchasing/Purchases/PurchasesConfiguringTests.swift
vegaro: This is the same as in Android so I think it was designed like this, but I am not sure. @aboedo might know better since he worked in Identity v2. I think it's to make cases like the following work better:

- configure with nil -> anonymous is generated
- login with for example a Firebase ID
- configure with nil -> it will reuse Firebase ID

vegaro (approved):

Tests/UnitTests/Purchasing/Purchases/PurchasesConfiguringTests.swift
NachoSoto: Cool 👍🏻 
Well at least we have a test for this now.

PR #2111, opened 2022-11-30
`IdentityManager`: extracted `anonymousPrefix`

NachoSoto:
This was duplicated.

### DIFF
--- a/Sources/Identity/IdentityManager.swift
+++ b/Sources/Identity/IdentityManager.swift
@@ -32,7 +32,8 @@ class IdentityManager: CurrentUserProvider {
     private let customerInfoManager: CustomerInfoManager
     private let attributeSyncing: AttributeSyncing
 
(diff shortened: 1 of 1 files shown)


Sources/Identity/IdentityManager.swift
```diff
@@ -32,7 +32,8 @@ class IdentityManager: CurrentUserProvider {
     private let customerInfoManager: CustomerInfoManager
     private let attributeSyncing: AttributeSyncing
 
-    internal static let anonymousRegex = #"\$RCAnonymousID:([a-z0-9]{32})$"#
+    internal static let anonymousPrefix = "$RCAnonymousID:"
+    internal static let anonymousRegex = #"\#(IdentityManager.anonymousPrefix)([a-z0-9]{32})$"#
```
NachoSoto, line 36: Never mind, this isn't quite right so it's not worth extracting.

PR #2113, opened 2022-12-02
Fix documentation typo

BasThomas:
### Checklist

### Motivation

Just a documentation header typo :)

### Description

`ccurrency` -> `currency`


vegaro (approved):

aboedo:
Thanks for doing this!

Merging since `danger` won't work because it's coming from a fork, but we do have the required tag `docs`, so we should be good to go.

PR #2115, opened 2022-12-05
`StoreKit1Wrapper`: process transactions in a background thread

### LINKED ISSUE #2107: Processing under the main thread hangs the UI

NachoSoto:
Fixes #2107 and [SDKONCALL-180]. That's a rare issue where a user ends up with thousands of transactions.
That exposed the fact that we're processing all of these transactions on the main thread. This fixes that.

[SDKONCALL-180]: https://revenuecats.atlassian.net/browse/SDKONCALL-180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/StoreKit1/StoreKit1Wrapper.swift
+++ b/Sources/Purchasing/StoreKit1/StoreKit1Wrapper.swift
@@ -63,17 +63,14 @@ class StoreKit1Wrapper: NSObject {
     private let paymentQueue: SKPaymentQueue
     private let operationDispatcher: OperationDispatcher
 
-    init(paymentQueue: SKPaymentQueue, operationDispatcher: OperationDispatcher) {
+    init(paymentQueue: SKPaymentQueue = .default(),
+         operationDispatcher: OperationDispatcher = .default) {
         self.paymentQueue = paymentQueue
         self.operationDispatcher = operationDispatcher
 
         super.init()
 
         Logger.verbose(Strings.purchase.storekit1_wrapper_init(self))
-    }
-
-    override convenience init() {
-        self.init(paymentQueue: .default(), operationDispatcher: .default)
     }
 
     deinit {
--- a/Sources/Purchasing/StoreKit1/StoreKit1Wrapper.swift
+++ b/Sources/Purchasing/StoreKit1/StoreKit1Wrapper.swift
@@ -61,9 +61,11 @@ class StoreKit1Wrapper: NSObject {
     }
 
     private let paymentQueue: SKPaymentQueue
+    private let operationDispatcher: OperationDispatcher
 
-    init(paymentQueue: SKPaymentQueue) {
+    init(paymentQueue: SKPaymentQueue, operationDispatcher: OperationDispatcher) {
         self.paymentQueue = paymentQueue
+        self.operationDispatcher = operationDispatcher
 
         super.init()
 
@@ -71,7 +73,7 @@ class StoreKit1Wrapper: NSObject {
     }
 
     override convenience init() {
-        self.init(paymentQueue: .default())
+        self.init(paymentQueue: .default(), operationDispatcher: .default)
     }
 
     deinit {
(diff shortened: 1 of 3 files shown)


Sources/Purchasing/StoreKit1/StoreKit1Wrapper.swift
```diff
@@ -61,17 +61,19 @@ class StoreKit1Wrapper: NSObject {
     }
 
     private let paymentQueue: SKPaymentQueue
+    private let operationDispatcher: OperationDispatcher
 
-    init(paymentQueue: SKPaymentQueue) {
+    init(paymentQueue: SKPaymentQueue, operationDispatcher: OperationDispatcher) {
```
NachoSoto, line 66: I need to pass the same one down from `Purchases` here.

Tests/UnitTests/Purchasing/StoreKit1WrapperTests.swift
```diff
@@ -253,3 +273,14 @@ class StoreKit1WrapperTests: TestCase, StoreKit1WrapperDelegate {
 #endif
 
 }
+
+private extension StoreKit1WrapperTests {
+
+    static func transaction(with: SKPayment) -> MockTransaction {
+        let transaction = MockTransaction()
+        transaction.mockPayment = SKPayment.init(product: SK1Product())
```
NachoSoto, line 281:
```suggestion
        transaction.mockPayment = SKPayment(product: SK1Product())
```

Tests/UnitTests/Purchasing/StoreKit1WrapperTests.swift
```diff
@@ -87,6 +93,20 @@ class StoreKit1WrapperTests: TestCase, StoreKit1WrapperDelegate {
...
+    func testCallsDelegateToProcessTransactionsOnWorkerThread() {
+        let payment = SKPayment(product: SK1Product())
+        self.wrapper?.add(payment)
+
+        let transactions = [
+            Self.transaction(with: payment),
+            Self.transaction(with: payment)
+        ]
+
+        self.wrapper?.paymentQueue(paymentQueue, updatedTransactions: transactions)
```
NachoSoto, line 105:
```suggestion
        self.wrapper.paymentQueue(paymentQueue, updatedTransactions: transactions)
```

Tests/UnitTests/Purchasing/StoreKit1WrapperTests.swift
```diff
@@ -87,6 +93,20 @@ class StoreKit1WrapperTests: TestCase, StoreKit1WrapperDelegate {
         expect(self.updatedTransactions).to(contain(transaction))
     }
 
+    func testCallsDelegateToProcessTransactionsOnWorkerThread() {
+        let payment = SKPayment(product: SK1Product())
+        self.wrapper?.add(payment)
```
NachoSoto, line 98:
```suggestion
        self.wrapper.add(payment)
```

aboedo (approved):
makes sense, good catch

PR #2117, opened 2022-12-05
`StoreKit1Wrapper`: added warning if receiving too many updated transactions

NachoSoto:
See #2107. We still don't know what caused that, but it leads to degraded performance (mildly reduced by #2115). To advice of that, in case users are wondering why the app might potentially take excessive CPU usage, I added this warning.

### DIFF
--- a/Sources/Logging/Strings/StoreKitStrings.swift
+++ b/Sources/Logging/Strings/StoreKitStrings.swift
@@ -42,6 +42,8 @@ enum StoreKitStrings {
 
     case no_cached_products_starting_store_products_request(identifiers: Set<String>)
 
+    case sk1_payment_queue_too_many_transactions(_ count: Int)
+
     case sk1_product_request_too_slow
 
     case sk2_product_request_too_slow
@@ -96,6 +98,10 @@ extension StoreKitStrings: CustomStringConvertible {
 
         case .no_cached_products_starting_store_products_request(let identifiers):
             return "No existing products cached, starting store products request for: \(identifiers)"
+
+        case let .sk1_payment_queue_too_many_transactions(count):
+            return "SKPaymentQueue sent \(count) updated transactions. " +
+            "This high number is unexpected and is likely due to bad state on your device."
 
         case .sk1_product_request_too_slow:
             return "StoreKit 1 product request took longer than expected"
--- a/Sources/Logging/Strings/StoreKitStrings.swift
+++ b/Sources/Logging/Strings/StoreKitStrings.swift
@@ -42,7 +42,7 @@ enum StoreKitStrings {
 
     case no_cached_products_starting_store_products_request(identifiers: Set<String>)
 
-    case sk1_payment_queue_too_many_transactions(_ count: Int)
+    case sk1_payment_queue_too_many_transactions(count: Int, isSandbox: Bool)
 
     case sk1_product_request_too_slow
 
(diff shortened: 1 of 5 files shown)


Sources/Logging/Strings/StoreKitStrings.swift
```diff
@@ -97,6 +99,10 @@ extension StoreKitStrings: CustomStringConvertible {
         case .no_cached_products_starting_store_products_request(let identifiers):
             return "No existing products cached, starting store products request for: \(identifiers)"
 
+        case let .sk1_payment_queue_too_many_transactions(count):
+            return "SKPaymentQueue sent \(count) updated transactions. " +
+            "This high number is unexpected and is likely due to bad state on your device."
```
aboedo, line 104: 🤔 
I mean... this would probably be the sandbox account, right? like, if you use it a lot for tests, it's gonna just grow and grow. 
Apple updated so they only keep transactions for 90 days now, but if you open up an old sandbox account on a new device, I wouldn't be surprised if you got a ton of transactions. 

I would mostly recommend using a different sandbox account here

aboedo: for example if I test with one account a ton, then go to a new device and log in with that account, the new device will get all of those transactions coming in at once

Sources/Purchasing/StoreKit1/StoreKit1Wrapper.swift
```diff
@@ -133,6 +133,10 @@ extension StoreKit1Wrapper: SKPaymentTransactionObserver {
     func paymentQueue(_ queue: SKPaymentQueue, updatedTransactions transactions: [SKPaymentTransaction]) {
         guard let delegate = self.delegate else { return }
 
+        if transactions.count >= 100 {
```
aboedo, line 136: let's extract 100 here into a constant so it's not a magic number in the middle of the code

aboedo (changes requested):

Sources/Logging/Strings/StoreKitStrings.swift
NachoSoto: > I would mostly recommend using a different sandbox account here

That makes sense 👍🏻 I'll update the message to point to that.

NachoSoto:
@aboedo updated.

Sources/Logging/Strings/StoreKitStrings.swift
aboedo: that looks great! One final nitpick: can we ensure that _that_ message is used only in sandbox, and have a different one for production? 
The production one could say something about there just being a very high number of transactions. 
I wouldn't necessarily call it _unexpected_, I can imagine that for a few apps they might actually have people with a ton of transactions (2 years of a weekly subscription is already over 100 renewals). I'd maybe just say something about there being a lot of transactions, so performance might be impacted

NachoSoto: Yup, good idea. I thought about it but I thought YAGNI because I had to inject `SystemInfo` to `StoreKit1Wrapper`. But I agree it's useful.

NachoSoto: @aboedo updated.

Sources/Logging/Strings/StoreKitStrings.swift
```diff
@@ -97,6 +99,14 @@ extension StoreKitStrings: CustomStringConvertible {
         case .no_cached_products_starting_store_products_request(let identifiers):
             return "No existing products cached, starting store products request for: \(identifiers)"
 
+        case let .sk1_payment_queue_too_many_transactions(count, isSandbox):
+            let messageSuffix = isSandbox
+            ? "This high number is unexpected and is likely due to using an old sandbox account on a new device. " +
+            "If this is impacting performance, using a new sandbox account is recommended."
+            : "This is a very high number and might impact performance."
+
+            return "SKPaymentQueue sent \(count) updated transactions. " + messageSuffix
```
aboedo, line 108: :chefskiss:

swrobel:
Out of curiosity, why wasn't this implemented for StoreKit 2?

NachoSoto:
[StoreKit 2 uses `StoreKit.Transaction.updates`](https://github.com/RevenueCat/purchases-ios/blob/4.16.0/Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift#L43), which is an `AsyncStream`. There's no easy way to detect the number of elements that will be sent through that stream.

PR #2118, opened 2022-12-05
`HTTPClient`: replaced `X-StoreKit2-Setting` with `X-StoreKit2-Enabled`

NachoSoto:
This will be used for https://github.com/RevenueCat/khepri/pull/4843, to fix [SDKONCALL-160] and #2020.

`X-StoreKit2-Setting` only told us if the apps had that setting enabled. However, when running on an older device, `SK2` would still be disabled.

[SDKONCALL-160]: https://revenuecats.atlassian.net/browse/SDKONCALL-160?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Misc/StoreKit2Setting.swift
+++ b/Sources/Misc/StoreKit2Setting.swift
@@ -60,6 +60,11 @@ extension StoreKit2Setting {
 
     /// - Returns: `true` iff SK2 is enabled and it's available.
     var shouldOnlyUseStoreKit2: Bool {
+        return self.isEnabledAndAvailable
+    }
+
+    /// - Returns: `true` iff SK2 is enabled and it's available.
+    var isEnabledAndAvailable: Bool {
         switch self {
         case .disabled, .enabledOnlyForOptimizations: return false
         case .enabledForCompatibleDevices: return Self.isStoreKit2Available
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -148,6 +148,7 @@ private extension HTTPClient {
             "X-Client-Build-Version": SystemInfo.buildVersion,
             "X-Client-Bundle-ID": SystemInfo.bundleIdentifier,
             "X-StoreKit2-Setting": "\(self.systemInfo.storeKit2Setting.debugDescription)",
+            "X-StoreKit2-Enabled": "\(self.systemInfo.storeKit2Setting.isEnabledAndAvailable)",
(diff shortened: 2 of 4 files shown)


Sources/Misc/StoreKit2Setting.swift
```diff
@@ -60,6 +60,11 @@ extension StoreKit2Setting {
 
     /// - Returns: `true` iff SK2 is enabled and it's available.
     var shouldOnlyUseStoreKit2: Bool {
```
NachoSoto, line 62: I decided to keep this other property too because it has slightly different implications, even if it has the same implementation (right now).
This is used to determine whether to _not_ enable SK1 stuff (see #1882 for example).

Sources/Misc/StoreKit2Setting.swift
```diff
@@ -60,6 +60,11 @@ extension StoreKit2Setting {
...
+        return self.isEnabledAndAvailable
+    }
+
+    /// - Returns: `true` iff SK2 is enabled and it's available.
```
vegaro, line 66:
```suggestion
    /// - Returns: `true` if SK2 is enabled and it's available.
```

Sources/Misc/StoreKit2Setting.swift
```diff
@@ -60,6 +60,11 @@ extension StoreKit2Setting {
 
     /// - Returns: `true` iff SK2 is enabled and it's available.
```
vegaro, line 61:
```suggestion
    /// - Returns: `true` if SK2 is enabled and it's available.
```

vegaro (approved):

NachoSoto:
Yeah I wish we had done this originally :/ I feel like long term we should deprecate the other one. If anything change in iOS I think the SDK is better suited to determine what's actually running and it seems scary to put that logic in the backend.

In the short term I agree that we should do that to fox promo purchases for existing users though. Since we're gonna do that, should we just close this?

NachoSoto:
Oh but as we discussed we need a client-side fix anyway for base64, so maybe this would be less risky that comparing version numbers?

aboedo:
if it ended up being impossible to fix older versions through a backend patch, then I'd vote for killing the old header in favor of the new one, I don't think it's _currently_ used for anything

NachoSoto:
Okay updated this to replace the header instead.

aboedo (approved):

PR #2124, opened 2022-12-09
Added a few performance improvements for `ReceiptParser`

NachoSoto:
See also #2123 and https://github.com/RevenueCat/purchases-ios/issues/2107#issuecomment-1342846583.

### DIFF
--- a/Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
+++ b/Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
@@ -151,7 +151,10 @@ private extension ASN1ContainerBuilder {
 
         if lengthDefinition == .indefinite {
             innerContainers = try buildInternalContainers(payload: data.dropFirst(bytesUsedForLength))
-            let innerContainersOverallLength = innerContainers.map { $0.totalBytesUsed }.reduce(0, +)
+            let innerContainersOverallLength = innerContainers
+                .lazy
+                .map { $0.totalBytesUsed }
+                .reduce(0, +)
             lengthValue = innerContainersOverallLength
         } else if isConstructed {
             let innerContainerData = data.dropFirst(bytesUsedForLength).prefix(lengthValue)
--- a/Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
+++ b/Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
@@ -152,7 +152,7 @@ private extension ASN1ContainerBuilder {
         if lengthDefinition == .indefinite {
             innerContainers = try buildInternalContainers(payload: data.dropFirst(bytesUsedForLength))
             let innerContainersOverallLength = innerContainers
-                .lazy
+                .lazy // Avoid creating intermediate arrays
                 .map { $0.totalBytesUsed }
                 .reduce(0, +)
             lengthValue = innerContainersOverallLength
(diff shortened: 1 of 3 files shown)


Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
```diff
@@ -151,7 +151,10 @@ private extension ASN1ContainerBuilder {
 
         if lengthDefinition == .indefinite {
             innerContainers = try buildInternalContainers(payload: data.dropFirst(bytesUsedForLength))
-            let innerContainersOverallLength = innerContainers.map { $0.totalBytesUsed }.reduce(0, +)
+            let innerContainersOverallLength = innerContainers
+                .lazy
```
NachoSoto, line 155: This avoids creating an array of N elements just to calculate the sum.

Sources/LocalReceiptParsing/DataConverters/UInt8+Extensions.swift
```diff
@@ -68,18 +68,8 @@ private extension UInt8 {
...
-        case 4: return 0b1111
-        case 5: return 0b11111
-        case 6: return 0b111111
-        case 7: return 0b1111111
-        case 8: return 0b11111111
-        default:
-            throw BitShiftError.unhandledRange
-        }
+
+        return 0b11111111 >> (8 - range)
```
NachoSoto, line 72: One liner, but also avoids jumps. I don't think the optimizer could have figured this out.

Sources/LocalReceiptParsing/ReceiptParser.swift
```diff
@@ -38,9 +38,7 @@ class ReceiptParser {
     }
 
     func parse(from receiptData: Data) throws -> AppleReceipt {
-        let intData = [UInt8](receiptData)
-
-        let asn1Container = try self.containerBuilder.build(fromPayload: ArraySlice(intData))
+        let asn1Container = try self.containerBuilder.build(fromPayload: ArraySlice(receiptData))
```
NachoSoto, line 41: I think this doesn't actually improve performance thanks to Swift's COW behavior, but it cleans up an unnecessary value.

Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
vegaro: I didn't know about this lazy thing. For whoever is interested this article explains it very well https://www.avanderlee.com/swift/lazy-collections-arrays/

Sources/LocalReceiptParsing/DataConverters/UInt8+Extensions.swift
vegaro: I like it, but it might be worth it to add a couple examples for clarification?

```
// If range 1, return `0b1`
// If range 2, return `0b11`
// ...
```

Sources/LocalReceiptParsing/ReceiptParser.swift
vegaro: I have no idea so I had to ask, does it matter we are creating a ArraySlice on a Data and we were passing a `[UInt8]` before?

Sources/LocalReceiptParsing/Builders/ASN1ContainerBuilder.swift
vegaro: Since this is not very well known, maybe adding a comment with a explanation would be worth it? Something like, `reduce` is applied without having to create an intermediate collection with `map`

NachoSoto: Done

Sources/LocalReceiptParsing/DataConverters/UInt8+Extensions.swift
NachoSoto: Done.

Sources/LocalReceiptParsing/ReceiptParser.swift
NachoSoto: `Data` is already a `Collection` of `UInt8` values, so this is equivalent to the previous implementation (and we have good test coverage to ensure that).

vegaro (approved):

PR #2136, opened 2022-12-12
`CallbackCache`: avoid exposing internal mutable cache

NachoSoto:
Small refactor to avoid users of this `class` change the content of the cache without using the methods.


vegaro (approved):
Makes sense!

PR #2138, opened 2022-12-12
`NetworkOperation`: added assertion to ensure that subclasses call completion

NachoSoto:
Looking into the leak in #2135, I realized that we have an operation which doesn't call `completion`, we would be leaking the completion block in the cache.
This ensures that it doesn't happen in tests. In production, we won't risk crashing.

### DIFF
--- a/Sources/Networking/Operations/NetworkOperation.swift
+++ b/Sources/Networking/Operations/NetworkOperation.swift
@@ -13,21 +13,41 @@
 
 import Foundation
 
-class CacheableNetworkOperation: NetworkOperation, CacheKeyProviding {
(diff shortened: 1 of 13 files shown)


Sources/Networking/Operations/NetworkOperation.swift
```diff
@@ -103,7 +106,25 @@ class NetworkOperation: Operation {
         super.init()
     }
 
+    deinit {
+        #if DEBUG
+        if ProcessInfo.isRunningRevenueCatTests {
+            precondition(
+                self.didStart,
```
NachoSoto, line 113: This depends on #2135.

tonidero (approved):

Issue #2132, opened 2022-12-11
Apple Watch Restore Purchases issue


Turacbey:
- [ *] I have updated Purchases SDK to the latest version
- [* ] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [ *] I have searched the [Community](https://community.revenuecat.com)
- [* ] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [* ] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

1. **Environment**
   1. Platform: WatchOS
   2. SDK version: 4.15.0 (the issue was happening on earlier versions)
   3. StoreKit 2 (_disabled with `useStoreKit2IfEnabled(false)`_) (Y/N): Y (it was happening when it was N)
   4. OS version: WatchOS 7-8-9 (nowadays we see MOSTLY watchOS 8-9)
   5. Xcode version: 14.0
   6. How widespread is the issue. Percentage of devices affected. A lot of (not known the exact number)
2. **[Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue:** NOT AVAILABLE 
3. **Steps to reproduce, with a description of expected vs. actual behavior**
When a users make a purchase using iPhone/iPad etc, they can't restore it on their Apple Watches. When they tap to Restore purchases button on their Apple Watches, they get "Receipt is missing" error.

**Extras:** We have been suffering this issue  more than 1 year and this problem has not been solved.

RCGitBot:
👀 [SDKONCALL-188] We've just linked this issue to our internal tracker and notified the team. Thank you for reporting, we're checking this out!

NachoSoto:
Also it would be extremely valuable for us to be able to help with this if you could attach debug logs.

Turacbey:
> Can you share how you're configuring the SDK on watchOS?

```
func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey : Any]? = nil) -> Bool {
        
 Purchases.configure(
            with: Configuration.Builder(withAPIKey: "appl_myKey")
[...]

PR #2140, opened 2022-12-12
`Purchase Tester`: added companion `watchOS` app

NachoSoto:
I added this in order to debug #2132.

### Changes:
- Extracted `Core` framework to be used by both apps
- Added new `watchOS` target
- Changed `fastlane match` to also create profiles for this new target
- Added a simple view in the watch app that displays a text if an entitlement is active, or a `restore` purchases button otherwise

This is just a starting point, we can add more to it in the future.

#### Note:
_Unfortunately I haven't been able to test this because of an Apple bug (`FB11873951`):_

incoming-722CC7FA-D970-4E23-AB88-50167248D98D

### DIFF
--- a/Tests/TestingApps/PurchaseTesterSwiftUI/Core/Constants.swift
+++ b/Tests/TestingApps/PurchaseTesterSwiftUI/Core/Constants.swift
@@ -11,12 +11,12 @@ import Foundation
  Configuration file for your app's RevenueCat settings.
  */
 
-struct Constants {
+public struct Constants {
 
     /*
      The API key for your app from the RevenueCat dashboard: https://app.revenuecat.com
      This is the default used by `ConfigurationView`
      */
-    static let apiKey = "REVENUECAT_API_KEY"
+    public static let apiKey = "REVENUECAT_API_KEY"
 
 }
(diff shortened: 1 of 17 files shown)


Tests/TestingApps/PurchaseTesterSwiftUI/Core/Constants.swift
```diff
@@ -11,12 +11,12 @@ import Foundation
...
 
-struct Constants {
+public struct Constants {
 
     /*
      The API key for your app from the RevenueCat dashboard: https://app.revenuecat.com
      This is the default used by `ConfigurationView`
      */
-    static let apiKey = "REVENUECAT_API_KEY"
+    public static let apiKey = "REVENUECAT_API_KEY"
```
tonidero, line 20: How will this work if using a dynamic API key?

tonidero (approved):
Code LGTM but should we hold off on this until that apple bug is fixed? I guess we can merge for now and revisit later. Maybe add a TODO somewhere indicating that this hasn't been fully tested?

NachoSoto:
This is just a starting point, like I said in the description we can keep improving it later.

Tests/TestingApps/PurchaseTesterSwiftUI/Core/Constants.swift
NachoSoto: Yeah we’ll need a way to either share the key from the main app or add the configuration screen on the watch.

PR #2148, opened 2022-12-13
Fixed race condition finishing `SK1` transactions

NachoSoto:
Fixes [CSDK-504].

`SKPaymentQueue.finishTransaction` is asynchronous, but we were assuming it wasn't.
This meant that when performing multiple purchases (like consumables) in quick succession, before the first transaction was finished, the second purchase was reusing the first transaction.

Example log:
```
[Purchases] - INFO: :moneybag: Finishing transaction ‘0’ for product ‘consumable.10_coins’
[Purchases] - INFO: :heart_eyes_cat::moneybag: Purchased product - ‘consumable.10_coins’
[Purchases] - INFO: :moneybag: Purchasing Product ‘consumable.10_coins’ from package in Offering ‘coins’
[Purchases] - DEBUG: :information_source: StoreKit1Wrapper (0x00006000008a9530) removedTransaction: consumable.10_coins 0 1
```

Notice how the "removed transaction" was notified _after_ we had already started the second purchase.

See also #2148 for an earlier fix for finishing transactions.

[CSDK-504]: https://revenuecats.atlassian.net/browse/CSDK-504?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/StoreKit1/PaymentQueueWrapper.swift
+++ b/Sources/Purchasing/StoreKit1/PaymentQueueWrapper.swift
@@ -31,7 +31,7 @@ protocol PaymentQueueWrapperDelegate: AnyObject, Sendable {
 @objc
 protocol PaymentQueueWrapperType: AnyObject {
 
-    func finishTransaction(_ transaction: SKPaymentTransaction)
+    func finishTransaction(_ transaction: SKPaymentTransaction, completion: @escaping () -> Void)
 
     #if os(iOS) || targetEnvironment(macCatalyst)
     @available(iOS 13.4, macCatalyst 13.4, *)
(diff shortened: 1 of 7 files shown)


Sources/Purchasing/StoreKit1/PaymentQueueWrapper.swift
```diff
@@ -77,8 +77,14 @@ class PaymentQueueWrapper: NSObject, PaymentQueueWrapperType {
         super.init()
     }
 
-    func finishTransaction(_ transaction: SKPaymentTransaction) {
+    func finishTransaction(_ transaction: SKPaymentTransaction, completion: @escaping () -> Void) {
+        // See `StoreKit1Wrapper.finishTransaction(:completion:)`.
+        // Technically this is a race condition, because `SKPaymentQueue.finishTransaction` is asynchronous
+        // In practice this method won't be used, because this class is only used in SK2 mode,
+        // and those transactions are finished through `SK2StoreTransaction`.
```
tonidero, line 84: Should we log an error if this is reached then? Also, not sure if we should finish transaction in this case, if we consider it an error... But I guess we should propagate the error somehow in that case.

tonidero (approved):
🙌

Sources/Purchasing/StoreKit1/PaymentQueueWrapper.swift
aboedo: having to have this here doesn't really say wonders about our class design, we might wanna start thinking about alternatives soon

NachoSoto: Yeah for sure. This is needed because we avoid using `StoreKit1Wrapper` when SK2 is enabled. But turned out we did need a payment queue wrapper for some things.
And the fact that `finishTransaction` needs a reference to _some_ wrapper, because SK1 transactions are finished through the queue, but SK2’s are finished directly with the transaction :/

So yeah I agree this is a smell, but due to those 2 limitations I can’t think of a way to avoid it.

PR #2161, opened 2022-12-15
`Fastfile`: changed `match` to `readonly`

NachoSoto:
We don't want CI jobs updating provisioning profiles.
If they're set up correctly, nothing should change when using `match` to run these jobs.

Also removed unnecessary `verbose` parameter.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -183,7 +183,7 @@ platform :ios do
 
   desc "archive"
   lane :archive do
-    match(type: "appstore")
+    match(type: "appstore", readonly: true)
     gym(export_method: "app-store")
   end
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -730,7 +730,14 @@ workflows:
           <<: *release-branches-and-main
       - release-checks:
           xcode_version: '14.1.0'
-          <<: *release-branches
+      - installation-tests-cocoapods:
```
NachoSoto, line 733: Will remove. Just making sure this PR works.

fastlane/Fastfile
```diff
@@ -491,7 +491,7 @@ platform :ios do
 
   desc "Build and deploy PurchaseTesterSwiftUI"
   lane :deploy_purchase_tester do
-    match(verbose: true, readonly: true)
```
NachoSoto, line 494: This was temporary for debugging.

vegaro (approved):

PR #2164, opened 2022-12-16
Fixes RevenueCat.xcframework.zip so that is compatible as a binaryTarget() in SPM.

### LINKED ISSUE #2163: RevenueCat.xcframework.zip not compatible with use in SPM binaryTarget()

**Describe the bug**
The `RevenueCat.xcframework.zip` that is produced for each release is not compatible for use as a binary target in swift package manager.  Using a pre-compiled xcframework can substantially cut down on build times for both development and CI environments and increase overall productivity.  For example one could define a `Package.swift` using a binary target and provide the correct checksum for that release:

```swift
let package = Package(
/// ....
  targets: [
    /// ...
    .binaryTarget(name: "RevenueCat",
                  url: "https://github.com/RevenueCat/purchases-ios/releases/download/4.15.4/RevenueCat.xcframework.zip",
                  checksum: "<the correct checksum>),
  ],
/// ...
)
```
SPM requires that the zip file contain only the framework and that the framework is available at the root level of the zip file per [Apple's documentation](https://developer.apple.com/documentation/xcode/distributing-binary-frameworks-as-swift-packages).  The current xcframework zip file contains a directory at the root named "xcframeworks" and then multiple sub-directories including the `RevenueCat.xcframework` below it as follows:

- (root)
  - xcframeworks
    - archives
    - RevenueCat.dSYMs
    - RevenueCat.xcframework

The directory structure, and the `archives` and `RevenueCat.dSYMs` seem to be artifacts of how the xcframework was built using `fastlane` and are certainly not needed in the final zip.  [#1118](https://github.com/RevenueCat/purchases-ios/pull/1118) already included the dSYMs info directly in the xcframework.

The fix is fairly straightforward which is to only include `RevenueCat.xcframework` directory in the final zip file so the structure would be as follows:

- (root)
  - RevenueCat.xcframework

This would then be compatible with both SPM and Carthage.  Note that I have not tested this with Carthage, but there are other projects that provide Carthage xcframework.zip files that are structured in this manner and compatible with SPM.

I already have a PR to fix the issue that I will submit shortly.

1. Environment
   1. Platform: iOS 16.x
   2. SDK version: 4.15.4
   3. StoreKit 2 (_disabled with `useStoreKit2IfEnabled(false)`_) (Y/N): N/A
   4. OS version: MacOS 13.1 (build environment)
   5. Xcode version: 14.2

scogeo:
### Checklist

### Motivation
This update changes the directory structure of the `RevenueCat.xcframework.zip` file so that is compatible for use with Swift Package Manager as a binary target.

Resolves #2163 

### Description
Only include the `RevenueCat.xcframework` in the final zip file.  Other directory are superfluous and not compatible with SPM.

#### Steps to verify.

Build the `RevenueCat.xcframework.zip` file using `bundle exec fastlane ios export_xcframework`.  Unzip the resulting `RevenueCat.xcframework.zip` file in a temporary directory and verify that `RevenueCat.xcframework` is the only contents.


NachoSoto (approved):
This is amazing, thank you so much!

I use pre-compiled binaries with Carthage myself Egidio is huge for reducing compile times, and had no idea it was this easy to do with SPM 🤯

aboedo:
@scogeo Oh wow I just read the issue and I hadn't seen the PR yet, this is amazing! We need to verify that it still works correctly with Carthage after the change, and then it's ready to 🚢 

We'll look into the separate repo approach, I didn't know about it!

aboedo:
We might want to consider vending this as a _different_ `.zip`, as in `RevenueCat_SPM.xcframework.zip`, just in case developers out there are using the current one and relying on its current folder structure in some automation

NachoSoto:
> We might want to consider vending this as a different .zip, as in RevenueCat_SPM.xcframework.zip

Actually I think that's a great idea. @scogeo would you be able to make that change?

scogeo:
> > We might want to consider vending this as a different .zip, as in RevenueCat_SPM.xcframework.zip
> 
> Actually I think that's a great idea. @scogeo would you be able to make that change?

@NachoSoto Yes, I can make that change.

However, I am a little wary as I'm not sure how Carthage downloads the xcframework.zip files.  I moved to SPM for all projects a while ago and haven't looked back. :)  But I do know from looking at other projects that build xcframeworks for carthage, that carthage does not care what the name of the file is.  I think it is likely that if you have two .xcframework.zip file that carthage will attempt to download both and probably create new issues.  Let me do a quick experiment or two with carthage in my forked repo and report back.

A final note, the change I am proposing is more consistent with how other 3rd party libraries structure their .xcframwork.zip file.  It really just seems to be an artifact of the using `fastlane` to build.  Most other 3rd party libraries are using `carthage` itself to build the xcframeworks.

scogeo:
I did a quick test of this PR with Carthage this evening.  So good news is this PR as it currently stands does work with carthage as expected.  Again, multiple 3rd party libraries ship their framework .zip files this way.

The proposed addition of adding another file in the release named `RevenueCat_SPM.xcframework.zip` does also work in carthage.  However, I think it is a bit of a fluke.  While only one version of `RevenueCat.xcframework` appears in the project's `Carthage/Build` directory, carthage appears to have downloaded both .zip files and copied them to the build directory.  So the last one to write the directory is the one that remains.  So it would result in potentially unreproducible build issues if there was every a discrepancy between the two zip files.  I can tell this by looking at Carthage's cache directory and can see it downloaded both .zip files.

Here's the Cartfile I used for this experiment and ran it with a trivial app that links with an initialized RevenueCat with an empty app id.

```
# Main repo
#github "revenuecat/purchases-ios" ~> 4.15.0

# With single unified RevenueCat.xcframework.zip (current PR)
#github "scogeo/purchases-ios" ~> 14.15.0

# With both original RevenueCat.xcframework.zip and RevenueCat_SPM.xcframework.zip in Release
github "scogeo/purchases-ios" ~> 14.16.0
```

I created two releases with a different major release version in my forked repo to test with.  I used different version numbers so I could inspect the carthage cache.  this is what it contains after running all repo options.

```
gscott@ragnarok:~/.../org.carthage.CarthageKit/binaries/purchases-ios
➜ pwd
/Users/gscott/Library/Caches/org.carthage.CarthageKit/binaries/purchases-ios

gscott@ragnarok:~/.../org.carthage.CarthageKit/binaries/purchases-ios
➜ ls -lR
total 0
drwxr-xr-x  3 gscott  staff   96 Dec 16 18:30 14.15.4/
drwxr-xr-x  4 gscott  staff  128 Dec 16 18:35 14.16.4/
drwxr-xr-x  3 gscott  staff   96 Dec 16 18:51 4.15.4/

./14.15.4:
total 97944
-rw-------  1 gscott  staff  50143876 Dec 16 18:30 ID<Asset>(rawValue: 88534491)-RevenueCat.xcframework.zip

./14.16.4:
total 459424
-rw-------  1 gscott  staff  175098967 Dec 16 18:35 ID<Asset>(rawValue: 88535242)-RevenueCat.xcframework.zip
-rw-------  1 gscott  staff   50143876 Dec 16 18:35 ID<Asset>(rawValue: 88535312)-RevenueCat_SPM.xcframework.zip

./4.15.4:
total 341992
-rw-------  1 gscott  staff  175098967 Dec 16 17:40 ID<Asset>(rawValue: 88361149)-RevenueCat.xcframework.zip
```

So I wouldn't recommend adding multiple xcframework.zip files in the release.

Anyway, I'll leave it up to you on how to proceed with the PR.  I think this provides maximum backward compatibility with carthage and forward compatibility with SPM binary targets.

scogeo:
Closing this as it has diverged from main quite a bit and I have an in-house solution to this as part of my build system.

PR #1275, opened 2022-02-09
Fixed `ErrorCode`'s domain in Swift

NachoSoto:
## Background
In version 3, `Objective-C` had the constant `RCPurchasesErrorDomain` (`Purchases.ErrorDomain` in Swift), manually defined in Objective-C.
This was also visible to Swift through the constant of the same name, and both had the value `RCPurchasesErrorDomain`.

In version 4, written in Swift, it's not possible to expose global constants to Objective-C. The way the Swift compiler exposes error domains is through the definition of the `Error` itself.
In our case, it's `ErrorCode`, which has an `@objc` name `RCPurchasesErrorCode`. The compiler appends the word "Domain" to that, creating the new `RCPurchasesErrorCodeDomain` which superseded `RCPurchasesErrorDomain`.

However, prior to this change, our implementation of `CustomNSError` was not overriding `errorDomain`, so the value of these constants was the default `RevenueCat.ErrorCode` which was a change (regression) from version 3.x.
Unfortunately because of a bug (https://bugs.swift.org/browse/SR-15841), this isn't enough to fix the domain when `ErrorCode` is converted to `NSError`. This PR includes a workaround, but unfortunately the workaround only provides the correct value to Swift.
`RCPurchasesErrorCodeDomain` will still be the "invalid" `RevenueCat.ErrorCode`, but this is better than nothing.

At least now we have an explicit domain value for Swift errors, and we're exposing a constant for them (`Purchases.ErrorDomain` which was already available in version 3, so it's back now).

## TLDR:
In version 4 we had lost `Purchases.ErrorDomain` (which is arguably still useful to Swift), and the value was the auto-generated `RevenueCat.ErrorCode` instead of the one in version 3 `RCPurchasesErrorDomain`.
Swift now has a new custom domain `RevenueCatErrorDomain` which matches Apple's domains like `NSCocoaErrorDomain`.

PR #2165, opened 2022-12-19
`ErrorUtils.purchasesError(withUntypedError:)`: handle `PublicError`s

NachoSoto:
See https://github.com/RevenueCat/purchases-hybrid-common/pull/278.
A valid error ended up as:
> failed: caught error: "Unknown error."

This change ensures that calling this method with an `NSError` does not return `ErrorCode.unknownError`, and instead we extract the correct error.

### DIFF
--- a/Sources/Error Handling/ErrorUtils.swift
+++ b/Sources/Error Handling/ErrorUtils.swift
@@ -296,6 +296,21 @@ enum ErrorUtils {
         withUntypedError error: Error,
         fileName: String = #fileID, functionName: String = #function, line: UInt = #line
     ) -> PurchasesError {
+        func handlePublicError(_ error: PublicError) -> PurchasesError {
+            if let errorCode = ErrorCode(rawValue: error.code) {
+                return .init(error: errorCode, userInfo: error.userInfo)
+            } else {
+                return createUnknownError()
+            }
+        }
+
+        func createUnknownError() -> PurchasesError {
+            ErrorUtils.unknownError(
+                error: error,
+                fileName: fileName, functionName: functionName, line: line
+            )
+        }
+
         if #available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *) {
             switch error {
             case let purchasesError as PurchasesError:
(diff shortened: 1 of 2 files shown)


Sources/Error Handling/ErrorUtils.swift
```diff
@@ -296,18 +296,32 @@ enum ErrorUtils {
         withUntypedError error: Error,
         fileName: String = #fileID, functionName: String = #function, line: UInt = #line
     ) -> PurchasesError {
+        func handlePublicError(_ error: PublicError) -> PurchasesError {
+            if let errorCode = ErrorCode(rawValue: error.code) {
+                return .init(error: errorCode, userInfo: error.userInfo)
```
vegaro, line 301: I might be wrong but when we convert this in `purchases-hybrid-common` https://github.com/RevenueCat/purchases-hybrid-common/blob/main/ios/PurchasesHybridCommon/PurchasesHybridCommon/ErrorContainer.swift#L23 the nsError.code code will be the backend's returned error code?

Sources/Error Handling/ErrorUtils.swift
```diff
@@ -320,11 +334,10 @@ enum ErrorUtils {
                 return backendError.asPurchasesError
             case let networkError as NetworkError:
                 return networkError.asPurchasesError
+            case let error as PublicError where error.domain == ErrorCode.errorDomain:
```
vegaro, line 337: This might be a stupid question but where do we define our `errorDomain`? I couldn't find it 😆

Sources/Error Handling/ErrorUtils.swift
NachoSoto: This isn’t changing the code though.
This method is used because Swift doesn’t have typed throws, so we need a way to convert from `Error` to a typed `PurchasesError`. But this method won’t change the error at all. You can see that in the tests.

Sources/Error Handling/ErrorUtils.swift
NachoSoto: Not a stupid question 😅 It’s Swift magic for `Swift.Error` conformances.
See also #1275.

vegaro (approved):

PR #2166, opened 2022-12-19
`DNSChecker`: simplified `NetworkError` initialization

NachoSoto:
`DNSChecker.errorWithBlockedHostFromError` already checks `isBlockedAPIError` when constructing this error, so we can simplify this to avoid calling `DNSChecker.resolvedHost` twice.

### DIFF
--- a/Sources/Networking/HTTPClient/DNSChecker.swift
+++ b/Sources/Networking/HTTPClient/DNSChecker.swift
@@ -45,18 +45,18 @@ enum DNSChecker: DNSCheckerType {
     }
 
     static func errorWithBlockedHostFromError(_ error: Error?) -> NetworkError? {
-        guard isBlockedAPIError(error),
+        guard self.isBlockedAPIError(error),
               let nsError = error as NSError?,
               let failedURL = nsError.userInfo[NSURLErrorFailingURLErrorKey] as? URL else {
             return nil
         }
 
-        let host = resolvedHost(fromURL: failedURL)
+        let host = self.resolvedHost(fromURL: failedURL)
         return .dnsError(failedURL: failedURL, resolvedHost: host)
     }
 
     static func isBlockedURL(_ url: URL) -> Bool {
-        guard let resolvedHostName = resolvedHost(fromURL: url) else {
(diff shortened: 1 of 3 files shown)


Sources/Networking/HTTPClient/DNSChecker.swift
```diff
@@ -45,18 +45,18 @@ enum DNSChecker: DNSCheckerType {
...
               let failedURL = nsError.userInfo[NSURLErrorFailingURLErrorKey] as? URL else {
             return nil
         }
 
-        let host = resolvedHost(fromURL: failedURL)
+        let host = self.resolvedHost(fromURL: failedURL)
         return .dnsError(failedURL: failedURL, resolvedHost: host)
     }
 
     static func isBlockedURL(_ url: URL) -> Bool {
```
tonidero, line 58: Should we make this private now? I believe it's not called outside this class other than it tests and we can test the public API for that.

tonidero (approved):
A suggestion but looks good.

Sources/Networking/HTTPClient/DNSChecker.swift
NachoSoto: Oh yeah let me check.

NachoSoto: We have specific tests for this method, which I think is useful.

tonidero: Well, I try to avoid making things public only for tests. I think if it needs to be tested because the logic is too complex and the public API of the class does too many things, it should be extracted to a different class where it can be tested better. This doesn't seem to be that complex of a logic though, so I don't think it warrants moving to a different class.

That said, I don't think this should block the PR.

aboedo: > Well, I try to avoid making things public only for tests. I think if it needs to be tested because the logic is too complex and the public API of the class does too many things, it should be extracted to a different class where it can be tested better.

💯

NachoSoto: I agree. The purpose of this PR was to avoid calling the DNS logic twice though.

Issue #2028, opened 2022-11-06
Error: A server with the specified hostname could not be found


jesus-mg-ios:
- [ ] I have updated Purchases SDK to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Development/CONTRIBUTING.md)
- [X] I have searched the [Community](https://community.revenuecat.com)
- [X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [X] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
We're receiving a random error on some devices `ErrorCode Code=10 "Error performing request."`,  please see more in additional context.
The thing is the error says that the path is satisfied. It's a little bit strange, I mean if the path is satisfied the host should be found. the source function is `init(_:dnsChecker:)` . Could be an error inside this function that makes it fail?

Thanks in advance. I really appreciate your hard work.
 
1. Environment
   1. Platform: iOS
   2. SDK version: 4.13.3
   3. StoreKit 2 (_disabled with `useStoreKit2IfEnabled(false)`_) (Y/N): useStoreKit2IfEnabled(false)
   4. OS version: iOS 15.X
   5. Xcode version: 13.4.1
   6. How widespread is the issue. Percentage of devices affected. -
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue -
3. Steps to reproduce, with a description of expected vs. actual behavior -
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

**Additional context**
```
readable_error_code: NETWORK_ERROR

Error Code: 10
source_function: init(_:dnsChecker:)

RevenueCat.ErrorCode.Error Domain=RevenueCat.ErrorCode Code=10 "Error performing request."
UserInfo={NSLocalizedDescription=Error performing request., source_file=RevenueCat/HTTPClient.swift:363, 
NSUnderlyingError=0x2805cea60 
{Error Domain=NSURLErrorDomain Code=-1003 "A server with the specified hostname could not be found."
 UserInfo={_kCFStreamErrorCodeKey=8,
 NSUnderlyingError=0x2805cf840 {Error Domain=kCFErrorDomainCFNetwork Code=-1003 "

NSUnderlyingError
[...]

PR #2167, opened 2022-12-19
`DNSChecker`: added log with resolved host

NachoSoto:
Example:
> DEBUG: ℹ️ Failing url 'https://0.0.0.0/offers' resolved to host '0.0.0.0'

Looking into #2028, it appears that the user is getting a `NSURLErrorCannotFindHost` with a `NSErrorFailingURLKey`. We didn't determine that this was a `DNS` error because whatever the host resolved to, wasn't one of `Set(["0.0.0.0", "127.0.0.1"])`.

This will add a log that will help us figure out why they got a `NSURLErrorCannotFindHost`.

### DIFF
--- a/Sources/Logging/Strings/NetworkStrings.swift
+++ b/Sources/Logging/Strings/NetworkStrings.swift
@@ -31,6 +31,7 @@ enum NetworkStrings {
     case retrying_request(httpMethod: String, path: String)
     case could_not_find_cached_response
     case could_not_find_cached_response_in_already_retried(response: String)
+    case failing_url_resolved_to_host(url: URL, resolvedHost: String)
     case blocked_network(url: URL, newHost: String?)
 
 }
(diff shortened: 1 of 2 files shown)


Sources/Logging/Strings/NetworkStrings.swift
```diff
@@ -86,7 +87,10 @@ extension NetworkStrings: CustomStringConvertible {
             return "We can't find the cached response, but call has already been retried. " +
                 "Returning result from backend \(response)"
 
-        case .blocked_network(let url, let newHost):
+        case let .failing_url_resolved_to_host(url, resolvedHost):
+            return "Failing url '\(url)' resolved to host '\(resolvedHost)'"
```
tonidero, line 91: Would we have any secret info in the URL? I would guess not but just checking we don't leak any info in the logs.

tonidero (approved):

Sources/Logging/Strings/NetworkStrings.swift
NachoSoto: I don't think so, all that data is normally in request bodies.

PR #2168, opened 2022-12-19
`HTTPClientTests`: refactored to use `waitUntil`

NachoSoto:
See also #2071 and #2144.

I broke these tests recently, and because we're using `toEventually` they each take over 1 second to fail, instead of failing immediately after the request finishes.

### DIFF
--- a/Tests/UnitTests/Networking/HTTPClientTests.swift
+++ b/Tests/UnitTests/Networking/HTTPClientTests.swift
@@ -16,12 +16,13 @@ import XCTest
 class HTTPClientTests: TestCase {
 
     private typealias EmptyResponse = HTTPResponse<HTTPEmptyResponseBody>.Result
-    let apiKey = "MockAPIKey"
-    let systemInfo = MockSystemInfo(finishTransactions: true)
-    var client: HTTPClient!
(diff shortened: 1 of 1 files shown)


Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -933,16 +947,17 @@ class HTTPClientTests: TestCase {
             return response
         }
 
-        let obtainedError: Atomic<NetworkError?> = nil
-        self.client.perform(.init(method: .get, path: path)) { (result: HTTPResponse<Data>.Result) in
-            obtainedError.value = result.error
+        let obtainedError: NetworkError? = waitUntilValue { completion in
```
vegaro, line 950: Does this need to be a `Atomic`?

vegaro (approved):

Tests/UnitTests/Networking/HTTPClientTests.swift
NachoSoto: Not with `waitUntilValue` anymore since it simply returns the value :) it doesn’t múrate anything.

PR #2172, opened 2022-12-19
`Purchase Tester`: added ability to purchase products directly with `StoreKit`

NachoSoto:
This will help test observer mode, mimicking how an app might use `RevenueCat` with observer mode.

When that setting is enabled, 2 new buttons will appear:
Simulator Screen Shot - iPhone 14 Pro Max - 2022-12-20 at 08 58 07

The implementation isn't the most maintainable, but we don't really need it to be. I favored simplicity instead of over-complicating the implementation. It's not meant to be a canonical example of the correct way to purchase with StoreKit, just a quick way of testing this behavior.

### DIFF
--- a/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Helpers/PurchasesOrchestrator.swift
+++ b/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Helpers/PurchasesOrchestrator.swift
@@ -49,7 +49,9 @@ final class PurchasesOrchestrator: NSObject {
             await transaction.finish()
 
             print("Successfully purchased SK2 product")
-        case let .success(.unverified(_, error)):
+        case let .success(.unverified(transaction, error)):
+            await transaction.finish()
+
             throw error
         case .userCancelled:
             return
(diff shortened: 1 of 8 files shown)


Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Helpers/PurchasesOrchestrator.swift
```diff
@@ -0,0 +1,112 @@
...
+
+    func purchase(sk2Product product: SK2Product) async throws {
+        let result = try await product.purchase()
+
+        switch result {
+        case let .success(.verified(transaction)):
+            await transaction.finish()
+
+            print("Successfully purchased SK2 product")
+        case let .success(.unverified(_, error)):
```
vegaro, line 52: do you not have to finish the transactions here?

NachoSoto: Hmm I guess we do yeah let me fix it.
It’s not a big deal cause this is just for internal testing, so I hope none of us is trying to cheat purchasing this 😅

PR #2174, opened 2022-12-20
`IntroEligibilityCalculator`: fixed logic for subscriptions in same group

NachoSoto:
- Fixes  [TRIAGE-221] for `SK1`.
- Filed `FB11889732` for `SK2`.

### Changes:
- Added 2 `IntroEligibilityCalculatorTests` unit tests to cover this scenario
- Added 1  `IntroEligibilityCalculatorTests` unit tests to cover the equivalent of the `testIneligibleForIntroAfterPurchaseExpires` integration test
- Added 2 `StoreKitIntegrationTests` to cover this scenario (one of them disabled in SK2)
- Updated `IntroEligibilityCalculator` logic to fix the new tests.
- Refactored `IntroEligibilityCalculatorTests` to avoid duplication and allow adding new cases more easily.
- Added a few more tests to ensure other cases are correct.

[TRIAGE-221]: https://revenuecats.atlassian.net/browse/TRIAGE-221?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/IntroEligibilityCalculator.swift
+++ b/Sources/Purchasing/IntroEligibilityCalculator.swift
@@ -41,12 +41,12 @@ class IntroEligibilityCalculator {
             let receipt = try self.receiptParser.parse(from: receiptData)
             Logger.debug(Strings.customerInfo.checking_intro_eligibility_locally_from_receipt(receipt))
 
-            let activeSubscriptionProductIdentifiers = receipt.activeSubscriptionProductIdentifiers
-            let purchasedProductIdsWithIntroOffersOrFreeTrials =
-            receipt.purchasedIntroOfferOrFreeTrialProductIdentifiers
+            let expiredSubscriptionProductIdentifiers = receipt.expiredSubscriptionProductIdentifiers
+            let activeSubscriptionsProductIdentifiers = receipt
+                .activeSubscriptionsProductIdentifiers
             let allProductIdentifiers = candidateProductIdentifiers
-                .union(activeSubscriptionProductIdentifiers)
-                .union(purchasedProductIdsWithIntroOffersOrFreeTrials)
+                .union(expiredSubscriptionProductIdentifiers)
+                .union(activeSubscriptionsProductIdentifiers)
 
             self.productsManager.products(withIdentifiers: allProductIdentifiers) {
                 let allProducts = $0.value ?? []
@@ -54,17 +54,17 @@ class IntroEligibilityCalculator {
                 let candidateProducts = allProducts.filter {
                     candidateProductIdentifiers.contains($0.productIdentifier)
                 }
-                let activeSubscriptionProducts = allProducts.filter {
-                    activeSubscriptionProductIdentifiers.contains($0.productIdentifier)
+                let expiredSubscriptionProducts = allProducts.filter {
+                    expiredSubscriptionProductIdentifiers.contains($0.productIdentifier)
                 }
-                let purchasedProductsWithIntroOffersOrFreeTrials = allProducts.filter {
-                    purchasedProductIdsWithIntroOffersOrFreeTrials.contains($0.productIdentifier)
+                let activeSubscriptionsProducts = allProducts.filter {
+                    activeSubscriptionsProductIdentifiers.contains($0.productIdentifier)
                 }
 
                 let eligibility = self.checkEligibility(
                     candidateProducts: candidateProducts,
-                    purchasedProductsWithIntroOffersOrFreeTrials: purchasedProductsWithIntroOffersOrFreeTrials,
-                    activeSubscriptionProducts: activeSubscriptionProducts
+                    activeSubscriptionsProducts: activeSubscriptionsProducts,
+                    expiredSubscriptionProducts: expiredSubscriptionProducts
                 )
                 result.merge(eligibility, strategy: .overwriteValue)
 
@@ -92,34 +92,31 @@ private extension IntroEligibilityCalculator {
 
     func checkEligibility(
         candidateProducts: Set<StoreProduct>,
-        purchasedProductsWithIntroOffersOrFreeTrials: Set<StoreProduct>,
-        activeSubscriptionProducts: Set<StoreProduct>
+        activeSubscriptionsProducts: Set<StoreProduct>,
+        expiredSubscriptionProducts: Set<StoreProduct>
     ) -> [String: IntroEligibilityStatus] {
         var result: [String: IntroEligibilityStatus] = [:]
 
         for candidate in candidateProducts {
             guard candidate.subscriptionPeriod != nil else {
-                result[candidate.productIdentifier] = IntroEligibilityStatus.unknown
+                result[candidate.productIdentifier] = .unknown
                 continue
             }
-            let usedIntroForProductIdentifier = (
+            let activeSubscriptionInGroup = (
                 candidate.subscriptionGroupIdentifier != nil &&
-                purchasedProductsWithIntroOffersOrFreeTrials.contains { purchased in
-                    candidate.subscriptionGroupIdentifier == purchased.subscriptionGroupIdentifier
+                activeSubscriptionsProducts.contains {
+                    $0.subscriptionGroupIdentifier == candidate.subscriptionGroupIdentifier
                 }
             )
-            let usedIntroInSubscriptionGroup = (
-                candidate.subscriptionGroupIdentifier != nil &&
-                activeSubscriptionProducts.contains { purchased in
-                    candidate.subscriptionGroupIdentifier == purchased.subscriptionGroupIdentifier
-                }
-            )
+            let expiredSubscriptionForProduct = expiredSubscriptionProducts
+                .lazy
+                .map(\.productIdentifier)
+                .contains(candidate.productIdentifier)
 
             if candidate.introductoryDiscount == nil {
                 result[candidate.productIdentifier] = .noIntroOfferExists
             } else {
-                // TODO: this isn't quite right, see failing tests
-                result[candidate.productIdentifier] = usedIntroForProductIdentifier || usedIntroInSubscriptionGroup
+                result[candidate.productIdentifier] = activeSubscriptionInGroup || expiredSubscriptionForProduct
                     ? .ineligible
                     : .eligible
             }
--- a/Sources/Purchasing/IntroEligibilityCalculator.swift
+++ b/Sources/Purchasing/IntroEligibilityCalculator.swift
@@ -36,33 +36,27 @@ class IntroEligibilityCalculator {
         }
         Logger.debug(Strings.customerInfo.checking_intro_eligibility_locally)
 
-        var result = candidateProductIdentifiers.reduce(into: [:]) { resultDict, productId in
-            resultDict[productId] = IntroEligibilityStatus.unknown
-        }
+        var result = candidateProductIdentifiers.dictionaryWithValues { _ in IntroEligibilityStatus.unknown }
         do {
             let receipt = try self.receiptParser.parse(from: receiptData)
             Logger.debug(Strings.customerInfo.checking_intro_eligibility_locally_from_receipt(receipt))
 
-            let purchasedProductIdsWithIntroOffersOrFreeTrials =
-                receipt.purchasedIntroOfferOrFreeTrialProductIdentifiers()
+            let activeSubscriptionProductIdentifiers = receipt.activeSubscriptionProductIdentifiers
+            let allProductIdentifiers = candidateProductIdentifiers.union(activeSubscriptionProductIdentifiers)
 
-            let allProductIdentifiers =
-                candidateProductIdentifiers.union(purchasedProductIdsWithIntroOffersOrFreeTrials)
-
-            productsManager.products(withIdentifiers: allProductIdentifiers) {
+            self.productsManager.products(withIdentifiers: allProductIdentifiers) {
                 let allProducts = $0.value ?? []
 
-                let purchasedProductsWithIntroOffersOrFreeTrials = allProducts.filter {
-                    purchasedProductIdsWithIntroOffersOrFreeTrials.contains($0.productIdentifier)
-                }
                 let candidateProducts = allProducts.filter {
                     candidateProductIdentifiers.contains($0.productIdentifier)
                 }
+                let activeSubscriptionProducts = allProducts.filter {
+                    activeSubscriptionProductIdentifiers.contains($0.productIdentifier)
+                }
 
-                let eligibility = self.checkEligibility(
-                    candidateProducts: candidateProducts,
-                    purchasedProductsWithIntroOffers: purchasedProductsWithIntroOffersOrFreeTrials)
-                result.merge(eligibility) { (_, new) in new }
+                let eligibility = self.checkEligibility(candidateProducts: candidateProducts,
+                                                        activeSubscriptionProducts: activeSubscriptionProducts)
+                result.merge(eligibility, strategy: .overwriteValue)
 
                 Logger.debug(
                     Strings.customerInfo.checking_intro_eligibility_locally_result(productIdentifiers: result)
@@ -86,8 +80,10 @@ extension IntroEligibilityCalculator: @unchecked Sendable {}
 @available(iOS 12.0, macOS 10.14, macCatalyst 13.0, tvOS 12.0, watchOS 6.2, *)
 private extension IntroEligibilityCalculator {
 
-    func checkEligibility(candidateProducts: Set<StoreProduct>,
-                          purchasedProductsWithIntroOffers: Set<StoreProduct>) -> [String: IntroEligibilityStatus] {
+    func checkEligibility(
+        candidateProducts: Set<StoreProduct>,
+        activeSubscriptionProducts: Set<StoreProduct>
+    ) -> [String: IntroEligibilityStatus] {
         var result: [String: IntroEligibilityStatus] = [:]
 
         for candidate in candidateProducts {
@@ -95,17 +91,17 @@ private extension IntroEligibilityCalculator {
                 result[candidate.productIdentifier] = IntroEligibilityStatus.unknown
                 continue
             }
-            let usedIntroForProductIdentifier = purchasedProductsWithIntroOffers
-                .contains { purchased in
-                    let foundByGroupId = (candidate.subscriptionGroupIdentifier != nil
-                        && candidate.subscriptionGroupIdentifier == purchased.subscriptionGroupIdentifier)
-                    return foundByGroupId
+            let usedIntroInSubscriptionGroup = (
+                candidate.subscriptionGroupIdentifier != nil &&
+                activeSubscriptionProducts.contains { purchased in
+                    candidate.subscriptionGroupIdentifier == purchased.subscriptionGroupIdentifier
                 }
+            )
 
             if candidate.introductoryDiscount == nil {
                 result[candidate.productIdentifier] = .noIntroOfferExists
             } else {
-                result[candidate.productIdentifier] = usedIntroForProductIdentifier
+                result[candidate.productIdentifier] = usedIntroInSubscriptionGroup
                     ? IntroEligibilityStatus.ineligible
                     : IntroEligibilityStatus.eligible
             }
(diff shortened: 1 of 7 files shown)


Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -314,6 +314,16 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
...
     }
 
+    func testIneligibleForIntroForDifferentProductInSameSubscriptionGroupAfterPurchase() async throws {
+        let product1 = try await self.annualPackage.storeProduct
+        let product2 = try await self.annualNoIntroProduct
+
+        _ = try await Purchases.shared.purchase(product: product2)
+
+        let eligibility = await Purchases.shared.checkTrialOrIntroDiscountEligibility(product: product1)
+        expect(eligibility) == .ineligible
```
NachoSoto, line 324: > A customer reported that they are calling `checkTrialOrIntroDiscountEligibility` and are receiving eligible for a trial product within the same subscription group of their current subscription. The payment sheet from Apple confirms that they should be receiving ineligible

Both SK1 and SK2 agree here that this is `eligible`, confirming that this could be an Apple bug.

vegaro (approved):

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
vegaro: Maybe add it as a comment so everyone knows why this is failing

aboedo: I left some comments on the jira ticket:

This should be ineligible IFF the purchase is active. Do we still have the bug where min time rate means instantly-expiring subscriptions? If so, then that might explain it. 

Also, we need to update the code for SK1 to account for that, it currently doesn't take into account a possibly active subscription. 

Maybe we can add that _other_ use case too? i.e.: purchase with no intro, cancel, expire, check that is eligible for different product in same subs group that has intro trial

aboedo: so... that's all to say that this test case is valid, but maybe we could add a line that checks that the subscription is currently active to make it more explicit? or rename the test case to include the fact that the subs is active?

NachoSoto: Yeah I'll add another test.
This one is failing so we can't keep it.

I'll fix the SK1 logic, and temporarily `skip` the test for SK2.

NachoSoto: > This should be ineligible IFF the purchase is active. Do we still have the bug where min time rate means instantly-expiring subscriptions? If so, then that might explain it.

No we don't case we moved from weekly to monthly subscriptions (and we add a debug (warning?) log to mention that this is happening).

NachoSoto: > Maybe we can add that other use case too? i.e.: purchase with no intro, cancel, expire, check that is eligible for different product in same subs group that has intro trial

Done:

```swift
func testEligibleForIntroForDifferentProductInSameSubscriptionGroupAfterExpiration() async throws {
    let product1 = try await self.annualNoIntroProduct
    let product2 = try await self.annualPackage.storeProduct

    let customerInfo = try await Purchases.shared.purchase(product: product1).customerInfo
    let entitlement = self.verifyEntitlementWentThrough(customerInfo)

    try await self.expireSubscription(entitlement)

    let eligibility = await Purchases.shared.checkTrialOrIntroDiscountEligibility(product: product2)
    expect(eligibility) == .eligible
}
```

NachoSoto: That test passes on both SK1 and SK2.

Sources/Purchasing/IntroEligibilityCalculator.swift
```diff
@@ -36,33 +36,27 @@ class IntroEligibilityCalculator {
...
+                    activeSubscriptionProductIdentifiers.contains($0.productIdentifier)
+                }
 
-                let eligibility = self.checkEligibility(
-                    candidateProducts: candidateProducts,
-                    purchasedProductsWithIntroOffers: purchasedProductsWithIntroOffersOrFreeTrials)
-                result.merge(eligibility) { (_, new) in new }
+                let eligibility = self.checkEligibility(candidateProducts: candidateProducts,
+                                                        activeSubscriptionProducts: activeSubscriptionProducts)
+                result.merge(eligibility, strategy: .overwriteValue)
```
NachoSoto, line 59: Simplified this using the `MergeStrategy` (see #2190)

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -265,6 +265,34 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
         expect(eligibility) == .eligible
     }
 
+    func testIneligibleForIntroForDifferentProductInSameSubscriptionGroupAfterPurchase() async throws {
+        try XCTSkipIf(
```
NachoSoto, line 269: Disabled for now.

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -265,6 +265,34 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
...
+        let productWithNoTrial = try await self.product(Self.group3MonthlyNoTrialProductID)
+        let productWithTrial = try await self.annualPackage.storeProduct
+
+        _ = try await Purchases.shared.purchase(product: productWithNoTrial)
+
+        let eligibility = await Purchases.shared.checkTrialOrIntroDiscountEligibility(product: productWithTrial)
+        expect(eligibility) == .ineligible
+    }
+
+    func testEligibleForIntroForDifferentProductInSameSubscriptionGroupAfterExpiration() async throws {
```
NachoSoto, line 283: This test used to pass before too, but it helped me make sure I didn't break this scenario, thanks for the suggestion @aboedo.

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -265,6 +265,34 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
...
+            "This test currently does not pass with SK2 (see FB11889732)"
+        )
+
+        let productWithNoTrial = try await self.product(Self.group3MonthlyNoTrialProductID)
+        let productWithTrial = try await self.annualPackage.storeProduct
+
+        _ = try await Purchases.shared.purchase(product: productWithNoTrial)
+
+        let eligibility = await Purchases.shared.checkTrialOrIntroDiscountEligibility(product: productWithTrial)
+        expect(eligibility) == .ineligible
```
NachoSoto, line 280: This test was failing before this fix.

Tests/UnitTests/Purchasing/IntroEligibilityCalculatorTests.swift
```diff
@@ -112,6 +112,79 @@ class IntroEligibilityCalculatorTests: TestCase {
         ]
     }
 
+    func testCheckTrialOrIntroDiscountEligibilityReturnsIneligibleWithActiveSubscriptionInSameGroup() {
```
NachoSoto, line 115: These 2 tests were failing before the fix.

NachoSoto:
@aboedo this is ready now, I think the logic is correct?

NachoSoto:
Thanks to all our test coverage I realized that the new logic wasn't quite right. I've added another failing unit test, but now I have to figure out how to reconcile both scenarios.

Sources/Purchasing/IntroEligibilityCalculator.swift
```diff
@@ -36,33 +36,37 @@ class IntroEligibilityCalculator {
...
             let receipt = try self.receiptParser.parse(from: receiptData)
             Logger.debug(Strings.customerInfo.checking_intro_eligibility_locally_from_receipt(receipt))
 
+            let activeSubscriptionProductIdentifiers = receipt.activeSubscriptionProductIdentifiers
             let purchasedProductIdsWithIntroOffersOrFreeTrials =
-                receipt.purchasedIntroOfferOrFreeTrialProductIdentifiers()
-
-            let allProductIdentifiers =
-                candidateProductIdentifiers.union(purchasedProductIdsWithIntroOffersOrFreeTrials)
+            receipt.purchasedIntroOfferOrFreeTrialProductIdentifiers
```
NachoSoto, line 46: @aboedo good news: turns out we still needed this. Bad news, not sure how to reconcile both scenarios.

Sources/Purchasing/IntroEligibilityCalculator.swift
```diff
@@ -86,26 +90,36 @@ extension IntroEligibilityCalculator: @unchecked Sendable {}
...
+                activeSubscriptionProducts.contains { purchased in
+                    candidate.subscriptionGroupIdentifier == purchased.subscriptionGroupIdentifier
                 }
+            )
 
             if candidate.introductoryDiscount == nil {
                 result[candidate.productIdentifier] = .noIntroOfferExists
             } else {
-                result[candidate.productIdentifier] = usedIntroForProductIdentifier
+                // TODO: this isn't quite right, see failing tests
```
NachoSoto, line 121: TODO: need to figure this out.

Sources/Purchasing/IntroEligibilityCalculator.swift
aboedo: lol

Sources/Purchasing/IntroEligibilityCalculator.swift
```diff
@@ -86,26 +90,36 @@ extension IntroEligibilityCalculator: @unchecked Sendable {}
...
+                purchasedProductsWithIntroOffersOrFreeTrials.contains { purchased in
+                    candidate.subscriptionGroupIdentifier == purchased.subscriptionGroupIdentifier
+                }
+            )
+            let usedIntroInSubscriptionGroup = (
+                candidate.subscriptionGroupIdentifier != nil &&
+                activeSubscriptionProducts.contains { purchased in
+                    candidate.subscriptionGroupIdentifier == purchased.subscriptionGroupIdentifier
                 }
+            )
```
aboedo, line 116: aren't we missing the ones with subsGroupID == nil?

NachoSoto: Maybe. Can we go through this together next week? My brain might have been fried last week, but couldn't figure out how to make all the tests pass.

aboedo: absolutely!

NachoSoto:
Okay I got all tests passing. I _think_ the new implementation is correct, but I'll write more tomorrow to cover the logic with products in multiple groups to make sure.

Tests/UnitTests/Purchasing/IntroEligibilityCalculatorTests.swift
```diff
@@ -74,67 +73,117 @@ class IntroEligibilityCalculatorTests: TestCase {
...
-        let product1 = MockSK1Product(mockProductIdentifier: "com.revenuecat.product1",
-                                      mockSubscriptionGroupIdentifier: "group1")
-        product1.mockDiscount = MockSKProductDiscount()
-        let product2 = MockSK1Product(mockProductIdentifier: "com.revenuecat.product2",
-                                      mockSubscriptionGroupIdentifier: "group2")
-        product2.mockDiscount = MockSKProductDiscount()
-
-        mockProductsManager.stubbedProductsCompletionResult = .success(
-            Set([product1, product2].map(StoreProduct.init(sk1Product:)))
+        self.testEligibility(
```
NachoSoto, line 82: 🧹 I've refactored all these tests to avoid the duplication and to be able to add more cases more easily.

Sources/Purchasing/IntroEligibilityCalculator.swift
```diff
@@ -86,28 +90,35 @@ extension IntroEligibilityCalculator: @unchecked Sendable {}
...
+                .map(\.productIdentifier)
+                .contains(candidate.productIdentifier)
 
             if candidate.introductoryDiscount == nil {
                 result[candidate.productIdentifier] = .noIntroOfferExists
             } else {
-                result[candidate.productIdentifier] = usedIntroForProductIdentifier
-                    ? IntroEligibilityStatus.ineligible
-                    : IntroEligibilityStatus.eligible
+                result[candidate.productIdentifier] = activeSubscriptionInGroup || expiredSubscriptionForProduct
```
NachoSoto, line 119: This is the new logic.

Sources/Purchasing/IntroEligibilityCalculator.swift
aboedo: how does this differ from `result += eligibility`?

Sources/Purchasing/IntroEligibilityCalculator.swift
```diff
@@ -86,28 +96,44 @@ extension IntroEligibilityCalculator: @unchecked Sendable {}
...
+            let activeSubscriptionInGroup = (
+                candidate.subscriptionGroupIdentifier != nil &&
+                activeSubscriptionsProducts.contains {
+                    $0.subscriptionGroupIdentifier == candidate.subscriptionGroupIdentifier
                 }
+            )
+            let expiredSubscriptionForProduct = expiredSubscriptionProducts
+                .lazy
+                .map(\.productIdentifier)
+                .contains(candidate.productIdentifier)
```
aboedo, line 121: We could potentially have a false negative here:
Expired subs for the product only matter if you have used a free trial for the product. 

For example, if you subscribe to `yearly 1`, and it has no free trial available: 
- you're not eligible while the subscription to `yearly 1` is active
- if the developer then _adds_ a free trial to `yearly 1` and you cancel the subscription, you're eligible because you haven't used a free trial for this product (or its group), and one is available

aboedo: image

https://developer.apple.com/documentation/storekit/in-app_purchase/original_api_for_in-app_purchase/subscriptions_and_offers/implementing_introductory_offers_in_your_app

aboedo (changes requested):

Sources/Purchasing/IntroEligibilityCalculator.swift
NachoSoto: Oh it's the same, I'll change it, thanks!
I improved this from the previous implementation after bae4a004 but didn't realize it can be even simpler now.

Sources/Purchasing/IntroEligibilityCalculator.swift
NachoSoto: Thanks a lot for the thorough review!

I think my TDD approach has kinda failed me here. I’ve been writing tests and finding the correct implementation that satisfies all tests, but it led me to the wrong implementation.

I misinterpreted the `is_in_trial_period` property in the receipt, I figured that wouldn't be `true` for an expired subscription, but now I see that it is (and verified through integration tests).
So your suggested implementation works after I fix the incorrect unit test 👍🏻 

I've also add a new unit test to cover this false positive you described, which failed before the fix:
```swift
func testCheckTrialOrIntroDiscountEligibilityReturnsEligibleForPreviouslyOwnedSubscriptionWithUnusedTrial() {
    self.testEligibility(
        purchaseExpirationsByProductIdentifier: [
            ("com.revenuecat.product1", Date().addingTimeInterval(-1000), false)
        ],
        productsInGroups: [
            "com.revenuecat.product1": (groupID: "group1", hasTrial: true)
        ],
        expectedResult: [
            "com.revenuecat.product1": .eligible
        ]
    )
}
```

NachoSoto:
@aboedo third time is the charm hopefully! Let me know if this is good now 🙏🏻

Sources/Purchasing/IntroEligibilityCalculator.swift
```diff
@@ -86,28 +90,39 @@ extension IntroEligibilityCalculator: @unchecked Sendable {}
...
+            let activeSubscriptionInGroup = (
+                candidate.subscriptionGroupIdentifier != nil &&
+                activeSubscriptionsProducts.contains {
+                    $0.subscriptionGroupIdentifier == candidate.subscriptionGroupIdentifier
                 }
+            )
+            let expiredTrialInGroup = (
+                candidate.subscriptionGroupIdentifier != nil &&
+                expiredTrialProducts.contains {
+                    $0.subscriptionGroupIdentifier == candidate.subscriptionGroupIdentifier
```
aboedo, line 114: this would miss eligibility for a product that doesn't have a subscription. In practice I'm not sure it's possible to have one anymore, but it'd be nice to cover for it. 
i.e.: 
- purchase `product1`, `subsGroupIdentifier = nil`
- eligibility should be: `ineligible` if `product1` is active subscription or expired but had trial, but this would return `unknown`

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -265,6 +265,46 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
         expect(eligibility) == .eligible
     }
 
+    func testIneligibleForIntroForDifferentProductInSameSubscriptionGroupAfterPurchase() async throws {
+        try XCTSkipIf(
+            Self.storeKit2Setting == .enabledForCompatibleDevices,
+            "This test currently does not pass with SK2 (see FB11889732)"
```
aboedo, line 271: is this one under the company account? I can't find it

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
aboedo: it's crazy that this fails. this is with timeRate > the minimum? just wondering if it's a bug related to the thing where subs would immediately expire

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -265,6 +265,46 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
...
+
+        let eligibility = await Purchases.shared.checkTrialOrIntroDiscountEligibility(product: productWithTrial)
+        expect(eligibility) == .ineligible
+    }
+
+    func testEligibleForIntroForDifferentProductInSameSubscriptionGroupAfterExpiration() async throws {
+        let productWithNoTrial = try await self.product(Self.group3MonthlyNoTrialProductID)
+        let productWithTrial = try await self.annualPackage.storeProduct
+
+        let customerInfo = try await Purchases.shared.purchase(product: productWithNoTrial).customerInfo
```
aboedo, line 287: for some reason I never realized that you can do this syntax (i.e.: `customerInfo = try await purchase().customerinfo`. That's awesome

aboedo (approved):
Looks great! 
I left a comment about an edge case that I'm not sure is possible to hit anymore, but it'd be nice to still consider.

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
NachoSoto: It is. Click on "Inbox" instead of "Submitted" because it got a response so it's there.

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
NachoSoto: The time rate is the minimum, but it's not a weekly subscription, it's monthly, so it's not the "immediate expiration" bug.

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
aboedo: Thanks!! That's exactly what happened 🤦 I was only checking "submitted"

Sources/Purchasing/IntroEligibilityCalculator.swift
NachoSoto: We're talking about this offline. I'll merge this for now since it's over a month old and it already fixes the main issue.
I'll that those 2 new tests in a separate PR.

NachoSoto: #2247

PR #2176, opened 2022-12-20
`BundleSandboxEnvironmentDetector`: fixed logic for `macOS`

NachoSoto:
I found out today that when running `PurchaseTester` on `macOS` (either native or with Catalyst), the SDK was sending `X-Is-Sandbox` = `true`.
The new examples in `MockBundle` illustrate this.

### DIFF
--- a/Sources/Misc/SandboxEnvironmentDetector.swift
+++ b/Sources/Misc/SandboxEnvironmentDetector.swift
@@ -30,11 +30,17 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
     }
 
     var isSandbox: Bool {
-        guard let url = self.bundle.appStoreReceiptURL else {
+        guard let path = self.bundle.appStoreReceiptURL?.path else {
             return false
         }
 
-        return url.path.contains("sandboxReceipt")
+        // `true` for either `macOS` or `Catalyst`
+        let isMASReceipt = path.contains("MASReceipt/receipt")
+        if isMASReceipt {
+            return path.contains("Xcode/DerivedData")
+        } else {
+            return path.contains("sandboxReceipt")
+        }
     }
 
     #if DEBUG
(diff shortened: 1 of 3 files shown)


Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -30,11 +30,17 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
...
-        guard let url = self.bundle.appStoreReceiptURL else {
+        guard let path = self.bundle.appStoreReceiptURL?.path else {
             return false
         }
 
-        return url.path.contains("sandboxReceipt")
+        // `true` for either `macOS` or `Catalyst`
+        let isMASReceipt = path.contains("MASReceipt/receipt")
+        if isMASReceipt {
+            return path.contains("Xcode/DerivedData")
```
aboedo, line 40: OOOHHHH right, great catch! 

could this just be simplified to "if contains sandbox or Xcode/DerivedData"? Feels like we wouldn't get a false positive, right?

aboedo (approved):

Sources/Misc/SandboxEnvironmentDetector.swift
NachoSoto: That would be a false positive for iOS as well. This is an example receipt URL for the iOS simulator:
```
/Users/nachosoto/Library/Developer/Xcode/DerivedData/RevenueCat-areulmnbtfydolgoennnwogndmkr/Build/Products/Debug-iphonesimulator/UnitTests.xctest/base64encodedreceiptsample1.txt
```

So we need to only do that for `macOS`. I think we should hardcode the least amount of strings though, in case this changes in the future, but I believe we can't simplify it more than this?

aboedo: ohhhh simulator, right. I'm so used to testing on device that I forgot 😅

PR #2177, opened 2022-12-20
`StoreKitTestHelpers`: cleaned up unnecessary log

NachoSoto:
See #2066.

I saw a bunch of these messages:
> DEBUG: ℹ️ Finishing 0 transactions before running tests

Which doesn't add anything if there aren't any leftover transactions.


aboedo (approved):
👌

PR #2178, opened 2022-12-20
`Restore purchases`: post product data when posting receipts

NachoSoto:
Fixes [CSDK-604].

When restore receipts, we don't post any price for products because we don't know what the user bought.
As a best-effort approach, this parses the `AppleReceipt` and looks for products that were purchased.

[CSDK-604]: https://revenuecats.atlassian.net/browse/CSDK-604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Tests/UnitTests/Purchasing/Purchases/PurchasesRestoreTests.swift
```diff
@@ -91,6 +91,91 @@ class PurchasesRestoreTests: BasePurchasesTests {
...
+        self.receiptFetcher.shouldReturnReceipt = true
+        self.mockReceiptParser.stubbedParseResult = .init(
+            bundleId: "",
+            applicationVersion: "",
+            originalApplicationVersion: nil,
+            opaqueValue: Data(),
+            sha1Hash: Data(),
+            creationDate: Date(),
+            expirationDate: nil,
+            inAppPurchases: [
```
NachoSoto, line 108: I added a bunch of products to test the sorting / filtering behavior .

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1107,6 +1114,32 @@ private extension PurchasesOrchestrator {
...
+    ) {
+        guard let receipt = try? self.receiptParser.parse(from: receiptData) else { completion(nil); return }
+
+        // Find the latest purchased subscription
+        guard let productIdentifier = receipt.inAppPurchases
+            .lazy
+            .filter({ $0.isActiveSubscription })
+            .sorted(by: { $0.purchaseDate > $1.purchaseDate })
+            .first?
+            .productId else { completion(nil); return }
```
aboedo, line 1132: this feels out of scope for this class, tbh. Like there should either be a separate object for it, or potentially the ReceiptParser itself should be able to answer the latest subscriptions found in the receipt. 

Also... This shouldn't be limited to subscriptions, we're really just looking for a set of all product ids the customer has ever purchased

NachoSoto: > this feels out of scope for this class, tbh. Like there should either be a separate object for it, or potentially the ReceiptParser itself should be able to answer the latest subscriptions found in the receipt.

Totally good point

> This shouldn't be limited to subscriptions

I can remove that filter, but....

> we're really just looking for a set of all product ids the customer has ever purchased

But we need to pick one, because we post only 1 `ProductRequestData` based on a single `StoreProduct`.

aboedo: ohh interesting. We should check with the backend team, maybe this is a limitation in khepri that we can work around. If not, then I think the active subscription approach is as good as any

swehner: We do not support multiple productIds by receipt. I think there are a lot of things that would have to change to support more than one product. [Receipt](https://github.com/RevenueCat/khepri/blob/main/khepri/models/receipt.py#L106) and [Transaction](https://github.com/RevenueCat/khepri/blob/main/khepri/models/apple_transaction.py#L67) are laid out to have a single product.
There was talk also to support multi lineItem receipts, but I don't think we have a concrete plan for this.

aboedo: @swehner that's good to know. We're not really looking to map this to a purchase of multiple products, really, we're just trying to provide price information for the products that this user has purchased in the past that they might be restoring. Would that be achievable _without_ trying to convert this from a restore event to a purchase event/

NachoSoto:
@aboedo I extracted the logic to `AppleReceipt` and added a few more tests.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
aboedo: @swehner @antoniobg could I get final confirmation that sending the price of the active subscription + the product id in a restore won't negatively affect any endpoints / webhooks?

antoniobg: @aboedo I think it should be fine. It might affect the prices for newer transactions for the receipts for which we didn't have cached prices before, but I think it will be more accurate

Sources/LocalReceiptParsing/BasicTypes/AppleReceipt.swift
```diff
@@ -90,6 +90,15 @@ extension AppleReceipt {
         )
     }
 
+    /// Returns the most recent subscription (see `InAppPurchase.isActiveSubscription`).
+    var lastPurchasedSubscription: InAppPurchase? {
```
aboedo, line 94: ❤️

aboedo: this doesn't specify that it's _active_, though, how about `mostRecentActiveSubscription` or `lastPurchasedActiveSubscription`?

aboedo (approved):

PR #2180, opened 2022-12-20
`Purchases` initialization: refactor to avoid multiple concurrent instances in memory

NachoSoto:
We allow calling `Purchases.configure` multiple times. Likely a programmer error, but it's common in tests, or for example in `Purchase Tester` where users can re-configure the SDK at runtime (#1999), which is where I saw this issue.

When this happens, the SDK was creating the new instance before clearing the old one from memory (clear thanks to #2082, notice how the first `deinit` happens after the second `init`):
```
[Purchases] - VERBOSE: Purchases.init: created new Purchases instance: Purchases (0x0000600003ca4200)
[Purchases] - INFO: ℹ️ Purchases instance already set. Did you mean to configure two Purchases objects?
[Purchases] - VERBOSE: Purchases.init: created new Purchases instance: Purchases (0x0000600003c8a100)
[Purchases] - VERBOSE: Purchases.deinit: Purchases (0x0000600003ca4200)
[Purchases] - VERBOSE: Purchases.deinit: Purchases (0x0000600003c8a100)
```

With this change, we avoid that, which could also avoid issues with concurrent requests ([CSDK-517]):
```
[Purchases] - VERBOSE: Purchases.init: created new Purchases instance: Purchases (0x0000600002b6e600)
[Purchases] - INFO: ℹ️ Purchases instance already set. Did you mean to configure two Purchases objects?
[Purchases] - VERBOSE: Purchases.deinit: Purchases (0x0000600002b6e600)
[Purchases] - VERBOSE: Purchases.init: created new Purchases instance: Purchases (0x0000600002b68100)
[Purchases] - VERBOSE: Purchases.deinit: Purchases (0x0000600002b68100)
```

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -497,8 +497,9 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
         Self.purchases.value = nil
     }
 
-    static func setDefaultInstance(_ purchases: Purchases) {
-        self.purchases.modify { currentInstance in
+    @discardableResult
+    static func setDefaultInstance(_ purchases: @autoclosure () -> Purchases) -> Purchases {
+        return self.purchases.modify { currentInstance in
             if currentInstance != nil {
                 #if DEBUG
                 if ProcessInfo.isRunningRevenueCatTests {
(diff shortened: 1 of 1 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -497,18 +497,24 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
         Self.purchases.value = nil
     }
 
-    static func setDefaultInstance(_ purchases: Purchases) {
-        self.purchases.modify { currentInstance in
+    @discardableResult
+    static func setDefaultInstance(_ purchases: @autoclosure () -> Purchases) -> Purchases {
```
vegaro, line 501: Do you mind adding a comment here why this is an `autoclosure` and not just an instance of `Purchases`?

vegaro (approved):

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: Good idea

PR #2187, opened 2022-12-22
`CircleCI`: use Xcode 14.2.0

NachoSoto:
See https://discuss.circleci.com/t/xcode-14-2-released-breaking-changes/46303

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -278,7 +278,7 @@ jobs:
           command: bundle exec fastlane test_ios
           no_output_timeout: 5m
           environment:
-            SCAN_DEVICE: iPhone 14 (16.1)
+            SCAN_DEVICE: iPhone 14 (16.2)
       - compress_result_bundle:
           directory: fastlane/test_output/xctest/ios
           bundle_name: RevenueCat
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -794,7 +794,7 @@ workflows:
         equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
     jobs:
       - prepare-next-version:
-          xcode_version: '14.1.0'
+          xcode_version: '14.2.0'
```
tonidero, line 797: Hmm I think we could probably run some of these jobs in non-mac machines... But that doesn't belong to this PR.

tonidero (approved):

.circleci/config.yml
NachoSoto: Yeah there's lots we could do to speed up some of these. Which ones were you thinking specifically?

NachoSoto: Oh like `prepare-next-version`, yeah. I think it's so we can have some basic stuff installed, but I'm sure they're lighter images we could use.

tonidero: Yeah, I was thinking on any like this, where it's just running some fastlane scripts.

PR #2190, opened 2022-12-22
Added `Dictionary.merge`

NachoSoto:
This simplifies the `+=` implementation by avoiding an intermediate dictionary, and allows a simpler way of merging dictionaries.

Will be used in #2174.

### DIFF
--- a/Sources/FoundationExtensions/Dictionary+Extensions.swift
+++ b/Sources/FoundationExtensions/Dictionary+Extensions.swift
@@ -81,7 +81,7 @@ extension Dictionary {
     /// - Parameters:
     ///   - lhs: A dictionary to merge.
     ///   - rhs: Another dictionary to merge.
-    /// - Returns: An dictionary with keys and values from both.
+    /// - Returns: A dictionary with keys and values from both.
     static func + (lhs: [Key: Value], rhs: [Key: Value]) -> [Key: Value] {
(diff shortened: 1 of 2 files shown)


Sources/FoundationExtensions/Dictionary+Extensions.swift
```diff
@@ -72,7 +83,7 @@ extension Dictionary {
     ///   - rhs: Another dictionary to merge.
     /// - Returns: An dictionary with keys and values from both.
```
vegaro, line 84: I know it's not from this PR, but just saw this:

```suggestion
    /// - Returns: A dictionary with keys and values from both.
```

Sources/FoundationExtensions/Dictionary+Extensions.swift
```diff
@@ -84,7 +95,7 @@ extension Dictionary {
...
```
vegaro, line 96:
```suggestion
    /// - Returns: A dictionary with keys and values from both.
```

vegaro (approved):

Sources/FoundationExtensions/Dictionary+Extensions.swift
NachoSoto: Oops thanks.

PR #2195, opened 2022-12-29
Fix path to storekit file in purchase tester

mcastany:
### Motivation
Fix issue on purchase tester

### Description

Fix path. When opening PurchaseTester and check the schema, I see this error

image


NachoSoto:
Weird, this looks correct to me without the change. What version of Xcode are you using?

I tested your branch and it works too, so I'm not against it. I'm just confused 😅

NachoSoto:
I think the existing version is correct, I imagine the path should be relative to the project itself.
If you're on an older version of Xcode I wonder if they changed this, so maybe we should keep it as is since the current version is compatible with the latest Xcode 14.2?

NachoSoto (changes requested):
Let me know if you agree with my assessment about Xcode 13 vs 14.

aboedo:
I get the path issue when using this branch and not in `main` ¯\_(ツ)_/¯

Screenshot 2023-01-17 at 10 35 47 AM

aboedo:
Just tried this again on a fresh clone just in case, same results. `Xcode 14.2 (14C18)`. 
@mcastany I'm closing this since I think the path was correct, but feel free to reopen if you think we should update

PR #2197, opened 2023-01-05
`PurchasesReceiptParser`: improved documentation to reference `default`

NachoSoto:
The most obvious way to create `PurchasesReceiptParser` is using `.init`.
In order to expose that we'd need to `override` `init` from an `extension` because each (RC / `ReceiptParser`) `target` needs a different `LoggerType`. However,  `NSObject.init` can't be overridden from an extension.

That's why the API exposes `default` instead. This will make the usage clearer:

Screenshot 2023-01-05 at 10 22 16


NachoSoto:
I think `APP_STORE_CONNECT_API_KEY_KEY` was recreated incorrectly.

aboedo (approved):

PR #2199, opened 2023-01-05
`PostReceiptDataOperation`: replaced receipt `base64` with `hash` for cache key

### LINKED ISSUE #2198: App Hang in `PostReceiptDataOperation`

Creating a [new issue per](https://github.com/RevenueCat/purchases-ios/issues/1877#issuecomment-1372606167) @NachoSoto 

**Describe the bug**
I have a React Native app that uses the react-native-purchases library, which I understand depends on this one. One of our testers is seeing frequent app hangs & terminations that all seem to indicate an error with `PostReceiptDataOperation`. My theory is that this tester has "purchased" our subscription so many times creating test accounts that there is a large volume of transactions to be downloaded that is hanging the app. It's worth noting that this is still happening even after I created a test build that removes all explicit calls to [getCustomerInfo()](https://revenuecat-docs.netlify.app/documentation/revenuecat/purchases/getcustomerinfo(completion:)), so this seems to be called when all I'm doing is setting up RevenueCat at app initialization, but not even making any explicit API calls. Removing all RevenueCat code & dependencies from my app prevents the hangs & associated crashes for this tester.

@NachoSoto has proposed that this is a string interpolation issue related to bad data from javascript. I use `Purchases.logIn(user.id)` from `react-native-purchases` and my user ids are UUIDs, ex: `4f5d5640-264d-4da2-82ea-ab4d6264f7bc`. Or maybe `appUserId` is something else? I know that the [React Native setup instructions](https://www.revenuecat.com/docs/getting-started#%EF%B8%8F-initialize-and-configure-the-sdk) don't mention it, though other SDK setup instructions (ie: Swift) do...

1. Environment
   1. Platform: iOS
   2. SDK versions:
       * react-native: 0.70.6
       * react-native-purchases: 5.4.0
       * PurchasesHybridCommon pod: 4.5.4
       * RevenueCat pod: 4.15.5
   3. StoreKit 2 (_disabled with `useStoreKit2IfEnabled(false)`_) (Y/N): N
   4. OS version: 16.1.1
   5. Xcode version: 14
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue - I can't provide these since it only occurs on a Testflight user's real device

**Additional context**
Representative crash report (all point to this same series of RevenueCat library calls):
```
Hardware Model:     iPhone13,2
Process:            Bakesy
Identifier:         app.bakesy.internal
Version:            2.3.6
Role:               Foreground
OS Version:         iOS 16.1.1

App Hang: The app was terminated while unresponsive

0  libsystem_pthread.dylib +0xba8 _thread_chkstk_darwin
1  libswiftCore.dylib +0x232b30   String.UTF8View._foreignSubscript(position:)
2  libswiftCore.dylib +0x2191ac   specialized __StringStorage.appendInPlace<A>(_:isASCII:)
3  libswiftCore.dylib +0x218e00   _StringGuts._foreignAppendInPlace(_:)
4  Bakesy +0x3c82cc               DefaultStringInterpolation.write(_:) (<compiler-generated>)
5  Bakesy +0x3408f0               static PostReceiptDataOperation.createFactory(configuration:postData:customerInfoResponseHandler:customerInfoCallbackCache:) (CustomerInfoResponseHandler.swift)
6  Bakesy +0x3f7678               Backend.post(receiptData:appUserID:isRestore:productData:presentedOfferingIdentifier:observerMode:initiationSource:subscriberAttributes:completion:) (Backend.swift:109:23)
7  Bakesy +0x3fd14c               closure #1 in PurchasesOrchestrator.fetchProductsAndPostReceipt(withTransaction:receiptData:initiationSource:storefront:) (PurchasesOrchestrator.swift:869:22)
8  Bakesy +0x3f17d8               closure #1 in closure #2 in PurchasesOrchestrator.products(withIdentifiers:completion:) (PurchasesOrchestrator.swift:189:17)
9  Bakesy +0x82198                thunk for @escaping @callee_guaranteed () -> () (<compiler-generated>)
10 libdispatch.dylib +0x24b0      __dispatch_call_block_and_release
11 libdispatch.dylib +0x3fd8      __dispatch_client_callout
12 libdispatch.dylib +0x127f0     __dispatch_main_queue_drain
13 libdispatch.dylib +0x12440     __dispatch_main_queue_callback_4CF
14 CoreFoundation +0x9a6f4        ___CFRUNLOOP_IS_SERVICING_THE_MAIN_DISPATCH_QUEUE__
15 CoreFoundation +0x7c054        ___CFRunLoopRun
16 CoreFoundation +0x80ed0        _CFRunLoopRunSpecific
17 GraphicsServices +0x1364       _GSEventRunModal
18 UIKitCore +0x3a23cc            -[UIApplication _run]
19 UIKitCore +0x3a2030            _UIApplicationMain
20 Bakesy +0x6cb8                 main (main.m:8:12)
21 dyld +0x1595c                  start
```

NachoSoto:
Should fix #2198 and [SDKONCALL-200]. This is a big performance optimization in time and memory, as we no longer need to convert the entire receipt to `base64` and concatenate it to the string, which is what was likely causing #2198.
This now uses `Data.hashString`, which avoids any data copies and efficiently creates a hash with the receipt content.

Apart from the new tests for this property, `BackendPostReceiptDataTests.testDoesntCacheForDifferentReceipts` ensures that different receipts have different cache keys, and therefore this new implementation still avoids collisions (other than the extremely unlikely hash collisions).

[SDKONCALL-200]: https://revenuecats.atlassian.net/browse/SDKONCALL-200?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/FoundationExtensions/Data+Extensions.swift
+++ b/Sources/FoundationExtensions/Data+Extensions.swift
@@ -68,8 +68,8 @@ extension Data {
             return Self.hexString(sha256.finalize().makeIterator())
         } else {
             let hash = self.withUnsafeBytes { (bytes: UnsafeRawBufferPointer) -> [UInt8] in
-                var hash = [UInt8](repeating: 0, count: Int(CC_SHA1_DIGEST_LENGTH))
-                CC_SHA1(bytes.baseAddress, CC_LONG(self.count), &hash)
+                var hash = [UInt8](repeating: 0, count: Int(CC_SHA256_DIGEST_LENGTH))
+                CC_SHA256(bytes.baseAddress, CC_LONG(self.count), &hash)
                 return hash
             }
 
--- a/Sources/FoundationExtensions/Data+Extensions.swift
+++ b/Sources/FoundationExtensions/Data+Extensions.swift
@@ -41,6 +41,19 @@ extension Data {
         return (self as NSData).asString()
     }
 
+    /// - Returns: `UUID` from the first 16 bytes of the underlying data.
+    var uuid: UUID? {
+        /// This implementation is equivalent to `return NSUUID(uuidBytes: [UInt8](self)) as UUID`
+        /// but ensures that the `Data` isn't unnecessarily copied in memory.
+        return self.withUnsafeBytes {
+            guard let baseAddress = $0.bindMemory(to: UInt8.self).baseAddress else {
+                return nil
+            }
+
+            return NSUUID(uuidBytes: baseAddress) as UUID
+        }
+    }
+
     /// - Returns: a string representing a fetch token.
     var asFetchToken: String {
         return self.base64EncodedString()
--- a/Sources/FoundationExtensions/Data+Extensions.swift
+++ b/Sources/FoundationExtensions/Data+Extensions.swift
@@ -13,6 +13,8 @@
 //  Created by Josh Holtz on 6/28/21.
 //
 
+import CommonCrypto
+import CryptoKit
 import Foundation
 
 extension NSData {
@@ -44,17 +46,26 @@ extension Data {
         return self.base64EncodedString()
     }
 
-    /// Creates an `UUID` from the underlying bytes.
-    var uuid: UUID? {
-        /// This implementation is equivalent to `return NSUUID(uuidBytes: [UInt8](self)) as UUID`
-        /// but ensures that the `Data` isn't unnecessarily copied in memory.
-        return self.withUnsafeBytes {
-            guard let baseAddress = $0.bindMemory(to: UInt8.self).baseAddress else {
-                return nil
+    /// - Returns: a hash representation of the underlying bytes.
+    var hashString: String {
+        if #available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *) {
+            var sha256 = SHA256()
+            sha256.update(data: self)
+
+            return Self.hexString(sha256.finalize().makeIterator())
+        } else {
+            let hash = self.withUnsafeBytes { (bytes: UnsafeRawBufferPointer) -> [UInt8] in
+                var hash = [UInt8](repeating: 0, count: Int(CC_SHA1_DIGEST_LENGTH))
+                CC_SHA1(bytes.baseAddress, CC_LONG(self.count), &hash)
+                return hash
             }
 
-            return NSUUID(uuidBytes: baseAddress) as UUID
+            return Self.hexString(hash.makeIterator())
         }
+    }
+
+    private static func hexString(_ iterator: Array<UInt8>.Iterator) -> String {
+        return iterator.map { String(format: "%02x", $0) }.joined()
     }
 
 }
--- a/Sources/FoundationExtensions/Data+Extensions.swift
+++ b/Sources/FoundationExtensions/Data+Extensions.swift
@@ -78,7 +78,10 @@ extension Data {
     }
 
     private static func hexString(_ iterator: Array<UInt8>.Iterator) -> String {
-        return iterator.map { String(format: "%02x", $0) }.joined()
+        return iterator
+            .lazy
+            .map { String(format: "%02x", $0) }
+            .joined()
     }
 
 }
--- a/Sources/FoundationExtensions/Data+Extensions.swift
+++ b/Sources/FoundationExtensions/Data+Extensions.swift
@@ -31,11 +31,6 @@ extension NSData {
         return deviceTokenString as String
     }
 
-    var uuid: UUID? {
-        let bytes = [UInt8](self)
-        return NSUUID(uuidBytes: bytes) as UUID
-    }
-
 }
 
 extension Data {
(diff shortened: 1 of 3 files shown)


Tests/UnitTests/FoundationExtensions/NSData+RCExtensionsTests.swift
```diff
@@ -29,14 +29,29 @@ class NSDataExtensionsTests: TestCase {
...
 
         expect(fetchToken).to(equal(storedReceiptText))
         expect(storedReceiptData.asFetchToken).to(equal(storedReceiptText))
     }
+
+    func testStringDataAsUUID() {
+        expect("sample string".asData.uuid) == UUID(uuidString: "73616D70-6C65-2073-7472-696E67000000")
+    }
+
+    func testDataAsUUIDHashesTheEntireData() {
```
NachoSoto, line 44: Hold on this new test fails ⚠️ 
I need to use an actual hashing algorithm. 

I luckily realized this was wrong by looking into the implementation of this method: https://opensource.apple.com/source/CF/CF-855.17/CFUUID.c.auto.html

Tests/UnitTests/FoundationExtensions/NSData+RCExtensionsTests.swift
```diff
@@ -29,14 +29,29 @@ class NSDataExtensionsTests: TestCase {
...
 
         expect(fetchToken).to(equal(storedReceiptText))
         expect(storedReceiptData.asFetchToken).to(equal(storedReceiptText))
     }
+
+    func testStringDataAsHashString() {
+        expect("sample string".asData.hashString) == "99ad9154f94977dd8913f3b7ea14091d00e52b8931c2bc1cfc7ea62b7c26727b"
+    }
+
+    func testDataAsHashStringHashesTheEntireData() {
```
NachoSoto, line 44: This test failed with the original `UUID` implementation, which only hashed the first 16 bytes (see https://opensource.apple.com/source/CF/CF-855.17/CFUUID.c.auto.html).

Sources/FoundationExtensions/Data+Extensions.swift
```diff
@@ -44,13 +41,44 @@ extension Data {
         return (self as NSData).asString()
     }
 
-    // Returns a string representing a fetch token.
+    /// - Returns: `UUID` from the first 16 bytes of the underlying data.
```
NachoSoto, line 44: I added this documentation to specify that this doesn't encode the entire data!

Sources/FoundationExtensions/Data+Extensions.swift
```diff
@@ -44,13 +41,44 @@ extension Data {
...
+    var uuid: UUID? {
+        /// This implementation is equivalent to `return NSUUID(uuidBytes: [UInt8](self)) as UUID`
+        /// but ensures that the `Data` isn't unnecessarily copied in memory.
+        return self.withUnsafeBytes {
+            guard let baseAddress = $0.bindMemory(to: UInt8.self).baseAddress else {
+                return nil
+            }
+
+            return NSUUID(uuidBytes: baseAddress) as UUID
+        }
```
NachoSoto, line 54: This is also a small optimization, to ensure that we don't copy `self`.

Sources/Networking/Operations/PostReceiptDataOperation.swift
```diff
@@ -41,7 +41,7 @@ final class PostReceiptDataOperation: CacheableNetworkOperation {
     ) -> CacheableNetworkOperationFactory<PostReceiptDataOperation> {
         let cacheKey =
         """
-        \(configuration.appUserID)-\(postData.isRestore)-\(postData.receiptData.asFetchToken)
+        \(configuration.appUserID)-\(postData.isRestore)-\(postData.receiptData.hashString)
```
NachoSoto, line 44: ✅ This is the core of this change

aboedo: this is a fantastic solution :chefskiss:
let's add a comment here explaining that we're explicitly using the `hashString` since the `fetchToken` could be huge if there are a lot of transactions in the receipt

aboedo (approved):

Sources/Networking/Operations/PostReceiptDataOperation.swift
NachoSoto: Good idea, I added this comment:
```swift
/// Cache key comprises of the following:
/// - `appUserID`
/// - `isRestore`
/// - Receipt (`hashString` instead of `fetchToken` to avoid big receipts leading to a huge cache key)
/// - `ProductRequestData.cacheKey`
/// - `presentedOfferingIdentifier`
/// - `observerMode`
/// - `subscriberAttributesByKey`
```

PR #2202, opened 2023-01-10
`ReceiptFetcher`: refactored implementation to log error when failing to fetch receipt

NachoSoto:
### Changes:

- Added `Optional.orThrow` to simplify implementation (this will be useful in many other parts of the code too)
- Added missing `OptinalExtensionsTests`
`DefaultFileReader` was implemented using `try?`, now it throws an error instead of returning `nil`
- `ReceiptFetcher`: logging specific error when loading receipt fails
- Refactored `ReceiptFetcher`:
  - No longer logging "loaded receipt" if the content was empty
  - No longer logging an error twice if the receipt failed to load
  - Simplified logic to handle errors only once


tonidero (approved):
LGTM!

PR #2204, opened 2023-01-10
Added new `ReceiptParser.fetchAndParseLocalReceipt`

NachoSoto:
Fixes [CSDK-622].

### Changes:
- Extracted `FileReader` to be able to use it in `ReceiptParser`
- Extracted `MockBundle` to be able to use it in `ReceiptParser`
- Added new `public` `ReceiptParser.fetchAndParseLocalReceipt()`, **only to `ReceiptParser`** (this is not available through `RevenueCat`).
- Added tests for the new API
- Added new API to `ReceiptParserAPITester`

Initially I thought about moving `ReceiptFetcher` to the `Receipt|Parser` target. Unfortunately that would require moving a whole bunch of other dependencies (`OperationDispatcher`, etc, etc).
We have no easy way to create a third "Core" framework in which to place these common dependencies, so that wouldn't be trivial.

This solution adds a much simpler version of `ReceiptFetcher` that simply reads from `Bundle.appStoreReceiptURL`. As it's documented there, under normal circumstances `SKReceiptRefreshRequest` is not required, so this implementation should work for 99% of use cases, and so this way we avoid increasing the.codebase complexity unnecessarily.

[CSDK-622]: https://revenuecats.atlassian.net/browse/CSDK-622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/LocalReceiptParsing/ReceiptParser-only-files/PurchasesReceiptParser+Fetching.swift
+++ b/Sources/LocalReceiptParsing/ReceiptParser-only-files/PurchasesReceiptParser+Fetching.swift
@@ -1,0 +1,60 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  ReceiptFetcher.swift
+//
+//  Created by Nacho Soto on 1/10/23.
+
+import Foundation
+
+public extension PurchasesReceiptParser {
+
+    // Note: this is a simplified version of `ReceiptFetcher`
+    // available for public use.
+
+    /// Fetches and parses the local receipt
+    /// - Returns: ``AppleReceipt`` of the parsed local receipt.
+    /// - Throws: ``PurchasesReceiptParser/Error`` if fetching or parsing failed.
+    ///
+    /// - Note: this method won't use ``SKReceiptRefreshRequest`` to
+    ///  fetch the receipt if it's not already available.
+    ///
+    /// ### Related Symbols
+    /// - ``SKReceiptRefreshRequest``
+    /// - ``Bundle/appStoreReceiptURL``
+    /// - ``PurchasesReceiptParser/parse(from:)``
+    func fetchAndParseLocalReceipt() throws -> AppleReceipt {
+        return try self.fetchAndParseLocalReceipt(reader: DefaultFileReader(),
+                                                  bundle: .main)
+    }
+
+    internal func fetchAndParseLocalReceipt(
+        reader: FileReader,
+        bundle: Bundle
+    ) throws -> AppleReceipt {
+        return try self.parse(from: self.fetchReceipt(reader, bundle))
+    }
+
+}
+
+private extension PurchasesReceiptParser {
+
+    func fetchReceipt(_ reader: FileReader, _ bundle: Bundle) throws -> Data {
+        guard let url = bundle.appStoreReceiptURL else {
+            throw Error.receiptNotPresent
+        }
+
+        do {
+            return try reader.contents(of: url)
+        } catch {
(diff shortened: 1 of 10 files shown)


Tests/ReceiptParserTests/ReceiptParserFetchingTests.swift
```diff
@@ -17,4 +17,117 @@ import XCTest
...
+    }
+
+    func testParseWithNilURLThrowsError() throws {
+        self.mockBundle.receiptURLResult = .nilURL
+
+        do {
+            _ = try self.fetchAndParse()
+            fail("Expected error")
+        } catch {
+            expect(error).to(matchError(PurchasesReceiptParser.Error.receiptNotPresent))
```
tonidero, line 54: I noticed here we are not using the same system as in the other 2 tests, where we catch the specific test we expect, and then a final catch to throw an error. Is there a reason for that?

Sources/LocalReceiptParsing/ReceiptParser-only-files/PurchasesReceiptParser+Fetching.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    // Note: this is a simplified version of `ReceiptFetcher`
+    // available for public use.
+
+    /// Fetches and parses the local receipt
+    /// - Returns: ``AppleReceipt`` of the parsed local receipt.
+    /// - Throws: ``PurchasesReceiptParser/Error`` if fetching or parsing failed.
+    ///
+    /// - Note: this method won't use ``SKReceiptRefreshRequest`` to
+    ///  fetch the receipt if it's not already available.
```
tonidero, line 26: Well I guess if missing this can cause issues in the functionality of the ReceiptParser library, it's possible we will get some questions from users using this... But I think it's ok to go with this for now and deal with it in the future if we are asked about this a lot.

tonidero (approved):
LGTM!

Sources/LocalReceiptParsing/ReceiptParser-only-files/PurchasesReceiptParser+Fetching.swift
aboedo: yeah, bummer that we don't the ability to refresh for now, but we can figure it out if we do get actual requests for it

Tests/ReceiptParserTests/ReceiptParserFetchingTests.swift
aboedo: +1, this seems like a nicer way to check for the error being the right type

aboedo (approved):

Sources/LocalReceiptParsing/ReceiptParser-only-files/PurchasesReceiptParser+Fetching.swift
NachoSoto: Agreed!

Tests/ReceiptParserTests/ReceiptParserFetchingTests.swift
NachoSoto: This approach produces a better error when it fails and it's one line instead of 2 different catches.

In the other test we're checking with a more specific error with an associated value which we wouldn't be able to do with `matchError`.

aboedo: ohhh the associated value, I missed that. What about `testParseWithNoReceiptThrowsError`?

NachoSoto: Oh yeah that can be simplified. Thanks!

NachoSoto: Oh I see now, that one has an associated value too: `case failedToLoadLocalReceipt(Swift.Error)`

Issue #2207, opened 2023-01-12
Incorrect `StoreProduct.price` in StoreKit 2


sochalewski:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
`StoreProduct.price` sometimes (depending on price) has wrong float precision when used with StoreKit 2. The issue does not exist when StoreKit 2 is disabled.

1. Environment
   1. Platform: iOS
   2. SDK version: 4.16.0
   3. StoreKit 2 (_disabled with `useStoreKit2IfEnabled(false)`_) (Y/N): Y
   4. OS version: 16.2
   5. Xcode version: 14.2
   6. How widespread is the issue. Percentage of devices affected. 100%
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue: N/A
3. Steps to reproduce, with a description of expected vs. actual behavior
Enable StoreKit 2 (`.with(usesStoreKit2IfAvailable: true)`), have a $23.99 product and use Polish locale (114.99 PLN).
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

StoreKit 2
```
(lldb) po priceLocale
pl_PL@currency=PLN (fixed)
(lldb) po price
114.98999999999997952 // <- THIS
(lldb) po priceDecimalNumber
114.98999999999997952 // <- THIS
(lldb) po localizedPriceString
114,99 zł
```

StoreKit
```
(lldb) po priceLocale
pl_PL@currency=PLN (fixed)
(lldb) po price
114.99
(lldb) po priceDecimalNumber
114.99
(lldb) po localizedPriceString
114,99 zł
```

**Additional context**
I'm aware that `localizedPriceString` is a preferred way over `price` to retrieve a product's price, but either way `price` should work exactly the same whether StoreKit 2 is enabled or not.

RCGitBot:
👀 [SDKONCALL-204] We've just linked this issue to our internal tracker and notified the team. Thank you for reporting, we're checking this out!

[...]

PR #2208, opened 2023-01-12
fix StoreKit 2 decimal rounding

aboedo:
This is an ugly solution meant as starter code to illustrate a fix for #2207 

It seems like StoreKit 2's decimals for prices aren't rounded like you'd expect, leading to this issue which makes for ugly paywalls. 

This adds an extra step of rounding in a price-friendly way.

### DIFF
--- a/Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
@@ -16,6 +16,13 @@ import StoreKit
 @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
 internal struct SK2StoreProduct: StoreProductType {
 
+    private static let decimalRoundingBehavior = NSDecimalNumberHandler(roundingMode: .up,
+                                                                        scale: 2,
+                                                                        raiseOnExactness: false,
+                                                                        raiseOnOverflow: false,
+                                                                        raiseOnUnderflow: false,
+                                                                        raiseOnDivideByZero: false)
+
     init(sk2Product: SK2Product) {
         #if swift(<5.7)
         self._underlyingSK2Product = sk2Product
@@ -58,7 +65,12 @@ internal struct SK2StoreProduct: StoreProductType {
         return offers?.first?["currencyCode"] as? String
     }
 
-    var price: Decimal { underlyingSK2Product.price }
+    var price: Decimal {
+        // See: https://github.com/RevenueCat/purchases-ios/issues/2207
+        // it seems like StoreKit 2 doesn't actually round decimals in a price-friendly fashion, so we do it
+        let priceNSDecimal = (underlyingSK2Product.price as NSDecimalNumber)
+        return priceNSDecimal.rounding(accordingToBehavior: Self.decimalRoundingBehavior) as Decimal
+    }
 
     var localizedPriceString: String { underlyingSK2Product.displayPrice }
 
--- a/Sources/Purchasing/StoreKitAbstractions/StoreProductDiscount.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/StoreProductDiscount.swift
@@ -29,6 +29,13 @@ public typealias SK2ProductDiscount = StoreKit.Product.SubscriptionOffer
 @objc(RCStoreProductDiscount)
 public final class StoreProductDiscount: NSObject, StoreProductDiscountType {
 
+    private static let decimalRoundingBehavior = NSDecimalNumberHandler(roundingMode: .up,
+                                                                        scale: 2,
+                                                                        raiseOnExactness: false,
+                                                                        raiseOnOverflow: false,
+                                                                        raiseOnUnderflow: false,
+                                                                        raiseOnDivideByZero: false)
+
     /// The payment mode for a `StoreProductDiscount`
     /// Indicates how the product discount price is charged.
     @objc(RCPaymentMode)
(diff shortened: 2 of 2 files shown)


Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
```diff
@@ -58,7 +65,12 @@ internal struct SK2StoreProduct: StoreProductType {
         return offers?.first?["currencyCode"] as? String
     }
 
-    var price: Decimal { underlyingSK2Product.price }
+    var price: Decimal {
+        // See: https://github.com/RevenueCat/purchases-ios/issues/2207
+        // it seems like StoreKit 2 doesn't actually round decimals in a price-friendly fashion, so we do it
```
aboedo, line 70: See the words from the [StoreKit whisperer](https://stackoverflow.com/a/47445051/5074358)

Sources/Purchasing/StoreKitAbstractions/StoreProductDiscount.swift
```diff
@@ -29,6 +29,13 @@ public typealias SK2ProductDiscount = StoreKit.Product.SubscriptionOffer
 @objc(RCStoreProductDiscount)
 public final class StoreProductDiscount: NSObject, StoreProductDiscountType {
 
+    private static let decimalRoundingBehavior = NSDecimalNumberHandler(roundingMode: .up,
+                                                                        scale: 2,
+                                                                        raiseOnExactness: false,
+                                                                        raiseOnOverflow: false,
+                                                                        raiseOnUnderflow: false,
+                                                                        raiseOnDivideByZero: false)
```
aboedo, line 37: yeah, not exactly clean to have this duplicated for both classes, I was going for "quick PR to show off a possible solution"

Sources/Purchasing/StoreKitAbstractions/StoreProductDiscount.swift
```diff
@@ -70,7 +77,13 @@ public final class StoreProductDiscount: NSObject, StoreProductDiscountType {
     @objc public var offerIdentifier: String? { self.discount.offerIdentifier }
     @objc public var currencyCode: String? { self.discount.currencyCode }
     // See also `priceDecimalNumber` for Objective-C
-    public var price: Decimal { self.discount.price }
+    public var price: Decimal {
+        // See: https://github.com/RevenueCat/purchases-ios/issues/2207
+        // it seems like StoreKit 2 doesn't actually round decimals in a price-friendly fashion, so we do it
+        let priceNSDecimal = (self.discount.price as NSDecimalNumber)
```
aboedo, line 83: ended up going back and forth to objc here, I bet there's a nicer solution

Sources/Purchasing/StoreKitAbstractions/StoreProductDiscount.swift
```diff
@@ -70,7 +77,13 @@ public final class StoreProductDiscount: NSObject, StoreProductDiscountType {
...
+        return priceNSDecimal.rounding(accordingToBehavior: Self.decimalRoundingBehavior) as Decimal
```
aboedo, line 84: for writing tests for this, we could try to reproduce using SKTest. There's a chance that SKTest doesn't actually reproduce, though, and that we end up having to just mock with unit tests.

Which... we can't do directly because SK2 so we'd have to move this logic to some other class or something and test that one instead, then inject it here and test that we're calling the class.

aboedo:
@NachoSoto would you mind taking it from here?

Sources/Purchasing/StoreKitAbstractions/StoreProductDiscount.swift
aboedo: here's code that I used to test that the approach works: 

```swift
let price = 114.98999999999997952

print(price)

let priceDecimal = NSDecimalNumber(floatLiteral: price)

print(priceDecimal)

let behavior = NSDecimalNumberHandler(roundingMode: .up, scale: 2, raiseOnExactness: false, raiseOnOverflow: false, raiseOnUnderflow: false, raiseOnDivideByZero: false)

let roundedDecimal = priceDecimal.rounding(accordingToBehavior: behavior)

let swiftDecimal = Decimal(floatLiteral: price)

print(swiftDecimal)

let roundedSwift = (swiftDecimal as NSDecimalNumber).rounding(accordingToBehavior: behavior)

let finalDecimal = roundedSwift as Decimal

print(finalDecimal)

```

NachoSoto: Based on this I think maybe the best solution is to extract the rounding so we can test it like this, instead of bothering with StoreKit.

Sources/Purchasing/StoreKitAbstractions/StoreProductDiscount.swift
NachoSoto: Hmmm yeah let me look at these.

NachoSoto:
Sure!

NachoSoto (changes requested):
The more I think about this, I don't think we should modify the actual value. `Decimal` is essentially a `Float` with a fancy type.
The fact that printing it shows `114.98999999999997952` is a typical consequence of binary float number representation.
`Decimal.description` doesn't say anything about formatting.

This is why we format it explicitly using `NSDecimalNumber` [when posting it to the backend](https://github.com/RevenueCat/purchases-ios/blob/4.16.0/Sources/Purchasing/ProductRequestData.swift#L87-L88).

```swift
NSDecimalNumber(floatLiteral: 114.98999999999997952) == NSDecimalNumber(floatLiteral: 114.99) // true
```

So in other words, we shouldn't modify the underlying `float` value. What if one of the locales uses a precision different than 2 decimals? We'd be incorrectly rounding the value.

aboedo:
> So in other words, we shouldn't modify the underlying float value. What if one of the locales uses a precision different than 2 decimals? We'd be incorrectly rounding the value.

Ooff, yeah, I forgot about that, good catch. There are a few currencies that use 3 decimals

NachoSoto:
I would say the issue is a won't-fix. Doing `p price` is relying on LLDB to format a `Float`, which is not how one would format a price in the UI, and we do offer different ways to format prices already.

PR #2203, opened 2023-01-10
Added `ReceiptParserTests`

NachoSoto:
For [CSDK-622], we need to be able to write tests specific to `ReceiptParser` implementations.
This sets up that target, and allows running those tests via Xcode or SPM.

### Changes:

- Added `ReceiptParserTests` target to Xcode
- Added `ReceiptParserTests` target to `Package.swift`
- Added `ReceiptParserTests` to "test" phase of `ReceiptParser` scheme
- Added `ReceiptParserTests` to All-Tests test plan
- Added `ReceiptParserTests` to CI-All-Tests test plan

[CSDK-622]: https://revenuecats.atlassian.net/browse/CSDK-622?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2211, opened 2023-01-12
`PurchaseTester`: added section to visualize `AppleReceipt`

NachoSoto:
Fixes [CSDK-626].

Depends on #2199, #2200, #2202, #2203, #2204, #2210.

Simulator Screen Shot - iPhone 14 Pro Max - 2023-01-12 at 09 59 26

[CSDK-626]: https://revenuecats.atlassian.net/browse/CSDK-626?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcodeproj/project.pbxproj
+++ b/Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcodeproj/project.pbxproj
@@ -940,7 +940,7 @@
 			repositoryURL = "https://github.com/RevenueCat/purchases-ios";
 			requirement = {
 				kind = upToNextMajorVersion;
-				minimumVersion = 4.14.0;
+				minimumVersion = 4.17.0;
 			};
 		};
 		577E0E3B29159A8B0071E063 /* XCRemoteSwiftPackageReference "purchases-ios" */ = {
--- a/Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcodeproj/project.pbxproj
+++ b/Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcodeproj/project.pbxproj
@@ -35,7 +35,6 @@
 		578DAA232947DD8C001700FD /* Core.h in Headers */ = {isa = PBXBuildFile; fileRef = 578DAA222947DD8C001700FD /* Core.h */; settings = {ATTRIBUTES = (Public, ); }; };
 		578DAA262947DD8C001700FD /* Core.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 578DAA202947DD8C001700FD /* Core.framework */; platformFilters = (ios, maccatalyst, macos, ); };
 		578DAA272947DD8C001700FD /* Core.framework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = 578DAA202947DD8C001700FD /* Core.framework */; platformFilters = (ios, maccatalyst, macos, ); settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
-		578DAA2C2947DD90001700FD /* RevenueCat in Frameworks */ = {isa = PBXBuildFile; productRef = 578DAA2B2947DD90001700FD /* RevenueCat */; };
 		578DAA2D2947DE37001700FD /* ConfiguredPurchases.swift in Sources */ = {isa = PBXBuildFile; fileRef = 57ED6A83290891AF009580C6 /* ConfiguredPurchases.swift */; };
 		578DAA2E2947DE59001700FD /* Logger.swift in Sources */ = {isa = PBXBuildFile; fileRef = 575642A1290C78DD00719219 /* Logger.swift */; };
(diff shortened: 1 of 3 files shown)


Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcodeproj/project.pbxproj
```diff
@@ -907,6 +917,14 @@
 /* End XCConfigurationList section */
 
 /* Begin XCRemoteSwiftPackageReference section */
+		5759B47329707153002472D5 /* XCRemoteSwiftPackageReference "purchases-ios" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/RevenueCat/purchases-ios";
+			requirement = {
+				branch = "receipt-parser-fetcher";
```
NachoSoto, line 924: ⚠️ TODO: need to change this back to the version.

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/LocalReceiptView.swift
```diff
@@ -0,0 +1,251 @@
...
+                self.form
+                    .scrollContentBackground(.hidden)
+            } else {
+                self.form
+            }
+
+            Spacer(minLength: 10)
+
+            Button {
+                Purchases.shared.restorePurchases { _, error in
```
tonidero, line 33: Hmm I guess we need to restorePurchases before refreshing the receipt in order to get the latest receipt from apple correct? Maybe we should change the label in the button to "Refresh receipt from apple" or something like that? It might be kind of implicit so I'm ok leaving it as is as well.

tonidero (approved):

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/LocalReceiptView.swift
aboedo: they're technically entirely separate things - restorePurchases will attach the latest purchases to the current user, but refresh receipt will refresh the receipt on device. restorePurchases may actually already refresh the receipt (depending on the status of the receipt on device)

aboedo: but with regards to the label... I think it's fine to not explicitly say "from apple" since the receipt = apple, entitlementInfo = revenuecat

aboedo: not entirely sure why we restorePurchases here, though, is it to prevent inconsistencies where the current user doesn't have the purchases on the receipt associated? I think it might be better not to have that as a side effect of the button, so you can actually see if the customerInfo doesn't have all purchases on the receipt yet

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/LocalReceiptView.swift
```diff
@@ -0,0 +1,251 @@
...
+                    }
+                }
+            } label: {
+                Text("Refresh receipt")
+            }
+            .buttonStyle(.borderedProminent)
+        }
+        .navigationTitle("Local Receipt")
+        .task {
+            await self.refreshReceipt()
```
aboedo, line 49: do we need this to happen automatically? If we're doing this when we load the view, then I think the refresh button will be mostly irrelevant, since you'll only be able to modify the receipt by either waiting for a subs to expire or going through manage subs in App Store or Xcode. 

I'd vote for this to be manual just so that the debugging tool doesn't modify the environment unless you ask it to, but I'm more than open to feedback on this

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/LocalReceiptView.swift
NachoSoto: Oh I forgot to explain this so valid confusion. We don't have any public API to refresh the receipt other than this.
#2204 introduces a way to fetch it but without refreshing it in case it's not there yet, so we need this mostly for the simulator...
I agree it doesn't make a lot of sense to call `restorePurchases` here though. Any suggestions?

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/LocalReceiptView.swift
NachoSoto: This is probably a bad name. `refreshReceipt` simply loads it, but the button calls `restorePurchases` first to use `SKReceiptRefreshRequest` before loading it again. Does that make sense?

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/LocalReceiptView.swift
aboedo: oohhhh right, that makes sense. I don't have a problem with it, but we should add a comment to explain the reasoning

NachoSoto: Will do 🙌🏻

aboedo (approved):

NachoSoto:
Will update this to point to a new release that includes the new necessary API introduced in #2204.

PR #2213, opened 2023-01-12
`CustomerInfo`: support parsing schema version 2 to restore SDK `v3.x` compatibility

NachoSoto:
Fixes [TRIAGE-240].

When we converted `CustomerInfo` to use `Codable` (#1496), the schema version was also changed from 2 to 3.
I don't remember if this was done on purpose, but turns out that it was not necessary.

This means that a user opening an app while offline, after upgrading from version `3.x` to `4.x` (>= `4.3`), would not have a cached `CustomerInfo` and therefore no entitlements.
This fixes that by restoring support and adding a specific test with a sample version 2 JSON (which is virtually identical).

[TRIAGE-240]: https://revenuecats.atlassian.net/browse/TRIAGE-240?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Identity/CustomerInfo.swift
+++ b/Sources/Identity/CustomerInfo.swift
@@ -221,6 +221,7 @@ extension CustomerInfo {
     static let currentSchemaVersion = "3"
 
     private static let compatibleSchemaVersions: Set<String> = [
+        // Version 3 is virtually identical to 2 (only difference is `Codable` vs manual decoding).
         "2",
         CustomerInfo.currentSchemaVersion
     ]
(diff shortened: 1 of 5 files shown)


aboedo (approved):
If this is also the case for Android, could you fix it there as well?

NachoSoto:
@aboedo Android has a schema version of 3, but looks like it's been that since 2020, so I'm guessing the last time it changed was deliberate.

PR #2214, opened 2023-01-13
`HTTPClient`: added support for sending request `UUID`s

NachoSoto:
Fixes [CSDK-631].


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -63,9 +63,17 @@ class HTTPClient {
...
+        return [Self.authorizationHeaderName: "Bearer \(apiKey)"]
     }
 
+    static func signatureUUIDHeader(with uuid: UUID) -> RequestHeaders {
+        return [Self.signatureHeaderName: uuid.uuidString]
+    }
+
+    static let authorizationHeaderName = "Authorization"
+    // TODO: confirm this name
+    static let signatureHeaderName = "X-Request-UIUD"
```
aboedo, line 75: typo `UIUD` -> `UUID`

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -63,9 +63,17 @@ class HTTPClient {
...
     static func authorizationHeader(withAPIKey apiKey: String) -> RequestHeaders {
-        return ["Authorization": "Bearer \(apiKey)"]
+        return [Self.authorizationHeaderName: "Bearer \(apiKey)"]
     }
 
+    static func signatureUUIDHeader(with uuid: UUID) -> RequestHeaders {
+        return [Self.signatureHeaderName: uuid.uuidString]
+    }
+
+    static let authorizationHeaderName = "Authorization"
```
aboedo, line 73: 👍

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto: Well good thing I waited to confirm the name 🤣

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -63,9 +63,17 @@ class HTTPClient {
...
+    // TODO: confirm this name
+    static let signatureHeaderName = "X-Request-UUID"
```
tonidero, line 75: From looking at the backend PR, I believe they are using `X-Nonce`

tonidero (commented):
It's looking good!

Sources/Networking/HTTPClient/HTTPRequest.swift
```diff
@@ -11,11 +11,30 @@
...
 
+import Foundation
+
 /// A request to be made by ``HTTPClient``
 struct HTTPRequest {
 
     let method: Method
     let path: Path
+    /// If present, this will be used by the server to compute a checksum of the response signed with a private key.
+    let signatureUUID: UUID?
```
tonidero, line 22: Hmm well, from the backend PR, seems like the backend is expecting a B64 encoded 16 bytes size string. We could also talk with backend team to change it on their end if we have to

NachoSoto: Ok let me change that, it's easy.

NachoSoto: Done

Sources/FoundationExtensions/Data+Extensions.swift
```diff
@@ -85,3 +85,36 @@ extension Data {
...
+
+    private static let nonceLength = 16
+
+}
+
+private extension UUID {
+
+    private var asUInt8Array: [UInt8] {
+        // swiftlint:disable:next identifier_name
+        let (u1, u2, u3, u4, u5, u6, u7, u8, u9, u10, u11, u12, u13, u14, u15, u16) = self.uuid
```
tonidero, line 112: Interesting, I didn't know you could unpack a UUID like this. But it looks good 👍

tonidero (approved):

Sources/Networking/HTTPClient/HTTPRequest.swift
```diff
@@ -11,11 +11,33 @@
...
     let path: Path
+    /// If present, this will be used by the server to compute a checksum of the response signed with a private key.
+    let nonce: Data?
+
+    init(method: Method, path: Path) {
+        self.init(method: method, path: path, nonce: nil)
+    }
+
+    init(method: Method, path: Path, nonce: Data?) {
+        assert(nonce == nil || nonce?.count == Data.nonceLength)
```
NachoSoto, line 29: @tonidero @aboedo thoughts on this? `assert` only compiles on `DEBUG` builds so it can help us catch issues in tests. Sending anything other than 12 bytes would fail so might as well catch it early.

Sources/FoundationExtensions/Data+Extensions.swift
```diff
@@ -85,3 +85,14 @@ extension Data {
     }
 
 }
+
+extension Data {
+
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    static func randomNonce() -> Data {
+        return Data(ChaChaPoly.Nonce())
```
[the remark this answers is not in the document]
tonidero: Love the name 🤣

tonidero (approved):

Sources/Networking/HTTPClient/HTTPRequest.swift
NachoSoto: The pre-built frameworks are release, yeah. But also for people integrating with CocoaPods / SPM, their production builds are release by default.

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -325,6 +332,24 @@ private extension HTTPClient {
...
+    func headers(with authHeaders: HTTPClient.RequestHeaders) -> HTTPClient.RequestHeaders {
+        var result: HTTPClient.RequestHeaders = [:]
+
+        if self.path.authenticated {
+            result += authHeaders
+        }
+
+        if let nonce = self.nonce {
+            result += HTTPClient.nonceHeader(with: nonce)
+        }
```
aboedo, line 346: is there any situation where the nonce should be `nil` but the `path.authenticated = true`? 
If not, we should try to consolidate the conditions and probably log if they're not simultaneously met

Sources/Networking/HTTPClient/HTTPRequest.swift
```diff
@@ -11,11 +11,34 @@
...
+        self.method = method
+        self.path = path
+        self.nonce = nonce
+    }
+
+    /// Creates an `HTTPRequest` with a `nonce`.
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    static func createIntegrityEnforcedRequestRequest(method: Method, path: Path) -> Self {
+        return .init(method: method, path: path, nonce: Data.randomNonce())
+    }
```
aboedo, line 41: technically we're verifying integrity and authenticity and the same time, right? maybe we could have the name reflect that? 
also, typo here: `RequestRequest`

Sources/FoundationExtensions/Data+Extensions.swift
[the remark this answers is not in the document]
aboedo: I want all features to be named like that

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto: The `/health` endpoint is one example.

Sources/Networking/HTTPClient/HTTPRequest.swift
NachoSoto: > technically we're verifying integrity and authenticity and the same time, right? maybe we could have the name reflect that?

How about `HTTPRequest.createWithResponseVerification`?

> also, typo here: RequestRequest
Thanks 🤦🏻  I noticed that in the next PR, I'll fix it.

PR #2215, opened 2023-01-14
`Configuration`: added (`internal` for now) API to load public key

NachoSoto:
Fixes [CSDK-632].

### Changes:
- Created `Signing` to encapsulate new signing logic and types
- Added abstractions like `Signing.Certificate` and `Signing.PublicKey` over `Security` C types
- Created `Configuration.EntitlementVerificationLevel` and `Signing.ResponseVerificationLevel` (the logic for the name difference is that publicly this is used to validate entitlements, but the internal implementation works for any response).
- Created `Signing.loadPublicKey`
- Added tests for loading public key embedded in SDK
- Added `internal` (for now) `Configuration.with(entitlementVerificationLevel:)`
- Added `SystemInfo.responseVerificationLevel` with the loaded key, if enabled

### For a future PR:
- Make the `Configuration.with(entitlementVerificationLevel:)` `public`
- Make `Configuration.EntitlementVerificationLevel` `public`
- Add those 2 to `APITesters`
- Improve documentation by adding a link to the docs
- Verify certificate issuer in test

[CSDK-632]: https://revenuecats.atlassian.net/browse/CSDK-632?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


aboedo:
🤔 my thinking is that it would be useful for this to exist for testing purposes, but a customer would never need to override, right? What alternatives could we explore that don't modify the public API? 
Perhaps we could have a file name where if the file exists, it overrides the key, and we use that for testing? Of course, that would mean that if the bundle is modified to include such file, it would override the key, which would make it a potential attack. Maybe we could have that only work in debug / sandbox?

NachoSoto:
Yeah when I did this I didn't have full context.
Should we make that `Configuration` method `internal`? (and `DEBUG` only?)

NachoSoto:
This should be ready now, without any new `public` API.

NachoSoto:
Hold on, I think we need to change the key loading code since it appears we're using a different type of key.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1221,6 +1228,10 @@ internal extension Purchases {
         return self.userDefaults
     }
 
+    var publicKey: Signing.PublicKey? {
+        return self.systemInfo.responseVerificationLevel.publicKey
+    }
```
tonidero, line 1233: I guess this will be used in future tests?

Tests/StoreKitUnitTests/TestHelpers/AvailabilityChecks.swift
```diff
@@ -19,6 +19,12 @@ import XCTest
 // everything in those classes will still be called by XCTest, and it will cause errors.
 enum AvailabilityChecks {
 
+    static func iOS12APIAvailableOrSkipTest() throws {
```
tonidero, line 22: Also unused, but maybe you plan to use it for future tests?

Sources/Misc/SystemInfo.swift
```diff
@@ -112,6 +115,7 @@ class SystemInfo {
          operationDispatcher: OperationDispatcher = .default,
          bundle: Bundle = .main,
          storeKit2Setting: StoreKit2Setting = .default,
+         responseVerificationLevel: Signing.ResponseVerificationLevel = .disabled,
```
tonidero, line 118: Should this use `.default`?

Sources/Purchasing/Configuration.swift
```diff
@@ -84,6 +87,7 @@ import Foundation
         private(set) var networkTimeout = Configuration.networkTimeoutDefault
         private(set) var storeKit1Timeout = Configuration.storeKitRequestTimeoutDefault
         private(set) var platformInfo: Purchases.PlatformInfo?
+        private(set) var responseVerificationLevel: Signing.ResponseVerificationLevel = .disabled
```
tonidero, line 90: Same here, should this be `.default`?

tonidero (approved):
A couple small things but looking good!

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: Oh I moved these tests somewhere else, but let me add another one for `Purchases.Configuration` that uses this too.
I moved them out on a refactor and also when I made this `internal`, but they're still useful for when we make it `public`.

Tests/StoreKitUnitTests/TestHelpers/AvailabilityChecks.swift
NachoSoto: Oh not anymore, since this feature is iOS 13 only, I'll remove it!

Sources/Misc/SystemInfo.swift
NachoSoto: Yes!! Great catch.

Sources/Purchasing/Configuration.swift
NachoSoto: Another good catch 👍🏻

PR #2217, opened 2023-01-16
`SPM`: added `APPLICATION_EXTENSION_API_ONLY` flag to `RevenueCat` and `ReceiptParser`

NachoSoto:
This flag is already set in `Xcode`'s target, so setting it here matches it when embedding the frameworks through `SPM`.

Screenshot 2023-01-16 at 10 09 30

This allows users to embed these package in non-app targets.
Unfortunately setting the flag directly is the only way to do this at the moment.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,6 +16,13 @@ if shouldIncludeDocCPlugin {
     dependencies.append(.package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"))
 }
 
+// Equivalent to `APPLICATION_EXTENSION_API_ONLY`
+// (see https://xcodebuildsettings.com/#application_extension_api_only)
+// This is the only way to set this flag for SPM at the moment.
+// The Xcode targets have this flag as well.
+// It allows users to embed these frameworks in a non-app target without receiving warnings.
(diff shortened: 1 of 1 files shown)


NachoSoto:
Modified this to set it to `RevenueCat` too.

aboedo (approved):

Package.swift
```diff
@@ -32,8 +32,14 @@ let package = Package(
...
-                exclude: ["Info.plist", "LocalReceiptParsing/ReceiptParser-only-files"]),
+                exclude: ["Info.plist", "LocalReceiptParsing/ReceiptParser-only-files"],
+                linkerSettings: [
+                    .unsafeFlags(["-Xlinker", "-no_application_extension"])
+                ]),
         .target(name: "ReceiptParser",
-                path: "LocalReceiptParsing")
+                path: "LocalReceiptParsing",
+                linkerSettings: [
+                    .unsafeFlags(["-Xlinker", "-no_application_extension"])
```
aboedo, line 42: let's add a comment here explaining what this is so maintainers don't have to google it

aboedo: we could link to this https://xcodebuildsettings.com/#application_extension_api_only and explain the need

aboedo (approved):

Package.swift
NachoSoto: Done

PR #2222, opened 2023-01-18
Produce a compilation error when using an old `Xcode` version

NachoSoto:
This will make it easier to figure out that the version is no longer compatible.
Aditionally, I've added `UpdatingRequirements.md` so we have a single place to reference when changing these requirements.

### DIFF
--- a/Contributing/UpdatingRequirements.md
+++ b/Contributing/UpdatingRequirements.md
@@ -1,0 +1,21 @@
+### Build Requirements
+
+The current list of build requirements is listed in `README.md`:
+- Xcode version
+- Deployment targets
+
+https://xcodereleases.com can be used as a reference.
+
+### Places to Update
+
+If these requirements change, there are several places that need to be updated:
+- `README.md`
+- https://www.revenuecat.com/docs/getting-started
+- Xcode `Project` build settings (at the project level; targets inherit these settings)
+- `RevenueCat.podspec`
+- `Package.swift`
+- `SwiftVersionCheck.swift`
+
+### Last
+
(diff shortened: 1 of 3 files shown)


Contributing/UpdatingRequirements.md
```diff
@@ -0,0 +1,21 @@
...
+- Xcode version
+- Deployment targets
+
+https://xcodereleases.com can be used as a reference.
+
+### Places to Update
+
+If these requirements change, there are several places that need to be updated:
+- `README.md`
+- https://www.revenuecat.com/docs/getting-started
```
NachoSoto, line 13: [This one](https://www.revenuecat.com/docs/ios#requirements) is currently wrong: 
Screenshot 2023-01-18 at 08 01 45

Sources/Support/SwiftVersionCheck.swift
```diff
@@ -0,0 +1,19 @@
...
+//  SwiftVersionCheck.swift
+//
+//  Created by Nacho Soto on 1/18/23.
+
+import Foundation
+
+#if swift(<5.5.2)
+// See https://xcodereleases.com
+#error("RevenueCat requires Xcode 13.2 with Swift 5.5.2 to compile.")
+#endif
```
aboedo, line 19: ❤️

aboedo (approved):

Contributing/UpdatingRequirements.md
NachoSoto: Fixed by @aboedo, thanks!

PR #2224, opened 2023-01-18
`CustomerInfoManager`: improved thread-safety

### LINKED ISSUE #1877: Crash in `CustomerInfoManager`
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Development/CONTRIBUTING.md)

**Describe the bug**
Just now received one more crash in RevenueCat lib

1. Environment
   1. Platform: iOS
   2. SDK version: 4.10.2
   3. StoreKit 2 (_enabled with `useStoreKit2IfEnabled`_) (Y/N): Y
   4. OS version: iOS 15.6.1
   5. Xcode version: 13.4.1
   6. idk, its first crash here
2. -
3. App crashed on start, on becoming active. I think, after update to new version.
4. Stack trace:

```
Crashed: com.apple.NSURLSession-delegate
0  CookingRecipes                 0x217cc4 closure #1 in CustomerInfoManager.sendUpdateIfChanged(customerInfo:) + 4330765508 (<compiler-generated>:4330765508)
1  CookingRecipes                 0x217748 CustomerInfoManager.cache(customerInfo:appUserID:) + 28 (Lock.swift:28)
2  CookingRecipes                 0x215268 closure #1 in CustomerInfoManager.fetchAndCacheCustomerInfo(appUserID:isAppBackgrounded:completion:) + 53 (Logger.swift:53)
3  CookingRecipes                 0x259230 closure #2 in GetCustomerInfoOperation.getCustomerInfo(completion:) + 35 (CustomerInfoResponseHandler.swift:35)
4  CookingRecipes                 0x26767c specialized closure #1 in HTTPClient.Request.init<A>(httpRequest:headers:completionHandler:) + 4331091580
5  CookingRecipes                 0x269244 partial apply for specialized closure #1 in HTTPClient.Request.init<A>(httpRequest:headers:completionHandler:) + 4331098692 (<compiler-generated>:4331098692)
6  CookingRecipes                 0x24d030 HTTPClient.handle(urlResponse:request:urlRequest:data:error:) + 237 (HTTPClient.swift:237)
7  CookingRecipes                 0x24e340 partial apply for closure #1 in HTTPClient.start(request:) + 4330988352 (<compiler-generated>:4330988352)
8  CookingRecipes                 0x24d9d0 thunk for @escaping @callee_guaranteed (@guaranteed Data?, @guaranteed NSURLResponse?, @guaranteed Error?) ->
```

NachoSoto:
Likely fixes #1877.

Thread-safety was done through a `Lock`, but nothing ensured that we didn't access the underlying data without locking.
This is done now using `Atomic<Data>` to guarantee that all reads and writes go through this lock.

### DIFF
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -17,21 +17,22 @@ class CustomerInfoManager {
 
     typealias CustomerInfoCompletion = @MainActor @Sendable (Result<CustomerInfo, BackendError>) -> Void
 
-    private(set) var lastSentCustomerInfo: CustomerInfo?
+    var lastSentCustomerInfo: CustomerInfo? { return self.data.value.lastSentCustomerInfo }
+
     private let operationDispatcher: OperationDispatcher
-    private let deviceCache: DeviceCache
     private let backend: Backend
     private let systemInfo: SystemInfo
-    private let customerInfoCacheLock = Lock()
+    /// Underlying synchronized data.
+    private let data: Atomic<Data>
 
     init(operationDispatcher: OperationDispatcher,
          deviceCache: DeviceCache,
          backend: Backend,
          systemInfo: SystemInfo) {
         self.operationDispatcher = operationDispatcher
-        self.deviceCache = deviceCache
         self.backend = backend
         self.systemInfo = systemInfo
+        self.data = .init(.init(deviceCache: deviceCache))
     }
 
     func fetchAndCacheCustomerInfo(appUserID: String,
@@ -41,7 +42,7 @@ class CustomerInfoManager {
                                      withRandomDelay: isAppBackgrounded) { result in
             switch result {
             case let .failure(error):
-                self.deviceCache.clearCustomerInfoCacheTimestamp(appUserID: appUserID)
+                self.withData { $0.deviceCache.clearCustomerInfoCacheTimestamp(appUserID: appUserID) }
                 Logger.warn(Strings.customerInfo.customerinfo_updated_from_network_error(error))
 
             case let .success(info):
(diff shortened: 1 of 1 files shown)


Sources/Identity/CustomerInfoManager.swift
```diff
@@ -41,7 +42,7 @@ class CustomerInfoManager {
                                      withRandomDelay: isAppBackgrounded) { result in
             switch result {
             case let .failure(error):
-                self.deviceCache.clearCustomerInfoCacheTimestamp(appUserID: appUserID)
+                self.withData { $0.deviceCache.clearCustomerInfoCacheTimestamp(appUserID: appUserID) }
```
tonidero, line 45: Hmm I believe in this case the behavior would be the same and considering we are not modifying data in the `Data` struct, this is correct, but considering this is clearing a cache, would it semantically be better to use `self.modifyData`?

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -178,17 +183,17 @@ class CustomerInfoManager {
     func cache(customerInfo: CustomerInfo, appUserID: String) {
         do {
             let jsonData = try JSONEncoder.default.encode(customerInfo)
-            self.deviceCache.cache(customerInfo: jsonData, appUserID: appUserID)
+            self.withData { $0.deviceCache.cache(customerInfo: jsonData, appUserID: appUserID) }
```
tonidero, line 186: Same here about `withData` vs `modifyData`.

tonidero (approved):
Just a question, but looks good otherwise.

Sources/Identity/CustomerInfoManager.swift
NachoSoto: In terms of naming semantics I agree, but then we'd have to ignore this `inout` parameter which I think would make it more confusing.

PR #2228, opened 2023-01-18
`PurchaseTester`: ignore errors when restoring purchases

NachoSoto:
See #2211.

If `restorePurchases` fails for any reason, we still want to try to load the local receipt. We're only interested in refreshing the receipt to load it.

### DIFF
--- a/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/LocalReceiptView.swift
+++ b/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/LocalReceiptView.swift
@@ -30,14 +30,13 @@ struct LocalReceiptView: View {
             Spacer(minLength: 10)
 
             Button {
-                // There's no public API to refresh the receipt other than this.
-                // This is the simplest way to force a receipt refresh before fetching it again.
-                Purchases.shared.restorePurchases { _, _ in
+                Task<Void, Never> {
+                    // There's no public API to refresh the receipt other than this.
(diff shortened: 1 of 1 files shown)


Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/LocalReceiptView.swift
```diff
@@ -32,13 +32,11 @@ struct LocalReceiptView: View {
...
-                        self.receipt = .failure(error)
-                    } else {
-                        Task<Void, Never> {
-                            await self.refreshReceipt()
-                        }
+                Purchases.shared.restorePurchases { _, _ in
+                    // Ignoring error, since the receipt might have actually been refreshed
+                    // and that's all we care about here.
+                    Task<Void, Never> {
+                        await self.refreshReceipt()
```
aboedo, line 39: if we're ignoring the error, is it nicer if we do something like
```swift
Task<void, Never> { 
    try await Purchases.shared.restorePurchases()
    await self.refreshReceipt()
}
```
?

aboedo (approved):

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/LocalReceiptView.swift
NachoSoto: Oh good point.

Issue #300, opened 2020-07-30
Main Thread Checker: UI API called on a background thread: -[UIApplication applicationState]


JMCPH:
**Describe the bug**
There is an issue in the latest framework version. UI API was called on a background thread on the line RCSystemInfo.m - 87.

1. Environment
    1. Platform: iOS
    2. SDK version: 3.5.2
    3. OS version: 13.5.1
    4. Xcode version: 11.6
    5. How widespread is the issue. Percentage of devices affected: 100 %
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue

```
Main Thread Checker: UI API called on a background thread: -[UIApplication applicationState]
PID: 908, TID: 139231, Thread name: (none), Queue name: NSOperationQueue 0x10bc0fdf0 (QOS: UNSPECIFIED), QoS: 0
Backtrace:
4   Purchases                           0x0000000105611a9c -[RCSystemInfo isApplicationBackgrounded] + 88
5   Purchases                           0x00000001055ff748 -[RCPurchases initWithAppUserID:requestFetcher:receiptFetcher:attributionFetcher:backend:storeKitWrapper:notificationCenter:userDefaults:systemInfo:offeringsFactory:deviceCache:identityManager:subscriberAttributesManager:] + 1500
6   Purchases                           0x00000001055feffc -[RCPurchases initWithAPIKey:appUserID:userDefaults:observerMode:platformFlavor:platformFlavorVersion:] + 912
7   Purchases                           0x00000001055feb00 +[RCPurchases configureWithAPIKey:appUserID:observerMode:userDefaults:platformFlavor:platformFlavorVersion:] + 244
8   Purchases                           0x00000001055fe9c8 +[RCPurchases configureWithAPIKey:appUserID:observerMode:userDefaults:] + 176
9   Purchases                           0x00000001055fe8e0 +[RCPurchases configureWithAPIKey:appUserID:observerMode:] + 136
```

3. Steps to reproduce, with a description of expected vs. actual behavior
`Purchases.configure(withAPIKey: "XXXX", appUserID: userID, observerMode: true)`

4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

**Additional context**

JMCPH:
@aboedo can you please take a look at this? We are planning to start paying from next week, so we are really hoping you can take a quick look at this :)

aboedo:
[...]

PR #2230, opened 2023-01-18
`SystemInfo.isApplicationBackgrounded`: added overload for `@MainActor`

NachoSoto:
Unfortunately we can't use this in any of the places that use the asynchronous method, but as we move more code to `async` this might be beneficial.

For now this also improves the documentation to explain why the API is asynchronous.

### DIFF
--- a/Sources/Misc/SystemInfo.swift
+++ b/Sources/Misc/SystemInfo.swift
@@ -126,10 +126,25 @@ class SystemInfo {
             : BundleSandboxEnvironmentDetector(bundle: bundle)
     }
 
+    /// Asynchronous API if caller can't ensure that it's invoked in the `@MainActor`
+    /// - Seealso: `isApplicationBackgrounded`
     func isApplicationBackgrounded(completion: @escaping (Bool) -> Void) {
-        self.operationDispatcher.dispatchOnMainThread {
+        self.operationDispatcher.dispatchOnMainActor {
             completion(self.isApplicationBackgrounded)
         }
+    }
+
+    /// Synchronous API for callers in `@MainActor`.
+    /// - Seealso: `isApplicationBackgrounded(completion:)`
+    @MainActor
+    var isApplicationBackgrounded: Bool {
+    #if os(iOS) || os(tvOS)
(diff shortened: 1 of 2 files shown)


Sources/Misc/SystemInfo.swift
```diff
@@ -126,12 +126,27 @@ class SystemInfo {
...
     func isApplicationBackgrounded(completion: @escaping (Bool) -> Void) {
-        self.operationDispatcher.dispatchOnMainThread {
+        self.operationDispatcher.dispatchOnMainActor {
             completion(self.isApplicationBackgrounded)
         }
     }
 
+    /// Synchronous API for callers in `@MainActor`.
+    /// - Seealso: `isApplicationBackgrounded(completion:)`
+    @MainActor
```
aboedo, line 139: this seems like a good idea, it basically guarantees that we won't be able to call this on a thread other than main, right? Which would crash if we did, so now it fails in compilation instead

aboedo (approved):

Sources/Misc/SystemInfo.swift
NachoSoto: Exactly.
Well it wouldn't have crashed I don't think, but `UIApplication` isn't thread safe so it wouldn't have been safe.

aboedo: sorry you're right, not a crash, but this [is what I was thinking about](https://github.com/RevenueCat/purchases-ios/issues/300). I remembered there being issues with it

NachoSoto:
Amazing :D glad I wrote this:
Screenshot 2023-01-18 at 14 12 17

NachoSoto:
This leak is super interesting.
Basically this lead to `isAppBackgrounded` returning `false` in integration tests for some reason (likely a timing difference with the previous implementation), which is changing the behavior in `Purchases`. Still looking into the specific reason.

Sources/Misc/SystemInfo.swift
NachoSoto: Oooh makes sense.

NachoSoto:
Okay that was a false positive, simply because that test wasn't waiting to finish.

PR #2232, opened 2023-01-18
`PurchaseTester`: setting `changelog` when submitting to `TestFlight`

NachoSoto:
This will provide some context when a new build is submitted.
Also changed the lane to use `upload_to_testflight` instead of `upload_to_appstore`.

IMG_C78B738B91AB-1

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -492,7 +492,9 @@ platform :ios do
 
   desc "Build and deploy PurchaseTesterSwiftUI"
   lane :deploy_purchase_tester do
-    def ios_build
+    changelog = "#{current_version_number} SDK release"
+
+    def ios_build(changelog)
       increment_build_number(
         build_number: latest_testflight_build_number(platform: 'ios') + 1,
         xcodeproj: 'PurchaseTester.xcodeproj'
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -659,5 +652,5 @@ lane :check_pods do
 end
 
 def current_version_number
-  File.read("../.version").strip
+  File.read("#{File.dirname(__FILE__)}/../.version").strip
```
NachoSoto, line 655: This was slightly wrong, now it works when called from any folder.

.circleci/config.yml
```diff
@@ -720,42 +720,8 @@ workflows:
       not:
         equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
     jobs:
-      - lint:
```
aboedo, line 723: all of these are coming back before merging, right?

fastlane/Fastfile
aboedo: yeah, my bad, I didn't really consider paths other than calling from root. Solid improvement

aboedo (approved):

.circleci/config.yml
NachoSoto: Removed commit 👍🏻  was making sure it was still working.

fastlane/Fastfile
NachoSoto: It wasn't used from any of the testers before :)

PR #2238, opened 2023-01-19
`TrialOrIntroPriceEligibilityChecker`: use `TimingUtil` to log when it takes too long

NachoSoto:
See #1893.

Apple is aware of this issue, but it appears that it's still not fixed. Using `TimingUtil` introduced in #2059 this adds a log when the requests take too long.


Sources/Purchasing/Configuration.swift
```diff
@@ -238,6 +238,7 @@ extension Configuration {
 
         case productRequest = 3
         case purchase = 7
+        case introEligibility = 2
```
NachoSoto, line 241: I've chosen 2 seconds but let me know if you think that's too low?

NachoSoto: > I'd make it a bit higher since we've had cases where it takes > 30 secs, but it's entirely arbitrary so I'm good with any value here. 

I added intro eligibility checking to my app and suddenly my prices were taking a lot longer to show up. I immediately knew why, but some users might think there's something wrong. My thought was that anything beyond 2 seconds might be useful to let the developer know where the delay in the SDK was.

> We should make sure that the logs do state that it's taking longer than X secs, though, so it's clear to a dev what "longer than expected" means

That's done automatically by `TimingUtil` :) this is an example log:

> WARN: 🍎⚠️ StoreKit 2 intro eligibility took longer than expected to determine. (2.03 seconds)

aboedo: oh nice, I forgot about that

aboedo (approved):

PR #2240, opened 2023-01-19
Improve logging for custom package durations

codykerns:
### Checklist

### Motivation
The previous logs were vague, seemed kind of scary mentioning that the packages would be ignored - but it was just referencing that they couldn't be accessed from an Offering's convenience methods for each duration.

### Description
Updates the logs to explain that packages can be referenced by their identifiers directly with a link to the docs.


aboedo (approved):
❤️

PR #2243, opened 2023-01-23
`PurchasesOrchestrator`: finish SK2 transactions from `StoreKit.Transaction.updates` after posting receipt

NachoSoto:
This won't affect the most common scenario (observer mode), since transactions are not finished there.
But it's possible that `StoreKit.Transaction.updates` will have unfinished transactions (maybe those that failed to post previously). In that case, we don't want to finish them until successfully posting the receipts.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -793,17 +793,17 @@ extension PurchasesOrchestrator: StoreKit2TransactionListenerDelegate {
         _ listener: StoreKit2TransactionListener,
         updatedTransaction transaction: StoreTransactionType
     ) async throws {
-        await Async.call { completion in
-            self.finishTransactionIfNeeded(transaction) { @MainActor in
-                completion(())
-            }
-        }
-
         let isRestore = self.systemInfo.observerMode
 
         _ = try await self.syncPurchases(receiptRefreshPolicy: .always,
                                          isRestore: isRestore,
                                          initiationSource: .queue)
+
+        await Async.call { completion in
+            self.finishTransactionIfNeeded(transaction) { @MainActor in
+                completion(())
+            }
+        }
     }
 
 }
--- a/Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
+++ b/Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
@@ -597,6 +597,34 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
     }
 
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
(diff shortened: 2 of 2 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -793,17 +793,19 @@ extension PurchasesOrchestrator: StoreKit2TransactionListenerDelegate {
...
     ) async throws {
-        await Async.call { completion in
-            self.finishTransactionIfNeeded(transaction) { @MainActor in
-                completion(())
-            }
-        }
-
         let isRestore = self.systemInfo.observerMode
 
+        precondition(!self.finishTransactions)
```
NachoSoto, line 798: This line I'm going to remove, I just want to see if any of the integration tests that aren't observer mode would be affected by this change.

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -596,6 +596,34 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
         expect(self.backend.invokedPostReceiptDataParameters?.isRestore) == false
     }
 
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func testStoreKit2TransactionListenerDoesNotFinishTransactionIfPostingReceiptFails() async throws {
```
aboedo, line 600: we need to also have a test that ensures that it happens _before_ posting. 
Like, even if you successfully post, if you finished the transaction and refresh the receipt and _then_ post successfully to the backend, a consumable is lost forever. 

We **cannot finish a transaction until it after gets successfully posted to the backend**.

aboedo (changes requested):
I think we need another test to ensure the order of operations, good to go afterwards

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
NachoSoto: > we need to also have a test that ensures that it happens before posting.

I assumed that checking that we don't finish the transaction if the receipt post fails was a way to check ordering, because we can't finish until we know that it finished posting.
But I agree an explicit ordering test would be helpful, let me think about how to best do that.

NachoSoto: @aboedo done, added a test :)
I love being able to create `enum`s inside the test, so the expectation is this:
```swift
expect(operations) == [
    .receiptPost,
    .finishTransaction
]
```

aboedo (approved):

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
aboedo: awesome! And the syntax is really nice :)
Yeah, to be clear I wasn't questioning the current code, I was just pointing out that without an explicit test for the ordering, we could break it unknowingly in the future.

PR #2244, opened 2023-01-25
`StoreKitIntegrationTests`: replaced `XCTSkipIf` with `XCTExpectFailure`

NachoSoto:
See https://developer.apple.com/documentation/xctest/expected_failures

I realized this going through the old PR. Better to actually run the test so that if/when it gets fixed in the future, we're notified because it will "fail".


NachoSoto:
Something's weird with these coverage diffs, I wonder if we have some randomness in some test that makes that fluctuate.

tonidero (approved):
Hmm yeah... Not sure about the coverage change... Code change looks ok though

PR #2247, opened 2023-01-25
`IntroEligibilityCalculator`: changed logic to handle products with no subscription group

NachoSoto:
Follow up to #2174.

### DIFF
--- a/Sources/Purchasing/IntroEligibilityCalculator.swift
+++ b/Sources/Purchasing/IntroEligibilityCalculator.swift
@@ -102,18 +102,12 @@ private extension IntroEligibilityCalculator {
                 result[candidate.productIdentifier] = .unknown
                 continue
             }
-            let activeSubscriptionInGroup = (
-                candidate.subscriptionGroupIdentifier != nil &&
-                activeSubscriptionsProducts.contains {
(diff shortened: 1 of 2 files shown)


Sources/Purchasing/IntroEligibilityCalculator.swift
```diff
@@ -102,18 +102,12 @@ private extension IntroEligibilityCalculator {
                 result[candidate.productIdentifier] = .unknown
                 continue
             }
-            let activeSubscriptionInGroup = (
-                candidate.subscriptionGroupIdentifier != nil &&
```
NachoSoto, line 106: @aboedo this is simpler, and I think it would be a valid implementation too, as a way to handle products with no group?

NachoSoto:
1. We haven't changed this file in a while
2. That diff makes no sense?
Screenshot 2023-01-25 at 10 44 29

Sources/Purchasing/IntroEligibilityCalculator.swift
aboedo: yep! I love it

aboedo (approved):

PR #2250, opened 2023-01-27
Replaced `Any` uses for workaround with `Box`

NachoSoto:
See https://github.com/apple/swift/issues/58099
This is fixed in newer versions of Xcode, but we still need the workaround for older ones. Instead of using `Any` and force-casting, `Box` is better since it's type safe.

### Other changes:
- Moved `Box` to `RevenueCat` to be able to use the workaround there too.
- Made `Box` immutable. It's very dangerous to have a reference value (`Box` is a `class`) with reference semantics be mutable. Instead, we force users to have a mutable property instead and re-create the `Box`, to avoid mutating that data under the hood.
- Updated the comments to point to the new `Swift` GitHub issue page.

### DIFF
--- a/Tests/UnitTests/TestHelpers/Box.swift
+++ b/Tests/UnitTests/TestHelpers/Box.swift
@@ -13,10 +13,10 @@
 
 import Foundation
 
-// Workaround for https://openradar.appspot.com/radar?id=4970535809187840 / https://bugs.swift.org/browse/SR-15825
+// Workaround for https://openradar.appspot.com/radar?id=4970535809187840 / https://github.com/apple/swift/issues/58099
 final class Box<T> {
 
-    var value: T
+    let value: T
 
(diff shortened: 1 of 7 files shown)


Tests/UnitTests/TestHelpers/Box.swift
```diff
@@ -13,10 +13,10 @@
 
 import Foundation
 
-// Workaround for https://openradar.appspot.com/radar?id=4970535809187840 / https://bugs.swift.org/browse/SR-15825
+// Workaround for https://openradar.appspot.com/radar?id=4970535809187840 / https://github.com/apple/swift/issues/58099
 final class Box<T> {
 
-    var value: T
+    let value: T
```
NachoSoto, line 19: ⚠️ No longer mutable

tonidero (approved):

PR #2257, opened 2023-02-02
`HTTPClientTests`: refactored tests to use `waitUntil`

NachoSoto:
See also #2168 and #2071.

I was looking into randomness in tests that might lead to test coverage fluctuations and saw that these tests could be simplified to avoid use of `toEventually`.

### DIFF
--- a/Tests/UnitTests/Networking/HTTPClientTests.swift
+++ b/Tests/UnitTests/Networking/HTTPClientTests.swift
@@ -838,7 +838,7 @@ class HTTPClientTests: TestCase {
                          headers: nil)
         }
 
-        let response: HTTPResponse<Data>.Result? = waitUntilValue(timeout: .seconds(1)) { completion in
+        let response: HTTPResponse<Data>.Result? = waitUntilValue { completion in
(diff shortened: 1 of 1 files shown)


Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -99,9 +107,11 @@ class HTTPClientTests: TestCase {
 
         let request = HTTPRequest(method: .post([:]), path: .mockPath)
 
-        self.client.perform(request) { (_: EmptyResponse) in }
+        waitUntil { completion in
+            self.client.perform(request) { (_: EmptyResponse) in completion() }
+        }
 
-        expect(headerPresent.value).toEventually(equal(true))
```
tonidero, line 104: Looks like a few of these didn't have a timeout and will have it now so it's a slight change, but I think those are ok.

tonidero (approved):
LGTM!

Tests/UnitTests/Networking/HTTPClientTests.swift
NachoSoto: `toEventually` had a default timeout as well.

PR #2258, opened 2023-02-02
`StoreKit1WrapperTests`: avoid using `Bool.random` to fix flaky code coverage

NachoSoto:
Fixes https://app.codecov.io/gh/RevenueCat/purchases-ios/pull/2257

Screenshot 2023-02-02 at 11 48 14

Depending on this value, we were evaluating a different string.


NachoSoto:
Ugh of course this breaks the test.
But wait, if we're always checking that this is invoked, those lines should always be evaluated... so why would it fluctuate?

vegaro (approved):

PR #2262, opened 2023-02-03
`HTTPClientTests`: fixed failing test with missing assertions

NachoSoto:
`HTTPClientTest.testRequestIsRetriedIfResponseFromETagManagerIsNil` was actually failing (the request wasn't successful), but because it had no expectations, we didn't notice until we introduced Code Coverage.
As you can see [in the Codecov diff](https://app.codecov.io/gh/RevenueCat/purchases-ios/pull/2258):
Screenshot 2023-02-06 at 09 11 37

Most of the time that code didn't have coverage (meaning we weren't testing it). The reason for why is uncovered in this PR.
The random execution order meant that sometimes (most of the time) tests like `testHandlesRealErrorConditions` were running first, which did this:
```swift
let response = HTTPStubsResponse.emptySuccessResponse
response.error = error
return response
```

That seemingly innocuous code was modifying `HTTPStubsResponse.emptySuccessResponse`, defined in an extension in that file. Being used to value types nobody thought much of that, but unfortunately `HTTPStubsResponse` is a class, therefore it's a mutable type with reference semantics. Because of that, every test executed _after_ any of these modifying the response would receive a failure, but it was going unnoticed.

This PR also introduces a couple improvements to how we wait for several concurrent requests (which I thought was the issue at first).

### DIFF
--- a/Tests/UnitTests/Networking/HTTPClientTests.swift
+++ b/Tests/UnitTests/Networking/HTTPClientTests.swift
@@ -40,9 +40,9 @@ class HTTPClientTests: TestCase {
     }
 
     override func tearDown() {
-        super.tearDown()
-
         HTTPStubs.removeAllStubs()
+
+        super.tearDown()
     }
 
     func testUsesTheCorrectHost() throws {
--- a/Tests/UnitTests/Networking/HTTPClientTests.swift
+++ b/Tests/UnitTests/Networking/HTTPClientTests.swift
@@ -27,22 +27,22 @@ class HTTPClientTests: TestCase {
     override func setUp() {
         super.setUp()
 
-        userDefaults = MockUserDefaults()
-        eTagManager = MockETagManager(userDefaults: userDefaults)
-        operationDispatcher = OperationDispatcher()
+        self.userDefaults = MockUserDefaults()
+        self.eTagManager = MockETagManager(userDefaults: self.userDefaults)
+        self.operationDispatcher = OperationDispatcher()
         MockDNSChecker.resetData()
 
-        client = HTTPClient(apiKey: apiKey,
-                            systemInfo: systemInfo,
-                            eTagManager: eTagManager,
-                            dnsChecker: MockDNSChecker.self,
-                            requestTimeout: 3)
+        self.client = HTTPClient(apiKey: self.apiKey,
+                                 systemInfo: self.systemInfo,
+                                 eTagManager: self.eTagManager,
+                                 dnsChecker: MockDNSChecker.self,
+                                 requestTimeout: 3)
     }
 
     override func tearDown() {
-        HTTPStubs.removeAllStubs()
-
         super.tearDown()
+
+        HTTPStubs.removeAllStubs()
     }
 
     func testUsesTheCorrectHost() throws {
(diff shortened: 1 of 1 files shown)


Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -799,24 +799,29 @@ class HTTPClientTests: TestCase {
...
-            self.client.perform(.init(method: .get, path: path)) { (_: HTTPResponse<Data>.Result) in
-                completion()
+        let result: HTTPResponse<Data>.Result? = waitUntilValue { completion in
+            self.client.perform(.init(method: .get, path: path)) {
+                completion($0)
             }
         }
+
+        expect(result).to(beSuccess())
+        expect(requests.value) == 2
```
NachoSoto, line 824: ⚠️ This test basically had no expectations. Most of the time it wasn't going through the part of the code that it was meant to test, but we didn't notice until we saw it in the test coverage diff!

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -27,22 +27,22 @@ class HTTPClientTests: TestCase {
...
+                                 dnsChecker: MockDNSChecker.self,
+                                 requestTimeout: 3)
     }
 
     override func tearDown() {
-        HTTPStubs.removeAllStubs()
-
         super.tearDown()
+
+        HTTPStubs.removeAllStubs()
```
aboedo, line 45: was the ordering here important?

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -1085,9 +1116,13 @@ private func isPath(_ path: HTTPRequest.Path) -> HTTPStubsTestBlock {
...
-    static let emptySuccessResponse: HTTPStubsResponse = .init(data: Data(),
-                                                               statusCode: .success,
-                                                               headers: nil)
+    static var emptySuccessResponse: HTTPStubsResponse {
+        // `HTTPStubsResponse` doesn't have value semantics, it's a mutable class!
+        // This creates a new response each time so modifications in one test don't affect others.
+        return .init(data: Data(),
+                     statusCode: .success,
+                     headers: nil)
+    }
```
aboedo, line 1125: honestly, I think this might be more intuitive as a function.
Otherwise, since the syntax looks exactly the same, it'd be reasonable if you looked only at the callsite to assume that this bug wasn't fixed. 

If you call it like `emptySuccessResponse()`, you expect to be getting a brand new variable each time

aboedo:
great catch!

Tests/UnitTests/Networking/HTTPClientTests.swift
NachoSoto: Oh no, let me revert this. It's one of the things I tested before I knew what was wrong.

Tests/UnitTests/Networking/HTTPClientTests.swift
NachoSoto: Agreed 👍🏻

NachoSoto:
Updated.

aboedo (approved):

PR #2265, opened 2023-02-06
`StoreKit2StorefrontListener`: added tests to fix flaky code coverage

NachoSoto:
Similarly to #2262, test execution ordering was leading to [flaky code coverage results](https://app.codecov.io/gh/RevenueCat/purchases-ios/pull/2262) with this.

Screenshot 2023-02-06 at 10 48 45

Because we didn't have any test coverage, only if some other test changed the `Storefront`, this test was calling the delegate code.

This adds explicit test coverage using a `MockAsyncSequence`.

### Other changes:
- Added `StorefrontType` to `StoreKit2StorefrontListenerDelegate` method
- `Storefront` initializers are no longer `Optional` `init?`
- Simplified `StoreKit2StorefrontListenerTests` by moving `AvailabilityChecks` to `setUp`

### DIFF
--- a/Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
+++ b/Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
@@ -16,7 +16,7 @@ import StoreKit
 
 protocol StoreKit2StorefrontListenerDelegate: AnyObject, Sendable {
 
-    func storefrontDidUpdate()
+    func storefrontDidUpdate(with storefront: StorefrontType)
 
 }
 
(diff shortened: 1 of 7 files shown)


Sources/Purchasing/StoreKit2/StoreKit2StorefrontListener.swift
```diff
@@ -32,17 +32,29 @@ class StoreKit2StorefrontListener {
...
 
-    init(delegate: StoreKit2StorefrontListenerDelegate?) {
+    convenience init(delegate: StoreKit2StorefrontListenerDelegate?) {
+        self.init(
+            delegate: delegate,
+            updates: StoreKit.Storefront.updates.map(Storefront.init(sk2Storefront:))
+        )
+    }
+
+    init<S: AsyncSequence>(
```
aboedo, line 44: let's add a description and sample usage here, this `init` won't be intuitive to folks not very used to Swift

NachoSoto: I added a comment explaining this.

PR #2271, opened 2023-02-08
Create SECURITY.md

Kelvmks93:
### Checklist


aboedo:
@Kelvmks93 could you share the motivation for this one? We do intend to add a security policy at one point, but we'd have to be very thoughtful about it

PR #2272, opened 2023-02-08
`HTTPClient`: added signature validation and introduced `ErrorCode.signatureVerificationFailed`

NachoSoto:
Fixes [CSDK-634].

## Changes:
- Created `HTTPResponseValidationResult`
- Added `HTTPResponseValidationResult` and `HTTPClient.ResponseHeaders` to `HTTPResponse`
- Added new `ErrorCode.signatureVerificationFailed`
- Extracted `SigningType` to create `MockSigning` to test `HTTPClient` without bothering with signatures
- Refactored `ETagManager` to use `HTTPResponse`, making it easier to pass through response headers and verification results
- Stored `HTTPResponseValidationResult` in `ETagManager` cache
- `HTTPClient` handles signatures from response based on `SystemInfo.responseVerificationLevel`
- Renamed `HTTPRequest.createIntegrityEnforcedRequestRequest` to `HTTPRequest.createWithResponseVerification`
- Added lots of tests for this new behavior in `HTTPClient`
- Added lots of new tests for `ETagManager`

[CSDK-634]: https://revenuecats.atlassian.net/browse/CSDK-634?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Misc/Signing+ResponseValidation.swift
+++ b/Sources/Misc/Signing+ResponseValidation.swift
@@ -15,25 +15,25 @@ import Foundation
 
 extension HTTPResponse where Body == Data {
 
-    static func createAndValidate(with response: HTTPURLResponse,
-                                  body: Data,
-                                  request: HTTPRequest,
-                                  publicKey: Signing.PublicKey?,
-                                  signing: SigningType.Type = Signing.self) -> Self {
-        return Self.createAndValidate(body: body,
-                                      statusCode: .init(rawValue: response.statusCode),
-                                      headers: response.allHeaderFields,
-                                      request: request,
-                                      publicKey: publicKey,
-                                      signing: signing)
+    static func create(with response: HTTPURLResponse,
+                       body: Data,
+                       request: HTTPRequest,
+                       publicKey: Signing.PublicKey?,
+                       signing: SigningType.Type = Signing.self) -> Self {
+        return Self.create(with: body,
+                           statusCode: .init(rawValue: response.statusCode),
+                           headers: response.allHeaderFields,
+                           request: request,
+                           publicKey: publicKey,
+                           signing: signing)
     }
 
-    static func createAndValidate(body: Data,
-                                  statusCode: HTTPStatusCode,
-                                  headers: HTTPClient.ResponseHeaders,
-                                  request: HTTPRequest,
-                                  publicKey: Signing.PublicKey?,
-                                  signing: SigningType.Type = Signing.self) -> Self {
+    static func create(with body: Data,
+                       statusCode: HTTPStatusCode,
+                       headers: HTTPClient.ResponseHeaders,
+                       request: HTTPRequest,
+                       publicKey: Signing.PublicKey?,
+                       signing: SigningType.Type = Signing.self) -> Self {
         return .init(
             statusCode: statusCode,
             responseHeaders: headers,
@@ -83,18 +83,16 @@ extension Result where Success == Data?, Failure == NetworkError {
 
     /// Converts a `Result<Data?, NetworkError>` into `Result<HTTPResponse<Data?>, NetworkError>`
     func mapToResponse(
-        responseHeaders: HTTPClient.ResponseHeaders,
-        statusCode: HTTPStatusCode,
+        response: HTTPURLResponse,
         request: HTTPRequest,
         signing: SigningType.Type,
         verificationLevel: Signing.ResponseVerificationLevel
     ) -> Result<HTTPResponse<Data?>, Failure> {
         return self.flatMap { body in
             if let body = body {
-                let response = HTTPResponse.createAndValidate(
+                let response = HTTPResponse.create(
+                    with: response,
                     body: body,
-                    statusCode: statusCode,
-                    headers: responseHeaders,
                     request: request,
                     publicKey: verificationLevel.publicKey,
                     signing: signing
@@ -110,8 +108,8 @@ extension Result where Success == Data?, Failure == NetworkError {
                 // so the response will be fetched from `ETagManager`.
                 return .success(
                     .init(
-                        statusCode: statusCode,
-                        responseHeaders: responseHeaders,
+                        statusCode: HTTPStatusCode(rawValue: response.statusCode),
+                        responseHeaders: response.allHeaderFields,
                         body: body,
                         validationResult: .notRequested
                     )
--- a/Sources/Misc/Signing+ResponseValidation.swift
+++ b/Sources/Misc/Signing+ResponseValidation.swift
@@ -1,0 +1,123 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  Signing+ResponseValidation.swift
+//
+//  Created by Nacho Soto on 2/8/23.
+
+import Foundation
+
+extension HTTPResponse where Body == Data {
+
+    static func createAndValidate(with response: HTTPURLResponse,
+                                  body: Data,
+                                  request: HTTPRequest,
+                                  publicKey: Signing.PublicKey?,
+                                  signing: SigningType.Type = Signing.self) -> Self {
+        return Self.createAndValidate(body: body,
+                                      statusCode: .init(rawValue: response.statusCode),
+                                      headers: response.allHeaderFields,
+                                      request: request,
+                                      publicKey: publicKey,
+                                      signing: signing)
+    }
+
+    static func createAndValidate(body: Data,
+                                  statusCode: HTTPStatusCode,
+                                  headers: HTTPClient.ResponseHeaders,
+                                  request: HTTPRequest,
+                                  publicKey: Signing.PublicKey?,
+                                  signing: SigningType.Type = Signing.self) -> Self {
+        return .init(
+            statusCode: statusCode,
+            responseHeaders: headers,
+            body: body,
+            validationResult: Self.validationResult(
+                body: body,
+                headers: headers,
+                request: request,
+                publicKey: publicKey,
+                signing: signing
+            )
+        )
+    }
+
+    private static func validationResult(
+        body: Data,
+        headers: HTTPClient.ResponseHeaders,
+        request: HTTPRequest,
+        publicKey: Signing.PublicKey?,
+        signing: SigningType.Type
+    ) -> HTTPResponseValidationResult {
+        guard let nonce = request.nonce,
+              let publicKey = publicKey,
+              #available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *) else {
+            return .notRequested
+        }
+
+        guard let signature = headers[HTTPClient.responseSignatureHeaderName] as? String else {
+            Logger.warn(Strings.signing.signature_was_requested_but_not_provided(request))
+
+            return .failedValidation
+        }
+
+        if signing.verify(message: body,
+                          nonce: nonce,
+                          hasValidSignature: signature,
+                          with: publicKey) {
+            return .validated
+        } else {
+            return .failedValidation
+        }
+    }
+
+}
+
+extension Result where Success == Data?, Failure == NetworkError {
+
+    /// Converts a `Result<Data?, NetworkError>` into `Result<HTTPResponse<Data?>, NetworkError>`
+    func mapToResponse(
+        responseHeaders: HTTPClient.ResponseHeaders,
+        statusCode: HTTPStatusCode,
+        request: HTTPRequest,
+        signing: SigningType.Type,
+        verificationLevel: Signing.ResponseVerificationLevel
+    ) -> Result<HTTPResponse<Data?>, Failure> {
+        return self.flatMap { body in
+            if let body = body {
+                let response = HTTPResponse.createAndValidate(
+                    body: body,
+                    statusCode: statusCode,
+                    headers: responseHeaders,
+                    request: request,
+                    publicKey: verificationLevel.publicKey,
+                    signing: signing
+                )
+
+                if response.validationResult == .failedValidation, case .enforced = verificationLevel {
+                    return .failure(.signatureVerificationFailed(path: request.path))
+                } else {
+                    return .success(response.mapBody(Optional.some))
+                }
+            } else {
+                // No body means that the status code `HTTPStatusCode.notModified`
+                // so the response will be fetched from `ETagManager`.
+                return .success(
+                    .init(
+                        statusCode: statusCode,
+                        responseHeaders: responseHeaders,
+                        body: body,
+                        validationResult: .notRequested
+                    )
+                )
+            }
+        }
+    }
+
+}
--- a/Sources/Networking/HTTPClient/ETagManager.swift
+++ b/Sources/Networking/HTTPClient/ETagManager.swift
@@ -35,28 +35,27 @@ class ETagManager {
 
     func eTagHeader(for urlRequest: URLRequest, refreshETag: Bool = false) -> [String: String] {
         var storedETag = ""
-        if !refreshETag, let storedETagAndResponse = storedETagAndResponse(for: urlRequest) {
+        if !refreshETag,
+           let storedETagAndResponse = self.storedETagAndResponse(for: urlRequest),
+           storedETagAndResponse.validationResult != .failedValidation {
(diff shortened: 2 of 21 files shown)


Tests/UnitTests/Networking/ETagManagerTests.swift
```diff
@@ -212,17 +286,214 @@ class ETagManagerTests: TestCase {
...
+            statusCode: .success,
+            data: actualResponse,
+            validationResult: .notRequested
+        ).asData()
+
+        let response = self.eTagManager.eTagHeader(for: request)
+        expect(response[ETagManager.eTagHeaderName]) == eTag
+    }
+
+    func testCachedResponseWithNoValidationResultIsIgnored() {
```
NachoSoto, line 380: This tests backwards compatibility with old responses. We handle that as a cache miss.

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -96,26 +106,6 @@ class HTTPClientTests: TestCase {
         expect(headers?.keys).toNot(contain(HTTPClient.nonceHeaderName))
     }
 
-    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
-    func testRequestIncludesRandomNonce() throws {
```
NachoSoto, line 100: This test was moved below to `SignatureVerificationHTTPClientTests`

Sources/Misc/Signing+ResponseValidation.swift
```diff
@@ -0,0 +1,123 @@
...
+//
+//  Created by Nacho Soto on 2/8/23.
+
+import Foundation
+
+extension HTTPResponse where Body == Data {
+
+    static func createAndValidate(with response: HTTPURLResponse,
+                                  body: Data,
+                                  request: HTTPRequest,
```
bisho, line 20: Can we extend `HTTPRequest` to have a `IntegrityVerifiedHTTPRequest` that guarantees contains a Nonce? And then make this signature expect that class in particular. Doesn't make sense to validate a `HTTPRequest` that is not sending Nonces.

NachoSoto: That's definitely an alternative approach, but that's not very Swift-y. We'd have to move away from it being a `struct` and lose the value semantics on this type.

I think the main issue is this is bad name. It returns `HTTPResponseValidationResult.notRequested` if the request doesn't need to be verified, so maybe instead of `createAndValidate` it should just be `HTTPResponse.create(with:...)`.

NachoSoto: Alternatively we'd have to make `HTTPResponse` a protocol and use a generic `<Response: HTTPResponseType>` everywhere to maintain type safety at compile time instead of doing that at runtime. But at that point it would be an overkill.

Sources/Networking/HTTPClient/HTTPResponse.swift
```diff
@@ -18,19 +18,48 @@ struct HTTPResponse<Body: HTTPResponseBody> {
...
+                return bodyDescription
+            } else {
+                return "\(type(of: self.body))"
+            }
+        }()
+
+        return """
+        HTTPResponse(" +
+        statusCode: \(self.statusCode.rawValue),
+        body: \(body),
```
NachoSoto, line 59: OMG code coverage pointing to a bug here, amazing.

NachoSoto:
Apologies for the big diff 🙏🏻 I couldn't really split the changes here into multiple PRs since they're all connected.
Most of the changes are a lot of new tests though

Sources/Misc/Signing+ResponseValidation.swift
bisho: Hummm... I see... I have no clue about swift :)

And something like createRequestAndValidator() that returns a request with the nonce, and a validator ready to use (contains the nonce)? It feels weird to have this generic create and validate that sometimes will not validate, plus there is not a lot of type safety, there is no way to enforce validation for some kind of request.

That way the request remains generic, you will just build a validator when requesting a request with validation. If not requesting validation, you will just have a create that builds a request.

Again, all this not knowing a thing about swift patterns :)

NachoSoto: I renamed this, I think it's clearer now. I do get your point of seemingly validating requests that don't require it.
`HTTPResponse` always has an `HTTPResponseValidationResult` (even if it's `.notRequested`).
So `HTTPResponse.create(with:body:request:..)` is the new API. It may or may not do any validation, but it'll return an `HTTPResponse` with the appropriate information. Does that sound less weird?

Tests/UnitTests/Networking/ETagManagerTests.swift
```diff
@@ -212,17 +286,214 @@ class ETagManagerTests: TestCase {
...
+        let cacheKey = url.absoluteString
+
+        let actualResponse = "response".asData
+
+        self.mockUserDefaults.mockValues[cacheKey] = """
+        {
+        "e_tag": "\(eTag)",
+        "status_code": 200,
+        "data": "\(actualResponse.asFetchToken)",
+        "validation_result": 1
```
thecadams, line 449: nit: maybe consider `"validation_result": \(.validated.rawValue)`

thecadams (commented):
Nice change @NachoSoto. Not a swift expert, but looks good to me. The API with the optional requesting of verification is going to prove handy if we decide to do this for more endpoint in future..

PS particularly liked how simple the `sign` and `createSalt` test methods got, the simple keypair generation with `private let (privateKey, publicKey) = SigningTests.createRandomKey()`, and how packing things into test classes creates clean encapsulation. Learned a bunch about Swift reading this, thanks for sending it through!

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -221,10 +226,12 @@ private extension HTTPClient {
...
+            .mapToResponse(response: httpURLResponse,
+                           request: request.httpRequest,
+                           signing: self.signing,
+                           verificationLevel: self.systemInfo.responseVerificationLevel)
             .map {
-                self.eTagManager.httpResultFromCacheOrBackend(with: httpURLResponse,
-                                                              data: $0.body,
+                self.eTagManager.httpResultFromCacheOrBackend(with: $0,
                                                               request: urlRequest,
                                                               retried: request.retried)
```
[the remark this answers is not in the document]
thecadams: This is based on the empty body case in https://github.com/RevenueCat/purchases-ios/pull/2272/files#diff-2a4a83f4e5d43494fe43c0b623281c54d24bd60193c7de185f37e565b0c72de3R107-R108 and the comment suggesting that code assumes a cache hit in the caller..

Tests/UnitTests/Networking/ETagManagerTests.swift
NachoSoto: In this particular case I wanted to hardcore it so we never break backwards compatibility if somebody changes the associated value. I'll add a comment to explain.

Sources/Networking/HTTPClient/HTTPClient.swift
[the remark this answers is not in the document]
NachoSoto: > If there's an empty response and no etag, looks like this will be nil due to nothing in cache. Is that ok for callers?

Yeah that's confusing, it took me a while to wrap my head around why that was done when I refactored this months ago.
The answer to that is in `HTTPClient.handle(urlResponse:request:urlRequest:data:error:)`. If this returns `nil` (the scenario you described), it gets retried. There's plenty of test coverage for that.

> Is that ok for callers?

As for users of `HTTPClient`, ultimately responses can't be `nil` thanks to the strong type safety provided by its API.
The `Completion` type is defined as
```swift
Completion<Value: HTTPResponseBody> = (HTTPResponse<Value>.Result) -> Void
```

Which is basically `(Result<HTTPResponse<Value>, NetworkError>) -> Void`. In other words, there will be a response or an error, but never an empty or `nil` response.

Tests/UnitTests/Networking/ETagManagerTests.swift
NachoSoto: Actually I'm adding plenty of test coverage for that serialization in #2277 so I'll follow your suggestion 👍🏻

NachoSoto:
I'm thinking of renaming the `ErrorCode` here instead if we agree with the name in #2277.

Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -35,28 +35,27 @@ class ETagManager {
 
     func eTagHeader(for urlRequest: URLRequest, refreshETag: Bool = false) -> [String: String] {
         var storedETag = ""
-        if !refreshETag, let storedETagAndResponse = storedETagAndResponse(for: urlRequest) {
+        if !refreshETag,
+           let storedETagAndResponse = self.storedETagAndResponse(for: urlRequest),
+           storedETagAndResponse.validationResult != .failedValidation {
```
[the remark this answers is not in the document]
aboedo: also... do we ever clean these up? 

It feels like if we have an eTag but it doesn't have validation information and validation has been activated, we should kill the old eTag, but I don't recall what our cleanup mechanism for this was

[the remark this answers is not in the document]
aboedo: also, we're not storing when validationResult = failed, so this would never be true, right?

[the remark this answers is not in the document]
aboedo: last question: what are we doing here when validation wasn't requested at all? Like, if validation is disabled, this should behave exactly as it always has

Tests/UnitTests/Mocks/MockHTTPClient.swift
```diff
@@ -31,7 +31,12 @@ class MockHTTPClient: HTTPClient {
             // swiftlint:disable:next force_try
             let data = try! JSONSerialization.data(withJSONObject: response)
 
-            self.init(response: .success(.init(statusCode: statusCode, body: data)),
+            self.init(response: .success(.init(
+                statusCode: statusCode,
+                responseHeaders: [:],
+                body: data,
+                validationResult: .notRequested
+            )),
```
tonidero, line 39: Hmm indentation here seems hard to read IMO... Maybe extract the `success` to a variable?

Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -166,7 +171,24 @@ extension ETagManager.Response {
...
+
+// MARK: -
+
+private extension HTTPResponse {
+
+    var shouldStore: Bool {
+        return (
+            self.statusCode != .notModified &&
+            !self.statusCode.isServerError &&
+            self.validationResult != .failedValidation
```
tonidero, line 190: 🎉

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -72,6 +76,7 @@ extension HTTPClient {
 
     static let authorizationHeaderName = "Authorization"
     static let nonceHeaderName = "X-Nonce"
+    static let responseSignatureHeaderName = "x-signature"
```
tonidero, line 79: Should this be 
```suggestion
    static let responseSignatureHeaderName = "X-Signature"
```

tonidero (commented):
Still need to review tests but looking good!

Sources/Networking/HTTPClient/ETagManager.swift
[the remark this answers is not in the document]
NachoSoto: > we're interchangeably using validation and verification throughout the feature, we should pick one and settle. I think verification is more technically appropriate for this

I fixed those inconsistencies in #2277. I went with `validation` but we can change it back.

> It feels like if we have an eTag but it doesn't have validation information and validation has been activated, we should kill the old eTag

Hmm good point, let me see if we have a test case for that and if not I'll add it.

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto: Hmm good point.

Sources/Misc/Signing+ResponseValidation.swift
```diff
@@ -0,0 +1,121 @@
...
+        body: Data,
+        headers: HTTPClient.ResponseHeaders,
+        request: HTTPRequest,
+        publicKey: Signing.PublicKey?,
+        signing: SigningType.Type
+    ) -> HTTPResponseValidationResult {
+        guard let nonce = request.nonce,
+              let publicKey = publicKey,
+              #available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *) else {
+            return .notRequested
```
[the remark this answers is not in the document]
NachoSoto: We talked about this offline.
We're going to solve this by making the "validated" APIs only available in iOS 13+, to force users to decide how to handle older versions.

The configuration API is already not available below, so it's also clear that you can't use it on older devices.

Tests/UnitTests/Mocks/MockHTTPClient.swift
NachoSoto: Done.

Sources/Networking/HTTPClient/ETagManager.swift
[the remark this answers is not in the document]
NachoSoto: Okay, updated `ETagManager` to ignore etags for the case where the stored cache had no validation but the new request requires a signature.

NachoSoto:
All comments addressed.

Sources/Misc/Signing+ResponseValidation.swift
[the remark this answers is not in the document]
tonidero: So this will change in a followup PR I imagine right?

Sources/Networking/HTTPClient/HTTPResponse.swift
```diff
@@ -18,19 +18,54 @@ struct HTTPResponse<Body: HTTPResponseBody> {
 
     typealias Result = Swift.Result<Self, NetworkError>
 
-    let statusCode: HTTPStatusCode
-    let body: Body
+    var statusCode: HTTPStatusCode
+    var responseHeaders: HTTPClient.ResponseHeaders
+    var body: Body
+    var validationResult: HTTPResponseValidationResult
```
tonidero, line 24: I may be missing something but why did we make these `var` instead of `let`?

Tests/UnitTests/Mocks/MockETagManager.swift
```diff
@@ -13,25 +13,39 @@ import Foundation
 // swiftlint:disable type_name
 class MockETagManager: ETagManager {
 
+    struct ETagHeaderRequest {
+        var urlRequest: URLRequest
+        var refreshETag: Bool
+        var signatureVerificationEnabled: Bool
```
tonidero, line 19: Same here, could these be let? I don't think callers of this mock would need to modify this?

tonidero (approved):
LGTM!

Sources/Networking/HTTPClient/HTTPResponse.swift
NachoSoto: So we can modify the validation result after it's initially created in `HTTPClient` (just an implementation detail).
Note that `struct`s have [value semantics](https://www.swiftbysundell.com/articles/utilizing-value-semantics-in-swift/), so this isn't like having mutable data in other languages.

Example:
```swift
struct Data {
  // Technically mutable
  var a: Int
  var b: String
}

let d = Data(a: 1, b: "test")
d.a +=1 // Error, `d` is immutable!

func f(_ d: Data) {
  d.a += 1 // Error, d is immutable
}
```

Additionally, `struct`s get implicitly copied (they have _value_ semantics, not _reference_), they're passed by _value_, so you can't lead to shared mutable state.

So basically it doesn't make a difference if these are `var`s, it just makes creating copies with different values easier without needing to do it manually calling the constructor again.

Tests/UnitTests/Mocks/MockETagManager.swift
NachoSoto: Ditto, it doesn't make a difference, it would only be mutable for a `var ETagHeaderRequest` anyway.

Sources/Networking/HTTPClient/HTTPResponse.swift
tonidero: Right, I asked about this before and I forgot 😅 thanks for explaining again!

PR #2276, opened 2023-02-09
`BundleSandboxEnvironmentDetector`: always return `true` when running on simulator

NachoSoto:
Thanks to @chrisvasselli for the find and initial implementation (#2275).

This supersedes the workaround introduced in #1730, now making `IntegrationTests` not override the `SandboxEnvironmentDetector`.

### DIFF
--- a/Sources/Misc/SandboxEnvironmentDetector.swift
+++ b/Sources/Misc/SandboxEnvironmentDetector.swift
@@ -24,12 +24,18 @@ protocol SandboxEnvironmentDetector: Sendable {
 final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
 
     private let bundle: Bundle
+    private let isRunningInSimulator: Bool
 
-    init(bundle: Bundle = .main) {
+    init(bundle: Bundle = .main, isRunningInSimulator: Bool = SystemInfo.isRunningInSimulator) {
         self.bundle = bundle
+        self.isRunningInSimulator = isRunningInSimulator
     }
 
     var isSandbox: Bool {
+        guard !self.isRunningInSimulator else {
+            return true
+        }
+
         guard let path = self.bundle.appStoreReceiptURL?.path else {
             return false
         }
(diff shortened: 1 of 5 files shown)


Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -43,12 +49,7 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
         }
     }
 
-    #if DEBUG
```
NachoSoto, line 46: No longer needed. This is the main change from #2275.

Tests/UnitTests/Misc/SystemInfoTests.swift
```diff
@@ -57,18 +57,6 @@ class SystemInfoTests: TestCase {
         expect(systemInfo.observerMode) == !finishTransactions
     }
 
-    func testIsSandbox() throws {
```
NachoSoto, line 60: I removed these tests so we don't need to also inject `isRunningInSimulator` to `SystemInfo`.
Sandbox environment is detected through `SandboxEnvironmentDetector`, so these tests were duplicated.

Tests/UnitTests/Misc/SandboxEnvironmentDetectorTests.swift
```diff
@@ -19,34 +19,43 @@ import XCTest
...
+        expect(try SystemInfo.with(receiptReceipt: .nilURL, inSimulator: true).isSandbox) == true
     }
 
 }
 
 private extension SandboxEnvironmentDetector {
 
-    static func withReceiptResult(_ result: MockBundle.ReceiptURLResult) throws -> SandboxEnvironmentDetector {
+    static func with(
+        receiptReceipt result: MockBundle.ReceiptURLResult,
```
tonidero, line 52: Typo?
```suggestion
        receiptResult result: MockBundle.ReceiptURLResult,
```

tonidero (approved):
LGTM!

Tests/UnitTests/Misc/SandboxEnvironmentDetectorTests.swift
NachoSoto: lol 🤦🏻‍♂️

tonidero:
Ah will need to fix the usages of that function as well @NachoSoto

PR #2274, opened 2023-02-09
`Signing`: updated to use production public key

NachoSoto:
Depends on #2272 and #2278.
Depends on https://github.com/RevenueCat/khepri/pull/5232.

PR #2267, opened 2023-02-06
`PurchasesDiagnostics`: added step to verify signature verification

NachoSoto:
⚠️ 🎉  This also changes integration tests to use `EntitlementVerificationLevel.enforced` so that integration tests fail if signatures are invalid.

#### Depends on:
- https://github.com/RevenueCat/khepri/pull/5191
- https://github.com/RevenueCat/khepri/pull/5204
- #2214
- #2215 
- #2216
- #2272

_Marking this as `feat`ure because it contains a new error in `PurchasesDiagnostics.Error`_

PR #2277, opened 2023-02-09
Introduced `Configuration.EntitlementVerificationMode` and `VerificationResult`

NachoSoto:
Fixes [CSDK-629].
Depends on #2274 and #2267.

Entitlement verification is accomplished by preventing MiTM attacks between the SDK and the RevenueCat server.
With verification enabled, the SDK ensures that the response created by the server was not modified by a third-party, and the entitlements received are exactly what was sent.

Entitlements are only verified if enabled using `Configuration.Builder.with(entitlementVerificationMode:)`, which is disabled by default.

### Changes:
- Added new `VerificationResult`, which replaces `HTTPResponseVerificationResult`
- Renamed `Configuration.EntitlementVerificationLevel` to `Configuration.EntitlementVerificationMode`
- `Configuration.with(entitlementVerificationMode:)` is now `public`
- Added `CustomerInfo.entitlementVerification`
- Added `EntitlementInfos.verification` and `EntitlementInfo.verification`
- Added tests to cover serialization of these new values
- `CustomerInfoResponseHandler` copies the `ValidationResult` to `CustomerInfo` (all operations that fetch `CustomerInfo` use this)
- Renamed "validation" to "verification" everywhere for consistency
- Finalized documentation

### TODO:
- [x] Figure out whether to use a separate type from `HTTPResponseValidationResult` or rename that one
- [x] Figure out final API type and property names
- [x] Add tests for `CustomerInfo.copy(with:)`
- [x] Ensure validation result is encoded and decoded (for caching)
- [x] Add missing snapshots

[CSDK-629]: https://revenuecats.atlassian.net/browse/CSDK-629?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Sources/Purchasing/EntitlementInfos.swift
```diff
@@ -31,8 +31,15 @@ import Foundation
         return self.all[key]
     }
 
+    // TODO: public
+    internal let validationResult: HTTPResponseValidationResult
```
aboedo, line 35: we should check with product. the original name for this was "trusted", values boolean (potentially optional to represent the not requested state)

aboedo (commented):
looking great!

Sources/Networking/HTTPClient/NetworkError.swift
```diff
@@ -25,7 +25,7 @@ enum NetworkError: Swift.Error, Equatable {
     case unableToCreateRequest(HTTPRequest.Path, Source)
     case unexpectedResponse(URLResponse?, Source)
     case errorResponse(ErrorResponse, HTTPStatusCode, Source)
-    case signatureVerificationFailed(HTTPRequest.Path, Source)
+    case signatureValidationFailed(HTTPRequest.Path, Source)
```
NachoSoto, line 28: This hasn't shipped so we can safely change it.

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -20,15 +20,15 @@ class CustomerInfoResponseHandler {
     func handle(customerInfoResponse response: HTTPResponse<Response>.Result,
                 completion: CustomerAPI.CustomerInfoResponseHandler) {
         let result: Result<CustomerInfo, BackendError> = response
-            .map {
+            .map { response in
```
thecadams, line 23: Nice clarification

Tests/APITesters/.swiftlint.yml
```diff
@@ -2,3 +2,4 @@ disabled_rules:
   - file_length
   - type_body_length
   - function_body_length
+  - force_try
```
thecadams, line 5: nit: Not an expert here, but only seeing one `try!` in this PR, so is this worth trying to leave enabled?
My research took me to https://realm.github.io/SwiftLint/force_try.html and https://cocoacasts.com/what-is-the-difference-between-try-try-and-try where it looks like they spell out good reasons to avoid `try!`.. so thought I would raise it.

thecadams (commented):
Looking good, but I wouldn't consider myself qualified to approve so just leaving a few comments

Tests/APITesters/.swiftlint.yml
NachoSoto: Oh I forgot to explain this. We don't actually _run_ APITester code, we only compile it (thats by design). `try!` would crash at runtime, but because it never runs it doesn't matter. It simplifies the type checking.

Sources/Identity/CustomerInfo.swift
```diff
@@ -45,6 +45,12 @@ import Foundation
      */
     @objc public let nonSubscriptions: [NonSubscriptionTransaction]
 
+    /// Whether the entitlements were validation.
+    ///
+    /// ### Related Symbols
+    /// - ``EntitlementValidation``
+    @objc public var entitlementValidation: EntitlementValidation { self.entitlements.validation }
```
aboedo, line 52: I'd go with "trusted" since it feels easier to wrap your head around (also, Trusted Entitlements as the official feature name has a nice ring to it)

NachoSoto: What would you name the type?

Sources/Identity/CustomerInfo.swift
```diff
@@ -45,6 +45,12 @@ import Foundation
      */
     @objc public let nonSubscriptions: [NonSubscriptionTransaction]
 
+    /// Whether the entitlements were validation.
```
tonidero, line 48: Typo:
```suggestion
    /// Whether the entitlements were validated.
```

Sources/Identity/CustomerInfo.swift
tonidero: Hmm maybe `EntitlementTrustStatus`? Might be too long...

Sources/Purchasing/Configuration.swift
```diff
@@ -207,19 +212,32 @@ import Foundation
 
 // MARK: - Public Keys
 
-internal extension Configuration {
+public extension Configuration {
 
-    /// Defines how strict ``EntitlementInfo`` verification ought to be.
-    @objc(RCEntitlementVerificationLevel)
-    enum EntitlementVerificationLevel: Int {
+    // TODO: link to docs
```
tonidero, line 217: Guessing this will be done after this is added to the docs right? We should just make sure we don't forget.

Sources/Purchasing/Configuration.swift
```diff
@@ -207,19 +212,32 @@ import Foundation
...
+    enum EntitlementValidationMode: Int {
+
+        /// The SDK will not perform any entitlement validation.
         case disabled = 0
 
-        /// The SDK will verify entitlements, but will not fail to parse them if verification failed.
+        /// Enable entitlement validation.
+        ///
+        /// If validation fails, this will be indicated with ``EntitlementValidation/failedValidation``
+        /// but parsing will not fail.
```
tonidero, line 234: Maybe clarify that entitlements will be granted? And maybe some description of the use cases. SOmething like:

```suggestion
        /// Enable entitlement validation but continue granting entitlements on failure.
        /// This can be useful if you want to customize your reaction to validation failures but still grant access.
        ///
        /// If validation fails, this will be indicated with ``EntitlementValidation/failedValidation``
        /// but parsing will not fail.
```

Sources/Purchasing/EntitlementInfo.swift
```diff
@@ -161,6 +161,9 @@ extension PeriodType: DefaultValueProvider {
      */
     @objc public var ownershipType: PurchaseOwnershipType { self.contents.ownershipType }
 
+    /// Whether this entitlement was validated.
```
tonidero, line 164: This makes me think of a boolean. Could we link to the `EntitlementValidation` type?

tonidero (approved):
Looks great! Just some small things

Sources/Purchasing/Configuration.swift
NachoSoto: I just filed a Jira for this 👍🏻

Sources/Purchasing/EntitlementInfo.swift
NachoSoto: `Docc` will automatically have a link on the return type, but we could add
```swift
/// ### Related Symbols
/// - ``EntitlementValidation``
```

@aboedo any thoughts on whether that's overkill?

Sources/Identity/CustomerInfo.swift
NachoSoto: We agreed on `VerificationResult`.

Sources/Purchasing/EntitlementInfo.swift
NachoSoto: Never mind, I see that I added this to `CustomerInfo`, so adding it here too 👍🏻

Sources/Purchasing/Configuration.swift
NachoSoto: Good idea, added.

NachoSoto:
Gotta fix the snapshots

PR #2278, opened 2023-02-10
`HTTPRequest.Path.health`: don't cache using `ETagManager`

NachoSoto:
This breaks signature validation in #2267 because `PurchasesDiagnostics` uses that endpoint with and without signature validation, so the first time it makes a request, it stores the cached response without a validated signature.

The best solution is to simply not cache this endpoint, since it's used precisely to make sure that requests can be made with no errors, and we wouldn't want false positives.


Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -490,6 +490,46 @@ class HTTPClientTests: TestCase {
         expect(result?.value?.statusCode) == .success
     }
 
+    func testCachedRequestsIncludeETagHeader() {
```
NachoSoto, line 493: Turns out we didn't have tests for this!

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -490,6 +490,46 @@ class HTTPClientTests: TestCase {
...
+
+        waitUntil { completion in
+            self.client.perform(request) { (_: HTTPResponse<Data>.Result) in completion() }
+        }
+
+        expect(headerPresent.value) == true
+        expect(self.eTagManager.invokedETagHeader) == true
+    }
+
+    func testNotCachedRequestsIncludeETagHeader() {
```
tonidero, line 516:
```suggestion
    func testNotCachedRequestsDontIncludeETagHeader() {
```

tonidero (approved):
I think this makes sense though I do wonder if we should rate limit this API request if we are not caching it... If we rate limit it from the backend, I imagine it will fail for a bunch of people, but if they use it correctly, I expect it not to be an issue so I think this should be ok. CC @aboedo in case you have any thoughts.

NachoSoto:
Good point we have to be careful with that. One thing to note however is that the SDK will never make requests to this endpoint by itself. The only public API for it is `PurchasesDiagnostics`, which is only meant for debugging purposes anyway.

Tests/UnitTests/Networking/HTTPClientTests.swift
aboedo: oops

Sources/Networking/HTTPClient/HTTPRequest.swift
```diff
@@ -119,6 +119,24 @@ extension HTTPRequest.Path {
...
+                .getIntroEligibility,
+                .logIn,
+                .postAttributionData,
+                .postOfferForSigning,
+                .postReceiptData,
+                .postSubscriberAttributes,
+                .postAdServicesToken:
+            return true
+        case .health:
+            return false
```
aboedo, line 136: nice! 
absolute nitpick, FFTI: I would name this `shouldCache` or something like that, to reflect that it's not something that _has_ already happened, rather, whether they should be. 
Even further, this isn't entirely a cache either - eTags are just about saving up some kilobytes of HTTP traffic, but it's not a traditional cache in that it doesn't have a specific TTL or anything.
So maybe something like `shouldSendEtag` would be more representative of that

aboedo (approved):

Sources/Networking/HTTPClient/HTTPRequest.swift
NachoSoto: > I would name this shouldCache or something like that, to reflect that it's not something that has already happened, rather, whether they should be.

Great point!

> `shouldSendEtag`

💯

Issue #2116, opened 2022-12-05
Error in first call, the user tries to purchase a product


jesus-mg-ios:
- [ ] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Development/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
Lastly, we're dealing with a first error thrown when we want to purchase something. The first time it throws a throttling error and the second time (not even a second after) the purchase works correctly.

What do we see? -> The purchase is canceled without the system sheet to pay is shown.

1. Environment
   1. Platform: iOS 
   2. SDK version: 3.14.3
   3. StoreKit 2 (_disabled with `useStoreKit2IfEnabled(false)`_) (Y/N): N
   4. OS version: 15.x, 16.x
   5. Xcode version: 14.3.1
   6. How widespread is the issue. Percentage of devices affected: Affects on App Store review, it is not demonstrated to happen in production. 
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
```logs
2022-12-03 12:31:51.715719+0100 app[673:61074] [Purchases] - DEBUG: 👤 Identifying App User ID
2022-12-03 12:31:51.732816+0100 app[673:61074] [Purchases] - DEBUG: ℹ️ Debug logging enabled
2022-12-03 12:31:51.733040+0100 app[673:61074] [Purchases] - DEBUG: ℹ️ SDK Version - 4.14.3
2022-12-03 12:31:51.733215+0100 app[673:61074] [Purchases] - DEBUG: 👤 Initial App User ID set
2022-12-03 12:31:51.747630+0100 app[673:61074] [Purchases] - DEBUG: ℹ️ setting reserved attribute: $AppInstanceId
2022-12-03 12:31:51.752473+0100 app[673:61074] [Purchases] - DEBUG: ℹ️ Delegate set
2022-12-03 12:31:51.776111+0100 app[673:61074] [APP] <RCPurchases: 0x280237a00>, info <CustomerInfo:
originalApplicationVersion=1.0,
latestExpirationDate=Optional(2022-12-03 00:49:22 +0000),
[...]

PR #2280, opened 2023-02-13
`PurchaseOrchestrator`: always refresh receipt purchasing in sandbox

### LINKED ISSUE #2260: 4.17.4 Break purchases

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

1. Environment
   1. Platform: iOS
   2. SDK version: 4.17.4 
   3. StoreKit 2 (_disabled with `useStoreKit2IfEnabled(false)`_) (Y/N):
   4. OS version: 16.x
   5. Xcode version: 14.2.0

**Additional context**

Hi:

We notice through our UITests a purchase flow break, changing the version from 4.17.3 to 4.17.4.

NachoSoto:
Fixes #2260 and [SDKONCALL-216].
Follow up to #2245.

The change in #2245 was meant to avoid receipt refresh throttling errors (#2116).
However, that was a regression when using StoreKit config files because those don't get refreshed in the backend.

Unfortunately the workaround introduced in #1945 prevented us from noticing this, because we use `DangerousSettings.InternalSettings.enableReceiptFetchRetry` in our own integration tests.
I added an integration test that explicitly covers this scenario. That fails without this change if disabling that dangerous setting. I thought about making that test run without the setting, but that might lead to flaky failures in CI, which is why we introduced that workaround in the first place. But at least now we have an explicit integration test for this, as well as 2 unit tests.

[SDKONCALL-216]: https://revenuecats.atlassian.net/browse/SDKONCALL-216?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -787,7 +787,8 @@ private extension PurchasesOrchestrator {
             //      We don't _want_ to always refresh receipts to avoid throttling errors
             //      We don't _need_ to because the receipt will be refreshed by the backend using /verifyReceipt
             // - Debug and sandbox builds (potentially using StoreKit config files):
-            //      We need to always refresh the receipt because the backend might not use /verifyReceipt
+            //      We need to always refresh the receipt because the backend does not use /verifyReceipt
+            //          when it was generated locally with SK config files.
 
             #if DEBUG
             return self.systemInfo.isSandbox
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -781,7 +781,21 @@ private extension PurchasesOrchestrator {
                                              maximumRetries: Self.receiptRetryCount,
                                              sleepDuration: Self.receiptRetrySleepDuration)
         } else {
+            // See https://github.com/RevenueCat/purchases-ios/pull/2245 and
+            // https://github.com/RevenueCat/purchases-ios/issues/2260
+            // - Release or production builds:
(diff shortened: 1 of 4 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -781,7 +781,21 @@ private extension PurchasesOrchestrator {
                                              maximumRetries: Self.receiptRetryCount,
                                              sleepDuration: Self.receiptRetrySleepDuration)
         } else {
+            // See https://github.com/RevenueCat/purchases-ios/pull/2245 and
+            // https://github.com/RevenueCat/purchases-ios/issues/2260
+            // - Release or production builds:
+            //      We don't _want_ to always refresh receipts to avoid throttling errors
+            //      We don't _need_ to because the receipt will be refreshed by the backend using /verifyReceipt
+            // - Debug and sandbox builds (potentially using StoreKit config files):
+            //      We need to always refresh the receipt because the backend might not use /verifyReceipt
```
aboedo, line 790: let's specify that the backend will not use `/verifyReceipt` when the receipt was generated locally with StoreKitTest / StoreKit Configuration files

aboedo (approved):

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: Done.

PR #2286, opened 2023-02-14
`HTTPClient`: automatically add `nonce` based on `HTTPRequest.Path`

NachoSoto:
Instead of manually choosing when to include `nonce` in requests, this data-driven approach ensures that all requests to endpoints that support signature verification will make `HTTPClient` do the right thing.

See also #2277 for how `CustomerInfo` and `EntitlementInfo` will use this.

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPRequest.swift
+++ b/Sources/Networking/HTTPClient/HTTPRequest.swift
@@ -21,11 +21,7 @@ struct HTTPRequest {
     /// If present, this will be used by the server to compute a checksum of the response signed with a private key.
     var nonce: Data?
 
-    init(method: Method, path: Path) {
-        self.init(method: method, path: path, nonce: nil)
-    }
-
-    init(method: Method, path: Path, nonce: Data?) {
+    init(method: Method, path: Path, nonce: Data? = nil) {
         assert(nonce == nil || nonce?.count == Data.nonceLength,
                "Invalid nonce: \(nonce?.description ?? "")")
 
@@ -129,7 +125,7 @@ extension HTTPRequest.Path {
         }
     }
 
-    /// Whether requests to this path can be cached using `ETagManager`
+    /// Whether requests to this path can be cached using `ETagManager`.
     var shouldSendEtag: Bool {
         switch self {
         case .getCustomerInfo,
@@ -143,6 +139,24 @@ extension HTTPRequest.Path {
                 .postAdServicesToken:
             return true
         case .health:
+            return false
+        }
+    }
+
+    // Whether the endpoint will perform signature validation.
+    var hasSignatureValidation: Bool {
+        switch self {
+        case .getCustomerInfo,
+                .logIn,
+                .postReceiptData,
+                .health:
+            return true
+        case .getOfferings,
+                .getIntroEligibility,
+                .postSubscriberAttributes,
+                .postAttributionData,
+                .postAdServicesToken,
+                .postOfferForSigning:
             return false
         }
     }
--- a/Sources/Networking/HTTPClient/HTTPRequest.swift
+++ b/Sources/Networking/HTTPClient/HTTPRequest.swift
@@ -144,7 +144,7 @@ extension HTTPRequest.Path {
     }
 
     // Whether the endpoint will perform signature validation.
-    var hasSignatureValidation: Bool {
+    var supportsSignatureValidation: Bool {
         switch self {
         case .getCustomerInfo,
                 .logIn,
--- a/Sources/Misc/Signing.swift
+++ b/Sources/Misc/Signing.swift
@@ -114,6 +114,13 @@ extension Signing {
             }
         }
 
+        var isEnabled: Bool {
+            switch self {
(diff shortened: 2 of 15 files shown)


bisho:
Does this makes sense? The requests with Nonce will include a validation field and might even error. Shouldn't the callers of this endpoints handle these situations? And if they need to handle them, shouldn't be an explicit HTTPValidatedClient that makes clear the interface?

NachoSoto:
> Shouldn't the callers of this endpoints handle these situations? 

We have 2 `EntitlementValidationMode`s (plus a third default `.disabled`):
- For `EntitlementValidationMode.enforced`, `HTTPClient` will produce `ErrorCode.signatureValidationFailed`, so it'll be automatically forwarded to the user.
- For `EntitlementValidationMode.informationOnly`, `HTTPResponse.validationResult` will contain the relevant information, and the callers need to handle them. You're right that callers need to handle it. `CustomerInfoResponseHandler` is responsible for doing that for all endpoints that return a `CustomerInfo`.

Do you have a specific scenario that you're worried about that's not handled by that?

Sources/Networking/HTTPClient/HTTPRequest.swift
```diff
@@ -147,6 +143,24 @@ extension HTTPRequest.Path {
...
+                .health:
+            return true
+        case .getOfferings,
+                .getIntroEligibility,
+                .postSubscriberAttributes,
+                .postAttributionData,
+                .postAdServicesToken,
+                .postOfferForSigning:
+            return false
+        }
```
aboedo, line 161: regarding @bisho's question... honestly, I think that's a valid concern. When looking at this diff, my first thought was 'oh, maybe we should just add signing to `getIntroEligibility` while we're at it". 

Then I realized that adding it here only works for the "enforced" mode, not "info only", which... tells me that there's a disconnect if it's not set up at the callsite.

bisho: If the validation were to be always strict, and the HTTP client will just raise, I wouldn't complain at adding Nonces at the HTTPClient level based on path.

But right now the HTTP client will most of the time return with some validation result. I think there is a disconnect between the code that will check the validation result and the http client that decides to add nonces to some requests in an obscure and implicit way. They need to be in sync, but there is not api forcing it.

I would prefer if the places that do need signing and will look&handle the validation result, explicitly call an HTTP client with validation. The default http client will not have the validation result. That makes the code solid, you will get an error if you try to check the validation of a request and you are not using a request with Nonce.

NachoSoto: I see what you're saying, it's a consequence of this being data-driven. I think it does simplify things.

I just had a conversation with @aboedo about this and I think we came up with a good middle ground.
The core of the issue comes from the fact that there are 2 reasons why we might want to validate:
1. Validation is enabled by the user
2. We want to validate all requests for a particular path (for example, the `/health` endpoint when we're testing that, or the future entitlement mapping request)

The current `HTTPClient` is mainly designed for the first case, as it's getting the user configuration from `SystemInfo.responseValidationMode`. This `Path` API facilitates that use case, but it does bring up the disconnect that you're describing.

> I would prefer if the places that do need signing and will look&handle the validation result, explicitly call an HTTP client with validation.

This is where we can solve your main concern, improve the API, and simplify the code in `HealthOperation` and future entitlement mapping requests. I'm going to add an optional parameter to `HTTPClient.perform(request:)` that allows overriding `Signing.ResponseValidationMode`. So it would look something like:
```swift
// This throws `ErrorCode.signatureValidationFailed`
let response = try await httpClient.perform(
  .init(method: .get, path: .entitlementMapping),
  validationMode: .enforced
)
```

> you will get an error if you try to check the validation of a request and you are not using a request with Nonce.

That "error" is already represented with `HTTPResponseValidationResult.notRequested`, so users of `HTTPClient` already have the ability to detect that case.

NachoSoto: If that's okay with you both, I'll handle this after this PR. I have 5 open PRs and it's already a lot of work to keep them all in sync.

It'll be a nice refactor to simplify `HealthOperation` 👍🏻

NachoSoto: Actually doing this in this PR, incoming.

Sources/Networking/HTTPClient/HTTPRequest.swift
```diff
@@ -147,6 +143,24 @@ extension HTTPRequest.Path {
         }
     }
 
+    // Whether the endpoint will perform signature validation.
+    var hasSignatureValidation: Bool {
```
tonidero, line 147: Maybe rename to `supportsSignatureValidation`?

tonidero (approved):
LGTM!

Sources/Networking/HTTPClient/HTTPRequest.swift
NachoSoto: That's better 👍🏻

PR #2288, opened 2023-02-15
`EntitlementInfo`: add a grace period limit to outdated entitlements

NachoSoto:
This prevents users from keeping an outdated active entitlement forever.
One could start a free trial then block all further requests from the backend in order to keep that active.

When this check fails, a new warning log is generated:
> WARN: ⚠️ Entitlement is no longer active (expired 2023-02-12 18:15:11 +0000) and it's outside grace period window (last updated 2023-02-12 18:13:31 +0000)

Depends on #2310

### DIFF
--- a/Sources/LocalReceiptParsing/DataConverters/DateFormatter+Extensions.swift
+++ b/Sources/LocalReceiptParsing/DataConverters/DateFormatter+Extensions.swift
@@ -45,7 +45,7 @@ extension ISO8601DateFormatter {
             }
 
             func string(from date: Date) -> String {
-                return ISO8601DateFormatter.withMilliseconds.string(from: date)
+                return ISO8601DateFormatter.noMilliseconds.string(from: date)
             }
         }
 
--- a/Sources/Purchasing/EntitlementInfo.swift
+++ b/Sources/Purchasing/EntitlementInfo.swift
@@ -302,8 +302,15 @@ private extension EntitlementInfo {
             return true
         }
 
-        let referenceDate = Self.referenceDate(for: requestDate)
-        return expirationDate.timeIntervalSince(referenceDate) >= 0
+        let (referenceDate, inGracePeriod) = Self.referenceDate(for: requestDate)
+        let isActive = expirationDate.timeIntervalSince(referenceDate) >= 0
+
+        if !inGracePeriod && !isActive {
+            Logger.warn(Strings.purchase.entitlement_expired_outside_grace_period(expiration: expirationDate,
+                                                                                  reference: requestDate ?? Date()))
+        }
+
+        return isActive
     }
 
     static func willRenewWithExpirationDate(expirationDate: Date?,
@@ -318,13 +325,14 @@ private extension EntitlementInfo {
         return !(isPromo || isLifetime || hasUnsubscribed || hasBillingIssues)
     }
 
-    private static func referenceDate(for requestDate: Date?) -> Date {
-        guard let requestDate = requestDate else { return Date() }
+    private static func referenceDate(for requestDate: Date?) -> (Date, inGracePeriod: Bool) {
+        guard let requestDate = requestDate else { return (Date(), true) }
 
         if Date().timeIntervalSince(requestDate) <= Self.requestDateGracePeriod.seconds {
-            return requestDate
+            return (requestDate, true)
         } else {
-            return Date()
+
+            return (Date(), false)
         }
     }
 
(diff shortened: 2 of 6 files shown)


Sources/Purchasing/EntitlementInfo.swift
```diff
@@ -297,6 +308,16 @@ private extension EntitlementInfo {
         return !(isPromo || isLifetime || hasUnsubscribed || hasBillingIssues)
     }
 
+    private static func referenceDate(for requestDate: Date?) -> Date {
+        guard let requestDate = requestDate else { return Date() }
+
+        if Date().timeIntervalSince(requestDate) <= Self.requestDateGracePeriod.seconds {
+            return requestDate
+        } else {
+            return Date()
```
NachoSoto, line 317: Forgot my own advice, need to add a log here.

NachoSoto:
Don't merge this yet, we need to solve one issue with `requestDate`s not getting updated from Etags.

NachoSoto:
That will be fixed by #2310.

tonidero:
Lmk if you need this reviewed then!

NachoSoto:
Need to finish #2310 before this.

Sources/LocalReceiptParsing/DataConverters/DateFormatter+Extensions.swift
```diff
@@ -45,7 +45,7 @@ extension ISO8601DateFormatter {
             }
 
             func string(from date: Date) -> String {
-                return ISO8601DateFormatter.withMilliseconds.string(from: date)
+                return ISO8601DateFormatter.noMilliseconds.string(from: date)
```
NachoSoto, line 48: JSON requires ISO8601 dates with no milliseconds.
Nothing else uses this, and technically this is more correct as `noMilliseconds` is implemented with only `.withInternetDateTime`.

Tests/UnitTests/Purchasing/CustomerInfoTests.swift
```diff
@@ -13,12 +13,19 @@ import XCTest
 @testable import RevenueCat
 
 class EmptyCustomerInfoTests: TestCase {
+
     func testEmptyDataFails() throws {
-        expect(try CustomerInfo(data: [:])).to(throwError())
+        expect(try CustomerInfo(data: [:])).to(throwError(ErrorCode.customerInfoError))
```
NachoSoto, line 18: Slightly more precise test.

Tests/UnitTests/Purchasing/CustomerInfoTests.swift
```diff
@@ -818,12 +833,12 @@ class BasicCustomerInfoTests: TestCase {
 
     func testCopyWithNewRequestDateUpdatesEntitlements() throws {
         expect(self.customerInfo.entitlements.active).to(haveCount(2))
-        expect(self.customerInfo.entitlements["old_pro"]?.isActive) == false
```
NachoSoto, line 821: This test was added in #2310.
`old_pro` no longer works because of the new grace period, so I've added a new `expired_pro` that expired not too far in the past.

NachoSoto:
@tonidero this is ready now.

tonidero (approved):
LGTM!

tonidero (commented):
@NachoSoto I noticed we didn't update the logic in `CustomerInfo` in https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Identity/CustomerInfo.swift#L359 . We probably should update it there as well?

Sources/Purchasing/EntitlementInfo.swift
```diff
@@ -282,23 +282,41 @@ public extension EntitlementInfo {
...
+    static let requestDateGracePeriod: DispatchTimeInterval = .days(3)
+
+}
+
 // MARK: - Private
 
 private extension EntitlementInfo {
 
-    class func isDateActive(expirationDate: Date?, forRequestDate requestDate: Date?) -> Bool {
+    static func isDateActive(expirationDate: Date?, for requestDate: Date?) -> Bool {
```
tonidero, line 300: Also, while working on the Android side, I noticed we could assume we always have a `requestDate`, so this wouldn't be nullable. I think it would be the same in iOS, but not sure if I'm missing something.

NachoSoto: I think you're right!

NachoSoto: #2325

NachoSoto:
@tonidero great catch! I'm going to make `CustomerInfo` use the same logic.

NachoSoto:
@tonidero done: #2327

Issue #2221, opened 2023-01-18
Error on compile -> Concurrently-executed local function 'complete()' must be marked as '@Sendable'


billypap1:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

1. Environment
   1. Platform: iOS
   2. SDK version: 4.17
   3. StoreKit 2 (_disabled with `useStoreKit2IfEnabled(false)`_) (Y/N): N
   4. OS version: 16.0
   5. Xcode version: 13.3.1
   6. How widespread is the issue. Percentage of devices affected : 100%
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue:  N/A
3. Steps to reproduce, with a description of expected vs. actual behavior
-> Just install the Swift SDK with CocoaPods
4. Other information (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.): N/A

**Additional context**
I have installed RevenueCat via Cocoapods ->   pod 'RevenueCat' (version 4.17) and I receive this error on compile:
**Concurrently-executed local function 'complete()' must be marked as '@Sendable'** in PurchasesOrchestrator.swift file

RCGitBot:
👀 [SDKONCALL-210] We've just linked this issue to our internal tracker and notified the team. Thank you for reporting, we're checking this out!

NachoSoto:
Thanks for the report.

Normally this happens when using an older version of Xcode, but 13.3.1 is still supported.
Could you share complete build logs and how you're building with CocoaPods?

Alternatively I'd recommend upgrading to Xcode 13.4.1.

PR #1994, opened 2022-10-24
Fixed Xcode 13.2.x / Swift 5.5 compatibility

NachoSoto:
Fixes [SDKONCALL-156].

This will be covered by #1990.

[SDKONCALL-156]: https://revenuecats.atlassian.net/browse/SDKONCALL-156?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2292, opened 2023-02-17
Fixed `PurchasesOrchestrator` compilation error on Xcode 14.3 beta 1

### LINKED ISSUE #2291: Xcode 14.3 support
image

NachoSoto:
Fixes #2291.

This has led to issues in the past too (#1994, #2221, 	https://github.com/RevenueCat/purchases-flutter/issues/516, https://github.com/RevenueCat/purchases-flutter/issues/507) because I tried to reuse that code to call a single completion method.
The new version of Swift makes that impossible now, because it detects that `@MainActor` isn't available in iOS 12.

To finally work around all those issues, this simply duplicates this single line of code, which should work across all versions of Swift.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -957,23 +957,12 @@ private extension PurchasesOrchestrator {
             case let .failure(error):
                 let purchasesError = error.asPublicError
 
-                #if swift(>=5.6)
-                @MainActor
-                func complete() {
-                    completion?(transaction, nil, purchasesError, false)
-                }
-                #else
-                @Sendable
-                @MainActor
-                func complete() {
-                    completion?(transaction, nil, purchasesError, false)
-                }
-                #endif
-
                 if finishable {
-                    self.finishTransactionIfNeeded(transaction) { complete() }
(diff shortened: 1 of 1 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -957,23 +957,12 @@ private extension PurchasesOrchestrator {
...
-                #endif
-
                 if finishable {
-                    self.finishTransactionIfNeeded(transaction) { complete() }
+                    self.finishTransactionIfNeeded(transaction) {
+                        completion?(transaction, nil, purchasesError, false)
+                    }
                 } else {
-                    complete()
+                    completion?(transaction, nil, purchasesError, false)
```
aboedo, line 965: honestly this is easier to read too

aboedo (approved):

tonidero (approved):

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: Yeah totally 🤷🏻‍♂️ not sure why I didn't realize it earlier that we could go back to this.

PR #2293, opened 2023-02-17
Deprecate `usesStoreKit2IfAvailable`

aboedo:
There's been a lot of customer confusion with the use SK2 parameter. 
In order to reduce confusion, and to improve the overall experience, we're moving into a more opinionated approach. 
Currently, we're defaulting to SK1-only. 

Once we get diagnostics data in, we'll be able to get more visibility into real-world, production performance of StoreKit 1 vs StoreKit 2 on a per-API basis. Using this data, we will implement remote configuration to ensure that the best implementation is utilized for each specific functionality, optimizing for maximum stability and performance. 

Until then, we can save our customers from confusion and headache by going back to the simplest approach. 

### What this looks like at the callsite

Screenshot 2023-02-17 at 2 38 02 PM

### What the relevant DocC page looks like

Screenshot 2023-02-17 at 2 28 20 PM

### DIFF
--- a/Sources/Misc/Deprecations.swift
+++ b/Sources/Misc/Deprecations.swift
@@ -402,8 +402,9 @@ public extension Configuration.Builder {
     RevenueCat currently uses StoreKit 1 for purchases, as its stability in production scenarios has
     proven to be more performant than StoreKit 2.
 
-    In the future the SDK will selectively use StoreKit 2 APIs for certain features where
-    StoreKit 2 APIs have been proven to perform better than StoreKit 1.
+    We're collecting more data on the best approach, but StoreKit 1 vs StoreKit 2 is an implementation detail
+    that you shouldn't need to care about.
+    
     Simply remove this method call to let RevenueCat decide for you which StoreKit implementation to use.
     """)
     @objc func with(usesStoreKit2IfAvailable: Bool) -> Configuration.Builder {
--- a/Sources/Misc/Deprecations.swift
+++ b/Sources/Misc/Deprecations.swift
@@ -391,7 +391,7 @@ extension CustomerInfo {
 }
 
 public extension Configuration.Builder {
-    ///
+
     /// Set `usesStoreKit2IfAvailable`. If `true`, the SDK will use StoreKit 2 APIs internally. If disabled, it will use StoreKit 1 APIs instead.
     /// - Parameter usesStoreKit2IfAvailable: enable StoreKit 2 on devices that support it.
     /// Defaults to  `false`.
@@ -404,13 +404,14 @@ public extension Configuration.Builder {
 
     We're collecting more data on the best approach, but StoreKit 1 vs StoreKit 2 is an implementation detail
     that you shouldn't need to care about.
-    
+
     Simply remove this method call to let RevenueCat decide for you which StoreKit implementation to use.
     """)
     @objc func with(usesStoreKit2IfAvailable: Bool) -> Configuration.Builder {
         self.storeKit2Setting = .init(useStoreKit2IfAvailable: usesStoreKit2IfAvailable)
         return self
     }
+
 }
 
 // swiftlint:enable line_length missing_docs file_length
--- a/Sources/Misc/Deprecations.swift
+++ b/Sources/Misc/Deprecations.swift
@@ -391,12 +391,19 @@ extension CustomerInfo {
 }
 
 public extension Configuration.Builder {
-    /**
-     * Set `usesStoreKit2IfAvailable`.
-     * - Parameter usesStoreKit2IfAvailable: opt in using StoreKit 2 on devices that support it.
-     * Defaults to  `false`.
-     */
-    @available(*, deprecated, message: "Use NonSubscriptionTransaction")
+    ///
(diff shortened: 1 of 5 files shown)


aboedo:
🤔 guess we don't have a great way to label deprecations for release trains, do we?

Sources/Misc/Deprecations.swift
```diff
@@ -389,3 +389,26 @@ extension CustomerInfo {
...
+    ///
+    /// Set `usesStoreKit2IfAvailable`. If `true`, the SDK will use StoreKit 2 APIs internally. If disabled, it will use StoreKit 1 APIs instead.
+    /// - Parameter usesStoreKit2IfAvailable: enable StoreKit 2 on devices that support it.
+    /// Defaults to  `false`.
+    /// - Important: This configuration flag has been deprecated, and will be replaced by automatic remote configuration in the future.
+    /// However, apps using it should work correctly.
+    ///
+    @available(*, deprecated, message: """
+    RevenueCat currently uses StoreKit 1 for purchases, as its stability in production scenarios has overall
+    proved more better than StoreKit 2.
```
yousefcat, line 403: I would change this:

> proved more better

to

> proved more performant

tonidero (approved):
> 🤔 guess we don't have a great way to label deprecations for release trains, do we?

Nope... Most times when I'm not sure how to label it I've just labeled it `refactor`... But I think this is fine as well.

Also should we rename the PR to `deprecate usesStoreKit2IfAvailable`? So it's clearer in the autogenerated changelog.

tonidero:
Though tests seems to be failing...

Sources/Misc/Deprecations.swift
```diff
@@ -389,3 +389,26 @@ extension CustomerInfo {
...
+    However, in the future the SDK will be opinionated and selectively use StoreKit 2 APIs for certain features where
```
yousefcat, line 404: I would make this,

> However, in the future the SDK will be opinionated and selectively use StoreKit 2 APIs for certain features where

more concise,

> In the future the SDK will selectively use StoreKit 2 APIs for certain features where

Sources/Misc/Deprecations.swift
```diff
@@ -389,3 +389,26 @@ extension CustomerInfo {
...
+public extension Configuration.Builder {
+    ///
+    /// Set `usesStoreKit2IfAvailable`. If `true`, the SDK will use StoreKit 2 APIs internally. If disabled, it will use StoreKit 1 APIs instead.
+    /// - Parameter usesStoreKit2IfAvailable: enable StoreKit 2 on devices that support it.
+    /// Defaults to  `false`.
+    /// - Important: This configuration flag has been deprecated, and will be replaced by automatic remote configuration in the future.
+    /// However, apps using it should work correctly.
+    ///
+    @available(*, deprecated, message: """
+    RevenueCat currently uses StoreKit 1 for purchases, as its stability in production scenarios has overall
```
yousefcat, line 402: I would just remove the word,

> overall

from the end.

yousefcat (commented):
Left a few comments on the message.

Sources/Misc/Deprecations.swift
aboedo: lol "more better" 
I just kept updating the copy until I totally messed it up

aboedo:
@tonidero optimally, we'd have a `deprecation` label, so we can put the related changes first and foremost - this should look pretty important in a changelog

aboedo:
tests are angry because we're using the deprecated method

aboedo:
@NachoSoto thoughts on addressing tests for this? It feels odd to not be able to use something for tests, even if it's deprecated

tonidero:
> @tonidero optimally, we'd have a `deprecation` label, so we can put the related changes first and foremost - this should look pretty important in a changelog

Good point. Until now, we did this manually, but if we can isolate the deprecation to a separate PR and add support for that label, that's probably better

NachoSoto:
I know of a workaround that we've used for other things, I can do this later.

Sources/Misc/Deprecations.swift
```diff
@@ -389,3 +389,28 @@ extension CustomerInfo {
     }
 
 }
+
+public extension Configuration.Builder {
+    ///
+    /// Set `usesStoreKit2IfAvailable`. If `true`, the SDK will use StoreKit 2 APIs internally. If disabled, it will use StoreKit 1 APIs instead.
+    /// - Parameter usesStoreKit2IfAvailable: enable StoreKit 2 on devices that support it.
+    /// Defaults to  `false`.
+    /// - Important: This configuration flag has been deprecated, and will be replaced by automatic remote configuration in the future.
```
NachoSoto, line 398: Can you explain what you mean by "automatic remote configuration"?

NachoSoto:
Actually I thought we'd need a different workaround. For those 2 tests we can just do this: #2295

NachoSoto (approved):

Sources/Misc/Deprecations.swift
tonidero: Hmm well, this might be going too ahead with this comment, but I believe the intention is to support changing this dynamically from the backend. So we could decide, according to the status returned by diagnostics or other sources, change whether apps use SK1 or SK2.

NachoSoto: Oooh that makes sense

PR #2294, opened 2023-02-17
clarifies error messages for storekit 1 bugs

aboedo:
We have a few error messages for StoreKit 1 bugs that look pretty scary. 

However, they really aren't - while they surface genuine bugs in StoreKit, they require no action on the developer's side, and our backend won't be affected. So in practice, they're not useful. 

This PR updates the copy so that: 
- it doesn't suggest using SK2
- it moves from `warn` to `verbose`, since there's really no action on the developer's side needed
- it clarifies what's going on under the hood

| Before | After |
| :-: | :-: |
| Screenshot 2023-02-17 at 3 43 02 PM | Screenshot 2023-02-17 at 3 52 05 PM |


NachoSoto (approved):
Agreed 👍🏻 these 3 aren't worth letting users know.

PR #2297, opened 2023-02-18
SwiftUI: Paywall View should respond to changes on the UserView model

ConfusedVorlon:
I hit this bug when I first tried the example. My offerings loaded slowly - so the paywall was empty, and didn't update when the offerings were retrieved.

### Checklist
nope

### Motivation
swiftui example app doesn't update paywall when offerings are retrieved

### Description
app stores current state in the shared UserViewModel, but the PaywallView doesn't currently observe changes on this model - so it only shows the state at the point where the view was shown.
This is potentially a state where offerings have not yet been retrieved.


NachoSoto (approved):
Thanks! 👍🏻

PR #2300, opened 2023-02-21
Fix typos (in comments)

MortenGregersen:
### Checklist

### Motivation
I found some extra f-ses.

### Description
I just removed them.

### DIFF
--- a/Sources/FoundationExtensions/Array+Extensions.swift
+++ b/Sources/FoundationExtensions/Array+Extensions.swift
@@ -26,7 +26,7 @@ extension Array {
 
 extension Collection {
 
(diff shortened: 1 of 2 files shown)


NachoSoto (changes requested):
Thanks, but this is on purpose: https://en.wikipedia.org/wiki/If_and_only_if

MortenGregersen:
Fine 👍 Never heard of it.

aboedo (commented):
maybe we could explicitly update to `if and only if`, since I'm not sure `iff` is that well-known

aboedo:
at least for public-facing stuff

PR #2302, opened 2023-02-21
Update `Gemfile.lock`

NachoSoto:


Gemfile.lock
```diff
@@ -338,7 +338,7 @@ GEM
     zeitwerk (2.6.6)
 
 PLATFORMS
-  ruby
+  arm64-darwin-22
```
tonidero, line 341: Hmm I'm not sure if this will have any issues in any of the CircleCI machines...

Gemfile.lock
```diff
@@ -88,18 +88,18 @@ GEM
     concurrent-ruby (1.2.0)
     cork (0.3.0)
       colored2 (~> 3.1)
-    danger (9.2.0)
+    danger (8.6.1)
```
tonidero, line 91: This seems like a downgrade, but I think it's probably ok, I know bumping dunger actually broke some things in Android, not sure if it's also causing any issues for you?

tonidero (approved):
If tests pass I think this should be fine, but I'm not sure with the Platforms change...

Gemfile.lock
NachoSoto: Yeah let me reverse this.

Gemfile.lock
NachoSoto: Yeah I was confused, but that's what I get when I do `gem install`?

tonidero: Hmm I see the `Gemfile` doesn't specify the version, so it should be using the latest... Maybe you are using a cached version? But that's weird, I don't know how that happened either 😅

PR #2309, opened 2023-02-21
`Signing`: added request time & eTag to signature verification

NachoSoto:
See https://github.com/RevenueCat/khepri/pull/5300

### DIFF
--- a/Sources/Misc/Signing+ResponseVerification.swift
+++ b/Sources/Misc/Signing+ResponseVerification.swift
@@ -85,13 +85,15 @@ extension HTTPResponse where Body == Data {
         let eTag = HTTPResponse.value(forCaseInsensitiveHeaderField: HTTPClient.ResponseHeader.eTag.rawValue,
                                       in: headers)
 
+        let messageToSign = statusCode == .notModified
+            ? eTag?.asData ?? .init()
+            : body
+
         if signing.verify(signature: signature,
                           with: .init(
-                            notModifiedResponse: statusCode == .notModified,
-                            message: body,
+                            message: messageToSign,
                             nonce: nonce,
-                            requestTime: requestTime,
-                            eTag: eTag
+                            requestTime: requestTime
                           ),
                           publicKey: publicKey) {
             return .verified
--- a/Sources/Misc/Signing+ResponseVerification.swift
+++ b/Sources/Misc/Signing+ResponseVerification.swift
@@ -40,6 +40,7 @@ extension HTTPResponse where Body == Data {
             body: body,
             verificationResult: Self.verificationResult(
                 body: body,
+                statusCode: statusCode,
                 headers: headers,
                 request: request,
                 publicKey: publicKey,
@@ -48,8 +49,10 @@ extension HTTPResponse where Body == Data {
         )
     }
 
+    // swiftlint:disable:next function_parameter_count
     private static func verificationResult(
         body: Data,
+        statusCode: HTTPStatusCode,
         headers: HTTPClient.ResponseHeaders,
         request: HTTPRequest,
         publicKey: Signing.PublicKey?,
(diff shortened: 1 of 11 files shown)


Sources/Misc/Signing+ResponseVerification.swift
```diff
@@ -90,31 +112,18 @@ extension Result where Success == Data?, Failure == NetworkError {
         verificationMode: Signing.ResponseVerificationMode
     ) -> Result<HTTPResponse<Data?>, Failure> {
         return self.flatMap { body in
-            if let body = body {
```
NachoSoto, line 93: No more special casing here 🎉

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -73,16 +73,28 @@ class HTTPClient {
...
-    static let responseSignatureHeaderName = "X-Signature"
+    enum RequestHeader: String {
+
+        case authorization = "Authorization"
+        case nonce = "X-Nonce"
+        case eTag = "X-RevenueCat-ETag"
+
+    }
+
+    enum ResponseHeader: String {
```
tonidero, line 91: I like this refactor 👍

Sources/Misc/Signing+ResponseVerification.swift
```diff
@@ -90,31 +112,18 @@ extension Result where Success == Data?, Failure == NetworkError {
...
-                )
-
-                if response.verificationResult == .failed, case .enforced = verificationMode {
-                    return .failure(.signatureVerificationFailed(path: request.path))
-                } else {
-                    return .success(response.mapBody(Optional.some))
-                }
+            let response = HTTPResponse.create(
+                with: response,
+                body: body ?? .init(),
```
tonidero, line 117: Hmm I wonder if we should make the `HTTPResponse.body` parameter nullable instead of setting to an empty `Data`... We would have to guard against that later on but seems like a better representation? NABD though.

Sources/Misc/Signing.swift
```diff
@@ -178,6 +187,23 @@ extension Signing {
 
 }
 
+extension Signing.SignatureParameters {
+
+    var asData: Data {
+        // Not modified responses sign the Etag
+        let payload: Data = self.notModifiedResponse
```
tonidero, line 194: `payload` doesn't seem like a good name if this could have the etag. Don't know how to name it though... maybe `signatureMessage`? NABD though, I think it's easy enough to understand

tonidero (approved):
Nice! Looks great!

Sources/Misc/Signing+ResponseVerification.swift
```diff
@@ -61,17 +64,36 @@ extension HTTPResponse where Body == Data {
...
 
-        if signing.verify(message: body,
-                          nonce: nonce,
-                          hasValidSignature: signature,
-                          with: publicKey) {
+        guard let requestTimeString = HTTPResponse.value(
+            forCaseInsensitiveHeaderField: HTTPClient.ResponseHeader.requestTime.rawValue,
+            in: headers
+        ), let requestTime = Int(requestTimeString) else {
+            Logger.warn(Strings.signing.request_time_missing_from_headers(request))
```
bisho, line 80: Only log if signature is requested?

Sources/Misc/Signing.swift
```diff
@@ -178,6 +187,23 @@ extension Signing {
...
+        ? self.eTag?.asData ?? Data()
+        : self.message
```
bisho, line 196: Given etag and message are exclusive, I would omit the etag from the signing parameters. For not modified responses the message IS the etag. That way you avoid having to have request-level logic in the signing code, gives you better abstraction IMHO. The signature for signature check should just be "local_nonce, server_timestamp, message, signature". And and leave the not-modified handling to the http request.

Tests/UnitTests/Misc/SigningTests.swift
```diff
@@ -68,76 +68,151 @@ class SigningTests: TestCase {
...
-                           nonce: "nonce".asData,
-                           hasValidSignature: "invalid signature".asData.base64EncodedString(),
-                           with: Signing.loadPublicKey())
+        _ = Signing.verify(signature: "invalid signature".asData.base64EncodedString(),
+                           with: .init(
+                            notModifiedResponse: false,
+                            message: "Hello World".asData,
+                            nonce: "nonce".asData,
+                            requestTime: 1677005916012,
+                            eTag: nil
```
bisho, line 112: Include etag for the text. 200 responses will do have etag field.

Sources/Misc/Signing+ResponseVerification.swift
NachoSoto: `HTTPResponse` is generic over its `Body` type.
Initially we do create an `HTTPResponse<Data?>`, and this method returns that, but ``Signing.verify` takes a body as `Data`. I think it's simpler to deal with only one valid representation of an empty body.
This is the same thing with `[]`, `{}, `""`. Ideally the language allows a way to represent in the type system a value that can't be empty, but since we can't, I think it's better to make the parameter to `Signing.verify` non-optional so we don't have to handle both cases.

Sources/Misc/Signing.swift
NachoSoto: I agree 👍🏻  I copied this from Khepri but I'll rename it.

Sources/Misc/Signing+ResponseVerification.swift
NachoSoto: This method starts with:
```swift
guard let nonce = request.nonce,
      let publicKey = publicKey,
      #available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *) else {
    return .notVerified
}
```
So this won't be executed if there is no signature requested.

Sources/Misc/Signing.swift
NachoSoto: Oh that's a great observation!
That was a nice small refactor, thanks: 59cdce772278473f458f357f883cb034629bf51d

Tests/UnitTests/Misc/SigningTests.swift
NachoSoto: No longer relevant with the other refactor 👍🏻

PR #2310, opened 2023-02-21
Update `CustomerInfo.requestDate` from 304 responses

NachoSoto:
### Purpose:

This is a prerequisite for #2288. The new grace period means that if we didn't update the date from the cached responses, entitlements would become stale and expired after 3 days.

### Changes:
- Added `HTTPResponseBody.copy(with newRequestDate:)` to be able to modify responses from the header request date
- `CustomerInfo` implements this method to modify its request date
- Added `HTTPResponse.requestDate` to be able to keep track of the request date from the server
- Added `VerificationResult.from(cache:response:)` to determine the most restrictive verification result based on what's cached or checked from a response
- Added `HTTPResponse.copy(with:)` to modify the verification result of a response using the previous method
- `HTTPClient` now uses the most restrictive verification result
- `HTTPClient` updates request date from server responses or cached responses (unless verification failed). This was the missing piece for #2288.

### Tests:
- Verify that `ETagManager` does not use an ETag if verification was previously not enabled
- Verify that `ETagManager` returns the request date from the server when returning a cached response
- Verify that `HTTPClient` updates the request date from the server or from a cached response
- Verify that `HTTPClient` does not update request date if verification failed
- Tests for `HTTPResponse` request date parsing
- Test for `CustomerInfoResponseHandler` updating request date
- Tests for `CustomerInfo.copy(with newRequestDate:)`

### Other smaller changes:
- Moved `CustomerInfo.asData()` into `Encodable.asJSONEncodedData()`
- Renamed `requestTime` to `requestDate` everywhere for consistency

### DIFF
--- a/Sources/Networking/HTTPClient/ETagManager.swift
+++ b/Sources/Networking/HTTPClient/ETagManager.swift
@@ -59,7 +59,7 @@ class ETagManager {
 
         guard let eTagInResponse = eTagInResponse else { return resultFromBackend }
         if self.shouldUseCachedVersion(responseCode: statusCode) {
-            if let storedResponse = self.storedHTTPResponse(for: request) {
+            if let storedResponse = self.storedHTTPResponse(for: request, withRequestDate: response.requestDate) {
                 return storedResponse
             }
             if retried {
@@ -107,8 +107,8 @@ private extension ETagManager {
         }
     }
 
-    func storedHTTPResponse(for request: URLRequest) -> HTTPResponse<Data>? {
-        return self.storedETagAndResponse(for: request)?.asResponse
+    func storedHTTPResponse(for request: URLRequest, withRequestDate requestDate: Date?) -> HTTPResponse<Data>? {
+        return self.storedETagAndResponse(for: request)?.asResponse(withRequestDate: requestDate)
     }
 
     func storeStatusCodeAndResponseIfNoError(for request: URLRequest,
@@ -185,11 +185,12 @@ extension ETagManager.Response {
         return try? JSONEncoder.default.encode(self)
     }
 
-    fileprivate var asResponse: HTTPResponse<Data> {
+    fileprivate func asResponse(withRequestDate requestDate: Date?) -> HTTPResponse<Data> {
         return HTTPResponse(
             statusCode: self.statusCode,
             responseHeaders: [:],
             body: self.data,
+            requestDate: requestDate,
             verificationResult: self.verificationResult
         )
     }
--- a/Sources/Purchasing/VerificationResult.swift
+++ b/Sources/Purchasing/VerificationResult.swift
@@ -69,3 +69,29 @@ extension VerificationResult: DefaultValueProvider {
     static let defaultValue: Self = .notVerified
 
 }
+
+extension VerificationResult {
+
+    /// - Returns: the most restrictive ``VerificationResult`` based on the cached verification and
+    /// the response verification.
+    static func from(cache cachedResult: Self, response responseResult: Self) -> Self {
+        switch (cachedResult, responseResult) {
+        case (.notVerified, .notVerified),
+            (.verified, .verified),
+            (.failed, .failed):
+            return cachedResult
+
+        // These shouldn't happen because `ETagManager` will ignore not verified cached responses
+        // if verification is enabled
+        case (.notVerified, .verified): return .notVerified
+        case (.notVerified, .failed): return .failed
+
+        case (.verified, .notVerified): return .notVerified
+        case (.verified, .failed): return .failed
+
+        case (.failed, .notVerified): return .failed
+        case (.failed, .verified): return .failed
+        }
+    }
+
+}
--- a/Sources/Purchasing/VerificationResult.swift
+++ b/Sources/Purchasing/VerificationResult.swift
@@ -81,14 +81,15 @@ extension VerificationResult {
             (.failed, .failed):
             return cachedResult
 
-        // These shouldn't happen because `ETagManager` will ignore not verified cached responses
-        // if verification is enabled
-        case (.notVerified, .verified): return .notVerified
-        case (.notVerified, .failed): return .failed
-
         case (.verified, .notVerified): return .notVerified
         case (.verified, .failed): return .failed
 
+        // These shouldn't happen because `ETagManager` will ignore not verified cached responses
+        // if verification is enabled.
+        case (.notVerified, .verified): return .notVerified
+        case (.notVerified, .failed): return .failed
+
+        // These shouldn't happen because `ETagManager` won't store responses with failed verification.
         case (.failed, .notVerified): return .failed
         case (.failed, .verified): return .failed
         }
--- a/Tests/UnitTests/Networking/HTTPClientTests.swift
+++ b/Tests/UnitTests/Networking/HTTPClientTests.swift
@@ -938,6 +938,7 @@ final class HTTPClientTests: BaseHTTPClientTests {
             "test": "data"
         ])
         let eTag = "tag"
+        let requestDate = Date().addingTimeInterval(-1000000)
 
         self.eTagManager.stubResponseEtag(eTag)
         self.eTagManager.shouldReturnResultFromBackend = false
@@ -945,7 +946,8 @@ final class HTTPClientTests: BaseHTTPClientTests {
             statusCode: .success,
             responseHeaders: [:],
             body: mockedCachedResponse,
-            verificationResult: .verified
+            requestDate: requestDate,
+            verificationResult: .notVerified
         )
 
         stub(condition: isPath(path)) { response in
@@ -965,7 +967,8 @@ final class HTTPClientTests: BaseHTTPClientTests {
         expect(response).toNot(beNil())
         expect(response?.value?.statusCode) == .success
         expect(response?.value?.body) == mockedCachedResponse
-        expect(response?.value?.verificationResult) == .verified
+        expect(response?.value?.requestDate) == requestDate
+        expect(response?.value?.verificationResult) == .notVerified
 
         expect(self.eTagManager.invokedETagHeaderParametersList).to(haveCount(1))
         expect(self.eTagManager.invokedETagHeaderParameters?.signatureVerificationEnabled) == false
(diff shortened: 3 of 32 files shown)


NachoSoto:
Will include this in #2309

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -1150,8 +1153,87 @@ final class HTTPClientTests: BaseHTTPClientTests {
             .toNot(contain(unexpectedDNSError.description))
     }
 
+    func testResponseFromServerUpdatesRequestDate() throws {
+        let path: HTTPRequest.Path = .mockPath
+        let mockedResponse = BodyWithDate(data: "test", requestDate: Date().addingTimeInterval(-3000000))
+        let encodedResponse = try mockedResponse.asJSONEncodedData()
+        let requestDate = Date().addingTimeInterval(-100000)
```
tonidero, line 1160: What do these numbers mean? We could also hardcode a date for this and the `mockedResponse` request date to make it more deterministic.

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -1458,6 +1540,49 @@ final class SignatureVerificationHTTPClientTests: BaseHTTPClientTests {
         expect(response?.value?.verificationResult) == .verified
     }
 
+    func testCachedResponseDoesNotUpdateRequestDateIfNewResponseIsNotVerified() throws {
```
tonidero, line 1543: Should we have a test testing that it does update if verified? It might be tricky though...

tonidero (approved):
Just a possible edge case I was thinking about but looking great!

bisho:
Can you add a test in case the server returns a 500 error? Want to ensure that is not treated as verification failure even if the date header is not present.

Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -185,11 +185,12 @@ extension ETagManager.Response {
...
     }
 
-    fileprivate var asResponse: HTTPResponse<Data> {
+    fileprivate func asResponse(withRequestDate requestDate: Date?) -> HTTPResponse<Data> {
         return HTTPResponse(
             statusCode: self.statusCode,
             responseHeaders: [:],
             body: self.data,
+            requestDate: requestDate,
             verificationResult: self.verificationResult
```
bisho, line 194: We should only update the requestDate if the verification result of the NOT_MODIFIED response is green, right?

Sources/Purchasing/VerificationResult.swift
```diff
@@ -69,3 +69,29 @@ extension VerificationResult: DefaultValueProvider {
...
+
+        // These shouldn't happen because `ETagManager` will ignore not verified cached responses
+        // if verification is enabled
+        case (.notVerified, .verified): return .notVerified
+        case (.notVerified, .failed): return .failed
+
+        case (.verified, .notVerified): return .notVerified
+        case (.verified, .failed): return .failed
+
+        case (.failed, .notVerified): return .failed
```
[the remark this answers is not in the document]
bisho: Perhaps we should just store wether the validation of the request was ok/bad and map that to the whole verified/notverified/failed options on the fly given the config setting.

It might simplify the logic about the signature validation in a few places...

Thus, changes in enforcement do not affect the cached data. The only path to care will be in case of enforcing and cached data is BAD, which is is initial upgrade path we have already considered (drop cache, depend on network retrieval)

Sources/Networking/HTTPClient/ETagManager.swift
bisho: And do we persist this date change in the cache?
Imagine:
1) 200 OK, valid signature, with request time t=1
2) 304 NOT_MODIFIED, valid signature, with request time t=2
3) Timeout or error 500... What the SDK will retrieve t=1 or t=2. Should be t=2 (and valid signature)

bisho: Similarly:
1) 200 OK, valid signature with request time t=1. SDK gets (valid signature, t=1)
2) 304 NOT_MODIFIED, INVALID signature, with request time t=2. SDK should get (valid signature, t=1) 
3) Timeout or error 500, SDK should get (valid signature, t=1) 

1) 200 OK valid signature, t=1. SDK gets (valid signature, t=1)
2) 200 OK INVALID SIGNATURE, t=2. SDK gets (valid signature, t=1)
3) Timeout or error 500, SDK should get (valid signature, t=1)

tonidero: Nacho can correct me but I believe we currently cache the request time not at the response level but at the model level. Right now we only cache it for the CustomerInfo model so we only store it in those models. We don't update the response cache on not modified responses either

Sources/Purchasing/VerificationResult.swift
```diff
@@ -69,3 +69,29 @@ extension VerificationResult: DefaultValueProvider {
     static let defaultValue: Self = .notVerified
 
 }
+
+extension VerificationResult {
+
+    /// - Returns: the most restrictive ``VerificationResult`` based on the cached verification and
+    /// the response verification.
+    static func from(cache cachedResult: Self, response responseResult: Self) -> Self {
```
bisho, line 77: Oh, I see that here you will return the response as not valid if a 304 is not valid. I don't think we should do this. We should act as unable to fetch. There is a legit situation for this, imagine you are in a airport wifi that intercepts and mangles requests. It should not break your cache.

I think I still have trouble with the logic for validation being in two places.

I would rather do something like:
- For validation errors (either 200, 403...), if there is cache, treat this as a 500, just return cache
- For 403 update request time in cache only if cache was checked valid and 304 was valid too (304 invalid would be covered in the above case anyway). Then return cache.
- For valid 200, update cache, return fresh data

Pseudocode:
```
# On requests:
if cache and (cacheValidationOk or SDKValidationDisabled):
    request_with_cache_etag()
else:
    request_without_etag()
```

```
def get_response_or_cache():
# On responses:
wrong_response = response_not_modified and (we_did_not_sent_etag or returned_etag_mistmatch)
  return FailedRequest  # This should not happen, we only expect not modified if we have a etag cached reponse. If the cache data is not trusted we should NOT send etag always to force a refetch (unless SDKValidationDisabled).

if failedRequest or wrong_response:
    if cache:
        return cache
    else:
        return FailedRequest

trust_response = responseValidationOk or SDKValidationDisabled

if trust_response:
    if not_modified:
        update_cache_request_time()
        # While we might trust the response if validation is disabled, we still track
        # wether the cached data is valid or not, and it will be only valid if cache
        # and requests are both valid
        update_cache_validation_to(responseValidationOk and cacheValidationOk)
        return cache
    else:
        store_new_response_in_cache()
        update_cache_validation_to(responseValidationOk)
        return response
elif cacheValidationOk:
    # If we are not trusting response, prefer valid cache always
    return cache
else:
    # We do not trust response nor cache
    if not_modified:
        return cache
    else:
        store_new_response_in_cache()
        update_cache_validation_to(validationFailed)
        return response
```

This code will handle properly the cache and validation status. In some cases this will return responses with validation failures, but it will try its best to use valid cache over a failed validation response (unless we are using the disable validation mode, in which case uses all responses valid or not (but still tracks validation state irregardless).

Above this, you can put the enforcer layer for strict mode. Having it separate simplifies the request / cache mangling layer.
```
    response = get_response_or_cache()
    if SDKValidationStrictMode and not response.isValid:
        return failedValidation
    return response
```

bisho: It took me a full hour write what I was thinking was going to be simple pseudocode. This request handling logic is truly complicated XDDD

Perhaps it is better to sync. Ping me if you are available.

Sources/Purchasing/VerificationResult.swift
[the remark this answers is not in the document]
NachoSoto: Responses that fail validation are never cached though (that's tested too).

Sources/Networking/HTTPClient/ETagManager.swift
NachoSoto: Short answer is t2.

@bisho you're conflating 2 separate layers of caching (which is normal because you're not familiar with the codebase).

If the server fails to return a 304 (500 or timeout), `HTTPClient` won't return the cached data.
That caching would be handled by `DeviceCache`, which would receive the updated request time.

NachoSoto: > We should only update the requestDate if the verification result of the NOT_MODIFIED response is green, right?

That I think I have a test for but let me check.

NachoSoto: > 304 NOT_MODIFIED, INVALID signature, with request time t=2. SDK should get (valid signature, t=1)

Oh that's a case I didn't consider.
Are we sure we want to do that? That would break an app without any indication that this is happening.

Sources/Purchasing/VerificationResult.swift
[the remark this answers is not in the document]
tonidero: Ah right, so these cases should never happen either. Maybe we should make it clearer grouping all the cases that should never happen. The current comment above makes me think the only cases shouldn't happen are the `notVerified` cached ones.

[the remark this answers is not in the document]
NachoSoto: Done.

Tests/UnitTests/Networking/HTTPClientTests.swift
NachoSoto: It's just a simpler way to make a date that's not "now" to avoid false negatives.

Tests/UnitTests/Networking/HTTPClientTests.swift
NachoSoto: That's tested by `testResponseFromServerUpdatesRequestDate`.

NachoSoto: But I guess it doesn't hurt to have another one where it's also verified, so let me do that.

NachoSoto: Done.

Sources/Networking/HTTPClient/ETagManager.swift
NachoSoto: > We should only update the requestDate if the verification result of the NOT_MODIFIED response is green, right?

That's already done:
```swift
fileprivate func copyWithNewRequestDate() -> Self {
    // Update request time from server unless it failed verification.
    guard self.verificationResult != .failed, let requestDate = self.requestDate else { return self }
```

NachoSoto:
> Can you add a test in case the server returns a 500 error? 

We already have a test for that and it's still passing 👍🏻

Sources/Purchasing/VerificationResult.swift
NachoSoto: > Oh, I see that here you will return the response as not valid if a 304 is not valid. I don't think we should do this. We should act as unable to fetch. There is a legit situation for this, imagine you are in an airport wifi that intercepts and mangles requests. It should not break your cache.

> I think I still have trouble with the logic for validation being in two places.

> I would rather do something like:

> For validation errors (either 200, 403...), if there is cache, treat this as a 500, just return cache

That's not changed. As per my other comment, that would be handled by `DeviceCache`. `HTTPClient` doesn't return a valid response if there is a 500.

> For 403 update request time in cache only if cache was checked valid and 304 was valid too (304 invalid would be covered in the above case anyway). Then return cache.
> For valid 200, update cache, return fresh data

That's how it works 👍🏻

NachoSoto:
I think I handled all comments, but let me know if I missed something.

Sources/Networking/HTTPClient/ETagManager.swift
bisho: >> 304 NOT_MODIFIED, INVALID signature, with request time t=2. SDK should get (valid signature, t=1)

> Oh that's a case I didn't consider.
> Are we sure we want to do that? That would break an app without any indication that this is happening.

The thing is that you can't use t=2 as it is not verified. And making the cache not verified in whole defeats the purpose of the cache. Note that there are legit cases of tampering (you are in an airport wifi that intercepts HTTPs. So you can't assume it is a bad actor and use t=2 but making it fail validation. We have a cache, use it until you can. The only case for using t=2 is if there app has disabled verification.

aboedo: I agree that we can't use t=2, otherwise that opens us up to an easy attack. 

But I'm not sure about the part about considering any result that we have "verified" if we get a signature that we can't verify - even if it's from a 304. Like, "_someone_ told us to use what's in cache but we don't really trust them" seems like a clear use case for "not verified". 

I get that we're not entirely sure if any valid tampering will happen in real life, but... this is all happening at the application level, right? Like, there really shouldn't be any modified content in the response content, datetime or request uuid that a public wifi modifies to have us fail validation. 

It feels like we might be setting ourselves up for one of those use cases that takes a long time to explain to someone, while just saying "if you got an invalid signature, you get an unverified entitlement" seems a lot simpler

NachoSoto:
This fixes #2470.

PR #2322, opened 2023-03-03
`HTTPClient`: don't verify 4xx/5xx responses

NachoSoto:


Sources/Misc/Signing+ResponseVerification.swift
```diff
@@ -65,6 +65,7 @@ extension HTTPResponse where Body == Data {
     ) -> VerificationResult {
         guard let nonce = request.nonce,
               let publicKey = publicKey,
+              !statusCode.isServerError,
```
tonidero, line 68: Should we also filter 4xx? Not sure if that would be caught anywhere else

NachoSoto: Yup! Let me add a test for that too 👍🏻

Sources/Misc/Signing+ResponseVerification.swift
```diff
@@ -65,6 +65,7 @@ extension HTTPResponse where Body == Data {
     ) -> VerificationResult {
         guard let nonce = request.nonce,
               let publicKey = publicKey,
+              statusCode.isSuccessfulResponse,
```
NachoSoto, line 68: Changed this.

Sources/Networking/HTTPClient/HTTPStatusCode.swift
```diff
@@ -78,7 +78,7 @@ extension HTTPStatusCode: Codable {}
 extension HTTPStatusCode {
 
     var isSuccessfulResponse: Bool {
-        return 200...299 ~= self.rawValue
+        return 200...399 ~= self.rawValue
```
NachoSoto, line 81: Changed this to match Android.
I confirmed nothing else uses this method, and the only other usage is unaffected.

NachoSoto (commented):
Finished this @tonidero

tonidero (approved):
Looks good!

PR #2323, opened 2023-03-03
replace `iff` with `if and only if`

aboedo:
Follow up to https://github.com/RevenueCat/purchases-ios/pull/2300, replace `iff` with the more explicit `if and only if`


aboedo:
cc @MortenGregersen

NachoSoto (approved):
This is much better thanks 👍🏻

PR #2324, opened 2023-03-06
Add additional logging on init

codykerns:
CSDK-258

Adds additional logging during init:
- System/OS version
- Environment (is/is not simulator)


NachoSoto (approved):
👏🏻

PR #2327, opened 2023-03-07
`CustomerInfo`: use same grace period logic for active subscriptions

NachoSoto:
Follow up to #2288.

### DIFF
--- a/Sources/Identity/CustomerInfo.swift
+++ b/Sources/Identity/CustomerInfo.swift
@@ -24,7 +24,7 @@ import Foundation
     @objc public let entitlements: EntitlementInfos
 
     /// All *subscription* product identifiers with expiration dates in the future.
-    @objc public var activeSubscriptions: Set<String> { self.activeKeys(dates: expirationDatesByProductId) }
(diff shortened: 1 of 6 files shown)


Sources/Identity/CustomerInfo.swift
```diff
@@ -360,15 +360,14 @@ private extension CustomerInfo {
...
-                    return self.isAfterReferenceDate(date: date)
-                }
+                .filter { self.isDateActive($1) }
                 .map { key, _ in key }
         )
     }
 
-    func isAfterReferenceDate(date: Date) -> Bool { date.timeIntervalSince(self.requestDate) > 0 }
+    private func isDateActive(_ date: Date?) -> Bool {
+        return EntitlementInfo.isDateActive(expirationDate: date, for: self.requestDate)
```
tonidero, line 369: This is a bit of a nitpick, but we are using a method in `EntitlementInfo` for subscriptions, which is not related to entitlements at all.

tonidero (approved):
Just one nitpick, but I don't think it's a blocker.

Sources/Identity/CustomerInfo.swift
NachoSoto: Yeah that's fair, let me move this somewhere else.

PR #2330, opened 2023-03-07
`IdentityManager`: clear `ETagManager` and `DeviceCache` if verification is enabled but cached `CustomerInfo` is not

NachoSoto:
Android counterpart: https://github.com/RevenueCat/purchases-android/pull/844

### DIFF
--- a/Sources/Networking/Backend.swift
+++ b/Sources/Networking/Backend.swift
@@ -143,6 +143,21 @@ extension Backend {
 // - Class is not `final` (it's mocked). This implicitly makes subclasses `Sendable` even if they're not thread-safe.
 extension Backend: @unchecked Sendable {}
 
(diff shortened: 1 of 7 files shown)


Sources/Networking/Backend.swift
```diff
@@ -143,6 +143,21 @@ extension Backend {
...
+
+extension Backend {
+
+    // For testing
+    var networkTimeout: TimeInterval {
+        return self.config.httpClient.timeout
+    }
+
+    @objc var signatureVerificationEnabled: Bool {
+        return self.config.httpClient.signatureVerificationEnabled
```
tonidero, line 156: I see what you mean, we can do the same in Android 👍

tonidero (approved):
LGTM!

PR #2331, opened 2023-03-07
`ETagManager`: don't ignore failed etags with `Signing.VerificationMode.informational`

NachoSoto:
Essentially we don't want to modify any behavior on `.informational` mode. Everything will continue behaving normally, and simply report verification failures.

### DIFF
--- a/Sources/Networking/HTTPClient/ETagManager.swift
+++ b/Sources/Networking/HTTPClient/ETagManager.swift
@@ -20,33 +20,43 @@ class ETagManager {
     static let eTagResponseHeaderName = HTTPClient.ResponseHeader.eTag.rawValue
 
     private let userDefaults: SynchronizedUserDefaults
+    private let verificationMode: Signing.ResponseVerificationMode
 
-    convenience init() {
+    convenience init(verificationMode: Signing.ResponseVerificationMode) {
         self.init(
             userDefaults: UserDefaults(suiteName: Self.suiteName)
             // This should never return `nil` for this known `suiteName`,
             // but `.standard` is a good fallback anyway.
-            ?? UserDefaults.standard
+            ?? UserDefaults.standard,
+            verificationMode: verificationMode
         )
     }
 
-    init(userDefaults: UserDefaults) {
+    init(userDefaults: UserDefaults, verificationMode: Signing.ResponseVerificationMode) {
         self.userDefaults = .init(userDefaults: userDefaults)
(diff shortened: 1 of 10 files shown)


Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -20,33 +20,43 @@ class ETagManager {
...
-        var storedETag = ""
-        if !refreshETag, let storedETagAndResponse = self.storedETagAndResponse(for: urlRequest) {
-            if !signatureVerificationEnabled || storedETagAndResponse.verificationResult == .verified {
-                storedETag = storedETagAndResponse.eTag
-            }
+        func eTag() -> String? {
+            if refreshETag { return nil }
+            guard let storedETagAndResponse = self.storedETagAndResponse(for: urlRequest) else { return nil }
+
+            let shouldUseETag = (
```
tonidero, line 50: Hmm I wonder, should we always use the Etag if it's available? 

It's true that once we support `enforced`, we would need to have either something like this or just clear the etag cache when switching from informational to enforced as well... I think that second options might be simpler? But I think we can do this for now too.

tonidero (commented):
Just one question... I think that the current way might be more future-proof, but I wonder if it's necessary.

Sources/Networking/HTTPClient/ETagManager.swift
NachoSoto: The answer depends on what we settle on from that spreadsheet. Looks like the new proposal might backtrack this so I'll wait.

NachoSoto: Since we're only focusing on `.informational` for now, what do you think about leaving this as is? Basically it'll always use the ETag on `.informational` as discussed.

tonidero: Yeah, I think this is ok as is for now. Maybe add a comment to revisit once enforced is added?

tonidero (approved):

Sources/Networking/HTTPClient/ETagManager.swift
NachoSoto: I think it's fine for now, the `.enforced` behavior is still not fully decided anyway.

tonidero: Sounds good 👍

PR #2333, opened 2023-03-13
Finished signature verification `HTTPClient` tests

NachoSoto:
See https://docs.google.com/spreadsheets/d/1kNcFNB8U47N5_vm8Ww57izMgdp60mdzUxZFK-cm46KU/edit#gid=0
Screenshot 2023-03-13 at 15 59 42

### Changes:
- Extracted signature verification `HTTPClient` tests into a separate file
- Added helper stubbing methods to simplify tests
- Created `InformationalSignatureVerificationHTTPClientTests` and `EnforcedSignatureVerificationHTTPClientTests`
- Changed `VerificationResult.from(cache:response:)` to match new expected tests
- Finished coverage for `ResponseVerificationMode.informational` tests

### DIFF
--- a/Sources/Security/VerificationResult.swift
+++ b/Sources/Security/VerificationResult.swift
@@ -83,7 +83,7 @@ extension VerificationResult {
         case (.verified, .notRequested): return .notRequested
         case (.verified, .failed): return .failed
 
-        case (.notRequested, .verified): return .notRequested
+        case (.notRequested, .verified): return .verified
         case (.notRequested, .failed): return .failed
 
         case (.failed, .notRequested): return .notRequested
--- a/Sources/Security/VerificationResult.swift
+++ b/Sources/Security/VerificationResult.swift
@@ -83,14 +83,11 @@ extension VerificationResult {
         case (.verified, .notRequested): return .notRequested
         case (.verified, .failed): return .failed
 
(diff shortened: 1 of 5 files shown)


thecadams:
Took a look, looks good to me, but I'll defer to folks with more Swift knowledge to approve.

Sources/Security/VerificationResult.swift
```diff
@@ -83,14 +83,11 @@ extension VerificationResult {
         case (.verified, .notRequested): return .notRequested
         case (.verified, .failed): return .failed
 
-        // These shouldn't happen because `ETagManager` will ignore not verified cached responses
-        // if verification is enabled.
         case (.notRequested, .verified): return .notRequested
```
tonidero, line 86: Should we also update this case to `verified`? Though in that case, we should just pick the responseResult always instead of this switch.

tonidero (approved):
LGTM!

Sources/Security/VerificationResult.swift
NachoSoto: Yeah I'll update it, that makes sense.
I think we might want to keep the switch and add a third argument: the mode. Because I predict we'll implement this differently for `.enforced`.

PR #2336, opened 2023-03-14
`RELEASING.md`: added GitHub rate limiting parameter

NachoSoto:
This is necessary to avoid rate limiting errors.


tonidero (approved):

tonidero:
I guess we could set it up as the default in fastlane... But this looks ok

PR #2337, opened 2023-03-14
`VerificationResult.notRequested`: removed caching reference

NachoSoto:
This is no longer true since #2330.


tonidero (approved):
Ah good catch! Will also update this in Android

Issue #2338, opened 2023-03-15
Crash after purchase (watchOS)


vincentneo:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

1. Environment
   1. Platform: Apple Watch
   2. SDK version: 4.17.7
   3. StoreKit 2 (_disabled with `useStoreKit2IfEnabled(false)`_) (Y/N): did not use `useStoreKit2IfEnabled` function
   4. OS version: watchOS 9.3.1
   5. Xcode version: Xcode 14.2
   6. How widespread is the issue. Percentage of devices affected. Tested on my device
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue

Slightly not following the template because there's 2 parts to this issue, also because I'm an idiot.

#### Part 1 
I observed this crash before (below is replicated today), before release of app:
Not sure how you get a crash log out of Xcode but here's a screenshot that maybe helps. (Note that this is RevenueCat 4.17.7, unmodified)
<img width="1900" alt="Screenshot 2023-03-15 at 12 04 01 PM" src="https://user-images.githubusercontent.com/23420208/225204619-b6deae37-5cea-4858-8113-263cbf107071.png">

The last few revenue cat console log:
```
2023-03-15 12:01:38.620022+0800 Pigeon WatchKit Extension[385:11853] [Purchases] - INFO: 💰 Purchasing Product 'pigeon.fullaccess.1year' from package in Offering 'full.access'
2023-03-15 12:01:38.622930+0800 Pigeon WatchKit Extension[385:11793] [Purchases] - DEBUG: ℹ️ StoreKit1Wrapper (0x16eeae40) updatedTransaction: pigeon.fullaccess.1year 0
[...]

PR #717, opened 2021-08-06
Support all platforms in the test targets

Juanpe:
Resolves #652 

The goal of this PR is to enable all platforms in the test targets.

PR #2340, opened 2023-03-16
`CI`: also run tests on `watchOS`

NachoSoto:
This is a follow up to #717.
In order to be able to reproduce #2338, which seems to be an issue only on `watchOS`, it's important that we now run tests on `watchOS` as well.
We probably won't be able to fully replicate that issue since we don't run tests on actual devices, so they'll still run on 64 bits. But this way we might be able to prevent issues that are only reproducible on `watchOS` in the future.

## Changes:
- Added CircleCI job
- Added lane
- Simplified `UnitTestsHostApp` definition using `SwiftUI`
- Increased _test only_ `watchOS` deployment target to 7.0 (this was necessary to be able to have a common `@main` for the host app)
- Increased _test only_ `tvOS` deployment target to `14.0` (same as above, plus we only run tvOS tests on the latest version anyway. We test backwards compatibility through iOS already)
- Fixed test compilation on `watchOS`
- Disabled `HTTPClient` tests on `watchOS` (see https://github.com/AliSoftware/OHHTTPStubs/issues/287)
- `StoreKitConfigTestCase` is only available from `watchOS 7.0` since that's when `SKTestSession` was introduced
- Added `watchOS` support to `UnitTests` target
- Added `watchOS` support to `ReceiptParserTests` target
- Fixed `isFamilySharable` `watchOS` version checks for version 8.0
- Created new `CI-RevenueCat` Test Plan to only run unit tests and not StoreKit tests on `watchOS`


Tests/StoreKitUnitTests/LocalReceiptParserStoreKitTests.swift
```diff
@@ -17,7 +17,7 @@ import Nimble
 import StoreKit
 import XCTest
 
-@available(iOS 14.0, tvOS 14.0, macOS 11.0, watchOS 6.2, *)
+@available(iOS 14.0, tvOS 14.0, macOS 11.0, watchOS 7.0, *)
```
aboedo, line 20: oh, [`SKTestSession` is watchOS 7+](https://developer.apple.com/documentation/storekittest/sktestsession#), I'm surprised this even worked before

Tests/StoreKitUnitTests/StoreProductTests.swift
```diff
@@ -107,14 +107,17 @@ class StoreProductTests: StoreKitConfigTestCase {
...
         expect(storeProduct.localizedTitle) == "Monthly Free Trial"
         // open the StoreKit Config file as source code to see the expected value
         expect(storeProduct.subscriptionGroupIdentifier) == "7096FF06"
 
         expect(storeProduct.subscriptionPeriod?.unit) == .month
         expect(storeProduct.subscriptionPeriod?.value) == 1
 
+        if #available(iOS 14.0, macOS 11.0, tvOS 14.0, watchOS 8.0, *) {
+            expect(storeProduct.isFamilyShareable) == true
+        }
```
aboedo, line 119: we don't need to fix this one right now, but we may have messed up here - [`isFamilyShareable` is available on watchOS 7](https://developer.apple.com/documentation/storekit/skproduct/3564805-isfamilyshareable#), not sure why we limited to 8 in StoreProduct but it was probably an oversight

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -27,6 +27,11 @@ class BaseHTTPClientTests: TestCase {
     override func setUpWithError() throws {
         try super.setUpWithError()
 
+        #if os(watchOS)
+        // See https://github.com/AliSoftware/OHHTTPStubs/issues/287
+        try XCTSkipIf(true, "OHHTTPStubs does not currently support watchOS")
+        #endif
```
aboedo, line 33: would love to replace OHHTTPStubs altogether some day

aboedo (approved):

Tests/UnitTests/Networking/HTTPClientTests.swift
NachoSoto: Yup..

Tests/StoreKitUnitTests/StoreProductTests.swift
NachoSoto: Huh thanks! I thought I saw this was watchOS 8.0 on `StoreKit` itself. I'll fix it!

Tests/StoreKitUnitTests/LocalReceiptParserStoreKitTests.swift
NachoSoto: We never compiled this on `watchOS`.

PR #2350, opened 2023-03-17
`Trusted Entitlements`: changed all APIs to `internal`

NachoSoto:
This isn't ready quite ready to be released yet. With this change, `main` can be released once again.

I've added a comment to all the changes so we can easily find them when ready to be made `public` again.


aboedo (approved):

tonidero (commented):
A bit late to this but should we also hide the error code for verification errors?

PR #2351, opened 2023-03-17
`Offline Entitlements`: created `ProductEntitlementMappingResponse`

NachoSoto:


Sources/Networking/Responses/ProductsEntitlementsResponse.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+//      https://opensource.org/licenses/MIT
+//
+//  ProductsEntitlementsResponse.swift
+//
+//  Created by Nacho Soto on 3/17/23.
+
+import Foundation
+
+struct ProductsEntitlementsResponse {
```
NachoSoto, line 16: I don't love this name, but changing it is easy, just wanted to get this blueprint structure out.

Tests/UnitTests/Networking/Responses/Fixtures/ProductsEntitlements.json
```diff
@@ -0,0 +1,23 @@
+{
+    "products": [
+        {
+            "id": "com.revenuecat.foo_1",
```
aboedo, line 4: maybe we should also change the endpoint response to be "identifier"? wdyt? I imagine with gzip compression enabled it wouldn't really make a difference in terms of data transfer

aboedo (approved):

tonidero (approved):

Tests/UnitTests/Networking/Responses/Fixtures/ProductsEntitlements.json
```diff
@@ -0,0 +1,23 @@
+{
+    "products": [
```
tonidero, line 2: Should this be `products_entitlements_mapping` or something like that?

Tests/UnitTests/Networking/Responses/Fixtures/ProductsEntitlements.json
NachoSoto: I agree. Is this endpoint ready? This is just a placeholder until we have the real format from the backend.

Tests/UnitTests/Networking/Responses/Fixtures/ProductsEntitlements.json
NachoSoto: I based this off of the suggested format from the doc.

NachoSoto:
Merging this since it matches the agreed response format for now, but we can change it.

PR #2352, opened 2023-03-17
Offline Entitlements: get SK2 purchase information

aboedo:
Just started work on this. Early draft getting purchase information from StoreKit 2 so we can create entitlements from it.

### DIFF
--- a/Sources/Logging/Strings/OfflineEntitlementsStrings.swift
+++ b/Sources/Logging/Strings/OfflineEntitlementsStrings.swift
@@ -1,0 +1,34 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  OfflineEntitlementsStrings.swift
+//
+//  Created by Andrés Boedo on 3/21/23.
+
+import Foundation
+import StoreKit
+
+// swiftlint:disable identifier_name
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+enum OfflineEntitlementsStrings {
+
+    case found_unverified_transactions_in_sk2(transaction: StoreKit.Transaction, error: Error)
+
+}
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+extension OfflineEntitlementsStrings: CustomStringConvertible {
+
+    var description: String {
+        switch self {
+        case .found_unverified_transactions_in_sk2(let transaction, let error):
+            return "Found an unverified transaction. It will be ignored and will not be a part of CustomerInfo. \nDetails: \(transaction.debugDescription)"
+        }
+    }
+}
--- a/Sources/Logging/Strings/OfflineEntitlementsStrings.swift
+++ b/Sources/Logging/Strings/OfflineEntitlementsStrings.swift
@@ -18,7 +18,7 @@ import StoreKit
 @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
 enum OfflineEntitlementsStrings {
 
-    case found_unverified_transactions_in_sk2(transaction: StoreKit.Transaction, error: Error)
+    case found_unverified_transactions_in_sk2(StoreKit.Transaction, Error)
 
 }
 
@@ -27,8 +27,14 @@ extension OfflineEntitlementsStrings: CustomStringConvertible {
 
     var description: String {
         switch self {
-        case .found_unverified_transactions_in_sk2(let transaction, let error):
-            return "Found an unverified transaction. It will be ignored and will not be a part of CustomerInfo. \nDetails: \(transaction.debugDescription)"
+        case let .found_unverified_transactions_in_sk2(transaction, error):
+            return """
+                Found an unverified transaction. It will be ignored and will not be a part of CustomerInfo.
+                    Details:
+                    Error: \(error.localizedDescription)
+                    Transaction: \(transaction.debugDescription)
+            """
         }
     }
+    
 }
--- a/Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
+++ b/Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
@@ -34,6 +34,10 @@ struct PurchasedSK2Product {
     let ownershipType: PurchaseOwnershipType
     let verification: VerificationResult = .verified
 
+}
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+extension PurchasedSK2Product {
     init(from transaction: StoreKit.Transaction,
          sandboxEnvironmentDetector: SandboxEnvironmentDetector = BundleSandboxEnvironmentDetector.default
     ) {
@@ -56,7 +60,7 @@ struct PurchasedSK2Product {
             self.periodType = .normal
         }
 
-        self.isActive = expirationDate == nil || expirationDate! > Date() // todo: check what we usually do for non-subs
+        self.isActive = expirationDate.map { $0 > Date() } ?? false // todo: check what we usually do for non-subs
         self.willRenew = true // best guess, StoreKit.Transaction does not provide this info.
         self.latestPurchaseDate = transaction.purchaseDate
         self.originalPurchaseDate = transaction.originalPurchaseDate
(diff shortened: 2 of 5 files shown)


Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
```diff
@@ -0,0 +1,75 @@
...
+        } else {
+            self.periodType = .normal
+        }
+    }
+}
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+struct PurchasedProductsManager {
+
+    func fetchPurchasedProducts() async throws -> [PurchasedSK2Product] {
```
NachoSoto, line 53: This would probably be `static`, right?

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
```diff
@@ -0,0 +1,75 @@
...
+import StoreKit
+
+
+/// This struct should have all the information we need from StoreKit  to create EntitlementInfo from a StoreKit 2 transaction.
+/// Other fields from other places might be needed.
+/// If not, then consider just having a constructor that creates EntitlementInfo from a Storekit.Transaction instead and using
+/// a Factory pattern to store the logic in one place
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+struct PurchasedSK2Product {
+    let productIdentifier: String
```
NachoSoto, line 24: Nit: line break inside types (it took me a while to get used to this but now it's hard not to hehe)

aboedo: I like it 💪🏻 thanks! FWIW, expect to see some of the ugliest code ever while this is in draft

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
aboedo: I went with instance out of sheer habit. Based on the env detector there, though, it looks like current style is protocol + static + initializer with default param? Happy to go with that instead

NachoSoto: Oh wait, this is a "manager" ignore my comment. I was thinking this was a product type.

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
```diff
@@ -0,0 +1,122 @@
...
+                                  originalPurchaseDate: purchasedSK2Product.originalPurchaseDate,
+                                  expirationDate: purchasedSK2Product.expirationDate,
+                                  store: purchasedSK2Product.store,
+                                  productIdentifier: purchasedSK2Product.productIdentifier,
+                                  isSandbox: purchasedSK2Product.isSandbox,
+                                  unsubscribeDetectedAt: purchasedSK2Product.unsubscribeDetectedAt,
+                                  billingIssueDetectedAt: purchasedSK2Product.billingIssueDetectedAt,
+                                  ownershipType: purchasedSK2Product.ownershipType,
+                                  verification: purchasedSK2Product.verification),
+                  rawData: [:]) // todo: should we just have the data be the transaction's raw data?
```
NachoSoto, line 120: Yeah ideally this matches the same format as our normal `EntitlementInfo`s, but that's going to be hard to guarantee. Given that we don't document it, I think it would be safe to have something else, like the transaction's raw data as you said 👍🏻

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
```diff
@@ -0,0 +1,122 @@
...
+    let originalPurchaseDate: Date?
+    let expirationDate: Date?
+    let store = Store.appStore
+    let isSandbox: Bool
+    let unsubscribeDetectedAt: Date?
+    let billingIssueDetectedAt: Date?
+    let ownershipType: PurchaseOwnershipType
+    let verification: VerificationResult = .verified
+
+    init(from transaction: StoreKit.Transaction,
```
NachoSoto, line 37: Small code clarity suggestion that I like doing: separate the `struct` definition and fields from this `init`, putting it in an extension. So it's easier to see at a glance what the `struct` has when using it, versus implementation detail on how it's constructed.

Sources/Logging/Strings/OfflineEntitlementsStrings.swift
```diff
@@ -0,0 +1,34 @@
...
+//  Created by Andrés Boedo on 3/21/23.
+
+import Foundation
+import StoreKit
+
+// swiftlint:disable identifier_name
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+enum OfflineEntitlementsStrings {
+
+    case found_unverified_transactions_in_sk2(transaction: StoreKit.Transaction, error: Error)
```
NachoSoto, line 21: Nit: I think these parameter names don't add anything useful that's not already encoded in the types:
```suggestion
    case found_unverified_transactions_in_sk2(StoreKit.Transaction, Error)
```

Sources/Logging/Strings/OfflineEntitlementsStrings.swift
```diff
@@ -0,0 +1,34 @@
...
+
+}
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+extension OfflineEntitlementsStrings: CustomStringConvertible {
+
+    var description: String {
+        switch self {
+        case .found_unverified_transactions_in_sk2(let transaction, let error):
```
NachoSoto, line 30: We don't have anything in the style guide, but if you agree with this maybe we should:
```suggestion
        case let .found_unverified_transactions_in_sk2(transaction, error):
```

One `let` instead of multiple?

NachoSoto:
This is a great start 💪🏻  about to branch off to work on exposing the `CustomerInfo` too.

Sources/Logging/Strings/OfflineEntitlementsStrings.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    var description: String {
+        switch self {
+        case let .found_unverified_transactions_in_sk2(transaction, error):
+            return """
+                Found an unverified transaction. It will be ignored and will not be a part of CustomerInfo.
+                    Details:
+                    Error: \(error.localizedDescription)
+                    Transaction: \(transaction.debugDescription)
+            """
```
aboedo, line 36: @NachoSoto I also updated this to multi-line string, I figured it's much easier to read now. WDTY?

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
aboedo: I like it! done

Sources/Logging/Strings/OfflineEntitlementsStrings.swift
NachoSoto: Love it

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
aboedo: so... to be clear, do you like this style better than the other one? I'm happy to go with either.

NachoSoto: Yeah this makes sense 👍🏻 
Both are the same thing TBH, but this makes mocking a bit easier.

NachoSoto: Actually this is a `struct`, so it can't be mocked. So might as well make this `static`. But this is fine for now, let's see how we need to use this and/or mock it and then we can change it.

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
```diff
@@ -0,0 +1,126 @@
...
+    let isSandbox: Bool
+    let unsubscribeDetectedAt: Date?
+    let billingIssueDetectedAt: Date?
+    let ownershipType: PurchaseOwnershipType
+    let verification: VerificationResult = .verified
+
+}
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+extension PurchasedSK2Product {
```
NachoSoto, line 40: I think we need a different representation here, so we can use the same code that constructs `EntitlementInfos` from the response data. This is useful for creating an individual `EntitlementInfo`, but we'd have to change a bunch more code. Instead, we can use the types in `CustomerInfoResponse` so that at the top level we can construct a `CustomerInfo` the same way.

Might not make a lot of sense, but I'll have a PR ready to show what I mean.

NachoSoto: My new suggested type:
```swift
struct PurchasedSK2Product {

    let subscription: CustomerInfoResponse.Subscription
    let transaction: CustomerInfoResponse.Transaction
    let entitlement: CustomerInfoResponse.Entitlement

}
```

NachoSoto: #2358

NachoSoto:
Continued on #2358

PR #2353, opened 2023-03-17
`Offline Entitlements`: added support for fetching `ProductEntitlementMappingResponse` in `OfflineEntitlementsAPI`

NachoSoto:
### TODO:
- [x] Update snapshots


aboedo (approved):

Sources/Networking/HTTPClient/HTTPRequest.swift
```diff
@@ -140,7 +142,7 @@ extension HTTPRequest.Path {
                 .postSubscriberAttributes,
                 .postAdServicesToken:
             return true
-        case .health:
```
tonidero, line 143: I guess this is because if this is going to live in a different server, we might not support etags in that place for now?

Sources/Networking/Operations/GetProductsEntitlementsOperation.swift
```diff
@@ -0,0 +1,66 @@
...
+    override func begin(completion: @escaping () -> Void) {
+        self.getResponse(completion: completion)
+    }
+
+}
+
+private extension GetProductsEntitlementsOperation {
+
+    func getResponse(completion: @escaping () -> Void) {
+        let request = HTTPRequest(method: .get, path: .getProductsEntitlements)
```
tonidero, line 50: I'm not sure but we might have a different base URL for this one @aboedo? Or would we just use DNS to point to whatever server we want to? If we need a different base URL, we will need to add support for that (in Android, that was added for diagnostics). In any case, that should be done in a different PR.

tonidero (approved):

Sources/Networking/HTTPClient/HTTPRequest.swift
NachoSoto: Exactly.

Sources/Networking/Operations/GetProductsEntitlementsOperation.swift
NachoSoto: Looks like the goal is to have it as the same domain, but it's not built yet.

Sources/Networking/HTTPClient/HTTPRequest.swift
```diff
@@ -138,7 +140,8 @@ extension HTTPRequest.Path {
                 .postOfferForSigning,
                 .postReceiptData,
                 .postSubscriberAttributes,
-                .postAdServicesToken:
+                .postAdServicesToken,
+                .getProductEntitlementMapping:
```
NachoSoto, line 144: @tonidero I enabled ETags for this again after talking to the server people. Once that works there, the SDK will support it.

tonidero: Thanks for letting me know. Will probably revert my PR since there won't be any requests using that right now... We can add it back in the future.

PR #2359, opened 2023-03-21
`OfferingsManager`: don't clear offerings cache timestamp when request fails

NachoSoto:
There is no reason to clear this and make the cache immediately stale. Instead, failed requests would eventually make the cache stale after a certain duration.


tonidero (commented):
Looks like this was also done in Android. I believe the reasoning is that (in Android at least), we set the cache timestamp right before we make the backend call. So if it fails, we want to retry immediately, and not wait for the non-existent cached timestamp to remain. If it's the same in iOS, we should keep this.

NachoSoto:
Oh that's wrong then, we should only update the timestamp after receiving the response. But I guess I see why... that's to avoid making multiple concurrent requests while stale. But the new callback cache fixes that now?

vegaro:
You two are right. This was there before to prevent making another request (the timestamp would indicate the cache is not stale). But I think as @NachoSoto says, the new cache callback fixes that and makes this more clear, so I think it's safe to remove

NachoSoto:
Cool, thanks!
We need to make sure that we also don't reset the timestamp until after the request finishes.

But again, let's not do this until after the next release, no reason to introduce more risk. I'll leave this open in the meantime and we can revisit it later.

NachoSoto:
@RevenueCat/coresdk this is rebased and ready to review again.

Sources/Caching/DeviceCache.swift
```diff
@@ -175,10 +175,6 @@ class DeviceCache {
         )
     }
 
-    func clearOfferingsCacheTimestamp() {
-        self.offeringsCachedObject.clearCacheTimestamp()
-    }
-
     func setOfferingsCacheTimestampToNow() {
```
tonidero, line 178: Hmm I was thinking, did we also change where we call this to only call it on the success case? Otherwise, if there is an error and we are calling this BEFORE the request happens, this change could cause some unintended behavior.

NachoSoto: Yeah this is only called by `preventPurchasePopupCallFromTriggeringCacheRefresh`.

NachoSoto: Which doesn't make a lot of sense btw, I filed SDK-3172 to address that.

NachoSoto: I removed that in #2623. Any other concerns here?

tonidero (approved):
After changing it in the other PR this seems safe now 👍

PR #2360, opened 2023-03-21
`ETagManager`: added test for 304 responses with no etag

NachoSoto:
Technically the backend would be breaking the contract with this, but it's worth checking that the SDK does something reasonable in this scenario.

### DIFF
--- a/Tests/UnitTests/Networking/ETagManagerTests.swift
+++ b/Tests/UnitTests/Networking/ETagManagerTests.swift
@@ -86,6 +86,30 @@ class ETagManagerTests: TestCase {
         expect(response?.requestDate) == requestDate
     }
 
+    func testStoredResponseIsNotUsedIfResponseCodeIs304ButContainsNoETag() throws {
+        let eTag = "an_etag"
+        let url = self.urlForTest()
+        let request = URLRequest(url: url)
+        _ = self.mockStoredETagResponse(for: url, statusCode: .success, eTag: eTag)
+        let responseObject = try JSONSerialization.data(withJSONObject: ["a": "response"])
+        let requestDate = Date().addingTimeInterval(-100000)
+
+        let response = self.eTagManager.httpResultFromCacheOrBackend(
+            with: self.responseForTest(url: url,
+                                       body: responseObject,
+                                       eTag: nil,
+                                       statusCode: .notModified,
+                                       requestDate: requestDate),
+            request: request,
+            retried: false
+        )
+
+        expect(response).toNot(beNil())
+        expect(response?.statusCode) == .notModified
+        expect(response?.body) == responseObject
+        expect(response?.requestDate) == requestDate
+    }
+
     func testBackendResponseIsReturnedIf304AndCantFindCachedAndItHasAlreadyRetried() throws {
         let eTag = "an_etag"
         let url = urlForTest()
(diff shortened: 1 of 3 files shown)


Tests/UnitTests/Networking/ETagManagerTests.swift
```diff
@@ -86,6 +86,30 @@ class ETagManagerTests: TestCase {
         expect(response?.requestDate) == requestDate
     }
 
+    func testStoredResponseIsNotUsedIfResponseCodeIs304ButContainsNoETag() throws {
+        let eTag = "an_etag"
+        let url = self.urlForTest()
+        let request = URLRequest(url: url)
+        _ = self.mockStoredETagResponse(for: url, statusCode: .success, eTag: eTag)
+        let responseObject = try JSONSerialization.data(withJSONObject: ["a": "response"])
+        let requestDate = Date().addingTimeInterval(-100000)
```
aboedo, line 95: why's the old request date needed?

Tests/UnitTests/Networking/ETagManagerTests.swift
```diff
@@ -86,6 +86,30 @@ class ETagManagerTests: TestCase {
...
+                                       eTag: nil,
+                                       statusCode: .notModified,
+                                       requestDate: requestDate),
+            request: request,
+            retried: false
+        )
+
+        expect(response).toNot(beNil())
+        expect(response?.statusCode) == .notModified
+        expect(response?.body) == responseObject
```
aboedo, line 109: the response _is_ used in this case, though, right? the name of the test says it's not used, I'm not entirely sure what that means in this context

NachoSoto: This is the response from the backend, not the stored one (the `ETagManager.httpResultFromCacheOrBackend` still confuses me too).

Tests/UnitTests/Networking/ETagManagerTests.swift
NachoSoto: Just to test that the request date returned matches the one from the response.

Tests/UnitTests/Networking/ETagManagerTests.swift
tonidero: Hmm I wonder, `304` responses usually won't have a body right? I know this test is for a hypothetical situation, but I wonder if the correct behavior here would be to just error the request... If we try to fallback to a hypothetical body in a 304 response it would make it much harder to catch this unexpected situation.

NachoSoto: Oooh I see what you're saying. Yeah in that case we'd have to fail cause we can't load the etag.
Let me update the test.

NachoSoto: Okay so our system is handling this well 👍🏻  I'll leave the updated test to cover this scenario though.
Here's what happens:
- First response: 304, but no etag
- Couldn't find stored response
- `HTTPClient` sees this, and retries the request, with `retried` = `true`
- Because it's retried, `ETagManager` doesn't use the etag, so subsequent request can't return 304

tonidero (approved):
Yup that makes sense!

PR #2363, opened 2023-03-21
`OfferingsManager`: moved log to common method to remove hardcoded string

NachoSoto:
Instead of having the log in `OfferingsManager.offerings()`, this is now in `OfferingsManager.updateOfferingsCache`, which is called by the former as well as `Purchases.updateAllCachesIfNeeded`.
This means that we no longer need that hardcoded string.

### DIFF
--- a/Sources/Purchasing/OfferingsManager.swift
+++ b/Sources/Purchasing/OfferingsManager.swift
@@ -58,10 +58,6 @@ class OfferingsManager {
 
         systemInfo.isApplicationBackgrounded { isAppBackgrounded in
             if self.deviceCache.isOfferingsCacheStale(isAppBackgrounded: isAppBackgrounded) {
-                Logger.debug(isAppBackgrounded
-                             ? Strings.offering.offerings_stale_updating_in_background
-                             : Strings.offering.offerings_stale_updating_in_foreground)
-
                 self.updateOfferingsCache(appUserID: appUserID,
                                           isAppBackgrounded: isAppBackgrounded,
                                           fetchPolicy: fetchPolicy,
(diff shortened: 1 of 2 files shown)


Sources/Purchasing/OfferingsManager.swift
```diff
@@ -58,16 +58,11 @@ class OfferingsManager {
...
-                             : Strings.offering.offerings_stale_updating_in_foreground)
-
                 self.updateOfferingsCache(appUserID: appUserID,
                                           isAppBackgrounded: isAppBackgrounded,
-                                          fetchPolicy: fetchPolicy,
-                                          completion: nil)
-
-                Logger.rcSuccess(Strings.offering.offerings_stale_updated_from_network)
+                                          fetchPolicy: fetchPolicy) { _ in
+                    Logger.rcSuccess(Strings.offering.offerings_stale_updated_from_network)
```
NachoSoto, line 64: This also needed to be in the completion block.

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -78,6 +73,10 @@ class OfferingsManager {
         fetchPolicy: FetchPolicy = .default,
         completion: (@MainActor @Sendable (Result<Offerings, Error>) -> Void)?
     ) {
+        Logger.debug(isAppBackgrounded
+                     ? Strings.offering.offerings_stale_updating_in_background
+                     : Strings.offering.offerings_stale_updating_in_foreground)
```
tonidero, line 78: I guess `isAppBackgrounded` could be a parameter to build the string... But this is how it was before so we can keep as is for now.

tonidero (approved):

PR #2365, opened 2023-03-22
`Offline Entitlements`: created `ProductEntitlementMapping`

NachoSoto:
This will help decouple the format of `ProductEntitlementMappingResponse` with the internal handling and caching.


Sources/OfflineEntitlements/ProductEntitlementMapping.swift
```diff
@@ -0,0 +1,48 @@
...
+    var entitlementsByProduct: [String: Set<String>]
+
+}
+
+extension ProductEntitlementMapping {
+
+    /// - Returns: entitlement identifiers associated to the given product identifier
+    func entitlements(for productIdentifier: String) -> Set<String> {
+        return self.entitlementsByProduct[productIdentifier] ?? []
+    }
```
tonidero, line 28: Maybe add a test or two hitting this?

Tests/UnitTests/OfflineEntitlements/ProductEntitlementMappingTests.swift
```diff
@@ -0,0 +1,80 @@
...
+                "pro_1",
+                "pro_2"
+            ],
+            "product_3": [
+                "pro_4"
+            ]
+        ]
+    }
+
+    func testResponseToMappingWithDuplicateProductIdentifiers() {
```
tonidero, line 43: Hmm interesting... We don't have this test right now in Android. I would say it would be unexpected to receive this case, but it's good to handle it nonetheless 👍

tonidero (approved):

Sources/OfflineEntitlements/ProductEntitlementMapping.swift
NachoSoto: Oh good point.

NachoSoto: Done.

PR #2367, opened 2023-03-23
Implemented `NetworkError.isServerDown`

NachoSoto:
This will be used to handle offline entitlements.

PR #2368, opened 2023-03-23
`Offline Entitlements`: use offline-computed `CustomerInfo` when server is down

NachoSoto:
### Changes:
- Created `PurchasedProductsFetcherType` to be able to mock and hold fetchers, even on older devices
- Created `ProductEntitlementMappingFetcher` to abstract injecting `ProductEntitlementMapping`
- Added logic to `CustomerInfoResponseHandler` to compute offline `CustomerInfo` after a server error (see #2367). **This is used by all `CustomerInfo` requests as well as `PostReceiptDataOperation`**.
- Wrote comprehensive `CustomerInfoResponseHandler` tests

### TODO:
- [x] Don't cache offline computed `CustomerInfo`: #2378
- [x] Integration tests: #2501.

### DIFF
--- a/Sources/Networking/Backend.swift
+++ b/Sources/Networking/Backend.swift
@@ -190,11 +190,7 @@ extension Backend {
     }
 
     static func createDefaultProductFetcher() -> PurchasedProductsFetcherType {
-        if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) {
-            return PurchasedProductsFetcher()
-        } else {
-            return VoidPurchasedProductsFetcher()
-        }
+        return PurchasedProductsFetcher()
     }
 
 }
--- a/Sources/Networking/Backend.swift
+++ b/Sources/Networking/Backend.swift
@@ -23,13 +23,17 @@ class Backend {
 
     private let config: BackendConfiguration
 
-    convenience init(apiKey: String,
-                     systemInfo: SystemInfo,
-                     httpClientTimeout: TimeInterval = Configuration.networkTimeoutDefault,
-                     eTagManager: ETagManager,
-                     operationDispatcher: OperationDispatcher,
-                     attributionFetcher: AttributionFetcher,
-                     dateProvider: DateProvider = DateProvider()) {
+    convenience init(
+        apiKey: String,
+        systemInfo: SystemInfo,
+        httpClientTimeout: TimeInterval = Configuration.networkTimeoutDefault,
+        eTagManager: ETagManager,
+        operationDispatcher: OperationDispatcher,
+        attributionFetcher: AttributionFetcher,
+        productEntitlementMappingFetcher: ProductEntitlementMappingFetcher,
+        purchasedProductsFetcher: PurchasedProductsFetcherType = Backend.createDefaultProductFetcher(),
+        dateProvider: DateProvider = DateProvider()
+    ) {
         let httpClient = HTTPClient(apiKey: apiKey,
                                     systemInfo: systemInfo,
                                     eTagManager: eTagManager,
@@ -37,8 +41,10 @@ class Backend {
         let config = BackendConfiguration(httpClient: httpClient,
                                           operationDispatcher: operationDispatcher,
                                           operationQueue: QueueProvider.createBackendQueue(),
-                                          dateProvider: dateProvider,
-                                          systemInfo: systemInfo)
+                                          systemInfo: systemInfo,
+                                          productEntitlementMappingFetcher: productEntitlementMappingFetcher,
+                                          purchasedProductsFetcher: purchasedProductsFetcher,
+                                          dateProvider: dateProvider)
         self.init(backendConfig: config, attributionFetcher: attributionFetcher)
     }
 
@@ -181,6 +187,14 @@ extension Backend {
             return operationQueue
         }
 
+    }
+
+    static func createDefaultProductFetcher() -> PurchasedProductsFetcherType {
+        if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) {
+            return PurchasedProductsFetcher()
+        } else {
+            return VoidPurchasedProductsFetcher()
+        }
     }
 
 }
--- a/Sources/Networking/Backend.swift
+++ b/Sources/Networking/Backend.swift
@@ -30,8 +30,7 @@ class Backend {
         eTagManager: ETagManager,
         operationDispatcher: OperationDispatcher,
         attributionFetcher: AttributionFetcher,
-        productEntitlementMappingFetcher: ProductEntitlementMappingFetcher,
-        purchasedProductsFetcher: PurchasedProductsFetcherType = Backend.createDefaultProductFetcher(),
+        offlineCustomerInfoCreator: OfflineCustomerInfoCreator?,
         dateProvider: DateProvider = DateProvider()
     ) {
         let httpClient = HTTPClient(apiKey: apiKey,
@@ -42,8 +41,7 @@ class Backend {
                                           operationDispatcher: operationDispatcher,
                                           operationQueue: QueueProvider.createBackendQueue(),
                                           systemInfo: systemInfo,
-                                          productEntitlementMappingFetcher: productEntitlementMappingFetcher,
-                                          purchasedProductsFetcher: purchasedProductsFetcher,
+                                          offlineCustomerInfoCreator: offlineCustomerInfoCreator,
                                           dateProvider: dateProvider)
         self.init(backendConfig: config, attributionFetcher: attributionFetcher)
     }
@@ -189,10 +187,6 @@ extension Backend {
             return operationQueue
         }
 
-    }
-
-    static func createDefaultProductFetcher() -> PurchasedProductsFetcherType {
-        return PurchasedProductsFetcher()
     }
 
 }
--- a/Sources/Networking/Backend.swift
+++ b/Sources/Networking/Backend.swift
@@ -102,9 +102,11 @@ class Backend {
 
     func getCustomerInfo(appUserID: String,
                          withRandomDelay randomDelay: Bool,
+                         allowComputingOffline: Bool = true,
                          completion: @escaping CustomerAPI.CustomerInfoResponseHandler) {
         self.customer.getCustomerInfo(appUserID: appUserID,
                                       withRandomDelay: randomDelay,
+                                      allowComputingOffline: allowComputingOffline,
                                       completion: completion)
     }
 
--- a/Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
+++ b/Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
@@ -15,10 +15,43 @@ import Foundation
 
 class CustomerInfoResponseHandler {
 
-    init() { }
+    typealias CustomerInfoCreator = ([PurchasedSK2Product],
+                                     ProductEntitlementMapping,
+                                     String) -> CustomerInfo
+
+    private let purchasedProductsFetcher: PurchasedProductsFetcherType
+    private let productEntitlementMapping: ProductEntitlementMapping?
+    private let customerInfoCreator: CustomerInfoCreator
+    private let userID: String
+
+    convenience init(
+        purchasedProductsFetcher: PurchasedProductsFetcherType,
+        productEntitlementMapping: ProductEntitlementMapping?,
+        userID: String
+    ) {
+        self.init(
+            purchasedProductsFetcher: purchasedProductsFetcher,
+            productEntitlementMapping: productEntitlementMapping,
+            customerInfoCreator: { products, mapping, userID in
+                CustomerInfo(from: products, mapping: mapping, userID: userID)
+            },
+            userID: userID)
+    }
+
+    init(
+        purchasedProductsFetcher: PurchasedProductsFetcherType,
+        productEntitlementMapping: ProductEntitlementMapping?,
+        customerInfoCreator: @escaping CustomerInfoCreator,
+        userID: String
+    ) {
+        self.purchasedProductsFetcher = purchasedProductsFetcher
+        self.productEntitlementMapping = productEntitlementMapping
+        self.customerInfoCreator = customerInfoCreator
+        self.userID = userID
+    }
 
     func handle(customerInfoResponse response: HTTPResponse<Response>.Result,
-                completion: CustomerAPI.CustomerInfoResponseHandler) {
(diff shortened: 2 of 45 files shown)


Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
```diff
@@ -43,3 +48,16 @@ class PurchasedProductsFetcher {
...
 }
+
+// MARK: -
+
+/// An empty implementation of `PurchasedProductsFetcherType` for platforms
+/// that don't support `PurchasedProductsFetcher`.
+final class VoidPurchasedProductsFetcher: PurchasedProductsFetcherType {
+
+    func fetchPurchasedProducts() async throws -> [PurchasedSK2Product] {
+        assertionFailure("This should never be used")
```
NachoSoto, line 59: This will only crash in `DEBUG` builds. We only create this prior to iOS15, where offline `CustomerInfo`s aren't computed.

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -59,3 +112,28 @@ extension CustomerInfoResponseHandler {
...
+            Logger.warn(Strings.offlineEntitlements.computing_offline_customer_info_with_no_entitlement_mapping)
+        }
+
+        let offlineCustomerInfo = self.customerInfoCreator(
+            products,
+            self.productEntitlementMapping ?? .empty,
+            self.userID
+        )
+
+        // TODO: merge with existing one?
```
NachoSoto, line 134: This is filed as a separate task.

NachoSoto:
The integration test failures seems unrelated to this

Sources/OfflineEntitlements/PurchasedSK2Product.swift
```diff
@@ -24,6 +23,8 @@ struct PurchasedSK2Product {
```
vegaro, line 23: I think this should support multiple entitlements correct? It's possible a product has N entitlements

aboedo: yep, that's a bug

aboedo: I'll update

NachoSoto: That's done later through the entitlement mapping.

NachoSoto:
Any pending comments here?

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -32,7 +65,27 @@ class CustomerInfoResponseHandler {
...
+        }
+
+        _ = Task<Void, Never> {
+            do {
+                completion(.success(try await self.computeOfflineCustomerInfo()))
+            } catch {
+                Logger.error(Strings.offlineEntitlements.computing_offline_customer_info_failed(error))
+                completion(result)
+            }
+        }
```
aboedo, line 88: The next time the app loads after having had entitlements computed offline, we'll be returning from cache first, right? Given that this only is used after a network request is made? 

If that's the case, then it's a bit of a problem - you make a purchase, get entitlements, next time you open the app the entitlements are gone until the notification comes in the delegate, and even then you'll only get them if the app listens to the delegate. 

We need to make sure that if offline entitlements were ever returned, we're still returning them until we get a response from the backend.

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -32,7 +65,27 @@ class CustomerInfoResponseHandler {
...
+    ) {
+        guard result.error?.isServerDown == true,
+        #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) else {
+            completion(result)
+            return
+        }
+
+        _ = Task<Void, Never> {
+            do {
+                completion(.success(try await self.computeOfflineCustomerInfo()))
```
aboedo, line 83: where are we disabling the feature? Meaning, if we merge to main, won't this actually start computing offline entitlements if we get a 5xx? I'm missing the feature flag

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -59,3 +112,52 @@ extension CustomerInfoResponseHandler {
     }
 
 }
+
+private extension CustomerInfoResponseHandler {
+
```
aboedo, line 117: having this whole feature at the response handling level honestly feels a bit off to me - none of the other handlers do anything this sophisticated, and this one already takes care of subscriber attributes errors, it feels like it might be handling a bit too much stuff. 
I'd also love to see all logic relevant to Offline Entitlements live in the OfflineEntitlements group, so that if a maintainer wants to see how that feature works, they can find everything they need in a single place

Sources/OfflineEntitlements/CustomerInfo+OfflineEntitlements.swift
```diff
@@ -48,7 +47,17 @@ extension CustomerInfo {
         )
     }
 
-    private static func createEntitlements(
+}
+
+extension CustomerInfo {
+
+    static let defaultManagementURL = URL(string: "https://apps.apple.com/account/subscriptions")!
```
aboedo, line 54: we already have this in SystemInfo, we can reuse it from there

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
```diff
@@ -14,9 +14,16 @@
 import Foundation
 import StoreKit
 
+protocol PurchasedProductsFetcherType {
+
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    func fetchPurchasedProducts() async throws -> [PurchasedSK2Product]
```
aboedo, line 20: Why is the void implementation needed? We're already checking API availability before calling `fetchPurchasedProducts`

Sources/Networking/Backend.swift
```diff
@@ -182,4 +188,12 @@ extension Backend {
 
     }
 
+    static func createDefaultProductFetcher() -> PurchasedProductsFetcherType {
+        if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) {
+            return PurchasedProductsFetcher()
+        } else {
+            return VoidPurchasedProductsFetcher()
```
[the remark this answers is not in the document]
NachoSoto: This is needed because we need to inject `PurchasedProductsFetcherType`, but `PurchasedProductsFetcher` isn't available. It won't be used if it's not available anyway. Can you think of a better way to do this?

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
NachoSoto: I moved 99% of this code out of here and into `OfflineEntitlements/CustomerInfo+OfflineEntitlements.swift`

As for whether this will be called from here or the logic will be used elsewhere, I'm looking at all the edge cases and how to handle them (and how Android did it) to decide that.

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
NachoSoto: Done, this is disabled by default now.

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
NachoSoto: Yeah I'm an idiot, I solved this in a better way so we don't need that void type :)

NachoSoto:
This should be ready now 🎉 
And integration tests added in #2501.

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -15,10 +15,46 @@ import Foundation
...
+    private let customerInfoCreator: CustomerInfo.OfflineCreator
+    private let userID: String
+    // Allows temporarily disabling the feature until it's fully ready
+    private let offlineEntitlementsEnabled: Bool
+
+    convenience init(
+        purchasedProductsFetcher: PurchasedProductsFetcherType,
+        productEntitlementMapping: ProductEntitlementMapping?,
+        userID: String,
+        offlineEntitlementsEnabled: Bool = false
```
NachoSoto, line 29: I'm gonna take this flag out now that this is fully functional.

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -15,10 +15,46 @@ import Foundation
 
 class CustomerInfoResponseHandler {
 
-    init() { }
+    private let purchasedProductsFetcher: PurchasedProductsFetcherType
+    private let productEntitlementMapping: ProductEntitlementMapping?
+    private let customerInfoCreator: CustomerInfo.OfflineCreator
```
tonidero, line 20: I wonder if we could abstract all these 3 objects to a single `OfflineCustomerInfoCreator` or `OfflineCustomerInfoCalculator` as in Android... That way we abstract all the process needed to create the offline customer info from these handlers and operations

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -32,7 +68,38 @@ class CustomerInfoResponseHandler {
...
+        guard self.offlineEntitlementsEnabled,
+              result.error?.isServerDown == true,
+              #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) else {
+            completion(result)
+            return
+        }
+
+        _ = Task<Void, Never> {
+            do {
+                completion(.success(try await self.computeOfflineCustomerInfo()))
```
tonidero, line 87: This is something that was only recently added to android but we also need to clear the cached customer info upon returning the offline entitlements customer info: https://github.com/RevenueCat/purchases-android/blob/main/common/src/main/java/com/revenuecat/purchases/common/offlineentitlements/OfflineEntitlementsManager.kt#L74. I believe we are not doing that here?

This is to avoid some possible edge cases where if a user enters offline entitlements mode, and restarts the app, we don't want them to use the cached customer info which might not contain the latest purchase information. Instead, we prefer to wait to try to fetch the customer info from the backend (and return an offline computed customer info from that if backend still down)

tonidero, line 87: A separate thing but somewhat related, I think this will call the customer info listener with the offline customer info correct? But since we are not caching that value, how are you handling users calling `getCustomerInfo` after that? 

In android, we currently would just return the offline customer info in cache (in memory) first, then initiate another fetch from the backend. 

I could see an argument for being stateless and just make it so we don't have cached customer info after we "enter" offline entitlements mode, so if trying to get the value from cache, it would fail and we should just initiate a backend request always... This might be slower in some conditions though so not sure if it's a good idea... Personally, I think caching the offline customer info would mimick real behavior more closely, but I can understand that stateless could also help avoiding some issues...

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -32,7 +68,38 @@ class CustomerInfoResponseHandler {
...
-        completion(result)
+        self.handle(result: result, completion: completion)
+    }
+
+    private func handle(
+        result: Result<CustomerInfo, BackendError>,
+        completion: @escaping CustomerAPI.CustomerInfoResponseHandler
+    ) {
+        guard self.offlineEntitlementsEnabled,
+              result.error?.isServerDown == true,
```
tonidero, line 79: I believe we are missing another condition here for when to calculate offline entitlements... When performing a "get customer info" request, we should only calculate offline entitlements if there is no cached customer info in disk. Otherwise, we should just rely on the cached customer info. Are you handling that somewhere else?

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
tonidero: This is already done! I missed that https://github.com/RevenueCat/purchases-ios/blob/3edef65e8557e6ea648f0a8917f3f15f44054b03/Sources/Identity/CustomerInfoManager.swift#L197

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
NachoSoto: Done, cleaned up this PR a lot.

tonidero (approved):
Great job on this! I think we are handling all cases pretty well now 👍

tonidero (approved):
Great job on this! I think we are handling all cases pretty well now 👍

Sources/OfflineEntitlements/OfflineCustomerInfoCreator.swift
```diff
@@ -0,0 +1,112 @@
...
+        self.productEntitlementMappingFetcher = productEntitlementMappingFetcher
+        self.creator = creator
+    }
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func create(for userID: String) async throws -> CustomerInfo {
+        Logger.info(Strings.offlineEntitlements.computing_offline_customer_info)
+
+        guard let mapping = self.productEntitlementMappingFetcher.productEntitlementMapping,
+              !mapping.entitlementsByProduct.isEmpty else {
```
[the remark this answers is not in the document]
tonidero: Ah followup PR: https://github.com/RevenueCat/purchases-ios/pull/2504. Nvm!

Tests/UnitTests/Mocks/MockOfflineEntitlementsManager.swift
```diff
@@ -7,7 +7,7 @@
 //
 //      https://opensource.org/licenses/MIT
 //
-//  MockOfflineEntitlementsManager.swift
+//  MockOfflineEntitlementsManager.swiftz
```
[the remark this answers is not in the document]
NachoSoto: Oops thanks

NachoSoto:
I did some initial sanity check and this is working well, so merging!
Will still do more manual testing before releasing.

Tests/UnitTests/Mocks/MockOfflineEntitlementsManager.swift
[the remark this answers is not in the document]
NachoSoto: Fixing in the next PR to avoid running CI just for this 😅

PR #2362, opened 2023-03-21
`Offline Entitlements`: created `LoadShedderIntegrationTests`

NachoSoto:
### TODO:

- [x] Set environment variable in CircleCI.
~~- [x] [Detect redirects](https://developer.apple.com/documentation/foundation/urlsessiontaskdelegate/1411626-urlsession) and log. Verify this is happening using `TestLogger`.~~

PR #2371, opened 2023-03-23
`HTTPClient`: debug log when performing redirects

NachoSoto:
This will be used to detect the load balancer redirecting to the load shedder. See also #2362.

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -42,7 +42,9 @@ class HTTPClient {
         config.httpMaximumConnectionsPerHost = 1
         config.timeoutIntervalForRequest = requestTimeout
         config.timeoutIntervalForResource = requestTimeout
-        self.session = URLSession(configuration: config)
+        self.session = URLSession(configuration: config,
+                                  delegate: RedirectLoggerSessionDelegate(),
+                                  delegateQueue: nil)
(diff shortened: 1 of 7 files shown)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -336,13 +337,17 @@ private extension HTTPClient {
...
             self.handle(urlResponse: urlResponse,
                         request: request,
                         urlRequest: urlRequest,
                         data: data,
                         error: error)
         }
+
+        if #available(iOS 15.0, *) {
+            task.delegate = RedirectLoggerTaskDelegate()
+        }
```
aboedo, line 350: we should always log - the load shedder happens automatically and isn't dependent on OS version. The only thing that is dependent on OS version is Offline Entitlements

aboedo (changes requested):
ready to go after we remove the os check!

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto: `URLSessionTask.delegate` is iOS 15 only, but I'll do this using `URLSession.delegate` instead.

NachoSoto: Oh wait, but that delegate doesn't have anything for following redirects...

aboedo (approved):

PR #2373, opened 2023-03-23
`ETagManager`/`HTTPClient`: sending new `X-RC-Last-Refresh-Time` header

NachoSoto:
Fixes SDK-3024 and SDK-3026.

### DIFF
--- a/Sources/Networking/HTTPClient/ETagManager.swift
+++ b/Sources/Networking/HTTPClient/ETagManager.swift
@@ -81,8 +81,11 @@ class ETagManager {
         }
 
         if self.shouldUseCachedVersion(responseCode: statusCode) {
-            if let storedResponse = self.storedHTTPResponse(for: request, withRequestDate: response.requestDate) {
-                return storedResponse
+            if let storedResponse = self.storedETagAndResponse(for: request) {
+                let newResponse = storedResponse.withUpdatedDate()
+
+                self.store(newResponse, for: request)
+                return newResponse.asResponse(withRequestDate: response.requestDate)
             }
             if retried {
                 Logger.warn(
@@ -119,7 +122,7 @@ private extension ETagManager {
 
     func storedETagAndResponse(for request: URLRequest) -> Response? {
         return self.userDefaults.read {
-            if let cacheKey = eTagDefaultCacheKey(for: request),
+            if let cacheKey = self.eTagDefaultCacheKey(for: request),
                let value = $0.object(forKey: cacheKey),
                let data = value as? Data {
                 return try? JSONDecoder.default.decode(Response.self, jsonData: data)
@@ -137,18 +140,24 @@ private extension ETagManager {
                                              response: HTTPResponse<Data?>,
                                              eTag: String) {
         if let data = response.body,
-           response.shouldStore(ignoreVerificationErrors: self.shouldIgnoreVerificationErrors),
-           let cacheKey = self.eTagDefaultCacheKey(for: request) {
-            let eTagAndResponse = Response(
-                eTag: eTag,
-                statusCode: response.statusCode,
-                data: data,
-                verificationResult: response.verificationResult
+           response.shouldStore(ignoreVerificationErrors: self.shouldIgnoreVerificationErrors) {
+            self.store(
+                Response(
+                    eTag: eTag,
+                    statusCode: response.statusCode,
+                    data: data,
+                    verificationResult: response.verificationResult
+                ),
+                for: request
             )
-            if let dataToStore = eTagAndResponse.asData() {
-                self.userDefaults.write {
-                    $0.set(dataToStore, forKey: cacheKey)
-                }
+        }
+    }
+
+    func store(_ response: Response, for request: URLRequest) {
+        if let cacheKey = self.eTagDefaultCacheKey(for: request),
+           let dataToStore = response.asData() {
+            self.userDefaults.write {
+                $0.set(dataToStore, forKey: cacheKey)
             }
         }
     }
@@ -223,6 +232,13 @@ extension ETagManager.Response {
             requestDate: requestDate,
             verificationResult: self.verificationResult
         )
+    }
+
+    fileprivate func withUpdatedDate() -> Self {
+        var copy = self
+        copy.lastUsed = Date()
+
+        return copy
     }
 
 }
--- a/Sources/Networking/HTTPClient/ETagManager.swift
+++ b/Sources/Networking/HTTPClient/ETagManager.swift
@@ -17,7 +17,7 @@ import Foundation
 class ETagManager {
 
     static let eTagRequestHeaderName = HTTPClient.RequestHeader.eTag.rawValue
-    static let eTagCreationTimeRequestHeaderName = HTTPClient.RequestHeader.eTagCreationTime.rawValue
+    static let eTagValidationTimeRequestHeaderName = HTTPClient.RequestHeader.eTagValidationTime.rawValue
     static let eTagResponseHeaderName = HTTPClient.ResponseHeader.eTag.rawValue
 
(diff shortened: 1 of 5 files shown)


Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -70,8 +81,11 @@ class ETagManager {
         }
 
         if self.shouldUseCachedVersion(responseCode: statusCode) {
-            if let storedResponse = self.storedHTTPResponse(for: request, withRequestDate: response.requestDate) {
-                return storedResponse
+            if let storedResponse = self.storedETagAndResponse(for: request) {
+                let newResponse = storedResponse.withUpdatedDate()
+
+                self.store(newResponse, for: request)
+                return newResponse.asResponse(withRequestDate: response.requestDate)
```
vegaro, line 88: I am a bit confused about this change. If I understand ti correctly, it is creating a new response with a date set to now, then saving that, then returning it with the request date of the request?

vegaro: Oh I see.. It saves the last used, then returns the response with the updated request date

Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -173,18 +193,22 @@ extension ETagManager {
         var eTag: String
         var statusCode: HTTPStatusCode
         var data: Data
+        @DefaultDecodable.Now
+        var lastUsed: Date
```
vegaro, line 197: What's this used for?

vegaro (commented):
Besides that little suggestion, it looks good

Sources/Networking/HTTPClient/ETagManager.swift
NachoSoto: You can see the full context here: https://linear.app/revenuecat/issue/SDK-3023/improve-load-shedder-performance-send-etag-validation-date

I know the code might not be the most clear, so suggestions for how to improve it are welcome.

Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -210,6 +234,13 @@ extension ETagManager.Response {
         )
     }
 
+    fileprivate func withUpdatedDate() -> Self {
```
vegaro, line 237: I would rename this to `withUpdatedLastDate`

vegaro (approved):

Sources/Networking/HTTPClient/ETagManager.swift
aboedo: I think this is about as clear as it gets, but I'll add a docstring to the actual property so its purpose is clear to future maintainers.

Sources/Networking/HTTPClient/ETagManager.swift
aboedo: agreed. updating

Sources/Networking/HTTPClient/ETagManager.swift
aboedo: marking as resolved, @vegaro lmk if there are still any concerns with this part

Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -126,18 +140,24 @@ private extension ETagManager {
...
-            if let dataToStore = eTagAndResponse.asData() {
-                self.userDefaults.write {
-                    $0.set(dataToStore, forKey: cacheKey)
-                }
+        }
+    }
+
+    func store(_ response: Response, for request: URLRequest) {
+        if let cacheKey = self.eTagDefaultCacheKey(for: request),
+           let dataToStore = response.asData() {
```
aboedo, line 158: this function might no-op, I'm going to rename it so it's clear from the callsite that it's not guaranteed to actually store

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -501,11 +501,14 @@ final class HTTPClientTests: BaseHTTPClientTests {
         let request = HTTPRequest(method: .get, path: .mockPath)
         let responseData = "{\"message\": \"something is great up in the cloud\"}".asData
         let eTag = "etag"
+        let eTagCreation = Date(timeIntervalSince1970: 1234567)
 
-        self.eTagManager.stubResponseEtag(eTag)
+        self.eTagManager.stubResponseEtag(eTag, creationTime: eTagCreation)
```
aboedo, line 506: this isn't etagCreation time anymore it's validation / update time, I'll rename accordingly

Sources/Networking/HTTPClient/ETagManager.swift
aboedo: maybe it'd be clearer if it's also called "validation" rather than "used", for consistency with the rest of the code?

Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -43,7 +44,7 @@ class ETagManager {
         withSignatureVerification: Bool,
         refreshETag: Bool = false
     ) -> [String: String] {
-        func eTag() -> String? {
+        func eTag() -> (tag: String, date: String)? {
```
aboedo, line 47: wait, how does this have a non-null date? how do we treat existing etags which don't have a date?

NachoSoto: The field is annotated with `@DefaultDecodable.Now`. I did that to maintain backwards compatibility, so stored etags don't fail to load when upgrading the SDK.

Tests/UnitTests/Networking/HTTPClientTests.swift
NachoSoto: I can rename this 👍🏻

Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -173,18 +193,23 @@ extension ETagManager {
         var eTag: String
         var statusCode: HTTPStatusCode
         var data: Data
+        /// Used by the backend for advanced load shedding techniques.
+        @DefaultDecodable.Now
+        var lastUsed: Date
```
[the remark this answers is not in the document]
NachoSoto: Done.

NachoSoto:
This should be ready now.

aboedo:
we received feedback from COIN saying that it'd be significantly preferable to reuse the existing eTag header, and append the epoch in seconds after a splitter (i.e.: `ETAG_VALUE:EPOCH_IN_SECONDS`). This means extra logic, sadly, but it saves some precious bytes on every request

aboedo:
could you update so that we don't send a new header and instead reuse the existing one with that format?

aboedo:
yet another update: we actually do need to use a separate header, because the Etag is used for signing, so mixing this in would complicate matters significantly. So... we're good to go here!

aboedo:
COIN asked that we use `X-RC-Last-Refresh-Time` (to avoid potential conflicts with other native ETag systems). We can update in this PR or merge this one instead and do the rename as a follow-up

NachoSoto:
Changed.

NachoSoto:
Let me know if you're good to merge this.

Sources/Networking/HTTPClient/ETagManager.swift
aboedo: so no date = now? I'm wondering if that's maybe worse than treating them as just old or something. Note that once the backend starts load shedding, it'll prefer to load shed requests that have recent cache, so no date will be more likely to be shed than something like "arbitrary old date"

NachoSoto: I totally missed this comment (saw it now catching up through my e-mail).

This is addressed in #2490.

PR #2378, opened 2023-03-24
`CustomerInfoManager`: don't cache offline `CustomerInfo`

NachoSoto:
This is meant to be a temporary `CustomerInfo` while the backend is offline.
See also #2368.

### DIFF
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -181,13 +181,18 @@ class CustomerInfoManager {
     }
 
     func cache(customerInfo: CustomerInfo, appUserID: String) {
-        do {
-            let jsonData = try JSONEncoder.default.encode(customerInfo)
-            self.withData { $0.deviceCache.cache(customerInfo: jsonData, appUserID: appUserID) }
-            self.sendUpdateIfChanged(customerInfo: customerInfo)
-        } catch {
-            Logger.error(Strings.customerInfo.error_encoding_customerinfo(error))
+        if customerInfo.shouldCache {
+            do {
+                let jsonData = try JSONEncoder.default.encode(customerInfo)
+                self.withData { $0.deviceCache.cache(customerInfo: jsonData, appUserID: appUserID) }
+            } catch {
+                Logger.error(Strings.customerInfo.error_encoding_customerinfo(error))
+            }
+        } else {
+            Logger.debug(Strings.customerInfo.not_caching_offline_customer_info)
         }
+
+        self.sendUpdateIfChanged(customerInfo: customerInfo)
     }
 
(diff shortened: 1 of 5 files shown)


Sources/Identity/CustomerInfoManager.swift
```diff
@@ -310,4 +315,29 @@ private extension CustomerInfoManager {
...
+    }
+
+}
+
+private extension VerificationResult {
+
+    var shouldCache: Bool {
+        switch self {
+        case .failed: return true
+        case .verified, .notRequested: return true
```
vegaro, line 338: Thoughts on this?

```suggestion
        case .failed, .verified, .notRequested: return true
```

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -181,13 +181,18 @@ class CustomerInfoManager {
...
+                let jsonData = try JSONEncoder.default.encode(customerInfo)
+                self.withData { $0.deviceCache.cache(customerInfo: jsonData, appUserID: appUserID) }
+            } catch {
+                Logger.error(Strings.customerInfo.error_encoding_customerinfo(error))
+            }
+        } else {
+            Logger.debug(Strings.customerInfo.not_caching_offline_customer_info)
         }
+
+        self.sendUpdateIfChanged(customerInfo: customerInfo)
```
vegaro, line 195: This will also send the updated customer info in case it goes in the catch.  We were not doing that before, and I am not sure we want to?

vegaro (approved):

Sources/Identity/CustomerInfoManager.swift
NachoSoto: Oh yeah, I meant to write a unit test for that.
I think it's still useful to notify the user, even if it fails to encode. I know that's changing behavior, but it made this easier so we didn't need 3 branches.

Sources/Identity/CustomerInfoManager.swift
NachoSoto: Oh yeah thanks.

Sources/Identity/CustomerInfoManager.swift
NachoSoto: Oh there's no easy way to test it cause we can't really represent a `CustomerInfo` that would fail to encode :/

vegaro: I guess it's fine, the reason it won't be able to be encoded is not that the customer info is broken, so I think it's safe to send it anyways

NachoSoto: Agreed.

NachoSoto:
Rebased this off of `main` so we can merge it before #2368.

PR #2379, opened 2023-03-24
Added new `VerificationResult.verifiedOnDevice`

NachoSoto:
This will be used to represent data that has been created and verified locally.

### DIFF
--- a/Sources/Security/VerificationResult.swift
+++ b/Sources/Security/VerificationResult.swift
@@ -54,6 +54,9 @@ internal enum VerificationResult: Int {
     /// Entitlements were verified with our server.
     case verified = 1
 
+    /// Entitlements were created and verified on device through `StoreKit 2`.
+    case verifiedOnDevice = 3
+
     /// Entitlement verification failed, possibly due to a MiTM attack.
     /// ### Related Symbols
     /// - ``ErrorCode/signatureVerificationFailed``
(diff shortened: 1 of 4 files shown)


vegaro (approved):

NachoSoto:
@aboedo thoughts?

Sources/Security/VerificationResult.swift
```diff
@@ -54,6 +54,9 @@ internal enum VerificationResult: Int {
     /// Entitlements were verified with our server.
     case verified = 1
 
+    /// Entitlements were created and verified locally through `StoreKit 2`.
+    case verifiedLocally = 3
```
aboedo, line 58: to me this reads like "I verified the results locally but I'd normally verify it on the server", which isn't entirely true... Also, I'd love it if we could reuse the result for when the load shedder is activated. Maybe something like `entitlementsComputedOffline`?

vegaro: What about `entitlementsComputedOnDevice`? I think offline implies there's no internet connection, which is not entirely true.

tonidero: Yup, makes sense!

vegaro: Maybe we are mixing two concepts, verification and computation? Maybe this should stay as `VERIFIED` and have another field for `origin` or something like that

vegaro: By the way, PR is up for Androd https://github.com/RevenueCat/purchases-android/pull/939 but will wait on merging until the name is clear

tonidero: > Maybe we are mixing two concepts, verification and computation?

That's probably the case... I guess when computed locally we are not really performing any verification, we are just trusting what Google/Apple give us.

tonidero: I'm wondering what use cases devs would have to know whether the entitlements come from our servers or were computed on device... Is it necessary to inform them? I could see as more useful to have the ability to turn off offline entitlements if devs want it.

aboedo: the verification still happens, it just happens through SK2 / GP. 
As for why inform devs, we should because some of the data might be inaccurate or a best guess, since we don't have all of the data available. 
Nacho makes a good point about this not being about entitlements, so I think we can go with either `verifiedLocally` or `verifiedOnDevice`. We don't need a final call for this until we ship, so we can just go with whichever you prefer for now and just make a final decision at the last min before shipping.

NachoSoto: What Andy said.

Another point, I do think this enum is the right place to encode this instead of another field. For offline computed data, none of the current `case`s really apply.

vegaro: let's go with `verifiedOnDevice`?

NachoSoto: Changed!

PR #2381, opened 2023-03-24
`SubscriberAttributesManagerIntegrationTests`: fixed flaky failures

NachoSoto:
Turns out we had already solved this race condition in #1492 that lead to flaky failures,
but it was only in `BaseStoreKitIntegrationTests`. This moves it up to `BaseBackendIntegrationTests` so that
`SubscriberAttributesManagerIntegrationTests` can use it too.

Fixes https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/10501/workflows/cb6dbf22-cda8-4c41-875b-04b489992631/jobs/61135


NachoSoto:
Integration tests passed with no retries 🎉

aboedo (approved):

PR #2383, opened 2023-03-27
Fix migration guide link

aboedo:
Fixes #2382

PR #2384, opened 2023-03-27
redirect to latest version of migration guide

aboedo:
From conversation in https://github.com/RevenueCat/purchases-ios/pull/2383, this creates a file with a redirect that will be used so that we can redirect customers to the latest version of the migration guide.


README.md
```diff
@@ -25,7 +25,7 @@ Sign up to [get started for free](https://app.revenuecat.com/signup).
 *RevenueCat* is the client for the [RevenueCat](https://www.revenuecat.com/) subscription and purchase tracking system. It's 100% `Swift` and compatible with `Objective-C`.
 
 ## Migrating from Purchases v3
-- See our [Migration guide](https://revenuecat-docs.netlify.app/documentation/revenuecat/v4_api_migration_guide)
+- See our [Migration guide](https://revenuecat.github.io/purchases-ios-docs/migration.html)
```
tonidero, line 28: Hmm if we have a migration doc in the future, there might be a conflict with this path... I guess ideally this would be something like `v4_api_migration_guide.html` as it was originally? But we can address that later

tonidero (approved):
Small comment but looking good

README.md
aboedo: yeah, agreed, updated

PR #2385, opened 2023-03-28
Purchases.configure: log warning if attempting to use a static appUserID

MarkVillacampa:
### Checklist
- [X] If applicable, unit tests
- [X] If applicable, create follow-up issues for `purchases-android` and hybrids

### Motivation

Hardcoding the user id is not desirable in almost any scenario, so we want to warn the developer when that happens.

### Description

Following what we did for `Purchases.logIn` in https://github.com/RevenueCat/purchases-ios/pull/1958, detect when `Builder.with(appUserId:)` is called with a static string and log a warning.

### DIFF
--- a/Sources/Purchasing/Configuration.swift
+++ b/Sources/Purchasing/Configuration.swift
@@ -113,8 +113,16 @@ import Foundation
          *
          * - Important: Set this property if you have your own user identifiers that you manage.
          */
+        @_disfavoredOverload
         @objc public func with(appUserID: String?) -> Builder {
             self.appUserID = appUserID
+            return self
+        }
+
+        // swiftlint:disable:next missing_docs
+        public func with(appUserID: StaticString) -> Configuration.Builder {
+            Logger.warn(Strings.identity.logging_in_with_static_string)
+            self.appUserID = "\(appUserID)"
             return self
         }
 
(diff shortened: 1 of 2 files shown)


Sources/Purchasing/Configuration.swift
```diff
@@ -113,11 +113,19 @@ import Foundation
...
+        @_disfavoredOverload
         @objc public func with(appUserID: String?) -> Builder {
             self.appUserID = appUserID
             return self
         }
 
+        // swiftlint:disable:next missing_docs
+        public func with(appUserID: StaticString) -> Configuration.Builder {
+            Logger.warn(Strings.identity.logging_in_with_static_string)
+            self.appUserID = "\(appUserID)"
```
NachoSoto, line 125: I'd delegate this to the other method:

```suggestion
            return self.with(appUserID: "\(appUserID)")
```

Tests/UnitTests/Purchasing/Purchases/PurchasesConfiguringTests.swift
```diff
@@ -158,6 +158,29 @@ class PurchasesConfiguringTests: BasePurchasesTests {
...
+    func testUserIdSringDoesNotLogMessage() {
+        let appUserID = "user ID"
+        let logger = TestLogHandler()
+
+        _ = Purchases.configure(
+            with: .init(withAPIKey: "")
+                .with(appUserID: appUserID)
+        )
+
+        logger.verifyMessageWasNotLogged(Strings.identity.logging_in_with_static_string, level: .warn)
```
NachoSoto, line 181: I recommend removing the level here to make sure it's not logged at _any_ level:

```suggestion
        logger.verifyMessageWasNotLogged(Strings.identity.logging_in_with_static_string)
```

NachoSoto (approved):
Nice 👍🏻

PR #2387, opened 2023-03-29
iOS v3 load shedder integration tests

aboedo:
Added a new test suite for integration tests for the load shedder using iOS v3. 

The way this is set up is a new project under Tests. 
This new project as a Swift Package Manager dependency on the latest iOS v3. 

It has some code that's duplicated from the main integration tests, but honestly it's very minimal, so I don't think it should be an issue. 

The new project includes a StoreKit Configuration file, and a couple of simple unit tests: one for testing getOfferings, one for purchasing the first available package. Dead simple. 

In a follow-up PR I'll actually set this up to run automatically on a schedule. But as it stands right now, this can be used. Just replace the API key with the one for the load shedder app and it works. 

I tested on my machine and it runs successfully. 

The actual setup:
| Project | StoreKit Config file | Running it |
| :-: | :-: | :-: |
| image | image | image |


aboedo:
note: these tests pass correctly locally. However, they're not in CI yet, so the test results you see are from the base PR. 
Also, the base PR will start passing correctly tomorrow, once the offerings cache catches up to a couple of changes I made today.

aboedo:
also, cc @tonidero, maybe this serves as inspiration for multi-version integration testing.

Tests/v3LoadShedderIntegration/v3LoadShedderIntegrationTests/v3LoadShedderIntegrationTests.swift
```diff
@@ -0,0 +1,92 @@
...
+//
+//  Created by Andrés Boedo on 3/29/23.
+
+@testable import Purchases
+import StoreKit
+import StoreKitTest
+@testable import v3LoadShedderIntegration
+import XCTest
+
+final class v3LoadShedderIntegrationTests: XCTestCase {
```
tonidero, line 20: Looks like tests don't like this name?

Tests/v3LoadShedderIntegration/v3LoadShedderIntegrationTests/v3LoadShedderIntegrationTests.swift
```diff
@@ -0,0 +1,92 @@
...
+    func testGetOfferings() async throws {
+        let offerings = try await Purchases.shared.offerings()
+        let offering = try XCTUnwrap(offerings.current)
+        XCTAssert(offering.availablePackages.count > 0)
+        let package = try XCTUnwrap(offering.availablePackages[0])
+        XCTAssert(package.product.productIdentifier == "com.revenuecat.loadShedder.monthly")
+        XCTAssert(offering.identifier == "default")
+    }
+
+    func testPurchasePackage() async throws {
```
tonidero, line 77: Should we have a test for `GetCustomerInfo` after a successful purchase? I guess the load shedder should also return the entitlement info right?

tonidero (commented):
Just a couple comments. As soon as we fix the tests in CI, I think this will be good to go though.

Tests/v3LoadShedderIntegration/v3LoadShedderIntegrationTests/v3LoadShedderIntegrationTests.swift
aboedo: the load shedder always returns 5xx for getCustomerInfo, it doesn't have enough context on any given subscriber to provide accurate data. So having a test for it won't really add anything. 
The idea is that if customerInfo is already in cache, the SDK will return the cached version anyway, so the system should work for all but fresh installs. 
And for the case of fresh installs, well, we're building Offline Entitlements

Tests/v3LoadShedderIntegration/v3LoadShedderIntegrationTests/v3LoadShedderIntegrationTests.swift
aboedo: forgot to push the line with the fix, it just doesn't like class names starting with lowercase

Tests/v3LoadShedderIntegration/v3LoadShedderIntegrationTests/v3LoadShedderIntegrationTests.swift
tonidero: Ah right, makes sense! 👍

Tests/v3LoadShedderIntegration/v3LoadShedderIntegrationTests/v3LoadShedderIntegrationTests.swift
```diff
@@ -0,0 +1,92 @@
...
+
+        Purchases.configure(withAPIKey: self.apiKey,
+                            appUserID: nil,
+                            observerMode: false,
+                            userDefaults: userDefaults)
+        clearReceiptIfExists()
+    }
+
+    func configureTestSession() throws {
+        assert(self.testSession == nil, "Attempted to configure session multiple times")
```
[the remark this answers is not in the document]
aboedo: 🤦 I was running these one at a time in my local machine, so I never ran into it. Great catch!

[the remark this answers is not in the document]
aboedo: surprisingly, it does pass when I run them all at once ¯\\_(ツ)_/¯ but I'll add the teardown to reset this anyway, doesn't hurt

aboedo:
I'm genuinely not sure why tests are failing (other than the linter), since this is a separate project being added and it doesn't modify existing tests. I'll rebase from main to see if anything changes.

aboedo:
oh I see, tests are failing because I accidentally added changes to files from `swiftlint --autocorrect`. Gotta say, quite the disconnect when swiftlint autocorrections cause tests to break.

aboedo:
@tonidero ready for another pass

Tests/v3LoadShedderIntegration/v3LoadShedderIntegrationTests/V3LoadShedderIntegrationTests.swift
```diff
@@ -0,0 +1,94 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  v3LoadShedderIntegrationTests.swift
+//
+//  Created by Andrés Boedo on 3/29/23.
+
+@testable import Purchases
+import StoreKit
+import StoreKitTest
+@testable import v3LoadShedderIntegration
```
tonidero, line 17: This also doesn't seem to follow convention of starting with uppercase? It might be ok though...

tonidero (approved):
LGTM!

PR #2388, opened 2023-03-30
Add lane for running iOS v3 load shedder integration tests

aboedo:
This adds a fastlane lane to execute load shedder integration tests. 
In a follow-up PR I'll call this lane automatically from CircleCI, on a schedule. 

It won't be a part of regular integration tests, since this logic is tied to v3 and static. So only the load shedder itself is getting tested here.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -252,7 +252,8 @@ platform :ios do
       previous_text: "REVENUECAT_LOAD_SHEDDER_API_KEY",
       new_text: ENV["REVENUECAT_LOAD_SHEDDER_API_KEY"],
       paths_of_files_to_update: [
-        './Tests/BackendIntegrationTests/Constants.swift'
+        './Tests/BackendIntegrationTests/Constants.swift',
+        './Tests/v3LoadShedderIntegration/v3LoadShedderIntegrationTests/V3LoadShedderIntegrationTests.swift'
       ]
     )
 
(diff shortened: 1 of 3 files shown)


fastlane/Fastfile
```diff
@@ -252,7 +252,8 @@ platform :ios do
       previous_text: "REVENUECAT_LOAD_SHEDDER_API_KEY",
       new_text: ENV["REVENUECAT_LOAD_SHEDDER_API_KEY"],
       paths_of_files_to_update: [
-        './Tests/BackendIntegrationTests/Constants.swift'
+        './Tests/BackendIntegrationTests/Constants.swift',
+        './Tests/v3LoadShedderIntegration/v3LoadShedderIntegrationTests/V3LoadShedderIntegrationTests.swift'
```
tonidero, line 256: Would it be worth separating it to a `Constants` file in the new project as well? Mostly to avoid replacing unintended things

fastlane/Fastfile
```diff
@@ -386,6 +387,22 @@ platform :ios do
...
+  lane :v3_loadshedder_integration_tests do |options|
+    replace_api_key_integration_tests
+    scan(
+      project: "./Tests/v3LoadShedderIntegration/v3LoadShedderIntegration.xcodeproj",
+      scheme: "v3LoadShedderIntegration",
+      derived_data_path: "scan_derived_data",
+      output_types: 'junit',
+      output_style: 'raw',
+      result_bundle: true,
+      configuration: 'Debug',
```
tonidero, line 400: I think running them on `Debug` is enough for now, since we are mostly just testing the interaction with the load shedder, which should be the same 👍

fastlane/Fastfile
```diff
@@ -386,6 +387,22 @@ platform :ios do
     )
   end
 
+  desc "Run LoadShedder tests"
+  lane :v3_loadshedder_integration_tests do |options|
```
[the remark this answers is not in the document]
tonidero: Should we also have a job for v4 to run them on a schedule? We can reuse the existing jobs though, so it's probably ok?

tonidero (approved):

fastlane/Fastfile
aboedo: maybe? It feels a bit YAGNI - the whole file has a single constant, and these are integration tests based on a version of the SDK that's in maintenance mode, so it's not like this test suite is really going to grow over time

fastlane/Fastfile
aboedo: +1. I'm not sure you even _could_ run them on release since we're using `@testable import`, but I haven't really checked. Won't make much of a difference in this case anyway.

fastlane/Fastfile
aboedo: also, the variable is called `REVENUECAT_LOAD_SHEDDER_API_KEY`, I don't feel like we're going to be running into a lot of conflicts with it. 
I'd vote to just keep it like this, as dead simple as possible, until the day comes where we actually start adding more to this and it becomes harder to manage.

PR #2392, opened 2023-04-08
`PurchaseTester`: fix memory leaks

kitwtnb:
### Checklist

### Motivation

Because memory leaks in `for await` loop.

### Description

Changed `self` to weak capture.


NachoSoto (approved):
Thanks!

PR #2399, opened 2023-04-10
`Xcode 14.3`: fixed compilation errors

NachoSoto:
See also #2397.
Xcode 14.3 introduced the following error:
> Heterogeneous collection literal could only be inferred to '[String : Any]'; add explicit type annotation if this is intentional

### Changes:
- Added explicit types to fix this new error
- Removed unnecessary `@preconcurrency` imports
- Added new `CustomerInfo.emptyInfo` to reduce duplication of same fake `CustomerInfo` creation

### DIFF
--- a/Tests/StoreKitUnitTests/BeginRefundRequestHelperTests.swift
+++ b/Tests/StoreKitUnitTests/BeginRefundRequestHelperTests.swift
@@ -233,9 +233,9 @@ private extension BeginRefundRequestHelperTests {
                 "original_app_user_id": "app_user_id",
                 "original_application_version": "2083",
                 "first_seen": "2019-06-17T16:05:33Z",
-                "non_subscriptions": [:],
+                "non_subscriptions": [:] as [String: Any],
                 "subscriptions": [
-                    "onemonth_freetrial": [:]
+                    "onemonth_freetrial": [:] as [String: Any]
                 ],
                 "entitlements": [
                     "\(mockEntitlementID)": [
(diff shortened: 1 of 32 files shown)


Tests/StoreKitUnitTests/BeginRefundRequestHelperTests.swift
```diff
@@ -233,9 +233,9 @@ private extension BeginRefundRequestHelperTests {
                 "original_app_user_id": "app_user_id",
                 "original_application_version": "2083",
                 "first_seen": "2019-06-17T16:05:33Z",
-                "non_subscriptions": [:],
+                "non_subscriptions": [:] as [String: Any],
```
tonidero, line 236: This is a slightly annoying check... But I guess it's safer 🤷‍♂️

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
```diff
@@ -29,10 +29,11 @@ class BaseCustomerInfoManagerTests: TestCase {
             "subscriber": [
                 "original_app_user_id": Self.appUserID,
                 "first_seen": "2019-06-17T16:05:33Z",
-                "subscriptions": [:],
-                "other_purchases": [:],
+                "subscriptions": [:]  as [String: Any],
+                "other_purchases": [:]  as [String: Any],
```
tonidero, line 33: Seems like we have a bunch like these in this PR?
```suggestion
                "subscriptions": [:] as [String: Any],
                "other_purchases": [:] as [String: Any],
```

tonidero (approved):
LGTM!

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
NachoSoto: Fixed, thanks!

PR #2405, opened 2023-04-13
SwiftUI Sample Project: Refactor Package terms method to a computed property

JOyo246:
Note: Not sure if PRs are wanted on Example code.

### Checklist

### Motivation
Noticed that the terms func was just passing the same `package`.

### Description
Testing: Looks like there are no tests for sample code.

This pull request refactors the terms(for package: Package) method in the Package extension to a computed property terms. The change not only improves readability and simplifies access to the package terms, but also eliminates the need to pass a Package instance to a function that belongs to the same Package type, which didn't make sense.
By making these changes, we provide a more streamlined and intuitive approach for developers to access package terms, and eliminate the need for an explicit method call.


aboedo (approved):
Thanks for doing this!! 🎉

PR #2407, opened 2023-04-14
Fix deadlock when sending notification center in customerInfoListener

aboedo:
Fixes an issue where a deadlock could happen if you fire a notification in the customerInfoListener. 

The issue stems from our usage of NotificationCenter to detect deletion of userDefaults. 

Since we've changed the default UserDefaults anyway, I don't feel like this adds much value anymore, and we've spent wayyyy too much time dealing with race conditions caused by it, so I'm just getting rid of it. 

I couldn't get it to reproduce on my device, but @codykerns was able to get a consistent repro and confirmed the fix with this branch.

### DIFF
--- a/Sources/Caching/DeviceCache.swift
+++ b/Sources/Caching/DeviceCache.swift
@@ -14,6 +14,7 @@
 
 import Foundation
 
+// swiftlint:disable file_length
 class DeviceCache {
 
     var cachedAppUserID: String? { return self._cachedAppUserID.value }
--- a/Sources/Caching/DeviceCache.swift
+++ b/Sources/Caching/DeviceCache.swift
@@ -14,7 +14,6 @@
 
 import Foundation
 
-// swiftlint:disable file_length type_body_length
 class DeviceCache {
 
     var cachedAppUserID: String? { return self._cachedAppUserID.value }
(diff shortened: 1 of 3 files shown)


aboedo:
Swiftlint and I seriously disagree on what constitutes a "serious" violation

image

image

Sources/Caching/DeviceCache.swift
```diff
@@ -33,52 +32,18 @@ class DeviceCache {
...
-
-    @objc private func handleUserDefaultsChanged(notification: Notification) {
-        guard let userDefaults = notification.object as? UserDefaults else {
-            return
-        }
-
-        // Note: this should never use `self.userDefaults` directly because this method
-        // might be synchronized, and `Atomic` is not reentrant.
-        if self.cachedAppUserID != nil && Self.cachedAppUserID(userDefaults) == nil {
-            fatalError(Strings.purchase.cached_app_user_id_deleted.description)
```
NachoSoto, line 72: I think this string can be deleted now.

NachoSoto (approved):
👏🏻

Sources/Caching/DeviceCache.swift
aboedo: good catch, deleted

PR #2416, opened 2023-04-19
`customEntitlementsComputation` integration PR

aboedo:
Adds a new dangerousSetting that disables some functionality in the SDK. The feature is specifically meant for apps that will do their own Entitlements computation, independently of RevenueCat.

it will:
- disable automatic cache refresh on getCustomerInfo, but if called directly, it does hit the backend
- disable anonymous users: if configure is called without appUserID, that’s an error. If logOut is called, that’s an error.
- disable the first call to the customerInfo listener (which is done automatically when you set it). This way you only get hit with a customerInfo listener if you post a receipt, or the first time you explicitly call getCustomerInfo.

it will not:
- stop you from using the rest of the APIs
- block subscriber attributes.


Sources/Misc/DangerousSettings.swift
```diff
@@ -45,14 +60,18 @@ import Foundation
...
+    @objc public convenience init(autoSyncPurchases: Bool = true, minimalImplementationOnly: Bool = false) {
+        self.init(autoSyncPurchases: autoSyncPurchases,
+                  minimalImplementationOnly: minimalImplementationOnly,
+                  internalSettings: .default)
+
     }
 
     /// Designated initializer
-    internal init(autoSyncPurchases: Bool, internalSettings: InternalSettings) {
+    internal init(autoSyncPurchases: Bool, minimalImplementationOnly: Bool, internalSettings: InternalSettings) {
```
aboedo, line 71: thought: we might want to split this into a few dangerous settings, making it much more modular. 
In particular, I'm thinking of: 
- one setting that disables logOut, and ideally also disables anonymous id creation. Basically it'd have to error out if you try to configure without appUserID. 
- one setting that disables getCustomerInfo, so that apps that do their own entitlements computation would be able to use that. 
- one or more settings to disable some other stuff, like attribution, subs attributes, trial eligibility

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -91,6 +91,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
             // Sends cached customer info (if exists) to delegate as latest
             // customer info may have already been observed and sent by the monitor
             self.sendCachedCustomerInfoToDelegateIfExists()
+            // TODO: do we need to send it the first time when the dangerousSetting is active?
```
aboedo, line 94: this is an interesting one. If you're doing your own entitlement computation, and would like to rely on the delegate as a proxy for knowing when you need to recompute, then firing right away after setting might not be useful. It might be more useful to only fire this one when an actual value comes in, which if everything else is disabled, would only happen when a post receipt goes through successfully

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -755,6 +784,15 @@ public extension Purchases {
     @objc(checkTrialOrIntroDiscountEligibility:completion:)
     func checkTrialOrIntroDiscountEligibility(productIdentifiers: [String],
                                               completion: @escaping ([String: IntroEligibility]) -> Void) {
+        guard !self.systemInfo.dangerousSettings.minimalImplementationOnly else {
+            var result: [String: IntroEligibility] = [:]
+            for product in productIdentifiers {
+                result[product] = IntroEligibility(eligibilityStatus: .unknown)
+            }
+            completion(result)
+            return
```
aboedo, line 793: I still need to add logging to these, but specifically _only_ to things that are called explicitly by the user, and not stuff that would otherwise have been called automatically.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -783,6 +821,8 @@ public extension Purchases {
 #endif
 
     @objc func invalidateCustomerInfoCache() {
+        guard !self.systemInfo.dangerousSettings.minimalImplementationOnly else { return }
```
aboedo, line 824: torn on this one, it might be useful

NachoSoto:
Another idea would be to make this a compile time flag like `RC_MINIMAL_IMPLEMENTATION` so it's configured at compile time.
### Pros:
- Reduced binary size
- Compile time guarantees of what's present vs what's missing
- Compile time guarantee that unavailable methods can't be called

### Cons:
- Increased complexity
- Requires separate builds for archiving and testing
- Higher complexity when integrating SDK, requiring clients to manually set the flag, which would be different based on how they integrate it

I think it's objectively better, but probably not worth the downsides.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1292,18 +1335,18 @@ private extension Purchases {
...
 
-            Logger.debug("Offerings cache is stale, updating caches")
-            self.offeringsManager.updateOfferingsCache(appUserID: self.appUserID,
-                                                       isAppBackgrounded: isAppBackgrounded,
-                                                       completion: nil)
+            if self.deviceCache.isOfferingsCacheStale(isAppBackgrounded: isAppBackgrounded) {
+                self.offeringsManager.updateOfferingsCache(appUserID: self.appUserID,
+                                                           isAppBackgrounded: isAppBackgrounded,
+                                                           completion: nil)
+            }
```
aboedo, line 1349: this just reorders the methods so that I could do an early return, but I could roll back the ordering and use an if instead

Tests/TestingApps/PurchaseTesterSwiftUI/Core/ConfiguredPurchases.swift
```diff
@@ -38,6 +38,7 @@ public final class ConfiguredPurchases {
             with: .builder(withAPIKey: apiKey)
                 .with(usesStoreKit2IfAvailable: useStoreKit2)
                 .with(observerMode: observerMode)
+                .with(dangerousSettings: DangerousSettings(minimalImplementationOnly: true))
```
aboedo, line 41: I'll roll this back before merging, it's just for testing

aboedo:
I'm using release/4.17.10 as a base for now because `main`'s version of the PurchaseTester is currently broken because of Trusted Entitlements. It'd be great to fix that and to ensure that `main` is working as well as possible.

Sources/Misc/DangerousSettings.swift
NachoSoto: Idea:
```swift
enum Feature {
  case attribution
  case authentication
  case entitlements
}

let disabledFeatures: Set<Feature> = []
```

NachoSoto:
> main's version of the PurchaseTester is currently broken because of Trusted Entitlements

Oops good catch. We don't build that in CI until release branches which is why we didn't catch it.

NachoSoto:
But if we want to allow granular configuration of enabled features, compile time flags is definitely not the way, so ignore :)

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: Yeah this type of stuff is scary. We might need to have a lot of _duplicate_ test coverage with this flag on so we can verify its behavior.

NachoSoto:
Fixed `PurchaseTester`: #2417

aboedo:
Update here: we're going to do something slightly different, where the new mode will be specifically geared towards apps who want to do entitlement computation independently of RevenueCat. 

it will:
- disable automatic cache refresh on getCustomerInfo, but if called directly, it does hit the backend
- disable anonymous users: if configure is called without appUserID, that’s an error. If logOut is called, that’s an error.
- disable the first call to the customerInfo listener (which is done automatically when you set it). This way you only get hit with a customerInfo listener if you post a receipt, or the first time you explicitly call getCustomerInfo.

it will not:
- stop you from using the rest of the APIs
- block subscriber attributes.

I'll work on that and open this PR for a first review pass when it's actually following the rules stated above

aboedo:
Updated to follow the rules outlined above

Tests/TestingApps/PurchaseTesterSwiftUI/Core/ConfiguredPurchases.swift
```diff
@@ -38,6 +38,7 @@ public final class ConfiguredPurchases {
             with: .builder(withAPIKey: apiKey)
                 .with(usesStoreKit2IfAvailable: useStoreKit2)
                 .with(observerMode: observerMode)
+                .with(dangerousSettings: DangerousSettings(customEntitlementComputation: true))
```
NachoSoto, line 41: This could be a UI setting :)

aboedo: I was thinking that too, I'll make a PR for it after I clean this up

aboedo: turns out a UI setting wouldn't really work here, for 2 reasons: 

1. the appUserID **must** be set if this mode is enabled. Totally fixable, though, and NaBD, we'd just add a field for it
2. and more important: PurchaseTester is really not built around this mode. So it'll still call customerInfo and other APIs, so it's not representative of this behavior in the slightest. 

I was thinking of just adding another TestingApp, single page, that just offers a button to initiate a purchase and maybe prints out customerInfo stream events when they come in, as well as offerings. Very simple. I already had to write it to test this out in the first place. 

What do you think?

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1318,14 +1343,20 @@ private extension Purchases {
         isAppBackgrounded: Bool,
         completion: ((Result<CustomerInfo, PublicError>) -> Void)?
     ) {
+        self.offeringsManager.updateOfferingsCache(appUserID: self.appUserID,
+                                                   isAppBackgrounded: isAppBackgrounded,
+                                                   completion: nil)
+
+        guard !self.systemInfo.dangerousSettings.customEntitlementComputation else {
```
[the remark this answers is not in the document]
aboedo: replied over slack. Features will be disabled at the backend layer instead of SDK

aboedo:
The logic in this branch won't change, but I'm going to leave it as draft until it has tests and the header in it

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -742,6 +743,49 @@ class HTTPClientTests: TestCase {
...
+        self.client = HTTPClient(apiKey: self.apiKey, systemInfo: systemInfo, eTagManager: self.eTagManager)
+
+        let request = HTTPRequest(method: .post([:]), path: .mockPath)
+
+        var headerPresent = false
+
+        stub(condition: isPath(request.path)) { request in
+            let headers =  request.allHTTPHeaderFields ?? [:]
+            headerPresent = headers["X-Custom-Entitlements-Computation"] != nil
+                && headers["X-Custom-Entitlements-Computation"] == "true"
```
NachoSoto, line 757: This can be simply
```suggestion
            headerPresent = headers["X-Custom-Entitlements-Computation"] == "true"
```

Tests/TestingApps/PurchaseTesterSwiftUI/Core/ConfiguredPurchases.swift
NachoSoto: > turns out a UI setting wouldn't really work here, for 2 reasons:

Oh yeah good point. Do you think it's worth maintaining a separate app though?

NachoSoto:
Should we merge this in `main` instead and do the hotfix after? We already released 4.17.11

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -418,6 +430,14 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
          purchasesOrchestrator: PurchasesOrchestrator,
          trialOrIntroPriceEligibilityChecker: TrialOrIntroPriceEligibilityCheckerType
     ) {
+
+        if systemInfo.dangerousSettings.customEntitlementComputation {
+            Logger.info("""
+            Entering customEntitlementComputation mode. CustomerInfo cache will not be automatically fetched. Anonymous
```
NachoSoto, line 436: Reminder to move this to strings

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1292,18 +1317,18 @@ private extension Purchases {
...
-            guard self.deviceCache.isOfferingsCacheStale(isAppBackgrounded: isAppBackgrounded) else {
-                return
+        self.systemInfo.isApplicationBackgrounded { isAppBackgrounded in
+            if !self.systemInfo.dangerousSettings.customEntitlementComputation {
+                self.customerInfoManager.fetchAndCacheCustomerInfoIfStale(appUserID: self.appUserID,
+                                                                          isAppBackgrounded: isAppBackgrounded,
+                                                                          completion: nil)
             }
 
-            Logger.debug("Offerings cache is stale, updating caches")
```
NachoSoto, line 1303: Did you mean to remove this?

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1318,14 +1343,20 @@ private extension Purchases {
...
+            completion?(.failure(error.asPublicError))
+            return
+        }
+
         self.customerInfoManager.fetchAndCacheCustomerInfo(appUserID: self.appUserID,
                                                            isAppBackgrounded: isAppBackgrounded) { @Sendable in
             completion?($0.mapError { $0.asPublicError })
         }
-
-        self.offeringsManager.updateOfferingsCache(appUserID: self.appUserID,
```
NachoSoto, line 1326: Did you mean to remove this?

Tests/APITesters/ObjCAPITester/ObjCAPITester/RCPurchasesErrorCodeAPI.m
```diff
@@ -58,6 +58,8 @@ + (void)checkEnums {
         case RCInvalidPromotionalOfferError:
         case RCOfflineConnectionError:
             NSLog(@"%ld", (long)errCode);
+        case RCFeatureNotAvailableInCustomEntitlementsComputationMode:
```
NachoSoto, line 61: Nit but you could move this up with the other cases to not need the `break`

NachoSoto:
Should we close this in favor of #2439?

aboedo:
yes, closing in favor of #2439

PR #2424, opened 2023-04-20
re-enable entitlement mapping integration test

aboedo:
Re-enables the entitlement mapping integration test on iOS. 

Note that it currently **doesn't work**, I'm not entirely sure why. 

We need to fix it before shipping.


aboedo:
closing in favor of @NachoSoto's branch

PR #2429, opened 2023-04-21
customEntitlementsComputation part 4: create convenience configuration method

aboedo:
Adds a new configuration method that's meant as a convenience for enabling customEntitlementsComputation. 

Syntax using this method is just: 

```swift
Purchases.configureInCustomEntitlementsComputationMode(apiKey: "<api_key>", appUserID: "<app_user_id>")
```

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1008,6 +1008,49 @@ public extension Purchases {
     }
 
     /**
+     * Configures an instance of the Purchases SDK with a specified API key and
+     * app user ID in Custom Entitlements Computation mode.
+
+     * - Warning: Configuring in Custom Entitlements Computation mode should only be enabled after
+     * being instructed to do so by the RevenueCat team.
+     * Apps configured in this mode will not have anonymous IDs, will not be able to use logOut methods,
+     * and will not have their CustomerInfo cache refreshed automatically.
+     *
+     * ## Custom Entitlements Computation mode
+     * This mode is intended for apps that will use RevenueCat to manage payment flows,
+     * but **will not** use RevenueCat's SDK to compute entitlements.
+     * Apps using this mode will instead rely on webhooks to get notified when purchases go through
+     * and to merge information between RevenueCat's servers
+     * and their own.
+     *
+     * In this mode, the RevenueCat SDK will never generate anonymous IDs. Instead, it can only be configured
+     * with a known appUserID, and the logOut methods
+     * will return an error if called. To change users, call ``logIn(_:)-arja``.
+     *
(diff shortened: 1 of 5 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1007,6 +1007,49 @@ public extension Purchases {
...
+     * - Parameter apiKey: The API Key generated for your app from https://app.revenuecat.com/
+     *
+     * - Parameter appUserID: The unique app user id for this user. This user id will allow users to share their
+     * purchases and subscriptions across devices. Pass `nil` or an empty string if you want ``Purchases``
+     * to generate this for you.
+     *
+     * - Returns: An instantiated ``Purchases`` object that has been set as a singleton.
+     */
+    @objc(configureInCustomEntitlementsModeWithApiKey:appUserID:)
+    @discardableResult static func configureInCustomEntitlementsComputationMode(apiKey: String,
```
aboedo, line 1044: Notable advantage to the convenience method: you can't forget to pass in `appUserID`. So it's a cleaner API in that you can't misuse it. 

couldn't find a much better way to test this method out with unit tests, since the paths where we call configure are not easy to mock. 
I will however be able to test it thoroughly with StoreKit Integration tests, coming up as a separate PR. 
In this one, I did add API tests.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1007,6 +1007,49 @@ public extension Purchases {
         Self.configure(withAPIKey: apiKey, appUserID: appUserID, observerMode: false)
     }
 
+    /**
+     * Configures an instance of the Purchases SDK with a specified API key and
```
aboedo, line 1011: this is that this section looks like when rendered in DocC:
image

image

NachoSoto (commented):
I love this, but what do you think about adding this helper to `Configuration.Builder` instead? That'll make testing trivial, and avoids increasing the size of the already huge `Purchases` class.
```
let purchases = Purchases.configure(
           .customEntitlementsComputationMode(appUserID: appUserID)
           .build()
)
```

aboedo:
TBH, I think the builder is great when there's a lot of parameters, but when it comes to creating as frictionless a path as possible for anyone trying out our SDK, there's definitely a decent amount of extra cognitive load in having to pass in a builder method. 

I don't love the idea of increasing the size of the megaPurchases class, but I think it's worth it in this case since we want to keep it as simple as humanly possible for the most simple use cases

NachoSoto:
Makes sense 👍🏻

aboedo:
@NachoSoto merging for now, happy to work on updating implementation for testability once we're done with the main implementation

PR #2430, opened 2023-04-21
`PurchaseOrchestrator`: fix incorrect `InitiationSource` for SK1 queue transactions

NachoSoto:
All StoreKit 1 transactions were being sent with `InitiationSource.purchased`.
#1957 introduced this, but it was only set correctly for SK2.

The consequence of this is that users might see "invalid receipt" errors when posting receipts due to updated transactions from the queue, which can only happen for `InitiationSource.purchased`.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -570,7 +570,9 @@ extension PurchasesOrchestrator: StoreKit1WrapperDelegate {
         let storeTransaction = StoreTransaction(sk1Transaction: transaction)
 
         switch transaction.transactionState {
-        case .restored: // for observer mode
+        // For observer mode. Should only come from calls to `restoreCompletedTransactions`,
+        // which the SDK does not currently use.
+        case .restored:
             self.handlePurchasedTransaction(storeTransaction,
                                             storefront: storeKit1Wrapper.currentStorefront,
                                             restored: true)
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -813,6 +813,8 @@ private extension PurchasesOrchestrator {
         for productIdentifier: String,
         restored: Bool
     ) -> ProductRequestData.InitiationSource {
+        // Having a purchase completed callback implies that the transation comes from an explicit call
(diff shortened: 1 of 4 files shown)


tonidero (approved):
LGTM!

NachoSoto:
@aboedo does this look good to you?

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -570,9 +570,14 @@ extension PurchasesOrchestrator: StoreKit1WrapperDelegate {
         let storeTransaction = StoreTransaction(sk1Transaction: transaction)
 
         switch transaction.transactionState {
-        case .restored, // for observer mode
-             .purchased:
-            self.handlePurchasedTransaction(storeTransaction, storefront: storeKit1Wrapper.currentStorefront)
+        case .restored: // for observer mode
```
aboedo, line 573: not from this PR, but we should probably expand with some comments here re: observer mode: 

When something comes through the `.restore` flow in the queue, that's actually not even from our SDK - it's from a call to [restoreCompletedTransactions](https://developer.apple.com/documentation/storekit/skpaymentqueue/1506123-restorecompletedtransactions#), which we don't use

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -799,6 +808,22 @@ private extension PurchasesOrchestrator {
             #endif
         }
     }
+
+    private func initiationSource(
+        for productIdentifier: String,
+        restored: Bool
+    ) -> ProductRequestData.InitiationSource {
+        // Having a purchase completed callback implies that the transation comes from an explicit call
+        // to `purchase()` instead of a StoreKit transaction notification.
```
aboedo, line 817: I think this works.

I got quite confused at first, since that restore isn't coming from us, and we do offer the ability to attach a completion block to restores when they get initiated through our SDK. 

But restores that _are_ coming from us won't actually go through this flow in the first place since they don't issue transactions in the queue. Instead, for our restores we just refresh the receipt and send it. 

Maybe we can document that behavior more in detail?

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: I'll add a comment 👍🏻

PR #2431, opened 2023-04-21
customEntitlementsMode: Sample app

aboedo:
Creates a sample app for CustomEntitlementsMode. 

It showcases how to configure and use the SDK, what it can do in this mode, and more. 

Sample screenshot

https://user-images.githubusercontent.com/3922667/233727555-1beaa352-4c79-428d-99d9-701bb6990fd7.MP4


Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
```diff
@@ -0,0 +1,275 @@
...
+            for await customerInfo in Purchases.shared.customerInfoStream {
+                DispatchQueue.main.async {
+                    print("got new customerInfo: \(customerInfo)")
+                    customerInfoStreamFires.append(CustomerInfoStreamFires(customerInfo: customerInfo, date: Date()))
+                }
+            }
+        }
+    }
+
+    func purchaseFirstOffering() async throws {
```
NachoSoto, line 149: I think you meant to make this not `throws` since you're logging the error inside.

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
```diff
@@ -0,0 +1,275 @@
...
+            .foregroundColor(.white)
+            .frame(maxWidth: .infinity)
+            .padding()
+            .background(Color.blue)
+            .cornerRadius(20)
+            .padding()
+
+            .onAppear {
+                Purchases.configureInCustomEntitlementsComputationMode(apiKey: Constants.apiKey,
+                                                                       appUserID: appUserID)
```
NachoSoto, line 81: This would get called multiple times. Maybe use `.task` instead of `.onAppear` for configuration, which also would allow you to use `async` methods.

NachoSoto (commented):
Nice 👍🏻

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
aboedo: good catch, I added the catch halfway through and forgot to update the signature

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
aboedo: it's what I get for pretending I know swiftUI 🙈

aboedo: updating

aboedo: done

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
```diff
@@ -0,0 +1,278 @@
...
+
+    func logIn(_ appUserID: String) {
+        print("Switching to user: \(appUserID)")
+        Task {
+            try await Purchases.shared.logIn(appUserID)
+        }
+    }
+
+    func subscribeToCustomerInfoStream() {
+        self.streamTask = Task {
```
NachoSoto, line 142: Maybe do this to ensure no errors are thrown:
```suggestion
        self.streamTask = Task<Void, Never> {
```

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
```diff
@@ -0,0 +1,278 @@
...
+            for await customerInfo in Purchases.shared.customerInfoStream {
+                DispatchQueue.main.async {
```
NachoSoto, line 144: This shouldn't be necessary, are you getting an error without it?

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
```diff
@@ -0,0 +1,278 @@
...
+            .padding()
+
+            .task {
+                Purchases.configureInCustomEntitlementsComputationMode(apiKey: Constants.apiKey,
+                                                                       appUserID: appUserID)
+                do {
+                    self.offerings = try await Purchases.shared.offerings()
+                    print("offerings: \(String(describing: offerings))")
+                }
+                catch {
```
NachoSoto, line 86: Nit per our style guide:
```suggestion
                } catch {
```

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
```diff
@@ -0,0 +1,278 @@
...
+                                    .bold()
+                                Text("\(customerInfoStreamFire.date, formatter: dateFormatter)")
+                            }
+                            HStack {
+                                Text("App User ID:")
+                                    .bold()
+                                Text(customerInfoStreamFire.customerInfo.originalAppUserId)
+                            }
+                        }
+                    }                }
```
NachoSoto, line 108: I think you need a line break in here

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
```diff
@@ -0,0 +1,278 @@
...
+                        VStack {
+                            Text("CustomerInfo stream values")
+                                .font(.headline)
+                                .foregroundColor(.primary)
+                                .padding()
+                            Text("List fills in below when new values arrive.")
+                                .foregroundColor(.primary)
+                        }
+                    }
+                }.task {
```
NachoSoto, line 121: Nit: I would put this in its own line for consistency with the other methods:
```suggestion
                }
                .task {
```

NachoSoto (approved):
Nothing blocking, this is just a test app after all.

Examples/testCustomEntitlementsComputation/README.md
```diff
@@ -0,0 +1,32 @@
+# Test Custom Entitlements Computation App
```
tonidero, line 1: Hmm the `test` part on the name of the app could be confusing, since this is not running tests or anything... How about `Custom entitlements computation sample app`?

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
```diff
@@ -0,0 +1,278 @@
...
+        .padding()
+        .navigationTitle("Customer Info Details")
+    }
+}
+
+struct ContentView_Previews: PreviewProvider {
+    static var previews: some View {
+        ContentView()
+    }
+}
```
tonidero, line 278: This file has quite a few things... But I think it's ok for a quick sample app.

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
```diff
@@ -0,0 +1,278 @@
...
+    let date: Date
+}
+
+struct ContentView: View {
+
+    @State private var streamTask: Task<Void, Never>?
+    @State private var offerings: Offerings?
+    @State private var customerInfoStreamFires: [CustomerInfoStreamFires] = []
+    @State private var showingLogInAlert = false
+    @State private var appUserID = Constants.defaultAppUserID
```
tonidero, line 23: I guess it could be useful to be able to select the initial app user id before configuring the SDK... But not needed for V1 of this 👍

tonidero (approved):
Looks great!

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
aboedo: I genuinely don't remember why I added it, removing

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
aboedo: good call

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
aboedo: I genuinely debated whether to split this up when writing it, since adding more files in swift is pretty trivial. Not entirely sure it's a net win for this - it feels like if you're using this as a reference, it's kinda nice for all the logic to be in one place so you don't need to figure out how to navigate it. Especially given that it's not that long of a file

aboedo: thoughts?

Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
aboedo: Agreed. I didn't do it to keep this as simple as possible, since having a way to set up an appUserID would almost instantly mean having some navigation where you start from the configuration screen then go to a second screen and we're trying to keep the scope as small as possible here

PR #2432, opened 2023-04-23
Fixed docs

kaunteya:
### Checklist
- No code changes

### Motivation
Updated the docs for async functions. These were missed probably because they were copied from their completion-handler based counterparts.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesType.swift
+++ b/Sources/Purchasing/Purchases/PurchasesType.swift
@@ -277,12 +277,14 @@ public protocol PurchasesType: AnyObject {
      * - Important: Call this method when a user has decided to purchase a product.
      * Only call this in direct response to user input.
      *
-     * From here ``Purchases`` will handle the purchase with `StoreKit` and call the ``PurchaseCompletedBlock``.
+     * From here ``Purchases`` will handle the purchase with `StoreKit` and return ``PurchaseResultData``.
      *
-     * - Note: You do not need to finish the transaction yourself in the completion callback, ``Purchases`` will
+     * - Note: You do not need to finish the transaction yourself after this, ``Purchases`` will
      * handle this for you.
      *
      * - Parameter product: The ``StoreProduct`` the user intends to purchase.
+     *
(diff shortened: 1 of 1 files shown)


Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -277,13 +277,15 @@ public protocol PurchasesType: AnyObject {
...
-     * From here ``Purchases`` will handle the purchase with `StoreKit` and call the ``PurchaseCompletedBlock``.
+     * From here ``Purchases`` will handle the purchase with `StoreKit` and return ``PurchaseResultData``.
      *
-     * - Note: You do not need to finish the transaction yourself in the completion callback, ``Purchases`` will
+     * - Note: You do not need to finish the transaction yourself after this, ``Purchases`` will
      * handle this for you.
      *
      * - Parameter product: The ``StoreProduct`` the user intends to purchase.
      *
+     * If the purchase was successful there will be a ``StoreTransaction`` and a ``CustomerInfo``.
```
kaunteya, line 287: Not sure if this is true!! for async API

NachoSoto: Yup that's part of the tuple 👍🏻 maybe put this and the next line about `userCancelled` inside of `- Returns:` blocks?

NachoSoto (approved):
Great thanks! Just a small suggestion to move those new lines inside of a new `Returns` block.

kaunteya:
Hi. Why is CustomerInfo optional in `PurchaseCompletedBlock` but non-optional in `PurchaseResultData`?
I need this info to make the doc more clear

NachoSoto:
The reason for that is that the completion block API returns _either_ customer info or an error. When it fails with an `error`, `CustomerInfo` is `nil`. For the `async throws` APIs this is made explicit since we can ensure that it either `throws` an error or it returns a value.

kaunteya:
@NachoSoto I have updated the docs based on your suggestions. Please have a look

NachoSoto (approved):
Thanks!

PR #2436, opened 2023-04-24
`customEntitlementsComputation`: disable automatic sync of `attConsentStatus`

aboedo:
Removed automatic syncing of attConsentStatus when sending  a POST /receipt for customEntitlementsComputation

### DIFF
--- a/Sources/Networking/BackendConfiguration.swift
+++ b/Sources/Networking/BackendConfiguration.swift
@@ -20,15 +20,18 @@ final class BackendConfiguration {
     let operationDispatcher: OperationDispatcher
     let operationQueue: OperationQueue
     let dateProvider: DateProvider
+    let systemInfo: SystemInfo
 
     init(httpClient: HTTPClient,
          operationDispatcher: OperationDispatcher,
          operationQueue: OperationQueue,
-         dateProvider: DateProvider = DateProvider()) {
+         dateProvider: DateProvider = DateProvider(),
+         systemInfo: SystemInfo) {
         self.httpClient = httpClient
         self.operationDispatcher = operationDispatcher
         self.operationQueue = operationQueue
         self.dateProvider = dateProvider
+        self.systemInfo = systemInfo
     }
 
     func clearCache() {
--- a/Sources/Networking/CustomerAPI.swift
+++ b/Sources/Networking/CustomerAPI.swift
@@ -91,12 +91,18 @@ final class CustomerAPI {
               initiationSource: ProductRequestData.InitiationSource,
               subscriberAttributes subscriberAttributesByKey: SubscriberAttribute.Dictionary?,
               completion: @escaping CustomerAPI.CustomerInfoResponseHandler) {
-        let attributionStatus = self.attributionFetcher.authorizationStatus
-        var subscriberAttributesByKey = subscriberAttributesByKey ?? [:]
-        let consentStatus = SubscriberAttribute(withKey: ReservedSubscriberAttribute.consentStatus.rawValue,
-                                                value: attributionStatus.description,
-                                                dateProvider: self.backendConfig.dateProvider)
-        subscriberAttributesByKey[ReservedSubscriberAttribute.consentStatus.key] = consentStatus
+        var subscriberAttributesToPost: SubscriberAttribute.Dictionary? = nil
+
+        if !self.backendConfig.systemInfo.dangerousSettings.customEntitlementComputation {
+            subscriberAttributesToPost = subscriberAttributesByKey ?? [:]
+            let attributionStatus = self.attributionFetcher.authorizationStatus
+            let consentStatus = SubscriberAttribute(withKey: ReservedSubscriberAttribute.consentStatus.rawValue,
+                                                    value: attributionStatus.description,
+                                                    dateProvider: self.backendConfig.dateProvider)
+            subscriberAttributesToPost?[ReservedSubscriberAttribute.consentStatus.key] = consentStatus
+        }
+
+
 
         let config = NetworkOperation.UserSpecificConfiguration(httpClient: self.backendConfig.httpClient,
                                                                 appUserID: appUserID)
@@ -108,7 +114,7 @@ final class CustomerAPI {
                                                          presentedOfferingIdentifier: offeringIdentifier,
                                                          observerMode: observerMode,
                                                          initiationSource: initiationSource,
-                                                         subscriberAttributesByKey: subscriberAttributesByKey)
+                                                         subscriberAttributesByKey: subscriberAttributesToPost)
         let factory = PostReceiptDataOperation.createFactory(configuration: config,
                                                              postData: postData,
                                                              customerInfoCallbackCache: self.customerInfoCallbackCache)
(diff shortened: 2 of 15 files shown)


Sources/Networking/CustomerAPI.swift
```diff
@@ -91,12 +91,18 @@ final class CustomerAPI {
...
-        let consentStatus = SubscriberAttribute(withKey: ReservedSubscriberAttribute.consentStatus.rawValue,
-                                                value: attributionStatus.description,
-                                                dateProvider: self.backendConfig.dateProvider)
-        subscriberAttributesByKey[ReservedSubscriberAttribute.consentStatus.key] = consentStatus
+        var subscriberAttributesToPost: SubscriberAttribute.Dictionary? = nil
+
+        if !self.backendConfig.systemInfo.dangerousSettings.customEntitlementComputation {
+            subscriberAttributesToPost = subscriberAttributesByKey ?? [:]
+            let attributionStatus = self.attributionFetcher.authorizationStatus
+            let consentStatus = SubscriberAttribute(withKey: ReservedSubscriberAttribute.consentStatus.rawValue,
```
NachoSoto, line 99: Turns out we have no tests for this.

NachoSoto: I guess we did through snapshots, never mind.

Sources/Networking/CustomerAPI.swift
```diff
@@ -108,7 +112,7 @@ final class CustomerAPI {
                                                          presentedOfferingIdentifier: offeringIdentifier,
                                                          observerMode: observerMode,
                                                          initiationSource: initiationSource,
-                                                         subscriberAttributesByKey: subscriberAttributesByKey)
+                                                         subscriberAttributesByKey: subscriberAttributesToPost)
```
tonidero, line 115: I guess this would be null if the new mode is on... probably ok but I thought I would mention it.

tonidero (approved):
LGTM!

Sources/Networking/BackendConfiguration.swift
```diff
@@ -20,15 +20,18 @@ final class BackendConfiguration {
...
     let operationQueue: OperationQueue
     let dateProvider: DateProvider
+    let systemInfo: SystemInfo
 
     init(httpClient: HTTPClient,
          operationDispatcher: OperationDispatcher,
          operationQueue: OperationQueue,
-         dateProvider: DateProvider = DateProvider()) {
+         dateProvider: DateProvider = DateProvider(),
+         systemInfo: SystemInfo) {
```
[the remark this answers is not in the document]
NachoSoto: Swift doesn't require that, that's an arbitrary requirement from lesser languages without named parameters 🤡

Sources/Networking/CustomerAPI.swift
NachoSoto: Yeah, since that's encoded in the type system, the method handles that explicitly. It's also tested now :)

aboedo: yeah I figured since the method took in a nullable value, it was safer to send in an null so that it wouldn't get translated into an empty list instead. I did test manually with POST receipt to make sure it worked fine

Sources/SubscriberAttributes/SubscriberAttribute.swift
```diff
@@ -32,6 +32,10 @@ struct SubscriberAttribute {
         self.init(withKey: key, value: value, isSynced: false, setTime: dateProvider.now())
     }
 
+    init(attribute: ReservedSubscriberAttribute, value: String?, dateProvider: DateProvider = DateProvider()) {
+        self.init(withKey: attribute.rawValue, value: value, dateProvider: dateProvider)
+    }
+
```
aboedo, line 38: do we need this?

Tests/UnitTests/Networking/Backend/BackendPostReceiptDataTests.swift
```diff
@@ -687,7 +691,45 @@ class BackendPostReceiptDataTests: BaseBackendTests {
...
+                              presentedOfferingIdentifier: nil,
+                              observerMode: false,
+                              initiationSource: .queue,
+                              subscriberAttributes: nil,
+                              completion: { _ in
+                completed()
+            })
+        }
+
+        expect(self.httpClient.calls).to(haveCount(1))
```
aboedo, line 727: I might be missing something, but I don't see anything here specifically checking that attConsentStatus isn't sent. Are we relying on snapshot testing for that part?

aboedo (commented):
can't approve because I opened but it looks great!

Tests/UnitTests/Networking/Backend/BackendPostReceiptDataTests.swift
NachoSoto: Yeah. Do you think that's too implicit and we should have something more clear? It would require introspecting the `[String: Any]` of the request post. We didn't have any explicit tests for the opposite case either.

Sources/SubscriberAttributes/SubscriberAttribute.swift
NachoSoto: It cleaned up the call so we sent the `enum` instead of a random `String`. I can remove it since it's only used in 1 place though.

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostReceiptDataTests/iOS12-testDoesNotPostConsentStatus.1.json
```diff
@@ -0,0 +1,16 @@
+{
+  "headers" : {
+    "Authorization" : "Bearer asharedsecret"
+  },
+  "request" : {
+    "body" : {
```
NachoSoto, line 6: @aboedo this is the "test". Equivalent to the normal one: https://github.com/RevenueCat/purchases-ios/blob/main/Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostReceiptDataTests/iOS16-testPostsReceiptDataCorrectly.1.json#L9

Tests/UnitTests/Networking/Backend/BackendPostReceiptDataTests.swift
aboedo: I was a bit confused about it when I read the test name tbh, since the test doesn't actually check for it. 
I think it's probably nice to have it since it might be clearer if you mess up the code, when you see exactly what test failed and how, but it's not the end of the world

NachoSoto: >>> rror: -[UnitTests.BackendPostReceiptCustomEntitlementsTests testDoesNotPostConsentStatus] : failed - Snapshot does not match reference.

@−
"Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostReceiptDataTests/iOS16-testDoesNotPostConsentStatus.1.json"
@+
"BackendPostReceiptDataTests/iOS16-testDoesNotPostConsentStatus.1.json"

@@ −4,8 +4,14 @@
   },
   "request" : {
     "body" : {
       "app_user_id" : "user",
+      "attributes" : {
+        "$attConsentStatus" : {
+          "updated_at_ms" : 1678307200000,
+          "value" : "authorized"
+        }
+      },
       "fetch_token" : "YW4gYXdlc29tZSByZWNlaXB0",
       "initiation_source" : "queue",
       "is_restore" : false,
       "observer_mode" : false

PR #2437, opened 2023-04-24
`customEntitlementsComputation`: new method to switch users

aboedo:
Adds a new method to switch users without posting to the backend for Custom Entitlements Computation mode.

### DIFF
--- a/Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
+++ b/Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
@@ -133,10 +133,7 @@ struct ContentView: View {
     }
 
     func logIn(_ appUserID: String) {
-        print("Switching to user: \(appUserID)")
-        Task {
-            Purchases.shared.switchUser(newAppUserID: appUserID)
-        }
+        Purchases.shared.switchUser(to: appUserID)
     }
 
     func subscribeToCustomerInfoStream() {
--- a/Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
+++ b/Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
@@ -135,7 +135,7 @@ struct ContentView: View {
     func logIn(_ appUserID: String) {
         print("Switching to user: \(appUserID)")
         Task {
-            try await Purchases.shared.logIn(appUserID)
+            Purchases.shared.switchUser(newAppUserID: appUserID)
         }
     }
 
--- a/Sources/Identity/IdentityManager.swift
+++ b/Sources/Identity/IdentityManager.swift
@@ -91,8 +91,9 @@ class IdentityManager: CurrentUserProvider {
         }
     }
 
-    func switchUser(newAppUserID: String) {
-        self.resetCacheAndSaveNewUserID(newUserID: newAppUserID)
+    func switchUser(to newAppUserID: String) {
+        Logger.debug(Strings.identity.switching_user(newUserId: newAppUserID))
+        self.resetCacheAndSave(newUserID: newAppUserID)
     }
 
     static func generateRandomID() -> String {
@@ -153,7 +154,7 @@ private extension IdentityManager {
             return
         }
 
-        self.resetCacheAndSaveNewUserID(newUserID: Self.generateRandomID())
+        self.resetCacheAndSave(newUserID: Self.generateRandomID())
         Logger.info(Strings.identity.log_out_success)
         completion(nil)
     }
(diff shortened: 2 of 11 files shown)


Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
```diff
@@ -135,7 +135,7 @@ struct ContentView: View {
     func logIn(_ appUserID: String) {
         print("Switching to user: \(appUserID)")
         Task {
```
NachoSoto, line 137: You don't need this `Task` anymore?

Sources/Identity/IdentityManager.swift
```diff
@@ -90,6 +90,10 @@ class IdentityManager: CurrentUserProvider {
         }
     }
 
+    func switchUser(newAppUserID: String) {
```
NachoSoto, line 93: The Swift-y way would be: `switchUser(to newAppuserID: String)`

Sources/Identity/IdentityManager.swift
```diff
@@ -162,8 +166,8 @@ extension IdentityManager: @unchecked Sendable {}
 
 private extension IdentityManager {
 
-    func resetUserIDCache() {
-        deviceCache.clearCaches(oldAppUserID: currentAppUserID, andSaveWithNewUserID: Self.generateRandomID())
+    func resetCacheAndSaveNewUserID(newUserID: String) {
```
NachoSoto, line 169: `newUserID` is repeated. This should be `resetCacheAndSave(newUserID:)`

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -630,6 +630,20 @@ public extension Purchases {
         return try await logOutAsync()
     }
 
+    ///
+    /// Updates the current appUserID to a new one, without associating the two.
+    /// This method is **only available** in Custom Entitlements Computation mode.
+    /// Receipts posted by the SDK to the RevenueCat backend after calling this method will be sent
+    /// with the newAppUserID.
+    /// 
+    func switchUser(newAppUserID: String) {
```
NachoSoto, line 639: Ditto:
```suggestion
    func switchUser(to newAppUserID: String) {
```

Will this be available to `Obj-C`? It'll probably need a custom name too.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -630,6 +630,20 @@ public extension Purchases {
         return try await logOutAsync()
     }
 
+    ///
+    /// Updates the current appUserID to a new one, without associating the two.
+    /// This method is **only available** in Custom Entitlements Computation mode.
```
NachoSoto, line 635: Maybe make this `- Important`? (I think that's the format?)

Sources/Identity/IdentityManager.swift
```diff
@@ -90,6 +90,10 @@ class IdentityManager: CurrentUserProvider {
         }
     }
 
+    func switchUser(newAppUserID: String) {
+        self.resetCacheAndSaveNewUserID(newUserID: newAppUserID)
```
NachoSoto, line 94: We might need a log here.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -630,6 +630,20 @@ public extension Purchases {
...
+    /// Receipts posted by the SDK to the RevenueCat backend after calling this method will be sent
+    /// with the newAppUserID.
+    /// 
+    func switchUser(newAppUserID: String) {
+        guard self.systemInfo.dangerousSettings.customEntitlementComputation else {
+            return
```
NachoSoto, line 641: I know this is WIP, but this could use a log here too :)

NachoSoto (commented):
Makes sense. Maybe we can link to this method in the new `Purchases.configure` method.

tonidero:
Just a heads up that I'm taking over this PR. Will address current comments first then test it manually and add unit tests.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -630,6 +630,22 @@ public extension Purchases {
...
     }
 
+    ///
+    /// Updates the current appUserID to a new one, without associating the two.
+    /// - Important: This method is **only available** in Custom Entitlements Computation mode.
+    /// Receipts posted by the SDK to the RevenueCat backend after calling this method will be sent
+    /// with the newAppUserID.
+    ///
+    @objc(switchUserTo:)
+    func switchUser(to newAppUserID: String) {
```
tonidero, line 640: It definitely feels some users might inadvertently use this method and it will error silently for them. I wonder if we should just crash if they are not in custom entitlement computation mode... Also, moving this to a different package could avoid this confusion.

tonidero:
Tested this manually and seems to work fine. Also added some unit tests though if we end up going with the approach of building a different package, we might have to move them as well... In any case, this is mostly final now.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -641,6 +641,22 @@ public extension Purchases {
         return try await logOutAsync()
     }
 
+    ///
+    /// Updates the current appUserID to a new one, without associating the two.
+    /// - Important: This method is **only available** in Custom Entitlements Computation mode.
+    /// Receipts posted by the SDK to the RevenueCat backend after calling this method will be sent
+    /// with the newAppUserID.
+    ///
+    @objc(switchUserTo:)
```
NachoSoto, line 650: I think in `Obj-C` this would be better as `switchUserToNewAppUserID:` (the conventions are different in both languages).

NachoSoto (approved):
Just one comment about the Obj-C method

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: Solved by #2440.

PR #2442, opened 2023-04-25
Custom Entitlement Computation: disabled unnecessary APIs

NachoSoto:
## Disabled API:

- `getCustomerInfo`
- `logIn`
- `logOut`
- `Purchases.attribution`
- `syncPurchases`
- `restorePurchases`
- Trial eligibility
- Promotional purchases
- `async` counterparts

PR #2443, opened 2023-04-25
`customEntitlementsComputation`: update the copy in the sample app to explain the new usage

aboedo:
Updated the copy in the sample app to explain the new usage


NachoSoto (approved):

NachoSoto:
Tests are fixed if you rebase.

Examples/testCustomEntitlementsComputation/README.md
```diff
@@ -31,4 +31,76 @@ To use this mode, ensure that you install the RevenueCat_CustomEntitlementComput
...
+}
+```
+
+### Observing changes to purchases:
+
+To ensure that your app reacts to changes to subscriptions in real time, you can use `customerInfoStream`. This stream will only fire when new `customerInfo` is registered
+in RevenueCat, like when a subscription is renewed. If there are no changes from the last value, it will not fire. This means it's not guaranteed to fire on every app open.
+
+```swift
+for await customerInfo in Purchases.shared.customerInfoStream {
```
NachoSoto, line 103: Oops I think I disabled this too #2442, fixing.

NachoSoto:
You can rebase this now.

PR #2447, opened 2023-04-26
Change user cancelled behavior in custom entitlement computation mode

tonidero:
### Description
Change user cancelled behavior in custom entitlement mode. Now for

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -413,9 +413,11 @@ final class PurchasesOrchestrator {
                     productIdentifier: product.id,
                     error: error
                 ))
+                let publicError = ErrorUtils.purchasesError(withUntypedError: error).asPublicError
+                let userCancelled = publicError.code == ErrorCode.purchaseCancelledError.rawValue
 
                 DispatchQueue.main.async {
-                    completion(nil, nil, ErrorUtils.purchasesError(withUntypedError: error).asPublicError, false)
+                    completion(nil, nil, publicError, userCancelled)
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


NachoSoto:
TODO: add an assertion to ensure there's no customer info request in this case.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -449,6 +451,9 @@ final class PurchasesOrchestrator {
                     return try await sk2Product.purchase(options: options)
                 }
         } catch StoreKitError.userCancelled {
+            if self.systemInfo.dangerousSettings.customEntitlementComputation {
+                throw ErrorUtils.purchaseCancelledError()
+            }
```
tonidero, line 456: I'm not sure if this error would be catched by the following catches... Still had to test each of these cases to make sure it behaved as I expected

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -413,9 +413,11 @@ final class PurchasesOrchestrator {
                     productIdentifier: product.id,
                     error: error
                 ))
+                let publicError = ErrorUtils.purchasesError(withUntypedError: error).asPublicError
+                let userCancelled = publicError.code == ErrorCode.purchaseCancelledError.rawValue
```
NachoSoto, line 417: Oh I see why you needed this now. We need to compare the error domain too though, I'll fix that.

NachoSoto: I'm changing it to this:
```suggestion
                let userCancelled = publicError.isCancelledError
```

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -466,6 +471,9 @@ final class PurchasesOrchestrator {
         // This detects if `Product.PurchaseResult.userCancelled` is true.
         let (userCancelled, sk2Transaction) = try await self.storeKit2TransactionListener
             .handle(purchaseResult: result)
+        if self.systemInfo.dangerousSettings.customEntitlementComputation {
```
NachoSoto, line 474: I think this was missing also checking `userCancelled`.

NachoSoto:
Closing in favor of #2449

PR #2448, opened 2023-04-26
Hide promotional offers APIs in custom entitlements computation package

tonidero:
### Description
This will remove the `promotionaOffer` purchase methods from the API in the custom entitlement computation package


tonidero (commented):
I thought about hiding the promotional offer related classes (e.g. `StoreProductDiscount`) as well... But that would complicate things a lot, and I think just hiding the entry methods should be ok

NachoSoto (approved):
Perfect.

PR #2449, opened 2023-04-26
Custom Entitlement Computation: avoid `getCustomerInfo` requests for cancelled purchases

NachoSoto:
Fixes SDK-3088
Supersedes #2447

Thanks to @tonidero for doing most of this.

Note that it's currently impossible to test this for SK2, so we only have SK1 tests (we have a Radar for that).

The next step will be to deprecate the current API to remove the `userCancelled` property and make this the new behavior.
See also #1910 for an explanation of the different return values of the purchase APIs.

### DIFF
--- a/Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
+++ b/Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
@@ -154,9 +154,16 @@ struct ContentView: View {
         }
 
         do {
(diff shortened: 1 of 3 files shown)


tonidero (approved):
LGTM! Planning to test this a bit more to make sure we didn't leave any edge cases

tonidero:
Actually should we also update the sample app so users of this mode can see how they should check if purchases were cancelled? So we can ask them to simply check out the sample code @NachoSoto

NachoSoto:
Done.

tonidero (approved):
❤️

PR #2450, opened 2023-04-26
Custom Entitlement Computation: mark `PurchaseResultData.userCancelled` as `unavailable`

NachoSoto:
Follow up to #2449.
This is step 2 of 3 towards getting rid of `userCancelled` in purchase results.

After #2449, `ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION` builds no longer send `userCancelled` `true` for cancellations, and instead throw `ErrorCode.purchaseCancelledError`.
To avoid users detecting cancellations looking at this property, this marks it `unavailable` with a message.

### How?

To achieve that, I converted the `tuple` into a `struct`. This also required moving the affected `purchase` methods from `PurchasesType` to `PurchasesSwiftType`, but that's an implementation detail.

The public API remains _mostly_ unchanged, with the only caveat that it's no longer possible to destructure the result tuple, so:
```swift
let (transaction, customerInfo, cancelled) = Purchases.shared.purchase(...)
```
Needs to become:
```swift
let result = Purchases.shared.purchase(...)
```

I updated the API tester to reflect that, as well as checking the properties of `PurchaseResultData`.

### Completion block API

As it's documented in #1910, completion block APIs already did send both `userCancelled` and the `error`, so we don't need to change those.

### Future changes

This paves the future for a full deprecation, where we mark it `deprecated` for normal builds as well, and change the behavior regardless of the `DangerousSettings`.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesType.swift
+++ b/Sources/Purchasing/Purchases/PurchasesType.swift
@@ -266,35 +266,10 @@ public protocol PurchasesType: AnyObject {
      * If the purchase was successful there will be a ``StoreTransaction`` and a ``CustomerInfo``.
      *
      * If the purchase was not successful, there will be an `NSError`.
-     *
      * If the user cancelled, `userCancelled` will be `true`.
      */
     @objc(purchaseProduct:withCompletion:)
     func purchase(product: StoreProduct, completion: @escaping PurchaseCompletedBlock)
-
-    /**
-     * Initiates a purchase of a ``StoreProduct``.
-     *
-     * Use this function if you are not using the ``Offerings`` system to purchase a ``StoreProduct``.
-     * If you are using the ``Offerings`` system, use ``Purchases/purchase(package:completion:)`` instead.
-     *
-     * - Important: Call this method when a user has decided to purchase a product.
-     * Only call this in direct response to user input.
-     *
-     * From here ``Purchases`` will handle the purchase with `StoreKit` and return ``PurchaseResultData``.
-     *
-     * - Note: You do not need to finish the transaction yourself after this, ``Purchases`` will
-     * handle this for you.
-     *
-     * - Parameter product: The ``StoreProduct`` the user intends to purchase.
-     *
-     * - Throws: An error of type ``ErrorCode`` is thrown if a failure occurs while purchasing
-     *
-     * - Returns: A tuple with ``StoreTransaction`` and a ``CustomerInfo`` if the purchase was successful.
-     * If the user cancelled the purchase, `userCancelled` will be `true`.
-     */
-    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
-    func purchase(product: StoreProduct) async throws -> PurchaseResultData
 
     /**
      * Initiates a purchase of a ``Package``.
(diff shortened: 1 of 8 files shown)


Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -262,36 +262,11 @@ public protocol PurchasesType: AnyObject {
      * If the purchase was successful there will be a ``StoreTransaction`` and a ``CustomerInfo``.
      *
      * If the purchase was not successful, there will be an `NSError`.
-     *
      * If the user cancelled, `userCancelled` will be `true`.
      */
     @objc(purchaseProduct:withCompletion:)
     func purchase(product: StoreProduct, completion: @escaping PurchaseCompletedBlock)
 
-    /**
```
NachoSoto, line 271: As explained in the description, these are moved from `PurchasesType` to `PurchasesSwiftType` because the result type is no longer representable in `Objective-C`. However, we didn't need these `async` APIs in Obj-C as we have the completion block ones.

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -613,9 +520,11 @@ public protocol PurchasesType: AnyObject {
     func checkTrialOrIntroDiscountEligibility(product: StoreProduct) async -> IntroEligibilityStatus
 
     /**
-     * Use this method to fetch ``PromotionalOffer``
-     *  to use in ``purchase(package:promotionalOffer:)`` or ``purchase(product:promotionalOffer:)``.
+     * Use this method to fetch ``PromotionalOffer`` to use in ``purchase(package:promotionalOffer:completion:)``
```
NachoSoto, line 523: This was producing a warning now because they're in the other `protocol`, but it made more sense to reference the completion block methods instead anyway.

Tests/APITesters/SwiftAPITester/SwiftAPITester/PurchasesAPI.swift
```diff
@@ -98,9 +98,13 @@ private func checkTypealiases(
...
-    let purchaseResultData: PurchaseResultData = (transaction: transaction,
-                                                  customerInfo: customerInfo,
-                                                  userCancelled: userCancelled)
+    let purchaseResultData = PurchaseResultData(transaction: transaction,
+                                                customerInfo: customerInfo,
+                                                userCancelled: userCancelled)
+
+    let _: StoreTransaction? = purchaseResultData.transaction
+    let _: CustomerInfo = purchaseResultData.customerInfo
+    let _: Bool = purchaseResultData.userCancelled
```
NachoSoto, line 107: These were missing from API tester.

aboedo:
> The public API remains mostly unchanged, with the only caveat that it's no longer possible to destructure the result tuple

Doesn't that make it a breaking change? Feels like a painful one given that: 
- it would likely look confusing if you don't know the implementation details, since when you look at the type, it seems just about the same
- unpacking the tuple is probably the nicest way to _currently_ use the async API

I'd be more inclined to have a new method + regular old fashioned deprecation tbh

NachoSoto:
> If that's the case I'd be more inclined to have a new method + regular old fashioned deprecation tbh

That's what I started with, but that was extremely painful because we'd need to have duplicate names for all **4** methods because we can't have overloads with different return types.

aboedo:
yeah, that's part of why I originally didn't want to update the purchase methods but just the behavior instead

NachoSoto:
I wonder how common destructuring of the result tuple is. Most people I'd imagine just need `let customerInfo = Purchases.shared.purchase(product).customerInfo` which would still work.

aboedo:
We sadly don't have a way of knowing, though. And it being the one you use to make purchases, it's likely that a lot of people would be using it, we have no guarantees that we won't break someone's build out of the blue with it

NachoSoto:
I did experiment with duplicating the purchase methods but that's really not feasible with the amount of duplication that it would produce, requiring slightly different purchase method names just for a slightly different result type.

NachoSoto:
This approach was too good to be valid. We talked about it offline, and we can't afford the breaking change.

PR #2454, opened 2023-04-26
updated example code for the sample app for Custom Entitlements

aboedo:
Updated example code based on latest usage

### DIFF
--- a/Examples/testCustomEntitlementsComputation/README.md
+++ b/Examples/testCustomEntitlementsComputation/README.md
@@ -85,7 +85,28 @@ After calling this method, you might need to call your backend to refresh entitl
 Call `purchase(package:)` through either the Async / Await or completion blocks alternatives:
 
 ```swift
-let (transaction, customerInfo, userCancelled) = try await Purchases.shared.purchase(package: package)
+do {
+    let (transaction, customerInfo, _) = try await Purchases.shared.purchase(package: package)
+    print(
+        """
+        Purchase finished:
+        Transaction: \(transaction.debugDescription)
+        CustomerInfo: \(customerInfo.debugDescription)
+        """
+    )
+} catch ErrorCode.receiptAlreadyInUseError {
+    print("The receipt is already in use by another subscriber. " +
+          "Log in with the previous account or contact support to get your purchases transferred to " +
+          "regain access")
+} catch ErrorCode.paymentPendingError {
+    print("The purchase is pending and may be completed at a later time." +
+          "This can happen when awaiting parental approval or going through extra authentication flows " +
+          "for credit cards in some countries.")
+} catch ErrorCode.purchaseCancelledError {
+    print("Purchase was cancelled by the user.")
+} catch {
+    print("FAILED TO PURCHASE: \(error.localizedDescription)")
+}
 ```
 
 ```swift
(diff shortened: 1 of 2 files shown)


Examples/testCustomEntitlementsComputation/README.md
```diff
@@ -85,7 +85,28 @@ After calling this method, you might need to call your backend to refresh entitl
...
+          "regain access")
+} catch ErrorCode.paymentPendingError {
+    print("The purchase is pending and may be completed at a later time." +
+          "This can happen when awaiting parental approval or going through extra authentication flows " +
+          "for credit cards in some countries.")
+} catch ErrorCode.purchaseCancelledError {
+    print("Purchase was cancelled by the user.")
+} catch {
+    print("FAILED TO PURCHASE: \(error.localizedDescription)")
+}
```
aboedo, line 109: we still need to add the completion block equivalent, I can tackle it in a bit but if anyone's around feel free to add it

Examples/testCustomEntitlementsComputation/README.md
```diff
@@ -85,12 +85,56 @@ After calling this method, you might need to call your backend to refresh entitl
...
 ```
 
 ```swift
 Purchases.shared.purchase(package: package) { (transaction, customerInfo, error, userCancelled) in
-    // code to handle here
+    if let error = error {
+        if error.code == ErrorCode.receiptAlreadyInUseError.rawValue {
```
NachoSoto, line 115: ⚠️ this is not a great practice, it's assuming the `error.domain`. Instead I recommend following the `Swift migration guide](https://github.com/RevenueCat/purchases-ios/blob/main/Sources/DocCDocumentation/DocCDocumentation.docc/V4_API_Migration_guide.md#error-handling):
```swift
if let error = error as? ErrorCode {
	switch error {
		case .purchaseInvalidError: ...
		default: ...
	}
```

NachoSoto (changes requested):
Looks good, only one comment about the completion block example.

Examples/testCustomEntitlementsComputation/README.md
aboedo: I had a vague recollection about this but didn't know where to look, thanks!

Examples/testCustomEntitlementsComputation/README.md
```diff
@@ -85,12 +85,57 @@ After calling this method, you might need to call your backend to refresh entitl
...
+            print("The purchase is pending and may be completed at a later time." +
+                  "This can happen when awaiting parental approval or going through extra authentication flows " +
+                  "for credit cards in some countries.")
+        case .purchaseCancelledError:
+            print("Purchase was cancelled by the user.")
+        default:
+            print("FAILED TO PURCHASE: \(error.localizedDescription)")
+        }
+        return
+    } else {
```
NachoSoto, line 130: Technically there's a third case where error isn't `Nil` and it's a different type.

NachoSoto (approved):

Examples/testCustomEntitlementsComputation/README.md
aboedo: oh I thought we were translating all errors we output into this format. If not if feels like we're putting a lot of burden onto the developer to know which types we're sending out?

aboedo: updated to consider that case, but we should probably reassess how we're doing these to make sure that we're keeping things as simple as possible for customers

NachoSoto: You're right, since #1879 we ensure that for the most part, but because of untyped throws in Swift we can't know that at compile time in this API.

PR #2455, opened 2023-04-26
Update offerings cache when switchUser(to:) is called

aboedo:
Added updating of offerings cache when `switchUser(to:)` is called. 
Also added checking that the new appUserID != the current one and no-op if it is.

### DIFF
--- a/Sources/Identity/IdentityManager.swift
+++ b/Sources/Identity/IdentityManager.swift
@@ -92,7 +92,7 @@ class IdentityManager: CurrentUserProvider {
     }
 
     func switchUser(to newAppUserID: String) {
-        Logger.debug(Strings.identity.switching_user(newUserId: newAppUserID))
+        Logger.debug(Strings.identity.switching_user(newUserID: newAppUserID))
         self.resetCacheAndSave(newUserID: newAppUserID)
     }
 
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -695,7 +695,18 @@ public extension Purchases {
     ///
     @objc(switchUserToNewAppUserID:)
     func switchUser(to newAppUserID: String) {
+        guard self.identityManager.currentAppUserID != newAppUserID else {
+            self.logger.warn(Strings.identity.switching_user_same_app_user_id(newUserID: newAppUserID))
+            return
+        }
+
         self.identityManager.switchUser(to: newAppUserID)
+
+        self.systemInfo.isApplicationBackgrounded { isBackgrounded in
+            self.offeringsManager.updateOfferingsCache(appUserID: self.appUserID,
+                                                       isAppBackgrounded: isBackgrounded,
+                                                       completion: nil)
+        }
     }
 
 }
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -696,7 +696,7 @@ extension Purchases {
     ///
     @objc(switchUserToNewAppUserID:)
     public func switchUser(to newAppUserID: String) {
-       internalSwitchUser(to: newAppUserID)
+        self.internalSwitchUser(to: newAppUserID)
     }
 #endif
 
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -681,12 +681,13 @@ public extension Purchases {
 
 }
 
-#else
+#endif
 
 // - MARK: - Custom entitlement computation API
 
-public extension Purchases {
+extension Purchases {
 
+#if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
     ///
     /// Updates the current appUserID to a new one, without associating the two.
     /// - Important: This method is **only available** in Custom Entitlements Computation mode.
@@ -694,7 +695,12 @@ public extension Purchases {
     /// with the newAppUserID.
     ///
     @objc(switchUserToNewAppUserID:)
-    func switchUser(to newAppUserID: String) {
+    public func switchUser(to newAppUserID: String) {
+       internalSwitchUser(to: newAppUserID)
+    }
+#endif
+
+    internal func internalSwitchUser(to newAppUserID: String) {
         guard self.identityManager.currentAppUserID != newAppUserID else {
             Logger.warn(Strings.identity.switching_user_same_app_user_id(newUserID: newAppUserID))
             return
@@ -710,8 +716,6 @@ public extension Purchases {
     }
 
 }
-
-#endif
 
 // MARK: Purchasing
 
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -696,7 +696,7 @@ public extension Purchases {
     @objc(switchUserToNewAppUserID:)
     func switchUser(to newAppUserID: String) {
         guard self.identityManager.currentAppUserID != newAppUserID else {
-            self.logger.warn(Strings.identity.switching_user_same_app_user_id(newUserID: newAppUserID))
+            Logger.warn(Strings.identity.switching_user_same_app_user_id(newUserID: newAppUserID))
             return
         }
 
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -687,7 +687,7 @@ public extension Purchases {
 
 extension Purchases {
 
-#if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+#if ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
     ///
     /// Updates the current appUserID to a new one, without associating the two.
     /// - Important: This method is **only available** in Custom Entitlements Computation mode.
--- a/Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
+++ b/Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
@@ -102,7 +102,7 @@ class PurchasesLogInTests: BasePurchasesTests {
 
     #if ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
 
-    func testSwitchUser() {
+    func testSwitchUserSwitchesUser() {
         self.systemInfo = MockSystemInfo(finishTransactions: true, customEntitlementsComputation: true)
         Purchases.clearSingleton()
         self.initializePurchasesInstance(appUserId: "old-test-user-id")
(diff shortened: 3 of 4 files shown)


Sources/Identity/IdentityManager.swift
```diff
@@ -92,7 +92,7 @@ class IdentityManager: CurrentUserProvider {
     }
 
     func switchUser(to newAppUserID: String) {
-        Logger.debug(Strings.identity.switching_user(newUserId: newAppUserID))
+        Logger.debug(Strings.identity.switching_user(newUserID: newAppUserID))
```
aboedo, line 95: renamed param for consistency with others

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -695,7 +695,18 @@ public extension Purchases {
     ///
     @objc(switchUserToNewAppUserID:)
     func switchUser(to newAppUserID: String) {
+        guard self.identityManager.currentAppUserID != newAppUserID else {
+            self.logger.warn(Strings.identity.switching_user_same_app_user_id(newUserID: newAppUserID))
+            return
+        }
```
aboedo, line 701: we missed this in the original implementation but it becomes a lot more important when a cache update is added to the mix

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
```diff
@@ -102,7 +102,7 @@ class PurchasesLogInTests: BasePurchasesTests {
 
     #if ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
```
aboedo, line 104: not sure how to run these locally

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
```diff
@@ -113,6 +113,30 @@ class PurchasesLogInTests: BasePurchasesTests {
...
+    func testRefreshesOfferingsCache() {
+        expect(self.mockOfferingsManager.invokedUpdateOfferingsCacheCount) == 0
+
+        self.systemInfo = MockSystemInfo(finishTransactions: true, customEntitlementsComputation: true)
+        Purchases.clearSingleton()
+        self.initializePurchasesInstance(appUserId: "old-test-user-id")
+
+        self.purchases.switchUser(to: "test-user-id")
+
+        expect(self.mockOfferingsManager.invokedUpdateOfferingsCacheCount).toEventually(equal(1))
```
aboedo, line 125: eventually because we need to check if the app is running in the background and that ironically requires a thread change IIRC

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
NachoSoto: We don't have a way right now :/
Before this, it was simply delegating to identity manager, but now it is worth testing.

We should set up a separate target that runs tests with this flag. Mind creating a Linear for that?

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
```diff
@@ -113,6 +113,30 @@ class PurchasesLogInTests: BasePurchasesTests {
         expect(self.identityManager.invokedSwitchUserParametersList) == ["test-user-id"]
     }
 
+    func testRefreshesOfferingsCache() {
+        expect(self.mockOfferingsManager.invokedUpdateOfferingsCacheCount) == 0
+
+        self.systemInfo = MockSystemInfo(finishTransactions: true, customEntitlementsComputation: true)
```
NachoSoto, line 119: This initialization is duplicated in several tests now, might be worth extracting.

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
NachoSoto: Not with the mock though? Not a big deal anyway.

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
```diff
@@ -113,6 +113,30 @@ class PurchasesLogInTests: BasePurchasesTests {
...
+        Purchases.clearSingleton()
+        self.initializePurchasesInstance(appUserId: "old-test-user-id")
+
+        self.purchases.switchUser(to: "test-user-id")
+
+        expect(self.mockOfferingsManager.invokedUpdateOfferingsCacheCount).toEventually(equal(1))
+    }
+
+    func testSwitchUserNoOpIfAppUserIDIsSameAsCurrent() {
+        self.systemInfo = MockSystemInfo(finishTransactions: true, customEntitlementsComputation: true)
```
NachoSoto, line 129: Ditto duplicated initialization.

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
```diff
@@ -113,6 +113,30 @@ class PurchasesLogInTests: BasePurchasesTests {
...
+        Purchases.clearSingleton()
+        let appUserId = "test-user-id"
+        self.initializePurchasesInstance(appUserId: appUserId)
+
+        self.purchases.switchUser(to: appUserId)
+
+        expect(self.identityManager.invokedSwitchUser) == false
+        expect(self.mockOfferingsManager.invokedUpdateOfferingsCacheCount).toEventually(equal(0))
```
NachoSoto, line 137: This eventually isn't needed, since it'll be 0 to begin with, right?

NachoSoto (approved):

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
aboedo: exactly the point, though: it'll be zero at first, but if there was a bug, it'd happen asynchronously. So it'd always be zero if you check right away, you need to check a little bit later to ensure that it wasn't triggered

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
aboedo: ohh right, maybe? I didn't check because I couldn't run this 😅

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
aboedo: on it

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
NachoSoto: But this `toEventually` will only check once synchronously, pass, and then not check again :)

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
aboedo: https://linear.app/revenuecat/issue/SDK-3090/create-a-new-test-target-to-test-logic-thats-specific-to

I think a good alternative though would be to just move the logic out of this class and either into PurchasesOrchestrator or a new class, then call it from here. That way we keep this class as just the public Facade, which is what it's supposed to be

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
aboedo: oh lol right

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
NachoSoto: Totally 👍🏻 then continue testing this with the runtime flag instead.

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: 👍🏻

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
```diff
@@ -113,6 +113,30 @@ class PurchasesLogInTests: BasePurchasesTests {
...
+    func testSwitchUserNoOpIfAppUserIDIsSameAsCurrent() {
+        self.systemInfo = MockSystemInfo(finishTransactions: true, customEntitlementsComputation: true)
+        Purchases.clearSingleton()
+        let appUserId = "test-user-id"
+        self.initializePurchasesInstance(appUserId: appUserId)
+
+        self.purchases.switchUser(to: appUserId)
+
+        expect(self.identityManager.invokedSwitchUser) == false
+        expect(self.mockOfferingsManager.invokedUpdateOfferingsCacheCount).toNever(equal(1))
```
NachoSoto, line 137: ⚠️ Wait hold on, the default `timeout` is 1 second, I don't think it's worth making tests take an extra second (that's like 50% increase) for this.

NachoSoto: `MockSystemInfo.isApplicationBackgrounded` runs synchronously, so you should be able to test this without any `eventually` or `never`, right?

aboedo:
@NachoSoto I ended up going with a slightly simpler path. And... I think we might wanna do this a bit more tbh, it feels a lot simpler to manage. 
Just have everything in internal methods, use compiler directive to decide whether to include a public version of the same method. 
Thoughts?

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
aboedo: removed now that I was able to test

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
aboedo: updated

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -681,27 +681,42 @@ public extension Purchases {
...
 
-#else
+#endif
 
 // - MARK: - Custom entitlement computation API
 
-public extension Purchases {
+extension Purchases {
 
+#if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
```
NachoSoto, line 690: I think this is backwards?

NachoSoto (approved):
Great idea! I like the simple approach. I think tests should catch that this is backwards though :)

NachoSoto:
Oh they don't yet because APITesters aren't merged.

NachoSoto:
I disabled auto merge just in case.

Sources/Purchasing/Purchases/Purchases.swift
aboedo: LOL I shouldn't be allowed to code

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -681,27 +681,42 @@ public extension Purchases {
...
     ///
     /// Updates the current appUserID to a new one, without associating the two.
     /// - Important: This method is **only available** in Custom Entitlements Computation mode.
     /// Receipts posted by the SDK to the RevenueCat backend after calling this method will be sent
     /// with the newAppUserID.
     ///
     @objc(switchUserToNewAppUserID:)
-    func switchUser(to newAppUserID: String) {
+    public func switchUser(to newAppUserID: String) {
+       internalSwitchUser(to: newAppUserID)
```
NachoSoto, line 699: Nit:
```suggestion
       self.internalSwitchUser(to: newAppUserID)
```

Sources/Purchasing/Purchases/Purchases.swift
aboedo: fixed

Sources/Purchasing/Purchases/Purchases.swift
aboedo: oh, woops, forgot

aboedo:
@NachoSoto thanks for the catches! Updated

NachoSoto (approved):

PR #2457, opened 2023-04-27
Re-enable subs attributes in custom entitlements mode

aboedo:
We're currently assessing whether we need subscriber attributes to be usable in Custom Entitlements Computation mode. 

This PR adds them so that they get sent along with the receipt when posting receipts, but they don't automatically sync when backgrounding / foregrounding the app.


aboedo:
closing since it might not be needed

PR #2458, opened 2023-04-28
Add message indicating tag doesn't exist

vegaro:
There's some confusion when checking if a tag exists since Fastlane prints a message in red if the tag doesn't exist.

Screenshot 2023-04-27 at 5 10 21 PM

I added a message if the tag doesn't exist, so it doesn't look like there's an error

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -673,9 +673,10 @@ end
 def check_no_git_tag_exists(version_number)
   if git_tag_exists(tag: version_number, remote: true, remote_name: 'origin')
     raise "git tag with version #{version_number} already exists!"
+  else
+    UI.message("git tag with version #{version_number} doesn't exist")
   end
 end
-
 
 def check_no_github_release_exists(version_number)
   found_release_number = get_github_release(url: "revenuecat/purchases-ios", version: version_number)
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -673,10 +673,11 @@ end
 def check_no_git_tag_exists(version_number)
   if git_tag_exists(tag: version_number, remote: true, remote_name: 'origin')
     raise "git tag with version #{version_number} already exists!"
+  else
+    UI.message("git tag with version #{version_number} doesn't exist")
```
aboedo, line 677: let's maybe add something like ✅  to the message to make it clearer that it's a good thing, since the message from the command looks ominous

aboedo: alternatively we could replace the fastlane command with a direct git command so there's no shell error code

aboedo (approved):
thanks for taking care of this!!

fastlane/Fastfile
vegaro: hmm I've been trying the alternative, and it looks like Fastlane's Action.sh will always log the non-zero status code. For example, I tried this command

```
Actions.sh("git ls-remote --tags origin #{version_number} | grep -q #{version_number}")
```

Which returns 1 if the tag doesn't exist and 0 if the tag exists and fastlane was still logging 

```
Exit status of command 'git ls-remote --tags origin 4.18.0-SNAPSHOT | grep -q 4.18.0-SNAPSHOT' was 1 instead of 0.
```

Maybe there's a way of preventing the log from showing, but I think I will just go with the checkmark for now ✅

tonidero (approved):

PR #2462, opened 2023-05-01
Add `StoreProduct.pricePerYear`

NachoSoto:


Sources/Purchasing/StoreKitAbstractions/SubscriptionPeriod.swift
```diff
@@ -102,18 +102,38 @@ extension SubscriptionPeriod.Unit: Sendable {}
...
+        case .month: return 1
+        case .year: return 12
+        }
+    }
+
+    private var unitsPerYear: Decimal {
+        switch self.unit {
+        case .day: return 1 / 365
+        case .week: return 1 / 52.14 // Number of weeks in a year
+        case .month: return 1 / 12
```
tonidero, line 127: This was interesting to me... We seem to be doing integer division here, but seems that the swift compiler is smart enough to do the conversion to Decimal automatically 👍

tonidero (approved):

PR #2412, opened 2023-04-17
`CustomerInfoManager`: fixed deadlock caused by reading `CustomerInfo` inside of observer

NachoSoto:
Fixes #2410 and https://github.com/RevenueCat/react-native-purchases/issues/579.

This adds 2 previously failing tests that reproduce this issue: one directly on `CustomerInfoManager` and another one utilizing `PurchasesDelegate`.
These illustrate the issue: if the observer (through `CustomerInfoManager.monitorChanges` or `PurchasesDelegate`) called a method that ended up reading `CustomerInfoManager`'s data inside of the observation, the SDK would deadlock.

This is because we were invoking the observers while inside the lock. This was unnecessary, so this fixes the issue by ensuring that the observers receive updates after the lock is released, using a new `OperationDispatcher.dispatchAsyncOnMainThread`.

PR #2463, opened 2023-05-01
`Integration Tests`: add `MainThreadMonitor` to ensure main thread is not blocked

NachoSoto:
We've had a few reports of deadlocks (#2412, #2375). This might have not detected them, but it would detect future issues, as well as busy operations running on the main thread.

Example:
Screenshot 2023-05-01 at 9 50 22 AM

### DIFF
--- a/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
@@ -34,6 +34,8 @@ class BaseBackendIntegrationTests: XCTestCase {
     // swiftlint:disable:next weak_delegate
     private(set) var purchasesDelegate: TestPurchaseDelegate!
 
+    private var mainThreadMonitor: MainThreadMonitor!
+
     class var storeKit2Setting: StoreKit2Setting { return .default }
     class var observerMode: Bool { return false }
     class var responseVerificationMode: Signing.ResponseVerificationMode {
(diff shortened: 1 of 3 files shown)


Tests/BackendIntegrationTests/MainThreadMonitor.swift
```diff
@@ -0,0 +1,52 @@
...
+            while self != nil {
+                let semaphore = DispatchSemaphore(value: 0)
+                DispatchQueue.main.async {
+                    semaphore.signal()
+                }
+
+                let deadline = DispatchTime.now() + Self.threshold
+                let result = semaphore.wait(timeout: deadline)
+
+                precondition(
```
NachoSoto, line 42: I couldn't make this a `XCTFail` because it wouldn't be reported if the main thread is completely blocked.

Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
```diff
@@ -34,6 +34,8 @@ class BaseBackendIntegrationTests: XCTestCase {
     // swiftlint:disable:next weak_delegate
     private(set) var purchasesDelegate: TestPurchaseDelegate!
 
+    private var mainThreadMonitor: MainThreadMonitor!
```
tonidero, line 37: This is a cool idea. Any thoughts on using this in other integration tests (not only backend integration tests)?

tonidero (approved):
Good idea!

Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
NachoSoto: This is the base class for all integration tests.

PR #2421, opened 2023-04-19
`CircleCI`: change all jobs to use `Xcode 14.x` and replace `xcode-install` with `xcodes`

NachoSoto:
`CircleCI` is deprecating `Xcode 13` soon.

[`CircleCI` still recommends using `xcode-install`](https://support.circleci.com/hc/en-us/articles/360011880714-How-do-I-add-different-simulators-to-my-iOS-builds-), but [it breaks with Xcode 14](https://github.com/xcpretty/xcode-install/issues/466), so [it's been deprecated now](https://github.com/xcpretty/xcode-install/blob/master/MIGRATION.md#known-limitations). This replaces the runtime installation with [`xcodes`](https://github.com/RobotsAndPencils/xcodes).

PR #2485, opened 2023-05-15
`SK2StoreProduct`: simplify `currencyCode` extraction

NachoSoto:
I noticed this new property thanks to [this tweet](https://twitter.com/ariskox/status/1658067009609904129?s=61&t=Zz639sMJdwHISm4AKYXoCw).

This changes the "future" implementation of `SK2StoreProduct.currencyCode` to use [the new (backdeployed) property](https://developer.apple.com/documentation/storekit/product/4044347-priceformatstyle).
We still need the existing implementation because `@_backDeploy` won't work if compiling with `Xcode 13.x`.

### DIFF
--- a/Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
@@ -94,15 +94,17 @@ internal struct SK2StoreProduct: StoreProductType {
 private extension SK2StoreProduct {
 
     var _currencyCode: String? {
-        if #available(iOS 16.0, tvOS 16.0, watchOS 9.0, macOS 13.0, *) {
+        #if swift(>=5.7)
+        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *) {
             return self.currencyCodeFromPriceFormat
-        } else {
-            // note: if we ever need more information from the jsonRepresentation object, we
-            // should use Codable or another decoding method to clean up this code.
-            let attributes = jsonDict["attributes"] as? [String: Any]
-            let offers = attributes?["offers"] as? [[String: Any]]
-            return offers?.first?["currencyCode"] as? String
         }
+        #endif
+
+        // note: if we ever need more information from the jsonRepresentation object, we
+        // should use Codable or another decoding method to clean up this code.
+        let attributes = jsonDict["attributes"] as? [String: Any]
+        let offers = attributes?["offers"] as? [[String: Any]]
+        return offers?.first?["currencyCode"] as? String
     }
 
     private var jsonDict: [String: Any] {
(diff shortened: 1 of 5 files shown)


NachoSoto:
We should also look into exposing the [`Decimal.FormatStyle.Currency`](https://developer.apple.com/documentation/foundation/decimal/formatstyle/currency)

Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
```diff
@@ -100,6 +85,33 @@ internal struct SK2StoreProduct: StoreProductType {
 
 }
 
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+private extension SK2StoreProduct {
+
+    var _currencyCode: String {
+        if #available(iOS 16.0, tvOS 16.0, watchOS 9.0, macOS 13.0, *) {
+            // This is marked as `@_backDeploy`, but it's not actually working before iOS 16.0
```
aboedo, line 93: lol why am I not surprised

aboedo: do we have a radar for this?

Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift
```diff
@@ -84,7 +84,7 @@ public typealias SK2Product = StoreKit.Product
     @available(iOS 12.0, macCatalyst 13.0, tvOS 12.0, macOS 10.14, watchOS 6.2, *)
     @objc public var subscriptionGroupIdentifier: String? { self.product.subscriptionGroupIdentifier}
 
-    @objc public var priceFormatter: NumberFormatter? { self.product.priceFormatter }
+    @objc public var priceFormatter: NumberFormatter { self.product.priceFormatter }
```
aboedo, line 87: does this make it a breaking change?

Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
NachoSoto: Actually I figured out the reason: this only works in whatever version of the SDK added this, but you can't use if it you're compiling with Xcode 13.x obviously... but the docs don't even say when it was added.

Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift
NachoSoto: Yeah which is why I made it a `feat`. But I had to revert this to fix Xcode 13.x compilation 🤷🏻‍♂️ 
Won't matter in the long term since you can't submit apps with that anymore, but we need to keep support for it in CI until I finish #2421.

NachoSoto:
This is ready now.

aboedo (approved):

PR #2486, opened 2023-05-15
`DangerousSettings`: debug-only `forceServerErrors`

NachoSoto:
Follow up to #2370. This removes `forceServerErrors` from release builds so it can't be accidentally used (even though `HTTPClient` was only checking it on `DEBUG` already).

Also allows changing the flag at runtime, so future integration tests can check the behavior when the server changes status throughout the lifetime of the app.

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -61,7 +61,7 @@ class HTTPClient {
         completionHandler: Completion<Value>?
     ) {
         #if DEBUG
-        guard !self.systemInfo.dangerousSettings.internalSettings.forceServerErrors else {
+        guard !self.systemInfo.forceServerErrors else {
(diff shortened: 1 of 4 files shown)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -61,7 +61,7 @@ class HTTPClient {
         completionHandler: Completion<Value>?
     ) {
         #if DEBUG
-        guard !self.systemInfo.dangerousSettings.internalSettings.forceServerErrors else {
+        guard !self.systemInfo.forceServerErrors else {
```
NachoSoto, line 64: `InternalSettings` aren't mutable, but stubbing this in `MockSystemInfo` will allow us to "change" the flag at runtime in tests.

tonidero (approved):
LGTM!

PR #2487, opened 2023-05-15
`HTTPClient`: log actual response status code

NachoSoto:
We were currently logging the status code retrieved from `ETagManager`, which made it seem like etags weren't working.

Example:
> API request completed: GET /v1/subscribers/identify (200)

With this change, we now log the actual response status code so we can verify the server is returning the correct status code:
> API request completed: GET /v1/subscribers/identify (304)

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -309,7 +309,10 @@ private extension HTTPClient {
         if let response = response {
             switch response {
             case let .success(response):
-                Logger.debug(Strings.network.api_request_completed(request.httpRequest, httpCode: response.statusCode))
+                Logger.debug(Strings.network.api_request_completed(
+                    request.httpRequest,
+                    httpCode: urlResponse?.httpStatusCode ?? response.statusCode
+                ))
             case let .failure(error):
                 Logger.debug(Strings.network.api_request_failed(request.httpRequest, error: error))
             }
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -311,6 +311,8 @@ private extension HTTPClient {
             case let .success(response):
                 Logger.debug(Strings.network.api_request_completed(
                     request.httpRequest,
+                    // Getting status code from the original response to detect 304s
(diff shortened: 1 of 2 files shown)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -309,7 +309,10 @@ private extension HTTPClient {
         if let response = response {
             switch response {
             case let .success(response):
-                Logger.debug(Strings.network.api_request_completed(request.httpRequest, httpCode: response.statusCode))
+                Logger.debug(Strings.network.api_request_completed(
+                    request.httpRequest,
+                    httpCode: urlResponse?.httpStatusCode ?? response.statusCode
```
aboedo, line 314: let's add a comment, it's not obvious at a glance why we're checking them in this order

aboedo: or make it a method with a clear name or something

Sources/Networking/HTTPClient/HTTPStatusCode.swift
```diff
@@ -93,3 +93,13 @@ extension HTTPStatusCode {
     }
 
 }
+
+extension URLResponse {
+
+    var httpStatusCode: HTTPStatusCode? {
+        guard let response = self as? HTTPURLResponse else { return nil }
+
+        return .init(rawValue: response.statusCode)
```
aboedo, line 102: [nitpick, FFTI] it might just be me, but I find this pattern in general to be less readable than 

```swift
    return HTTPStatusCode(rawValue: response.statusCode)
```

since I have to do a double-take and check what the return type of the method is to figure out what we're creating

aboedo (approved):

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto: Yup good point 👍🏻

Sources/Networking/HTTPClient/HTTPStatusCode.swift
NachoSoto: Yeah it's definitely a tradeoff, especially for more complex functions where the context of the type might be lost to the reader. In this particular case my thinking is just "let the compiler figure it out".

PR #2488, opened 2023-05-15
`CarthageInstallation`: removed workaround

NachoSoto:
That was only required for Xcode 12, which is no longer supported.

Also added `--verbose` parameter to the job to avoid timeouts from slow builds (see https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/11009/workflows/9a750760-2a24-4f0e-adeb-b464bfc8d25e/jobs/66974)


NachoSoto:
This is required for #2483.

aboedo (approved):
rest in peace, horrible workaround

PR #2489, opened 2023-05-15
`CocoapodsInstallation`: fixed `Xcode 14.3.0` issue

NachoSoto:
See https://github.com/CocoaPods/CocoaPods/issues/11808. That's fixed in CocoaPods 1.12.1, but that's not included in [CircleCI's 14.3.0 image](https://circle-macos-docs.s3.amazonaws.com/image-manifest/v11629/manifest.txt).

This changes it to use the `CocoaPods` from `Bundle` to use `1.12.1`.

Fixes https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/11009/workflows/9a750760-2a24-4f0e-adeb-b464bfc8d25e/jobs/66978


NachoSoto:
This is required for #2483.

aboedo (approved):
I think we can also force a cocoapods version by including it in the bundle and then calling it through `bundle exec`

NachoSoto:
Oh good point, I see that `Gemfile.lock` has the correct `1.12.1`. Let me try that instead.

.circleci/config.yml
```diff
@@ -604,7 +604,7 @@ jobs:
           name: Install Pods
           working_directory: Tests/InstallationTests/CocoapodsInstallation/
           command: |
-              pod install
+              bundle exec pod install
```
NachoSoto, line 607: @aboedo this should work.

PR #2490, opened 2023-05-15
`ETagManager`: don't send validation time if not present

NachoSoto:
This is not desirable behavior: it will lead to the backend assuming that the contents are very fresh, when they might be super old. This will in turn lead to bad prioritization of requests. 

If we don't have a validation date for an eTag, we should just not send it.


NachoSoto:
This is required for #2483.

vegaro (approved):

tonidero (approved):

PR #2495, opened 2023-05-16
`OfferingsManager`: return `Offerings` from new disk cache when server is down

NachoSoto:
This new cache is used only when the server is down, so paywalls can continue to work in that case.

### Example:
```
- DEBUG: ℹ️ Configuring SDK using RevenueCat's UserDefaults suite.
- DEBUG: ℹ️ No cached Offerings, fetching from network
- DEBUG: ℹ️ Offerings cache is stale, updating from network in foreground
- WARN: ⚠️ Error fetching offerings: server appears down
- DEBUG: ℹ️ MockProductsManager: no stubbed products, returning fake products for ["monthly_freetrial"]
- DEBUG: ℹ️ Vending Offerings from disk cache
```

### DIFF
--- a/Sources/Caching/DeviceCache.swift
+++ b/Sources/Caching/DeviceCache.swift
@@ -67,6 +67,7 @@ class DeviceCache {
 
             // Clear offerings cache.
             self.offeringsCachedObject.clearCache()
+            userDefaults.removeObject(forKey: CacheKey.offerings(oldAppUserID))
 
             // Delete attributes if synced for the old app user id.
             if Self.unsyncedAttributesByKey(userDefaults, appUserID: oldAppUserID).isEmpty {
@@ -113,10 +114,6 @@ class DeviceCache {
         }
     }
 
-    func clearCachedOfferings() {
-        self.offeringsCachedObject.clearCache()
-    }
-
     func clearCustomerInfoCacheTimestamp(appUserID: String) {
         self.userDefaults.write {
             Self.clearCustomerInfoCacheTimestamp($0, appUserID: appUserID)
@@ -142,25 +139,48 @@ class DeviceCache {
         }
     }
 
-    // MARK: - offerings
+    // MARK: - Offerings
 
-    func cache(offerings: Offerings) {
-        offeringsCachedObject.cache(instance: offerings)
+    func cachedOfferingsData(appUserID: String) -> Data? {
+        return self.userDefaults.read {
+            $0.data(forKey: CacheKey.offerings(appUserID))
+        }
+    }
+
+    func cache(offerings: Offerings, appUserID: String) {
+        self.cacheInMemory(offerings: offerings)
+
+        if let jsonData = try? JSONEncoder.default.encode(value: offerings.response, logErrors: true) {
+            self.userDefaults.write {
+                $0.set(jsonData, forKey: CacheKey.offerings(appUserID))
+            }
+        }
+    }
+
+    func cacheInMemory(offerings: Offerings) {
+        self.offeringsCachedObject.cache(instance: offerings)
+    }
+
+    func clearOfferingsCache(appUserID: String) {
+        self.offeringsCachedObject.clearCache()
+        self.userDefaults.write {
+            $0.removeObject(forKey: CacheKey.offerings(appUserID))
+        }
     }
 
     func isOfferingsCacheStale(isAppBackgrounded: Bool) -> Bool {
-        return offeringsCachedObject.isCacheStale(
+        return self.offeringsCachedObject.isCacheStale(
             durationInSeconds: self.cacheDurationInSeconds(isAppBackgrounded: isAppBackgrounded,
                                                            isSandbox: self.sandboxEnvironmentDetector.isSandbox)
         )
     }
 
     func clearOfferingsCacheTimestamp() {
-        offeringsCachedObject.clearCacheTimestamp()
+        self.offeringsCachedObject.clearCacheTimestamp()
     }
 
     func setOfferingsCacheTimestampToNow() {
-        offeringsCachedObject.updateCacheTimestamp(date: Date())
+        self.offeringsCachedObject.updateCacheTimestamp(date: Date())
     }
 
     // MARK: - subscriber attributes
@@ -330,6 +350,7 @@ class DeviceCache {
 
         case customerInfo(String)
         case customerInfoLastUpdated(String)
+        case offerings(String)
         case legacySubscriberAttributes(String)
         case attributionDataDefaults(String)
 
@@ -337,6 +358,7 @@ class DeviceCache {
             switch self {
             case let .customerInfo(userID): return "\(Self.base)purchaserInfo.\(userID)"
             case let .customerInfoLastUpdated(userID): return "\(Self.base)purchaserInfoLastUpdated.\(userID)"
+            case let .offerings(userID): return "\(Self.base)offerings.\(userID)"
             case let .legacySubscriberAttributes(userID): return "\(Self.legacySubscriberAttributesBase)\(userID)"
             case let .attributionDataDefaults(userID): return "\(Self.base)attribution.\(userID)"
             }
@@ -547,7 +569,7 @@ fileprivate extension UserDefaults {
             return nil
         }
 
-        return try? JSONDecoder.default.decode(jsonData: data)
+        return try? JSONDecoder.default.decode(jsonData: data, logErrors: true)
     }
 
     func set(_ value: Any?, forKey key: CacheKeyType) {
--- a/Sources/Caching/DeviceCache.swift
+++ b/Sources/Caching/DeviceCache.swift
@@ -141,7 +141,7 @@ class DeviceCache {
 
     // MARK: - Offerings
 
-    func cachedOfferingsData(appUserID: String) -> Data? {
+    func cachedOfferingsResponseData(appUserID: String) -> Data? {
         return self.userDefaults.read {
             $0.data(forKey: CacheKey.offerings(appUserID))
         }
--- a/Sources/Purchasing/Offerings.swift
+++ b/Sources/Purchasing/Offerings.swift
@@ -45,14 +45,32 @@ import Foundation
         return all[currentOfferingID]
     }
 
+    internal let response: OfferingsResponse
+
     private let currentOfferingID: String?
+
+    init(
+        offerings: [String: Offering],
+        currentOfferingID: String?,
+        response: OfferingsResponse
+    ) {
+        self.all = offerings
+        self.currentOfferingID = currentOfferingID
+        self.response = response
+    }
+
+}
+
+extension Offerings: Sendable {}
+
+public extension Offerings {
 
     /**
      Retrieves a specific offering by its identifier, use this to access additional offerings configured in the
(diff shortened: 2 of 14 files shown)


Sources/Caching/DeviceCache.swift
```diff
@@ -547,7 +570,7 @@ fileprivate extension UserDefaults {
             return nil
         }
 
-        return try? JSONDecoder.default.decode(jsonData: data)
+        return try? JSONDecoder.default.decode(jsonData: data, logErrors: true)
```
NachoSoto, line 573: This is just to be explicit, so reader doesn't wonder why we're ignoring errors with `try?`

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -80,9 +81,24 @@ class OfferingsManager {
             switch result {
             case let .success(response):
                 self.handleOfferingsBackendResult(with: response,
+                                                  appUserID: appUserID,
                                                   fetchPolicy: fetchPolicy,
                                                   completion: completion)
 
+            case let .failure(.networkError(networkError)) where networkError.isServerDown:
```
NachoSoto, line 88: ℹ️  This is the meat of this PR: special case for these errors: instead of forwarding the error, we try to load from disk cache. If that still fails, we return the error anyway.

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -111,25 +127,53 @@ class OfferingsManager {
...
+                    completion(offerings)
+
+                case .failure:
+                    completion(nil)
+                }
+            }
+        )
+    }
+
+    func createOfferings(
```
NachoSoto, line 159: Extracted this to be used from both `fetchCachedOfferingsFromDisk` and `handleOfferingsBackendResult`.

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -111,25 +127,53 @@ class OfferingsManager {
...
+
+        self.createOfferings(
+            from: response,
+            fetchPolicy: fetchPolicy,
+            completion: { [cache = self.deviceCache] result in
+                switch result {
+                case let .success(offerings):
+                    Logger.debug(Strings.offering.vending_offerings_cache_from_disk)
+
+                    cache.cacheInMemory(offerings: offerings)
```
NachoSoto, line 149: Important to cache in memory in this case too I think. Covered by a test.

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -80,9 +81,24 @@ class OfferingsManager {
...
+            case let .failure(.networkError(networkError)) where networkError.isServerDown:
+                Logger.warn(Strings.offering.fetching_offerings_failed_server_down)
+
+                // If unable to fetch offerings when server is down, attempt to load them from disk cache.
+                self.fetchCachedOfferingsFromDisk(appUserID: appUserID,
+                                                  fetchPolicy: fetchPolicy) { offerings in
+                    if let offerings = offerings {
+                        self.dispatchCompletionOnMainThreadIfPossible(completion, value: .success(offerings))
+                    } else {
+                        self.handleOfferingsUpdateError(.backendError(.networkError(networkError)),
```
NachoSoto, line 97: If it fails or we didn't have anything in the cache, we forward the original error.

Tests/UnitTests/Mocks/MockProductsManager.swift
```diff
@@ -31,6 +31,8 @@ class MockProductsManager: ProductsManager {
                 completion(result)
             }
         } else {
+            Logger.debug("\(type(of: self)): no stubbed products, returning fake products for \(identifiers)")
```
NachoSoto, line 34: This was confusing me, I had no idea where the products were coming from in a new test. It should help make it clear they're default stubs.

Tests/UnitTests/Networking/Responses/OfferingsDecodingTests.swift
```diff
@@ -60,4 +60,8 @@ class OfferingsDecodingTests: BaseHTTPResponseTest {
         expect(package.platformProductIdentifier) == "com.revenuecat.other_product"
     }
 
+    func testEncoding() throws {
+        expect(try self.response.encodeAndDecode()) == self.response
```
NachoSoto, line 64: Love how easy this is 😍

Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
```diff
@@ -357,24 +357,34 @@ extension OfferingsManagerTests {
...
+            self.offeringsManager.offerings(appUserID: MockData.anyAppUserID) { result in
+                completed(result)
             }
         }
 
         // then
         expect(self.mockOfferings.invokedGetOfferingsForAppUserIDCount) == expectedCallCount
         expect(self.mockDeviceCache.cacheOfferingsCount) == expectedCallCount
         expect(self.mockOfferings.invokedGetOfferingsForAppUserIDParameters?.randomDelay) == true
+        expect(result).to(beSuccess())
```
NachoSoto, line 378: This test wasn't verifying the output.

Sources/Caching/DeviceCache.swift
```diff
@@ -142,25 +139,48 @@ class DeviceCache {
...
+        }
+    }
+
+    func cache(offerings: Offerings, appUserID: String) {
+        self.cacheInMemory(offerings: offerings)
+
+        if let jsonData = try? JSONEncoder.default.encode(value: offerings.response, logErrors: true) {
+            self.userDefaults.write {
+                $0.set(jsonData, forKey: CacheKey.offerings(appUserID))
+            }
```
tonidero, line 156: I mentioned it in Linear, not a fan of duplicating the information. We should be careful that the memory and the disk caches don't become out of sync...

But probably not a reason to block the PR either.

Sources/Caching/DeviceCache.swift
```diff
@@ -142,25 +139,48 @@ class DeviceCache {
         }
     }
 
-    // MARK: - offerings
+    // MARK: - Offerings
 
-    func cache(offerings: Offerings) {
-        offeringsCachedObject.cache(instance: offerings)
+    func cachedOfferingsData(appUserID: String) -> Data? {
```
tonidero, line 144: I think these function names can be confusing... As in, there is nothing there that tells us if it's using the memory or the disk cache and you need to look at the implementation.

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -42,9 +42,10 @@ class OfferingsManager {
         fetchPolicy: FetchPolicy = .default,
         completion: (@MainActor @Sendable (Result<Offerings, Error>) -> Void)?
     ) {
-        guard let cachedOfferings = self.deviceCache.cachedOfferings else {
+        guard let memoryCachedOfferings = self.deviceCache.cachedOfferings else {
```
tonidero, line 45: Maybe rename DeviceCache's `cachedOfferings` to `memoryCachedOfferings`?

Sources/Purchasing/OfferingsManager.swift
tonidero: Mentioned in Linear but I wonder if we should just use the cache (if available) in all errors... While I agree that users won't be able to purchase anyway if there is no connection, I think we can still help showing the paywall with the cached data so the issue is delayed... I think this alternative is also acceptable though.

Sources/Purchasing/Offerings.swift
```diff
@@ -45,14 +45,32 @@ import Foundation
         return all[currentOfferingID]
     }
 
+    internal let response: OfferingsResponse
```
tonidero, line 48: Just curious, seems like we are storing the response to be able to store it in UserDefaults because `Offerings` isn't `Codeable` correct? I wonder if that's something we could add to avoid having to parse it again later? In any case, nothing change here.

Sources/Purchasing/OfferingsManager.swift
tonidero: This makes sense but we should keep the offerings in memory cache as "stale" in this case so we retry getting the data from the backend as soon as possible.

tonidero (commented):
I think this makes sense! My main concern is that some methods in `DeviceCache` are now not clear on whether they are using a memory or disk cache... Also, I think having 2 caches for the same data is confusing... But we can talk about it more!

tonidero (commented):
I think this makes sense! My main concern is that some methods in `DeviceCache` are now not clear on whether they are using a memory or disk cache... Also, I think having 2 caches for the same data is confusing... But we can talk about it more!

Sources/Purchasing/Offerings.swift
NachoSoto: We can't make `Offerings` `Codable` because it has the full `StoreProduct`s inside, so we need to load those from the `OfferingsResponse` data.

Sources/Caching/DeviceCache.swift
NachoSoto: Good call, will improve.

Sources/Caching/DeviceCache.swift
tonidero: Ahh yeah makes sense, I missed that. We will have the same situation in Android actually. So yeah I think this approach makes sense.

One thing we could do to improve this is have an "offerings cache" and a "offerings http response cache" to try to differentiate them more?

Sources/Purchasing/OfferingsManager.swift
tonidero: I still think we might want to keep the in memory offerings as "stale" in this case, so we retry fetching as soon as possible. Currently this will store the offerings in memory and keep them there for the full duration of the cache. Thoughts?

Sources/Purchasing/OfferingsManager.swift
tonidero: Bump to this comment.

tonidero (approved):
Just a couple of comments but nothing blocking. Looks good!

tonidero (approved):
Just a couple of comments but nothing blocking. Looks good!

Sources/Purchasing/OfferingsManager.swift
NachoSoto: Oh sorry I missed it.

I'm not sure about _all_ errors though. What about "offerings are empty"?
If you're working on your app, and break offerings, it might be better to show "offerings are empty" rather than showing the previously cached ones, so you know a fresh install won't show them.
Maybe we can do it for all network errors only? /cc @aboedo

Sources/Purchasing/OfferingsManager.swift
NachoSoto: Oh great point!

NachoSoto: 26b79868841546137d0da5ad897507530ddbe1d8

Sources/Purchasing/OfferingsManager.swift
NachoSoto: Merging this for now, since it solves the issue of being able to load it when the server is down.
Happy to extend to other errors in a separate PR.

aboedo: yeah, Offerings is conceptually different than CustomerInfo for this. I think only serving from cache on 5xx is the safest approach in that it's also the closest to current behavior

PR #2496, opened 2023-05-16
Performance: new check to ensure serialization / deserialization is done from background thread

NachoSoto:


aboedo (approved):

NachoSoto:
Cross-posting from Linear:

Looks like the only time we decode `CustomerInfo` in the main thread is when we initialize the SDK and we load it from cache. We could change that, but it would slightly change the behavior since we wouldn't be sending the first `CustomerInfo` to the delegate until after the next iteration of the run loop, which **might potentially be breaking for some apps**.

@aboedo thoughts on leaving that behavior?

NachoSoto:
Merging this for now. We can talk about the decoding change on Linear.

PR #2497, opened 2023-05-17
Tests: grouped all `Matcher`s into one file

NachoSoto:
Just a small refactor. I needed some of these matchers in a new test, and realized we could reuse them in other places too.


Tests/UnitTests/TestHelpers/Matchers.swift
tonidero, line 1: FFTI. I'm not sure if we will plan to add other matchers here. If not, we could rename this to be `DateMatchers` to be more specific?

tonidero (approved):

Tests/UnitTests/TestHelpers/Matchers.swift
NachoSoto: Actually just noticed we have an `ErrorMatcher`s file too, so I grouped them all here :)

PR #2498, opened 2023-05-17
`Offering` metadata

joshdholtz:
### Motivation

Added `metadata` to `Offering` 😊 

### Description

- Added `metadata` to `Offering`
  - Had to use a private internal `Metadata` struct to make Swift compiler happy
- Updated unit tests
- Updated backend integration tests
- Updated load shedder tests


Tests/BackendIntegrationTests/LoadShedderIntegrationTests.swift
```diff
@@ -39,6 +39,7 @@ class LoadShedderStoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
     }
 
     func testCanGetOfferings() async throws {
+        // JOSH HERE - add one for meta data
```
NachoSoto, line 42: Actually won't need to anymore after #2499 :) simply run the test and update the snapshot 🎉

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -38,6 +38,7 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
     }
 
     func testCanGetOfferings() async throws {
+        // JOSH HERE - add one for meta data
```
NachoSoto, line 41: Ditto.

Tests/BackendIntegrationTests/Constants.swift
```diff
@@ -14,7 +14,7 @@ enum Constants {
     static let loadShedderApiKey = "REVENUECAT_LOAD_SHEDDER_API_KEY"
 
     // Server URL for the tests. If set to empty string, we'll use the default URL.
-    static let proxyURL = "REVENUECAT_PROXY_URL"
+    static let proxyURL = ""
```
NachoSoto, line 17: Just a reminder to make sure we revert this.

Sources/Purchasing/Offering.swift
```diff
@@ -14,6 +14,12 @@
 
 import Foundation
 
+extension Offering {
+    struct Metadata {
+        let data: [String: Any]
+    }
+}
+
```
NachoSoto, line 22: Nit: our style guide requires spaces here:
```suggestion
extension Offering {

    struct Metadata {
        let data: [String: Any]
    }
    
}

```

Tests/UnitTests/Networking/Responses/Fixtures/OfferingsWithMetadata.json
```diff
@@ -0,0 +1,49 @@
...
+            "identifier": "alternate",
+            "metadata": {
+                "int": 5,
+                "double": 5.5,
+                "boolean": true,
+                "string": "five",
+                "array": ["five"],
+                "dictionary": {
+                    "string": "five"
+                }
```
NachoSoto, line 39: I love that you added all types.

joshdholtz: Yeah yeah, I wrote these for future proofing so we (hopefully) don't need to make any model changes 🤷‍♂️

Sources/Purchasing/Offering.swift
joshdholtz: Oh yeah, I just didn't get to fixing this yet because BC5 work 😛  But will fix soooooooon!

Tests/BackendIntegrationTests/__Snapshots__/StoreKitIntegrationTests/testCanGetOfferings.1.json
```diff
@@ -4,6 +4,9 @@
     {
       "description" : "standard set of packages",
       "identifier" : "default",
+      "metadata" : {
```
NachoSoto, line 7: Amazing

Tests/UnitTests/Networking/Responses/OfferingsDecodingTests.swift
```diff
@@ -48,10 +49,48 @@ class OfferingsDecodingTests: BaseHTTPResponseTest {
     }
 
     func testDecodesSecondOffering() throws {
-        let offering = try XCTUnwrap(self.response.offerings.last)
+        let offering = try XCTUnwrap(self.response.offerings[1])
```
NachoSoto, line 52: This will make the test crash.  I'll help you on Monday, we'll need to define a new method that returns an Optional for an index.

Tests/UnitTests/Networking/Responses/OfferingsDecodingTests.swift
```diff
@@ -48,10 +49,48 @@ class OfferingsDecodingTests: BaseHTTPResponseTest {
...
+        expect(offering.packages).to(haveCount(1))
+
+        let package = offering.packages[0]
+
+        expect(package.identifier) == PackageType.lifetime.description
+        expect(package.platformProductIdentifier) == "com.revenuecat.other_product"
+    }
+
+    func testDecodesMetadataOffering() throws {
+        let offering = try XCTUnwrap(self.response.offerings[2])
```
NachoSoto, line 66: Ditto

Tests/UnitTests/Networking/Responses/OfferingsDecodingTests.swift
```diff
@@ -48,10 +49,48 @@ class OfferingsDecodingTests: BaseHTTPResponseTest {
...
     func testDecodesSecondOffering() throws {
-        let offering = try XCTUnwrap(self.response.offerings.last)
+        let offering = try XCTUnwrap(self.response.offerings[1])
 
         expect(offering.identifier) == "alternate"
         expect(offering.description) == "alternate offering"
+        expect(offering.metadata) == [:]
+        expect(offering.packages).to(haveCount(1))
+
+        let package = offering.packages[0]
```
NachoSoto, line 59: To avoid the test crashing (instead of failing):
```suggestion
        let package = try XCTUnwrap(offering.packages.onlyElement)
```

Tests/UnitTests/Networking/Responses/OfferingsDecodingTests.swift
```diff
@@ -48,10 +49,48 @@ class OfferingsDecodingTests: BaseHTTPResponseTest {
...
+            "boolean": true,
+            "string": "five",
+            "array": ["five"],
+            "dictionary": [
+                "string": "five"
+            ]
+        ]
+        expect(offering.packages).to(haveCount(1))
+
+        let package = offering.packages[0]
```
NachoSoto, line 82: Ditto:
```suggestion
        try XCTUnwrap(offering.packages.onlyElement)
```

Tests/UnitTests/Networking/Responses/OfferingsDecodingTests.swift
```diff
@@ -48,10 +49,48 @@ class OfferingsDecodingTests: BaseHTTPResponseTest {
...
+
+        expect(package.identifier) == PackageType.lifetime.description
+        expect(package.platformProductIdentifier) == "com.revenuecat.other_product"
+    }
+
+    func testDecodesNullMetadataOffering() throws {
+        let offering = try XCTUnwrap(self.response.offerings[3])
```
NachoSoto, line 89: Ditto

NachoSoto (commented):
I think this is ready 👍🏻  Just need to fix those potential crashes in the test, I'll send a PR on Monday with a new method on `Collection` so we can do that.

joshdholtz:
@NachoSoto I think I fixed all the safety things that could crash the tests! But let me know if you want me to do it a different way or move code around 😇

Tests/UnitTests/Networking/Responses/OfferingsDecodingTests.swift
```diff
@@ -65,3 +104,12 @@ class OfferingsDecodingTests: BaseHTTPResponseTest {
     }
 
 }
+
+private extension Collection {
+
+    /// Returns the element at the specified index if it exists, otherwise nil.
+    subscript (safe index: Index) -> Element? {
```
NachoSoto, line 111: Cool this is exactly what I had in mind 👍🏻 
I can extract this an add unit tests for it in a separate PR since I think it's useful to have.

Tests/UnitTests/Networking/Responses/OfferingsDecodingTests.swift
```diff
@@ -48,13 +49,51 @@ class OfferingsDecodingTests: BaseHTTPResponseTest {
...
+
+    func testDecodesNullMetadataOffering() throws {
+        let offering = try XCTUnwrap(self.response.offerings[safe: 3])
+
+        expect(offering.identifier) == "nullmetadata"
+        expect(offering.description) == "offering with null metadata"
+        expect(offering.metadata) == [:]
+        expect(offering.packages).to(haveCount(1))
+
+        let package = try XCTUnwrap(offering.packages[safe: 0])
```
NachoSoto, line 96: I'd use `first` here which is the same:
```suggestion
        let package = try XCTUnwrap(offering.packages.first)
```

Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
```diff
@@ -495,6 +495,7 @@ private extension OfferingsManagerTests {
                     Offering(
                         identifier: offering.identifier,
                         serverDescription: offering.description,
+                        metadata: [:],
```
NachoSoto, line 498: The other tests only check `OfferingsResponse`. Let's add some metadata here so you can check in another test in this file that they get copied correctly to `Offering` (basically that`offering.metadata.mapValues(\.asAny)`).

NachoSoto (commented):
This is ready :) just one more suggested test.

Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
```diff
@@ -465,14 +466,27 @@ private extension OfferingsManagerTests {
...
         static let anyBackendOfferingsResponse: OfferingsResponse = .init(
             currentOfferingId: "base",
             offerings: [
                 .init(identifier: "base",
                       description: "This is the base offering",
                       packages: [
                         .init(identifier: "$rc_monthly", platformProductIdentifier: "monthly_freetrial")
-                      ])
+                      ],
+                      metadata: _metadata)
```
NachoSoto, line 489: Nit per our style guide:
```suggestion
                      metadata: Self.metadata)
```

Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
```diff
@@ -465,14 +466,27 @@ private extension OfferingsManagerTests {
     enum MockData {
         static let anyAppUserID = ""
 
+        @DefaultDecodable.EmptyDictionary
```
NachoSoto, line 469: Hmm this shouldn't be here since this type isn't `Decodable`.

NachoSoto (approved):

Tests/UnitTests/Purchasing/OfferingsTests.swift
```diff
@@ -153,6 +153,55 @@ class OfferingsTests: TestCase {
...
+                from: products,
+                data: .init(
+                    currentOfferingId: "offering_a",
+                    offerings: [
+                        .init(identifier: "offering_a",
+                              description: "This is the base offering",
+                              packages: [
+                                .init(identifier: "$rc_six_month", platformProductIdentifier: "com.myproduct.annual")
+                              ],
+                             metadata: _metadata),
```
joshdholtz, line 186: Needing `@DefaultDecodable.EmptyDictionary var metadata: [String: AnyDecodable]` and then `_metadata` seems really weird to me but it was the only thing that made Xcode happy 🤷‍♂️ 

Here are screenshots of the errors I was getting otherwise 👇 

# 1. Without anything 

Screenshot 2023-05-23 at 5 57 56 AM

# 2. Without the underscore

Screenshot 2023-05-23 at 5 58 11 AM

NachoSoto: Oh. The right way to solve this is by creating a custom constructor for `OfferingsResponse.Offering`. But simpler is to change this line to `metadata: .init(wrappedValue: metadata)`, which is creating a `DefaultValue`.

Tests/UnitTests/Purchasing/OfferingsTests.swift
```diff
@@ -153,6 +153,55 @@ class OfferingsTests: TestCase {
         expect(offerings.current) == offerings["offering_a"]
     }
 
+    func testOfferingsWithMetadataIsCreated() throws {
```
NachoSoto, line 156: 👍🏻

Sources/Purchasing/Offering.swift
```diff
@@ -165,6 +173,14 @@ import Foundation
 
 }
 
+extension Offering {
+
+    struct Metadata {
+        let data: [String: Any]
```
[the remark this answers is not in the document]
NachoSoto: Yeah, or maybe as an additional API 👍🏻 
```swift
extension Offering {
  func getMetadataValue<T>(for key: String, defaultValue: T) -> T {
    return self.metadata.data[key] as? T ?? defaultValue
  }
} 
```

Sources/Purchasing/Offering.swift
```diff
@@ -38,6 +38,13 @@ import Foundation
      */
     @objc public let serverDescription: String
 
+    private let _metadata: Metadata
+
+    /**
+     Offering metadata defined in RevenueCat dashboard.
+     */
+    @objc public var metadata: [String: Any] { self._metadata.data }
```
NachoSoto, line 46: I just realized we need to put this in `OfferingAPI.swift` and `RCOfferingAPI.m`

Tests/UnitTests/Purchasing/OfferingsTests.swift
NachoSoto: I know this is confusing, let me know if you need help with this.

Sources/Purchasing/Offering.swift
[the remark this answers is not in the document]
tonidero: Also, we might want to have different accessors so people don't need to handle `Any`. Things like `getStringValue`, `getIntValue`, `getBoolValue`,... Not sure if we want to support different types though!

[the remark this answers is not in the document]
NachoSoto: That's not very Swifty.
For Obj-C the current API works well.
For Swift I'd go with my suggested generic API.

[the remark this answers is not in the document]
joshdholtz: Do we want `getMetadataValue()` on `Offering`? Or should we make the `Metadata` object public with a subscript thing?

Something like this on `Metadata`... 🤷‍♂️ 

```swift
public subscript<T>(key: String, defaultValue defaultValue: T) -> T {
    return self.data[key] as? T ?? defaultValue
}
```

Sources/Purchasing/Offering.swift
```diff
@@ -165,6 +173,14 @@ import Foundation
 
 }
 
+extension Offering {
```
NachoSoto, line 176: I'd make this `private` since it doesn't need to be visible outside this file:
```suggestion
private extension Offering {
```

NachoSoto (changes requested):
Marking this as request changes since I want to look at the final version with the pending changes :)

Sources/Purchasing/Offering.swift
```diff
@@ -165,6 +173,17 @@ import Foundation
...
+
+    /**
+     Gets a value from `metadata` and falls back to a specified default value.
```
NachoSoto, line 179: Maybe:
```suggestion
     - Returns: the `metadata` value associated to `key` for the expected type,
       or `defaultValue` if not found, or it's not the expected type.
```

Sources/Purchasing/Offering.swift
```diff
@@ -165,6 +173,17 @@ import Foundation
...
+     */
+    public func getMetadataValue<T>(for key: String, defaultValue: T) -> T {
```
NachoSoto, line 181: Being nit-picky since this is `public` so we want to get it right from the beginning.
Maybe:
```suggestion
    public func getMetadataValue<T>(for key: String, default: T) -> T {
```
Swift guidelines require avoiding that alliteration.

Tests/APITesters/SwiftAPITester/SwiftAPITester/OfferingAPI.swift
```diff
@@ -29,6 +29,10 @@ func checkOfferingAPI() {
     var pPack: Package? = off.package(identifier: "")
     pPack = off.package(identifier: nil)
     let package: Package? = off[""]
+    let metadata: [String: Any] = off.metadata
+    let metadataString: String = off.getMetadataValue(for: "", defaultValue: "")
+    let metadataInt: Int = off.getMetadataValue(for: "", defaultValue: 0)
```
NachoSoto, line 34: I know you were following the pattern here, but it's leftover from the first implementation which was kind of messy.
We should get rid of all these `print` calls.

To avoid increasing the noise, I'd change these to:
```suggestion
    let _: String = off.getMetadataValue(for: "", defaultValue: "")
    let _: Int = off.getMetadataValue(for: "", defaultValue: 0)
```

Tests/UnitTests/Purchasing/OfferingsTests.swift
```diff
@@ -153,6 +153,55 @@ class OfferingsTests: TestCase {
...
+                )
+            )
+        )
+
+        expect(offerings["offering_a"]).toNot(beNil())
+        expect(offerings["offering_b"]).toNot(beNil())
+        expect(offerings.current) == offerings["offering_a"]
+
+        let offeringA = try XCTUnwrap(offerings["offering_a"])
+        expect(offeringA.metadata.count) == metadata.count
```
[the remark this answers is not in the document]
NachoSoto: Nit: I'd change this to `haveCount`, otherwise the test is great 👍🏻

NachoSoto (commented):
Almost ready! Just some nit-picks about the API.

Tests/UnitTests/Purchasing/OfferingsTests.swift
```diff
@@ -153,6 +153,59 @@ class OfferingsTests: TestCase {
         expect(offerings.current) == offerings["offering_a"]
     }
 
+    func testOfferingsWithMetadataIsCreated() throws {
+        @DefaultDecodable.EmptyDictionary
```
NachoSoto, line 157: Also reminder to remove this

joshdholtz: I think this is needed though 🤔 I couldn't get it to compile without that 😬

NachoSoto (approved):
Ship it!

Sources/Purchasing/Offering.swift
```diff
@@ -165,6 +173,21 @@ import Foundation
 
 }
 
+extension Offering {
+
+    /**
+     - Returns: the `metadata` value associated to `key` for the expected type,
+     or `defaultValue` if not found, or it's not the expected type.
+     */
+    public func getMetadataValue<T>(for key: String, default: T) -> T {
```
tonidero, line 182: Would this also support null default values? Like `let metadataInt: Int? = offering.getMetadataValue(for: "key", default: nil)`. I think that could be useful.

tonidero (approved):
Looks good!

Sources/Purchasing/Offering.swift
joshdholtz: It should be let me write another test for that! 😊

joshdholtz: Yup, this works!

```swift
let optionalInt: Int? = offeringA.getMetadataValue(for: "optionalInt", default: nil)
expect(optionalInt).to(beNil())
```

PR #1332, opened 2022-02-22
`IntegrationTests`: simplified tests using `async` APIs

NachoSoto:
### Improvements:
- No more `expect(completed).toEventually(beTrue())`.
- No more `expect(error).to(beNil())`. Using `try` and `throws` tests.
- No more `expect(result).toNot(beNil())` or `XCTUnwrap`. The new APIs already return non-optional data when not failed.
- No more `toEventually`s + polling. Tests run serially now without busy waiting, which will make them faster.
- No more completion block trees, tests are linear now.

### Other changes:
- Removed empty optional `purchases(_:, shouldPurchasePromoProduct:, defermentBlock:)`.
- Using modern `==` Nimble methods.
- Removed redundant `syncPurchases` call when purchasing.

PR #2501, opened 2023-05-17
`Offline Entitlements`: integration tests

NachoSoto:
## Example tests:
```swift
func testOfferingsAreCachedInMemory() async throws {
    let onlineOfferings = try await Purchases.shared.offerings()
    expect(onlineOfferings.all).toNot(beEmpty())

    self.serverDown()

    let offlineOfferings = try await Purchases.shared.offerings()
    expect(offlineOfferings) === onlineOfferings
}

func testOfferingsAreCachedOnDisk() async throws {
    let onlineOfferings = try await Purchases.shared.offerings()
    expect(onlineOfferings.all).toNot(beEmpty())

    self.serverDown()
    await self.resetPurchases()

    let offlineOfferings = try await Purchases.shared.offerings()
    expect(offlineOfferings.response) == onlineOfferings.response

    let offering = try XCTUnwrap(offlineOfferings.current)
    expect(offering.availablePackages.count) == onlineOfferings.current?.availablePackages.count
    expect(offering.monthly?.storeProduct.productIdentifier) == "com.revenuecat.monthly_4.99.1_week_intro"
}
```

### DIFF
--- a/Sources/Misc/DangerousSettings.swift
+++ b/Sources/Misc/DangerousSettings.swift
@@ -13,14 +13,11 @@ import Foundation
  */
 @objc(RCDangerousSettings) public final class DangerousSettings: NSObject {
 
-    /// Dangerous settings not exposed outside of the SDK.
-    internal struct InternalSettings {
+    internal struct Internal: InternalDangerousSettingsType {
 
-        /// Whether `ReceiptFetcher` can retry fetching receipts.
         let enableReceiptFetchRetry: Bool
 
         #if DEBUG
-        /// Whether `HTTPClient` will fake server errors
         let forceServerErrors: Bool
 
         init(enableReceiptFetchRetry: Bool = false, forceServerErrors: Bool = false) {
@@ -60,7 +57,7 @@ import Foundation
      */
     @objc public let customEntitlementComputation: Bool
 
-    internal let internalSettings: InternalSettings
+    internal let internalSettings: InternalDangerousSettingsType
 
     @objc public override convenience init() {
         self.init(autoSyncPurchases: true)
@@ -84,14 +81,14 @@ import Foundation
     @objc internal convenience init(autoSyncPurchases: Bool = true, customEntitlementComputation: Bool) {
         self.init(autoSyncPurchases: autoSyncPurchases,
                   customEntitlementComputation: customEntitlementComputation,
-                  internalSettings: .default)
+                  internalSettings: Internal.default)
 
     }
 
     /// Designated initializer
     internal init(autoSyncPurchases: Bool,
                   customEntitlementComputation: Bool = false,
-                  internalSettings: InternalSettings) {
+                  internalSettings: InternalDangerousSettingsType) {
         self.autoSyncPurchases = autoSyncPurchases
         self.internalSettings = internalSettings
         self.customEntitlementComputation = customEntitlementComputation
@@ -100,3 +97,16 @@ import Foundation
 }
 
 extension DangerousSettings: Sendable {}
+
+/// Dangerous settings not exposed outside of the SDK.
+internal protocol InternalDangerousSettingsType: Sendable {
+
+    /// Whether `ReceiptFetcher` can retry fetching receipts.
+    var enableReceiptFetchRetry: Bool { get }
+
+    #if DEBUG
+    /// Whether `HTTPClient` will fake server errors
+    var forceServerErrors: Bool { get }
+    #endif
+
+}
--- a/Sources/Purchasing/StoreKit1/StoreKit1Wrapper.swift
+++ b/Sources/Purchasing/StoreKit1/StoreKit1Wrapper.swift
@@ -55,7 +55,7 @@ class StoreKit1Wrapper: NSObject {
     weak var delegate: StoreKit1WrapperDelegate? {
         didSet {
             if self.delegate != nil {
-                self.notifyDelegateOfExistingTransactions()
+                self.notifyDelegateOfExistingTransactionsIfNeeded()
 
                 self.paymentQueue.add(self)
             } else {
@@ -114,7 +114,7 @@ class StoreKit1Wrapper: NSObject {
         return payment
     }
 
-    private func notifyDelegateOfExistingTransactions() {
+    private func notifyDelegateOfExistingTransactionsIfNeeded() {
         // Here be dragons. Explanation:
         // When initializing the SDK after an app opens, `SKPaymentQueue` notifies its
         // transaction observers of _existing_ transactions, so this method is normally not required.
--- a/Sources/Purchasing/StoreKit1/StoreKit1Wrapper.swift
+++ b/Sources/Purchasing/StoreKit1/StoreKit1Wrapper.swift
@@ -55,6 +55,8 @@ class StoreKit1Wrapper: NSObject {
     weak var delegate: StoreKit1WrapperDelegate? {
         didSet {
             if self.delegate != nil {
+                self.notifyDelegateOfExistingTransactions()
+
                 self.paymentQueue.add(self)
             } else {
                 self.paymentQueue.remove(self)
@@ -110,6 +112,34 @@ class StoreKit1Wrapper: NSObject {
         let payment = self.payment(with: product)
         payment.paymentDiscount = discount
         return payment
+    }
+
+    private func notifyDelegateOfExistingTransactions() {
+        // Here be dragons. Explanation:
+        // When initializing the SDK after an app opens, `SKPaymentQueue` notifies its
+        // transaction observers of _existing_ transactions, so this method is normally not required.
+        //
+        // However: `BaseOfflineStoreKitIntegrationTests` simulates restarting apps.
+        // When it re-creates `Purchases` to do that, `StoreKit 1` doesn't know to re-notify
+        // its observers. This does so manually.
+        // This isn't required in StoreKit 2 because resubscribing to
+        // `StoreKit.Transaction.updates` does forward existing transactions.
+
+        #if DEBUG
+        guard ProcessInfo.isRunningIntegrationTests, let delegate = self.delegate else { return }
+
+        let transactions = self.paymentQueue.transactions
+        guard !transactions.isEmpty else { return }
+
+        Logger.appleWarning(
+            "StoreKit1Wrapper: sending delegate \(transactions.count) existing transactions " +
+            "for Integration Tests."
+        )
+
+        for transaction in transactions {
+            delegate.storeKit1Wrapper(self, updatedTransaction: transaction)
+        }
+        #endif
     }
 
 }
--- a/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
@@ -89,7 +89,7 @@ class BaseBackendIntegrationTests: XCTestCase {
     }
 
     /// Simulates closing the app and re-opening with a fresh instance of `Purchases`.
-    final func resetPurchases() async {
+    final func resetSingleton() async {
         Logger.warn("Resetting Purchases.shared")
 
         Purchases.clearSingleton()
--- a/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
@@ -15,15 +15,18 @@ import Nimble
 @testable import RevenueCat
 import XCTest
 
-final class TestPurchaseDelegate: NSObject, PurchasesDelegate {
+final class TestPurchaseDelegate: NSObject, PurchasesDelegate, Sendable {
 
-    var customerInfo: CustomerInfo?
-    var customerInfoUpdateCount = 0
+    private let _customerInfo: Atomic<CustomerInfo?> = nil
+    private let _customerInfoUpdateCount: Atomic<Int> = .init(0)
 
     func purchases(_ purchases: Purchases, receivedUpdated customerInfo: CustomerInfo) {
-        self.customerInfo = customerInfo
-        customerInfoUpdateCount += 1
+        self._customerInfo.value = customerInfo
+        self._customerInfoUpdateCount.value += 1
     }
+
+    var customerInfo: CustomerInfo? { return self._customerInfo.value }
+    var customerInfoUpdateCount: Int { return self._customerInfoUpdateCount.value }
 
 }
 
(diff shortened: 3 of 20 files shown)


Sources/Misc/DangerousSettings.swift
```diff
@@ -100,3 +97,16 @@ import Foundation
 }
 
 extension DangerousSettings: Sendable {}
+
+/// Dangerous settings not exposed outside of the SDK.
+internal protocol InternalDangerousSettingsType: Sendable {
```
NachoSoto, line 102: This allows having a custom type in integration tests so we can swap the return value, without having to make the actual `DangerousSettings.Internal` `struct` mutable.

Sources/Misc/DangerousSettings.swift
```diff
@@ -13,14 +13,11 @@ import Foundation
  */
 @objc(RCDangerousSettings) public final class DangerousSettings: NSObject {
 
-    /// Dangerous settings not exposed outside of the SDK.
-    internal struct InternalSettings {
+    internal struct Internal: InternalDangerousSettingsType {
```
NachoSoto, line 16: Changed this from `DangerousSettings.InternalSettings` to `DangerousSettings.Internal`.

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -26,7 +26,7 @@ class CustomerInfoResponseHandler {
         purchasedProductsFetcher: PurchasedProductsFetcherType,
         productEntitlementMapping: ProductEntitlementMapping?,
         userID: String,
-        offlineEntitlementsEnabled: Bool = false
+        offlineEntitlementsEnabled: Bool = true
```
NachoSoto, line 29: ⚠️  This PR enables offline entitlements (so we can test them on integration tests).
I think instead I'll get rid of the flag?

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
```diff
@@ -0,0 +1,169 @@
...
+
+        self.serverDown()
+        try await self.purchaseMonthlyProduct()
+
+        logger.verifyMessageWasLogged(Strings.offlineEntitlements.computing_offline_customer_info, level: .info)
+        logger.verifyMessageWasNotLogged("Finishing transaction")
+    }
+
+    @available(iOS 15.2, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func testPurchaseWhileServerIsDownPostsReceiptAfterServerComesBack() async throws {
```
NachoSoto, line 98: I'm glad I'm running these in both SK1 and SK2. Turns out this is failing in SK1 only.

NachoSoto: For some reason, after restarting the SDK and recreating `SKPaymentQueue`, the SDK isn't getting notified about the transaction again, unlike with SK2. Need to figure that out.

Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
```diff
@@ -86,6 +88,14 @@ class BaseBackendIntegrationTests: XCTestCase {
         self.mainThreadMonitor = nil
     }
 
+    /// Simulates closing the app and re-opening with a fresh instance of `Purchases`.
+    final func resetPurchases() async {
```
winstondu, line 92: Nit: naming could be a bit better, as it could be semantically confused with resetting "purchases of entitlements" rather than the "Purchases Singleton"

NachoSoto: I agree 👍🏻 I think I'll rename to `resetSingleton`

Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
```diff
@@ -156,7 +172,14 @@ private extension BaseBackendIntegrationTests {
...
-                     internalSettings: .init(enableReceiptFetchRetry: true))
+                     internalSettings: self)
     }
 
 }
+
+extension BaseBackendIntegrationTests: InternalDangerousSettingsType {
+
+    var enableReceiptFetchRetry: Bool { return true }
+    var forceServerErrors: Bool { return false }
```
winstondu, line 183: I think composition over inheritance is the way to go here.

It feels unnatural for the test class itself, which is fairly large and complex, to conform to this protocol and being what gets "dependency-injected" into the `DangerousSettings` initializer.

Why not have `BaseBackendIntegrationTests` instances hold a member variable of a new type, say a `MutableInternalDangerousSettings` class, that conforms to this protocol? 

The `serverUp()` and `serverDown()` methods can then just mutate `forceServerErrors` that variable.

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
```diff
@@ -0,0 +1,169 @@
...
+
+    var serverIsDown: Bool = false
+    override var forceServerErrors: Bool { return self.serverIsDown }
+
+    override func setUp() async throws {
+        try await super.setUp()
+
+        await self.waitForPendingCustomerInfoRequests()
+        if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) {
+            try await self.ensureEntitlementMappingIsAvailable()
```
tonidero, line 35: Quick question, I guess `serverIsDown` is reset to `false` automatically on each test right?

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
```diff
@@ -0,0 +1,169 @@
...
+        self.serverDown()
+        await self.resetPurchases()
+
+        let offlineOfferings = try await Purchases.shared.offerings()
+        expect(offlineOfferings.response) == onlineOfferings.response
+
+        let offering = try XCTUnwrap(offlineOfferings.current)
+        expect(offering.availablePackages.count) == onlineOfferings.current?.availablePackages.count
+        expect(offering.monthly?.storeProduct.productIdentifier) == "com.revenuecat.monthly_4.99.1_week_intro"
+    }
```
tonidero, line 62: I think these offerings tests are not directly related to offline entitlements so not sure if they should be separate... But this is fine for now.

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
```diff
@@ -0,0 +1,169 @@
...
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func testOfflineCustomerInfoWithNoPurchases() async throws {
+        Purchases.shared.invalidateCustomerInfoCache()
```
tonidero, line 66: Not directly related to this PR but looking at this comment https://github.com/RevenueCat/purchases-ios/blob/main/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift#L144-L154. Looks like that waits for get offerings to finish, but not for the get customer info request to finish I think? If so, there could possibly be race conditions of getting the result for that request after this line is executed... There might be something I'm missing though!

tonidero (commented):
Really love these tests! I think I should be able to copy part of it for the Android tests 👍. Though we don't support coroutines there yet so they won't look nearly as nice 😬

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
tonidero: Nvm! I didn't see `waitForPendingCustomerInfoRequests`

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
NachoSoto: Oops good catch, I don't think it was!

NachoSoto: Done.

Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
NachoSoto: > I think composition over inheritance is the way to go

Normally I agree, but I fail to see how that would improve this particular case. That would semantically be the same.
This set up is much simpler this way. The way these tests are set up is such that the tests themselves become the configuration, which allows us to override flags like [this](https://github.com/RevenueCat/purchases-ios/blob/662940100f0cd1f4005934d2a99b4ba16e9ea6d3/Tests/BackendIntegrationTests/StoreKitObserverModeIntegrationTests.swift#L25).

What benefit does your suggestion provide?

NachoSoto:
> Though we don't support coroutines there yet so they won't look nearly as nice 

@tonidero migrating tests to async (#1332) was huge, it allowed us to iterate, maintain, and write new integration tests much faster. Before it was callback hell.
Might be worth investing on that in iOS.

Tests/UnitTests/TestHelpers/AsyncTestHelpers.swift
```diff
@@ -90,4 +90,13 @@ func asyncWait(
         line: line,
         foundCorrectValue
     ).to(beTrue(), description: description)
+
+    if !foundCorrectValue {
+        struct ConditionFailedError: Error {}
+
+        // Because this method is `async`, for some reason Swift is continuing execution of the test
```
NachoSoto, line 97: This was required to avoid failing tests from crashing instead of failing.

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
NachoSoto: It's related to being able to purchase while offline :P

Sources/Purchasing/StoreKit1/StoreKit1Wrapper.swift
```diff
@@ -112,6 +114,34 @@ class StoreKit1Wrapper: NSObject {
         return payment
     }
 
+    private func notifyDelegateOfExistingTransactions() {
```
tonidero, line 117: Should we add something to the name of this method to indicate that it's only going to be run on tests? Like `notifyDelegateOfExistingTransactionsInTests`?

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
```diff
@@ -0,0 +1,276 @@
...
+        // 2. "Re-open" the app
+        await self.resetSingleton()
+
+        // 3. `CustomerInfo` should contain offline purchase
+        let info = try await Purchases.shared.customerInfo()
+        try await self.verifyEntitlementWentThrough(info)
+    }
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func testPurchaseAgainAfterServerRecovers() async throws {
```
tonidero, line 187: I believe apple/Google shouldn't allow you to purchase the same subscription a second time right? So not sure if this is a valid scenario.

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
```diff
@@ -0,0 +1,276 @@
...
+}
+
+class OfflineWithNoMappingStoreKitIntegrationTests: BaseOfflineStoreKitIntegrationTests {
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func testOfflineCustomerInfoFailsIfNoEntitlementMapping() async throws {
+        self.serverDown()
+
+        do {
+            _ = try await Purchases.shared.customerInfo(fetchPolicy: .fetchCurrent)
```
tonidero, line 249: Hmm so this will fail but not try to enter offline entitlements mode because it already has a cached customer info. I think we are missing a `invalidateCustomerInfoCache` before this?

tonidero (approved):
Looks great! This brought a great point in favor of supporting coroutines in Android 🙏

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
```diff
@@ -0,0 +1,276 @@
...
+        Purchases.shared.invalidateCustomerInfoCache()
+        await self.resetSingleton()
+
+        // 6. To ensure (with a clean cache) that the receipt was posted
+        let info = try await Purchases.shared.customerInfo()
+        try await self.verifyEntitlementWentThrough(info)
+    }
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func testReopeningAppWithOfflineEntitlementsDoesNotReturnStaleCache() async throws {
```
tonidero, line 173: Should we add a test case for foregrounding the app sends unsynced purchases?

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
```diff
@@ -0,0 +1,276 @@
...
+    override func setUp() async throws {
+        try await super.setUp()
+
+        if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) {
+            try await self.ensureEntitlementMappingIsAvailable()
+        }
+    }
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func testOfflineCustomerInfoFailsIfNoEntitlementMapping() async throws {
```
tonidero, line 52: Oh wait, I think this test should be removed since it's in its own class

Sources/Purchasing/StoreKit1/StoreKit1Wrapper.swift
NachoSoto: Good idea. Maybe `notifyDelegateOfExistingTransactionsIfNeeded`

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
NachoSoto: Oh yeah, this is failing because it's not even trying to compute it because there's something in the cache 👍🏻 
Great catch! I'll add that.

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
NachoSoto: Thanks for catching it.

PR #2502, opened 2023-05-17
`MainThreadMonitor`: don't monitor thread if debugger is attached

NachoSoto:
See #2463.
If you're actively debugging and interrupt the main thread, it'll end up asserting, which isn't very convenient.

Copied the logic from https://stackoverflow.com/a/33177600/401024. It's not the safest code, but this only runs in tests.

### DIFF
--- a/Tests/BackendIntegrationTests/MainThreadMonitor.swift
+++ b/Tests/BackendIntegrationTests/MainThreadMonitor.swift
@@ -29,6 +29,11 @@ final class MainThreadMonitor {
     }
 
     func run() {
+        guard !Self.debuggerIsAttached else {
+            Logger.verbose("Debugger is attached, ignoring")
+            return
+        }
+
         self.queue.async { [weak self] in
             while self != nil {
                 let semaphore = DispatchSemaphore(value: 0)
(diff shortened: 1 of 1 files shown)


Tests/BackendIntegrationTests/MainThreadMonitor.swift
```diff
@@ -50,3 +55,22 @@ final class MainThreadMonitor {
...
+private extension MainThreadMonitor {
+
+    // From https://stackoverflow.com/a/33177600/401024
+    static var debuggerIsAttached: Bool {
+        // Buffer for "sysctl(...)" call's result.
+        var info = kinfo_proc()
+        // Counts buffer's size in bytes (like C/C++'s `sizeof`).
+        var size = MemoryLayout.stride(ofValue: info)
+        // Tells we want info about own process.
+        var mib: [Int32] = [CTL_KERN, KERN_PROC, KERN_PROC_PID, getpid()]
```
tonidero, line 68: Not sure but since this info isn't supposed to change, could `mib` be a `let`?

Tests/BackendIntegrationTests/MainThreadMonitor.swift
```diff
@@ -29,6 +29,11 @@ final class MainThreadMonitor {
     }
 
     func run() {
+        guard !Self.debuggerIsAttached else {
+            Logger.verbose("\(type(of: self)): debugger is attached, ignoring")
+            return
+        }
```
tonidero, line 35: So when you run tests through Xcode, the debugger will be attached by default I think. I guess that means this monitor will only run when executing tests through the command line tool correct? We should just be aware that tests results may vary between executing them through xcode and through command line.

tonidero (approved):

Tests/BackendIntegrationTests/MainThreadMonitor.swift
NachoSoto: Yeah or in CI.

Tests/BackendIntegrationTests/MainThreadMonitor.swift
NachoSoto: It's passed as a reference in the next line so it can be modified. Swift would actually suggest this needs to be immutable otherwise :)

PR #2506, opened 2023-05-18
`PurchasedProductsFetcher`: don't throw errors if purchased products were found

NachoSoto:
This is a best-effort implementation. We can still compute the offline `CustomerInfo` if we found some purchased products.
Extracted from #2368 and added tests.

### DIFF
--- a/Tests/StoreKitUnitTests/OfflineEntitlements/PurchasedProductsFetcherTests.swift
+++ b/Tests/StoreKitUnitTests/OfflineEntitlements/PurchasedProductsFetcherTests.swift
@@ -30,7 +30,7 @@ class BasePurchasedProductsFetcherTests: StoreKitConfigTestCase {
 
         self.sandboxDetector = MockSandboxEnvironmentDetector(isSandbox: .random())
         self.fetcher = PurchasedProductsFetcher(
-            appStoreSync: self.syncAppStore,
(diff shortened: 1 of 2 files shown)


Tests/StoreKitUnitTests/OfflineEntitlements/PurchasedProductsFetcherTests.swift
```diff
@@ -18,20 +18,32 @@ import XCTest
...
 
         self.sandboxDetector = MockSandboxEnvironmentDetector(isSandbox: .random())
-        self.fetcher = PurchasedProductsFetcher(sandboxDetector: self.sandboxDetector)
+        self.fetcher = PurchasedProductsFetcher(
+            appStoreSync: self.syncAppStore,
+            sandboxDetector: self.sandboxDetector
+        )
     }
 
+    fileprivate func syncAppStore() async throws {
```
tonidero, line 38: Could this be `appStoreSync`, same as in the other places?

NachoSoto: My thinking is that the closure is a noun "App Store sync closure", versus this is an action / verb: "sync App Store".
Let me know if that's confusing though.

PR #2518, opened 2023-05-19
`PurchaseTesterSwiftUI`: added new window on Mac to manage proxy

NachoSoto:
This allows managing the proxy at runtime through the UI:
Screenshot 2023-05-19 at 2 07 18 PM

See also https://github.com/RevenueCat/khepri-mitm-proxy/pull/6

### DIFF
--- a/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Proxy/ProxyManager.swift
+++ b/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Proxy/ProxyManager.swift
@@ -1,0 +1,109 @@
+//
+//  ProxyManager.swift
+//  PurchaseTester
+//
+//  Created by Nacho Soto on 5/19/23.
+//
+
+import Foundation
+
+#if os(macOS) || targetEnvironment(macCatalyst)
+
+enum ProxyStatus {
+
+    enum Mode: String, Decodable, CaseIterable {
+
+        case disabled               = "OFF"
+        case serverDown             = "SERVER_DOWN"
+        case overrideEntitlements   = "OVERRIDE_ENTITLEMENTS"
+
+    }
+
+    case disabled
+    case enabled(Mode)
+
+}
+
+final class ProxyManager {
+
+    static func fetchProxyStatus(proxyURL: URL) async -> ProxyStatus {
+        do {
+            let response: ProxyStatusResponse = try await URLSession.shared
+                .fetch(proxyURL: proxyURL,
+                       path: "status")
+
+            return .enabled(response.mode)
+        } catch {
+            print("Error fetching proxy status: \(error.localizedDescription)")
+            return .disabled
+        }
+    }
+
+    /// Throws if mode failed to change
+    static func changeMode(proxyURL: URL, to newMode: ProxyStatus.Mode) async throws {
+        let _: ProxyChangeModeResponse = try await URLSession.shared
+            .fetch(proxyURL: proxyURL, path: newMode.pathToChange)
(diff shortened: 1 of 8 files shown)


Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/HomeView.swift
```diff
@@ -273,6 +273,8 @@ private struct CustomerInfoHeaderView: View {
                 #if targetEnvironment(macCatalyst) || os(macOS)
                 if #available(macCatalyst 16.0, *) {
                     OpenWindowButton()
+
+                    OpenProxyWindowButton()
```
tonidero, line 277: I'm wondering, should this button also exist in the configuration screen? In case we want to switch the proxy mode before we configure the SDK.

tonidero (approved):
Left some comments, but I don't think they are blockers

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Proxy/ProxyManager.swift
```diff
@@ -0,0 +1,109 @@
...
+//  Created by Nacho Soto on 5/19/23.
+//
+
+import Foundation
+
+#if os(macOS) || targetEnvironment(macCatalyst)
+
+enum ProxyStatus {
+
+    enum Mode: String, Decodable, CaseIterable {
```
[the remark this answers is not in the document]
NachoSoto: That's what `ProxyStatus.disabled` represents.
This has 2 status: `ProxyStatus.disabled` (if we weren't able to load the mode, like you described), or `ProxyStatus.enabled`, with an associated `Mode`.

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/HomeView.swift
NachoSoto: Yeah good point! Once you open the window it remembers it and stays open every time you open the app, but technically there's no way to open it on first launch, so I'll do that.

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Proxy/ProxyManager.swift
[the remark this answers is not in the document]
tonidero: Ahh right makes sense!

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/HomeView.swift
tonidero: Hmm so the app will remember the proxy mode across launches? I would have guessed it should just use the `/status` endpoint to get the current mode and use that... I think modifying the proxy mode on app launch can be confusing IMO. Maybe I understood you incorrectly though.

NachoSoto: > Hmm so the app will remember the proxy mode across launches? 

No it doesn't, it'll refetch it. It just remembers that the window was open.

NachoSoto: > I'm wondering, should this button also exist in the configuration screen? In case we want to switch the proxy mode before we configure the SDK.

This is a bit of extra work so I'll punt for now. You can still open the window, then click "reconfigure" to change the mode before starting the SDK.

PR #2522, opened 2023-05-22
`Offline Entitlements`: don't compute offline `CustomerInfo` when purchasing a consumable products

NachoSoto:
This will be logged when doing so:
```
ERROR: 😿 ❗❗ Error computing offline CustomerInfo. Will return original error.
Creation error: Unable to compute offline CustomerInfo when purchasing consumable products.
```

### Changes:
- Added unit tests for `PostReceiptDataOperation`
- Added integration test for this new scenario
- `PostReceiptDataOperation` won't use `OfflineCustomerInfoCreator` when purchasing consumable products


NachoSoto:
I ended up going with this which I think is much more elegant. Avoids having to parse the local receipt unnecessarily.

Sources/Networking/Operations/PostReceiptDataOperation.swift
```diff
@@ -39,11 +39,16 @@ final class PostReceiptDataOperation: CacheableNetworkOperation {
         customerInfoCallbackCache: CallbackCache<CustomerInfoCallback>,
         offlineCustomerInfoCreator: OfflineCustomerInfoCreator?
     ) -> CacheableNetworkOperationFactory<PostReceiptDataOperation> {
+        // Can't compute offline CustomerInfo when purchasing non-consumables
+        let offlineCreator: OfflineCustomerInfoCreator? = postData.isNonSubscriptionProduct
```
tonidero, line 43: Hmm so this will only work when posting a receipt right? I don't think this will handle this scenario:
- User purchases a consumable while offline (the post fails)
- For some reason, they invalidate customer info or uninstall/reinstall (so no customer info cached)
- They restart the app
- During restart, get customer info request fails with a 5xx, so we try to enter offline entitlements even though it has a consumable.

Admittedly, this seems like an uncommon case (users invalidating customer info in the middle), but ideally we would also handle it.

NachoSoto: Thanks for pointing this out. The new implementation fixes that 👍🏻 
Also I think it's much cleaner moving the check to use `StoreKit.Transaction.unfinished`.

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
```diff
@@ -97,4 +101,48 @@ class PurchasedProductsFetcher: PurchasedProductsFetcherType {
         return result
     }
 
+    private static var hasPendingConsumablePurchase: Bool {
```
tonidero, line 104: Is it possible to unit test these? I guess we would have to mock `StoreKit`. We do have integration tests so it's probably okish, but I think it's good to test things at different granularities to try to have as much coverage as possible.

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
```diff
@@ -97,4 +101,48 @@ class PurchasedProductsFetcher: PurchasedProductsFetcherType {
...
+        get async {
+            return await StoreKit.Transaction.unfinished.contains {
```
tonidero, line 106: Just to check, but I'm guessing SK2 will cache the transactions internally right? So this shouldn't take too long. 

Also, will this be executed in a background thread? I'm guessing since the API is using async/await that might be handled internally by SK2 though.

tonidero (approved):
LGTM!

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
NachoSoto: It's _indirectly_ tested, but I agree it would be useful to test independently 👍🏻  which we can do with `SKTestSession`.

I added a type in #2533 to do something similar, so I'm going to combine both things and test them.

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
NachoSoto: The short answer is: it doesn't matter. Concurrency in Swift is different from what you'd expect. Even if this takes 1 hour, it's not stopping the thread, it creates a continuation.

Sources/OfflineEntitlements/PurchasedProductsFetcher.swift
NachoSoto: Extracted and tested in #2539 like you suggested.

PR #2528, opened 2023-05-23
`DefaultDecodable`: moved to `Misc/Codable/DefaultDecodable.swift`

NachoSoto:
Non functional change.

I was looking for these types and couldn't find them. Turns out it was under "FoundationExtensions", even though they're new definitions and not a foundation extension.
This creates a new `DefaultDecodable.swift`, and combines `JSONDecoder+Extensions` and `Decoder+Extensions`.


Sources/FoundationExtensions/Decoder+Extensions.swift
```diff
@@ -26,226 +45,142 @@ extension Decoder {
...
-
-    var wrappedValue = Source.defaultValue
-
-}
-
-extension DefaultValue: Equatable where Value: Equatable {}
-extension DefaultValue: Hashable where Value: Hashable {}
-
-extension DefaultValue: Decodable where Value: Decodable {
+extension JSONDecoder {
```
NachoSoto, line 48: This diff is very hard to read. It's simply moving the definitions around.

PR #2529, opened 2023-05-23
`HTTPClient`: don't assume error responses are JSON

NachoSoto:
When receiving 500 responses, `HTTPClient` was always calling `ErrorResponse.from(_ data: Data)`, which tried to decode the content as if it was JSON.
However, if the response had something completely different, it would fail to decode.

This wasn't really a bug because `HTTPClient` was still forwarding the right result, but now it will do so without even trying to parse it.

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -528,7 +528,7 @@ private extension HTTPResponse {
 
 }
 
-private extension HTTPResponse<Data> {
+private extension HTTPResponse where Body == Data {
 
     func parseUnsuccessfulResponse() -> NetworkError {
         let isJSON = self.value(forHeaderField: HTTPClient.ResponseHeader.contentType.rawValue) == "application/json"
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -111,6 +111,7 @@ extension HTTPClient {
         case eTag = "X-RevenueCat-ETag"
         case signature = "X-Signature"
         case requestDate = "X-RevenueCat-Request-Time"
+        case contentType = "Content-Type"
 
     }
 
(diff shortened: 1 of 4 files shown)


Tests/UnitTests/TestHelpers/TestLogHandler.swift
```diff
@@ -134,7 +134,10 @@ extension TestLogHandler {
             line: line,
             self.messages
         )
-        .toNot(containElementSatisfying(Self.entryCondition(message: message, level: level)))
+        .toNot(
+            containElementSatisfying(Self.entryCondition(message: message, level: level)),
```
NachoSoto, line 138: This provides a better failed assertion message, similar to `verifyMessageWasLogged`.

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -413,6 +419,8 @@ final class HTTPClientTests: BaseHTTPClientTests {
         expect(error.isServerDown) == true
 
         expect(MockSigning.requests).to(beEmpty())
+
+        logger.verifyMessageWasNotLogged("Couldn't decode data from json")
```
NachoSoto, line 423: This was failing.

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -531,3 +527,18 @@ extension HTTPResponse {
     }
 
 }
+
+private extension HTTPResponse where Body == Data {
+
+    func parseUnsuccessfulResponse() -> NetworkError {
+        let isJSON = self.value(forHeaderField: HTTPClient.ResponseHeader.contentType.rawValue) == "application/json"
```
tonidero, line 534: My guess is yes but should we confirm the response will always have this header? (When the server is up)

tonidero (approved):

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto: If it doesn't we still produce an error, we just won't try to parse the specific backend error, so in practice it doesn't really matter that much.

PR #2532, opened 2023-05-23
`Integration Tests`: workaround for `XCTest` crash after a test failure

NachoSoto:
See https://app.circleci.com/jobs/github/RevenueCat/purchases-ios/69881

Assertion failures in that method were leading to _crashes_ in `Purchases.shared`, instead of a simple test failure that could be retried:
image

For some reason, `XCTest` is continuing execution after an assertion failure in an `async` method, despite having `self.continueAfterFailure = false`.
This might even be a Swift mis-compilation. This PR is a workaround: instead of relying on `XCTest` stopping execution of the test, we instead throw an `Error` to force that at compile time.

Note that this still produces nice errors when an assertion fails inside of `verifyEntitlementWentThrough`:
Screenshot 2023-05-23 at 13 27 55

### DIFF
--- a/Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
@@ -147,19 +147,27 @@ extension BaseStoreKitIntegrationTests {
         file: FileString = #file,
         line: UInt = #line
     ) async throws -> EntitlementInfo {
-        let entitlements = customerInfo.entitlements.all
+        // This is used to throw an error when the test fails.
+        // For some reason XCTest is continuing execution even after a test failure
+        // despite having `self.continueAfterFailure = false`
+        //
+        // By doing this, instead of only calling `fail`, we ensure that
+        // Swift stops executing code when an assertion has failed,
+        // and therefore avoid code running after the test has already failed.
+        // This prevents test crashes from code calling `Purchases.shared` after the test has ended.
+        func failTest(_ message: String) async throws {
+            struct ExpectationFailure: Swift.Error {}
 
-        if entitlements.isEmpty {
             await self.printReceiptContent()
+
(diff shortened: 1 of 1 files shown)


Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
```diff
@@ -147,19 +147,27 @@ extension BaseStoreKitIntegrationTests {
...
-        let entitlements = customerInfo.entitlements.all
+        // This is used to throw an error when the test fails.
+        // For some reason XCTest is continuing execution even after a test failure
+        // despite having `self.continueAfterFailure = false`
+        //
+        // By doing this, instead of only calling `fail`, we ensure that
+        // Swift stops executing code when an assertion has failed,
+        // and therefore avoid code running after the test has already failed.
+        // This prevents test crashes from code calling `Purchases.shared` after the test has ended.
+        func failTest(_ message: String) async throws {
```
tonidero, line 158: Is this happening in other tests?

tonidero (approved):
Just a question but looks good! Seems weird that it continues executing after an assertion but 🤷‍♂️. Considering you said this happens in async/await functions, I wonder if this is a threading issue, like maybe the `fail` function changes threads at some point before failing?

NachoSoto:
> I wonder if this is a threading issue, like maybe the fail function changes threads at some point before failing?

These tests are marked as `@MainActor` so I don't think so. I could investigate it but I didn't think it was worth our time right now, especially since that Radar might not go anywhere.

Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
NachoSoto: Not that I know. This particular case was happening a fair amount in CI though, because we use this method in pretty much all integration tests.

PR #2533, opened 2023-05-23
`GetCustomerInfo` posts receipts if there are pending transactions

NachoSoto:
This fixes the last edge-case for offline entitlements. Solves the following scenario:
```swift
func testPurchaseWhileServerIsDownPostsReceiptWhenForegroundingApp() async throws {
    let logger = TestLogHandler()

    // 1. Purchase while server is down
    self.serverDown()

    logger.verifyMessageWasNotLogged("Finishing transaction")

    // 2. Server is back
    self.serverUp()

    // 3. Request current CustomerInfo
    let info1 = try await Purchases.shared.customerInfo()
    try await self.verifyEntitlementWentThrough(info1)

    // 4. Ensure transaction is finished
    logger.verifyMessageWasLogged("Finishing transaction", level: .info)

    // 5. Restart app
    Purchases.shared.invalidateCustomerInfoCache()
    await self.resetSingleton()

    // 6. To ensure (with a clean cache) that the receipt was posted
    let info2 = try await Purchases.shared.customerInfo()
    try await self.verifyEntitlementWentThrough(info2)
}
```

This race condition happened when fetching `CustomerInfo` after we had computed an offline `CustomerInfo` with a purchase: if the server is still not aware of the purchase, then we lose that until some arbitrary amount of time when `StoreKit` might decide to send the pending transactions in the queue.
For that reason, instead of waiting, we proactively check now whenever `CustomerInfoManager` fetches new info.

The refactors #2540 and #2542 where required to break a retain cycle.
Also thanks to #2534 we know for sure that there aren't any cycles.

Note that this is an iOS 15.0+ only feature. It would be useful to better ensure consistency in older versions, but it's not strictly required, as those versions don't support offline entitlements either.

### DIFF
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -13,6 +13,8 @@
 
 import Foundation
 
+// swiftlint:disable file_length
+
 class CustomerInfoManager {
 
     typealias CustomerInfoCompletion = @MainActor @Sendable (Result<CustomerInfo, BackendError>) -> Void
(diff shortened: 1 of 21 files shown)


Sources/Identity/CustomerInfoManager.swift
```diff
@@ -279,20 +284,89 @@ class CustomerInfoManager {
...
+                                     withRandomDelay: isAppBackgrounded,
+                                     allowComputingOffline: allowComputingOffline,
+                                     completion: completion)
+    }
+
+    // Note: this is just a best guess.
+    private static let sourceForUnfinishedTransaction: PurchaseSource = .init(
+        isRestore: false,
+        // This might have been in theory a `.purchase`. The only downside of this is that the server
+        // won't validate that the product is present in the receipt.
```
tonidero, line 366: Hmm can you explain more in what situation this transaction could originate from a `purchase`? If we are in the middle of posting the transaction after a purchase, this should wait for that I guess?

tonidero (approved):
Just a question but looks great! I think we should also implement this behavior in Android soon

Sources/Identity/CustomerInfoManager.swift
NachoSoto: 1. Server offline
2. Purchase fails
3. Compute customer info (but transaction still in queue)
4. Next time you launch the app, server is back up
5. Pending transaction -> post receipt

Technically that _was_ a `.purchase` when it was done originally, but we're sending it as `.queue` now.

tonidero: Ahh right, that would also happen currently if a POST receipt fails and it's retried later on right? But yeah, that makes sense 👍

PR #2104, opened 2022-11-30
`BasePurchasesTests`: added assertion to ensure `Purchases` does not leak

NachoSoto:
To continue debugging [CSDK-517], this adds an assertion to make sure that we don't leak `Purchases`, either because of a retain cycle, or due to any `XCTest` issue.

[CSDK-517]: https://revenuecats.atlassian.net/browse/CSDK-517?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2534, opened 2023-05-23
`BasePurchasesTests`: fixed leak detection

NachoSoto:
While working on #2533, I knew that my proof of concept had a retain cycle. I however was very confused when it wasn't detected by this mechanism introduced first introduced in #2104.

Turns out that this had been wrong the whole time: the current implementation, as explained by the comment, was executing those blocks in LIFO order, which meant that the references were being set to `nil` before the assertions could be created.
This meant that by the time the `except { [weak purchases = self.purchases] ... }` lines were called, `self.purchases` itself was `nil`, so nothing was being checked.

This simplifies the implementation by inlining the checks, and saving a `weak` reference before clearing the singleton and all the local references.

Luckily there was only one failing test that had to do with how `Purchases` was being created, and implementation details of `expectFatalError`. To avoid this, I changed the test to use the `Purchases.init` method directly.

### DIFF
--- a/Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
+++ b/Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
@@ -101,54 +101,20 @@ class BasePurchasesTests: TestCase {
         // this level it should be moved to `StoreKitUnitTests`, which runs serially.
         Purchases.logLevel = .verbose
 
-        // See `addTeardownBlock` docs:
-        // - These run *before* `tearDown`.
-        // - They run in LIFO order.
         self.addTeardownBlock {
-            expect { [weak purchases = self.purchases] in purchases }
-                .toEventually(beNil(), description: "Purchases has leaked")
-        }
-        self.addTeardownBlock {
-            expect { [weak orchestrator = self.purchasesOrchestrator] in orchestrator }
-                .toEventually(beNil(), description: "PurchasesOrchestrator has leaked")
-        }
-        self.addTeardownBlock {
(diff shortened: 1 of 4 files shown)


NachoSoto:
And now this correctly detects the retain cycle I introduced 🎉

Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
```diff
@@ -101,54 +101,20 @@ class BasePurchasesTests: TestCase {
...
-            expect { [weak orchestrator = self.purchasesOrchestrator] in orchestrator }
-                .toEventually(beNil(), description: "PurchasesOrchestrator has leaked")
-        }
-        self.addTeardownBlock {
-            expect { [weak deviceCache = self.deviceCache] in deviceCache }
-                .toEventually(beNil(), description: "DeviceCache has leaked: \(self)")
-        }
+            weak var purchases = self.purchases
+            weak var orchestrator = self.purchasesOrchestrator
+            weak var deviceCache = self.deviceCache
```
tonidero, line 107: Should we rename these 3 so there is less confusion with the global variables?

tonidero (approved):

Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
NachoSoto: They're limited to the scope of this method so I think it's clear that they're shadowing those variables. But let me know if you disagree.

PR #2535, opened 2023-05-24
`BackendError`/`NetworkError`: removed `Equatable` conformances

NachoSoto:
This wasn't required, likely since the introduction of #1879, which added `matchError` as a much better way to check errors.
I noticed this when adding a new `case` with an associated value that wasn't `Equatable`.

Tiny change but should make compilation faster, and reduce the binary size.


NachoSoto:
Actually this is useful, it is indeed used by several tests, so closing.

PR #2536, opened 2023-05-24
`PurchasesOrchestrator`: simplified `StoreKit2TransactionListenerDelegate` transaction handling

NachoSoto:
While working on #2533 I noticed that the implementation for `StoreKit2TransactionListenerDelegate` was unnecessarily complex:
```swift
extension PurchasesOrchestrator: StoreKit2TransactionListenerDelegate {
    func storeKit2TransactionListener(
        _ listener: StoreKit2TransactionListener,
        updatedTransaction transaction: StoreTransactionType
    ) async throws {
        let isRestore = self.systemInfo.observerMode

        _ = try await self.syncPurchases(receiptRefreshPolicy: .always,
                                         isRestore: isRestore,
                                         initiationSource: .queue)

        await Async.call { completion in
            self.finishTransactionIfNeeded(transaction) { @MainActor in
                completion(())
            }
        }
    }
}
```

Instead of only posting the one transaction, we were calling `syncPurchases`, which is inconsistent with how we handle SK1 transactions.
The new logic in #2533 will also make use of this single-transaction handling code, so it makes sense to unify this as well.

Note that since #1704, **this method is only invoked for transactions detected outside of a manual purchase flow**.

This only required an overload to specify the `InitiationSource`, which uncovered one incorrect `isRestore` parameter in one of our tests.
I verified that this is expected based on other tests, and the implementation of `initiationSource(for productIdentifier:restored:)`, as well as checking with @antoniobg.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -177,7 +177,7 @@ final class PurchasesOrchestrator {
 
     func syncPurchases(completion: (@Sendable (Result<CustomerInfo, PurchasesError>) -> Void)? = nil) {
         self.syncPurchases(receiptRefreshPolicy: .never,
-                           isRestore: allowSharingAppStoreAccount,
+                           isRestore: self.allowSharingAppStoreAccount,
                            initiationSource: .restore,
                            completion: completion)
     }
@@ -481,7 +481,7 @@ final class PurchasesOrchestrator {
         let customerInfo: CustomerInfo
 
         if let transaction = transaction {
-            customerInfo = try await self.handlePurchasedTransaction(transaction)
+            customerInfo = try await self.handlePurchasedTransaction(transaction, .purchase)
         } else {
             // `transaction` would be `nil` for `Product.PurchaseResult.pending` and
             // `Product.PurchaseResult.userCancelled`.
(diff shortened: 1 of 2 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -858,17 +868,7 @@ extension PurchasesOrchestrator: StoreKit2TransactionListenerDelegate {
...
-
-        _ = try await self.syncPurchases(receiptRefreshPolicy: .always,
-                                         isRestore: isRestore,
-                                         initiationSource: .queue)
-
-        await Async.call { completion in
-            self.finishTransactionIfNeeded(transaction) { @MainActor in
-                completion(())
-            }
-        }
```
NachoSoto, line 871: Goodbye manual code here.

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -725,7 +726,7 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
 
         expect(transaction.finishInvoked) == false
         expect(self.backend.invokedPostReceiptData) == true
-        expect(self.backend.invokedPostReceiptDataParameters?.isRestore) == true
+        expect(self.backend.invokedPostReceiptDataParameters?.isRestore) == false
```
NachoSoto, line 729: I explained this in the PR description, this was wrong.

NachoSoto:
This breaks `testResubscribeAfterExpiration` because `PurchasesOrchestrator` gets notified from the queue right before the second purchase :/
I'll have to look into it, so putting this on hold for now.

NachoSoto:
This will probably benefit from #2540.

NachoSoto:
Done in #2540 instead.

PR #2539, opened 2023-05-24
Created `StoreKit2TransactionFetcher`

NachoSoto:
This extracts the logic from `PurchasedProductsFetcher` and adds tests. It will be used in #2536.

Also exposed and combined the duplicated `underlyingTransaction` implementations.


Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -0,0 +1,78 @@
...
+        get async {
+            return await StoreKit.Transaction
+                .unfinished
+                .compactMap { $0.verifiedTransaction }
+                .map { StoreTransaction(sk2Transaction: $0) }
+                .extractValues()
+        }
+    }
+
+    var hasPendingConsumablePurchase: Bool {
```
tonidero, line 40: I see we are not using `verifiedTransaction` here. Should we filter to only consider verified transactions here as well? Not sure when a purchase will be unverified in SK2 though, so this might be ok.

tonidero (approved):
Just a question but I think it's probably ok so :shipit:

PR #2540, opened 2023-05-24
Extracted `TransactionPoster` from `PurchasesOrchestrator`

NachoSoto:
Necessary refactor for #2533. This allows `CustomerInfoManager` to have an instance of `TransactionPosterType` instead of the whole `PurchasesOrchestrator`.

Essentially this becomes the hierarchy now:
IMG_B8C56C11B7E5-1

### Changes:
- Created stateless `TransactionPoster`
- Extracted `handlePurchasedTransaction` into `TransactionPoster`
- The new implementation there is _mostly_ the same, with the one difference that it's stateless. Instead of holding `purchaseCompleteCallbacksByProductID`, that's done by `PurchasesOrchestrator`. The methods there all take a callback, so they're easier to use and don't require keeping a callback around.
- Created `PurchaseSource` to abstract the combo of `isRestore` and `InitiationSource`. This is still messy using the deprecated `allowSharingAppStoreAccount`, but at least that's abstracted out and the new `TransactionPoster` won't deal with that deprecated property.
- I didn't mock `TransactionPoster`, which means that all existing tests that check `PurchasesOrchestrator` continue working the same way. This was on purpose to keep this refactor as simple as possible.
- This supersedes #2536: the change made there was now possible here without any integration test failures, thanks to the fact that we can process transactions independently without messing with the `purchaseCompleteCallbacksByProductID` state now.

### Future improvements
This is just a start, there's a lot more than can be done to simplify the `PurchasesOrchestrator` monster. The main thing would be to remove all the custom code for `restorePurchases`, which is largely the same as what's in `TransactionPoster` now.

### Notes:
Just posting this here for posteriority. I had to trace the call hierarchy for both SK1 and SK2. This "block" is what's now in `TransactionPoster`:
IMG_C60274393F7F-1

### DIFF
--- a/Sources/Misc/Concurrency/OperationDispatcher.swift
+++ b/Sources/Misc/Concurrency/OperationDispatcher.swift
@@ -68,7 +68,5 @@ extension OperationDispatcher {
 
 }
 
-#if swift(<5.8)
-// `DispatchQueue` is not `Sendable` as of Swift 5.7, but this class performs no mutations.
+// `DispatchQueue` is not `Sendable` as of Swift 5.8, but this class performs no mutations.
 extension OperationDispatcher: @unchecked Sendable {}
-#endif
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -50,9 +50,6 @@ final class PurchasesOrchestrator {
         self.attribution.unsyncedAttributesByKey(appUserID: self.appUserID)
     }
 
-    static let receiptRetryCount: Int = 3
-    static let receiptRetrySleepDuration: DispatchTimeInterval = .seconds(5)
-
     private let productsManager: ProductsManagerType
     private let paymentQueueWrapper: EitherPaymentQueueWrapper
     private let systemInfo: SystemInfo
@@ -62,6 +59,7 @@ final class PurchasesOrchestrator {
     private let receiptParser: PurchasesReceiptParser
     private let customerInfoManager: CustomerInfoManager
     private let backend: Backend
+    private let transactionPoster: TransactionPosterType
     private let currentUserProvider: CurrentUserProvider
     private let transactionsManager: TransactionsManager
     private let deviceCache: DeviceCache
@@ -97,6 +95,7 @@ final class PurchasesOrchestrator {
                      receiptParser: PurchasesReceiptParser,
                      customerInfoManager: CustomerInfoManager,
                      backend: Backend,
+                     transactionPoster: TransactionPoster,
                      currentUserProvider: CurrentUserProvider,
                      transactionsManager: TransactionsManager,
                      deviceCache: DeviceCache,
@@ -116,6 +115,7 @@ final class PurchasesOrchestrator {
             receiptParser: receiptParser,
             customerInfoManager: customerInfoManager,
             backend: backend,
+            transactionPoster: transactionPoster,
             currentUserProvider: currentUserProvider,
             transactionsManager: transactionsManager,
             deviceCache: deviceCache,
@@ -145,6 +145,7 @@ final class PurchasesOrchestrator {
          receiptParser: PurchasesReceiptParser,
          customerInfoManager: CustomerInfoManager,
          backend: Backend,
+         transactionPoster: TransactionPoster,
          currentUserProvider: CurrentUserProvider,
          transactionsManager: TransactionsManager,
          deviceCache: DeviceCache,
@@ -160,6 +161,7 @@ final class PurchasesOrchestrator {
         self.receiptParser = receiptParser
         self.customerInfoManager = customerInfoManager
         self.backend = backend
+        self.transactionPoster = transactionPoster
         self.currentUserProvider = currentUserProvider
         self.transactionsManager = transactionsManager
         self.deviceCache = deviceCache
@@ -487,7 +489,7 @@ final class PurchasesOrchestrator {
         let customerInfo: CustomerInfo
 
         if let transaction = transaction {
-            customerInfo = try await self.handlePurchasedTransaction(transaction)
+            customerInfo = try await self.handlePurchasedTransaction(transaction, .purchase)
         } else {
             // `transaction` would be `nil` for `Product.PurchaseResult.pending` and
             // `Product.PurchaseResult.userCancelled`.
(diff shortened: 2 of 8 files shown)


NachoSoto:
I hated this new name, but just realized it matches our other `AttributionPoster` 🤝

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -0,0 +1,306 @@
...
+    }
+
+    func finishTransactionIfNeeded(
+        _ transaction: StoreTransactionType,
+        completion: @escaping @Sendable @MainActor () -> Void
+    ) {
+        @Sendable
+        func complete() {
+            self.operationDispatcher.dispatchOnMainActor(completion)
+        }
```
aboedo, line 117: I realize this was already there, but I'm not sure it's necessarily easier to read having this one-line function called twice vs just calling the same line twice manually

aboedo: lmk if it was done for `@Sendable` purposes

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -0,0 +1,306 @@
...
+        self.attribution.markAttributesAsSynced(subscriberAttributes, appUserID: self.appUserID)
+    }
+
+}
+
+// MARK: - Implementation
+
+private extension TransactionPoster {
+
+    /// Called as a result a purchase.
```
aboedo, line 152: ... or a renewal, right? 

I realize a renewal is a type of purchase, but given that in parts of the codebase we make distinctions between one and the other to determine initiation source, I think it might be good to clarify

aboedo: and it might also be good to clarify that this won't be called from restores, since restores aren't tied to specific transactions

aboedo (approved):
Looks great! When we're done with the refactors it'd be great to get a tests specifically for TransactionPoster, even though we're currently reusing the existing tests

NachoSoto:
> When we're done with the refactors it'd be great to get a tests specifically for TransactionPoster, even though we're currently reusing the existing tests

Good idea 👍🏻  doing that right now in a separate PR.

Sources/Misc/Concurrency/OperationDispatcher.swift
```diff
@@ -68,7 +68,5 @@ extension OperationDispatcher {
 
 }
 
-#if swift(<5.8)
-// `DispatchQueue` is not `Sendable` as of Swift 5.7, but this class performs no mutations.
+// `DispatchQueue` is not `Sendable` as of Swift 5.8, but this class performs no mutations.
```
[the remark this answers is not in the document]
NachoSoto: I forgot to explain this.
Until this PR, turns out we weren't relying on `OperationDispatcher` being `Sendable`. This broke on Swift 5.8, because we didn't have an `#else`.
I don't think they're ever going to make `DispatchQueue` `Sendable`, but our `@unchecked Sendable` is still valid, hence why I removed the conditional.

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -0,0 +1,306 @@
...
+/// Determines what triggered a purchase and whether it comes from a restore.
+struct PurchaseSource {
+
+    let isRestore: Bool
+    let initiationSource: ProductRequestData.InitiationSource
+
+}
+
+/// A type that can post receipts as a result of a purchased transaction.
+protocol TransactionPosterType: AnyObject, Sendable {
```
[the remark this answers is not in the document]
NachoSoto: Totally. It's so weird that some of our mocks rely only partially on the parent implementation. Plus what we lose on performance by requiring dynamic dispatch versus being able to make the classes `final` (and being able to make them `Sendable`).

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -864,15 +824,19 @@ extension PurchasesOrchestrator: StoreKit2TransactionListenerDelegate {
...
+            self.transactionPoster.handlePurchasedTransaction(
+                StoreTransaction.from(transaction: transaction),
+                presentedOfferingID: nil,
+                storefront: storefront,
+                source: .init(
+                    isRestore: self.allowSharingAppStoreAccount,
+                    initiationSource: .queue
+                )
+            ) { _, customerInfo, error, _ in
+                completed(Result(customerInfo, error))
```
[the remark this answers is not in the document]
NachoSoto: Good point. I think it's still useful to make this method `wait` for this to finish, but we don't actually care about the result so I'll remove that.

Sources/Purchasing/Purchases/TransactionPoster.swift
NachoSoto: I don't think so, it was to remove the duplication so we didn't accidentally use `dispatchOnMainQueue` or something in one of them.

Sources/Purchasing/Purchases/TransactionPoster.swift
NachoSoto: This was an old comment but yeah I'll update 👍🏻

NachoSoto:
Tests: #2557

NachoSoto:
Tests: #2557

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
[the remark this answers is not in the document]
NachoSoto: Actually this is used because of the `try`. This becomes a bit simpler in #2542:
```swift
_ = try await Async.call { completed in
    self.transactionPoster.handlePurchasedTransaction(
        StoreTransaction.from(transaction: transaction),
        data: .init(
            appUserID: self.appUserID,
            presentedOfferingID: nil,
            unsyncedAttributes: self.unsyncedAttributes,
            storefront: storefront,
            source: .init(
                isRestore: self.allowSharingAppStoreAccount,
                initiationSource: .queue
            )
        )
    ) { result in
        completed(result.mapError(\.asPurchasesError))
    }
}
```

Sources/Purchasing/Purchases/TransactionPoster.swift
NachoSoto: Actually I'll remove this comment, it's not that relevant in this context anymore.

PR #2541, opened 2023-05-25
`Backend.ResponseHandler` is now `@Sendable`

NachoSoto:
This was necessary for a follow-up refactor to #2540. The main change is in `Backend`, the rest are test changes that are required now that the closure is `@Sendable`.

### DIFF
--- a/Tests/UnitTests/SubscriberAttributes/BackendSubscriberAttributesTests.swift
+++ b/Tests/UnitTests/SubscriberAttributes/BackendSubscriberAttributesTests.swift
@@ -78,8 +78,6 @@ class BackendSubscriberAttributesTests: TestCase {
     // MARK: PostReceipt with subscriberAttributes
 
     func testPostReceiptWithSubscriberAttributesSendsThemCorrectly() throws {
-        var completionCallCount = 0
-
         let subscriberAttributesByKey: [String: SubscriberAttribute] = [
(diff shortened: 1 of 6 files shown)


Tests/UnitTests/SubscriberAttributes/BackendSubscriberAttributesTests.swift
```diff
@@ -78,8 +78,6 @@ class BackendSubscriberAttributesTests: TestCase {
     // MARK: PostReceipt with subscriberAttributes
 
     func testPostReceiptWithSubscriberAttributesSendsThemCorrectly() throws {
-        var completionCallCount = 0
```
tonidero, line 81: Hmm so we were not verifying this right? Instead of removing it, should we verify it instead?

tonidero (approved):

Tests/UnitTests/SubscriberAttributes/BackendSubscriberAttributesTests.swift
NachoSoto: No haha. Yeah I can add it back

NachoSoto: Actually it's checked by other tests already, so I think it's better to keep this test focused on testing this one thing.

PR #2542, opened 2023-05-25
Refactored `TransactionPoster`: removed 2 dependencies and abstracted parameters

NachoSoto:
Follow up to #2540.


Tests/UnitTests/SubscriberAttributes/PurchasesSubscriberAttributesTests.swift
```diff
@@ -700,10 +698,10 @@ class PurchasesSubscriberAttributesTests: TestCase {
 
         expect(self.mockBackend.invokedPostReceiptData).to(beTrue())
         expect(self.mockStoreKit1Wrapper.finishCalled).toEventually(beTrue())
-        expect(self.mockSubscriberAttributesManager.invokedMarkAttributes) == true
-        expect(self.mockSubscriberAttributesManager.invokedMarkAttributesParameters!.syncedAttributes) == mockAttributes
-        expect(self.mockSubscriberAttributesManager.invokedMarkAttributesParameters!.appUserID) ==
-        mockIdentityManager.currentAppUserID
+        expect(self.mockSubscriberAttributesManager.invokedMarkAttributes).toEventually(beTrue())
```
NachoSoto, line 701: This is the only semantic change. Because of this refactor, attributes are marked as synced after the transactions are finished. The test below (like 732) already had `toEventually`. This is purely an ordering change and shouldn't have any impact.

Sources/Networking/CustomerAPI.swift
```diff
@@ -109,16 +104,14 @@ final class CustomerAPI {
...
-        let postData = PostReceiptDataOperation.PostData(appUserID: appUserID,
-                                                         receiptData: receiptData,
-                                                         isRestore: isRestore,
-                                                         productData: productData,
-                                                         presentedOfferingIdentifier: offeringIdentifier,
-                                                         observerMode: observerMode,
-                                                         initiationSource: initiationSource,
-                                                         subscriberAttributesByKey: subscriberAttributesToPost)
+        let postData = PostReceiptDataOperation.PostData(
+            transactionData: transactionData.withAttributesToPost(subscriberAttributesToPost),
```
tonidero, line 110: So we are recreating the transaction data here since the `consentStatus` could be added to the attributes correct? If so, I wonder if we could extract that piece of code adding that attribute to the moment where we create the `PurchasedTransactionData` to avoid this recreation. Not a blocker though.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -964,41 +988,56 @@ private extension PurchasesOrchestrator {
...
-        ) { transaction, customerInfo, error, cancelled in
-            let completion = self.getAndRemovePurchaseCompletedCallback(forTransaction: purchasedTransaction)
-
-            if let customerInfo = customerInfo {
-                self.customerInfoManager.cache(customerInfo: customerInfo, appUserID: self.appUserID)
-            }
+            data: .init(
+                appUserID: self.appUserID,
+                presentedOfferingID: offeringID,
+                unsyncedAttributes: unsyncedAttributes,
```
tonidero, line 1024: It seems that we were not syncing unsyncedAttributes in some of these posts. I think it shouldn't cause any issues, but it seems to be a behavior change so I just wanted to bring it up.

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -21,16 +21,25 @@ struct PurchaseSource {
...
 
     /// Starts a `PostReceiptDataOperation` for the transaction.
     func handlePurchasedTransaction(
         _ transaction: StoreTransaction,
-        presentedOfferingID: String?,
-        storefront: StorefrontType?,
-        source: PurchaseSource,
-        completion: @escaping PurchaseCompletedBlock
+        data: PurchasedTransactionData,
+        completion: @escaping CustomerAPI.CustomerInfoResponseHandler
```
tonidero, line 42: Hmm it's a bit weird to have a reference to `CustomerAPI` in the `TransactionPoster`... I think it's ok though.

tonidero (approved):
Great refactor! Looks great!

Sources/Purchasing/Purchases/TransactionPoster.swift
NachoSoto: `CustomerAPI` also has the `post(receiptData:)` method.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: That was being done inside by `TransactionPoster` because it had access to the unsynced attributes there.

PR #2549, opened 2023-05-25
`PostReceiptOperation`: added ability to also post `AdServices` token

NachoSoto:
### Changes:
- Added `aadAttributionToken` to `PostReceiptDataOperation`
- Exposed `AttributionPoster.adServicesTokenToPostIfNeeded`
- Added snapshot test to verify it's sent
- Added `PurchasesOrchestrator` tests (SK1/SK2) for sending the attribution token
- Added missing `PurchasesOrchestrator` tests for sending attributes
- Added log when marking `AdServices` token as synced
- Exposed `Purchases.attribution` for custom entitlement computation framework (and added to API tester)
- Exposed `Purchases.enableAdServicesAttributionTokenCollection`  for custom entitlement computation framework (and added to API tester)

### DIFF
--- a/Sources/Attribution/AttributionPoster.swift
+++ b/Sources/Attribution/AttributionPoster.swift
@@ -128,15 +128,37 @@ final class AttributionPoster {
     @available(tvOS, unavailable)
     @available(watchOS, unavailable)
     func postAdServicesTokenIfNeeded() {
-        guard latestNetworkIdAndAdvertisingIdentifierSent(network: .adServices) == nil else {
-            return
-        }
-
-        guard let attributionToken = attributionFetcher.adServicesToken else {
-            return
-        }
+        guard let attributionToken = self.adServicesTokenToPostIfNeeded else { return }
 
         self.post(adServicesToken: attributionToken)
+    }
+
+    var adServicesTokenToPostIfNeeded: String? {
+        #if os(tvOS) || os(watchOS)
+        return nil
+        #else
+        guard #available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *) else {
+            return nil
+        }
+
+        guard self.latestNetworkIdAndAdvertisingIdentifierSent(network: .adServices) == nil else {
+            return nil
+        }
+
+        return self.attributionFetcher.adServicesToken
+        #endif
+    }
+
+    @discardableResult
+    func markAdServicesToken(_ token: String, asSyncedFor userID: String) -> [AttributionNetwork: String] {
+        Logger.info(Strings.attribution.adservices_marking_as_synced(appUserID: userID))
+
+        var newDictToCache = self.deviceCache.latestAdvertisingIdsByNetworkSent(appUserID: userID)
+        newDictToCache[AttributionNetwork.adServices] = token
+
+        self.deviceCache.set(latestAdvertisingIdsByNetworkSent: newDictToCache, appUserID: userID)
+
+        return newDictToCache
     }
 
     func postPostponedAttributionDataIfNeeded() {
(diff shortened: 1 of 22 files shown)


Sources/Attribution/AttributionPoster.swift
```diff
@@ -167,9 +177,8 @@ final class AttributionPoster {
         let currentAppUserID = self.currentUserProvider.currentAppUserID
 
         // set the cache in advance to avoid multiple post calls
-        var newDictToCache = self.deviceCache.latestAdvertisingIdsByNetworkSent(appUserID: currentAppUserID)
-        newDictToCache[AttributionNetwork.adServices] = adServicesToken
-        self.deviceCache.set(latestAdvertisingIdsByNetworkSent: newDictToCache, appUserID: currentAppUserID)
+        var newDictToCache = self.subscriberAttributesManager.markAdServicesToken(adServicesToken,
+                                                                                  asSyncedFor: currentAppUserID)
```
aboedo, line 181: feels odd for subs attributes manager to be in charge for this - there's no subs attribute involved, right?

aboedo: I realize that this is piping through deviceCache anyway, but I feel like going directly might be more appropriate in this case - there's no subs attribute involved, but if I were reading this code with no context, I'd assume that the token is stored as a subs attribute

If we want to share the code, I think this could be either a part of DeviceCache itself or its own object, it just feels arbitrary to add it to subs attributes manager since it's not really related

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -489,6 +578,85 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
...
+
+        expect(self.receiptFetcher.receiptDataCalled) == true
+        expect(self.receiptFetcher.receiptDataReceivedRefreshPolicy) == .always
+
+        expect(self.backend.invokedPostReceiptDataCount) == 1
+        expect(self.backend.invokedPostReceiptDataParameters?.aadAttributionToken).to(beNil())
+    }
+
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, *)
+    func testPurchaseSK2PackagePostsAdServicesTokenAndAttributes() async throws {
```
aboedo, line 612: feels like we need another set of 2 tests for "doesn't post the token if it's already been posted, sk1 / sk2"

aboedo: I suppose that's tested through AttributionPoster, though, right?

aboedo (commented):
looks great! Left a couple of small comments

aboedo (commented):
looks great! Left a couple of small comments

Sources/Attribution/AttributionPoster.swift
NachoSoto: Hmm I agree. That's what I had originally, I _think_ I moved this so I could mock it, let me see.

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
NachoSoto: Tested by `testAdServicesTokenToPostIfNeededReturnsNilIfAlreadySent`

Sources/Attribution/AttributionPoster.swift
NachoSoto: It's mainly because we don't mock `AttributionPoster`, and I wanted to verify this is called, but I'll see if I can do it in another way.

NachoSoto: Done

NachoSoto:
Updated this to point to `release/4.19.0`. Will merge there and then make `4.19.1`.

Sources/Attribution/AttributionPoster.swift
```diff
@@ -128,15 +128,37 @@ final class AttributionPoster {
...
+        #if os(tvOS) || os(watchOS)
+        return nil
+        #else
+        guard #available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *) else {
+            return nil
         }
 
-        guard let attributionToken = attributionFetcher.adServicesToken else {
-            return
+        guard self.latestNetworkIdAndAdvertisingIdentifierSent(network: .adServices) == nil else {
```
tonidero, line 144: I've not been involved in the specs so this might be correct, but with this, only the first purchase for a user will send the token right? As in, if a user unsubscribes, the purchase expires and then resubscribes, the second time it won't be sent. Not sure if this token can change, so it might be ok to just send it the first time, but I thought I would mention it.

tonidero (commented):
LGTM, but will leave it to @aboedo to give the final approval since I'm not completely familiar with the requirements here.

Sources/Attribution/AttributionPoster.swift
aboedo: yeah, that should be fine. That's the same behavior we have for adservices in other places. 
And in practice, attribution tokens expire quickly, and a user will only be attributed once

aboedo (approved):
thanks for the fast work here!

NachoSoto:
Post-refactor version: #2566.

PR #2551, opened 2023-05-25
`enableAdServicesAttributionTokenCollection`: added integration test

NachoSoto:
Because `AAAttribution.attributionToken()` isn't available on simulators, we currently had no coverage for this entire part of the codebase.
By mocking the token (on simulator + debug + integration tests only), we can verify posting to the server works correctly.

As a follow-up, we can also verify that posting it as part of the receipt with #2549 works correctly as well.

### DIFF
--- a/Sources/Attribution/AttributionFetcher.swift
+++ b/Sources/Attribution/AttributionFetcher.swift
@@ -85,6 +85,13 @@ class AttributionFetcher {
 #if canImport(AdServices)
         do {
             #if targetEnvironment(simulator)
+            #if DEBUG
+            if let mockToken = ProcessInfo.mockAdServicesToken {
+                Logger.warn(Strings.attribution.adservices_mocking_token(mockToken))
+                return mockToken
+            }
+            #endif
+
                 // See https://github.com/RevenueCat/purchases-ios/issues/2121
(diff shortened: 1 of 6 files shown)


Sources/FoundationExtensions/ProcessInfo+Extensions.swift
```diff
@@ -13,11 +13,14 @@
 
 import Foundation
 
+#if DEBUG
```
NachoSoto, line 16: It's now impossible to read these on release builds.

Sources/Attribution/AttributionFetcher.swift
```diff
@@ -85,6 +85,13 @@ class AttributionFetcher {
 #if canImport(AdServices)
         do {
             #if targetEnvironment(simulator)
+            #if DEBUG
+            if let mockToken = ProcessInfo.mockAdServicesToken {
+                Logger.warn(Strings.attribution.adservices_mocking_token(mockToken))
```
NachoSoto, line 90: Added this so it becomes obvious if we mess it up and mock it when we shouldn't.

tonidero (approved):

aboedo (approved):

PR #2563, opened 2023-05-30
`CustomerInfoManager`: post all unfinished transactions

NachoSoto:
This expands the logic added in #2533 to post not only the first transaction, but all of them.

This is important for cases where there might be a consumable transaction. By handling _all_ those, we ensure that we are able to process offline `CustomerInfo` later on, instead of potentially failing due to a consumable transaction still unfinished.

### DIFF
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -324,11 +324,11 @@ private extension CustomerInfoManager {
                     let storefront = await Storefront.currentStorefront
                     var lastResult: Result<CustomerInfo, BackendError>?
 
-                    for transaction in transactions {
-                        Logger.debug(
-                            Strings.customerInfo.posting_transaction_in_lieu_of_fetching_customerinfo(transaction)
-                        )
+                    Logger.debug(
+                        Strings.customerInfo.posting_transactions_in_lieu_of_fetching_customerinfo(transactions)
+                    )
 
+                    for transaction in transactions {
                         lastResult = await self.transactionPoster.handlePurchasedTransaction(
                             transaction,
                             data: .init(appUserID: appUserID,
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -325,7 +325,9 @@ private extension CustomerInfoManager {
                     var lastResult: Result<CustomerInfo, BackendError>?
 
                     for transaction in transactions {
-                        Logger.debug(Strings.customerInfo.posting_transaction_in_lieu_of_fetching_customerinfo(transaction))
+                        Logger.debug(
+                            Strings.customerInfo.posting_transaction_in_lieu_of_fetching_customerinfo(transaction)
+                        )
 
                         lastResult = await self.transactionPoster.handlePurchasedTransaction(
                             transaction,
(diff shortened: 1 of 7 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1136,23 +1136,23 @@ extension PurchasesOrchestrator {
...
-                    appUserID: self.appUserID,
-                    presentedOfferingID: offeringID,
-                    unsyncedAttributes: unsyncedAttributes,
-                    storefront: storefront,
-                    source: .init(isRestore: self.allowSharingAppStoreAccount,
-                                  initiationSource: initiationSource)
-                )
-            ) { result in
-                self.handlePostReceiptResult(result, subscriberAttributes: unsyncedAttributes)
+        let result = await self.transactionPoster.handlePurchasedTransaction(
```
NachoSoto, line 1139: This is using the new `async` overload to simplify it.

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
```diff
@@ -304,6 +304,40 @@ class OfflineStoreKit1IntegrationTests: BaseOfflineStoreKitIntegrationTests {
         try await self.verifyEntitlementWentThrough(info2)
     }
 
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func testPurchasingMultipleProductsWhileServerIsDownHandlesAllTransactionsWhenForegroundingApp() async throws {
```
NachoSoto, line 308: New test 🤪 /cc @aboedo

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -319,20 +319,28 @@ private extension CustomerInfoManager {
...
+                if !transactions.isEmpty {
+                    var lastResult: Result<CustomerInfo, BackendError> = .failure(.missingCachedCustomerInfo())
+                    let storefront = await Storefront.currentStorefront
+
+                    Logger.debug(
+                        Strings.customerInfo.posting_transactions_in_lieu_of_fetching_customerinfo(transactions)
                     )
+
+                    for transaction in transactions {
+                        lastResult = await self.transactionPoster.handlePurchasedTransaction(
```
tonidero, line 333: Hmm I think this way we would override the previous result even if it was an error right? So if the last request succeeded but any of the previous ones failed, this would return success.
Should we post the first error if any? So the logic I think would be best would be:
- If any post fails, keep posting but keep track of the errors.
- Once all post end, if there were no errors, continue with the last success, or with the first error if any.

tonidero (commented):
One comment about how to handle partial errors. Looks good otherwise

Sources/Identity/CustomerInfoManager.swift
NachoSoto: > Once all post end, if there were no errors, continue with the last success, or with the first error if any.

Considering that every post will post the entirety of the receipt, I think it would be best to be more optimistic with the results. This comes from calling `getCustomerInfo`. If we could post any of these, and get a `CustomerInfo`, it would be up-to-date.

I'll update the logic to pick any of the successes instead of "randomly" the last one.

tonidero: Ahhh right, it's different in iOS true. Then yeah, that makes sense 👍

NachoSoto: Done: f9ed157599c60f8912e3a4fa684270d33e9acd42

tonidero (approved):
Looks good!

PR #1794, opened 2022-08-05
`SK2StoreProduct`: conditionally removed `@available` workaround

NachoSoto:
Looks like this is fixed in Xcode 14.0.
`Any` is not `Sendable`, so we need to remove this workaround for [CSDK-379].

[CSDK-379]: https://revenuecats.atlassian.net/browse/CSDK-379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #2565, opened 2023-05-30
`SK2StoreProduct`: fix crash on iOS 12

NachoSoto:
Even though https://bugs.swift.org/browse/SR-15825 is supposedly fixed on Xcode 14.x, and we validated that on CI (#1794), that's still crashing for some reason on iOS 12 devices.
Fixes https://github.com/RevenueCat/purchases-unity/issues/278, [SDKONCALL-267], SDK-3154.

[SDKONCALL-267]: https://revenuecats.atlassian.net/browse/SDKONCALL-267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Sources/Misc/Box.swift
+++ b/Sources/Misc/Box.swift
@@ -31,3 +31,5 @@ final class WeakBox<T: AnyObject> {
     init(_ value: T) { self.value = value }
 
 }
+
+extension Box: Sendable where T: Sendable {}
--- a/Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
@@ -17,22 +17,15 @@ import StoreKit
 internal struct SK2StoreProduct: StoreProductType {
 
     init(sk2Product: SK2Product) {
-        #if swift(<5.7)
         self._underlyingSK2Product = .init(sk2Product)
-        #else
-        self.underlyingSK2Product = sk2Product
-        #endif
     }
 
-    #if swift(<5.7)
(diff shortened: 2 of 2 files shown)


Sources/Misc/Box.swift
```diff
@@ -31,3 +31,5 @@ final class WeakBox<T: AnyObject> {
     init(_ value: T) { self.value = value }
 
 }
+
+extension Box: Sendable where T: Sendable {}
```
NachoSoto, line 35: @aboedo this is all that was needed :)

aboedo (approved):
awesome!

Sources/Misc/Box.swift
ncooke3: Hi @NachoSoto, do you recall _why_ this did the trick? We're seeing a very similar crash in the firebase-ios-sdk and I'm thinking this solution will also work for our "Box" type.

NachoSoto: You mean the `Sendable` part, or `Box` in general?

ncooke3: The `Sendable` part

NachoSoto: That wasn't fixing a crash, that was necessary to use this type inside another Sendable type.

ncooke3: Oh, my bad. I re-read the issue and see that now. Sorry for the pings.

NachoSoto: No problem!

PR #2567, opened 2023-05-30
Introduced `debugRevenueCatOverlay()`: new SwiftUI debug overlay

NachoSoto:
simulator_screenshot_DBF551B9-3BA8-440D-9B3F-16DB1D47A0A7

### DIFF
--- a/Sources/Support/DebugUI/DebugContentViews.swift
+++ b/Sources/Support/DebugUI/DebugContentViews.swift
@@ -11,11 +11,11 @@
 //
 //  Created by Nacho Soto on 5/30/23.
 
-#if DEBUG && canImport(SwiftUI) && !os(macOS)
+#if DEBUG && os(iOS)
 
 import SwiftUI
 
-@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+@available(iOS 16.0, *)
 struct DebugSwiftUIRootView: View {
 
     @StateObject
--- a/Sources/Support/DebugUI/DebugContentViews.swift
+++ b/Sources/Support/DebugUI/DebugContentViews.swift
@@ -94,6 +94,7 @@ private struct DebugSummaryView: View {
                 Text("Loading...")
 
             case let .loaded(config):
+                LabeledContent("SDK version", value: SystemInfo.frameworkVersion)
                 LabeledContent("Observer mode", value: config.observerMode.description)
                 LabeledContent("Sandbox", value: config.sandbox.description)
                 LabeledContent("StoreKit 2", value: config.storeKit2Enabled ? "on" : "off")
--- a/Sources/Support/DebugUI/DebugContentViews.swift
+++ b/Sources/Support/DebugUI/DebugContentViews.swift
@@ -1,0 +1,222 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  DebugViewContent.swift
+//
+//  Created by Nacho Soto on 5/30/23.
+
+#if DEBUG && canImport(SwiftUI) && !os(macOS)
+
+import SwiftUI
+
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+struct DebugSwiftUIRootView: View {
+
+    @StateObject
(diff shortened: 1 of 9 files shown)


Sources/Support/DebugUI/DebugViewSheetPresentation.swift
```diff
@@ -0,0 +1,65 @@
...
+
+#if DEBUG && canImport(SwiftUI) && !os(macOS)
+
+import SwiftUI
+
+@available(iOS 16.0, tvOS 16.0, *)
+extension View {
+
+    @ViewBuilder
+    func bottomSheet(
```
NachoSoto, line 22: I took this from my own app. It's necessary because `SwiftUI`'s sheets still don't expose everything that UIKit has. Maybe in iOS 17?

Sources/Support/DebugUI/DebugContentViews.swift
```diff
@@ -0,0 +1,222 @@
...
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+struct DebugSwiftUIRootView: View {
+
+    @StateObject
+    private var model = DebugViewModel()
+
+    var body: some View {
+        NavigationStack(path: self.$model.navigationPath) {
+            DebugSummaryView(model: self.model)
+                .navigationDestination(for: DebugViewPath.self) { path in
```
NachoSoto, line 27: New SwiftUI navigation API 🤓

NachoSoto:
Currently stuck on this issue: https://developer.apple.com/forums/thread/724068

NachoSoto:
This is ready.

Sources/Support/DebugUI/DebugContentViews.swift
```diff
@@ -0,0 +1,255 @@
...
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  DebugViewContent.swift
+//
+//  Created by Nacho Soto on 5/30/23.
+
+#if DEBUG && os(iOS) && swift(>=5.8)
```
tonidero, line 14: Should we support macos as well? It probably shouldn't be a bottom sheet though... We can think about it and add in a future PR though

Sources/Support/DebugUI/DebugViewModel.swift
```diff
@@ -0,0 +1,130 @@
...
+
+import Foundation
+
+#if DEBUG && os(iOS) && swift(>=5.8)
+
+import SwiftUI
+
+@MainActor
+@available(iOS 16.0, *)
+final class DebugViewModel: ObservableObject {
```
tonidero, line 22: Should we exclude these new files from coverage?

Sources/Support/DebugUI/DebugViewSheetPresentation.swift
```diff
@@ -0,0 +1,48 @@
...
+        presentationDetents: Set<PresentationDetent>,
+        isPresented: Binding<Bool>,
+        largestUndimmedIdentifier: PresentationDetent = .large,
+        cornerRadius: CGFloat,
+        interactiveDismissDisabled: Bool = false,
+        @ViewBuilder content: @escaping () -> some View
+    ) -> some View {
+        self
+            .sheet(isPresented: isPresented) {
```
tonidero, line 31: I'm not familiar with iOS UI development 😅 . So just to check, if `isPresented` is false, is the view created? I just have some concerns of loading the debug view model triggering some extra operations in the SDK... I don't think it should be a big issue right now though.

tonidero (approved):
I'm not very familiar with iOS UI development but looks great to me! Will approve it tentatively.

Sources/Support/DebugUI/DebugViewModel.swift
NachoSoto: I want to write UI snapshot tests for this since it's a public facing feature :)

Sources/Support/DebugUI/DebugViewSheetPresentation.swift
NachoSoto: Good point let me check

NachoSoto: Confirmed :) it doesn't load until it's visible.

Sources/Support/DebugUI/DebugViewModel.swift
NachoSoto: Done: #2630

PR #2582, opened 2023-06-05
`Dictionary.MergeStrategy`: fixed Xcode 15 compilation

NachoSoto:
> Generic parameter 'Value' shadows generic parameter from outer scope with the same name; this is an error in Swift 6


joshdholtz (approved):
LGTM 💪

PR #2586, opened 2023-06-06
Fixed `Catalyst` build with Xcode 15 beta 1

NachoSoto:
Fixes this error:
> Property cannot be more available than enclosing scope
> Enclosing scope requires availability of application extensions for Mac Catalyst 13.1 or newer

### DIFF
--- a/Sources/FoundationExtensions/UIApplication+RCExtensions.swift
+++ b/Sources/FoundationExtensions/UIApplication+RCExtensions.swift
@@ -18,6 +18,7 @@ import UIKit
 extension UIApplication {
 
     @available(iOS 13.0, *)
+    @available(macCatalystApplicationExtension 13.1, *)
     @available(macOS, unavailable)
     @available(watchOS, unavailable)
     @available(watchOSApplicationExtension, unavailable)


Sources/FoundationExtensions/UIApplication+RCExtensions.swift
```diff
@@ -18,6 +18,7 @@ import UIKit
 extension UIApplication {
 
     @available(iOS 13.0, *)
+    @available(macCatalystApplicationExtension 13.1, *)
```
tonidero, line 21: Hmm should this also be available in `macCatalyst` (not an application extension)?

tonidero (approved):
Just a question, but looks good!

Sources/FoundationExtensions/UIApplication+RCExtensions.swift
NachoSoto: `iOS` already satisfies that requirement implicitly (`iOS` is basically `macCatayst` too unless you add a specific `@available` for it).
Which means that this is available for `macCatalyst 13.0`. I have no idea why Apple changed the requirement here in Xcode 15, but either way this property is only used from places that require iOS 15, which is why everything else comiles.

PR #2589, opened 2023-06-06
`StoreKitObserverModeIntegrationTests`: fixed and disabled SK2 `testPurchaseInDevicePostsReceipt`

NachoSoto:
Turns out this test was only passing because it was waiting for the receipt to be posted during a renewal, not the initial purchase!
This updates the test to reflect the correct expectation (disabling renewals), and added the `XCTExpectFailure` so we can detect if it starts passing in a future iOS version.

Also changed the test to not use `customerInfo()` because of #2563.


NachoSoto:
This is passing locally :( (correctly ignoring the error)

PR #2591, opened 2023-06-06
`CI`: added `iOS 17` job

NachoSoto:
https://discuss.circleci.com/t/xcode-15-beta-1-released/48298

### Depends on :
- #2603
- #2602
- #2605
- #2606
- #2607

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -315,6 +315,32 @@ jobs:
           command: swift build -c release --target ReceiptParser
           no_output_timeout: 30m
 
+  run-test-ios-17:
+    <<: *base-job
+    steps:
+      - checkout
+      - install-dependencies
+      - run:
+          name: Run tests
+          command: bundle exec fastlane test_ios
+          no_output_timeout: 5m
+          environment:
+            SCAN_DEVICE: iPhone 14 (17.0)
+      - compress_result_bundle:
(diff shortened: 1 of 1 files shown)


NachoSoto:
I'd suggest merging that even though we still have some failures, so we can keep an eye until it's all resolved (I have a bunch of open PRs that solve them).

.circleci/config.yml
```diff
@@ -315,6 +315,32 @@ jobs:
...
+          condition: << pipeline.parameters.generate_snapshots >>
+          steps:
+            - run:
+                name: Run create_snapshot_pr 
+                command: bundle exec fastlane create_snapshot_pr version:"ios-17"
+      - store_test_results:
+          path: fastlane/test_output
+      - store_artifacts:
+          path: fastlane/test_output/xctest
+          destination: scan-test-output
```
tonidero, line 342: Should we extract this to be a command? Seems like it's almost repeated (with some small differences) for the other iOS versions... We can do that in a different PR though

tonidero (approved):
I'm a bit concerned about having tests not passing... even if it's to monitor them... Should we expect failures of them instead so we notice when they start passing? Though please ignore if tests are already passing with other PRs!

NachoSoto:
> Should we expect failures of them instead so we notice when they start passing

That's a good idea, is that a CircleCI thing?

PR #2593, opened 2023-06-07
Created `PaywallExtensions`: `StoreView` and `SubscriptionStoreView` overloads for `Offering`

NachoSoto:
Example:
```swift
struct App: View {

    var offering: Offering

    var body: some View {
        Content()
            .sheet(isPresented: self.$showingSubscriptionSheet) {
                SubscriptionStoreView(offering: self.offering)
            }
    }

}
```


Sources/Support/PaywallExtensions.swift
```diff
@@ -0,0 +1,153 @@
...
+#if swift(>=5.9)
+
+@available(iOS 17.0, macOS 14.0, tvOS 17.0, watchOS 10.0, *)
+extension StoreView {
+
+    /// Creates a view to load a collection of products from the App Store, and merchandise them.
+    /// When the user purchases products through this paywall, the `RevenueCat` SDK will handle
+    /// the result automatically. All you need to do is to dismiss the paywall
+    public init(
+        offering: Offering,
```
tonidero, line 26: This is probably slightly more complex but I was thinking, could we provide another override to a view that actually does the `getOfferings` call and displayes the `StoreView`/`SubscriptionStoreView` once those have loaded? The complex think about that is that we would need to write loading and error UI states ourselves (would need design input). Good thing is that it would make it even easier to use RevenueCat (with this you still would need to use `getOfferings` first and pass the current offering). CC @aboedo

Sources/Support/PaywallExtensions.swift
```diff
@@ -0,0 +1,153 @@
...
+        prefersPromotionalIcon: Bool = false
+    ) where Icon == EmptyView, PlaceholderIcon == EmptyView {
+        self.init(
+            ids: offering.allProductIdentifiers,
```
tonidero, line 30: I wonder... would these products be ordered by the iOS UI? Just it would be great to avoid showing the yearly price above the monthly for example. 

If it doesn't and it depends on the order of the product ids passed maybe we should try to make that customizable.

NachoSoto: Oh good point! In fact I noticed the ordering wasn't stable. I'll sort by duration 👍🏻

tonidero: I think that works for a V1. For the future, it would be great to allow devs to customize the order (so we don't limit what they can do with the iOS widget), but that's probably not trivial, so I think we can leave that for later

Sources/Support/PaywallExtensions.swift
NachoSoto: That's what `CurrentOfferingSubscriptionStoreView` does below 😄

Sources/Support/PaywallExtensions.swift
NachoSoto: They can always use the paywall API directly in that case :)

NachoSoto: Sorting done 👍🏻

Sources/Support/PaywallExtensions.swift
tonidero: 🤦 I got confused, but yeah that's what I meant 🙌

Sources/Support/PaywallExtensions.swift
```diff
@@ -0,0 +1,165 @@
...
+            if let currentOffering {
+                if let marketingContent {
+                    SubscriptionStoreView(offering: currentOffering,
+                                          marketingContent: marketingContent)
+                } else {
+                    SubscriptionStoreView(offering: currentOffering)
+                }
+            } else {
+                ProgressView()
+                    .progressViewStyle(.circular)
```
tonidero, line 123: 2 questions:
- Will the `ProgressView` use the full available space? Or more importantly, will the transition between this and the `SubscriptionStoreView` once it's loaded look good?
- Did you/should we get a design review on this?

Sources/Support/PaywallExtensions.swift
```diff
@@ -0,0 +1,165 @@
...
+            }
+        }
+        .task {
+            if let offering = try? await Purchases.shared.offerings().current {
```
tonidero, line 127: So if getting offerings fail, we just keep the loading state? I think we should show some error UI. Thoughts?

Sources/Support/PaywallExtensions.swift
```diff
@@ -0,0 +1,165 @@
...
+
+    // swiftlint:disable:next missing_docs
+    public var body: some View {
+        Group {
+            if let currentOffering {
+                if let marketingContent {
+                    SubscriptionStoreView(offering: currentOffering,
+                                          marketingContent: marketingContent)
+                } else {
+                    SubscriptionStoreView(offering: currentOffering)
```
tonidero, line 119: A couple more questions about these views:
- When purchasing through this view, does it automatically finish the transaction? Or does the developer still have to do it?
- If the user purchases an offering through this view, we won't be getting the `presentedOfferingIdentifier` setup in the SDK... We need to figure out a way to set that.

tonidero: Will add to this, that anything we do in the `purchasePackage/Product` methods won't happen when they use this view. We should check if there are any other unintended side effects of this. Another one I see is that we would not be calling `preventPurchasePopupCallFromTriggeringCacheRefresh` so we would never set the offerings cache as stale I think.

Sources/Support/PaywallExtensions.swift
NachoSoto: Yeah this is very WIP. 
I'm going to extract `CurrentOfferingSubscriptionStoreView` into a separate PR so we can merge the rest, and handle this and the offering failure separately.

Sources/Support/PaywallExtensions.swift
NachoSoto: > When purchasing through this view, does it automatically finish the transaction? Or does the developer still have to do it?

Our SDK handles that :) (though we do need SDK-3170)

> If the user purchases an offering through this view, we won't be getting the presentedOfferingIdentifier setup in the SDK... We need to figure out a way to set that.

Oh great call. Just filed SDK-3176 for that.

NachoSoto: > Will add to this, that anything we do in the purchasePackage/Product methods won't happen when they use this view. We should check if there are any other unintended side effects of this.

I think that's fine. This is basically like our SDK acting in observer mode for a purchase being done from outside.

> Another one I see is that we would not be calling `preventPurchasePopupCallFromTriggeringCacheRefresh` so we would never set the offerings cache as stale I think.

Another reason to do SDK-3172 and get rid of that method.

NachoSoto:
@tonidero updated this to leave just the overloads, and listed the 3 tasks that need to be done (working on those today)
This is just overloads for SwiftUI so they don't have to wait for those fixes.

PR #2595, opened 2023-06-07
`debugRevenueCatOverlay`: added ability to display new `SubscriptionStoreView`

NachoSoto:
See #2593

Simulator Screenshot - iPhone 14 Pro Max - 2023-06-07 at 13 07 58

### DIFF
--- a/Sources/Support/DebugUI/DebugContentViews.swift
+++ b/Sources/Support/DebugUI/DebugContentViews.swift
@@ -13,6 +13,7 @@
 
 #if DEBUG && os(iOS) && swift(>=5.8)
 
+import StoreKit
 import SwiftUI
 
 @available(iOS 16.0, *)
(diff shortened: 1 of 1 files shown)


Sources/Support/DebugUI/DebugContentViews.swift
```diff
@@ -175,10 +193,70 @@ private struct DebugOfferingView: View {
...
+                    Text("RevenueCat Demo Paywall")
+                }
+                .font(.title)
+
+                Text(self.offering.getMetadataValue(for: "title", default: "Premium Access"))
+                    .font(.title2)
+                    .foregroundStyle(.primary)
+
+                Text(self.offering.getMetadataValue(for: "subtitle",
+                                                    default: "Unlimited access to premium content."))
```
joshdholtz, line 244: 😍

joshdholtz (approved):

Sources/Support/DebugUI/DebugContentViews.swift
```diff
@@ -159,9 +160,26 @@ private struct DebugSummaryView: View {
 @available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
 private struct DebugOfferingView: View {
 
+    @State private var showingSubscriptionSheet = false
+    @State private var showingStoreSheet = false
```
tonidero, line 164: I wonder if this could be an enum (since we can only show one of the 2 states or none right?)

tonidero (approved):

Sources/Support/DebugUI/DebugContentViews.swift
NachoSoto: `SwiftUI` could handle having more than one sheet :)

tonidero: Oh interesting! That's fine then

PR #2600, opened 2023-06-08
Refactor: extracted all log strings

NachoSoto:
I'm working on improving our logging setup, which will be made easier if we don't use hardcoded strings in our logs.


RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -3479,7 +3478,6 @@
 				574A2F4F282D7B9E00150D40 /* PostOfferDecodingTests.swift in Sources */,
 				2DDF41CE24F6F4C3005BC22D /* InAppPurchaseBuilderTests.swift in Sources */,
 				573A10DB2800AF4700F976E5 /* PurchaseErrorTests.swift in Sources */,
-				B300E4BF26D436F900B22262 /* LogIntent.swift in Sources */,
```
NachoSoto, line 3482: This was unnecessarily being compiled for the `Unit Tests` target as well.

Sources/Logging/Strings/DiagnosticsStrings.swift
```diff
@@ -0,0 +1,36 @@
...
+//
+//  DiagnosticsStrings.swift
+//
+//  Created by Nacho Soto on 6/8/23.
+
+import Foundation
+
+// swiftlint:disable identifier_name
+
+enum DiagnosticsStrings {
```
NachoSoto, line 18: I started this, which will grow more with the upcoming diagnostics.

Sources/Logging/Strings/RestoreStrings.swift
```diff
@@ -1,35 +0,0 @@
...
-//
-//  Created by Tina Nguyen on 12/11/20.
-//
-
-import Foundation
-
-enum RestoreStrings {
-
-    // swiftlint:disable identifier_name
-    case restorepurchases_called_with_allow_sharing_appstore_account_false_warning
```
NachoSoto, line 20: This only had one string, so I combined it in `PurchaseStrings`.

Sources/Logging/Strings/PurchaseStrings.swift
```diff
@@ -76,6 +76,10 @@ enum PurchaseStrings {
     case sk2_transactions_update_received_transaction(StoreTransaction)
     case sk1_purchase_too_slow
     case sk2_purchase_too_slow
+    case payment_queue_wrapper_delegate_call_sk1_enabled
+
+    // swiftlint:disable:next identifier_name
+    case restorepurchases_called_with_allow_sharing_appstore_account_false
```
tonidero, line 82: Nitpick and it was like this before but should this be `restore_purchases_called_with_allow_sharing_appstore_account_false`

tonidero (approved):

Sources/Logging/Strings/PurchaseStrings.swift
NachoSoto: I think it was like this because the method is `restorePurchases`, so this is one word to signal that.

PR #2605, opened 2023-06-08
Changed tests to work around `URL` decoding differences in `iOS 17`

NachoSoto:
The "invalid" test URLs we were using are no longer invalid on iOS 17. This fixes those tests by using an `Int` instead of a `String`.

image

### DIFF
--- a/Tests/UnitTests/FoundationExtensions/DecoderExtensionTests.swift
+++ b/Tests/UnitTests/FoundationExtensions/DecoderExtensionTests.swift
@@ -89,7 +89,7 @@ class DecoderExtensionsIgnoreErrorsTests: TestCase {
     }
 
     func testIgnoresErrors() throws {
-        let json = "{\"url\": \"not a! valid url@\"}"
+        let json = "{\"url\": 1}"
         let data = try Data.decode(json)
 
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/FoundationExtensions/DecoderExtensionTests.swift
```diff
@@ -89,7 +89,7 @@ class DecoderExtensionsIgnoreErrorsTests: TestCase {
     }
 
     func testIgnoresErrors() throws {
-        let json = "{\"url\": \"not a! valid url@\"}"
```
tonidero, line 92: Hmm so it accepts this as a URL? 🤯 

I wonder if we should add back this test (passing) so we get notified if they revert this change in future betas...

tonidero (approved):
Just a suggestion... But we can get this merged to get tests to pass.

Tests/UnitTests/FoundationExtensions/DecoderExtensionTests.swift
NachoSoto: I mean it doesn't really matter 😅 these tests were just to verify that we handle parsing errors, which we're still testing.

PR #2607, opened 2023-06-08
`Encodable.jsonEncodedData`: fixed tests on iOS 17 due to inconsistent key ordering

NachoSoto:
A bunch of tests were failing on Xcode 15 due to [a change in Swift](https://github.com/apple/swift/issues/66466), because we relying on being able to verify that the encoded `Data` saved to caches is consistent.
Starting on Swift 5.9 `JSONEncoder` has optimizations that mean we can't rely on it.

To fix this, this PR does 2 things:
- Changed `JSONEncoder.prettyPrinted` to have `.sortedKeys`
- Changed test to verify the decoded response `struct`, not the raw encoded `Data`

Note that I didn't make `JSONEncoder.default` use `.sortedKeys` because of the potential performance impact. `Encodable.prettyPrintedData` is used in tests like `ReceiptFetcherTests` (which this PR fixes) and for debug purposes, but `Encodable.encodedJSON` is used everywhere.

### DIFF
--- a/Sources/LocalReceiptParsing/DataConverters/Codable+Extensions.swift
+++ b/Sources/LocalReceiptParsing/DataConverters/Codable+Extensions.swift
@@ -39,6 +39,8 @@ extension Encodable {
         }
     }
 
+    /// - Note: beginning with iOS 17, the output of this is not guaranteed to be consistent due to key ordering.
+    /// For tests, it's better to compare `prettyPrintedData` which does sort keys.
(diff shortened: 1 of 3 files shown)


tonidero (approved):

NachoSoto:
@tonidero mind taking another look? I reworked this PR to actually fix the issue, now that I got a response and we know it's not really a regression.

tonidero (approved):
Looks good!

PR #2611, opened 2023-06-09
`StoreTransaction`: read `Storefront` from `StoreKit.Transaction`

NachoSoto:
This was added in iOS 17, and it's potentially more accurate than reading `Storefront.currentStorefront` later.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -824,7 +824,7 @@ extension PurchasesOrchestrator: StoreKit2TransactionListenerDelegate {
         _ listener: StoreKit2TransactionListener,
         updatedTransaction transaction: StoreTransactionType
     ) async throws {
-        let storefront = await Storefront.currentStorefront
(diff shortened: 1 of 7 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -824,7 +824,7 @@ extension PurchasesOrchestrator: StoreKit2TransactionListenerDelegate {
         _ listener: StoreKit2TransactionListener,
         updatedTransaction transaction: StoreTransactionType
     ) async throws {
-        let storefront = await Storefront.currentStorefront
+        let storefront = await transaction.storefront ??? (await Storefront.currentStorefront)
```
tonidero, line 827: FFTI I wonder if we should move this inside the `transaction.storefront` calculation, in case we ever need it somewhere else.

tonidero (approved):

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: Yeah let me put that in an extension 👍🏻

PR #2590, opened 2023-06-06
`StoreKitObserverModeIntegrationTests`: added test for posting renewals

NachoSoto:
Follow up to #2589.
Depends on #2612.

PR #2612, opened 2023-06-09
Always initialize `StoreKit2TransactionListener` even on SK1 mode

NachoSoto:
This is necessary for the new paywall screens to work.
Tests for this added in #2590.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -130,8 +130,9 @@ final class PurchasesOrchestrator {
         storeKit2TransactionListener.delegate = self
         storeKit2StorefrontListener.delegate = self
 
+        storeKit2TransactionListener.listenForTransactions()
+
         if systemInfo.storeKit2Setting == .enabledForCompatibleDevices {
-            storeKit2TransactionListener.listenForTransactions()
             storeKit2StorefrontListener.listenForStorefrontChanges()
         }
     }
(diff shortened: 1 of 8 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -130,8 +130,9 @@ final class PurchasesOrchestrator {
         storeKit2TransactionListener.delegate = self
         storeKit2StorefrontListener.delegate = self
 
+        storeKit2TransactionListener.listenForTransactions()
```
tonidero, line 133: Is there any danger of this sending the same transaction multiple times and/or with different data to our backend?

tonidero (commented):
Code looks good, but this seems like a change we should be careful with so will hold on approving until @aboedo has a chance to look at it.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: The only danger is increased load. @aboedo and I talked about this yesterday.
#2590 adds a test that verifies this particular scenario and checks that we only post once.

aboedo: Yeah, definitely a compromise. Our system is pretty good at de-duping post receipts that are in flight, but there's always a chance that we get slightly delayed signals from SK1 and SK2 and the response from one gets back before we post the other one and we do a double post. 

But on the other hand, that seems like a better scenario than if we miss a purchase in observer mode because we were observing the wrong object.

Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
```diff
@@ -24,20 +24,26 @@ protocol StoreKit2TransactionListenerDelegate: AnyObject {
 
 }
 
+/// Observes `StoreKit.Transaction.updates`, which receives:
+/// - Updates from outside `Product.purchase()`, like renewals
```
aboedo, line 28:
```suggestion
/// - Updates from outside `Product.purchase()`, like renewals and purchases made on other devices
```

aboedo (approved):

PR #2616, opened 2023-06-09
`StoreKitConfigTestCase`/`BaseStoreKitIntegrationTests`: also clear transactions after every test

NachoSoto:
Trying to work around `FB11799675`, this was suggested by @joshdholtz and it seems to be working really well. This now calls `deleteAllTransactions` and `SKTestSession.clearTransactions` _after_ every test, not only before.

The problem is that `clearTransactions` only seems to delete stuff from `SKTestSession.allTransactions`, but not from `StoreKit.Transaction.unfinished`. Doing this at the end of every test seems to do a much better job at ensuring that each test invocation has no leftover transaction, hopefully leading to fewer flaky failures.


joshdholtz (approved):
LGTM 💪

PR #2619, opened 2023-06-09
Added `.xcprivacy` for Xcode 15

NachoSoto:
Example generated report:

Screenshot 2023-06-09 at 15 16 11

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -36,13 +36,22 @@ let package = Package(
     targets: [
         .target(name: "RevenueCat",
                 path: "Sources",
-                exclude: ["Info.plist", "LocalReceiptParsing/ReceiptParser-only-files"]),
+                exclude: ["Info.plist", "LocalReceiptParsing/ReceiptParser-only-files"],
+                resources: [
+                    .copy("../Sources/PrivacyInfo.xcprivacy")
+                ]),
         .target(name: "RevenueCat_CustomEntitlementComputation",
                 path: "CustomEntitlementComputation",
                 exclude: ["Info.plist", "LocalReceiptParsing/ReceiptParser-only-files"],
+                resources: [
+                    .copy("PrivacyInfo.xcprivacy")
+                ],
                 swiftSettings: [.define("ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION")]),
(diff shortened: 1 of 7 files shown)


Package.swift
```diff
@@ -36,13 +36,22 @@ let package = Package(
     targets: [
         .target(name: "RevenueCat",
                 path: "Sources",
-                exclude: ["Info.plist", "LocalReceiptParsing/ReceiptParser-only-files"]),
+                exclude: ["Info.plist", "LocalReceiptParsing/ReceiptParser-only-files"],
+                resources: [
+                    .copy("../Sources/PrivacyInfo.xcprivacy")
```
NachoSoto, line 41: This kind of breaks (cc @neonacho)
Screenshot 2023-06-09 at 15 40 13

/cc @

jefago:
Here are our related docs. It really is a bit tricky because a lot depends on how the developer is using RevenueCat. Theoretically we should just mark "Purchase history" as being collected: https://www.revenuecat.com/docs/apple-app-privacy

NachoSoto:
Okay just updated it to only purchase history.

Package.swift
NachoSoto: I'll leave this for now since it's the only way to add it.

jefago (approved):

Package.swift
```diff
@@ -36,13 +36,22 @@ let package = Package(
...
                 exclude: ["Info.plist", "LocalReceiptParsing/ReceiptParser-only-files"],
+                resources: [
+                    .copy("PrivacyInfo.xcprivacy")
+                ],
                 swiftSettings: [.define("ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION")]),
         .target(name: "ReceiptParser",
-                path: "LocalReceiptParsing"),
+                path: "LocalReceiptParsing",
+                resources: [
+                    .copy("../Sources/PrivacyInfo.xcprivacy")
```
tonidero, line 53: Hmm does the receipt parser also use UserIds?

tonidero (approved):

Package.swift
NachoSoto: Oops good catch, let me remove it from this one.

PR #2622, opened 2023-06-11
`PurchaseTester`: make CI job always point to current version

NachoSoto:
Not sure how we didn't notice this before, but the build we were submitting wasn't being updated! It was stuck at 4.17.0 🤦🏻


Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcodeproj/project.pbxproj
```diff
@@ -971,8 +971,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/RevenueCat/purchases-ios";
 			requirement = {
-				kind = upToNextMajorVersion;
```
NachoSoto, line 974: Oh I guess depending on when the tag was created this could have found the last release. But this PR will be better, allowing us to use the job from any branch too.

tonidero (approved):

PR #2623, opened 2023-06-12
Removed `preventPurchasePopupCallFromTriggeringCacheRefresh`, update caches on `willEnterForeground`

NachoSoto:
This method was a hack to deal with apps getting `didBecomeActive` notifications in the middle of a purchase when the popups were displayed.
This had several problems:
- It was only being used for StoreKit 1 purchases
- It wasn't used for StoreKit 2 purchases through the SDK
- It wasn't used for purchases initiated outside the SDK (like using the upcoming paywall screens)
- It updated the cache timestamp to "now", including offerings, despite not actually updating data
- It was an indirect and implicit way of solving a problem instead of an explicit one

This solves the same by observing `willEnterForeground`, which only happens when the app actually comes back from the background, and not when it simply loses focus.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -361,7 +361,6 @@ final class PurchasesOrchestrator {
         }
 
         payment.applicationUsername = self.appUserID
-        self.preventPurchasePopupCallFromTriggeringCacheRefresh(appUserID: self.appUserID)
 
         self.cachePresentedOfferingIdentifier(package: package, productIdentifier: productIdentifier)
 
@@ -1054,11 +1053,6 @@ private extension PurchasesOrchestrator {
                 }
             }
         }
-    }
-
-    func preventPurchasePopupCallFromTriggeringCacheRefresh(appUserID: String) {
-        deviceCache.setCacheTimestampToNowToPreventConcurrentCustomerInfoUpdates(appUserID: appUserID)
-        deviceCache.setOfferingsCacheTimestampToNow()
     }
 
     func purchase(
--- a/Tests/UnitTests/Purchasing/Purchases/PurchasesDelegateTests.swift
+++ b/Tests/UnitTests/Purchasing/Purchases/PurchasesDelegateTests.swift
@@ -35,11 +35,11 @@ class PurchasesDelegateTests: BasePurchasesTests {
         expect(self.storeKit1Wrapper.delegate).toNot(beNil())
     }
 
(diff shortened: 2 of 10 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -361,7 +361,6 @@ final class PurchasesOrchestrator {
         }
 
         payment.applicationUsername = self.appUserID
-        self.preventPurchasePopupCallFromTriggeringCacheRefresh(appUserID: self.appUserID)
```
NachoSoto, line 364: This is no longer needed, we won't update caches in the middle of a purchase.

aboedo:
I went through the trouble of spelunking through RC history to see when that was added, to see if we were missing anything... 

Jan 2019 🤯  https://github.com/RevenueCat/purchases-ios/commit/64e59c98dc4a2eab48331428e4910746d72054a7

aboedo:
> It was only being used for StoreKit 1 purchases
Could you clarify this? Why would SK2 not trigger didBecomeActive after the modal?

Tests/UnitTests/Purchasing/Purchases/PurchasesDelegateTests.swift
```diff
@@ -35,11 +35,11 @@ class PurchasesDelegateTests: BasePurchasesTests {
         expect(self.storeKit1Wrapper.delegate).toNot(beNil())
     }
 
-    func testSubscribesToUIApplicationDidBecomeActive() throws {
+    func testSubscribesToUIApplicationWillEnterForeground() throws {
         expect(self.notificationCenter.observers).to(haveCount(2))
 
         let (_, _, name, _) = try XCTUnwrap(self.notificationCenter.observers.first)
-        expect(name) == SystemInfo.applicationDidBecomeActiveNotification
+        expect(name) == SystemInfo.applicationWillEnterForegroundNotification
```
aboedo, line 42: maybe we can also test "does not subscribe to app did become active", since doing so would actually introduce a regression where we double-post?

aboedo (approved):
Godspeed, old hack
godspeed

NachoSoto:
> Could you clarify this? Why would SK2 not trigger didBecomeActive after the modal? Or other stuff? They all show the same UI

StoreKit 2 purchases wasn't calling `preventPurchasePopupCallFromTriggeringCacheRefresh`

Tests/UnitTests/Purchasing/Purchases/PurchasesDelegateTests.swift
NachoSoto: 👍🏻  will add.

PR #2626, opened 2023-06-12
Simplified `Purchases.updateAllCachesIfNeeded`

NachoSoto:
This is only called when coming back from the background, so we didn't need to call `isApplicationBackgrounded`.


NachoSoto:
Need to rebase this

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1507,23 +1507,21 @@ private extension Purchases {
         }
     }
 
-    func updateAllCachesIfNeeded() {
-        self.systemInfo.isApplicationBackgrounded { isAppBackgrounded in
```
NachoSoto, line 1511: Changed to a parameter instead.

tonidero (approved):

PR #2635, opened 2023-06-13
Add `InternalDangerousSettings.forceSignatureFailures`

NachoSoto:
Similar to #2486. This will be used for SDK-3181.

```
- WARN: ⚠️ Returning fake signature verification failure for 'GET /v1/subscribers/$RCAnonymousID:fc0508bea4a74a41bfc9a69db30d925f'
- ERROR: 😿‼️ Request failed signature verification. Request to /v1/subscribers/$RCAnonymousID%3Afc0508bea4a74a41bfc9a69db30d925f failed verification
```

PR #2639, opened 2023-06-13
`copy(with: VerificationResult)`: optimization to avoid copies

NachoSoto:
I noticed this could be improved while working on #2635.


Sources/Networking/HTTPClient/HTTPResponse.swift
```diff
@@ -98,6 +98,8 @@ extension HTTPResponse {
     }
 
     func copy(with newVerificationResult: VerificationResult) -> Self {
+        guard newVerificationResult != self.verificationResult else { return self }
```
NachoSoto, line 101: I'm writing tests for this cause we had none.

tonidero (approved):

PR #2645, opened 2023-06-13
`PaywallExtensions`: post purchases with `Offering` identifier

NachoSoto:
### Changes:

- Exposed `internal`-only `Purchases.cachePresentedOfferingIdentifier`
- Changed `SwiftUI` extensions to `static` methods: this is required to be able to call an instance method on them
- Added integration test to verify this offering identifier is used

Technically this is an API change, but these APIs are implicitly beta because they're only available in Xcode 15 beta.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1330,6 +1330,21 @@ extension Purchases: @unchecked Sendable {}
 
 // MARK: Internal
 
+extension Purchases {
+
+    /// Used when purchasing through `SwiftUI` paywalls.
+    func cachePresentedOfferingIdentifier(_ identifier: String, productIdentifier: String) {
+        Logger.debug(Strings.purchase.caching_presented_offering_identifier(offeringID: identifier,
+                                                                            productID: productIdentifier))
+
+        self.purchasesOrchestrator.cachePresentedOfferingIdentifier(
+            identifier,
+            productIdentifier: productIdentifier
+        )
+    }
+
+}
+
 extension Purchases: InternalPurchasesType {
 
     @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
--- a/Sources/Support/PaywallExtensions.swift
+++ b/Sources/Support/PaywallExtensions.swift
@@ -16,41 +16,53 @@ import SwiftUI
 
 #if swift(>=5.9)
 
+// MARK: - StoreView
+
 @available(iOS 17.0, macOS 14.0, tvOS 17.0, watchOS 10.0, *)
 extension StoreView {
-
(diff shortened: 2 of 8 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1329,6 +1329,18 @@ extension Purchases: @unchecked Sendable {}
 
 // MARK: Internal
 
+extension Purchases {
+
+    /// Used when purchasing through `SwiftUI` paywalls.
+    func cachePresentedOfferingIdentifier(_ identifier: String, productIdentifier: String) {
+        self.purchasesOrchestrator.cachePresentedOfferingIdentifier(
```
NachoSoto, line 1336: I should log something here.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1329,6 +1329,18 @@ extension Purchases: @unchecked Sendable {}
...
+            identifier,
+            productIdentifier: productIdentifier
+        )
+    }
```
NachoSoto, line 1340: This is covered by integration tests, but I'll add a unit test too.

Sources/Support/PaywallExtensions.swift
```diff
@@ -16,65 +16,101 @@ import SwiftUI
...
+
+// MARK: - Private
+
+@available(iOS 17.0, macOS 14.0, tvOS 17.0, watchOS 10.0, *)
+private extension View {
+
+    func handlePurchases(_ offering: Offering) -> some View {
+        return self
+            .onInAppPurchaseStart { product in
+                Purchases.shared.cachePresentedOfferingIdentifier(
```
NachoSoto, line 109: I should check if `Purchases.isConfigured` to prevent an unnecessary crash.

Sources/Support/PaywallExtensions.swift
```diff
@@ -16,65 +16,101 @@ import SwiftUI
...
+}
+
+// MARK: - Private
+
+@available(iOS 17.0, macOS 14.0, tvOS 17.0, watchOS 10.0, *)
+private extension View {
+
+    func handlePurchases(_ offering: Offering) -> some View {
+        return self
+            .onInAppPurchaseStart { product in
```
tonidero, line 108: Do we need to clear the cached value on purchase cancellation? 

EDIT: Though thinking about it, it's probably not needed... If the user tries to purchase again with a different offering (super uncommon scenario) it would get set again I guess.

NachoSoto: Are you good with this then?

tonidero (approved):

Sources/Support/PaywallExtensions.swift
tonidero: Yup, sorry for the delay. I thought I had approved it 😅

PR #2647, opened 2023-06-14
`Trusted Entitlements`: added debug log with `ResponseVerificationMode`

NachoSoto:
I was going through some logs, and realized I couldn't tell what mode was configured, and I thought it would be useful.


tonidero (approved):
That's a good point, I will add it as well!

PR #2654, opened 2023-06-14
Support `product_plan_identifier` for purchased subscriptions from `Google Play`

joshdholtz:
### Motivation

Addresses https://github.com/RevenueCat/purchases-flutter/issues/692

### Description

Shows the product id as `<product>:<base_plan>` if the subscription was purchased from Google Play. This uses the `product_plan_identifier` that is returned with the subscription.

### DIFF
--- a/Sources/Identity/CustomerInfo+ActiveDates.swift
+++ b/Sources/Identity/CustomerInfo+ActiveDates.swift
@@ -55,6 +55,10 @@ extension CustomerInfo {
                         .map { productID, subscription in
                             let key: String
                             let value = subscription.expiresDate
+
+                            // Products purchased from Google Play will have a product plan identifier (base plan)
+                            // These products get mapped as "productId:productPlanIdentifier" in the Android SDK
+                            // so the same mapping needs to be handled here for cross platform purchases
                             if let productPlanIdentfier = subscription.productPlanIdentifier {
                                 key = "\(productID):\(productPlanIdentfier)"
                             } else {
--- a/Sources/Identity/CustomerInfo+ActiveDates.swift
+++ b/Sources/Identity/CustomerInfo+ActiveDates.swift
@@ -48,7 +48,17 @@ extension CustomerInfo {
     }
 
     static func extractExpirationDates(_ subscriber: CustomerInfoResponse.Subscriber) -> [String: Date?] {
-        return subscriber.subscriptions.mapValues { $0.expiresDate }
+        var map = [String: Date?]()
+
+        for (productId, subscription) in subscriber.subscriptions {
+            if let productPlanIdentfier = subscription.productPlanIdentifier {
+                map["\(productId):\(productPlanIdentfier)"] = subscription.expiresDate
+            } else {
+                map[productId] = subscription.expiresDate
+            }
+        }
+
+        return map
     }
 
     static func extractPurchaseDates(_ subscriber: CustomerInfoResponse.Subscriber) -> [String: Date?] {
--- a/Sources/Identity/CustomerInfo+ActiveDates.swift
+++ b/Sources/Identity/CustomerInfo+ActiveDates.swift
@@ -48,17 +48,21 @@ extension CustomerInfo {
     }
 
     static func extractExpirationDates(_ subscriber: CustomerInfoResponse.Subscriber) -> [String: Date?] {
-        var map = [String: Date?]()
-
-        for (productId, subscription) in subscriber.subscriptions {
-            if let productPlanIdentfier = subscription.productPlanIdentifier {
-                map["\(productId):\(productPlanIdentfier)"] = subscription.expiresDate
-            } else {
-                map[productId] = subscription.expiresDate
-            }
-        }
-
-        return map
+        return Dictionary(
+                    uniqueKeysWithValues: subscriber
+                        .subscriptions
+                        .lazy
+                        .map { productID, subscription in
+                            let key: String
+                            let value = subscription.expiresDate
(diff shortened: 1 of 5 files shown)


NachoSoto:
Ooooh this makes a lot of sense 👍🏻

Sources/Identity/CustomerInfo+ActiveDates.swift
```diff
@@ -48,7 +48,17 @@ extension CustomerInfo {
...
+
+        for (productId, subscription) in subscriber.subscriptions {
+            if let productPlanIdentfier = subscription.productPlanIdentifier {
+                map["\(productId):\(productPlanIdentfier)"] = subscription.expiresDate
+            } else {
+                map[productId] = subscription.expiresDate
+            }
+        }
+
+        return map
```
NachoSoto, line 61: Nit, but you can make this functional too:
```suggestion
        return Dictionary(
            uniqueKeysWithValues: subscriber
                .subscriptions
                .lazy
                .map { productID, subscription in
                    let key: String
                    let value = subscription.expiresDate

                    if let productPlanIdentfier = subscription.productPlanIdentifier {
                        key = "\(productID):\(productPlanIdentfier)"
                    } else {
                        key = productID
                    }

                    return (key, value)
                }
        )
```

I think that's a bit more clear that every value is the same (`expiresDate`), and we just change the key pro subscriptions with a plan.

Sources/Purchasing/EntitlementInfo.swift
```diff
@@ -133,6 +133,11 @@ extension PeriodType: DefaultValueProvider {
      */
     @objc public var productIdentifier: String { self.contents.productIdentifier }
 
+    /**
+     The product plan identifier that unlocked this entitlement
```
NachoSoto, line 137: Can you add a note that this is just for Android? In fact I wonder if this even needs to be exposed?

Tests/UnitTests/Purchasing/CustomerInfoTests.swift
```diff
@@ -116,13 +127,14 @@ class BasicCustomerInfoTests: TestCase {
     }
 
     func testListActiveSubscriptions() {
-        expect(self.customerInfo.activeSubscriptions) == ["onemonth_freetrial"]
+        expect(self.customerInfo.activeSubscriptions) == ["onemonth_freetrial", "gold:monthly"]
```
NachoSoto, line 130: 👍🏻

NachoSoto (commented):
I think this makes sense. My only important question is regarding the need to expose `EntitlementInfo.productPlanIdentifier`.

Tests/UnitTests/Purchasing/CustomerInfoTests.swift
```diff
@@ -116,13 +127,14 @@ class BasicCustomerInfoTests: TestCase {
...
     }
 
     func testAllPurchasedProductIdentifier() {
         let allPurchased = self.customerInfo.allPurchasedProductIdentifiers
 
-        expect(allPurchased) == ["onemonth", "onemonth_freetrial", "threemonth_freetrial", "onetime_purchase"]
+        expect(allPurchased) == ["onemonth", "onemonth_freetrial",
+                                 "threemonth_freetrial", "gold:monthly", "onetime_purchase"]
```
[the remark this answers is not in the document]
NachoSoto: But the more I think about it, I do worry about breaking the semantics of `activeSubscriptions` and `allPurchasedProductIdentifiers`. They'd no longer be "product identifiers" as their documented to be.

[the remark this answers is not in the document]
NachoSoto: So I wonder if the solution would be to have a separate property instead that combines both. At the end of the day this is something that users of the iOS SDK won't care about, only hybrids, right?

Sources/Purchasing/EntitlementInfo.swift
NachoSoto: Note that if we _do_ need to expose it you need to change this from `fix` to `feat` and add this new method to the API testers.

Tests/UnitTests/Purchasing/CustomerInfoTests.swift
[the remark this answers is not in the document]
joshdholtz: Yeah yeah, this is generally something that hybrids or people using native iOS and Android will care about. I'd _ideally_ like this to be modeled in a new way in somehow in a future version of the SDK because I don't really love this 😛 But 🤷‍♂️

Sources/Purchasing/EntitlementInfo.swift
joshdholtz: I made it internal! Doesn't need to be public at all yet (unless some users think it should be 🤷‍♂️)

Sources/Identity/CustomerInfo+ActiveDates.swift
```diff
@@ -48,7 +48,21 @@ extension CustomerInfo {
...
     static func extractExpirationDates(_ subscriber: CustomerInfoResponse.Subscriber) -> [String: Date?] {
-        return subscriber.subscriptions.mapValues { $0.expiresDate }
+        return Dictionary(
+                    uniqueKeysWithValues: subscriber
+                        .subscriptions
+                        .lazy
+                        .map { productID, subscription in
+                            let key: String
+                            let value = subscription.expiresDate
+                            if let productPlanIdentfier = subscription.productPlanIdentifier {
```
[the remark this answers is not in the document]
joshdholtz: Done!

NachoSoto (approved):
Awesome 🚀

PR #2659, opened 2023-06-15
`Integration Tests`: added tests to verify 304 behavior

NachoSoto:
We had no end-to-end tests for ETags.

### DIFF
--- a/Tests/BackendIntegrationTests/OtherIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/OtherIntegrationTests.swift
@@ -24,6 +24,45 @@ class OtherIntegrationTests: BaseBackendIntegrationTests {
         expect(info.isComputedOffline) == false
     }
 
+    func testGetCustomerInfoReturnsNotModified() async throws {
+        // 1. Fetch user once
+        _ = try await Purchases.shared.customerInfo(fetchPolicy: .fetchCurrent)
+
+        let logger = TestLogHandler()
+
+        // 2. Re-fetch user
+        _ = try await Purchases.shared.customerInfo(fetchPolicy: .fetchCurrent)
+
+        let expectedRequest = HTTPRequest(method: .get,
+                                          path: .getCustomerInfo(appUserID: Purchases.shared.appUserID))
+
+        // 3. Verify response was 304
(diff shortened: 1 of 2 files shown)


Tests/BackendIntegrationTests/OtherIntegrationTests.swift
```diff
@@ -24,6 +24,45 @@ class OtherIntegrationTests: BaseBackendIntegrationTests {
...
+        // 3. Re-fetch user
+        _ = try await Purchases.shared.customerInfo(fetchPolicy: .fetchCurrent)
+
+        let expectedRequest = HTTPRequest(method: .get,
+                                          path: .getCustomerInfo(appUserID: Purchases.shared.appUserID))
+
+        // 4. Verify response was 304
+        logger.verifyMessageWasLogged(
+            Strings.network.api_request_completed(expectedRequest, httpCode: .notModified)
+        )
```
NachoSoto, line 63: ⚠️ This test currently fails /cc @RevenueCat/backend. The backend is not using the ETag, and it's returning an invalid signature. The request has ETag X, and the server is responding with a 200 instead of 304, and putting the same ETag X in the header.

Tests/BackendIntegrationTests/OtherIntegrationTests.swift
```diff
@@ -24,6 +24,45 @@ class OtherIntegrationTests: BaseBackendIntegrationTests {
...
+        let expectedRequest = HTTPRequest(method: .get,
+                                          path: .getCustomerInfo(appUserID: Purchases.shared.appUserID))
+
+        // 3. Verify response was 304
+        logger.verifyMessageWasLogged(
+            Strings.network.api_request_completed(expectedRequest, httpCode: .notModified)
+        )
+    }
+
+    func testGetCustomerInfoAfterLogInReturnsNotModified() async throws {
```
NachoSoto, line 45: This has exposed a bug in signature validation on some 304 responses. I'm still debugging why.

NachoSoto: Figured it out: #2668.

tonidero (approved):

PR #2661, opened 2023-06-15
Documented `HTTPStatusCode.isSuccessfullySynced`

NachoSoto:
Adding this new `HTTPStatusCode.unauthorized` has exposed the fact that we were treating all `4xx` as "successfully synced".

### DIFF
--- a/Tests/UnitTests/Networking/HTTPStatusCodeTests.swift
+++ b/Tests/UnitTests/Networking/HTTPStatusCodeTests.swift
@@ -57,6 +57,7 @@ class HTTPStatusCodeTests: TestCase {
 
     func testIsNotSuccessfulResponse() {
         expect(HTTPStatusCode.invalidRequest.isSuccessfulResponse) == false
+        expect(HTTPStatusCode.unauthorized.isSuccessfulResponse) == false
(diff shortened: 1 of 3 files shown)


Tests/UnitTests/Networking/HTTPStatusCodeTests.swift
```diff
@@ -90,7 +92,6 @@ class HTTPStatusCodeTests: TestCase {
         expect(HTTPStatusCode.redirect.isSuccessfullySynced) == true
         expect(HTTPStatusCode.notModified.isSuccessfullySynced) == true
         expect(HTTPStatusCode.temporaryRedirect.isSuccessfullySynced) == true
-        expect(HTTPStatusCode.invalidRequest.isSuccessfullySynced) == true
```
NachoSoto, line 93: This was also wrong?

NachoSoto: I wonder if we were doing this so that if there's bad subscriber attribute data was just mark is as synced to avoid posting it forever?
If that's the case, we should change the implementation to allow `400` but not all `4xx`.

PR #2517, opened 2023-05-19
`MainThreadMonitor`: fixed flakiness in CI

NachoSoto:
Fixes https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/11230/workflows/78444bf6-22b8-40fc-ad92-1f29279377d0/jobs/69663
This is meant to detect deadlocks. 1 second is unfortunately too low for CI with limited resources.

PR #2662, opened 2023-06-15
`MainThreadMonitor`: increased threshold

NachoSoto:
See also #2517.

We keep getting flaky failures with this in CI, which is annoying because it marks the whole test run as failed and it doesn't retry.
The reason for that is because we have to use `fatalError` and can't use `fail()` because this detection has to happen outside the main thread.
If the main thread is blocked, well, we can't rely `XCTest` to run in the main thread.

Ultimately, we can't guarantee that CI machines will be fast enough to ensure they don't get blocked.
The purpose of this class was to detect deadlocks. By increasing it to 30 seconds, we pretty much avoid flaky failures for slow CI machines, but still have the ability to detect if the main thread is deadlocked due to a locking issue.

### DIFF
--- a/Tests/BackendIntegrationTests/MainThreadMonitor.swift
+++ b/Tests/BackendIntegrationTests/MainThreadMonitor.swift
@@ -52,7 +52,7 @@ final class MainThreadMonitor {
         }
     }
 
-    private static let threshold: DispatchTimeInterval = .seconds(5)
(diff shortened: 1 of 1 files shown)


NachoSoto:
I think this is extra obvious today because for some reason CircleCI seems to be taking 2x the time building.

aboedo (approved):

PR #2663, opened 2023-06-15
`LoadShedderIntegrationTests`: verify requests are actually handled by load shedder

NachoSoto:
This is important to avoid false positives if the API key redirect stops working.

### DIFF
--- a/Tests/BackendIntegrationTests/LoadShedderIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/LoadShedderIntegrationTests.swift
@@ -37,26 +37,42 @@ class LoadShedderStoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
         return Signing.enforcedVerificationMode()
     }
 
-    func testCanGetOfferings() async throws {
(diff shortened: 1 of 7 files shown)


Tests/BackendIntegrationTests/LoadShedderIntegrationTests.swift
```diff
@@ -37,15 +37,39 @@ class LoadShedderStoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
...
+
     func testCanGetOfferings() async throws {
         let receivedOfferings = try await Purchases.shared.offerings()
 
         expect(receivedOfferings.all).toNot(beEmpty())
         assertSnapshot(matching: receivedOfferings.response, as: .formattedJson)
     }
 
+    func testOfferingsComeFromLoadShedder() async throws {
+        self.logger.verifyMessageWasLogged(
```
tonidero, line 58: Having these kind of tests would be very useful in Android. It's difficult to test this without it

tonidero (approved):

PR #2665, opened 2023-06-15
`ETagManager`: add API key to `ETag` content

NachoSoto:
This fixes the following scenario:

- User makes initial requests with invalid API key
- 401 response
- We store that response (`ErrorResponse`)
- User fixes the API key
- Server returns 304
- We return the cached body for the `ErrorResponse`

This is a proposed fix for COIN-595. Instead of this, we might want to ensure that 401 responses don't return etags.


NachoSoto:
Closing this since the backend is no longer sending etags in 401s.

PR #2666, opened 2023-06-16
`ETagManager.httpResultFromCacheOrBackend`: return response headers

NachoSoto:
When working on #2663, the new `isLoadShedder` was not getting the right value for 304 responses because we were returning `[:]`, since we don't store headers in `ETagManager`.

### DIFF
--- a/Sources/Networking/HTTPClient/ETagManager.swift
+++ b/Sources/Networking/HTTPClient/ETagManager.swift
@@ -84,6 +84,8 @@ class ETagManager {
             .compactMapValues { $0 }
     }
 
+    /// - Returns: `response` if a cached response couldn't be fetched,
+    /// or the cached `HTTPResponse`, always including the headers in `response`.
     func httpResultFromCacheOrBackend(with response: HTTPResponse<Data?>,
                                       request: URLRequest,
                                       retried: Bool) -> HTTPResponse<Data>? {
--- a/Sources/Networking/HTTPClient/ETagManager.swift
+++ b/Sources/Networking/HTTPClient/ETagManager.swift
@@ -99,7 +99,8 @@ class ETagManager {
                 let newResponse = storedResponse.withUpdatedValidationTime()
 
                 self.storeIfPossible(newResponse, for: request)
-                return newResponse.asResponse(withRequestDate: response.requestDate)
+                return newResponse.asResponse(withRequestDate: response.requestDate,
+                                              headers: response.responseHeaders)
             }
             if retried {
                 Logger.warn(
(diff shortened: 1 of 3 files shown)


Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -133,10 +134,6 @@ private extension ETagManager {
         }
     }
 
-    func storedHTTPResponse(for request: URLRequest, withRequestDate requestDate: Date?) -> HTTPResponse<Data>? {
```
NachoSoto, line 136: This wasn't used.

Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -226,10 +223,13 @@ extension ETagManager.Response {
...
 
-    fileprivate func asResponse(withRequestDate requestDate: Date?) -> HTTPResponse<Data> {
+    fileprivate func asResponse(
+        withRequestDate requestDate: Date?,
+        headers: HTTPClient.ResponseHeaders
+    ) -> HTTPResponse<Data> {
         return HTTPResponse(
             statusCode: self.statusCode,
-            responseHeaders: [:],
+            responseHeaders: headers,
```
NachoSoto, line 232: This is the core of the change.

Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -86,7 +86,8 @@ class ETagManager {
                 let newResponse = storedResponse.withUpdatedValidationTime()
 
                 self.storeIfPossible(newResponse, for: request)
-                return newResponse.asResponse(withRequestDate: response.requestDate)
+                return newResponse.asResponse(withRequestDate: response.requestDate,
+                                              headers: response.responseHeaders)
```
tonidero, line 90: This will use the cached response but with the 304 response headers... I think this could be confusing... 

I was thinking, we could maybe add additional properties to `HTTPResponse` to have the original headers and new headers (only on 304 responses). It would still be a bit confusing, but clearer that there are 2 sets of headers.

Another option would be to rename the property in HTTPResponse to `latestHeaders` or something and add some documentation... Thoughts?

NachoSoto: > This will use the cached response but with the 304 response headers... I think this could be confusing...

I mean this method is `httpResultFromCacheOrBackend`, it returns the cached value if getting a 304.
Another way to think about it is, this method returns the 304 response + the "missing body".

I'll add a comment mentioning that.

NachoSoto: Done: 25510ab
And also updated a test to reflect this when using the mock: d6ac1be

tonidero: Yeah I think that helps 👍

tonidero (approved):

PR #2642, opened 2023-06-13
`Signature Verification`: new `Integration Tests`

NachoSoto:

PR #2667, opened 2023-06-16
`Offline Entitlements`: added signature validation to entitlement mapping

NachoSoto:
Integration tests will be added in #2642.


Sources/Networking/HTTPClient/HTTPRequest.swift
```diff
@@ -154,15 +154,15 @@ extension HTTPRequest.Path {
         case .getCustomerInfo,
                 .logIn,
                 .postReceiptData,
+                .getProductEntitlementMapping,
```
tonidero, line 157: Hmm don't we need to align on the mechanism for signing these requests?

NachoSoto: Yeah I’m doing that separately. I thought these already had signing with nonces but no, that’s what tests are failing,

NachoSoto:
This is ready now. Didn't need to do any code changes 🎉  simply covered this behavior in tests.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1373,8 +1373,9 @@ extension Purchases: InternalPurchasesType {
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
     func productEntitlementMapping() async throws -> ProductEntitlementMapping {
         let response = try await Async.call { completion in
-            self.backend.offlineEntitlements.getProductEntitlementMapping(withRandomDelay: false,
-                                                                          completion: completion)
+            self.backend.offlineEntitlements.getProductEntitlementMapping(withRandomDelay: false) { result in
+                completion(result.mapError(\.asPublicError))
```
NachoSoto, line 1377: This was forwarding the wrong error.

tonidero (approved):

tonidero:
Mentioned this over Slack but I believe we need to add a check to make sure that, when verification is enabled and the signature is missing in offerings/product-entitlement-mapping, it fails to verify. Currently, since those requests won't have a nonce, it won't fail I believe.

NachoSoto:
@tonidero thanks! Yeah, I'll do that in a separate PR.

NachoSoto:
@tonidero done in #2752

PR #2561, opened 2023-05-29
add DNSSEC for iOS 16

aboedo:
Addresses #2560. 

Sets DNSSEC for URLSession. I need to test more in detail, but so far it seems to work just fine with no surprises. 
I also ended up doing a few things for the sake of testing: 
- extracted the URLSessionConfiguration creation into a factory
- created tests for it and a mock
- passed it in through DI to the HTTPClient
- added availability checks for iOS 16 for tests

PR #2668, opened 2023-06-16
`HTTPClient`: disable `URLSession` cache

NachoSoto:
Turns out this was breaking an HTTP 304 integration test in #2659. I'm very confused why we hadn't seen it before.

Possibly it didn't matter for normal requests, but it was breaking signature verification because the `nonce` was different on the new request, and therefore it wouldn't match the cached signature.

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -42,6 +42,7 @@ class HTTPClient {
         config.httpMaximumConnectionsPerHost = 1
         config.timeoutIntervalForRequest = requestTimeout
         config.timeoutIntervalForResource = requestTimeout
+        config.urlCache = nil // We implement our own caching with `ETagManager`.
         self.session = URLSession(configuration: config,
                                   delegate: RedirectLoggerSessionDelegate(),
                                   delegateQueue: nil)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -42,6 +42,7 @@ class HTTPClient {
         config.httpMaximumConnectionsPerHost = 1
         config.timeoutIntervalForRequest = requestTimeout
         config.timeoutIntervalForResource = requestTimeout
+        config.urlCache = nil // We implement our own caching with `ETagManager`.
```
tonidero, line 45: Hmm I wonder, if this was the case before, did we ever receive 304 responses in our code? As in, would the urlCache always kick in? Not sure how it works internally though.

tonidero (commented):
I think this makes sense... But still wonder why we receive 304 responses sometimes with this... Any thoughts?

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto: We did, but we didn’t really have any actual coverage for it other that using the HTTP mocks which could have messed with this.

NachoSoto:
> I think this makes sense... But still wonder why we receive 304 responses sometimes with this... Any thoughts?

@aboedo figured it out: 
> Since origin servers do not always provide explicit expiration times, a cache MAY assign a heuristic expiration time when an explicit time is not specified

Source: https://www.rfc-editor.org/rfc/rfc9111.html#heuristic.freshness

So the answer is: it's an implementation detail 🤷🏻‍♂️

tonidero (approved):
Interesting, thanks for investigating!

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto, line 45: BTW let's add a test for this when we finish #2561.

PR #2669, opened 2023-06-16
`HTTPClient`: improved log for failed requests

NachoSoto:
Changed this log:
```
[network] DEBUG: ℹ️ API request failed: GET /v1/subscribers/17571811-B0A4-49DC-B326-572E602BF195: Request failed signature verification.
```

To this:
```
[network] DEBUG: ℹ️ API request failed: GET /v1/subscribers/BD9F38FF-33C6-4A77-86F7-6AEA92E6717D (200): Request failed signature verification.
```

This was useful when debugging #2668 and #2659.

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -320,7 +320,9 @@ private extension HTTPClient {
                     httpCode: urlResponse?.httpStatusCode ?? response.statusCode
                 ))
             case let .failure(error):
-                Logger.debug(Strings.network.api_request_failed(request.httpRequest, error: error))
+                Logger.debug(Strings.network.api_request_failed(request.httpRequest,
+                                                                httpCode: urlResponse?.httpStatusCode,
+                                                                error: error))
             }
 
             request.completionHandler?(response)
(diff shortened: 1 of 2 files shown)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -320,7 +320,9 @@ private extension HTTPClient {
                     httpCode: urlResponse?.httpStatusCode ?? response.statusCode
                 ))
             case let .failure(error):
-                Logger.debug(Strings.network.api_request_failed(request.httpRequest, error: error))
+                Logger.debug(Strings.network.api_request_failed(request.httpRequest,
+                                                                httpCode: urlResponse?.httpStatusCode,
```
tonidero, line 324: In the success case we fallback to the cached status code. Is there a reason we need that? Just wondering if it's the same here.

tonidero (approved):

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto: Because `urlResponse` is optional here, for cases when the response fails catastrophically and there’s nothing. In the case of an error we don’t have a parsed response anyway, only this optional.

PR #2670, opened 2023-06-16
`CI`: changed `Codecov` to `informational`

NachoSoto:
For some PRs where we only change a few lines, we sometimes get failures (example: #2668).
It's extra noise, since we're only using coverage data as information.

See documentation: https://docs.codecov.com/docs/codecovyml-reference


tonidero (approved):
This makes sense. We should probably also add this to the other repos

PR #2671, opened 2023-06-16
`ETagManager`: refactored e-tag creation and tests

NachoSoto:
We had lots of different parts of the code doing `request.url?.absoluteString` and also using different values for e-tags.
This refactor combines all that to use a single `eTag(for:)` method.

I did this as part of #2665, which is getting closed, but it will still be useful if we ever decide to change the content of e-tags.

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -100,7 +100,6 @@ extension HTTPClient {
     enum RequestHeader: String {
 
         case authorization = "Authorization"
-        case location = "Location"
         case nonce = "X-Nonce"
         case eTag = "X-RevenueCat-ETag"
         case eTagValidationTime = "X-RC-Last-Refresh-Time"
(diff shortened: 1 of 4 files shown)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -99,7 +99,6 @@ extension HTTPClient {
     enum RequestHeader: String {
 
         case authorization = "Authorization"
-        case location = "Location"
```
NachoSoto, line 102: This was in the wrong `enum`: it's a response header not request.

Tests/UnitTests/Networking/ETagManagerTests.swift
```diff
@@ -280,8 +290,9 @@ class ETagManagerTests: TestCase {
     func testResponseIsNotStoredIfVerificationFailedWithEnforcedMode() throws {
         try self.setEnforcedETagManager()
 
-        let eTag = "an_etag"
```
NachoSoto, line 283: What was I thinking, we still need to have a separate eTag value...

PR #2673, opened 2023-06-16
`Diagnostics`: new `FileHandler` for abstracting file operations

NachoSoto:
Implemented as an `actor` to easily synchronize operations on a file.

Simple interface:
Screenshot 2023-06-16 at 16 45 14

### DIFF
--- a/Sources/Diagnostics/FileHandler.swift
+++ b/Sources/Diagnostics/FileHandler.swift
@@ -1,0 +1,256 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  FileHandler.swift
+//
+//  Created by Nacho Soto on 6/16/23.
+
+import Foundation
+
+/// A wrapper that allows basic operations on a file, synchronized as an `actor`.
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+actor FileHandler {
+
+    private var fileHandle: FileHandle
+
+    let url: URL
+
+    init(_ fileURL: URL) throws {
+        try Self.createFileIfNecessary(fileURL)
+
+        self.url = fileURL
+        self.fileHandle = try FileHandle(fileURL)
+    }
+
+    deinit {
+        let url = self.url
+        Logger.verbose(Message.closing_handle(url))
+
+        self.fileHandle.closeAndLogErrors()
+    }
+
+    /// - Note: this loads the entire file in memory
+    /// For newer versions, consider using `readLines` instead.
+    func readFile() throws -> Data {
+        try self.moveToBeginningOfFile()
+
+        return self.fileHandle.availableData
+    }
+
+    /// Returns an async sequence for every line in the file
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    func readLines() throws -> AsyncLineSequence<FileHandle.AsyncBytes> {
+        try self.moveToBeginningOfFile()
+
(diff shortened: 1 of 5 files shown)


Sources/Diagnostics/FileHandler.swift
```diff
@@ -0,0 +1,252 @@
...
+    func emptyFile() throws {
+        do {
+            try self.fileHandle.truncate(atOffset: 0)
+        } catch {
+            throw Error.failedEmptyingFile(error)
+        }
+    }
+
+    /// Deletes the first N lines from the file, without loading the entire file in memory.
+    func removeFirstLines(_ count: Int) async throws {
```
NachoSoto, line 75: This implementation avoids loading the entire file in memory.

Sources/Diagnostics/FileHandler.swift
```diff
@@ -0,0 +1,256 @@
...
+
+    func replaceHandler(with otherURL: URL) throws {
+        do {
+            try Self.fileManager.removeItem(at: self.url)
+            try Self.fileManager.moveItem(at: otherURL, to: self.url)
+        } catch {
+            throw Error.failedMovingNewFile(from: otherURL, toURL: self.url, error)
+        }
+
+        self.fileHandle.closeAndLogErrors()
```
tonidero, line 187: So closing the fileHandle AFTER we've removed the original item won't be a problem? I'd imagine we would need to close it before deleting the file but I might be wrong.

Sources/Diagnostics/FileHandler.swift
```diff
@@ -0,0 +1,256 @@
...
+    static func createTemporaryFile() throws -> URL {
+        let result = Self.fileManager.temporaryDirectory
+            .appendingPathComponent("com.revenenuecat")
+            .appendingPathComponent(UUID().uuidString)
+
+        try Self.createFileIfNecessary(result)
+        return result
+    }
+
+    func replaceHandler(with otherURL: URL) throws {
```
tonidero, line 179: I wonder if this function should be executed atomically, so if there are multiple threads using this handler, this doesn't cause any issues (We don't do this in Android either)

Tests/UnitTests/Diagnostics/FileHandlerTests.swift
```diff
@@ -0,0 +1,301 @@
...
+
+    // MARK: - append
+
+    func testAppendOneLine() async throws {
+        let content = Self.sampleLine()
+
+        await self.handler.append(line: content)
+
+        let data = try await self.handler.readFile()
+        expect(data).to(matchLines(content))
```
tonidero, line 59: Should we have a test making sure the file ends with a new line break?

tonidero (approved):
Looks great!

Sources/Diagnostics/FileHandler.swift
NachoSoto: That's handled by the fact that this type is an `actor` :) 
https://www.swiftbysundell.com/articles/swift-actors/

Tests/UnitTests/Diagnostics/FileHandlerTests.swift
NachoSoto: I thought it wasn't that important to add an explicit test, as long as the other logic works as expected (appending an existing file, etc). What do you think?

Sources/Diagnostics/FileHandler.swift
tonidero: Ahh right!

Tests/UnitTests/Diagnostics/FileHandlerTests.swift
tonidero: Yeah that makes sense 👍

Sources/Diagnostics/FileHandler.swift
NachoSoto: This is basically just a file descriptor inside. I did this after moving it because we don't want to leave an inconsistent state if any of the move operations fail.

PR #2675, opened 2023-06-20
[WIP] `Configuration`: added parameter to allow overriding `Bundle`

NachoSoto:
This is necessary for app extensions that require configuring the SDK from outside the main bundle where the App Store receipt lives.

Example:
```swift
Purchases.configure(
    with: .init(withAPIKey: "api key")
        .with(bundle: Bundle(identifier: "com.your_main.app_bundle"))
)
```


NachoSoto:
Closing this since it's no longer required. Will update the docs to explain how to better handle extensions

PR #2679, opened 2023-06-21
`Trusted Entitlements`: new Signature format

NachoSoto:
### New format:
 - 32 bytes: intermediate public key
 - 4 bytes: Expiration (in days since epoch)
 - 64 bytes: intermediate public key signature, signed with the root private key
 - 16 bytes: salt
 - 64 bytes: payload signature:
   - salt
   - nonce (if present)
   - request time (as int string)
   - etag (if present)
   - payload

This also adds support for optional nonces for "static" signatures, which is required for #2667.

### DIFF
--- a/Sources/Security/Signing+ResponseVerification.swift
+++ b/Sources/Security/Signing+ResponseVerification.swift
@@ -70,20 +70,23 @@ extension HTTPResponse where Body == Data {
         publicKey: Signing.PublicKey?,
         signing: SigningType.Type
     ) -> VerificationResult {
-        guard let nonce = request.nonce,
-              let publicKey = publicKey,
+        guard let publicKey = publicKey,
               statusCode.isSuccessfulResponse,
               #available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *) else {
             return .notRequested
         }
 
         guard let signature = HTTPResponse.value(
-            forCaseInsensitiveHeaderField: HTTPClient.ResponseHeader.signature.rawValue,
+            forCaseInsensitiveHeaderField: .signature,
             in: headers
         ) else {
-            Logger.warn(Strings.signing.signature_was_requested_but_not_provided(request))
-
-            return .failed
+            let signatureRequested = request.nonce != nil
+            if signatureRequested {
+                Logger.warn(Strings.signing.signature_was_requested_but_not_provided(request))
+                return .failed
+            } else {
+                return .notRequested
+            }
         }
 
         guard let requestDate = requestDate else {
@@ -92,17 +95,11 @@ extension HTTPResponse where Body == Data {
             return .failed
         }
 
-        let eTag = HTTPResponse.value(forCaseInsensitiveHeaderField: HTTPClient.ResponseHeader.eTag.rawValue,
-                                      in: headers)
-
-        let messageToSign = statusCode == .notModified
-            ? eTag?.asData ?? .init()
-            : body
-
         if signing.verify(signature: signature,
                           with: .init(
-                            message: messageToSign,
-                            nonce: nonce,
+                            message: body,
+                            nonce: request.nonce,
+                            etag: HTTPResponse.value(forCaseInsensitiveHeaderField: .eTag, in: headers),
                             requestDate: requestDate.millisecondsSince1970
                           ),
                           publicKey: publicKey) {
--- a/Sources/Security/Signing.swift
+++ b/Sources/Security/Signing.swift
@@ -35,7 +35,8 @@ enum Signing: SigningType {
     struct SignatureParameters {
 
         let message: Data
-        let nonce: Data
+        let nonce: Data?
+        let etag: String?
         let requestDate: UInt64
 
     }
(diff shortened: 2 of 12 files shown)


Sources/Security/Signing.swift
```diff
@@ -195,4 +207,53 @@ extension Signing.SignatureParameters {
...
+            case .payload: return 64
+            }
+        }
+
+        var offset: Int {
+            return Self.allCases
+                .prefix(while: { $0 != self })
+                .map(\.size)
+                .sum()
+        }
```
NachoSoto, line 233: TODO: cache these lazily.

Tests/UnitTests/Networking/SignatureVerificationHTTPClientTests.swift
```diff
@@ -216,6 +216,8 @@ final class InformationalSignatureVerificationHTTPClientTests: BaseSignatureVeri
     }
 
     func testValidSignatureWithETagResponse() throws {
+        XCTExpectFailure("Not yet implemented")
```
NachoSoto, line 219: The fix for this is quite long so it's coming in a separate PR.

Sources/Security/Signing+ResponseVerification.swift
```diff
@@ -81,9 +80,13 @@ extension HTTPResponse where Body == Data {
...
         ) else {
-            Logger.warn(Strings.signing.signature_was_requested_but_not_provided(request))
-
-            return .failed
+            let signatureRequested = request.nonce != nil
+            if signatureRequested {
+                Logger.warn(Strings.signing.signature_was_requested_but_not_provided(request))
+                return .failed
+            } else {
+                return .notRequested
```
tonidero, line 88: This might not be true for endpoints that don't need a nonce I think? I'm not too clear on how the backend will know whether to sign those endpoints though...

Sources/Security/Signing.swift
```diff
@@ -185,14 +201,74 @@ extension Signing.SignatureParameters {
...
+        case intermediatePublicKey
+        case expiration
+        case intermediatePublicKeySignature
+        case salt
+        case payload
+
+        var size: Int {
+            switch self {
+            case .intermediatePublicKey: return 32
+            case .expiration: return 4
```
tonidero, line 227: Should we rename this to `intermediatePublicKeyExpiration`?

Sources/Security/Signing.swift
```diff
@@ -185,14 +201,74 @@ extension Signing.SignatureParameters {
...
         )
     }
 
 }
 
 // MARK: - Private
 
+extension Signing {
+
+    enum SignatureComponent: CaseIterable, Comparable {
```
tonidero, line 216: This is nice! ❤️

tonidero (commented):
Should we hold off merging this to main until everything is ready so we don't break things in main? This looks good though!

Sources/Security/Signing.swift
```diff
@@ -103,9 +120,8 @@ enum Signing: SigningType {
 
     // MARK: -
 
-    private static let publicKey = "UC1upXWg5QVmyOSwozp755xLqquBKjjU+di6U8QhMlM="
+    private static let publicKey = "nVoKJjLhhTNo19Mkjr5DEmgMf361HWxxMyctC10Ob7c="
```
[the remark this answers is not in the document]
NachoSoto: I still need to figure out that part.

Sources/Security/Signing.swift
NachoSoto: Oh yeah

Sources/Security/Signing+ResponseVerification.swift
NachoSoto: The endpoints that can have signatures are marked with a decorator, either a normal signature or static. Static ones don’t need a nonce to produce a signature.

Sources/Security/Signing.swift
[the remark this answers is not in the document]
NachoSoto: I left this here so tests pass, but in the last PR this will be reverted, since the actual public key hasn't changed.

Sources/Security/Signing+ResponseVerification.swift
```diff
@@ -92,17 +95,10 @@ extension HTTPResponse where Body == Data {
...
-
-        let messageToSign = statusCode == .notModified
-            ? eTag?.asData ?? .init()
-            : body
-
         if signing.verify(signature: signature,
                           with: .init(
-                            message: messageToSign,
-                            nonce: nonce,
+                            message: body,
```
tonidero, line 100: So this will be true even if the body is empty right? (in the case of 304 responses I'm thinking)

tonidero: Ah ignore me that was in the next PR 😅

Sources/Security/Signing.swift
```diff
@@ -70,11 +71,28 @@ enum Signing: SigningType {
...
         }
 
-        let salt = signature.subdata(in: 0..<Self.saltSize)
-        let signatureToVerify = signature.subdata(in: Self.saltSize..<signature.count)
+        guard signature.count == SignatureComponent.totalSize else {
+            Logger.warn(Strings.signing.signature_invalid_size(signature))
+            return false
+        }
+
+        // Fixme: verify public key
```
tonidero, line 79: Is this still needed?

tonidero (approved):
Looks good!

Sources/Security/Signing.swift
NachoSoto: That's done in #2715 :)

Tests/BackendIntegrationTests/SignatureVerificationIntegrationTests.swift
```diff
@@ -25,6 +25,8 @@ class BaseSignatureVerificationIntegrationTests: BaseStoreKitIntegrationTests {
     fileprivate var invalidSignature: Bool = false
 
     override func setUp() async throws {
+        throw XCTSkip("Tests disabled until backend deploys new signature format")
```
NachoSoto, line 28: Disabling for now.

Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
```diff
@@ -49,7 +49,9 @@ class BaseBackendIntegrationTests: XCTestCase {
     class var storeKit2Setting: StoreKit2Setting { return .default }
     class var observerMode: Bool { return false }
     class var responseVerificationMode: Signing.ResponseVerificationMode {
-        return .enforced(Signing.loadPublicKey())
+        // Disabled until the backend deploys the new signature format
+        // return .enforced(Signing.loadPublicKey())
+        return .disabled
```
NachoSoto, line 54: ⚠️ Disabling for now.

PR #2681, opened 2023-06-21
`xrOS`: fixed compilation by disabling debug overlay

NachoSoto:
We can look into making them work in the future, but for now this helps compile the SDK for `xrOS`.

### DIFF
--- a/Sources/Support/DebugUI/DebugContentViews.swift
+++ b/Sources/Support/DebugUI/DebugContentViews.swift
@@ -11,7 +11,8 @@
 //
 //  Created by Nacho Soto on 5/30/23.
 
-#if DEBUG && swift(>=5.8) && (os(iOS) || os(macOS)) && !(swift(>=5.9) && os(xrOS))
+#if DEBUG && swift(>=5.8) && (os(iOS) || os(macOS))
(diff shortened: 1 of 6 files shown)


NachoSoto:
I guess Xcode doesn’t stop evaluating the condition early 🫤

Sources/Support/DebugUI/DebugContentViews.swift
```diff
@@ -12,6 +12,7 @@
 //  Created by Nacho Soto on 5/30/23.
 
 #if DEBUG && swift(>=5.8) && (os(iOS) || os(macOS))
+#if !(swift(>=5.9) && os(xrOS))
```
tonidero, line 15: Hmm can it be `xrOS` after the previous line (`(os(iOS) || os(macOS)`)? Same in the other places in this file

MarkVillacampa: I think `os(xrOS)` is only available in `swift(>=5.9)` so that would mean this code would not compile in Xcode 14 (@NachoSoto correct me if I'm wrong)

MarkVillacampa (approved):

tonidero (approved):

PR #2698, opened 2023-06-23
`Trusted Entitlements`: update handling of 304 responses

NachoSoto:
### Changes:
- `ETagManager` no longer stores the verification mode
- Created new `VerifiedHTTPResponse`, which has the `VerificationResult`, extracted from `HTTPResponse`
- Removed `VerificationResult.from(cache:response:)`
- Changed `SignatureVerificationHTTPClientTests` to test using the real `ETagManager`, so they become more "end to end" unit tests. A lot of them were relying on implementation details of the mock instead of the real `ETagManager`.
- Removed several `HTTPClient` tests that were checking a behavior that was impossible (for example, no verification result despite it being enabled), or that checked behavior based on the cached `VerificationResult`

Depends on #2679 and #2697.

### DIFF
--- a/Sources/Networking/HTTPClient/ETagManager.swift
+++ b/Sources/Networking/HTTPClient/ETagManager.swift
@@ -21,27 +21,23 @@ class ETagManager {
     static let eTagResponseHeaderName = HTTPClient.ResponseHeader.eTag.rawValue
 
     private let userDefaults: SynchronizedUserDefaults
-    private let verificationMode: Signing.ResponseVerificationMode
 
-    convenience init(verificationMode: Signing.ResponseVerificationMode) {
+    convenience init() {
         self.init(
             userDefaults: UserDefaults(suiteName: Self.suiteName)
             // This should never return `nil` for this known `suiteName`,
             // but `.standard` is a good fallback anyway.
-            ?? UserDefaults.standard,
-            verificationMode: verificationMode
+            ?? UserDefaults.standard
         )
     }
 
-    init(userDefaults: UserDefaults, verificationMode: Signing.ResponseVerificationMode) {
+    init(userDefaults: UserDefaults) {
         self.userDefaults = .init(userDefaults: userDefaults)
-        self.verificationMode = verificationMode
     }
 
     /// - Parameter withSignatureVerification: whether the request contains a nonce.
     func eTagHeader(
         for urlRequest: URLRequest,
-        withSignatureVerification: Bool,
         refreshETag: Bool = false
     ) -> [String: String] {
         func eTag() -> (tag: String, date: String?)? {
@@ -51,28 +47,12 @@ class ETagManager {
                 return nil
             }
 
-            let shouldUseETag = (
-                !withSignatureVerification ||
-                self.shouldIgnoreVerificationErrors ||
-                storedETagAndResponse.verificationResult == .verified
-            )
+            Logger.verbose(Strings.etag.using_etag(urlRequest,
+                                                   storedETagAndResponse.eTag,
+                                                   storedETagAndResponse.validationTime))
 
-            if shouldUseETag {
-                Logger.verbose(Strings.etag.using_etag(urlRequest,
-                                                       storedETagAndResponse.eTag,
-                                                       storedETagAndResponse.validationTime))
-
-                return (tag: storedETagAndResponse.eTag,
-                        date: storedETagAndResponse.validationTime?.millisecondsSince1970.description)
-            } else {
-                Logger.verbose(Strings.etag.not_using_etag(
-                    urlRequest,
-                    storedETagAndResponse.verificationResult,
-                    needsSignatureVerification: withSignatureVerification
-
-                ))
-                return nil
-            }
+            return (tag: storedETagAndResponse.eTag,
+                    date: storedETagAndResponse.validationTime?.millisecondsSince1970.description)
         }
 
         let (etag, date) = eTag() ?? ("", nil)
@@ -166,13 +146,12 @@ private extension ETagManager {
                                              response: HTTPResponse<Data?>,
                                              eTag: String) {
         if let data = response.body {
-            if response.shouldStore(ignoreVerificationErrors: self.shouldIgnoreVerificationErrors) {
+            if response.shouldStore() {
                 self.storeIfPossible(
                     Response(
                         eTag: eTag,
                         statusCode: response.statusCode,
-                        data: data,
-                        verificationResult: response.verificationResult
+                        data: data
                     ),
                     for: request
                 )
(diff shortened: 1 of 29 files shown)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -253,7 +253,7 @@ private extension HTTPClient {
                request: Request,
                urlRequest: URLRequest,
                data: Data?,
-               error networkError: Error?) -> HTTPResponse<Data>.Result? {
+               error networkError: Error?) -> VerifiedHTTPResponse<Data>.Result? {
```
tonidero, line 256: Hmm I find a bit confusing that the `parse` returns a `VerifiedHTTPResponse`... As in, if verification is disabled, we would still return a `VerifiedHttpResponse`, but it wouldn't be verified right? 

I was trying to think of other names like `HttpResponseWithVerificationResult` but not sure... Thoughts?

NachoSoto: Yeah I agree, I thought of the same, but I think if you consider the fact that “not requested” is a type of verification result, this name might be consistent?

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -262,36 +262,55 @@ private extension HTTPClient {
...
-            .mapToResponse(response: httpURLResponse,
-                           request: request.httpRequest,
-                           signing: self.signing(for: request.httpRequest),
-                           verificationMode: request.verificationMode)
+            .success(data)
+            .mapToResponse(response: httpURLResponse, request: request.httpRequest)
+            // Fetch from ETagManager if available
             .map { (response) -> HTTPResponse<Data>? in
-                guard let cachedResponse = self.eTagManager.httpResultFromCacheOrBackend(
+                return self.eTagManager.httpResultFromCacheOrBackend(
```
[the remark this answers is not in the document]
NachoSoto: > This might be ok though... if we receive a 304 on a request that already failed to verify, it makes sense that it still fails to verify

Yeah 👍🏻 
Although as discussed this has to change again.

Sources/Networking/HTTPClient/HTTPResponse.swift
```diff
@@ -14,48 +14,108 @@
...
+
+}
+
+struct HTTPResponse<Body: HTTPResponseBody>: HTTPResponseType {
+
+    var statusCode: HTTPStatusCode
     var responseHeaders: HTTPClient.ResponseHeaders
     var body: Body
     var requestDate: Date?
-    var verificationResult: VerificationResult
```
NachoSoto, line 28: This is removed from `HTTPResponse` now, so it no longer defaults to `.notRequested`.

Sources/Networking/HTTPClient/HTTPResponse.swift
```diff
@@ -14,48 +14,108 @@
...
 }
 
-extension HTTPResponse {
+// MARK: - Extensions
+
+extension HTTPResponseType {
 
     /// Equivalent to `HTTPURLResponse.value(forHTTPHeaderField:)`
     /// In keeping with the HTTP RFC, HTTP header field names are case-insensitive.
-    func value(forHeaderField field: String) -> String? {
```
NachoSoto, line 58: A nice refactor here: this is now defined on the protocol `HTTPResponseType` so it's available for both response types.
Also no longer accepting any `String`, enforcing that we use this only with `HTTPClient.ResponseHeader`s and not `HTTPClient.RequestHeader`s

Sources/Security/VerificationResult.swift
```diff
@@ -72,31 +72,16 @@ extension VerificationResult: DefaultValueProvider {
 
 }
 
-extension VerificationResult {
+extension VerificationResult: CustomDebugStringConvertible {
 
-    /// - Returns: the most restrictive ``VerificationResult`` based on the cached verification and
-    /// the response verification.
-    static func from(cache cachedResult: Self, response responseResult: Self) -> Self {
```
NachoSoto, line 79: This is gone 🎉

NachoSoto:
This is ready again 🎉

Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -150,6 +143,14 @@ private extension ETagManager {
         responseCode == .notModified
     }
 
+    func shouldUseETag(_ response: Response, withSignatureVerification: Bool) -> Bool {
+        switch response.verificationResult {
+        case .verified: return true
+        case .notRequested: return !withSignatureVerification
```
tonidero, line 149: In Android, what we do, is that, when we detect that the cached customer info has a `NOT_REQUESTED` verification result and signature verification is enabled (with either informational or enforced), we clear both the customer info cache and the etag cache. I believe we do the same in iOS. Shouldn't that mean this shouldn't happen?

tonidero: Oh well, I guess that would only work for the customer info endpoint... There could be edge cases where this might not be tru for other verified endpoints. Will implement this in Android 👍

Sources/Networking/HTTPClient/HTTPResponse.swift
```diff
@@ -14,48 +14,108 @@
...
+            requestDate: \(self.requestDate?.description ?? "<>")
+        )
+        """
+    }
+
+}
+
+// MARK: - VerifiedHTTPResponse
+
+struct VerifiedHTTPResponse<Body: HTTPResponseBody>: HTTPResponseType {
```
tonidero, line 67: In Android, I'm still keeping only the `HTTPResponse` object (in android that's `HTTPResult`. In Android, that's created within the `ETagManager`, (which I admit, it's a weird responsibility to have there), but we don't need to default to `NOT_REQUESTED`.

tonidero (approved):
Everything makes sense!

Sources/Networking/HTTPClient/ETagManager.swift
tonidero: Done in https://github.com/RevenueCat/purchases-android/pull/1114

Sources/Networking/HTTPClient/HTTPResponse.swift
NachoSoto: I kept this refactor cause it made things cleaner on iOS.

Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -150,6 +143,14 @@ private extension ETagManager {
...
+        case .failed, .verifiedOnDevice: return false
```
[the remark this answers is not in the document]
NachoSoto: Correct. Let me add a comment

PR #2608, opened 2023-06-08
Changed `Logger` to use `os_log`

NachoSoto:
Fixes SDK-3173.

### Motivation
`os_log` (and the new `Logger.log`) is the canonical way of logging in Apple platforms, replacing the old `NSLog`. This was made even more useful with the [updates to log printing in Xcode 15](https://developer.apple.com/videos/play/wwdc2023/10226/).
Additionally, this allows us in the future to log `OSLogMessage` (using `OSLogInterpolation`), and therefore being able to mark certain parts of logs as `privacy: .private`.

### Changes:
- Introduced `LogMessage` to ensure every logged message has a `category`
- Created `LoggerStore` to abstract `os.Logger` creation
- Changed `Logger` to use `os.Logger` if available (defaulting to the old `NSLog` otherwise)
- Updated tests to always use `LogMessage` instances (same as #2600)

With these new categories, log provide extra context. Example
```
[storeKit] DEBUG: 😻 Store products request received response
[storeKit] DEBUG: ℹ️ Store products request finished
[offering] DEBUG: 😻 Offerings updated from network.
[offering] DEBUG: 😻 Offerings updated from network.
[network] DEBUG: ℹ️ API request completed: GET /v1/product_entitlement_mapping (200)
[offline_entitlements] DEBUG: ℹ️ ProductEntitlementMapping cache updated from network.
[network] DEBUG: ℹ️ GetProductEntitlementMappingOperation: Finished
[network] DEBUG: ℹ️ Serial request done: GET product_entitlement_mapping, 0 requests left in the queue
[network] DEBUG: ℹ️ GetCustomerInfoOperation: Started
[network] DEBUG: ℹ️ There are no requests currently running, starting request GET subscribers/$RCAnonymousID%3A7f0e7d8de70745ada839e66a66ec2d86
[network] DEBUG: ℹ️ API request started: GET /v1/subscribers/$RCAnonymousID:7f0e7d8de70745ada839e66a66ec2d86
[network] DEBUG: ℹ️ API request completed: GET /v1/subscribers/$RCAnonymousID:7f0e7d8de70745ada839e66a66ec2d86 (201)
[customer] VERBOSE: Updating CustomerInfo '$RCAnonymousID:7f0e7d8de70745ada839e66a66ec2d86' request date: 2023-06-13 21:33:34 +0000
[customer] DEBUG: ℹ️ Sending latest CustomerInfo to delegate.
[customer] DEBUG: 😻 CustomerInfo updated from network.
[...]

PR #2699, opened 2023-06-23
`Purchases.restoreLogHandler`

NachoSoto:
`Purchases.verboseLogHandler` changed in #2608 with the introduction of a new `internalLogHandler`.
Restoring the log handler doing `Purchases.verboseLogHandler = Purchases.verboseLogHandler` now loses the message category.

We didn't have test coverage for `Purchases.logHandler` either which this adds as well.

This also fixes tests in https://github.com/RevenueCat/purchases-hybrid-common/pull/435

### DIFF
--- a/Tests/UnitTests/TestHelpers/TestLogHandler.swift
+++ b/Tests/UnitTests/TestHelpers/TestLogHandler.swift
@@ -75,6 +75,12 @@ final class TestLogHandler {
 
     deinit { Self.sharedHandler.remove(observer: self) }
 
+    /// If a test overrides `Purchases.verboseLogHandler` or `Logger.internalLogHandler`
+    /// this needs to be called to re-install the test handler.
+    static func restore() {
+        Self.sharedHandler.install()
+    }
+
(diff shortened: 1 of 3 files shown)


Tests/UnitTests/TestHelpers/TestLogHandler.swift
```diff
@@ -75,6 +75,12 @@ final class TestLogHandler {
 
     deinit { Self.sharedHandler.remove(observer: self) }
 
+    /// If a test overrides `Purchases.verboseLogHandler` or `Logger.internalLogHandler`
+    /// this needs to be called to re-install the test handler.
```
tonidero, line 79: Is it worth doing this before every test? Just so we don't forget in the future

tonidero (approved):

Tests/UnitTests/TestHelpers/TestLogHandler.swift
NachoSoto: That would break with `TestLogHandler` because it gets installed only once.

PR #2703, opened 2023-06-26
Fix google play purchases missing purchase date

tonidero:
### Description 
Reported in https://github.com/RevenueCat/purchases-flutter/issues/738. 

After https://github.com/RevenueCat/purchases-ios/pull/2654, we were using the new Google Play identifiers that include the plan id in expiration dates, but we didn't change the system that parses purchase dates. Later on, when we map purchases, we iterate over the expiration dates map, and since that id wasn't found in the purchase dates, it was null, causing the issue in flutter.

This makes the purchase date calculation use the same algorithm that we use for expiration dates to get the new google play product identifiers.

### DIFF
--- a/Sources/Identity/CustomerInfo+ActiveDates.swift
+++ b/Sources/Identity/CustomerInfo+ActiveDates.swift
@@ -53,24 +53,24 @@ extension CustomerInfo {
                         .subscriptions
                         .lazy
                         .map { productID, subscription in
-                            let key: String
+                            let key = Self.productID(productID: productID, purchase: subscription)
                             let value = subscription.expiresDate
-
-                            // Products purchased from Google Play will have a product plan identifier (base plan)
-                            // These products get mapped as "productId:productPlanIdentifier" in the Android SDK
-                            // so the same mapping needs to be handled here for cross platform purchases
-                            if let productPlanIdentfier = subscription.productPlanIdentifier {
-                                key = "\(productID):\(productPlanIdentfier)"
-                            } else {
-                                key = productID
-                            }
                             return (key, value)
                         }
                 )
     }
 
     static func extractPurchaseDates(_ subscriber: CustomerInfoResponse.Subscriber) -> [String: Date?] {
-        return subscriber.allTransactionsByProductId.mapValues { $0.purchaseDate }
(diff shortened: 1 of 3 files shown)


vegaro (approved):

Sources/Identity/CustomerInfo+ActiveDates.swift
```diff
@@ -79,6 +79,17 @@ extension CustomerInfo {
 
 private extension CustomerInfo {
 
+    static func productID(productID: String, purchase: CustomerInfoResponse.Subscription) -> String {
```
aboedo, line 82: I'm a bit late to the game, but I think a more swift-y name for this would be something like: 
```swift
static func productID(_ productID: String, from purchase: CustomerInfoResponse.Subscription)
```
and given that this is called from a function called `extractPurchaseDates`, I'd follow the same scheme here, i.e.: 

```swift
static func extractProductIDAndBasePlan(from productID: String, purchase: CustomerInfoResponse.Subscription)
```

- making clear that it's also doing something with a base plan, since having a function called productID that takes the productID seems confusing unless you look at the implementation
- using `extract` for consistency
- using parameter names to make the name slightly cleaner

aboedo (commented):
Thanks for taking care of this so quickly!

Tests/UnitTests/Purchasing/CustomerInfoTests.swift
```diff
@@ -447,6 +448,11 @@ class BasicCustomerInfoTests: TestCase {
         expect(purchaseDate) == Date(timeIntervalSince1970: 1526797490)
     }
 
+    func testPurchaseDateForGooglePlayProductIdentifier() throws {
```
NachoSoto, line 451: 👌🏻

NachoSoto (commented):
Thanks for fixing this!
I agree with @aboedo’s recommendation but you can refactor this in a separate PR, I’m glad you released the fix so quickly.

Sources/Identity/CustomerInfo+ActiveDates.swift
tonidero: Done in #2708

PR #2709, opened 2023-06-26
`XCFramework`: sign archive for `Xcode 15`

NachoSoto:
_I filed https://github.com/fastlane/fastlane/issues/21359, but for now we can do this ourselves._

## Motivation
This is what happens with Xcode 15 if you try to build with an unsigned `XCFramework`:
image

That breaks the build until you accept it.

### With the signature
This can be inspected in Xcode:
image

When using Xcode 15, this modifies the project to save the known signature:
image

But the `.xcframework` remains compatible with Xcode 14.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -278,7 +278,6 @@ platform :ios do
     version_number = current_version_number
     # making it its own lane until https://github.com/CocoaPods/CocoaPods/issues/11621 is fixed.
     # push_pods 
-    carthage_archive
     export_xcframework
     github_release(version: version_number)
   end
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -415,20 +415,26 @@ platform :ios do
         'macOS',
         'maccatalyst',
         'tvOS',
-        'watchOS'
+        'watchOS',
+        # Fixme: add xrOS
       ]
 
     create_xcframework(
       destinations: platforms,
       scheme: 'RevenueCat',
       xcframework_output_directory: output_directory,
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -278,7 +278,6 @@ platform :ios do
     version_number = current_version_number
     # making it its own lane until https://github.com/CocoaPods/CocoaPods/issues/11621 is fixed.
     # push_pods 
-    carthage_archive
```
NachoSoto, line 281: I've also removed this from `release`. It's useful to have it in `release_checks`, to make sure that our framework can be built with Carthage`.
However, what we actually release is using `export_xcframework` which builds the frameworks. This will make `release` MUCH faster since we no longer build everything twice as many times (and for each platform / architecture combination).

joshdholtz (approved):
Looks good! And will work on adding this into _fastlane_ 💪

fastlane/Fastfile
```diff
@@ -502,6 +507,25 @@ platform :ios do
...
   end
 
+  private_lane :sign_xcframework do |options|
+    match(
+      readonly: true,
+      git_url: ENV['CERTS_REPO_URL'],
+      type: 'appstore',
+      skip_provisioning_profiles: true,
+      # This isn't important, we just need to fetch the certificate
+      app_identifier: "com.revenuecat.CarthageIntegration"
```
NachoSoto, line 517: BTW @joshdholtz how come this is required if we’re skipping provisioning profiles?

Issue #2193, opened 2022-12-28
Issue in UITests with current offerings property.


jesus-mg-ios:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
We are using the current property from offerings to get the required purchase offer from the Product team.
This is a great feature given by revenuecat because you don't need to know the identifier associated with the product.

But when you land to tests with storekit tests, if the configuration doesn't match with the configuration (offerings and packages) in revenuecat, the current offer property is not filled. 

1. Environment
   1. Platform: iOS
   2. SDK version: 4.15.x
   3. StoreKit 2 (_disabled with `useStoreKit2IfEnabled(false)`_) (Y/N): N
   4. OS version: 16.x
   5. Xcode version: 14.1.0
   6. How widespread is the issue. Percentage of devices affected: No users, because it's only on  UI tests.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue
3. Steps to reproduce, with a description of expected vs. actual behavior (described above)
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

**Additional context**
Maybe there's any way to fill the current offer for tests.

```Swift
 Purchases.shared.getOfferings { offerings, error in
    ...

   print(offerings?.current) // Nil
   ...
  }

```

RCGitBot:
👀 [SDKONCALL-197] We've just linked this issue to our internal tracker and notified the team. Thank you for reporting, we're checking this out!

michaelAtRC:
Thanks for reaching out! 

Can you confirm that you have the package that you want to display set as 'current' on the RevenueCat dashboard? Also, I suggest following this [guide](https://www.revenuecat.com/docs/displaying-products#fetching-offerings) on fetching offerings. Once you create the package you should be able to log that to the console.

michaelAtRC:
If your offerings, products, or available packages are empty, it's due to some configuration issue in App Store Connect or the Play Console.

[...]

PR #2711, opened 2023-06-26
New `TestStoreProduct` for creating mock `StoreProduct`s and `Offering`s

### LINKED ISSUE #1563: Public init() for Package and Offerings
Dear RevenueCat developers,

We just migrated to 4.0 and really like the swift experience. Well done!

Currently, I am building some Paywalls we want to A/B test.
SwiftUI & Previews are a great tool, therefore!

However, we initialize our view model with our different packages.
To be able to use Previews, a public init would be really handy to create the mock data.

Is there a reason why they are fileprivate at the moment?
Of course, I could create my own wrapper object, but that does not feel native to me.

Best regards,
Jan André

NachoSoto:
Fixes #1563 and #2193.

This introduces **`DEBUG` and `Swift` only** types that allow creating a whole `Offering` for tests and SwiftUI previews.

Example:
Screenshot 2023-06-26 at 12 36 40

By making it available in only `DEBUG` builds we ensure that apps can't accidentally try to purchase these products, which would be impossible. If this is attempted in `DEBUG` builds, we return `ErrorCode`.

When trying to compile this API in release builds, you get an error with a suggestion:
image

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -289,6 +289,8 @@ final class PurchasesOrchestrator {
                           package: package,
                           promotionalOffer: nil,
                           completion: completion)
+        } else if product.isTestProduct {
+            self.handleTestProduct(completion)
         } else {
             fatalError("Unrecognized product: \(product)")
         }
@@ -315,6 +317,8 @@ final class PurchasesOrchestrator {
                           package: package,
                           promotionalOffer: promotionalOffer,
                           completion: completion)
+        } else if product.isTestProduct {
+            self.handleTestProduct(completion)
         } else {
             fatalError("Unrecognized product: \(product)")
         }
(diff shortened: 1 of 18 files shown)


Sources/Purchasing/StoreKitAbstractions/Test Data/TestStoreProduct.swift
```diff
@@ -0,0 +1,147 @@
...
+extension TestStoreProduct: StoreProductType {
+
+    internal var productCategory: StoreProduct.ProductCategory { return self.productType.productCategory }
+
+    internal var currencyCode: String? {
+        // Test currency defaults to current locale
+        return Locale.current.rc_currencyCode
+    }
+
+    internal var isFamilyShareable: Bool { return false }
```
aboedo, line 104: maybe later we could expose this one as a stored property, so that you can also check flows where you have a paywall that states whether the subs is family shareable or not

aboedo (approved):
love it! Tagging @tonidero since it's related to some of the stuff we were talking about for android

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -289,6 +289,8 @@ final class PurchasesOrchestrator {
                           package: package,
                           promotionalOffer: nil,
                           completion: completion)
+        } else if product.isTestProduct {
+            self.handleTestProduct(completion)
```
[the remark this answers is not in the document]
NachoSoto: If I understand your point, that also applies to how we handle SK1 and SK2 products above, right?

Sources/Purchasing/StoreKitAbstractions/Test Data/TestStoreProduct.swift
NachoSoto: Oh let me add it. I removed this parameter cause I didn't think it was useful to mock, but you make a good point.

PR #2715, opened 2023-06-26
`Signing`: extract and verify intermediate key

NachoSoto:
Last step of the new Signature format. Follow up to #2679 and #2698.

This reverts the public key change in #2679, since that was the intermediate key.
This now extracts the new intermediate public key from the signature, and verifies it using the public key.

### DIFF
--- a/Sources/Security/Signing.swift
+++ b/Sources/Security/Signing.swift
@@ -55,7 +55,7 @@ enum Signing: SigningType {
         }
 
         do {
-            return try Curve25519.Signing.PublicKey(rawRepresentation: key)
+            return try Self.createPublicKey(with: key)
         } catch {
             fail(Strings.signing.invalid_public_key(error.localizedDescription))
         }
@@ -76,7 +76,12 @@ enum Signing: SigningType {
             return false
         }
 
-        // Fixme: verify public key
+        guard let intermediatePublicKey = Self.extractAndVerifyIntermediateKey(
+            from: signature,
+            publicKey: publicKey
+        ) else {
+            return false
+        }
 
         let salt = signature.component(.salt)
         let payload = signature.component(.payload)
@@ -85,6 +90,7 @@ enum Signing: SigningType {
         #if DEBUG
         Logger.verbose(Strings.signing.verifying_signature(
             signature: signature,
+            publicKey: intermediatePublicKey.rawRepresentation,
             parameters: parameters,
             salt: salt,
             payload: payload,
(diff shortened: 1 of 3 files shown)


Sources/Security/Signing.swift
```diff
@@ -259,7 +240,50 @@ extension Signing.SignatureParameters {
...
+        guard expirationDate.timeIntervalSince(Date()) >= 0 else {
+            Logger.warn(Strings.signing.intermediate_key_expired(expirationDate, intermediateKeyExpiration))
+            return nil
+        }
+
+        Logger.verbose(Strings.signing.intermediate_key_creating(expiration: expirationDate,
+                                                                 data: intermediatePublicKey))
+
+        do {
+            return try Self.createPublicKey(with: intermediatePublicKey)
```
NachoSoto, line 277: We considered caching this, but I just ran a benchmark on my device and the whole process (creating this key and verifying the original signature) takes 0.17 seconds.

Sources/Security/Signing.swift
```diff
@@ -259,7 +240,65 @@ extension Signing.SignatureParameters {
...
+        let intermediateKeySignature = signature.component(.intermediateKeySignature)
+
+        guard publicKey.isValidSignature(intermediateKeySignature,
+                                         for: intermediateKeyExpiration + intermediatePublicKey) else {
+            Logger.warn(Strings.signing.intermediate_key_failed_verification(signature: intermediateKeySignature))
+            return nil
+        }
+
+        guard let expirationDate = Self.extractAndVerifyIntermediateKeyExpiration(intermediateKeyExpiration) else {
+            return nil
```
tonidero, line 267: So if the intermediate key is expired, we assume it's an attacker correct? As long as the backend makes sure this doesn't happen, this should be fine 👍

tonidero (commented):
Love this! Going to try to get this done in Android ASAP

tonidero (commented):
Love this! Going to try to get this done in Android ASAP

Sources/Security/Signing.swift
NachoSoto: Yup /cc @bisho

Sources/Security/Signing.swift
NachoSoto: Sorry I meant 0.17ms!

tonidero (approved):
Looks good!

NachoSoto:
> Some tests seem to be failing though

Yeah that's `testVerifyKnownSignatureWithNoNonceAndNoEtag` because I need a signature signed with the real private root key, so I'm waiting for `offerings` to have static signatures.

PR #2717, opened 2023-06-26
`TrustedEntitlements`: measure performance using `TimingUtil`

NachoSoto:
Depends on #2715 and #2716.

### DIFF
--- a/Sources/Logging/Strings/SigningStrings.swift
+++ b/Sources/Logging/Strings/SigningStrings.swift
@@ -34,6 +34,8 @@ enum SigningStrings {
 
     case request_date_missing_from_headers(HTTPRequest)
 
+    case verification_too_slow
+
     #if DEBUG
     case verifying_signature(signature: Data,
                              publicKey: Data,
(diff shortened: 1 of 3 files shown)


NachoSoto:
I'll close this since I don't think it's that useful now that we know it's really fast.

aboedo (commented):
in practice this shouldn't happen, right? 
I wonder if we can just actually test this with the slowest supported device and an artificially huge receipt (or a modified version of a really large receipt) to see how it goes.

aboedo:
mostly wondering because the 0.5 seconds threshold seems arbitrary and device-specific

NachoSoto:
Yeah I debated whether we actually need it. I only did it to measure the performance myself, and was thinking that if it happens to be slow in the wild in an old device, this would be a way to know.
I was envisioning the possibility of maybe `Curve25519` being too slow in some version of iOS 13 or something.

> with the slowest supported device
We'd have to do it with a physical device though.

NachoSoto:
I'll just close this since I don't think it's that useful.

PR #2718, opened 2023-06-27
`Integration Tests`: override `SystemInfo.identifierForVendor` on each test

NachoSoto:
This allows us to ensure that the backend doesn't produce conflicts when processings similar receipts across separate test invocations.


NachoSoto:
Actually this isn't feasible, closing.

PR #2719, opened 2023-06-27
`Integration Tests`: fix clearing `UserDefaults` before each test

NachoSoto:
For some reason `synchronize()` is needed so that subsequent reads and writes actually see that we cleared the domain.
I added a new expectation to make sure if this breaks we catch it early instead of a weird failure later in the test.

### DIFF
--- a/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
@@ -127,6 +127,9 @@ private extension BaseBackendIntegrationTests {
     func createUserDefaults() {
         self.userDefaults = UserDefaults(suiteName: Constants.userDefaultsSuiteName)
         self.userDefaults.removePersistentDomain(forName: Constants.userDefaultsSuiteName)
+        // See also `SynchronizedUserDefaults`.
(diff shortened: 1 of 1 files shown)


Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
```diff
@@ -128,12 +124,26 @@ private extension BaseBackendIntegrationTests {
         }
     }
 
+    func createUserDefaults() {
+        self.userDefaults = UserDefaults(suiteName: Constants.userDefaultsSuiteName)
+        self.userDefaults.removePersistentDomain(forName: Constants.userDefaultsSuiteName)
+        self.userDefaults.synchronize()
```
aboedo, line 130: let's add a comment here in case a future dev wonders why this is needed, reads Apple's docs and decides to remove it as "not needed"

Screenshot 2023-06-27 at 2 54 02 PM

aboedo (approved):

Tests/BackendIntegrationTests/BaseBackendIntegrationTests.swift
NachoSoto: 👍🏻 
That's also why I added the assertion below, but I'll add a comment too.

PR #2606, opened 2023-06-08
[WIP] `StoreProductTests`: update tests for `iOS 17`

NachoSoto:
`StoreKit.Product.displayPrice` changed format in `iOS 17`.

I'm making this a draft for now in case this is just a beta 1 thing, but otherwise we'll need to deal with these differences in tests.
image

PR #2720, opened 2023-06-27
`Tests`: disabled `iOS 11.x` tests to fix `Xcode 15` tests

NachoSoto:
We weren't running iOS 11.x tests anyway because it's no longer possible to install those simulators.
This simply changes the deployment target of **_only_ test targets** to iOS 12.0 so that running those in `Xcode 15.0` doesn't fail.

Note that iOS 17 tests are still disabled in CI since we still have a few failures (see #2606).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -848,9 +848,8 @@ workflows:
   generate-snapshot:
     when: << pipeline.parameters.generate_snapshots >>
     jobs:
-      # Disabled until we drop support for iOS 11
-      # - run-test-ios-17:
-      #     xcode_version: '15.0.0'
+      - run-test-ios-17:
+          xcode_version: '15.0.0'
       - run-test-ios-16:
           xcode_version: '14.3.0'
       - run-test-ios-15:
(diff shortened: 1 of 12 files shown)


.circleci/config.yml
```diff
@@ -879,9 +878,8 @@ workflows:
           xcode_version: '14.3.0'
       - spm-receipt-parser:
           xcode_version: '14.3.0'
-      # Disabled until we drop support for iOS 11
-      # - run-test-ios-17:
-      #     xcode_version: '15.0.0'
+      - run-test-ios-17:
+          xcode_version: '15.0.0'
```
NachoSoto, line 882: Now we can enable these!

Tests/APITesters/ObjCAPITester/ObjCAPITester/RCEntitlementInfoAPI.m
```diff
@@ -31,7 +31,7 @@ + (void)checkAPI {
     RCPurchaseOwnershipType ot = [ri ownershipType];
     NSDictionary<NSString *, id> *rawData = [ri rawData];
 
-    if (@available(iOS 13.0, *)) {
+    if (@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)) {
```
NachoSoto, line 34: The `tvOS` deployment target of API testers was set to iOS 14, so we weren't checking these.

tonidero (approved):

PR #2730, opened 2023-06-28
`Trusted Entitlements`: add integration tests to verify `CustomerInfo` cache invalidation

NachoSoto:
I remembered this behavior while going through the code, so I thought it would be useful to encode as integration tests.


tonidero (approved):
Nice tests!

PR #2731, opened 2023-06-28
Refactor: simplified `PurchasesOrchestrator.syncPurchases`

NachoSoto:
I found this while looking into SDK-3097.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -963,7 +963,6 @@ private extension PurchasesOrchestrator {
         }
     }
 
-    // swiftlint:disable:next function_body_length
     func syncPurchases(receiptRefreshPolicy: ReceiptRefreshPolicy,
                        isRestore: Bool,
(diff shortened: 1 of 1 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -998,17 +997,14 @@ private extension PurchasesOrchestrator {
...
 
-                if !hasTransactions && hasOriginalPurchaseDate {
+                if !hasTransactions, let customerInfo = cachedCustomerInfo, customerInfo.originalPurchaseDate != nil {
                     if let completion = completion {
                         self.operationDispatcher.dispatchOnMainThread {
-                            completion(
-                                Result(cachedCustomerInfo,
-                                       ErrorUtils.customerInfoError(withMessage: "No cached customer info"))
-                            )
+                            completion(.success(customerInfo))
```
NachoSoto, line 1004: @aboedo this new implementation is 100% equivalent. You can see that it wasn't even possible to send an error in this particular case.

tonidero (approved):

PR #2742, opened 2023-06-30
`Documentation`: removed references to deprecated methods on index

NachoSoto:
The documentation is still there, but I don't think we need to reference them in the index anymore.


aboedo (commented):
I added them explicitly because if you don't have them, they end up in this section without anything telling you not to use them

image

I honestly feel like having them in a deprecated section is clearer

NachoSoto:
Oh totally, I forgot about that. Thanks!

PR #2746, opened 2023-06-30
`Trusted Entitlements`: add API key and `HTTPRequest.Path` to signature

NachoSoto:
Depends on #2747.

### Changes:
- Changed `Signing` to a `final class`, so it can be instantiated with the api key.
-  Updated all `SigningTests` to include the correct signatures

### DIFF
--- a/Sources/Security/Signing.swift
+++ b/Sources/Security/Signing.swift
@@ -17,7 +17,7 @@ import Foundation
 /// A type that can verify signatures.
 protocol SigningType {
 
-    static func verify(
+    func verify(
         signature: String,
         with parameters: Signing.SignatureParameters,
         publicKey: Signing.PublicKey
@@ -26,7 +26,7 @@ protocol SigningType {
 }
 
 /// Utilities for handling signature verification.
-enum Signing: SigningType {
+final class Signing: SigningType {
 
     /// An object that represents a cryptographic key.
     typealias PublicKey = SigningPublicKey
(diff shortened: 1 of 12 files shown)


Tests/UnitTests/Security/SigningTests.swift
```diff
@@ -276,6 +295,15 @@ class SigningTests: TestCase {
         )) == true
     }
 
+    /*
```
NachoSoto, line 298: @aboedo I added this for future reference :)
(This PR already has the correct signatures since @bisho got all the changes in the canary 🎉)

Sources/Security/Signing.swift
```diff
@@ -238,6 +246,15 @@ extension Signing.SignatureParameters {
 
 }
 
+private extension HTTPRequest.Path {
+
+    // Signatures remove percent encoding from the paths.
+    var unescapedPath: String {
+        return self.relativePath.removingPercentEncoding ?? ""
+    }
```
NachoSoto, line 254: This is slightly scary because of 
image

I'd vote to have the backend not unescape. Thoughts @bisho?

tonidero: Yeah, that would be my preference as well.

tonidero (approved):
LGTM!

Sources/Security/Signing.swift
NachoSoto: Actually this is an implementation detail of the backend, I can handle it here without having to call `removingPercentEncoding`.

PR #2747, opened 2023-06-30
`HTTPRequest.Path`: escape `appUserID`

NachoSoto:
We were escaping users for every (hopefully) request. This refactor moves that responsibility to the path creation in `HTTPRequest.Path` to ensure it's done consistently.

Additionally, this is required so that `SigningTests` produce a consistent path (see #2746).

### DIFF
--- a/Sources/FoundationExtensions/String+Extensions.swift
+++ b/Sources/FoundationExtensions/String+Extensions.swift
@@ -16,27 +16,14 @@ import Foundation
 
 extension String {
 
-    enum Error: Swift.Error {
-
-        case escapingEmptyString
-
-    }
-
     func rot13() -> String {
         ROT13.string(self)
     }
 
-    func escapedOrError() throws -> String {
-        let trimmedAndEscaped = self
+    var trimmedAndEscaped: String {
+        return self
             .trimmingCharacters(in: .whitespacesAndNewlines)
             .addingPercentEncoding(withAllowedCharacters: .urlHostAllowed)!
-
-        guard trimmedAndEscaped.count > 0 else {
-            Logger.error("Attempting to escape an empty string")
-            throw Error.escapingEmptyString
-        }
-
-        return trimmedAndEscaped
     }
 
     /// Returns `nil` if `self` is an empty string.
--- a/Sources/FoundationExtensions/String+Extensions.swift
+++ b/Sources/FoundationExtensions/String+Extensions.swift
@@ -22,8 +22,8 @@ extension String {
 
     var trimmedAndEscaped: String {
         return self
-            .trimmingCharacters(in: .whitespacesAndNewlines)
-            .addingPercentEncoding(withAllowedCharacters: .urlHostAllowed)!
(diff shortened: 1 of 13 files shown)


Sources/FoundationExtensions/String+Extensions.swift
```diff
@@ -16,27 +16,14 @@ import Foundation
...
-        let trimmedAndEscaped = self
+    var trimmedAndEscaped: String {
+        return self
             .trimmingCharacters(in: .whitespacesAndNewlines)
             .addingPercentEncoding(withAllowedCharacters: .urlHostAllowed)!
-
-        guard trimmedAndEscaped.count > 0 else {
-            Logger.error("Attempting to escape an empty string")
-            throw Error.escapingEmptyString
-        }
```
NachoSoto, line 37: Throwing was completely unnecessary here.

Tests/BackendIntegrationTests/LoadShedderIntegrationTests.swift
```diff
@@ -65,7 +65,7 @@ class LoadShedderStoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
     func testOfferingsComeFromLoadShedder() async throws {
         self.logger.verifyMessageWasLogged(
             Strings.network.request_handled_by_load_shedder(
-                .getOfferings(appUserID: try Purchases.shared.appUserID.escapedOrError())
+                .getOfferings(appUserID: Purchases.shared.appUserID)
```
NachoSoto, line 68: This shows the benefit of this approach: we we can always pass a normal userID value now.

Tests/UnitTests/Networking/Backend/BackendGetCustomerInfoTests.swift
```diff
@@ -74,22 +74,6 @@ class BackendGetCustomerInfoTests: BaseBackendTests {
         }
     }
 
-    func testEncodesCustomerUserID() {
```
NachoSoto, line 77: This test is no longer required, we test this at the `HTTPRequest.Path` level.

Sources/Networking/Operations/GetIntroEligibilityOperation.swift
```diff
@@ -63,7 +63,9 @@ private extension GetIntroEligibilityOperation {
             return
         }
 
-        guard let appUserID = try? self.configuration.appUserID.escapedOrError() else {
+        let appUserID = self.configuration.appUserID
+
+        guard !self.configuration.appUserID.trimmedAndEscaped.isEmpty else {
```
NachoSoto, line 68: Actually no need to escape here, just check that it’s not empty.

tonidero, line 68: I think this could be:
```suggestion
        guard !appUserID.trimmedAndEscaped.isEmpty else {
```

tonidero (approved):
Makes sense!

PR #2750, opened 2023-07-05
[WIP] New `Purchases.errorHandler` for detecting any `NSError`

NachoSoto:
### Description

On Android `LogHandler` has a separate `e` method that receives a `Throwable`. However, it's impossible to receive all logged errors from iOS, since the `LogHandler` definition only has a `message: String`.

This is useful when integrating with other tools like `Sentry`, since you can pipe those errors directly.

### Example usage :

```swift
Purchases.errorHandler = { error in
  SentrySDK.reportError(error)
}
```

### Other benefits
- Makes it easier to detect internal errors when writing tests using `TestLogHandler`
- Makes it easy for users to detect any occurrence of `ErrorCode.signatureVerificationFailed`, even those coming from internal requests.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -161,6 +161,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
      * - Note:``verboseLogHandler`` provides additional information.
      *
      * #### Related Symbols
+     * - ``errorHandler``
      * - ``verboseLogHandler``
      * - ``logLevel``
      */
@@ -188,6 +189,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
      *
      * #### Related Symbols
      * - ``logHandler``
+     * - ``errorHandler``
      * - ``logLevel``
      */
     @objc public static var verboseLogHandler: VerboseLogHandler {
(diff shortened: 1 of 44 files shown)


Tests/APITesters/SwiftAPITester/SwiftAPITester/PurchasesAPI.swift
```diff
@@ -78,7 +78,16 @@ func checkPurchasesEnums() {
 
 private func checkStaticMethods() {
     let logHandler: (LogLevel, String) -> Void = { _, _ in }
+    let verboseLogHandler: (_ level: LogLevel,
```
NachoSoto, line 81: This was missing.

Sources/Security/VerificationResult.swift
```diff
@@ -36,10 +36,13 @@ import Foundation
 /// }
 /// ```
 ///
+/// Verification failures will also be forwarded to ``Purchases/errorHandler``.
```
tonidero, line 39: 👍

tonidero (approved):
Left a question in Slack about the approach but I think it's ok. The code looks good.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -204,9 +206,26 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
...
+     * Set a custom handler for redirecting errors to your own logging system.
+     * This does nothing by default, but it can be overriden to detect and handle
+     * any error occuring in the SDK.
+     *
+     * #### Related Symbols
+     * - ``logHandler``
+     * - ``verboseLogHandler``
+     * - ``logLevel``
+     */
+    @objc public static var errorHandler: ErrorHandler {
```
aboedo, line 219: my only question is: without reading the docstring, would it be immediately obvious to a developer that this doesn't actually handle the errors, but rather error _logs_?
I.e.: I might see this and assume that we do a mega try-catch where everything that's caught is handled by this, so that I have a single place to show a "something went wrong" alert or toast. 
Rather than realize that this is actually about errors. 
Maybe errorLogHandler or something might help?

NachoSoto: Great point 👍🏻  I'll change that.

NachoSoto:
I'm going to hold off of this for a bit. We need to fix the fact that we log errors on creation, which is often leading to us logging them multiple times. Not the worst thing if it's just going to the console, but we shouldn't be sending duplicates to this new API.

PR #2752, opened 2023-07-05
`Trusted Entitlements`: produce verification failures for static endpoints with no signature

NachoSoto:
Equivalent to https://github.com/RevenueCat/purchases-android/pull/1119

### DIFF
--- a/Sources/Security/Signing+ResponseVerification.swift
+++ b/Sources/Security/Signing+ResponseVerification.swift
@@ -64,7 +64,7 @@ extension HTTPResponse where Body == Data? {
             forCaseInsensitiveHeaderField: .signature,
             in: headers
         ) else {
-            if request.requiresSignature {
+            if request.path.supportsSignatureVerification {
                 Logger.warn(Strings.signing.signature_was_requested_but_not_provided(request))
                 return .failed
             } else {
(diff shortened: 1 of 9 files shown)


Sources/Security/Signing+ResponseVerification.swift
```diff
@@ -95,3 +94,16 @@ extension HTTPResponse where Body == Data? {
...
+
+private extension HTTPRequest {
+
+    var requiresSignature: Bool {
+        guard self.path.supportsSignatureVerification else {
+            return false
+        }
+
+        let needsNonce = self.path.needsNonceForSigning
+        return !needsNonce || (needsNonce && self.nonce != nil)
```
tonidero, line 106: Hmm I wonder, is this part needed? In Android I'm only checking that `supportsSignatureVerification` is true and that entitlement verification is enabled. In that situation, we should always require a signature I think? Whether the path requires a nonce or not, shouldn't affect whether we require a signature I think.

Sources/Security/Signing+ResponseVerification.swift
```diff
@@ -95,3 +94,16 @@ extension HTTPResponse where Body == Data? {
     }
 
 }
+
+private extension HTTPRequest {
+
+    var requiresSignature: Bool {
```
tonidero, line 100: Should this also check if entitlement verification is enabled?

Sources/Security/Signing+ResponseVerification.swift
NachoSoto: Oh the issue was some of the failing tests were wrong. I thought I needed this check but I don't! Fixed.

Sources/Security/Signing+ResponseVerification.swift
NachoSoto: That's done by the calling method. But I removed this property so it's more clear now.

PR #2756, opened 2023-07-06
`Testing`: add new `SimpleApp`

NachoSoto:
Just like `PurchaseTester` this can be used to debug the SDK, but as a much simpler setup.

This depends on the SPM as a relative path, so it's easier to iterate on the main SDK library(s) while using this app.

### DIFF
--- a/Tests/TestingApps/SimpleApp/SimpleApp/SimpleApp.swift
+++ b/Tests/TestingApps/SimpleApp/SimpleApp/SimpleApp.swift
@@ -8,8 +8,9 @@
 import RevenueCat
 import SwiftUI
 
-#warning("This needs to be set")
+#warning("This needs to be configured.")
 private let apiKey = ""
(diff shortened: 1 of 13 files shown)


Tests/TestingApps/SimpleApp/SimpleApp/SimpleApp.swift
```diff
@@ -0,0 +1,33 @@
...
+//
+//  Created by Nacho Soto on 5/30/23.
+//
+
+import RevenueCat
+import SwiftUI
+
+#warning("This needs to be set")
+private let apiKey = ""
+private let proxyURL = ""
```
aboedo, line 13: let's add another warning here that explains that this one is optional and you can set it to your own instance of the RC backend or leave empty for prod

aboedo (approved):

PR #2758, opened 2023-07-06
`Trusted Entitlements`: update handling of escaped URLs

NachoSoto:
The backend no longer unescapes URLs when creating signatures.

### DIFF
--- a/Sources/Security/Signing.swift
+++ b/Sources/Security/Signing.swift
@@ -239,7 +239,7 @@ extension Signing.SignatureParameters {
     var asData: Data {
         return (
             (self.nonce ?? .init()) +
-            self.path.unescapedPath.asData +
+            self.path.relativePath.asData +
             String(self.requestDate).asData +
             (self.etag ?? "").asData +
             (self.message ?? .init())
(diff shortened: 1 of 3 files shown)


Sources/Networking/HTTPClient/HTTPRequest.swift
```diff
@@ -234,3 +234,12 @@ extension HTTPRequest.Path: CustomStringConvertible {
     }
 
 }
+
+extension HTTPRequest.Path {
+
+    /// - Returns: the `relativePath` without percent encoding characters.
+    var unescapedRelativePath: String {
```
NachoSoto, line 241: I moved this here and added tests for it specifically.

Tests/UnitTests/Security/SigningTests.swift
```diff
@@ -310,7 +310,7 @@ class SigningTests: TestCase {
         curl -v 'https://api.revenuecat.com/v1/subscribers/login' \
         -X GET \
         -H 'X-Nonce: MTIzNDU2Nzg5MGFi' \
-        -H 'Authorization: Bearer {api_key}'
+        -H 'Authorization: Bearer appl_fFVBVAoYujMZJnepIziGKVjnZBz'
```
NachoSoto, line 313: This is already exposed in the constant below, so makes it easier to copy these `curl` commands and ensure that you use the right key.

Sources/Security/Signing.swift
```diff
@@ -248,11 +248,18 @@ extension Signing.SignatureParameters {
 
 }
 
-private extension HTTPRequest.Path {
+extension Signing.SignatureParameters: CustomDebugStringConvertible {
```
NachoSoto, line 251: I've implemented this to make debugging easier.

aboedo (approved):

PR #2769, opened 2023-07-07
`AnyEncodable`: also implement `Decodable`

NachoSoto:
This will be used for an upcoming test that uses `AnyEncodable` and needs to extract the contents.

### DIFF
--- a/Sources/Misc/Codable/AnyEncodable.swift
+++ b/Sources/Misc/Codable/AnyEncodable.swift
@@ -23,9 +23,10 @@ struct AnyEncodable {
 
 }
 
+// swiftlint:disable cyclomatic_complexity
+
 extension AnyEncodable: Encodable {
 
-    // swiftlint:disable:next cyclomatic_complexity
     func encode(to encoder: Encoder) throws {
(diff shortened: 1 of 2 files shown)


Sources/Misc/Codable/AnyEncodable.swift
```diff
@@ -69,3 +70,43 @@ extension AnyEncodable: Encodable {
     }
 
 }
+
+/// `AnyEncodable` can also be decoded
+extension AnyEncodable: Decodable {
```
aboedo, line 75: feels like we should make this AnyCodable now?

aboedo (approved):

Sources/Misc/Codable/AnyEncodable.swift
NachoSoto: I'm doing that in a separate PR cause it ended up a bit more work.

NachoSoto:
#2770.

Sources/Misc/Codable/AnyEncodable.swift
NachoSoto: #2770

PR #2771, opened 2023-07-07
`Signing`: inject `ClockType` to ensure hardcoded signatures don't fail when intermediate key expires

NachoSoto:
Thanks to @aboedo for catching this and providing this simple idea.

### DIFF
--- a/Sources/Security/Signing.swift
+++ b/Sources/Security/Signing.swift
@@ -43,9 +43,11 @@ final class Signing: SigningType {
     }
 
     private let apiKey: String
+    private let clock: ClockType
 
-    init(apiKey: String) {
+    init(apiKey: String, clock: ClockType = Clock.default) {
         self.apiKey = apiKey
+        self.clock = clock
     }
 
     /// Parses the binary `key` and returns a `PublicKey`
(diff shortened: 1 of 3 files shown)


Sources/Security/Signing.swift
```diff
@@ -312,7 +317,10 @@ private extension Signing {
     }
 
     /// - Returns: `nil` if the key is expired or has an invalid expiration date.
-    private static func extractAndVerifyIntermediateKeyExpiration(_ expirationData: Data) -> Date? {
+    private static func extractAndVerifyIntermediateKeyExpiration(
+        _ expirationData: Data,
+        _ clock: ClockType
```
tonidero, line 322: Hmm seems `clock` is unused? I'm guessing we want to use it to get the current date in line 332?

tonidero (commented):
FWIW, What I did was adding a day very far into the future (Year 2106) in Android for the unit tests. But this also works 👍

aboedo:
> FWIW, What I did was adding a day very far into the future (Year 2106) in Android for the unit tests. But this also works 👍

Some poor dev is going to really hate you in 2107

aboedo (approved):

Sources/Security/Signing.swift
aboedo: good catch

NachoSoto: 🤦‍♂️ 
This would have started failing in like 10 days 😅

PR #2777, opened 2023-07-10
`Tests`: fixed flaky failure with asynchronous check

NachoSoto:
The `getCustomerInfo` call is asynchronous, so this expectation should be too.

I also renamed the old property name to "customer".


tonidero (approved):

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
```diff
@@ -65,7 +65,7 @@ class PurchasesLogInTests: BasePurchasesTests {
         self.identityManager.mockLogOutError = nil
         self.backend.overrideCustomerInfoResult = .success(Self.mockLoggedOutInfo)
 
-        expect(self.backend.getSubscriberCallCount) == 1
+        expect(self.backend.getCustomerInfoCallCount).toEventually(equal(1))
```
NachoSoto, line 68: This is the only change

PR #2778, opened 2023-07-10
`IgnoreDecodeErrors`: log decoding error

NachoSoto:
This will be used for decoding the upcoming paywall data.
If we end up with values that we can't parse (like an unknown template), we need to be able to see that in the logs instead of just decoding no paywalls.

I made this `debug` since we expected it and recovered from it.

This also improves the error.
#### Before:
```
DEBUG: ℹ️ Couldn't decode data from json.
Error: The data couldn’t be read because it isn’t in the correct format.
```

#### After:
```
[codable] DEBUG: ℹ️ Couldn't decode 'E' from json.
Error: Swift.DecodingError.dataCorrupted(Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "e", intValue: nil)], debugDescription: "Cannot initialize E from invalid String value e3", underlyingError: nil))
```

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -24,7 +24,6 @@ custom_rules:
   xctestcase_superclass:
     included: ".*\\.swift"
     excluded:
-      - Tests/BackendIntegrationTests
       - Tests/ReceiptParserTests
       - Tests/v3LoadShedderIntegration
     regex: "\\: XCTestCase \\{"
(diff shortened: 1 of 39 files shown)


tonidero (approved):

NachoSoto:
This has exposed the `TestLogHandler` issue even more. Hopefully #2784 will fix it.

PR #2782, opened 2023-07-11
`Trusted Entitlements`: improved documentation

NachoSoto:
I've also added a link to our new documentation page: https://rev.cat/trusted-entitlements

Screenshot 2023-07-11 at 08 47 19

### DIFF
--- a/Sources/DocCDocumentation/DocCDocumentation.docc/Purchases.md
+++ b/Sources/DocCDocumentation/DocCDocumentation.docc/Purchases.md
@@ -115,13 +115,14 @@ Most features require configuring the SDK before using it.
 - ``Purchases/setOnesignalID(_:)``
 
 ### Advanced Configuration
-- ``Purchases/finishTransactions``
+- ``Configuration/Builder/with(observerMode:)``
 - ``Purchases/invalidateCustomerInfoCache()``
 - ``Purchases/forceUniversalAppStore``
 - ``Purchases/proxyURL``
 - ``Purchases/verboseLogs``
 - ``Purchases/verboseLogHandler``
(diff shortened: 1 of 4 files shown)


Sources/Purchasing/Configuration.swift
```diff
@@ -172,18 +172,21 @@ import Foundation
...
 
         /// Set ``Configuration/EntitlementVerificationMode``.
+        ///
         /// Defaults to ``Configuration/EntitlementVerificationMode/disabled``.
         ///
         /// The result of the verification can be obtained from ``EntitlementInfos/verification`` or
         /// ``EntitlementInfo/verification``.
-        ///
-        /// - Note: This requires iOS 13+.
-        /// - Important: This feature is currently in beta.
```
NachoSoto, line 181: I've also removed this.

Sources/DocCDocumentation/DocCDocumentation.docc/Purchases.md
```diff
@@ -115,13 +115,14 @@ Most features require configuring the SDK before using it.
 - ``Purchases/setOnesignalID(_:)``
 
 ### Advanced Configuration
-- ``Purchases/finishTransactions``
```
tonidero, line 118: Was this removal intentional? I guess we now need to set this up with observer mode, so this is probably ok. Just confirming we don't want to keep the old one.

NachoSoto: Oh yeah forgot to mention this. But in going to explain it I realized my logic was flawed, so let me revert it.

PR #2794, opened 2023-07-12
`TimingUtil`: improved tests by using `Clock`

NachoSoto:
These tests will be a lot less flaky now since they won't depend on sleeping the current thread, so they won't be susceptible to slow running CI.
Additionally, we can use `seconds` instead of `milliseconds` which will remove imprecision errors when determining if the time was over the threshold.

### DIFF
--- a/Tests/UnitTests/Misc/TimingUtilTests.swift
+++ b/Tests/UnitTests/Misc/TimingUtilTests.swift
@@ -19,18 +19,23 @@ import XCTest
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
 class TimingUtilAsyncTests: TestCase {
 
+    private var clock: TestClock!
+
     override func setUpWithError() throws {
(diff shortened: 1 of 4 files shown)


Tests/UnitTests/Misc/TimingUtilTests.swift
```diff
@@ -175,22 +187,32 @@ class TimingUtilAsyncTests: TestCase {
             level: level
         )
     }
+
+    private static func asyncMethod() async {}
```
tonidero, line 191: Hmm is this used for something?

tonidero (approved):
Just a question but looks good

Tests/UnitTests/Misc/TimingUtilTests.swift
NachoSoto: Simply to verify that the calls to `measure` in those tests take an `async` closure. They do `await Self.asyncMethod` to verify that.

PR #2797, opened 2023-07-12
`PackageType: Codable` implementation

NachoSoto:
This will be used for decoding it inside of `PaywallData`.

### DIFF
--- a/Tests/UnitTests/Purchasing/PackageTypeTests.swift
+++ b/Tests/UnitTests/Purchasing/PackageTypeTests.swift
@@ -1,0 +1,43 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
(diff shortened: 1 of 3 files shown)


Tests/UnitTests/Purchasing/PackageTypeTests.swift
```diff
@@ -0,0 +1,43 @@
...
+                )
+            } catch {
+                fail("Failed encoding '\(type.debugDescription)': \(error)")
+            }
+        }
+    }
+
+    func testEncodingCustom() throws {
+        // We don't have a way to tell this appart from `.unknown`
+        expect(try PackageType.custom.encodeAndDecode()) == .unknown
```
NachoSoto, line 40: This isn't great, but we don't need anything better at the moment, so I figured I'd least I'd document this limitation in a test.

PR #2798, opened 2023-07-12
`Paywalls`: added `packages` to configuration

NachoSoto:
This allows configuring which package(s) will be displayed in each paywall.


Sources/Networking/HTTPClient/ETagManager.swift
```diff
@@ -33,6 +33,8 @@ class ETagManager {
 
     init(userDefaults: UserDefaults) {
         self.userDefaults = .init(userDefaults: userDefaults)
+
+        self.clearCaches()
```
joshdholtz, line 37: We don't want to commit this, right? 👀

joshdholtz (changes requested):
Looks good but I think I found one thing we don't want to commit 🤷‍♂️

Sources/Networking/HTTPClient/ETagManager.swift
NachoSoto: Fixed.

joshdholtz (approved):

RevenueCatUI/PaywallView.swift
```diff
@@ -26,7 +26,8 @@ public struct PaywallView: View {
 struct PaywallView_Previews: PreviewProvider {
 
     static var previews: some View {
-        PaywallView(offering: TestData.offering, paywall: TestData.paywall)
+        let offering = TestData.offeringWithIntroOffer
+        PaywallView(offering: offering, paywall: offering.paywall!)
```
aboedo, line 30: might as well do an if let here and have a text explaining that there's no offering so the preview doesn't ugly-crash? 

I.e.: 

```swift
        let offering = TestData.offeringWithIntroOffer
        if let paywall = offering.paywall {
            PaywallView(offering: offering, paywall: paywall)
        } else {
            Text("Preview not correctly set up! The offering's paywall need to be non-nil")
        }
```

aboedo (commented):
❤️

RevenueCatUI/PaywallView.swift
NachoSoto: 👍🏻  I guess debugging crashes in previews is annoying, so yeah.

NachoSoto: Done in the next PR.

PR #2800, opened 2023-07-12
`Paywalls`: added header image to configuration

NachoSoto:


RevenueCatUI/Templates/Example1Template.swift
```diff
@@ -71,17 +61,25 @@ private struct Example1TemplateContent: View {
...
-                .resizable()
-                .frame(maxWidth: .infinity)
-                .aspectRatio(1, contentMode: .fill)
-                .edgesIgnoringSafeArea(.top)
-                .padding(.bottom)
-                .mask(alignment: .top) {
-                    Circle()
-                        .offset(y: -160)
-                        .scale(2.5)
+            AsyncImage(url: self.data.headerImageURL) { phase in
```
aboedo, line 64: this is fine for now, but I think sooner or later we'll have to move into a system that pre-downloads and caches the image. Paywalls need to have the smallest possible wait time to load

RevenueCatUI/Templates/Example1Template.swift
```diff
@@ -133,6 +131,8 @@ private struct Example1TemplateContent: View {
         .controlSize(.large)
     }
 
+    private static let imageAspectRatio = 0.7
```
aboedo, line 134: cc @guido732 @lburdock FYI we're hardcoding the aspect ratio here for now

aboedo (commented):
great work on this!

RevenueCatUI/Templates/Example1Template.swift
NachoSoto: Yeah once we figure out what we want to do for the "actual" template we can revisit this.

RevenueCatUI/Templates/Example1Template.swift
NachoSoto: Yup 👍🏻

PR #2801, opened 2023-07-13
`Tests`: unified default timeouts

NachoSoto:
I did this as part of looking into issues running tests on CI.
Ultimately I found the root cause, but this is a useful refactor to make sure all tests have a shared timeout.

### DIFF
--- a/Tests/UnitTests/Networking/HTTPClientTests.swift
+++ b/Tests/UnitTests/Networking/HTTPClientTests.swift
@@ -908,7 +908,7 @@ final class HTTPClientTests: BaseHTTPClientTests<MockETagManager> {
             }
         }
 
-        self.waitForExpectations(timeout: defaultTimeout.seconds * TimeInterval(serialRequests))
+        self.waitForExpectations(timeout: defaultTimeout.seconds)
         expect(completionCallCount.value) == serialRequests
     }
 
--- a/Tests/UnitTests/Networking/HTTPClientTests.swift
+++ b/Tests/UnitTests/Networking/HTTPClientTests.swift
@@ -62,7 +62,7 @@ class BaseHTTPClientTests<ETag: ETagManager>: TestCase {
                           eTagManager: self.eTagManager,
                           signing: self.signing,
                           dnsChecker: MockDNSChecker.self,
-                          requestTimeout: 3)
+                          requestTimeout: defaultTimeout.seconds)
     }
 
 }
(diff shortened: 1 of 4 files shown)


Tests/UnitTests/Purchasing/Purchases/PurchasesPurchasingTests.swift
```diff
@@ -544,25 +544,28 @@ class PurchasesPurchasingTests: BasePurchasesTests {
             try XCTUnwrap(self.offeringsFactory.createOfferings(from: [:], data: .mockResponse))
         )
 
-        var package: Package!
+        let result: Package? = waitUntilValue { completion in
```
NachoSoto, line 547: This replaces `toEventually` with `waitUntilValue` to avoid polling.

Tests/UnitTests/TestHelpers/AsyncTestHelpers.swift
```diff
@@ -88,3 +98,7 @@ func asyncWait(
         throw ConditionFailedError()
     }
 }
+
+// Higher value required to avoid slow CI failing tests.
+let defaultTimeout: DispatchTimeInterval = .seconds(2)
```
NachoSoto, line 103: Also increased the default timeout we were using from 1 second to 2, to avoid flaky failures when CircleCI is being slow.

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -908,7 +908,7 @@ final class HTTPClientTests: BaseHTTPClientTests<MockETagManager> {
             }
         }
 
-        self.waitForExpectations(timeout: 1)
+        self.waitForExpectations(timeout: defaultTimeout.seconds * TimeInterval(serialRequests))
```
tonidero, line 911: I guess this will increase the timeout a lot right? (up to 20 seconds). Just want to confirm we want to do that.

tonidero (approved):

Tests/UnitTests/Networking/HTTPClientTests.swift
NachoSoto, line 911:
```suggestion
        self.waitForExpectations(timeout: defaultTimeout.seconds)
```

NachoSoto: Not needed anymore thanks to #2802

PR #2780, opened 2023-07-10
`Paywalls`: initial configuration types

NachoSoto:
### Changes:
- Added `Offering.paywall`
- Decoding `PaywallData` in `OfferingsResponse`, using `IgnoreDecodeErrors` (this will be better after #2778)
- New `PaywallData` `struct`
- Added new APIs to testers
- Testing paywall deserialization from `Offerings`
- Testing paywall deserialization separately to check edge cases

### Other changes:
- Made `DefaultDecodable` and `AnyDecodable` `Sendable`

PR #2802, opened 2023-07-13
Replaced `FatalErrorUtil` with `Nimble`

NachoSoto:
Turns out that all the testing issues in CI with #2780 were due to these functions.

Screenshot 2023-07-12 at 16 45 37

Even though I wasn't able to reproduce the slowness, I was able to detect the root cause using Instruments, which detects this "severe hang".
With this commit, all the slow test issues go away :tada:

The benefit is we don't have to maintain that hacky code and instead rely on `Nimble` with a much better implementation that use this hack:
```swift
func unreachable() {
    repeat {
        NSRunLoop.currentRunLoop().run()
    } while (true)
}
```

Downsides:
- We lose the ability to verify the assertion message.
- This does not work in `watchOS`


Tests/StoreKitUnitTests/UserDefaultsDefaultTests.swift
```diff
@@ -25,10 +25,7 @@ final class UserDefaultsDefaultTests: TestCase {
     override func setUp() {
         super.setUp()
 
-        // We don't care if `DeviceCache` detects the modification of `UserDefaults` from this test.
-        self.ignoreFatalErrors {
```
NachoSoto, line 29: #2407 got rid of this assertion.

tonidero (approved):

aboedo (approved):
so long, FatalErrorUtil

aboedo:
amazing find!!

PR #2806, opened 2023-07-13
`Tests`: avoid race-condition in leak detection

NachoSoto:
Fixes https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/12839/workflows/2b1e24b9-428a-40ed-bcfd-a94d7591236d/jobs/89720/steps

I noticed when downloading the `.xcresult` that the tests had crashed with this:
```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libswiftCore.dylib            	       0x10bc64589 swift::runtime::AccessSet::insert(swift::runtime::Access*, void*, void*, swift::ExclusivityFlags) + 73
1   libswiftCore.dylib            	       0x10bc647e2 swift_beginAccess + 66
2   BackendIntegrationTests       	       0x13ab37c76 default argument 1 of Expectation.toEventually(_:timeout:pollInterval:description:) + 54
3   BackendIntegrationTests       	       0x13ab37b44 closure #1 in BaseBackendIntegrationTests.verifyPurchasesDoesNotLeak() + 276 (BaseBackendIntegrationTests.swift:167)
```

Turns out that because of the way it was written, normally `Nimble` tries to print the instance during a failure:
```
BasePurchasesTests.swift:116: error: -[UnitTests.PurchasesConfiguringTests testSettingTheDelegateAfterInitializationSendsCachedCustomerInfo] : failed - Purchases has leaked
expected to eventually be nil, got <<RCPurchases: 0x125d13bf0>>
```

That crash points to a race condition inside Swift's runtime. To avoid that, this changes the expectation so `Nimble` would only try to print "expected true, got false", with our own provided description. Therefore making sure that it doesn't try to access the instance while it's being deallocated.

### DIFF
--- a/Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
+++ b/Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
@@ -112,12 +112,14 @@ class BasePurchasesTests: TestCase {
             Purchases.clearSingleton()
             self.clearReferences()
 
-            expect(purchases)
-                .toEventually(beNil(), description: "Purchases has leaked")
-            expect(orchestrator)
-                .toEventually(beNil(), description: "PurchasesOrchestrator has leaked")
-            expect(deviceCache)
-                .toEventually(beNil(), description: "DeviceCache has leaked: \(self)")
+            // Note: this captures the boolean to avoid race conditions when Nimble tries
+            // to print the instances while they're being deallocated.
+            expect { purchases == nil }
+                .toEventually(beTrue(), description: "Purchases has leaked")
+            expect { orchestrator == nil }
+                .toEventually(beTrue(), description: "PurchasesOrchestrator has leaked")
+            expect { deviceCache == nil }
+                .toEventually(beTrue(), description: "DeviceCache has leaked")
         }
     }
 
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
```diff
@@ -113,12 +113,14 @@ class BasePurchasesTests: TestCase {
             Purchases.clearSingleton()
             self.clearReferences()
 
-            expect(purchases)
-                .toEventually(beNil(), description: "Purchases has leaked")
-            expect(orchestrator)
-                .toEventually(beNil(), description: "PurchasesOrchestrator has leaked")
-            expect(deviceCache)
-                .toEventually(beNil(), description: "DeviceCache has leaked: \(self)")
```
NachoSoto, line 121: Bonus points for also not capturing`self` here.

tonidero (approved):

Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
```diff
@@ -113,12 +113,14 @@ class BasePurchasesTests: TestCase {
...
+            // Note: this captures the boolean to avoid race conditions when Nimble tries
+            // to print the instances while they're being deallocated.
```
aboedo, line 117: didn't know it was possible to capture this as a boolean, I thought this block would capture strongly instead. How does this work? like, doesn't it need the object in order to reference when re-evaluating?

NachoSoto: The references are `weak` though (defined above). The previous implementation used an `@autoclosure`, so `expect(purchases)` was converted to `expect { purchases }`.
Internally it was using reflection (possibly `String.init(describing:)`) to print a value for that object.

Now when it generates a description from the subject of the expectation, it's just a boolean.

Let me know if that makes sense.

PR #2807, opened 2023-07-13
`PackageTypeTests`: fixed iOS 12

NachoSoto:
Follow up to #2797.
These tests are failing there because:
```
failed - Failed encoding 'PackageType.unknown': invalidValue(PackageType.unknown, Swift.EncodingError.Context(codingPath: [], debugDescription: "Top-level PackageType encoded as null JSON fragment.", underlyingError: nil))
```

### DIFF
--- a/Tests/UnitTests/Purchasing/PackageTypeTests.swift
+++ b/Tests/UnitTests/Purchasing/PackageTypeTests.swift
@@ -23,9 +23,9 @@ class PackageTypeTests: TestCase {
     func testCodable() throws {
         for type in PackageType.allCases where type != .custom {
             do {
-                let encoded = try type.encodeAndDecode()
+                let encoded = try Data(type: type).encodeAndDecode()
 
-                expect(encoded).to(
+                expect(encoded.type).to(
                     equal(type),
                     description: "Failed encoding '\(type.debugDescription)'"
                 )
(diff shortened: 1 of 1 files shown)


Tests/UnitTests/Purchasing/PackageTypeTests.swift
```diff
@@ -18,8 +18,16 @@ import XCTest
 
 @testable import RevenueCat
 
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
 class PackageTypeTests: TestCase {
 
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+
+        // iOS 12 does not support encoding top level objects as JSON
```
aboedo, line 27: Could you elaborate on this? 
how does the extension work in older versions?

NachoSoto: It works, we just can't test it with these as top level objects.
We'd have to test it encoding something like `[String: PackageType]`.

So solutions:
- Update test so it works in iOS 12 too (I didn't do it because I didn't think it was worth it)
- Improve comment to make that more clear

What do you think?

NachoSoto: @aboedo done.

aboedo (approved):

PR #2810, opened 2023-07-13
`Paywalls`: added tests for `PackageType` filtering

NachoSoto:
Some missing tests for the logic added in #2798.
I've also moved this function to be fined in `PaywallData` instead of the `TemplateViewType` protocol, which makes more sense.

### DIFF
--- a/Tests/RevenueCatUITests/PackageFilteringTests.swift
+++ b/Tests/RevenueCatUITests/PackageFilteringTests.swift
@@ -1,0 +1,72 @@
+//
+//  PackageFilteringTests.swift
+//  
+//
+//  Created by Nacho Soto on 7/13/23.
+//
+
+import Nimble
+import RevenueCat
+@testable import RevenueCatUI
+import XCTest
+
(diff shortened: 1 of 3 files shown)


Tests/RevenueCatUITests/PackageFilteringTests.swift
```diff
@@ -0,0 +1,72 @@
...
+    func testFilterOutSinglePackge() {
+        expect(PaywallData.filter(packages: [Self.monthly], with: [.annual])) == []
+    }
+
+    func testFilterOutNonSubscriptions() {
+        expect(PaywallData.filter(packages: [Self.consumable], with: [.custom])) == []
+    }
+
+    func testFilterByPackageType() {
+        expect(PaywallData.filter(packages: [Self.monthly, Self.annual], with: [.monthly])) == [Self.monthly]
```
aboedo, line 33: with these being lists, perhaps we should also test behavior for repeated values. 
Also I'd add one case where we have more than one value being returned, so that if we had code that looked for `first`, a test would fail

aboedo (approved):
left a couple of ideas for more test cases, but good to go!

PR #2812, opened 2023-07-13
`Paywalls`: added support for purchasing

NachoSoto:
This adds a new `PurchaseHandler` type, which can be instantiated with `Purchases` or with a mock implementation for previews. That gets injected, and templates can now purchase and dismiss themselves.

I've improved the `SimpleApp` setup to better work with the lifetime of purchases. Now it only displays the paywall if the user doesn't have an entitlement.

I've also added some basic error handling to the new `AsyncButton`:
Simulator Screenshot - iPhone 14 Pro - 2023-07-13 at 15 32 06


RevenueCatUI/Templates/Example1Template.swift
```diff
@@ -173,8 +178,12 @@ private struct Example1TemplateContent: View {
 
     @ViewBuilder
     private var button: some View {
-        Button {
+        AsyncButton {
+            let cancelled = try await self.purchaseHandler.purchase(package: self.data.package).userCancelled
 
+            if !cancelled {
+                await self.dismiss()
```
NachoSoto, line 185: This should work regardless of how users display `PaywallView`:
Per the docs, it can:
- Dismiss a modal presentation, like a sheet or a popover.
- Pop the current view from a `NavigationStack
- Close a window that you create with `WindowGroup`

Tests/TestingApps/SimpleApp/SimpleApp/Views/PaywallScreen.swift
```diff
@@ -0,0 +1,31 @@
...
+
+    var offering: Offering
+    var paywall: PaywallData
+
+    @State
+    private var isDisplayed = false
+
+    var body: some View {
+        Rectangle()
+            .hidden()
```
NachoSoto, line 22: Just a placeholder view that can present a sheet.

Tests/TestingApps/SimpleApp/SimpleApp/Views/PaywallScreen.swift
```diff
@@ -0,0 +1,31 @@
...
+            .sheet(isPresented: self.$isDisplayed) {
+                PaywallView(offering: self.offering, paywall: self.paywall)
+            }
+            .onAppear {
+                self.isDisplayed = true
```
NachoSoto, line 27: Displayed automatically.

Tests/TestingApps/SimpleApp/SimpleApp/Views/AppContentView.swift
```diff
@@ -0,0 +1,59 @@
...
+            }
+            .padding()
+
+            Rectangle()
+                .foregroundStyle(.orange.gradient)
+                .opacity(0.2)
+                .frame(maxWidth: .infinity, maxHeight: .infinity)
+                .edgesIgnoringSafeArea(.all)
+        }
+    }
```
NachoSoto, line 41: This is this simple screen:
simulator_screenshot_4F32B1F0-9F4A-4DBE-8821-0233B1540DEF

RevenueCatUI/Views/AsyncButton.swift
```diff
@@ -0,0 +1,95 @@
...
+    init(error: NSError) {
+        self.underlyingError = error
+    }
+
+    var errorDescription: String? {
+        return "\(self.underlyingError.domain) \(self.underlyingError.code)"
+    }
+
+    var failureReason: String? {
+        if let errorCode = self.underlyingError as? ErrorCode {
```
NachoSoto, line 84: The good thing about these being our own errors is that we can even improve this error handling.
Eventually we should provide localizations for each of our errors too.

RevenueCatUI/PaywallView.swift
```diff
@@ -8,29 +8,38 @@ public struct PaywallView: View {
     private let offering: Offering
```
aboedo, line 8: I feel like we might have to rethink the naming for these at one point to minimize collisions with paywalls a customer might already have. 

Like, say you have a paywall in place and are looking to migrate into ours, it'd suck if you have to rename or remove the one you already have right away in order to test out our SDK

aboedo (approved):

RevenueCatUI/Templates/Example1Template.swift
NachoSoto: Good point, especially from what we talked about today.
Especially for that I think we should add a sort of "event handler" / SwiftUI property like StoreKit views.

RevenueCatUI/PaywallView.swift
NachoSoto: Good point. I'd say before we merge this and ship it, we can go through the entire public API (won't be much) and make sure it's solid and documentation is good.

PR #2814, opened 2023-07-14
`MockDeviceCache`: avoid using real `UserDefaults`

NachoSoto:
I noticed that in some tests that shouldn't be using real `UserDefaults` this was being logged, even multiple times per test:
> Configuring SDK using RevenueCat's UserDefaults suite.


aboedo (approved):
😅

PR #2818, opened 2023-07-14
`PaywallData`: ignore empty strings in `LocalizedConfiguration`

NachoSoto:
Uses the new wrapper added by #2819.

### DIFF
--- a/Sources/Paywalls/PaywallData.swift
+++ b/Sources/Paywalls/PaywallData.swift
@@ -51,6 +51,8 @@ public protocol PaywallLocalizedConfiguration {
 
 }
 
+// swiftlint:disable identifier_name
+
 extension PaywallData {
 
     /// Defines the necessary localized information for a paywall.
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/Networking/Responses/Fixtures/PaywallData-Sample1.json
```diff
@@ -13,7 +13,8 @@
             "title": "Tienda",
             "subtitle": "Descripción",
             "call_to_action": "Comprar",
-            "offer_details": "{{ price_per_month }} cada mes"
+            "offer_details": "{{ price_per_month }} cada mes",
+            "offer_details_with_intro_offer": " "
```
NachoSoto, line 17: This makes the existing test still validate that this is becoming `nil`.

Sources/Paywalls/PaywallData.swift
```diff
@@ -62,13 +64,21 @@ extension PaywallData {
         public var title: String
         public var subtitle: String
         public var callToAction: String
-        public var callToActionWithIntroOffer: String?
         public var offerDetails: String
-        public var offerDetailsWithIntroOffer: String?
-
-        // swiftlint:enable missing_docs
+        @NonEmptyStringDecodable
+        var _callToActionWithIntroOffer: String?
```
NachoSoto, line 69: These can't be `public` because the property wrapper isn't, so they get exposed separately from the property that's decoded/encoded.

aboedo (approved):

PR #2819, opened 2023-07-14
New `@NonEmptyStringDecodable`

NachoSoto:
This allows decorating a `String` property in a `Decodable` type so it ensures it gets converted to `nil` if it's empty.
It will be used for `PaywallData`.


aboedo (approved):
❤️

PR #2821, opened 2023-07-14
`Paywalls`: disable `macOS`/`macCatalyst`/`watchOS` for now

NachoSoto:
In the future we'll be able to polish support for those platforms.


RevenueCatUI/PaywallView.swift
```diff
@@ -3,6 +3,8 @@ import SwiftUI
 
 /// A full-screen SwiftUI view for displaying a `PaywallData` for an `Offering`.
 @available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
```
NachoSoto, line 5: This version is still required, otherwise the compiler throws compatibility errors with APIs used inside 🤷🏻

joshdholtz (approved):

PR #2825, opened 2023-07-17
`Paywalls`: update `codecov.yml`

NachoSoto:
None of this code will be built / tested by the job uploading code coverage (`run-test-ios-16`). So it doesn't make sense to report it.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,6 +1,17 @@
 codecov:
   require_ci_to_pass: yes
 
+ignore:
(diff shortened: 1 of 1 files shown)


vegaro (approved):

NachoSoto:
I see that `Codecov` already only has `Sources`:https://app.codecov.io/gh/RevenueCat/purchases-ios/tree/paywalls

Probably because those are the only files in the coverage result. So closing this.

PR #2828, opened 2023-07-17
`CI`: make all `Codecov` jobs `informational`

NachoSoto:
Follow up to #2670. Reading through [the docs](https://docs.codecov.com/docs#tips-and-tricks) I realized we were missing this too.


vegaro (approved):
I think it makes sense and we shouldn't block on coverage changes

PR #2831, opened 2023-07-17
`@EnsureNonEmptyArrayDecodable`

NachoSoto:
This allows us to define properties in a `Decodable` type and ensure that they're not empty.
Example
```swift
struct Data: Codable {
    @EnsureNonEmptyArrayDecodable
    var value: [String]
}
```

Used for #2832.

### DIFF
--- a/Sources/Misc/Codable/EnsureNonEmptyArrayDecodable.swift
+++ b/Sources/Misc/Codable/EnsureNonEmptyArrayDecodable.swift
@@ -1,0 +1,69 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  EnsureNonEmptyArrayDecodable.swift
+//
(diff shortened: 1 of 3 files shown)


Sources/Misc/Codable/EnsureNonEmptyArrayDecodable.swift
```diff
@@ -0,0 +1,69 @@
...
+extension EnsureNonEmptyArrayDecodable: Hashable where Value: Hashable {}
+
+extension EnsureNonEmptyArrayDecodable: Decodable {
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.singleValueContainer()
+        let array = try container.decode([Value].self)
+
+        if array.isEmpty {
+            throw Error()
```
tonidero, line 42: This seems a bit of a generic error... Would it be possible to include the property that is causing this error?

tonidero (approved):
Looks good, just a suggestion

Sources/Misc/Codable/EnsureNonEmptyArrayDecodable.swift
NachoSoto: It’s `EnsureNonEmptyArray.Error`. When you decode the top level object `JSONDecoder` will include the information for what key failed :)

PR #2832, opened 2023-07-17
`Paywalls`: add support for multiple images in template configuration

NachoSoto:
Changed `PaywallData.Configuration.headerImageName` to `imageNames`.
This more flexible definition allows future templates to display a carrousel of images instead of a single image.

Depends on


Tests/UnitTests/Networking/Responses/Fixtures/PaywallData-Sample1.json
```diff
@@ -20,7 +20,7 @@
     "default_locale": "en_US",
     "config": {
         "packages": ["$rc_monthly", "$rc_annual"],
-        "header_image": "asset_name.png",
```
NachoSoto, line 23: @guido732 @lburdock as discussed in the doc, this is now an array. The current template only supports 1 but allows us to display multiple images in future templates.
I've added validation when decoding it to ensure that this is not empty. It would be nice if the frontend did the same to make sure we never save a template without an image.

Tests/UnitTests/Paywalls/PaywallDataTests.swift
```diff
@@ -79,6 +81,12 @@ class PaywallDataTests: BaseHTTPResponseTest {
         expect(localization.title) == "Tienda"
     }
 
+    func testFailsToDecodeWithNoImages() throws {
+        expect {
+            let _: PaywallData = try self.decodeFixture("PaywallData-empty_images")
+        }.to(throwError(EnsureNonEmptyArrayDecodable<String>.Error()))
```
NachoSoto, line 87: Hmm this should probably be a more precise error instead.

PR #2834, opened 2023-07-18
`Paywalls`: support for multiple `PaywallViewMode`s

NachoSoto:
This adds 2 new modes: `.square` and `.banner`:

### `.card`:
iOS16-testSquarePaywall 1
### `.banner`:
iOS16-testBannerPaywall 1

-------

I've updated `Example1Template` to support both as an example, and added that as snapshot tests.

This also adds a new constructor to `PaywallView` that doesn't require passing an `Offering` or `PaywallData`, it automatically loads that from `Purchases.shared`. With that, embedding these mini-paywalls is trivial:

```swift
struct MyApp: View {
  var body: some View {
    MyOtherViews()
    if !customerInfo.hasPro {
      PaywallView(mode: .card)
    }
  }
}
```

I've added these as examples to the testing app:

simulator_screenshot_84270E1A-F7C6-4C5D-8773-F4F2E646CB0F


Tests/TestingApps/SimpleApp/SimpleApp/Views/SampleChart.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  SampleChart.swift
+//  SimpleApp
+//
+//  Created by Nacho Soto on 7/17/23.
+//
+
+import SwiftUI
+
+struct BarChartView: View {
```
NachoSoto, line 10: This is just for the `SimpleApp`, to have a "main app view" that's not too empty and makes having mini-paywalls there make a bit more sense.

RevenueCatUI/PaywallView.swift
```diff
@@ -6,52 +6,145 @@ import SwiftUI
...
+                        purchaseHandler: purchaseHandler
+                    )
+                } else {
+                    DebugErrorView("Offering '\(offering.identifier)' has no configured paywall",
+                                   releaseBehavior: .emptyView)
+                }
+            } else {
+                self.loadingView
+                    .task {
+                        // TODO: error handling
```
NachoSoto, line 76: This is the only thing left in the PR. Need to figure out what to do in this case if loading offerings fails.

RevenueCatUI/PaywallView.swift
```diff
@@ -6,52 +6,145 @@ import SwiftUI
...
+                    LoadedOfferingPaywallView(
+                        offering: offering,
+                        paywall: paywall,
+                        mode: mode,
+                        introEligibility: checker,
+                        purchaseHandler: purchaseHandler
+                    )
+                } else {
+                    DebugErrorView("Offering '\(offering.identifier)' has no configured paywall",
+                                   releaseBehavior: .emptyView)
```
NachoSoto, line 71: For cases when developers do this:
```swift
PaywallView(mode: .banner)
```

But the offering doesn't have any paywalls.
In debug builds this will show a red banner with an error.
In release builds, it'll just be empty.

RevenueCatUI/Data/PaywallViewMode.swift
```diff
@@ -13,4 +13,26 @@ public enum PaywallViewMode {
     /// Paywall is displayed full-screen, with as much information as available.
     case fullScreen
 
+    /// Paywall is displayed with a square aspect ratio. It can be embedded inside any other SwiftUI view.
+    case square
```
joshdholtz, line 17: I'd also like this be renamed to like `card`? 🤔 `square` feels a little to concrete on the design and not the use case 🤷‍♂️

joshdholtz (approved):
Code looks good! Left a small thought on the `square` naming but not a blocker 😊

RevenueCatUI/Data/PaywallViewMode.swift
NachoSoto: I love `card` 👌🏻

RevenueCatUI/PaywallView.swift
NachoSoto: I'll do this in a separate PR.

PR #2839, opened 2023-07-19
`Paywalls`: extracted `PurchaseButton`

NachoSoto:
Another small refactor, follow up to #2837.
Slowly taking implementation details from `Example1Template` to make it easier to implement more templates.

### DIFF
--- a/RevenueCatUI/Data/TestData.swift
+++ b/RevenueCatUI/Data/TestData.swift
@@ -114,7 +114,7 @@ internal enum TestData {
     )
 
     #if canImport(SwiftUI) && canImport(UIKit)
-    static let colors: PaywallData.Configuration.Colors = .comine(light: Self.lightColors, dark: Self.darkColors)
+    static let colors: PaywallData.Configuration.Colors = .combine(light: Self.lightColors, dark: Self.darkColors)
     #endif
 
     static let customerInfo: CustomerInfo = {
--- a/RevenueCatUI/Data/TestData.swift
+++ b/RevenueCatUI/Data/TestData.swift
@@ -203,10 +203,11 @@ extension PurchaseHandler {
 
     /// Creates a copy of this `PurchaseHandler` with a delay.
     func with(delay: Duration) -> Self {
-        return .init { [purchaseBlock = self.purchaseBlock] in
-            try? await Task.sleep(for: delay)
+        return self.map { purchaseBlock in {
+                try? await Task.sleep(for: delay)
 
-            return try await purchaseBlock($0)
+                return try await purchaseBlock($0)
+            }
         }
     }
 }
--- a/RevenueCatUI/Data/TestData.swift
+++ b/RevenueCatUI/Data/TestData.swift
@@ -10,6 +10,7 @@ import RevenueCat
 
 #if DEBUG
 
+@available(iOS 14.0, macOS 11.0, tvOS 14.0, watchOS 7.0, *)
 internal enum TestData {
 
     static let productWithIntroOffer = TestStoreProduct(
@@ -112,6 +113,10 @@ internal enum TestData {
         callToActionForeground: "#000000"
     )
 
+    #if canImport(SwiftUI) && canImport(UIKit)
+    static let colors: PaywallData.Configuration.Colors = .comine(light: Self.lightColors, dark: Self.darkColors)
+    #endif
+
     static let customerInfo: CustomerInfo = {
         let json = """
         {
@@ -145,11 +150,11 @@ internal enum TestData {
         return try! decoder.decode(CustomerInfo.self, from: Data(json.utf8))
     }()
 
-    private static let localization: PaywallData.LocalizedConfiguration = .init(
+    static let localization: PaywallData.LocalizedConfiguration = .init(
         title: "Ignite your child's curiosity",
         subtitle: "Get access to all our educational content trusted by thousands of parents.",
         callToAction: "Purchase for {{ price }}",
-        callToActionWithIntroOffer: nil,
+        callToActionWithIntroOffer: "Purchase for {{ price_per_month }} per month",
         offerDetails: "{{ price_per_month }} per month",
         offerDetailsWithIntroOffer: "Start your {{ intro_duration }} trial, then {{ price_per_month }} per month"
     )
--- a/RevenueCatUI/Helpers/ColorInformation+MultiScheme.swift
+++ b/RevenueCatUI/Helpers/ColorInformation+MultiScheme.swift
@@ -21,7 +21,7 @@ extension PaywallData.Configuration.ColorInformation {
             return light
         }
 
-        return .comine(light: light, dark: dark)
+        return .combine(light: light, dark: dark)
     }
 
 }
@@ -29,7 +29,7 @@ extension PaywallData.Configuration.ColorInformation {
 extension PaywallData.Configuration.Colors {
 
     @available(iOS 14.0, macOS 11.0, tvOS 14.0, watchOS 7.0, *)
-    static func comine(light: Self, dark: Self) -> Self {
+    static func combine(light: Self, dark: Self) -> Self {
         return .init(
             background: .init(light: light.background, dark: dark.background),
             foreground: .init(light: light.foreground, dark: dark.foreground),
--- a/RevenueCatUI/Helpers/ColorInformation+MultiScheme.swift
+++ b/RevenueCatUI/Helpers/ColorInformation+MultiScheme.swift
@@ -21,6 +21,15 @@ extension PaywallData.Configuration.ColorInformation {
             return light
         }
 
+        return .comine(light: light, dark: dark)
+    }
+
+}
+
+extension PaywallData.Configuration.Colors {
+
+    @available(iOS 14.0, macOS 11.0, tvOS 14.0, watchOS 7.0, *)
+    static func comine(light: Self, dark: Self) -> Self {
         return .init(
             background: .init(light: light.background, dark: dark.background),
             foreground: .init(light: light.foreground, dark: dark.foreground),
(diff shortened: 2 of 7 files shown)


RevenueCatUI/Views/AsyncButton.swift
```diff
@@ -11,7 +11,7 @@ import SwiftUI
 @available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
 struct AsyncButton<Label>: View where Label: View {
 
-    typealias Action = @Sendable () async throws -> Void
+    typealias Action = @Sendable @MainActor () async throws -> Void
```
NachoSoto, line 14: This ensures that the action is always `@MainActor` so usages don't have to make it explicit.

RevenueCatUI/Views/PurchaseButton.swift
```diff
@@ -0,0 +1,158 @@
...
+    private var purchased = false
+    @Environment(\.dismiss)
+    private var dismiss
+
+    var body: some View {
+        // Workaround for SwiftUI bug
+        VStack {
+            if self.purchased {
+                // TODO: what to display?
+                Text("Thanks!")
```
NachoSoto, line 31: We need to figure out a good UX for this.

RevenueCatUI/Views/PurchaseButton.swift
```diff
@@ -0,0 +1,158 @@
...
+    private var button: some View {
+        AsyncButton {
+            let cancelled = try await self.purchaseHandler.purchase(package: self.package).userCancelled
+
+            if !cancelled {
+                if case .fullScreen = self.mode {
+                    self.dismiss()
+                } else {
+                    self.purchased = true
+                }
```
NachoSoto, line 47: @aboedo you mentioned handling this for not-fullscreen paywalls.

RevenueCatUI/Views/PurchaseButton.swift
```diff
@@ -0,0 +1,158 @@
...
+    let introEligibility: IntroEligibilityStatus?
+    let mode: PaywallViewMode
+
+    @State
+    private var purchased = false
+    @Environment(\.dismiss)
+    private var dismiss
+
+    var body: some View {
+        // Workaround for SwiftUI bug
```
NachoSoto, line 27: TODO: file a Radar for https://twitter.com/nachosoto/status/1681447887077801989?s=61&t=xbQZk0UV1WUriSSGU7EvJw

RevenueCatUI/Views/PurchaseButton.swift
```diff
@@ -0,0 +1,158 @@
...
+        private static let eligibilityChecker: TrialOrIntroEligibilityChecker = .producing(eligibility: .eligible)
+            .with(delay: .seconds(0.3))
+        private static let purchaseHandler: PurchaseHandler = .mock()
+            .with(delay: .seconds(0.5))
+
+    }
+
+    static var previews: some View {
+        ForEach(PaywallViewMode.allCases, id: \.self) { mode in
+            Preview(mode: mode)
```
NachoSoto, line 151: Having small previewable views makes it easy to iterate on each small component.

RevenueCatUI/PaywallView.swift
```diff
@@ -129,6 +132,16 @@ private struct LoadedOfferingPaywallView: View {
...
+    private var shouldHidePaywall: Bool {
+        switch self.mode {
+        case .fullScreen:
+            return false
+
+        case .card, .banner:
+            return self.purchaseHandler.purchased
+        }
+    }
+
```
NachoSoto, line 144: @aboedo this looks pretty good for now :) and it gets animated.

RevenueCatUI/Data/TestData.swift
```diff
@@ -198,10 +203,11 @@ extension PurchaseHandler {
...
     func with(delay: Duration) -> Self {
-        return .init { [purchaseBlock = self.purchaseBlock] in
-            try? await Task.sleep(for: delay)
+        return self.map { purchaseBlock in {
+                try? await Task.sleep(for: delay)
 
-            return try await purchaseBlock($0)
+                return try await purchaseBlock($0)
+            }
         }
```
NachoSoto, line 211: Small refactor because I made the `block` `private`.

RevenueCatUI/PaywallView.swift
aboedo: wonder if we should even consider making that variable public so that you can use it in the app

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -27,11 +31,24 @@ final class PurchaseHandler: ObservableObject {
...
 
     func purchase(package: Package) async throws -> PurchaseResultData {
-        return try await self.purchaseBlock(package)
+        let result = try await self.purchaseBlock(package)
+
+        if !result.userCancelled {
+            withAnimation(Constants.defaultAnimation) {
+                self.purchased = true
+            }
+        }
```
aboedo, line 44: I love that even this bit is something that we're handling now

RevenueCatUI/Views/PurchaseButton.swift
```diff
@@ -0,0 +1,144 @@
...
+
+    private var button: some View {
+        AsyncButton {
+            let cancelled = try await self.purchaseHandler.purchase(package: self.package).userCancelled
+
+            if !cancelled, case .fullScreen = self.mode {
+                self.dismiss()
+            }
+        } label: {
+            IntroEligibilityStateView(
```
aboedo, line 36: Feels like this could be more appropriately named something like CTAText view

RevenueCatUI/Views/PurchaseButton.swift
```diff
@@ -0,0 +1,144 @@
...
+                ? .infinity
+                : nil
+            )
+        }
+        .font(self.mode.buttonFont.weight(.semibold))
+        .tint(self.colors.callToActionBackgroundColor.gradient)
+        .buttonBorderShape(self.mode.buttonBorderShape)
+        .controlSize(self.mode.buttonSize)
+        .buttonStyle(.borderedProminent)
+        .frame(maxWidth: .infinity)
```
aboedo, line 54: feels somewhat confusing that `self.mode.fullWidthButton` doesn't refer to this width but rather the one for the button's label

RevenueCatUI/Views/PurchaseButton.swift
```diff
@@ -0,0 +1,144 @@
...
+
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+private extension PaywallViewMode {
+
+    var buttonFont: Font {
+        switch self {
+        case .fullScreen, .card: return .title2
+        case .banner: return .footnote
+        }
+    }
```
aboedo, line 67: how do you envision this scaling to different templates?

RevenueCatUI/Helpers/ColorInformation+MultiScheme.swift
```diff
@@ -21,6 +21,15 @@ extension PaywallData.Configuration.ColorInformation {
...
 
+        return .comine(light: light, dark: dark)
+    }
+
+}
+
+extension PaywallData.Configuration.Colors {
+
+    @available(iOS 14.0, macOS 11.0, tvOS 14.0, watchOS 7.0, *)
+    static func comine(light: Self, dark: Self) -> Self {
```
aboedo, line 32: what's `comine`? I've never heard the term before. Is this meant to be `combine`?

aboedo (approved):
left a few comments, looks good overall

RevenueCatUI/Views/PurchaseButton.swift
NachoSoto: Probably would need to inject this font to `PurchaseButton` instead.

RevenueCatUI/Helpers/ColorInformation+MultiScheme.swift
NachoSoto: It's a common term in some secret functional programming circles......

But yes 💩

RevenueCatUI/Views/PurchaseButton.swift
NachoSoto: Yeah it's a bad name, but it's not just for the CTA. 
From the docstring:
> /// A view that can process intro eligibility and display different data based on the result.

It's used for CTA but also for other labels. It's just a label that uses a loading intro eligibility and then decides what to display. Better name suggestions welcome.

NachoSoto: Actually that's what you approved: #2837
I'll hold on in case you have a suggestion for a better name though.

RevenueCatUI/Data/TestData.swift
```diff
@@ -112,6 +113,10 @@ internal enum TestData {
         callToActionForeground: "#000000"
     )
 
+    #if canImport(SwiftUI) && canImport(UIKit)
+    static let colors: PaywallData.Configuration.Colors = .comine(light: Self.lightColors, dark: Self.darkColors)
+    #endif
```
joshdholtz, line 118: What happens here if these aren't supported? 🤔

RevenueCatUI/Helpers/ColorInformation+MultiScheme.swift
joshdholtz: I have also never heard of it and I know feel better that @aboedo also doesn't know 😛

joshdholtz (approved):

RevenueCatUI/Data/TestData.swift
NachoSoto: The views don’t support it either. But when/if we do, the compiler will fail. So it’ll force us to use something else (in practice we’ll expose an NSColor for example).

RevenueCatUI/Views/PurchaseButton.swift
NachoSoto: Merging this, but let me know if you can think of a better name.

RevenueCatUI/Views/PurchaseButton.swift
NachoSoto: That's just an implementation detail for how SwiftUI's layout works.

PR #2841, opened 2023-07-19
`TransactionPoster`: don't finish transactions for non-subscriptions if they're not processed

NachoSoto:
This ensures that if the server returns a 200, but didn't actually process the transaction, we don't lose it forever.

When this happens, a warning will be logged:
```
[purchases] DEBUG: ℹ️ TransactionPoster: handling transaction for product 'BA92DCD8-74C2-45FF-AEE4-2FFD036AF827'
[purchases] WARN: ⚠️ Transaction '19B3BD23-181B-442A-A5B8-05D686ED98E2' will not be finished: it's a non-subscription and it's missing in CustomerInfo list: []
```

#### Dependencies:
- #3008
- #3009

### DIFF
--- a/Sources/Identity/CustomerInfo+NonSubscriptions.swift
+++ b/Sources/Identity/CustomerInfo+NonSubscriptions.swift
@@ -1,0 +1,24 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  CustomerInfo+NonSubscriptions.swift
+//
+//  Created by Nacho Soto on 7/18/23.
+
+import Foundation
+
+extension CustomerInfo {
+
+    func containsNonSubscription(_ transation: StoreTransactionType) -> Bool {
+        return self.nonSubscriptions.contains {
+            $0.transactionIdentifier == transation.transactionIdentifier
+        }
+    }
+
+}
--- a/Sources/Logging/Strings/PurchaseStrings.swift
+++ b/Sources/Logging/Strings/PurchaseStrings.swift
@@ -31,6 +31,8 @@ enum PurchaseStrings {
     case entitlements_revoked_syncing_purchases(productIdentifiers: [String])
     case entitlement_expired_outside_grace_period(expiration: Date, reference: Date)
     case finishing_transaction(StoreTransactionType)
+    case finish_transaction_skipped_because_its_missing_in_non_subscriptions(StoreTransactionType,
+                                                                             [NonSubscriptionTransaction])
     case purchasing_with_observer_mode_and_finish_transactions_false_warning
     case paymentqueue_revoked_entitlements_for_product_identifiers(productIdentifiers: [String])
     case paymentqueue_adding_payment(SKPaymentQueue, SKPayment)
@@ -129,6 +131,10 @@ extension PurchaseStrings: LogMessage {
         case let .finishing_transaction(transaction):
             return "Finishing transaction '\(transaction.transactionIdentifier)' " +
             "for product '\(transaction.productIdentifier)'"
+
+        case let .finish_transaction_skipped_because_its_missing_in_non_subscriptions(transaction, nonSubscriptions):
+            return "Transaction '\(transaction.transactionIdentifier)' will not be finished: " +
+            "it's a non-subscription and it's missing in CustomerInfo list: \(nonSubscriptions)"
 
         case .purchasing_with_observer_mode_and_finish_transactions_false_warning:
             return "Observer mode is active (finishTransactions is set to false) and " +
--- a/Sources/Purchasing/Purchases/TransactionPoster.swift
+++ b/Sources/Purchasing/Purchases/TransactionPoster.swift
@@ -137,6 +137,8 @@ final class TransactionPoster: TransactionPosterType {
 
         switch product.productCategory {
         case .subscription:
+            // Note: this includes non-renewing subscriptions. Those are included in `.nonSubscriptions`,
+            // but we can't tell them apart using `product.productType` because that's unknown for SK1 products.
             return true
 
         case .nonSubscription:
--- a/Sources/Purchasing/Purchases/TransactionPoster.swift
+++ b/Sources/Purchasing/Purchases/TransactionPoster.swift
@@ -124,6 +124,40 @@ final class TransactionPoster: TransactionPosterType {
         transaction.finish(self.paymentQueueWrapper.paymentQueueWrapperType, completion: complete)
     }
 
+    static func shouldFinish(
+        transaction: StoreTransactionType,
+        for product: StoreProductType?,
+        customerInfo: CustomerInfo
+    ) -> Bool {
+        // Don't finish transactions if CustomerInfo was computed offline
+        guard !customerInfo.isComputedOffline else { return false }
+
+        // If we couldn't find the product, we can't determine if it's a consumable
+        guard let product = product else { return true }
+
+        switch product.productCategory {
+        case .subscription:
+            return true
+
+        case .nonSubscription:
+            // Only finish consumables if the server actually processed it.
+            let shouldFinish = (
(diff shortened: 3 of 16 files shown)


Sources/Identity/CustomerInfo+NonSubscriptions.swift
```diff
@@ -0,0 +1,25 @@
...
+//
+//  Created by Nacho Soto on 7/18/23.
+
+import Foundation
+
+extension CustomerInfo {
+
+    func containsNonSubscription(_ transation: StoreTransactionType) -> Bool {
+        return self.nonSubscriptions.contains {
+            $0.transactionIdentifier == transation.transactionIdentifier
```
NachoSoto, line 20: @tonycosentini this is ready. Just gotta change to use the real store transaction ID not the RC ID.
Also I wonder if we should restrict this by sandbox and store too /cc @aboedo

Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
```diff
@@ -291,6 +324,14 @@ extension BaseStoreKitIntegrationTests {
...
     }
 
+    // This should be "Finishing transaction".
+    static let finishingTransactionLog: String = Strings.purchase
+        .finishing_transaction(MockStoreTransaction())
+        .description
+        .components(separatedBy: " ")
+        .prefix(2)
+        .joined(separator: " ")
+
```
NachoSoto, line 334: Welcome to feedback if anybody thinks this is overkill.

NachoSoto:
Waiting on @tonycosentini to change transaction identifiers for StoreKit tests so we can verify this in integration tests.

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -140,6 +140,8 @@ final class TransactionPoster: TransactionPosterType {
             return true
 
         case .nonSubscription:
+            // TODO: if the underlying transaction doesn't have a real identifier
```
NachoSoto, line 143: Tests are passing now. Just gotta do these TODOs.

NachoSoto:
This should be ready.

Sources/Identity/CustomerInfo+NonSubscriptions.swift
joshdholtz: Does non subscription here mean both a consumable and a non-consumable? 🤔 

Can the answer to this be added as a comment to this function?

Sources/Logging/Strings/PurchaseStrings.swift
```diff
@@ -130,6 +132,10 @@ extension PurchaseStrings: LogMessage {
             return "Finishing transaction '\(transaction.transactionIdentifier)' " +
             "for product '\(transaction.productIdentifier)'"
 
+        case let .finish_transaction_skipped_for_missing_consumable(transaction, nonSubscriptions):
+            return "Transaction '\(transaction.transactionIdentifier)' will not be finished: " +
+            "it's consumable, but it's missing from list of non-subscriptions: \(nonSubscriptions)"
```
joshdholtz, line 137:
```suggestion
            "it's a consumable, but it's missing from list of non-subscriptions: \(nonSubscriptions)"
```

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -124,6 +124,40 @@ final class TransactionPoster: TransactionPosterType {
...
+            return true
+
+        case .nonSubscription:
+            // Only finish consumables if the server actually processed it.
+            let shouldFinish = (
+                !transaction.hasKnownTransactionIdentifier ||
+                customerInfo.nonSubscriptions.contains {
+                    $0.storeTransactionIdentifier == transaction.transactionIdentifier
+                }
+            )
```
joshdholtz, line 149: My head is still getting stuck here if `nonSubscription` means consumable or non-consumable and if they should be treated the same 🤷‍♂️

joshdholtz (commented):
This makes sense to me! I think I really have questions about if and how consumables and non-consumables are being treated here 🤷‍♂️ The title of the PR mentions consumables but the code refers to non-subscriptions.

Sources/Purchasing/NonSubscriptionTransaction.swift
```diff
@@ -23,22 +23,38 @@ public final class NonSubscriptionTransaction: NSObject {
     /// The date that App Store charged the user’s account.
     @objc public let purchaseDate: Date
 
-    /// The unique identifier for the transaction.
+    /// The unique identifier for the transaction created by RevenueCat.
     @objc public let transactionIdentifier: String
 
+    /// The unique identifier for the transaction created by the store.
+    @objc public let storeTransactionIdentifier: String
```
tonidero, line 30: Should we add this to the PR description as well?

Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
aboedo: lol yeah, kinda. I'd either just look for "Finishing transaction" or maybe update the string in the log so it's meaningful to us (like adding `action: finish_transaction`) to the log or something

Like, if you wanted to change that string it'd be easier to understand that a change like "will finish transaction" breaks a place where we're explicitly looking for "Finishing Transaction", than it is to understand why it broke this line

Sources/Purchasing/Purchases/TransactionPoster.swift
NachoSoto: Answer to this (will test to make sure), `non_subscriptions` include:
- Consumables
- Non-consumables
- Non-renewing subscriptions

Sources/Logging/Strings/PurchaseStrings.swift
NachoSoto: TODO: update this.

Sources/Identity/CustomerInfo+NonSubscriptions.swift
NachoSoto: TODO: we discovered that they're included, so test those too.

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -124,6 +124,40 @@ final class TransactionPoster: TransactionPosterType {
...
+        switch product.productCategory {
+        case .subscription:
+            return true
+
+        case .nonSubscription:
+            // Only finish consumables if the server actually processed it.
+            let shouldFinish = (
+                !transaction.hasKnownTransactionIdentifier ||
+                customerInfo.nonSubscriptions.contains {
+                    $0.storeTransactionIdentifier == transaction.transactionIdentifier
```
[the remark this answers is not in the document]
NachoSoto: Totally a good point, but I'm not sure it's worth deprecating it.
I extracted this to #3009, and you can see I made that clear in the docstring for both properties. What do you think?

Sources/Identity/CustomerInfo+NonSubscriptions.swift
NachoSoto: I updated the PR, logs, and code to make this more clear. I was mixing consumables with non-subscriptions everywhere.

> We might wanna do it for non-renewing subscriptions as well, since they have the same characteristics as consumables in that they disappear from the receipt forever.
> 
> And lastly, non-consumables don't actually need this, do they get the same treatment?

See `TransactionPoster.shouldFinish(transaction:for:customerInfo:)`. That new logic is based on `StoreProduct.productCategory` (subscription / non-subscriptions). We can't use `ProductType` because we can't determine that for `SK1` products (see `SK1StoreProduct.productType`).

NachoSoto:
This is ready now.

Sources/Identity/CustomerInfo+NonSubscriptions.swift
aboedo: Just a note that we _could_ in theory distinguish between non-renewing and auto-renewing subs by mixing in the subs vs non subs distinction + expiration date != nil, but it's a non-trivial change that feels a bit more risky, and it isn't worth the effort.

aboedo (approved):

PR #2847, opened 2023-07-20
`Paywalls`: fuzzy `Locale` lookups

NachoSoto:
`Locale.current` might only report a language and not a region. For example when configuring the scheme like this:
Screenshot 2023-07-21 at 10 23 37

iOS localization supports this, and so should we.

### DIFF
--- a/Sources/Paywalls/PaywallData.swift
+++ b/Sources/Paywalls/PaywallData.swift
@@ -99,8 +99,12 @@ extension PaywallData {
     }
 
     /// - Returns: ``PaywallData/LocalizedConfiguration-swift.struct`` for the given `Locale`, if found.
-    public func config(for locale: Locale) -> LocalizedConfiguration? {
-        return self.localization[locale.identifier]
+    /// - Note: this allows searching by `Locale` with only language code and missing region (like `en`, `es`, etc).
+    public func config(for requiredLocale: Locale) -> LocalizedConfiguration? {
+        self.localization[requiredLocale.identifier] ??
+        self.localization.first { locale, _ in
+            Locale(identifier: locale).sharesLanguageCode(with: requiredLocale)
+        }?.value
     }
 
     /// The default `Locale` used if `Locale.current` is not configured for this paywall.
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/Paywalls/PaywallDataTests.swift
```diff
@@ -71,6 +71,14 @@ class PaywallDataTests: BaseHTTPResponseTest {
         expect(paywall.config(for: Locale(identifier: "gl_ES"))).to(beNil())
     }
 
+
+    func testFindsLocaleWithOnlyLanguage() throws {
+        let paywall: PaywallData = try self.decodeFixture("PaywallData-Sample1")
+
+        let esConfig = try XCTUnwrap(paywall.config(for: Locale(identifier: "es")))
```
NachoSoto, line 78: @aboedo we'll need to make this work so it's consistent with how iOS works.

Sources/Paywalls/PaywallData.swift
```diff
@@ -99,8 +99,12 @@ extension PaywallData {
     }
 
     /// - Returns: ``PaywallData/LocalizedConfiguration-swift.struct`` for the given `Locale`, if found.
-    public func config(for locale: Locale) -> LocalizedConfiguration? {
-        return self.localization[locale.identifier]
+    /// - Note: this allows searching by `Locale` with only language code and missing region (like `en`, `es`, etc).
+    public func config(for requiredLocale: Locale) -> LocalizedConfiguration? {
+        self.localization[requiredLocale.identifier] ??
+        self.localization.first { locale, _ in
```
NachoSoto, line 105: @joshdholtz we'll need to implement the same logic in the backend when requesting a specific locale (or did we decide we weren't gonna do that?)

joshdholtz: @NachoSoto We decided against that for now so nothing to do there 😊

joshdholtz (approved):
🔥

PR #2848, opened 2023-07-20
`Paywalls`: disable animations during unit tests

NachoSoto:
This improves snapshot tests, and means we can reduce the polling time.


joshdholtz (approved):
This looks good! But can't animations also be disabled at a simulator level? 🤔

NachoSoto:
Hmm yeah, that’s usually for UIKit though but I can check if it would make sense to use that instead.

NachoSoto:
@joshdholtz thanks, replaced this with just one line:
```swift
UIView.setAnimationsEnabled(false)
```

PR #2851, opened 2023-07-20
`Paywalls`: basic localization support

NachoSoto:
Most of the changes are to support the creation of the new `PaywallViewLocalizationTests` by being able to override the environment `Locale`.

Simulator Screenshot - iPhone 14 Pro - 2023-07-20 at 18 10 20


RevenueCatUI/Views/AsyncButton.swift
```diff
@@ -47,7 +47,7 @@ struct AsyncButton<Label>: View where Label: View {
             Button {
                 self.error = nil
             } label: {
-                Text("OK")
+                Text("OK", bundle: .module)
```
joshdholtz, line 50: Does `.module` work with cocoapods? Asking this if we decided to try and call `RevenueCatUI` views from React Native, Flutter, etc 🤷‍♂️

RevenueCatUI/Data/Localizable.xcstrings
joshdholtz, line 1: 😍

joshdholtz (approved):

RevenueCatUI/Views/AsyncButton.swift
NachoSoto: Yes :) it’d be equivalent to `Bundle(forClass:)`

NachoSoto: But we'd have to manually add support for CocoaPods anyway, which we won't initially.

PR #2853, opened 2023-07-20
`Paywalls`: added test to ensure package selection maintains order

NachoSoto:
I thought I saw a bug because this wasn't true, and realized that it's important to cover in a test. The order of package should depend on the setting, not the offering order. This test ensures that.


lburdock (approved):
I approve this test!

joshdholtz (approved):

PR #2846, opened 2023-07-20
`Paywalls`: `TrialOrIntroEligibilityChecker.eligibility(for packages:)`

NachoSoto:
This will be used for the multi-package template.

Includes #2858.

PR #2858, opened 2023-07-20
New `Purchases.checkTrialOrIntroDiscountEligibility(packages:)`

NachoSoto:
This new overload helps the implementation of `RevenueCatUI` (#2846), so we can get a complete dictionary of all requested `Packages`, regardless of whether they might refer to the same products, or if the implementation does not contain that in the result.

I also implemented `IntroEligibility`'s `Hashable`'s overrides as well as `debugDescription`, which were needed to debug this.
Turns out there was no test coverage for the other check trial eligibility methods, so this tests those indirectly as well.


NachoSoto:
Changed this to target `paywalls` so we don't unnecessary ship a minor release just with this.

NachoSoto:
Actually this was included in #2846, so closing.

PR #2859, opened 2023-07-21
`Paywalls`: send presented `PaywallViewMode` with purchases

NachoSoto:
I've moved `PaywallViewMode` from `RevenueCatUI` to `RevenueCat`. This uses the same approach as #2645. This method is `public` so it can be used from `RevenueCatUI`, but the documentation stays its purpose.

The behavior is covered with integration tests and a snapshot test.


NachoSoto:
I need to rebase this and handle a conflict. Will do on Monday.

NachoSoto:
Just rebased this

joshdholtz (approved):
LGTM

PR #2862, opened 2023-07-23
`SimpleApp`: updates for TestFlight compatibility

NachoSoto:
### Changes:

- Added missing `Package.resolved`
- Add icon
- Updated deployment target to iOS 16.0
- Added compliance check


.gitignore
```diff
@@ -39,6 +39,7 @@ scan_derived_data/
 generated_docs/
 Examples/**/Package.resolved
 Tests/**/Package.resolved
+!Tests/TestingApps/SimpleApp/SimpleApp.xcworkspace/xcshareddata/swiftpm/Package.resolved
```
tonidero, line 42: Hmm this might be slightly dangerous since, if we ever move this app somewhere else, we might forget to update this. How about updating the line above to be multiple entries for any other occurences of `Package.resolved`?

tonidero (approved):
Looks great!

.gitignore
NachoSoto: Good point, good idea.

NachoSoto: Done.

PR #2864, opened 2023-07-24
`Codable`: debug log entire JSON when decoding fails

NachoSoto:
This is a simple example, but on a larger JSON it becomes easier for us to debug decoding the entire JSON rather than seeing only the part that failed:
```
[codable] DEBUG: ℹ️ Encountered error when decoding JSON: {"e": "e3"}
[codable] ERROR: 😿‼️ Couldn't decode data from json, it was corrupted: Context(codingPath: [CodingKeys(stringValue: "e", intValue: nil)], debugDescription: "Cannot initialize E from invalid String value e3", underlyingError: nil)
```

### DIFF
--- a/Sources/FoundationExtensions/Decoder+Extensions.swift
+++ b/Sources/FoundationExtensions/Decoder+Extensions.swift
@@ -63,7 +63,7 @@ extension JSONDecoder {
             return try self.decode(type, from: jsonData)
         } catch {
             if logErrors {
-                ErrorUtils.logDecodingError(error, type: type)
+                ErrorUtils.logDecodingError(error, type: type, data: jsonData)
             }
             throw error
         }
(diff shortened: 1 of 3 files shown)


NachoSoto:
I love these little QoL improvements. I was trying to figure out what was wrong in a test, this helped.

Sources/FoundationExtensions/Decoder+Extensions.swift
```diff
@@ -123,7 +123,11 @@ extension JSONEncoder {
 
 extension ErrorUtils {
 
-    static func logDecodingError(_ error: Error, type: Any.Type) {
+    static func logDecodingError(_ error: Error, type: Any.Type, data: Data? = nil) {
+        if let data = data {
+            Logger.debug(Strings.codable.invalid_data_when_decoding(data))
```
tonidero, line 128: Hmm so this will be a separate log. I wonder if it should be part of the error log below...

tonidero (approved):
Left a comment... But can see it working either way

Sources/FoundationExtensions/Decoder+Extensions.swift
NachoSoto: I did it this way for 2 reasons, let me know what you think:

- The implementation of this method uses like 5 different strings so I'd have to modify all of them
- The JSON might contain sensitive data, so we might not want the whole thing to be in the console. In release builds, `debug` logs aren't enabled by default, therefore eliminating this.

tonidero: Ah ok makes sense 👍

PR #2860, opened 2023-07-21
`Paywalls`: pre-fetch intro eligibility for paywalls

NachoSoto:
This will help provide a better experience when launching paywalls, as hopefully we will have already computed intro eligibility by then.
This is even better than when using `StoreKit` paywalls.

Note that thanks to `CachingTrialOrIntroPriceEligibilityChecker` (#2007) this is cached automatically, and also it handles multiple requests with the same or different product identifiers.

Depends on #2865.

PR #2865, opened 2023-07-24
`Perf`: update `CustomerInfo` cache before anything else

NachoSoto:
I noticed the order of cache updates on SDK initialization was:
- `Offerings`
- `ProductEntitlementMapping`
- `CustomerInfo`

This is inefficient because fetching `CustomerInfo` first is the most important. Also, #2860 will pre-warm intro eligibility, so we want to do that _after_ we might have potentially cleared the intro eligibility cache after a `CustomerInfo` update.
Example log from before:
```
[eligibility] DEBUG: ℹ️ Warming up intro eligibility cache for packages in paywall: [PackageType.$rc_monthly, PackageType.$rc_annual]
[customer] DEBUG: ℹ️ Sending latest CustomerInfo to delegate.
[customer] DEBUG: 😻 CustomerInfo updated from network.
[eligibility] DEBUG: ℹ️ Detected active subscriptions changed. Clearing trial or intro eligibility cache.
```

The new order is:
- `CustomerInfo`
- `ProductEntitlementMapping`
- `Offerings`

_Note: I recommend reviewing the diff ignoring whitespace._

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -750,9 +750,7 @@ extension Purchases {
         self.identityManager.switchUser(to: newAppUserID)
 
         self.systemInfo.isApplicationBackgrounded { isBackgrounded in
-            self.offeringsManager.updateOfferingsCache(appUserID: self.appUserID,
-                                                       isAppBackgrounded: isBackgrounded,
-                                                       completion: nil)
+            self.updateOfferingsCache(isAppBackgrounded: isBackgrounded)
         }
     }
 
@@ -1553,9 +1551,7 @@ private extension Purchases {
         }
 
         if self.deviceCache.isOfferingsCacheStale(isAppBackgrounded: isAppBackgrounded) {
-            self.offeringsManager.updateOfferingsCache(appUserID: self.appUserID,
-                                                       isAppBackgrounded: isAppBackgrounded,
-                                                       completion: nil)
+            self.updateOfferingsCache(isAppBackgrounded: isAppBackgrounded)
         }
     }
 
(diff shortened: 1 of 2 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1604,6 +1599,12 @@ private extension Purchases {
         self.delegate?.purchases?(self, receivedUpdated: info)
     }
 
+    private func updateOfferingsCache(isAppBackgrounded: Bool) {
```
NachoSoto, line 1602: _I also extracted this into a method because it's going to grow more in #2860._

Tests/UnitTests/Purchasing/Purchases/PurchasesGetOfferingsTests.swift
```diff
@@ -26,6 +26,14 @@ class PurchasesGetOfferingsTests: BasePurchasesTests {
         expect(self.mockOfferingsManager.invokedUpdateOfferingsCacheCount).toEventually(equal(1))
     }
 
+    func testFirstInitializationGetsOfferingsIfAppActiveInCustomEntitlementComputation() {
```
NachoSoto, line 29: ⚠️  It was important to add this test because a naive re-ordering of `updateAllCachesIfNeeded` would have meant that this would have now been skipped.

tonidero (commented):
Changes look good to me... With the jittering, I imagine these requests can still happen in different order, but I think that's ok. Leaving others to review first though.

Another question I have. To calculate intro-eligibility, don't we need offerings first? So we can know what products to use to calculate intro eligibility.

NachoSoto:
> With the jittering, I imagine these requests can still happen in different order, but I think that's ok. Leaving others to review first though.

We only do jitter if updating caches in the background. This method I'm modifying is called when launching the app **and it's in the foreground**, so the order will be deterministic.

> To calculate intro-eligibility, don't we need offerings first? So we can know what products to use to calculate intro eligibility.

Yeah that's why pre-warming that is done last, after fetching offerings. Without this change, that might have been done before updating the `CustomerInfo`. When that was done (at the end), it could have invalidated the eligibility cache we had just pre-warmed.

aboedo (approved):

PR #2869, opened 2023-07-24
`Paywalls`: new API for easily displaying `PaywallView` with just one line

NachoSoto:
Self-explanatory. This removes a lot of boilerplate, reducing the required code to just:
```swift
var body: some View {
   YourApp()
      .presentPaywallIfNecessary(requiredEntitlementIdentifier: "pro")
}
```


RevenueCatUI/View+PresentPaywall.swift
```diff
@@ -0,0 +1,86 @@
...
+    /// ```
+    /// - Note: If loading the `CustomerInfo` fails (for example, if Internet is offline),
+    /// the paywall won't be displayed.
+    public func presentPaywallIfNecessary(
+        mode: PaywallViewMode = .default,
+        requiredEntitlementIdentifier: String
+    ) -> some View {
+        return self.presentPaywallIfNecessary { info in
+            !info.entitlements
+                .activeInCurrentEnvironment
```
joshdholtz, line 30: Can we make this `.activeInCurrentEnvironment`  a parameter here? I know some people don't like using this. I think its good to have it as the default but I think I'd like this to be something the user can change 🤷‍♂️

joshdholtz (approved):
Love this! ❤️ Small question/request on `activeInCurrentEnvironment` but otherwise 👍

NachoSoto:
Feedback on the method name welcome.

RevenueCatUI/View+PresentPaywall.swift
NachoSoto: That's why I added the other overload (that takes a `CustomerInfo -> Bool` closure), so you can customize it.
I thought about the parameter to change that but I think it's more awkward. I documented that it has to be active in the _current_ environment in the docstring for this one.
```swift
.presentPaywallIfNecessary(entitlement: "", inCurrentEnvironment: true)
```

It's kinda confusing if you don't fully understand it IMO, and I think it's one of those cases where making the behavior depend on a `Bool` parameter is not a great interface.

What do you think?

PR #2871, opened 2023-07-24
`Paywalls`: added new `defaultPackage` configuration

NachoSoto:
It's optional, so it defaults to the first package if not set.

### DIFF
--- a/Sources/Paywalls/PaywallData.swift
+++ b/Sources/Paywalls/PaywallData.swift
@@ -11,6 +11,8 @@
 //
 //  Created by Nacho Soto on 7/10/23.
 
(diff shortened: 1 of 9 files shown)


Sources/Paywalls/PaywallData.swift
```diff
@@ -141,6 +143,9 @@ extension PaywallData {
         /// The list of package types this paywall will display
         public var packages: [PackageType]
 
+        /// The package to be selected by default.
+        public var defaultPackage: PackageType?
```
NachoSoto, line 147: cc @guido732 @lburdock new field.

aboedo (approved):

PR #2877, opened 2023-07-25
`Paywalls`: added `offerName` parameter

NachoSoto:
This is now used in `MultiPackageTemplate`, defaulting to the `productName` if none is set.

### DIFF
--- a/RevenueCatUI/Templates/MultiPackageTemplate.swift
+++ b/RevenueCatUI/Templates/MultiPackageTemplate.swift
@@ -142,7 +142,7 @@ private struct MultiPackageTemplateContent: View {
                         : .gray
                     )
 
-                Text(package.content.productName)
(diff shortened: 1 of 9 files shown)


RevenueCatUI/Templates/MultiPackageTemplate.swift
```diff
@@ -142,7 +142,7 @@ private struct MultiPackageTemplateContent: View {
                         : .gray
                     )
 
-                Text(package.content.productName)
+                Text(self.localization(for: package.content).offerName ?? package.content.productName)
```
NachoSoto, line 145: `offerName` isn't required (since not all templates will have it), so I'm making it default to the product name which is always available /cc @guido732 @lburdock

joshdholtz (approved):

PR #2879, opened 2023-07-25
`CustomEntitlementComputation`: use custom API key

tonidero:
### Description
We were using the default API key to test the custom entitlement computation package installation and integration tests. This changes that so we use a special API key with some special behavior in the backend to mimic real scenarios.


NachoSoto:
Thanks for doing this!

tonidero:
We do need to fix the project 😅 Will do so after the meetings today

NachoSoto:
Mainly I think it's the bundle ID in that app is different from the integration tests. Might need to create a separate app with the same integration test bundle ID that's configured in the same way.

PR #2889, opened 2023-07-26
`PostReceiptDataOperation`: add `InitiationSource` to cache key

NachoSoto:
The backend can produce "receipt not valid" errors depending on this value, so we don't want to share requests if they have different initiation sources.

### DIFF
--- a/Sources/Networking/Operations/PostReceiptDataOperation.swift
+++ b/Sources/Networking/Operations/PostReceiptDataOperation.swift
@@ -64,13 +64,14 @@ final class PostReceiptDataOperation: CacheableNetworkOperation {
         /// - `isRestore`
         /// - Receipt (`hashString` instead of `fetchToken` to avoid big receipts leading to a huge cache key)
         /// - `ProductRequestData.cacheKey`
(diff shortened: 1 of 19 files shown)


aboedo:
looks like we need to update tests?

NachoSoto:
Yup, `testCachesRequestsForSameReceipt` had 2 different initiation sources so it was now failing.

NachoSoto:
Fixed.

aboedo:
🤔 this would also in turn meant that we'd fire 2 requests if we have one that's queue and one that's purchase, right? I wonder if we should instead have "priorities", i.e.: if you have one queue and one purchase, the purchase one goes through

NachoSoto:
> 🤔 this would also in turn meant that we'd fire 2 requests if we have one that's queue and one that's purchase, right? I wonder if we should instead have "priorities", i.e.: if you have one queue and one purchase, the purchase one goes through

We can't really do that, they get inserted in whatever order.
But now that I think about this, and with the context of the optimizations we've been doing yesterday (I did this 2 days ago), this is probably counter productive, so closing.

PR #2890, opened 2023-07-26
`PurchasesOrchestrator`: disambiguate transactions from the queue

NachoSoto:
This adds tests and a fix for a scenario discovered by @aboedo and @mcastany.

### Timeline:
- Make a purchase of product_1
- Close app
- Renewals happen in the background
- Subscription expires
- Open app, renewals are in the queue but not processed yet
- Start a purchase of `product_1` (`SKPaymentQueue.add`)
- StoreKit queue shows renewal from step 3 as finished, it's expired
- SDK sees finished transaction in the queue for `product_1`
- SDK sees completion block for `product_1`, considers it a successful `.purchase`
- SDK posts receipt, `initiation_source: .purchase` 💥
- Backend replies with 422 because StoreKit hasn't processed the transaction from 6

### Fix:
The fix looks at the date when the completion block was added (the time `.purchase()` is called).
If the transaction was created _before_ this, we know it didn't come from the actual purchase, and therefore it should be identified as `.queue`.

### DIFF
--- a/Sources/LocalReceiptParsing/Helpers/ReceiptStrings.swift
+++ b/Sources/LocalReceiptParsing/Helpers/ReceiptStrings.swift
@@ -28,7 +28,7 @@ enum ReceiptStrings {
     case parsing_receipt
     case refreshing_empty_receipt
     case unable_to_load_receipt(Error)
-    case posting_receipt(AppleReceipt)
+    case posting_receipt(AppleReceipt, initiationSource: String)
     case receipt_subscription_purchase_equals_expiration(
         productIdentifier: String,
         purchase: Date,
@@ -87,8 +87,8 @@ extension ReceiptStrings: LogMessage {
             return "Unable to load receipt, ensure you are logged in to a valid Apple account.\n" +
             "Error: \(error)"
 
-        case let .posting_receipt(receipt):
-            return "Posting receipt (note: the contents might not be up-to-date, " +
+        case let .posting_receipt(receipt, initiationSource):
+            return "Posting receipt (source: '\(initiationSource)') (note: the contents might not be up-to-date, " +
             "but it will be refreshed with Apple's servers):\n\(receipt.debugDescription)"
 
         case let .receipt_subscription_purchase_equals_expiration(
--- a/Sources/Logging/Strings/PurchaseStrings.swift
+++ b/Sources/Logging/Strings/PurchaseStrings.swift
@@ -40,6 +40,9 @@ enum PurchaseStrings {
                                                              SKPaymentTransaction)
     case paymentqueue_updated_transaction(SKPaymentTransactionObserver,
                                           SKPaymentTransaction)
+    case paymentqueue_ignoring_callback_for_older_transaction(PurchasesOrchestrator,
+                                                              StoreTransactionType,
+                                                              Date)
     case presenting_code_redemption_sheet
     case unable_to_present_redemption_sheet
     case purchases_synced
@@ -167,6 +170,11 @@ extension PurchaseStrings: LogMessage {
             ]
                 .compactMap { $0 }
                 .joined(separator: " ")
+
+        case let .paymentqueue_ignoring_callback_for_older_transaction(observer, transaction, date):
+            return "\(Strings.objectDescription(observer)): will not notify callback for transaction " +
+            "'\(transaction.transactionIdentifier)'. " +
+            "Transaction date '\(transaction.purchaseDate)' - callback date '\(date)'"
 
         case .presenting_code_redemption_sheet:
             return "Presenting code redemption sheet."
--- a/Sources/Logging/Strings/PurchaseStrings.swift
+++ b/Sources/Logging/Strings/PurchaseStrings.swift
@@ -51,7 +51,7 @@ enum PurchaseStrings {
     case product_purchase_failed(productIdentifier: String, error: Error)
     case skpayment_missing_from_skpaymenttransaction
     case skpayment_missing_product_identifier
-    case sktransaction_missing_transaction_date
+    case sktransaction_missing_transaction_date(SKPaymentTransactionState)
     case sktransaction_missing_transaction_identifier
     case could_not_purchase_product_id_not_found
     case payment_identifier_nil
@@ -208,11 +208,12 @@ extension PurchaseStrings: LogMessage {
             return "There is a problem with the SKPayment missing " +
             "a product identifier - this is an issue with the App Store."
 
-        case .sktransaction_missing_transaction_date:
+        case let .sktransaction_missing_transaction_date(transactionState):
             return """
             The SKPaymentTransaction has a nil value for transaction date - this is a bug in StoreKit.
             Unix Epoch will be used instead for the transaction within the app.
             Transactions in the backend and in webhooks are unaffected and will have the correct timestamps.
+            Transaction state: \(transactionState)
             """
 
         case .sktransaction_missing_transaction_identifier:
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -956,9 +956,21 @@ private extension PurchasesOrchestrator {
     func getAndRemovePurchaseCompletedCallback(
         forTransaction transaction: StoreTransaction
     ) -> PurchaseCompletedBlock? {
-        return self.purchaseCompleteCallbacksByProductID.modify {
-            $0.removeValue(forKey: transaction.productIdentifier)
-        }?.completion
+        return self.purchaseCompleteCallbacksByProductID.modify { callbacks -> PurchaseCompletedBlock? in
+            guard let value = callbacks[transaction.productIdentifier] else { return nil }
+
+            if value.date <= transaction.purchaseDate {
+                callbacks.removeValue(forKey: transaction.productIdentifier)
+                return value.completion
+            } else {
+                Logger.verbose(Strings.purchase.paymentqueue_ignoring_callback_for_older_transaction(
+                    self,
+                    transaction,
+                    value.date
+                ))
+                return nil
+            }
+        }
     }
 
     func markSyncedIfNeeded(
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -970,7 +970,7 @@ private extension PurchasesOrchestrator {
         return self.purchaseCompleteCallbacksByProductID.modify { callbacks -> PurchaseCompletedBlock? in
             guard let value = callbacks[transaction.productIdentifier] else { return nil }
 
-            if value.creationDate <= transaction.purchaseDate {
+            if !transaction.hasKnownPurchaseDate || value.creationDate <= transaction.purchaseDate {
                 callbacks.removeValue(forKey: transaction.productIdentifier)
                 return value.completion
             } else {
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -43,9 +43,7 @@ final class PurchasesOrchestrator {
 
     private let _allowSharingAppStoreAccount: Atomic<Bool?> = nil
     private let presentedOfferingIDsByProductID: Atomic<[String: String]> = .init([:])
-    private let purchaseCompleteCallbacksByProductID: Atomic<[
-        String: (completion: PurchaseCompletedBlock, date: Date)
-    ]> = .init([:])
+    private let purchaseCompleteCallbacksByProductID: Atomic<[String: CallbackData]> = .init([:])
 
     private var appUserID: String { self.currentUserProvider.currentAppUserID }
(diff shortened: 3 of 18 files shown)


Sources/LocalReceiptParsing/Helpers/ReceiptStrings.swift
```diff
@@ -87,8 +87,8 @@ extension ReceiptStrings: LogMessage {
             return "Unable to load receipt, ensure you are logged in to a valid Apple account.\n" +
             "Error: \(error)"
 
-        case let .posting_receipt(receipt):
-            return "Posting receipt (note: the contents might not be up-to-date, " +
+        case let .posting_receipt(receipt, initiationSource):
+            return "Posting receipt (source: '\(initiationSource)') (note: the contents might not be up-to-date, " +
```
NachoSoto, line 91: Logging this for awareness next time we have to debug these logs.

Sources/Logging/Strings/PurchaseStrings.swift
```diff
@@ -208,11 +208,12 @@ extension PurchaseStrings: LogMessage {
             return "There is a problem with the SKPayment missing " +
             "a product identifier - this is an issue with the App Store."
 
-        case .sktransaction_missing_transaction_date:
+        case let .sktransaction_missing_transaction_date(transactionState):
             return """
             The SKPaymentTransaction has a nil value for transaction date - this is a bug in StoreKit.
             Unix Epoch will be used instead for the transaction within the app.
             Transactions in the backend and in webhooks are unaffected and will have the correct timestamps.
+            Transaction state: \(transactionState)
```
NachoSoto, line 216: I added this for awareness, but we'll eventually need to deal with this and remove this log.
[The docs](https://developer.apple.com/documentation/storekit/skpaymenttransaction/1411273-transactiondate) make this clear:
> The contents of this property are undefined except when `transactionState` is set to `SKPaymentTransactionState.purchased` or `SKPaymentTransactionState.restored`.

Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
```diff
@@ -204,7 +204,7 @@ extension BaseStoreKitIntegrationTests {
         }
 
         if !entitlement.isActive {
-            try await failTest("Entitlement is not active")
+            try await failTest("Entitlement is not active: \(entitlement)")
```
NachoSoto, line 207: This helps debug these failures.

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -421,6 +421,39 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
...
+
+        // 4. Expire subscription
+        try await self.expireSubscription(entitlement)
+
+        // 5. Re-open app
+        await self.resetSingleton()
+        self.logger.clearMessages()
+
+        // 6. Purchase again
+        _ = try await self.purchases.purchase(package: self.monthlyPackage)
```
NachoSoto, line 445: ❗  One thing I don't understand even after looking at the receipt posted: this call to purchase returns an inactive entitlement ❓ 
We should look into it either in this PR or in a separate test.

MarkVillacampa (approved):

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
NachoSoto: I'll wait to talk to @aboedo about this cause I'm really confused.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -43,7 +43,9 @@ final class PurchasesOrchestrator {
 
     private let _allowSharingAppStoreAccount: Atomic<Bool?> = nil
     private let presentedOfferingIDsByProductID: Atomic<[String: String]> = .init([:])
-    private let purchaseCompleteCallbacksByProductID: Atomic<[String: PurchaseCompletedBlock]> = .init([:])
+    private let purchaseCompleteCallbacksByProductID: Atomic<[
+        String: (completion: PurchaseCompletedBlock, date: Date)
```
aboedo, line 47: maybe we should have this typed into a struct rather than a tuple? that'd be a little more expressive + if we ever need to add anything more it's easier to add

aboedo: also maybe we can name this `callbackCreationDate` instead of `date` for clarity, otherwise you might think this is the date of the corresponding transaction or something

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -823,19 +830,27 @@ private extension PurchasesOrchestrator {
...
+                    return .queue
+                }
 
-            switch (hasPurchaseCallback, restored) {
-            case (true, false): return .purchase
                 // Note that restores initiated through the SDK with `restorePurchases`
                 // won't use this method since those set the initiation source explicitly.
-            case (true, true): return .restore
-            case (false, _): return .queue
+            case (.some, true): return .restore
```
aboedo, line 852: kinda feels like we're losing a little bit of readability in the transition from `true` -> `some`

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -933,7 +948,7 @@ private extension PurchasesOrchestrator {
                 return false
             }
 
-            callbacks[productIdentifier] = completion
+            callbacks[productIdentifier] = (completion, Date())
```
aboedo, line 951: wonder if we'll want to end up passing a date provider for this for tests

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
aboedo: yeah, that's unexpected - that's what this whole PR is meant to address. 
well, actually... 
I mean, purchase is going to return right after seeing one of the `queue` values return. like, this case is still true: 

image

The problem isn't just that the `initiation_source` was marked incorrectly, it's that we're returning that value for the purchase, when in reality we should ignore transactions that don't have that date entirely. 

I'll comment in the corresponding line

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -823,19 +830,27 @@ private extension PurchasesOrchestrator {
...
             // Having a purchase completed callback implies that the transation comes from an explicit call
             // to `purchase()` instead of a StoreKit transaction notification.
-            let hasPurchaseCallback = self.purchaseCompleteCallbacksByProductID.value.keys.contains(productIdentifier)
+            // As long as the transaction date is _after_ the completion block was added.
+            let completionBlockDate = self.purchaseCompleteCallbacksByProductID.value[productIdentifier]?.date
+
+            switch (completionBlockDate, restored) {
+            case let (.some(completionBlockDate), false):
+                if transaction.purchaseDate > completionBlockDate {
+                    return .purchase
```
aboedo, line 845: we shouldn't just change the `initiation_source`, we should entirely not use this callback if the transaction is older than when you started the purchase. 

It's somewhat delicate in that we still need to handle transactions with no dates (like for user cancels), but if we can definitely say "this transaction isn't the one I started the purchase for" then the best path would be to just not call completion for it, right?

aboedo (commented):
Left a few comments. Looking great! But we need to think about the case where we're calling completion too early. Maybe we can pair up on this tomorrow?

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: Both good suggestions 👍🏻  will do.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: Oooooh of course. Let's talk through together this tomorrow if you can 🙏🏻

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: That's the only way to both check for it being `nil` and extract the value from it though. Any suggestion?

Tests/UnitTests/Mocks/MockTransaction.swift
```diff
@@ -24,7 +24,12 @@ class MockTransaction: SKPaymentTransaction {
 
     var mockTransactionDate: Date? = Date()
     override var transactionDate: Date? {
-        mockTransactionDate
+        // This matches the behavior of `SKPaymentTransaction`.
+        guard self.transactionState == .purchased || self.transactionState == .restored else {
+            return nil
+        }
```
NachoSoto, line 30: I'm glad I did this because this exposed how this is breaking some scenarios.

NachoSoto:
@aboedo all tests passing 🎉

Sources/LocalReceiptParsing/Helpers/ReceiptStrings.swift
```diff
@@ -28,7 +28,7 @@ enum ReceiptStrings {
     case parsing_receipt
     case refreshing_empty_receipt
     case unable_to_load_receipt(Error)
-    case posting_receipt(AppleReceipt)
+    case posting_receipt(AppleReceipt, initiationSource: String)
```
joshdholtz, line 31: Might be a nit but any reason why this is `String` instead of `InitiationSource`? 🤷‍♂️

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -933,16 +959,28 @@ private extension PurchasesOrchestrator {
...
+            if !transaction.hasKnownPurchaseDate || value.creationDate <= transaction.purchaseDate {
+                callbacks.removeValue(forKey: transaction.productIdentifier)
+                return value.completion
+            } else {
+                Logger.verbose(Strings.purchase.paymentqueue_ignoring_callback_for_older_transaction(
+                    self,
+                    transaction,
+                    value.creationDate
+                ))
+                return nil
```
joshdholtz, line 982: Why does this not need to remove the callback? When would this callback get removed? 🤔 

Just want to make sure its not there for forever 🤷‍♂️

joshdholtz (approved):
A few small questions/nits but looks good to ship 👍

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -435,6 +435,36 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
...
+        self.logger.clearMessages()
+
+        // 6. Purchase again
+        try await self.purchaseMonthlyProduct()
+
+        self.logger.clearMessages()
+
+        // 7. Verify no transaction is posted as a purchase.
+        self.logger.verifyMessageWasNotLogged("Posting receipt (source: 'purchase')",
+                                              allowNoMessages: true)
```
aboedo, line 465: shouldn't we eventually post the transaction from 6 as a purchase?

Sources/LocalReceiptParsing/Helpers/ReceiptStrings.swift
NachoSoto: Good suggestion, changed.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: Done.

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
NachoSoto: LOL yeah this is a bad refactor from me changing the test `clearMessages` shouldn't be there.

Sources/LocalReceiptParsing/Helpers/ReceiptStrings.swift
NachoSoto: Oh I know why: this is used by `ReceiptParser` too, which doesn't have that type

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -933,16 +959,30 @@ private extension PurchasesOrchestrator {
...
 
     func getAndRemovePurchaseCompletedCallback(
         forTransaction transaction: StoreTransaction
     ) -> PurchaseCompletedBlock? {
-        return self.purchaseCompleteCallbacksByProductID.modify {
-            $0.removeValue(forKey: transaction.productIdentifier)
+        return self.purchaseCompleteCallbacksByProductID.modify { callbacks -> PurchaseCompletedBlock? in
+            guard let value = callbacks[transaction.productIdentifier] else { return nil }
+
+            if !transaction.hasKnownPurchaseDate || value.creationDate <= transaction.purchaseDate {
```
tonidero, line 973: I know I'm late here but maybe we could extract this condition to a function `transactionDateMatchCallbackDate`, `transactionMatchesCallback` or something like that? Just so it's easier to read

PR #2896, opened 2023-07-26
`Logging`: avoid logging "updating offerings" when request is cached

NachoSoto:
I noticed this in our logs:
```
[offering] DEBUG: ℹ️ No cached Offerings, fetching from network
[purchases] VERBOSE: Updating all caches
[offering] DEBUG: ℹ️ Offerings cache is stale, updating from network in foreground
[offering] DEBUG: ℹ️ Offerings cache is stale, updating from network in foreground
[network] DEBUG: ℹ️ Network operation 'GetOfferingsOperation' found with the same cache key 'GetOfferingsOpe...'. Skipping request.
[network] DEBUG: ℹ️ GetOfferingsOperation: Started
```

The double "updating from network" can be confusing, so this fixes it so it's only logged if the request will actually be started. Also added/updated our tests to verify the correct logs as well as caching behavior.

### DIFF
--- a/Sources/Networking/OfferingsAPI.swift
+++ b/Sources/Networking/OfferingsAPI.swift
@@ -28,7 +28,7 @@ class OfferingsAPI {
     }
 
     func getOfferings(appUserID: String,
-                      withRandomDelay randomDelay: Bool,
+                      isAppBackgrounded: Bool,
                       completion: @escaping OfferingsResponseHandler) {
         let config = NetworkOperation.UserSpecificConfiguration(httpClient: self.backendConfig.httpClient,
                                                                 appUserID: appUserID)
@@ -40,9 +40,15 @@ class OfferingsAPI {
         let offeringsCallback = OfferingsCallback(cacheKey: factory.cacheKey, completion: completion)
         let cacheStatus = self.offeringsCallbacksCache.add(offeringsCallback)
 
+        if cacheStatus == .firstCallbackAddedToList {
+            Logger.debug(isAppBackgrounded
+                         ? Strings.offering.offerings_stale_updating_in_background
+                         : Strings.offering.offerings_stale_updating_in_foreground)
+        }
+
         self.backendConfig.addCacheableOperation(
             with: factory,
-            withRandomDelay: randomDelay,
+            withRandomDelay: isAppBackgrounded,
             cacheStatus: cacheStatus
         )
     }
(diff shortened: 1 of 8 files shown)


Sources/Networking/OfferingsAPI.swift
```diff
@@ -40,9 +40,15 @@ class OfferingsAPI {
...
+        if cacheStatus == .firstCallbackAddedToList {
+            Logger.debug(isAppBackgrounded
+                         ? Strings.offering.offerings_stale_updating_in_background
+                         : Strings.offering.offerings_stale_updating_in_foreground)
+        }
+
         self.backendConfig.addCacheableOperation(
             with: factory,
-            withRandomDelay: randomDelay,
+            withRandomDelay: isAppBackgrounded,
```
NachoSoto, line 51: This decision makes more sense in this "API" type than in `OfferingsManager`.

Tests/BackendIntegrationTests/OtherIntegrationTests.swift
```diff
@@ -59,6 +66,15 @@ class OtherIntegrationTests: BaseBackendIntegrationTests {
         )
     }
 
+    func testOfferingsAreOnlyFetchedOnceOnSDKInitialization() async throws {
```
NachoSoto, line 69: New test.

Tests/BackendIntegrationTests/OtherIntegrationTests.swift
```diff
@@ -59,6 +66,15 @@ class OtherIntegrationTests: BaseBackendIntegrationTests {
...
+        self.logger.verifyMessageWasLogged(Strings.offering.offerings_stale_updating_in_foreground,
+                                           level: .debug,
+                                           expectedCount: 1)
```
NachoSoto, line 72: This was the failing test that this fixes.

Tests/BackendIntegrationTests/OtherIntegrationTests.swift
```diff
@@ -59,6 +66,15 @@ class OtherIntegrationTests: BaseBackendIntegrationTests {
...
+        self.logger.verifyMessageWasLogged("GetOfferingsOperation: Started",
+                                           level: .debug,
+                                           expectedCount: 1)
```
NachoSoto, line 75: But this also doesn't hurt to verify.

aboedo (approved):

PR #2899, opened 2023-07-26
`Paywalls`: added "features" to `LocalizedConfiguration`

NachoSoto:
To be used by upcoming templates.

### DIFF
--- a/Sources/Paywalls/PaywallData.swift
+++ b/Sources/Paywalls/PaywallData.swift
@@ -74,6 +74,8 @@ extension PaywallData {
         var _callToActionWithIntroOffer: String?
         @NonEmptyStringDecodable
         var _offerDetailsWithIntroOffer: String?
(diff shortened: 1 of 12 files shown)


Sources/Paywalls/PaywallData.swift
```diff
@@ -140,6 +148,32 @@ extension PaywallData {
 
 }
 
+extension PaywallData.LocalizedConfiguration {
+
+    /// An item to be showcased in a paywall.
+    public struct Feature {
```
aboedo, line 154: gotta love the ability to namespace things like this

aboedo (approved):

Sources/Paywalls/PaywallData.swift
NachoSoto: 👍🏻  to avoid with a nested mess hehe

PR #2900, opened 2023-07-26
`Paywalls`: changed `subtitle` to be optional

NachoSoto:
Not all templates require it, so it's optional from here on.


lburdock:
For frontend validation, should we still require `subtitle` on templates where it's in the mockup (like the first two templates)? Or do you think we should make subtitle optional for all templates?

NachoSoto:
@lburdock we can't encode that knowledge in the data type because it's shared. Because of that I'm making it optional in the template rendering. We could display an error, but easier to validate that in the frontend instead.

guido732 (approved):

aboedo (approved):

PR #2901, opened 2023-07-26
`Paywalls`: extracted intro eligibility out of templates

NachoSoto:
This simplifies template views further. Before they needed to get the `TrialOrIntroEligibilityChecker` environment and call the appropriate method.

Now they just need to inject the new `IntroEligibilityViewModel`, and extract either `.allEligibility` or `singleEligibility` based on the type of template. Everything else is done automatically outside of templates.

### DIFF
--- a/RevenueCatUI/Templates/MultiPackageBoldTemplate.swift
+++ b/RevenueCatUI/Templates/MultiPackageBoldTemplate.swift
@@ -5,12 +5,8 @@ import SwiftUI
 struct MultiPackageBoldTemplate: TemplateViewType {
 
     private let configuration: TemplateViewConfiguration
-
     @EnvironmentObject
-    private var introEligibilityChecker: TrialOrIntroEligibilityChecker
-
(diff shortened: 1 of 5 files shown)


RevenueCatUI/Templates/MultiPackageBoldTemplate.swift
```diff
@@ -5,25 +5,16 @@ import SwiftUI
 struct MultiPackageBoldTemplate: TemplateViewType {
 
     private let configuration: TemplateViewConfiguration
-
     @EnvironmentObject
-    private var introEligibilityChecker: TrialOrIntroEligibilityChecker
-
-    @State
-    private var introEligibility: [Package: IntroEligibilityStatus] = [:]
+    private var introEligibility: IntroEligibilityViewModel
```
NachoSoto, line 9: This is all that templates need to get now.

aboedo (approved):

PR #2904, opened 2023-07-26
`Logging`: avoid logging "updating offerings" when request is cached

NachoSoto:
I noticed this in our logs:
```
[offering] DEBUG: ℹ️ No cached Offerings, fetching from network
[purchases] VERBOSE: Updating all caches
[offering] DEBUG: ℹ️ Offerings cache is stale, updating from network in foreground
[offering] DEBUG: ℹ️ Offerings cache is stale, updating from network in foreground
[network] DEBUG: ℹ️ Network operation 'GetOfferingsOperation' found with the same cache key 'GetOfferingsOpe...'. Skipping request.
[network] DEBUG: ℹ️ GetOfferingsOperation: Started
```

The double "updating from network" can be confusing, so this fixes it so it's only logged if the request will actually be started. Also added/updated our tests to verify the correct logs as well as caching behavior.

### DIFF
--- a/Sources/Networking/OfferingsAPI.swift
+++ b/Sources/Networking/OfferingsAPI.swift
@@ -28,7 +28,7 @@ class OfferingsAPI {
     }
 
     func getOfferings(appUserID: String,
-                      withRandomDelay randomDelay: Bool,
+                      isAppBackgrounded: Bool,
                       completion: @escaping OfferingsResponseHandler) {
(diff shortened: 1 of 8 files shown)


NachoSoto:
@aboedo sorry I merged #2896 too early. Do you mind approving this again? 😇

aboedo (approved):

PR #2905, opened 2023-07-27
`Integration Tests`: added more observer mode tests

NachoSoto:
This extracts the purchasing code into a new `ExternalPurchasesManager` and `ObserverModeManager`, which can be used to perform purchases without using the SDK.

I've also bumped the deployment target of integration tests to 15.0 to simplify the code.

PR #2909, opened 2023-07-27
`StoreKit2TransactionListener`: converted into an `actor`

NachoSoto:
This is a preparatory refactor for an upcoming fix.
This improves the thread-safety of this class, and also refactors its mock to use a `protocol` instead.


aboedo (approved):

NachoSoto:
Note that this depends on #2905 so can't merge it yet.

PR #2911, opened 2023-07-27
Created `PostedTransactionCache`

NachoSoto:
This will be used to skip posting the same transaction multiple times, as a way to work around StoreKit reporting SK1 notifications from SK2 APIs.

### DIFF
--- a/Sources/Caching/DeviceCache.swift
+++ b/Sources/Caching/DeviceCache.swift
@@ -46,6 +46,26 @@ class DeviceCache {
         Logger.verbose(Strings.purchase.device_cache_deinit(self))
     }
 
+    // MARK: - generic methods
+
+    func update<Key: DeviceCacheKeyType, Value: Codable>(
+        key: Key,
+        default defaultValue: Value,
+        updater: @Sendable (inout Value) -> Void
+    ) {
+        self.userDefaults.write {
+            var value: Value = $0.value(forKey: key) ?? defaultValue
+            updater(&value)
+            $0.set(value, forKey: key)
+        }
+    }
+
+    func value<Key: DeviceCacheKeyType, Value: Codable>(for key: Key) -> Value? {
+        self.userDefaults.read {
+            $0.value(forKey: key)
+        }
+    }
+
     // MARK: - appUserID
 
     func cache(appUserID: String) {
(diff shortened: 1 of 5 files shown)


Sources/Caching/PostedTransactionCache.swift
```diff
@@ -0,0 +1,58 @@
...
+
+    func savePostedTransaction(_ transaction: StoreTransactionType)
+    func hasPostedTransaction(_ transaction: StoreTransactionType) -> Bool
+
+}
+
+// TODO: tests
+// TODO: integrate
+
+final class PostedTransactionCache: PostedTransactionCacheType {
```
NachoSoto, line 26: This uses `DeviceCache` underneath (composition, not inheritance). It also doesn't make `DeviceCache` have any more domain-specific caching code.

This is how I want to refactor `DeviceCache` so it's split in smaller classes.

Tests/UnitTests/Caching/PostedTransactionCacheTests.swift
```diff
@@ -0,0 +1,57 @@
...
+    private var deviceCache: DeviceCache!
+    private var cache: PostedTransactionCache!
+
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+
+        self.userDefaults = try XCTUnwrap(.init(suiteName: UUID().uuidString))
+        self.deviceCache = .init(sandboxEnvironmentDetector: MockSandboxEnvironmentDetector(isSandbox: true),
+                                 userDefaults: self.userDefaults)
+        self.cache = .init(deviceCache: self.deviceCache)
```
NachoSoto, line 32: I decided to test with with real `DeviceCache` and `UserDefaults` which is a lot more accurate and less painful.

Sources/Caching/PostedTransactionCache.swift
```diff
@@ -0,0 +1,55 @@
...
+    private typealias StoredTransactions = Set<String>
+
+    private let deviceCache: DeviceCache
+
+    init(deviceCache: DeviceCache) {
+        self.deviceCache = deviceCache
+    }
+
+    func savePostedTransaction(_ transaction: StoreTransactionType) {
+        self.deviceCache.update(key: CacheKey.transactions,
```
NachoSoto, line 34: TODO: do a quick test on device with N x 1000 transactions.

Sources/Caching/DeviceCache.swift
```diff
@@ -552,47 +569,56 @@ private extension DeviceCache {
         _ userDefaults: UserDefaults,
         productEntitlementMapping mapping: ProductEntitlementMapping
     ) {
-        guard let data = try? JSONEncoder.default.encode(value: mapping, logErrors: true) else {
-            return
+        if userDefaults.set(codable: mapping,
```
joshdholtz, line 572: Is there any sort of error that we should be logging if this fails?

joshdholtz (approved):
Small question but not show stopping. Looks good 👍

PR #2912, opened 2023-07-27
`Offering`: improved description

NachoSoto:


Sources/Purchasing/Offering.swift
```diff
@@ -92,11 +92,18 @@ import Foundation
 
     public override var description: String {
         return """
-        <Offering {\n\tidentifier=\(identifier)\n\tserverDescription=\(serverDescription)\n"
-        \tavailablePackages=\(valueOrEmpty(availablePackages))\n\tlifetime=\(valueOrEmpty(lifetime))\n
```
NachoSoto, line 96: This is using `"""`, so no need for manual `\n`/`\t`, etc.

Sources/Purchasing/Offering.swift
```diff
@@ -92,11 +92,18 @@ import Foundation
...
-        \tannual=\(valueOrEmpty(annual))\n\tsixMonth=\(valueOrEmpty(sixMonth))\n
-        \tthreeMonth=\(valueOrEmpty(threeMonth))\n\ttwoMonth=\(valueOrEmpty(twoMonth))\n
-        \tmonthly=\(valueOrEmpty(monthly))\n\tweekly=\(valueOrEmpty(weekly))\n}>
+        <Offering {
+            identifier=\(self.identifier)\n\tserverDescription=\(self.serverDescription)"
```
tonidero, line 96: Could we also separate the `serverDescription` to a new line?

NachoSoto: Oh thanks I missed that

PR #2914, opened 2023-07-28
`TransactionPoster`: avoid posting transactions multiple times

NachoSoto:
This is a performance optimization that will help avoid duplicate posts after #2612.

### TODO:
- [x] Tests for finishing (if posted) / not finishing
- [x] Observer mode does not finish if already posted
- [x] Test for get customer info + already posted avoids infinite recursion
- [x] Integration tests

### DIFF
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -374,21 +374,6 @@ private extension CustomerInfoManager {
         }
     }
 
-    private func cachedOrRequestCustomerInfoFetcher(
-        appUserID: String,
-        isAppBackgrounded: Bool
-    ) -> TransactionPosterResult.CustomerInfoFetcher {
-        return { completion in
-            if let cachedCustomerInfo = self.cachedCustomerInfo(appUserID: appUserID) {
-                completion(.success(cachedCustomerInfo))
-            } else {
-                self.requestCustomerInfo(appUserID: appUserID,
-                                         isAppBackgrounded: isAppBackgrounded,
-                                         completion: completion)
-            }
-        }
-    }
-
     private func requestCustomerInfo(appUserID: String,
                                      isAppBackgrounded: Bool,
                                      completion: @escaping CustomerAPI.CustomerInfoResponseHandler) {
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -320,10 +320,14 @@ private extension CustomerInfoManager {
         if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
             _ = Task<Void, Never> {
                 let transactions = await self.transactionFetcher.unfinishedVerifiedTransactions
+                let unpostedTransactions = self.transactionPoster.unpostedTransactions(in: transactions)
 
-                if let transactionToPost = transactions.first {
+                if let transactionToPost = unpostedTransactions.first {
                     Logger.debug(
-                        Strings.customerInfo.posting_transactions_in_lieu_of_fetching_customerinfo(transactions)
+                        Strings.customerInfo.posting_transactions_in_lieu_of_fetching_customerinfo(
+                            allTransactions: transactions,
+                            unpostedTransactions: unpostedTransactions
+                        )
                     )
 
                     let transactionData = PurchasedTransactionData(
@@ -336,7 +340,7 @@ private extension CustomerInfoManager {
 
                     // Post everything but the first transaction in the background
                     // in parallel so they can be de-duped
-                    let otherTransactionsToPostInParalel = Array(transactions.dropFirst())
+                    let otherTransactionsToPostInParalel = Array(unpostedTransactions.dropFirst())
                     Task.detached(priority: .background) {
                         await self.postTransactions(otherTransactionsToPostInParalel, transactionData)
                     }
@@ -344,10 +348,19 @@ private extension CustomerInfoManager {
                     // Return the result of posting the first transaction.
                     // The posted receipt will include the content of every other transaction
                     // so we don't need to wait for those.
-                    completion(await self.transactionPoster.handlePurchasedTransaction(
+                    let result = await self.transactionPoster.handlePurchasedTransaction(
                         transactionToPost,
                         data: transactionData
-                    ))
+                    )
+
+                    result.toResult(completion: completion) {
+                        // There is a chance due to race-conditions that even though it was initially determined that
+                        // `transactionToPost` had not been posted,
+                        // `handlePurchasedTransaction` returns `.alreadyPosted`. This falls back to
+                        self.requestCustomerInfo(appUserID: appUserID,
+                                                 isAppBackgrounded: isAppBackgrounded,
+                                                 completion: $0)
+                    }
                 } else {
                     self.requestCustomerInfo(appUserID: appUserID,
                                              isAppBackgrounded: isAppBackgrounded,
@@ -358,6 +371,21 @@ private extension CustomerInfoManager {
             return self.requestCustomerInfo(appUserID: appUserID,
                                             isAppBackgrounded: isAppBackgrounded,
                                             completion: completion)
+        }
+    }
+
+    private func cachedOrRequestCustomerInfoFetcher(
+        appUserID: String,
+        isAppBackgrounded: Bool
+    ) -> TransactionPosterResult.CustomerInfoFetcher {
+        return { completion in
+            if let cachedCustomerInfo = self.cachedCustomerInfo(appUserID: appUserID) {
+                completion(.success(cachedCustomerInfo))
+            } else {
+                self.requestCustomerInfo(appUserID: appUserID,
+                                         isAppBackgrounded: isAppBackgrounded,
+                                         completion: completion)
+            }
         }
     }
 
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -1116,40 +1116,42 @@ private extension PurchasesOrchestrator {
         let unsyncedAttributes = self.unsyncedAttributes
         let adServicesToken = self.attribution.unsyncedAdServicesToken
 
-        self.transactionPoster.handlePurchasedTransaction(
-            purchasedTransaction,
-            data: .init(
-                appUserID: self.appUserID,
-                presentedOfferingID: offeringID,
-                unsyncedAttributes: unsyncedAttributes,
-                aadAttributionToken: adServicesToken,
-                storefront: storefront,
-                source: self.purchaseSource(for: purchasedTransaction.productIdentifier,
-                                            transaction: purchasedTransaction,
-                                            restored: restored)
-            )
-        ) { result in
-            let completion: (@Sendable(Result<CustomerInfo, PurchasesError>) -> Void)? =
-            self.getAndRemovePurchaseCompletedCallback(forTransaction: purchasedTransaction)
-                .map { completion in
-                    return { @Sendable result in
-                        self.operationDispatcher.dispatchOnMainActor {
-                            completion(
-                                purchasedTransaction,
-                                result.value,
-                                result.error?.asPublicError,
-                                result.error?.isCancelledError ?? false
-                            )
+        self.operationDispatcher.dispatchOnWorkerThread {
+            self.transactionPoster.handlePurchasedTransaction(
+                purchasedTransaction,
+                data: .init(
+                    appUserID: self.appUserID,
+                    presentedOfferingID: offeringID,
+                    unsyncedAttributes: unsyncedAttributes,
+                    aadAttributionToken: adServicesToken,
+                    storefront: storefront,
+                    source: self.purchaseSource(for: purchasedTransaction.productIdentifier,
+                                                transaction: purchasedTransaction,
+                                                restored: restored)
+                )
+            ) { result in
+                let completion: (@Sendable(Result<CustomerInfo, PurchasesError>) -> Void)? =
+                self.getAndRemovePurchaseCompletedCallback(forTransaction: purchasedTransaction)
+                    .map { completion in
+                        return { @Sendable result in
+                            self.operationDispatcher.dispatchOnMainActor {
+                                completion(
+                                    purchasedTransaction,
+                                    result.value,
+                                    result.error?.asPublicError,
+                                    result.error?.isCancelledError ?? false
+                                )
+                            }
                         }
                     }
-                }
 
-            self.handleReceiptPost(
-                result: result,
-                subscriberAttributes: unsyncedAttributes,
-                adServicesToken: adServicesToken,
-                completion: completion
-            )
+                self.handleReceiptPost(
+                    result: result,
+                    subscriberAttributes: unsyncedAttributes,
+                    adServicesToken: adServicesToken,
+                    completion: completion
+                )
+            }
         }
     }
 
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -865,7 +865,7 @@ extension PurchasesOrchestrator: StoreKit2TransactionListenerDelegate {
         let subscriberAttributes = self.unsyncedAttributes
         let adServicesToken = self.attribution.unsyncedAdServicesToken
 
-        let result: Result<CustomerInfo, BackendError> = await Async.call { completed in
+        let result: TransactionPosterResult = await Async.call { completed in
             self.transactionPoster.handlePurchasedTransaction(
                 StoreTransaction.from(transaction: transaction),
                 data: .init(
@@ -1058,7 +1058,7 @@ private extension PurchasesOrchestrator {
                                         source: .init(isRestore: isRestore, initiationSource: initiationSource)
                                       ),
                                       observerMode: self.observerMode) { result in
-                        self.handleReceiptPost(result: result,
+                        self.handleReceiptPost(result: .init(result),
                                                subscriberAttributes: unsyncedAttributes,
                                                adServicesToken: adServicesToken,
                                                completion: completion)
@@ -1068,7 +1068,7 @@ private extension PurchasesOrchestrator {
         }
     }
 
-    func handleReceiptPost(result: Result<CustomerInfo, BackendError>,
+    func handleReceiptPost(result: TransactionPosterResult,
                            subscriberAttributes: SubscriberAttribute.Dictionary,
                            adServicesToken: String?,
                            completion: (@Sendable (Result<CustomerInfo, PurchasesError>) -> Void)?) {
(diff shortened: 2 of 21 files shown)


NachoSoto:
This is ready now

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -66,41 +109,59 @@ final class TransactionPoster: TransactionPosterType {
...
         self.productsManager = productsManager
         self.receiptFetcher = receiptFetcher
         self.backend = backend
+        self.cache = cache
         self.paymentQueueWrapper = paymentQueueWrapper
         self.systemInfo = systemInfo
         self.operationDispatcher = operationDispatcher
     }
 
+    // TODO: consumables
```
NachoSoto, line 126: Gotta consider how this works with #2841.

joshdholtz (approved):

Sources/Purchasing/Purchases/TransactionPoster.swift
aboedo: maybe we should get both into an integration branch before shipping and test them together?

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -172,13 +260,19 @@ private extension TransactionPoster {
...
                 if customerInfo.isComputedOffline {
                     completion(result)
                 } else {
                     self.finishTransactionIfNeeded(transaction) {
+                        Logger.debug(Strings.purchase.transaction_poster_storing_posted_transaction(
+                            productID: transaction.productIdentifier,
+                            transactionID: transaction.transactionIdentifier
+                        ))
+
+                        cache.savePostedTransaction(transaction)
```
aboedo, line 275: we don't need to wait for `finishTransactionIfNeeded` to save the transaction, right? Also, we'd have to wait for the main thread? 
I think we should be able to mark it as posted before finishing. In a worst-case scenario where the app crashes before finishing, then we still finish when we see that it has already been posted

Tests/UnitTests/Caching/PostedTransactionCacheTests.swift
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    func testSavesFirstTransaction() {
+        let transaction = MockStoreTransaction()
+
+        self.cache.savePostedTransaction(transaction)
+        expect(self.cache.hasPostedTransaction(transaction)) == true
+    }
+
+    func testSaveMultipleTransactions() {
```
aboedo, line 46: we should add a test to verify the behavior if the same transaction gets posted over and over (like if it's not finished)

aboedo: basically to ensure that we don't grow the cache forever

Tests/UnitTests/Identity/CustomerInfoManagerPostReceiptTests.swift
```diff
@@ -75,7 +75,57 @@ class CustomerInfoManagerPostReceiptTests: BaseCustomerInfoManagerTests {
...
+
+        expect(parameters.transaction as? StoreTransaction) === transaction
+        expect(parameters.data.appUserID) == Self.appUserID
+        expect(parameters.data.presentedOfferingID).to(beNil())
+        expect(parameters.data.unsyncedAttributes).to(beEmpty())
+        expect(parameters.data.source.isRestore) == false
+        expect(parameters.data.source.initiationSource) == .queue
+    }
+
+    func testFetchesCustomerInfoIfTransactionWasAlreadyPostedButNoCachedCustomerInfo() async throws {
```
aboedo, line 110: 🤔 
This is a weird case to think about. Not sure fetching customer info is the best path here? 

I can only imagine arriving here if: 
- we've posted the transaction before
- we call invalidate customer info cache
- we get notified again for an unfinished transaction

Or similar thinking but someone calls `restoreCompletedTransactions`. 

Feels like it'd be safer to just post in this case? Would that be a huge refactor? WDYT?

Sources/Purchasing/Purchases/TransactionPoster.swift
NachoSoto: > we don't need to wait for finishTransactionIfNeeded to save the transaction, right? 
> I think we should be able to mark it as posted before finishing. In a worst-case scenario where the app crashes before finishing, then we still finish when we see that it has already been posted

Yeah I can reorder this.

> ? Also, we'd have to wait for the main thread?
Can you explain?

Tests/UnitTests/Caching/PostedTransactionCacheTests.swift
NachoSoto: It's stored as a `Set`, but yeah I can add that.

Tests/UnitTests/Identity/CustomerInfoManagerPostReceiptTests.swift
NachoSoto: That complicates the code unnecessarily. I'd have to give `TransactionPoster` knowledge of whether there is a cached customer info or not.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1088,7 +1090,7 @@ private extension PurchasesOrchestrator {
             if let completion = self.getAndRemovePurchaseCompletedCallback(forTransaction: purchasedTransaction) {
                 self.operationDispatcher.dispatchOnMainActor {
                     completion(purchasedTransaction,
-                               result.value,
```
NachoSoto, line 1091: There's a possibility that this ends up with `nil` customer info and `nil` error, leading to a crash: https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/13951/workflows/3b380d52-bc83-4449-aa91-cf6b55235ec8/jobs/103818

Need to write a unit test for this and fix it.

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -375,6 +381,21 @@ private extension CustomerInfoManager {
         }
     }
 
+    private func cachedOrRequestCustomerInfoFetcher(
```
tonidero, line 384: I've been thinking, shouldn't this also try to request customer info first if cache is stale, instead of relying directly on the cache? 

I think this matters mostly In the event that all purchases have already been posted, this will just pick the cached version, which might be stale.

tonidero (commented):
Looks pretty good, just a comment

Sources/Identity/CustomerInfoManager.swift
NachoSoto: Good point, need to check if cache is stale. But actually this depends on the fetch policy.
TODO: need to make this depend on `CacheFetchPolicy`.

NachoSoto: Solution:
Replace `customerInfoFetcher` with a call to the top-level `customerInfo(fetchPolicy:)` but with a flag preventing it from post any unfinished transactions anymore.

NachoSoto: > but with a flag preventing it from post any unfinished transactions anymore.

I did that in `customer-info-manager-optional-post-transactions`, but I think we can't do that.
I'm going to update this with a comment explaining why.

Tests/UnitTests/Caching/PostedTransactionCacheTests.swift
NachoSoto: Done:
```swift
func testHandlePurchasedTransactionMultipleTimesPostsItOnce() throws {
    let transactionData = PurchasedTransactionData(
        appUserID: "user",
        source: .init(isRestore: false, initiationSource: .queue)
    )
    let count = 10

    self.backend.stubbedPostReceiptResult = .success(self.createCustomerInfo(nonSubscriptionProductID: nil))

    for _ in 0..<count {
        _ = try self.handleTransaction(transactionData)
    }

    expect(self.mockTransaction.finishInvokedCount) == count

    expect(self.cache.postedTransactions) == [self.mockTransaction.transactionIdentifier]
}
```

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
NachoSoto: Fixed.

Sources/Purchasing/Purchases/TransactionPoster.swift
NachoSoto: Addressed, this no longer does it in the main thread.

Sources/Caching/PostedTransactionCache.swift
```diff
@@ -32,15 +35,41 @@ final class PostedTransactionCache: PostedTransactionCacheType {
     }
 
     func savePostedTransaction(_ transaction: StoreTransactionType) {
+        RCIntegrationTestAssertNotMainThread()
```
NachoSoto, line 38: Having a large cache can make this unnecessarily slow because we need to load them all in memory. This ensures at least we don't do that in the main thread.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1117,19 +1129,27 @@ private extension PurchasesOrchestrator {
...
+                                result.value,
+                                result.error?.asPublicError,
+                                result.error?.isCancelledError ?? false
+                            )
+                        }
+                    }
                 }
-            }
+
+            self.handleReceiptPost(
```
NachoSoto, line 1147: This uses `handleReceiptPost` which simplifies handling `.alreadyPosted`.

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -33,16 +35,61 @@ struct PurchasedTransactionData {
...
 
+/// The possible results when posting a transaction.
+enum TransactionPosterResult {
+
+    /// Transaction was handled and backend returned a `CustomerInfo`
+    /// or it was computed offline.
+    case success(CustomerInfo)
+
+    /// Transaction had already been posted.
+    case alreadyPosted
```
NachoSoto, line 46: New type helps ensure that we handle this scenario everywhere that used `Result<CustomerInfo, BackendError>`.

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -208,32 +304,39 @@ private extension TransactionPoster {
...
+        switch result {
+        case let .success((customerInfo, product)):
+            self.cache.savePostedTransaction(transaction)
+
+            if Self.shouldFinish(
+                transaction: transaction,
+                for: product,
+                customerInfo: customerInfo
+            ) {
+                self.finishTransactionIfNeeded(transaction) {
```
NachoSoto, line 322: A subtle change here, suggested by @aboedo: we no longer call `finishTransactionIfNeeded` from the main actor, we only call the completion (see `complete`) above block _after_ we're done with everything.

Tests/UnitTests/Identity/CustomerInfoManagerPostReceiptTests.swift
```diff
@@ -75,25 +75,83 @@ class CustomerInfoManagerPostReceiptTests: BaseCustomerInfoManagerTests {
...
+
+        expect(parameters.transaction as? StoreTransaction) === transaction
+        expect(parameters.data.appUserID) == Self.appUserID
+        expect(parameters.data.presentedOfferingID).to(beNil())
+        expect(parameters.data.unsyncedAttributes).to(beEmpty())
+        expect(parameters.data.source.isRestore) == false
+        expect(parameters.data.source.initiationSource) == .queue
+    }
+
+    func testPostsFirstUnpostedTransaction() async throws {
```
NachoSoto, line 134: Updated this test to have some previously posted transactions too.

Tests/UnitTests/Purchasing/Purchases/PurchasesTransactionHandlingTests.swift
```diff
@@ -20,28 +20,15 @@ import XCTest
...
-    }
-
-    func testDelegateIsCalledForRandomPurchaseSuccess() throws {
-        let customerInfo = try CustomerInfo(data: Self.emptyCustomerInfoData)
-        self.backend.postReceiptResult = .success(customerInfo)
-
-        let payment = SKPayment(product: self.product)
-
-        let customerInfoBeforePurchase = try CustomerInfo(data: [
+        self.customerInfoBeforePurchase = try CustomerInfo(data: [
```
NachoSoto, line 31: This is extracting these 2 `CustomerInfo`s to avoid duplicate code.

Tests/UnitTests/Purchasing/Purchases/PurchasesTransactionHandlingTests.swift
```diff
@@ -102,27 +104,35 @@ class PurchasesTransactionHandlingTests: BasePurchasesTests {
...
+        let transaction1 = MockTransaction()
+        transaction1.mockPayment = payment
+        transaction1.mockState = .purchasing
+
+        let transaction2 = MockTransaction()
+        transaction2.mockPayment = payment
+        transaction2.mockState = .purchasing
+
+        try self.delegate.storeKit1Wrapper(self.storeKit1Wrapper, updatedTransaction: transaction1)
+        try self.delegate.storeKit1Wrapper(self.storeKit1Wrapper, updatedTransaction: transaction2)
```
NachoSoto, line 120: This test no longer had the same semantics because it used the same transaction, which was a shortcut. I made this reflect a more normal flow, and added another one below that uses the same transaction, for which the second one we don't post

Tests/UnitTests/Purchasing/Purchases/TransactionPosterTests.swift
```diff
@@ -48,8 +49,27 @@ class TransactionPosterTests: TestCase {
                 source: .init(isRestore: false, initiationSource: .queue)
             )
         )
-        expect(result).to(beFailure())
         expect(result.error) == BackendError.missingReceiptFile(self.receiptFetcher.mockReceiptURL)
+        expect(result.error?.finishable) == false
+
+        expect(self.cache.postedTransactions).to(beEmpty())
```
NachoSoto, line 55: `testHandlePurchasedTransactionWithMissingReceipt` is basically testing a not-finishable error. In this case we don't save the posted transaction.

Tests/UnitTests/Purchasing/Purchases/TransactionPosterTests.swift
```diff
@@ -48,8 +49,27 @@ class TransactionPosterTests: TestCase {
...
+
+        let result = try self.handleTransaction(
+            .init(
+                appUserID: "user",
+                source: .init(isRestore: false, initiationSource: .queue)
+            )
+        )
+        expect(result.error?.finishable) == true
+
+        self.verifyTransactionWasCached()
```
NachoSoto, line 72: For finishable errors we do.

Tests/UnitTests/Purchasing/Purchases/TransactionPosterTests.swift
```diff
@@ -64,13 +84,57 @@ class TransactionPosterTests: TestCase {
...
+        let count = 10
+
+        self.backend.stubbedPostReceiptResult = .success(self.createCustomerInfo(nonSubscriptionProductID: nil))
+
+        for _ in 0..<count {
+            _ = try self.handleTransaction(transactionData)
+        }
+
+        expect(self.backend.invokedPostReceiptDataCount) == 1
+        expect(self.mockTransaction.finishInvokedCount) == count
```
NachoSoto, line 135: Calling it 10 times to ensure that we only post it once (and finish 10 times).

Tests/UnitTests/Purchasing/Purchases/TransactionPosterTests.swift
```diff
@@ -140,14 +206,17 @@ class TransactionPosterTests: TestCase {
...
         let result = try self.handleTransaction(transactionData)
-        expect(result).to(beSuccess())
-        expect(result.value) === Self.mockCustomerInfo
+        expect(result.customerInfo) === Self.mockCustomerInfo
 
         expect(self.backend.invokedPostReceiptData) == true
         expect(self.backend.invokedPostReceiptDataParameters?.observerMode) == true
         expect(self.mockTransaction.finishInvoked) == false
+
+        self.verifyTransactionWasCached()
```
NachoSoto, line 215: Added these to all tests that are relevant.

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
```diff
@@ -104,12 +104,13 @@ class OfflineStoreKit1IntegrationTests: BaseOfflineStoreKitIntegrationTests {
...
 
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
-    func testPurchaseWhileServerIsDownSucceedsButDoesNotFinishTransaction() async throws {
+    func testPurchaseWhileServerIsDownSucceedsButDoesNotFinishOrStoreTransaction() async throws {
         self.serverDown()
         try await self.purchaseMonthlyProduct()
 
         self.logger.verifyMessageWasLogged(Strings.offlineEntitlements.computing_offline_customer_info, level: .info)
         self.verifyNoTransactionsWereFinished()
+        self.verifyNoTransactionsWereStored()
```
NachoSoto, line 113: I actually broke this, so now I'm covering it with a test.

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -60,6 +60,8 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
                 offeringID: package.offeringIdentifier
             )
         )
+        self.verifyTransactionWasFinished(count: 1)
+        self.verifyTransactionWasStored(count: 1)
```
NachoSoto, line 64: Improved the basic "can purchase package" test with these 2 assertions.

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -361,6 +374,21 @@ private extension CustomerInfoManager {
...
```
tonidero, line 377: Hmm this seems unused now after the latest changes?

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -162,15 +231,42 @@ final class TransactionPoster: TransactionPosterType {
...
-/// Async extension
+// MARK: - Extensions
+
+extension TransactionPosterResult {
+
+    typealias CustomerInfoFetcher = (@escaping @Sendable (Result<CustomerInfo, BackendError>) -> Void) -> Void
+
+    /// Converts the `TransactionPosterResult` into `Result<CustomerInfo, BackendError>`
+    /// by using `customerInfoFetcher` if the result is `.alreadyPosted`.
+    func toResult(
```
tonidero, line 242: The comment is clear but I'm concerned the name of the function doesn't indicate that this might initiate a fetch of the customer info... Maybe we could rename to `toResultOrFetchCustomerInfo`? Or something along those lines?

Sources/Caching/PostedTransactionCache.swift
```diff
@@ -32,15 +35,43 @@ final class PostedTransactionCache: PostedTransactionCacheType {
...
+    func unpostedTransactions<T: StoreTransactionType>(in transactions: [T]) -> [T] {
+        RCIntegrationTestAssertNotMainThread()
+
+        return Self.unpostedTransactions(in: transactions, with: self.storedTransactions)
+    }
+
+    // MARK: -
+
+    private var storedTransactions: StoredTransactions {
+        return self.deviceCache.value(for: CacheKey.transactions) ?? []
```
tonidero, line 63: Kinda unrelated to this PR but I was wondering, do we clear the cache values at any point? In android we clear them when we have a token but not a corresponding active purchase anymore (so the purchase has expired). We perform this check on every app foreground.

tonidero (approved):
Left a few comments but nothing major. I think this looks good.

Sources/Caching/PostedTransactionCache.swift
NachoSoto: Done in #3023 and rebased this.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1104,31 +1116,41 @@ private extension PurchasesOrchestrator {
...
-                storefront: storefront,
-                source: self.purchaseSource(for: purchasedTransaction.productIdentifier,
-                                            transaction: purchasedTransaction,
-                                            restored: restored)
-            )
-        ) { result in
-            self.handlePostReceiptResult(result,
-                                         subscriberAttributes: unsyncedAttributes,
-                                         adServicesToken: adServicesToken)
+        self.operationDispatcher.dispatchOnWorkerThread {
```
NachoSoto, line 1119: ⚠️ Another change here. We were handling transactions in the main thread without this, which now made us read/write cache in the main thread.
Thanks to the new assertions we're avoiding this now.

NachoSoto:
Closing this. As shown in #3020 this won't work with that 😢

PR #2918, opened 2023-07-28
`Paywalls`: work around SwiftUI bug to allow embedding `PaywallView` inside `NavigationStack`

NachoSoto:
Turns out that this:
```swift
NavigationStack {
    PaywallView()
}
```

Exposed `FB12674350` (https://twitter.com/nachosoto/status/1681447887077801989?s=61&t=gy8vpJJPeGJzezleoRlJiw). Except this also reproduced on simulator AND device.

Notice how it's the same thing: an `if` statement changing branches after a `@State` change. Luckily the same workaround works.

### DIFF
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -64,29 +64,31 @@ public struct PaywallView: View {
 
     @ViewBuilder
     private var content: some View {
-        if let checker = self.introEligibility, let purchaseHandler = self.purchaseHandler {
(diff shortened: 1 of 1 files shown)


NachoSoto:
I'll try to see if I have time to file a separate Radar for this. The other one only reproduced on previews but this is worse.

NachoSoto:
Filed: `FB12787354`

joshdholtz (approved):

PR #2923, opened 2023-07-31
`Paywalls`: moved purchasing state to `PurchaseHandler`

NachoSoto:
This allows `PaywallView` to handle state handling during a purchase: instead of the previous behavior where only `PurchaseButton` would disable itself, now the entire `PaywallView` is disabled.

I extracted `PackageButtonStyle` so any template with multi-package selection can get this same behavior: the package selection also becomes disabled, as well as any other button (like restore purchases).

I've also cleaned up the templates since we no longer needed the "Content" inside view, and they can be simplified with a single type.

By making `PurchaseHandler` `@MainActor` we also ensure that all these state transitions lead to UI changes happening exclusively on the main thread.

### DIFF
--- a/RevenueCatUI/Purchasing/PurchaseHandler.swift
+++ b/RevenueCatUI/Purchasing/PurchaseHandler.swift
@@ -10,6 +10,7 @@ import StoreKit
 import SwiftUI
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+@MainActor
 final class PurchaseHandler: ObservableObject {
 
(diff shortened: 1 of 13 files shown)


RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -10,6 +10,7 @@ import StoreKit
 import SwiftUI
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+@MainActor
```
NachoSoto, line 13: This fixes issues with updates happening outside the main thread.

joshdholtz (approved):

PR #2924, opened 2023-07-31
`Paywalls` added individual previews for templates

NachoSoto:
This makes it easier to preview and iterate on each individual template instead of doing that through `PaywallView`. I extracted a lot of the common code into a new `PreviewHelpers` for this.

### DIFF
--- a/RevenueCatUI/Helpers/PaywallData+Default.swift
+++ b/RevenueCatUI/Helpers/PaywallData+Default.swift
@@ -7,6 +7,7 @@
 
 import Foundation
 import RevenueCat
+import SwiftUI
 
 #if canImport(SwiftUI) && swift(>=5.7)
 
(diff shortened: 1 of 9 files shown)


RevenueCatUI/Helpers/PaywallData+Default.swift
```diff
@@ -64,4 +65,32 @@ private extension PaywallData {
...
 
+// MARK: -
+
+#if DEBUG
+
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
+@available(watchOS, unavailable)
+@available(macOS, unavailable)
+@available(macCatalyst, unavailable)
+struct DefaultPaywall_Previews: PreviewProvider {
```
NachoSoto, line 76: Also added this so we can easily iterate on this.

joshdholtz (approved):

PR #2926, opened 2023-07-31
`Paywalls`: improved default paywall configuration

NachoSoto:
This could still use more polish, but some basic fixes:
- Fixed colors
- Added application name as a title
- Using `total_price_and_per_month` instead of `price_per_month` so it also works with monthly subscriptions
- Added snapshot test for its dark mode version

### DIFF
--- a/RevenueCatUI/Data/TestData.swift
+++ b/RevenueCatUI/Data/TestData.swift
@@ -321,7 +321,7 @@ internal enum TestData {
         background: Self.paywallBackgroundImageName,
         icon: Self.paywallHeaderImageName
     )
-    static let paywallAssetBaseURL = URL(string: "https://d35rwhxn1vk1te.cloudfront.net")!
+    static let paywallAssetBaseURL = URL(string: "https://assets.pawwalls.com")!
 
     private static let offeringIdentifier = "offering"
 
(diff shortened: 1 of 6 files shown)


RevenueCatUI/Data/TestData.swift
```diff
@@ -321,7 +321,7 @@ internal enum TestData {
         background: Self.paywallBackgroundImageName,
         icon: Self.paywallHeaderImageName
     )
-    static let paywallAssetBaseURL = URL(string: "https://d35rwhxn1vk1te.cloudfront.net")!
+    static let paywallAssetBaseURL = URL(string: "assets.pawwalls.com")!
```
NachoSoto, line 324: Not strictly required for this PR but updating this now :)

RevenueCatUI/Data/Variables.swift
```diff
@@ -85,6 +87,7 @@ private extension VariableDataProvider {
 
     func value(for variableName: String, locale: Locale) -> String {
         switch variableName {
+        case "app_name": return self.applicationName
```
NachoSoto, line 90: I made this a "secret" variable so the default template can display the app name as a title.
I don't think it makes sense to expose it in the docs / frontend, since users can just write whatever their app name is there, it doesn't have to be dynamic.

Thoughts @RevenueCat/cashnip?

joshdholtz (approved):

PR #2930, opened 2023-07-31
`Paywalls`: new `.onPurchaseCompleted` modifier

NachoSoto:
This allows users to present a `PaywallView` and be notified of purchasing events. Example:
```swift
VStack {
    YourViews()

    if !self.didPurchase {
        PaywallView()
    }
}
    .onPurchaseCompleted { customerInfo in
        print("Purchase completed: \(customerInfo.entitlements)")
        self.didPurchase = true
    }
```

Additionally, an explicit optional closure is added to `.presentPaywallIfNecessary`:
```swift
YourApp()
.presentPaywallIfNecessary {
    !$0.entitlements.active.keys.contains("entitlement_identifier")
} purchaseCompleted: { customerInfo in
    print("Customer info unlocked entitlement: \(customerInfo.entitlements)")
}
```

These are now tested through new unit tests.

PR #2931, opened 2023-07-31
`Paywalls`: more unit tests for purchasing state

NachoSoto:
Follow up to #2930.

Added `PresentIfNecessaryTests` and `PurchaseHandlerTests`.
This also removes the last source of "crashes" whenever `Purchases` isn't configured, leading to errors instead.

### DIFF
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -26,8 +26,8 @@ public struct PaywallView: View {
         self.init(
             offering: nil,
             mode: mode,
-            introEligibility: Purchases.isConfigured ? .init() : nil,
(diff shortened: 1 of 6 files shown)


RevenueCatUI/PaywallView.swift
```diff
@@ -80,6 +80,10 @@ public struct PaywallView: View {
                         .transition(Self.transition)
                         .task {
                             do {
+                                guard Purchases.isConfigured else {
+                                    throw PaywallError.purchasesNotConfigured
+                                }
```
NachoSoto, line 85: @aboedo handled this too, so no more crashes from `Purchases` not being configured :)

joshdholtz (approved):

PR #2932, opened 2023-07-31
`Paywalls`: more polish from design feedback

NachoSoto:


NachoSoto:
No idea why the snapshots don't have the rounded mask in the header image, but it's still there.

PR #2933, opened 2023-08-02
`Paywalls`: initial iOS 15 support

NachoSoto:
`PaywallView` and `PaywallViewController` (along with the extensions) still require iOS 16, but this paves the way to support iOS 15 down the road if we need to.

### Main Changes:

- Changed all `@available` annotations
- Added CI tests
- Replaced `ViewThatFits` with the "biggest" version
- Because of that, `scrollableIfNeeded` is always scrollable
- Double `Regex` implementation (keeping the modern one), so once we drop support for iOS 15 we can just keep the modern one.

### TODO: 
- [x] Add snapshots
- [x] Test paywalls
- [x] Updated APITester to verify this


RevenueCatUI/Templates/OnePackageStandardTemplate.swift
```diff
@@ -21,9 +21,7 @@ struct OnePackageStandardTemplate: TemplateViewType {
         VStack(spacing: self.configuration.mode.verticalSpacing) {
             self.scrollableContent
                 .scrollableIfNecessary()
-                .scrollContentBackground(.hidden)
```
joshdholtz, line 24: Was this not needed anymore?

RevenueCatUI/Views/FooterView.swift
```diff
@@ -71,9 +71,13 @@ private struct RestorePurchasesButton: View {
...
         } label: {
-            ViewThatFits {
+            if #available(iOS 16.0, macOS 13.0, tvOS 16.0, *) {
+                ViewThatFits {
+                    Text("Restore purchases", bundle: .module)
+                    Text("Restore", bundle: .module)
+                }
+            } else {
                 Text("Restore purchases", bundle: .module)
-                Text("Restore", bundle: .module)
```
joshdholtz, line 76: Is the "Restore" not needed for iOS 15? 🤔

RevenueCatUI/Views/FooterView.swift
```diff
@@ -101,20 +105,28 @@ private struct LinkButton: View {
...
-                    destination: self.url
-                )
+        if #available(iOS 16.0, macOS 13.0, tvOS 16.0, *) {
+            ViewThatFits {
+                ForEach(self.titles, id: \.self) { title in
+                    self.link(for: title, bundle: bundle)
+                }
             }
+        } else if let first = self.titles.first {
+            self.link(for: first, bundle: bundle)
```
joshdholtz, line 115: Why is this only showing one link? 🤔

joshdholtz (commented):
Some small questions but I'm really happy that we could drop down to iOS 15 support 😊

RevenueCatUI/Views/FooterView.swift
NachoSoto: This is an optimization to make it look better if it doesn't fit. We can't use this in iOS 15 so we gotta rely on `Text` clipping the end of the word if it doesn't fit.

RevenueCatUI/Views/FooterView.swift
NachoSoto: It's the same as above, we can't use `ViewThatFits`.

NachoSoto:
> I'm really happy that we could drop down to iOS 15 support 😊

Still gotta try it and make sure that these compromises aren't too much.

RevenueCatUI/Templates/OnePackageStandardTemplate.swift
NachoSoto: It's hidden by default in a scroll view (it's not a `List`). This isn't available in iOS 15, so I just removed it.

joshdholtz (approved):

PR #2934, opened 2023-08-02
`Paywalls`: UIKit `PaywallViewController`

NachoSoto:


NachoSoto:
Feedback from @codykerns:
> - Before, I was initializing PaywallView with a specific offering (depending on some custom logic), but it doesn’t look like I can do that with this view controller - maybe make it optional as part of the init? (unless I shouldn’t be doing that)

I'm going to add the optional parameter too.

> - It feels just a little weird for customer info to be called “with” in the delegate

I'll improve the delegate method name.

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+//
+
+import RevenueCat
+import SwiftUI
+import UIKit
+
+/// A view controller for displaying `PaywallData` for an `Offering`.
+///
+/// - Seealso: ``PaywallView`` for `SwiftUI`.
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
```
joshdholtz, line 15: Will this need to get dropped to `iOS 15.0` after #2933 gets merged?

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -0,0 +1,68 @@
...
+    public override func viewWillLayoutSubviews() {
+        super.viewWillLayoutSubviews()
+
+        self.hostingController.view.frame = self.view.bounds
+    }
+
+}
+
+/// Delegate for ``PaywallViewController``.
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
```
joshdholtz, line 59: Same here about #2933

joshdholtz (approved):

RevenueCatUI/UIKit/PaywallViewController.swift
NachoSoto: Yes (**if** it gets merged, we'll see)

RevenueCatUI/UIKit/PaywallViewController.swift
NachoSoto: Yup.

PR #2941, opened 2023-08-02
`Paywalls`: added support for custom and lifetime products

NachoSoto:
The main change is that `VariableHandler` no longer cashes when trying to determine price per month for non-subscriptions. I added a test to cover this behavior.


NachoSoto:
Need to enable `.custom` too.

PR #2942, opened 2023-08-02
`Paywalls`: new `subscription_duration` variable

NachoSoto:
This will be used by template 4, but also useful in others.


guido732 (approved):
I created a PR to add this to the frontend as well 👌🏻

PR #2945, opened 2023-08-02
`Paywalls`: extracted `TemplateBackgroundImageView`

NachoSoto:
To be used by future templates.

### DIFF
--- a/RevenueCatUI/Views/TemplateBackgroundImageView.swift
+++ b/RevenueCatUI/Views/TemplateBackgroundImageView.swift
@@ -1,0 +1,36 @@
+//
+//  TemplateBackgroundImageView.swift
+//  
+//
+//  Created by Nacho Soto on 8/1/23.
+//
+
+import RevenueCat
+import SwiftUI
+
+@available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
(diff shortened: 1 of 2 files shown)


RevenueCatUI/Views/TemplateBackgroundImageView.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    var configuration: TemplateViewConfiguration
+
+    var body: some View {
+        if let url = self.configuration.backgroundImageURL {
+            self.image(url)
+                .unredacted()
+                .frame(maxWidth: .infinity, maxHeight: .infinity)
+                .edgesIgnoringSafeArea(.all)
+        }
```
joshdholtz, line 22: Does this need to get wrapped in a `VStack` or something to prevent any of those weird SwiftUI bugs that you've been seeing?

joshdholtz (approved):

RevenueCatUI/Views/TemplateBackgroundImageView.swift
NachoSoto: Good question, but this one only has one branch and no button to change that state so the but wouldn't apply 😅

PR #2963, opened 2023-08-03
`Paywalls`: changed `offerDetails` to be optional

NachoSoto:
In template 4 this is optional, so all templates will support this now.

### DIFF
--- a/RevenueCatUI/Views/IntroEligibilityStateView.swift
+++ b/RevenueCatUI/Views/IntroEligibilityStateView.swift
@@ -12,14 +12,14 @@ import SwiftUI
 @available(iOS 16.0, macOS 13.0, tvOS 16.0, *)
 struct IntroEligibilityStateView: View {
 
-    var textWithNoIntroOffer: String
+    var textWithNoIntroOffer: String?
     var textWithIntroOffer: String?
     var introEligibility: IntroEligibilityStatus?
     var foregroundColor: Color?
     var alignment: Alignment
 
     init(
-        textWithNoIntroOffer: String,
+        textWithNoIntroOffer: String?,
         textWithIntroOffer: String?,
         introEligibility: IntroEligibilityStatus?,
         foregroundColor: Color? = nil,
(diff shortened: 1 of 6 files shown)


Sources/Paywalls/PaywallData.swift
```diff
@@ -81,6 +82,10 @@ extension PaywallData {
             get { return self._callToActionWithIntroOffer }
             set { self._callToActionWithIntroOffer = newValue }
         }
+        public var offerDetails: String? {
```
NachoSoto, line 85: cc @guido732 this is optional now.

RevenueCatUI/Views/IntroEligibilityStateView.swift
```diff
@@ -38,7 +38,7 @@ struct IntroEligibilityStateView: View {
             // only if there is a custom intro text.
             .withPendingData(self.needsToWaitForIntroEligibility, alignment: self.alignment)
             // Hide if there is no intro but we have no text to ensure layout does not change.
-            .hidden(if: self.isNotEligibleForIntro && self.textWithNoIntroOffer.isEmpty)
+            .hidden(if: self.isNotEligibleForIntro && self.textWithNoIntroOffer == nil)
```
joshdholtz, line 41: Should this test for nil or empty? 🤔

RevenueCatUI/Views/IntroEligibilityStateView.swift
```diff
@@ -49,7 +49,7 @@ struct IntroEligibilityStateView: View {
         } else {
             // Display text with intro offer as a backup to ensure layout does not change
             // when switching states.
-            return self.textWithNoIntroOffer.notEmpty ?? self.textWithIntroOffer ?? ""
+            return self.textWithNoIntroOffer ?? self.textWithIntroOffer ?? ""
```
joshdholtz, line 52: Same thing here with testing for empty still for safety?

joshdholtz (approved):
Looks good! Just a few questions if we should still be checking for empty string or nil in places 🤷‍♂️

RevenueCatUI/Views/IntroEligibilityStateView.swift
NachoSoto: `PaywallData` now uses `@NonEmptyStringDecodable` for this, so it's guaranteed to not be empty if it's not `nil`.

RevenueCatUI/Views/IntroEligibilityStateView.swift
NachoSoto: Ditto.

PR #2964, opened 2023-08-03
`Paywalls`: changed default `PaywallData` to display available packages

NachoSoto:
Currently it was doing a best effort and showing `weekly`/`monthly`/`annual` if those existed in the offering. This makes it so it doesn't have to guess, and it creates a `Paywall` using the `availablePackages`.


joshdholtz (approved):
Yeesssssssssssssss 🙌

PR #2967, opened 2023-08-03
[DO NOT MERGE] Post SK2 transaction JWS instead of SK1 receipt to the receipts endpoint

MarkVillacampa:
This is just a Proof of Concept to post SK2 transaction JWS instead of SK1 receipt to the /receipts endpoint


NachoSoto:
Can we close this?

PR #2968, opened 2023-08-04
`AdServices`: Rename `postAdServicesTokenIfNeeded()` to `postAdServicesTokenOncePerInstallIfNeeded()`

joshdholtz:
### Motivation

It was not clear what `postAdServicesTokenIfNeeded()` was doing for anybody unfamiliar with our codebase or AdServices. This method is in `applicationWillEnterForeground` and without digging into any of the implementation, it looks like it could send the attribution token every time the app enters the foreground. 

The behavior was actually only sending it once per install (as it should) so the method rename reflects that.

### Description
Renames `postAdServicesTokenIfNeeded()` to `postAdServicesTokenOncePerInstallIfNeeded()`

### DIFF
--- a/Sources/Attribution/AttributionPoster.swift
+++ b/Sources/Attribution/AttributionPoster.swift
@@ -127,7 +127,7 @@ final class AttributionPoster {
     @available(iOS 14.3, macOS 11.1, macCatalyst 14.3, *)
     @available(tvOS, unavailable)
     @available(watchOS, unavailable)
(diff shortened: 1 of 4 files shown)


NachoSoto (approved):
I think `postAdServicesTokenOncePerInstallIfNeeded` (without only) is a bit clearer, but this is good too.

NachoSoto:
Did you mean to make this a "fix"? I changed it to `refactor` 😇

PR #2978, opened 2023-08-07
`Paywalls`: warm-up image cache

NachoSoto:
This relies on `AsyncImage` using `URLSession.shared`

I've tested this, and the app doesn't even make a request the second time around because the images are caches on disk.


joshdholtz (approved):
:shipit:

PR #2954, opened 2023-08-02
`CustomerInfoManager`: post transactions in parallel to POST receipts only once

NachoSoto:
The existing implementation was posting them in sequence, which meant that it was always going to make N requests.
With this new implementation we can ensure that only one request is made for all transactions.

PR #2983, opened 2023-08-08
`Performance`: throttle cache updates, and avoid double updates on app launch

NachoSoto:
Possibly as a consequence of #2623 we were calling `updateAllCachesIfNeeded` on app launch together with `updateCachesIfInForeground`, meaning that we were updating all caches twice.
Thanks to our de-duping logic this is normally not a big issue, but this issue combined with what #2954 fixes, means that if there are pending transactions, opening the app would probably end up posting receipts duplicated times.

This adds a new log as well to detect this:
> [purchases] DEBUG: ℹ️ Throttling cache update, only 0.8 seconds elapsed

And a new integration test that posts `UIApplication.willEnterForegroundNotification` to test this.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -233,6 +233,8 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
         set { self.systemInfo.finishTransactions = newValue }
     }
 
+    private let lastCacheUpdateTime: Atomic<DispatchTime?> = nil
+
     private let attributionFetcher: AttributionFetcher
     private let attributionPoster: AttributionPoster
     private let backend: Backend
@@ -1487,10 +1489,13 @@ private extension Purchases {
     @objc func applicationWillEnterForeground() {
         Logger.debug(Strings.configure.application_foregrounded)
 
+        guard !self.shouldSkipCacheUpdate() else { return }
+
         // Note: it's important that we observe "will enter foreground" instead of
         // "did become active" so that we don't trigger cache updates in the middle
         // of purchases due to pop-ups stealing focus from the app.
         self.updateAllCachesIfNeeded(isAppBackgrounded: false)
+        self.setLastCacheUpdateTime()
         self.dispatchSyncSubscriberAttributes()
 
         #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
@@ -1535,6 +1540,8 @@ private extension Purchases {
                 self.operationDispatcher.dispatchOnWorkerThread {
                     self.updateAllCaches(isAppBackgrounded: isBackgrounded, completion: nil)
                 }
+
+                self.setLastCacheUpdateTime()
             }
         }
     }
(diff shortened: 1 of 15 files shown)


Tests/UnitTests/SubscriberAttributes/PurchasesSubscriberAttributesTests.swift
```diff
@@ -218,30 +222,42 @@ class PurchasesSubscriberAttributesTests: TestCase {
...
 
         expect(self.mockNotificationCenter.observers).toNot(beEmpty())
 
         expect(self.mockNotificationCenter.observers).to(containElementSatisfying {
             $0.notificationName == SystemInfo.applicationDidEnterBackgroundNotification
         })
 
+        let previousCount = self.mockSubscriberAttributesManager.invokedSyncAttributesForAllUsersCount
         self.mockNotificationCenter.fireNotifications()
-        expect(self.mockSubscriberAttributesManager.invokedSyncAttributesForAllUsersCount) == 2
```
NachoSoto, line 244: This was only 2 because `fireNotifications` was firing the `didEnterForeground` and `didEnterBackground`, and this test was detecting the double update.
Now it's `1` because it only gets updated automatically (regardless of throttling) when entering the background.

Tests/UnitTests/SubscriberAttributes/PurchasesSubscriberAttributesTests.swift
```diff
@@ -256,7 +272,7 @@ class PurchasesSubscriberAttributesTests: TestCase {
 
         self.mockNotificationCenter.fireNotifications()
 
-        expect(self.mockSubscriberAttributesManager.invokedSyncAttributesForAllUsersCount) == 2
```
NachoSoto, line 259: This was only 2 because `fireNotifications` was firing the `didEnterForeground` and `didEnterBackground`, and this test was detecting the double update.
Now it's `1` because it only gets updated automatically (regardless of throttling) when entering the background.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1487,10 +1489,13 @@ private extension Purchases {
     @objc func applicationWillEnterForeground() {
         Logger.debug(Strings.configure.application_foregrounded)
 
+        guard !self.shouldSkipCacheUpdate() else { return }
+
         // Note: it's important that we observe "will enter foreground" instead of
         // "did become active" so that we don't trigger cache updates in the middle
         // of purchases due to pop-ups stealing focus from the app.
         self.updateAllCachesIfNeeded(isAppBackgrounded: false)
+        self.setLastCacheUpdateTime()
```
joshdholtz, line 1498: I think this naming might be a little confusing here 🤔 `setLastCacheUpdateTime()` makes it seem like this method should be called inside of `updateAllCachesIfNeeded()` when its completed. Should it some like... `setLastCacheFetchedTime()` 🤷‍♂️

joshdholtz (approved):
This looks good to me! Just a little thought on naming 🤷‍♂️

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1487,10 +1489,13 @@ private extension Purchases {
...
         self.dispatchSyncSubscriberAttributes()
```
tonidero, line 1499: This is now throttled in the `applicationWillEnterForeground` but not on the `applicationDidEnterBackground`... Not a blocker but wanted to mention it.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1487,10 +1489,13 @@ private extension Purchases {
     @objc func applicationWillEnterForeground() {
         Logger.debug(Strings.configure.application_foregrounded)
 
+        guard !self.shouldSkipCacheUpdate() else { return }
```
tonidero, line 1492: I feel we should extract all the code here to an `updateCachesOnAppForeground` or something like that, and move this there... Just in case we want to do something else on app foreground and we don't want to skip it. NABD though.

tonidero (approved):
A couple comments but nothing blocking

NachoSoto:
Decided that we don't need this.
Completion-block-based de-dupping + integration tests (`testSimultanousCallsToGetCustomerInfoWithPendingTransactionPostsReceiptOnlyOnce` / #3013) are enough.

PR #2986, opened 2023-08-08
`HTTPClient`: extracted `HTTPRequestPath` protocol

NachoSoto:
This paves the way to add a second set of paths with a separate server URL to be able to shed requests to offerings and product entitlement mapping to a separate server.

### Changes:
- New `HTTPRequestPath` with the same properties as `HTTPRequest.Path` + a new `serverHostURL`
- Moved proxy logic from `SystemInfo`/`HTTPClient` to `HTTPRequestPath`
- Simplified `HTTPRequest` implementation

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -418,7 +418,10 @@ private extension HTTPClient {
     }
 
     func convert(request: Request) -> URLRequest? {
-        var urlRequest = URLRequest(url: request.httpRequest.path.url(proxyURL: SystemInfo.proxyURL))
+        guard let requestURL = request.httpRequest.path.url(proxyURL: SystemInfo.proxyURL) else {
+            return nil
+        }
+        var urlRequest = URLRequest(url: requestURL)
         urlRequest.httpMethod = request.method.httpMethod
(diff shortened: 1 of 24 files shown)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -418,11 +418,7 @@ private extension HTTPClient {
     }
 
     func convert(request: Request) -> URLRequest? {
-        guard let requestURL = request.httpRequest.path.url else {
```
NachoSoto, line 421: `url` is no longer optional which is a win.

Sources/Networking/HTTPClient/HTTPRequestPath.swift
```diff
@@ -0,0 +1,187 @@
...
+        return baseURL.appendingPathComponent(self.relativePath)
+    }
+
+}
+
+// MARK: - Main paths
+
+extension HTTPRequest {
+
+    enum Path: Hashable {
```
NachoSoto, line 51: It is now trivial to add a second `enum` for the new endpoints, which can be passed to `HTTPClient` and it'll make a request to the right place.
That way it will be impossible to make requests to the new server for paths that shouldn't go there.

PR #2989, opened 2023-08-08
`visionOS`: support for `Xcode 15 beta 6`

NachoSoto:
### Context:

`#if os(iOS)` no longer evaluates to `true` when compiling for `visionOS`
Which means we need `|| os(xrOS)`.
But to support older Xcode we need `#if swift(>=5.9) && os(xrOS)`.
But, fun fact, short-circuiting doesn’t work in `Xcode 13`.
And we still need that to run iOS <15 tests (`FB12223404` / `FB12805891`)

💩 

This works around that by defining our own `VISION_OS` flag.

### Changes:
- Added `VISION_OS` compile time flag to determine SDK
- Replaced `#if os(xrOS)` with `#if VISION_OS`
- Added flag to `Package.swift` (thanks @neonichu for helping with this)


Package@swift-5.5.swift
```diff
@@ -0,0 +1,56 @@
+// swift-tools-version:5.5
```
NachoSoto, line 1: This remains here now for Xcode 13/14 compatibility. Moving forward, the main `Package.swift` is what Xcode 15 will use.

MarkVillacampa (approved):
Nice one!

PR #2990, opened 2023-08-08
`CI`: add `visionOS` build

NachoSoto:


NachoSoto:
Forgot to mention that I'm only building and not running tests because Nimble doesn't support visionOS and we can't update it without losing iOS 11 support.

PR #2991, opened 2023-08-09
`Paywalls`: prevent multiple concurrent purchases

NachoSoto:
Tapping multiple times on the `PurchaseButton` can lead to "purchase already in progress", because of a race condition with SwiftUI not updating the `.disabled` state on time. This works around that.


thillsman (approved):
Thanks! 🙌

joshdholtz (approved):
:shipit:

PR #3005, opened 2023-08-11
`Xcode 15`: fix non-`visionOS` build by replacing `.custom` platform

### LINKED ISSUE #2998: RevenueCat fails to build with Xcode 15 Beta 6
I am unable to build RevenueCat 4.25.3 using Xcode 15 Beta 6.

When building for macOS I get the error "No such module 'UIKit'" in `UIApplication+RCExtensions.swift`, and when building for iOS I get the error "'purchase(confirmIn:options:)' is only available in iOS 17.0 or newer" in `PurchasesOrchestrator.swift`.  It seems like there is an issue with the changes made in https://github.com/RevenueCat/purchases-ios/pull/2989 and the VISION_OS flag is always being set.

NachoSoto:
Fixes #2998. See #2989 for why this is needed in the first place.
This is now fixed in SPM: https://github.com/apple/swift-package-manager/pull/6794, but we need a workaround for now.

Since `.custom` doesn't work properly (it gets evaluated in Xcode even outside of that platform), this splits `Package.swift` based on Xcode version.

Thanks @neonichu for helping with this.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -16,6 +16,10 @@ if shouldIncludeDocCPlugin {
     dependencies.append(.package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"))
 }
 
+// See https://github.com/RevenueCat/purchases-ios/pull/2989
+// #if os(xrOS) can't really be used in Xcode 13, so we use this instead.
+let visionOSSetting: SwiftSetting = .define("VISION_OS", .when(platforms: [.visionOS]))
+
 let package = Package(
     name: "RevenueCat",
     platforms: [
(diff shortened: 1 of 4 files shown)


Package.swift
```diff
@@ -16,21 +16,14 @@ if shouldIncludeDocCPlugin {
...
-
 let package = Package(
     name: "RevenueCat",
     platforms: [
         .macOS(.v10_13),
         .watchOS("6.2"),
         .tvOS(.v11),
-        .iOS(.v11)
+        .iOS(.v11),
+        .visionOS(.v1)
```
NachoSoto, line 26: An advantage of this is being able to define this too, even though for some reason it's not actually required.

tonidero (approved):
Not very familiar with this, but makes sense 🚢

Package@swift-5.8.swift
```diff
@@ -0,0 +1,56 @@
+// swift-tools-version:5.8
```
aboedo, line 1: can this be 5.6 again?

NachoSoto: Just added a separate `Package@swift-5.7.swift` to keep Xcode 14.1 compatibility (the min to submit to the App Store), and added a CI job to verify that.

PR #3009, opened 2023-08-14
Added `internal` `NonSubscriptionTransaction.storeTransactionIdentifier`

NachoSoto:
This will be used for #2841.

### DIFF
--- a/Sources/Purchasing/NonSubscriptionTransaction.swift
+++ b/Sources/Purchasing/NonSubscriptionTransaction.swift
@@ -28,11 +28,15 @@ public final class NonSubscriptionTransaction: NSObject {
     /// The date that App Store charged the user’s account.
     @objc public let purchaseDate: Date
 
-    /// The unique identifier for the transaction.
+    /// The unique identifier for the transaction created by RevenueCat.
     @objc public let transactionIdentifier: String
+
+    /// The unique identifier for the transaction created by the Store.
+    @objc internal let storeTransactionIdentifier: String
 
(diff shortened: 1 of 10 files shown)


Sources/Purchasing/NonSubscriptionTransaction.swift
```diff
@@ -23,22 +23,38 @@ public final class NonSubscriptionTransaction: NSObject {
     /// The date that App Store charged the user’s account.
     @objc public let purchaseDate: Date
 
-    /// The unique identifier for the transaction.
+    /// The unique identifier for the transaction created by RevenueCat.
     @objc public let transactionIdentifier: String
 
+    /// The unique identifier for the transaction created by the Store.
+    @objc public let storeTransactionIdentifier: String
```
aboedo, line 30: I'd make it internal for now, because: 
a) we don't currently expose it for subscriptions, so maybe it's better to avoid confusion
b) it's also not exposed on android, so maybe it's better for parity?
c) we're doing related work with external purchases API, we should ensure that we unify naming before exposing

aboedo (commented):
looks great, but I'd maybe keep it as internal until we have a clear reason to expose it as public

Sources/Purchasing/NonSubscriptionTransaction.swift
NachoSoto: 👍🏻 

Will do.

aboedo (approved):

PR #3012, opened 2023-08-14
`CustomerInfoManager`: allow disabling post-transaction behavior

NachoSoto:
See https://github.com/RevenueCat/purchases-ios/pull/2914#discussion_r1291605284
This will be used by that PR as a fallback whenever we try to post a transaction that had already been posted.


NachoSoto:
Actually this might not be needed, making it a draft for now.

PR #3013, opened 2023-08-14
`Integration Tests`: add tests to verify `CustomerInfo`+`Offerings` request de-dupping

NachoSoto:
This adds more coverage for why we don't actually need #2983.

### DIFF
--- a/Tests/BackendIntegrationTests/OtherIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/OtherIntegrationTests.swift
@@ -31,6 +31,74 @@ class OtherIntegrationTests: BaseBackendIntegrationTests {
         expect(info.isComputedOffline) == false
     }
 
+    func testGetCustomerInfoMultipleTimesInParallel() async throws {
+        let requestCount = 3
+
+        let purchases = try self.purchases
+
+        // 1. Make sure any existing customer info requests finish
+        _ = try await purchases.customerInfo()
+        // 2. Invalidate cache
+        purchases.invalidateCustomerInfoCache()
+        self.logger.clearMessages()
+
+        // 3. Request customer info multiple times in parallel
+        await withThrowingTaskGroup(of: Void.self) {
+            for _ in 0..<requestCount {
+                $0.addTask { _ = try await purchases.customerInfo() }
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


Tests/BackendIntegrationTests/OtherIntegrationTests.swift
```diff
@@ -31,6 +31,74 @@ class OtherIntegrationTests: BaseBackendIntegrationTests {
         expect(info.isComputedOffline) == false
     }
 
+    func testGetCustomerInfoMultipleTimesInParallel() async throws {
```
aboedo, line 34: maybe we could have another test that also verifies the behavior when you call `customerInfo()` right after the response came back? i.e.: verifying that we're still respecting the TTL

Tests/BackendIntegrationTests/OtherIntegrationTests.swift
```diff
@@ -31,6 +31,74 @@ class OtherIntegrationTests: BaseBackendIntegrationTests {
...
+                $0.addTask { _ = try await purchases.customerInfo() }
+            }
+        }
+
+        // 4. Verify N-1 requests were de-duped
+        self.logger.verifyMessageWasLogged(
+            "Network operation 'GetCustomerInfoOperation' found with the same cache key",
+            level: .debug,
+            expectedCount: requestCount - 1
+        )
```
aboedo, line 57: nothing to do for this PR, but I feel like relying on logs is somewhat brittle. Perhaps we should consider having a spy for the HTTPClient, such that we don't modify behavior but we can check exactly what was sent to it

aboedo (approved):

Tests/BackendIntegrationTests/OtherIntegrationTests.swift
NachoSoto: Hmm yeah I don't think we have any _integration_ test for that, let me add it.

Tests/BackendIntegrationTests/OtherIntegrationTests.swift
NachoSoto: Yeah totally, this has been an extremely useful hack, but a hack after all. It's become our default "spy" for anything, but I agree especially for something like `HTTPClient` we should expose something instead.

Issue #3020, opened 2023-08-16
The completion block is not working when downgrade plan


Abbott-Deng:
-  I have updated Purchases SDK to the latest version
-  I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-android/blob/main/CONTRIBUTING.md)
-  I have searched the [Community](https://community.revenuecat.com)
-  I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
-  I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-android/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

1. Environment
   1. Platform: iOS
   2. SDK version:4.25.3
   3. OS version:iOS 17
2. Steps to reproduce, I purchase / upgrade plan is working fine, But It's reproduced when I try to downgrade my plan.

**Additional context**
```
2023-08-15 10:40:29.153671+0800 Hobnob-Staging[1667:246401] [purchases] INFO: 💰 Purchasing Product 'staging_sub_premium_2_yearly'
2023-08-15 10:40:29.153911+0800 Hobnob-Staging[1667:246401] [purchases] DEBUG: ℹ️ Adding payment for product 'staging_sub_premium_2_yearly'. 0 transactions already in the queue.
2023-08-15 10:40:29.156959+0800 Hobnob-Staging[1667:247367] [purchases] DEBUG: ℹ️ StoreKit1Wrapper (0x00000002813a89c0) updatedTransaction: staging_sub_premium_2_yearly 0
2023-08-15 10:40:29.165165+0800 Hobnob-Staging[1667:247367] [purchases] VERBOSE: The SKPaymentTransaction has a nil value for transaction date - this is a bug in StoreKit.
2023-08-15 10:40:29.165220+0800 Hobnob-Staging[1667:247367] [purchases] VERBOSE: The SKPaymentTransaction has a nil value for transaction identifier - this is a bug in StoreKit.
2023-08-15 10:40:56.451599+0800 Hobnob-Staging[1667:247796] [purchases] DEBUG: ℹ️ StoreKit1Wrapper (0x00000002813a89c0) updatedTransaction: staging_sub_premium_2_yearly 2000000390174047 2000000390162515 1
[...]

PR #3023, opened 2023-08-16
`PostedTransactionCache`: add a TTL to every transaction

NachoSoto:
This was suggested in #2914 as a way to prevent the cache from growing forever.

### Other changes:
- Added `PostedTransactionCache.unpostedTransactions`: this was added in #2914, so I put it here to prevent merge conflicts.
- `PostedTransactionCache` now verifies that cache interaction doesn't happen in the main thread
- Added debug log when saving transactions
- Added debug log when pruning cache
- Added `ClockType` tests

### DIFF
--- a/Sources/Caching/PostedTransactionCache.swift
+++ b/Sources/Caching/PostedTransactionCache.swift
@@ -19,28 +19,78 @@ protocol PostedTransactionCacheType: Sendable {
     func savePostedTransaction(_ transaction: StoreTransactionType)
     func hasPostedTransaction(_ transaction: StoreTransactionType) -> Bool
 
+    /// - Returns: the subset of `transactions` that have not been posted.
+    func unpostedTransactions<T: StoreTransactionType>(in transactions: [T]) -> [T]
+
 }
 
 final class PostedTransactionCache: PostedTransactionCacheType {
 
-    private typealias StoredTransactions = Set<String>
+    /// The format of the cache: `StoreTransactionType.transactionIdentifier` -> `Date`.
+    private typealias StoredTransactions = [String: Date]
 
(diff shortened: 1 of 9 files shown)


Sources/Caching/PostedTransactionCache.swift
```diff
@@ -19,28 +19,78 @@ protocol PostedTransactionCacheType: Sendable {
...
+        RCIntegrationTestAssertNotMainThread()
+
+        Logger.debug(Strings.purchase.saving_posted_transaction(transaction))
+
         self.deviceCache.update(key: CacheKey.transactions,
-                                default: Set<String>()) { transactions in
-            transactions.insert(transaction.transactionIdentifier)
+                                default: StoredTransactions()) { transactions in
+            self.pruneOldTransactions(from: &transactions)
+            transactions[transaction.transactionIdentifier] = self.clock.now
```
NachoSoto, line 48: This entire modification is done _in place_ so we don't end up with 2 dictionary copies in memory.

Tests/UnitTests/Caching/PostedTransactionCacheTests.swift
```diff
@@ -54,4 +56,105 @@ class PostedTransactionCacheTests: TestCase {
         expect(self.cache.hasPostedTransaction(transaction2)) == true
     }
 
+    func testHasPostedTransactionDoesNotTakeDateIntoAccount() {
```
NachoSoto, line 59: This makes `hasPostedTransaction` faster, we only prune when writing.

NachoSoto:
Closing this. As shown in #3020 this won't work with that 😢

PR #3025, opened 2023-08-16
`Integration Tests`: improved `testCanPurchaseMultipleSubscriptions`

NachoSoto:
I was looking into #3020, and saw that we already have this test to cover downgrades. I added an extra assertion to verify entitlements.
I also ran this test on iOS 17 and it passes.


Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -182,6 +182,8 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
         _ = try await self.purchases.purchase(product: product1)
         let info = try await self.purchases.purchase(product: product2).customerInfo
 
+        try await verifyEntitlementWentThrough(info)
```
NachoSoto, line 185:
```suggestion
        try await self.verifyEntitlementWentThrough(info)
```

aboedo (approved):

PR #3028, opened 2023-08-16
`PurchasesOrchestrator`: fixed callback not invoked regression during downgrades

NachoSoto:
Fixes #3020.
This essentially reverts #2890.

As shown in #3020, downgrades are reported with the same transaction identifier as the original transaction. Therefore our approach in #2890 is invalid.
We considered checking whether the `SKPayment` in the `SKPaymentTransaction` matches the one added to the queue as a way to disambiguate these transactions, but it doesn't. So we have no way of telling the difference between queue transactions and purchases.

I've left the original tests as a way to at least document this behavior.

PR #3034, opened 2023-08-17
`Xcode 15`: fixed release build

NachoSoto:
Fixes https://github.com/RevenueCat/purchases-ios/pull/3028#issuecomment-1681533213
This was wrong in #3032 because `Purchases.storeKit2Setting` is only available in `DEBUG` builds, but we didn't catch it because we have no CI for Xcode 15 yet, which this adds as well.


joshdholtz (approved):
🚀

PR #3042, opened 2023-08-17
`visionOS`: changed CI job to Release

NachoSoto:
The `build_tv_watch_mac` job already builds on Release mode (`Carthage`'s default).
This would have also helped detect the issue fixed by #3034.

I've also changed the API tester jobs to compile on release, though those aren't ran by CI.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -203,6 +203,7 @@ platform :ios do
       workspace: 'RevenueCat.xcworkspace',
       scheme: 'RevenueCat',
       destination: 'generic/platform=visionOS',
+      configuration: 'release',
       xcargs: "CODE_SIGN_IDENTITY='' CODE_SIGNING_REQUIRED=NO"
     )
   end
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -237,7 +238,8 @@ platform :ios do
     xcodebuild(
       workspace: 'RevenueCat.xcworkspace',
       scheme: 'SwiftAPITester',
-      destination: 'generic/platform=iOS Simulator'
+      destination: 'generic/platform=iOS Simulator',
+      configuration: 'release',
```
tonidero, line 242: I wonder if we should build the api tests in both debug and release, in case we have any debug-only apis... I agree release is more important though, so this makes sense.

tonidero (approved):

fastlane/Fastfile
NachoSoto: We compile API testers in debug when running tests.

PR #3045, opened 2023-08-17
`StoreKit 2`: added warning to paywall constructors

NachoSoto:
This became a requirement with #3032.

### DIFF
--- a/Sources/Support/PaywallExtensions.swift
+++ b/Sources/Support/PaywallExtensions.swift
@@ -24,7 +24,10 @@ extension StoreView {
     /// Creates a view to load a collection of products from the App Store, and merchandise them using an
     /// icon and custom placeholder icon.
     /// When the user purchases products through this paywall, the `RevenueCat` SDK will handle
-    /// the result automatically. All you need to do is to dismiss the paywall
+    /// the result automatically. All you need to do is to dismiss the paywall.
+    ///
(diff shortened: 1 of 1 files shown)


Sources/Support/PaywallExtensions.swift
```diff
@@ -48,7 +51,10 @@ extension StoreView where Icon == EmptyView, PlaceholderIcon == EmptyView {
 
     /// Creates a view to load a collection of products from the App Store, and merchandise them.
     /// When the user purchases products through this paywall, the `RevenueCat` SDK will handle
-    /// the result automatically. All you need to do is to dismiss the paywall
+    /// the result automatically. All you need to do is to dismiss the paywall.
+    ///
+    /// - Warning: In order to use StoreKit paywalls you must configure the `RevenueCat` SDK
```
tonidero, line 56: I haven't been tracking the Paywall changes, but would it make sense to log an error if a user tries to display the paywall using SK1? I was also thinking of actually erroring out, but not sure how that would work, so a log might be enough.

tonidero (approved):

Sources/Support/PaywallExtensions.swift
NachoSoto: That's done already :) https://github.com/RevenueCat/purchases-ios/blob/4.25.6/Sources/Support/PaywallExtensions.swift#L112

PR #3046, opened 2023-08-17
`Paywalls`: handle already purchased state

NachoSoto:
This adds `currentlySubscribed` to `TemplateViewConfiguration.Package` so that `PurchaseButton` can access that information.

`PurchaseButton` now gets automatically disabled for products that are already subscribed.

### DIFF
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -101,10 +101,12 @@ public struct PaywallView: View {
                                     throw PaywallError.purchasesNotConfigured
                                 }
 
-                                guard let offering = try await Purchases.shared.offerings().current else {
-                                    throw PaywallError.noCurrentOffering
+                                if self.offering == nil {
+                                    guard let offering = try await Purchases.shared.offerings().current else {
+                                        throw PaywallError.noCurrentOffering
+                                    }
+                                    self.offering = offering
                                 }
-                                self.offering = offering
 
                                 if self.customerInfo == nil {
                                     self.customerInfo = try await Purchases.shared.customerInfo()
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -20,6 +20,8 @@ public struct PaywallView: View {
     @State
     private var offering: Offering?
     @State
+    private var customerInfo: CustomerInfo?
+    @State
     private var error: NSError?
 
     /// Create a view that loads the `Offerings.current`.
@@ -32,6 +34,7 @@ public struct PaywallView: View {
     ) {
         self.init(
             offering: nil,
+            customerInfo: nil,
             fonts: fonts,
             introEligibility: .default(),
             purchaseHandler: .default()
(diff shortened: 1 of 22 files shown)


NachoSoto:
Just gotta actually use this new flag, want to discuss the best way to visualize this in paywalls /cc @aboedo

RevenueCatUI/PaywallView.swift
```diff
@@ -83,8 +89,9 @@ public struct PaywallView: View {
     private var content: some View {
         VStack { // Necessary to work around FB12674350 and FB12787354
             if let checker = self.introEligibility, let purchaseHandler = self.purchaseHandler {
-                if let offering = self.offering {
+                if let offering = self.offering, let customerInfo = self.customerInfo {
                     self.paywallView(for: offering,
+                                     activelySubscribedProductIdentifiers: customerInfo.activeSubscriptions,
```
joshdholtz, line 94: This _may_ not actually be enough information to accurately show a purchased products with paywalls due to cross platform 😬 

One case for this is if your offering looks like...
- Monthly
  -  Apple - `pro_annually`
  - Google - `pro_yearly`
  
If the user is logged in and sharing purchases a crossed platforms, the paywall would incorrectly show their state and they _could_ accidentally purchase another subscription. 

It _could_ be on the developer to make sure the paywall is shown correctly but it still could happen 🤷‍♂️ 

The thing with this is... I don't think we have a solution for this right now 😅 We would essentially need a way to get "packages purchased" from the customer info so that it correctly maps to the packages that we are showing in the Paywall UI

joshdholtz (commented):
Left a thought on why `customerInfo.activeSubscriptions` might not be sufficient enough 😬

RevenueCatUI/PaywallView.swift
NachoSoto: Yeah that's right... 
Given that this is just a "nice to have" I don't think it's a big deal if we don't detect that particular scenario.

RevenueCatUI/Views/PurchaseButton.swift
```diff
@@ -90,6 +90,7 @@ struct PurchaseButton: View {
         .buttonStyle(.borderedProminent)
         .frame(maxWidth: .infinity)
         .dynamicTypeSize(...Constants.maximumDynamicTypeSize)
+        .disabled(self.package.currentlySubscribed)
```
NachoSoto, line 93: How it's used right now.

joshdholtz (approved):

RevenueCatUI/View+PresentPaywall.swift
```diff
@@ -115,14 +115,15 @@ private struct PresentingPaywallModifier: ViewModifier {
     var purchaseHandler: PurchaseHandler?
 
     @State
-    private var isDisplayed = false
+    private var state: (displayed: Bool, customerInfo: CustomerInfo?) = (false, nil)
```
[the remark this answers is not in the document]
NachoSoto: It could be, but given that this is only used here I'd say it's overkill.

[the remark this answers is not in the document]
NachoSoto: Actually I simplified this by using a single value now :)

PR #3049, opened 2023-08-21
`Paywalls`: rename card to footer

joshdholtz:
### Motivation

Rename "card" to "footer" to because that is how we expect developers to use the current implementation of the non-fullscreen modes as a footer to their custom paywalls.

### Description
- Renamed `card` to `footer`
- Renamed `condensedCard` to `condensedFooter`


NachoSoto (approved):
I think this makes much more sense 👍🏻 
FYI I changed the tags. If you make this `breaking` that would trigger purchases-ios 5.0 which I don't think you want to 😅 since we haven't shipped paywalls we can make any breaking changes.

NachoSoto:
Hmm not sure about footer. Let's talk about it tomorrow?

joshdholtz:
> Hmm not sure about footer. Let's talk about it tomorrow?

@NachoSoto Was planning on chatting with you tomorrow about this 😊 Had some realizations with Charlie and Dan today about a footer view (and I have a followup PR that coming soon) that makes will hopefully make this change make more sense

Tests/RevenueCatUITests/Templates/Template2ViewTests.swift
```diff
@@ -23,20 +23,20 @@ class Template2ViewTests: BaseSnapshotTest {
         .snapshot(size: Self.fullScreenSize)
     }
 
-    func testCardPaywall() {
+    func testFooterPaywall() {
```
NachoSoto, line 26: Need to update the snapshot names

NachoSoto:
Merging this.

PR #3051, opened 2023-08-22
`Paywalls`: new `PaywallFooterView` to replace `modes`

joshdholtz:
## Motivation

It was a little unclear how and when to use the different `PaywallView` modes (`.fullscreen`, `.card`, and `.condensedCard`). The actual functionality of the modes is great but they seemed like three different components in one view and it was just trying to do too much for one public API.

This change will...
- Remove modes from being a public parameter in `PaywallView` and only allow it to be full screen
- Create a new `PaywallFooterView` which will combine `.footer` (previously `.card`) and `.condensedFooter` (previously `.condensedCard`)
- Removed the `modes` parameter from `PaywallView`

### New `PaywallFooterView`

- Wraps `PaywallView` and will only choose between `.footer` and `.footerCondensed`
- Currently offers a `condensed` parameter but this might be switched to being a remote setting inside of `PaywalData`

### New `.paywallFooter()` view modifier

- Easy way for developers to place `PaywallFooterView` where it is supposed to be placed
- This places the view in `.safeAreaInset(edge: .bottom)`
  - This works on any view but is ideal in `ScrollView` as it will automatically handle adjusting the scroll insets with the paywalls size

### New footer structure and animation (with some hacks)

- No parts of `PaywallFooterView` uses `.overlay` anymore
  - The whole grows in `.condensedFooter` when "All Plans" is toggled

### My sample app

https://github.com/RevenueCat/purchases-ios/assets/401294/17c52bba-956d-404e-b991-a2272e7fc4f1

### Simple sample app 

https://github.com/RevenueCat/purchases-ios/assets/401294/4ea94156-8367-4dea-9640-0cce8ffa1deb


RevenueCatUI/PaywallFooterView.swift
```diff
@@ -0,0 +1,91 @@
...
+            condensed: condensed,
+            fonts: fonts,
+            introEligibility: .default(),
+            purchaseHandler: .default()
+        )
+    }
+
+    init(
+        offering: Offering?,
+        condensed: Bool = false,
```
joshdholtz, line 69: Going with `condensed: Bool` for now to make it easier on which type of footer is wanted but this might get moved to a remote configuration in `PaywallData`

RevenueCatUI/Data/Constants.swift
```diff
@@ -12,10 +12,12 @@ enum Constants {
 
     static let defaultAnimation: Animation = .easeInOut(duration: 0.2)
     static let fastAnimation: Animation = .easeInOut(duration: 0.1)
-    static let displayAllPlansAnimation: Animation = .easeInOut(duration: 0.2)
+    static let toggleAllPlansAnimation: Animation = .spring(response: 0.35, dampingFraction: 0.7)
 
     static let defaultCornerRadius: CGFloat = 20
 
+    static let maxFooterHeight: CGFloat = 400
```
joshdholtz, line 19: Leaving unlimited height for footer felt bad so just putting this at 400 for now? But not sold on this at all 🙃

RevenueCatUI/Data/Constants.swift
```diff
@@ -12,10 +12,12 @@ enum Constants {
 
     static let defaultAnimation: Animation = .easeInOut(duration: 0.2)
     static let fastAnimation: Animation = .easeInOut(duration: 0.1)
-    static let displayAllPlansAnimation: Animation = .easeInOut(duration: 0.2)
+    static let toggleAllPlansAnimation: Animation = .spring(response: 0.35, dampingFraction: 0.7)
```
NachoSoto, line 15: 😍

RevenueCatUI/Modifiers/FooterHidingModifier.swift
```diff
@@ -41,27 +38,12 @@ private struct FooterHidingModifier: ViewModifier {
...
-                        )
-                        .frame(height: self.hide ? 0 : nil)
-                        .blur(radius: self.hide ? Self.blurRadius : 0)
-                }
+            content
+                .onSizeChange(.vertical) { self.height = $0 }
+                .opacity(self.hide ? 0 : 1)
+                .frame(height: self.hide ? 0 : nil, alignment: .top)
+                .clipped()
+                .transition(.move(edge: .bottom))
```
NachoSoto, line 46: This is so much simpler

RevenueCatUI/Templates/Template2View.swift
```diff
@@ -83,20 +86,35 @@ struct Template2View: TemplateViewType {
...
+            VStack {
+                self.packagesScrollHack
+
                 Spacer()
-            } else {
-                self.packages
-                    .onSizeChange(.vertical) { if $0 > 0 { self.containerHeight = $0 } }
-                    .hideFooterContent(self.configuration,
-                                       hide: !self.displayingAllPlans,
-                                       offset: self.containerHeight)
```
NachoSoto, line 94: We can remove this property now, right?

RevenueCatUI/Templates/Template2View.swift
```diff
@@ -83,20 +86,35 @@ struct Template2View: TemplateViewType {
...
             }
+            // Sets a max footer height so there is enough room for paywall content above
+            .frame(maxHeight: configuration.mode == .fullScreen ? nil : Constants.maxFooterHeight)
```
NachoSoto, line 95: Why not use `.infinite`? I think it accomplishes the same goal without hardcoding a magic number.

RevenueCatUI/Templates/Template2View.swift
```diff
@@ -83,20 +86,35 @@ struct Template2View: TemplateViewType {
...
         .frame(maxHeight: .infinity)
     }
 
+    /// Hack because scrollableIfNecessary() doesn't work on iOS 15 because of unknown reason
+    private var packagesScrollHack: some View {
+        Group {
+            if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
+                self.packages
+                    // Another hack because bounce animation temporarily clips top package a little bits
+                    .padding(.top, 5)
```
NachoSoto, line 108: :( let's look at a better way of doing this.

RevenueCatUI/Templates/Template2View.swift
```diff
@@ -83,20 +86,35 @@ struct Template2View: TemplateViewType {
...
+            .frame(maxHeight: configuration.mode == .fullScreen ? nil : Constants.maxFooterHeight)
+            .hideFooterContent(self.configuration,
+                               hide: !self.configuration.mode.shouldDisplayPackages && !self.displayingAllPlans)
         }
         .frame(maxHeight: .infinity)
     }
 
+    /// Hack because scrollableIfNecessary() doesn't work on iOS 15 because of unknown reason
+    private var packagesScrollHack: some View {
+        Group {
```
NachoSoto, line 104: The recommended way of doing this is:
```suggestion
    @ViewBuilder
    private var packagesScrollHack: some View {
```

RevenueCatUI/View+PresentPaywallFooter.swift
```diff
@@ -0,0 +1,117 @@
...
+    /// - Note: If loading the `CustomerInfo` fails (for example, if Internet is offline),
+    /// the paywall won't be displayed.
+    ///
+    /// ### Related Articles
+    /// [Documentation](https://rev.cat/paywalls)
+    public func paywallFooter(
+        condensed: Bool = false,
+        fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
+        purchaseCompleted: PurchaseCompletedHandler? = nil
+    ) -> some View {
```
NachoSoto, line 32: 👏🏻  I think this is great.

RevenueCatUI/View+PresentPaywallFooter.swift
```diff
@@ -0,0 +1,117 @@
...
+
+    /// Presents a ``PaywallFooterView`` at the bottom of a view with the given offering.
+    /// ```swift
+    /// var body: some View {
+    ///    YourPaywall()
+    ///      .paywallFooter(offering: offering)
+    /// }
+    /// ```
+    /// - Note: If loading the `CustomerInfo` fails (for example, if Internet is offline),
+    /// the paywall won't be displayed.
```
NachoSoto, line 50: Ditto

RevenueCatUI/View+PresentPaywallFooter.swift
```diff
@@ -0,0 +1,117 @@
...
+                introEligibility: introEligibility,
+                purchaseHandler: purchaseHandler
+            ))
+    }
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+struct PresentingPaywallFooterModifier: ViewModifier {
```
NachoSoto, line 92: This can be `private`.

Tests/RevenueCatUITests/PaywallFooterTests.swift
```diff
@@ -0,0 +1,51 @@
...
+@MainActor
+class PaywallFooterTests: TestCase {
+
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+
+        try AvailabilityChecks.iOS16APIAvailableOrSkipTest()
+    }
+
+    func testPresentWithPurchaseHandler() throws {
```
NachoSoto, line 26: Cool.
Let's add a snapshot test too.

NachoSoto (commented):
I haven't tested this, but assuming it hasn't broken any of the permutations (iOS / template / etc) this is awesome 👏🏻 
- Much better API
- Simpler implementation
- More discoverable
- More intuitive to use

One thing is that we can make `PaywallViewMode` `internal`, and "hide" the parameter from the `PaywallView` public constructors, right?

RevenueCatUI/PaywallFooterView.swift
```diff
@@ -0,0 +1,92 @@
...
+/// A SwiftUI view for displaying a `PaywallData` for an `Offering`.
+///
+/// ### Related Articles
+/// [Documentation](https://rev.cat/paywalls)
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
+@available(watchOS, unavailable, message: "RevenueCatUI does not support watchOS yet")
+@available(macOS, unavailable, message: "RevenueCatUI does not support macOS yet")
+@available(tvOS, unavailable, message: "RevenueCatUI does not support tvOS yet")
+@available(macCatalyst, unavailable, message: "RevenueCatUI does not support Catalyst yet")
+public struct PaywallFooterView: View {
```
NachoSoto, line 20: Now that I think about it, does this need to be `public`?

RevenueCatUI/PaywallFooterView.swift
```diff
@@ -0,0 +1,92 @@
...
+
+    private let mode: PaywallViewMode
+    private let fonts: PaywallFontProvider
+    private let introEligibility: TrialOrIntroEligibilityChecker?
+    private let purchaseHandler: PurchaseHandler?
+
+    @State
+    private var offering: Offering?
+    @State
+    private var error: NSError?
```
NachoSoto, line 30: This isn't used.

RevenueCatUI/PaywallFooterView.swift
```diff
@@ -0,0 +1,92 @@
...
+@available(tvOS, unavailable, message: "RevenueCatUI does not support tvOS yet")
+@available(macCatalyst, unavailable, message: "RevenueCatUI does not support Catalyst yet")
+public struct PaywallFooterView: View {
+
+    private let mode: PaywallViewMode
+    private let fonts: PaywallFontProvider
+    private let introEligibility: TrialOrIntroEligibilityChecker?
+    private let purchaseHandler: PurchaseHandler?
+
+    @State
```
NachoSoto, line 27: This shouldn't be `@State`.

RevenueCatUI/Templates/Template2View.swift
```diff
@@ -41,7 +41,10 @@ struct Template2View: TemplateViewType {
             Spacer()
 
             self.scrollableContent
-                .scrollableIfNecessary()
+                // Disabling in overlay mode here because of animation issues with ViewThatFits
+                // Scrolling is enabled in overlay mode in the packages
+                // Bonus is that CTA stays visible and doesn't scroll
+                .scrollableIfNecessary(enabled: configuration.mode == .fullScreen)
```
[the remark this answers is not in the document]
NachoSoto: Yeah I don't think we can do this, this probably breaks with larger dynamic types.

RevenueCatUI/Templates/Template2View.swift
```diff
@@ -83,20 +86,35 @@ struct Template2View: TemplateViewType {
...
-                self.packages
-                    .onSizeChange(.vertical) { if $0 > 0 { self.containerHeight = $0 } }
-                    .hideFooterContent(self.configuration,
-                                       hide: !self.displayingAllPlans,
-                                       offset: self.containerHeight)
             }
+            // Sets a max footer height so there is enough room for paywall content above
+            .frame(maxHeight: configuration.mode == .fullScreen ? nil : Constants.maxFooterHeight)
+            .hideFooterContent(self.configuration,
+                               hide: !self.configuration.mode.shouldDisplayPackages && !self.displayingAllPlans)
```
NachoSoto, line 97: Why this change?

RevenueCatUI/PaywallFooterView.swift
NachoSoto: Changed.

RevenueCatUI/PaywallFooterView.swift
NachoSoto: Changed.

RevenueCatUI/PaywallFooterView.swift
NachoSoto: Removed

RevenueCatUI/PaywallFooterView.swift
NachoSoto: Nah this makes sense, I think this should be configured from the SDK as it depends on how it's used in the app.

RevenueCatUI/Templates/Template2View.swift
NachoSoto: Removed.

RevenueCatUI/Templates/Template2View.swift
NachoSoto: Removed this. We can add a parameter for the maximum height if users request it.

RevenueCatUI/PaywallFooterView.swift
```diff
@@ -0,0 +1,89 @@
...
+/// A SwiftUI view for displaying a `PaywallData` for an `Offering`.
+///
+/// ### Related Articles
+/// [Documentation](https://rev.cat/paywalls)
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
+@available(watchOS, unavailable, message: "RevenueCatUI does not support watchOS yet")
+@available(macOS, unavailable, message: "RevenueCatUI does not support macOS yet")
+@available(tvOS, unavailable, message: "RevenueCatUI does not support tvOS yet")
+@available(macCatalyst, unavailable, message: "RevenueCatUI does not support Catalyst yet")
+internal struct PaywallFooterView: View {
```
NachoSoto, line 20: Made this `internal` now.

NachoSoto (approved):

PR #3052, opened 2023-08-22
`README`: added `visionOS` to list of supported platforms

NachoSoto:
I also added `Mac Catalyst`.


README.md
```diff
@@ -58,6 +58,7 @@ Or view our iOS sample apps:
 | tvOS | 11.0+ |
 | macOS | 10.13+ |
 | watchOS | 6.2+ |
+| visionOS | 1.0+ |
```
tonidero, line 61: 1.0 still hasn't launched right? 😛. But well, I guess it's good to be ready

tonidero (approved):

README.md
NachoSoto: True, 1.0 beta, and we support it 😇
I suppose I can change this to `1.0 beta X` but we're going to keep supporting this and we'll make sure it works in Xcode 15 RC.

PR #3053, opened 2023-08-22
`Paywalls`: internal documentation for implementing templates

NachoSoto:
This is a start of some basic documentation for some basic principles to follow when implementing new templates.


aboedo (commented):
looks great. This PR also updates deployment target to iOS 15, was that meant to be a separate thing?

PR #3064, opened 2023-08-23
`Paywalls`: removed unnecessary `PaywallFooterView`

NachoSoto:
Since this is no longer `public`, it was just duplicating all the API surface area of `PaywallView` in an unnecessary way. This became clear in #3046 when a lot of the same changes had to be made in both view types.


joshdholtz (approved):
This much nicer!

PR #3066, opened 2023-08-23
`Configuration`: log warning if attempting to use observer mode with StoreKit 2

NachoSoto:
See also https://github.com/RevenueCat/revenuecat-docs/pull/299.

Since #3032 developers using observer mode need to configure the SDK with the correct StoreKit 2 setting.

### DIFF
--- a/Sources/Misc/Deprecations.swift
+++ b/Sources/Misc/Deprecations.swift
@@ -400,7 +400,7 @@ extension CustomerInfo {
 
 public extension Configuration.Builder {
 
-    /// Set `usesStoreKit2IfAvailable`. If `true`, the SDK will use StoreKit 2 APIs internally. If disabled, it will use StoreKit 1 APIs instead.
+    /// Set `storeKit2Setting`. If `true`, the SDK will use StoreKit 2 APIs internally. If disabled, it will use StoreKit 1 APIs instead.
     /// - Parameter usesStoreKit2IfAvailable: enable StoreKit 2 on devices that support it.
     /// Defaults to  `false`.
     /// - Important: This configuration flag has been deprecated, and will be replaced by automatic remote configuration in the future.
--- a/Sources/Purchasing/Configuration.swift
+++ b/Sources/Purchasing/Configuration.swift
@@ -53,6 +53,7 @@ import Foundation
 
     private init(with builder: Builder) {
         Self.verify(apiKey: builder.apiKey)
+        Self.verify(observerMode: builder.observerMode, storeKit2Setting: builder.storeKit2Setting)
 
         self.apiKey = builder.apiKey
         self.appUserID = builder.appUserID
@@ -129,12 +130,14 @@ import Foundation
          * Set `observerMode`.
          * - Parameter observerMode: Set this to `true` if you have your own IAP implementation and want to use only
          * RevenueCat's backend. Default is `false`.
+         *
+         * - Warning: This assumes your IAP implementation uses StoreKit 1.
+         * Observer mode is not compatible with StoreKit 2.
          */
-        @objc public func with(observerMode: Bool) -> Builder {
+        @objc public func with(observerMode: Bool) -> Configuration.Builder {
             self.observerMode = observerMode
             return self
         }
-
         /**
          * Set `userDefaults`.
          * - Parameter userDefaults: Custom `UserDefaults` to use
@@ -286,6 +289,12 @@ extension Configuration {
         case .validApplePlatform: break
         case .legacy: Logger.debug(Strings.configure.legacyAPIKey)
         case .otherPlatforms: Logger.error(Strings.configure.invalidAPIKey)
+        }
+    }
+
+    fileprivate static func verify(observerMode: Bool, storeKit2Setting: StoreKit2Setting) {
+        if observerMode, storeKit2Setting.usesStoreKit2IfAvailable {
+            Logger.warn(Strings.configure.observer_mode_with_storekit2)
         }
     }
 
--- a/Tests/APITesters/SwiftAPITester/SwiftAPITester/ConfigurationAPI.swift
+++ b/Tests/APITesters/SwiftAPITester/SwiftAPITester/ConfigurationAPI.swift
@@ -14,7 +14,7 @@ func checkConfigurationAPI() {
         .with(apiKey: "")
         .with(appUserID: "")
         .with(appUserID: nil)
-        .with(observerMode: false)
+        .with(observerMode: true)
         .with(userDefaults: UserDefaults.standard)
         .with(dangerousSettings: DangerousSettings())
         .with(dangerousSettings: DangerousSettings(autoSyncPurchases: true))
@@ -29,4 +29,10 @@ func checkConfigurationAPI() {
             .with(entitlementVerificationMode: .informational)
             .build()
     }
+}
+
+@available(*, deprecated)
+func checkDeprecatedConfiguration(_ builder: Configuration.Builder) {
+    _ = builder
+        .with(usesStoreKit2IfAvailable: false)
 }
--- a/Sources/Logging/Strings/ConfigureStrings.swift
+++ b/Sources/Logging/Strings/ConfigureStrings.swift
@@ -33,6 +33,8 @@ enum ConfigureStrings {
 
     case observer_mode_enabled
 
+    case observer_mode_with_storekit2
+
     case response_verification_mode(Signing.ResponseVerificationMode)
 
     case delegate_set
@@ -98,6 +100,8 @@ extension ConfigureStrings: LogMessage {
             return "StoreKit 2 support enabled"
         case .observer_mode_enabled:
             return "Purchases is configured in observer mode"
+        case .observer_mode_with_storekit2:
+            return "Observer mode is not currently compatible with StoreKit 2"
         case let .response_verification_mode(mode):
             switch mode {
             case .disabled:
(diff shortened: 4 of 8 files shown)


Sources/Purchasing/Configuration.swift
```diff
@@ -129,9 +129,23 @@ import Foundation
...
+
+        /**
+         * Set `observerMode` with a corresponding `StoreKit 2` setting.
+         * - Parameter observerMode: Set this to `true` if you have your own IAP implementation and want to use only
+         * RevenueCat's backend. Default is `false`.
+         * - Parameter storeKit2: Set this to `true` if your own IAP implementation uses StoreKit 2. Default is `false`.
+         */
+        @objc public func with(observerMode: Bool, storeKit2: Bool) -> Builder {
             self.observerMode = observerMode
+            self.storeKit2Setting = .init(useStoreKit2IfAvailable: storeKit2)
```
tonidero, line 148: I'm wondering whether we really need to keep `useStoreKit2IfAvailable` deprecated... Considering we still need to use it here. Also, devs can potentially do `.with(observerMode: false, storeKit2: true)`, to bypass the deprecation and achieve the same result, so it's pretty much as if it wasn't deprecated...

Sources/Purchasing/Configuration.swift
```diff
@@ -129,9 +129,23 @@ import Foundation
          * Set `observerMode`.
          * - Parameter observerMode: Set this to `true` if you have your own IAP implementation and want to use only
          * RevenueCat's backend. Default is `false`.
+         *
+         * - Note: This assumes your IAP implementation uses StoreKit 1.
+         * If you use StoreKit 2, use ``with(observerMode:storeKit2:)`` instead.
```
tonidero, line 134: I wonder if we should deprecate this method and ask people to use the other one to make it more explicit...

tonidero (approved):
I'm wondering about whether we need to undeprecate the storekit 2 setting... But I understand linking observer mode to the storeKit2 setting, so I think this is probably better

Sources/Purchasing/Configuration.swift
vegaro: I think that would be a good way to notify developers that are currently using this function

vegaro (approved):

Sources/Purchasing/Configuration.swift
NachoSoto: Yeah I'll do that.

Sources/Purchasing/Configuration.swift
NachoSoto: > Also, devs can potentially do .with(observerMode: false, storeKit2: true)

Yeah maybe the method should have never been `.with(observerMode: Bool)` but instead `.enableObserverMode()`.

Sources/Purchasing/Configuration.swift
NachoSoto: Done.

Tests/APITesters/SwiftAPITester/SwiftAPITester/ConfigurationAPI.swift
```diff
@@ -30,3 +30,10 @@ func checkConfigurationAPI() {
             .build()
     }
 }
+
+@available(*, deprecated)
+func checkDeprecatedConfiguration(_ builder: Configuration.Builder) {
+    _ = builder
+        .with(usesStoreKit2IfAvailable: false)
```
NachoSoto, line 37: This was missing from the API tester.

Sources/Purchasing/Configuration.swift
aboedo: Feels a little odd to have to select "storekit 2 false" now, like, I'm not sure that tells you all that much about what's happening. 

How about 

```swift

enum StoreKitVersion {
    case StoreKit1, StoreKit2
}

.with(observerMode: Bool, storeKitVersion: StoreKitVersion) -> Builder {
```

(well, with changes for objc compatibility for the enum and stuff)

NachoSoto: Isn't that the same for `usesStoreKit2IfAvailable`? We made that a `Bool` too.
While we're at it, we could add an overload of that one to use this new enum. What do you think?

aboedo: > we could add an overload of that one to use this new enum
to the deprecated one? would the overload also be deprecated?

NachoSoto: Yeah, both. It would also have to need a new name, probably `.with(storeKit2Version:)`

aboedo: I wouldn't add a new API that's already deprecated when we add it. 
I was only suggesting that 

```swift
.with(observerMode: false, storeKitVersion: storeKit1)
```

might be slightly more expressive than
```swift
.with(observerMode: false, storeKit2: false)
```

Sources/Misc/Deprecations.swift
```diff
@@ -420,6 +420,19 @@ public extension Configuration.Builder {
         return self
     }
 
+    /**
+     * Set `observerMode`.
+     * - Parameter observerMode: Set this to `true` if you have your own IAP implementation and want to use only
+     * RevenueCat's backend. Default is `false`.
+     *
+     * - Note: This assumes your IAP implementation uses StoreKit 1.
```
aboedo, line 428:
```suggestion
     * - Warn: This assumes your IAP implementation uses StoreKit 1.
```

Can't remember if `Warn` is the exact keyword, but I'm shooting for the yellow warning sign

Sources/Purchasing/Configuration.swift
aboedo: That way the docstring can read something like 

```swift
* Set `observerMode` with a corresponding `StoreKit` implementation
* - Parameter storeKitVersion: Set this to the StoreKit implementation 
* your app uses for purchases. 
* I.e.: if your app uses StoreKit 1, set it to `.storeKit1`, 
* and if your app uses StoreKit 2, set it to `.storeKit2`. 
* Apps using StoreKit 1 use `SKPaymentQueue` for transactions, 
* whereas StoreKit 2 uses `StoreKit.Product.Purchase`. 
```

Added some more text in case folks don't know which is which. Apple removed many of the SK1 docs, so I think it's likely that folks will start getting confused about this over time as SK1 slowly fades away

aboedo (approved):

Sources/Purchasing/Configuration.swift
NachoSoto: I'd avoid adding a new enum + docs + Obj-C equivalent etc just for one API.
But I do think it makes sense to change the deprecated method (which arguably won't be once we fully support SK2) to use it as well.

Sources/Misc/Deprecations.swift
NachoSoto: Oh yeah. It's warning, I'll change it:
```suggestion
     * - Warning: This assumes your IAP implementation uses StoreKit 1.
```

NachoSoto:
Planning to ship this with the paywalls release since they both require a minor.

Sources/Purchasing/Configuration.swift
aboedo: > I'd avoid adding a new enum + docs + Obj-C equivalent etc just for one API.
But I do think it makes sense to change the deprecated method (which arguably won't be once we fully support SK2) to use it as well.

I missed this comment but that makes a lot of sense

Sources/Purchasing/Configuration.swift
aboedo: If a developer is _currently_ using this function, it's almost certain that they're using SK1, though, right? Like, it seems fairly unlikely that you: 
- have an existing SK1 implementation that you don't want to touch, but you want RC, so you add observer mode
- decide that you _do_ want to move into SK2, but still want to keep RC in observer mode

Right? 

So I'd vote to just keep a single method with a default to sk1

NachoSoto:
Updated this per our conversation:
- Not deprecated `.with(observerMode:)`
- Not deprecated `.with(observerMode:usesStoreKit2:)`
- Deprecated `.with(usesStoreKit2IfAvailable:)`

NachoSoto:
Changing this since we can't support observer mode + SK2.

tonidero (approved):

slavasemeniuk:
> Changing this since we can't support observer mode + SK2.

Hey, is there any insights on availability of SK 2 support for the observer mode?
Thanks

slavasemeniuk:
> > Changing this since we can't support observer mode + SK2.
> 
> Hey, is there any insights on availability of SK 2 support for the observer mode?
> Thanks

Also is there any way to just manually call something like ‘trackPurchase(transaction)’?

NachoSoto:
> Also is there any way to just manually call something like ‘trackPurchase(transaction)’?

@slavasemeniuk 
We are indeed looking into adding that exact API in our continuous effort to better support StoreKit 2 :)

Sources/Misc/Deprecations.swift
```diff
@@ -400,7 +400,7 @@ extension CustomerInfo {
 
 public extension Configuration.Builder {
 
-    /// Set `usesStoreKit2IfAvailable`. If `true`, the SDK will use StoreKit 2 APIs internally. If disabled, it will use StoreKit 1 APIs instead.
+    /// Set `storeKit2Setting`. If `true`, the SDK will use StoreKit 2 APIs internally. If disabled, it will use StoreKit 1 APIs instead.
```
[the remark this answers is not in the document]
aboedo: should we use double backticks to create a code reference then, so mismatches get caught at the docs generation step?

aboedo (commented):
do we plan on doing the syncObserverModePurchase thing? Is this an intermediate step?

NachoSoto:
> do we plan on doing the syncObserverModePurchase thing? Is this an intermediate step?

Yup that's the plan (@MarkVillacampa do you think you can take on that? I think you're pretty familiar with the setup now, and I'm happy to support you in the process :))

This is an intermediate step. Once the API is available we can remove the warning and update the documentation

PR #3067, opened 2023-08-23
`Purchases`: don't clear intro eligibility / purchased products cache on first launch

NachoSoto:
These caches are important, especially for `RevenueCatUI`.
Without this fix, launching the app was doing this:
- Pre-warming cache
- Pre-warming cache a second time (to be fixed by a separate PR)
- Clearing cache

Which meant that the cache wasn't really warm when launching paywalls.

To fix that, this only clears the cache after receiving an actual change.

I've also removed `CustomerInfoManager.sendCachedCustomerInfoIfAvailable` because it wasn't used.

### DIFF
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -19,8 +19,6 @@ class CustomerInfoManager {
 
     typealias CustomerInfoCompletion = @MainActor @Sendable (Result<CustomerInfo, BackendError>) -> Void
 
-    var lastSentCustomerInfo: CustomerInfo? { return self.data.value.lastSentCustomerInfo }
-
     private let offlineEntitlementsManager: OfflineEntitlementsManager
     private let operationDispatcher: OperationDispatcher
     private let backend: Backend
@@ -207,7 +205,6 @@ class CustomerInfoManager {
     func clearCustomerInfoCache(forAppUserID appUserID: String) {
         self.modifyData {
             $0.deviceCache.clearCustomerInfoCache(appUserID: appUserID)
-            $0.lastSentCustomerInfo = nil
         }
     }
 
@@ -250,6 +247,9 @@ class CustomerInfoManager {
             }
         }
     }
+
+    // Visible for tests
+    var lastSentCustomerInfo: CustomerInfo? { return self.data.value.lastSentCustomerInfo }
 
     private func removeObserver(with identifier: Int) {
         self.modifyData {
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -100,14 +100,6 @@ class CustomerInfoManager {
         }
     }
 
-    func sendCachedCustomerInfoIfAvailable(appUserID: String) {
-        guard let info = self.cachedCustomerInfo(appUserID: appUserID) else {
-            return
-        }
-
-        self.sendUpdateIfChanged(customerInfo: info)
-    }
-
     // swiftlint:disable:next function_body_length
     func customerInfo(
         appUserID: String,
@@ -219,6 +211,12 @@ class CustomerInfoManager {
         }
     }
 
+    func setLastSentCustomerInfo(_ info: CustomerInfo) {
+        self.modifyData {
+            $0.lastSentCustomerInfo = info
+        }
+    }
+
     @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
     var customerInfoStream: AsyncStream<CustomerInfo> {
         return AsyncStream(bufferingPolicy: .bufferingNewest(1)) { continuation in
@@ -226,15 +224,17 @@ class CustomerInfoManager {
                 continuation.yield(lastSentCustomerInfo)
             }
 
-            let disposable = self.monitorChanges { continuation.yield($0) }
+            let disposable = self.monitorChanges { _, new in continuation.yield(new) }
 
             continuation.onTermination = { @Sendable _ in disposable() }
         }
     }
 
+    typealias CustomerInfoChangeClosure = (_ old: CustomerInfo?, _ new: CustomerInfo) -> Void
+
     /// Allows monitoring changes to the active `CustomerInfo`.
     /// - Returns: closure that removes the created observation.
-    func monitorChanges(_ changes: @escaping (CustomerInfo) -> Void) -> () -> Void {
+    func monitorChanges(_ changes: @escaping CustomerInfoChangeClosure) -> () -> Void {
         self.modifyData {
             let lastIdentifier = $0.customerInfoObserversByIdentifier.keys
                 .sorted()
@@ -258,11 +258,12 @@ class CustomerInfoManager {
     }
 
     private func sendUpdateIfChanged(customerInfo: CustomerInfo) {
-        self.modifyData {
-            guard !$0.customerInfoObserversByIdentifier.isEmpty,
-                  $0.lastSentCustomerInfo != customerInfo else {
-                      return
-                  }
+        return self.modifyData {
+            let lastSentCustomerInfo = $0.lastSentCustomerInfo
+
+            guard !$0.customerInfoObserversByIdentifier.isEmpty, lastSentCustomerInfo != customerInfo else {
+                return
+            }
 
             if $0.lastSentCustomerInfo != nil {
                 Logger.debug(Strings.customerInfo.sending_updated_customerinfo_to_delegate)
@@ -276,7 +277,7 @@ class CustomerInfoManager {
             // this class' data. By making it async, the closure is invoked outside of the lock.
             self.operationDispatcher.dispatchAsyncOnMainThread { [observers = $0.customerInfoObserversByIdentifier] in
                 for closure in observers.values {
-                    closure(customerInfo)
+                    closure(lastSentCustomerInfo, customerInfo)
                 }
             }
         }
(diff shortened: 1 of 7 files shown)


Sources/Identity/CustomerInfoManager.swift
```diff
@@ -100,14 +98,6 @@ class CustomerInfoManager {
         }
     }
 
-    func sendCachedCustomerInfoIfAvailable(appUserID: String) {
```
NachoSoto, line 103: This wasn't being used by anything other than in tests.

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -19,8 +19,6 @@ class CustomerInfoManager {
 
     typealias CustomerInfoCompletion = @MainActor @Sendable (Result<CustomerInfo, BackendError>) -> Void
 
-    var lastSentCustomerInfo: CustomerInfo? { return self.data.value.lastSentCustomerInfo }
```
NachoSoto, line 22: This was only visible for tests so I moved it below.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1478,10 +1478,13 @@ internal extension Purchases {
 
 private extension Purchases {
 
-    func handleCustomerInfoChanged(_ customerInfo: CustomerInfo) {
-        self.trialOrIntroPriceEligibilityChecker.clearCache()
-        self.purchasedProductsFetcher?.clearCache()
-        self.delegate?.purchases?(self, receivedUpdated: customerInfo)
+    func handleCustomerInfoChanged(from old: CustomerInfo?, to new: CustomerInfo) {
+        if old != nil {
```
NachoSoto, line 1482: This is the core of this change. We no longer clear the caches for the first time we receive a customer info change.

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
```diff
@@ -196,51 +207,10 @@ class CustomerInfoManagerTests: BaseCustomerInfoManagerTests {
         expect(self.mockBackend.invokedGetSubscriberDataCount) == 1
     }
 
-    func testSendCachedCustomerInfoIfAvailableForAppUserIDSendsIfNeverSent() throws {
```
NachoSoto, line 199: Removed all these tests for the unused method.

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
```diff
@@ -448,15 +418,45 @@ class CustomerInfoManagerTests: BaseCustomerInfoManagerTests {
     func testCacheCustomerInfoSendsToDelegateIfChanged() {
         self.customerInfoManager.cache(customerInfo: self.mockCustomerInfo, appUserID: "myUser")
         expect(self.customerInfoManagerChangesCallCount).toEventually(equal(1))
-        expect(self.customerInfoManagerLastCustomerInfo) == self.mockCustomerInfo
+        expect(self.customerInfoManagerLastCustomerInfoChange) == (old: nil, new: self.mockCustomerInfo)
+    }
+
+    func testCacheCustomerInfoSendsMultipleUpdatesIfChange() throws {
```
NachoSoto, line 424: We didn't have this test.

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
```diff
@@ -448,15 +418,45 @@ class CustomerInfoManagerTests: BaseCustomerInfoManagerTests {
...
     func testCacheCustomerInfoSendsToDelegateWhenComputedOnDevice() {
         let info = self.mockCustomerInfo.copy(with: .verifiedOnDevice)
 
         self.customerInfoManager.cache(customerInfo: info, appUserID: "myUser")
         expect(self.customerInfoManagerChangesCallCount).toEventually(equal(1))
-        expect(self.customerInfoManagerLastCustomerInfo) == info
+        expect(self.customerInfoManagerLastCustomerInfoChange) == (old: nil, new: info)
+    }
+
+    func testCacheCustomerInfoSendsToDelegateAfterCachingComputedOnDevice() {
```
NachoSoto, line 451: The first implementation broke integration tests in a way that this test now captures.

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -215,7 +205,12 @@ class CustomerInfoManager {
     func clearCustomerInfoCache(forAppUserID appUserID: String) {
         self.modifyData {
             $0.deviceCache.clearCustomerInfoCache(appUserID: appUserID)
-            $0.lastSentCustomerInfo = nil
```
NachoSoto, line 218: This didn't make sense. It didn't matter because the property wasn't used by anything, but now this is used for sending the "old" value. Resetting this broke an integration test, which is what I captured in the new `testCacheCustomerInfoSendsToDelegateAfterCachingComputedOnDevice` test.

joshdholtz (approved):

Sources/Identity/CustomerInfoManager.swift
aboedo: interesting! I guess it's a holder from this: https://github.com/RevenueCat/purchases-ios/pull/1828/

aboedo (approved):
this reminds me of this old gem: https://linear.app/revenuecat/issue/SDK-2871/new-customerinfo-listener-with-diff, feels like this change would make that pretty much trivial?

Sources/Identity/CustomerInfoManager.swift
NachoSoto: Oh that's right 😇 this paves the way for that.

NachoSoto:
Oh that's right 😇 this paves the way for that.

rafaelnobrekz:
Hey @NachoSoto , is there even a way to clear these caches from the user side at all ? I'm having stale cache issues that persist even across a cold app start

NachoSoto:
There is not currently. What are you observing? Could you share some logs so we can take a look?

rafaelnobrekz:
We're hoping to add some trial introductory offers a few weeks from now, and that would go along with some UI updates on the paywall to showcase it to the users. But it takes a long time for the SDK to catch up with any updates we do on App Store Connect, it comes back from disk cache it seems:

```
2023-11-01 22:20:23.887923-0300 Kinzoo[92535:24609323] [offering] DEBUG: ℹ️ Skipping products request for these products because they were already cached: ["com.kinzoo.app.staging.club.yearly", "com.kinzoo.app.staging.club.monthly"]
2023-11-01 22:20:23.921121-0300 Kinzoo[92535:24609323] [customer] DEBUG: ℹ️ Local intro eligibility computed locally. Result: ["com.kinzoo.app.staging.club.monthly": IntroEligibilityStatus.eligible, "com.kinzoo.app.staging.club.yearly": IntroEligibilityStatus.noIntroOfferExists]
2023-11-01 22:20:23.921359-0300 Kinzoo[92535:24609323] [eligibility] DEBUG: ℹ️ Caching trial or intro eligibility for products: ["com.kinzoo.app.staging.club.yearly", "com.kinzoo.app.staging.club.monthly"]
```

Could you share with me what is the cache window duration if any? I believe it would have cleared if I killed the app, but apparently it does not

rafaelnobrekz:
The test I'm doing is actually adding/removing a free trial on App Store Connect, so product caching might also play a factor

NachoSoto:
>  it comes back from disk cache it seems:

Intro eligibility is never cached on disk though, only in memory (so re-computed when launching the app). The cache is cleared whenever the `CustomerInfo` changes (after a purchase, or when fetching an updated value from the server.

> But it takes a long time for the SDK to catch up with any updates we do on App Store Connect

Interesting. What types of changes are you making and what do you observe? Also can confirm if you're using SK1 or SK2?

rafaelnobrekz:
> Intro eligibility is never cached on disk though, only in memory (so re-computed when launching the app). The cache is cleared whenever the `CustomerInfo` changes (after a purchase, or when fetching an updated value from the server.
Cool, thanks for the assurance. I did notice the `clearCache()` call in `handleCustomerChanged`, would calling `Purchases.shared.getCustomerInfo(fetchPolicy: .fetchCurrent)` have any effect when the customer data does not change in an attempt to clear the cache through that means? Not that it would be any efficient. Anyway, the changes surviving restart suggest this is not my issue, possibly RC product cache (is it also memory-only?) or even StoreKit product caching.

> Interesting. What types of changes are you making and what do you observe? Also can confirm if you're using SK1 or SK2?
I am removing an existing Introductory Offer from App Store Connect (or adding one if it did not exist). I assume it wouldn't be a very common scenario, but it was caught by our QA team and we're trying to understand exactly what is the criteria for this to update or nudge it in the right way to refresh.
Regarding StoreKit, we're using the defaults for latest RC 4.29.0 so I assume it's SK2 for eligibility checking and SK1 for product listing and purchasing?

NachoSoto:
> Anyway, the changes surviving restart suggest this is not my issue, possibly RC product cache (is it also memory-only?) or even StoreKit product caching.

The SDK also doesn't have any on-disk product cache.

> Regarding StoreKit, we're using the defaults for latest RC 4.29.0 so I assume it's SK2 for eligibility checking and SK1 for product listing and purchasing?

The default is SK1 for everything. 

> we're trying to understand exactly what is the criteria for this to update or nudge it in the right way to refresh.

Our current guess is that there's some propagation lag by Apple when you update offers in ASC.
For testing this behavior I'd recommend using SK config files where you can simulate these changes more effectively than relying on ASC servers catching up 👍🏻

PR #3075, opened 2023-08-23
`Paywalls`: only dismiss `PaywallView` when explicitly presenting it with `.presentPaywallIfNeeded`

NachoSoto:
This makes more sense than assuming we can call the `\.dismiss` environment variable.


joshdholtz (approved):
Oh I like this!

PR #3081, opened 2023-08-24
`CustomerInfo`: add `VerificationResult` to `description`

NachoSoto:
I needed this to debug a test failure, so it can be useful.

### DIFF
--- a/Sources/Identity/CustomerInfo.swift
+++ b/Sources/Identity/CustomerInfo.swift
@@ -141,6 +141,14 @@ import Foundation
 
         let allEntitlementsDescription = self.entitlements.all.mapValues { $0.description }
 
+        let verificationResult: String
+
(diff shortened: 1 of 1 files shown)


Sources/Identity/CustomerInfo.swift
```diff
@@ -141,6 +141,14 @@ import Foundation
 
         let allEntitlementsDescription = self.entitlements.all.mapValues { $0.description }
 
+        let verificationResult: String
+
+        if #available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *) {
+            verificationResult = self.entitlements.verification.debugDescription
```
tonidero, line 147: Hmm I guess verificationResult is not technically part of the `customerInfo` but of `EntitlementInfos`... But I guess it's also related 1-1 currently, so I'm ok adding it to the description.

tonidero (approved):

PR #3086, opened 2023-08-24
`Integration Tests`: fixed race condition in flaky test

NachoSoto:
This test had a race condition because it was assuming that the prior transactions had already been posted.
Otherwise the only post receipt that happens is detected as `.queue`.

### DIFF
--- a/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
@@ -513,12 +513,15 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
 
         // 5. Re-open app
         await self.resetSingleton()
-        self.logger.clearMessages()
 
-        // 6. Purchase again
+        // 6. Wait for pending transactions to be posted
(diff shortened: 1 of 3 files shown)


Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -513,12 +513,15 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
 
         // 5. Re-open app
         await self.resetSingleton()
-        self.logger.clearMessages()
 
-        // 6. Purchase again
+        // 6. Wait for pending transactions to be posted
+        try await self.waitUntilUntilNoUnfinishedTransactions()
```
aboedo, line 518: typo UntilUntil
```suggestion
        try await self.waitUntilNoUnfinishedTransactions()
```

aboedo (approved):

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
NachoSoto: Oops thanks.

PR #3088, opened 2023-08-25
`Paywalls`: temporarily disable `PaywallTemplate.template4`

NachoSoto:
It's not available in the dashboard so we don't want the first SDK releases to support it yet.

I was going to remove the `enum` `case` but that required commenting out a lot of extra code.
This is a much simpler one-line change.


joshdholtz (approved):
🥷

Issue #3048, opened 2023-08-19
`Purchases.logLevel = .verbose` does not affect levels available in Console.app


codeactual:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

1. Environment
   1. Platform: iOS (iPad)
   2. SDK version: 4.25.6
   3. StoreKit version:
      - [x] StoreKit 1
      - [ ] StoreKit 2 (_enabled with `usesStoreKit2IfAvailable(true)`_)
   4. OS version: 16.4 (simulator, iPad 10th) / 16.6 (iPad 9th)
   5. Xcode version: 14.3.1
   6. Device and/or simulator:
      - [x] Device
      - [x] Simulator
   7. Environment:
      - [x] Sandbox
      - [x] TestFight
      - [x] Production
   8. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue. **Complete logs with `Purchases.logLevel = .verbose` will help us debug this issue.**

```
WARN: 🍎‼️ Unable to load receipt, ensure you are logged in to a valid Apple account. <snip>
WARN: 🍎‼️ App running on sandbox without a receipt file. Unable to determine into eligibility unless you've purchased before and there is a receipt available.
```

3. Steps to reproduce, with a description of expected vs. actual behavior

Expected:

With `Purchases.logLevel = .verbose`, I expected to see logs with levels up to that maximum in `Console.app`.

Actual:

`DEBUG` and `VERBOSE` logs that appear in the xcode console do not also appear in `Console.app`.

What I was trying to do:

Investigate a defect observed in TestFlight and production using `Console.app` to see the logs of the physical device in developer mode, but could only see a limited set of RC SDK logs (error/warn level).

Comments:

My simplest repro case (for this issue, not the defect I was investigating earlier) is running a debug build in the simulator and just looking at what logs appear when the app starts and just shows the product offerings on a paywall (no UI interaction), then comparing what's shown in `Console.app` versus xcode console.

[...]

PR #3108, opened 2023-08-28
`Integration Tests`: workaround to not lose debug logs

NachoSoto:
`OSLogType.debug` logs are purged and not saved:
> The system only captures debug-level messages in memory when you enable debug logging through a configuration change, and purges them in accordance with the configuration’s persistence setting.

I haven't been able to find a way to override this behavior. This means than when reading the logs on CI or when downloading the `xcresult` archive, we can't see these, making it really hard to debug failures.

See also #3048.

### DIFF
--- a/Sources/FoundationExtensions/ProcessInfo+Extensions.swift
+++ b/Sources/FoundationExtensions/ProcessInfo+Extensions.swift
@@ -49,7 +49,9 @@ extension ProcessInfo {
     }
 
     static var mockAdServicesToken: String? {
-        return self[.RCMockAdServicesToken]?.notEmptyOrWhitespaces
+        guard let token = self[.RCMockAdServicesToken], !token.isEmpty else { return nil }
+
+        return token
     }
 
 }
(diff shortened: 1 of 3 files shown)


Sources/FoundationExtensions/ProcessInfo+Extensions.swift
```diff
@@ -49,7 +49,9 @@ extension ProcessInfo {
     }
 
     static var mockAdServicesToken: String? {
-        return self[.RCMockAdServicesToken]?.notEmptyOrWhitespaces
+        guard let token = self[.RCMockAdServicesToken], !token.isEmpty else { return nil }
+
+        return token
```
NachoSoto, line 54: This file is now in `ReceiptParser`, which doesn't have `Strings+Extensions`.

Sources/LocalReceiptParsing/Helpers/LoggerType.swift
```diff
@@ -177,8 +177,17 @@ private extension LogLevel {
 
     var logType: OSLogType {
         switch self {
-        case .verbose: return .debug
-        case .debug: return .debug
+        case .verbose, .debug:
```
NachoSoto, line 180: Hold on I should cache this so we won't evaluate it every time.

tonidero (approved):

PR #3115, opened 2023-08-28
`Paywalls`: reference test snapshots from submodule

NachoSoto:
See https://github.com/RevenueCat/purchases-ios-snapshots

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -80,7 +80,7 @@ let package = Package(
                         "Nimble",
                         .product(name: "SnapshotTesting", package: "swift-snapshot-testing")
                     ],
-                    exclude: ["Templates/__Snapshots__"],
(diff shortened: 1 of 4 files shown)


joshdholtz (approved):
I think this should be good ☺️🤷‍♂️

Package.swift
```diff
@@ -80,7 +80,7 @@ let package = Package(
                         "Nimble",
                         .product(name: "SnapshotTesting", package: "swift-snapshot-testing")
                     ],
-                    exclude: ["Templates/__Snapshots__"],
+                    exclude: ["Templates/__Snapshots__", "Data/__Snapshots__", "TestPlans"],
```
NachoSoto, line 83: I ran `swift build` and got some warnings, so I realized these were missing from here.

PR #3118, opened 2023-08-29
`CI`: enable `use_system_scm`

NachoSoto:
See https://stackoverflow.com/questions/47842479/how-to-use-swift-package-manager-with-private-repos


NachoSoto:
More context on why I'm doing this when I validate the solution.

PR #3123, opened 2023-08-30
`PurchasesOrchestrator`: return early if receipt has no transactions when checking for promo offers

MarkVillacampa:
Avoids an extra request to the /offers endpoint in the common case where receipt has no trasactions.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -226,7 +226,7 @@ final class PurchasesOrchestrator {
     @available(iOS 12.2, macOS 10.14.4, watchOS 6.2, macCatalyst 13.0, tvOS 12.2, *)
     func promotionalOffer(forProductDiscount productDiscount: StoreProductDiscountType,
                           product: StoreProductType,
-                          completion: @escaping (Result<PromotionalOffer, PurchasesError>) -> Void) {
+                          completion: @escaping @Sendable (Result<PromotionalOffer, PurchasesError>) -> Void) {
         guard let discountIdentifier = productDiscount.offerIdentifier else {
             completion(.failure(ErrorUtils.productDiscountMissingIdentifierError()))
             return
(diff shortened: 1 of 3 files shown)


Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -426,6 +427,32 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
...
+          _ = try await Async.call { completion in
+              self.orchestrator.promotionalOffer(forProductDiscount: storeProductDiscount,
+                                                 product: StoreProduct(sk1Product: product),
+                                                 completion: completion)
+          }
+      } catch {
+          expect(error).to(matchError(ErrorCode.ineligibleError))
+      }
+
+      expect(self.offerings.invokedPostOffer) == false
```
NachoSoto, line 453: 👏🏻

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -426,6 +427,32 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
         expect(self.offerings.invokedPostOffer) == false
     }
 
+    func testGetPromotionalOfferFailsWithIneligibleIfReceiptHasNoTransactions() async throws {
```
aboedo, line 430: let's add another explicit test for the opposite case (doesn't fail if there are transactions). It's already indirectly tested, but having a direct one might help ensure that if other tests are refactored in some way, this one would still fail with a clear error

NachoSoto (commented):
This is perfect 👍🏻 

We already have `testGetPromotionalOfferWithNoPurchasesReturnsIneligible` in `StoreKitIntegrationTests`.
I would update that to verify that no requests are made:
```swift
self.logger.verifyMessageWasNotLogged("API request started")
```

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -249,6 +249,13 @@ final class PurchasesOrchestrator {
                 return
             }
 
+            if !self.receiptParser.receiptHasTransactions(receiptData: receiptData) {
```
NachoSoto, line 252: 😇 our integration tests caught that this is now running in the main thread, which is bad for performance
```
0   libswiftCore.dylib            	       0x10fe98e81 _assertionFailure(_:_:file:line:flags:) + 353
1   RevenueCat                    	       0x1402cee1a static PurchasesReceiptParser.ensureRunningOutsideOfMainThread() + 842 (PurchasesReceiptParser.swift:127)
2   RevenueCat                    	       0x1402ce459 PurchasesReceiptParser.parse(from:) + 217 (PurchasesReceiptParser.swift:43)
```

NachoSoto (approved):

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -543,6 +543,7 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
         } catch {
             expect(error).to(matchError(ErrorCode.ineligibleError))
         }
+        self.logger.verifyMessageWasNotLogged("API request started")
```
NachoSoto, line 546: You might have to clear the logs before calling the method because this contains the entire test lifetime, including `setUp`

NachoSoto (approved):

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
MarkVillacampa: Fixed!

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
MarkVillacampa: Good catch tests! :D

Issue #3124, opened 2023-08-30
Error adding beta package


pedromamare:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
I am following the following step by step: https://www.revenuecat.com/docs/paywalls

However, when adding the package, it returns the following error: https://share.cleanshot.com/CW0mmJwl

![logo](https://media.cleanshot.cloud/media/15578/qYpFFNB41klrGcHu86ygw6YcpDqg5m5FntUsZBrt.jpeg?Expires=1693434653&Signature=Nf~x4ckb8OBqX-66aLLCkrBVCSuhl-zG-IaektESvlRBMwIIOU~KBP7qSthqITRqNBsJIZs0rpPPupPeEKZZg8Wq7iec2yGeyqE5f1KRXW3O~ZXDhCs4Ow5DsCXS4cpc4b4MxKRn5cYZzixqdzkVU6VJO9wyj1aGaXTkr2LYlmTt4QwyBQchcKfJ3cNa9KwlneSmaHXH45v-VY9dUKZfjBX05WMJv7-jqmqxXNNZ7R8rdhlD9zke9D4ibA-LdIR5Odm9trH92xm6SH5g2u2eXCPv4AhjFUsx24ZkfEImc2UN3c~HeHbScKbfpsazZjmuCDDUcLz14B3FDNzFrMJUvg__&Key-Pair-Id=K269JMAT9ZF4GZ "logo")

1. Environment
   1. Platform: SwiftUI
   2. SDK version: 4.26.0-beta.1
   3. StoreKit version:
      - [] StoreKit 1
      - [x] StoreKit 2 (_enabled with `usesStoreKit2IfAvailable(true)`_)
   4. OS version: 16
   5. Xcode version: 14.3.1
   6. Device and/or simulator:
      - [X] Device
      - [X] Simulator
   7. Environment:
      - [X] Sandbox
      - [ ] TestFight
      - [ ] Production
   8. How widespread is the issue. Percentage of devices affected.
[...]

PR #3116, opened 2023-08-28
`Paywalls`: update snapshot generation with new separate git repo

NachoSoto:
Follow up to #3115.

Note that this won't update the commit that the branch is pointing to. But one all PRs are merged, one can easily update that one commit in your local PR.

Example: https://github.com/RevenueCat/purchases-ios-snapshots/pull/4/files

PR #3125, opened 2023-08-30
`Paywalls`: replaced submodule with `gitignore`d reference

NachoSoto:
The `submodule` approach had 2 issues:
- Inconsistent results if users don't have a valid SSH key (#3124)
- `SPM` actually clones the submodule, which users don't need

This new approach only fetches the snapshots through a new `fastlane fetch_snapshots` lane.

#3116 will be updated to support this.


fastlane/Fastfile
```diff
@@ -757,6 +760,38 @@ platform :ios do
...
+
+        sh "git", "clone", "git@github.com:RevenueCat/#{snapshots_repo_name}.git"
+      end
+
+      commit = current_snapshots_commit
+      UI.message("Updating snapshots repo to #{commit}")
+      Dir.chdir(snapshots_repo_name) do
+        sh "git", "fetch", "origin"
+        sh "git", "checkout", commit
+      end
```
joshdholtz, line 777: Do we need to check if the repo is dirty at error out before this? 🤔 

Or do we need to do a hard reset so its clean?

joshdholtz (approved):
One question about if we need to check if its dirty before updating 🤷‍♂️  But not a blocker! We can keep iterating 😊

fastlane/Fastfile
NachoSoto: Good point that would make it more resilient.

NachoSoto: Actually I'm gonna leave it like this. If it's dirty then you want to handle that error so we don't make you lose whatever local changes you have.

PR #3128, opened 2023-08-31
[WIP] `Paywalls`: `macOS` support

NachoSoto:
Some initial work to support it, but still a fair amount of work left.


Jeehut:
@NachoSoto I just wanted to see how this current state renders on my macOS-only app, but the branch doesn't seem to even download (or it takes very long). I'd be willing to beta-test this in a new app I'm about to release even if it's not 100% ready yet. Please let me know if it's in a "kinda" or "probably" working state.

I will have to build my own paywall for now, but I'm really looking forward to this. You can DM me on X to get in touch.

PR #3131, opened 2023-08-31
`Paywalls`: new `{{ sub_relative_discount }}` variable

NachoSoto:


RevenueCatUI/Resources/en.lproj/Localizable.strings
```diff
@@ -14,3 +14,4 @@
 "Monthly" = "Monthly";
 "Weekly" = "Weekly";
 "Lifetime" = "Lifetime";
+"%d%% off" = "%d%% off";
```
NachoSoto, line 17: The only thing missing in this PR: we need this for the other languages.

Tests/RevenueCatUITests/Data/PackageVariablesTests.swift
```diff
@@ -107,6 +107,16 @@ class PackageVariablesTests: TestCase {
...
     }
 
+    func testEnglishRelativeDiscount() {
+        expect(TestData.weeklyPackage.localizedRelativeDiscount(nil, Self.english)).to(beNil())
+        expect(TestData.weeklyPackage.localizedRelativeDiscount(0.372, Self.english)) == "37% off"
+    }
+
+    func testSpanishRelativeDiscount() {
+        expect(TestData.weeklyPackage.localizedRelativeDiscount(nil, Self.spanish)).to(beNil())
+        expect(TestData.weeklyPackage.localizedRelativeDiscount(0.372, Self.english)) == "ahorra 37%"
```
NachoSoto, line 117: This is failing right now.

joshdholtz (commented):
Looks good! One question (which I may have missed) is... how does this work if there is also a lifetime in the offerings? It looks like it may return an empty string because no monthly price but just wanted to make sure 😇

NachoSoto:
Yup! I added an explicit test though: 9703ae134e412d679204ae8697bf529e26a34f70

joshdholtz (approved):

PR #3133, opened 2023-09-01
`Paywalls`: `{{ price_per_period }}` now takes `SubscriptionPeriod.value` into account

NachoSoto:
This was incorrect for `PackageType.threeMonth` and `PackageType.sixMonth`.

image

### DIFF
--- a/Sources/Purchasing/StoreKitAbstractions/Test Data/TestStoreProduct.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/Test Data/TestStoreProduct.swift
@@ -120,3 +120,8 @@ extension TestStoreProduct {
     }
 
 }
+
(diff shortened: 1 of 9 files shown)


Sources/Purchasing/StoreKitAbstractions/Test Data/TestStoreProduct.swift
```diff
@@ -59,6 +59,7 @@ public struct TestStoreProduct {
     public var isFamilyShareable: Bool
     public var introductoryDiscount: StoreProductDiscount?
     public var discounts: [StoreProductDiscount]
+    public var locale: Locale
```
NachoSoto, line 62: This is a new optional parameter too.

aboedo (approved):

PR #3141, opened 2023-09-05
`Paywalls`: fixed error view being displayed on release builds

NachoSoto:
Thanks @aboedo for catching this.

I've also improved the API to avoid using `AnyView`.


RevenueCatUI/Views/DebugErrorView.swift
```diff
@@ -55,10 +55,14 @@ struct DebugErrorView: View {
                 #endif
 
             case let .replacement(view):
+                #if DEBUG
```
NachoSoto, line 58: This is the core of the fix.

aboedo (approved):

PR #3150, opened 2023-09-06
`PaywallData`: added `@NonEmptyString` to `subtitle` and `offerName`

NachoSoto:
Thanks to @vegaro for pointing this out. It's not required, but it helps to ensure that these properties don't end up with empty strings.

### DIFF
--- a/RevenueCatUI/Data/Strings.swift
+++ b/RevenueCatUI/Data/Strings.swift
@@ -14,7 +14,7 @@
 import Foundation
 import RevenueCat
 
-// swiftlint:disable variable_name
+// swiftlint:disable identifier_name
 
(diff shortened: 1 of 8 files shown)


RevenueCatUI/Data/Strings.swift
```diff
@@ -14,7 +14,7 @@
 import Foundation
 import RevenueCat
 
-// swiftlint:disable variable_name
+// swiftlint:disable identifier_name
```
NachoSoto, line 17: > 'variable_name' has been renamed to 'identifier_name' and will be completely removed in a future release.

I didn't want to create a PR just for this.

vegaro (approved):

PR #3157, opened 2023-09-06
`Paywalls`: created `PaywallEventStore`

NachoSoto:
### Changes:

- New `PaywallEventStore` using `FileHandler` (#2673)
- `MockFileHandler`
- `PaywallStoredEvent`: this contains the necessary information to store an event: `PaywallEvent` + user ID
- Created `PaywallEventSerializer`

### DIFF
--- a/Sources/Diagnostics/FileHandler.swift
+++ b/Sources/Diagnostics/FileHandler.swift
@@ -15,7 +15,29 @@ import Foundation
 
 /// A wrapper that allows basic operations on a file, synchronized as an `actor`.
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
-actor FileHandler {
+protocol FileHandlerType: Sendable {
+
+    /// - Note: this loads the entire file in memory
+    /// For newer versions, consider using `readLines` instead.
+    func readFile() async throws -> Data
+
+    /// Returns an async sequence for every line in the file
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    func readLines() async throws -> AsyncLineSequence<FileHandle.AsyncBytes>
+
+    /// Adds a line at the end of the file
+    func append(line: String) async
+
+    /// Removes the contents of the file
+    func emptyFile() async throws
+
+    /// Deletes the first N lines from the file, without loading the entire file in memory.
+    func removeFirstLines(_ count: Int) async throws
+
+}
+
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+actor FileHandler: FileHandlerType {
 
     private var fileHandle: FileHandle
 
--- a/Sources/Diagnostics/FileHandler.swift
+++ b/Sources/Diagnostics/FileHandler.swift
@@ -17,10 +17,6 @@ import Foundation
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
 protocol FileHandlerType: Sendable {
 
-    /// - Note: this loads the entire file in memory
-    /// For newer versions, consider using `readLines` instead.
-    func readFile() async throws -> Data
-
     /// Returns an async sequence for every line in the file
     @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
     func readLines() async throws -> AsyncLineSequence<FileHandle.AsyncBytes>
--- a/Sources/Paywalls/Events/PaywallEventStore.swift
+++ b/Sources/Paywalls/Events/PaywallEventStore.swift
@@ -62,6 +62,8 @@ internal actor PaywallEventStore: PaywallEventStoreType {
         }
     }
 
+    // - Note: If removing these `count` events fails, it will attempt to
+    // remove the entire file. This ensures that the same events again aren't sent again.
     func clear(_ count: Int) async {
         assert(count > 0, "Invalid count: \(count)")
 
@@ -70,8 +72,6 @@ internal actor PaywallEventStore: PaywallEventStoreType {
         } catch {
             Logger.error(PaywallEventStoreStrings.error_removing_first_lines(count: count, error))
 
-            // If removing these `count` events fails, try removing the entire file.
-            // This ensures that we don't try to send the same events again.
             do {
                 try await self.handler.emptyFile()
             } catch {
--- a/Sources/Paywalls/Events/PaywallEventStore.swift
+++ b/Sources/Paywalls/Events/PaywallEventStore.swift
@@ -1,0 +1,162 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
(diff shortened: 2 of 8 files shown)


Sources/Paywalls/Events/PaywallEventStore.swift
```diff
@@ -0,0 +1,162 @@
...
+            .appendingPathComponent("revenuecat")
+            .appendingPathComponent("paywall_event_store")
+
+        return try .init(handler: FileHandler(url))
+    }
+
+    private static var documentsDirectory: URL {
+        get throws {
+            if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
+                return URL.documentsDirectory
```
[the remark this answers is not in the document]
NachoSoto: I know seriously 🙈
Since we need the fallback we could not have this, but this API is so much more reliable...

Sources/Paywalls/Events/PaywallEventStore.swift
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    func fetch(_ count: Int) async -> [PaywallStoredEvent] {
+        assert(count > 0, "Invalid count: \(count)")
+
+        do {
+            return try await self.handler.readLines()
+                .prefix(count)
+                .compactMap { try? PaywallEventSerializer.decode($0) }
```
NachoSoto, line 57: This lazily decodes each event line by line without creating copies of everything in memory all at once :)

Sources/Diagnostics/FileHandler.swift
```diff
@@ -15,7 +15,29 @@ import Foundation
...
+
+    /// Adds a line at the end of the file
+    func append(line: String) async
+
+    /// Removes the contents of the file
+    func emptyFile() async throws
+
+    /// Deletes the first N lines from the file, without loading the entire file in memory.
+    func removeFirstLines(_ count: Int) async throws
+
```
aboedo, line 36: protocols + mock implementations is what we should always have done, I'm glad we're moving away from inherit / replace for these. Much easier to isolate behavior

Sources/Diagnostics/FileHandler.swift
```diff
@@ -15,7 +15,29 @@ import Foundation
...
+}
+
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+actor FileHandler: FileHandlerType {
```
aboedo, line 40: Can't comment on the code that's not modified in the PR, but it looks great. The only way I can think of for improving it would be using a circular buffer for removing lines instead of an aux file to avoid extra disk reads

Sources/Paywalls/Events/PaywallEventStore.swift
```diff
@@ -0,0 +1,162 @@
...
+    func clear(_ count: Int) async {
+        assert(count > 0, "Invalid count: \(count)")
+
+        do {
+            try await self.handler.removeFirstLines(count)
+        } catch {
+            Logger.error(PaywallEventStoreStrings.error_removing_first_lines(count: count, error))
+
+            // If removing these `count` events fails, try removing the entire file.
+            // This ensures that we don't try to send the same events again.
```
aboedo, line 74: maybe this should be a docstring for the method instead? feels like a pretty important detail

Sources/Diagnostics/FileHandler.swift
aboedo: which maybe we can revisit one day if we start having a lot of events

aboedo (approved):

Sources/Diagnostics/FileHandler.swift
```diff
@@ -15,7 +15,29 @@ import Foundation
 
 /// A wrapper that allows basic operations on a file, synchronized as an `actor`.
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
-actor FileHandler {
+protocol FileHandlerType: Sendable {
+
+    /// - Note: this loads the entire file in memory
+    /// For newer versions, consider using `readLines` instead.
+    func readFile() async throws -> Data
```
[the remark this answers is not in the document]
NachoSoto: Oh yeah thanks. I did this on the initial implementation when I wasn't sure if I was gonna be able to use the stream method (and yeah not so relevant on the protocol).

I'll delete the method now.

Sources/Paywalls/Events/PaywallEventStore.swift
```diff
@@ -0,0 +1,162 @@
...
+
+    /// Removes the first `count` events from the store.
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    func clear(_ count: Int) async
+
+}
+
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+internal actor PaywallEventStore: PaywallEventStoreType {
+
```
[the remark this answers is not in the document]
NachoSoto: We have to do it manually creating a separate `Task`. I could put it at the top level (`PaywallEventsManager`), but I wonder if internally the other actors will override that priority 🤔

Sources/Paywalls/Events/PaywallEventStore.swift
```diff
@@ -0,0 +1,162 @@
...
+    func store(_ storedEvent: PaywallStoredEvent) async {
+        do {
+            Logger.verbose(PaywallEventStoreStrings.storing_event(storedEvent.event))
+
+            await self.handler.append(line: try PaywallEventSerializer.encode(storedEvent))
+        } catch {
+            Logger.error(PaywallEventStoreStrings.error_storing_event(error))
+        }
+    }
```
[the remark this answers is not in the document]
NachoSoto: This uses `FileHandle.write(contentsOf:)` (the Foundation type). 
I obviously haven't done a lot of performance testing yet, but based on the docs it seems like it doesn't immediately write to disk.

See also:
```
func synchronize()
Causes all in-memory data and attributes of the file represented by the file handle to write to permanent storage.
Programs that require the file to always be in a known state should call this method. An invocation of this method doesn’t return until memory is flushed.
```

We don't call it, so I think it's safe to assume it gets put in a buffer first.

Sources/Diagnostics/FileHandler.swift
NachoSoto: Totally. And these being actors + async made mocking these and creating fake in-memory implementations so easy to work with.

Sources/Paywalls/Events/PaywallEventStore.swift
NachoSoto: 👍🏻

Sources/Diagnostics/FileHandler.swift
NachoSoto: The way removing lines is implemented is sort of like a circular buffer though (it reads and writes in chunks to prevent the entire file in memory all at once).

Sources/Diagnostics/FileHandler.swift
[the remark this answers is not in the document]
NachoSoto: Actually I'm going to leave this on `FileHandler` (just not the protocol) because it makes testing it a lot simpler.

Sources/Paywalls/Events/PaywallEventStore.swift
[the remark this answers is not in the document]
aboedo: oh awesome!

PR #3158, opened 2023-09-06
`Paywalls`: implemented `PostPaywallEventsOperation`

NachoSoto:
This adds the operation + `PaywallEventsRequest` and associated tests. Also added method to `InternalAPI` for posting events.

### DIFF
--- a/Sources/Networking/InternalAPI.swift
+++ b/Sources/Networking/InternalAPI.swift
@@ -13,29 +13,56 @@
 
 import Foundation
 
-final class InternalAPI {
+class InternalAPI {
 
     typealias ResponseHandler = (BackendError?) -> Void
 
     private let backendConfig: BackendConfiguration
-    private let callbackCache: CallbackCache<HealthOperation.Callback>
+    private let healthCallbackCache: CallbackCache<HealthOperation.Callback>
 
     init(backendConfig: BackendConfiguration) {
         self.backendConfig = backendConfig
-        self.callbackCache = .init()
+        self.healthCallbackCache = .init()
     }
 
     func healthRequest(signatureVerification: Bool, completion: @escaping ResponseHandler) {
         let factory = HealthOperation.createFactory(httpClient: self.backendConfig.httpClient,
-                                                    callbackCache: self.callbackCache,
+                                                    callbackCache: self.healthCallbackCache,
                                                     signatureVerification: signatureVerification)
 
         let callback = HealthOperation.Callback(cacheKey: factory.cacheKey, completion: completion)
-        let cacheStatus = self.callbackCache.add(callback)
+        let cacheStatus = self.healthCallbackCache.add(callback)
 
         self.backendConfig.addCacheableOperation(with: factory,
                                                  delay: .none,
                                                  cacheStatus: cacheStatus)
+    }
+
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+    func postPaywallEvents(events: [PaywallStoredEvent], completion: @escaping ResponseHandler) {
+        guard !events.isEmpty else {
+            self.backendConfig.operationDispatcher.dispatchOnMainThread {
+                completion(nil)
+            }
+            return
+        }
+
+        let operation = PostPaywallEventsOperation(configuration: .init(httpClient: self.backendConfig.httpClient),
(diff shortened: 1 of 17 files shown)


Sources/Paywalls/Events/Networking/PaywallHTTPRequestPath.swift
```diff
@@ -0,0 +1,55 @@
...
+//
+//      https://opensource.org/licenses/MIT
+//
+//  PaywallHTTPRequestPath.swift
+//
+//  Created by Nacho Soto on 9/5/23.
+
+import Foundation
+
+extension HTTPRequest.PaywallPath: HTTPRequestPath {
```
NachoSoto, line 16: This is thanks to the refactor in #2986

Sources/Paywalls/Events/Networking/PaywallHTTPRequestPath.swift
```diff
@@ -0,0 +1,55 @@
...
+
+    // TODO: TBD
+    static let serverHostURL = URL(string: "https://api-paywalls.revenuecat.com")!
```
NachoSoto, line 19: @joshdholtz let me know when we know the final domain :)

joshdholtz: It will _probably_ be this but will confirm once its actually created 😛

joshdholtz (approved):
Looks good! 👍

Sources/Paywalls/Events/Networking/PaywallHTTPRequestPath.swift
```diff
@@ -0,0 +1,55 @@
...
+        switch self {
+        case .postEvents:
+            return false
+        }
+    }
+
+    var pathComponent: String {
+        switch self {
+        case .postEvents:
+            return "paywall_events"
```
tonidero, line 52: I was wondering whether we wanted to make this endpoint more generic in case we wanted to track other types of events in the future? CC @joshdholtz

NachoSoto: Good question. In that case we might want to make the domain something like `events.revenuecat.com` too

Sources/Networking/InternalAPI.swift
```diff
@@ -13,29 +13,56 @@
...
     }
 
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+    func postPaywallEvents(events: [PaywallStoredEvent], completion: @escaping ResponseHandler) {
+        guard !events.isEmpty else {
+            self.backendConfig.operationDispatcher.dispatchOnMainThread {
+                completion(nil)
+            }
+            return
+        }
```
aboedo, line 48: why do we need main thread for _anything_ in paywall events?

Sources/Paywalls/Events/Networking/PaywallHTTPRequestPath.swift
```diff
@@ -0,0 +1,55 @@
...
+
+extension HTTPRequest.PaywallPath: HTTPRequestPath {
+
+    static let serverHostURL = URL(string: "https://api-paywalls.revenuecat.com")!
+
+    var authenticated: Bool {
+        switch self {
+        case .postEvents:
+            return true
+        }
```
aboedo, line 24: I'm sure this makes sense, but why do all of these do switch self case .postEvents return hardcoded value? why not just return the harcoded value?

Sources/Networking/InternalAPI.swift
NachoSoto: Good point. I did that out of habit from other operations, but `ResponseHandler` isn't `@MainActor` anyway.
I'll fix in the next PR.

Sources/Paywalls/Events/Networking/PaywallHTTPRequestPath.swift
NachoSoto: So that when/if we add a new path the compiler fails and forces us to implement these for the new `case`.

Sources/Networking/InternalAPI.swift
NachoSoto: 10da09b89b146875800bc09bec22ccc7a9462e3e

PR #3159, opened 2023-09-06
`Paywalls`: created `PaywallEventsManager`

NachoSoto:
This combines the types created in prior PRs (#3157, #3158) to store and post events.

### DIFF
--- a/Sources/Logging/Strings/PaywallsStrings.swift
+++ b/Sources/Logging/Strings/PaywallsStrings.swift
@@ -21,6 +21,16 @@ enum PaywallsStrings {
     case warming_up_images(imageURLs: Set<URL>)
     case error_prefetching_image(URL, Error)
 
+    // MARK: - Events
+
+    case event_manager_not_initialized_not_available
+    case event_manager_failed_to_initialize(Error)
+
+    case event_flush_already_in_progress
+    case event_flush_with_empty_store
+    case event_flush_starting(count: Int)
+    case event_flush_failed(BackendError)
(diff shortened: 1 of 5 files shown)


Sources/Logging/Strings/PaywallsStrings.swift
```diff
@@ -35,6 +45,27 @@ extension PaywallsStrings: LogMessage {
...
+            return "Paywall event flushing already in progress. Skipping."
+
+        case .event_flush_with_empty_store:
+            return "Paywall event flushing requested with empty store."
+
+        case let .event_flush_starting(count):
+            return "Paywall event flush: posting \(count) events."
+
+        case let .event_flush_failed(error):
+            return "Paywall event flushing failed, will retry. Error: \(error.localizedDescription)"
```
joshdholtz, line 68: I'm a big fan of your event/error messages 😀  I really need to want to do this more myself 😛

joshdholtz (approved):
Looks good! I had a few questions but then I saw and read the tests and it cleared my questions up 😊 

LGTM 💪 🚀

Sources/Logging/Strings/PaywallsStrings.swift
NachoSoto: As an SDK with no UI (well, you know what I mean 🤣) logs are so important. I love having `verbose` logs now that can have a lot of this information and most people won't see, but we can still look into when looking into bugs or when running tests, etc.

PR #3160, opened 2023-09-07
`Paywalls`: added `Purchases.track(paywallEvent:)`

NachoSoto:
This uses `PaywallEventsManager` (#3159) to track events, and saves the presented `PaywallEvent` to be send along with `PostReceiptDataOperation`.


RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -4805,7 +4809,7 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/RevenueCat/purchases-ios";
 			requirement = {
-				branch = main;
+				branch = paywalls;
```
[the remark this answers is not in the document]
NachoSoto: OMG thanks!

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1031,6 +1053,28 @@ public extension Purchases {
...
+    /// Used by `RevenueCatUI` to keep track of ``PaywallEvent``s.
+    func track(paywallEvent: PaywallEvent) async {
+        switch paywallEvent {
+        case let .view(data):
+            self.purchasesOrchestrator.cachePresentedPaywall(data)
+
+        case .close:
+            self.purchasesOrchestrator.clearPresentedPaywall()
+
+        case .cancel: break
```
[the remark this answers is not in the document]
NachoSoto: I'll add a comment. Nothing special to do other than tracking it with the line below.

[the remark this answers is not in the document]
NachoSoto: Done.

RevenueCat.xcodeproj/project.pbxproj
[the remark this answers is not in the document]
NachoSoto: Removed.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1031,6 +1054,30 @@ public extension Purchases {
 
 // swiftlint:enable missing_docs
 
+// MARK: - Paywalls
+
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+public extension Purchases {
```
aboedo, line 1060: man I wish we could use `package`

aboedo: We used to use objective-c headers added to the dependency for this kind of thing. I can't think of any clean way of doing this without having to expose the method.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1031,6 +1054,30 @@ public extension Purchases {
...
+
+        case .close:
+            self.purchasesOrchestrator.clearPresentedPaywall()
+
+        case .cancel:
+            // No special handling, simply track the event below.
+            break
+        }
+
+        await self.paywallEventsManager?.track(paywallEvent: paywallEvent)
```
aboedo, line 1076: should we be moving this to the orchestrator layer? Ideally we do as little as possible at the Facade layer

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1031,6 +1054,30 @@ public extension Purchases {
...
+public extension Purchases {
+
+    /// Used by `RevenueCatUI` to keep track of ``PaywallEvent``s.
+    func track(paywallEvent: PaywallEvent) async {
+        switch paywallEvent {
+        case let .view(data):
+            self.purchasesOrchestrator.cachePresentedPaywall(data)
+
+        case .close:
+            self.purchasesOrchestrator.clearPresentedPaywall()
```
aboedo, line 1069: feels like the orchestrator should be the one deciding what to do based on the params

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: Yeah...

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: This has nothing to do with _purchases_ though. `PaywallEventsManager` was my way of extracting this out, though I agree ideally methods in `Purchases` are no more than 1 line.

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: Good point, I'll pass the event instead.

joshdholtz (approved):

PR #3163, opened 2023-09-07
`Integration Tests`: improve flaky tests

NachoSoto:
We get a lot of flaky failures on `OfflineStoreKitIntegrationTests` (example: https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/14843/workflows/a3cffa7c-da25-4811-931d-fc241befee0f/jobs/118214/tests).
I managed to reproduce locally running some of these 100+ times on a loop until they failed.

I traced the issue to this sequence:
- SK2 purchase
- `StoreKit.Transaction.unfinished` is empty

This is likely due to a race condition. In practice, when computing offline `CustomerInfo`, we never check it right away, since we need to wait for the server to return a 500.

I filed a Radar for this: `FB13133387`.

To make tests more realistic, I introduced a delay in `HTTPClient` returning fake 5xx errors. I haven't been able to reproduce this issue again, so this should make `OfflineStoreKitIntegrationTests` more reliable.

I've also added a new debug log that could help debug similar issues, because `Strings.offline.computed_offline_customer_info` was not very complete.

### DIFF
--- a/Sources/Logging/Strings/NetworkStrings.swift
+++ b/Sources/Logging/Strings/NetworkStrings.swift
@@ -113,7 +113,7 @@ extension NetworkStrings: LogMessage {
 
         #if DEBUG
         case let .api_request_forcing_server_error(request):
-            return "Returning fake HTTP 500 error for '\(request.description)'"
+            return "Returning fake HTTP 500 error for \(request.description)"
 
         case let .api_request_forcing_signature_failure(request):
             return "Returning fake signature verification failure for '\(request.description)'"
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -64,9 +64,12 @@ class HTTPClient {
         #if DEBUG
         guard !self.systemInfo.dangerousSettings.internalSettings.forceServerErrors else {
             Logger.warn(Strings.network.api_request_forcing_server_error(request))
-            completionHandler?(
-                .failure(.errorResponse(Self.serverErrorResponse, .internalServerError))
-            )
+            DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(300)) {
+                completionHandler?(
+                    .failure(.errorResponse(Self.serverErrorResponse, .internalServerError))
+                )
+            }
+
             return
         }
         #endif
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -64,6 +64,10 @@ class HTTPClient {
         #if DEBUG
         guard !self.systemInfo.dangerousSettings.internalSettings.forceServerErrors else {
             Logger.warn(Strings.network.api_request_forcing_server_error(request))
+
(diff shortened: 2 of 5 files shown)


Sources/Logging/Strings/NetworkStrings.swift
```diff
@@ -113,7 +113,7 @@ extension NetworkStrings: LogMessage {
 
         #if DEBUG
         case let .api_request_forcing_server_error(request):
-            return "Returning fake HTTP 500 error for '\(request.description)'"
+            return "Returning fake HTTP 500 error for \(request.description)"
```
NachoSoto, line 116: This ended up with double quotes:
> Returning fake HTTP 500 error for 'POST '/v1/receipts''

Sources/OfflineEntitlements/OfflineCustomerInfoCreator.swift
```diff
@@ -80,7 +80,12 @@ class OfflineCustomerInfoCreator {
 
         let offlineCustomerInfo = creator(products, mapping, userID)
 
-        Logger.info(Strings.offlineEntitlements.computed_offline_customer_info(offlineCustomerInfo.entitlements))
+        Logger.info(Strings.offlineEntitlements.computed_offline_customer_info(
+            products, offlineCustomerInfo.entitlements
+        ))
+        Logger.debug(Strings.offlineEntitlements.computed_offline_customer_info_details(
```
NachoSoto, line 86: I improved these logs to be able to debug future issues here. Initially I wasn't sure if `PurchasedProductsFetcher` was working correctly.

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -64,9 +64,12 @@ class HTTPClient {
         #if DEBUG
         guard !self.systemInfo.dangerousSettings.internalSettings.forceServerErrors else {
             Logger.warn(Strings.network.api_request_forcing_server_error(request))
-            completionHandler?(
-                .failure(.errorResponse(Self.serverErrorResponse, .internalServerError))
-            )
+            DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(300)) {
```
aboedo, line 67: the intent behind this isn't obvious at first glance so let's make sure to add a comment here

aboedo:
> I traced the issue to this sequence:
> 
> SK2 purchase
> StoreKit.Transaction.unfinished is empty

Do we have a radar for this? It doesn't seem like expected behavior. Just to clarify, we check unfinished after we get the response from await purchase, right?

aboedo (approved):

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto: Good point, I always think it's in git history but that's easily lost.

NachoSoto:
> Just to clarify, we check unfinished after we get the response from await purchase, right?

Yup

> Do we have a radar for this? It doesn't seem like expected behavior. 

Nope. I'll file one 👍🏻

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto: Done.

PR #3164, opened 2023-09-07
`Paywalls`: send events to `Purchases`

NachoSoto:
Last part, follow up to #3160.

This now makes `PaywallView` track events through the `RevenueCat` SDK.

### DIFF
--- a/RevenueCatUI/Data/TestData.swift
+++ b/RevenueCatUI/Data/TestData.swift
@@ -186,7 +186,8 @@ internal enum TestData {
             colors: .init(light: Self.lightColors, dark: Self.darkColors)
         ),
         localization: Self.localization1,
-        assetBaseURL: Self.paywallAssetBaseURL
+        assetBaseURL: Self.paywallAssetBaseURL,
+        revision: 5
     )
 
     static let offeringWithIntroOffer = Offering(
@@ -560,19 +561,19 @@ extension PurchaseHandler {
             )
         } restorePurchases: {
             return TestData.customerInfo
+        } trackEvent: { event in
+            Logger.debug("Tracking event: \(event)")
         }
     }
 
     static func cancelling() -> Self {
-        return self.init { _ in
-            return (
-                transaction: nil,
-                customerInfo: TestData.customerInfo,
-                userCancelled: true
-            )
-        } restorePurchases: {
-            return TestData.customerInfo
-        }
+        return .mock()
+            .map { block in {
+                    var result = try await block($0)
+                    result.userCancelled = true
+                    return result
+                }
+            } restore: { $0 }
     }
 
     /// Creates a copy of this `PurchaseHandler` with a delay.
@@ -589,7 +590,10 @@ extension PurchaseHandler {
         }
         }
     }
+
 }
+
+// MARK: -
 
 extension PaywallColor: ExpressibleByStringLiteral {
 
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -187,6 +187,9 @@ struct LoadedOfferingPaywallView: View {
     private let mode: PaywallViewMode
     private let fonts: PaywallFontProvider
 
+    @State
+    private var session: (lastPaywall: DisplayedPaywall, id: PaywallEvent.SessionID)
+
     @StateObject
     private var introEligibility: IntroEligibilityViewModel
     @ObservedObject
@@ -194,6 +197,9 @@ struct LoadedOfferingPaywallView: View {
 
     @Environment(\.locale)
     private var locale
+
+    @Environment(\.colorScheme)
+    private var colorScheme
 
     init(
         offering: Offering,
@@ -215,6 +221,10 @@ struct LoadedOfferingPaywallView: View {
             wrappedValue: .init(introEligibilityChecker: introEligibility)
         )
         self._purchaseHandler = .init(initialValue: purchaseHandler)
+        self._session = .init(initialValue: (
+            lastPaywall: .init(offering: offering, paywall: paywall),
+            id: .init()
+        ))
     }
 
     var body: some View {
@@ -233,6 +243,8 @@ struct LoadedOfferingPaywallView: View {
             .preference(key: RestoredCustomerInfoPreferenceKey.self,
                         value: self.purchaseHandler.restoredCustomerInfo)
             .disabled(self.purchaseHandler.actionInProgress)
+            .onAppear { self.purchaseHandler.trackPaywallView(self.eventData) }
+            .onDisappear { self.purchaseHandler.trackPaywallClose() }
 
         switch self.mode {
         case .fullScreen:
@@ -242,6 +254,44 @@ struct LoadedOfferingPaywallView: View {
             view
                 .fixedSize(horizontal: false, vertical: true)
                 .edgesIgnoringSafeArea(.bottom)
+        }
+    }
+
+    private var eventData: PaywallEvent.Data {
+        self.updateSessionIfNeeded()
+
+        return .init(
+            offering: self.offering,
+            paywall: self.paywall,
+            sessionID: self.session.id,
+            displayMode: self.mode,
+            locale: .current,
+            darkMode: self.colorScheme == .dark
+        )
+    }
+
+    private func updateSessionIfNeeded() {
+        let newPaywall: DisplayedPaywall = .init(offering: self.offering, paywall: self.paywall)
+        guard self.session.lastPaywall != newPaywall else { return }
+
+        self.session.lastPaywall = newPaywall
+        self.session.id = .init()
+    }
+
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
+private extension LoadedOfferingPaywallView {
+
+    struct DisplayedPaywall: Equatable {
+        var offeringIdentifier: String
+        var paywallTemplate: String
+        var revision: Int
+
+        init(offering: Offering, paywall: PaywallData) {
+            self.offeringIdentifier = offering.identifier
+            self.paywallTemplate = paywall.templateName
+            self.revision = paywall.revision
         }
     }
 
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -221,6 +221,9 @@ struct LoadedOfferingPaywallView: View {
             wrappedValue: .init(introEligibilityChecker: introEligibility)
         )
         self._purchaseHandler = .init(initialValue: purchaseHandler)
+
+        // Each `PaywallView` impression gets its own session.
+        // See also `updateSessionIfNeeded`.
         self._session = .init(initialValue: (
             lastPaywall: .init(offering: offering, paywall: paywall),
             id: .init()
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -244,7 +244,7 @@ struct LoadedOfferingPaywallView: View {
                         value: self.purchaseHandler.restoredCustomerInfo)
             .disabled(self.purchaseHandler.actionInProgress)
             .onAppear { self.purchaseHandler.trackPaywallView(self.eventData) }
-            .onDisappear { self.purchaseHandler.trackPaywallClose() }
+            .onDisappear { self.purchaseHandler.trackPaywallClose(self.eventData) }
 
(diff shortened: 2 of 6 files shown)


RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -96,13 +106,62 @@ extension PurchaseHandler {
...
+}
+
+// MARK: - Private
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
+private extension PurchaseHandler {
+
+    func trackEvent(_ eventCreator: (PaywallEvent.Data) -> PaywallEvent) {
+        guard let data = self.eventData else {
+            Logger.warning(Strings.attempted_to_track_event_with_missing_data)
```
NachoSoto, line 156: This is covered by tests. I tried different approaches of where to create the session ID to avoid this, but this was the simplest.

RevenueCatUI/Views/LoadingPaywallView.swift
```diff
@@ -71,6 +71,9 @@ private extension LoadingPaywallView {
         },
         restorePurchases: {
             fatalError("Should not be able to purchase")
+        },
+        trackEvent: { _ in
+            // Ignoring events from loading paywall view
```
NachoSoto, line 76: I love that the compiler made me implement this. And realized this was the best solution.

Tests/RevenueCatUITests/PaywallEventsTests.swift
```diff
@@ -0,0 +1,194 @@
...
+import RevenueCat
+@testable import RevenueCatUI
+import SwiftUI
+import XCTest
+
+#if !os(macOS)
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@MainActor
+class PaywallEventsTests: TestCase {
```
NachoSoto, line 24: These tests cover everything 🎉

RevenueCatUI/Data/TestData.swift
```diff
@@ -186,7 +186,8 @@ internal enum TestData {
             colors: .init(light: Self.lightColors, dark: Self.darkColors)
         ),
         localization: Self.localization1,
-        assetBaseURL: Self.paywallAssetBaseURL
+        assetBaseURL: Self.paywallAssetBaseURL,
+        revision: 5
```
aboedo, line 190: this quickly became a BIG file. Maybe we should move it to a folder and split into smaller things?

aboedo: it'll only grow over time

RevenueCatUI/PaywallView.swift
```diff
@@ -227,6 +236,8 @@ struct LoadedOfferingPaywallView: View {
             .preference(key: PurchasedCustomerInfoPreferenceKey.self,
                         value: self.purchaseHandler.purchasedCustomerInfo)
             .disabled(self.purchaseHandler.actionInProgress)
+            .onAppear { self.purchaseHandler.trackPaywallView(self.eventData) }
+            .onDisappear { self.purchaseHandler.trackPaywallClose() }
```
aboedo, line 240: does onDisappear get called when the app gets backgrounded? And when the iOS payment window pops up?

aboedo: wondering whether we need to [do something like this](https://lyons.app/2020/08/31/how-to-only-run-ondisappear-when-a-view-is-unloaded-in-swiftui/#:~:text=How%20to%20only%20run%20onDisappear%20when%20a%20view,code%20when%20the%20view%20is%20dismissed%20or%20unloaded.)

RevenueCatUI/PaywallView.swift
```diff
@@ -239,6 +250,44 @@ struct LoadedOfferingPaywallView: View {
...
+            darkMode: self.colorScheme == .dark
+        )
+    }
+
+    private func updateSessionIfNeeded() {
+        let newPaywall: DisplayedPaywall = .init(offering: self.offering, paywall: self.paywall)
+        guard self.session.lastPaywall != newPaywall else { return }
+
+        self.session.lastPaywall = newPaywall
+        self.session.id = .init()
```
aboedo, line 271: Note: we're baking in an assumption here: 
We're saying that if SwiftUI keeps the PaywallView in memory but calls onAppear more than once, we consider it to be the same session, right? 
If that's the case, maybe we should explicitly call it out here?

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -96,13 +106,62 @@ extension PurchaseHandler {
         return result
     }
 
+    func trackPaywallView(_ eventData: PaywallEvent.Data) {
```
aboedo, line 109: I'm super late to this, but given that we already use "view" and "paywallView" a bunch because SwiftUI, I'd call this a Paywall Impression event

RevenueCatUI/Purchasing/PurchaseHandler.swift
aboedo: Could you elaborate on what problem this approach solves? I'm trying to understand how we ended up with trackEvent taking a function that turns data into an event, and I'm sure there's a reasonable explanation but Fried Brain Fridays

RevenueCatUI/Data/TestData.swift
NachoSoto: Yup 👍🏻  I'll do it next week.

RevenueCatUI/Purchasing/PurchaseHandler.swift
NachoSoto: Good idea 👍🏻

RevenueCatUI/Purchasing/PurchaseHandler.swift
NachoSoto: This method relies on `self.eventData`, because that needs to be saved throughout the lifetime of a `PaywallView` (since it has the session ID, dark mode, locale, etc).
The closure parameter is just a way to avoid duplicated code. It takes `self.eventData` (which is a `PaywallEvent.Data`) and wraps it into a `PaywallEvent`.
So it gets used like this:
```
self.trackEvent(PaywallEvent.view)
self.trackEvent(PaywallEvent.cancel)
self.trackEvent(PaywallEvent.close)
```

RevenueCatUI/PaywallView.swift
NachoSoto: Good point, let me test this + that approach.

RevenueCatUI/Purchasing/PurchaseHandler.swift
aboedo: it took me a few mins to understand the implementation tbh. I got confused with the closure, the fact that there are a few things called trackEvent and the mapping on the init. 

Would this work? 

- always pass in the full paywall data for events, including cancel and close. The event reporting layer has that data after all, and that protects against unexpected cases where cancel comes before view or something (we're delegating into background priority tasks, and although I imagine things happen in order, in case of a freak accident maybe it'd be safer to always have the data). 
- remove the closures, making the implementation of this method something like
```swift
    func track(event: PaywallEvent) {
        Task.detached(priority: .background) { [block = self.trackEventBlock] in
            await purchases.track(paywallEvent: event)
        }
    }
```

and called like 
```swift
func trackPaywallView(_ eventData: PaywallEvent.Data) {
    self.track(event: PaywallEvent.view(eventData))
}
...
self.purchaseHandler.trackPaywallView(self.eventData)
```

aboedo: we'd need to also pass in the paywall data when calling purchase and restore, but I think that'd be fine? the caller knows the paywall information anyway

NachoSoto: Refactored

RevenueCatUI/PaywallView.swift
NachoSoto: Yup, I'll document it (I've also confirmed with @joshdholtz that repeated view events will be handled in the backend).

RevenueCatUI/Purchasing/PurchaseHandler.swift
NachoSoto: I'll make a PR renaming this in the code. I'll keep the backend event identifier as "view" unless @joshdholtz is also okay changing that.

RevenueCatUI/PaywallView.swift
NachoSoto: The answer to both is no :) so the simple approach here works perfectly without extra hacks :)

RevenueCatUI/PaywallView.swift
```diff
@@ -215,6 +221,13 @@ struct LoadedOfferingPaywallView: View {
             initialValue: .init(introEligibilityChecker: introEligibility)
         )
         self._purchaseHandler = .init(initialValue: purchaseHandler)
+
+        // Each `PaywallView` impression gets its own session.
+        // See also `updateSessionIfNeeded`.
+        self._session = .init(initialValue: (
```
aboedo, line 227: ❤️

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -15,17 +15,19 @@ import RevenueCat
 import StoreKit
 import SwiftUI
 
-@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
 final class PurchaseHandler: ObservableObject {
 
     typealias PurchaseBlock = @Sendable (Package) async throws -> PurchaseResultData
     typealias RestoreBlock = @Sendable () async throws -> CustomerInfo
+    typealias TrackEventBlock = @Sendable (PaywallEvent) async -> Void
```
aboedo, line 23: These are now only needed for testing purposes, right? 
Is it worth keeping them around for that? Having them also forces us to have fatalError workarounds in the loading view IIRC?

Would it suffice to have a purchases mock (spy) and pass that in, so that we can directly call the methods from purchases? Seems like we only use them to check that we're calling the right methods with the right params

NachoSoto: These serve several purposes:
- Abstract implementation from `Purchases`
- Previews: we can have custom handlers that don't do anything or that complete a successful purchase
- Snapshots: same thing
- Unit tests (`PaywallFooterTests` for example)
- `LoadingPaywallView`: this one doesn't use real blocks for obvious reasons

> Having them also forces us to have fatalError workarounds in the loading view IIRC?

We have a `fatalError` in case you try to purchase from the loading view (which is `.disabled`, so it's impossible). Other than that, the `trackEvent` one is a no-op now.

Hopefully that answers your question why we can't just use a `Purchases` instance. We could create a `PurchasesType` implementation, but that would require implementing dozens of methods instead of just 3.

aboedo: Yeah, that's what I was getting at, essentially for DI. 

I don't feel strong enough about this to continue to block this PR, but I do think there are alternative approaches worth studying, that wouldn't necessitate building a mock of all methods in `PurchasesType` from scratch. 

I worry about us adding a different approach for testing (instead of DI of a class, DI at the closure level). It makes things more complex for a brand new maintainer. And I'd argue that the closures are harder to read and wrap your head around than just passing in a purchases object, although that bit is subjective. 

Alternate approaches that I think would also do the trick: 
- find a way to share the purchases mock instance we use for other tests. We already share some tests code, this wouldn't fall too far from that
- create a new, much smaller protocol, that only has the methods that we'd need (purchase, restore, track events). Refactor PurchasesType so that it also conforms to this protocol. Then we pass in an instance of this smaller protocol, which is easy to mock. Then PurchasesHandler just takes a `purchases: NewSmallType = Purchases.shared` as param, and the rest of the class is dead simple.

aboedo (approved):

RevenueCatUI/Purchasing/PurchaseHandler.swift
NachoSoto: Thanks for clarifying what you meant! It does indeed lead to a simpler implementation :)

Done in #3196

RevenueCatUI/Data/TestData.swift
NachoSoto: Also done in #3196

PR #3166, opened 2023-09-07
`Xcode`: removed `purchases-ios` SPM reference

NachoSoto:
The project no longer depends on `purchases-ios` through SPM.
This was needed for 2 reasons:
- `v3LoadShedderIntegration`
- `BackendCustomEntitlementsIntegrationTests`

Both of these depend on `RevenueCat_CustomEntitlementComputation`, which is only an SPM library.

This solution allows us to build it but without having to clone the repo separately through a dependency, and instead embedding a reference to the _local_ Package.
`v3LoadShedderIntegration` depended on the 3.x release, so I've simply removed it from the workspace. It can still be modified and ran directly by opening the project.

This solves at least 2 major current annoyances:
- Compiling `BackendIntegrationTests` lead to compile errors because the local changes weren't seen when building `RevenueCat_CustomEntitlementComputation`.
- Switching branches was always frustrating, as Xcode tried to modify `RevenueCat.xcworkspace/xcshareddata/swiftpm/Package.resolved` to point to the latest commit in `main`, sometimes even making it crash.

⚠️  *This might also help with the timeouts when cloning the repo, since SPM was probably resolving our repo twice through the transitive dependency.*

#### New setup:
Screenshot 2023-09-07 at 15 01 53

### DIFF
--- a/RevenueCat.xcodeproj/project.pbxproj
+++ b/RevenueCat.xcodeproj/project.pbxproj
@@ -702,20 +702,6 @@
 			remoteGlobalIDString = 2DC5621524EC63420031F69B;
 			remoteInfo = RevenueCat;
 		};
-		2D9E3CBE29D4D60600DB4730 /* PBXContainerItemProxy */ = {
-			isa = PBXContainerItemProxy;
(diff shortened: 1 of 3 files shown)


tonidero (approved):
Nice!

PR #3171, opened 2023-09-08
`Paywalls`: improved `FooterView`

charliemchapman:
### Motivation
The buttons at the bottom of the paywall templates underneath the CTA are smaller than 44pt at the default font size and visually are too tight to the CTA button.

### Description
I set the minimum height to 44pt for all of the buttons in the footer view and then updated the template layouts to account for the change.

I tested using the SimpleApp in multiple simulators. Unfortunately I do not have iOS 15 simulators and can't seem to download them in Xcode so I wasn't able to test this against iOS 15 yet which I know we've run into layout issues with before.

### Screenshots
image

image

image

image

image

image


RevenueCatUI/Templates/Template3View.swift
```diff
@@ -76,7 +76,6 @@ struct Template3View: TemplateViewType {
                 introEligibility: self.introEligibility,
                 purchaseHandler: self.purchaseHandler
             )
-            .padding(.bottom)
```
NachoSoto, line 79: Thanks for cleaning up all these paddings to be consistent 🙏🏻

RevenueCatUI/Views/FooterView.swift
```diff
@@ -27,6 +27,7 @@ struct FooterView: View {
     var boldPreferred: Bool
     var purchaseHandler: PurchaseHandler
     var displayingAllPlans: Binding<Bool>?
+    static let buttonMinHeight: CGFloat = 44
```
NachoSoto, line 30: This is a great idea 👏🏻 

Suggestion:
- Let's extract this to `Constants.swift`
- We can define it like this to be consistent with the [design guidelines](https://developer.apple.com/design/human-interface-guidelines/buttons#Best-practices) (maybe even add a comment to reference it):

```swift
// See https://developer.apple.com/design/human-interface-guidelines/buttons#Best-practices
#if swift(>=5.9) && os(visionOS)
static let minimumButtonHeight: CGFloat = 60
#else
static let minimumButtonHeight: CGFloat = 44
#endif
```

RevenueCatUI/Views/FooterView.swift
```diff
@@ -114,6 +115,7 @@ struct FooterView: View {
         } label: {
             Text("All subscriptions", bundle: .module)
         }
+        .frame(minHeight: FooterView.buttonMinHeight)
```
NachoSoto, line 118: Should we apply this to the entire view instead?

charliemchapman: That was my original thought, but I want to make sure the button tap area itself is 44pt. The buttons won't always expand to fill the HStack right?

NachoSoto: Good point. You'd need `.frame(maxHeight: .infinity)` but that will cause other issues so this is fine 👍🏻

RevenueCatUI/Views/FooterView.swift
```diff
@@ -147,10 +148,10 @@ private struct SeparatorView: View {
             .accessibilityHidden(true)
     }
 
-    @ScaledMetric(relativeTo: .caption)
```
NachoSoto, line 150: Oh thanks 👏🏻 
Unfortunately this can't use a common constant so it has to be hardcoded again.

NachoSoto (approved):

Tests/purchases-ios-snapshots-commit
```diff
@@ -1 +1 @@
-82d84e2e241db3853e487cb8718a22d112dfba8f
+58e1e2b348e67f180562a49541247d0f0796050e
```
NachoSoto, line 1: @charliemchapman updated :)

PR #3172, opened 2023-09-08
`Paywalls`: fixed `FooterView` horizontal centering

NachoSoto:
`hidden` is not the right approach here because it's still part of the layout. Obvious in retrospect, but this was one of the first few elements done for paywalls and never went back to it.


charliemchapman (approved):
yep, that makes sense 👍

PR #3187, opened 2023-09-11
`Paywalls`: fixed data flow resulting in multiple `PurchaseHandler` instances

NachoSoto:
This was breaking event notifications as well as `OnPurchaseCompletedModifier`, because there were multiple `PurchaseHandler` instances created.

The solution is to make `PaywallView` store these as `@StateObject`. In order to do that, the type can't be optional, so I instead encoded the "is configured" information inside.

### DIFF
--- a/RevenueCatUI/View+PresentPaywall.swift
+++ b/RevenueCatUI/View+PresentPaywall.swift
@@ -142,12 +142,12 @@ private struct PresentingPaywallModifier: ViewModifier {
                     .onPurchaseCompleted {
                         self.purchaseCompleted?($0)
 
-                        self.data = nil
+                        self.close()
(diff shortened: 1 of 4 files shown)


RevenueCatUI/View+PresentPaywall.swift
```diff
@@ -142,12 +142,12 @@ private struct PresentingPaywallModifier: ViewModifier {
                     .onPurchaseCompleted {
                         self.purchaseCompleted?($0)
 
-                        self.data = nil
+                        self.close()
```
NachoSoto, line 145: This is just a small refactor

joshdholtz (approved):

PR #3190, opened 2023-09-11
`Paywalls`: new `onRestoreCompleted` handler

NachoSoto:
This is available on `.paywallFooter()`, `.presentPaywallIfNeeded`, and as a `.onRestoreCompleted` modifier.


RevenueCatUI/View+PresentPaywall.swift
```diff
@@ -144,6 +154,13 @@ private struct PresentingPaywallModifier: ViewModifier {
 
                         self.close()
                     }
+                    .onPurchaseCompleted { customerInfo in
+                        self.restoreCompleted?(customerInfo)
+
+                        if !self.shouldDisplay(customerInfo) {
+                            self.close()
+                        }
```
NachoSoto, line 162: Data driven UI FTW: we can make it dismiss automatically as well if the condition becomes `false` after restoring 🎉

RevenueCatUI/View+PurchaseRestoreCompleted.swift
```diff
@@ -0,0 +1,115 @@
...
+    ///  }
+    /// ```
+    ///
+    /// - Warning: Receiving a ``CustomerInfo``does not imply that the user has any entitlements,
+    /// simply that the process was successful. You must verify the ``CustomerInfo/entitlements``
+    /// to confirm that they are active.
+    ///
+    /// ### Related Articles
+    /// [Documentation](https://rev.cat/paywalls)
+    public func onRestoreCompleted(
```
tonidero, line 77: I wonder, would it make sense to group all these callbacks together? In android, I was planning to create a "Listener" interface, that can be implemented by the developer to listen to any callbacks they need to.

RevenueCatUI/View+PresentPaywall.swift
```diff
@@ -144,6 +154,13 @@ private struct PresentingPaywallModifier: ViewModifier {
 
                         self.close()
                     }
+                    .onPurchaseCompleted { customerInfo in
```
NachoSoto, line 157:
```suggestion
                    .onRestoreCompleted { customerInfo in
```

joshdholtz (approved):
YESSSSSS 🥳

RevenueCatUI/View+PurchaseRestoreCompleted.swift
NachoSoto: That would be the `UIKit` way, with a delegate `protocol`. But no in SwiftUI...

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -76,8 +80,17 @@ public protocol PaywallViewControllerDelegate: AnyObject {
...
+    optional func paywallViewController(_ controller: PaywallViewController,
+                                        didFinishPurchasingWith customerInfo: CustomerInfo)
+
+    /// Notifies that the restore operation has completed in a ``PaywallViewController``.
+    ///
+    /// - Warning: Receiving a ``CustomerInfo``does not imply that the user has any entitlements,
+    /// simply that the process was successful. You must verify the ``CustomerInfo/entitlements``
+    /// to confirm that they are active.
+    @objc(paywallViewController:didFinishRestoringWithCustomerInfo:)
+    optional func paywallViewController(_ controller: PaywallViewController,
```
NachoSoto, line 92: Added this too just for @codykerns

PR #3194, opened 2023-09-12
`Paywalls`: fixed `IntroEligibilityViewModel` data lifetime

NachoSoto:
Follow up to #3187.
Thanks to @charliemchapman for reporting.

`IntroEligibilityViewModel` was being recreated, which meant that we were losing the internal state on redraws.

I've also simplified the initialization of `TrialOrIntroEligibilityChecker` and `PurchaseHandler` to avoid duplicated calls to `.default()`.

### DIFF
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -52,8 +52,8 @@ public struct PaywallView: View {
             offering: nil,
             customerInfo: nil,
             fonts: fonts,
-            introEligibility: .default(),
-            purchaseHandler: .default()
+            introEligibility: nil,
+            purchaseHandler: nil
         )
     }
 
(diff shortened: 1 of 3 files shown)


RevenueCatUI/PaywallView.swift
```diff
@@ -187,7 +187,7 @@ struct LoadedOfferingPaywallView: View {
     private let mode: PaywallViewMode
     private let fonts: PaywallFontProvider
 
-    @ObservedObject
+    @StateObject
```
NachoSoto, line 190: This was the main issue. We can use `@ObservedObject` on `PurchasesHandler` because it's injected, but here we create a new `IntroEligibilityViewModel`.

joshdholtz (approved):

PR #3195, opened 2023-09-12
`Paywalls`: convert empty images into `nil`

NachoSoto:
Ideally the backend would never send `""` for images, but better to be defensive.

### DIFF
--- a/Sources/Paywalls/PaywallData.swift
+++ b/Sources/Paywalls/PaywallData.swift
@@ -262,13 +262,29 @@ extension PaywallData.Configuration {
     public struct Images {
 
         /// Image displayed as a header in a template.
-        public var header: String?
+        public var header: String? {
+            get { self._header }
+            set { self._header = newValue }
+        }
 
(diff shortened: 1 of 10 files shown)


Sources/Paywalls/PaywallData.swift
```diff
@@ -262,13 +262,29 @@ extension PaywallData.Configuration {
...
+            get { self._icon }
+            set { self._icon = newValue }
+        }
+
+        @NonEmptyStringDecodable
+        var _header: String?
+        @NonEmptyStringDecodable
+        var _background: String?
+        @NonEmptyStringDecodable
+        var _icon: String?
```
NachoSoto, line 287: cc @vegaro we'll need to do the same in Android

joshdholtz (approved):

Sources/Paywalls/PaywallData.swift
NachoSoto: https://github.com/RevenueCat/purchases-android/pull/1398

PR #3183, opened 2023-09-08
`Paywalls`: polish template 4

NachoSoto:
image
image

PR #3202, opened 2023-09-13
`Paywalls`: fixed broken layout on template 4

NachoSoto:
I broke this on #3183.

image

We need to reset the height calculation when the inputs change, but setting it to 0 was making it layout with a 0 height instead of unlimited height to measure it.

This also now enables template 4 since it's ready!

### DIFF
--- a/RevenueCatUI/Templates/Template4View.swift
+++ b/RevenueCatUI/Templates/Template4View.swift
@@ -25,7 +25,7 @@ struct Template4View: TemplateViewType {
     private var selectedPackage: TemplateViewConfiguration.Package
 
     @State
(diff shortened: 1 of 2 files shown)


RevenueCatUI/Templates/Template4View.swift
```diff
@@ -102,6 +102,7 @@ struct Template4View: TemplateViewType {
 
             self.subscribeButton
                 .defaultHorizontalPadding()
+                .padding(.bottom, Self.verticalPadding / -2)
```
NachoSoto, line 105: Feedback from @bab-s

joshdholtz (approved):

PR #3203, opened 2023-09-13
`Paywalls`: fixed SimpleApp release build

NachoSoto:
We're currently making debug builds for it, but this fixes the release build in case we want to do that (like for profiling).


Tests/TestingApps/SimpleApp/SimpleApp/SamplePaywalls.swift
```diff
@@ -5,6 +5,8 @@
 //  Created by Nacho Soto on 7/27/23.
 //
 
+#if DEBUG
+
 import Foundation
 import RevenueCat
 @testable import RevenueCatUI
```
NachoSoto, line 12: Can't use `@testable` on release builds.

joshdholtz (approved):

PR #3205, opened 2023-09-14
`Paywalls`: fixed mock intro eligibility on snapshot tests

NachoSoto:
This become broken in #3046.

I noticed that the "ineligible" snapshots didn't show the ineligible state. For example: https://github.com/RevenueCat/purchases-ios-snapshots/blob/58e1e2b348e67f180562a49541247d0f0796050e/Templates/__Snapshots__/Template1ViewTests/iOS16-testSamplePaywallWithIneligibleIntroOffer.1.png


Tests/RevenueCatUITests/BaseSnapshotTest.swift
```diff
@@ -55,7 +55,7 @@ class BaseSnapshotTest: TestCase {
                            customerInfo: TestData.customerInfo,
                            mode: mode,
                            fonts: fonts,
-                           introEligibility: eligibleChecker,
+                           introEligibility: introEligibility,
```
NachoSoto, line 58: This needs to be the parameter. This silently worked because of implicit `self` 😢

NachoSoto:
I'm generating snapshots with this

joshdholtz (approved):

PR #3206, opened 2023-09-14
Fix runtime crash in SK2TransactionListener in iOS < 15

tonidero:
### Description
We found that there was another runtime crash in Capacitor in iOS < 15 similar to https://github.com/RevenueCat/purchases-ios/pull/3139. This applies the same fix to this new crash.

### DIFF
--- a/Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
+++ b/Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
@@ -62,11 +62,10 @@ actor StoreKit2TransactionListener: StoreKit2TransactionListenerType {
     // We can't directly store instances of `AsyncStream`, since that causes runtime crashes when
     // loading this type in iOS <= 15, even with @available checks correctly in place.
     // See https://openradar.appspot.com/radar?id=4970535809187840 / https://github.com/apple/swift/issues/58099
-    private let _updates: Any
+    private let _updates: Box<AsyncStream<TransactionResult>>
 
     var updates: AsyncStream<TransactionResult> {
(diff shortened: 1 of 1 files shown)


Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
```diff
@@ -58,7 +58,16 @@ actor StoreKit2TransactionListener: StoreKit2TransactionListenerType {
...
+
+    // We can't directly store instances of `AsyncStream`, since that causes runtime crashes when
+    // loading this type in iOS <= 15, even with @available checks correctly in place.
+    // See https://openradar.appspot.com/radar?id=4970535809187840 / https://github.com/apple/swift/issues/58099
+    private let _updates: Any
+
+    var updates: AsyncStream<TransactionResult> {
+        // swiftlint:disable:next force_cast
+        return self._updates as! AsyncStream<TransactionResult>
+    }
```
NachoSoto, line 70: In this case we can use `Box` (we couldn't with `NavigationPath` because of the `@Published`).
Can you try changing this to
```suggestion
    private let _updates: Box<AsyncStream<TransactionResult>>

    var updates: AsyncStream<TransactionResult> {
        return self._updates.value
    }
```

NachoSoto:

Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
tonidero: Done!

NachoSoto (approved):
👏🏻

PR #3121, opened 2023-08-29
`Paywalls`: add support for CTA button gradients

NachoSoto:
Screenshot 2023-08-29 at 15 41 31

PR #3217, opened 2023-09-14
`Paywalls`: loading indicator for in-progress purchases

### LINKED ISSUE #3216: [ RevenueCatUI ] - Custom Loading View for PaywallView and PaywallFooter

**The solution**
Create an *extension view* similar to `onPurchaseCompleted/onRestoreCompleted´ that returns Bool to control the display of a Custom Loading View.

I have a custom loading throughout my application, but in RevenueCatUI I lose this UI consistency, because I did not find an "uncomplicated" way to do this simple action and receive the status of actionInBlock.
In my specific case I have an extension for the View called `.loading´ that receives a bool parameter with this implementation I could easily add my custom loading.

Por ex: 

```swift
  @State
  private var displayPaywall: Bool = true

  @State
  private var isLoading: Bool = true

  var body: some View {
     ContentView()
         .sheet(isPresented: self.$displayPaywall) {
             PaywallView()
                 .onLoadingStatus { status in
                     self.isLoading = status
                }
                .loading(self.isLoading)
         }
  }
```

A positive point for this approach also works for `paywallFooter´ and taking complexity and adding customization to each project.

Already taking the opportunity to say that a `onUserCancel´ is also positive so that you can have the opportunity to track the event and also display an offer to the user, an opportunity in view of the fact that many checkouts are started and canceled.

1. Environment
   1. Platform: iOS
   2. SDK version: 4.26.0-beta.4
   3. StoreKit version:
      - [x] StoreKit 1
      - [ ] StoreKit 2 (_enabled with `usesStoreKit2IfAvailable(true)`_)
   4. OS version: 16
   5. Xcode version: 14
   6. Device and/or simulator:
      - [x] Device
      - [x] Simulator
   7. Environment:
      - [x] Sandbox
      - [x] TestFight
      - [x] Production
   8. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue. **Complete logs with `Purchases.logLevel = .verbose` will help us debug this issue.**

Not necessary.

4. Another context that creates an inconsistency is when the paywall is displayed via shet where the user can close even after clicking buy due to the delay of the apple popup. 
It can be corrected with the developer obtaining the status and blocking the closure of the shet

NachoSoto:
Fixes #3216.

This likely got broken with #3121 because we stopped using `.buttonStyle`. This implements a manual style when the button is `disabled` (when `PurchaseHandler.actionInProgress` is `true`).

https://github.com/RevenueCat/purchases-ios/assets/685609/0e23bcf1-7e1d-4d57-adf7-61a76b1e997e

### DIFF
--- a/RevenueCatUI/Views/PurchaseButton.swift
+++ b/RevenueCatUI/Views/PurchaseButton.swift
@@ -59,9 +59,6 @@ struct PurchaseButton: View {
         self.purchaseHandler = purchaseHandler
     }
 
-    @Environment(\.userInterfaceIdiom)
-    var userInterfaceIdiom
-
     var body: some View {
         self.button
     }
@@ -69,29 +66,22 @@ struct PurchaseButton: View {
     private var button: some View {
         AsyncButton {
             guard !self.purchaseHandler.actionInProgress else { return }
+            guard !self.package.currentlySubscribed else { return }
 
             _ = try await self.purchaseHandler.purchase(package: self.package.content)
         } label: {
-            IntroEligibilityStateView(
-                textWithNoIntroOffer: self.package.localization.callToAction,
-                textWithIntroOffer: self.package.localization.callToActionWithIntroOffer,
+            PurchaseButtonLabel(
+                package: self.package,
+                colors: self.colors,
                 introEligibility: self.introEligibility,
-                foregroundColor: self.colors.callToActionForegroundColor
+                mode: self.mode
             )
-                .frame(
-                    maxWidth: self.mode.fullWidthButton
-                       ? .infinity
-                        : nil
-                )
-                .padding()
-                .padding(.vertical, self.userInterfaceIdiom == .pad ? 10 : 0)
         }
         .font(self.fonts.font(for: self.mode.buttonFont).weight(.semibold))
         .background(self.backgroundView)
         .tint(.clear)
         .frame(maxWidth: .infinity)
         .dynamicTypeSize(...Constants.maximumDynamicTypeSize)
-        .disabled(self.package.currentlySubscribed)
     }
 
     @ViewBuilder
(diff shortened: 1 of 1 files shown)


RevenueCatUI/Views/PurchaseButton.swift
```diff
@@ -59,39 +59,29 @@ struct PurchaseButton: View {
...
-    var userInterfaceIdiom
-
     var body: some View {
         self.button
     }
 
     private var button: some View {
         AsyncButton {
             guard !self.purchaseHandler.actionInProgress else { return }
+            guard !self.package.currentlySubscribed else { return }
```
NachoSoto, line 69: Doing this instead of marking the button as `disabled`, which would display a spinner now.

RevenueCatUI/Views/PurchaseButton.swift
```diff
@@ -116,6 +106,46 @@ struct PurchaseButton: View {
...
+            foregroundColor: self.colors.callToActionForegroundColor
+        )
+            .frame(
+                maxWidth: self.mode.fullWidthButton
+                   ? .infinity
+                    : nil
+            )
+            .padding()
+            .padding(.vertical, self.userInterfaceIdiom == .pad ? 10 : 0)
+            .hidden(if: !self.isEnabled)
```
NachoSoto, line 137: Always displaying this to keep layout consistent, but hiding it when the button is disabled.

RevenueCatUI/Views/PurchaseButton.swift
```diff
@@ -116,6 +106,46 @@ struct PurchaseButton: View {
...
+            .overlay {
+                if !self.isEnabled {
+                    ProgressView()
```
aboedo, line 140: Super minor, FFTI: but given that it goes into a loading state when it's not enabled, maybe the flag should be something like `isLoading` or `hasOperationInProgress` or something to reflect that

aboedo (approved):

joshdholtz (approved):

RevenueCatUI/Views/PurchaseButton.swift
NachoSoto: But it's coming from the `.isEnabled` environment variable, that's why I named it as such 🤔

PR #3218, opened 2023-09-14
`Integration Tests`: fixed more flaky failures

NachoSoto:
Same as #3167. These can fail when verifying that no transactions are finished because there might be leftover transactions before the test even begins. In order to ignore those, we clear `TestLogHandler` at the beginning. We don't do that for every test because in some test we want to look at what was logged during SDK initialization.


joshdholtz (approved):

NachoSoto:
Merging this to unblock `paywalls`.

aboedo:
> We don't do that for every test because in some test we want to look at what was logged during SDK initialization.

Maybe we should always clear for every test and trigger initialization _within_ the test for the ones where we want those logs. 
Having an unclean environment when testing starts seems like a recipe for future stress when a new dev comes in and starts writing a test and runs into this again

PR #3222, opened 2023-09-18
`Paywalls`: improved `footerView` to use `.continuous` rounded corners

NachoSoto:
This API is iOS 16+ and Xcode 15 only but it's worth it to make it more consistent with everything else.

This also improves Template 4.


RevenueCatUI/Modifiers/ViewExtensions.swift
```diff
@@ -295,3 +314,23 @@ private extension Axis {
     }
 
 }
+
+// MARK: -
+
+#if swift(>=5.9)
```
NachoSoto, line 320: `UnevenRoundedRectangle` was only added in `Xcode 15` (backported to iOS 16)

vegaro (approved):

PR #3233, opened 2023-09-20
`Paywalls`: don't display "Purchases restored successfully" if nothings was restored

NachoSoto:
We can improve the logic here, but this is a simple improvement for the obvious case where nothing was actually restored.

### DIFF
--- a/RevenueCatUI/Views/FooterView.swift
+++ b/RevenueCatUI/Views/FooterView.swift
@@ -167,8 +167,11 @@ private struct RestorePurchasesButton: View {
 
     var body: some View {
         AsyncButton {
-            _ = try await self.purchaseHandler.restorePurchases()
-            self.displayRestoredAlert = true
+            let customerInfo = try await self.purchaseHandler.restorePurchases()
+
+            if !customerInfo.entitlements.active.isEmpty {
+                self.displayRestoredAlert = true
+            }
         } label: {
             if #available(iOS 16.0, macOS 13.0, tvOS 16.0, *) {
                 ViewThatFits {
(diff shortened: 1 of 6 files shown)


RevenueCatUI/Views/FooterView.swift
```diff
@@ -167,8 +167,11 @@ private struct RestorePurchasesButton: View {
 
     var body: some View {
         AsyncButton {
-            _ = try await self.purchaseHandler.restorePurchases()
-            self.displayRestoredAlert = true
+            let customerInfo = try await self.purchaseHandler.restorePurchases()
+
+            if !customerInfo.entitlements.active.isEmpty {
+                self.displayRestoredAlert = true
+            }
```
aboedo, line 174: This might not be accurate - it assumes that the developer is using entitlements, which is encouraged but not required. 
We should check nonSubscriptionTransactions as well

aboedo (changes requested):
let's play it safe and generalize to all transactions, not just active entitlements

RevenueCatUI/Views/FooterView.swift
NachoSoto: Good point.

RevenueCatUI/Views/FooterView.swift
```diff
@@ -167,8 +167,10 @@ private struct RestorePurchasesButton: View {
 
     var body: some View {
         AsyncButton {
-            _ = try await self.purchaseHandler.restorePurchases()
-            self.displayRestoredAlert = true
+            let success = try await self.purchaseHandler.restorePurchases().success
+            if success {
+                self.displayRestoredAlert = true
+            }
```
aboedo, line 173: this looks good to me as a stopgap, but I think we should try to get copy for saying something about "no purchases found" and localize it. 
No-op is kind of an awkward experience to me

aboedo (approved):

RevenueCatUI/Views/FooterView.swift
NachoSoto: Filed a Linear for this.

PR #3239, opened 2023-09-21
Make revisionID private in PaywallData+Default

vegaro:
Noticed this property can be private


NachoSoto:
Thanks! You'll need to rebase from the latest `main` to fix Xcode 15 tests :)

NachoSoto (approved):

NachoSoto:
(I changed the tags since this isn't really a "fix" that needs to be called out to devs, just an internal refactor)

PR #3243, opened 2023-09-21
`Paywalls`: renamed `SimpleApp` to `PaywallsTester`

NachoSoto:
The name was picked before we were ready to announce RevenueCat paywalls publicly.

This is a better name for what's become of this app now, and it will help to reference it from our docs.


Tests/TestingApps/PaywallsTester/ci_scripts/ci_pre_xcodebuild.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash -e
```
NachoSoto, line 1: Gotta make sure that I update `Xcode Cloud` to make it work with the new project.

aboedo (approved):

PR #3245, opened 2023-09-22
[WIP] `Paywalls`: polished default margins

NachoSoto:
### Before:
image
image

### After:
image
image

### Comparison with design:
image
image


NachoSoto:
I'm not a fan, and we probably won't go with this, but there's a few refactors in here with how the padding is distributed that I want to keep.

PR #3246, opened 2023-09-22
`Paywalls`: created `ConsistentPackageContentView` to improve package change transitions

NachoSoto:
This abstracts away the logic to ensure layouts are consistent when selecting different packages in multi-package templates, in a way that the content doesn't move around.

### DIFF
--- a/RevenueCatUI/Data/IntroEligibility/IntroEligibilityViewModel.swift
+++ b/RevenueCatUI/Data/IntroEligibility/IntroEligibilityViewModel.swift
@@ -42,6 +42,7 @@ extension IntroEligibilityViewModel {
         switch packages {
         case let .single(package):
             self.singleEligibility = await self.introEligibilityChecker.eligibility(for: package.content)
+            self.allEligibility[package.content] = self.singleEligibility
 
         case let .multiple(_, _, packages):
             self.allEligibility = await self.introEligibilityChecker.eligibility(for: packages.map(\.content))
--- a/RevenueCatUI/Modifiers/ConsistentPackageContentView.swift
+++ b/RevenueCatUI/Modifiers/ConsistentPackageContentView.swift
@@ -1,0 +1,48 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  ConsistentPackageContentView.swift
+//
+//  Created by Nacho Soto on 9/22/23.
+//
+
+import SwiftUI
+
+/// A wrapper view that can display content based on a selected package
+/// and maintain a consistent layout when that selected package changes.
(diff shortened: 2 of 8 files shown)


RevenueCatUI/Templates/Template4View.swift
```diff
@@ -55,8 +55,6 @@ struct Template4View: TemplateViewType {
                 Spacer()
 
                 self.footerContent
-                    .edgesIgnoringSafeArea(.bottom)
-                    .frame(maxWidth: .infinity, alignment: .bottom)
```
NachoSoto, line 59: This was unnecessary and it was pushing the content below the safe area.

RevenueCatUI/Templates/Template4View.swift
```diff
@@ -214,9 +257,13 @@ struct Template4View: TemplateViewType {
                 return 1.5
             }
         }()
+        let maximumPackagesToDisplay = 3
```
NachoSoto, line 260: We weren't enforcing a limit, which would have shrunk packages unnecessarily when displaying more than 3, instead of making it scrollable.

RevenueCatUI/Modifiers/ConsistentPackageContentView.swift
```diff
@@ -0,0 +1,48 @@
...
+//
+//  Created by Nacho Soto on 9/22/23.
+//
+
+import SwiftUI
+
+/// A wrapper view that can display content based on a selected package
+/// and maintain a consistent layout when that selected package changes.
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
+struct ConsistentPackageContentView<Content: View>: View {
```
NachoSoto, line 20: I've extracted the logic here so it can be used by all templates.

NachoSoto:
This is not quite ready, there is a bad transition on `PurchaseButton` when switching packages.

RevenueCatUI/Data/IntroEligibility/IntroEligibilityViewModel.swift
```diff
@@ -42,6 +42,7 @@ extension IntroEligibilityViewModel {
         switch packages {
         case let .single(package):
             self.singleEligibility = await self.introEligibilityChecker.eligibility(for: package.content)
+            self.allEligibility[package.content] = self.singleEligibility
```
NachoSoto, line 45: This allows using `allEligibility` even on templates that use only one package.

RevenueCatUI/Modifiers/ConsistentPackageContentView.swift
```diff
@@ -0,0 +1,48 @@
...
+        self.selected = selected
+        self.creator = creator
+    }
+
+    var body: some View {
+        // We need to layout all possible packages to accomodate for the longest text
+        return ZStack {
+            ForEach(self.packages, id: \.self.content) { package in
+                self.creator(package)
+                    .opacity(package.content == self.selected.content ? 1 : 0)
```
aboedo, line 43: 👍

aboedo (approved):

PR #3252, opened 2023-09-27
StoreKit In App messages support

tonidero:
### Description
StoreKit displays different types of in-app messages on app launch, but it provides an API to disable that and show these messages at a different time. This PR wraps those APIs to make them easier to use through our SDK.

#### New APIs
- Add new configuration parameter to defer StoreKit's automatic in-app messages
- Add new method in `Purchases` to display a list of specific message types
- New `StoreKitMessageType` enum mapping to the types used by StoreKit


Sources/Support/StoreKitMessagesHelper.swift
```diff
@@ -0,0 +1,62 @@
...
+
+import StoreKit
+
+class StoreKitMessagesHelper {
+
+    private let systemInfo: SystemInfo
+    private let showStoreKitMessagesAutomatically: Bool
+
+    // TODO: Need to improve this :/
+    private var deferredMessages: [Any] = []
```
MarkVillacampa, line 22: Should this type be `Message` instead of `Any`?

Tests/TestingApps/SimpleApp/SimpleApp.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>IDEDidComputeMac32BitWarning</key>
+	<true/>
+</dict>
+</plist>
```
MarkVillacampa, line 8: Can we remove this file?

Sources/Purchasing/Configuration.swift
```diff
@@ -174,6 +177,13 @@ import Foundation
             return self
         }
 
+        /// Set `showStoreKitMessagesAutomatically`. Enabled by default. If you want to disable, make sure
+        /// you're configuring the SDK during the `didFinishLaunchingWithOptions` delegate call.
```
aboedo, line 181: Let's not mention `didFinishLaunchingWithOptions` here - [it's not the SwiftUI way to do it](https://www.revenuecat.com/docs/configuring-sdk), so it wouldn't apply to a lot of apps. 

Let's instead keep it more high level here, and say "configuring the SDK as early as possible in your app flow, otherwise messages will be displayed automatically". 

Also, I think we could improve this copy significantly - it doesn't really tell you _anything_ about what this is. 

How about: 

```suggestion
        /// Set `showStoreKitMessagesAutomatically`. Enabled by default.
        ///  If enabled, if the user has billing issues, has yet to accept a price increase consent or 
        /// there are other messages from StoreKit, they will be displayed automatically when the app is initialized. 
        /// 
        ///  If you want to disable this behavior so that you can customize when these messages are shown, make sure
        /// you configure the SDK as early as possible in the app's lifetime, 
        /// otherwise messages will be displayed automatically. 
        /// Then use the ``<other method name here>`` method to display the messages. 
        /// More information: <rev.cat link that points [to this page](https://developer.apple.com/documentation/storekit/message)>
```

Sources/Support/StoreKitMessageType.swift
```diff
@@ -0,0 +1,59 @@
...
+//  Created by Antonio Rico Diez on 27/9/23.
+
+import StoreKit
+
+@available(iOS 16.4, *)
+@available(macOS, unavailable)
+@available(watchOS, unavailable)
+@available(tvOS, unavailable)
+/// Types of messages available in StoreKit
+@objc public enum StoreKitMessageType: Int, CaseIterable {
```
aboedo, line 21: we're making these very StoreKit-dependent, what do we gain by abstracting them? 
I'd either: 
- make all of these "StoreMessage" such that if we have a new backing store (cough, DMA, cough) they still work 
- just vend the StoreKit types directly

I guess they're done this way for testing purposes?

Sources/Support/StoreKitMessagesHelper.swift
aboedo: +1

tonidero: Yeah, that's why I have the `TODO`. Since `Message` is iOS 16.0+, I need to make this whole class `@available` that, which was a bit annoying since I needed to pass that around. On the other hand, I could make this like that and make it nullable in the `PurchasesOrchestrator`, which is probably the best way to go for this.

Tests/TestingApps/SimpleApp/SimpleApp.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
tonidero: Oh right, I checked this in accidentally. Thanks!

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,187 @@
...
+
+    @available(iOS 16.4, *)
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(tvOS, unavailable)
+    func deferMessagesIfNeeded() async throws {
+        guard !self.showStoreMessagesAutomatically else {
+            return
+        }
+        Task(priority: .background) {
```
tonidero, line 51: I moved this inside a background task. Otherwise this will keep listening for messages forever, getting stuck. This did cause some issues in the tests though, since we have to wait until the messages are cached before showing them.

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -0,0 +1,164 @@
...
+    @available(iOS 16.4, *)
+    private func waitForDeferredMessages(messages: [StoreMessage]) async throws {
+        try await self.helper.deferMessagesIfNeeded()
+
+        try await Task.sleep(nanoseconds: DispatchTimeInterval.milliseconds(50).nanoseconds)
+
+        for message in messages {
+            self.storeMessagesProvider.updatesPublisher.send(message)
+
+            try await Task.sleep(nanoseconds: DispatchTimeInterval.milliseconds(50).nanoseconds)
```
tonidero, line 113: I really didn't like this... But without this, and since caching the messages is asynchronous, it would cause us to miss the messages, failing the tests... Any other suggestions are welcome

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,187 @@
...
+
+    #endif
+}
+
+// @unchecked because:
+// - Class is not `final` (it's mocked). This implicitly makes subclasses `Sendable` even if they're not thread-safe.
+// - It has mutable `_deferredMessages` which is necessary due to the availability annotations.
+extension StoreMessagesHelper: @unchecked Sendable {}
+
+protocol StoreMessagesProvider {
```
tonidero, line 85: Had to move a lot of stuff behind protocols in order to be able to test it, which made the code much more complex IMO... Honestly, I kinda miss `mockk` even with its defects.

tonidero (commented):
Still need to add some unit tests and test it on a device, but I would appreciate feedback on these questions @aboedo @MarkVillacampa

tonidero (commented):
I was able to test and confirm that it works on sandbox. Lmk if there is anything weird... I did run into a lot of issues with compatibilities everywhere 😓

Sources/Support/StoreMessagesHelper.swift
MarkVillacampa: We may want to perform the append in the main thread just to be safe since we're accessing it from the main actor in `showStoreMessages`. Something like:

```
await MainActor.run {
    self.deferredMessages.append(message)
}
```

MarkVillacampa: Also, I think there may be a retain cycle here, we can break it by making self weak inside the task:

`Task(priority: .background) { [weak self]`

and then calling it optionally:

`self?.deferredMessages.append(message)`

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
MarkVillacampa: We maaay be able to do something with a MockAsyncSequence with static values but I don't have a lot of experience with it and don't know how it would work

https://github.com/RevenueCat/purchases-ios/blob/main/Tests/UnitTests/Mocks/MockAsyncSequence.swift

tonidero: I tried but couldn't get it to work :(

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -134,6 +137,14 @@ final class PurchasesOrchestrator {
         }
 
         Task {
+            #if os(iOS)
+
+            if #available(iOS 16.4, *) {
+                try? await storeMessagesHelper.deferMessagesIfNeeded()
```
aboedo: I suspect this is why it's not working in tests, we need to also check swift version here

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -134,6 +137,14 @@ final class PurchasesOrchestrator {
         }
 
         Task {
+            #if os(iOS)
```
aboedo, line 140: I can't remember what the latest was, does this include catalyst and visionOS?

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,191 @@
...
+
+    @available(iOS 16.4, *)
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(tvOS, unavailable)
+    func deferMessagesIfNeeded() async throws {
+        guard !self.showStoreMessagesAutomatically else {
+            return
+        }
+        Task(priority: .background) { [weak self] in
```
aboedo, line 50: not sure this should have low priority, since if this takes long to execute we won't start listening in time, right?

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,191 @@
...
+            guard let storeMessagesProvider = self?.storeMessagesProvider else {
+                return
+            }
+            for try await message in storeMessagesProvider.messages {
+                await MainActor.run { [weak self] in
+                    self?.deferredMessages.append(message)
```
aboedo, line 56: why do these need to be in the main actor?

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,191 @@
...
+
+    @available(iOS 16.4, *)
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(tvOS, unavailable)
+    @MainActor
+    func showStoreMessages(types: Set<StoreMessageType>) {
+        for message in self.deferredMessages {
+            if let messageType = message.reason.messageType {
+                if types.contains(messageType) {
```
aboedo, line 70: you can combine these ifs into one with a comma like 
```suggestion
            if let messageType = message.reason.messageType,
               if types.contains(messageType) {
```

Sources/Support/StoreMessagesHelper.swift
aboedo: hoping that macros and the new testing direction eventually solve this

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -0,0 +1,165 @@
...
+        try await waitForDeferredMessages(messages: [message1, message2])
+
+        await self.helper.showStoreMessages(types: [StoreMessageType.billingIssue])
+
+        expect(message1.displayCalled) == true
+        expect(message2.displayCalled) == false
+    }
+
+    @available(iOS 16.4, *)
+    func testShowMessagesAfterDeferMessagesAndShowingMessagesAutomaticallyDoesNotShowsMessages() async throws {
```
aboedo, line 72: typo
```suggestion
    func testShowMessagesAfterDeferMessagesAndShowingMessagesAutomaticallyDoesNotShowMessages() async throws {
```

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
aboedo: @NachoSoto thoughts?

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -0,0 +1,165 @@
...
+
+    let reason: Message.Reason
+
+    init(reason: Message.Reason) {
+        self.reason = reason
+    }
+
+    var displayCalled = false
+    var displayCallCount = 0
+    @MainActor func display(in scene: UIWindowScene) throws {
```
aboedo, line 129: nitpick, but I think our style guide says to have a line break in between these
```suggestion
    var displayCallCount = 0
    
    @MainActor func display(in scene: UIWindowScene) throws {
```

aboedo (commented):
looking good! left a few minor comments

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
tonidero: Tried that in https://github.com/RevenueCat/purchases-ios/pull/3252/commits/67895d1bec5102400f14a9f3722d3d4d84e83e5b but didn't work. Will try to download an older version of Xcode to try to repro this

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
tonidero: I believe visionOS is not included. Will add it 👍

Sources/Support/StoreMessagesHelper.swift
tonidero: Mostly to make sure we only access `deferredMessages` from the main actor and avoid any concurrency issues.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
tonidero: Actually, I'm not sure if catalyst is included, since it seems it's added only in some places, even for APIs that seem supported there (like `beginRefundRequest` APIs)... Will try to figure that out.

Sources/Support/StoreMessagesHelper.swift
tonidero: Hmm that's a good point... Will increase it to medium priority 👍

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,196 @@
...
+    @available(iOS 16.4, *)
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(tvOS, unavailable)
+    func deferMessagesIfNeeded() async throws {
+        guard !self.showStoreMessagesAutomatically else {
+            return
+        }
+        // Compilation fails in older xcode versions with a similar issue to https://github.com/apple/swift/issues/57379
+        #if swift(>=5.8)
```
tonidero, line 51: This is a hack to fix some compilation issues in Xcode <=14.2. Without this, it fails with a bunch of errors like `call can throw but is not marked with 'try'` in a bunch of unrelated files: https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/15473/workflows/e249ae88-bffe-4482-8b75-6a2194aff297/jobs/129352

This is not ideal, but don't want to block this from shipping anymore. Please lmk if you have other ideas @aboedo @NachoSoto @MarkVillacampa

Sources/Logging/Strings/ConfigureStrings.swift
```diff
@@ -75,6 +75,8 @@ enum ConfigureStrings {
 
     case sk2_required_for_swiftui_paywalls
 
+    case could_not_defer_store_messages(errorMessage: String)
```
NachoSoto, line 78: To avoid passing a random string here, we typically make these take an Error and format it inside:
```suggestion
    case could_not_defer_store_messages(Error)
```

Sources/Misc/Concurrency/Purchases+nonasync.swift
```diff
@@ -62,4 +62,48 @@ public extension Purchases {
...
+            completion()
+        }
+    }
+
+    @available(iOS 16.4, *)
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(tvOS, unavailable)
+    func showStoreMessages(
+        forTypes types: Set<StoreMessageType>,
```
NachoSoto, line 84: This should be 
```suggestion
        for types: Set<StoreMessageType>,
```

Sources/Support/StoreMessageType.swift
```diff
@@ -0,0 +1,56 @@
...
+//
+//      https://opensource.org/licenses/MIT
+//
+//  StoreMessageType.swift
+//
+//  Created by Antonio Rico Diez on 27/9/23.
+
+import StoreKit
+
+/// Types of messages available in StoreKit
```
NachoSoto, line 16: Nit: an enum, represents one type
```suggestion
/// Type of messages available in StoreKit
```

Sources/Support/StoreMessageType.swift
```diff
@@ -0,0 +1,56 @@
...
+///
+/// #### Related Symbols
+/// - ``Purchases/showStoreMessages(forTypes:)``
+@objc(RCStoreMessageType) public enum StoreMessageType: Int, CaseIterable, Sendable {
+    /// Message shown when there are billing issues in a subscription
```
NachoSoto, line 21: Nit: types need a space before and after (see the style guide):
```suggestion

    /// Message shown when there are billing issues in a subscription
```

Sources/Support/StoreMessageType.swift
```diff
@@ -0,0 +1,56 @@
...
+}
+
+#if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+
+@available(iOS 16.4, *)
+@available(macOS, unavailable)
+@available(watchOS, unavailable)
+@available(tvOS, unavailable)
+extension Message.Reason {
+    var messageType: StoreMessageType? {
```
NachoSoto, line 40: Ditto: space before and after:
```suggestion

    var messageType: StoreMessageType? {
```

Sources/Support/StoreMessageType.swift
```diff
@@ -0,0 +1,56 @@
...
+    case generic
+
+    var numberValue: NSNumber {
+        return NSNumber(value: self.rawValue)
+    }
+}
+
+#if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+
+@available(iOS 16.4, *)
```
NachoSoto, line 35: Shouldn't this be 16.0?

Sources/Support/StoreMessageType.swift
```diff
@@ -0,0 +1,56 @@
...
+/// - ``Purchases/showStoreMessages(forTypes:)``
+@objc(RCStoreMessageType) public enum StoreMessageType: Int, CaseIterable, Sendable {
+    /// Message shown when there are billing issues in a subscription
+    case billingIssue = 0
+    /// Message shown when there is a price increase in a subscription that requires consent
+    case priceIncreaseConsent
+    /// Generic Store messages
+    case generic
+
+    var numberValue: NSNumber {
```
NachoSoto, line 28: Can you explain why you need this?

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+        Task(priority: .medium) { [weak self] in
+            guard let storeMessagesProvider = self?.storeMessagesProvider else {
```
NachoSoto, line 53: There is no possible retain cycle here. Instead of using `weak self` I would capture the provider directly, then you don't need this early return:
```suggestion
        Task(priority: .medium) { [provider = self.storeMessagesProvider] in
```

Sources/Support/StoreMessagesHelper.swift
NachoSoto: Let's make `StoreMessagesHelper` an `actor` then, this type of synchronization is how one would do it _in the old days_. That would simplify this a lot and remove the need for `weak self` and everything.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+    private let storeMessagesProvider: StoreMessagesProvider
+
+    private var deferredMessages: [StoreMessage] = []
+
+    init(systemInfo: SystemInfo,
+         showStoreMessagesAutomatically: Bool) {
+        self.systemInfo = systemInfo
+        self.showStoreMessagesAutomatically = showStoreMessagesAutomatically
+        self.storeMessagesProvider = StoreMessagesProviderWrapper()
+    }
```
NachoSoto, line 29: This constructor is unnecessary right? You already have a default value in the second one.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+         showStoreMessagesAutomatically: Bool,
+         storeMessagesProvider: StoreMessagesProvider = StoreMessagesProviderWrapper()) {
+        self.systemInfo = systemInfo
+        self.showStoreMessagesAutomatically = showStoreMessagesAutomatically
+        self.storeMessagesProvider = storeMessagesProvider
+    }
+
+    #if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+
+    @available(iOS 16.4, *)
```
NachoSoto, line 42: Why 16.4?

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+                    self?.deferredMessages.append(message)
+                }
+            }
+        }
+        #else
+        Logger.error("Defer messages are not enabled when compiling using Xcode <14.2")
+        #endif
+    }
+
+    @available(iOS 16.4, *)
```
NachoSoto, line 67: Ditto, 16.4?

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(tvOS, unavailable)
+    @MainActor
+    func showStoreMessages(types: Set<StoreMessageType>) {
+        for message in self.deferredMessages {
```
NachoSoto, line 73: Does `deferredMessages` never get reset?

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+            if let messageType = message.reason.messageType,
+               types.contains(messageType) {
+                    do {
+                        try message.display(in: self.systemInfo.currentWindowScene)
```
NachoSoto, line 77: Have you tried displaying multiple messages? My bet is that the second one would fail. I wonder if we should just display the first one only.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+                    } catch {
+                        Logger.error("Error displaying StoreKit message: \(error)")
```
NachoSoto, line 79: Can you move this to `StoreKitStrings`?

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+        }
+    }
+
+    #endif
+}
+
+// @unchecked because:
+// - Class is not `final` (it's mocked). This implicitly makes subclasses `Sendable` even if they're not thread-safe.
+// - It has mutable `_deferredMessages` which is necessary due to the availability annotations.
+extension StoreMessagesHelper: @unchecked Sendable {}
```
NachoSoto, line 91: You won't need this hack when you make `StoreMessagesHelper` an `actor`.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+    @available(tvOS, unavailable)
+    @MainActor func display(in scene: UIWindowScene) throws
+
+    #endif
+}
+
+#if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+
+@available(iOS 13.0, *)
+@rethrows protocol StoreMessageAsyncIteratorProtocol: AsyncIteratorProtocol where Element == StoreMessage { }
```
NachoSoto, line 128: Whoa was does this `@rethrows` do here?

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+
+@available(iOS 13.0, *)
+@rethrows protocol StoreMessageAsyncSequence: AsyncSequence where AsyncIterator: StoreMessageAsyncIteratorProtocol { }
+
+@available(iOS 16.0, *)
+@available(macOS, unavailable)
+@available(watchOS, unavailable)
+@available(tvOS, unavailable)
+private final class StoreMessageWrapper: StoreMessage {
```
NachoSoto, line 137: I don't think this is needed, can you make `Message` conform to `StoreMessage` instead?

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+
+    }
+
+    let underlyingSequence: Message.Messages
+
+    func makeAsyncIterator() -> AsyncIterator {
+        AsyncIterator(iterator: underlyingSequence.makeAsyncIterator())
+    }
+
+}
```
NachoSoto, line 178: I think there's an alternative way that doesn't require this manual creation of a sequence. I'll ping you in Slack to help you with that.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+
+#endif
+
+private final class StoreMessagesProviderWrapper: StoreMessagesProvider {
```
NachoSoto, line 182: Ditto, I think there is a way we can get rid of this.

Tests/APITesters/SwiftAPITester/SwiftAPITester/PurchasesAPI.swift
```diff
@@ -177,6 +187,15 @@ private func checkPurchasesSupportAPI(purchases: Purchases) {
         _ = purchases.delegate?.shouldShowPriceConsent
     }
     #endif
+    #if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+    if #available(iOS 16.4, *) {
```
NachoSoto, line 191: Ditto, I think this can be 16.0?

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -0,0 +1,168 @@
...
+import Combine
+import Foundation
+import Nimble
+import StoreKit
+import XCTest
+
+@testable import RevenueCat
+
+#if swift(>=5.8)
+#if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
```
NachoSoto, line 23: 👍🏻

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -0,0 +1,168 @@
...
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+
+        self.systemInfo = MockSystemInfo(finishTransactions: true)
+        self.storeMessagesProvider = MockStoreMessagesProvider()
+    }
+
+    @available(iOS 16.4, *)
+    func testShowMessagesAfterDeferMessagesAndNotShowingMessagesAutomaticallyShowsAllDeferredMessages() async throws {
+        try AvailabilityChecks.iOS16_4APIAvailableOrSkipTest()
```
NachoSoto, line 42: You have this in every test, so it can be moved to `setUpWithError`.

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -0,0 +1,168 @@
...
+        self.createHelper(showStoreMessagesAutomatically: false)
+
+        let message1 = MockStoreMessage(reason: Message.Reason.billingIssue)
+        let message2 = MockStoreMessage(reason: Message.Reason.priceIncreaseConsent)
+
+        try await waitForDeferredMessages(messages: [message1, message2])
```
NachoSoto, line 48: Nit: explicit self
```suggestion
        try await self.waitForDeferredMessages(messages: [message1, message2])
```

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -0,0 +1,168 @@
...
+        try AvailabilityChecks.iOS16_4APIAvailableOrSkipTest()
+        self.createHelper(showStoreMessagesAutomatically: true)
+
+        try await waitForDeferredMessages(messages: [])
+
+        await self.helper.showStoreMessages(types: Set(StoreMessageType.allCases))
+    }
+
+    @available(iOS 16.4, *)
+    private func createHelper(showStoreMessagesAutomatically: Bool) {
```
NachoSoto, line 99: Nit: I recommend moving these helpers to an extension of `StoreMessagesHelperTests` so it's easier to find the tests versus the helper methods.

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -0,0 +1,168 @@
...
+        self.helper = StoreMessagesHelper(systemInfo: self.systemInfo,
+                                          showStoreMessagesAutomatically: showStoreMessagesAutomatically,
+                                          storeMessagesProvider: self.storeMessagesProvider)
+    }
+
+    @available(iOS 16.4, *)
+    private func waitForDeferredMessages(messages: [StoreMessage]) async throws {
+        try await self.helper.deferMessagesIfNeeded()
+
+        try await Task.sleep(nanoseconds: DispatchTimeInterval.milliseconds(50).nanoseconds)
```
NachoSoto, line 109: Why do you need this?

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
NachoSoto: Can you explain why you needed to wait?

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -0,0 +1,168 @@
...
+        self.displayCalled = true
+        self.displayCallCount += 1
+    }
+}
+
+@available(iOS 15.0, *)
+extension AsyncPublisher<PassthroughSubject<StoreMessage, Never>>.Iterator: StoreMessageAsyncIteratorProtocol {}
+
+@available(iOS 15.0, *)
+private struct MockStoreMessagesAsyncSequence: StoreMessageAsyncSequence {
```
NachoSoto, line 141: I think you should be able to use `MockAsyncSequence` instead?

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -0,0 +1,168 @@
...
+        for message in messages {
+            self.storeMessagesProvider.updatesPublisher.send(message)
+
+            try await Task.sleep(nanoseconds: DispatchTimeInterval.milliseconds(50).nanoseconds)
+        }
+    }
+}
+
+@available(iOS 16.0, *)
+private final class MockStoreMessage: @unchecked Sendable, StoreMessage {
```
NachoSoto, line 120: I'd make this an `actor` so you don't need `@unchecked Sendable`.

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -0,0 +1,168 @@
...
+private final class MockStoreMessagesProvider: StoreMessagesProvider {
+
+    let updatesPublisher = PassthroughSubject<StoreMessage, Never>()
+
+    @available(iOS 16.0, *)
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(tvOS, unavailable)
+    var messages: any StoreMessageAsyncSequence {
+        MockStoreMessagesAsyncSequence(publisher: self.updatesPublisher)
```
NachoSoto, line 163: Yup, this could simply be `MockAsyncSequence` I think.

Tests/StoreKitUnitTests/TestHelpers/AvailabilityChecks.swift
```diff
@@ -55,6 +55,12 @@ enum AvailabilityChecks {
         }
     }
 
+    static func iOS16_4APIAvailableOrSkipTest() throws {
```
NachoSoto, line 58: Probably don't need it since I think we can use 16.0 for everything (other than `billingIssue`)

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -0,0 +1,168 @@
...
+import Nimble
+import StoreKit
+import XCTest
+
+@testable import RevenueCat
+
+#if swift(>=5.8)
+#if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+
+@available(iOS 15.0, *)
```
NachoSoto, line 25: Since you're not testing this below 16.0, I'd make this available only on 16.0.

Sources/Purchasing/Configuration.swift
```diff
@@ -174,6 +177,20 @@ import Foundation
             return self
         }
 
+        /// Set `showStoreMessagesAutomatically`. Enabled by default.
+        /// If enabled, if the user has billing issues, has yet to accept a price increase consent or
+        /// there are other messages from StoreKit, they will be displayed automatically when the app is initialized.
+        ///
+        /// If you want to disable this behavior so that you can customize when these messages are shown, make sure
+        /// you configure the SDK as early as possible in the app's lifetime, otherwise messages will be displayed
```
NachoSoto, line 185: Can you explain this part? I don't think that's important, because the SDK won't display them until it's initialized?

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -736,6 +736,41 @@ public protocol PurchasesType: AnyObject {
 
     #endif
 
+    #if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+
+    /**
+     * Displays all store in-app message types to the user if there are any available to be shown.
+     * - Important: This should only be used if you disabled these messages from showing automatically
```
NachoSoto, line 743: Love it.

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -736,6 +736,41 @@ public protocol PurchasesType: AnyObject {
...
+     * during SDK configuration using ``Configuration/Builder/with(showStoreMessagesAutomatically:)``
+     *
+     * ### Related Symbols
+     * - ``Purchases/showStoreMessages(forRawValues:)``
+     * - ``Purchases/showStoreMessages(forTypes:)``
+     */
+    @available(iOS 16.4, *)
```
NachoSoto, line 750: Ditto 16.0

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -736,6 +736,41 @@ public protocol PurchasesType: AnyObject {
...
+     * Displays the specified store in-app message types as raw values to the user if there are any
+     * available to be shown. Obj-C compatible version of this method.
+     * - Important: This should only be used if you disabled these messages from showing automatically
+     * during SDK configuration using ``Configuration/Builder/with(showStoreMessagesAutomatically:)``
+     *
+     * ### Related Symbols
+     * - ``Purchases/showStoreMessages()``
+     * - ``Purchases/showStoreMessages(forTypes:)``
+     */
+    @available(iOS 16.4, *)
```
NachoSoto, line 766: Ditto 16.0

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -736,6 +736,41 @@ public protocol PurchasesType: AnyObject {
...
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(tvOS, unavailable)
+    func showStoreMessages(forRawValues rawValues: Set<NSNumber>) async
```
NachoSoto, line 770: I'm still confused about why this is needed, since `Message.Reason` isn't available in Obj-C anyway.

Sources/Support/StoreMessagesHelper.swift
NachoSoto: We need to keep supporting 14.1+ though, otherwise this would silently fail.

Let me compile this locally with Xcode 14.2 and see why it's failing.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,195 @@
...
+
+protocol StoreMessagesProvider {
+
+    #if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+
+    @available(iOS 16.0, *)
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(tvOS, unavailable)
+    var messages: any StoreMessageAsyncSequence { get }
```
NachoSoto, line 101: You should be able to get rid of these async sequence types by changing this to `AsyncStream<StoreKit.Message>`. That's essentially type-erasing the sequence, and it would simplify the code a lot. Let me know if I can help with that.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1289,7 +1334,9 @@ public extension Purchases {
                   storeKit2Setting: storeKit2Setting,
                   storeKitTimeout: storeKitTimeout,
                   networkTimeout: networkTimeout,
-                  dangerousSettings: dangerousSettings)
+                  dangerousSettings: dangerousSettings,
+                  showStoreMessagesAutomatically: showStoreMessagesAutomatically
+                 )
```
NachoSoto, line 1339: Super-nit: since the opening `(` is in the same line as the first parameter, this should match and be on the same line as the last.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -134,6 +137,19 @@ final class PurchasesOrchestrator {
         }
 
         Task {
+            #if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+            #if swift(>=5.8)
+            if #available(iOS 16.4, *) {
```
NachoSoto, line 142: Ditto so we don't forget
```suggestion
            if #available(iOS 16.0, *) {
```

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -134,6 +137,19 @@ final class PurchasesOrchestrator {
         }
 
         Task {
+            #if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+            #if swift(>=5.8)
```
NachoSoto, line 141: Is this needed?

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -134,6 +137,19 @@ final class PurchasesOrchestrator {
...
+            if #available(iOS 16.4, *) {
+                do {
+                    try await storeMessagesHelper.deferMessagesIfNeeded()
+                } catch {
+                    Logger.error(Strings.configure.could_not_defer_store_messages(errorMessage:
```
NachoSoto, line 146: I don't think this is a configuration error. I'd move it to `StoreKitStrings`.

NachoSoto (commented):
Great job with this 👏🏻 
The overall setup is perfect, just some recommendations to simplify the code.

NachoSoto:
Addressed a lot of these in #3264

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
tonidero: Since we were changing threads/tasks to a bakckground task and then back to the main actor, if I didn't sleep here, the test ended before it had the chance to move the receive the message. I think that's fixed with your PR though

Sources/Support/StoreMessageType.swift
tonidero: I think you figured it out already but did this since `billingIssue` was added in 16.4. I see you already fixed it though.

Sources/Support/StoreMessageType.swift
tonidero: Right, I didn't think about `intValue` 😓. Can probably be removed now. thanks!

Sources/Support/StoreMessagesHelper.swift
tonidero: Made this change here: https://github.com/RevenueCat/purchases-ios/pull/3252/commits/7a8586f8d478fc85411f247b4f3d7fea1708613b

Sources/Support/StoreMessagesHelper.swift
tonidero: It wasn't. After refactoring to make it only display the first message, I changed it though: https://github.com/RevenueCat/purchases-ios/pull/3252/commits/7a8586f8d478fc85411f247b4f3d7fea1708613b

Sources/Purchasing/Configuration.swift
MarkVillacampa: By default, if you don't do anything, iOS will show the messages. iOS stops showing these messages if you start listening for messages, but if you don't do it early enough in the app's lifecycle, it's possible iOS may have already shown some messages. This is why they recommend setting up the listener as early as possible: `If you want to defer or suppress App Store messages, set up the message listener in your app when your app launches`.

https://developer.apple.com/documentation/storekit/message

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -973,6 +1008,24 @@ public protocol PurchasesSwiftType: AnyObject {
...
+     * - ``Purchases/showStoreMessages()``
+     * - ``Purchases/showStoreMessages(for:)-79d0o``
+     */
+    @available(iOS 16.0, *)
+    @available(macOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(tvOS, unavailable)
+    func showStoreMessages(for types: Set<StoreMessageType>) async
+
+    #endif
```
MarkVillacampa, line 1027: nitpick but why not put this method alongside the others above? would save an #if

tonidero: This extension is separated since it's for swift only methods. Since this method is for swift only, we need to move it here unfortunately.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -0,0 +1,131 @@
...
+
+private final class StoreMessagesProvider: StoreMessagesProviderType {
+
+    #if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+
+    @available(iOS 16.0, *)
+    var messages: AsyncStream<StoreMessage> {
+        return Message.messages
+            .map { $0 as StoreMessage }
+            .toAsyncStream()
```
MarkVillacampa, line 127: Nice!

Sources/Support/StoreMessagesHelper.swift
aboedo: @NachoSoto could you elaborate? Why would you expect the second one to fail? 

Apple's docs suggest this as the way to do it https://developer.apple.com/documentation/storekit/displaymessageaction

And it seems like others are already doing it like that
https://github.com/brave/brave-ios/blob/509191082e749fd6e96b87e3172b4528e4eabc98/Sources/Brave/Frontend/Browser/BrowserViewController/BrowserViewController%2BCallout.swift#L273

tonidero: Reverted the changes so we display all messages again: https://github.com/RevenueCat/purchases-ios/pull/3252/commits/b65d22d5c35ee3732db27a5b8c580526e7983a5c

aboedo: I think that's the safest unless we have a clear reason why we know it would fail. This is pretty hard to test in practice, so following Apple's example as close as we can seems like a good bet

MarkVillacampa (approved):

tonidero:
@NachoSoto @aboedo going to merge this and make a release with this. If there are any concerns, I can address them later

PR #3255, opened 2023-09-27
PaywallsTester: improve navigation on macOS and iPadOS

aboedo:
The navigation is slightly awkward because we're rendering a list on the left side, but then we display a sheet, so the right side pane is never filled. This just updates it so that it always looks like a full screen list instead. The changes don't apply to iPhone, just iPad landscape and macOS

| Before | After |
| :-: | :-: |
| Screenshot 2023-09-27 at 5 53 31 PM | Screenshot 2023-09-27 at 6 06 55 PM |


tonidero (approved):

NachoSoto (commented):
<3 I've been meaning to do this, thank you!

PR #3256, opened 2023-09-27
PaywallsTester: replace opening default paywall automatically with button

aboedo:
This one is a total pet peeve, but it was annoying to me that every time you open the PaywallsTester it opens up the modal, since most of the time I'm not actually testing the default paywall. 

This replaces that behavior with an explicit button for it, but it otherwise retains the rest of the behavior. 

Happy to close this one out if others preferred the default paywall opening up by itself every time. 

IMG_3F18253B39B7-1

### DIFF
--- a/Tests/TestingApps/PaywallsTester/PaywallsTester/Views/AppContentView.swift
+++ b/Tests/TestingApps/PaywallsTester/PaywallsTester/Views/AppContentView.swift
@@ -27,7 +27,7 @@ struct AppContentView: View {
     private var customerInfo: CustomerInfo?
 
     @State
-    private var didPurchase: Bool = false
+    private var showingDefaultPaywall: Bool = false
 
     var body: some View {
         TabView {
@@ -42,6 +42,7 @@ struct AppContentView: View {
                 .tabItem {
                     Label("App", systemImage: "iphone")
                 }
+                .navigationViewStyle(StackNavigationViewStyle())
             }
 
             #if DEBUG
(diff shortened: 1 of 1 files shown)


Tests/TestingApps/PaywallsTester/PaywallsTester/Views/AppContentView.swift
```diff
@@ -58,11 +59,6 @@ struct AppContentView: View {
                     }
             }
         }
-        .presentPaywallIfNeeded {
-            !$0.hasPro
```
tonidero, line 62: Are we testing this somewhere else? To make sure it works as expected, not sure if we could move it somewhere else...

tonidero (approved):
I have once concern about not testing the `presentPaywallIfNeeded` modiifer anywhere with this... But in general it's a better behavior yeah... I do think it's a bit weird to show the paywall always

Tests/TestingApps/PaywallsTester/PaywallsTester/Views/AppContentView.swift
aboedo: gonna to it as a separate PR

Tests/TestingApps/PaywallsTester/PaywallsTester/Views/AppContentView.swift
```diff
@@ -80,7 +76,7 @@ struct AppContentView: View {
                 Text(verbatim: "You're signed in: \(info.originalAppUserId)")
                     .font(.callout)
 
-                if self.didPurchase {
+                if self.customerInfo?.activeSubscriptions.count ?? 0 > 0 {
```
NachoSoto, line 79: This would be simpler as
```
self.customerInfo?.activeSubscriptions.isEmpty == false
```

PR #3257, opened 2023-09-27
PaywallsTester: fix StoreKit Configuration scheme

aboedo:
The SK config file scheme was broken after the rename from SimpleApp to PaywallsTester, this fixes it.


tonidero (approved):

NachoSoto:
Oh thanks great catch!

PR #3258, opened 2023-09-27
PaywallsTester: allow resizing on macOS

aboedo:
Refactored so that in mac catalyst, instead of presenting as a sheet, we present as part of navigation, so that resizing the view also resizes the paywall. 

This is useful for quick testing of different sizes on macOS.

iOS remains unchanged

https://github.com/RevenueCat/purchases-ios/assets/3922667/8de24c5d-80ef-4ed6-8813-eca72af00008

### DIFF
--- a/Tests/TestingApps/PaywallsTester/PaywallsTester/Views/OfferingsList.swift
+++ b/Tests/TestingApps/PaywallsTester/PaywallsTester/Views/OfferingsList.swift
@@ -41,23 +41,13 @@ struct OfferingsList: View {
         switch self.offerings {
         case let .success(offerings):
             self.list(with: offerings)
+            #if !targetEnvironment(macCatalyst)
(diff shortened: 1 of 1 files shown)


Tests/TestingApps/PaywallsTester/PaywallsTester/Views/OfferingsList.swift
```diff
@@ -41,23 +41,13 @@ struct OfferingsList: View {
         switch self.offerings {
         case let .success(offerings):
             self.list(with: offerings)
+            #if !targetEnvironment(macCatalyst)
```
tonidero, line 44: Hmm I wonder if we should provide the option of presenting as a sheet somewhere... But I guess we can do that later. This does provide more flexibility 👍

tonidero (approved):

PR #3264, opened 2023-10-02
Refactored `StoreMessagesHelper`

NachoSoto:
Improvements for #3252

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -399,12 +399,16 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
 
         let storeMessagesHelper: StoreMessagesHelperType?
 
+        #if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
         if #available(iOS 16.0, *) {
             storeMessagesHelper = StoreMessagesHelper(systemInfo: systemInfo,
                                                       showStoreMessagesAutomatically: showStoreMessagesAutomatically)
         } else {
             storeMessagesHelper = nil
         }
+        #else
+        storeMessagesHelper = nil
+        #endif
 
         let purchasesOrchestrator: PurchasesOrchestrator = {
             if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1083,8 +1083,8 @@ public extension Purchases {
     @available(watchOS, unavailable)
     @available(tvOS, unavailable)
     func showStoreMessages(forRawValues rawValues: Set<NSNumber>) async {
-        let storeMessageTypes = rawValues.map { number in
-            StoreMessageType(rawValue: number.intValue) ?? .generic
+        let storeMessageTypes = rawValues.compactMap { number in
+            StoreMessageType(rawValue: number.intValue)
         }
         await self.storeMessagesHelper?.showStoreMessages(types: Set(storeMessageTypes))
     }
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -254,7 +254,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
     private let requestFetcher: StoreKitRequestFetcher
     private let paymentQueueWrapper: EitherPaymentQueueWrapper
     private let systemInfo: SystemInfo
-    private let storeMessagesHelper: StoreMessagesHelper
+    private let storeMessagesHelper: StoreMessagesHelperType
     private var customerInfoObservationDisposable: (() -> Void)?
 
     // swiftlint:disable:next function_body_length
@@ -517,7 +517,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
          purchasesOrchestrator: PurchasesOrchestrator,
          purchasedProductsFetcher: PurchasedProductsFetcherType?,
          trialOrIntroPriceEligibilityChecker: CachingTrialOrIntroPriceEligibilityChecker,
-         storeMessagesHelper: StoreMessagesHelper
+         storeMessagesHelper: StoreMessagesHelperType
     ) {
 
         if systemInfo.dangerousSettings.customEntitlementComputation {
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -254,7 +254,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
     private let requestFetcher: StoreKitRequestFetcher
     private let paymentQueueWrapper: EitherPaymentQueueWrapper
     private let systemInfo: SystemInfo
-    private let storeMessagesHelper: StoreMessagesHelperType
+    private let storeMessagesHelper: StoreMessagesHelperType?
     private var customerInfoObservationDisposable: (() -> Void)?
 
     // swiftlint:disable:next function_body_length
@@ -397,8 +397,14 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
                                                                 customerInfoManager: customerInfoManager,
                                                                 currentUserProvider: identityManager)
 
-        let storeMessagesHelper = StoreMessagesHelper(systemInfo: systemInfo,
+        let storeMessagesHelper: StoreMessagesHelperType?
+
+        if #available(iOS 16.0, *) {
+            storeMessagesHelper = StoreMessagesHelper(systemInfo: systemInfo,
                                                       showStoreMessagesAutomatically: showStoreMessagesAutomatically)
+        } else {
+            storeMessagesHelper = nil
+        }
 
         let purchasesOrchestrator: PurchasesOrchestrator = {
             if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
@@ -517,7 +523,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
          purchasesOrchestrator: PurchasesOrchestrator,
          purchasedProductsFetcher: PurchasedProductsFetcherType?,
          trialOrIntroPriceEligibilityChecker: CachingTrialOrIntroPriceEligibilityChecker,
-         storeMessagesHelper: StoreMessagesHelperType
+         storeMessagesHelper: StoreMessagesHelperType?
     ) {
 
         if systemInfo.dangerousSettings.customEntitlementComputation {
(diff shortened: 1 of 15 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -136,20 +136,21 @@ final class PurchasesOrchestrator {
             storeKit2StorefrontListener.listenForStorefrontChanges()
         }
 
-        Task {
-            #if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
-            #if swift(>=5.8)
-            if #available(iOS 16.4, *) {
+        #if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
+        if #available(iOS 16.4, *) {
+            Task {
```
NachoSoto, line 141: It's important that this happens in 2 separate `Tasks` so that this doesn't block `storeKit2TransactionListener`

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -13,25 +13,37 @@
 
 import StoreKit
 
-class StoreMessagesHelper {
+protocol StoreMessagesHelperType {
```
NachoSoto, line 16: Extracted a `protocol` for this so we don't do `@unchecked Sendable` hacks.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -13,25 +13,37 @@
...
+actor StoreMessagesHelper: StoreMessagesHelperType {
 
     private let systemInfo: SystemInfo
     private let showStoreMessagesAutomatically: Bool
-    private let storeMessagesProvider: StoreMessagesProvider
+    private let storeMessagesProvider: StoreMessagesProviderType
 
     private var deferredMessages: [StoreMessage] = []
 
-    init(systemInfo: SystemInfo,
```
NachoSoto, line 24: Like I pointed out, this extra constructor wasn't needed.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -47,62 +59,47 @@ class StoreMessagesHelper {
         guard !self.showStoreMessagesAutomatically else {
             return
         }
-        // Compilation fails in older xcode versions with a similar issue to https://github.com/apple/swift/issues/57379
-        #if swift(>=5.8)
-        Task(priority: .medium) { [weak self] in
```
NachoSoto, line 52: This didn't need to create a `Task`.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -47,62 +59,47 @@ class StoreMessagesHelper {
...
-                return
-            }
-            for try await message in storeMessagesProvider.messages {
-                await MainActor.run { [weak self] in
-                    self?.deferredMessages.append(message)
-                }
-            }
+
+        for try await message in self.storeMessagesProvider.messages {
+            self.deferredMessages.append(message)
```
NachoSoto, line 64: Synchronization guaranteed through the `actor` now.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -47,62 +59,47 @@ class StoreMessagesHelper {
...
-        #endif
     }
 
     @available(iOS 16.4, *)
     @available(macOS, unavailable)
     @available(watchOS, unavailable)
     @available(tvOS, unavailable)
-    @MainActor
-    func showStoreMessages(types: Set<StoreMessageType>) {
+    func showStoreMessages(types: Set<StoreMessageType>) async {
```
NachoSoto, line 72: This didn't need to be `@MainActor`. Since the method is already `async`, we can `await` on the call to `display(in:)`.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -47,62 +59,47 @@ class StoreMessagesHelper {
...
+protocol StoreMessagesProviderType {
 
     #if os(iOS) || targetEnvironment(macCatalyst) || VISION_OS
 
     @available(iOS 16.0, *)
     @available(macOS, unavailable)
     @available(watchOS, unavailable)
     @available(tvOS, unavailable)
-    var messages: any StoreMessageAsyncSequence { get }
+    var messages: AsyncStream<StoreMessage> { get }
```
NachoSoto, line 97: No more custom sequence types.

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -124,71 +121,26 @@ protocol StoreMessage: Sendable {
...
     @available(iOS 16.0, *)
     @available(macOS, unavailable)
     @available(watchOS, unavailable)
     @available(tvOS, unavailable)
-    var messages: any StoreMessageAsyncSequence {
-        return StoreMessageSequence(underlyingSequence: Message.messages)
+    var messages: AsyncStream<StoreMessage> {
+        return Message.messages
+            .map { $0 as StoreMessage }
+            .toAsyncStream()
```
NachoSoto, line 143: Using `toAsyncStream` to create a custom stream.

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -104,65 +103,47 @@ class StoreMessagesHelperTests: TestCase {
...
-
-            try await Task.sleep(nanoseconds: DispatchTimeInterval.milliseconds(50).nanoseconds)
-        }
+        try await self.helper.deferMessagesIfNeeded()
     }
 }
 
 @available(iOS 16.0, *)
-private final class MockStoreMessage: @unchecked Sendable, StoreMessage {
+private final class MockStoreMessage: StoreMessage {
```
NachoSoto, line 113: Made this type thread safe instead.

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -104,65 +103,47 @@ class StoreMessagesHelperTests: TestCase {
...
+    var stubbedMessages: [StoreMessage] = []
 
     @available(iOS 16.0, *)
     @available(macOS, unavailable)
     @available(watchOS, unavailable)
     @available(tvOS, unavailable)
-    var messages: any StoreMessageAsyncSequence {
-        MockStoreMessagesAsyncSequence(publisher: self.updatesPublisher)
+    var messages: AsyncStream<StoreMessage> {
+        MockAsyncSequence(with: self.stubbedMessages).toAsyncStream()
```
NachoSoto, line 145: Creating a fake sequence this way

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -104,65 +103,47 @@ class StoreMessagesHelperTests: TestCase {
 
     @available(iOS 16.4, *)
     private func waitForDeferredMessages(messages: [StoreMessage]) async throws {
-        try await self.helper.deferMessagesIfNeeded()
-
-        try await Task.sleep(nanoseconds: DispatchTimeInterval.milliseconds(50).nanoseconds)
```
NachoSoto, line 109: No more sleep hacks (cc @aboedo since you were asking)

tonidero (approved):
Nice! This all makes sense to me... I tried to use the `AsyncStream` at one point but I did it wrong and ended up going in the complex path 😓. This makes sense! Looks like the tests are failing due to some compatibility issues but I can take a look tomorrow from this. Many thanks!

aboedo:
what's the story with tests?

NachoSoto:
I just fixed the last issue (for some reason it doesn't fail when compiling locally): `actor` doesn't support pre-iOS 13.

I did some more refactors to fix the compatibility and lowered everything to 16.0 as well.

Sources/Support/StoreMessageType.swift
```diff
@@ -28,29 +29,35 @@ import StoreKit
...
         case .priceIncreaseConsent: return .priceIncreaseConsent
         case .generic: return .generic
         default:
+            // billingIssue message reason was added in iOS 16.4, but it's not recognized by older xcode versions.
+            // https://developer.apple.com/documentation/xcode-release-notes/xcode-14_3-release-notes
+            #if swift(>=5.8)
+            if #available(iOS 16.4, *), case .billingIssue = self {
+                return .billingIssue
+            }
+            #endif
```
NachoSoto, line 54: I refactored this to allow using 16.0 everywhere @tonidero 
This is unnecessarily complex because `Message.Reason` isn't an `enum`.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1064,31 +1070,31 @@ public extension Purchases {
...
+    @available(iOS 16.0, *)
     @available(macOS, unavailable)
     @available(watchOS, unavailable)
     @available(tvOS, unavailable)
     func showStoreMessages(forRawValues rawValues: Set<NSNumber>) async {
-        let storeMessageTypes = rawValues.map({ number in
-            StoreMessageType(rawValue: number.intValue) ?? StoreMessageType.billingIssue
-        })
-        await self.storeMessagesHelper.showStoreMessages(types: Set(storeMessageTypes))
+        let storeMessageTypes = rawValues.compactMap { number in
```
NachoSoto, line 1086: @tonidero I changed this because I don't think we want to assume the type?

NachoSoto: I'm still confused why we need this raw value API since this isn't available in Obj-C, but we can address that in your PR.

aboedo:
tests didn't run on the last commit?

Sources/Purchasing/Purchases/Purchases.swift
tonidero: The idea behind this was to support Obj-C yeah, specially to support the hybrids still using Obj-C... Why do you say this isn't available in Obj-C? Afaik, the APIs should work there but maybe I'm missing something.

aboedo: > I'm still confused why we need this raw value API since this isn't available in Obj-C
We should still be able to make it available for Obj-C devs, right? Historically we've tried to make StoreKit's Swift-only APIs available in Obj-C through our own API

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
aboedo: good catch

Sources/Support/StoreMessageType.swift
aboedo: clever approach

Sources/Support/StoreMessagesHelper.swift
aboedo: Could you elaborate here and explain why and how this works? Feel like the workaround will be hard to understand for anyone unfamiliar with Sendable. 

Might even be valuable to make it a comment

Sources/Support/StoreMessagesHelper.swift
aboedo: I suppose the value add may have been to set the priority explicitly, which can only be done with a separate Task?

Tests/StoreKitUnitTests/StoreMessagesHelperTests.swift
```diff
@@ -33,136 +32,107 @@ class StoreMessagesHelperTests: TestCase {
...
 
-    @available(iOS 16.4, *)
     func testShowMessagesAfterDeferMessagesAndShowingMessagesAutomaticallyDoesNotShowMessages() async throws {
-        try AvailabilityChecks.iOS16_4APIAvailableOrSkipTest()
         self.createHelper(showStoreMessagesAutomatically: true)
 
-        let message1 = MockStoreMessage(reason: Message.Reason.billingIssue)
-        let message2 = MockStoreMessage(reason: Message.Reason.priceIncreaseConsent)
+        let message1 = MockStoreMessage(reason: .generic)
+        let message2 = MockStoreMessage(reason: .priceIncreaseConsent)
```
aboedo, line 73: given that we have specific separate logic for `billingIssue`, feels like we should have at least one test that explicitly uses it

NachoSoto: Yeah we don't have explicit tests for `Message.Reason.messageType` /cc @tonidero

tonidero: Oh ok, will try to add it in a followup PR

Sources/Support/StoreMessagesHelper.swift
NachoSoto: non-final classes can't be Sendable. We use inheritance for mocking, which eliminates the ability to get automatic Sendable conformance. By using a protocol for mocking that's no longer a factor.

Sources/Support/StoreMessagesHelper.swift
NachoSoto: Oh I see. This method is already `async`. If we want to change the priority we should do it when we call this method in `PurchasesOrchestrator` /cc @tonidero

PR #3269, opened 2023-10-03
Remove ObjC showStoreMessages API

tonidero:
### Description
This removes the extra APIs that were added to support showing store messages in ObjC. We will add these in PHC. If there is a demand for it we can always add it later.

### DIFF
--- a/Sources/Purchasing/Configuration.swift
+++ b/Sources/Purchasing/Configuration.swift
@@ -186,7 +186,7 @@ import Foundation
         /// automatically.
         /// Then use the ``Purchases/showStoreMessages(for:)`` method to display the messages.
         /// More information:  https://rev.cat/storekit-message
-        /// - Important: Set this property only if you're using swift. If you're using ObjC, you won't be able to call
+        /// - Important: Set this property only if you're using Swift. If you're using ObjC, you won't be able to call
         /// the related methods
         @objc public func with(showStoreMessagesAutomatically: Bool) -> Builder {
             self.showStoreMessagesAutomatically = showStoreMessagesAutomatically
--- a/Sources/Purchasing/Configuration.swift
+++ b/Sources/Purchasing/Configuration.swift
@@ -186,6 +186,8 @@ import Foundation
         /// automatically.
         /// Then use the ``Purchases/showStoreMessages(for:)`` method to display the messages.
         /// More information:  https://rev.cat/storekit-message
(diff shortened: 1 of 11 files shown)


Sources/Purchasing/Configuration.swift
```diff
@@ -184,8 +184,10 @@ import Foundation
         /// If you want to disable this behavior so that you can customize when these messages are shown, make sure
         /// you configure the SDK as early as possible in the app's lifetime, otherwise messages will be displayed
         /// automatically.
-        /// Then use the ``Purchases/showStoreMessages(for:)-8kw87`` method to display the messages.
+        /// Then use the ``Purchases/showStoreMessages(for:)`` method to display the messages.
         /// More information:  https://rev.cat/storekit-message
+        /// - Important: Set this property only if you're using swift. If you're using ObjC, you won't be able to call
+        /// the related methods
```
NachoSoto, line 190:
```suggestion
        /// - Important: Set this property only if you're using Swift. If you're using ObjC, you won't be able to call
        /// the related methods
```

Tests/StoreKitUnitTests/StoreMessageTypeTests.swift
```diff
@@ -0,0 +1,42 @@
...
+
+    func testMessageReasonIsMappedToCorrectStoreMessageType() async throws {
+        #if swift(>=5.8)
+        if #available(iOS 16.4, *) {
+            expect(Message.Reason.billingIssue.messageType) == .billingIssue
+        }
+        #endif
+        expect(Message.Reason.priceIncreaseConsent.messageType) == .priceIncreaseConsent
+        expect(Message.Reason.generic.messageType) == .generic
+    }
```
NachoSoto, line 39: Nit:
```suggestion
    }
    
```

NachoSoto (approved):
Thank you! Looks great.

Tests/StoreKitUnitTests/StoreMessageTypeTests.swift
tonidero: I keep forgetting these... It would be so great to have a linter for this kind of code style stuff

NachoSoto: Yeah it took me a while to get used to it. I don't have a big preference other than consistency. I think it would be simpler to get rid of this and lint it.

PR #3274, opened 2023-10-03
`Purchases`: new `cachedCustomerInfo` and `cachedOfferings`

### LINKED ISSUE #2391: Consider adding a synchronous `customerInfo` method that returns a cached `customerInfo` if it's available
I just upgraded from 3.x to 4.x. In the process of doing that, I experimented with using `customerInfoStream` in place of the delegate.

I'd like to suggest that you add a synchronous `customerInfo` method that returns a cached `customerInfo` if it's available.

Let me explain why.

The first screen of my app has features that show up only for subscribed users. During initialization I try to determine the subscription status as early as possible. In particular, I try to determine the subscription status before the first view is shown. If I discover the status too late, the view is initially drawn in the unsubscribed state, and then re-drawn in the subscribed state.

If I use a delegate, I get the status early enough. If I use `customerInfoStream` I don't.

It turns out the difference is [here](https://github.com/RevenueCat/purchases-ios/blob/583714f4822ae978f359e29f8923f14fd0483b24/Sources/Purchasing/Purchases/Purchases.swift#L93):

```swift
            // Sends cached customer info (if exists) to delegate as latest
            // customer info may have already been observed and sent by the monitor
            self.sendCachedCustomerInfoToDelegateIfExists()
```

When the delegate is assigned, it's immediately — synchronously — sent the cached `customerInfo` (if it exists). It seems there's *no other way* for me to get the cached `customerInfo` synchronously (if it exists)!

Most apps would probably have the same issue, no? Or... maybe I'm thinking about this the wrong way?

Thoughts?

P.S. Of course, I could cache the subscription status myself, etc., but I don't want to add a cache on top of your cache.

NachoSoto:
This allows fixing #3283.
Fixes #2391.

### DIFF
--- a/Sources/Purchasing/OfferingsManager.swift
+++ b/Sources/Purchasing/OfferingsManager.swift
@@ -42,7 +42,7 @@ class OfferingsManager {
         fetchPolicy: FetchPolicy = .default,
         completion: (@MainActor @Sendable (Result<Offerings, Error>) -> Void)?
     ) {
-        guard let memoryCachedOfferings = self.deviceCache.cachedOfferings else {
+        guard let memoryCachedOfferings = self.cachedOfferings else {
             Logger.debug(Strings.offering.no_cached_offerings_fetching_from_network)
 
             self.systemInfo.isApplicationBackgrounded { isAppBackgrounded in
@@ -65,6 +65,10 @@ class OfferingsManager {
                                           completion: nil)
             }
         }
+    }
+
+    var cachedOfferings: Offerings? {
+        return self.deviceCache.cachedOfferings
     }
 
     func updateOfferingsCache(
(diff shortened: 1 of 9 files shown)


RevenueCatUI/PaywallView.swift
```diff
@@ -355,19 +383,4 @@ private extension PaywallViewMode {
 
 }
 
-@available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
-private extension PaywallTemplate {
-
-    var name: String {
```
NachoSoto, line 361: This wasn't used and was triggering a "file length" lint violation.

aboedo (approved):

aboedo (approved):

PR #3278, opened 2023-10-05
`SK2StoreProduct.priceFormatter`: use locale from `StoreKit.Product`

NachoSoto:
This only works in iOS 16+ (because it needs `StoreKit.Product.priceFormatStyle`), but it's a good improvement which ensures consistent formatting of prices in paywalls, where sometimes `StoreProduct.localizedPrice` would have a different format compared to formatting `pricePerMonth` using `StoreProduct.priceFormatter`.

### DIFF
--- a/Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
@@ -39,7 +39,7 @@ internal struct SK2StoreProduct: StoreProductType {
 
     var localizedDescription: String { underlyingSK2Product.description }
 
-    var currencyCode: String? { self._currencyCode }
+    var currencyCode: String? { self._currencyCodeAndLocale.code }
 
     var price: Decimal { underlyingSK2Product.price }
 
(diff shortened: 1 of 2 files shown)


Tests/StoreKitUnitTests/StoreProductTests.swift
```diff
@@ -299,6 +299,33 @@ class StoreProductTests: StoreKitConfigTestCase {
         expect(storeProduct.currencyCode) == "USD"
     }
 
+    @available(iOS 16.0, tvOS 16.0, macOS 13.0, watchOS 9.0, *)
+    func testSk2PriceFormatterGetsLocaleFromStoreKit() async throws {
```
NachoSoto, line 303: This is equivalent to the previous test `testSk2PriceFormatterReactsToStorefrontChanges`, except it doesn't need to manually change the locale in the `priceFormatter`

Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
```diff
@@ -86,28 +91,31 @@ internal struct SK2StoreProduct: StoreProductType {
...
     }
 
     private var jsonDict: [String: Any] {
         let decoded = try? JSONSerialization.jsonObject(with: self.underlyingSK2Product.jsonRepresentation, options: [])
         return decoded as? [String: Any] ?? [:]
     }
 
-    // This is marked as `@_backDeploy`, but it's only visible when compiling with Xcode 14.x
-    // and for some reason only returning a non-empty string on iOS 16+.
+    // This is marked as `@_backDeploy` but for some reason only returns a non-empty string on iOS 16+.
```
aboedo, line 115: lol classic storekit

aboedo (approved):

PR #3289, opened 2023-10-07
`PaywallEventStore`: changed container to use `URL.applicationSupportDirectory`

NachoSoto:
See https://nemecek.be/blog/57/making-files-from-your-app-available-in-the-ios-files-app#writing-files-into-the-application-support-directory

Unfortunately the choice of `URL.documentsDirectory` meant that any app wanting to make its documents accessible via the Files app,
would have exposed these events.

### DIFF
--- a/Sources/Paywalls/Events/PaywallEventStore.swift
+++ b/Sources/Paywalls/Events/PaywallEventStore.swift
@@ -85,23 +85,46 @@ internal actor PaywallEventStore: PaywallEventStoreType {
 @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
 extension PaywallEventStore {
 
-    static func createDefault(applicationSupportDirectory: URL?) throws -> PaywallEventStore {
-        let applicationSupportDirectory = try applicationSupportDirectory ?? Self.applicationSupportDirectory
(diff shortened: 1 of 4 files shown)


aboedo (approved):

Sources/Paywalls/Events/PaywallEventStore.swift
```diff
@@ -85,23 +85,61 @@ internal actor PaywallEventStore: PaywallEventStoreType {
...
+    static func createDefault(
+        applicationSupportDirectory: URL?,
+        documentsDirectory: URL? = nil
+    ) throws -> PaywallEventStore {
+        let url = Self.url(in: try applicationSupportDirectory ?? Self.applicationSupportDirectory)
+        Logger.verbose(PaywallEventStoreStrings.initializing(url))
+
         let documentsDirectory = try documentsDirectory ?? Self.documentsDirectory
-        let url = documentsDirectory
+        Self.removeStoreIfExists(Self.url(in: documentsDirectory))
```
aboedo, line 96: maybe we can call this something like `removeLegacyDirectoryIfExists` so it's obvious at first glance what we're doing and why?

aboedo (approved):

PR #3290, opened 2023-10-09
Expose `productPlanIdentifier` in `EntitlementInfo`

tonidero:
### Description
Exposes `productPlanIdentifier` in `EntitlementInfo`. This will be helpful in the hybrids so they have actual values for users purchasing Google play products and using iOS apps.

### DIFF
--- a/Sources/Purchasing/EntitlementInfo.swift
+++ b/Sources/Purchasing/EntitlementInfo.swift
@@ -134,7 +134,7 @@ extension PeriodType: DefaultValueProvider {
     @objc public var productIdentifier: String { self.contents.productIdentifier }
 
     /**
-     The product plan identifier that unlocked this entitlement (usually for a Google Play purchase)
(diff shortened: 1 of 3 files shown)


Sources/Purchasing/EntitlementInfo.swift
```diff
@@ -136,7 +136,7 @@ extension PeriodType: DefaultValueProvider {
     /**
      The product plan identifier that unlocked this entitlement (usually for a Google Play purchase)
```
NachoSoto, line 137: Should we remove the word "usually" here?

NachoSoto (approved):

Sources/Purchasing/EntitlementInfo.swift
tonidero: I pushed the change but it's not appearing in Github 😞. Will wait for a bit to see if it appears

PR #3291, opened 2023-10-09
`Paywalls`: `watchOS` support

NachoSoto:
simulator_screenshot_38A61F03-8048-40E2-AD3E-384DF4112771
simulator_screenshot_ADD7F6B5-8A39-473B-85E2-5175027F641C


Tests/UnitTests/Paywalls/Events/PaywallEventsBackendTests.swift
```diff
@@ -72,7 +72,7 @@ private extension BackendPaywallEventTests {
         offeringIdentifier: "offering_1",
         paywallRevision: 5,
         sessionID: .init(uuidString: "98CC0F1D-7665-4093-9624-1D7308FFF4DB")!,
-        displayMode: .condensedFooter,
+        displayMode: .fullScreen,
```
NachoSoto, line 75: `.fullScreen` is the only available on in `watchOS`, so I changed this to fix those tests.

RevenueCatUI/Data/TestData.swift
```diff
@@ -521,6 +529,14 @@ internal enum TestData {
             type: .introductory
         )
     }
+
+    #if os(watchOS)
+    // `Locale.current` in watchOS produces `en_001` when running tests
```
NachoSoto, line 534: https://twitter.com/nachosoto/status/1724890382776066080?s=61&t=FOp7rqqtZdYUZnPvV5nkBw

RevenueCatUI/Templates/Template5View.swift
```diff
@@ -15,6 +15,8 @@ import RevenueCat
 import SwiftUI
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@available(tvOS, unavailable)
```
tonidero, line 18: Do we need to remove the `watchOS` from the lines above each of these?

RevenueCatUI/Views/TemplateBackgroundImageView.swift
```diff
@@ -54,6 +54,15 @@ struct TemplateBackgroundImageView: View {
...
                 .blur(radius: 40)
                 .opacity(0.7)
+                .background {
+                    // Simulate dark materials in pre-watchOS 10.0
+                    // where `Material` isn't available.
+                    #if os(watchOS)
+                    if #unavailable(watchOS 10.0) {
+                        Color.black
+                    }
+                    #endif
```
tonidero, line 64: Just to confirm, in the else case, I guess this will be nil, so there won't be any background, correct?

NachoSoto: That's correct.

RevenueCatUI/Templates/Template5View.swift
NachoSoto: If you don't add it it assumes the requirement matches the deployment target in the target which is 6.2 so it fails to compile.

Tests/RevenueCatUITests/Data/__Snapshots__/PaywallDataValidationTests/testUnrecognizedTemplateNameGeneratesDefaultPaywall-watchOS.1.json
```diff
@@ -24,7 +24,7 @@
     "tos_url" : null
   },
   "localized_strings" : {
-    "en_001" : {
```
NachoSoto: This is broken on watchOS + Xcode 15

PR #3312, opened 2023-10-16
`CI`: run all iOS 17 tests

NachoSoto:
Xcode 15.0.1 (#3311) fixed performance issues (partially), so we should be able to run all tests on iOS 17.

See https://discuss.circleci.com/t/severe-performance-problems-with-xcode-15/49205/65?u=nachosoto


NachoSoto:
Unfortunately the performance issue is still there when running tests that require a host app :(

.circleci/config.yml
```diff
@@ -424,10 +424,10 @@ jobs:
           name: Run tests
           # StoreKit tests currently get stuck.
           # See https://discuss.circleci.com/t/severe-performance-problems-with-xcode-15/49205/11?u=nachosoto
```
tonidero, line 426: Can we remove this comment now?

tonidero (approved):

.circleci/config.yml
NachoSoto: Oops yeah

.circleci/config.yml
```diff
@@ -958,7 +956,7 @@ workflows:
     when: << pipeline.parameters.generate_snapshots >>
     jobs:
       - run-test-ios-17:
-          xcode_version: '15.0.0'
+          xcode_version: '15.0.1'
```
NachoSoto, line 959: CircleCI initially shipped the update under `Xcode 15.0.0`, but now it's its own `15.0.1`

PR #3315, opened 2023-10-18
`Integration Tests`: removed sleep to wait for expirations

NachoSoto:
I wonder if this is fixed in iOS 17.

We filed a Radar about this almost 1 year ago:
Screenshot 2023-10-17 at 17 21 03


NachoSoto:
Nope, still required.

PR #3323, opened 2023-10-23
`Paywalls`: new `.onPurchaseCompleted` overload with `StoreTransaction`

NachoSoto:
The current API only provided a `CustomerInfo`.

Note that this does not include `userCanceled`. We can expose separately with another `.onPurchaseCancelled` modifier in the future.


RevenueCatUI/View+PurchaseRestoreCompleted.swift
```diff
@@ -34,7 +38,7 @@ extension View {
     ///             PaywallView()
     ///                 .onPurchaseCompleted { customerInfo in
     ///                     print("Purchase completed: \(customerInfo.entitlements)")
-    ///                     self.didPurchase = false
+    ///                     self.displayPaywall = false
```
NachoSoto, line 41: This was wrong

joshdholtz (approved):
This is a good change 🔥

PR #3325, opened 2023-10-23
`Paywalls`: enable all iOS 17 tests

NachoSoto:
These were failing in some of the earlier Xcode 15 betas, but they all seem to be passing now. `PresentIfNeededTests` are still not working though. But I just manually verified that these modifiers do work correctly on iOS 17 on an app.

### DIFF
--- a/Tests/RevenueCatUITests/BaseSnapshotTest.swift
+++ b/Tests/RevenueCatUITests/BaseSnapshotTest.swift
@@ -87,7 +87,9 @@ extension BaseSnapshotTest {
 extension View {
 
     /// Adds the receiver to a view hierarchy to be able to test lifetime logic.
-    func addToHierarchy() throws {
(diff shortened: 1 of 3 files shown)


Tests/RevenueCatUITests/BaseSnapshotTest.swift
```diff
@@ -88,10 +88,6 @@ extension View {
 
     /// Adds the receiver to a view hierarchy to be able to test lifetime logic.
     func addToHierarchy() throws {
-        if #available(iOS 17.0, *) {
-            try XCTSkipIf(true, "This is currently not working on iOS 17")
-        }
-
```
NachoSoto, line 94: Now we'll be running a lot more tests that were disabled on iOS 17.

joshdholtz (approved):
LGTM

PR #3327, opened 2023-10-23
`ProductFetcherSK1`: enable `TimingUtil` log

NachoSoto:
`ProductFetcherSK1` had 2 public methods for fetching products: `products(withIdentifiers:completion:)` and `sk1Products(withIdentifiers:completion:)`.
Only one of them was using `TimingUtil` to log slow requests, and that one happened to not be used by `ProductsManager`, which means that we weren't getting these logs.

This is evident in https://github.com/RevenueCat/react-native-purchases/issues/755. With this change, we'll now be able to see how long the product request took.

### DIFF
--- a/Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
+++ b/Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
@@ -40,35 +40,6 @@ final class ProductsFetcherSK1: NSObject {
         self.requestTimeout = requestTimeout
     }
 
-    func sk1Products(withIdentifiers identifiers: Set<String>,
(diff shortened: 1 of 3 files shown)


Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
```diff
@@ -101,6 +72,35 @@ final class ProductsFetcherSK1: NSObject {
         }
     }
 
+    private func sk1Products(withIdentifiers identifiers: Set<String>,
```
NachoSoto, line 75: This is private now, so the other method is the only `public` one.

aboedo (approved):

PR #3332, opened 2023-10-24
`Paywalls`: add identifier to events

joshdholtz:
### Motivation

Send a unique `id` on each paywall event

### Description

- Added `id: ID?` to `PaywallEvent.Data`
  - Optional so its backwards compatible with events recorded before this change (see actual tests from below)
- Add `withNewID()` function that will mutate the `id` when tracked for cancel and close 

#### Events recored before `id` changes but posted with `id` changes

Screenshot 2023-10-25 at 9 33 27 AM

#### Events recorded and posted with `id` changes

Screenshot 2023-10-25 at 9 37 53 AM


NachoSoto:
How are you thinking of this not being backwards compatible? Just drop previous unsent events?

joshdholtz:
> How are you thinking of this not being backwards compatible? Just drop previous unsent events?

@NachoSoto The backend has this event id being optional so that its backwards compatible 🤷‍♂️ I _could_ have made new "version 2" validation models on the backend but it wouldn't provide that much value and more code to maintain

NachoSoto:
What i mean is that `PaywallStoredEvent`/`PaywallEvent` now has a new mandatory field, so `PaywallEventStore.fetch` will not decode them anymore.

Sources/Paywalls/Events/PaywallEvent.swift
```diff
@@ -16,6 +16,9 @@ import Foundation
 /// An event to be sent by the `RevenueCatUI` SDK.
 public enum PaywallEvent {
 
+    /// An identifier that represents a paywall event.
+    public typealias EventID = UUID
```
NachoSoto, line 20: Maybe
```suggestion
    public typealias ID = UUID
```
So this becomes `PaywallEvent.ID`?

Sources/Paywalls/Events/PaywallEvent.swift
```diff
@@ -54,6 +58,7 @@ extension PaywallEvent {
             darkMode: Bool
         ) {
             self.init(
+                id: EventID(),
```
NachoSoto, line 61: I'd change it to:
```suggestion
                id: .init(),
```

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -122,7 +122,7 @@ extension PurchaseHandler {
             return false
         }
 
-        self.track(.close(data.withCurrentDate()))
+        self.track(.close(data.withNewID().withCurrentDate()))
```
joshdholtz, line 125: Need this otherwise the `id` was the same for each event in the session 😅

Sources/Paywalls/Events/PaywallEvent.swift
```diff
@@ -36,6 +40,7 @@ extension PaywallEvent {
     public struct Data {
 
         // swiftlint:disable missing_docs
+        public var id: ID?
```
joshdholtz, line 43: Optional to make backwards compatible with previously stored events

RevenueCatUI/Purchasing/PurchaseHandler.swift
NachoSoto: This is a smell in how we're using the type.

I think to prevent this we should put the `Event.ID` inside of `PaywallEvent` not `PaywallEvent.Data`. What do you think?

Sources/Paywalls/Events/PaywallEvent.swift
NachoSoto: Let's leave a comment explaining that.

Sources/Paywalls/Events/Networking/PaywallEventsRequest.swift
```diff
@@ -63,6 +64,7 @@ extension PaywallEventsRequest.Event {
         let data = storedEvent.event.data
 
         self.init(
+            id: data.id?.uuidString,
```
NachoSoto, line 67: Let's add a test verifying that we can send events with no ID.

NachoSoto (commented):
Almost ready, a few comments:
- Re-using event data and having to change IDs
- Test for posting events with no ID
- Also can you add a test in `PaywallEventStoreTests` that verifies that we can fetch events with no ID?

NachoSoto:
Nit: I'd mark this as a "refactor" since it's not a user facing change that users need to know about in the release notes IMO.

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -135,7 +135,7 @@ extension PurchaseHandler {
             return false
         }
 
-        self.track(.cancel(data.withCurrentDate()))
+        self.track(.cancel(.init(), data))
```
NachoSoto, line 138: Much better 👍🏻

NachoSoto (approved):

PR #3339, opened 2023-10-24
`Paywall Tester`: fixed paywall presentation

NachoSoto:
Having the `.sheet` deep inside a `List` + `Section` + `if` was making `SwiftUI` re-evaluate it over and over.
This cleans up `PaywallPresenter` and moves it to the top level of the view to avoid that.


joshdholtz (approved):
LGTM 💪

PR #3342, opened 2023-10-24
`Paywalls`: new `onDismiss` parameter for `presentPaywallIfNeeded`

NachoSoto:
Without this there is no way to know when the paywall was dismissed.

### DIFF
--- a/RevenueCatUI/View+PresentPaywall.swift
+++ b/RevenueCatUI/View+PresentPaywall.swift
@@ -38,7 +38,8 @@ extension View {
         requiredEntitlementIdentifier: String,
         fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
         purchaseCompleted: PurchaseOrRestoreCompletedHandler? = nil,
-        restoreCompleted: PurchaseOrRestoreCompletedHandler? = nil
+        restoreCompleted: PurchaseOrRestoreCompletedHandler? = nil,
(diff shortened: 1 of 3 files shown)


RevenueCatUI/View+PresentPaywall.swift
```diff
@@ -188,9 +198,9 @@ private struct PresentingPaywallModifier: ViewModifier {
     }
 
     private func close() {
-        self.data = nil
-
         Logger.debug(Strings.dismissing_paywall)
+
+        self.data = nil
```
NachoSoto, line 203: Just re-ordered this because the log says "dismissing", not "dismissed".

RevenueCatUI/View+PresentPaywall.swift
```diff
@@ -140,7 +150,7 @@ private struct PresentingPaywallModifier: ViewModifier {
 
     func body(content: Content) -> some View {
         content
-            .sheet(item: self.$data) { data in
+            .sheet(item: self.$data, onDismiss: self.onDismiss) { data in
```
NachoSoto, line 153: I tested this and it gets called when closing the sheet manually by the user and programatically.

Unfortunately I wasn't able to write a unit test that gets called because of the way we hack adding SwiftUI views to the hierarchy without a host app in the test target.

joshdholtz (approved):
💯

PR #3357, opened 2023-10-27
`Paywalls`: improve conversion from `Color`/`UIColor` to `PaywallColor`

NachoSoto:
This wasn't fully required, but now we'll correctly encode `stringRepresentation`.
Also fixed the math (dividing by 255 instead of 256), which was technically producing slightly incorrect colors.

Also adds an overload for `UIColor` which allows us to use color literals in Xcode:
image


Sources/Paywalls/PaywallColor.swift
```diff
@@ -116,10 +116,10 @@ private extension PaywallColor {
                 hexNumber |= 0xff
             }
 
-            red = CGFloat((hexNumber & 0xff000000) >> 24) / 256
-            green = CGFloat((hexNumber & 0x00ff0000) >> 16) / 256
-            blue = CGFloat((hexNumber & 0x0000ff00) >> 8) / 256
-            alpha = CGFloat(hexNumber & 0x000000ff) / 256
+            red = CGFloat((hexNumber & 0xff000000) >> 24) / 255
```
NachoSoto, line 119: These were wrong

joshdholtz (approved):
👍

PR #3359, opened 2023-10-27
`Paywalls`: new optional `displayCloseButton` parameter

NachoSoto:
The [documentation](https://www.revenuecat.com/docs/displaying-paywalls#close-button) points out that this is only automatic when using `presentPaywallIfNeeded`, but it can also be useful to have `RevenueCatUI` do this automatically.

### DIFF
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -26,9 +26,12 @@ public struct PaywallView: View {
 
     private let mode: PaywallViewMode
     private let fonts: PaywallFontProvider
+    private let displayCloseButton: Bool
 
     @Environment(\.locale)
     private var locale
+    @Environment(\.dismiss)
+    private var dismissAction
 
     @StateObject
     private var purchaseHandler: PurchaseHandler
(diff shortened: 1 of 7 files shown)


RevenueCatUI/PaywallView.swift
```diff
@@ -26,9 +26,12 @@ public struct PaywallView: View {
 
     private let mode: PaywallViewMode
     private let fonts: PaywallFontProvider
+    private let displayCloseButton: Bool
```
tonidero, line 29: I used `shouldDisplayDismissButton` as the naming for this in [Android](https://github.com/RevenueCat/purchases-android/pull/1390)... Honestly I could go either way, but I think ideally we have the same naming.

Lmk if you prefer this and I can update it in Android

RevenueCatUI/PaywallView.swift
```diff
@@ -44,34 +47,40 @@ public struct PaywallView: View {
     private var error: NSError?
 
     /// Create a view that loads the `Offerings.current`.
+    /// - Parameter displayCloseButton: Set this to `true` to automatically include a close button.
```
tonidero, line 50: Just a question but I noticed we don't have docs for the `fonts` parameter. Should we have it as well?

RevenueCatUI/PaywallView.swift
```diff
@@ -179,6 +191,19 @@ public struct PaywallView: View {
...
     }
 
+    @ToolbarContentBuilder
+    private var toolbar: some ToolbarContent {
+        ToolbarItemGroup(placement: .destructiveAction) {
+            if self.displayCloseButton {
+                Button {
+                    self.dismiss()
+                } label: {
+                    Image(systemName: "xmark")
```
tonidero, line 201: In Android I used our own `close` button vector image for this, but I think it should be ok (probably preferrable) to use the system one.

tonidero (approved):

RevenueCatUI/PaywallView.swift
NachoSoto: Naming conventions in Swift and Kotlin are different, I don't think we can strive to maintain 100% API parity while keeping our APIs "native".
For reference, I checked and there's 0 methods of `SwiftUI.View` that contain the word "should".

RevenueCatUI/PaywallView.swift
NachoSoto: I'll add it now 👍🏻

RevenueCatUI/PaywallView.swift
NachoSoto: Yeah this uses SFSymbols which is definitely preferred.

PR #3361, opened 2023-10-30
`Paywalls`: avoid animating `PurchaseButton` labels when text does not change

NachoSoto:
Equivalent to https://github.com/RevenueCat/purchases-android/pull/1416

### DIFF
--- a/RevenueCatUI/Views/PurchaseButton.swift
+++ b/RevenueCatUI/Views/PurchaseButton.swift
@@ -85,6 +85,11 @@ struct PurchaseButton: View {
         .tint(.clear)
         .frame(maxWidth: .infinity)
         .dynamicTypeSize(...Constants.maximumDynamicTypeSize)
+        .transaction { transaction in
+            if !self.packagesProduceDifferentLabels {
+                transaction.animation = nil
+            }
+        }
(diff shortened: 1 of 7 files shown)


tonidero (approved):

RevenueCatUI/Views/PurchaseButton.swift
```diff
@@ -109,6 +114,29 @@ struct PurchaseButton: View {
...
+                .lazy
+                .map {
+                    IntroEligibilityStateView.text(
+                        for: .callToAction,
+                        localization: $0.localization,
+                        introEligibility: self.introEligibilityViewModel.allEligibility[$0.content]
+                    )
+                }
+        )
+        .count > 1
```
tonidero, line 133: This means we need to process the intro eligibility text for all packages in order to calculate the value (in android we would early exit the moment we find one that is different, but I think it's totally fine and probably cleaner 👍

NachoSoto: Oh yeah forgot to mention. Technically your implementation is more efficient, but we’re talking about <10 iterations if that, and each of those is just a few if statements.

PR #3364, opened 2023-10-31
`Paywalls`: improve Japanese localization

NachoSoto:


tonidero (approved):
Still haven't learned enough japanese 😆

PR #3365, opened 2023-10-31
[SK2] Add support for StoreKit Config files in SK2

MarkVillacampa:
When using StoreKit Config file, we do not have access to the transaction history or subscription status server-side.

To work around this, we introduce a new `EncodedAppleReceipt` type that closely resembles the structure of an SK1 receipt and contains all the transaction history and subscription status for the customer. It can be sent in a single request to `receipt`


Sources/Purchasing/StoreKit2/StoreKit2Receipt.swift
```diff
@@ -0,0 +1,47 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  StoreKit2Receipt.swift
+//
+//  Created by MarkVillacampa on 26/10/23.
+
+import StoreKit
+
+/// A type that resembles the structure of a StoreKit 1 receipt using StoreKit 2 data.
+struct StoreKit2Receipt: Equatable, Codable {
```
NachoSoto, line 17: 🤯

Sources/Purchasing/StoreKit2/StoreKit2Receipt.swift
```diff
@@ -0,0 +1,47 @@
...
+    /// The bundle identifier of the app.
+    let bundleId: String
+
+    /// The app version that the user originally purchased from the App Store.
+    let originalApplicationVersion: String?
+
+    /// The date the user originally purchased the app from the App Store.
+    let originalPurchaseDate: Date?
+
+    private enum CodingKeys: String, CodingKey {
```
NachoSoto, line 39: Nit: to make reading the code easier, I recommend separating the actual type definition with conformances:
```swift

struct Data {

  // fields

}

// MARK: -

extension Data: Codable {

  private enum CodingKeys: String, CodingKey { 

     // ... 

  }

}

```

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -92,15 +95,27 @@ final class TransactionPoster: TransactionPosterType {
...
         if systemInfo.dangerousSettings.internalSettings.usesStoreKit2JWS,
            let jwsRepresentation = transaction.jwsRepresentation {
-            self.fetchProductsAndPostReceipt(
-                transaction: transaction,
-                data: data,
-                receipt: .jws(jwsRepresentation),
-                completion: completion
-            )
+            if transaction.environment == .xcode, #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
+                _ = Task<Void, Never> {
```
MarkVillacampa, line 99: Any better way to avoid duplication? The enclosing function is not async

Sources/Purchasing/StoreKitAbstractions/EncodedAppleReceipt.swift
```diff
@@ -29,7 +30,13 @@ extension EncodedAppleReceipt {
         case .jws(let jws):
             return jws
         case .receipt(let data):
-            return data.asFetchToken
+            return data.base64EncodedString()
+        case .sk2receipt(let receipt):
+            do {
+                return try receipt.jsonEncodedData.base64EncodedString()
+            } catch {
+                return ""
```
NachoSoto: Let's log an error here.

Sources/Purchasing/StoreKitAbstractions/SK2AppTransaction.swift
```diff
@@ -0,0 +1,32 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  SK2AppTransaction.swift
+//
+//  Created by MarkVillacampa on 26/10/23.
+
+import StoreKit
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+internal struct SK2AppTransaction {
```
NachoSoto, line 17: Nice, good idea abstracting this. Let's add a docstring with something like "wrapper for `StoreKit.AppTransaction`".

Sources/Purchasing/StoreKitAbstractions/SK2AppTransaction.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, visionOS 1.0, *)
+    init(appTransaction: AppTransaction) {
+        self.bundleId = appTransaction.bundleID
+        self.originalApplicationVersion = appTransaction.originalAppVersion
+        self.originalPurchaseDate = appTransaction.originalPurchaseDate
+        self.environment = StoreEnvironment.init(rawValue: appTransaction.environment.rawValue.lowercased())
```
NachoSoto, line 24: Instead of putting the logic to convert here I'd add a constructor in `StoreEnvironment` that takes an `AppStore.Environment` type. Then in there we should log if the type is unrecognized too.

And I'd strongly recommend adding tests for that because this is implicitly assuming that our raw values match StoreKit's.

Sources/Purchasing/StoreKit2/StoreKit2Receipt.swift
```diff
@@ -0,0 +1,47 @@
...
+/// A type that resembles the structure of a StoreKit 1 receipt using StoreKit 2 data.
+struct StoreKit2Receipt: Equatable, Codable {
+
+    /// The server environment where the receipt was generated.
+    let environment: StoreEnvironment
+
+    /// The current subscription status for each subscription group, including the renewal information.
+    /// The keys of the dictionary represent Subscription Group Identifiers, where the values are arrays 
+    /// of JWS tokens of the renewal information.
+    let subscriptionStatus: [String: [String]]
```
MarkVillacampa, line 25: Thinking about adding some typealias to String, to improve clarity, thoughts?

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -73,6 +76,78 @@ final class StoreKit2TransactionFetcher: StoreKit2TransactionFetcherType {
...
+                    .extractValues()
+                    .reduce(into: [:]) {result, value in
+                        result[value.groupID] = value.statuses
+                    }
+            } else {
+                // `StoreKit.Product.SubscriptionInfo.Status.all` is only available starting in iOS 17.0
+                // For previous versions, we retrieve all the previously purchased transactions,
+                // and fetch the subscription status only once per subscription group.
+                var subscriptionGroups: [String: [Product.SubscriptionInfo.Status]] = [:]
+                for await transaction in StoreKit.Transaction.all {
```
MarkVillacampa, line 102: Thought about using `StoreKit.Transaction.currentEntitlements` but that only returns current subscriptions whereas `StoreKit.Product.SubscriptionInfo.Status.all` also returns inactive subscriptions.

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -73,6 +76,78 @@ final class StoreKit2TransactionFetcher: StoreKit2TransactionFetcherType {
...
+            } catch {
+                return nil
+            }
+        }
+    }
+
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    var receipt: StoreKit2Receipt {
+        get async {
+            async let transactions = verifiedTransactions.compactMap(\.jwsRepresentation)
```
NachoSoto, line 136: `verifiedTransactions` is using `extractAll` on a potentially very large array of a lot of data.
I'd recommend changing the property to return `jwsRepresentation`s instead and moving this `map` before calling `extractAll` to lower the memory footprint.

Sources/Purchasing/StoreKitAbstractions/SK2StoreTransaction.swift
```diff
@@ -16,14 +16,17 @@ import StoreKit
 @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
 internal struct SK2StoreTransaction: StoreTransactionType {
 
-    init(sk2Transaction: SK2Transaction, jwsRepresentation: String) {
+    init(sk2Transaction: SK2Transaction, jwsRepresentation: String, environment: StoreEnvironment? = nil) {
```
MarkVillacampa, line 19: Needed a way to force the environment of a StoreTransaction since all transactions in tests have the `xcode` environment.

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -92,15 +95,27 @@ final class TransactionPoster: TransactionPosterType {
...
+                    )
+                }
+            } else {
+                self.fetchProductsAndPostReceipt(
+                    transaction: transaction,
+                    data: data,
+                    receipt: .jws(jwsRepresentation),
+                    completion: completion
+                )
+            }
```
NachoSoto, line 115: This has so many branches is kinda hard to follow now. This new `if transaction.environent == .xcode` doesn't need to be inside of the `if let jwsRepresentation = transaction.jwsRepresentation`, right?

Maybe we can extract a method that produces the `EncodedAppleReceipt`, since all these branches end up doing the same:
```swift
self.fetchProductsAndPostReceipt(
    transaction: transaction,
    data: data,
    receipt: encodedReceipt,
    completion: completion
)
```

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -66,13 +67,15 @@ final class TransactionPoster: TransactionPosterType {
     init(
         productsManager: ProductsManagerType,
         receiptFetcher: ReceiptFetcher,
+        transactionFetcher: StoreKit2TransactionFetcherType,
```
NachoSoto, line 70: This type keeps getting more dependencies 😅
I'm thinking we could wrap `ReceiptFetcher` and `StoreKit2ReceiptFetcher` into a common type, but I think that's probably more complex than it's worth it. But it is weird that `TransactionPoster` now needs a way to also fetch transactions :/

NachoSoto (changes requested):
Awesome! Lots of small things, but the overall approach makes a lot of sense <3

Sources/Purchasing/StoreKit2/StoreKit2Receipt.swift
NachoSoto: A typealias doesn't provide any more guarantees. Can this by its own `struct`, and convert it to this format only when encoding it?

Sources/Purchasing/StoreKitAbstractions/SK2AppTransaction.swift
MarkVillacampa: Added constructor and now using exhaustive switch instead of converting from raw values

Sources/Purchasing/StoreKitAbstractions/SK2AppTransaction.swift
```diff
@@ -0,0 +1,32 @@
...
+//  SK2AppTransaction.swift
+//
+//  Created by MarkVillacampa on 26/10/23.
+
+import StoreKit
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+internal struct SK2AppTransaction {
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, visionOS 1.0, *)
```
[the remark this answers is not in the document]
MarkVillacampa: Removed all references to visionOS 👍

Sources/Purchasing/StoreKitAbstractions/SK2StoreTransaction.swift
```diff
@@ -16,14 +16,17 @@ import StoreKit
...
         self.underlyingSK2Transaction = sk2Transaction
 
         self.productIdentifier = sk2Transaction.productID
         self.purchaseDate = sk2Transaction.purchaseDate
         self.transactionIdentifier = String(sk2Transaction.id)
         self.quantity = sk2Transaction.purchasedQuantity
         self.jwsRepresentation = jwsRepresentation
+        self.environment = environment ?? StoreEnvironment(
+            rawValue: sk2Transaction.environmentStringRepresentation.lowercased()
+        )
```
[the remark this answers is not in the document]
MarkVillacampa: `sk2Transaction ` is a `StoreKit.Transaction` and its `environment` field is immutable. It will always be `Xcode` in tests, but we want to be able to test the behavior when the environment is `Sandbox`/`Production`.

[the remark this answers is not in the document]
NachoSoto: > we want to be able to test the behavior when the environment is Sandbox/Production.

That's something we can override with `MockTransaction` though, right?

Sources/Purchasing/StoreKitAbstractions/EncodedAppleReceipt.swift
MarkVillacampa: Added a log here and in `PostReceiptDataOperation.swift` when there's an error encoding an SK2 receipt

Sources/Networking/Operations/PostReceiptDataOperation.swift
```diff
@@ -197,6 +197,11 @@ private extension PostReceiptDataOperation {
                 content,
                 initiationSource: self.postData.initiationSource.rawValue
             ))
+        case .sk2receipt(let receipt):
+            self.log(Strings.receipt.posting_jws(
```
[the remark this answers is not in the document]
MarkVillacampa: Good catch! added proper logging

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1102,6 +1102,13 @@ private extension PurchasesOrchestrator {
                     return
                 }
 
+                var receipt: EncodedAppleReceipt = .jws(jwsRepresentation)
+
+                if transaction.environment == .xcode {
+                    let sk2receipt = await self.transactionFetcher.receipt
+                    receipt = .sk2receipt(sk2receipt)
+                }
+
```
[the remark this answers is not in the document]
MarkVillacampa: Simplified 👍 but not sure where in `handlePurchasedTransaction` is it duplicated tho?

Sources/Purchasing/StoreKitAbstractions/SK2StoreTransaction.swift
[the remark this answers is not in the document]
MarkVillacampa: Not in tests where we're using a transaction created via StoreKit Testing in a `StoreKitConfigTestCase`, unless I'm missing something?

Sources/Purchasing/StoreKit2/StoreKit2Receipt.swift
aboedo: +1 for structs

aboedo: also nitpick if it stays as dict: name dictionaries `valuesByKey` so it's clear what to expect from both sides, i.e.: `subscriptionStatusBySubscriptionGroupId`

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -90,15 +111,87 @@ extension StoreKit.VerificationResult where SignedType == StoreKit.Transaction {
...
+    private var verifiedTransactionsJWS: [String] {
+        get async {
+            return await StoreKit.Transaction.all
+                .compactMap { $0.verifiedStoreTransaction?.jwsRepresentation }
+                .extractValues()
+        }
+    }
+
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    private var subscriptionStatus: [String: [Product.SubscriptionInfo.Status]] {
```
aboedo, line 146: same note here re: `valuesByKey`

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -90,15 +111,87 @@ extension StoreKit.VerificationResult where SignedType == StoreKit.Transaction {
...
+        get async {
+            #if swift(>=5.9)
+            if #available(iOS 17.0, macOS 14.0, tvOS 17.0, watchOS 10.0, *) {
+                return await StoreKit.Product.SubscriptionInfo.Status.all
+                    .extractValues()
+                    .reduce(into: [:]) {result, value in
+                        result[value.groupID] = value.statuses
+                    }
+            } else {
+                // `StoreKit.Product.SubscriptionInfo.Status.all` is only available starting in iOS 17.0
```
aboedo, line 156: I'm confused, though - do you mean StoreKit.Transaction.all? Also, StoreKit.Transaction.all is available on 15+

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -90,15 +111,87 @@ extension StoreKit.VerificationResult where SignedType == StoreKit.Transaction {
...
+                // For previous versions, we retrieve all the previously purchased transactions,
+                // and fetch the subscription status only once per subscription group.
+                var subscriptionGroups: [String: [Product.SubscriptionInfo.Status]] = [:]
+                for await transaction in StoreKit.Transaction.all {
+                    if let verifiedTransaction = transaction.verifiedTransaction,
+                       let subscriptionGroup = verifiedTransaction.subscriptionGroupID,
+                       !subscriptionGroups.keys.contains(subscriptionGroup),
+                       let status = await verifiedTransaction.subscriptionStatus {
+                        subscriptionGroups[subscriptionGroup] = [status]
```
aboedo, line 165: I suppose there's no way to batch this

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
MarkVillacampa: Check the previous branch of the if/else statement. `StoreKit.Product.SubscriptionInfo.Status.all` is new in iOS 17

Sources/Purchasing/Purchases/TransactionPoster.swift
NachoSoto: Suggested below.

Sources/Purchasing/StoreKitAbstractions/StoreEnvironment.swift
```diff
@@ -0,0 +1,62 @@
...
+            self = .production
+        case "Sandbox":
+            self = .sandbox
+        case "Xcode":
+            self = .xcode
+        default:
+            Logger.appleWarning(Strings.storeKit.skunknown_payment_mode(environment))
+            return nil
+        }
+    }
```
NachoSoto, line 60: These strings are duplicated because you defined `StoreEnvironment: String`. Maybe remove the `: String` part?

Sources/Purchasing/StoreKitAbstractions/StoreEnvironment.swift
```diff
@@ -0,0 +1,62 @@
...
+    init?(environment: StoreKit.AppStore.Environment) {
+        switch environment {
+        case .production:
+            self = .production
+        case .sandbox:
+            self = .sandbox
+        case .xcode:
+            self = .xcode
+        default:
+            Logger.appleWarning(Strings.storeKit.skunknown_payment_mode(String.init(describing: environment)))
```
NachoSoto, line 43: These aren't payment modes 😇

Sources/Purchasing/StoreKitAbstractions/StoreEnvironment.swift
```diff
@@ -0,0 +1,62 @@
...
+    init?(environment: String) {
+        switch environment {
+        case "Production":
+            self = .production
+        case "Sandbox":
+            self = .sandbox
+        case "Xcode":
+            self = .xcode
+        default:
+            Logger.appleWarning(Strings.storeKit.skunknown_payment_mode(environment))
```
NachoSoto, line 57: These aren't payment modes 😇

Sources/Purchasing/StoreKit2/StoreKit2Receipt.swift
```diff
@@ -0,0 +1,74 @@
...
+    private enum CodingKeys: String, CodingKey {
+        case environment
+        case subscriptionStatus = "subscription_status"
+        case transactions
+        case bundleId = "bundle_id"
+        case originalApplicationVersion = "original_application_version"
+        case originalPurchaseDate = "original_purchase_date"
+    }
+
+    func encode(to encoder: Encoder) throws {
```
NachoSoto, line 61: Can you explain why you had to implement this manually?

MarkVillacampa: After introducing the `SubscriptionStatus` struct so it's more obvious what each field represents, I needed to customize the encoding so the JSON format for `subscription_status` fits what the API will expect (it's also less verbose):

```
"subscription_status" : {
  "{SUBSCRIPTION_GROUP_ID}": ["jws_token"]
}
```

instead of 

```
"subscription_status" : [{
   "subscriptionGroupId": "foo",
  "renewalInfoJWSTokens": ["bar"]
}]
```

Sources/Purchasing/StoreKitAbstractions/StoreEnvironment.swift
MarkVillacampa: The strings in the `case` statement represent the possible values for `StoreKit.AppStore.Environment`, so they are not really duplicated in the `StoreEnvironment` enum (e.g. first letter is capitalized)

Moreover, if I remove the `: String`, the JSON-encoded value of the enum becomes something like `{"production":{}}` instead of just `"production"` 😅

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -90,15 +116,109 @@ extension StoreKit.VerificationResult where SignedType == StoreKit.Transaction {
...
+                    }
+
+                    return await taskGroup.reduce(
+                        into: [String: [Product.SubscriptionInfo.Status]]()
+                    ) { result, value  in
+                        if let value = value {
+                            result[value.0] = [value.1]
+                        }
+                    }
+                }
```
MarkVillacampa, line 193: Tried using a task group to parallelize fetching the `subscriptionStatus` for all transactions, @NachoSoto care to double check?

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
MarkVillacampa: Implemented a possible solution to parallelize using taskGroup 👍

Tests/StoreKitUnitTests/StoreKit2/StoreKit2TransactionFetcherTests.swift
```diff
@@ -148,6 +148,22 @@ class StoreKit2TransactionFetcherTests: StoreKitConfigTestCase {
         expect(result) == transaction
     }
 
+    // MARK: - receipt
+
+    func testGeneratesReceipt() async throws {
+        try AvailabilityChecks.skipIfXcode14orEarlier()
```
MarkVillacampa, line 154: Even thought [subscriptionStatus](https://developer.apple.com/documentation/storekit/transaction/4137250-subscriptionstatus) (used in `receipt`) is back-deployed to iOS 15, it is not available in Xcode 14 :(

Sources/Purchasing/StoreKitAbstractions/StoreEnvironment.swift
```diff
@@ -0,0 +1,62 @@
...
+
+// MARK: - Private
+
+extension StoreEnvironment {
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    init?(environment: StoreKit.AppStore.Environment) {
+        switch environment {
+        case .production:
+            self = .production
```
NachoSoto, line 37: It would be nice to add tests for these.

Sources/Purchasing/StoreKitAbstractions/StoreEnvironment.swift
```diff
@@ -0,0 +1,62 @@
...
+//  StoreEnvironment.swift
+//
+//  Created by MarkVillacampa on 26/10/23.
+//
+
+import Foundation
+import StoreKit
+
+/// A wrapper for `StoreKit.AppStore.Environment`.
+enum StoreEnvironment: String {
```
NachoSoto, line 19: I see you're using `: String` for the `Codable` implementation. Are you using the `Codable` implementation for this type though?
I can't really find where it's used. But if so, maybe add a comment mentioning what the `rawValue` is used for? Basically We want to make sure we don't inadvertently change it. Alternatively if we are using it somewhere like a request, it would be useful to add it to the snapshot tests to capture that.

Tests/StoreKitUnitTests/TestHelpers/AvailabilityChecks.swift
```diff
@@ -74,4 +74,10 @@ enum AvailabilityChecks {
         #endif
     }
 
+    static func skipIfXcode14orEarlier() throws {
```
NachoSoto, line 77: Nit:
```suggestion
    static func skipIfXcode14OrEarlier() throws {
```

Tests/StoreKitUnitTests/TestHelpers/AvailabilityChecks.swift
```diff
@@ -74,4 +74,10 @@ enum AvailabilityChecks {
...
+        #if swift(<5.9)
+        throw XCTSkip("Test for for Xcode 14 or earlier")
```
NachoSoto, line 79: This should be 
```suggestion
        throw XCTSkip("Test for for Xcode 15 or earlier")
```

Tests/StoreKitUnitTests/StoreKit2/StoreKit2TransactionFetcherTests.swift
NachoSoto: Let's add a test to verify the behavior on Xcode 14 then.

Sources/Purchasing/StoreKit2/SK2AppTransaction.swift
```diff
@@ -0,0 +1,33 @@
...
+/// A wrapper for `StoreKit.AppTransaction`.
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+internal struct SK2AppTransaction {
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    init(appTransaction: AppTransaction) {
+        self.bundleId = appTransaction.bundleID
+        self.originalApplicationVersion = appTransaction.originalAppVersion
+        self.originalPurchaseDate = appTransaction.originalPurchaseDate
+        self.environment = StoreEnvironment.init(environment: appTransaction.environment)
```
NachoSoto, line 25: Nit: no need to repeat the type
```suggestion
        self.environment = .init(environment: appTransaction.environment)
```

Sources/Purchasing/StoreKit2/StoreKit2Receipt.swift
```diff
@@ -0,0 +1,74 @@
...
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(self.environment, forKey: .environment)
+        try container.encode(self.transactions, forKey: .transactions)
+        try container.encode(self.bundleId, forKey: .bundleId)
+        try container.encode(self.originalApplicationVersion, forKey: .originalApplicationVersion)
+        try container.encode(self.originalPurchaseDate, forKey: .originalPurchaseDate)
+        let statuses = Dictionary(uniqueKeysWithValues: self.subscriptionStatus.map {
+            ($0.subscriptionGroupId, $0.renewalInfoJWSTokens)
```
NachoSoto, line 69: Is this guaranteed to be unique? If not, it will crash with duplicate subscription group IDs

Sources/Purchasing/StoreKit2/StoreKit2Receipt.swift
NachoSoto: There is but it's not trivial to implement.

Let's at least add test coverage for this using snapshot tests.

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -73,6 +76,29 @@ final class StoreKit2TransactionFetcher: StoreKit2TransactionFetcherType {
         }
     }
 
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    var receipt: StoreKit2Receipt {
```
NachoSoto, line 80: This isn't tested?

Sources/Logging/Strings/StoreKitStrings.swift
```diff
@@ -59,6 +59,16 @@ enum StoreKitStrings {
...
 
+    case sk2_unknown_environment(String)
+
+    case sk2_error_encoding_receipt(Error)
+
+    case sk2_error_fetching_app_transaction(Error)
+
+    case sk2_app_transaction_unavailable
+
+    case sk2_unverified_transaction(String, Error)
```
NachoSoto, line 70: Let's put a label to `String` since it's not obvious what it is based on the type alone:
```suggestion
    case sk2_unverified_transaction(identifier: String, Error)
```

Sources/Logging/Strings/StoreKitStrings.swift
```diff
@@ -149,6 +159,21 @@ extension StoreKitStrings: LogMessage {
         case .sk2_observing_transaction_updates:
             return "Observing StoreKit.Transaction.updates"
 
+        case let .sk2_unknown_environment(environment):
```
NachoSoto, line 162: All these 5 warnings for untested logs is code paths that could have unit tests 🙏🏻

Sources/Networking/Operations/PostReceiptDataOperation.swift
```diff
@@ -332,6 +337,13 @@ private extension EncodedAppleReceipt {
             return content.asData.hashString
         case let .receipt(data):
             return data.hashString
+        case .sk2receipt(let receipt):
```
NachoSoto, line 340: Nit for consistency with the rest:
```suggestion
        case let .sk2receipt(receipt):
```

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -90,15 +116,109 @@ extension StoreKit.VerificationResult where SignedType == StoreKit.Transaction {
...
+                return await StoreKit.Product.SubscriptionInfo.Status.all
+                    .extractValues()
+                    .reduce(into: [:]) {result, value in
+                        result[value.groupID] = value.statuses
+                    }
+            } else {
+                // `StoreKit.Product.SubscriptionInfo.Status.all` is only available starting in iOS 17.0
+                // For previous versions, we retrieve all the previously purchased transactions,
+                // and fetch the subscription status only once per subscription group.
+                var transactionBySubscriptionGroup: [String: StoreKit.Transaction] = [:]
```
NachoSoto, line 164: This could really use tests to verify the behavior on iOS 17 and also below.

NachoSoto (commented):
Looks good but I'm worried about all the untested code.
Especially we should have new tests on `BackendPostReceiptDataTests` that verify the format of how we're posting these new receipts.

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
MarkVillacampa: Its tested (see `testGeneratesReceipt`) but I wonder how Codecov generated coverage, and if it is only from the latest run test (e.g. if it the last run test was iOS 14, this method would not have been tested)

Sources/Purchasing/StoreKit2/StoreKit2Receipt.swift
MarkVillacampa: Good catch, it should be but it's not guaranteed at compile time, so I've changed this to the initializer that accepts non-unique keys

Tests/StoreKitUnitTests/TestHelpers/AvailabilityChecks.swift
MarkVillacampa: Renoved this helper as it's no longer needed

Sources/Purchasing/StoreKitAbstractions/StoreEnvironment.swift
MarkVillacampa: It's used when serializing `StoreKit2Receipt`. Added snapshot tests for the request where it's used

Sources/Purchasing/StoreKitAbstractions/StoreEnvironment.swift
MarkVillacampa: Added unit tests 👍

Sources/LocalReceiptParsing/Helpers/ReceiptStrings.swift
```diff
@@ -95,6 +96,9 @@ extension ReceiptStrings: LogMessage {
         case let .posting_jws(token, initiationSource):
             return "Posting JWS token (source: '\(initiationSource)'):\n\(token)"
 
+        case let .posting_sk2_receipt(receipt, initiationSource):
+            return "Posting StoreKit 2 receipt (source: '\(initiationSource)'):\n\(receipt)"
```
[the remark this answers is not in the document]
MarkVillacampa: Added test in `BackendPostReceiptDataTests` which should cover this.

Sources/Networking/Operations/PostReceiptDataOperation.swift
```diff
@@ -332,6 +337,13 @@ private extension EncodedAppleReceipt {
             return content.asData.hashString
         case let .receipt(data):
             return data.hashString
+        case let .sk2receipt(receipt):
+            do {
+                return try receipt.prettyPrintedData.hashString
```
MarkVillacampa, line 342: Switched to `prettyPrintedData` to generate consistent hashes on iOS 17+

Sources/Purchasing/StoreKitAbstractions/EncodedAppleReceipt.swift
```diff
@@ -29,7 +30,14 @@ extension EncodedAppleReceipt {
         case .jws(let jws):
             return jws
         case .receipt(let data):
-            return data.asFetchToken
+            return data.base64EncodedString()
+        case .sk2receipt(let receipt):
+            do {
+                return try receipt.prettyPrintedData.base64EncodedString()
```
MarkVillacampa, line 36: Switched to `prettyPrintedData` to generate consistent JSON for snapshot tests.

Sources/Networking/Operations/PostReceiptDataOperation.swift
NachoSoto: Oooh great catch

Sources/Purchasing/StoreKit2/StoreKit2Receipt.swift
NachoSoto: I just realized that this class is `Codable`, you've changed how it's _encoded_, but now how it's _decoded_, so doing:
```swift
expect(try receipt.encodeAndDecode()) == receipt
```

Would fail. I think you can change this to conform to `Encodable` only instead?

Tests/StoreKitUnitTests/StoreKit2/StoreKit2TransactionFetcherTests.swift
```diff
@@ -148,6 +148,78 @@ class StoreKit2TransactionFetcherTests: StoreKitConfigTestCase {
...
+            expect(receipt.bundleId) == Bundle.main.bundleIdentifier
+            expect(receipt.originalApplicationVersion).notTo(beEmpty())
+            expect(receipt.originalPurchaseDate).notTo(beNil())
+        } else {
+            // AppTransaction is not available on iOS 15
+            expect(receipt.bundleId).to(beEmpty())
+            expect(receipt.originalApplicationVersion).to(beNil())
+            expect(receipt.originalPurchaseDate).to(beNil())
+
+            self.logger.verifyMessageWasLogged(
```
NachoSoto, line 173: 👍🏻

Tests/StoreKitUnitTests/StoreKit2/StoreKit2TransactionFetcherTests.swift
```diff
@@ -148,6 +148,78 @@ class StoreKit2TransactionFetcherTests: StoreKitConfigTestCase {
...
+        _ = try await self.createTransaction(productID: Self.product1,
+                                             finished: false)
+        _ = try await self.createTransaction(productID: Self.product2,
+                                             finished: false)
+        let receipt = await self.fetcher.receipt
+        expect(receipt.transactions.count) ==  2
+        expect(receipt.subscriptionStatus.count) == 2
+        expect(receipt.environment) == .xcode
+
+        if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
```
NachoSoto, line 163: Awesome

Tests/StoreKitUnitTests/StoreKit2/StoreKit2TransactionFetcherTests.swift
```diff
@@ -148,6 +148,78 @@ class StoreKit2TransactionFetcherTests: StoreKitConfigTestCase {
...
+    // MARK: - receipt
+
+    func testGeneratesReceipt() async throws {
+        _ = try await self.createTransaction(productID: Self.product1,
+                                             finished: false)
+        _ = try await self.createTransaction(productID: Self.product2,
+                                             finished: false)
+        let receipt = await self.fetcher.receipt
+        expect(receipt.transactions.count) ==  2
+        expect(receipt.subscriptionStatus.count) == 2
```
NachoSoto, line 160: There's a better way to write this that provides more informative error messages:
```suggestion
        expect(receipt.transactions).to(haveCount(2))
        expect(receipt.subscriptionStatus).to(haveCount(2))
```

Tests/StoreKitUnitTests/StoreKit2/StoreKit2TransactionFetcherTests.swift
```diff
@@ -148,6 +148,78 @@ class StoreKit2TransactionFetcherTests: StoreKitConfigTestCase {
...
+            self.logger.verifyMessageWasLogged(
+                Strings.storeKit.sk2_app_transaction_unavailable,
+                level: .warn
+            )
+        }
+    }
+
+    // MARK: - unverified transactions
+
+    func testHandlesUnverifiedTransactions() async throws {
```
NachoSoto, line 182: Can you rename these to include the method that is being tested?
I'd prefix these with `testVerifiedTransaction...`

Tests/StoreKitUnitTests/StoreKit2/StoreKit2TransactionFetcherTests.swift
```diff
@@ -148,6 +148,78 @@ class StoreKit2TransactionFetcherTests: StoreKitConfigTestCase {
...
+            expect(receipt.originalPurchaseDate).to(beNil())
+
+            self.logger.verifyMessageWasLogged(
+                Strings.storeKit.sk2_app_transaction_unavailable,
+                level: .warn
+            )
+        }
+    }
+
+    // MARK: - unverified transactions
```
NachoSoto, line 180: Maybe 
```suggestion
    // MARK: - verifiedTransaction
```

Tests/StoreKitUnitTests/StoreKit2/StoreEnvironmentTests.swift
```diff
@@ -0,0 +1,66 @@
...
+    override func setUp() async throws {
+        try await super.setUp()
+        try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
+    }
+
+    // MARK: - StoreEnvironment
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    func testFromAppStoreEnvironment() throws {
+        try AvailabilityChecks.iOS16APIAvailableOrSkipTest()
```
NachoSoto, line 33: Nit: new line after these so the actual test is easier to distinguish from this
```suggestion
        try AvailabilityChecks.iOS16APIAvailableOrSkipTest()
        
```

Tests/StoreKitUnitTests/StoreKit2/StoreEnvironmentTests.swift
```diff
@@ -0,0 +1,66 @@
...
+        expect(StoreEnvironment(environment: .production)) == .production
+        expect(StoreEnvironment(environment: .sandbox)) == .sandbox
+        expect(StoreEnvironment(environment: .xcode)) == .xcode
+    }
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    func testFromUnknownAppStoreEnvironment() throws {
+        try AvailabilityChecks.iOS16APIAvailableOrSkipTest()
+        let environment = AppStore.Environment.init(rawValue: "revenuecat")
```
NachoSoto, line 42: I'm surprised this doesn't produce a lint error:
```suggestion
        let environment = AppStore.Environment(rawValue: "revenuecat")
```

Tests/StoreKitUnitTests/StoreKit2/StoreEnvironmentTests.swift
```diff
@@ -0,0 +1,66 @@
...
+        expect(StoreEnvironment(environment: environment)) == nil
+
+        self.logger.verifyMessageWasLogged(
+            Strings.storeKit.sk2_unknown_environment(String.init(describing: environment)),
```
NachoSoto, line 46:
```suggestion
            Strings.storeKit.sk2_unknown_environment(String(describing: environment)),
```

NachoSoto: Or does this work instead?
```suggestion
            Strings.storeKit.sk2_unknown_environment(environment.rawValue),
```

Tests/StoreKitUnitTests/StoreKit2/StoreEnvironmentTests.swift
```diff
@@ -0,0 +1,66 @@
...
+        expect(StoreEnvironment(environment: .production)) == .production
+        expect(StoreEnvironment(environment: .sandbox)) == .sandbox
+        expect(StoreEnvironment(environment: .xcode)) == .xcode
+    }
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    func testFromUnknownAppStoreEnvironment() throws {
+        try AvailabilityChecks.iOS16APIAvailableOrSkipTest()
+        let environment = AppStore.Environment.init(rawValue: "revenuecat")
+        expect(StoreEnvironment(environment: environment)) == nil
```
NachoSoto, line 43: You need to use `.to(beNil())` instead of comparing with `nil` (just a Nimble implementation detail).

Tests/StoreKitUnitTests/StoreKit2/StoreEnvironmentTests.swift
```diff
@@ -0,0 +1,66 @@
...
+    }
+
+    func testStoreEnvironmentFromString() {
+        expect(StoreEnvironment(environment: "Production")) == .production
+        expect(StoreEnvironment(environment: "Sandbox")) == .sandbox
+        expect(StoreEnvironment(environment: "Xcode")) == .xcode
+    }
+
+    func testFromUnknownString() {
+        expect(StoreEnvironment(environment: "revenuecat")) == nil
```
NachoSoto, line 58: Ditto (IIRC if the actual value is not nil this will fail with a confusing error):
```suggestion
        expect(StoreEnvironment(environment: "revenuecat")).to(beNil())
```

Tests/StoreKitUnitTests/StoreKit2/StoreEnvironmentTests.swift
```diff
@@ -0,0 +1,66 @@
...
+
+        self.logger.verifyMessageWasLogged(
+            Strings.storeKit.sk2_unknown_environment(String.init(describing: "revenuecat")),
```
NachoSoto, line 61:
```suggestion
            Strings.storeKit.sk2_unknown_environment("revenuecat"),
```

Tests/StoreKitUnitTests/StoreTransactionTests.swift
```diff
@@ -64,6 +64,7 @@ class StoreTransactionTests: StoreKitConfigTestCase {
         expect(transaction.hasKnownPurchaseDate) == false
         expect(transaction.hasKnownTransactionIdentifier) == true
         expect(transaction.jwsRepresentation).to(beNil())
+        expect(transaction.environment).to(beNil())
```
NachoSoto, line 67: Maybe add this to `testSK1DetailsWrapCorrectly` too?

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostReceiptDataTests/iOS15-testPostsSK2XcodeReceiptWithProductDataCorrectly.1.json
```diff
@@ -0,0 +1,26 @@
+{
```
NachoSoto, line 1: Awesome!

FYI looks like this is missing iOS 12 through 14 snapshots. Unfortunately that's not caught by CI because when they're missing, they get generated (making the test fail), but then the test is retried and it passes as a false positive.

_I have a [PR to fix that](https://github.com/pointfreeco/swift-snapshot-testing/pull/768)_

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostReceiptDataTests/iOS16-testPostsSK2XcodeReceiptWithProductDataCorrectly.1.json
```diff
@@ -0,0 +1,26 @@
...
+    "body" : {
+      "app_user_id" : "user",
+      "attributes" : {
+        "$attConsentStatus" : {
+          "updated_at_ms" : 1678307200000,
+          "value" : "authorized"
+        }
+      },
+      "currency" : "USD",
+      "fetch_token" : "ewogICJidW5kbGVfaWQiIDogIjEyM19idW5kbGVfaWQiLAogICJlbnZpcm9ubWVudCIgOiAieGNvZGUiLAogICJvcmlnaW5hbF9hcHBsaWNhdGlvbl92ZXJzaW9uIiA6ICIxMjNfb3JpZ2luYWxfYXBwbGljYXRpb25fdmVyc2lvbiIsCiAgIm9yaWdpbmFsX3B1cmNoYXNlX2RhdGUiIDogIjE5NzAtMDEtMDFUMDA6MDI6MDNaIiwKICAic3Vic2NyaXB0aW9uX3N0YXR1cyIgOiB7CiAgICAiMTIzX3N1YnNjcmlwdGlvbl9pZCIgOiBbCiAgICAgICIxMjNfcmVuZXdhbF9pbmZvX2p3c190b2tlbiIKICAgIF0KICB9LAogICJ0cmFuc2FjdGlvbnMiIDogWwogICAgIjEyM190cmFuc2FjdGlvbl9qd3NfdG9rZW4iCiAgXQp9",
```
NachoSoto, line 15: Awesome, now we have coverage for this

Tests/UnitTests/Networking/Backend/BackendPostReceiptDataTests.swift
```diff
@@ -984,6 +1015,17 @@ private extension BaseBackendPostReceiptDataTests {
     static let receipt = EncodedAppleReceipt.receipt("an awesome receipt".asData)
     static let receipt2 = EncodedAppleReceipt.receipt("an awesomeer receipt".asData)
     static let jws = EncodedAppleReceipt.jws("an awesomer jws token")
+    static let sk2receipt = EncodedAppleReceipt.sk2receipt(StoreKit2Receipt(
```
NachoSoto, line 1018: Nit, but this might be clearer avoiding the repetition:
```suggestion
    static let sk2receipt = EncodedAppleReceipt.sk2receipt(.init(
```

Tests/UnitTests/Networking/Backend/BackendPostReceiptDataTests.swift
```diff
@@ -984,6 +1015,17 @@ private extension BaseBackendPostReceiptDataTests {
...
+        environment: .xcode,
+        subscriptionStatus: [StoreKit2Receipt.SubscriptionStatus(
```
NachoSoto, line 1020: Ditto you can rely on `.init` to avoid the extra types:
```suggestion
        subscriptionStatus: [
          .init(
```

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -270,6 +249,30 @@ private extension TransactionPoster {
...
+            ) { receiptData, receiptURL in
+                if let receiptData = receiptData, !receiptData.isEmpty {
+                    completion(.success(.receipt(receiptData)))
+                } else {
+                    completion(.failure(BackendError.missingReceiptFile(receiptURL)))
+                }
+            }
+        }
+    }
+
```
MarkVillacampa, line 275: Extracted the logic to craft an `EncodedAppleReceipt` to its own method as it was getting complex

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -90,31 +93,29 @@ final class TransactionPoster: TransactionPosterType {
...
+                                   subscriberAttributes: nil,
+                                   completion: completion)
+            return
+        }
+
+        self.fetchEncodedReceipt(transaction: transaction) { result in
+            switch result {
+            case .success(let encodedReceipt):
+                self.product(with: productIdentifier) { product in
+                    self.postReceipt(transaction: transaction,
```
NachoSoto, line 108: Much better 👏🏻

Sources/Purchasing/StoreKitAbstractions/SK2StoreTransaction.swift
```diff
@@ -25,6 +25,19 @@ internal struct SK2StoreTransaction: StoreTransactionType {
...
+
+        } else {
+            #if VISION_OS
+            self.environment = environment ?? StoreEnvironment(environment: sk2Transaction.environment)
+            #else
+            self.environment = environment ?? StoreEnvironment(
+                environment: sk2Transaction.environmentStringRepresentation
+            )
+            #endif
+        }
```
NachoSoto, line 39: What do you think about extracting this to a new `StoreEnvironment` constructor that takes a `SK2Transaction`?

Then this becomes one line:
```swift
self.environment = environment ?? .init(sk2Transaction)
```

Tests/StoreKitUnitTests/StoreKit2/StoreEnvironmentTests.swift
```diff
@@ -0,0 +1,67 @@
...
+import Nimble
+@testable import RevenueCat
+import StoreKitTest
+import XCTest
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+@MainActor
+class StoreEnvironmentTests: StoreKitConfigTestCase {
+
+    @MainActor
```
NachoSoto, line 23: This isn't needed cause it's in the parent type already.

Tests/UnitTests/Purchasing/Purchases/TransactionPosterTests.swift
```diff
@@ -127,6 +127,46 @@ class TransactionPosterTests: TestCase {
         expect(self.mockTransaction.finishInvoked) == true
     }
 
+    func testHandlePurchasedTransactionSendsSK2Receipt() throws {
```
NachoSoto, line 130: <3

NachoSoto (approved):
Reminder that we should add integration tests for this as well, but we can do it in a separate PR. Let me know if you need help with that.

Awesome job with this! 👏🏻

Sources/Purchasing/StoreKit2/StoreKit2Receipt.swift
```diff
@@ -0,0 +1,115 @@
...
+//  StoreKit2Receipt.swift
+//
+//  Created by MarkVillacampa on 26/10/23.
+
+import StoreKit
+
+/// A type that resembles the structure of a StoreKit 1 receipt using StoreKit 2 data.
+struct StoreKit2Receipt: Equatable {
+
+    struct SubscriptionState: RawRepresentable, Equatable {
```
NachoSoto, line 19: Why not use a regular enum?

MarkVillacampa: This wraps a [RenewalState](https://developer.apple.com/documentation/storekit/product/subscriptioninfo/renewalstate/) which is itself a rawrepresentable int struct.

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostReceiptDataTests/iOS15-testPostsSK2XcodeReceiptWithProductDataCorrectly.1.json
MarkVillacampa: Added

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -299,6 +306,12 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
 
         try self.testSession.approveAskToBuyTransaction(identifier: transaction.identifier)
 
+        // In JWS mode, transaction takes a bit longer to be processed after `approveAskToBuyTransaction`
+        // We need to wait so `restorePurchases` actually posts it.
+        if self.usesStoreKit2JWS {
+            try? await Task.sleep(nanoseconds: DispatchTimeInterval.milliseconds(500).nanoseconds)
```
NachoSoto, line 312: 🥶🙈 This is a recipe for more flakiness. 
Can we use something like `waitUntilUnfinishedTransactions`? (or a similar method)

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -629,7 +642,8 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
         try await self.purchaseMonthlyProduct()
 
         // 8. Verify transaction is posted as a purchase.
-        self.logger.verifyMessageWasLogged("Posting receipt (source: 'purchase')")
+        try await self.verifyReceiptIsEventuallyPosted()
```
NachoSoto, line 645: 👍🏻

Tests/BackendIntegrationTests/StoreKitObserverModeIntegrationTests.swift
```diff
@@ -81,13 +81,21 @@ class StoreKit2ObserverModeIntegrationTests: StoreKit1ObserverModeIntegrationTes
 
 }
 
+class StoreKit2JWSObserverModeIntegrationTests: StoreKit2ObserverModeIntegrationTests {
```
NachoSoto, line 84: Awesome

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -22,6 +22,13 @@ class StoreKit2IntegrationTests: StoreKit1IntegrationTests {
 
 }
 
+class StoreKit2JWSIntegrationTests: StoreKit1IntegrationTests {
```
NachoSoto, line 25: Should this inherit `StoreKit12ntegrationTests` instead?

Tests/BackendIntegrationTests/StoreKitObserverModeIntegrationTests.swift
```diff
@@ -81,13 +81,21 @@ class StoreKit2ObserverModeIntegrationTests: StoreKit1ObserverModeIntegrationTes
...
 class StoreKit1ObserverModeIntegrationTests: BaseStoreKitObserverModeIntegrationTests {
 
     override class var storeKit2Setting: StoreKit2Setting { return .disabled }
 
     func testPurchaseOutsideTheAppPostsReceipt() async throws {
         try self.testSession.buyProduct(productIdentifier: Self.monthlyNoIntroProductID)
 
+        // In JWS mode, transaction takes a bit longer to be processed after `buyProduct`
+        // We need to wait so `restorePurchases` actually posts it.
+        try? await Task.sleep(nanoseconds: DispatchTimeInterval.seconds(2).nanoseconds)
```
NachoSoto, line 97: Ditto, let's avoid these sleeps. Let's replace it with an explicit wait for whatever we need (See `StoreKitTestHelpers` for examples).

Tests/StoreKitUnitTests/StoreKit2/StoreEnvironmentTests.swift
```diff
@@ -0,0 +1,66 @@
...
+import StoreKitTest
+import XCTest
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+@MainActor
+class StoreEnvironmentTests: StoreKitConfigTestCase {
+
+    override func setUp() async throws {
+        try await super.setUp()
+        try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
```
NachoSoto, line 25: `StoreEnvironment` is available on all versions. I know some tests do require iOS 16, but others like `testStoreEnvironmentFromString` can run on all, right?

Tests/StoreKitUnitTests/StoreKit2/StoreEnvironmentTests.swift
```diff
@@ -0,0 +1,66 @@
...
+//  Created by Mark Villacampa on 11/15/23.
+
+import Nimble
+@testable import RevenueCat
+import StoreKitTest
+import XCTest
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+@MainActor
+class StoreEnvironmentTests: StoreKitConfigTestCase {
```
NachoSoto, line 21: I don't think this needs to inherit `StoreKitConfigTestCase`, it can be a regular unit test (and moved to `UnitTests`), right?

Tests/UnitTests/Mocks/MockStoreKit2TransactionListener.swift
```diff
@@ -61,7 +63,7 @@ final class MockStoreKit2TransactionListener: StoreKit2TransactionListenerType {
         self.invokedHandleParametersList.append((.init(purchaseResult), ()))
 
         let transaction: StoreTransaction? = self.mockTransaction.value.map {
-            StoreTransaction(sk2Transaction: $0, jwsRepresentation: "")
+            StoreTransaction(sk2Transaction: $0, jwsRepresentation: mockJWSToken, environment: mockEnvironment)
```
NachoSoto, line 66: Explicit self
```suggestion
            StoreTransaction(sk2Transaction: $0, jwsRepresentation: self.mockJWSToken, environment: self.mockEnvironment)
```

Tests/UnitTests/Purchasing/Purchases/TransactionPosterTests.swift
```diff
@@ -127,6 +127,50 @@ class TransactionPosterTests: TestCase {
...
+        self.transactionFetcher.stubbedReceipt = receipt
+        self.productsManager.stubbedProductsCompletionResult = .success([StoreProduct(sk1Product: product)])
+        self.backend.stubbedPostReceiptResult = .success(Self.mockCustomerInfo)
+
+        let result = try self.handleTransaction(transactionData)
+        expect(result).to(beSuccess())
+        expect(result.value) === Self.mockCustomerInfo
+
+        expect(self.backend.invokedPostReceiptData) == true
+        expect(self.backend.invokedPostReceiptDataParameters?.transactionData).to(match(transactionData))
```
NachoSoto, line 168: Why `match`?

MarkVillacampa: Copied this bit from other tests in the file. It's a custom matcher we have:

https://github.com/RevenueCat/purchases-ios/blob/main/Tests/UnitTests/Purchasing/Purchases/TransactionPosterTests.swift#L392-L402

NachoSoto: Oh I forgot 🙈

Sources/Purchasing/StoreKitAbstractions/StoreEnvironment.swift
```diff
@@ -0,0 +1,78 @@
...
+        }
+    }
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    init?(sk2Transaction: SK2Transaction) {
+        if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
+            self.init(environment: sk2Transaction.environment)
+
+        } else {
+            #if VISION_OS
```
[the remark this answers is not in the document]
MarkVillacampa: My bad, thought it would not run the `#available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)` branch. Removed.

Sources/Purchasing/StoreKitAbstractions/SK2StoreTransaction.swift
```diff
@@ -16,14 +16,20 @@ import StoreKit
 @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
 internal struct SK2StoreTransaction: StoreTransactionType {
 
-    init(sk2Transaction: SK2Transaction, jwsRepresentation: String) {
+    /// - Parameter environmentOverride: Overrides the environment from the StoreKit 2 transaction.
+    /// Used to override the default `Xcode` environment when running tests.
+    init(sk2Transaction: SK2Transaction,
+         jwsRepresentation: String,
+         environmentOverride: StoreEnvironment? = nil
+    ) {
```
NachoSoto, line 24: Nit: for symmetry the closing `)` should be in the same line as the last parameter, or put the first parameter on its own line too:
```suggestion
    init(
         sk2Transaction: SK2Transaction,
         jwsRepresentation: String,
         environmentOverride: StoreEnvironment? = nil
    ) {
```

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -299,6 +305,12 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
 
         try self.testSession.approveAskToBuyTransaction(identifier: transaction.identifier)
 
+        // In JWS mode, transaction takes a bit longer to be processed after `approveAskToBuyTransaction`
+        // We need to wait so `restorePurchases` actually posts it.
+        if self.usesStoreKit2JWS {
```
NachoSoto, line 310: Actually I think it would make sense to do this always, not only in this mode?

Tests/UnitTests/Purchasing/StoreKitAbstractions/StoreEnvironmentTests.swift
```diff
@@ -0,0 +1,60 @@
...
+//
+//  Created by Mark Villacampa on 11/15/23.
+
+import Nimble
+@testable import RevenueCat
+import StoreKit
+import XCTest
+
+@MainActor
+class StoreEnvironmentTests: TestCase {
```
NachoSoto, line 20: 👍🏻

Tests/UnitTests/Purchasing/StoreKitAbstractions/StoreEnvironmentTests.swift
```diff
@@ -0,0 +1,60 @@
...
+
+    // MARK: - StoreEnvironment
```
NachoSoto, line 22: Nit: this isn't needed since it's already inside `StoreEnvironmentTests`.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1309,11 +1311,13 @@ private extension PurchasesOrchestrator {
                 return
             }
 
+            let receipt = await encodedReceipt(transaction: transaction, jwsRepresentation: jwsRepresentation)
```
NachoSoto, line 1314:
```suggestion
            let receipt = await self.encodedReceipt(transaction: transaction, jwsRepresentation: jwsRepresentation)
```

NachoSoto (approved):

NachoSoto (changes requested):
I'm concerned about the sleep delay. Are you sure the production implementation needs that? Look at how we use `.retryUntilProductIsFound` only in integration tests to work around `SKTestConfig` doing this.

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -90,15 +115,147 @@ extension StoreKit.VerificationResult where SignedType == StoreKit.Transaction {
...
+    private var verifiedTransactions: [StoreTransaction] {
+        get async {
+            return await StoreKit.Transaction.all
+                .compactMap { $0.verifiedStoreTransaction }
+                .extractValues()
+        }
+    }
+
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    func verifiedTransactions(containing transaction: StoreTransactionType) async -> [StoreTransaction] {
```
NachoSoto, line 166: Let's add a docstring explaining what this does, I don't think it's obvious reading the signature.

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -90,15 +115,147 @@ extension StoreKit.VerificationResult where SignedType == StoreKit.Transaction {
...
+            return nil
+        }
+    }
+
+}
+
+extension StoreKit2TransactionFetcher {
+
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    private var verifiedTransactions: [StoreTransaction] {
```
NachoSoto, line 157: Nit: move the `private` methods below so the public ones are more visible at the top.

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -90,15 +115,147 @@ extension StoreKit.VerificationResult where SignedType == StoreKit.Transaction {
...
+            retries += 1
+            let verifiedTransactions = await verifiedTransactions
+            if verifiedTransactions.contains(where: { $0.id == transaction.transactionIdentifier }) {
+                return verifiedTransactions
+            } else {
+                Logger.appleWarning(
+                    Strings.storeKit.sk2_receipt_missing_purchase(transactionId: transaction.transactionIdentifier)
+                )
+            }
+            try? await Task.sleep(nanoseconds: UInt64(sleepDuration.nanoseconds))
```
NachoSoto, line 182: Let's add a comment in this implementation to clarify that even though it's not efficient to introduce arbitrary delays, it's only applicable for SKTestSession / SK config files.

NachoSoto (commented):
Thanks for clarifying that this is only for test environments 👍🏻

Sources/FoundationExtensions/AsyncExtensions.swift
```diff
@@ -112,6 +112,44 @@ internal enum Async {
...
+    ///     } else {
+    ///         return (false, receipt)
+    ///     }
+    /// }
+    /// ```
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    static func retry<T>(
+        maximumRetries: Int = 5,
+        pollInterval: DispatchTimeInterval = DispatchTimeInterval.milliseconds(300),
+        until value: @Sendable () async -> (Bool, T)
```
NachoSoto, line 134: Let's give these names so it's more clear:
```suggestion
        until value: @Sendable () async -> (shouldRetry: Bool, result: T)
```

Sources/FoundationExtensions/AsyncExtensions.swift
```diff
@@ -112,6 +112,44 @@ internal enum Async {
...
+    ) async -> T {
+        var lastValue: T
+        var retries = 0
+
+        repeat {
+            retries += 1
+            let currentValue = await value()
```
NachoSoto, line 141: Then here you can do 
```suggestion
            let (shouldRetry, value) = await value()
            lastValue = value
```

Sources/FoundationExtensions/AsyncExtensions.swift
```diff
@@ -112,6 +112,44 @@ internal enum Async {
...
+    ///         return (true, receipt)
+    ///     } else {
+    ///         return (false, receipt)
+    ///     }
+    /// }
+    /// ```
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    static func retry<T>(
+        maximumRetries: Int = 5,
+        pollInterval: DispatchTimeInterval = DispatchTimeInterval.milliseconds(300),
```
NachoSoto, line 133:
```suggestion
        pollInterval: DispatchTimeInterval =.milliseconds(300),
```

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -90,15 +116,144 @@ extension StoreKit.VerificationResult where SignedType == StoreKit.Transaction {
...
+    }
+
+    /// Returns an an array of  all verified `StoreTransaction`s, making sure `transaction` is contained in it.
+    ///
+    /// When using SKTestSession / SK config files, a newly generated transaction may not appear immediately
+    /// in `Transaction.all`, so we need to retry and introduce a small synthetic delay to make sure it is.s
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    private func verifiedTransactions(containing transaction: StoreTransactionType) async -> [StoreTransaction] {
+
+        return await Async.retry {
```
NachoSoto, line 173: Much better 👍🏻

Sources/FoundationExtensions/AsyncExtensions.swift
```diff
@@ -112,6 +112,44 @@ internal enum Async {
...
+
+        repeat {
+            retries += 1
+            let (shouldRetry, result) = await value()
+            if shouldRetry {
+                return result
+            } else {
+                lastValue = result
+                try? await Task.sleep(nanoseconds: UInt64(pollInterval.nanoseconds))
+            }
```
NachoSoto, line 147: This is backwards now (sorry I should have mentioned that `shouldRetry` was the opposite logic you had).

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -90,15 +116,144 @@ extension StoreKit.VerificationResult where SignedType == StoreKit.Transaction {
...
+            let verifiedTransactions = await verifiedTransactions
+            if verifiedTransactions.contains(where: { $0.id == transaction.transactionIdentifier }) {
+                return  (true, verifiedTransactions)
+            } else {
+                Logger.appleWarning(
+                    Strings.storeKit.sk2_receipt_missing_purchase(transactionId: transaction.transactionIdentifier)
+                )
+                return  (false, verifiedTransactions)
```
NachoSoto, line 181: Then here I'd write it like this so it's more readable. Otherwise it's not clear what `true` or `false` means.
```suggestion
                return (shouldRetry: true, verifiedTransactions)
```

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
NachoSoto, line 173: This could check a total of 16 times which seems like a lot. Should we increase the delay or reduce the timeout?

MarkVillacampa: It's not a timeout it's the maximum number of retries, so it should try max 5 times by default.

NachoSoto: Oh yeah what am I saying 🤦🏻 this is good then.

Sources/FoundationExtensions/AsyncExtensions.swift
```diff
@@ -139,7 +139,7 @@ internal enum Async {
         repeat {
             retries += 1
             let (shouldRetry, result) = await value()
-            if shouldRetry {
+            if !shouldRetry {
```
NachoSoto, line 142: Nit: change this to `shouldRetry` and move this to the else and viceversa. Easier to read than the reverse logic.

Sources/FoundationExtensions/AsyncExtensions.swift
```diff
@@ -112,6 +112,44 @@ internal enum Async {
...
+    ///     let receipt = fetchReceipt()
+    ///     if receipt.contains(transaction) {
+    ///         return (true, receipt)
+    ///     } else {
+    ///         return (false, receipt)
+    ///     }
+    /// }
+    /// ```
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    static func retry<T>(
```
NachoSoto, line 131: It would be nice to add tests for this in `AsyncExtensionsTests` but we can do that in a separate PR.

Sources/FoundationExtensions/AsyncExtensions.swift
```diff
@@ -112,6 +112,44 @@ internal enum Async {
...
+    /// Runs the given block `maximumRetries` times at most, at `pollInterval` times until the
+    /// block returns a tuple where the first argument is true, and the second is the expected value.
+    /// After the maximum retries, returns the last seen value.
+    ///
+    /// Example:
+    /// ```swift
+    /// let receipt = await Async.retry {
+    ///     let receipt = fetchReceipt()
+    ///     if receipt.contains(transaction) {
+    ///         return (true, receipt)
```
NachoSoto, line 124: Maybe update this with the tuple name too so it's clear?

NachoSoto (approved):

PR #3373, opened 2023-11-01
`Offering`: new `getMetadataValue` with `Decodable` type

NachoSoto:
This allows to easily extract data from `Offering.metadata` in a structured way.

Example:
```swift
struct Data: Decodable {
    var number: Int
}

let elements: [Data] = offering.getMetadataValue(for: "elements") ?? []
let element: Data? = offering.getMetadataValue(for: "element")
```

### DIFF
--- a/Sources/Purchasing/Offering.swift
+++ b/Sources/Purchasing/Offering.swift
@@ -230,6 +230,7 @@ extension Offering {
     /// - Returns: the `metadata` value associated to `key` for the expected `Decodable` type,
     /// or `nil` if not found.
     /// - Throws: Error if the content couldn't be deserialized to the expected type.
+    /// - Note: This decodes JSON using `JSONDecoder.KeyDecodingStrategy.convertFromSnakeCase`.
     ///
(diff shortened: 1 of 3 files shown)


Sources/Purchasing/Offering.swift
```diff
@@ -215,17 +215,37 @@ import Foundation
...
             return `default`
         }
         return value
     }
 
+    ///
+    /// - Returns: the `metadata` value associated to `key` for the expected `Decodable` type,
+    /// or `nil` if not found.
+    /// - Throws: Error if the content couldn't be deserialized to the expected type.
+    ///
```
joshdholtz, line 233: Should we mention which `KeyDecodingStrategy` its using?

joshdholtz (commented):
This looks amazing! Just one quick question about `KeyDecodingStrategy`

Sources/Purchasing/Offering.swift
NachoSoto: Done.

joshdholtz (approved):
:shipit:

PR #3381, opened 2023-11-02
`Paywalls`: fix template 4 layout bug on iOS 16

NachoSoto:
This only reproduces on iOS 16 and using the new `displayCloseButton: true` (because it embeds `PaywallView` in a `NavigationView`).

simulator_screenshot_EB70C623-426F-44C5-92BA-0BDA0D3F5AFE

The problem was that the template used `ViewDimentionPreferenceKey` in several parts of the hierarchy and `SwiftUI` mixes both.
I work around it by separating `onWidthChange` and `onHeightChange`.

### DIFF
--- a/RevenueCatUI/Modifiers/ViewExtensions.swift
+++ b/RevenueCatUI/Modifiers/ViewExtensions.swift
@@ -192,9 +192,9 @@ extension View {
             .onPreferenceChange(ViewSizePreferenceKey.self, perform: closure)
     }
 
-    /// Invokes the given closure with the dimension specified by `axis` changes whenever it changes.
-    func onSizeChange(
-        _ axis: Axis,
+    /// Invokes the given closure with the view width whenever it changes.
+    @ViewBuilder
+    func onWidthChange(
         _ closure: @escaping (CGFloat) -> Void
     ) -> some View {
         self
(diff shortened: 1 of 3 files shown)


RevenueCatUI/Modifiers/ViewExtensions.swift
```diff
@@ -272,14 +288,29 @@ private struct RoundedCorner: Shape {
...
+private protocol ViewDimensionPreferenceKey: PreferenceKey where Value == CGFloat {}
 
-    typealias Value = CGFloat
+/// `PreferenceKey` for keeping track of a view width.
+private struct ViewWidthPreferenceKey: ViewDimensionPreferenceKey {
 
     static var defaultValue: Value = 10
 
-    static func reduce(value: inout Value, nextValue: () -> Value) {
+    static func reduce(value: inout CGFloat, nextValue: () -> CGFloat) {
```
tonidero, line 299: Is it possible to move this static method to the protocol considering it's repeated between width and height?

tonidero (approved):
Just a question, but I'm not sure if that's possible. Looks good though

RevenueCatUI/Modifiers/ViewExtensions.swift
NachoSoto: Done!

PR #3388, opened 2023-11-06
`Paywalls`: fix Turkish translation

iremkaraoglu:
I updated translation of `Privacy`, `Restore`, `Terms`, `Lifetime` and `Continue`.

### DIFF
--- a/RevenueCatUI/Resources/tr.lproj/Localizable.strings
+++ b/RevenueCatUI/Resources/tr.lproj/Localizable.strings
@@ -1,11 +1,11 @@
 "OK" = "TAMAM";
 "All subscriptions" = "Tüm abonelikler";
-"Privacy" = "Mahremiyet";
+"Privacy" = "Gizlilik";
 "Privacy policy" = "Gizlilik Politikası";
 "Purchases restored successfully!" = "Satın almalar başarıyla geri yüklendi!";
-"Restore" = "Eski haline getirmek";
+"Restore" = "Geri yükle";
 "Restore purchases" = "Satın alınanları geri yükle";
-"Terms" = "şartlar";
+"Terms" = "Şartlar";
 "Terms and conditions" = "Şartlar ve koşullar";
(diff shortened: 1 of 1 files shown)


RevenueCatUI/Resources/tr.lproj/Localizable.strings
```diff
@@ -1,19 +1,19 @@
 "OK" = "TAMAM";
 "All subscriptions" = "Tüm abonelikler";
-"Privacy" = "Mahremiyet";
+"Privacy" = "Gizlilik";
 "Privacy policy" = "Gizlilik Politikası";
 "Purchases restored successfully!" = "Satın almalar başarıyla geri yüklendi!";
-"Restore" = "Eski haline getirmek";
+"Restore" = "Geri yükle";
```
NachoSoto, line 6: Awesome thank you so much for improving these! 🙏🏻 

Just one question: is the longer one below still correct?
> "Restore purchases" = "Satın alınanları geri yükle";

iremkaraoglu: Thank you @NachoSoto 🙏🏻 It seems accurate and meaningful. If a better and clearer wording comes to my mind, I'll let you know.

NachoSoto: Awesome thanks again for your contribution!

mavideniz: Hi @iremkaraoglu, what do you think `"Satın almaları geri yükle"`? Because it is used as "satın alma" in this line, I thought it would be better if it was the same on both sides.
`"Purchases restored successfully!" = "Satın almalar başarıyla geri yüklendi!";`

iremkaraoglu: Hey @mavideniz, I see your point. Yes, it makes sense. 🙌🏻

PR #3392, opened 2023-11-07
`RevenueCatUI`: added support to other deployment targets

NachoSoto:
Follow up to #3378.

This allows us to link both `RevenueCat` and `RevenueCatUI` from `PurchasesHybridCommon` supporting the same platforms.
The actual types might not be available in some platforms / versions, but we can detect that at compile time.

See https://app.circleci.com/pipelines/github/RevenueCat/purchases-hybrid-common/2564/workflows/ea6a4e24-e1eb-4c7a-a930-173945732999/jobs/10352

### DIFF
--- a/RevenueCatUI.podspec
+++ b/RevenueCatUI.podspec
@@ -16,7 +16,8 @@ Pod::Spec.new do |s|
   s.framework      = 'SwiftUI'
   s.swift_version  = '5.7'
 
-  # Technically PaywallView isn't available in all these platforms / versions but this can be detected at compile time.
+  # RevenueCatUI APIs are not available in all these platforms / versions, however retaining this support at the Pod level 
+  # allows us to depend on it in the same platforms as RevenueCat.
   # Opening support allows us to depend on it in the same platforms as RevenueCat.
   s.ios.deployment_target = '11.0'
   s.watchos.deployment_target = '6.2'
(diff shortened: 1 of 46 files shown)


fastlane/Fastfile
```diff
@@ -960,7 +960,7 @@ lane :check_pods do
     verbose: true,
     podspec: 'RevenueCatUI.podspec',
     include_podspecs: 'RevenueCat.podspec',
-    platforms: 'ios',
+    platforms: 'ios,osx,tvos,watchos',
```
NachoSoto, line 963: This will be linted now

tonidero (approved):

RevenueCatUI.podspec
```diff
@@ -16,9 +16,12 @@ Pod::Spec.new do |s|
   s.framework      = 'SwiftUI'
   s.swift_version  = '5.7'
 
-  # Technically PaywallView isn't available until iOS 15,
-  # but this can be detected at compile time.
+  # Technically PaywallView isn't available in all these platforms / versions but this can be detected at compile time.
```
aboedo, line 19: the language here makes it seem like a hack, but it is actually a reasonable way of managing this. Perhaps we can make this clearer? 
```suggestion
  # RevenueCatUI APIs are not available in all these platforms / versions, however retaining this support at the Pod level 
  # allows us to depend on it in the same platforms as RevenueCat.
```

aboedo (approved):

PR #3404, opened 2023-11-13
Mark Attribution.setOnesignalUserID as deprecated

Raquel10-RevenueCat:
We need to support OneSignal aliases as a prerequisite to allow using the setOneSignalUserId function in the intended way. Supporting OneSignal aliases is not yet prioritized, so we will mark setOneSignalUserId as deprecated until this functionality is fully supported.

### Checklist


NachoSoto (dismissed):
This is part of the public API now so we can't remove it without a major release.

aboedo:
are API tests mixed in with backend integration tests? I'd have expected this to have an explicit failed test

NachoSoto:
@aboedo they run as part of the regular tests, but this PR is also reverting the APITester changes.

Raquel10-RevenueCat:
> This is part of the public API now so we can't remove it without a major release.

If we want to mark this method as deprecated, is adding a comment enough or should I do something else? (I see we have a Deprecations file but not sure if I should do anything there)

Sources/Purchasing/Purchases/Attribution.swift
```diff
@@ -241,13 +241,14 @@ public extension Attribution {
...
      * Subscriber attribute associated with the OneSignal User ID for the user.
-     * Required for the RevenueCat OneSignal integration with versions v11.0 and above.
+     * Deprecated. Use `setOnesignalID` instead.
      *
      * #### Related Articles
      * - [OneSignal RevenueCat Integration](https://docs.revenuecat.com/docs/onesignal)
      *
      *- Parameter onesignalUserID: Empty String or `nil` will delete the subscriber attribute.
      */
+    @available(*, deprecated, message: "This API is not currently available. Contact support if you need to use it.")
```
NachoSoto, line 251: Nice 👍🏻 

You can add a reference to the other one too:
```suggestion
    @available(
      *, 
      deprecated,
      message: "This API is not currently available. Contact support if you need to use it.",
      renamed: "setOnesignalID"
      )
```

NachoSoto (approved):
Thank you!

I think we'll need to update API testers, I can help with that.
Also let's decide on whether this is a `feat` to force a minor bump before merging this.

Tests/APITesters/SwiftAPITester/SwiftAPITester/AttributionAPI.swift
```diff
@@ -43,9 +43,6 @@ func checkAttributionAPI() {
     attribution.setOnesignalID("")
     attribution.setOnesignalID(nil)
 
-    attribution.setOnesignalUserID("")
```
NachoSoto, line 46: We need to keep these and the tests because they're still part of the API (so we want to verify the API and have tests).
Instead you can move the API tests to a method that'd also marked with `@available(*, deprecated)`, and mark the tests with the same annotation.

NachoSoto, line 46: We need to keep these and the tests because they're still part of the API (so we want to verify the API and have tests).
Instead you can move the API tests to a method that'd also marked with `@available(*, deprecated)`, and mark the tests with the same annotation.

Tests/APITesters/SwiftAPITester/SwiftAPITester/AttributionAPI.swift
```diff
@@ -81,3 +78,11 @@ func checkAttributionAPI() {
     }
     #endif
 }
+
+@available(*, deprecated)
+func checkAttributionAPIDeprecated() {
+    attribution.setAttributes([String: String]())
```
NachoSoto, line 84: This isn't deprecated though?

Raquel10-RevenueCat (commented):
I think the tests that are failing are marked as flaky and not related with this change

Tests/UnitTests/SubscriberAttributes/PurchasesSubscriberAttributesTests.swift
```diff
@@ -409,16 +409,6 @@ class PurchasesSubscriberAttributesTests: TestCase {
             .to(equal((nil, purchases.appUserID)))
     }
 
-    func testSetAndClearOnesignalUserID() {
```
NachoSoto, line 412: We need to bring the tests back. You can mark the tests as `@(*, deprecated)` as well so it compiles, and they'll still run.

NachoSoto:
@Raquel10-RevenueCat is this still needed?

PR #3405, opened 2023-11-13
Adds `X-Storefront` request header for App Store Storefront

joshdholtz:
### Motivation

Allow backend to determine Storefront of customer

### Description

- Adds `X-Storefront` as a default header
- Uses StoreKit 1 implementation since its not async
  - StoreKit 2 implementation is async and would require a some restructuring that we don't need right now

### DIFF
--- a/Sources/Misc/SystemInfo.swift
+++ b/Sources/Misc/SystemInfo.swift
@@ -53,6 +53,14 @@ class SystemInfo {
         return self.sandboxEnvironmentDetector.isSandbox
     }
 
+    var storefront: StorefrontType? {
+        if #available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, macCatalyst 13.1, *) {
+            return Storefront.sk1CurrentStorefrontType
+        } else {
+            return nil
+        }
+    }
+
     static var frameworkVersion: String {
         return "4.31.0-SNAPSHOT"
     }
--- a/Tests/UnitTests/Networking/HTTPClientTests.swift
+++ b/Tests/UnitTests/Networking/HTTPClientTests.swift
@@ -306,24 +306,23 @@ final class HTTPClientTests: BaseHTTPClientTests<MockETagManager> {
     }
 
     func testRequestsWithoutStorefrontDoNotSendHeader() {
-        let headerName = "X-Storefrontt"
-        self.systemInfo.stubbedStorefront = nil
-
+        let headerName = "X-Storefront"
         let request = HTTPRequest(method: .post([:]), path: .mockPath)
 
-        var headerPresent = true
+        self.systemInfo.stubbedStorefront = nil
+
+        let header: Atomic<(value: String?, set: Bool)> = .init((nil, false))
 
         stub(condition: isPath(request.path)) { request in
-            let headers =  request.allHTTPHeaderFields ?? [:]
-            headerPresent = headers[headerName] != nil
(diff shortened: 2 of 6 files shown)


Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -285,6 +285,47 @@ final class HTTPClientTests: BaseHTTPClientTests<MockETagManager> {
...
+        }
+
+        expect(header.value) == "USA"
+    }
+
+    func testRequestsWithoutStorefrontDoNotSendHeader() {
+        let headerName = "X-Storefront"
+        self.systemInfo.stubbedStorefront = nil
+
+        var headerPresent = false
```
NachoSoto, line 312: This can be a false positive if the value is never changed. Maybe use the same approach as the other test?

joshdholtz: @NachoSoto I copied this way from the `testRequestsWithoutStorefrontDoNotSendHeader` test 😅 I tried the same approach as the other new test that I wrote but the header was still _there_ but just had a `nil` value. 

I was trying to find a way to test that the header wasn't sent at all 🤔 

Can I change it to `var headerPresent = true` so that this will default fail?

joshdholtz: Pushed a commit that I _think_ is better 🤷‍♂️  And also fixed one test that could have resulted in a false positive

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -915,7 +956,7 @@ final class HTTPClientTests: BaseHTTPClientTests<MockETagManager> {
 
         let request = HTTPRequest(method: .post([:]), path: .mockPath)
 
-        var headerPresent = false
+        var headerPresent = true
```
joshdholtz, line 959: This test is testing for `false` so changed the `headerPresent` to default to `true` so it will default fail

Tests/UnitTests/Networking/HTTPClientTests.swift
NachoSoto: >  I tried the same approach as the other new test that I wrote but the header was still there but just had a nil value.

Probably because of the `X-Storefrontt` typo 🙃 

This is fine too, but I think the other approach is more explicit? Starting with `headerPresent = true` can be confusing to future readers of this test.

Tests/UnitTests/Networking/HTTPClientTests.swift
NachoSoto: 👍🏻  You can leave this test as is since you didn't add it, and this does prevent a false positive

Sources/Misc/SystemInfo.swift
```diff
@@ -53,6 +53,14 @@ class SystemInfo {
         return self.sandboxEnvironmentDetector.isSandbox
     }
 
+    var storefront: StorefrontType? {
```
NachoSoto, line 56: Can you add a quick test for this in `SystemInfoTests`?

NachoSoto: Done

Tests/UnitTests/Networking/HTTPClientTests.swift
NachoSoto: Fixed

NachoSoto (approved):

PR #3418, opened 2023-11-15
Enable explicit_init lint rule and fix issues

MarkVillacampa:
This rule will warn when calling `init` with the explicit type where it can be inferred, or where the `.init` part can be omitted.

### DIFF
--- a/Sources/Purchasing/StoreKitAbstractions/SK1StoreProductDiscount.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/SK1StoreProductDiscount.swift
@@ -75,7 +75,7 @@ private extension StoreProductDiscount.PaymentMode {
         case .freeTrial:
             self = .freeTrial
         @unknown default:
-            Logger.appleWarning(Strings.storeKit.skunknown_payment_mode(String.init(describing: paymentMode)))
+            Logger.appleWarning(Strings.storeKit.skunknown_payment_mode(.init(describing: paymentMode)))
             return nil
         }
     }
--- a/Sources/Purchasing/StoreKitAbstractions/SK2StoreProductDiscount.swift
+++ b/Sources/Purchasing/StoreKitAbstractions/SK2StoreProductDiscount.swift
@@ -60,7 +60,7 @@ private extension StoreProductDiscount.PaymentMode {
         case .freeTrial:
             self = .freeTrial
         default:
-            Logger.appleWarning(Strings.storeKit.skunknown_payment_mode(String.init(describing: paymentMode)))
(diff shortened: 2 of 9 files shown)


Sources/Purchasing/StoreKitAbstractions/SK1StoreProductDiscount.swift
```diff
@@ -75,7 +75,7 @@ private extension StoreProductDiscount.PaymentMode {
         case .freeTrial:
             self = .freeTrial
         @unknown default:
-            Logger.appleWarning(Strings.storeKit.skunknown_payment_mode(String.init(describing: paymentMode)))
+            Logger.appleWarning(Strings.storeKit.skunknown_payment_mode(.init(describing: paymentMode)))
```
NachoSoto, line 78: Can we do
```suggestion
            Logger.appleWarning(Strings.storeKit.skunknown_payment_mode(paymentMode.rawValue))
```

That doesn't use reflection so it's a lot faster

Sources/Purchasing/StoreKitAbstractions/SK2StoreProductDiscount.swift
```diff
@@ -60,7 +60,7 @@ private extension StoreProductDiscount.PaymentMode {
         case .freeTrial:
             self = .freeTrial
         default:
-            Logger.appleWarning(Strings.storeKit.skunknown_payment_mode(String.init(describing: paymentMode)))
+            Logger.appleWarning(Strings.storeKit.skunknown_payment_mode(.init(describing: paymentMode)))
```
NachoSoto, line 63: Ditto

NachoSoto (approved):
Thanks for doing this! 👏🏻

Sources/Purchasing/StoreKitAbstractions/SK1StoreProductDiscount.swift
MarkVillacampa: Sadle it's a `UInt` enum :(

NachoSoto: Then `String(enum.rawValue)`. Describing uses runtime reflection and it's A LOT slower.

NachoSoto: You already merged so it's fine. It was already like this anyway.

NachoSoto: And it's not like this is a hot path.

PR #3424, opened 2023-11-16
`Trusted Entitlements`: add support for signing request headers

NachoSoto:
This adds support for including arbitrary headers in the signature verification, therefore preventing tampering of them.

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -148,6 +148,14 @@ extension HTTPClient {
         }
     }
 
+    static func headerParametersForSignatureHeader(with headers: RequestHeaders) -> RequestHeaders {
+        if let header = HTTPRequest.headerParametersForSignatureHeader(headers: headers) {
+            return [RequestHeader.headerParametersForSignature.rawValue: header]
+        } else {
+            return [:]
+        }
+    }
+
     enum RequestHeader: String {
 
         case authorization = "Authorization"
@@ -155,6 +163,7 @@ extension HTTPClient {
         case eTag = "X-RevenueCat-ETag"
         case eTagValidationTime = "X-RC-Last-Refresh-Time"
         case postParameters = "X-Post-Params-Hash"
+        case headerParametersForSignature = "X-Header-Params-Hash"
         case sandbox = "X-Is-Sandbox"
 
     }
@@ -313,6 +322,7 @@ private extension HTTPClient {
                 return cachedResponse.verify(
                     signing: self.signing(for: request.httpRequest),
                     request: request.httpRequest,
+                    requestHeaders: request.headers,
                     publicKey: request.verificationMode.publicKey
                 )
             }
@@ -532,10 +542,13 @@ extension HTTPRequest {
 
         if #available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *),
            verificationMode.isEnabled,
-           self.path.supportsSignatureVerification,
-           let body = self.requestBody {
+           self.path.supportsSignatureVerification {
+            result += HTTPClient.headerParametersForSignatureHeader(with: defaultHeaders)
+
+            if let body = self.requestBody {
                 result += HTTPClient.postParametersHeaderForSigning(with: body)
             }
+        }
 
         return result
     }
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -163,7 +163,7 @@ extension HTTPClient {
         case eTag = "X-RevenueCat-ETag"
         case eTagValidationTime = "X-RC-Last-Refresh-Time"
         case postParameters = "X-Post-Params-Hash"
-        case headerParametersForSignature = "X-Header-Params-Hash"
+        case headerParametersForSignature = "X-Headers-Hash"
         case sandbox = "X-Is-Sandbox"
 
     }
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -148,8 +148,14 @@ extension HTTPClient {
         }
     }
 
-    static func headerParametersForSignatureHeader(with headers: RequestHeaders) -> RequestHeaders {
-        if let header = HTTPRequest.headerParametersForSignatureHeader(headers: headers) {
+    static func headerParametersForSignatureHeader(
+        with headers: RequestHeaders,
+        path: HTTPRequestPath
+    ) -> RequestHeaders {
+        if let header = HTTPRequest.headerParametersForSignatureHeader(
+            headers: headers,
+            path: path
+        ) {
             return [RequestHeader.headerParametersForSignature.rawValue: header]
         } else {
             return [:]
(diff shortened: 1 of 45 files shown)


Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,14 @@
+{
```
NachoSoto, line 1: TODO: remove this

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -148,6 +148,14 @@ extension HTTPClient {
         }
     }
 
+    static func headerParametersForSignatureHeader(with headers: RequestHeaders) -> RequestHeaders {
+        if let header = HTTPRequest.headerParametersForSignatureHeader(headers: headers) {
+            return [RequestHeader.headerParametersForSignature.rawValue: header]
```
NachoSoto, line 153: This is tested by snapshot tests.

Tests/UnitTests/Mocks/MockHTTPClient.swift
```diff
@@ -83,6 +83,8 @@ class MockHTTPClient: HTTPClient {
             .requestAddingNonceIfRequired(with: verificationMode)
             .withHardcodedNonce
 
+isRecording = true
```
NachoSoto, line 86: TODO: remove

Sources/Security/HTTPRequest+Signing.swift
```diff
@@ -0,0 +1,91 @@
...
+    ) -> String {
+        return [
+            keys.joined(separator: ","),
+            postParameterHashingAlgorithmName,
+            hash
+        ].joined(separator: ":")
+    }
+
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    static func signingParameterHash(_ values: [String]) -> String {
```
NachoSoto, line 63: This is now a shared method for both POST body and header parameters.

Sources/Security/Signing.swift
```diff
@@ -276,6 +283,7 @@ extension Signing.SignatureParameters {
             nonce +
             path +
             postParameterHash +
+            headerParametersHash +
```
NachoSoto, line 286: Thanks to unit tests I was able to get this right. Initially I had this before the post hash.

Tests/UnitTests/Security/SigningTests.swift
```diff
@@ -502,10 +502,100 @@ class SigningTests: TestCase {
...
+        -H 'X-Nonce: MTIzNDU2Nzg5MGFi' \
+        -H 'X-Is-Sandbox: true' \
+        -H 'X-Headers-Hash: X-Is-Sandbox:sha256:b5bea41b6c623f7c09f1bf24dcae58ebab3c0cdd90ad966bc43a45b44867e12b' \
+        -H 'Authorization: Bearer appl_fFVBVAoYujMZJnepIziGKVjnZBz' \
+         */
+
+        let response = """
+        {"request_date":"2023-12-08T19:17:04Z","request_date_ms":1702063024731,"subscriber":{"entitlements":{},"first_seen":"2023-12-08T19:13:02Z","last_seen":"2023-12-08T19:13:02Z","management_url":null,"non_subscriptions":{},"original_app_user_id":"$RCAnonymousID:6ca4535c42714f88abc99c563703f113","original_application_version":null,"original_purchase_date":null,"other_purchases":{},"subscriptions":{}}}\n
+        """
+        let expectedSignature = "x2qnlHOl5WuzGi4TbSUVHxzlKELRCfrRYG9XAiso7ucZTQAAYEZqbguA3X0YfCJqCKh2hnTLSdEr4R+t23xBlTxceWZu2TJjK3461UJKpUnrwXDv+tYo2K54IoS3/tsEr3VmB5ppKAq0P2CR7SwbsDPpxUlHBcl5/4XJvb/DHOnTKjIVd4WJ+57LLWvIV9sDHnj9XxiBez+p5cEjez1RtUis0XdCfAFXU8XfAq6ggiEJKX4F"
```
NachoSoto, line 520: I've generated these from the backend, so we can confirm that these real signatures are valid.

Sources/Security/HTTPRequest+Signing.swift
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    /// - Returns: `nil` if none of the requested headers are found
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
+    private static func postParameterHash(_ headers: Headers) -> String? {
+        let headersToSign = Self.headersToSign.map(\.rawValue)
+        let values = headers
+            .filter { headersToSign.contains($0.key) }
+            .map(\.value)
```
bisho, line 40: Will this respect the order? we need values to be in the order of headers to sign, don't we?

NachoSoto: Good eye! This is backwards, it needs to be derived from `headersToSign` which is ordered (`headers` is a `Map` so it's not).
It didn't matter right now because we only sign one header, but it would have been non-deterministic as soon as we added a second one, so thanks! I just fixed it :)

Sources/Security/HTTPRequest+Signing.swift
```diff
@@ -0,0 +1,87 @@
...
+//  HTTPRequest+Signing.swift
+//
+//  Created by Nacho Soto on 11/16/23.
+
+import CryptoKit
+import Foundation
+
+extension HTTPRequest {
+
+    static func headerParametersForSignatureHeader(headers: Headers) -> String? {
```
tonidero, line 19: Hmm this name sounds weird to me... Maybe `headerParametersHeaderForSigning`? (To mimick `postParametersHeaderForSigning`

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -532,10 +542,13 @@ extension HTTPRequest {
 
         if #available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *),
            verificationMode.isEnabled,
-           self.path.supportsSignatureVerification,
-           let body = self.requestBody {
+           self.path.supportsSignatureVerification {
+            result += HTTPClient.headerParametersForSignatureHeader(with: defaultHeaders)
```
tonidero, line 546: FFTI I was wondering whether we should pass the `result` headers instead of only the `defaultHeaders` to future-proof it in case we need to sign any of the other headers... This should work for now though.

tonidero (approved):
Looks great!

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto: Yeah if we need to that will get caught by tests, so we can change this to support that.

Sources/Security/HTTPRequest+Signing.swift
NachoSoto: I was trying to avoid the double "header" word in `headerParametersHeaderForSigning`, which I think reads kinda confusingly :/

tonidero: Hmm yeah... Not sure... maybe it's ok as it is.

NachoSoto:
This should be ready, but the signature is still wrong on the load shedder

NachoSoto:
Merging this! 🎉

aboedo (commented):
I'm late to the game but great work on this!

PR #3425, opened 2023-11-16
`HTTPClient`: test all request headers

NachoSoto:
This allows us to verify exactly what headers are included in all requests.

It also simplifies the implementation for #3424.

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -80,6 +80,7 @@ class HTTPClient {
 
         self.perform(request: .init(httpRequest: request,
                                     authHeaders: self.authHeaders,
+                                    defaultHeaders: self.defaultHeaders,
                                     verificationMode: verificationMode ?? self.systemInfo.responseVerificationMode,
                                     completionHandler: completionHandler))
     }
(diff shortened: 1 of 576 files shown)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -92,6 +93,41 @@ class HTTPClient {
...
+            "X-Client-Version": SystemInfo.appVersion,
+            "X-Client-Build-Version": SystemInfo.buildVersion,
+            "X-Client-Bundle-ID": SystemInfo.bundleIdentifier,
+            "X-StoreKit2-Enabled": "\(self.systemInfo.storeKit2Setting.isEnabledAndAvailable)",
+            "X-Observer-Mode-Enabled": "\(self.systemInfo.observerMode)",
+            RequestHeader.sandbox.rawValue: "\(self.systemInfo.isSandbox)"
+        ]
+
+        if let storefront = self.systemInfo.storefront {
+            headers["X-Storefront"] = storefront.countryCode
```
tonidero, line 113: FFTI but in android we add all the headers to the same map and then filter out null values. That might be slightly more legible?

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -119,6 +155,8 @@ extension HTTPClient {
         case eTag = "X-RevenueCat-ETag"
         case eTagValidationTime = "X-RC-Last-Refresh-Time"
         case postParameters = "X-Post-Params-Hash"
+        case headerParametersForSignature = "X-Header-Params-Hash"
```
tonidero, line 158: Hmm this enum seems unused?

tonidero (approved):
❤️ We should add backend request snapshot tests working on Android soon

Sources/Networking/HTTPClient/HTTPClient.swift
NachoSoto: Oops thanks that's for #3424.

PR #3444, opened 2023-11-17
Some `APITester` fixes

NachoSoto:
Not sure why but sometimes Xcode gets in a state where it won't compile these without this.

### DIFF
--- a/Tests/APITesters/SwiftAPITester/SwiftAPITester/VerificationResultAPI.swift
+++ b/Tests/APITesters/SwiftAPITester/SwiftAPITester/VerificationResultAPI.swift
@@ -8,7 +8,7 @@
 import Foundation
 import RevenueCat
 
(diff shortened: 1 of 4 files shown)


Tests/APITesters/SwiftAPITester/SwiftAPITester/VerificationResultAPI.swift
```diff
@@ -8,7 +8,7 @@
 import Foundation
 import RevenueCat
 
-func checkVerificationResultAPI(_ mode: EntitlementVerificationMode = .disabled,
+func checkVerificationResultAPI(_ mode: Configuration.EntitlementVerificationMode = .disabled,
```
NachoSoto, line 11: This is technically the correct way. I don't understand why it compiles without it. CMD+clicking it finds the Obj-C symbol 🤔

tonidero (approved):

Issue #3465, opened 2023-11-21
Paywall not displaying right on an iPad & landscape iPhone Pro Max


QuentinRth:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
UIKit iOS/iPad OS app.
When presenting a PaywallViewController either on an iPad (both landscape and portrait) or on an iPhone Pro Max in landscape mode, it will enable some Split-view behaviour resulting in an invisible or partial paywall view.
It happens only when you put this parameter: displayCloseButton: true

```
let controller = PaywallViewController(offering: offering, displayCloseButton: true)
controller.delegate = self
self.present(controller, animated: true)
```

1. Environment
   1. Platform: iPhone 15 Pro Max & iPad Pro 11 inch
   2. SDK version: 4.30.5
   3. StoreKit version:
      - [x] StoreKit 1
      - [ ] StoreKit 2 (_enabled with `usesStoreKit2IfAvailable(true)`_)
   4. OS version: iOS 17.1.1 & iPad OS 17.0.1
   5. Xcode version: 15.0.1
   6. Device and/or simulator:
      - [x] Device
      - [x] Simulator
   7. Environment:
      - [x] Sandbox
      - [x] TestFight
      - [x] Production
   8. How widespread is the issue. Percentage of devices affected. 100% of Pro Max / iPad
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue. **Complete logs with `Purchases.logLevel = .verbose` will help us debug this issue.**

```
Nothing is happening in the logs, no constraints break or anything
```

3. Steps to reproduce, with a description of expected vs. actual behavior

I haven't tried on a SwiftUI project but with a UIKit project with different orientations enabled.
On a regular UIViewController, running on an iPad Pro or a iPhone 15 Pro Max in landscape mode, simply do this and it should reproduce the issue.
`self.present(PaywallViewController(displayCloseButton: true), animated: true)`
 
[...]

PR #3466, opened 2023-11-21
Fix navigation with close button in UIKit

aboedo:
Addresses #3465 

| Before | After |
| :-: | :-: |
|  IMG_0213 | IMG_0214 |

### DIFF
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -311,6 +311,8 @@ struct LoadedOfferingPaywallView: View {
                 view
                     .toolbar { Self.toolbar(dismiss) }
             }
+            .navigationViewStyle(.stack)
+            
         } else {
             view
         }


joshdholtz (approved):
:shipit:

RevenueCatUI/PaywallView.swift
```diff
@@ -311,6 +311,8 @@ struct LoadedOfferingPaywallView: View {
                 view
                     .toolbar { Self.toolbar(dismiss) }
             }
+            .navigationViewStyle(.stack)
```
NachoSoto, line 314: Makes a lot of sense to put this here 👍🏻  thanks!

NachoSoto:
Thanks for doing this! I'm surprised I haven't see it before.
Suggestions for the release notes:
- Change `fix` to `RevenueCatUI` so it's grouped there
- Add some reference to paywalls in the title so it's clear what this is about?

aboedo:
@NachoSoto thanks for the suggestions, changes made!

I feel like we should try to make it so that you can mix the tags (i.e.: fix + RevenueCat UI creates a section for paywalls bug fixes)

PR #3467, opened 2023-11-21
`Optimization`: changed `first` to `first(where:)`

NachoSoto:
See https://github.com/apple/swift/issues/55374

`.first` has been broken since 2020.

### DIFF
--- a/RevenueCatUI/Data/Localization.swift
+++ b/RevenueCatUI/Data/Localization.swift
@@ -31,7 +31,7 @@ enum Localization {
         return self.unitAbbreviationLengthPriorities
             .lazy
             .compactMap { length in options.first { $0.count == length } }
-            .first
+            .first { _ in true } // See https://github.com/apple/swift/issues/55374
         ?? options.last!
     }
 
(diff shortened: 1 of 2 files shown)


RevenueCatUI/Data/Localization.swift
```diff
@@ -31,7 +31,7 @@ enum Localization {
         return self.unitAbbreviationLengthPriorities
             .lazy
             .compactMap { length in options.first { $0.count == length } }
-            .first
+            .first { _ in true } // See https://github.com/apple/swift/issues/55374
```
tonidero, line 34: Should we create our own extension `firstElement` or something like that for this?

tonidero (approved):

aboedo (approved):

RevenueCatUI/Data/Localization.swift
NachoSoto: We'd need to duplicate it in both `RevenueCat` and `RevenueCatUI` (since we can't use the `package` access type yet).

PR #3471, opened 2023-11-21
[WIP] `Paywalls`: improve landscape support of all templates

NachoSoto:
#### Before
Simulator Screenshot - iPhone 15 Pro Max - 2023-11-22 at 07 29 06

#### After
simulator_screenshot_171E3E61-7182-4B4A-80C5-770F9B0803E6

#### Before
simulator_screenshot_528FF616-C89A-4F58-9983-732AE4D57A40

#### After
simulator_screenshot_8A95B6D9-2C44-41F4-ACAB-ADD5DE0663EE

#### Before
simulator_screenshot_71F4CF71-A49C-4252-AB17-87D39B94CE9F

#### After
simulator_screenshot_55A7A235-DF58-40FC-96E4-01B1F0070F39

#### Before
simulator_screenshot_30C0580D-D2D7-42C1-930E-D95D0C000115

#### After
simulator_screenshot_C75156C2-5237-4D22-920F-61A4D0452780


RevenueCatUI/Data/Constants.swift
```diff
@@ -74,4 +74,21 @@ extension TemplateViewType {
         return Constants.defaultVerticalPaddingLength(self.userInterfaceIdiom)
     }
 
+    var isVerticalSizeCompact: Bool {
+        #if os(tvOS)
```
NachoSoto, line 78: We don't support `tvOS` yet, but this is what we'll want to do anyway.

RevenueCatUI/Templates/Template3View.swift
```diff
@@ -24,60 +24,99 @@ struct Template3View: TemplateViewType {
...
+                self.horizontalFullScreenContent
+            } else {
+                self.verticalContent
+            }
+        }
+            .defaultHorizontalPadding()
+            .padding(.top, self.defaultVerticalPaddingLength)
+    }
+
+    // TODO: snapshots
```
NachoSoto, line 55: TODO: need snapshots for landscape paywalls.

RevenueCatUI/Templates/Template3View.swift
```diff
@@ -90,8 +129,39 @@ struct Template3View: TemplateViewType {
...
+            selectedPackage: self.configuration.packages.default,
+            configuration: self.configuration
+        )
+        .matchedGeometryEffect(id: Geometry.purchaseButton, in: self.namespace)
+    }
+
+    private var footer: some View {
+        FooterView(configuration: self.configuration,
+                   purchaseHandler: self.purchaseHandler)
+        .matchedGeometryEffect(id: Geometry.footer, in: self.namespace)
```
NachoSoto, line 160: TODO: this isn't quite working yet.

guido732:
first of all, looks great! 

I'm just wondering about that close button X and the alignment so it doesn't overlap with the content at any point. Is there a chance to adjust it or move it to the side outside of the content? 

image

aboedo:
ohhh that is really cool

NachoSoto:
> I'm just wondering about that close button X and the alignment so it doesn't overlap with the content at any point. Is there a chance to adjust it or move it to the side outside of the content?

Thanks for taking a close look at this!

I'll check in more detail, but that's actually the default alignment for both the close button (navigation bar buttons) and the main content.

RevenueCatUI/Templates/TemplateViewType.swift
```diff
@@ -41,6 +41,11 @@ protocol TemplateViewType: SwiftUI.View {
     var configuration: TemplateViewConfiguration { get }
     var userInterfaceIdiom: UserInterfaceIdiom { get }
 
+    /// `UserInterfaceSizeClass` is only available for macOS/watchOS/tvOS with Xcode 15
+    #if swift(>=5.9) || (!os(macOS) && !os(watchOS) && !os(tvOS))
```
NachoSoto, line 45: Ideally we mark `TemplateViewType` as unavailable in these platforms (since it is), but that leads to a SwiftUI warning:
> _modify accessor cannot be more available than unavailable enclosing scope

And having a warning breaks CocoaPods linting.

RevenueCatUI/Data/Constants.swift
```diff
@@ -74,4 +74,21 @@ extension TemplateViewType {
...
+        // tvOS never reports UserInterfaceSizeClass.compact
+        // but for the purposes of template layouts, we consider landscape
+        // on tvOS as compact to produce horizontal layouts.
```
aboedo, line 81: this is a fair point, but it makes me wonder whether the flag should be named something more representative, like `shouldUseLandscapeLayout` or something, given that it's not really "compact" for tvOS and that on macOS it's neither compact nor not compact, right?

RevenueCatUI/Templates/Template2View.swift
```diff
@@ -63,21 +78,49 @@ struct Template2View: TemplateViewType {
...
+
+                    Spacer(minLength: self.defaultVerticalPaddingLength)
+
+                    self.subscribeButton
+                }
+                .frame(maxHeight: .infinity)
+            }
+
+            self.footer
+        }
```
aboedo, line 123: this just looks so nice

RevenueCatUI/Templates/TemplateViewType.swift
aboedo: what's the story with visionOS for these? wondering if it'd be easier or safer to allow-list iOS rather than deny-list the rest

aboedo (approved):
looking great!

RevenueCatUI/Templates/TemplateViewType.swift
NachoSoto: It is available in visionOS.
I went with deny-list because presumedly we'll remove these eventually when adding support for them (other than watchOS which uses a separate template)

PR #3474, opened 2023-11-27
Paywalls: disable the close button when an action is in progress

aboedo:
Disabled and lowered the opacity of the close button while an action is in progress, to prevent a potentially bad UX state when you exit the paywall.

### DIFF
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -309,7 +309,7 @@ struct LoadedOfferingPaywallView: View {
         if let dismiss = self.dismiss {
             NavigationView {
                 view
-                    .toolbar { Self.toolbar(dismiss) }
+                    .toolbar { self.toolbar(dismiss) }
             }
             .navigationViewStyle(.stack)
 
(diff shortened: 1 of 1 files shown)


rkotzy:
I think the case when the purchase flow takes a long time that's exactly when you'd want to prevent someone from dismissing the view. It's like when you buy something on a website and the page says "Payment in process, don't go back or refresh your browser" 😅. 

Swapping out the "X" for a loading indicator could be slick enhancement or even making an entire loading state so it's more clear that something is happening. I've seen apps swap out the title of the primary CTA button with a loading indicator and that looks good. Maybe we should give it more thought to make something 🪄

rkotzy:
Nevermind - it looks like there already is a loading state with dimmed package buttons and a loading indicator over the CTA button.

tonidero:
Right, that's a good point... It just might be weird to have the user locked in the screen for a few seconds, but might be better than having the purchase flow appear suddenly later 👍

NachoSoto (approved):
I think this makes sense as a default behavior. Users can always disable our close button and add their own otherwise.

PR #3475, opened 2023-11-27
`Paywalls`: open Terms and Privacy Policy links in-app

aboedo:
Updated the terms and Privacy Policy links to open within the app, without exiting to Safari

https://github.com/RevenueCat/purchases-ios/assets/685609/e10df16d-1016-4271-9270-35c4ae8a87d4

### DIFF
--- a/RevenueCatUI/Views/FooterView.swift
+++ b/RevenueCatUI/Views/FooterView.swift
@@ -13,7 +13,10 @@
 
 import RevenueCat
 import SwiftUI
+
+#if canImport(WebKit)
 import WebKit
+#endif
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 struct FooterView: View {
@@ -225,6 +228,7 @@ private struct LinkButton: View {
     }
 
     var body: some View {
+        #if canImport(WebKit)
         Button {
             self.displayLink = true
         } label: {
@@ -247,6 +251,11 @@ private struct LinkButton: View {
                     }
             }
         }
+        #else
+        Link(destination: self.url) {
+            self.content
+        }
+        #endif
     }
 
     @ViewBuilder
@@ -284,8 +293,8 @@ private struct LinkButton: View {
 
 }
 
+#if canImport(WebKit)
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
-@available(watchOS, unavailable)
 @available(macOS, unavailable)
 @available(tvOS, unavailable)
 private struct WebView: UIViewRepresentable {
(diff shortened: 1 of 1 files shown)


tonidero:
> what if we always open the links in a fullscreen modal?

I agree that makes more sense. Just not sure if you can present a modal on top of a modal, in case the developer is presenting the paywall as a modal already?

aboedo:
Yeah, the modal thing might feel weird if you're already in a modal. One alternative is to make this remotely controllable from a variable

NachoSoto:
How does this look on iPad, and visionOS?

NachoSoto (commented):
I think this makes sense and it's a better UX 👍🏻

RevenueCatUI/Views/FooterView.swift
```diff
@@ -257,6 +257,23 @@ private struct LinkButton: View {
 
 }
 
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@available(watchOS, unavailable)
```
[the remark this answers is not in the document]
NachoSoto: And maybe `visionOS`?

NachoSoto:
Just watched the video. Perhaps we should present this as a sheet. Right now it shows a navigation bar which doesn't make a lot of sense since the controller doesn't have a title.

RevenueCatUI/Views/FooterView.swift
```diff
@@ -13,6 +13,7 @@
 
 import RevenueCat
 import SwiftUI
+import WebKit
```
NachoSoto, line 16: This doesn't exist on watchOS 😅

RevenueCatUI/Views/FooterView.swift
```diff
@@ -221,12 +228,44 @@ private struct LinkButton: View {
...
+                                self.displayLink = false
+                            } label: {
+                                Image(systemName: "xmark")
+                            }
+                        }
+                    }
+            }
+        }
+        #else
+        Link(destination: self.url) {
```
NachoSoto, line 255: Keeping the existing implementation if `WebKit` isn't supported.

NachoSoto:
Updated to use a sheet. Presenting a sheet on top of a sheet is a common pattern supported by iOS (see video)

NachoSoto (approved):

stuart-j-williams:
This is broken on iPad as it opens the target in the sidebar of a split screen.  Do I need to log a new issue for this?  Thanks
IMG_0003 2

NachoSoto:
Oh thanks! We'll get that fixed 🙏🏻

NachoSoto:
@stuart-j-williams I'm unable to reproduce this, could you please file an issue including:
- iOS version
- Xcode version
- Code with how you're presenting the paywall.

Thanks!

PR #3476, opened 2023-11-28
`Paywalls`: adjusted German translations

tensei:
Special abbreviation style for German and some corrections to the German translation

### Motivation

Just fixing/adjusting some stuff for all germans


jefago (approved):
Sounds good. Changing "Geschäftsbedingungen" to "Nutzungsbedingungen" isn't super unambiguous, but I found more examples of apps using "Nutzungsbedingungen" so it's probably a good solution.

RevenueCatUI/Data/Localization.swift
```diff
@@ -164,6 +164,8 @@ private extension Localization {
         switch locale.languageCodeIdentifier {
         // Abbreviated does not fully work with Japanese
         case "ja": return [.brief]
+        // Abbreviated is too short for German
+        case "de": return [.full, .brief]
```
vegaro, line 168: Can you provide more context for this change? Thanks!

tensei: This change is for the variable {{ price_per_period }} and {{ total_price_and_per_month }} in Paywalls

Weekly Before `9,99 €/Wo.` Now `no change`
Monthly Before `9,99 €/M` Now `9,99 €/Mon.`
Yearly Before `9,99 €/J` Now `no change`

NachoSoto:
This is great! Let's add tests like we have for Japanese :)

tensei:
Tests added 👍

NachoSoto (approved):

PR #3477, opened 2023-11-28
Paywalls: explicitly cache images for paywalls 

aboedo:
This adds explicit caching to images for Paywalls. 

In my testing we seem to be mostly re-fetching images on every load. We're studying also improving the cache headers and that might (hopefully) render this PR unnecessary, but opening this up in the meantime just in case. 

This PR fixes the re-fetching as far as I can tell.

### DIFF
--- a/RevenueCatUI/Helpers/ImageLoader.swift
+++ b/RevenueCatUI/Helpers/ImageLoader.swift
@@ -24,9 +24,12 @@ class ImageLoader: ObservableObject {
     @Published var error: Error?
     private var cancellable: AnyCancellable?
     private let cache: URLCache
+    private let urlSessionConfiguration: URLSessionConfiguration
 
-    init(url: URL, cache: URLCache = .shared) {
+    init(url: URL, cache: URLCache) {
         self.cache = cache
+        self.urlSessionConfiguration = URLSessionConfiguration.default
+        urlSessionConfiguration.urlCache = cache
         load(url: url)
     }
 
--- a/RevenueCatUI/Helpers/ImageLoader.swift
+++ b/RevenueCatUI/Helpers/ImageLoader.swift
@@ -1,0 +1,59 @@
+//
+//  ImageLoader.swift
+//
+//
+//  Created by Andrés Boedo on 11/29/23.
+//
+
+import Combine
+import Foundation
+import UIKit
+
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+extension URLCache {
+    static let imageCache = URLCache(memoryCapacity: 5 * 1000 * 1000, diskCapacity: 30 * 1000 * 1000)
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@MainActor
+class ImageLoader: ObservableObject {
+    @Published var image: UIImage?
+    @Published var error: Error?
+    private var cancellable: AnyCancellable?
+    private let cache: URLCache
+
+    init(url: URL, cache: URLCache = .shared) {
+        self.cache = cache
+        load(url: url)
+    }
+
+    deinit {
+        cancellable?.cancel()
+    }
+
+    func load(url: URL) {
+        let request = URLRequest(url: url, cachePolicy: .returnCacheDataElseLoad)
+        cancellable = URLSession.shared.dataTaskPublisher(for: request)
+            .tryMap { output in
+                guard let httpResponse = output.response as? HTTPURLResponse, httpResponse.statusCode == 200 else {
+                    throw URLError(.badServerResponse)
+                }
+                return UIImage(data: output.data)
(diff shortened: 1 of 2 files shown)


aboedo:
this is ready to go from a code perspective... but I don't love the idea of not having any tests in it?

Looks like our setup is mostly snapshot testing, which makes sense, but for stuff that is Paywalls-side logic it'd be great to have a test target. I'll make that as a separate PR if I get the time

aboedo:
image
uhh wat

aboedo:
nevermind we do have tests and I even modified test cases earlier today, what am I saying

aboedo:
going back to draft to write tests for it

RevenueCatUI/Views/RemoteImage.swift
```diff
@@ -8,57 +8,64 @@
...
+//
 //  Created by Nacho Soto on 7/19/23.
 
 import SwiftUI
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 struct RemoteImage: View {
 
+    @StateObject private var loader: ImageLoader
+    private let cache: URLCache
```
tonidero, line 20: Small thing but do we need to store this here? Seems like this is unused.

aboedo:
Update here: we're actually supposed to be caching this stuff here https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Paywalls/PaywallCacheWarming.swift, so we should dig into why that's not happening correctly before merging this, and unify the caching for both

RevenueCatUI/Views/RemoteImage.swift
aboedo: good catch

aboedo:
PaywallCacheWarming uses a different instance of URLCache, so we'll have to unify them. Also, we're using the shared cache, which isn't great because we're also setting the actual values for the cache in the whole app. A bit of a bigger refactor, though, since the warming happens in RC and the image loading currently happens in UI

RevenueCatUI/Helpers/ImageLoader.swift
```diff
@@ -0,0 +1,64 @@
...
+//
+
+import Combine
+import Foundation
+import UIKit
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+extension URLCache {
+
+    static let imageCache = URLCache(memoryCapacity: 5 * 1000 * 1000, diskCapacity: 30 * 1000 * 1000)
```
NachoSoto, line 15: We'll need to update `PaywallCacheWarming` to use this too, so we'll probably have to expose it from `Purchases`.

RevenueCatUI/Helpers/ImageLoader.swift
```diff
@@ -0,0 +1,64 @@
...
+
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@MainActor
+class ImageLoader: ObservableObject {
+
+    @Published var image: UIImage?
+    @Published var error: Error?
+    private var cancellable: AnyCancellable?
```
NachoSoto, line 25: We're only storing one of these, but you can call `load` multiple times, so only the last one is cancelled?

RevenueCatUI/Helpers/ImageLoader.swift
```diff
@@ -0,0 +1,64 @@
...
+    static let imageCache = URLCache(memoryCapacity: 5 * 1000 * 1000, diskCapacity: 30 * 1000 * 1000)
+
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@MainActor
+class ImageLoader: ObservableObject {
+
+    @Published var image: UIImage?
+    @Published var error: Error?
```
NachoSoto, line 24: Should this be a `Result<UIImage, Error>` instead?

NachoSoto (commented):
It's unfortunate that it means moving away from `AsyncImage`, but I guess it's time to finally implement this manually and make it work consistently.

RevenueCatUI/Helpers/ImageLoader.swift
aboedo: agreed! I commented the same thing

RevenueCatUI/Helpers/ImageLoader.swift
aboedo: that'd make a lot more sense

RevenueCatUI/Helpers/ImageLoader.swift
aboedo: yep!

NachoSoto: I'll update this so that overriding a cancellable cancels the previous one.

NachoSoto:
Replaced by #3498

PR #3478, opened 2023-11-29
[SK2] Send original application version & purchase date as HTTP header

MarkVillacampa:
Previously, this information was contained in the SK1 receipt. Now, it needs to be fetched as an AppTransaction from StoreKit 2.


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1721,6 +1727,22 @@ private extension Purchases {
...
+           self.systemInfo.originalAppVersion == nil,
+           self.systemInfo.originalAppPurchaseDate == nil {
+            if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
+                _ = Task<Void, Never> {
+                    let appTransaction = await self.transactionFetcher.appTransaction
+                    systemInfo.originalAppVersion = appTransaction?.originalApplicationVersion
+                    systemInfo.originalAppPurchaseDate = appTransaction?.originalPurchaseDate
+                }
+            }
+        }
```
MarkVillacampa, line 1745: AppTransaction will be fetched on first launch (if not backgrounded) and on each subsequent foregrounding, if it's not already cached.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1516,8 +1520,6 @@ internal extension Purchases {
         return self.systemInfo.storeKit2Setting
     }
 
-    #if DEBUG
-
```
MarkVillacampa, line 1520: `fetchReceipt` is already in an `internal` extension.

NachoSoto: This was only used for tests, and I think we didn't want to use it in production until we verified if it was too slow.
Let's at least make sure we don't call this in the main thread with `RCTestAssertNotMainThread`

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1721,6 +1723,27 @@ private extension Purchases {
...
+        self.fetchAndCacheAppTransactionIfNeeded()
+    }
+
+    func fetchAndCacheAppTransactionIfNeeded() {
+        if self.systemInfo.dangerousSettings.internalSettings.usesStoreKit2JWS,
+           self.systemInfo.originalAppVersion == nil,
+           self.systemInfo.originalAppPurchaseDate == nil {
+            if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
+                _ = Task<Void, Never> {
+                    if let appTransaction = await self.transactionFetcher.appTransaction {
```
NachoSoto, line 1736: Let's add `RCTestAssertNotMainThread` here

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1721,6 +1723,27 @@ private extension Purchases {
...
+                        systemInfo.originalAppVersion = appTransaction.originalApplicationVersion
```
NachoSoto, line 1737: Explicit `self` plz.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1721,6 +1727,27 @@ private extension Purchases {
...
+                        systemInfo.originalAppPurchaseDate = appTransaction.originalPurchaseDate
+                    } else {
+                        let receipt = try? await self.fetchReceipt(.onlyIfEmpty)
+                        systemInfo.originalAppVersion = receipt?.originalApplicationVersion
+                        systemInfo.originalAppPurchaseDate = nil
```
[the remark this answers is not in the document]
NachoSoto: And it's tested 🙌🏻

Sources/Purchasing/StoreKit2/SK2AppTransaction.swift
```diff
@@ -25,6 +24,18 @@ internal struct SK2AppTransaction {
         self.environment = .init(environment: appTransaction.environment)
     }
 
+    init(
```
NachoSoto, line 27: This is only used for tests, right?

In that case, I'd move it to an extension in tests only so we don't ship this, since it's only for creating mock transactions.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1721,6 +1723,27 @@ private extension Purchases {
...
+        if self.systemInfo.dangerousSettings.internalSettings.usesStoreKit2JWS,
+           self.systemInfo.originalAppVersion == nil,
+           self.systemInfo.originalAppPurchaseDate == nil {
+            if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) {
+                _ = Task<Void, Never> {
+                    if let appTransaction = await self.transactionFetcher.appTransaction {
+                        systemInfo.originalAppVersion = appTransaction.originalApplicationVersion
+                        systemInfo.originalAppPurchaseDate = appTransaction.originalPurchaseDate
+                    } else {
+                        let receipt = try? await self.fetchReceipt(.onlyIfEmpty)
```
NachoSoto, line 1740: I love that this fallback is tested too 👍🏻

Tests/UnitTests/Purchasing/Purchases/PurchasesConfiguringTests.swift
```diff
@@ -391,6 +391,73 @@ class PurchasesConfiguringTests: BasePurchasesTests {
         expect(self.mockPaymentQueueWrapper.delegate) === self.purchasesOrchestrator
     }
 
+    func testCachesAppTransactionIfInStoreKit2JWSMode() throws {
+        try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
```
NachoSoto, line 395: Can you add tests pre-iOS 15 too with a mock receipt?

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1516,8 +1520,6 @@ internal extension Purchases {
         return self.systemInfo.storeKit2Setting
     }
 
-    #if DEBUG
-
     /// - Returns: the parsed `AppleReceipt`
     ///
     /// - Warning: this is only meant for integration tests, as a way to debug purchase failures.
```
NachoSoto, line 1525: This comment is still here, but I guess we need to remove it now?

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto, line 1740: ⚠️ (cc @aboedo so he can confirm), but if we use `.onlyIfEmpty`, users would see a login prompt if the receipt is empty. We need to use `.never`

NachoSoto (changes requested):
Looks good in general, just 2 issues:
- Make sure we don't fetch and parse receipts on the main thread
- Don't refresh receipts to avoid login prompts.

PR #3480, opened 2023-11-29
Improve price per month accuracy for weekly subscriptions

aboedo:
We were calculating the price per month for weekly products as price * 4, but there are on average 4.345 weeks in a month, so this updates the value accordingly.


tonidero (approved):

NachoSoto (approved):
Makes sense!

PR #3487, opened 2023-12-04
[SK2] Add `Configuration.with(storeKitVersion:)` to select the version of StoreKit to use

MarkVillacampa:
This PR introduces a new configuration option to select the StoreKit version to use.

It supersedes the deprecated `.with(usesStoreKit2IfAvailable: true)` and the internal `usesStoreKit2JWS` configuration options. 

Uses JWS tokens instead of SK1 receipts when in StoreKit 2 mode where applicable.

Example usage:
```swift
Purchases.configure(
                .with(storeKitVersion: .storeKit2)
                .build()
```


Sources/Misc/StoreKitVersion.swift
```diff
@@ -0,0 +1,61 @@
...
+//
+//  StoreKitVersion.swift
+//
+//  Created by Mark Villacampa on 4/13/23.
+
+import Foundation
+
+/// Defines which version of StoreKit may be used
+@objc(RCStoreKitVersion)
+public enum StoreKitVersion: Int {
```
NachoSoto, line 18: Have you considered removing `StoreKit2Setting`? The confusing part of this PR is that we now have both but since we're not using `.enabledOnlyForOptimizations`, we don't actually need that anymore?

Sources/Misc/StoreKitVersion.swift
```diff
@@ -0,0 +1,61 @@
...
+        switch self {
+        case .storeKit1, .default: return false
+        case .storeKit2: return Self.isStoreKit2Available
+        }
+    }
+
+}
+
+extension StoreKitVersion {
+    var versionString: String {
```
NachoSoto, line 54: Nit: spacing here

Sources/Misc/StoreKitVersion.swift
```diff
@@ -0,0 +1,61 @@
...
+        if isStoreKit2EnabledAndAvailable {
```
NachoSoto, line 55: Explicit `self`:
```suggestion
        if self.isStoreKit2EnabledAndAvailable {
```

Tests/UnitTests/Misc/StoreKitVersionTests.swift
```diff
@@ -0,0 +1,65 @@
...
+//  StoreKitVersionTests.swift
+//
+//  Created by Mark Villacampa on 4/12/23.
+
+import Nimble
+import XCTest
+
+@testable import RevenueCat
+
+class StoreKitVersionTests: TestCase {
```
NachoSoto, line 19: <3

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendGetCustomerInfoTests/iOS14-testHandlesInvalidJSON.1.json
```diff
@@ -12,6 +12,7 @@
     "X-Platform-Flavor" : "native",
     "X-Platform-Version" : "Version 17.0.0 (Build 21A342)",
     "X-Storefront" : "USA",
+    "X-StoreKit-Version" : "1",
```
NachoSoto, line 15: This is a better header than `X-StoreKit2-Enabled` 👍🏻

Sources/Purchasing/Configuration.swift
```diff
@@ -218,6 +221,27 @@ import Foundation
...
+        ///
+        /// Defaults to ``StoreKitVersion/default`` which lets the SDK select
+        /// the most appropriate version of StoreKit. Currently defaults to StoreKit 1.
+        ///
+        /// - Note: StoreKit 2 is only available on iOS 15+. StoreKit 1 will be used for previous iOS versions
+        /// regardless of this setting.
+        ///
+        /// ### Related Symbols
+        /// - ``StoreKitVersion``
+        @objc public func with(storeKitVersion version: StoreKitVersion) -> Builder {
```
NachoSoto, line 234: I started doing this for #3066 but ended up removing it. I think it's a better API 👍🏻

Tests/APITesters/SwiftAPITester/SwiftAPITester/ConfigurationAPI.swift
```diff
@@ -21,6 +21,7 @@ func checkConfigurationAPI() {
         .with(networkTimeout: 1)
         .with(storeKit1Timeout: 1)
         .with(platformInfo: Purchases.PlatformInfo(flavor: "", version: ""))
+        .with(storeKitVersion: .default)
```
NachoSoto, line 24: We're missing a few API testers:
- This in Obj-C
- `StoreKitVersion`

NachoSoto:
Looks good in general, but I want to discuss the duplicate flag.

NachoSoto:
Let's print the configured SK version on SDK initialization too: https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/Purchases/Purchases.swift#L556

Sources/Misc/StoreKitVersion.swift
NachoSoto: Bump

Sources/Logging/Strings/ConfigureStrings.swift
```diff
@@ -37,6 +37,8 @@ enum ConfigureStrings {
 
     case response_verification_mode(Signing.ResponseVerificationMode)
 
+    case storekit_version(String)
```
NachoSoto, line 40: This `String` could be anything. Better to use `StoreKitVersion` to make it type safe.

Sources/Misc/StoreKitVersion.swift
```diff
@@ -0,0 +1,63 @@
...
+    /// Always use StoreKit 1. StoreKit 2 will be used (if available in the current device) only for certain APIs
+    /// that provide a better implementation. For example: intro eligibility, determining if a receipt has
+    /// purchases, managing subscriptions.
+    case storeKit1
+
+    /// Always use StoreKit 2.
+    case storeKit2
+
+    /// Let RevenueCat use the most appropiate version of StoreKit
+    case `default`
```
NachoSoto, line 29: Instead of making this a `case`, I think it would make more sense to define it like this:
```
extension StoreKitVersion {

  public static `default`: Self = .storeKit1

}
```

Since it's technically not a different version.

Then we don't need to consider it separately in all the logic, and to update it we only need to change it in one place.

Sources/Misc/StoreKitVersion.swift
```diff
@@ -0,0 +1,63 @@
...
+        case .storeKit1, .default: return false
+        case .storeKit2: return Self.isStoreKit2Available
+        }
+    }
+
+}
+
+extension StoreKitVersion {
+
+    var versionString: String {
```
NachoSoto, line 55: Nit: putting the type of the value in the name isn't a thing anymore 😅

Maybe we can do this:
```
extension StoreKitVersion: CustomDebugStringConvertible {

  public var debugDescription: {
    switch self {} // Let's use a switch, so if we ever add more cases, the compiler reminds us to handle it here.
  }

}
```

Sources/Misc/SystemInfo.swift
```diff
@@ -33,6 +33,7 @@ class SystemInfo {
     }
 
     let storeKit2Setting: StoreKit2Setting
+    let storeKitVersion: StoreKitVersion
```
NachoSoto, line 36: Per my other comment, this is the main thing I'm concerned about: we now have 2 sources of truth.

I'd remove `StoreKit2Setting`, and instead derive the existing logic from the new `StoreKitVersion`.

NachoSoto (changes requested):
Let's agree one what to do with `StoreKit2Setting`

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -449,6 +447,51 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
         expect(self.receiptFetcher.receiptDataReceivedRefreshPolicy) == .onlyIfEmpty
     }
 
+    func testPurchaseSK1PackageRetriesReceiptFetchIfEnabled() async throws {
```
MarkVillacampa, line 450: Functionally equivalent to the removed test below, but uses SK1, since SK2 does not post receipts now.

Sources/Misc/StoreKitVersion.swift
MarkVillacampa: Renamed it to `effectiveVersion` instead of using the `debugDescription` mechanism since it does not represent the enum value, but the effective version of StoreKit that's use despite what's configured. Added a docstring comment too.

Sources/Misc/StoreKitVersion.swift
```diff
@@ -0,0 +1,70 @@
...
+    case storeKit1
+
+    /// Always use StoreKit 2.
+    case storeKit2
+}
+
+public extension StoreKitVersion {
+
+    /// Let RevenueCat use the most appropiate version of StoreKit
+    static var `default` = Self.storeKit1
```
NachoSoto, line 32: Oops! This is mutable , meaning you can do
```
StoreKitVersion.default = .storeKit2 😬
```

```suggestion
    static let `default`: Self = .storeKit1
```

Sources/Misc/StoreKitVersion.swift
```diff
@@ -0,0 +1,70 @@
...
+    var isStoreKit2EnabledAndAvailable: Bool {
+        switch self {
+        case .storeKit1: return false
+        case .storeKit2: return Self.isStoreKit2Available
+        }
+    }
+
+    /// Returns a string representing the effective version of StoreKit used.
+    /// This can be different from the configured version if StoreKit 2 is not available on the current device.
+    var effectiveVersion: String {
```
NachoSoto, line 57: It seems like this is doing 2 things:
- Detecting the effective `StoreKitVersion`
- Converting to `String`

I'd change this to
```suggestion
    var effectiveVersion: StoreKitVersion {
```

Then maybe using my `CustomDebugStringConvertible` suggestion?

Sources/Misc/Deprecations.swift
```diff
@@ -403,24 +403,9 @@ extension CustomerInfo {
...
-    @available(*, deprecated, message: """
-    RevenueCat currently uses StoreKit 1 for purchases, as its stability in production scenarios has
-    proven to be more performant than StoreKit 2.
-
-    We're collecting more data on the best approach, but StoreKit 1 vs StoreKit 2 is an implementation detail
-    that you shouldn't need to care about.
-
-    Simply remove this method call to let RevenueCat decide for you which StoreKit implementation to use.
-    """)
+    @available(*, deprecated, message: "Use .with(storeKitVersion:) to enable StoreKit 2")
```
NachoSoto, line 406: <3

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -548,7 +548,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
         }
 
         Logger.debug(Strings.configure.debug_enabled, fileName: nil)
-        if systemInfo.storeKit2Setting == .enabledForCompatibleDevices {
+        if systemInfo.storeKitVersion.isStoreKit2EnabledAndAvailable {
```
NachoSoto, line 551: Maybe we can remove this now that we have the generic `storekit_version` below?

Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,50 @@
+{
```
NachoSoto, line 1: Oops, can you remove this file?

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -788,7 +788,7 @@ final class HTTPClientTests: BaseHTTPClientTests<MockETagManager> {
 
         let headerPresent: Atomic<Bool> = false
 
-        let enabled = self.systemInfo.storeKit2Setting.isEnabledAndAvailable.description
+        let enabled = self.systemInfo.storeKitVersion.isStoreKit2EnabledAndAvailable.description
```
NachoSoto, line 791: Maybe add a test for the new header too?

Tests/UnitTests/Purchasing/ConfigurationTests.swift
```diff
@@ -35,19 +35,18 @@ class ConfigurationTests: TestCase {
         let configuration = Configuration.Builder(withAPIKey: "test").build()
 
         expect(configuration.observerMode) == false
-        expect(configuration.storeKit2Setting) == .enabledOnlyForOptimizations
+        expect(configuration.storeKitVersion) == .storeKit1
```
NachoSoto, line 38: Let's leave 2 extra tests (marked as deprecated) that test the `usesStoreKit2IfAvailable` configuration. Since it's still available, we need to make sure that it continues to work.

Tests/UnitTests/Purchasing/ConfigurationTests.swift
```diff
@@ -67,14 +66,39 @@ class ConfigurationTests: TestCase {
     func testObserverModeWithStoreKit2() {
         let configuration = Configuration.Builder(withAPIKey: "test")
             .with(observerMode: true)
-            .with(usesStoreKit2IfAvailable: true)
+            .with(storeKitVersion: .storeKit2)
```
NachoSoto, line 69: This test is still marked as deprecated

Tests/UnitTests/Purchasing/ConfigurationTests.swift
NachoSoto, line 38: Let's leave 2 extra tests (marked as deprecated) that test the `usesStoreKit2IfAvailable` configuration. Since it's still available, we need to make sure that it continues to work.

Edit: never mind I see you did that :D

Tests/UnitTests/Purchasing/ConfigurationTests.swift
```diff
@@ -67,14 +66,39 @@ class ConfigurationTests: TestCase {
...
+        expect(configuration.storeKitVersion) == .storeKit2
+    }
+
+    @available(*, deprecated)
+    func testLegacyFlagSetsStoreKitVersionWhenStoreKit1Enabled() {
+        let configuration = Configuration.Builder(withAPIKey: "test")
+            .with(usesStoreKit2IfAvailable: false)
+            .build()
+
+        expect(configuration.storeKitVersion) == StoreKitVersion.default
```
NachoSoto, line 101: Nit: 
```suggestion
        expect(configuration.storeKitVersion) == .default
```

Tests/UnitTests/Purchasing/Purchases/BasePurchasesTests.swift
```diff
@@ -313,10 +313,8 @@ class BasePurchasesTests: TestCase {
         self.storeKit1Wrapper.delegate?.storeKit1Wrapper(self.storeKit1Wrapper, updatedTransaction: transaction)
     }
 
-    var storeKit2Setting: StoreKit2Setting {
-        // Even though the new default is StoreKit 2, most of the tests from this parent class
-        // were written for SK1. Therefore we want to default to it being disabled.
-        return .enabledOnlyForOptimizations
+    var storeKitVersion: StoreKitVersion {
+        return .storeKit1
```
NachoSoto, line 317: Let's leave the comment?

NachoSoto (commented):
I love this PR 🤩

Just a few small details.

Tests/APITesters/SwiftAPITester/SwiftAPITester/ConfigurationAPI.swift
NachoSoto: Bump

Sources/Misc/StoreKitVersion.swift
```diff
@@ -17,12 +17,13 @@ import Foundation
 @objc(RCStoreKitVersion)
 public enum StoreKitVersion: Int {
 
-    /// Always use StoreKit 1. StoreKit 2 will be used (if available in the current device) only for certain APIs
-    /// that provide a better implementation. For example: intro eligibility, determining if a receipt has
-    /// purchases, managing subscriptions.
+    /// Always use StoreKit 1.
```
MarkVillacampa, line 20: The comment about StoreKit 2 was no longer true. We dont use StoreKit 2 at all if `.storeKit1` is selected.

Sources/Misc/StoreKitVersion.swift
```diff
@@ -17,12 +17,13 @@ import Foundation
...
-    /// that provide a better implementation. For example: intro eligibility, determining if a receipt has
-    /// purchases, managing subscriptions.
+    /// Always use StoreKit 1. 
     case storeKit1
 
-    /// Always use StoreKit 2.
+    /// Always use StoreKit 2 (StoreKit 1 will be used if StoreKit 2 is not available in the current device.)
+    ///
+    /// Make sure you have an In-App Purchase Key configured in your app.
+    /// Please see https://rev.cat/in-app-purchase-key-configuration for more info.
```
MarkVillacampa, line 26: Added the link to configure the In-App Purchase Key as it is required when using `.storeKit2`

Sources/Misc/StoreKitVersion.swift
NachoSoto: Oh that's right

Sources/Misc/Deprecations.swift
```diff
@@ -118,7 +118,7 @@ public extension Purchases {
             userDefaults: userDefaults,
             platformInfo: nil,
             responseVerificationMode: .default,
-            storeKit2Setting: .init(useStoreKit2IfAvailable: useStoreKit2IfAvailable),
+            storeKitVersion: useStoreKit2IfAvailable ? .storeKit2 : .default,
```
NachoSoto, line 121: I know this is deprecated, but should this be
```suggestion
            storeKitVersion: useStoreKit2IfAvailable ? .storeKit2 : .storeKit1,
```

If you pass `false`, it might still use it SK2 if/when that becomes the default.

Sources/Misc/StoreKitVersion.swift
```diff
@@ -0,0 +1,82 @@
...
+public enum StoreKitVersion: Int {
+
+    /// Always use StoreKit 1. 
+    case storeKit1
+
+    /// Always use StoreKit 2 (StoreKit 1 will be used if StoreKit 2 is not available in the current device.)
+    ///
+    /// Make sure you have an In-App Purchase Key configured in your app.
+    /// Please see https://rev.cat/in-app-purchase-key-configuration for more info.
+    case storeKit2
```
NachoSoto, line 27: Nit spacing below
```suggestion
    case storeKit2
    
```

Sources/Misc/StoreKitVersion.swift
```diff
@@ -0,0 +1,82 @@
...
+//
+//  StoreKitVersion.swift
+//
+//  Created by Mark Villacampa on 4/13/23.
+
+import Foundation
+
+/// Defines which version of StoreKit may be used
+@objc(RCStoreKitVersion)
+public enum StoreKitVersion: Int {
```
NachoSoto, line 18: Do you need this to have a `rawValue`? I'd remove it unless we're using it. And if we are, I'd explicitly set the values (right now `storeKit1` is `0`, and `storeKit2` is `1`)
```suggestion
public enum StoreKitVersion {
```

Sources/Misc/StoreKitVersion.swift
```diff
@@ -0,0 +1,82 @@
...
+    }
+
+    /// Returns the effective version of StoreKit used.
+    /// This can be different from the configured version if StoreKit 2 is not available on the current device.
+    var effectiveVersion: StoreKitVersion {
+        switch self {
+        case .storeKit1:
+            return .storeKit1
+        case .storeKit2:
+            if self.isStoreKit2EnabledAndAvailable {
```
NachoSoto, line 74: I think it makes more sense to check `isStoreKit2Available` instead.

Sources/Purchasing/Configuration.swift
```diff
@@ -311,8 +324,8 @@ extension Configuration {
         }
     }
 
-    fileprivate static func verify(observerMode: Bool, storeKit2Setting: StoreKit2Setting) {
-        if observerMode, storeKit2Setting.usesStoreKit2IfAvailable {
+    fileprivate static func verify(observerMode: Bool, storeKitVersion: StoreKitVersion) {
+        if observerMode, storeKitVersion == .storeKit2 {
             Logger.warn(Strings.configure.observer_mode_with_storekit2)
```
NachoSoto, line 329: Let's make sure we remove this when it is compatible :)

Tests/APITesters/ObjCAPITester/ObjCAPITester/RCConfigurationAPI.m
```diff
@@ -23,6 +23,7 @@ + (void)checkAPI {
                                            withStoreKit1Timeout: 1]
                                           withPlatformInfo:[[RCPlatformInfo alloc] initWithFlavor:@"" version:@""]]
                                          withUsesStoreKit2IfAvailable:false]
+                                        withStoreKitVersion:RCStoreKitVersionStoreKit2]
```
NachoSoto, line 26: Let's also add tests (Swift + Obj-C) for the enum with the 2 cases + default.

NachoSoto (approved):

Sources/Misc/StoreKitVersion.swift
MarkVillacampa: `@objc' enum must declare an integer raw type`

PR #3489, opened 2023-12-04
Paywalls: Improve Chinese localization

aboedo:
Improvements to Chinese Traditional and Chinese Simplified localizations based on user feedback.

PR #3490, opened 2023-12-04
Fix Chinese localizations

francisfeng:
### Motivation
Resolves #3489

### Description
Another issue I found after looking at the [Localizable.strings](https://github.com/RevenueCat/purchases-ios/pull/3489/files): You can’t directly translate "%d%% off” as "%d%%折” in Chinese. In Chinese, if one item is 20% off, we say it’s “8折” (8 comes from (1-20%)*10).

So if you insist on this translation you will need to change the related code. I would suggest you to simply translate "%d%% off" as "优惠%d%%"(Simplified Chinese) and "優惠%d%%" (Traditional Chinese), which aligns with the word ”off”.

Last issue regarding formatting: "Default_offer_details_with_intro_offer" = "开始 {{ sub_offer_duration }} 试用，然后 {{total_price_and_per_month }}。";  I don’t think we need spaces around “{{ sub_offer_duration }}” and ”{{total_price_and_per_month }}”.

### DIFF
--- a/RevenueCatUI/Resources/zh-Hans.lproj/Localizable.strings
+++ b/RevenueCatUI/Resources/zh-Hans.lproj/Localizable.strings
@@ -14,6 +14,6 @@
 "Monthly" = "每月";
 "Weekly" = "每周";
 "Lifetime" = "永久";
-"%d%% off" = "%d%%折";
+"%d%% off" = "优惠%d%%";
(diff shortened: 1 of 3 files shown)


aboedo:
@francisfeng thanks for opening this up!

aboedo (approved):

PR #3493, opened 2023-12-05
Paywalls: Add dismissal method in `PaywallViewControllerDelegate`

tonidero:
### Description
This adds a new method to the `PaywallViewControllerDelegate` to indicate when the `PaywallViewController` is dismissed.

This will be useful to listen to responses in the hybrids, since we need to present the paywall and wait until it's been purchased/dismissed before returning a value for the hybrids.

This will be needed to fix: https://github.com/RevenueCat/purchases-flutter/issues/886

### DIFF
--- a/RevenueCatUI/UIKit/PaywallViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallViewController.swift
@@ -86,8 +86,8 @@ public final class PaywallViewController: UIViewController {
     }
 
     public override func viewDidDisappear(_ animated: Bool) {
-        if isBeingDismissed {
-            self.delegate?.paywallViewControllerDismissed?(self)
+        if self.isBeingDismissed {
+            self.delegate?.paywallViewControllerWasDismissed?(self)
         }
         super.viewDidDisappear(animated)
     }
@@ -120,7 +120,7 @@ public protocol PaywallViewControllerDelegate: AnyObject {
 
     /// Notifies that the ``PaywallViewController`` was dismissed.
     @objc(paywallViewControllerDismissed:)
-    optional func paywallViewControllerDismissed(_ controller: PaywallViewController)
+    optional func paywallViewControllerWasDismissed(_ controller: PaywallViewController)
 
 }
 
--- a/RevenueCatUI/UIKit/PaywallViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallViewController.swift
@@ -85,6 +85,12 @@ public final class PaywallViewController: UIViewController {
         self.hostingController.view.frame = self.view.bounds
     }
 
+    public override func viewDidDisappear(_ animated: Bool) {
+        if isBeingDismissed {
+            self.delegate?.paywallViewControllerDismissed?(self)
+        }
+        super.viewDidDisappear(animated)
+    }
 }
 
 /// Delegate for ``PaywallViewController``.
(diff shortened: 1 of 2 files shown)


RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -85,6 +85,12 @@ public final class PaywallViewController: UIViewController {
         self.hostingController.view.frame = self.view.bounds
     }
 
+    public override func viewDidDisappear(_ animated: Bool) {
+        if isBeingDismissed {
+            self.delegate?.paywallViewControllerDismissed?(self)
```
tonidero, line 90: I'm not certain this is the best way to do this. Afaik, in iOS, the presenter is usually in charge of dismissing the presented VC, but not sure if the same applies to modals, where it can be dismissed by user interaction. And we want to know when it's dismissed so the hybrids can return a result at that point in time. Any feedback would be welcome @aboedo @NachoSoto

tonidero: I guess in the hybrids, it's probably not a modal though, so this might not work...

aboedo: the footer wouldn't own the modal anyway, right? I think we should just leave it up to the owner of the view that has a footer in it to dismiss the whole thing, which in turn will kill the footer

tonidero: Right, this is a bit confusing, but this fix is for the old approach of presenting the paywall as a modal in the hybrids and fix the issue I mentioned in the description... But yeah, if we end up moving to using views, this becomes unnecessary.

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -112,6 +118,10 @@ public protocol PaywallViewControllerDelegate: AnyObject {
     optional func paywallViewController(_ controller: PaywallViewController,
                                         didFinishRestoringWith customerInfo: CustomerInfo)
 
+    /// Notifies that the ``PaywallViewController`` was dismissed.
+    @objc(paywallViewControllerDismissed:)
+    optional func paywallViewControllerDismissed(_ controller: PaywallViewController)
```
NachoSoto, line 123: This makes sense 👍🏻 
I'd rename this though:
```suggestion
    optional func paywallViewControllerWasDismissed(_ controller: PaywallViewController)
```

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -85,6 +85,12 @@ public final class PaywallViewController: UIViewController {
         self.hostingController.view.frame = self.view.bounds
     }
 
+    public override func viewDidDisappear(_ animated: Bool) {
+        if isBeingDismissed {
```
NachoSoto, line 89: Explicit self so it's clear this isn't a global:
```suggestion
        if self.isBeingDismissed {
```

NachoSoto (commented):
This makes sense :+1

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -112,6 +118,10 @@ public protocol PaywallViewControllerDelegate: AnyObject {
     optional func paywallViewController(_ controller: PaywallViewController,
                                         didFinishRestoringWith customerInfo: CustomerInfo)
 
+    /// Notifies that the ``PaywallViewController`` was dismissed.
+    @objc(paywallViewControllerDismissed:)
```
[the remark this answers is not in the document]
tonidero: 🤦‍♂️ No, I just forgot to update it in objc. Fixed and thanks!

NachoSoto (approved):

PR #3495, opened 2023-12-06
`Paywalls`: fix empty description when using `custom` package type and `{{ sub_period }}`

aboedo:
If you select a custom package type and set Offer Period, the paywall would have an empty entry for that package. 

image

| Before | After |
| :-: | :-: |
| IMG_908CED5B663B-1 (1) | IMG_B1CED574DF20-1 |

### DIFF
--- a/RevenueCatUI/Helpers/Package+VariableDataProvider.swift
+++ b/RevenueCatUI/Helpers/Package+VariableDataProvider.swift
@@ -40,11 +40,8 @@ extension Package: VariableDataProvider {
     }
 
     func periodName(_ locale: Locale) -> String {
-        if packageType == .custom {
-            return self.identifier
-        }
         return Localization.localized(packageType: self.packageType,
-                                      locale: locale)
+                                      locale: locale) ?? self.identifier
     }
 
     func subscriptionDuration(_ locale: Locale) -> String? {
--- a/RevenueCatUI/Helpers/Package+VariableDataProvider.swift
+++ b/RevenueCatUI/Helpers/Package+VariableDataProvider.swift
@@ -40,6 +40,9 @@ extension Package: VariableDataProvider {
     }
 
     func periodName(_ locale: Locale) -> String {
+        if packageType == .custom {
+            return self.identifier
+        }
         return Localization.localized(packageType: self.packageType,
                                       locale: locale)
     }
--- a/RevenueCatUI/Helpers/Package+VariableDataProvider.swift
+++ b/RevenueCatUI/Helpers/Package+VariableDataProvider.swift
@@ -39,14 +39,14 @@ extension Package: VariableDataProvider {
         return self.storeProduct.localizedTitle
     }
 
-    func periodName(_ locale: Locale) -> String {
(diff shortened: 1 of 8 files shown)


Tests/RevenueCatUITests/Data/PackageVariablesTests.swift
```diff
@@ -134,6 +134,7 @@ class PackageVariablesTests: TestCase {
         expect(TestData.sixMonthPackage.periodName(Self.english)) == "6 Month"
         expect(TestData.annualPackage.periodName(Self.english)) == "Annual"
         expect(TestData.lifetimePackage.periodName(Self.english)) == "Lifetime"
+        expect(TestData.customPackage.periodName(Self.english)) == "Custom"
```
NachoSoto, line 137: <3

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
```diff
@@ -40,6 +40,9 @@ extension Package: VariableDataProvider {
     }
 
     func periodName(_ locale: Locale) -> String {
+        if packageType == .custom {
```
NachoSoto, line 43: The issue seems to be that `PackageType.localizationKey` returns `nil` for `.custom` **but also `.unknown`**. 
I'd suggest a slightly different implementation to fix `.unknown` too: change `Localization.localized(packageType:locale:)` to return an `Optional`, and provide `self.identifier` as the fallback here.

NachoSoto (commented):
Awesome, nice find!

Just one comment about `.unknown`.

RevenueCatUI/Templates/Template4View.swift
```diff
@@ -325,7 +325,7 @@ private struct PackageButton: View {
         // Placeholder to make sure consistent layout
         self.offerText(firstRow: "12",
                        secondRow: Localization.localized(packageType: .monthly,
-                                                         locale: self.locale))
+                                                         locale: self.locale) ?? "Monthly")
```
NachoSoto, line 328: Oh this should be `Localization.localizedDuration(for: .monthly, locale: self.locale)` instead

NachoSoto: Just pushed this 😇

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
```diff
@@ -41,7 +41,7 @@ extension Package: VariableDataProvider {
 
     func periodName(_ locale: Locale) -> String {
         return Localization.localized(packageType: self.packageType,
-                                      locale: locale)
+                                      locale: locale) ?? self.identifier
     }
```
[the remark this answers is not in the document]
NachoSoto: That would lead to duplicated code, but I agree it can be confusing. I'll add a commend.
I do think the implementation makes sense, this is used like: `package.periodName()`. So it's kind of an implementation detail that if the package type doesn't have a "period" we default to the package identifier.

NachoSoto (approved):

PR #3498, opened 2023-12-07
`Paywalls`: improve image caching

NachoSoto:
Supersedes #3477.

This replaces `AsyncImage` with a custom implementation that allows us to ensure caching works reliably by explicitly using a custom `URLSession` with a shared `URLCache`.


RevenueCatUI/Helpers/ImageLoader.swift
```diff
@@ -0,0 +1,119 @@
...
+                return nil
+            }
+
+            if let httpResponse = response as? HTTPURLResponse, httpResponse.statusCode != 200 {
+                return .failure(.badResponse(.init(.badServerResponse)))
+            }
+
+            // Load images in a background thread
+            return await Task<Value, Never>
+                .detached(priority: .utility) { data.toImage() }
```
NachoSoto, line 89: This is new

RevenueCatUI/Helpers/ImageLoader.swift
```diff
@@ -0,0 +1,119 @@
...
+private extension Data {
+
+    @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
+    func toImage() -> ImageLoader.Value {
+        #if os(macOS)
+        if let image = NSImage(data: self) {
+            return .success(.init(nsImage: image))
+        } else {
+            return .failure(.invalidImage)
+        }
```
NachoSoto, line 109: We don't support `macOS` yet, but this makes the class compile, and it'll be ready when we do.

RevenueCatUI/Views/RemoteImage.swift
```diff
@@ -20,51 +20,67 @@ struct RemoteImage: View {
...
                     image
                         .resizable()
                 }
+
+            case let .failure(error):
+                self.emptyView(error: error)
+            }
+        }
+        .transition(Self.transition)
+        .task(id: self.url) {
```
NachoSoto, line 54: This automatically cancels the task when the url changes.

RevenueCatUI/Helpers/ImageLoader.swift
aboedo: Feels like `utility` might be a bit low? 

I'd imagined we'd have the pre-warming loader use `utility` or `background` and then ~`userInitiated`~ `default` when loading the paywall, given the potential impact on conversion for an image taking too long to load, and that there's probably not a lot more going on that's more important than the image at the time of displaying a paywall. 

What do you think?

RevenueCatUI/Helpers/ImageLoader.swift
aboedo: great catch, I keep forgetting macOS

RevenueCatUI/Views/RemoteImage.swift
aboedo: 🤔 let's maybe add a comment? Feels like that would be easy to miss

Tests/RevenueCatUITests/ImageLoaderTests.swift
```diff
@@ -0,0 +1,213 @@
...
+
+        await self.loader.load(url: Self.url)
+
+        expect(self.loader.result).to(beSuccess())
+        expect(self.loader.result?.value?.pngData()) == UIImage(data: response.0)?.pngData()
+        expect(self.urlSession.requestedImage) == Self.url
+        expect(self.urlSession.cachePolicy) == .returnCacheDataElseLoad
+    }
+
+    func testLoadingImageResetsPreviousResult() async throws {
```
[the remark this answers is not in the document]
aboedo: makes sense. At least the code is very clear

aboedo:
couple of small questions, looking great though

RevenueCatUI/Data/Strings.swift
```diff
@@ -16,6 +16,7 @@ import RevenueCat
 
 // swiftlint:disable identifier_name
 
+@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
```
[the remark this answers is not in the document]
NachoSoto: Oops thanks, not sure where I got this.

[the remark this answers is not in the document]
NachoSoto: Oh this is because some method I needed from URLSession wasn't available until specifically these versions. So having a lower watchOS one doesn't matter, it's just overly available. But I'll update for consistency.

RevenueCatUI/Helpers/ImageLoader.swift
NachoSoto: The pre-warming is actually happening with `.default` QOS because that's what `OperationDispatcher.workerQueue` is set up with :/ 
We should talk about changing that, maybe having multiple queues.

But I agree, I'll change this to `default`.

RevenueCatUI/Views/RemoteImage.swift
NachoSoto: It's the canonical way of implementing task: making the ID whatever parameter the task depends on. But yeah I'll add a comment 👍🏻

NachoSoto:
@aboedo all comments addressed.

tonidero (approved):
LGTM!

aboedo (approved):

PR #3510, opened 2023-12-10
`Paywalls`: fix PaywallViewControllerDelegate access from Objective-C

noncenz:
Public variables must be exposed with `@objc` to be accessible outside of Swift. This change is necessary to set delegate from host project so that we may dismiss the paywall.

### Checklist

### Motivation
Paywall doesn't seem to be usable in an Objective-C project as there is no way to receive the callback to drop the paywall other than by setting a delegate. 

### Description
Added code to expose the public variable `delegate` in PaywallViewController to Objective-C.
Tested by setting delegate and receiving delegate method callbacks in Obj-C code.


NachoSoto (approved):
Thank you!! :pray:

PR #3512, opened 2023-12-11
Fix typo in zh-Hans localization of RevenueCatUI

francisfeng:
“終” is Traditional Chinese while “终” is Simplified Chinese.


NachoSoto (approved):
Thank you!

PR #3526, opened 2023-12-19
Fix `RevenueCatUI` snapshot tests

NachoSoto:
Turns out that because neither https://github.com/pointfreeco/swift-snapshot-testing/pull/702 or https://github.com/pointfreeco/swift-snapshot-testing/pull/666 had been merged, we had been getting false positives on snapshot tests on CI since the very beginning.

The `perceptualPrecision` implementation requires `Metal` support, and without it `swift-snapshot-testing` was silently passing.

This fixes that, and prevents it from happening again.

### Changes:
- Changed `CircleCI` to use `M1` instances, which do support `Metal`
- Added assertion to catch this from happening again
- Replaced `ProgressView` with a static version in snapshots to prevent failures from the spinning indicator
- Re-generated snapshots from M1 instances


NachoSoto:
TODO: replace LoadingView with something else for snapshots to test

.circleci/config.yml
```diff
@@ -1028,56 +1031,12 @@ workflows:
...
-          xcode_version: '14.3.0'
-      - spm-receipt-parser:
-          xcode_version: '14.3.0'
       - spm-revenuecat-ui-ios-15:
           xcode_version: '14.3.0'
       - spm-revenuecat-ui-ios-16:
           xcode_version: '14.3.0'
       - spm-revenuecat-ui-ios-17:
           xcode_version: '15.1'
-      - spm-revenuecat-ui-watchos:
```
NachoSoto, line 1049: TODO: remove commit

Tests/RevenueCatUITests/BaseSnapshotTest.swift
```diff
@@ -33,10 +33,16 @@ import XCTest
...
-    // swiftlint:disable:next unneeded_override
     override class func setUp() {
         super.setUp()
 
+        // See https://github.com/pointfreeco/swift-snapshot-testing/pull/702 and
+        // https://github.com/pointfreeco/swift-snapshot-testing/pull/666
+        expect(MTLCreateSystemDefaultDevice()).toNot(
+            beNil(),
+            description: "Metal is required for perceptuallyCompare, but not available on this machine."
+        )
```
MarkVillacampa, line 44: 😍

Tests/UnitTests/TestHelpers/SnapshotTesting+Extensions.swift
```diff
@@ -155,8 +155,8 @@ private let traits: UITraitCollection = .init(displayScale: 1)
 
 #endif
 
-private let perceptualPrecision: Float = 0.97
-private let timeout: DispatchTimeInterval = .seconds(5)
+private let perceptualPrecision: Float = 0.94
```
MarkVillacampa, line 158: Just out of curiosity, what failed if we keep it at 97?

MarkVillacampa (approved):

Tests/UnitTests/TestHelpers/SnapshotTesting+Extensions.swift
NachoSoto: No idea, the diff produced showing the inconsistency was completely black to the naked eye. So just very tiny color differences.

PR #3533, opened 2023-12-20
`Tests`: running tests on `macOS`

NachoSoto:
Launching simulators continues to be a struggle.

Our SDK is compatible with `macOS`, so this PR allows us to run tests without needing to launch a simulator.


Tests/StoreKitUnitTests/LocalReceiptParserStoreKitTests.swift
```diff
@@ -47,7 +47,14 @@ class LocalReceiptParserStoreKitTests: StoreKitConfigTestCase {
         let receipt = try self.parser.parse(from: data)
 
         expect(receipt.bundleId) == "com.revenuecat.StoreKitUnitTestsHostApp"
+        #if os(macOS)
+        expect(receipt.applicationVersion) == SystemInfo.frameworkVersion.replacingOccurrences(
+            of: "-SNAPSHOT",
+            with: ""
+        )
```
aboedo, line 54: why is it tied to framework version on macOS only?

aboedo (approved):
looks good overall, left a question

PR #3534, opened 2023-12-20
`Paywalls`: improve template 5 landscape layout

NachoSoto:
I didn't notice this initially because I had been testing with 3 packages that happened to fill the screen

### Before
Simulator Screenshot - iPhone 15 Pro Max - 2023-12-20 at 14 34 35

### After
Simulator Screenshot - iPhone 15 Pro Max - 2023-12-20 at 14 34 24


RevenueCatUI/Templates/Template5View.swift
```diff
@@ -82,6 +82,8 @@ struct Template5View: TemplateViewType {
                     .scrollableIfNecessary()
             }
 
+            Spacer()
```
tonidero, line 85: Just curious but how does this look on an iPad in landscape? Just wondering whether the main content will be moved too much to the top, since I think we might want it centered within the available space?

tonidero (approved):
Just a question, but this should be good for now I think.

RevenueCatUI/Templates/Template5View.swift
NachoSoto: iPad landcape doesn't have "vertical height condensed", so this layout isn't used.

PR #3547, opened 2023-12-29
Paywalls: improve footer view UIKit support for hybrid SDKs

aboedo:
### Motivation
Improves support for FooterView by adding layout constraints so that the footer can automatically size itself, and adding a clear background for the cutout to show properly

Simulator Screenshot - iPhone 15 - 2023-12-29 at 18 31 38

### DIFF
--- a/RevenueCatUI/UIKit/PaywallFooterViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallFooterViewController.swift
@@ -51,7 +51,6 @@ public final class PaywallFooterViewController: PaywallViewController {
     public required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
-
 }
 
 #endif
--- a/RevenueCatUI/UIKit/PaywallFooterViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallFooterViewController.swift
@@ -52,24 +52,6 @@ public final class PaywallFooterViewController: PaywallViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
-    public override func loadView() {
-        super.loadView()
-
-        self.addChild(self.hostingController)
-        self.view.addSubview(self.hostingController.view)
-
-        NSLayoutConstraint.activate([
-            hostingController.view.bottomAnchor.constraint(equalTo: view.bottomAnchor),
-            hostingController.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
-            hostingController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor)
-        ])
-
-        self.hostingController.didMove(toParent: self)
-
-        hostingController.view.translatesAutoresizingMaskIntoConstraints = false
-        hostingController.view.backgroundColor = .clear
-    }
-
 }
 
 #endif
--- a/RevenueCatUI/UIKit/PaywallFooterViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallFooterViewController.swift
@@ -51,6 +51,25 @@ public final class PaywallFooterViewController: PaywallViewController {
     public required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+
+    public override func loadView() {
+        super.loadView()
+
+        self.addChild(self.hostingController)
+        self.view.addSubview(self.hostingController.view)
+
+        NSLayoutConstraint.activate([
+            hostingController.view.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+            hostingController.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            hostingController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor)
+        ])
+
+        self.hostingController.didMove(toParent: self)
+
+        hostingController.view.translatesAutoresizingMaskIntoConstraints = false
+        hostingController.view.backgroundColor = .clear
+    }
+
 }
 
 #endif
--- a/RevenueCatUI/UIKit/PaywallViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallViewController.swift
@@ -104,13 +104,6 @@ public class PaywallViewController: UIViewController {
         // make the background of the container clear so that if there are cutouts, they don't get
         // overridden by the hostingController's view's background.
         hostingController.view.backgroundColor = .clear
-
-    }
-
-    public override func viewWillLayoutSubviews() {
-        super.viewWillLayoutSubviews()
-
-        self.hostingController.view.frame = self.view.bounds
     }
 
     public override func viewDidDisappear(_ animated: Bool) {
--- a/RevenueCatUI/UIKit/PaywallViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallViewController.swift
@@ -58,7 +58,7 @@ public class PaywallViewController: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
-    lazy var hostingController: UIHostingController<some View> = {
+    private lazy var hostingController: UIHostingController<some View> = {
         let view = PaywallView(offering: self.offering,
                                customerInfo: nil,
                                mode: self.mode,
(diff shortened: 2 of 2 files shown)


aboedo:
@vegaro the full screen paywall isn't working correctly for me, but it also wasn't working correctly for me before this PR so maybe I'm missing something or I'm using some WIP stuff? I'm basing off of the branches we talked about on Friday

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -54,7 +54,7 @@ public class PaywallViewController: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
-    private lazy var hostingController: UIHostingController<some View> = {
+    lazy var hostingController: UIHostingController<some View> = {
```
NachoSoto, line 57:
```suggestion
    private(set) lazy var hostingController: UIHostingController<some View> = {
```

RevenueCatUI/UIKit/PaywallFooterViewController.swift
```diff
@@ -51,6 +51,25 @@ public final class PaywallFooterViewController: PaywallViewController {
     public required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+
+    public override func loadView() {
+        super.loadView()
+
+        self.addChild(self.hostingController)
+        self.view.addSubview(self.hostingController.view)
```
NachoSoto, line 59: The parent class is already doing this though?

RevenueCatUI/UIKit/PaywallFooterViewController.swift
```diff
@@ -51,6 +51,25 @@ public final class PaywallFooterViewController: PaywallViewController {
...
+
+        NSLayoutConstraint.activate([
```
NachoSoto, line 61: Hmm won't the parent's layout mess with this?

RevenueCatUI/UIKit/PaywallFooterViewController.swift
```diff
@@ -51,6 +51,25 @@ public final class PaywallFooterViewController: PaywallViewController {
...
+            hostingController.view.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+            hostingController.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            hostingController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor)
+        ])
+
+        self.hostingController.didMove(toParent: self)
+
+        hostingController.view.translatesAutoresizingMaskIntoConstraints = false
+        hostingController.view.backgroundColor = .clear
```
NachoSoto, line 70: This is probably not needed right?

RevenueCatUI/UIKit/PaywallFooterViewController.swift
NachoSoto: I would change the parent implementation to use constraints, expose the top constraint, and deactivate it in this subclass.

RevenueCatUI/UIKit/PaywallFooterViewController.swift
aboedo: 🤦 I was rushing through things and missed it, thanks

RevenueCatUI/UIKit/PaywallFooterViewController.swift
aboedo: should have added a note, but it's needed so that you can actually see the cutout of the footer. otherwise the view from the hosting controller, which doesn't have the cutout and has a white background, will show its background behind the cutout

RevenueCatUI/UIKit/PaywallFooterViewController.swift
aboedo: it doesn't in my testing, but that sounds like a good approach

RevenueCatUI/UIKit/PaywallViewController.swift
aboedo: good call

aboedo:
update on the full screen paywalls thing: it works correctly with the changes from this PR and https://github.com/RevenueCat/react-native-purchases/pull/822

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -31,6 +31,10 @@ public class PaywallViewController: UIViewController {
         return .fullScreen
     }
 
+    var shouldAddTopConstraint: Bool {
```
tonidero, line 34: This seems unused?

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -82,12 +86,24 @@ public class PaywallViewController: UIViewController {
...
+        hostingController.view.translatesAutoresizingMaskIntoConstraints = false
+
+        NSLayoutConstraint.activate([
+            hostingController.view.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+            hostingController.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            hostingController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor)
+        ])
+
+        // for footers we only want to tie to the sides and the bottom.
+        if self.mode != .footer && self.mode != .condensedFooter {
```
tonidero, line 98: Would it be easier to check if it's full screen instead?

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -82,12 +86,24 @@ public class PaywallViewController: UIViewController {
         self.addChild(self.hostingController)
         self.view.addSubview(self.hostingController.view)
         self.hostingController.didMove(toParent: self)
-    }
-
-    public override func viewWillLayoutSubviews() {
-        super.viewWillLayoutSubviews()
+        hostingController.view.translatesAutoresizingMaskIntoConstraints = false
```
NachoSoto, line 89: Explicit self here and below:
```suggestion
        self.hostingController.view.translatesAutoresizingMaskIntoConstraints = false
```

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -82,6 +86,25 @@ public class PaywallViewController: UIViewController {
...
+            hostingController.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            hostingController.view.trailingAnchor.constraint(equalTo: view.trailingAnchor)
+        ])
+
+        // for footers we only want to tie to the sides and the bottom.
+        if self.mode != .footer && self.mode != .condensedFooter {
+            NSLayoutConstraint.activate([
+                hostingController.view.topAnchor.constraint(equalTo: view.topAnchor)
+            ])
+        }
```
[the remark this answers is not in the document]
NachoSoto: Much better 👍🏻

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -82,6 +86,25 @@ public class PaywallViewController: UIViewController {
...
+
+        // make the background of the container clear so that if there are cutouts, they don't get
+        // overridden by the hostingController's view's background.
+        hostingController.view.backgroundColor = .clear
```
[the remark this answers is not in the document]
NachoSoto: 👍🏻

NachoSoto (approved):
Much better. Only main thing is the unused property.

RevenueCatUI/UIKit/PaywallViewController.swift
aboedo: lol my bad, I added it and then realized that I already had the info in the mode variable

RevenueCatUI/UIKit/PaywallViewController.swift
aboedo: That was my original code, and yeah, it looks simpler. 

I modified it to this because I felt like it was more "correct" to say that footers don't need to be constrained to the top, rather than full screen being the special case. Like, if we had yet another type (i.e.: banner, badge, etc), then we'd have to also special case those. 

But that's super easy to refactor again in the future, and we don't know as of right now what types we'll have, so I'll go with checking only full screen for now. Updating

PR #3549, opened 2024-01-02
Improve sandbox detector for macOS apps

MarkVillacampa:
Checking for the presence of `Xcode/DerivedData` in the receipt path was failing for TestFlight builds or if the .app bundle generated by Xcode was moved from the folder.

Additionally, checking for `MASReceipt/receipt` seems less robust than simply using compiler directives.


Tests/UnitTests/Misc/SandboxEnvironmentDetectorTests.swift
```diff
@@ -30,14 +30,6 @@ class SandboxEnvironmentDetectorTests: TestCase {
         expect(SystemInfo.with(receiptResult: .nilURL, inSimulator: false).isSandbox) == false
     }
 
-    func testMacSandboxReceiptIsSandbox() {
-        expect(SystemInfo.with(receiptResult: .macOSSandboxReceipt, inSimulator: false).isSandbox) == true
-    }
-
-    func testMacAppStoreReceiptIsNotSandbox() {
-        expect(SystemInfo.with(receiptResult: .macOSAppStoreReceipt, inSimulator: false).isSandbox) == false
-    }
```
MarkVillacampa, line 39: Removed because 1. we dont currently run tests on mac targets 2. the new way of testing for sandbox environment on macos does not depend on the receipt path

Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -40,14 +40,51 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
...
+    /// This routine is inspired by the source code from ProcInfo, the underlying library
+    /// of the WhatsYourSign code signature checking tool developed by Objective-See. Initially,
+    /// it checked the common name but was changed to an extension check to make it more
+    /// future-proof.
+    ///
+    /// For more information, see the following references:
+    /// - https://github.com/objective-see/ProcInfo/blob/master/procInfo/Signing.m#L184-L247
+    /// - https://gist.github.com/lukaskubanek/cbfcab29c0c93e0e9e0a16ab09586996#gistcomment-3993808
+    #if os(macOS) || targetEnvironment(macCatalyst)
+    var isMacAppStore: Bool {
```
NachoSoto, line 62: This could be `static`

Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -40,14 +40,51 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
...
+        var status = noErr
+
+        var code: SecStaticCode?
+        status = SecStaticCodeCreateWithPath(Bundle.main.bundleURL as CFURL, [], &code)
+
+        guard status == noErr, let code = code else { return false }
```
NachoSoto, line 68: Should we log the error?

Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -40,14 +40,51 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
...
+
+    /// Returns whether the bundle was signed for Mac App Store distribution by checking
+    /// the existence of a specific extension (marker OID) on the code signing certificate.
+    ///
+    /// This routine is inspired by the source code from ProcInfo, the underlying library
+    /// of the WhatsYourSign code signature checking tool developed by Objective-See. Initially,
+    /// it checked the common name but was changed to an extension check to make it more
+    /// future-proof.
+    ///
+    /// For more information, see the following references:
```
NachoSoto, line 58: 🙏🏻

Sources/Misc/SandboxEnvironmentDetector.swift
NachoSoto: And should be `private`. Maybe move it to an extension?

Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -40,14 +40,51 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
...
+        guard status == noErr, let code = code else { return false }
+
+        var requirement: SecRequirement?
+        status = SecRequirementCreateWithString(
+            "anchor apple generic and certificate leaf[field.1.2.840.113635.100.6.1.9]" as CFString,
+            [], // default
+            &requirement
+        )
+
+        guard status == noErr, let requirement = requirement else { return false }
```
NachoSoto, line 77: Let's log a warning here too?

Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -40,14 +40,51 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
             return false
         }
 
-        // `true` for either `macOS` or `Catalyst`
-        let isMASReceipt = path.contains("MASReceipt/receipt")
-        if isMASReceipt {
-            return path.contains("Xcode/DerivedData")
-        } else {
+        #if os(macOS) || targetEnvironment(macCatalyst)
+            return !isMacAppStore
```
NachoSoto, line 44:
```suggestion
            return !Self.isMacAppStore
```

NachoSoto (commented):
Awesome thanks for doing this!

Sources/LocalReceiptParsing/BasicTypes/AppleReceipt.swift
```diff
@@ -18,6 +18,24 @@ import Foundation
...
+        case production = "Production"
+
+        /// Development build or downloaded from TestFlight
+        case sandbox = "ProductionSandbox"
+
+        /// StoreKit Testing in Xcode
+        case xcode = "Xcode"
+
+        /// Unknown environment
+        case unknown = "Unknown"
```
MarkVillacampa, line 33: There is also `ProductionVPP` and `ProductionVPPSandbox` for Apple's Volume Purchase Program, but I believe those apps cannot have in-app purchases anyway.

Sources/LocalReceiptParsing/BasicTypes/AppleReceipt.swift
```diff
@@ -18,6 +18,24 @@ import Foundation
...
+    }
+
+    /// The receipt's environment.
+    public let environment: Environment?
```
MarkVillacampa, line 37: Making it optional since the value is not publicly documented by Apple as a receipt field

Sources/Misc/SystemInfo.swift
```diff
@@ -57,8 +57,12 @@ class SystemInfo {
     private static let _forceUniversalAppStore: Atomic<Bool> = false
     private static let _proxyURL: Atomic<URL?> = nil
 
-    var isSandbox: Bool {
+    private lazy var _isSandbox: Bool = {
         return self.sandboxEnvironmentDetector.isSandbox
+    }()
```
MarkVillacampa, line 62: Caching the result since the sandbox detector is a bit more CPU-intensive, and does not change.

Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -40,12 +44,19 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
             return false
```
MarkVillacampa, line 44: I'd argue if `appStoreReceiptUR` is nil then we are in sandbox environment, so this should be true?

I have not been able to find a case where `appStoreReceiptURL` is nil tho. The URL is always returned, but then the file does not exist in some cases.

Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -25,10 +25,14 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
 
     private let bundle: Atomic<Bundle>
     private let isRunningInSimulator: Bool
+    private let receiptParser: PurchasesReceiptParser
 
-    init(bundle: Bundle = .main, isRunningInSimulator: Bool = SystemInfo.isRunningInSimulator) {
+    init(bundle: Bundle = .main,
+         isRunningInSimulator: Bool = SystemInfo.isRunningInSimulator,
+         receiptParser: PurchasesReceiptParser = PurchasesReceiptParser.default) {
```
NachoSoto, line 32: What do you think of injecting a protocol type so we don't tie this to receipt parser or fetcher (and make it easier to test)?

Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -40,12 +44,19 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
...
-            return path.contains("Xcode/DerivedData")
-        } else {
+        #if os(macOS) || targetEnvironment(macCatalyst)
+            return !isProductionReceipt || !Self.isMacAppStore
+        #else
             return path.contains("sandboxReceipt")
+        #endif
+    }
+
+    var isProductionReceipt: Bool {
```
NachoSoto, line 54: This can be private?

Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -40,12 +44,19 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
             return false
         }
 
-        // `true` for either `macOS` or `Catalyst`
-        let isMASReceipt = path.contains("MASReceipt/receipt")
-        if isMASReceipt {
-            return path.contains("Xcode/DerivedData")
-        } else {
+        #if os(macOS) || targetEnvironment(macCatalyst)
+            return !isProductionReceipt || !Self.isMacAppStore
```
NachoSoto, line 48:
```suggestion
            return !self.isProductionReceipt || !Self.isMacAppStore
```

Sources/Misc/SandboxEnvironmentDetector.swift
NachoSoto: Maybe move it to the extension below?

Sources/Misc/SystemInfo.swift
NachoSoto: I was worried this was suddenly a lot less efficient, but:
- It's macOS only
- It's cached 👏🏻

Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -40,12 +44,19 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
             return false
         }
 
-        // `true` for either `macOS` or `Catalyst`
-        let isMASReceipt = path.contains("MASReceipt/receipt")
-        if isMASReceipt {
-            return path.contains("Xcode/DerivedData")
-        } else {
+        #if os(macOS) || targetEnvironment(macCatalyst)
```
NachoSoto, line 47: Feel free to do it on a separate PR (maybe pointing to the macOS tests branch), but this could really use tests. I'm afraid that if we don't add them now we'll forget.

Sources/LocalReceiptParsing/Builders/AppleReceiptBuilder.swift
```diff
@@ -70,6 +71,11 @@ class AppleReceiptBuilder {
             let payload = valueContainer.internalPayload
 
             switch attributeType {
+            case .environment:
+                let internalContainer = try containerBuilder.build(fromPayload: payload)
+                if let environmentString = internalContainer.internalPayload.toString() {
+                    environment = .init(rawValue: environmentString) ?? .unknown
```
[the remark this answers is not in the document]
NachoSoto: But do we need those 2 cases? Isn't "not present in the receipt" the same as unknown?

Sources/LocalReceiptParsing/LocalReceiptFetcher.swift
```diff
@@ -7,13 +7,17 @@
...
+//  Created by Mark Villacampa on 8/1/24.
 
 import Foundation
 
-public extension PurchasesReceiptParser {
+internal protocol LocalReceiptFetcherType: Sendable {
+    func fetchAndParseLocalReceipt() throws -> AppleReceipt
+}
+
+internal final class LocalReceiptFetcher: LocalReceiptFetcherType {
```
MarkVillacampa, line 20: `ReceiptFetcher` depends on `StoreKitRequestFetcher` and `SystemInfo` (which would create a circular dependency BundleSandboxEnvironmentDetector).

So instead of adding these methods as an extension to `PurchasesReceiptParser`, I've created `LocalReceiptFetcher` as a simplified version of `ReceiptFetcher`, with an accompanying protocol so it can be tested.

Ideally we would be able to use `ReceiptFetcher` but the dependencies make it hard :/ any other suggestion to simplify this?

Tests/UnitTests/Misc/SandboxEnvironmentDetectorTests.swift
```diff
@@ -16,46 +16,96 @@ import XCTest
...
-        expect(SystemInfo.with(receiptResult: .receiptWithData, inSimulator: true).isSandbox) == true
-        expect(SystemInfo.with(receiptResult: .nilURL, inSimulator: true).isSandbox) == true
+    func testIsSandboxIfReceiptIsNotProduction() throws {
+        try AvailabilityChecks.skipIfNotMacOS()
+
+        expect(SystemInfo.with(receiptEnvironment: .sandbox).isSandbox) == true
+    }
+
+    func testIsSandboxIfReceiptParsingFailsAndBundleSignatureIsNotMacAppStore() throws {
+        try AvailabilityChecks.skipIfNotMacOS()
```
MarkVillacampa, line 70: Added some tests for macOS, which do not currently run so I cannot properly test them. Once the macOS tests PR is ready I'll make sure to fix any errors if tests don't pass.

Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -59,3 +61,61 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
 }
 
 extension BundleSandboxEnvironmentDetector: Sendable {}
+
+private extension BundleSandboxEnvironmentDetector {
+
+    var isProductionReceipt: Bool {
```
NachoSoto, line 67: Maybe move this inside of the `#if os(macOS)` check too? It's potentially very inefficient for mobile devices, so that way we make sure we don't use it there.

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/LocalReceiptView.swift
```diff
@@ -84,6 +84,7 @@ private struct ReceiptDataView: View {
     init(receipt: AppleReceipt) {
         self.init(
             values: [
+                .init("Environment", (receipt.environment ?? .unknown).rawValue),
```
NachoSoto, line 87: Case in point :P we don't need optional + unknown

Tests/UnitTests/LocalReceiptParsing/Builders/AppleReceiptBuilderTests.swift
```diff
@@ -108,8 +109,21 @@ class AppleReceiptBuilderTests: TestCase {
         }
     }
 
+    func testBuildDoesntThrowIfEnvironmentIsMissing() {
```
NachoSoto, line 112: <3

Tests/UnitTests/Misc/SandboxEnvironmentDetectorTests.swift
```diff
@@ -16,46 +16,96 @@ import XCTest
 
 @testable import RevenueCat
 
+final class MockLocalReceiptFetcher: LocalReceiptFetcherType {
```
NachoSoto, line 19: This can be `private`. Can you move it to the bottom of the file, since it's just an implementation detail? That way the first thing in the file is the actual tests.

Tests/UnitTests/Misc/SandboxEnvironmentDetectorTests.swift
```diff
@@ -16,46 +16,96 @@ import XCTest
...
+        self.mockReceipt = mockReceipt
+        self.failReceiptParsing = failReceiptParsing
+    }
+
+    func fetchAndParseLocalReceipt() throws -> RevenueCat.AppleReceipt {
+        if failReceiptParsing {
+            throw PurchasesReceiptParser.Error.receiptParsingError
+        }
+        return mockReceipt
+    }
```
NachoSoto, line 34:
```suggestion
    }
    
```

Tests/UnitTests/Misc/SandboxEnvironmentDetectorTests.swift
```diff
@@ -16,46 +16,96 @@ import XCTest
...
+    init(mockReceipt: AppleReceipt, failReceiptParsing: Bool) {
+        self.mockReceipt = mockReceipt
+        self.failReceiptParsing = failReceiptParsing
+    }
+
+    func fetchAndParseLocalReceipt() throws -> RevenueCat.AppleReceipt {
+        if failReceiptParsing {
+            throw PurchasesReceiptParser.Error.receiptParsingError
+        }
+        return mockReceipt
```
NachoSoto, line 33:
```suggestion
        return self.mockReceipt
```

Sources/LocalReceiptParsing/LocalReceiptFetcher.swift
NachoSoto: Yeah that's why we had this. In the future we'll be able to share code using the `package` visibility.

NachoSoto (commented):
This looks great 👍🏻  The only thing left is the unknown / nil. Do you still think we need optional?

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Views/LocalReceiptView.swift
MarkVillacampa: Made it non-optional, with `.unknown` as default, and overridden by whatever comes in the receipt :D

Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -59,3 +61,62 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
 }
 
 extension BundleSandboxEnvironmentDetector: Sendable {}
+
+private extension BundleSandboxEnvironmentDetector {
+
+    #if os(macOS) || targetEnvironment(macCatalyst)
+    var isProductionReceipt: Bool {
+        do {
+            return try receiptFetcher.fetchAndParseLocalReceipt().environment == .production
```
NachoSoto, line 70:
```suggestion
            return try self.receiptFetcher.fetchAndParseLocalReceipt().environment == .production
```

NachoSoto (approved):
🎉

PR #3552, opened 2024-01-04
Tests: improve test flakiness

aboedo:
Attempting to fix some test flakiness. 

Not the cleanest fixes, but flaky tests are dangerous. 

My methodology was to run tests over and over and fix the ones that would occasionally fail.

### DIFF
--- a/Sources/Purchasing/Purchases/Attribution.swift
+++ b/Sources/Purchasing/Purchases/Attribution.swift
@@ -25,13 +25,15 @@ import Foundation
  */
 @objc(RCAttribution) public final class Attribution: NSObject {
 
+    // internal for testing purposes
+    var automaticAdServicesAttributionTokenCollection: Bool = false
+
     private let subscriberAttributesManager: SubscriberAttributesManager
     private let currentUserProvider: CurrentUserProvider
     private let attributionPoster: AttributionPoster
     private let systemInfo: SystemInfo
 
     private var appUserID: String { self.currentUserProvider.currentAppUserID }
-    private var automaticAdServicesAttributionTokenCollection: Bool = false
 
     weak var delegate: AttributionDelegate?
 
--- a/Tests/UnitTests/Paywalls/Events/PaywallEventsManagerTests.swift
+++ b/Tests/UnitTests/Paywalls/Events/PaywallEventsManagerTests.swift
@@ -189,10 +189,6 @@ class PaywallEventsManagerTests: TestCase {
     }
 
     func testCannotFlushMultipleTimesInParallel() async throws {
-        // The way this test is written does not work in iOS 15.
-        // The second Task does not start until the first one is done.
-        try AvailabilityChecks.iOS16APIAvailableOrSkipTest()
-
         let event1 = await self.storeRandomEvent()
         _ = await self.storeRandomEvent()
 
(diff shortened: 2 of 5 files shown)


Sources/Purchasing/Purchases/Attribution.swift
```diff
@@ -25,13 +25,15 @@ import Foundation
  */
 @objc(RCAttribution) public final class Attribution: NSObject {
 
+    // internal for testing purposes
+    var automaticAdServicesAttributionTokenCollection: Bool = false
+
```
aboedo, line 30: there are cleaner ways to do this, I was going for fast

Tests/UnitTests/Paywalls/Events/PaywallEventsManagerTests.swift
```diff
@@ -199,8 +199,7 @@ class PaywallEventsManagerTests: TestCase {
         let task1 = Task<Int, Error> { [manager = self.manager!] in try await manager.flushEvents(count: 1) }
         let task2 = Task<Int, Error> { [manager = self.manager!] in try await manager.flushEvents(count: 1) }
 
-        let result1 = try await task1.value
-        let result2 = try await task2.value
+        let (result1, result2) = try await (task1.value, task2.value)
```
aboedo, line 202: not entirely sure that this fixes it? but I did want to ensure that we don't run one and _then_ the other

Tests/UnitTests/Purchasing/Purchases/PurchasesLogInTests.swift
```diff
@@ -251,29 +251,43 @@ class ExistingUserPurchasesLogInTests: BasePurchasesLogInTests {
     }
 
     func testLogInClearsTrialEligibilityCache() {
-        expect(self.cachingTrialOrIntroPriceEligibilityChecker.invokedClearCache) == false
+        // set up updates customer info, which clears cache once. wait until it's done before checking the rest.
```
aboedo, line 254: these two tests weren't actually working - commenting out the logIn line would still work. This is because we're mocking the identity manager, which bypasses the customerInfomanager, which bypasses actually triggering a customer info update, which in turn skips the actual cache clearing we're trying to test.

Tests/UnitTests/SubscriberAttributes/PurchasesSubscriberAttributesTests.swift
```diff
@@ -794,7 +794,9 @@ class PurchasesSubscriberAttributesTests: TestCase {
         let token = "token"
 
         self.mockAttributionFetcher.adServicesTokenToReturn = token
```
aboedo, line 796: this test worked even if you comment out the purchasing part, because ad services tokens get posted right away, once per install. This removes that behavior so we can actually test what happens in post receipt

vegaro (approved):
The changes make sense to me

aboedo:
merging for now since these changes do fix the flakiness we have specifically for unit tests, but @NachoSoto if you have some time next week please take a look

Sources/Purchasing/Purchases/Attribution.swift
```diff
@@ -25,13 +25,15 @@ import Foundation
  */
 @objc(RCAttribution) public final class Attribution: NSObject {
 
+    // internal for testing purposes
+    var automaticAdServicesAttributionTokenCollection: Bool = false
```
NachoSoto, line 29: Maybe make it `internal` only `#if DEBUG`?

NachoSoto (commented):
Thanks for doing this!

PR #3554, opened 2024-01-05
`PurchaseTester`: improved `ReceiptInspector` so it accepts receipts with escape sequences

aboedo:
Receipts in logs have extra escape sequences now (forward slashes), so this updates the ReceiptInspector tool to be able to accept them for easier usage. 

It also accepts extra `"` so that if you copied slightly incorrectly it still works.

### DIFF
--- a/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/ReceiptInspector.swift
+++ b/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/ReceiptInspector.swift
@@ -73,7 +73,8 @@ struct ReceiptInspectorView: View {
     func inspectReceipt() async {
         do {
             guard !encodedReceipt.isEmpty else { return }
-            parsedReceipt = try PurchasesReceiptParser.default.parse(base64String: encodedReceipt).debugDescription
+            let receiptWithoutForwardSlashes = encodedReceipt.replacingOccurrences(of: "\\", with: "")
+            parsedReceipt = try PurchasesReceiptParser.default.parse(base64String: receiptWithoutForwardSlashes).debugDescription
             verifyReceiptResult = await ReceiptVerifier().verifyReceipt(base64Encoded: encodedReceipt,
                                                                         sharedSecret: sharedSecret)
         } catch {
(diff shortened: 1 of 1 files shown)


Tests/TestingApps/PurchaseTesterSwiftUI/Shared/ReceiptInspector.swift
```diff
@@ -73,8 +73,12 @@ struct ReceiptInspectorView: View {
     func inspectReceipt() async {
         do {
             guard !encodedReceipt.isEmpty else { return }
-            parsedReceipt = try PurchasesReceiptParser.default.parse(base64String: encodedReceipt).debugDescription
-            verifyReceiptResult = await ReceiptVerifier().verifyReceipt(base64Encoded: encodedReceipt,
+            // in kibana, receipts get encoded with extra `\`s
+            let receiptWithoutForwardSlashes = encodedReceipt.replacingOccurrences(of: "\\", with: "")
+            // just in case you accidentally copied with extra double-quotations
+            let receiptWithoutQuotations = receiptWithoutForwardSlashes.replacingOccurrences(of: "\"", with: "")
```
tonidero, line 79: FFTI, should we only remove start/end characters instead of all occurences? I think this should also work though, but in order to minify possible changes.

tonidero (approved):

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/ReceiptInspector.swift
aboedo: I thought about it, but I don't think we stand to gain much? 

Like, the receipt won't work locally unless it's a valid receipt (save for the double-quotes), and it won't validate against `/verifyReceipt` unless the contents are unaltered, so it's pretty safe to just assume you can remove them all in the easiest way possible

tonidero: Yeah... I guess if it ever could be a valid character it would be a problem, but I don't see that happening. It's fine as it is 👍

NachoSoto:
Rebased.

PR #3555, opened 2024-01-08
Set StoreKit 2 appAccountToken if our appUserID is an UUID

MarkVillacampa:
We already set this for SK1 purchases:

https://github.com/RevenueCat/purchases-ios/blob/daa0e1f3d0b8eaf653d81ef99396caa2a5e95506/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift#L385

However, as explained in the documentation, if this value is not a UUID, it basically gets ignored:

https://developer.apple.com/documentation/storekit/skmutablepayment/1506088-applicationusername

This PR makes sure we do the equivalent in SK2 purchases.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -465,6 +465,10 @@ final class PurchasesOrchestrator {
                 .simulatesAskToBuyInSandbox(Purchases.simulatesAskToBuyInSandbox)
             ]
 
+            if let uuid = UUID(uuidString: self.appUserID) {
+                options.insert(.appAccountToken(uuid))
+            }
+
             if let signedData = promotionalOffer {
                 Logger.debug(
                     Strings.storeKit.sk2_purchasing_added_promotional_offer_option(signedData.identifier)
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -466,6 +466,9 @@ final class PurchasesOrchestrator {
             ]
 
             if let uuid = UUID(uuidString: self.appUserID) {
+                Logger.debug(
+                    Strings.storeKit.sk2_purchasing_added_uuid_option(uuid)
+                )
                 options.insert(.appAccountToken(uuid))
             }
 
--- a/Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
+++ b/Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
@@ -1137,6 +1137,7 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
         try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
 
         self.customerInfoManager.stubbedCustomerInfoResult = .success(self.mockCustomerInfo)
+        self.mockStoreKit2TransactionListener?.mockResult = .init(.userCancelled)
 
         let product = try await self.fetchSk2Product()
 
--- a/Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
+++ b/Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
@@ -1173,6 +1173,39 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
     }
 
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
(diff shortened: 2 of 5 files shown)


Tests/UnitTests/Mocks/MockStoreKit2TransactionListener.swift
```diff
@@ -62,10 +62,17 @@ final class MockStoreKit2TransactionListener: StoreKit2TransactionListenerType {
...
+
+        if let mockTransaction = self.mockTransaction.value {
+            transaction = StoreTransaction(sk2Transaction: mockTransaction,
+                                           jwsRepresentation: self.mockJWSToken,
+                                           environmentOverride: self.mockEnvironment)
+        } else {
+            let result = purchaseResult.verificationResult!
+            transaction = StoreTransaction(sk2Transaction: result.underlyingTransaction,
+                                           jwsRepresentation: result.jwsRepresentation,
+                                           environmentOverride: self.mockEnvironment)
```
MarkVillacampa, line 75: In order to properly test this, we need to return the actual transaction made by the orchestrator, instead of providing a mock.

It cannot be tested with `invokedHandleParameters` because, if we have set a mock transaction using `simulateAnyTransaction`, the orchestrator would return the same mocked transaction:

image

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -465,6 +465,10 @@ final class PurchasesOrchestrator {
                 .simulatesAskToBuyInSandbox(Purchases.simulatesAskToBuyInSandbox)
             ]
 
+            if let uuid = UUID(uuidString: self.appUserID) {
```
NachoSoto, line 468: Let's add a debug log like we o below for the promotional offer option?

Tests/UnitTests/Mocks/MockStoreKit2TransactionListener.swift
```diff
@@ -62,10 +62,17 @@ final class MockStoreKit2TransactionListener: StoreKit2TransactionListenerType {
         self.invokedHandleParameters = (.init(purchaseResult), ())
         self.invokedHandleParametersList.append((.init(purchaseResult), ()))
 
-        let transaction: StoreTransaction? = self.mockTransaction.value.map {
-            StoreTransaction(sk2Transaction: $0,
-                             jwsRepresentation: self.mockJWSToken,
-                             environmentOverride: self.mockEnvironment)
+        var transaction: StoreTransaction
```
NachoSoto, line 65: This can be `let`.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -465,6 +465,10 @@ final class PurchasesOrchestrator {
...
+                options.insert(.appAccountToken(uuid))
```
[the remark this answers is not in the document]
NachoSoto: So I think we should make sure that we have prom offer purchases (SK1 and SK2) with user IDs that are and aren't UUIDs to make sure both work.

Tests/UnitTests/Mocks/MockStoreKit2TransactionListener.swift
```diff
@@ -62,10 +62,17 @@ final class MockStoreKit2TransactionListener: StoreKit2TransactionListenerType {
...
+
+        if let mockTransaction = self.mockTransaction.value {
+            transaction = StoreTransaction(sk2Transaction: mockTransaction,
+                                           jwsRepresentation: self.mockJWSToken,
+                                           environmentOverride: self.mockEnvironment)
+        } else {
+            let result = purchaseResult.verificationResult!
```
NachoSoto, line 72: This can crash the tests. Let's use this instead:
```suggestion
            let result = try XCTUnwrap(purchaseResult.verificationResult)
```

NachoSoto (changes requested):
Marking this as reviewed since there's still a few pending changes.

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -1137,6 +1137,7 @@ class PurchasesOrchestratorTests: StoreKitConfigTestCase {
         try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
 
         self.customerInfoManager.stubbedCustomerInfoResult = .success(self.mockCustomerInfo)
+        self.mockStoreKit2TransactionListener?.mockResult = .init(.userCancelled)
```
MarkVillacampa, line 1140: This test was failing after the changes to `MockStoreKit2TransactionListener` because it was returning the purchased transaction.

NachoSoto (approved):
👏🏻

PR #3563, opened 2024-01-09
`Paywalls`: new `PaywallViewControllerDelegate.paywallViewController(_:didChangeSizeTo:)`

vegaro:
- Add `paywallViewControlleSizeDidChange` to notify of size changes
- Added back top anchor to footers for auto layout to size correctly

### DIFF
--- a/RevenueCatUI/UIKit/PaywallViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallViewController.swift
@@ -74,7 +74,7 @@ public class PaywallViewController: UIViewController {
             }
             .onSizeChange { [weak self] in
                 guard let self = self else { return }
-                self.delegate?.paywallViewController?(self, didChangeSize: $0)
+                self.delegate?.paywallViewController?(self, didChangeSizeTo: $0)
             }
 
         return .init(rootView: view)
@@ -138,9 +138,9 @@ public protocol PaywallViewControllerDelegate: AnyObject {
     optional func paywallViewControllerWasDismissed(_ controller: PaywallViewController)
 
     /// For internal use only.
-    @objc(paywallViewControlle:didChangeSize:)
+    @objc(paywallViewControlle:didChangeSizeTo:)
     optional func paywallViewController(_ controller: PaywallViewController,
-                                        didChangeSize size: CGSize)
+                                        didChangeSizeTo size: CGSize)
 
 }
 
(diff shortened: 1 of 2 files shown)


RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -139,6 +136,9 @@ public protocol PaywallViewControllerDelegate: AnyObject {
     @objc(paywallViewControllerWasDismissed:)
     optional func paywallViewControllerWasDismissed(_ controller: PaywallViewController)
 
+    @objc(paywallViewControlleSizeDidChange:)
+    optional func paywallViewControlleSizeDidChange(_ size: CGSize)
```
NachoSoto, line 140: Let's add a doc to this and the sender controller like the others have:
```suggestion
    /// For internal use only.
    @objc(paywallViewControlle:didChangeSize:)
    optional func paywallViewController(_controller: PaywallViewController, didChangeSize size: CGSize)
```

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -72,6 +72,9 @@ public class PaywallViewController: UIViewController {
                 guard let self = self else { return }
                 self.delegate?.paywallViewController?(self, didFinishRestoringWith: customerInfo)
             }
+            .onSizeChange { [weak self] in
+                self?.delegate?.paywallViewControlleSizeDidChange?($0)
+            }
```
NachoSoto, line 77: Then this becomes:
```suggestion
            .onSizeChange { [weak self] in
                guard let self = self else { return }
                self.delegate?.paywallViewController(self, didChangeSize: $0)
            }
```

NachoSoto (commented):
🎉 

Just a couple of comments from our initial prototype implementations.
Also this needs to go into the API tester: https://github.com/RevenueCat/purchases-ios/blob/main/Tests/APITesters/RevenueCatUIAPITester/SwiftAPITester/PaywallViewControllerAPI.swift#L31

NachoSoto:
Changed the label to `refactor` since this is meant for internal use only.

tonidero (commented):
Looks great! After Nacho's comments are fixed will rereview.

RevenueCatUI/UIKit/PaywallViewController.swift
NachoSoto: Maybe `didChangeSizeTo`?

NachoSoto (approved):

PR #3567, opened 2024-01-10
`PaywallViewController`: add `PaywallFontProvider` parameter

NachoSoto:
This allows creating a paywall through the UIKit API and provide a custom font provider.

### DIFF
--- a/RevenueCatUI/UIKit/PaywallFooterViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallFooterViewController.swift
@@ -48,7 +48,7 @@ public final class PaywallFooterViewController: PaywallViewController {
         displayCloseButton: Bool = false
     ) {
         super.init(offering: offering,
-                   fonts: DefaultPaywallFontProvider(),
+                   fonts: fonts,
                    displayCloseButton: false)
     }
 
--- a/RevenueCatUI/UIKit/PaywallFooterViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallFooterViewController.swift
@@ -36,15 +36,20 @@ public final class PaywallFooterViewController: PaywallViewController {
     /// `Offerings.current` will be used by default.
     @objc
     public init(offering: Offering? = nil) {
-        super.init(offering: offering, displayCloseButton: false)
+        super.init(offering: offering,
+                   fonts: DefaultPaywallFontProvider(),
+                   displayCloseButton: false)
     }
 
     @available(*, unavailable)
     override init(
         offering: Offering? = nil,
+        fonts: PaywallFontProvider,
         displayCloseButton: Bool = false
     ) {
-        super.init(offering: offering, displayCloseButton: false)
+        super.init(offering: offering,
+                   fonts: DefaultPaywallFontProvider(),
+                   displayCloseButton: false)
     }
 
     // swiftlint:disable:next missing_docs
--- a/RevenueCatUI/UIKit/PaywallViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallViewController.swift
@@ -32,6 +32,7 @@ public class PaywallViewController: UIViewController {
     }
 
     private let offering: Offering?
+    private let fonts: PaywallFontProvider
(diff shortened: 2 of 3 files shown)


RevenueCatUI/UIKit/PaywallFooterViewController.swift
```diff
@@ -36,15 +36,20 @@ public final class PaywallFooterViewController: PaywallViewController {
...
 
     @available(*, unavailable)
     override init(
         offering: Offering? = nil,
+        fonts: PaywallFontProvider,
         displayCloseButton: Bool = false
     ) {
-        super.init(offering: offering, displayCloseButton: false)
+        super.init(offering: offering,
+                   fonts: DefaultPaywallFontProvider(),
```
tonidero, line 51: I think this should be 
```suggestion
                   fonts: fonts,
```

RevenueCatUI/UIKit/PaywallFooterViewController.swift
```diff
@@ -36,15 +36,20 @@ public final class PaywallFooterViewController: PaywallViewController {
...
+                   displayCloseButton: false)
     }
 
     @available(*, unavailable)
     override init(
         offering: Offering? = nil,
+        fonts: PaywallFontProvider,
```
tonidero, line 47: IMO It's weird to have a parameter without a default value after one with one... should it be the first parameter? Or should we give it a default value `DefaultPaywallFontProvider`?

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -32,18 +32,37 @@ public class PaywallViewController: UIViewController {
...
+    }
+
+    /// Initialize a `PaywallViewController` with an optional `Offering` and ``PaywallFontProvider``.
+    /// - Parameter offering: The `Offering` containing the desired `PaywallData` to display.
+    /// `Offerings.current` will be used by default.
+    /// - Parameter fonts: An optional ``PaywallFontProvider``.
+    /// - Parameter displayCloseButton: Set this to `true` to automatically include a close button.
     public init(
         offering: Offering? = nil,
+        fonts: PaywallFontProvider,
```
tonidero, line 61: Same here about a non-optional parameter after an optional one. We should also make sure these work well in objc I think.

tonidero (approved):
Couple of comments but looking good

RevenueCatUI/UIKit/PaywallFooterViewController.swift
NachoSoto: I know it's unusual in other languages, but argument labels are required in swift. If you look at the API tester, the different combinations all make sense.

I ordered it this way to be consistent with the `PaywallView` argument order. It also can't have a default value here because calling a constructor with no arguments would be ambiguous.

RevenueCatUI/UIKit/PaywallFooterViewController.swift
NachoSoto: OMG thanks.

RevenueCatUI/UIKit/PaywallViewController.swift
NachoSoto: This is Swift only because `PaywallFontProvider` can't be represented in Obj-C.

RevenueCatUI/UIKit/PaywallFooterViewController.swift
NachoSoto: Fixed. Though it wouldn't have mattered cause this method is unavailable.

RevenueCatUI/UIKit/PaywallViewController.swift
tonidero: Ohh ok... yeah we probably need to figure out a way to do this later in the hybrids anyway, so this works 👍

NachoSoto: Yeah we'll need to wrap the type somehow. The font parameter is a SwiftUI type.

PR #3570, opened 2024-01-11
`PaywallFooterViewController`: ensure that safe area inset is set on `PaywallView`

NachoSoto:
Turns out that `UIHostingController` ignores the safe area, so SwiftUI thinks there is no safe area.

Alternative fix in case https://github.com/RevenueCat/react-native-purchases/pull/834 doesn't work


vegaro:
Screenshot 2024-01-12 at 12 33 26

Tried this fix but it doesn't fix it

NachoSoto:
Closing since the other fix is more correct 🎉

PR #3580, opened 2024-01-17
[SK2] Enable Observer Mode for StoreKit 2

MarkVillacampa:


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1101,6 +1101,22 @@ public extension Purchases {
 
 #endif
 
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+    func handleObserverModeTransaction(
+        purchaseResult: StoreKit.Product.PurchaseResult
+    ) async throws -> StoreTransaction? {
+        guard !self.systemInfo.observerMode else {
+            throw NewErrorUtils.observerModeNotEnabledError()
```
MarkVillacampa, line 1109: Throw early, throw often. 😄

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1101,6 +1101,22 @@ public extension Purchases {
...
+        }
+        let (userCancelled, transaction) = try await self.purchasesOrchestrator.storeKit2TransactionListener.handle(
+            purchaseResult: purchaseResult, fromTransactionUpdate: true)
+
+        if userCancelled, self.systemInfo.dangerousSettings.customEntitlementComputation {
```
MarkVillacampa, line 1114: I'm not entirely sure about the `customEntitlementComputation` check here. This snippet was taken from `PurchasesOrchestrator`:

https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift#L501-L502

MarkVillacampa: Removed it for now

Sources/Purchasing/Purchases/Purchases.swift
NachoSoto: I would prevent creating yet another error for this. That's yet another error that users need to handle not just on this API, but in **all** of our APIs, and maybe localize to different languages, etc. 
I would instead throw `unsupportedError` or `configurationError`.

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -1008,6 +1009,35 @@ public protocol PurchasesSwiftType: AnyObject {
...
+     * - Throws: An error of type ``ErrorCode`` is thrown if a failure occurs while handling the purchase.
+     *
+     * - Returns: A ``StoreTransaction`` if there was a transacton found and handled for the provided product ID.
+     *
+     * - Important: This should only be used if you have enabled observer mode during SDK configuration using 
+     * ``Configuration/Builder/with(observerMode:)``
+     */
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+    func handleObserverModeTransaction(
+        purchaseResult: StoreKit.Product.PurchaseResult
```
NachoSoto, line 1038: I think the swifty way would be
```suggestion
        _ purchaseResult: StoreKit.Product.PurchaseResult
```

Tests/BackendIntegrationTests/StoreKitObserverModeIntegrationTests.swift
```diff
@@ -81,6 +81,33 @@ class StoreKit2ObserverModeIntegrationTests: StoreKit1ObserverModeIntegrationTes
 
 }
 
+class StoreKit2ObserverModeDisabledIntegrationTests: StoreKit1ObserverModeIntegrationTests {
```
NachoSoto, line 84: This is going to inherit and run every single test in the super class.

MarkVillacampa: Ugh. This can be tested as unit in `StoreKitTests` (because we need a `Transaction` object) but we don't have any test that instantiates a Purchases object. Is it worth it creating one?

NachoSoto: > but we don't have any test that instantiates a Purchases object. Is it worth it creating one?

Yeah I think it would be useful, but also mock all the dependencies. So I think we can extract the `Purchases` setup from `BasePurchasesTest` for that.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1101,6 +1101,18 @@ public extension Purchases {
 
 #endif
 
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+    func handleObserverModeTransaction(
+        purchaseResult: StoreKit.Product.PurchaseResult
+    ) async throws -> StoreTransaction? {
```
joshdholtz, line 1107: I know this is Observer Mode but does it matter what they have set for the StoreKit versioN? Like... should this also require `.with(storeKitVersion: .storeKit2)`?

Tests/APITesters/SwiftAPITester/SwiftAPITester/PurchasesAPI.swift
```diff
@@ -263,6 +263,11 @@ private func checkAsyncMethods(purchases: Purchases) async {
         let _: CustomerInfo = try await purchases.restorePurchases()
         let _: CustomerInfo = try await purchases.syncPurchases()
 
+        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *) {
+            let result = try await StoreKit.Product.products(for: [""]).first!.purchase()
```
NachoSoto, line 267: I like that you did it this way to ensure that whatever we pass to the next method is the result of this.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1101,6 +1101,18 @@ public extension Purchases {
...
+        guard self.systemInfo.observerMode else {
+            throw NewErrorUtils.observerModeNotEnabledError()
+        }
+        let (_, transaction) = try await self.purchasesOrchestrator.storeKit2TransactionListener.handle(
+            purchaseResult: purchaseResult, fromTransactionUpdate: true)
```
NachoSoto, line 1112: I wish we linted this stuff 😇 
I'd format it with symmetrical parenthesis:
```suggestion
        let (_, transaction) = try await self.purchasesOrchestrator.storeKit2TransactionListener.handle(
            purchaseResult: purchaseResult, fromTransactionUpdate: true
        )
```

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -1008,6 +1009,35 @@ public protocol PurchasesSwiftType: AnyObject {
...
+     * You only need to use this method with *new* purchases. Subscription updates are observed automatically.
+     *
+     * #### Example:
+     *
+     * ```swift
+     * let result = try await product.purchase(options: options)
+     * Purchases.shared.handleObserverModeTransaction(purchaseResult: result)
+     * ```
+     *
+     * - Note: You need to finish the transaction yourself after calling this method.
```
NachoSoto, line 1025: Maybe turn this into a `Warning` since it's pretty important? Maybe put that in the code example above too?

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -1008,6 +1009,35 @@ public protocol PurchasesSwiftType: AnyObject {
...
+     *
+     * - Parameter purchaseResult: The ``StoreKit.Product.PurchaseResult`` of the product that was just purchased.
```
NachoSoto, line 1027: I don't think you can reference types outside the module with "``", does this not give a warning when building?

Tests/UnitTests/Purchasing/ConfigurationTests.swift
```diff
@@ -31,50 +31,6 @@ class ConfigurationTests: TestCase {
...
-    func testObserverModeWithStoreKit2() {
-        let configuration = Configuration.Builder(withAPIKey: "test")
-            .with(observerMode: true)
-            .with(storeKitVersion: .storeKit2)
-            .build()
-
-        expect(configuration.observerMode) == true
-        expect(configuration.storeKitVersion) == .storeKit2
-
-        self.logger.verifyMessageWasLogged(Strings.configure.observer_mode_with_storekit2,
```
NachoSoto, line 74: Maybe instead we could log a different message now pointing out that transactions need to be handled manually?

Sources/Logging/Strings/ConfigureStrings.swift
```diff
@@ -187,6 +185,8 @@ extension ConfigureStrings: LogMessage {
             return "Purchases is not configured with StoreKit 2 enabled. This is required in order to detect " +
             "transactions coming from SwiftUI paywalls. You must use `.with(storeKitVersion: .storeKit2)` " +
             "when configuring the SDK."
+        case .sk2_required_observer_mode:
+            return "Observer mode must be enabled. You must use `.with(observerMode: true)` when configuring the SDK."
```
NachoSoto, line 189: Maybe: 
```suggestion
            return "Attempted to manually handle transactions with observer mode not enabled. You must use `.with(observerMode: true)` when configuring the SDK."
```

Sources/Purchasing/Purchases/Purchases.swift
MarkVillacampa: Good catch. Indeed that's a requirement.

Tests/BackendIntegrationTests/StoreKitObserverModeIntegrationTests.swift
```diff
@@ -81,6 +81,33 @@ class StoreKit2ObserverModeIntegrationTests: StoreKit1ObserverModeIntegrationTes
...
+    override class var observerMode: Bool { false }
+
+    override func setUp() async throws {
+        try await super.setUp()
+
+        try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
+    }
+
+    @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+    func testObservingTransactionThrowsIfObserverModeNotEnabled() async throws {
```
NachoSoto, line 97: We'll also need an integration test to verify that the transaction is posted, etc.

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -1008,6 +1009,35 @@ public protocol PurchasesSwiftType: AnyObject {
...
+     * ```swift
+     * let result = try await product.purchase(options: options)
+     * Purchases.shared.handleObserverModeTransaction(result)
+     * ```
+     *
+     * - Note: You need to finish the transaction yourself after calling this method.
+     *
+     * - Parameter purchaseResult: The ``StoreKit.Product.PurchaseResult`` of the product that was just purchased.
+     *
+     * - Throws: An error of type ``ErrorCode`` is thrown if a failure occurs while handling the purchase.
```
NachoSoto, line 1029: This isn't guaranteed because Swift doesn't have typed throws yet.
See `PurchasesError.swift`, I've done a lot of work to make sure that we always ensure this.

You need to do this:
```swift
do {
    try purchasesOrchestratorMethod()
} catch {
  throw ErrorUtils.purchasesError(withUntypedError: self).asPublicError
}
```

Sources/Purchasing/Purchases/PurchasesType.swift
MarkVillacampa: I haven't noticed any warning, but replaced with single ticks since it cannot link anywhere ayway.

NachoSoto: Screenshot 2024-01-17 at 12 18 25

Tests/BackendIntegrationTests/StoreKitObserverModeIntegrationTests.swift
MarkVillacampa: true! I moved it to `StoreKit2IntegrationTests` and created `StoreKit2NotEnabledObserverModeIntegrationTests` for the sk2 not enabled one.

Tests/BackendIntegrationTests/StoreKitObserverModeIntegrationTests.swift
MarkVillacampa: I think that should be covered by `testObservingTransactionUnlocksEntitlement`?

Sources/Logging/Strings/ConfigureStrings.swift
```diff
@@ -187,6 +189,18 @@ extension ConfigureStrings: LogMessage {
             return "Purchases is not configured with StoreKit 2 enabled. This is required in order to detect " +
             "transactions coming from SwiftUI paywalls. You must use `.with(storeKitVersion: .storeKit2)` " +
             "when configuring the SDK."
+
+        case .observer_mode_required:
```
NachoSoto, line 193: This is specific to handling transactions, maybe:
```suggestion
        case .handle_transaction_observer_mode_required:
```

Sources/Logging/Strings/ConfigureStrings.swift
```diff
@@ -187,6 +189,18 @@ extension ConfigureStrings: LogMessage {
...
+            return "Attempted to manually handle transactions with observer mode not enabled. " +
+            "You must use `.with(observerMode: true)` when configuring the SDK."
+
+        case .sk2_required:
+            return "StoreKit 2 must be enabled. You must use `.with(storeKitVersion: .storeKit2)` " +
+            "when configuring the SDK."
+
+        case .observer_mode_with_storekit2:
```
NachoSoto, line 201: :+1

Sources/Logging/Strings/PurchaseStrings.swift
```diff
@@ -86,6 +86,7 @@ enum PurchaseStrings {
     case caching_presented_offering_identifier(offeringID: String, productID: String)
     case payment_queue_wrapper_delegate_call_sk1_enabled
     case restorepurchases_called_with_allow_sharing_appstore_account_false
+    case sk2_observer_mode_error_processing_transaction(error: Error)
```
NachoSoto, line 89: Nit: the `error` label isn't required since it doesn't add anything to the type:
```suggestion
    case sk2_observer_mode_error_processing_transaction(Error)
```

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -1008,6 +1009,56 @@ public protocol PurchasesSwiftType: AnyObject {
...
+     * You only need to use this method with *new* purchases. Subscription updates are observed automatically.
+     *
+     * #### Example:
+     *
+     * ```swift
+     * // Fetch and purchase the product
+     * let product = try await StoreKit.Product.products(for: ["my_product_id"]).first
+     * guard let product = product else { return }
+     * let result = try await product.purchase()
+     * // Let RevenueCat obserbe the transaction result
```
NachoSoto, line 1025:
```suggestion
     * // Let RevenueCat observe the transaction result
```

NachoSoto: Or maybe
```suggestion
     * // Let RevenueCat handle the transaction result
```

Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
```diff
@@ -37,10 +37,11 @@ protocol StoreKit2TransactionListenerType: Sendable {
 
     func set(delegate: StoreKit2TransactionListenerDelegate) async
 
-    /// - Returns: `nil` `CustomerInfo` if purchases were not synced
-    /// - Throws: Error if purchase was not completed successfully
+    /// - Throws: ``ErrorCode`` if purchase was not completed successfully
```
NachoSoto, line 40: Technically these throw `PurchasesError` (for the most part, since it's impossible to guarantee that there aren't any other types).

Tests/BackendIntegrationTests/StoreKitObserverModeIntegrationTests.swift
```diff
@@ -45,13 +45,13 @@ class StoreKit2ObserverModeIntegrationTests: StoreKit1ObserverModeIntegrationTes
...
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
-    func testPurchaseInDevicePostsReceipt() async throws {
+    func testObservingTransactionUnlocksEntitlement() async throws {
         let result = try await self.manager.purchaseProductFromStoreKit2()
         let transaction = try XCTUnwrap(result.verificationResult?.underlyingTransaction)
 
         try self.testSession.disableAutoRenewForTransaction(identifier: UInt(transaction.id))
 
-        XCTExpectFailure("This test currently does not pass (see FB12231111)")
+        _ = try await Purchases.shared.handleObserverModeTransaction(result)
```
NachoSoto, line 54: OMG OMG I missed this! So excited :D 

I think we can simplify the test though. Since this is asynchronous, the check below can be done instantly, right?

NachoSoto:
Can you make sure that the new Integration Tests are added to the appropriate test plans? (`other`/`SK2`/etc)

Sources/Purchasing/Purchases/PurchasesType.swift
MarkVillacampa: Clearly I can't write "observe" 😝

Tests/UnitTests/Purchasing/ConfigurationTests.swift
```diff
@@ -100,4 +56,16 @@ class ConfigurationTests: TestCase {
...
 
+    func testObserverModeWithStoreKit2() {
+        let configuration = Configuration.Builder(withAPIKey: "test")
+            .with(observerMode: true)
+            .with(storeKitVersion: .storeKit2)
+            .build()
+
+        expect(configuration.observerMode) == true
+        expect(configuration.storeKitVersion) == .storeKit2
+
```
NachoSoto, line 67: Nit: extra space

Tests/UnitTests/Purchasing/ConfigurationTests.swift
```diff
@@ -31,50 +31,6 @@ class ConfigurationTests: TestCase {
         expect(Configuration.validate(apiKey: "swRTCezdEzjnJSxdexDNJfcfiFrMXwqZ")) == .legacy
     }
 
-    func testNoObserverModeWithStoreKit1() {
```
NachoSoto, line 34: You brought back `observer_mode_with_storekit2`. Can you bring back these tests instead? I think you can just reset this file, right?

NachoSoto (commented):
Looks good, just one more question about the removed tests.

PR #3140, opened 2023-09-05
`CircleCI`: change all jobs to M1

NachoSoto:
`CircleCI` is deprecating Intel machines, so we need to get this working: https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718

~~Deadline for this is January 31 2024.~~ New deadline is 28 June 2024.

Note that Codecov isn't compatible ([yet?](https://github.com/codecov/feedback/issues/51)) so I'm turning it off for now.

PR #3582, opened 2024-01-17
`CI`: remove `.framework` archives from releases

NachoSoto:
Depends on #3140.

[`Carthage` cannot build fat `.framework`s on Apple Silicon](https://github.com/Carthage/Carthage?tab=readme-ov-file#migrating-a-project-from-framework-bundles-to-xcframeworks):
> require `XCFrameworks` when building on an Apple Silicon Mac

It fails with this error:
> RevenueCat have the same architectures (arm64) and can't be in the same fat output file

To address that, we're planning to remove these on the next major release.


fastlane/Fastfile
```diff
@@ -86,7 +86,7 @@ platform :ios do
       repo_name: repo_name,
       github_api_token: ENV["GITHUB_TOKEN"],
       changelog_latest_path: changelog_latest_path,
-      upload_assets: ['RevenueCat.framework.zip', 'RevenueCat.xcframework.zip']
```
tonidero, line 89: Not sure if a lot of people are using those but  just in case we should mention this in the changelog once we release this

tonidero (approved):

fastlane/Fastfile
NachoSoto: Yup: 
> remove `.framework` archives from releases

Also this is `breaking` so it'll be highlighted as a breaking change.

MarkVillacampa (approved):

NachoSoto:
@MarkVillacampa feel free to merge this too

PR #3584, opened 2024-01-17
`Integration Tests`: improve flaky tests

NachoSoto:
Fixes https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/17208/workflows/729d01a6-d644-446c-8514-80b54177ff43/jobs/162052/tests
With #3140 and faster executors in CI, it's become clear that this test is very flaky because it relies on requests not finishing quickly enough.

For the first test, the new implementation ensures that all `customerInfo` calls happen in parallel.
For the second test, it's important that if the first `CustomerInfo` request had finished, we don't make a second one, which would make the test fail.


NachoSoto:
Included in #3140

PR #3586, opened 2024-01-19
`Paywalls`: extracted `PaywallViewConfiguration`

NachoSoto:
The number of parameters was already very large. This refactor combines the different values into its own type.

This is also a preliminary refactor to be able to configure a paywall with an offering identifier instead of an offering, which will be used by the hybrid SDKs.

### DIFF
--- a/RevenueCatUI/Data/PaywallViewConfiguration.swift
+++ b/RevenueCatUI/Data/PaywallViewConfiguration.swift
@@ -1,0 +1,24 @@
+//
+//  PaywallViewConfiguration.swift
+//
+//
+//  Created by Nacho Soto on 1/19/24.
+//
+
+import Foundation
+
+import RevenueCat
+
+/// Parameters needed to configure a ``PaywallView``.
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+struct PaywallViewConfiguration {
+
+    var offering: Offering?
+    var customerInfo: CustomerInfo?
+    var mode: PaywallViewMode = .default
+    var fonts: PaywallFontProvider = DefaultPaywallFontProvider()
+    var displayCloseButton: Bool = false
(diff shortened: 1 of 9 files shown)


RevenueCatUI/Data/PaywallViewConfiguration.swift
```diff
@@ -0,0 +1,24 @@
...
+
+/// Parameters needed to configure a ``PaywallView``.
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+struct PaywallViewConfiguration {
+
+    var offering: Offering?
+    var customerInfo: CustomerInfo?
+    var mode: PaywallViewMode = .default
+    var fonts: PaywallFontProvider = DefaultPaywallFontProvider()
+    var displayCloseButton: Bool = false
```
NachoSoto, line 20: The default value for these is now in a single place instead of being duplicated in several places.

vegaro (approved):

RevenueCatUI/Data/PaywallViewConfiguration.swift
```diff
@@ -0,0 +1,24 @@
...
+//  Created by Nacho Soto on 1/19/24.
+//
+
+import Foundation
+
+import RevenueCat
+
+/// Parameters needed to configure a ``PaywallView``.
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+struct PaywallViewConfiguration {
```
tonidero, line 14: In android, this would be similar to `PaywallOptions`, which is public and it's how developers can customize most things for the paywall. We used the builder pattern there, but I'm not sure if that's so common in swift.

NachoSoto: Yeah we could think about exposing this (and only the public types). We can do that in a separate PR.

PR #3587, opened 2024-01-19
`PaywallViewController`: new initializer with `Offering` identifier

NachoSoto:
Depends on #3586.

This will be used by the hybrid SDKs to be able to load paywalls with an offering identifier.

### DIFF
--- a/RevenueCatUI/UIKit/PaywallViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallViewController.swift
@@ -27,13 +27,7 @@ public class PaywallViewController: UIViewController {
     /// See ``PaywallViewControllerDelegate`` for receiving purchase events.
     @objc public final weak var delegate: PaywallViewControllerDelegate?
 
-    var mode: PaywallViewMode {
-        return .fullScreen
-    }
-
-    private let offering: Offering?
-    private let fonts: PaywallFontProvider
-    private let displayCloseButton: Bool
+    private let configuration: PaywallViewConfiguration
 
     /// Initialize a `PaywallViewController` with an optional `Offering`.
     /// - Parameter offering: The `Offering` containing the desired `PaywallData` to display.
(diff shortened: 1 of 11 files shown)


Sources/FoundationExtensions/Optional+Extensions.swift
```diff
@@ -44,14 +44,3 @@ extension OptionalType {
     }
 
 }
-
-// MARK: -
-
-internal extension Optional where Wrapped == String {
```
NachoSoto, line 50: Moved this to `String+Extensions` since it makes more sense there. This file is also now sym-linked so it can be used by `RevenueCatUI`.

aboedo (approved):

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -56,14 +50,45 @@ public class PaywallViewController: UIViewController {
     /// `Offerings.current` will be used by default.
     /// - Parameter fonts: An optional ``PaywallFontProvider``.
     /// - Parameter displayCloseButton: Set this to `true` to automatically include a close button.
-    public init(
+    public convenience init(
```
vegaro, line 53: just a question I have, there's only one `@objc` annotated method in this file, but we don't use it in phc, because it's written in Swift... is it exposed for folks that are using objective c? If so, should we expose another constructor?

vegaro: Also in the docs of this constructor it says "an optional PaywallFontProvider", but it's not optional

vegaro (approved):

RevenueCatUI/UIKit/PaywallViewController.swift
NachoSoto: > is it exposed for folks that are using objective c?

Yeah that's correct, because `PaywallFontProvider` isn't available in Swift.
I'll add another one with offering identifier for ObjC

NachoSoto: And fixed the docs, thanks!

NachoSoto: Actually I think I'm not going to do that. The ideal way is passing an `Offering` to minimize the chances that you'll pass something that won't be found. The `offeringIdentifier` one is just for us for the hybrids. Other users should pass `Offering`s

PR #3588, opened 2024-01-19
Add `getCurrentOffering(forPlacement: String)` to `Offerings`

guido732:
WIP

### Checklist


Sources/Purchasing/Offerings.swift
```diff
@@ -93,4 +96,12 @@ public extension Offerings {
         return description
     }
 
+    /**
+     Retrieves a current offering for a placement identifier, use this to access offerings defined by targeting
+     placements configured in the RevenueCat dashboard, e.g. `offerings.getCurrentOffering(for: "placement_id")`.
+     */
+    @objc(getCurrentOfferingForPlacement:)
+    func getCurrentOffering(for placement: String) -> Offering? {
```
NachoSoto, line 104: I know I suggested this, but I think this should be
```suggestion
    func getCurrentOffering(forPlacement placement: String) -> Offering? {
```

Since the type (`String`) doesn't provide enough context on what the parameter is.

guido732: sure!

PR #3592, opened 2024-01-22
`PaywallViewController`: methods for reconfiguring paywall with new offering

NachoSoto:
For use by `react-native-purchases`.

This is not the most efficient, but it's required because we don't get the props on React Native until after we've already returned the `UIView`.

### DIFF
--- a/RevenueCatUI/UIKit/PaywallViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallViewController.swift
@@ -210,8 +210,29 @@ public protocol PaywallViewControllerDelegate: AnyObject {
 private extension PaywallViewController {
 
     func createHostingController() -> UIHostingController<PaywallContainerView> {
-        let controller = UIHostingController(rootView: PaywallContainerView(configuration: self.configuration,
-                                                                            controller: self))
+        let container = PaywallContainerView(
+            configuration: self.configuration,
+            purchaseCompleted: { [weak self] transaction, customerInfo in
+                guard let self else { return }
+                self.delegate?.paywallViewController?(self, didFinishPurchasingWith: customerInfo)
+                self.delegate?.paywallViewController?(self,
+                                                      didFinishPurchasingWith: customerInfo,
+                                                      transaction: transaction)
+            },
+            purchaseCancelled: { [weak self] in
+                guard let self else { return }
+                self.delegate?.paywallViewControllerDidCancelPurchase?(self)},
+            restoreCompleted: { [weak self] customerInfo in
+                guard let self else { return }
+                self.delegate?.paywallViewController?(self, didFinishRestoringWith: customerInfo)
+            },
+            onSizeChange: { [weak self] in
+                guard let self else { return }
+                self.delegate?.paywallViewController?(self, didChangeSizeTo: $0)
+            }
+        )
+
+        let controller = UIHostingController(rootView: container)
 
         // make the background of the container clear so that if there are cutouts, they don't get
         // overridden by the hostingController's view's background.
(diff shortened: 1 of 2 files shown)


RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -196,4 +202,54 @@ public protocol PaywallViewControllerDelegate: AnyObject {
...
+        return controller
+    }
+
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
+private struct PaywallContainerView: View {
+
+    var configuration: PaywallViewConfiguration
+    var controller: PaywallViewController
```
NachoSoto, line 228: TODO: need to test to make sure this doesn't lead to a leak due to this retain cycle.

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -150,12 +115,53 @@ public class PaywallViewController: UIViewController {
...
 
+    /// - Warning: For internal use only
+    public func update(with offering: Offering) {
+        self.configuration.content = .offering(offering)
+    }
+
+    /// - Warning: For internal use only
+    public func update(with offeringIdentifier: String) {
+        self.configuration.content = .offeringIdentifier(offeringIdentifier)
+    }
```
NachoSoto, line 126: @vegaro this is the new API.

vegaro: Thanks. Couple things I've noticed while testing. This is how it gets called from objc:

```
- (void)setOptions:(NSDictionary *)options{
    NSDictionary *offering = options[@"offering"];
    if (offering && ![offering isKindOfClass:[NSNull class]]) {
        NSString *identifier = offering[@"identifier"];
        if (identifier) {
            [self.footerViewController updateWith:identifier];
        }
    }
}
```

It needs `@objc`. And we might want to update the name at least in objc because right now, if you try to add `@objc` to the one that accepts an `Offering` there's a name colision. There will probably be other parameters added soon.

vegaro: Another thing, when we create the paywall in full screen, we will also need to be able to update the offering. Right now `createPaywallView` in `PaywallProxy` returns a `UIViewController`, we might want to update it to return a `PaywallViewController` instead.

```
    @objc
    public func createPaywallView() -> UIViewController {
        let controller = PaywallViewController()
        controller.delegate = self
        return controller
    }
```

vegaro:
@NachoSoto this works fine, I made some comments regarding its usage, which you can see in
https://github.com/RevenueCat/react-native-purchases/pull/850/files#diff-009f073d996d3fadbe53dd880828b32ed98406e22be941947288a8d093e97459
and 
https://github.com/RevenueCat/react-native-purchases/pull/850/files#diff-4c098c9f4efa9edf7579aae6e47a211ce08715082ee3b1009876e90423346117

RevenueCatUI/UIKit/PaywallViewController.swift
vegaro: If we end up modifying `createPaywallView` we will need https://github.com/RevenueCat/react-native-purchases/pull/860

NachoSoto: I added the `@objc` annotations 👍🏻  

> Right now createPaywallView in PaywallProxy returns a UIViewController

Yeah let's change that to return `PaywallViewController`.

NachoSoto: I did that in https://github.com/RevenueCat/purchases-hybrid-common/pull/675

RevenueCatUI/UIKit/PaywallViewController.swift
NachoSoto: Tested with Cesar and we fixed it.

tonidero (approved):
Pretty complex stuff 😅, but since we tested it and it works, it looks good to me!

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -150,12 +115,55 @@ public class PaywallViewController: UIViewController {
...
 
+    /// - Warning: For internal use only
+    @objc(updateWithOffering:)
+    public func update(with offering: Offering) {
+        self.configuration.content = .offering(offering)
+    }
+
+    /// - Warning: For internal use only
+    @objc(updateWithOfferingIdentifier:)
+    public func update(with offeringIdentifier: String) {
```
aboedo, line 126: almost wonder whether we should prefix these with `_` or something, or name them something that you're less likely to want to use

aboedo: like, you might be inclined to use the `update(with:)` method if you want to swap the offering under the hood

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -150,12 +115,55 @@ public class PaywallViewController: UIViewController {
...
+        willSet {
+            guard let oldValue = self.hostingController else { return }
+
+            oldValue.willMove(toParent: nil)
+            oldValue.view.removeFromSuperview()
+            oldValue.removeFromParent()
+        }
+
+        didSet {
+            guard let newController = self.hostingController else { return }
```
aboedo, line 148: given that this is not the most simple or intuitive implementation, we should document _why_ we're doing this

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -196,4 +204,65 @@ public protocol PaywallViewControllerDelegate: AnyObject {
 
 }
 
+// MARK: - Private
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
+private extension PaywallViewController {
+
```
aboedo, line 211: we can do it as a separate step, but it is kind of a bummer that because react-native is a pain, all of our hybrids and our UIKit paywalls would suffer. 
might be worth just special-casing this one out

aboedo (approved):

RevenueCatUI/UIKit/PaywallViewController.swift
NachoSoto: `_` methods are reserved for Apple.

Ultimately there's nothing wrong with others using them though, there's just little reason to need them instead of creating a new `PaywallViewController`, so was just trying to disincentivize potential users of it.

RevenueCatUI/UIKit/PaywallViewController.swift
NachoSoto: 👍🏻

RevenueCatUI/UIKit/PaywallViewController.swift
NachoSoto: > all of our hybrids and our UIKit paywalls would suffer.

What makes you say that?
If you don't use the `update` methods, this is functionally equivalent to the previous implementation.

PR #3605, opened 2024-01-26
`Paywalls`: improve `PaywallData.config(for:)` disambiguation

NachoSoto:
This fixes an issue with `PaywallData` picking the wrong `Locale` when the configuration has multiple translations sharing the same language code.

The fix replaces comparing `Language.languageCode` with the smarter `Language.isEquivalent(to:)`.

### Other changes:
- Removed the duplicate `zh_TW` since it was the same as `zh_Hant` (we don't have that specific language in the dashboard either)
- Added configuration to be able to test this on `PaywallsTester`
- Renamed `zh_HK` to `zh_Hant` for consistency with `Hans`

Simulator Screenshot - iPhone 15 Pro Max - 2024-01-26 at 12 57 29
Simulator Screenshot - iPhone 15 Pro Max - 2024-01-26 at 11 17 14


Sources/Paywalls/PaywallData.swift
```diff
@@ -524,7 +524,7 @@ private extension Locale {
 
     func sharesLanguageCode(with other: Locale) -> Bool {
         if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
-            return self.language.languageCode == other.language.languageCode
+            return self.language.isEquivalent(to: other.language)
```
NachoSoto, line 527: This is the core of the fix.

tonidero (approved):

PR #3631, opened 2024-02-05
removed background color for visionOS

aboedo:
| Before | After |
| :-: | :-: |
| Screenshot 2024-02-05 at 12 29 45 PM | Screenshot 2024-02-05 at 12 27 25 PM |

### DIFF
--- a/RevenueCatUI/Templates/TemplateViewType.swift
+++ b/RevenueCatUI/Templates/TemplateViewType.swift
@@ -161,7 +161,11 @@ extension View {
 
     func adaptTemplateView(with configuration: TemplateViewConfiguration) -> some View {
         self
+        #if os(visionOS)
+            .background(Material.regular)
+        #else
             .background(configuration.backgroundView)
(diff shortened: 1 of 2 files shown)


RevenueCatUI/Templates/TemplateViewType.swift
```diff
@@ -161,7 +161,11 @@ extension View {
 
     func adaptTemplateView(with configuration: TemplateViewConfiguration) -> some View {
         self
+        #if os(visionOS)
+            .background(Material.regular)
```
NachoSoto, line 165: We might want to do this below on `backgroundContent` so we only use this if blurred background is enabled. Otherwise it'd ignore paywalls configured with a color and the foreground text might become ineligible.

NachoSoto:
Maybe we should use https://developer.apple.com/documentation/realitykit/model3dplaceholdercontent/glassbackgroundeffect(displaymode:)?

PR #3633, opened 2024-02-05
`Paywalls`: fix finding locales with different regions

NachoSoto:
This was a regression introduced by #3605.

### Example:

Consider a paywall configuration with these locales: `en_US`, `de_DE`.
Running on a device with locale `en_IN` for example.

Before #3605, `PaywallData.localizedConfiguration` would have returned the `en_US` localization. However, that change made it so that it could no longer find a matching localization with a different region.

This fix improves the locale lookup so that if it doesn't find a matching localization with both language AND region, it then tries to look it up with only the language, before falling back to the default localization.

### DIFF
--- a/Tests/UnitTests/Paywalls/PaywallDataTests.swift
+++ b/Tests/UnitTests/Paywalls/PaywallDataTests.swift
@@ -122,6 +122,25 @@ class PaywallDataTests: BaseHTTPResponseTest {
         expect(esConfig.title) == "Tienda"
     }
 
+    func testLocalizedConfigurationFallsBackToLanguageWithDifferentRegion() throws {
+        let paywall: PaywallData = try self.decodeFixture("PaywallData-Sample1")
+
+        let enConfig = try XCTUnwrap(paywall.localizedConfiguration(for: [
+            .init(identifier: "en_IN"),
+            .init(identifier: "en-IN"),
+            .init(identifier: "en-IN").removingRegion
+        ].compactMap { $0 }))
+        expect(enConfig.title) == "Paywall"
+    }
+
+    func testLocalesOrderedByPriority() {
+        expect(PaywallData.localesOrderedByPriority) == [
+            .init(identifier: "en_US"),
+            .init(identifier: "en-US"),
+            .init(identifier: "en")
+        ]
+    }
+
     func testDoesNotFindLocaleWithMissingLanguage() throws {
         let paywall: PaywallData = try self.decodeFixture("PaywallData-Sample1")
 
--- a/Tests/UnitTests/Paywalls/PaywallDataTests.swift
+++ b/Tests/UnitTests/Paywalls/PaywallDataTests.swift
@@ -134,10 +134,10 @@ class PaywallDataTests: BaseHTTPResponseTest {
     }
 
     func testLocalesOrderedByPriority() {
-        expect(PaywallData.localesOrderedByPriority) == [
-            .init(identifier: "en_US"),
-            .init(identifier: "en-US"),
-            .init(identifier: "en")
+        expect(PaywallData.localesOrderedByPriority.map(\.identifier)) == [
+            "en_US",
+            "en-US",
+            "en"
         ]
     }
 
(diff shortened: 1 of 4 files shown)


Tests/UnitTests/Paywalls/PaywallDataTests.swift
```diff
@@ -122,6 +122,25 @@ class PaywallDataTests: BaseHTTPResponseTest {
...
+
+        let enConfig = try XCTUnwrap(paywall.localizedConfiguration(for: [
+            .init(identifier: "en_IN"),
+            .init(identifier: "en-IN"),
+            .init(identifier: "en-IN").removingRegion
+        ].compactMap { $0 }))
+        expect(enConfig.title) == "Paywall"
+    }
+
+    func testLocalesOrderedByPriority() {
```
NachoSoto, line 136: Now we also have a test for this logic

Tests/UnitTests/Paywalls/PaywallDataTests.swift
```diff
@@ -122,6 +122,25 @@ class PaywallDataTests: BaseHTTPResponseTest {
         expect(esConfig.title) == "Tienda"
     }
 
+    func testLocalizedConfigurationFallsBackToLanguageWithDifferentRegion() throws {
+        let paywall: PaywallData = try self.decodeFixture("PaywallData-Sample1")
+
+        let enConfig = try XCTUnwrap(paywall.localizedConfiguration(for: [
```
NachoSoto, line 128: And also have a test for this method, which we didn't have before.

Tests/UnitTests/Paywalls/PaywallDataTests.swift
```diff
@@ -122,6 +122,25 @@ class PaywallDataTests: BaseHTTPResponseTest {
...
+            .init(identifier: "en_IN"),
+            .init(identifier: "en-IN"),
+            .init(identifier: "en-IN").removingRegion
```
NachoSoto, line 131: The fact that this is included in `Paywall.localizedConfiguration` is tested below.

Tests/UnitTests/Paywalls/PaywallDataTests.swift
```diff
@@ -122,6 +122,25 @@ class PaywallDataTests: BaseHTTPResponseTest {
...
+        ].compactMap { $0 }))
+        expect(enConfig.title) == "Paywall"
```
NachoSoto, line 133: This was a failing test before the fix.

tonidero (approved):
LGTM!

joshdholtz (approved):
Ship ship!

samkass:
After upgrading to this version, I still see this problem. RevenueCat only allows German localization for "German (Germany)" and my iOS application (and my iPhone's settings) is localized for "German". Even after 4.36.0 is applied, I am seeing the English Paywall when my phone is set to "German". When I set the phone to "German (Germany)" I see German. When I set it to "German (Switzerland)" I also see English. I sent an extensive steps to reproduce, expected and observed behavior, and test cases via email attached to my support ticket on this topic.

NachoSoto:
Sorry this is causing you trouble. I haven't been able to reproduce that exact setup, it uses the german localization for me with `de` and `de_DE`.

Do you mind filing a separate GitHub issue and providing more details to help us look into it? Specifically wondering what iOS version you're using.

I've also sent #3649 which adds some verbose logs to help us diagnose it. Would you be able to run the app using that branch? Make sure you set `Purchases.logLevel = .verbose`.

Thanks 👍🏻

samkass:
> Sorry this is causing you trouble. I haven't been able to reproduce that exact setup, it uses the german localization for me with `de` and `de_DE`.
> 
> Do you mind filing a separate GitHub issue and providing more details to help us look into it? Specifically wondering what iOS version you're using.
> 
> I've also sent #3649 which adds some verbose logs to help us diagnose it. Would you be able to run the app using that branch? Make sure you set `Purchases.logLevel = .verbose`.
> 
> Thanks 👍🏻

Added #3655. To your specific question, I'm using the latest iOS 17.2.1. In the Simulator, this happens on iOS 17.x and 16.x, but not 15.x.

PR #3632, opened 2024-02-05
`Paywalls`: new `presentationMode` parameter

Lascorbe:
### TLDR
- Added a way to present `PaywallView` either as sheet or full screen
- No breaking changes have been made

### Motivation
I wanted to easily present the PaywallView on full screen.

### Description
I added a new enum `PaywallPresentationMode` which currently has 2 options: `.sheet` and `.fullScreen `. I added this as a parameter to the current public `presentPaywallIfNeeded` modifiers. The default is `sheet` so there are no breaking changes with the current public API.

| Sheet             | Full Screen    |
| ------------- | ------------- |
| Simulator Screenshot - iPhone 15 Pro - 2024-02-05 at 18 08 51  |  Simulator Screenshot - iPhone 15 Pro - 2024-02-05 at 18 44 15 |

PD: I looked for info on how to contribute but it's unclear if I have to open an issue first or not (contributing says open issue, then when opening an issue it says please don't open but contact us at https://app.revenuecat.com/settings/support). I hope this is valid, otherwise let me know how I can proceed.

PR #3638, opened 2024-02-06
`Paywalls`: new `presentationMode` parameter (by @Lascorbe)

NachoSoto:
Original PR: #3632 by [Lascorbe](https://github.com/Lascorbe).

### TLDR
- Added a way to present `PaywallView` either as sheet or full screen
- No breaking changes have been made

### Motivation
I wanted to easily present the PaywallView on full screen.

### Description
I added a new enum `PaywallPresentationMode` which currently has 2 options: `.sheet` and `.fullScreen`. I added this as a parameter to the current public `.presentPaywallIfNeeded` modifiers. The default is `sheet` so there are no breaking changes with the current public API.

| Sheet             | Full Screen    |
| ------------- | ------------- |
| Simulator Screenshot - iPhone 15 Pro - 2024-02-05 at 18 08 51  |  Simulator Screenshot - iPhone 15 Pro - 2024-02-05 at 18 44 15 |

### DIFF
--- a/RevenueCatUI/View+PresentPaywall.swift
+++ b/RevenueCatUI/View+PresentPaywall.swift
@@ -16,6 +16,19 @@ import SwiftUI
 
 #if !os(macOS) && !os(tvOS)
 
+/// Presentation options to use with the [presentPaywallIfNeeded](x-source-tag://presentPaywallIfNeeded) View modifiers.
+///
+/// ### Related Articles
+/// [Documentation](https://rev.cat/paywalls)
+public enum PaywallPresentationMode {
+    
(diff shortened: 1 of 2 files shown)


tonidero (approved):
Oh nice addition!  I was thinking whether this could apply to Android, but we're already presenting it as full screen when using the `PaywallActivityLauncher`. With the `PaywallDialog` composable we are full screen on mobile, and modal on tablets, but people can always use the `Paywall` composable directly if they have to, so I think we should be fine there.

NachoSoto:
Yeah I think this is sort of specific to this platform where SwiftUI offers these 2 options.

PR #3641, opened 2024-02-06
[SK2] Set StoreKit version when enabling observer mode

MarkVillacampa:
Default to StoreKit 1 when enabling observer mode.

### DIFF
--- a/Sources/Purchasing/Configuration.swift
+++ b/Sources/Purchasing/Configuration.swift
@@ -132,15 +132,11 @@ import Foundation
          * - Parameter observerMode: Set this to `true` if you have your own IAP implementation and want to use only
          * RevenueCat's backend. Default is `false`.
          * - Parameter storeKitVersion: Set the StoreKit version you're using to make purchases.
-         * Default is `StoreKitVersion.storeKit1`.
          */
         @objc public func with(
             observerMode: Bool,
-            storeKitVersion: StoreKitVersion = .storeKit1
+            storeKitVersion: StoreKitVersion
         ) -> Configuration.Builder {
-            if self.storeKitVersion != StoreKitVersion.default {
-                Logger.warn(Strings.configure.observer_mode_with_storekit_version)
-            }
             self.observerMode = observerMode
             self.storeKitVersion = storeKitVersion
             return self
@@ -233,9 +229,6 @@ import Foundation
         /// ### Related Symbols
         /// - ``StoreKitVersion``
         @objc public func with(storeKitVersion version: StoreKitVersion) -> Builder {
-            if self.observerMode {
-                Logger.warn(Strings.configure.observer_mode_with_storekit_version)
-            }
             self.storeKitVersion = version
             return self
         }
--- a/Sources/Purchasing/Configuration.swift
+++ b/Sources/Purchasing/Configuration.swift
@@ -131,12 +131,18 @@ import Foundation
          * Set `observerMode`.
          * - Parameter observerMode: Set this to `true` if you have your own IAP implementation and want to use only
          * RevenueCat's backend. Default is `false`.
-         *
-         * - Warning: This assumes your IAP implementation uses StoreKit 1.
-         * Observer mode is not compatible with StoreKit 2.
+         * - Parameter storeKitVersion: Set the StoreKit version you're using to make purchases.
+         * Default is `StoreKitVersion.storeKit1`.
          */
-        @objc public func with(observerMode: Bool) -> Configuration.Builder {
+        @objc public func with(
+            observerMode: Bool,
+            storeKitVersion: StoreKitVersion = .storeKit1
+        ) -> Configuration.Builder {
+            if self.storeKitVersion != StoreKitVersion.default {
+                Logger.warn(Strings.configure.observer_mode_with_storekit_version)
+            }
             self.observerMode = observerMode
+            self.storeKitVersion = storeKitVersion
             return self
         }
         /**
@@ -227,6 +233,9 @@ import Foundation
         /// ### Related Symbols
         /// - ``StoreKitVersion``
         @objc public func with(storeKitVersion version: StoreKitVersion) -> Builder {
+            if self.observerMode {
+                Logger.warn(Strings.configure.observer_mode_with_storekit_version)
+            }
             self.storeKitVersion = version
             return self
         }
--- a/Tests/APITesters/ObjCAPITester/ObjCAPITester/RCConfigurationAPI.m
+++ b/Tests/APITesters/ObjCAPITester/ObjCAPITester/RCConfigurationAPI.m
@@ -14,7 +14,7 @@
 + (void)checkAPI {
     RCConfigurationBuilder *builder = [RCConfiguration builderWithAPIKey:@""];
     RCConfiguration *config __unused = [[[[[[[[[[[[builder withApiKey:@""]
-                                                 withObserverMode:false]
+                                                 withObserverMode:false storeKitVersion:RCStoreKitVersionStoreKit2]
                                                 withUserDefaults:NSUserDefaults.standardUserDefaults]
                                                withAppUserID:@""]
                                               withAppUserID:nil]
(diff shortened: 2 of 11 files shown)


Sources/Purchasing/Configuration.swift
```diff
@@ -227,6 +233,9 @@ import Foundation
         /// ### Related Symbols
         /// - ``StoreKitVersion``
         @objc public func with(storeKitVersion version: StoreKitVersion) -> Builder {
+            if self.observerMode {
+                Logger.warn(Strings.configure.observer_mode_with_storekit_version)
+            }
```
NachoSoto, line 238: This only works if you happen to set the version _after_ setting observer mode. Instead, let's do this check in `build()`.

Tests/APITesters/ObjCAPITester/ObjCAPITester/RCConfigurationAPI.m
```diff
@@ -14,7 +14,7 @@ @implementation RCConfigurationAPI
 + (void)checkAPI {
     RCConfigurationBuilder *builder = [RCConfiguration builderWithAPIKey:@""];
     RCConfiguration *config __unused = [[[[[[[[[[[[builder withApiKey:@""]
-                                                 withObserverMode:false]
+                                                 withObserverMode:false storeKitVersion: RCStoreKitVersionStoreKit2]
```
NachoSoto, line 17: Objective-C doesn't get a version with the default parameter by default unless we make it an override, but I guess this is fine?
```suggestion
                                                 withObserverMode:false storeKitVersion:RCStoreKitVersionStoreKit2]
```

Sources/Purchasing/Configuration.swift
```diff
@@ -131,12 +131,18 @@ import Foundation
          * Set `observerMode`.
          * - Parameter observerMode: Set this to `true` if you have your own IAP implementation and want to use only
          * RevenueCat's backend. Default is `false`.
-         *
-         * - Warning: This assumes your IAP implementation uses StoreKit 1.
-         * Observer mode is not compatible with StoreKit 2.
+         * - Parameter storeKitVersion: Set the StoreKit version you're using to make purchases.
+         * Default is `.storeKit1`.
```
NachoSoto, line 135: Maybe
```suggestion
         * Default is ``StoreKitVersion.storeKit1``.
```

Sources/Purchasing/Configuration.swift
```diff
@@ -131,12 +131,18 @@ import Foundation
...
          */
-        @objc public func with(observerMode: Bool) -> Configuration.Builder {
+        @objc public func with(
+            observerMode: Bool,
+            storeKitVersion: StoreKitVersion = .storeKit1
+        ) -> Configuration.Builder {
+            if self.storeKitVersion != StoreKitVersion.default {
+                Logger.warn(Strings.configure.observer_mode_with_storekit_version)
```
NachoSoto, line 142: Did you mean to use a different string here? This would tell you to use this method when using it with `.storeKit1` as well.

Tests/BackendIntegrationTests/Constants.swift
```diff
@@ -10,12 +10,12 @@ import Foundation
 
 enum Constants {
 
-    static let apiKey = "REVENUECAT_API_KEY"
-    static let loadShedderApiKey = "REVENUECAT_LOAD_SHEDDER_API_KEY"
-    static let customEntitlementComputationApiKey = "REVENUECAT_CUSTOM_ENTITLEMENT_COMPUTATION_API_KEY"
+    static let apiKey = "appl_gyJuOxwYgimOegOmmobUGlRoJGV"
```
NachoSoto: Oops, you might want to setup the pre-commit hook: https://github.com/RevenueCat/purchases-ios/blob/main/scripts/pre-commit.sh

Tests/TestingApps/PurchaseTesterSwiftUI/Core/ConfiguredPurchases.swift
```diff
@@ -45,7 +45,7 @@ public final class ConfiguredPurchases {
 
         let purchases = Purchases.configure(
             with: .builder(withAPIKey: apiKey)
-                .with(observerMode: observerMode)
+                .with(observerMode: observerMode, storeKitVersion: useStoreKit2 ? .storeKit2 : .storeKit1)
                 .with(entitlementVerificationMode: entitlementVerificationMode)
                 .with(storeKitVersion: useStoreKit2 ? .storeKit2 : .storeKit1)
```
NachoSoto, line 50: You can remove this now?

Sources/Purchasing/Configuration.swift
MarkVillacampa: Fixed, meant to use with `self.`, Basically this means: if the currently configured storeKitVersion is different than the default, it means the developer must have used `.with(sotreKitVersion:)` before calling this method.

Sources/Purchasing/Configuration.swift
MarkVillacampa: This checks for usages of this method _after_ setting the observer mode, and the warning log in the observer mode method checks for usages if this method _before_ setting the observer mode (as long as the value is different than the default, otherwise we cannot know)

RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -1001,8 +1001,8 @@
 		4D6ABB0B2AF13F9400BB2A08 /* StoreKit2Receipt.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = StoreKit2Receipt.swift; sourceTree = "<group>"; };
 		4D6ABB0D2AF13FB100BB2A08 /* StoreEnvironment.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = StoreEnvironment.swift; sourceTree = "<group>"; };
 		4D6ABB0F2AF13FBD00BB2A08 /* SK2AppTransaction.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = SK2AppTransaction.swift; sourceTree = "<group>"; };
-		4DBF1F352B4D572400D52354 /* LocalReceiptFetcher.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = LocalReceiptFetcher.swift; sourceTree = "<group>"; };
 		4DBC30952B1DFA97001D33C7 /* StoreKitVersion.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = StoreKitVersion.swift; sourceTree = "<group>"; };
+		4DBF1F352B4D572400D52354 /* LocalReceiptFetcher.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = LocalReceiptFetcher.swift; sourceTree = "<group>"; };
```
MarkVillacampa, line 1005: Xcode 15.2 keeps making this change locally so might as well commit it

Sources/Purchasing/Configuration.swift
NachoSoto: This is a smell that something is wrong with the API then.
Since we're making this a major release, I'd solve this issue by removing the default value for StoreKit version in this method. That way we don't need an easy to miss warning and instead developers know this at compile time.

Sources/Purchasing/Configuration.swift
NachoSoto: See my comment above. I think it's a smell that something is wrong with the API.

Sources/Purchasing/Configuration.swift
MarkVillacampa: I think this makes sense! It indeed felt weird while implementing it. Didn't want to break the API for developers currently using StoreKit 1, but explicit is better than implicit.

NachoSoto (approved):
Awesome 👍🏻 

Let's add a obsoletion in `Obsoletions.swift` for `Configuration.Builder.with(observerMode:)` explaining why that method is gone?

MarkVillacampa:
Oh nice I didn't think that would be possible with just a parameter overloading!

NachoSoto (approved):

PR #3647, opened 2024-02-07
Allow force dark mode.

Jjastiny:
### Checklist

### Motivation
* Allow force dark mode, when darkMode boolean is not null
* Currently I have a RN project using paywall and it does not respect the dark mode user preference.

### Description
* Allow devs to change darkmode scheme when isDarkMode is provided.
* If it's null it uses the system colorScheme env variable

Need to check lint and unit test

https://github.com/RevenueCat/purchases-ios/assets/6516487/07637747-f234-4a6a-916a-ebe58ce7a585

### DIFF
--- a/RevenueCatUI/Data/PaywallViewConfiguration.swift
+++ b/RevenueCatUI/Data/PaywallViewConfiguration.swift
@@ -20,6 +20,7 @@ struct PaywallViewConfiguration {
     var displayCloseButton: Bool
     var introEligibility: TrialOrIntroEligibilityChecker?
     var purchaseHandler: PurchaseHandler?
+    var isDarkMode: Bool?
 
     init(
         content: Content,
@@ -28,7 +29,8 @@ struct PaywallViewConfiguration {
         fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
         displayCloseButton: Bool = false,
         introEligibility: TrialOrIntroEligibilityChecker? = nil,
-        purchaseHandler: PurchaseHandler? = nil
+        purchaseHandler: PurchaseHandler? = nil,
+        isDarkMode: Bool? = nil
     ) {
         self.content = content
         self.customerInfo = customerInfo
@@ -37,6 +39,7 @@ struct PaywallViewConfiguration {
         self.displayCloseButton = displayCloseButton
         self.introEligibility = introEligibility
         self.purchaseHandler = purchaseHandler
+        self.isDarkMode = isDarkMode
     }
 
 }
@@ -67,7 +70,8 @@ extension PaywallViewConfiguration {
         fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
         displayCloseButton: Bool = false,
         introEligibility: TrialOrIntroEligibilityChecker? = nil,
-        purchaseHandler: PurchaseHandler? = nil
+        purchaseHandler: PurchaseHandler? = nil,
+        isDarkMode: Bool? = nil
     ) {
         self.init(
             content: .optionalOffering(offering),
@@ -76,7 +80,8 @@ extension PaywallViewConfiguration {
             fonts: fonts,
             displayCloseButton: displayCloseButton,
             introEligibility: introEligibility,
-            purchaseHandler: purchaseHandler
+            purchaseHandler: purchaseHandler,
+            isDarkMode: isDarkMode
         )
     }
 
--- a/RevenueCatUI/UIKit/PaywallViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallViewController.swift
@@ -134,6 +134,12 @@ public class PaywallViewController: UIViewController {
     public func updateFont(with fontName: String) {
         self.configuration.fonts = CustomPaywallFontProvider(fontName: fontName)
     }
+    
+    /// - Warning: For internal use only
+    @objc(updateDarkMode:)
+    public func updateDarkMode(with isDarkMode: Bool) {
+        self.configuration.isDarkMode = isDarkMode
+    }
 
     // MARK: - Internal
 
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -29,6 +29,7 @@ public struct PaywallView: View {
     private let mode: PaywallViewMode
     private let fonts: PaywallFontProvider
     private let displayCloseButton: Bool
+    private let isDarkMode: Bool?
 
     @Environment(\.locale)
     private var locale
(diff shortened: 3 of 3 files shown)


NachoSoto (changes requested):
Have you tried overriding `environment(\.colorScheme)` instead? We don't need a parameter on `PaywallView`.

Jjastiny:
@NachoSoto 
Gotcha, that's an awesome suggestion
which would u prefer? adding `.environment` on `PaywallContainerView` or in the initialization of `UIHostingController`?

RevenueCatUI/Data/PaywallViewConfiguration.swift
```diff
@@ -28,7 +29,8 @@ struct PaywallViewConfiguration {
         fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
         displayCloseButton: Bool = false,
         introEligibility: TrialOrIntroEligibilityChecker? = nil,
-        purchaseHandler: PurchaseHandler? = nil
+        purchaseHandler: PurchaseHandler? = nil,
+        isDarkMode: Bool? = nil
```
NachoSoto, line 33: Let's inject a `UIUserInterfaceStyle?` and then we can convert that internally to `ColorScheme`.

Jjastiny: https://github.com/RevenueCat/purchases-ios/pull/3647/commits/112eec4d8863affb43e31487e0a3d59aaac6f99d

thanks for the feedback, @NachoSoto i've updated the PR here. Also removed the isDarkMode boolean to accept a style from the viewcontroller

RevenueCatUI/Data/PaywallViewConfiguration.swift
```diff
@@ -20,6 +21,7 @@ struct PaywallViewConfiguration {
     var displayCloseButton: Bool
     var introEligibility: TrialOrIntroEligibilityChecker?
     var purchaseHandler: PurchaseHandler?
+    var uiUserInterfaceStyle: UIUserInterfaceStyle?
```
NachoSoto, line 24: This can be simpler:
```suggestion
    var userInterfaceStyle: UIUserInterfaceStyle?
```

RevenueCatUI/Data/PaywallViewConfiguration.swift
```diff
@@ -28,7 +30,8 @@ struct PaywallViewConfiguration {
         fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
         displayCloseButton: Bool = false,
         introEligibility: TrialOrIntroEligibilityChecker? = nil,
-        purchaseHandler: PurchaseHandler? = nil
+        purchaseHandler: PurchaseHandler? = nil,
+        uiUserInterfaceStyle: UIUserInterfaceStyle? = nil
```
NachoSoto, line 34: Ditto all these:
```suggestion
        userInterfaceStyle: UIUserInterfaceStyle? = nil
```

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -134,6 +134,12 @@ public class PaywallViewController: UIViewController {
     public func updateFont(with fontName: String) {
         self.configuration.fonts = CustomPaywallFontProvider(fontName: fontName)
     }
+    
+    /// - Warning: For internal use only
+    @objc(updateUserInterface:)
+    public func updateUserInterface(with style: UIUserInterfaceStyle){
```
NachoSoto, line 140: How about:
```suggestion
    @objc(updateWithUserInterfaceStyle:)
    public func update(with style: UIUserInterfaceStyle) {
```

RevenueCatUI/Data/PaywallViewConfiguration.swift
```diff
@@ -8,6 +8,7 @@
 import Foundation
 
 import RevenueCat
+import UIKit
```
NachoSoto, line 11: This will break for platforms that don't support `UIKit`. Let's not add this setting to `PaywallViewConfiguration` since it's only used for `PaywallViewController`.

NachoSoto (commented):
Thanks for your help with this! It's almost ready.

RevenueCatUI/Data/PaywallViewConfiguration.swift
Jjastiny: gotcha if that's the case then I'll have to move userInterfaceStyle out of PaywallViewConfiguration.
I'm assuming that you'd want me to create a variable userInterfaceStyle inside viewController?

NachoSoto: Yes exactly, since that parameter is exclusive to `PaywallViewController`.

Jjastiny: Cool then I'll go ahead and resolve the comments about renaming the variable in the configuration file!

Jjastiny: @NachoSoto 
One more question. 
So updating the configuration here actually recreates the hosting controller
   ```swift 
private var configuration: PaywallViewConfiguration {
        didSet {
            // Overriding the configuration requires re-creating the `HostingViewController`.
            // This is used by some Hybrid SDKs that require modifying the content after creation.
            self.hostingController = self.createHostingController()
        }
    }
````
if we were to exclude the userInterfaceStyle out of configuration, then it's most likely that it wouldn't rerender right? WDYT about this?

NachoSoto: Have you tried doing this overriding the trait in the `UIViewController`? Does that get forwarded to `SwiftUI`?

NachoSoto: That's an option, but better yet would be to avoid recreating it if `UIHostingController` detects the change in the `UITraitCollection`, maybe using `overrideTraitCollection(forChild childViewController: [UIViewController])`.

Jjastiny: @NachoSoto 
not sure if i understood you correctly, but 
without the code above,
it updates the userInterfaceStyle but does not propagate to swift UI.
with the code above,
it recreates the hosting controller which does get propagated to swiftUI

Jjastiny: Gotcha, I think our comments are out of sync haha. I'll try that!

Jjastiny: @NachoSoto 
Looks like it's working great!

https://github.com/RevenueCat/purchases-ios/assets/6516487/cbb94fe4-ac1b-417f-992a-aa080700dfa7

Added a timer to switch back and forth dark and light automatically from view controller, and it works great.

NachoSoto (approved):

PR #3661, opened 2024-02-09
`PaywallData`: change all `Color`s to be optional

NachoSoto:
This will allow the implementation of _color overrides_ by tier for the upcoming multi-tier paywall.

### DIFF
--- a/RevenueCatUI/Helpers/ColorInformation+MultiScheme.swift
+++ b/RevenueCatUI/Helpers/ColorInformation+MultiScheme.swift
@@ -101,16 +101,18 @@ import SwiftUI
 // Helpful acessors
 extension PaywallData.Configuration.Colors {
 
-    var backgroundColor: Color { self.background.underlyingColor }
-    var text1Color: Color { self.text1.underlyingColor }
-    var text2Color: Color { self.text2?.underlyingColor ?? self.text1.underlyingColor }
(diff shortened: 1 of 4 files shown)


RevenueCatUI/Helpers/ColorInformation+MultiScheme.swift
```diff
@@ -101,17 +101,19 @@ import SwiftUI
 // Helpful acessors
 extension PaywallData.Configuration.Colors {
 
-    var backgroundColor: Color { self.background.underlyingColor }
-    var text1Color: Color { self.text1.underlyingColor }
-    var text2Color: Color { self.text2?.underlyingColor ?? self.text1.underlyingColor }
+    var backgroundColor: Color { self.background?.underlyingColor ?? Self.defaultColor }
+    var text1Color: Color { self.text1?.underlyingColor ?? Self.defaultColor }
```
tonidero, line 105: Hmm seems strange for the background color and text color to default to the same color? (You wouldn't be able to see the text).

I guess the backend should perform the validations to make sure these colors are given, but I thought I would mention it.

tonidero (approved):

RevenueCatUI/Helpers/ColorInformation+MultiScheme.swift
NachoSoto: Yeah good call. We can have better defaults here. Changed.

PR #3662, opened 2024-02-09
`Paywalls`: allow empty `packages`

NachoSoto:
`PaywallData.packages` won't be used by multi-tier paywalls, so we need to allow that.

### DIFF
--- a/RevenueCatUI/Data/PaywallData+Validation.swift
+++ b/RevenueCatUI/Data/PaywallData+Validation.swift
@@ -48,11 +48,13 @@ extension Offering {
                 return (paywall, template, nil)
 
             case let .failure(error):
-                // If there are any errors, create a default paywall
-                // with only the configured packages.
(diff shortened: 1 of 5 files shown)


RevenueCatUI/Data/PaywallData+Validation.swift
```diff
@@ -48,11 +48,13 @@ extension Offering {
                 return (paywall, template, nil)
 
             case let .failure(error):
-                // If there are any errors, create a default paywall
-                // with only the configured packages.
-                return (.createDefault(with: paywall.config.packages, locale: locale),
-                        PaywallData.defaultTemplate,
-                        error)
+                let paywall: PaywallData = paywall.config.packages.isEmpty
```
NachoSoto, line 51: This change is covered by a test in #3664.

NachoSoto:
The validation code that replaces this won't be in `main` so I don't want to release this change without that.

PR #3665, opened 2024-02-09
`Paywalls`: allow overriding `UIUserInterfaceStyle` in `PaywallViewController` (by @Jjastiny)

NachoSoto:
Original PR: #3647.

`PaywallView` allows overriding it with `.environment(\.colorScheme)`.
This allows doing the same with `PaywallViewController`.

https://github.com/RevenueCat/purchases-ios/assets/6516487/07637747-f234-4a6a-916a-ebe58ce7a585

### DIFF
--- a/RevenueCatUI/UIKit/PaywallViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallViewController.swift
@@ -134,10 +134,12 @@ public class PaywallViewController: UIViewController {
     public func updateFont(with fontName: String) {
         self.configuration.fonts = CustomPaywallFontProvider(fontName: fontName)
     }
-    
-    /// - Warning: For internal use only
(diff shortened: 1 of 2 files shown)


Jjastiny:
@NachoSoto 
First of all I want to apologize for this.
I realized that our RN project was updating the color pallete using redux, without updating the system appearance settings (Appearance.setColorScheme()) which caused an issue with the paywall's dark mode. I believe we did this b/c of performance issues where the dark mode switch is very slow.

But after changing the colorScheme in the RN project, it looks like the dark mode on the paywall is updated correctly even without this change... Unless your team needs this for anything else, I think this could be closed :)

NachoSoto:
That makes sense! Glad you got it working. Thanks for contributing this anyway 👍🏻

PR #3664, opened 2024-02-09
`Paywalls`: `PaywallData.localizedConfiguration` is optional

NachoSoto:
This is a breaking change required to support multi-tier paywalls.

Instead of failing to decode `PaywallData`, this is now validated when processing the data before displaying the paywall.

PR #3666, opened 2024-02-09
`Paywalls`: added tier information to `PaywallData`

NachoSoto:
This will be used for multi-tier paywalls

Depends on #3661 and #3664.


Sources/Paywalls/PaywallData.swift
```diff
@@ -396,6 +488,33 @@ extension PaywallData.Configuration {
...
+extension PaywallData {
+
+    /// A group of packages that can be displayed together in a multi-tier paywall template.
+    public struct Tier {
+
+        /// The identifier for this tier.
+        public var id: String
+
+        /// The list of package identifiers this tier will display
+        public var packages: [String]
```
tonidero, line 502: Should we rename this to `packageIds`? 

I was actually wondering if we should put the actual packages here as well instead even if it means duplicating the references... But I guess I'm ok with this considering we're likely the ones that will be using this.

tonidero (approved):
LGTM! Just a comment

Sources/Paywalls/PaywallData.swift
NachoSoto: > Should we rename this to packageIds?

In retrospect, yes. But I think it's clearer that this matches `PaywallData.packages` since it serves the same purpose.

> I was actually wondering if we should put the actual packages here as well instead even if it means duplicating the references...

What do you mean by the actual packages?

PR #3673, opened 2024-02-12
Run load shedder integration tests on release branches

tonidero:
### Description
This will run the load shedder integration tests on release branches.

PR #3675, opened 2024-02-12
`CI`: split load shedder integration tests

NachoSoto:
Follow up to #3673.

### Changes:
- Split `loadshedder-integration-tests` into `integration-tests-all` and `loadshedder-integration-tests-v3`
- Removed `loadshedder-integration-tests` from `build_test`: this reverts #3673. We already run loadshedder tests as part of `SK1` and `SK2` test suites.
- Add `BackendIntegrationTests-Shedder` test plane. This isn't used by CI (see above point), but it can be useful to run all of the load shedder tests.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -977,7 +977,7 @@ jobs:
           name: Create automatic PR
           command: bundle exec fastlane automatic_bump github_rate_limit:10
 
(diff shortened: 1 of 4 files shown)


tonidero (approved):
Makes sense, thanks for doing this!

PR #3680, opened 2024-02-13
`CI`: skip `RevenueCatUI` API tests when generating snapshots

NachoSoto:
Small performance optimization. This isn't needed when creating snapshots.


joshdholtz (approved):
🫡

PR #3517, opened 2023-12-13
`Paywalls`: always dismiss paywalls automatically after a purchase

NachoSoto:
Fixes #3516.

This logic only existed for `.presentPaywallIfNeeded`. However, it should apply to `PaywallView` when presented manually as well.

This is also consistent with how Android works, reduces boilerplate code required by users, and removes the possibility that a paywall is still presented after a purchase, which doesn't make sense since our paywalls don't have a mode for "already subscribed".

PR #3683, opened 2024-02-13
`Paywalls`: don't dismiss footer paywalls automatically

NachoSoto:
Follow up to #3517.

This behavior is the default and not opt-in, and it doesn't make sense on footer paywalls.


joshdholtz (approved):
:shipit:

PR #3684, opened 2024-02-13
`Paywalls`: template 7 allows image overrides by tier

NachoSoto:
https://github.com/RevenueCat/purchases-ios/assets/685609/2d761ce1-cd06-44ac-999b-5d4d78087ce6

### DIFF
--- a/Sources/Paywalls/PaywallData.swift
+++ b/Sources/Paywalls/PaywallData.swift
@@ -238,16 +238,22 @@ extension PaywallData {
         }
 
         /// The images for each of the tiers.
-        public var imagesByTier: [String: Images] {
-            let images = self.images
+        public internal(set) var imagesByTier: [String: Images] {
+            get {
+                let images = self.images
 
-            return Set(self.tiers.map(\.id))
-                .dictionaryWithValues { tier in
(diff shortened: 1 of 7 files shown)


joshdholtz (approved):
:shipit:

Sources/Paywalls/PaywallData.swift
```diff
@@ -238,16 +238,22 @@ extension PaywallData {
...
 
-            return Set(self.tiers.map(\.id))
-                .dictionaryWithValues { tier in
-                    return Self.merge(
-                        source: self._imageOverridesByTier[tier],
-                        fallback: images
-                    )
-                }
+            set {
+                self._imageOverridesByTier = newValue
```
vegaro, line 255: not sure in which conditions this setter will be called, but I think it's a bit confusing that it sets the overrides someone calling it will probably set the whole `imagesByTier`, which includes the base ones (not only the overrides)

NachoSoto: Yeah I implemented this "correctly" for `images` since it's a derived property too.
That's why I made this `setter` `internal` though. But I'll leave a comment.

PR #3685, opened 2024-02-13
`Paywalls`: fix `PaywallColor.init(light:dark:)`

NachoSoto:
We probably didn't notice this because `PaywallData.Configuration.ColorInformation.multiScheme` checks for the existence of `dark` before.

However, this new implementation helps with the upcoming tier color overrides.


joshdholtz (approved):
:shipit:

PR #3627, opened 2024-02-02
`Paywalls`: new `.onPurchaseStarted` modifier

NachoSoto:
Similar to #3622.

This API is available in:
- `PaywallView()`
- `.presentPaywallIfNeeded`
- `.paywallFooter`
- `PaywallViewControllerDelegate`

PR #3693, opened 2024-02-15
`Paywalls`: new `.onPurchaseStarted(package)` modifier

vegaro:
Similar to #3627.

This API is available in:
- `PaywallView()`
- `.presentPaywallIfNeeded`
- `.paywallFooter`
- `PaywallViewControllerDelegate`

### DIFF
--- a/RevenueCatUI/View+PresentPaywall.swift
+++ b/RevenueCatUI/View+PresentPaywall.swift
@@ -37,12 +37,72 @@ extension PaywallPresentationMode {
 
 }
 
+// swiftlint:disable file_length
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 @available(macOS, unavailable, message: "RevenueCatUI does not support macOS yet")
 @available(tvOS, unavailable, message: "RevenueCatUI does not support tvOS yet")
 extension View {
 
     typealias CustomerInfoFetcher = @Sendable () async throws -> CustomerInfo
+
+    // swiftlint:disable line_length
+    /// Presents a ``PaywallView`` if the given entitlement identifier is not active
+    /// in the current environment for the current `CustomerInfo`.
+    /// ```swift
+    /// var body: some View {
+    ///    YourApp()
+    ///      .presentPaywallIfNeeded(requiredEntitlementIdentifier: "pro")
+    /// }
+    /// ```
+    /// - Note: If loading the `CustomerInfo` fails (for example, if Internet is offline),
+    /// the paywall won't be displayed.
+    ///
+    /// - Parameter offering: The `Offering` containing the desired `PaywallData` to display.
+    /// If `nil` (the default), `Offerings.current` will be used. Note that specifying this parameter means
+    /// that it will ignore the offering configured in an active experiment.
+    /// - Parameter fonts: An optional ``PaywallFontProvider``.
+    /// - Parameter presentationMode: The desired presentation mode of the paywall. Defaults to `.sheet`.
+    ///
+    /// ### Related Articles
+    /// [Documentation](https://rev.cat/paywalls)
+    ///
+    /// - Tag: presentPaywallIfNeeded
+    @available(iOS, deprecated: 1, renamed: "presentPaywallIfNeeded(requiredEntitlementIdentifier:offering:fonts:presentationMode:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:onDismiss:)")
+    @available(tvOS, deprecated: 1, renamed: "presentPaywallIfNeeded(requiredEntitlementIdentifier:offering:fonts:presentationMode:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:onDismiss:)")
+    @available(watchOS, deprecated: 1, renamed: "presentPaywallIfNeeded(requiredEntitlementIdentifier:offering:fonts:presentationMode:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:onDismiss:)")
+    @available(macOS, deprecated: 1, renamed: "presentPaywallIfNeeded(requiredEntitlementIdentifier:offering:fonts:presentationMode:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:onDismiss:)")
+    @available(macCatalyst, deprecated: 1, renamed: "presentPaywallIfNeeded(requiredEntitlementIdentifier:offering:fonts:presentationMode:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:onDismiss:)")
+    // swiftlint:enable line_length
+    public func presentPaywallIfNeeded(
+        requiredEntitlementIdentifier: String,
+        offering: Offering? = nil,
+        fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
+        presentationMode: PaywallPresentationMode = .default,
+        purchaseStarted: @escaping PurchaseStartedHandler,
+        purchaseCompleted: PurchaseOrRestoreCompletedHandler? = nil,
+        purchaseCancelled: PurchaseCancelledHandler? = nil,
+        restoreCompleted: PurchaseOrRestoreCompletedHandler? = nil,
+        purchaseFailure: PurchaseFailureHandler? = nil,
+        restoreFailure: PurchaseFailureHandler? = nil,
+        onDismiss: (() -> Void)? = nil
+    ) -> some View {
+        return self.presentPaywallIfNeeded(
+            requiredEntitlementIdentifier: requiredEntitlementIdentifier,
+            offering: offering,
+            fonts: fonts,
+            presentationMode: presentationMode,
+            purchaseStarted: { _ in
+                purchaseStarted()
+            },
+            purchaseCompleted: purchaseCompleted,
+            purchaseCancelled: purchaseCancelled,
+            restoreStarted: nil,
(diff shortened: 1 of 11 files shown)


RevenueCatUI/View+PresentPaywall.swift
```diff
@@ -65,18 +67,77 @@ extension View {
...
+    @available(macOS, deprecated: 1, renamed: "presentPaywallIfNeeded(requiredEntitlementIdentifier:offering:fonts:presentationMode:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:onDismiss:)")
+    @available(macCatalyst, deprecated: 1, renamed: "presentPaywallIfNeeded(requiredEntitlementIdentifier:offering:fonts:presentationMode:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:onDismiss:)")
+    // swiftlint:enable line_length
     public func presentPaywallIfNeeded(
         requiredEntitlementIdentifier: String,
         offering: Offering? = nil,
         fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
         presentationMode: PaywallPresentationMode = .default,
-        purchaseStarted: PurchaseStartedHandler? = nil,
+        purchaseStarted: @escaping PurchaseStartedHandler,
```
vegaro, line 81: Made this non nullable so the function wouldn't collide if someone was only passing `requiredEntitlementIdentifier`. I also had to made it `@escaping`. Is that considered a breaking change?

RevenueCatUI/View+PresentPaywall.swift
```diff
@@ -134,12 +196,18 @@ extension View {
...
+    @available(macOS, deprecated: 1, renamed: "presentPaywallIfNeeded(offering:fonts:presentationMode:shouldDisplay:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:onDismiss:)")
+    @available(macCatalyst, deprecated: 1, renamed: "presentPaywallIfNeeded(offering:fonts:presentationMode:shouldDisplay:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:onDismiss:)")
+    // swiftlint:enable line_length
     public func presentPaywallIfNeeded(
         offering: Offering? = nil,
         fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
         presentationMode: PaywallPresentationMode = .default,
         shouldDisplay: @escaping @Sendable (CustomerInfo) -> Bool,
-        purchaseStarted: PurchaseStartedHandler? = nil,
+        purchaseStarted: @escaping PurchaseStartedHandler,
```
vegaro, line 210: Same here. If someone was not passing anything or passing let's say just `offering`, it would collide with the new version

RevenueCatUI/View+PresentPaywallFooter.swift
```diff
@@ -19,6 +19,7 @@ import SwiftUI
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
 extension View {
 
+    // swiftlint:disable line_length
```
tonidero, line 22: Hmm I might be wrong but I think we need to reenable this check after the affected block, so we still check the rest of the file?

RevenueCatUI/View+PurchaseRestoreCompleted.swift
```diff
@@ -53,6 +56,28 @@ extension View {
     /// [Documentation](https://rev.cat/paywalls)
     public func onPurchaseStarted(
         _ handler: @escaping PurchaseStartedHandler
+    ) -> some View {
+        return self.modifier(OnPurchaseStartedModifier(handler: { _ in
+            handler()
+        }))
+    }
```
tonidero, line 63: Should we mar this one as deprecated as well?

RevenueCatUI/View+PurchaseRestoreCompleted.swift
```diff
@@ -24,6 +24,9 @@ public typealias PurchaseCompletedHandler = @MainActor @Sendable (_ transaction:
 /// A closure used for notifying of purchase initiation.
 public typealias PurchaseStartedHandler = @MainActor @Sendable () -> Void
```
tonidero, line 25: Same here, not sure if we can mark a typealias as deprecated 🤔

tonidero (approved):
LGTM! I don't think the escaping change is breaking but might be good to confirm with @joshdholtz and/or @MarkVillacampa

RevenueCatUI/View+PresentPaywall.swift
joshdholtz: @vegaro I'm not certain if the `@escaping` is breaking changing (I'd have to experiment) but making this non-nullable is a breaking change, isn't it? 😬 

If somebody was using this function without `purchaseStarted`, it will now require them to have `purchaseStarted`?

RevenueCatUI/View+PresentPaywallFooter.swift
joshdholtz: Agreed... I think this has to be enabled ☝️

RevenueCatUI/View+PresentPaywallFooter.swift
```diff
@@ -29,15 +30,57 @@ extension View {
...
+    @available(tvOS, deprecated: 1, renamed: "paywallFooter(condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    @available(watchOS, deprecated: 1, renamed: "paywallFooter(condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    @available(macOS, deprecated: 1, renamed: "paywallFooter(condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    @available(macCatalyst, deprecated: 1, renamed: "paywallFooter(condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    // swiftlint:enable line_length
     public func paywallFooter(
         condensed: Bool = false,
         fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
-        purchaseStarted: PurchaseStartedHandler? = nil,
+        purchaseStarted: @escaping PurchaseStartedHandler,
```
joshdholtz, line 42: Same here... I think making this non-nullable is a breaking change too if somebody wasn't already calling it 🤔

Tests/RevenueCatUITests/PurchaseCompletedHandlerTests.swift
```diff
@@ -35,13 +36,17 @@ class PurchaseCompletedHandlerTests: TestCase {
             .onPurchaseStarted {
                 started = true
             }
+            .onPurchaseStarted { package in
```
joshdholtz, line 39: Does this work for a variable name?! Even though `package` is a reserved word? 😅

RevenueCatUI/View+PresentPaywall.swift
joshdholtz: Did the API tester for this break? That would really let us know, wouldn't it? 🤷‍♂️

vegaro: It didn't fail, that's why I am not sure, but I think it's not

> If somebody was using this function without purchaseStarted, it will now require them to have purchaseStarted?
If they were not passing it, it will use the new non-deprecated function automatically (that one has it as nullable)

joshdholtz: Ahhh, okay! That makes sense then 😊 I think this should be good then!

joshdholtz (approved):

RevenueCatUI/View+PurchaseRestoreCompleted.swift
vegaro: I tried and it is possible but there's a big gotcha, which is that the warnings can't be silenced for types only for functions, or a least I couldn't get that to work and the api tests were warning all over the place

RevenueCatUI/View+PurchaseRestoreCompleted.swift
vegaro: Oh yes, good catch

RevenueCatUI/View+PurchaseRestoreCompleted.swift
tonidero: Ok, then I think it's fine to just deprecate the usages 👍

RevenueCatUI/View+PresentPaywallFooter.swift
vegaro: Adding same comment as above for visibility

If they were not passing it, it will use the new non-deprecated function automatically (that one has it as nullable)

RevenueCatUI/View+PurchaseRestoreCompleted.swift
vegaro: added in [f5c21b5](https://github.com/RevenueCat/purchases-ios/pull/3693/commits/f5c21b56a8baf64ea314e43f54f87908602d52fb) and [e5b935f](https://github.com/RevenueCat/purchases-ios/pull/3693/commits/e5b935fe84a7bf111dc838fc5a6a76d5cef7baf1)

RevenueCatUI/View+PurchaseRestoreCompleted.swift
vegaro: managed to add it in [f5c21b5](https://github.com/RevenueCat/purchases-ios/pull/3693/commits/f5c21b56a8baf64ea314e43f54f87908602d52fb) and [e5b935f](https://github.com/RevenueCat/purchases-ios/pull/3693/commits/e5b935fe84a7bf111dc838fc5a6a76d5cef7baf1)

PR #3694, opened 2024-02-15
`Paywalls`: new `.onRestoreStarted` modifier

vegaro:
Similar to #3627.

This API is available in:
- `PaywallView()`
- `.presentPaywallIfNeeded`
- `.paywallFooter`
- `PaywallViewControllerDelegate`

PR #3698, opened 2024-02-20
Remove `restoreStarted` from extensions

vegaro:
Since we are deprecating the old function in https://github.com/RevenueCat/purchases-ios/pull/3693/files 

I thought it will be simpler to just add both `restoreStarted` and the new `purchaseStarted(package:)`  in the same release. We haven't made a release since I merged #3694 so these are safe to remove for now.

I open this PR in case we want to make a release before #3693 is finished.


tonidero:
I wouldn't consider this a `fix` but more of a `refactor`? Since nothing was wrong

tonidero (approved):
I guess this is not removed from the VC delegate, but that's probably ok. This looks good. Should we hold this until the other PR is ready or we make a release? (which should happen tomorrow)

joshdholtz (approved):

vegaro:
I think we can merge and it's fine to released tomorrow. Hoping to finish that other PR today though. Just wanted to merge this in case we needed to do a release today.

PR #3700, opened 2024-02-21
`[SK2]`: Fix syncPurchases

MarkVillacampa:
We can use any transaction token available, not only auto-renewable ones

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -1094,7 +1094,7 @@ private extension PurchasesOrchestrator {
 
         self.attribution.unsyncedAdServicesToken { adServicesToken in
             _ = Task<Void, Never> {
-                let transaction = await self.transactionFetcher.firstVerifiedAutoRenewableTransaction
+                let transaction = await self.transactionFetcher.firstVerifiedTransaction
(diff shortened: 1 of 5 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1307,7 +1307,7 @@ private extension PurchasesOrchestrator {
                              completion: @escaping @Sendable (Result<PromotionalOffer, PurchasesError>) -> Void) {
 
         _ = Task<Void, Never> {
-            let transaction = await self.transactionFetcher.firstVerifiedAutoRenewableTransaction
+            let transaction = await self.transactionFetcher.firstVerifiedTransaction
```
tonidero, line 1310: Should this only be subscriptions? Not sure if we can have promotional purchases for non-subscriptions?

joshdholtz (approved):
:shipit:

tonidero (approved):
Left a comment about the promotionalOffers but should be ok for the syncPurchases I think

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
MarkVillacampa: Changed back to return only subscriptions for promotional offers 👍

PR #3701, opened 2024-02-21
Retry flaky tests

vegaro:
I've noticed we have a bunch of flaky tests, and it's making merging PRs very time consuming. I added a retry to failing tests, hoping it reduces the amount of times we have to retry in CircleCI


vegaro:
I think this mainly helps with UI tests rather than the backend integration tests, because I just retried them twice and they failed

tonidero (approved):

PR #3707, opened 2024-02-22
Add `currentOffering(forPlacement: String)` to `Offerings`

guido732:
### Motivation

New function to get `Offering` by a placement identifier

### Description

New `currentOffering(forPlacement: String): Offering?` method on `Offerings`
- Added demo button for this in purchase tester app

```swift
// Example usage
if let offering = offerings.currentOffering(forPlacement: "onboarding") {
  showPaywall(offering)
}
```

#### Logic flow

- If a placement identifier explicitly exists within the `placements.offeringIdsByPlacement` dictionary, the method attempts to fetch the corresponding `Offering` without resorting to a fallback
  - Example: when there is a `nil` / null value for a placement, we don't want to use the fallback
- If the placement identifier does not exist, the method then attempts to fetch a fallback offering defined by `placements.fallbackOfferingId`

#### New placements object in offerings response

```js
{
    // other offerings stuff
    "placements": {
        "fallback_offering_id": "offering_a", // can be null
        "offering_ids_by_placement": {
            "onboarding_start": null,
            "onboarding_end": "offering_b",
        }
  }
}
```


Sources/Purchasing/Offerings.swift
```diff
@@ -93,4 +96,48 @@ public extension Offerings {
...
+        }
+
+        let returnOffering: Offering?
+        if let explicitOfferingId: String? = placements.offeringIdsByPlacement[placementIdentifier] {
+            // Don't use fallback since placement id was explicity set in the dictionary
+            returnOffering = explicitOfferingId.flatMap { self.all[$0] }
+        } else {
+            // Use fallback since the placement didn't exist
+            returnOffering =  placements.fallbackOfferingId.flatMap { self.all[$0]}
+        }
```
joshdholtz, line 117: `placements.offeringIdsByPlacement[placementIdentifier]` returns at `Offering??` because `placements.offeringIdsByPlacement` is a `[String: String?]`

The result of could be an explicit `nil` value. In that case, we don't want to use the fallback offering.

If `placements.offeringIdsByPlacement[placementIdentifier]` is not found and goes into the `else`, it means no key was found and we should use the fallback offering.

Sources/Purchasing/Placements.swift
```diff
@@ -0,0 +1,29 @@
...
+//
+//      https://opensource.org/licenses/MIT
+//
+//  Placements.swift
+//
+//  Created by Guido Torres on 15/02/2024.
+
+import Foundation
+
+internal final class Placements: NSObject {
```
tonidero, line 16: I'm wondering if this should be inside `Offering` as well, so it's clearer it's refering to that. Right now this is a bit too generic IMO

Sources/Purchasing/Offerings.swift
```diff
@@ -93,4 +96,48 @@ public extension Offerings {
...
+    func copyWith(placementIdentifier: String?) -> Offering {
+        let updatedPackages = self.availablePackages.map { pkg in
+            let newContext = PresentedOfferingContext(
+                offeringIdentifier: pkg.presentedOfferingContext.offeringIdentifier,
+                placementIdentifier: placementIdentifier
+            )
+            return Package(identifier: pkg.identifier,
+                           packageType: pkg.packageType,
+                           storeProduct: pkg.storeProduct,
+                           presentedOfferingContext: newContext
```
tonidero, line 133: Nitpick but should we have an internal copy constructor from package so we can create a new package with a new context without having to pass all the parameters again?

Tests/APITesters/CustomEntitlementComputationSwiftAPITester/SwiftAPITester/PresentedOfferingContextAPI.swift
```diff
@@ -14,5 +14,5 @@ func checkPresentedOfferingContextAPI(context: PresentedOfferingContext! = nil)
 }
 
 private func checkCreatePresentedOfferingContextAPI() {
-    let _: PresentedOfferingContext = .init(offeringIdentifier: "")
+    let _: PresentedOfferingContext = .init(offeringIdentifier: "", placementIdentifier: "")
```
tonidero, line 17: We should have API tests without the new parameter as well, to make sure we don't break it.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1170,13 +1171,14 @@ private extension PurchasesOrchestrator {
...
                                     restored: Bool) {
-        let offeringContext = self.getAndRemovePresentedOfferingIdentifier(for: purchasedTransaction)
+        let offeringContext = self.getAndRemovePresentedOfferingContext(for: purchasedTransaction)
         let paywall = self.getAndRemovePresentedPaywall()
         let unsyncedAttributes = self.unsyncedAttributes
         self.attribution.unsyncedAdServicesToken { adServicesToken in
             let transactionData: PurchasedTransactionData = .init(
                 appUserID: self.appUserID,
                 presentedOfferingID: offeringContext?.offeringIdentifier,
+                presentedPlacementID: offeringContext?.placementIdentifier,
```
tonidero, line 1181: Nitpick but should we have a constructor that accepts the whole `offeringContext`?

tonidero (commented):
Have a meeting, will continue reviewing after

Sources/Support/PaywallExtensions.swift
```diff
@@ -124,8 +124,16 @@ private extension View {
                     Logger.appleWarning(Strings.configure.sk2_required_for_swiftui_paywalls)
                 }
 
+                // Find offering context from a matching package
+                let offeringContext = offering.availablePackages.first {
+                    $0.storeProduct.productIdentifier == product.id
```
tonidero, line 129: Hmm this could potentially be wrong if the offering has multiple packages with the same product. It would likely be a misconfiguration so maybe it's ok to assume this in almost all cases though...

Tests/APITesters/ObjCAPITester/ObjCAPITester/RCPresentedOfferingContextAPI.m
```diff
@@ -0,0 +1,24 @@
...
+#import "RCPresentedOfferingContextAPI.h"
+
+@import StoreKit;
+@import RevenueCat;
+
+@implementation RCPresentOfferingContextAPI
+
++ (void)checkAPI {
+    RCPresentedOfferingContext *poc = [[RCPresentedOfferingContext alloc] initWithOfferingIdentifier:@"" placementIdentifier:nil];
+    poc = [[RCPresentedOfferingContext alloc] initWithOfferingIdentifier:@"" placementIdentifier:@""];
```
tonidero, line 17: We should have API tests without passing a `placementIdentifier`.

Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
```diff
@@ -498,6 +502,7 @@ private extension OfferingsManagerTests {
                 }
                 .dictionaryWithKeys(\.identifier),
             currentOfferingID: MockData.anyBackendOfferingsResponse.currentOfferingId,
+            placements: .init(fallbackOfferingId: "", offeringIdsByPlacement: [:]),
```
tonidero, line 505: This doesn't have the `.init(wrappedValue: [:])`, is that intentional? 

Also, since this seems to be kinda repeated, should we extract it to a constant?

tonidero (approved):
Some non-blocking comments. Looks good!

Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
joshdholtz: Some of these should just be `nil`! I'll go change those

But the `.init(wrappedValue: [:])` ones are only needed when initializing the `OfferingsResponse.Placements`. That one has a property wrapper in it called `@DefaultDecodable.EmptyDictionary` which is why the wrapped value is needed 🤷‍♂️ So yup yup, its intential!

Tests/APITesters/ObjCAPITester/ObjCAPITester/RCPresentedOfferingContextAPI.m
joshdholtz: Objective-C doesn't have that API 🤷‍♂️  It only has this one signature where `placementIdentifier`  nullable so this test passes a `nil` and a `@""`

Sources/Support/PaywallExtensions.swift
joshdholtz: This is true! However, both packages would actually have the same `PresentedOfferingContext` in this situation so this is "technically" okay 🤷‍♂️  I'll make a note of that in the comments. But this is also only used by the StoreKit Views so it probably doesn't get used too often.

Tests/APITesters/ObjCAPITester/ObjCAPITester/RCPresentedOfferingContextAPI.m
tonidero: Ahh ok, Since we haven't released the other one it should be ok 👍

joshdholtz: I lied... I just added that convince initializer because why not 🤷‍♂️ So I'm adding it to the API tester now 😇

Sources/Purchasing/Offerings.swift
```diff
@@ -93,4 +110,50 @@ public extension Offerings {
         return description
     }
 
+    /**
+     Retrieves a current offering for a placement identifier, use this to access offerings defined by targeting
+     placements configured in the RevenueCat dashboard, 
+     e.g. `offerings.getCurrentOffering(forPlacement: "placement_id")`.
+     */
+    @objc(getCurrentOfferingForPlacement:)
+    func getCurrentOffering(forPlacement placementIdentifier: String) -> Offering? {
```
aboedo, line 119: is it more swift-style to make this `(for placement placementIdentifier: String)` instead? 
so the callsite is 
```swift 
getCurrentOffering(for: "HomePageBanner")
```

instead of

```swift 
getCurrentOffering(forPlacement: "HomePageBanner")
```

or even 

```swift
currentOffering(for: "HomePageBanner")
```

aboedo: we've historically reserved verbs like `get` or `fetch` as ways to indicate that it'll issue some work like network requests

Sources/Purchasing/Package.swift
```diff
@@ -19,18 +19,31 @@ import Foundation
 ///
 @objc(RCPresentedOfferingContext) public final class PresentedOfferingContext: NSObject {
 
-    /// The identifier of the ``Offering`` containing this Package.
+    /// The identifier of the ``Offering`` containing this ``Package``.
```
aboedo, line 22: I'm late to the game here, but... what do I do with this class as a developer? Is it so I can use it for analytics? 
The docstring for the class doesn't really provide any context on it

aboedo (commented):
looking great, left a couple of questions

Sources/Purchasing/Package.swift
joshdholtz: @aboedo Oops, missed this comment from yesterday! This is mainly for internal usage. Instead of passing `offeringIdentifier` from get offerings to post receipt, this objects gets passed around instead. It holds offering identifier, placement information, and targeting information.

PR #3709, opened 2024-02-22
New `syncAttributesAndOfferingsIfNeeded` method

lburdock:
### Motivation

Adding new method to manually sync attributes and fetch fresh offerings 

### Description

- New `Purchases.shared.syncAttributesAndOfferingsIfNeeded()` method
    - Syncs subscriber attributes and then fetches fresh offerings from network
    - This is designed to be a blocking call since developer will want fresh offerings after syncing subscriber attributes
    - The method is rate limited to 5 calls per minute (new `RateLimiter` class()

### DIFF
--- a/Sources/Misc/RateLimiter.swift
+++ b/Sources/Misc/RateLimiter.swift
@@ -14,25 +14,33 @@
 import Foundation
 
 internal class RateLimiter {
-    private var timestamps: [Date] = []
+    private var timestamps: [Date?]
+    private var index: Int = 0
+    private let maxCallsInclusive: Int
 
     let maxCalls: Int
-    let period: TimeInterval // Period in seconds
+    let period: TimeInterval
 
     init(maxCalls: Int, period: TimeInterval) {
         self.maxCalls = maxCalls
+        self.maxCallsInclusive = self.maxCalls + 1
         self.period = period
+
+        self.timestamps = Array(repeating: nil, count: maxCallsInclusive)
     }
 
     func shouldProceed() -> Bool {
         let now = Date()
-        timestamps = timestamps.filter { now.timeIntervalSince($0) <= period }
+        let oldestIndex = (index + 1) % maxCallsInclusive
+        let oldestTimestamp = timestamps[oldestIndex]
 
-        if timestamps.count < maxCalls {
-            timestamps.append(now)
-            return true
-        } else {
+        // Check if the oldest timestamp is outside the rate limiting period or if it's nil
+        if let oldestTimestamp = oldestTimestamp, now.timeIntervalSince(oldestTimestamp) <= period {
             return false
+        } else {
+            timestamps[index] = now
+            index = oldestIndex
+            return true
         }
     }
 }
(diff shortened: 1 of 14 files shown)


Sources/Purchasing/OfferingsManager.swift
```diff
@@ -40,17 +42,16 @@ class OfferingsManager {
     func offerings(
         appUserID: String,
         fetchPolicy: FetchPolicy = .default,
+        fetchCurrent: Bool = false,
```
tonidero, line 45: I nearly confused the `fetchPolicy` parameter above with the `CacheFetchPolicy` 😅. We should eventually move to that, but I'm ok with a boolean parameter for now.

Sources/Misc/RateLimiter.swift
```diff
@@ -0,0 +1,38 @@
...
+    let period: TimeInterval // Period in seconds
+
+    init(maxCalls: Int, period: TimeInterval) {
+        self.maxCalls = maxCalls
+        self.period = period
+    }
+
+    func shouldProceed() -> Bool {
+        let now = Date()
+        timestamps = timestamps.filter { now.timeIntervalSince($0) <= period }
```
tonidero, line 29: Oh I kinda liked the approach in android since it avoided this loop every time we check this... But it shouldn't take long at all, so I'm ok with this.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -793,6 +798,23 @@ public extension Purchases {
         return try await logOutAsync()
     }
 
+    @objc func syncAttributesAndOfferingsIfNeeded(completion: @escaping (Offerings?, PublicError?) -> Void) {
```
tonidero, line 801: Hmm I've been thinking whether it would be worth moving the logic inside this method to the `PurchasesOrchestrator` (mostly for organization purposes)... However, seems that the `getOfferings` method already was not moved there, so I'm ok with this for now.

Sources/Misc/RateLimiter.swift
```diff
@@ -0,0 +1,38 @@
...
+    let maxCalls: Int
+    let period: TimeInterval // Period in seconds
+
+    init(maxCalls: Int, period: TimeInterval) {
+        self.maxCalls = maxCalls
+        self.period = period
+    }
+
+    func shouldProceed() -> Bool {
+        let now = Date()
```
tonidero, line 28: Since we are getting the current time here, it makes it more difficult to mock the values... Should we inject a provider-type class in the constructor of the limiter? That would avoid the `sleep` in the tests.

tonidero (approved):
Left some comments, but nothing blocking

Sources/Purchasing/OfferingsManager.swift
joshdholtz: Same, TBH 😅  We thought about adding a similar enum didn't want to over complicate it. Should be easy to switch to an enum later if we want!

PR #3710, opened 2024-02-23
Avoid updating docs default version on prereleases

tonidero:
### Description
We should not update the main docs page on prereleases. This makes sure we are pointing to the latest stable version docs.


tonidero:
Closing in favor of #3711

PR #3738, opened 2024-02-28
PaywallViewController: option to not automatically dismiss and add `onRequestedDismissal` 

vegaro:
Please read https://github.com/RevenueCat/react-native-purchases/pull/913

### DIFF
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -29,6 +29,7 @@ public struct PaywallView: View {
     private let mode: PaywallViewMode
     private let fonts: PaywallFontProvider
     private let displayCloseButton: Bool
+    private let shouldAutomaticallyDismiss: Bool
 
     @Environment(\.locale)
     private var locale
@@ -104,6 +105,7 @@ public struct PaywallView: View {
         self.mode = configuration.mode
         self.fonts = configuration.fonts
         self.displayCloseButton = configuration.displayCloseButton
+        self.shouldAutomaticallyDismiss = configuration.shouldAutomaticallyDismiss
     }
 
     // swiftlint:disable:next missing_docs
@@ -122,10 +124,13 @@ public struct PaywallView: View {
                                      activelySubscribedProductIdentifiers: customerInfo.activeSubscriptions,
                                      fonts: self.fonts,
                                      checker: self.introEligibility,
-                                     purchaseHandler: self.purchaseHandler)
+                                     purchaseHandler: self.purchaseHandler,
+                                     shouldAutomaticallyDismiss: self.shouldAutomaticallyDismiss)
                     .transition(Self.transition)
                 } else {
-                    LoadingPaywallView(mode: self.mode, displayCloseButton: self.displayCloseButton)
+                    LoadingPaywallView(mode: self.mode,
+                                       displayCloseButton: self.displayCloseButton,
+                                       shouldAutomaticallyDismiss: self.shouldAutomaticallyDismiss)
                         .transition(Self.transition)
                         .task {
                             do {
@@ -157,7 +162,8 @@ public struct PaywallView: View {
         activelySubscribedProductIdentifiers: Set<String>,
         fonts: PaywallFontProvider,
         checker: TrialOrIntroEligibilityChecker,
-        purchaseHandler: PurchaseHandler
+        purchaseHandler: PurchaseHandler,
+        shouldAutomaticallyDismiss: Bool
     ) -> some View {
         let (paywall, template, error) = offering.validatedPaywall(locale: self.locale)
 
@@ -170,7 +176,8 @@ public struct PaywallView: View {
             fonts: fonts,
             displayCloseButton: self.displayCloseButton,
             introEligibility: checker,
-            purchaseHandler: purchaseHandler
+            purchaseHandler: purchaseHandler,
+            shouldAutomaticallyDismiss: shouldAutomaticallyDismiss
         )
 
         if let error {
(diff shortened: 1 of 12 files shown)


RevenueCatUI/PaywallView.swift
```diff
@@ -268,6 +277,9 @@ struct LoadedOfferingPaywallView: View {
     @Environment(\.locale)
     private var locale
 
+    @Environment(\.onRequestedDismissal)
+    private var onRequestedDismissal: (() -> Void)?
```
tonidero, line 281: Hmm maybe I'm wrong but I think we could make this not nullable now? (We always pass a value I think)

tonidero (approved):
LGTM!

RevenueCatUI/PaywallView.swift
vegaro: there will only be a value if `onRequestedDismissal` on PaywallView is called, and it's not even a public function (but PaywallView is), so this needs to be nullable

tonidero: Ohh ok makes sense

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -281,6 +300,10 @@ private extension PaywallViewController {
                 guard let self else { return }
                 self.delegate?.paywallViewController?(self, didFailRestoringWith: error)
             },
+            requestedDismissal: { [weak self] in
```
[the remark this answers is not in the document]
joshdholtz: I think I agree with this too ☝️ We should only set this if the delegate is set? Or... manually dismiss this if the delegate isn't set.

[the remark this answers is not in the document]
vegaro: yes, you both are right. Made some changes in https://github.com/RevenueCat/purchases-ios/pull/3738/commits/9efc7252f4b231b4a67b5296c3dbc273ee497588. Let me know if they make sense. It looks like I broke something though, but looking into it.

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -27,6 +27,10 @@ public class PaywallViewController: UIViewController {
     /// See ``PaywallViewControllerDelegate`` for receiving purchase events.
     @objc public final weak var delegate: PaywallViewControllerDelegate?
 
+    /// See ``PaywallViewControllerRequestedDismissalDelegate`` for dismissing the paywall. 
+    /// If this is not set, the paywall will close itself automatically after a successful purchase.
+    @objc public final weak var dismissalDelegate: PaywallViewControllerDismissalDelegate?
```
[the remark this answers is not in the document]
vegaro: done!

RevenueCatUI/UIKit/PaywallFooterViewController.swift
```diff
@@ -34,15 +34,17 @@ public final class PaywallFooterViewController: PaywallViewController {
     public init(offering: Offering? = nil) {
         super.init(content: .optionalOffering(offering),
                    fonts: DefaultPaywallFontProvider(),
-                   displayCloseButton: false)
+                   displayCloseButton: false,
+                   dismissalRequest: nil)
```
tonidero, line 38: Hmm I think we should still accept a dismissalRequest parameter for the footer, since there might be a successful purchase? Alternatively, we can just devs use the purchaseCompleted callback, which would be called in the same location... But not sure if we want to be consistent with that behavior between the full screen and footer.

RevenueCatUI/UIKit/PaywallFooterViewController.swift
```diff
@@ -51,18 +53,21 @@ public final class PaywallFooterViewController: PaywallViewController {
...
     }
 
     @available(*, unavailable)
     override init(
         content: PaywallViewConfiguration.Content,
         fonts: PaywallFontProvider,
-        displayCloseButton: Bool = false
+        displayCloseButton: Bool = false,
+        dismissalRequest: ((_ controller: PaywallViewController) -> Void)? = nil
```
tonidero, line 65: Hmm while I think we need to add it to the other constructors here, why do we add it to the unavailable constructor?

tonidero (commented):
Just that question about the footer. Otherwise looks good!

RevenueCatUI/UIKit/PaywallFooterViewController.swift
vegaro: oh actually, this one is overriden from `PaywallViewController`, so it's required :/

RevenueCatUI/UIKit/PaywallFooterViewController.swift
vegaro: hmmm, yes. I think it makes sense, I will add it to `PaywallFooterViewController`

vegaro: Ok... so I added it, but the callback is not being called because there's this code in `PaywallView`

```
              if self.mode.isFullScreen, purchased {
                    Logger.debug(Strings.dismissing_paywall)
                    guard let onRequestedDismissal = self.onRequestedDismissal else {
                        self.dismiss()
                        return
                    }
                    onRequestedDismissal()
                }
```

I am going to see what happens in Android... but if we don't call the `onDismiss` in Android, I think we shouldn't do it here either

vegaro:
@joshdholtz can you take a look? I can't merge until I get your approval 😄

joshdholtz (approved):
LGTM 💪

PR #3739, opened 2024-02-29
 Mark methods with StaticString for appUserID as deprecated

MarkVillacampa:


Sources/Purchasing/Configuration.swift
```diff
@@ -123,6 +123,11 @@ import Foundation
             return self
         }
 
+        @available(*, deprecated, message: """
+        The appUserID passed to logIn is a constant string known at compile time.
+        This is likely a programmer error. This ID is used to identify the current user.
+        See https://docs.revenuecat.com/docs/user-ids for more information.
```
MarkVillacampa, line 129: I wish there was a way to not repeat these strings but I have not found one :/

vegaro: what was the behavior before this? After this, it will add a warning right?

MarkVillacampa: Before this, we were only logging a warning int he console. Now, it will show a compilation warning in the IDE.

aboedo (approved):

PR #3751, opened 2024-03-06
Fix for passing `TargetingContext` when using `currentOffering(forPlacement:)`

joshdholtz:
### Motivation

Targeting context was not being passed when getting current offering for placement

### Description

- Passing `Offerings.Targeting` when copying `Offering` with a placement identifier
- Added test

### DIFF
--- a/Sources/Purchasing/Offerings.swift
+++ b/Sources/Purchasing/Offerings.swift
@@ -131,7 +131,8 @@ public extension Offerings {
             returnOffering =  placements.fallbackOfferingId.flatMap { self.all[$0]}
         }
 
-        return returnOffering?.copyWith(placementIdentifier: placementIdentifier)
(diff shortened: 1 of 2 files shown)


Sources/Purchasing/Offerings.swift
```diff
@@ -131,7 +131,8 @@ public extension Offerings {
             returnOffering =  placements.fallbackOfferingId.flatMap { self.all[$0]}
         }
 
-        return returnOffering?.copyWith(placementIdentifier: placementIdentifier)
+        return returnOffering?.copyWith(placementIdentifier: placementIdentifier, 
+                                        targeting: self.targeting)
```
tonidero, line 135: Ohh right... it would be nice if we had something like the copy constructor, that uses the existing values by default unless passed as parameter. We could write it ourselves, but this is ok for now I think.

tonidero (approved):

PR #3757, opened 2024-03-08
Add auto sync of custom attributes with `CustomAttributesManager`

joshdholtz:
### Motivation

Auto sync custom attributes and fetch new offerings

### Decisions (Please read 🙏)

If the attribute being set is a custom attribute, this change will sync attributes while the app is in the foreground.

This _could_ be moved to when the app is foreground/background but this current approach was decided because:
- The design of the get offerings operation doesn't allow it to be run while the app is in the background.
- This is a better experience for the developer and customer

Some potential drawbacks:
- It could result in significantly more subscriber attribute posts and offerings refreshes

Analyzing and mitigating this behavior:
- We will send headers in these requests so we can monitor the reasons for sync and refreshes and see if its a significant more amount
- We can always turn off the custom attributes from being included in the offerings response to stop this new behavior all together if needed

### Description

Use a list of subscriber attributes in the offerings response to auto sync specific attributes after they are set. This is used to provide a current offerings response that uses custom attributes in targeting. 

- Added custom attribute information into `Offerings.Targeting`
- New `CustomAttributesManager` that handles all custom attributes syncing and logic
  - Manual `startManualSync()` (called by `Purchases.shared.syncAttributesAndOfferingsIfNeeded()`)
  - Automatic `startAutomaticSync` (called when an attribute is updated)
- New delegate method on `SubscriberAttributesManager` when attributes are set
  - `Attribution` listens to this and will then call `startAutomaticSync()`
- `GetOfferingsOperation` adds additional header called `X-RC-Offerings-Refresh-Reason`
  - This will allow us to determine how often and which form of the refresh method gets called

#### HTTP Request and Client Changes
- `HTTPRequest` has new `additionalHeaders` property
- `HTTPClient` will send these on the request
  - Adds these to the `defaultHeaders` for signature signing

### Todo
- [ ] Add `X-RC-Sync-Attributes-Reason` header to post attributes (?????)
- [ ] More tests
- [ ] Verify that we need salt and/or hash

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPRequest.swift
+++ b/Sources/Networking/HTTPClient/HTTPRequest.swift
@@ -22,22 +22,24 @@ struct HTTPRequest {
     var path: HTTPRequestPath
     /// If present, this will be used by the server to compute a checksum of the response signed with a private key.
     var nonce: Data?
+    var additionalHeaders: Headers
 
-    init(method: Method, path: HTTPRequest.Path, nonce: Data? = nil) {
-        self.init(method: method, requestPath: path, nonce: nonce)
+    init(method: Method, path: HTTPRequest.Path, nonce: Data? = nil, additionalHeaders: Headers = [:]) {
+        self.init(method: method, requestPath: path, nonce: nonce, additionalHeaders: additionalHeaders)
     }
 
-    init(method: Method, path: HTTPRequest.PaywallPath, nonce: Data? = nil) {
-        self.init(method: method, requestPath: path, nonce: nonce)
+    init(method: Method, path: HTTPRequest.PaywallPath, nonce: Data? = nil, additionalHeaders: Headers = [:]) {
+        self.init(method: method, requestPath: path, nonce: nonce, additionalHeaders: additionalHeaders)
     }
 
-    private init(method: Method, requestPath: HTTPRequestPath, nonce: Data? = nil) {
+    private init(method: Method, requestPath: HTTPRequestPath, nonce: Data? = nil, additionalHeaders: Headers = [:]) {
         assert(nonce == nil || nonce?.count == Data.nonceLength,
                "Invalid nonce: \(nonce?.description ?? "")")
 
         self.method = method
         self.path = requestPath
         self.nonce = nonce
+        self.additionalHeaders = additionalHeaders
     }
 
 }
--- a/Sources/Networking/Operations/GetOfferingsOperation.swift
+++ b/Sources/Networking/Operations/GetOfferingsOperation.swift
@@ -16,15 +16,18 @@ import Foundation
 final class GetOfferingsOperation: CacheableNetworkOperation {
 
     private let offeringsCallbackCache: CallbackCache<OfferingsCallback>
+    private let fetchReason: String?
     private let configuration: AppUserConfiguration
 
     static func createFactory(
         configuration: UserSpecificConfiguration,
+        fetchReason: String?,
         offeringsCallbackCache: CallbackCache<OfferingsCallback>
     ) -> CacheableNetworkOperationFactory<GetOfferingsOperation> {
         return .init({ cacheKey in
                     .init(
                         configuration: configuration,
+                        fetchReason: fetchReason,
                         offeringsCallbackCache: offeringsCallbackCache,
                         cacheKey: cacheKey
                     )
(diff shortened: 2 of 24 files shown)


Sources/Networking/HTTPClient/HTTPRequest.swift
```diff
@@ -22,22 +22,24 @@ struct HTTPRequest {
...
 
-    private init(method: Method, requestPath: HTTPRequestPath, nonce: Data? = nil) {
+    private init(method: Method, requestPath: HTTPRequestPath, nonce: Data? = nil, additionalHeaders: Headers = [:]) {
         assert(nonce == nil || nonce?.count == Data.nonceLength,
                "Invalid nonce: \(nonce?.description ?? "")")
 
         self.method = method
         self.path = requestPath
         self.nonce = nonce
+        self.additionalHeaders = additionalHeaders
```
joshdholtz, line 42: This allows `getOfferings()` to send up a specific header for refresh reason

Sources/Networking/Operations/GetOfferingsOperation.swift
```diff
@@ -61,7 +66,14 @@ private extension GetOfferingsOperation {
             return
         }
 
-        let request = HTTPRequest(method: .get, path: .getOfferings(appUserID: appUserID))
+        var additionalHeaders: HTTPRequest.Headers = [:]
+        if let fetchReason = self.fetchReason {
+            additionalHeaders["X-RC-Offerings-Refresh-Reason"] = fetchReason
```
joshdholtz, line 71: Unsure about this naming? 🤷‍♂️

Sources/Purchasing/Offerings.swift
```diff
@@ -59,7 +68,7 @@ import Foundation
 
     private let currentOfferingID: String?
     private let placements: Placements?
-    private let targeting: Targeting?
+    internal let targeting: Targeting?
```
joshdholtz, line 71: Changing to internal so the custom attribute keys can be used in `CustomAttributesManager`

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -289,6 +301,32 @@ extension OfferingsManager {
...
+            case .fetchCurrent(reason: let reason):
+                return reason.rawValue
+            }
+        }
+    }
+
+    internal enum FetchReason: String {
+        case manualSyncCustomAttributes = "custom_attributes_sync_manual"
+        case automaticSyncCustomAttributes = "custom_attributes_sync_automatic"
+    }
```
joshdholtz, line 328: Replaced our boolean with a enum for fetching behavior. Don't know if I love the name `FetchBehavior` (it differs from our customer one) but it felt different enough that our existing `FetchPolicy` in this same file.

`fetchCurrent` takes a `FetchReason` because we want to know why we are going straight for new offerings

Sources/SubscriberAttributes/CustomAttributesManager.swift
joshdholtz, line 1: This whole file could use rework to make prettier but its at least isolated right now 🤷‍♂️

Sources/SubscriberAttributes/CustomAttributesManager.swift
```diff
@@ -0,0 +1,161 @@
...
+
+    func throttle(_ block: @escaping () -> Void) {
+        workItem?.cancel()
+
+        let task = DispatchWorkItem { block() }
+        self.workItem = task
+
+        queue.asyncAfter(deadline: .now() + delayInSeconds, execute: task)
+    }
+}
```
joshdholtz, line 109: This could be moved into its own file where `RateLimiter` is. Not sure if this is the best approach for this (or if we have something existing that I could use) but it wanted to keep it simple at first

Sources/SubscriberAttributes/CustomAttributesManager.swift
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        let unsyncedKeys = attribution.unsyncedAttributesByKey(appUserID: appUserID).keys
+        guard !unsyncedKeys.isEmpty else {
+            return false
+        }
+
+        let saltedAndHashedKeys = Set(unsyncedKeys.map { key in
+            return "\(customAttributes.salt)\(key)".asData.hashString
+        })
```
joshdholtz, line 128: Salt _might_ not be needed so stay tuned

joshdholtz:
Closing for now

PR #3759, opened 2024-03-12
Pin xcbeautify version for xcode 14 tests

vegaro:
Xcode 14 tests broke with xcbeautify 2.0.0, which needs xcode 15. Earlier versions of xcbeautify are not available in homebrew, so we need to install it using mint


tonidero (approved):
Makes sense!

PR #3773, opened 2024-03-22
Add `RC_BILLING` store

tonidero:
### Description
iOS part of adding `RC_BILLING` store to the mobile SDKs

### DIFF
--- a/Sources/Purchasing/EntitlementInfo.swift
+++ b/Sources/Purchasing/EntitlementInfo.swift
@@ -40,6 +40,9 @@ import Foundation
     /// For entitlements granted via the Amazon Store.
     @objc(RCAmazon) case amazon = 6
 
+    /// For entitlements granted via RC Billing
+    @objc(RCRCBilling) case rcBilling = 7
+
 }
 
 extension Store: CaseIterable {}
--- a/Sources/Purchasing/EntitlementInfo.swift
+++ b/Sources/Purchasing/EntitlementInfo.swift
@@ -41,7 +41,7 @@ import Foundation
     @objc(RCAmazon) case amazon = 6
 
     /// For entitlements granted via RC Billing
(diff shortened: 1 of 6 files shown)


Sources/Purchasing/EntitlementInfo.swift
```diff
@@ -40,6 +40,9 @@ import Foundation
     /// For entitlements granted via the Amazon Store.
     @objc(RCAmazon) case amazon = 6
 
+    /// For entitlements granted via RC Billing
+    @objc(RCRCBilling) case rcBilling = 7
```
tonidero, line 44: I wasn't really sure about the ObjC name, but I didn't want to call it `RCBilling` thoughts?

tonidero: Hmm well... I guess as long as we don't have any other `Billing` stores in the future, it should be ok 😅. Will rename it.

joshdholtz: Ha uh... I do agree that `RCBilling` is _probably_ the right way to go even though it feels weird because different pattern 🤷‍♂️

joshdholtz (approved):
👍

Kaspik:
Can we get some info into documentation what `rcBilling` actually is? Just seeing warnings for not-exhaustive switch and have no idea where to categorize that.

tonidero:
> Can we get some info into documentation what `rcBilling` actually is? Just seeing warnings for not-exhaustive switch and have no idea where to categorize that.

Hi @Kaspik,

RCBilling is the new web subscription service provided by RC (currently on beta). You can check more information in our blogpost: https://www.revenuecat.com/blog/company/introducing-revenuecat-billing/. So this new store will be returned if a purchase is done through this new web subscription service and using the same account in your mobile apps.

Please, let us know if you have any more questions!

PR #3772, opened 2024-03-19
[EXTERNAL] Cocoapods support for privacy manifest

sdurban:
### Motivation
The library already includes Privacy.xcprivacy in SPM, but when used with CocoaPods (mandatory in Flutter), these files are not exported as the podspec does not include that file as a resource bundle.

### Description
Listing the PrivacyInfo.xcprivacy file inside the resource_bundles specification in the podspec allows to distribute the file correctly. Ref. https://github.com/CocoaPods/CocoaPods/issues/10325#issuecomment-1718723762

This GitHub issue contains the entire discussion on how to add these xcprivacy files and is referenced in multiple places; many libraries already follow this approach:

https://github.com/firebase/firebase-ios-sdk/blob/main/FirebaseCrashlytics.podspec
https://github.com/SDWebImage/SDWebImage/pull/3649
https://github.com/flutter/packages/pull/5846
https://github.com/Baseflow/flutter-permission-handler/pull/1291
https://github.com/Baseflow/flutter-geolocator/pull/1462

Thank you guys for Revenuecat ❤️ !

PR #3775, opened 2024-03-26
[EXTERNAL] Cocoapods support for privacy manifest (#3772) via @sdurban

aboedo:
### Motivation
The library already includes Privacy.xcprivacy in SPM, but when used with CocoaPods (mandatory in Flutter), these files are not exported as the podspec does not include that file as a resource bundle.

### Description
Listing the PrivacyInfo.xcprivacy file inside the resource_bundles specification in the podspec allows to distribute the file correctly. Ref.
https://github.com/CocoaPods/CocoaPods/issues/10325#issuecomment-1718723762

This GitHub issue contains the entire discussion on how to add these xcprivacy files and is referenced in multiple places; many libraries already follow this approach:

https://github.com/firebase/firebase-ios-sdk/blob/main/FirebaseCrashlytics.podspec https://github.com/SDWebImage/SDWebImage/pull/3649 https://github.com/flutter/packages/pull/5846
https://github.com/Baseflow/flutter-permission-handler/pull/1291 https://github.com/Baseflow/flutter-geolocator/pull/1462

### DIFF
--- a/RevenueCat.podspec
+++ b/RevenueCat.podspec
@@ -31,5 +31,5 @@ Pod::Spec.new do |s|
   s.source_files = 'Sources/**/*.swift'
   s.exclude_files = 'Sources/LocalReceiptParsing/ReceiptParser-only-files/**'
   
-  
+  s.resource_bundles = {'RevenueCat' => ['Sources/PrivacyInfo.xcprivacy']}
 end


tonidero (approved):

RevenueCat.podspec
```diff
@@ -31,5 +31,5 @@ Pod::Spec.new do |s|
   s.source_files = 'Sources/**/*.swift'
   s.exclude_files = 'Sources/LocalReceiptParsing/ReceiptParser-only-files/**'
   
-  
+  s.resource_bundles = {'RevenueCat' => ['Sources/PrivacyInfo.xcprivacy']}
```
tonidero, line 34: Hmm I was wondering if we also need this for `RevenueCatUI` but I think it shouldn't be needed, since `RevenueCat` is also a dependency for `RevenueCatUI` right?

tonidero (approved):

RevenueCat.podspec
aboedo: Yeah, that's my understanding. Having it only in `RevenueCat` would also match our SPM setup. 
In any case even if we did need to add it we could do it as a separate PR, it doesn't really take value away from merging this one

aboedo:
given that the SPM test that failed is unrelated to this PR, I'm gonna go ahead and merge this one

aboedo:
oh, turns out I can't force merge and I can't re-run tests either. Taking this to slack

PR #3781, opened 2024-04-01
iOS append events to JSONL file and get diagnostics events

vegaro:
Based off https://github.com/RevenueCat/purchases-ios/pull/2672/

Equivalent to https://github.com/RevenueCat/purchases-android/pull/784/

Adds DiagnosticsEntry classes and implements adding an entry to the file, getting all entries, and deleting X first entries

### DIFF
--- a/Sources/Diagnostics/DiagnosticsEntry.swift
+++ b/Sources/Diagnostics/DiagnosticsEntry.swift
@@ -14,8 +14,10 @@
 import Foundation
 
 protocol DiagnosticsEntry: Codable, Equatable {
+
     var diagnosticType: String { get }
     var version: Int { get }
+
 }
 
 extension DiagnosticsEntry {
@@ -32,6 +34,7 @@ extension DiagnosticsEntry {
 }
 
 struct DiagnosticsEvent: DiagnosticsEntry {
+    
     var diagnosticType: String = "event"
     let name: String
     let properties: [String: AnyEncodable]
@@ -41,18 +44,22 @@ struct DiagnosticsEvent: DiagnosticsEntry {
         case diagnosticType = "type"
         case name, properties, timestamp
     }
+    
 }
 
 extension DiagnosticsEvent {
+
     static func == (lhs: DiagnosticsEvent, rhs: DiagnosticsEvent) -> Bool {
         return lhs.diagnosticType == rhs.diagnosticType &&
                lhs.name == rhs.name &&
                lhs.properties == rhs.properties &&
                lhs.timestamp == rhs.timestamp
     }
+
 }
 
 struct Counter: DiagnosticsEntry {
+
     var diagnosticType: String = "counter"
     let name: String
     let tags: [String: String]
@@ -62,9 +69,11 @@ struct Counter: DiagnosticsEntry {
         case diagnosticType = "type"
         case name, tags, value
     }
+
 }
 
 struct Histogram: DiagnosticsEntry {
+
     var diagnosticType: String = "histogram"
     let name: String
     let tags: [String: String]
@@ -74,4 +83,5 @@ struct Histogram: DiagnosticsEntry {
         case diagnosticType = "type"
         case name, tags, values
     }
+    
 }
--- a/Sources/Diagnostics/DiagnosticsEntry.swift
+++ b/Sources/Diagnostics/DiagnosticsEntry.swift
@@ -13,24 +13,15 @@
 
 import Foundation
 
-protocol DiagnosticsEntry: Codable, Equatable {
-
-    var diagnosticType: String { get }
-    var version: Int { get }
-
-}
-
-struct DiagnosticsEvent: DiagnosticsEntry {
+struct DiagnosticsEvent: Codable, Equatable {
 
     let version: Int = 1
-    let diagnosticType: String = "event"
     let name: String
     let properties: [String: AnyEncodable]
     let timestamp: Date
 
     enum CodingKeys: String, CodingKey {
-        case diagnosticType = "type"
-        case name, properties, timestamp, version
+        case version, name, properties, timestamp
     }
 
 }
@@ -38,7 +29,7 @@ struct DiagnosticsEvent: DiagnosticsEntry {
 extension DiagnosticsEvent {
 
     static func == (lhs: DiagnosticsEvent, rhs: DiagnosticsEvent) -> Bool {
-        return lhs.diagnosticType == rhs.diagnosticType &&
+        return lhs.version == rhs.version &&
                lhs.name == rhs.name &&
                lhs.properties == rhs.properties &&
                lhs.timestamp == rhs.timestamp
--- a/Sources/Diagnostics/DiagnosticsEntry.swift
+++ b/Sources/Diagnostics/DiagnosticsEntry.swift
@@ -24,7 +24,7 @@ extension DiagnosticsEntry {
     func toJSONString() -> String? {
         let encoder = JSONEncoder()
         encoder.keyEncodingStrategy = .convertToSnakeCase
-        encoder.dateEncodingStrategy = .millisecondsSince1970 // Example date encoding strategy
+        encoder.dateEncodingStrategy = .iso8601
 
         guard let jsonData = try? encoder.encode(self) else { return nil }
         return String(data: jsonData, encoding: .utf8)
--- a/Sources/Diagnostics/DiagnosticsEntry.swift
+++ b/Sources/Diagnostics/DiagnosticsEntry.swift
@@ -21,16 +21,9 @@ protocol DiagnosticsEntry: Codable, Equatable {
 }
 
 extension DiagnosticsEntry {
+
     var version: Int { 1 }
 
-    func toJSONString() -> String? {
-        let encoder = JSONEncoder()
-        encoder.keyEncodingStrategy = .convertToSnakeCase
-        encoder.dateEncodingStrategy = .iso8601
-
-        guard let jsonData = try? encoder.encode(self) else { return nil }
-        return String(data: jsonData, encoding: .utf8)
-    }
 }
 
 struct DiagnosticsEvent: DiagnosticsEntry {
--- a/Sources/Diagnostics/DiagnosticsEntry.swift
+++ b/Sources/Diagnostics/DiagnosticsEntry.swift
@@ -13,7 +13,7 @@
 
 import Foundation
 
-protocol DiagnosticsEntry: Codable {
+protocol DiagnosticsEntry: Codable, Equatable {
     var diagnosticType: String { get }
     var version: Int { get }
 }
@@ -40,6 +40,15 @@ struct DiagnosticsEvent: DiagnosticsEntry {
     enum CodingKeys: String, CodingKey {
         case diagnosticType = "type"
         case name, properties, timestamp
+    }
+}
+
+extension DiagnosticsEvent {
+    static func == (lhs: DiagnosticsEvent, rhs: DiagnosticsEvent) -> Bool {
+        return lhs.diagnosticType == rhs.diagnosticType &&
+               lhs.name == rhs.name &&
+               lhs.properties == rhs.properties &&
+               lhs.timestamp == rhs.timestamp
     }
 }
 
--- a/Sources/Diagnostics/DiagnosticsEntry.swift
+++ b/Sources/Diagnostics/DiagnosticsEntry.swift
@@ -27,7 +27,7 @@ extension DiagnosticsEntry {
 }
 
 struct DiagnosticsEvent: DiagnosticsEntry {
-    
+
     var diagnosticType: String = "event"
     let name: String
     let properties: [String: AnyEncodable]
@@ -37,7 +37,7 @@ struct DiagnosticsEvent: DiagnosticsEntry {
         case diagnosticType = "type"
         case name, properties, timestamp
     }
-    
+
 }
 
 extension DiagnosticsEvent {
@@ -76,5 +76,5 @@ struct Histogram: DiagnosticsEntry {
         case diagnosticType = "type"
         case name, tags, values
     }
-    
+
 }
--- a/Sources/Diagnostics/DiagnosticsEntry.swift
+++ b/Sources/Diagnostics/DiagnosticsEntry.swift
@@ -45,33 +45,3 @@ extension DiagnosticsEvent {
     }
 
 }
-
-struct Counter: DiagnosticsEntry {
-
-    let version: Int = 1
-    let diagnosticType: String = "counter"
-    let name: String
-    let tags: [String: String]
-    let value: Int
-
-    enum CodingKeys: String, CodingKey {
-        case diagnosticType = "type"
-        case name, tags, value, version
-    }
-
-}
-
-struct Histogram: DiagnosticsEntry {
-
-    let version: Int = 1
-    let diagnosticType: String = "histogram"
-    let name: String
-    let tags: [String: String]
-    let values: [Double]
-
-    enum CodingKeys: String, CodingKey {
-        case diagnosticType = "type"
-        case name, tags, values, version
-    }
-
-}
--- a/Sources/Diagnostics/DiagnosticsEntry.swift
+++ b/Sources/Diagnostics/DiagnosticsEntry.swift
@@ -20,22 +20,17 @@ protocol DiagnosticsEntry: Codable, Equatable {
 
 }
 
-extension DiagnosticsEntry {
-
-    var version: Int { 1 }
-
-}
-
 struct DiagnosticsEvent: DiagnosticsEntry {
 
-    var diagnosticType: String = "event"
+    let version: Int = 1
+    let diagnosticType: String = "event"
     let name: String
     let properties: [String: AnyEncodable]
     let timestamp: Date
 
     enum CodingKeys: String, CodingKey {
         case diagnosticType = "type"
-        case name, properties, timestamp
+        case name, properties, timestamp, version
     }
 
 }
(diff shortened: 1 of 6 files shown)


Tests/UnitTests/Diagnostics/DiagnosticsFileHandlerTests.swift
```diff
@@ -0,0 +1,134 @@
...
+//    func testAppendEventWithProperties() async throws {
+//        let content = DiagnosticsEvent(name: "HTTP_REQUEST_PERFORMED",
+//                                       properties: ["key": AnyEncodable("value")],
+//                                       timestamp: Date())
+//
+//        await self.handler.appendEvent(diagnosticsEvent: content)
+//
+//        let entries = await self.handler.getEntries()
+//        expect(entries[0]).to(equal(content))
+//    }
```
vegaro, line 56: Been spending a good amount of time trying to fix this and I don't know what's up but it's erroring comparing the dates:

```
failed - expected to equal <DiagnosticsEvent(diagnosticType: "event", name: "HTTP_REQUEST_PERFORMED", properties: ["key": RevenueCat.AnyEncodable(value: "value")], timestamp: 2024-04-03 10:53:19 +0000)>, got <DiagnosticsEvent(diagnosticType: "event", name: "HTTP_REQUEST_PERFORMED", properties: ["key": RevenueCat.AnyEncodable(value: "value")], timestamp: 2024-04-03 10:53:19 +0000)>
```

I removed the timestamp from the comparison and it started to work. I think it's related to:

```
The issue with the timestamp comparison you're encountering in your test case is a common problem when dealing with dates in unit tests, especially when serialization and deserialization are involved. The problem arises because when you serialize and then deserialize a Date, the resulting Date might not be equal to the original due to precision loss or differences in the date encoding and decoding strategies.
```

tonidero: Hmm can we encode the timestamp as milliseconds since epoch? (That's how we have it in android right now). Not sure if that would help here.

vegaro: I was doing that and it was encoding it with decimals `1712142586558.343` that's why I changed to iso. I think the decimals makes it clear that it's some precision issue

Sources/Diagnostics/DiagnosticsEntry.swift
```diff
@@ -0,0 +1,77 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  DiagnosticsEntry.swift
+//
+//  Created by Cesar de la Vega on 1/4/24.
+
+import Foundation
+
+protocol DiagnosticsEntry: Codable, Equatable {
+    var diagnosticType: String { get }
```
aboedo, line 17: let's add line breaks after type declarations and at the end of them
https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/SwiftStyleGuide.swift#L18

Sources/Diagnostics/DiagnosticsEntry.swift
```diff
@@ -0,0 +1,77 @@
...
+    var version: Int { 1 }
+
+    func toJSONString() -> String? {
+        let encoder = JSONEncoder()
+        encoder.keyEncodingStrategy = .convertToSnakeCase
+        encoder.dateEncodingStrategy = .iso8601
+
+        guard let jsonData = try? encoder.encode(self) else { return nil }
+        return String(data: jsonData, encoding: .utf8)
+    }
```
aboedo, line 31: you can simplify this by using `JSONEncoder.default.encode` or even `JSONEncoder.prettyPrinted.encode`. 
Declaration here: https://github.com/RevenueCat/purchases-ios/blob/main/Sources/LocalReceiptParsing/DataConverters/Codable%2BExtensions.swift#L52
Example usage: https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Identity/CustomerInfoManager.swift#L192

Or even better, since `DiagnosticsEntry` is `Codable`, you can leverage the other extension to directly do stuff like calling `prettyPrintedJSON` and `encodedJSON` on the entry itself, i.e.: `let myJSONString = entry.encodedJSON`

Sources/Diagnostics/DiagnosticsFileHandler.swift
```diff
@@ -0,0 +1,137 @@
...
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  DiagnosticsFileHandler.swift
+//
+//  Created by Nacho Soto on 6/16/23.
```
aboedo, line 12: lol
I love that Nacho is still adding RC code

Sources/Diagnostics/DiagnosticsFileHandler.swift
```diff
@@ -0,0 +1,137 @@
...
+            return nil
+        }
+    }
+
+    init(_ fileHandler: FileHandler) {
+        self.fileHandler = fileHandler
+    }
+
+    func appendEvent(diagnosticsEvent: DiagnosticsEvent) async {
+        guard let jsonString = diagnosticsEvent.toJSONString() else {
```
aboedo, line 45: I imagine this was probably written before we had the Codable extensions, here's one instance where you could just make it `diagnosticsEvent.prettyPrintedJSON` or `diagnosticsEvent.encodedJSON`

Sources/Diagnostics/DiagnosticsFileHandler.swift
```diff
@@ -0,0 +1,137 @@
...
+        decoder.dateDecodingStrategy = .iso8601
+
+        do {
+            guard let data = line.data(using: .utf8) else { return nil }
+            let event = try decoder.decode(DiagnosticsEvent.self, from: data)
+            return event
+        } catch {
+            return nil
+        }
+    }
```
aboedo, line 136: nitpick: add extra line break at end of type declaration

aboedo, line 136: we can probably kill this whole bit in favor of `JSONEncoder.default`

Tests/UnitTests/Diagnostics/DiagnosticsFileHandlerTests.swift
aboedo: I'm not sure that's it - I think there might be some data precision getting lost in the encoding / decoding. 

For example, if you do `expect(try content.jsonEncodedData).to(equal(try content.jsonEncodedData))`, that actually works. So something between appending and getting the entries is losing precision

aboedo: yeah, that'll be it - we're comparing the timestamp of the DiagnosticsEvent post-encoding (`entries[0].timestamp`) vs pre-encoding `content.timestamp`, and there'll be precision loss because the encoder doesn't take nanoseconds

Sources/Misc/Codable/AnyEncodable.swift
```diff
@@ -109,4 +109,22 @@ extension AnyEncodable: Decodable {
 
 }
 
+extension AnyEncodable: Equatable {
+    static func == (lhs: AnyEncodable, rhs: AnyEncodable) -> Bool {
+        let lhsEncoder = JSONEncoder()
+        let rhsEncoder = JSONEncoder()
+
+        lhsEncoder.dateEncodingStrategy = .iso8601
+        rhsEncoder.dateEncodingStrategy = .iso8601
```
aboedo, line 118: we can simplify this bit by just doing something like 
```swift
        do {
            let lhsData = try lhs.encodedJSON
            let rhsData = try rhs.encodedJSON
            return lhsData == rhsData
            return lhsData == rhsData
        } catch {
            return false
        }
```

Tests/UnitTests/Diagnostics/DiagnosticsFileHandlerTests.swift
aboedo: easiest fix: update this line to
```swift
        let encodedContent = try content.encodeAndDecode()
        expect(entries[0]).to(equal(encodedContent))
```

or even
```swift
        let encodedContent = try content.encodeAndDecode()
        expect(entries[0]) == encodedContent
```

vegaro: That indeed fixed it!!!! Thank you so much

Sources/Diagnostics/DiagnosticsEntry.swift
vegaro: ooh that's nice!

Sources/Diagnostics/DiagnosticsFileHandler.swift
vegaro: thanks! much better :)

Sources/Misc/Codable/AnyEncodable.swift
vegaro: so clean!

Sources/Diagnostics/DiagnosticsEntry.swift
```diff
@@ -0,0 +1,77 @@
...
+//
+//  DiagnosticsEntry.swift
+//
+//  Created by Cesar de la Vega on 1/4/24.
+
+import Foundation
+
+protocol DiagnosticsEntry: Codable, Equatable {
+
+    var diagnosticType: String { get }
```
tonidero, line 18: We can remove the type and counters/histograms now, that's not going to be used by the backend

Sources/Diagnostics/DiagnosticsEntry.swift
```diff
@@ -0,0 +1,77 @@
...
+    static func == (lhs: DiagnosticsEvent, rhs: DiagnosticsEvent) -> Bool {
+        return lhs.diagnosticType == rhs.diagnosticType &&
+               lhs.name == rhs.name &&
+               lhs.properties == rhs.properties &&
+               lhs.timestamp == rhs.timestamp
+    }
+
+}
+
+struct Counter: DiagnosticsEntry {
```
tonidero, line 49: As mentioned, we can remove this and the histogram

Sources/Diagnostics/DiagnosticsFileHandler.swift
```diff
@@ -0,0 +1,142 @@
...
+    func appendEvent(diagnosticsEvent: DiagnosticsEvent) async {
+        guard let jsonString = try? diagnosticsEvent.encodedJSON else {
+            Logger.error("Failed to serialize diagnostics event to JSON")
+            return
+        }
+
+        await fileHandler.append(line: jsonString)
+    }
+
+    func getEntries() async -> [DiagnosticsEvent] {
```
tonidero, line 56: I think this should be fine, but note that it shouldn't be necessarily needed to parse the entries back to a list of `DiagnosticsEntry`. We just need a list of jsons to send in the backend request. But I don't think this should be a performance issue, so we can do this 👍

Sources/Diagnostics/DiagnosticsFileHandler.swift
```diff
@@ -0,0 +1,142 @@
...
+            do {
+                for try await line in try await fileHandler.readLines() {
+                    if let event = decodeDiagnosticsEvent(from: line) {
+                        entries.append(event)
+                    }
+                }
+            } catch {
+                Logger.error("Failed to read lines from file: \(error.localizedDescription)")
+            }
+        } else {
```
tonidero, line 69: I wonder if we should just not support diagnostics in iOS < 15... I think it should be fine?

Tests/UnitTests/Diagnostics/DiagnosticsFileHandlerTests.swift
```diff
@@ -0,0 +1,210 @@
...
+    // MARK: - append
+
+    func testAppendEventWithProperties() async throws {
+        let content = DiagnosticsEvent(name: "HTTP_REQUEST_PERFORMED",
+                                       properties: ["key": AnyEncodable("value")],
+                                       timestamp: Date())
+
+        await self.handler.appendEvent(diagnosticsEvent: content)
+
+        let entries = await self.handler.getEntries()
```
tonidero, line 54: Small thing but should we check that no entries are returned before adding the event?

Tests/UnitTests/Diagnostics/DiagnosticsFileHandlerTests.swift
```diff
@@ -0,0 +1,210 @@
...
+        entries = await self.handler.getEntries()
+        expect(entries.count).to(equal(0))
+    }
+
+    // MARK: - getEntries
+
+    func testGetEntries() async throws {
+        let line1 =
+            "{\"properties\":{\"key\":\"value\"}," +
+            "\"timestamp\":\"2024-04-04T12:55:59Z\"," +
```
tonidero, line 84: We should try to sync up on this so we send the same values in iOS and Android... Should we send this in Android as well? I kinda like sending millis since epoch, though we lose timezone information, so I'm ok sending this if you prefer.

Tests/UnitTests/Diagnostics/DiagnosticsFileHandlerTests.swift
```diff
@@ -0,0 +1,210 @@
...
+        await self.handler.emptyDiagnosticsFile()
+
+        data = try await self.fileHandler.readFile()
+        expect(data).to(beEmpty())
+    }
+
+}
+
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+class ModernDiagnosticsFileHandlerTests: BaseDiagnosticsFileHandlerTests {
```
tonidero, line 142: Interesting name 😅. Maybe `iOS15DiagnosticsFileHandlerTests`? I know it affects other systems, but usually we follow iOS I think. This might not be needed if we just drop support for diagnostics in iOS < 15

tonidero (commented):
It looks good to me! Just left some comments. Only thing blocking I think is deciding the format for the timestamp property.

Sources/Diagnostics/DiagnosticsFileHandler.swift
aboedo: yeah, I'd be fine with that if there's any extra cost to supporting iOS < 15. If we already have the code I wouldn't remove it, but if we need to add anything else, we might as well cut it

Tests/UnitTests/Diagnostics/DiagnosticsFileHandlerTests.swift
```diff
@@ -0,0 +1,210 @@
...
+            "\"timestamp\":\"2024-04-04T12:55:59Z\"," +
+            "\"name\":\"HTTP_REQUEST_PERFORMED\"," +
+            "\"type\":\"event\"," +
+            "\"version\":1}"
+        let line2 =
+            "{\"properties\":{\"key\":\"value\"}," +
+            "\"timestamp\":\"2024-04-04T13:55:59Z\"," +
+            "\"name\":\"HTTP_REQUEST_PERFORMED\"," +
+            "\"type\":\"event\"," +
+            "\"version\":1}"
```
aboedo, line 162: If you don't want to deal with all of these escape chars and quotes, you can use triple quotes and clean this up: 

```suggestion
        let line1 = """
        {
          "properties": {"key": "value"},
          "timestamp": "2024-04-04T12:55:59Z",
          "name": "HTTP_REQUEST_PERFORMED",
          "type": "event",
          "version": 1
        }
        """

        let line2 = """
        {
          "properties": {"key": "value"},
          "timestamp": "2024-04-04T13:55:59Z",
          "name": "HTTP_REQUEST_PERFORMED",
          "type": "event",
          "version": 1
        }
        """
```

aboedo (approved):
looking great so far! Consider @tonidero's suggestions, and let's make sure that we send the same data on iOS vs android, but other than that it's ready to go

Tests/UnitTests/Diagnostics/DiagnosticsFileHandlerTests.swift
vegaro: I tried that but I think there is an issue when adding the lines to the JSONL file because the events are not decoded properly, so I rolled back to this. Haven't looked much into it though...

Sources/Diagnostics/DiagnosticsEntry.swift
vegaro: right, I had this from before and forgot to remove. Thanks for pointing that out

Tests/UnitTests/Diagnostics/DiagnosticsFileHandlerTests.swift
vegaro: This is what the default encoder encodes to. I am leaning more towards this since it's readable but not sure if we will ever need to read this directly

Sources/Diagnostics/DiagnosticsFileHandler.swift
vegaro: I see! Just noticed Android is also returning a stream of jsons

Tests/UnitTests/Diagnostics/DiagnosticsFileHandlerTests.swift
tonidero: I think that's ok, I can change it in the backend/Android 👍

Tests/UnitTests/Diagnostics/DiagnosticsFileHandlerTests.swift
vegaro: the lines had `\n` which were causing issues. I created a function to remove them and this now works. Thanks!

Tests/UnitTests/Diagnostics/DiagnosticsFileHandlerTests.swift
vegaro: Thanks!

Sources/Diagnostics/DiagnosticsEntry.swift
```diff
@@ -0,0 +1,47 @@
...
+//
+//  DiagnosticsEntry.swift
+//
+//  Created by Cesar de la Vega on 1/4/24.
+
+import Foundation
+
+protocol DiagnosticsEntry: Codable, Equatable {
+
+    var diagnosticType: String { get }
```
tonidero, line 18: This can still be removed I think?

tonidero (approved):

Sources/Diagnostics/DiagnosticsEntry.swift
vegaro: this is getting removed in the json we send to the backend right?

tonidero: Yup, that's correct, this property won't be read by the backend.

Sources/Diagnostics/DiagnosticsFileHandler.swift
vegaro: ok, I cut the support to 15, it will simplify the code, and it is probably overengineering

Sources/Diagnostics/DiagnosticsFileHandler.swift
NachoSoto: 👀

PR #3784, opened 2024-04-04
Create `DiagnosticsTracker`

vegaro:
Create `DiagnosticsTracker` and adds it to the `Purchases` initialization. I don't expose the `diagnosticsEnabled` yet

### DIFF
--- a/Sources/Diagnostics/DiagnosticsTracker.swift
+++ b/Sources/Diagnostics/DiagnosticsTracker.swift
@@ -13,14 +13,14 @@
 
 import Foundation
 
-protocol DiagnosticsTrackerType: AnyObject {
-    
+protocol DiagnosticsTrackerType {
+
     func track(_ event: DiagnosticsEvent) async
 
 }
 
-@available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
-final class DiagnosticsTracker {
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+final class DiagnosticsTracker: DiagnosticsTrackerType {
 
     private let diagnosticsFileHandler: DiagnosticsFileHandler
 
--- a/Sources/Diagnostics/DiagnosticsTracker.swift
+++ b/Sources/Diagnostics/DiagnosticsTracker.swift
@@ -1,0 +1,34 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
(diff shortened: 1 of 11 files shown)


Sources/Diagnostics/DiagnosticsTracker.swift
```diff
@@ -0,0 +1,35 @@
...
+final class DiagnosticsTracker: DiagnosticsTrackerType {
+
+    private let diagnosticsFileHandler: DiagnosticsFileHandler
+
+    init(diagnosticsFileHandler: DiagnosticsFileHandler) {
+        self.diagnosticsFileHandler = diagnosticsFileHandler
+    }
+
+    func track(_ event: DiagnosticsEvent) async {
+        await diagnosticsFileHandler.appendEvent(diagnosticsEvent: event)
```
tonidero, line 33: Hmm I know `DiagnosticsFileHandler` is an `actor`, but I think we still need to specify the thread here? We can do this in a separate PR though.

Sources/Purchasing/Configuration.swift
```diff
@@ -223,6 +226,18 @@ import Foundation
...
         }
 
+        /// Enabling diagnostics will send some performance and debugging information from the SDK to our servers.
+        /// Examples of this information include response times, cache hits or error codes.
+        /// This information will be anonymous so it can't be traced back to the end-user
+        /// 
+        /// Defaults to ``false``
+        ///
+        @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+        @objc internal func with(diagnosticsEnabled: Bool) -> Builder {
```
tonidero, line 236: Added new method as `internal` for now.

tonidero (approved):

Sources/Diagnostics/DiagnosticsTracker.swift
```diff
@@ -0,0 +1,35 @@
...
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+final class DiagnosticsTracker: DiagnosticsTrackerType {
+
+    private let diagnosticsFileHandler: DiagnosticsFileHandler
+
+    init(diagnosticsFileHandler: DiagnosticsFileHandler) {
+        self.diagnosticsFileHandler = diagnosticsFileHandler
+    }
+
+    func track(_ event: DiagnosticsEvent) async {
```
tonidero, line 32: We should also add some tests here, just in case. Will do that in follow-up PRs

PR #3788, opened 2024-04-09
Change `onChangeOf` to be internal

joshdholtz:
### Motivation

`onChangeOf` was accidentally made public in #3517 

### Description

- Change scope of `onChangeOf` from `public` to `internal`
- Added breaking change in v5 migration doc

### DIFF
--- a/RevenueCatUI/Modifiers/ViewExtensions.swift
+++ b/RevenueCatUI/Modifiers/ViewExtensions.swift
@@ -35,7 +35,7 @@ extension View {
     /// Wraps the 2 `onChange(of:)` implementations in iOS 17+ and below depending on what's available
     @inlinable
     @ViewBuilder
-    public func onChangeOf<V>(
+    internal func onChangeOf<V>(
         _ value: V,
         perform action: @escaping (_ newValue: V) -> Void
     ) -> some View where V: Equatable {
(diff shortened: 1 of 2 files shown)


RevenueCatUI/Modifiers/ViewExtensions.swift
```diff
@@ -35,7 +35,7 @@ extension View {
     /// Wraps the 2 `onChange(of:)` implementations in iOS 17+ and below depending on what's available
     @inlinable
     @ViewBuilder
-    public func onChangeOf<V>(
+    internal func onChangeOf<V>(
```
aboedo, line 38: I think you can also skip the `internal` here since it's the default, but I won't go against keeping it explicit

Sources/DocCDocumentation/DocCDocumentation.docc/V5_API_Migration_guide.md
```diff
@@ -83,3 +83,7 @@ The minimum targets have been raised to the folllowing:
 ## Release Assets
 
 Pre-built `.frameworks` are no longer included in releases, only `.xcframeworks`: https://github.com/RevenueCat/purchases-ios/pull/3582
+
+## Breaking Changes
+
+- The scope of `onChangeOf` is changed from `public` to `internal`
```
aboedo, line 89: 🤔 This would be something like: 
```suggestion
- The scope of the View extension`onChangeOf(value:perform:)` is changed from `public` to `internal`
```
right? 

Also, maybe we can also add a note stating that you most likely wouldn't have used it? Just wondering if we can save folks a few minutes of spelunking their codebase for usage of a thing they probably never touched

aboedo (approved):

PR #3789, opened 2024-04-10
Prevent Template 4 from wrapping Lifetime

joshdholtz:
### Motivation

Prevent "Lifetime" from wrapping in Template 4

### Description

- Added `.frame(maxWidth: .infinity)` to the package button textfields
  -  They were not expanding all the way due to calculations on their parent's width

| Before | After |
|--------|--------|
| Screenshot 2024-04-09 at 9 12 57 PM | Screenshot 2024-04-09 at 9 12 16 PM |

### DIFF
--- a/RevenueCatUI/Templates/Template4View.swift
+++ b/RevenueCatUI/Templates/Template4View.swift
@@ -359,6 +359,7 @@ private struct PackageButton: View {
             }
 
             Text(secondRow)
+                .frame(maxWidth: .infinity)
                 .font(self.font(for: .title3).weight(.regular))
         }
     }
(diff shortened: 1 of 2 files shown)


tonidero (approved):

NachoSoto:
Does this work when there are more than 3 and it scrolls horizontally?

joshdholtz:
> Does this work when there are more than 3 and it scrolls horizontally?

Yup yup, it does! 

https://github.com/RevenueCat/purchases-ios/assets/401294/980f3af8-1baf-4045-9f35-17c22b0c7644

PR #3790, opened 2024-04-10
Add name property to HTTPRequestPath

vegaro:
This will be used for diagnostics

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPRequestPath.swift
+++ b/Sources/Networking/HTTPClient/HTTPRequestPath.swift
@@ -33,6 +33,9 @@ protocol HTTPRequestPath {
     /// The path component for this endpoint.
     var pathComponent: String { get }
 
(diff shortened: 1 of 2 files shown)


Sources/Networking/HTTPClient/HTTPRequestPath.swift
```diff
@@ -189,6 +192,45 @@ extension HTTPRequest.Path: HTTPRequestPath {
...
+            return "post_offer_for_signing"
+
+        case .postReceiptData:
+            return "post_receipt"
+
+        case .postSubscriberAttributes:
+            return "post_attributes"
+
+        case .health:
+            return "post_health"
```
tonidero, line 229: I need to add some of these to the backend validation

tonidero (approved):
Looks good! I will add the new endpoint names to the backend validation

PR #3791, opened 2024-04-10
Add HTTPRequest.DiagnosticsPath

vegaro:
Create DiagnosticsHTTPRequestPath for Diagnostics

Depends on #3790 which adds the name


tonidero (approved):

tonidero (commented):
I'm not sure if we should mark all of these as `feat`, since it won't be added until we actually enable the public API

PR #3793, opened 2024-04-10
Add origin to HTTPResponseType

vegaro:
This is needed for Diagnostics

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPResponse.swift
+++ b/Sources/Networking/HTTPClient/HTTPResponse.swift
@@ -235,7 +235,6 @@ extension VerifiedHTTPResponse {
 enum HTTPResponseOrigin {
 
     case backend
-
     case cache
 
 }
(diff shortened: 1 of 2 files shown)


Sources/Networking/HTTPClient/HTTPResponse.swift
```diff
@@ -226,3 +231,11 @@ extension VerifiedHTTPResponse {
     }
 
 }
+
+enum HTTPResponseOrigin {
+
+    case backend
+
```
tonidero, line 238: Total nitpick but should this line be removed?

tonidero (approved):

Sources/Networking/HTTPClient/HTTPResponse.swift
vegaro: right. According to the style guide it should. Thanks!

PR #3795, opened 2024-04-10
Create DiagnosticsPostOperation

vegaro:
Add DiagnosticsPostOperation and DiagnosticsAPI


Sources/Diagnostics/Networking/DiagnosticsPostOperation.swift
```diff
@@ -0,0 +1,70 @@
...
+//
+//      https://opensource.org/licenses/MIT
+//
+//  DiagnosticsPostOperation.swift
+//
+//  Created by Cesar de la Vega on 8/4/24.
+
+import Foundation
+
+typealias DiagnosticsEntries = [String]
```
tonidero, line 16: Oh so the whole body of each event is sent as a string then? We should be careful with encoding characters then, but maybe `Body` is already taking care of that? (not sure)

Sources/Networking/DiagnosticsAPI.swift
```diff
@@ -0,0 +1,40 @@
...
+
+    typealias ResponseHandler = Backend.ResponseHandler<Void>
+
+    private let backendConfig: BackendConfiguration
+
+    init(backendConfig: BackendConfiguration) {
+        self.backendConfig = backendConfig
+    }
+
+    func postDiagnostics(
```
tonidero, line 26: Is this something we can add tests for now? (Can also happen in a separate PR)

tonidero (commented):
Just a couple of questions but looking good!

Sources/Diagnostics/Networking/DiagnosticsPostOperation.swift
vegaro: I noticed `PostAttributionDataOperation` is doing the same, so I think it is being taken care of, but I will try to find some docs around that

vegaro: So `Body` is `Encodable` and `Body` has the entries, so I think this should work. But I am going to change it to the way `PostAttributionDataOperation` works. It stores the data as `AnyEncodable` instead of `[String:Any]`

Sources/Networking/DiagnosticsAPI.swift
vegaro: added a test that verifies we are calling the http client (similar to `BackendPostAdServicesTokenTests`)

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostDiagnosticsTests/iOS17-testPostDiagnosticsPutsDataInDataKey.1.json
```diff
@@ -0,0 +1,28 @@
...
+    "X-Platform-Flavor" : "native",
+    "X-Platform-Version" : "Version 17.0.0 (Build 21A342)",
+    "X-Storefront" : "USA",
+    "X-StoreKit2-Enabled" : "false",
+    "X-Version" : "4.0.0"
+  },
+  "request" : {
+    "body" : {
+      "data" : [
+        "{\n  \"properties\": {\"key\": \"value\"},\n  \"timestamp\": \"2024-04-04T12:55:59Z\",\n  \"name\": \"HTTP_REQUEST_PERFORMED\",\n  \"version\": 1\n}",
```
tonidero, line 21: Hmm we should send these as json objects, instead of strings. Maybe we need to process it back to an object after all?

vegaro: hmm good point, I hadn't noticed that! Good catch

vegaro: Ended up doing a refactor to make the code similar to Paywall Events, and now they are sent as json with the correct date format. lmkwyt

tonidero (approved):
LGTM!

PR #3797, opened 2024-04-10
Override PaywallViewController touch events to avoid propagating them

tonidero:
### Description
There was an issue reporting with paywalls in flutter: https://github.com/RevenueCat/purchases-flutter/issues/1023

From investigating, it was related to this issue: https://github.com/flutter/flutter/issues/14720

This is a possible solution as suggested in this comment: https://github.com/flutter/flutter/issues/35784#issuecomment-516243057

PR #3798, opened 2024-04-10
Make `PaywallViewController` open

tonidero:
### Description
Following a different approach to #3797. We want to move some logic to the hybrids and we need to make the `PaywallViewController` open for that.


tonidero:
Closing in favor of #3801

PR #3805, opened 2024-04-11
Paywalls: Allow closed button color to be configured

joshdholtz:
### Motivation

Allow presented paywall view to have a custom close button color from the dashboard

### Description

- New `close_button` color

### DIFF
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -321,14 +321,20 @@ struct LoadedOfferingPaywallView: View {
 
     @ViewBuilder
     private var content: some View {
+        let configuration = self.paywall.configuration(
+            for: self.offering,
+            activelySubscribedProductIdentifiers: self.activelySubscribedProductIdentifiers,
+            template: self.template,
+            mode: self.mode,
+            fonts: self.fonts,
+            locale: self.locale
+        )
+
         let view = self.paywall
             .createView(for: self.offering,
-                        activelySubscribedProductIdentifiers: self.activelySubscribedProductIdentifiers,
                         template: self.template,
-                        mode: self.mode,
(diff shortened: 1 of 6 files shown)


RevenueCatUI/PaywallView.swift
```diff
@@ -350,7 +356,11 @@ struct LoadedOfferingPaywallView: View {
         if self.displayCloseButton {
             NavigationView {
                 view
-                    .toolbar { self.toolbar }
+                    .toolbar {
+                        self.makeToolbar(
```
tonidero, line 360: I'm wondering, does this mean it will recreate the toolbar every time it needs to redraw? Do we need to mark `makeToolbar` with the `@ViewBuilder` annotation? (Not superfamiliar with swiftUI so it might be able to optimize it somehow...)

RevenueCatUI/PaywallView.swift
```diff
@@ -321,14 +321,20 @@ struct LoadedOfferingPaywallView: View {
 
     @ViewBuilder
     private var content: some View {
+        let configuration = self.paywall.configuration(
```
tonidero, line 324: Nitpick, but I wonder if we should move creating the configuration outside the view... Having said that, it would need all these parameters, so I think it's ok as is, at least for now.

tonidero (approved):
A couple questions but looking good!

RevenueCatUI/PaywallView.swift
joshdholtz: This can't be `@ViewBuilder` because weirdly enough, `ToolbarContent` isn't a `View` 😅  This will recreate the toolbar content object but won't necessarily redraw the UI since SwiftUI does those diffs under the hood

PR #3806, opened 2024-04-11
[PoC] cloud synced anonymous ids

aboedo:
This allows an app that doesn't have an account system to automatically sync anonymous IDs. 

This is helpful in a few ways: 
- it allows automatic restores on different devices under the same iCloud account
- it prevents sharing app store accounts to attach them to anonymous user IDs and share purchases

This is achieved by using `NSUbiquituousKeyValueStorage`, which syncs data through iCloud. Each app only gets 1mb worth of data, but we're only storing a single string. 

Missing: 
- tests (tested on device and it works as expected though)
- watchOS compatibility (watchOS only supports key value storage starting on watchOS 9.0)
- a delegate to inform developer when an app store account change is detected and the user is logged out as a result

### DIFF
--- a/Sources/Identity/IdentityManager.swift
+++ b/Sources/Identity/IdentityManager.swift
@@ -143,11 +143,9 @@ extension IdentityManager {
 private extension IdentityManager {
 
     func performLogIn(appUserID: String, completion: @escaping IdentityAPI.LogInResponseHandler) {
-        // todo: update warning message, we want to warn that you're logging in in this mode,
-        // which maybe is an accident, and that you should check that you're using the right
-        // tranfer behavior
         if cloudSyncedAnonymousIDMode {
-            Logger.warn(Strings.identity.deleting_attributes_none_found)
+            Logger.warn(Strings.identity.login_called_in_cloud_synced_appuserid_mode(oldUserID: currentAppUserID,
+                                                                                     newUserID: appUserID))
         }
 
         let oldAppUserID = self.currentAppUserID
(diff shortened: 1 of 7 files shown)


Sources/Identity/IdentityManager.swift
```diff
@@ -39,27 +41,51 @@ class IdentityManager: CurrentUserProvider {
...
+        deviceCache.cache(appUserID: finalAppUserID)
         deviceCache.cleanupSubscriberAttributes()
-        self.invalidateCachesIfNeeded(appUserID: appUserID)
+        self.invalidateCachesIfNeeded(appUserID: finalAppUserID)
+
+        // todo: inject store and notification center
+        NotificationCenter.default.addObserver(
+            self,
+            selector: #selector(self.ubiquitousKeyValueStoreDidChange),
+            name: NSUbiquitousKeyValueStore.didChangeExternallyNotification,
```
tonidero, line 79: Should we only subscribe to this if in `cloudSyncedAnonymousIDMode`? Maybel also the `applicationWillEnterForegroundNotification` subscribe since right now it's only for this use case as well

Sources/Logging/Strings/IdentityStrings.swift
```diff
@@ -80,7 +84,25 @@ extension IdentityStrings: LogMessage {
...
+            logIn() called with cloud synced appUserIDs enabled. This shouldn't be an issue, but the purchases
+            for the user will be transferred or blocked depending on the configuration in RevenueCat Dashboard.
+            Old user ID: \(oldUserID)
+            New user ID: \(newUserID)
+            More info: https://rev.cat/auth
+            """
+        case let .logout_called_in_cloud_synced_appuserid_mode(oldUserID, newUserID):
+            return """
+            logOut() called with cloud synced appUserIDs enabled. The current user will be logged out, and
+            a new appUserID will be generated.
```
tonidero, line 98: This is actually the same that happens in a normal logout right? Maybe we should mention that the iCloud user id has been reset?

aboedo: oops, probably a copy/pasted string that I meant to updaet

PR #3814, opened 2024-04-17
4. Paywalls Tester: Sign in persistence + Sign Out

jamesrb1:
Signed in status is persisted and adds the ability to sign out.

### DIFF
--- a/Tests/TestingApps/PaywallsTester/PaywallsTester/Data/ApplicationData.swift
+++ b/Tests/TestingApps/PaywallsTester/PaywallsTester/Data/ApplicationData.swift
@@ -34,6 +34,13 @@ final class ApplicationData {
         self.manager.signOut()
         self.authentication = .signedOut
     }
+    
+    var isSignedIn: Bool {
+        if case .signedIn = self.authentication {
+            return true
+        }
+        return false
+    }
 
     @ObservationIgnored
     private var manager: ApplicationManagerType = ApplicationManager()
--- a/Tests/TestingApps/PaywallsTester/PaywallsTester/Data/ApplicationData.swift
+++ b/Tests/TestingApps/PaywallsTester/PaywallsTester/Data/ApplicationData.swift
@@ -12,27 +12,27 @@ import OSLog
 @Observable
 final class ApplicationData {
 
-    private(set) var authentication: Authentication = .unknown {
+    private(set) var authenticationStatus: AuthenticationStatus = .unknown {
         didSet {
-            Self.logger.info("Changed authentication: \(String(describing: self.authentication))")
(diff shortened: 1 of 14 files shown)


Tests/TestingApps/PaywallsTester/PaywallsTester/PaywallsTester.entitlements
```diff
@@ -4,6 +4,10 @@
 <dict>
 	<key>com.apple.security.app-sandbox</key>
 	<true/>
+	<key>com.apple.security.application-groups</key>
+	<array>
+		<string>group.com.revenuecat</string>
+	</array>
```
jamesrb1, line 10: Not sure why this is showing up here again, something I did when setting up the stacked PRs.

Tests/TestingApps/PaywallsTester/PaywallsTester/Data/ApplicationData.swift
```diff
@@ -34,6 +34,13 @@ final class ApplicationData {
         self.manager.signOut()
         self.authentication = .signedOut
     }
+    
+    var isSignedIn: Bool {
+        if case .signedIn = self.authentication {
+            return true
+        }
+        return false
+    }
```
jamesrb1, line 43: This is here so it's simple to get a bool of if the user is signed in or not, without worrying about the signed in enum's associated data.

Tests/TestingApps/PaywallsTester/PaywallsTester/Networking/HTTPClient.swift
```diff
@@ -73,7 +73,11 @@ public final class HTTPClient {
     }
 
     func removeCookies() {
-        self.session.configuration.httpCookieStorage?.removeCookies(since: .distantPast)
+        // the following call only persists the deletion on device, does not work on simulators
+        // self.session.configuration.httpCookieStorage?.removeCookies(since: .distantPast)
```
aboedo, line 77: Feels like this would be a weird thing for a maintainer to see without context on why we even keep the old line around, maybe we can make that super clear in the comment like
```suggestion
        // we need to delete cookies one by one because deleting them all directly from 
        // `httpCookieStorage.removeCookies(since:)` only persists the deletion on device
        // and does not work on simulators (FB13733433)
```

aboedo (approved):
looking great, left one suggestion

Tests/TestingApps/PaywallsTester/PaywallsTester/Networking/HTTPClient.swift
jamesrb1: Committed!

PR #3711, opened 2024-02-23
Do not update the documentation index.html file if version is prerelease

MarkVillacampa:
### Checklist
- [ ] If applicable, unit tests
- [ ] If applicable, create follow-up issues for `purchases-android` and hybrids

PR #3815, opened 2024-04-17
Only update docs index on latest stable releases

tonidero:
### Description
iOS equivalent of: https://github.com/RevenueCat/purchases-android/pull/1676

We updated the plugin to support a new parameter that will only be updated on stable versions (no prereleases) and as the latest version (no bigger semver versions exist)


tonidero:
This might cause some conflicts with https://github.com/RevenueCat/purchases-ios/pull/3711 that was merged in the 5.0 branch.

I still think it's worth using this approach since it will allow us to forget about issues with the docs if we have to make a hotfix for 4.x. After this is merged, I can try to fix the collisions with the 5.0-dev branch cc @fire-at-will

aboedo (approved):

PR #3817, opened 2024-04-17
Use SK2 in Traditional Configure SDK Functions

fire-at-will:
### Motivation
The `Purchases.configure()` functions that don't use the builder patter weren't yet using StoreKit2. This PR updates them to use the default StoreKit version (in this case, StoreKit 2).

### Testing
TODO: Test locally through the purchase tester app and in my own apps.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1325,7 +1325,7 @@ public extension Purchases {
             with: Configuration
                 .builder(withAPIKey: apiKey)
                 .with(appUserID: appUserID)
-                .with(observerMode: observerMode, storeKitVersion: .storeKit1)
+                .with(observerMode: observerMode, storeKitVersion: .default)
                 .build()
         )
     }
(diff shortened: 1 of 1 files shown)


joshdholtz (commented):
I still have some concern with making these use `.default` because of the `observerMode: Bool` parameter  🤔 

When set to `true` with `.default` (and which could be `.storeKit2`), Observer Mode would break because the the requirment to do 👇 

```
let result = try await product.purchase()
Purchases.shared.handleObserverModeTransaction(result)
```

I think these methods need to either be:
1) Deprecated and keep `.storeKit1`
2) Add `storeKitVersion` as a parameter that defaults to `.storeKit1`

We need the use of `.storeKit2` to be intentional

aboedo:
+1, I think when you set observer mode you should also _have_ to set the storekit version yourself.
If you already have a StoreKit implementation then it should be very easy for you to know which version it's using, so you won't have a hard time getting it to match

fire-at-will:
Closing in favor of https://github.com/RevenueCat/purchases-ios/pull/3826

PR #3824, opened 2024-04-19
Clear diagnostics file if we detect it's too big during SDK initialization

tonidero:
### Description
This makes it so, if we detect the diagnostics file gets too big during SDK initialization, we just clear it and start over. In a follow-up PR, we will add a tracking event when this happens so we can see occurences of this

### DIFF
--- a/Sources/Diagnostics/DiagnosticsFileHandler.swift
+++ b/Sources/Diagnostics/DiagnosticsFileHandler.swift
@@ -24,6 +24,8 @@ protocol DiagnosticsFileHandlerType: Sendable {
 
     func emptyDiagnosticsFile() async
 
+    func isDiagnosticsFileTooBig() async -> Bool
+
 }
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
@@ -50,7 +52,7 @@ actor DiagnosticsFileHandler: DiagnosticsFileHandlerType {
             return
         }
 
-        await fileHandler.append(line: jsonString)
+        await self.fileHandler.append(line: jsonString)
     }
 
     func getEntries() async -> [DiagnosticsEvent] {
@@ -71,7 +73,7 @@ actor DiagnosticsFileHandler: DiagnosticsFileHandlerType {
         }
 
         do {
-            try await fileHandler.removeFirstLines(diagnosticsSentCount)
+            try await self.fileHandler.removeFirstLines(diagnosticsSentCount)
         } catch {
             Logger.error("Failed to clean sent diagnostics: \(error.localizedDescription)")
         }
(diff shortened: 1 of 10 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -165,6 +165,7 @@ final class PurchasesOrchestrator {
         }
 
         Task {
+            await diagnosticsSynchronizer?.clearDiagnosticsFileIfTooBig()
             await syncDiagnosticsIfNeeded()
```
aboedo, line 169: When we do actually clear diagnostics if the file is too large, we'll only have one event - the "file cleared event". 
It's kind of a bummer because we're losing a bunch of events unnecessarily. 

I was thinking that we could optimize this a bit by doing the "clear if too big" part at event generation time instead of at syncing time. This would cut down on the events thrown away. 

This would optimize cases where for example you: 
before:
- have max file size 10
- have 15 events
- sync, now you delete 15 events and generate the "file nuked" event and sync that

after: 
- have max file size 10
- when you try to add event 11, you see that you've gone over and kill the previous 10
- add the next 4
- sync, now you have 5 events that you can sync

bonus points: file size is guaranteed to never go over the limit, regardless of any other logic.

tonidero: That is a good point. Will update android to this new logic and update this PR as well. Thanks!

Sources/Diagnostics/FileHandler.swift
```diff
@@ -127,6 +132,14 @@ actor FileHandler: FileHandlerType {
         try self.replaceHandler(with: tempURL)
     }
 
+    func fileSizeInKB() async throws -> Double {
+        let attributes = try FileManager.default.attributesOfItem(atPath: self.url.path)
```
tonidero, line 136: I took a look and I think there are no privacy labels that need to be added to use this but just fyi.

Sources/Diagnostics/DiagnosticsFileHandler.swift
```diff
@@ -74,20 +77,30 @@ actor DiagnosticsFileHandler: DiagnosticsFileHandlerType {
...
         } catch {
             Logger.error("Failed to empty diagnostics file: \(error.localizedDescription)")
         }
     }
 
+    func isDiagnosticsFileTooBig() async -> Bool {
+        do {
+            return try await self.fileHandler.fileSizeInKB() > Self.maxFileSizeInKb
+        } catch {
+            Logger.error("Failed to check whether diagnostics file is too big: \(error.localizedDescription)")
```
aboedo, line 98: kinda surprised we haven't extracted the strings for error messages in this file like we do for all the others

aboedo (approved):

Sources/Diagnostics/DiagnosticsFileHandler.swift
tonidero: Good point! Addressed in https://github.com/RevenueCat/purchases-ios/pull/3824/commits/b452ebdf21ca1067772332f5655c47aba374c58d

PR #3826, opened 2024-04-19
Support SK2 in Basic Configure Functions

fire-at-will:
### Motivation
One of the main goals of the 5.0.0 release of `purchases-ios` is to handle purchases with StoreKit 2 by default. This PR gets us closer to that goal by introducing new parameters to the simpler SDK `configure` functions that support StoreKit 2 by default when possible.

### Changes Introduced
- Introduced a new parameter, `storeKitVersion: StoreKitVersion` to the non-builder observer mode configure function.
   - A default value is not provided, forcing developers to choose their StoreKit version.

- Obsoleted the old observer mode configure function without the `storeKitVersion` parameter.
   - Developers using `Purchases.configure(withAPIKey:appUserID:observerMode:)` will hit the old obsoleted configure function, and will be pushed to specify a StoreKit version:
Screenshot 2024-04-19 at 3 29 18 PM

- Updated the Swift & Objective-C API testers to account for the new changes

- Updated to our DocC documentation to reflect these new function signatures

- Included a warning in `StoreKitVersion.storeKit2`'s documentation instructing developers using observer mode to call `Purchases/handleObserverModeTransaction(_:)` after making a purchase.

### Breaking Changes
- Developers that were using `Purchases.configure(withAPIKey:appUserID:observerMode:)` will be forced to migrate to `Purchases.configure(withAPIKey:appUserID:observerMode:storeKitVersion)`.

### Other Approaches Considered
See [this Google doc](https://docs.google.com/document/d/1Kwri7qcSMQRlyNyn4Dyckqxd8eVf_JqEkH1fJJsk994/edit#heading=h.m1h0u03xo2zt) for a list of other approaches considered.


Sources/DocCDocumentation/DocCDocumentation.docc/Purchases.md
```diff
@@ -22,7 +22,7 @@ Most features require configuring the SDK before using it.
 - ``Purchases/logHandler``
 
 ### Configuring the SDK
-- ``Purchases/configure(withAPIKey:)``
+- ``Purchases/configure(withAPIKey:storeKitVersion:)``
```
aboedo, line 25: honestly I think this kind of defeats the purpose of having these functions in the first place - they're meant to keep configuration as easy as possible for the majority of the customer base. 

Most customers should not have to think about the storekit version, they pay us to do it for them.

Sources/Misc/Obsoletions.swift
```diff
@@ -526,6 +526,115 @@ public extension Purchases {
...
+     * ``Purchases`` will generate a unique identifier for the current device and persist it to `NSUserDefaults`.
+     * This also affects the behavior of ``Purchases/restorePurchases(completion:)``.
+     *
+     * - Parameter apiKey: The API Key generated for your app from https://app.revenuecat.com/
+     *
+     * - Returns: An instantiated ``Purchases`` object that has been set as a singleton.
+     */
+    @objc(configureWithAPIKey:)
+    @available(iOS, obsoleted: 1, message: """
+    If you're using observer mode with StoreKit 2, ensure that you call Purchases.handleObserverModeTransaction(result:) after making a purchase.
```
aboedo, line 545: this note is great, let's also add it to the docstring though. In the docstring we can even use double backticks syntax to create a link to the other method

Sources/Misc/Obsoletions.swift
```diff
@@ -526,6 +526,115 @@ public extension Purchases {
...
+     *
+     * - Note: Use this initializer if your app does not have an account system.
+     * ``Purchases`` will generate a unique identifier for the current device and persist it to `NSUserDefaults`.
+     * This also affects the behavior of ``Purchases/restorePurchases(completion:)``.
+     *
+     * - Parameter apiKey: The API Key generated for your app from https://app.revenuecat.com/
+     *
+     * - Returns: An instantiated ``Purchases`` object that has been set as a singleton.
+     */
+    @objc(configureWithAPIKey:)
```
aboedo, line 543: I don't understand why we need to break this API - this is for folks who don't use observer mode. Why should they have to change anything? Isn't it fine for them to just switch seamlessly from SK1 to SK2?

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1312,20 +1322,22 @@ public extension Purchases {
...
      * - Warning: This assumes your IAP implementation uses StoreKit 1.
      */
-    @_disfavoredOverload
-    @objc(configureWithAPIKey:appUserID:observerMode:)
+    @objc(configureWithAPIKey:appUserID:observerMode:storeKitVersion:)
     @discardableResult static func configure(withAPIKey apiKey: String,
                                              appUserID: String?,
-                                             observerMode: Bool) -> Purchases {
+                                             observerMode: Bool,
+                                             storeKitVersion: StoreKitVersion) -> Purchases {
```
aboedo, line 1335: this is the only one that I actually understand why we need to break

aboedo:
> A default value is not provided for the configure functions that include the ability to set observerMode, forcing developers to choose their StoreKit version.

This seems fine and is consistent with the builder version of it, but we're also not providing a default value for the ones that do not provide the ability to set observer mode, which has me confused

fire-at-will:
@aboedo Thanks for the review, your comments make a lot of sense! I've made a few changes based on the feedback:
- Removed the optional parameter `storeKitVersion` from `Purchases.configure(withAPIKey:)` and `Purchases.configure(withAPIKey:appUserID)` and set it so that they use StoreKit 2
   - Updated the docs to reflect these changes
- Remove the obsoletions on the non observer-mode configure functions
- Added the warning on the obsoleted observer mode configure function to its docstring
- Fixed the docstring reference on the observer-mode constructor. Previously, it stated that you could use that function to set a custom UserDefaults, but the function didn't allow developers pass one in.

Now the only breaking change in the PR is for the observer mode configure function. 😎

Sources/Misc/Obsoletions.swift
```diff
@@ -526,6 +526,48 @@ public extension Purchases {
...
+     * purchases and subscriptions across devices. Pass `nil` or an empty string if you want ``Purchases``
+     * to generate this for you.
+     *
+     * - Parameter observerMode: Set this to `true` if you have your own IAP implementation and want to use only
+     * RevenueCat's backend. Default is `false`.
+     *
+     * - Returns: An instantiated ``Purchases`` object that has been set as a singleton.
+     *
+     * - Warning: This assumes your IAP implementation uses StoreKit 1.
+     * - Warning: If you're using observer mode with StoreKit 2, ensure that you call ``Purchases/handleObserverModeTransaction(_:)`` after making a purchase.
```
aboedo, line 549: The second warning is kind of a no-op since this API is obsoleted, right? Meaning you can't use it in any case, so you would still have to move to the SK2 version of it? 

Assuming that's correct, maybe we can tell folks to move to that method as well, like
```suggestion
     * - Warning: This assumes your IAP implementation uses StoreKit 1.
     * - Warning: If you're using observer mode with StoreKit 2, configure the SDK with `configure(withAPIKey:appUserID:observerMode:storeKitVersion:)` passing in `.storeKit2` as the `storeKitVersion` and ensure that you call ``Purchases/handleObserverModeTransaction(_:)`` after making a purchase.
```

aboedo (approved):

aboedo:
left a minor suggestion but good to go!

PR #3827, opened 2024-04-19
Sync repo in an SPM-friendly way

jamesrb1:
Creates a copy of this repo at `/RevenueCat/purchases-ios-spm` that is much faster for integrating RC via SPM.

Before merge:

- [x] Delete and recreate purchases-ios-spm repo
- [x] Set up branch restrictions to require a PR for everyone except the RCGitBot Team
- [x] Assign purchases-ios-spm an owner in Vanta

Before or after merge:
- [x] Manually run the script once to populate the spm repo

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1007,6 +1007,7 @@ jobs:
       - run:
           name: Create automatic PR
           command: bundle exec fastlane automatic_bump github_rate_limit:10
+      - push-to-spm
 
   push-to-spm:
     docker:
@@ -1234,10 +1235,6 @@ workflows:
               branches:
                 # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
                 ignore: /pull\/[0-9]+/
-
-  deploy-spm:
-    jobs:
-      - push-to-spm
 
   deploy:
     when:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -340,9 +340,6 @@ commands:
           destination: scan-test-output
 
   push-to-spm:
-    docker:
-      - image: cimg/base:stable
-    resource_class: small
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -339,6 +339,20 @@ commands:
...
+      - setup-git-credentials
+      - run:
+          name: Clone purchases-ios and push to purchases-ios-spm
+          command: |
+            git clone https://github.com/RevenueCat/purchases-ios.git
+            cd purchases-ios
+            git fetch --tags
+            git remote set-url origin https://github.com/RevenueCat/purchases-ios-spm.git
+            git push origin
+            git push --tags
```
aboedo, line 354: [not needed for this PR] ideally we'd move this to fastlane, we try to keep the CI stuff as simple as possible and move stuff into fastlane which is much easier to test. 
Also it feels like the URLs might be a good thing to move into env variables in case we rename them, but not a dealbreaker and we might never need it in any case.

.circleci/config.yml
```diff
@@ -339,6 +339,20 @@ commands:
...
 
+  push-to-spm:
+    steps:
+      - checkout
+      - setup-git-credentials
+      - run:
+          name: Clone purchases-ios and push to purchases-ios-spm
+          command: |
+            git clone https://github.com/RevenueCat/purchases-ios.git
+            cd purchases-ios
```
aboedo, line 350: this will always be running _from_ a clone, right? why do we do this instead of adding the other repo as an alt remote to the existing repo?

aboedo (approved):

.circleci/config.yml
jamesrb1: As discussed, merging now because it's working and we are biased to shipping, will return to this as a possible future improvement.

PR #3833, opened 2024-04-22
Fix a few flaky tests

aboedo:
Looking into our flaky tests, I noticed that calls to [forceRenewalOfSubscription](https://developer.apple.com/documentation/storekittest/sktestsession/3579486-forcerenewalofsubscription) don't always trigger calls from `StoreKit.product.updates`, it's flaky as hell. 

I tried instead using an accelerated timeRate value and adding explicit waits, and that seems to consistently work in local testing. Opening this up to see if this helps with the flakiness of these particular tests. 

### Notes from Josh

- This also contains a cherry-picked `swiftlint` fix from https://github.com/RevenueCat/purchases-ios/pull/3909 
- This also contains a cleanup of the _fastlane_ code for started in #3902 (which is what this branch is merging into)

### DIFF
--- a/Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
@@ -90,6 +90,7 @@ extension BaseStoreKitIntegrationTests {
     static let group3MonthlyTrialProductID = "com.revenuecat.monthly.1.99.1_free_week"
     static let group3MonthlyNoTrialProductID = "com.revenuecat.monthly.1.99.no_intro"
     static let group3YearlyTrialProductID = "com.revenuecat.annual.10.99.1_free_week"
+    static let weeklyWith3DayTrial = "shortest_duration"
 
     var currentOffering: Offering {
         get async throws {
--- a/Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
@@ -174,7 +174,7 @@ extension BaseStoreKitIntegrationTests {
         let logger = TestLogHandler()
         let product = try await StoreKit.Product.products(for: ["shortest_duration"]).first!
 
-        let data = try await self.purchases.purchase(product: product)
+        let data = try await self.purchases.purchase(product: StoreProduct(sk2Product: product))
 
         try await self.verifyEntitlementWentThrough(data.customerInfo,
                                                     file: file,
--- a/Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
@@ -164,6 +164,29 @@ extension BaseStoreKitIntegrationTests {
 
         return data
     }
+    
+    @discardableResult
+    func purchaseShortestDuration(
+        allowOfflineEntitlements: Bool = false,
+        file: FileString = #file,
+        line: UInt = #line
+    ) async throws -> PurchaseResultData {
+        let logger = TestLogHandler()
+        let product = try await StoreKit.Product.products(for: ["shortest_duration"]).first!
+
+        let data = try await self.purchases.purchase(product: product)
+
+        try await self.verifyEntitlementWentThrough(data.customerInfo,
+                                                    file: file,
+                                                    line: line)
+
+        if !allowOfflineEntitlements {
+            // Avoid false positives if the API returned a 500 and customer info was computed offline
+            self.verifyCustomerInfoWasNotComputedOffline(logger: logger, file: file, line: line)
+        }
+
+        return data
+    }
 
     @discardableResult
     func purchaseConsumablePackage(
--- a/Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
@@ -178,7 +178,7 @@ extension BaseStoreKitIntegrationTests {
         line: UInt = #line
     ) async throws -> PurchaseResultData {
         let logger = TestLogHandler()
-        let product = try await StoreKit.Product.products(for: ["shortest_duration"]).first!
+        let product = try await StoreKit.Product.products(for: [Self.weeklyWith3DayTrial]).first!
 
         let data = try await self.purchases.purchase(product: StoreProduct(sk2Product: product))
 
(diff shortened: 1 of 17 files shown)


Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift
```diff
@@ -164,6 +171,29 @@ extension BaseStoreKitIntegrationTests {
...
     }
 
+    @discardableResult
+    func purchaseShortestDuration(
+        allowOfflineEntitlements: Bool = false,
+        file: FileString = #file,
+        line: UInt = #line
+    ) async throws -> PurchaseResultData {
+        let logger = TestLogHandler()
+        let product = try await StoreKit.Product.products(for: ["shortest_duration"]).first!
```
fire-at-will, line 181: Nit: use the constant here

```suggestion
        let product = try await StoreKit.Product.products(for: [Self.weeklyWith30DayTrial]).first!
```

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -356,27 +361,34 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
...
         // 2. Change to user 2
         let (identifiedCustomerInfo, _) = try await self.purchases.logIn(userID2)
         self.assertNoPurchases(identifiedCustomerInfo)
 
         // 3. Renew subscription
         self.logger.clearMessages()
-
-        try self.testSession.forceRenewalOfSubscription(productIdentifier: productIdentifier)
+        // swiftlint:disable:next force_try
+        try! await Task.sleep(nanoseconds: 3 * 1_000_000_000)
```
fire-at-will, line 391: I never thought I'd say this, but sleeping in this test actually makes more sense than relying on the old flaky StoreKit behavior 😂

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
```diff
@@ -241,16 +243,23 @@ class OfflineStoreKit1IntegrationTests: BaseOfflineStoreKitIntegrationTests {
 
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
     func testCallToGetCustomerInfoWithPendingTransactionsPostsReceiptOnlyOnce() async throws {
+        // forceRenewalOfSubscription doesn't work well, so we use this instead
+        if #available(iOS 16.4, *) {
```
fire-at-will, line 247: It looks like we have this code duplicated in several tests, could it be abstracted out to a common function in `BaseStoreKitIntegrationTests`?

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -641,21 +683,21 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
         await waitForNewPurchaseDate()
 
         // 4. Expire subscription
-        try await self.expireSubscription(entitlement)
+//        try await self.expireSubscription(entitlement)
```
fire-at-will, line 686: Is commenting out the latter part of this test intentional?

Tests/StoreKitUnitTests/StoreKit2/StoreKit2TransactionListenerTests.swift
```diff
@@ -217,11 +217,13 @@ class StoreKit2TransactionListenerTransactionUpdatesTests: StoreKit2TransactionL
...
     @available(iOS 16.4, macOS 13.3, tvOS 16.4, watchOS 9.4, *)
     func testNotifiesDelegateForRenewals() async throws {
+        self.testSession.timeRate = .oneRenewalEveryTwoSeconds
         try await self.simulateAnyPurchase(finishTransaction: true)
 
         await self.listener.listenForTransactions()
 
-        try? self.testSession.forceRenewalOfSubscription(productIdentifier: Self.productID)
+        // swiftlint:disable:next force_try
+        try! await Task.sleep(nanoseconds: 2 * 1_000_000_000)
```
fire-at-will, line 226: Nit: it might be good to sleep for a little longer than 2 seconds to give the renewal a little bit of breathing room to finish before verifying

```suggestion
        try! await Task.sleep(nanoseconds: 2 * 1_100_000_000)
```

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
```diff
@@ -241,16 +243,23 @@ class OfflineStoreKit1IntegrationTests: BaseOfflineStoreKitIntegrationTests {
...
+            self.testSession.timeRate = .oneRenewalEveryTwoSeconds
+        } else {
+            self.testSession.timeRate = SKTestSession.TimeRate.monthlyRenewalEveryThirtySeconds
```
fire-at-will, line 250: Will this fail if we use `SKTestSession.TimeRate.monthlyRenewalEveryThirtySeconds` on iOS <16.4 and then sleep for 3 seconds later in the test?

fire-at-will:
@joshdholtz Looks good! I had a few suggestions/questions and then I think it'll be good to go!

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
joshdholtz: Nope! I noticed this a bit ago and am fixing 😛

Tests/BackendIntegrationTests/OfflineStoreKitIntegrationTests.swift
joshdholtz: @fire-at-will  We only run backend tests on iOS 17 right now so... odds are this will never happen 😛

fire-at-will: 🤪

fastlane/actions/retry_scan_save_failed_tests.rb
```diff
@@ -0,0 +1,157 @@
+require 'nokogiri'
+
+module Fastlane
+  module Actions
+    module SharedValues
+      RETRY_SCAN_SAVE_FAILED_TESTS_CUSTOM_VALUE = :RETRY_SCAN_SAVE_FAILED_TESTS_CUSTOM_VALUE
+    end
+
+    class RetryScanSaveFailedTestsAction < Action
```
tonidero, line 9: Wonder if we should move this to the plugin, so it can also be tested... But since I think it will only be used here for now, I think this is ok 👍

PR #3834, opened 2024-04-22
Push to SPM after release has been made

jamesrb1:
Previously the sync with SPM occurred before the release PR had been merged. This PR should correct that.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1240,6 +1240,10 @@ workflows:
                 # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
                 ignore: /pull\/[0-9]+/
 
+  run-spm-workflow:
+    jobs:
+      - deploy-to-spm
+
   deploy:
     when:
       not:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1121,6 +1120,13 @@ jobs:
                 ]
               }
 
+  deploy-to-spm:
+    docker:
+      - image: cimg/base:stable
+    resource_class: small
+    steps:
+      - push-to-spm
```
vegaro, line 1128: nitpick, but since the `push-to-spm` command is only used in this job, I think we could just have it here directly, without the need of a command:

```
deploy-to-spm:
    docker:
      - image: cimg/base:stable
    resource_class: small
    steps:
      - checkout
      - setup-git-credentials
      - run:
          name: Clone purchases-ios and push to purchases-ios-spm
          command: |
            git clone https://github.com/RevenueCat/purchases-ios.git
            cd purchases-ios
            git fetch --tags
            git remote set-url origin https://github.com/RevenueCat/purchases-ios-spm.git
            git push origin
            git push --tags
 ```

vegaro (approved):

.circleci/config.yml
jamesrb1: Done!

PR #3837, opened 2024-04-23
Add bad binary file to test

jamesrb1:
The purpose of this is to test changes to the danger file. It will never be merged and I will close it later.


vegaro:
Awesome!

PR #3838, opened 2024-04-24
Update SDK docs link

jamesrb1:
This is in reference to https://revenuecat.slack.com/archives/C03HPN9UE80/p1713915105266199

I'm unsure if the updated link is the correct one (what did this used to point to)?

This outdated link show up in a number of places - two others in this repo and in a number of other repos - once I get confirmation the new link is the correct one, I can update them too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Or view our iOS sample apps:
 | visionOS | 1.0+ |
 
 ## SDK Reference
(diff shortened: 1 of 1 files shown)


vegaro (approved):

tonidero:
I believe the existing link is correct? We don't want to update to the installation docs, since that's a link to the SDK reference docs.

PR #3845, opened 2024-04-24
Replace CircleCI x86 machines with M1

joshdholtz:
### Motivation

CircleCI removed `macos.x86.medium.gen2`

### Description

Replace `macos.x86.medium.gen2` with `macos.m1.medium.gen1`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -582,7 +582,7 @@ jobs:
   run-test-ios-15:
     <<: *base-job
     # Fix-me: running on M1 makes these tests crash
-    resource_class: macos.x86.medium.gen2
+    resource_class: macos.m1.medium.gen1
     steps:
       - checkout
       - install-dependencies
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -785,7 +785,7 @@ jobs:
   release-checks:
     <<: *base-job
     # Fix-me: Carthage can't build fat frameworks on Apple Silicon. See https://github.com/RevenueCat/purchases-ios/pull/3582
-    resource_class: macos.x86.medium.gen2
+    resource_class: macos.m1.medium.gen1
```
tonidero, line 788: We should remove all these comments if it works (I imagine we might need to do some changes?)

.circleci/config.yml
```diff
@@ -582,7 +582,7 @@ jobs:
   run-test-ios-15:
     <<: *base-job
     # Fix-me: running on M1 makes these tests crash
-    resource_class: macos.x86.medium.gen2
+    resource_class: macos.m1.medium.gen1
```
vegaro, line 585: this is already the resource _class in the `base-job` so I think we can just remove this?

vegaro (changes requested):

PR #3853, opened 2024-04-26
Seamless SK2 Observer Mode

fire-at-will:
### Motivation
The current 5.0 betas require developers using observer mode with StoreKit 2 to call `Purchases.shared.handleObserverModeTransaction(result)` after processing a purchase. While this works, it’s an additional step for developers and, if a developer doesn’t include it, will likely lead to purchases being missed. We’ve been looking for a way around this, and until now haven’t been able to come up with one until recently.

This POC demonstrates a method that might allow us to achieve "seamless" observer mode with StoreKit 2 without calling `Purchases.shared.handleObserverModeTransaction(result)`.

### Approach 
In this approach, the SDK detects when the app is foregrounded after being backgrounded by the purchase modals. When it is foregrounded, it performs the following to observe the purchase:
1. Check the on-device cache to obtain a list of transaction IDs that have been previously observed by this method.
2. Get a list of all transactions for the user by querying StoreKit 2’s [Transaction.all](https://developer.apple.com/documentation/storekit/transaction/3851203-all) API.
3. Observe the latest transaction by sending it to Khepri. If there are multiple unsynced transactions, Khepri will find them when it fetches the list of most recent transactions from the list transactions API.

#### Notable New Classes
- `StoreKit2ObserverModeManager`: A class owned by `PurchasesOrchestrator` that determines when observer mode should check for new transactions. If any new transactions are found, it forwards the most recent transaction to the `PurchasesOrchestrator` via its conformance to the `StoreKit2ObserverModeManagerDelegate` protocol to be processed.
- `StoreKit2ObserverModePurchaseDetector`: An actor that checks for unsynced transactions, and if any are found, sends them the provided `StoreKit2ObserverModeManagerDelegate` for processing. It's an actor to prevent any potential race condition schenanigans with how it caches the list of synced transaction IDs. Owned by the `StoreKit2ObserverModeManager`.

### TODO
- Remove the requirement that non-builder based configure functions force the developer to specify a `StoreKitVersion`, and instead just default to StoreKit2. This was intentionally excluded from this PR since it's already pretty large in size.


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1725,6 +1734,44 @@ private extension Purchases {
...
+                        }
+
+                        if !cachedSyncedSK2TransactionIDs.contains(verifiedTransaction.id) && purchaseOrLegacyiOS {
+
+                            do {
+                                try await self.handleObserverModeTransaction(transaction: verifiedTransaction,
+                                                                             jwsRepresentation: transaction.jwsRepresentation)
+
+                                cachedSyncedSK2TransactionIDs.insert(verifiedTransaction.id)
+                                self.deviceCache.cacheSyncedSK2TransactionIDs(syncedSK2TransactionIDs: Array(cachedSyncedSK2TransactionIDs),
```
aboedo, line 1761: This code is fine for a PoC, but in order to productionize we'll have to make it actually work with concurrency. 
We'd have to: 
- ensure that operations are performed atomically (here we're reading, then doing other things including a network operation, then writing original list + new transaction, but with task.detached there are no guarantees that this will work correctly). Easiest fix would be to ensure that all related read/write operations run from the same thread, and that we're performing them atomically when needed. I'd move all the logic to a single class (maybe even device cache) and have a method that only adds a single token to cache instead to make that easier. Let me know if you need a hand with this
- add considerations for the case where the token isn't successfully cached at the last step, although come to think about it the current approach where it'll just get re-synced if the process ends before writing is probably fine

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1725,6 +1734,44 @@ private extension Purchases {
         #endif
     }
 
+    @objc func applicationDidBecomeActive() {
```
aboedo, line 1737: ❤️

aboedo (commented):
👏👏👏

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
```diff
@@ -0,0 +1,109 @@
...
+        } catch {
+            Logger.error(Strings.purchase.sk2_observer_mode_error_processing_transaction(error))
+        }
+    }
+}
+
+/// A wrapper protocol that allows for abstracting out calls to an `AsyncSequence<VerificationResult<Transaction>>`.
+/// This will usually be `Transaction.all` in production but allows us to inject custom AsyncSequences for testing.
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+protocol AllTransactionsProviderType: Sendable {
```
fire-at-will, line 98: I initially tried to inject `Transaction.all` into the `StoreKit2ObserverModePurchaseDetector` directly into the actor's initializer (similar to [how it's done in StoreKit2TransactionLister](https://github.com/RevenueCat/purchases-ios/blob/2cff931fa3ddd7f5471cf2bfc641bc6b27651557/Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift#L79)), and while this compiled, for some reason it caused `Transaction.all` to not return the initial purchase when the app was foregrounded, so I abandoned that approach.

While it's a bit more verbose, this approach avoids that issue!

Sources/Caching/DeviceCache.swift
```diff
@@ -338,6 +338,21 @@ class DeviceCache {
...
+                $0.uint64Value
+            }
+        }
+    }
+
+    func cacheSyncedSK2TransactionIDs(syncedSK2TransactionIDs: [UInt64], appUserID: String) {
+        self.userDefaults.write {
+            $0.set(syncedSK2TransactionIDs, forKey: CacheKey.syncedSK2TransactionIDs(appUserID))
+        }
+    }
```
MarkVillacampa, line 354: Since we're only checking for the most recent transaction now, we store a single value instead of an array?

Sources/Caching/DeviceCache.swift
```diff
@@ -368,6 +384,7 @@ class DeviceCache {
             case let .offerings(userID): return "\(Self.base)offerings.\(userID)"
             case let .legacySubscriberAttributes(userID): return "\(Self.legacySubscriberAttributesBase)\(userID)"
             case let .attributionDataDefaults(userID): return "\(Self.base)attribution.\(userID)"
+            case let .syncedSK2TransactionIDs(userID): return "\(Self.base)syncedSK2TransactionIDs.\(userID)"
```
MarkVillacampa, line 387: I'd add something regarding observer mode to this key so its purpose is clearer (e.g. `latestSyncedSK2ObserverModeTransactionID`)

Tests/BackendIntegrationTests/OtherIntegrationTests.swift
```diff
@@ -173,7 +173,7 @@ class OtherIntegrationTests: BaseBackendIntegrationTests {
         try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
 
         let result = try await self.purchases.productEntitlementMapping().entitlementsByProduct
-        expect(result).to(haveCount(17))
+        expect(result).to(haveCount(18))
```
fire-at-will, line 176: Bringing in the change from https://github.com/RevenueCat/purchases-ios/commit/7f6baa9aa4edd5504e600c155bbdc5315c4336bd to make CI pass

Sources/Misc/SystemInfo.swift
```diff
@@ -261,6 +261,21 @@ extension SystemInfo {
...
+    static var applicationDidBecomeActiveNotification: Notification.Name? {
+        #if os(iOS) || os(tvOS) || VISION_OS || targetEnvironment(macCatalyst)
+            return UIApplication.didBecomeActiveNotification
+        #elseif os(macOS)
+            return NSApplication.didBecomeActiveNotification
+        #elseif os(watchOS)
+            if #available(watchOS 7.0, *) {
+                return WKApplication.didBecomeActiveNotification
+            } else {
+                // There's no equivalent notification available on watchOS <7.
```
MarkVillacampa, line 273: We'll probably need to document the limitation of macOS (does not become active automatically after purchase) and watchOS (only supported for 7.0 and later, while we support 6.2 and later for the SDK)

fire-at-will: Agreed! We should document it in the DocC comments for this property in case it's used for something else in the future.

I don't think it'll be an issue for SK2 observer mode on watchOS - wasn't StoreKit2 first available with watchOS8?

MarkVillacampa: Did you try if we can use [NSExtensionHostDidBecomeActiveNotification](https://developer.apple.com/documentation/foundation/nsextensionhostdidbecomeactivenotification?language=objc) for this use case?

fire-at-will: Yeah,I tried that first before `WKApplication.didBecomeActiveNotification`. Unfortunately it was never fired during the IAP flow so I had to move on to something else. 🤷‍♂️

aboedo: FWIW, IAPs were introduced in watchOS 6.2, and in a fairly broken state (the `appStoreReceiptURL` even points to the wrong file, we added our own workaround for that), with very few apps using it in watchOS < 7. So I wouldn't really stress too much about that limitation.

aboedo: I wouldn't even mind bumping the requirement for watchOS to 7.0, really.

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
```diff
@@ -0,0 +1,109 @@
...
+                return (
+                    verifiedTransaction: verifiedTransaction,
+                    jwsRepresentation: transaction.jwsRepresentation
+                )
+            }
+            .sorted(by: { $0.verifiedTransaction.purchaseDate > $1.verifiedTransaction.purchaseDate })
+            .first
+        ) else {
+            return
+        }
```
aboedo, line 65: I'll be the first to admit that I had to read this a couple of times because I'm not used to the mix of very idiomatic swift features, from the tuple to the chain of functional calls afterwards

aboedo: maybe we can dumb it down or move the tuple to a named struct, or just do one thing at a time?

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
```diff
@@ -0,0 +1,109 @@
...
+                guard let verifiedTransaction = transaction.verifiedTransaction else {
+                    return nil
+                }
+                return (
+                    verifiedTransaction: verifiedTransaction,
+                    jwsRepresentation: transaction.jwsRepresentation
+                )
+            }
+            .sorted(by: { $0.verifiedTransaction.purchaseDate > $1.verifiedTransaction.purchaseDate })
+            .first
```
aboedo, line 62: 🤔 how will this get along with consumables? 
A consumable would show up only until its corresponding transaction is finished, so would this mean that we detect twice per consumable (once upon purchase + one false positive after you foreground the app next time after finishing)?

Sources/Caching/DeviceCache.swift
aboedo: I left a comment in another place, but it being only the most recent transaction scares me a bit since a single transaction can disappear from history if: 
- consumable transaction is finished
- family sharing access is revoked

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
fire-at-will: We decided to get rid of processing everything returned from Transaction.all to avoid making `n` requests to Khepri for `n` number of transactions the first time the app is opened in observer mode. We figured that if we missed uploading one somehow that Khepri would find it when it queries the list transactions API.

I'm not sure off the top of my head how those consumables would be treated but will do some testing this afternoon! :D

fire-at-will: Okay, spent some time testing this this afternoon! Here's how it went:

The consumable showed up in `Transaction.all` before it was finished, so it was processed by observer mode before the app's execution resumed from calling `await purchase()`. Then, the app immediately finished the transaction, the app was background/foregrounded again, and it wasn't in Transaction.all since it was finished. Here was the full timeline:

1. User purchases consumable
2. App is background/foregrounded
3. SDK queries `Transaction.all`
4. SDK sploads transaction to Khepri
5. Execution returns to app
6. App finishes the consumable
7. App is background/foregrounded
8. SDK queries `Transaction.all`, but it's empty and nothing is uploaded

Interestingly, when there was an existing subscription when the consumable was purchased, the consumable was in  `Transaction.all` the second time. It was still only processed once though since its ID was cached.

In both tests, the consumable was only processed once and appeared accurately in the RC dashboard

Sources/Caching/DeviceCache.swift
fire-at-will: @MarkVillacampa and I decided via DM to keep the caching to all of the IDs via an array. Going to resolve this conversation and continue the discussion about consumables/family sharing to the other thread!

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
fire-at-will: Not a problem! I've changed it so it all happens in discrete steps (and got rid of the tuple!) 😄

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
fire-at-will: For the family sharing option, I _think_ it would probably happen in a similar way where it's all processed before execution flow resumes after `await purchase()`, I'll look at it more tomorrow!

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModeManager.swift
```diff
@@ -0,0 +1,78 @@
...
+    func beginObservingPurchases() {
+        if let applicationDidBecomeActiveNotification = SystemInfo.applicationDidBecomeActiveNotification {
+            notificationCenter.addObserver(self,
+                                           selector: #selector(applicationDidBecomeActive),
+                                           name: applicationDidBecomeActiveNotification,
+                                           object: nil)
+        }
+    }
+
+    @objc func applicationDidBecomeActive() {
```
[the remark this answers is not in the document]
fire-at-will: I moved things around so that the app listening is in `subscribeToAppStateNotifications`, and it made it so that I was able to completely remove `StoreKit2ObserverModeManager`! 😄

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
aboedo: > SDK queries Transaction.all, but it's empty and nothing is uploaded

This will only be true when the user has purchased nothing but a single consumable, though, right? 
Like, if I have purchased a subscription, then some time goes by and I purchase a consumable, we will post receipt after the consumable purchase twice: once when the consumable is unfinished, and once again when the consumable is finished, since the latest transaction will be the one for the subscription. 

i.e.: 
1) buy subscription, generate transaction T1
1) foreground app, check history, find transaction T1, send to backend, all is good. Finish transaction. 
1) foreground the app again (like, just open the app). Check history, find transaction T1, not posted to backend because that's the latest one on record. 
1) buy consumable, generate transaction T2
1) foreground app, check history, find transaction T2, send to backend, all is good. Finish transaction. 
1) foreground the app again (like, just open the app). Check history, find transaction T1, post to backend again since the latest one on record is T2 != T1.

fire-at-will: We decided to discard Mark's suggestion to only cache the most recent transaction ID to avoid this problem. The timeline in your last comment is in line with what I'm seeing in my testing, except for step 6:

Here, we find transaction T1, but we don't post it to the backend again since the cache  is an array of each transaction ID observer mode has previously processed. Since the cache contains the IDs `[T1.id, T2.id]`, observer mode won't ever re-post T1 or T2.

Let me know if that makes sense - I'm happy to hop in a huddle to discuss!

jamesrb1 (approved):

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
```diff
@@ -0,0 +1,117 @@
...
+    ) async {
+        let allTransactions = await allTransactionsProvider.getAllTransactions()
+        var verifiedTransactions = allTransactions.filter { transaction in
+            return transaction.verifiedTransaction != nil
+        }
+        verifiedTransactions.sort {
+            $0.verifiedTransaction?.purchaseDate ?? .distantPast > $1.verifiedTransaction?.purchaseDate ?? .distantPast
+        }
+        guard let mostRecentTransaction = verifiedTransactions.first else {
+            return
```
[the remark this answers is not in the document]
fire-at-will: Ooooh that's _fantastic_! :D Will definitely use this 👍

Tests/BackendIntegrationTests/StoreKitObserverModeIntegrationTests.swift
```diff
@@ -51,25 +51,34 @@ class StoreKit2ObserverModeIntegrationTests: StoreKit1ObserverModeIntegrationTes
 
         try self.testSession.disableAutoRenewForTransaction(identifier: UInt(transaction.id))
 
-        _ = try await Purchases.shared.handleObserverModeTransaction(result)
+        try await simulateAppDidBecomeActive()
+        try await Task.sleep(nanoseconds: 3_000_000_000)    // 3 sec
```
vegaro, line 55: Can we maybe observe customer info changes instead of sleeping here? Feels like this could potentially make the test flaky?

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1481,3 +1489,26 @@ extension PurchasesOrchestrator {
     }
 
 }
+
+// MARK: - Application Lifecycle
+extension PurchasesOrchestrator {
+    func handleApplicationDidBecomeActive() {
+        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *),
+           self.observerMode && self.systemInfo.storeKitVersion == .storeKit2 {
+            Task(priority: .high) {
```
vegaro, line 1498: So this will get notified every time the app becomes active, not only after becoming active after a purchase, do you think there could be any problem because we are checking for purchases everytime the app becomes active? I think it should be fine and it will actually check for purchases made outside the app in the Appstore not only via a regular purchase flow.

fire-at-will: I haven't been able to think of any reasons that it should cause any issues yet. The app shouldn't be backgrounded/foregrounded super quickly (like in a loop or something), so we won't be spamming SK2 with repeated requests, and each time this is called, we're only making at most 1 network request our backend, so I don't see an issue there either.

And yeah, I love that this will pick up transactions made outside the app! 😄

aboedo: not sure about task priority being high for observer mode, though, tbh. 
I wish there were better docs as to how this actually translates, but given that in observer mode you'd probably not be using our Entitlements in the first place, I kinda feel like `utility` might be a better fit?

aboedo: I mean, we can go with `high` and it's more likely than not that we'll be happy about it, I just want to ensure that we're being good citizens in people's apps

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
```diff
@@ -0,0 +1,124 @@
...
+    ) {
+        self.deviceCache = deviceCache
+        self.currentUserProvider = currentUserProvider
+        self.allTransactionsProvider = allTransactionsProvider
+    }
+
+    /// Detects unobserved transactions and forwards them to the StoreKit2ObserverModeManagerDelegate
+    /// for processing.
+    func detectUnobservedTransactions(
+        delegate: StoreKit2ObserverModePurchaseDetectorDelegate?
```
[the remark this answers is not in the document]
aboedo: if we're expecting it to never be `nil`, we should maybe look for a safer approach

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
```diff
@@ -0,0 +1,117 @@
...
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+protocol StoreKit2ObserverModePurchaseDetectorType {
+    func detectUnobservedTransactions(
+        delegate: StoreKit2ObserverModePurchaseDetectorDelegate?
+    ) async
+}
+
+/// Actor responsibile for detecting purchases from StoreKit2 that should be processed by observer mode.
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+actor StoreKit2ObserverModePurchaseDetector: StoreKit2ObserverModePurchaseDetectorType {
```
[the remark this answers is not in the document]
aboedo: it would have been such a pain without actors 🙌

Tests/BackendIntegrationTests/StoreKitObserverModeIntegrationTests.swift
fire-at-will: Absolutely, I've modified it to work without the sleep! 💪

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
fire-at-will: I think that `utility` is appropriate given that observer mode usually isn't in most app's (if any) critical paths. I'll make the change!

RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -4415,7 +4447,7 @@
 				CLANG_WARN_QUOTED_INCLUDE_IN_FRAMEWORK_HEADER = YES;
 				CODE_SIGN_STYLE = Automatic;
 				DEVELOPMENT_ASSET_PATHS = "\"Tests/BackendIntegrationTestApp/Preview Content\"";
-				DEVELOPMENT_TEAM = "";
+				DEVELOPMENT_TEAM = 8SXR2327BM;
```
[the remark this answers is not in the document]
fire-at-will: No, this and the `Package.resolved` changes somehow slipped into one of my commits. I'll undo them!

Sources/Caching/DeviceCache.swift
```diff
@@ -338,6 +338,23 @@ class DeviceCache {
         return self.userDefaults.read(Self.productEntitlementMapping)
     }
 
+    // MARK: - SK2
+    func cachedSyncedSK2ObserverModeTransactionIDs(appUserID: String) -> [UInt64]? {
+        return self.userDefaults.read { userDefaults in
+            userDefaults.array(
+                forKey: CacheKey.syncedSK2ObserverModeTransactionIDs(appUserID).rawValue) as? [UInt64]
```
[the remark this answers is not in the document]
fire-at-will: The only reason I tied it to the app user ID is that it looked like several other items in the `DeviceCache` were doing it. Your point about the ghost transfers sounds like something that would probably happen with this though, so I'll definitely remove it. Great catch! 🤩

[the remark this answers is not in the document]
fire-at-will: I also love that this removes the `StoreKit2ObserverModePurchaseDetector`'s dependency on `CurrentUserProvider` 😎

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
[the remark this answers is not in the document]
fire-at-will: I've made the delegate a non-optional param here! If we want to simplify this further, this could be made into a callback param, but I think it's probably fine the way it is.

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
```diff
@@ -0,0 +1,124 @@
...
+
+        var cachedSyncedSK2ObserverModeTransactionIDs = Set(
+            self.deviceCache.cachedSyncedSK2ObserverModeTransactionIDs(
+                appUserID: currentUserProvider.currentAppUserID
+            ) ?? []
+        )
+
+        if cachedSyncedSK2ObserverModeTransactionIDs.contains(transaction.id) { return }
+
+        do {
```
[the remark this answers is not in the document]
fire-at-will: > This only handles the case where there is a single unknown transaction, and it is the latest one, right?

Yeah, that's right! The idea is that when it's sent up to the backend, the backend will pick up any other unprocessed transactions when it refreshes the receipt. That way, even if a transaction is skipped over by observer mode, it's still detected by the backend.

If we were to send up all unprocessed transactions, that'd probably result in some unnecessary network requests.

If we wanted to loosen this up a bit, we could make it send up the most recent unprocessed transaction, even if it isn't the newest in the queue instead of only transactions that are the most recent transaction in the queue **and** unprocessed.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1481,3 +1489,26 @@ extension PurchasesOrchestrator {
     }
 
 }
+
+// MARK: - Application Lifecycle
+extension PurchasesOrchestrator {
+    func handleApplicationDidBecomeActive() {
```
aboedo, line 1495: total nitpick, but to match [our Style Guide](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/SwiftStyleGuide.swift) we should have line breaks right after the start and before the end of these definitions

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
```diff
@@ -0,0 +1,111 @@
...
+//
+//  StoreKit2ObserverModePurchaseDetector.swift
+//
+//  Created by Will Taylor on 5/1/24.
+
+import Foundation
+import StoreKit
+
+/// A delegate protocol for handling verified transactions in observer mode.
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
```
aboedo, line 18: With Swift Macros we'll eventually be able to re-replace these with stuff like `"STOREKIT_2_COMPATIBLE_OS"` right? I know they're not as powerful as C pre-processor macros so I've been wondering

aboedo: we used to have that back when the SDK was ObjC and I definitely miss it

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
```diff
@@ -0,0 +1,111 @@
...
+    /// Detects unobserved transactions and forwards them to the StoreKit2ObserverModeManagerDelegate
+    /// for processing.
+    func detectUnobservedTransactions(
+        delegate: StoreKit2ObserverModePurchaseDetectorDelegate
+    ) async {
+        guard let mostRecentTransaction = await allTransactionsProvider.getMostRecentVerifiedTransaction() else {
+            return
+        }
+
+        let jwsRepresentation = mostRecentTransaction.jwsRepresentation
```
aboedo, line 64: The old comment is now hidden because the code changed, but it's a good point that in practice we only need to detect a single change in the history to tell the backend and it can go and find any / all transactions that might have been missed. 

One thing that I do still wonder about, wouldn't we still have false positives if multiple transactions happen while the app is closed? 

I.e.:
1. make purchase, T1
1. come back after a few months. T2, T3 happened
1. re-open app, find T3, post, mark it synced
1. buy consumable, T4, sync, mark it synced. 
1. re-open app, we look at latest, T4, sync, mark it synced. T3 is now latest.
1. T3 revoked through family sharing
1. T2 is now latest, we post it, but it's been synced since step 3


Obviously this is a very edge-casey thing, but I imagine there's probably a lot of cases that I'm not thinking about off the top of my head. 

Maybe the safest approach is to consider all existing jwts at the time of posting the receipt as synced? 

In terms of how it differs to this approach, it'd mean: 
- instead of getting the most recent transaction, get all transactions and see if there are _any_ that we haven't marked synced (we're already fetching all from StoreKit and DeviceCache so it wouldn't be less efficient other than the in-memory comparison of two sets). 
- store the list of all "unsynced" transaction ids
- after getting a <5xx response from the backend, mark all unsynced transaction ids as synced

You can see an example of us doing a similar thing (minus the StoreKit part) that for subscriber attributes: 
https://github.com/RevenueCat/purchases-ios/blob/main/Sources/SubscriberAttributes/SubscriberAttributesManager.swift#L148

aboedo (commented):
Looking great!

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
fire-at-will: I think this could also occur with a consumable that is finished in addition to a revoked family sharing transaction. I'll  get this implemented!

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
fire-at-will: That we be amazing - it'd be way too easy to accidentally copy/paste the wrong one somewhere. Hopefully we can make that work with Macros!

Sources/Purchasing/StoreKit2/Observer Mode/StoreKit2ObserverModePurchaseDetector.swift
fire-at-will: @aboedo We're now marking all found transactions as synced to avoid this edge case, see https://github.com/RevenueCat/purchases-ios/pull/3853/commits/8a470de1a4183dec17c71632a63ec1666cc7b1ef 😎

aboedo: would it still make sense to look at all transactions from SK2 to find an unsynced one rather than latest in this case?

aboedo (approved):
great work on this!

PR #3855, opened 2024-04-29
[v5 migration guide] simplified observer mode configuration calls

aboedo:
simplified a couple of calls to configure the SDK in the migration guide


Sources/DocCDocumentation/DocCDocumentation.docc/V5_API_Migration_guide.md
```diff
@@ -39,9 +39,7 @@ If you're using the Firebase SDK, you'll need to follow [these instructions](htt
 Version 5.0 of the SDK introduces support for observer mode when making purchases with StoreKit 2. You can enable it when configuring the SDK:
 
 ```swift
-Purchases.configure(with: .builder(withAPIKey: apiKey)
-		.with(observerMode: true, storeKitVersion: .storeKit2)
-    .build()
+Purchases.configure(withAPIKey: apiKey, observerMode: true, storeKitVersion: .storeKit2)
```
fire-at-will, line 42: We don't currently have a configure function with this signature - currently if you want to use observerMode with this type of configure function you need to provide an appUserID. Perhaps we should add a configure function with this signature in the future, but for now this won't build.

FWIW, you can configure observer mode without an appUserID, but you need to use the builder pattern:
```swift
Purchases.configure(with: .builder(withAPIKey: apiKey)
         .with(observerMode: true, storeKitVersion: .storeKit2)
         .build()
     )
```

```suggestion
Purchases.configure(withAPIKey: apiKey, appUserID: appUserID, observerMode: true, storeKitVersion: .storeKit2)
```

Sources/DocCDocumentation/DocCDocumentation.docc/V5_API_Migration_guide.md
```diff
@@ -54,9 +52,7 @@ Purchases.shared.handleObserverModeTransaction(result)
 If you're using observer mode with StoreKit 1, you will need to explicitly configure the SDK to use StoreKit 1:
 
 ```swift
-Purchases.configure(with: .builder(withAPIKey: apiKey)
-		.with(observerMode: true, storeKitVersion: .storeKit1)
-    .build()
+Purchases.configure(withAPIKey: apiKey, observerMode: true, storeKitVersion: .storeKit1)
```
fire-at-will, line 55:
```suggestion
Purchases.configure(withAPIKey: apiKey, appUserID: appUserID, observerMode: true, storeKitVersion: .storeKit1)
```

fire-at-will (changes requested):
I made a small change, but otherwise looks good to me!

We should consider adding a `configure` function with the signature `Purchases.configure(withAPIKey:observerMode:storeKitVersion:)` - it makes a lot of sense to use observer mode with anonymous users and it'd be nice to do that without using the builder configure function.

PR #3850, opened 2024-04-25
Paywalls Tester: Universal links (#1 in queue)

jamesrb1:
This adds support for universal links.

Format was
app.revenuecat.com/paywallstester/?pw=ofrng71bdfc2037

but should soon be
app.revenuecat.com/pwt/?pw=ofrng71bdfc2037

QR code for the second one:

image

PR #3856, opened 2024-04-29
PaywallsTester: Improves OfferingList UI  (#2 in queue)

jamesrb1:
Base branch is https://github.com/RevenueCat/purchases-ios/pull/3850

This improves the Offerings list UI:

- Offerings with paywalls are grouped into a single section
- Offering paywall buttons are named by their offering, with the sub-title containing the template number and main content.
- Paywall buttons get a ••• menu that is the same as the long-press context menu, but easier to discover
- If there are offerings without configured paywalls, they are listed in a separate section.
- Offerings have links out to the revenue cat dashboard for quick edits (or even to create a new one), which may be handy for use on the phone or for sending the URL to a desktop.

The PR also contain quite a bit of refactoring - view decomposition, simplification in places as well as better data organization and flow.

Before:

https://github.com/RevenueCat/purchases-ios/assets/109382862/fc4abf5c-b779-41fa-be61-bcd885b74fc9

After:

https://github.com/RevenueCat/purchases-ios/assets/109382862/7f7a3c77-08e5-4b89-9cc0-309ea9841bb5

### DIFF
--- a/Tests/TestingApps/PaywallsTester/PaywallsTester/Data/OfferingsPaywallsViewModel.swift
+++ b/Tests/TestingApps/PaywallsTester/PaywallsTester/Data/OfferingsPaywallsViewModel.swift
@@ -68,8 +68,10 @@ final class OfferingsPaywallsViewModel {
             let paywalls = try await appPaywalls
 
             let offeringPaywallData = OfferingPaywallData(offerings: offerings, paywalls: paywalls)
-            let listData = PaywallsData(offeringsAndPaywalls: offeringPaywallData.paywallsByOffering(), offeringsWithoutPaywalls: offeringPaywallData.offeringsWithoutPaywalls())
-            self.hasMultipleTemplates = Set(listData.offeringsAndPaywalls.map { $0.paywall.data.templateName }).count > 1
+            let listData = PaywallsData(offeringsAndPaywalls: offeringPaywallData.paywallsByOffering(),
+                                        offeringsWithoutPaywalls: offeringPaywallData.offeringsWithoutPaywalls())
+            let templateNames = listData.offeringsAndPaywalls.map { $0.paywall.data.templateName }
+            self.hasMultipleTemplates = Set(templateNames).count > 1
             self.hasMultipleOfferingsWithPaywalls = listData.offeringsAndPaywalls.count > 1
             self.listData = listData
             self.state = .success
--- a/Tests/TestingApps/PaywallsTester/PaywallsTester/Data/OfferingsPaywallsViewModel.swift
+++ b/Tests/TestingApps/PaywallsTester/PaywallsTester/Data/OfferingsPaywallsViewModel.swift
@@ -61,9 +61,6 @@ final class OfferingsPaywallsViewModel {
             let listData = PaywallsListData(offeringsAndPaywalls: offeringPaywallData.paywallsByOffering(), offeringsWithoutPaywalls: offeringPaywallData.offeringsWithoutPaywalls())
 
             self.listData = .success(listData)
-
-
-
         } catch let error as NSError {
             self.listData = .failure(error)
             Self.logger.log(level: .error, "Could not fetch offerings/paywalls: \(error)")
--- a/Tests/TestingApps/PaywallsTester/PaywallsTester/Data/OfferingsPaywallsViewModel.swift
+++ b/Tests/TestingApps/PaywallsTester/PaywallsTester/Data/OfferingsPaywallsViewModel.swift
@@ -8,6 +8,11 @@
 import Foundation
 import RevenueCat
 
+struct PaywallsListData: Hashable {
+    let offeringsAndPaywalls: [OfferingPaywall]
+    let offeringsWithoutPaywalls: [OfferingsResponse.Offering]
+}
+
 struct OfferingPaywall: Hashable {
     let offering: OfferingsResponse.Offering
     let paywall: PaywallsResponse.Paywall
@@ -22,13 +27,16 @@ struct PresentedPaywall: Hashable {
 @Observable
 final class OfferingsPaywallsViewModel {
 
-    var apps: [DeveloperResponse.App]
+    var singleApp: DeveloperResponse.App? {
+        guard apps.count == 1 else { return nil }
+        return apps.first
+    }
 
     init(apps: [DeveloperResponse.App]) {
         self.apps = apps
     }
 
-    var offeringsPaywalls: Result<[OfferingPaywall], NSError>? {
+    var listData: Result<PaywallsListData, NSError>? {
         didSet {
             Task { @MainActor in
                 refreshPresentedPaywall()
@@ -50,12 +58,14 @@ final class OfferingsPaywallsViewModel {
 
             let offeringPaywallData = OfferingPaywallData(offerings: offerings, paywalls: paywalls)
 
-            self.offeringsPaywalls = .success(
-                offeringPaywallData.paywallsByOffering()
-            )
+            let listData = PaywallsListData(offeringsAndPaywalls: offeringPaywallData.paywallsByOffering(), offeringsWithoutPaywalls: offeringPaywallData.offeringsWithoutPaywalls())
+
+            self.listData = .success(listData)
+
+
 
         } catch let error as NSError {
-            self.offeringsPaywalls = .failure(error)
+            self.listData = .failure(error)
             Self.logger.log(level: .error, "Could not fetch offerings/paywalls: \(error)")
         }
     }
@@ -72,6 +82,8 @@ final class OfferingsPaywallsViewModel {
     }
 
     private static var logger = Logging.shared.logger(category: "Paywalls Tester")
+
+    private var apps: [DeveloperResponse.App]
 
 }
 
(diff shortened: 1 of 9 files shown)


Tests/TestingApps/PaywallsTester/PaywallsTester/Data/OfferingsPaywallsViewModel.swift
```diff
@@ -49,30 +68,37 @@ final class OfferingsPaywallsViewModel {
             let paywalls = try await appPaywalls
 
             let offeringPaywallData = OfferingPaywallData(offerings: offerings, paywalls: paywalls)
-
-            self.offeringsPaywalls = .success(
-                offeringPaywallData.paywallsByOffering()
-            )
-
+            let listData = PaywallsData(offeringsAndPaywalls: offeringPaywallData.paywallsByOffering(), offeringsWithoutPaywalls: offeringPaywallData.offeringsWithoutPaywalls())
```
aboedo, line 71: nitpick: let's keep line lengths at <= 120 chars so we can control the wrapping a little nicer than github does. I'd recommend setting up Xcode's "Page guide at column 120" to make it easier. 
PaywallsTester is a bit of an exception but the rest of our SDK has swiftlint set up to enforce it

Tests/TestingApps/PaywallsTester/PaywallsTester/Data/OfferingsPaywallsViewModel.swift
```diff
@@ -84,32 +110,41 @@ extension OfferingsPaywallsViewModel {
...
-    private func showPaywallForID(_ id: String) {
-        switch self.offeringsPaywalls {
-        case let .success(data):
-            if let dataForRequestedID = data.first(where: { $0.offering.id == id }) {
-                let newRCOffering = dataForRequestedID.paywall.convertToRevenueCatPaywall(with: dataForRequestedID.offering)
+    private func showPaywallForID(_ id: String, mode: PaywallViewMode = .default) {
+        switch self.state {
+        case .notloaded:
+            Self.logger.log(level: .info, "Could not show paywall for id \(id), data not loaded.")
+            self.presentedPaywall = nil
```
aboedo, line 139: from a user's perspective, are we displaying any kind of an error message? A log wouldn't be enough if the user isn't running it from Xcode

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/ManagePaywallButton.swift
```diff
@@ -0,0 +1,92 @@
...
+        case new
+
+        var systemImageName: String {
+            switch self {
+            case .edit:
+                return "slider.horizontal.2.square.on.square"
+            case .new:
+                return "plus.square.on.square"
+            }
+        }
```
aboedo, line 23: if we're going in this direction, might as well lean in and also add a `var buttonName: String` so you don't need the `buttonName ?? "Edit Paywall"` line later on

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/ManagePaywallButton.swift
```diff
@@ -0,0 +1,92 @@
...
+                    Spacer()
+                    Image(systemName: kind.systemImageName)
+                }
+            case .new:
+                HStack {
+                    Text(buttonName ?? "New Paywall")
+                        .font(.headline)
+                    Spacer()
+                    Image(systemName: "escape")
+                        .rotationEffect(Angle(degrees: 90))
```
aboedo, line 45: should this be `Image(systemName: kind.systemImageName)` instead? 
If so, you could unify the name into a var (see previous suggestion) and then remove the `switch` here, where the same button works for both kinds

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/OfferingButton.swift
```diff
@@ -0,0 +1,101 @@
...
+    }
+}
+
+private extension OfferingButton {
+
+    @ViewBuilder
+    private func label() -> some View {
+        let templateName = rcOffering.paywall?.templateName
+        let paywallTitle = rcOffering.paywall?.localizedConfiguration.title
+        let decorator = viewModel.hasMultipleOfferingsWithPaywalls && self.selectedItemID == responseOffering.id ? "▶ " : ""
```
aboedo, line 56: 👍

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/OfferingButton.swift
```diff
@@ -0,0 +1,101 @@
...
+        HStack {
+            VStack(alignment:.leading, spacing: 5) {
+                Text(decorator + responseOffering.displayName)
+                    .font(.headline)
+                if let title = paywallTitle, let name = templateName {
+                    let text = viewModel.hasMultipleTemplates ? "Style \(name): \(title)" : title
+                    Text(text)
+                        .font(.footnote)
+                        .foregroundColor(.secondary)
```
aboedo, line 65: super unrelated, but... we should maybe think about updating the tint colors to RevenueCat Red rather than the orange that we added I can't remember why

aboedo (approved):

Tests/TestingApps/PaywallsTester/PaywallsTester/Data/OfferingsPaywallsViewModel.swift
jamesrb1: Done!

Tests/TestingApps/PaywallsTester/PaywallsTester/Data/OfferingsPaywallsViewModel.swift
jamesrb1: I haven't put in any user-facing error, good idea, will do so as part of another PR. I made [a linear ticket](https://linear.app/revenuecat/issue/PWL-458/user-facing-error-dialogs).

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/ManagePaywallButton.swift
jamesrb1: Good idea, I added a new default name here. We need to keep the `??` though, because the constructor takes an optional button name, which we use for Offerings without paywalls, where the button name is the name of the Offering.

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/ManagePaywallButton.swift
jamesrb1: This button in "edit" mode is used in two locations, on the main screen for Offerings without paywalls, and also in the Menu that you get by tapping the ellipsis on the Offerings with Paywalls. It uses a different icon in each location. "escape" is used in the Offerings list and I like it best because it communicates that it is going to exit the app. Unfortunately the `.rotationEffect` modifier has not effect when used on an Image in Menu, and without rotating it the icon points up and to the left, which looks wrong. So I kept it for the Offerings list, where I can rotate it, and used a different icon in the Menu, and coding it as the "standard" icon because it will work everywhere.

The best thing to do here (imo) would be to create a custom escape icon that is pre-rotated that we can use in both locations. But I didn't think it was worth the time at this point.

Thoughts?

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/OfferingButton.swift
jamesrb1: I had thought the same thing! Works in both light and dark mode. Done!
image
image

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/ManagePaywallButton.swift
aboedo: I'd agree that it's not worth the effort. It did feel pretty awkward to me to see a `Kind` enum that includes a system image and then it not being used for 1 of the 2 possible values

aboedo: maybe it should be a third kind?

jamesrb1: Actually some of what I said was outdated! Reworked this, much better, no more switching the view body. https://github.com/RevenueCat/purchases-ios/pull/3856/commits/9ee4f800633bdca325d5a47f9479e04bee76e2d7

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/OfferingButton.swift
aboedo: nice, love it!!

PR #3858, opened 2024-04-30
Change deploy-purchase-tester to use xcode 15

vegaro:
It is now required to use xcode 15 to push to the store

https://developer.apple.com/news/?id=fxu2qp7b

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1298,7 +1298,7 @@ workflows:
           xcode_version: '14.3.0'
           <<: *release-tags
       - deploy-purchase-tester:
-          xcode_version: '14.3.0'
+          xcode_version: '15.2'
           <<: *release-tags
       - notify-on-non-patch-release-branches:
           requires:


.circleci/config.yml
```diff
@@ -1298,7 +1298,7 @@ workflows:
           xcode_version: '14.3.0'
           <<: *release-tags
       - deploy-purchase-tester:
-          xcode_version: '14.3.0'
+          xcode_version: '15.2'
```
tonidero, line 1301: I wonder if we should update all the others (except the tests in older versions of course)... I think ideally we default to this one and just overwrite it where needed... But this looks good for now!

tonidero (approved):

.circleci/config.yml
vegaro: yes, I think we should. This is the one that failed when deploying so fixing this for now

vegaro: https://github.com/RevenueCat/purchases-ios/pull/3870

PR #3863, opened 2024-05-01
Paywalls Tester: Intro eligibility picker (#3 in Queue)

jamesrb1:
* Adds the ability to toggle intro offers. The user's selection is persisted in UserDefaults. (I would like to add the ability to edit details that would normally come from the App Store, but this is not part of this PR.)
* The template type is now communicated via its name rather than number, and as a pill next to the offering name.
* Only two tabs are shown (The App and Upsell tabs are disabled, am unsure what we want to do with these.)

Resolves PWL-446

image


Tests/TestingApps/PaywallsTester/PaywallsTester/SimpleApp.swift
```diff
@@ -64,6 +71,18 @@ extension SimpleApp {
...
+            if value == "1" {
+                return .eligible
+            } else if value == "0" {
+                return .ineligible
+            } else if value == "n" {
+                return .noIntroOfferExists
+            } else {
+                return .unknown
+            }
+        } ?? .unknown
```
aboedo, line 84: feels like this is complex / specific enough to be in its own function? 
I'm also curious about the naming for these - we're going with the shortest possible values because these are deeplinks? If so, maybe we can make them constants or something for readability rather than magic strings? 

I'm kinda torn on the suggestion because adding indirection might not necessarily make this easier to read in any case, given that a simple if / else block here is still fairly clear

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/AppContentView.swift
```diff
@@ -25,7 +25,7 @@ struct AppContentView: View {
 
     var body: some View {
         TabView {
-            if Purchases.isConfigured {
+            if false && Purchases.isConfigured {
```
aboedo, line 28: I'd add a comment above this explaining the reasoning behind it being disabled, otherwise to a maintainer this looks like an accident

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/AppContentView.swift
```diff
@@ -51,8 +53,7 @@ struct AppContentView: View {
                     Label("My Apps", systemImage: "network")
                 }
 
-            if Purchases.isConfigured {
-
+            if false && Purchases.isConfigured {
```
aboedo, line 56: same thought here

aboedo: in terms of what to do with these... I'd just kill them. The App tab was a workaround because we didn't have log in / log out, and the Upsell tab was just a quick way to show a single paywall, but it's fairly useless

aboedo: (I'd probably do it on a separate PR though)

aboedo (approved):

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/AppContentView.swift
jamesrb1: Done!

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/AppContentView.swift
jamesrb1: OK, comment left, will delete in future!

Tests/TestingApps/PaywallsTester/PaywallsTester/SimpleApp.swift
aboedo: oh I like that. Yeah, I kinda figured we're going for min length for deeplink purposes, sounds good

PR #3866, opened 2024-05-03
CAT-1509: Update SDK to reflect new SK2 Observer Mode Changes

fire-at-will:
### Motivation
In light of https://github.com/RevenueCat/purchases-ios/pull/3853, we no longer require developers to call `handleObserverModePurchase()` when using Observer Mode with StoreKit2. To reflect these changes, a few changes needed to be made to the SDK's logging and documentation

### Changes:
- Removed the warning log that appears when a developer uses StoreKit 2 with observer mode, instructing them to call `handleObserverModePurchase()` after making a purchase.
- Updated documentation to reflect these changes, including the Version 5 migration guide.
- Updated the tests to reflect these changes

# Other Notes
This PR should only be merged if https://github.com/RevenueCat/purchases-ios/pull/3853 is also merged.


aboedo:
Sorry to keep going back and forth on this, but I'm confused: 

> Un-obsoleted the configure(apiKey:appUserID:observerMode) function and changed it to use the .default StoreKitVersion (which currently uses StoreKit 2)

Would this lead to the worst-case scenario? We need to know the version in order to set up the observing code, and if we default to SK2 then we're assuming that the customer's implementation is also SK2, which the vast majority of the time will be wrong. 

Also we would make that API _not_ breaking, which in turn means they wouldn't even know about it, right? 

Shouldn't we be forcing anyone who's using observer mode to tell us which version to use?

fire-at-will:
@aboedo You're totally right! I've updated the PR to leave the observer mode `configure` functions/builders alone and just update the documentation and log messages 😄

Sources/Misc/Obsoletions.swift
```diff
@@ -548,18 +548,18 @@ public extension Purchases {
...
+               renamed: "configure(withAPIKey:appUserID:observerMode:storeKitVersion:)")
+    @available(tvOS, obsoleted: 1,
+               message: "Explicitly setting the StoreKit version is now required.",
+               renamed: "configure(withAPIKey:appUserID:storeKitVersion:)")
+    @available(watchOS, obsoleted: 1,
+               message: "Explicitly setting the StoreKit version is now required.",
+               renamed: "configure(withAPIKey:appUserID:observerMode:storeKitVersion:)")
+    @available(macOS, obsoleted: 1,
+               message: "Explicitly setting the StoreKit version is now required.",
+               renamed: "configure(withAPIKey:appUserID:observerMode:storeKitVersion:)")
```
aboedo, line 562: may be we could have the message explicitly say "for observer mode", i.e.: `Explicitly setting the StoreKit version is now required when using observer mode.`

Tests/APITesters/SwiftAPITester/SwiftAPITester/PurchasesAPI.swift
```diff
@@ -299,13 +299,16 @@ func checkNonAsyncMethods(_ purchases: Purchases) {
...
+        .with(observerMode: true, storeKitVersion: .default)
+        .build())
     Purchases.configure(with: Configuration.Builder(withAPIKey: "")
         .with(showStoreMessagesAutomatically: false)
         .build())
 
     Purchases.configure(withAPIKey: "")
     Purchases.configure(withAPIKey: "", appUserID: nil)
-    Purchases.configure(withAPIKey: "", appUserID: nil, observerMode: true, storeKitVersion: .storeKit2)
+    Purchases.configure(withAPIKey: "", appUserID: nil, observerMode: true, storeKitVersion: .default)
```
aboedo, line 311: almost wonder whether having `.default` be an option for this specific API is dangerous and we should limit the options to 1 vs 2 with a different enum

aboedo (approved):

PR #3868, opened 2024-05-03
Paywalls Tester 0.1

jamesrb1:
This code has been reviewed in other branches.


aboedo:
I'd recommend removing the `feat` tag and replacing it with `test`, since it's technically not going to be a new feature in the SDK itself but rather a better way to test it. 

In practice, the difference will be that `test` implies a patch rather than a minor version in terms of the release train, and it'll show up in the "other" section of the changelog rather than as an API change

aboedo (approved):

PR #3869, opened 2024-05-03
Paywalls Tester: Use key defined in CI

jamesrb1:
Use key defined in Xcode Cloud.

### DIFF
--- a/Tests/TestingApps/PaywallsTester/ci_scripts/ci_pre_xcodebuild.sh
+++ b/Tests/TestingApps/PaywallsTester/ci_scripts/ci_pre_xcodebuild.sh
@@ -12,6 +12,7 @@ else
     echo "Replacing API keys on PaywallsTester"
 
     file="$SCRIPT_DIR/../PaywallsTester/Configuration.swift"
+    sed -i.bak 's/private static let apiKey = ""/private static let apiKey = "'$REVENUECAT_XCODE_CLOUD_SIMPLE_APP_API_KEY'"/g' $file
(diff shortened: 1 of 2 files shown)


Tests/TestingApps/PaywallsTester/ci_scripts/ci_pre_xcodebuild.sh
```diff
@@ -12,6 +12,7 @@ else
     echo "Replacing API keys on PaywallsTester"
 
     file="$SCRIPT_DIR/../PaywallsTester/Configuration.swift"
+    sed -i.bak 's/private static let apiKey = ""/private static let apiKey = "'$REVENUECAT_XCODE_CLOUD_SIMPLE_APP_API_KEY'"/g' $file
```
tonidero, line 15: Should we still be using the `Simple App` naming for this? I think we decided to change it... Also, I see we have a couple existing API keys: `REVENUECAT_XCODE_CLOUD_SIMPLE_APP_API_KEY_FOR_TESTING` and `REVENUECAT_XCODE_CLOUD_SIMPLE_APP_API_KEY_FOR_DEMOS`. Can we reuse one of those?

tonidero (approved):
Just a question but looks good!

Tests/TestingApps/PaywallsTester/ci_scripts/ci_pre_xcodebuild.sh
jamesrb1: This isn't a new key, it was defined as well as the others as part of the Xcode cloud build environment. I guess if we're truly done with the others (probably?) we could delete them and the code that uses them, and rename the key while we're at it.

image

PR #3874, opened 2024-05-07
Add NSPrivacyAccessedAPICategoryFileTimestamp to privacy report

vegaro:
As reported in https://github.com/RevenueCat/purchases-flutter/issues/1064 we were missing an entry

Added C617.1 as reason which reads:

> Declare this reason to access the timestamps, size, or other metadata of files inside the app container, app group container, or the app’s CloudKit container.

https://developer.apple.com/documentation/bundleresources/privacy_manifest_files/describing_use_of_required_reason_api#4278393

### DIFF
--- a/Sources/PrivacyInfo.xcprivacy
+++ b/Sources/PrivacyInfo.xcprivacy
@@ -6,6 +6,14 @@
 	<array>
 		<dict>
 			<key>NSPrivacyAccessedAPIType</key>
+			<string>NSPrivacyAccessedAPICategoryFileTimestamp</string>
+			<key>NSPrivacyAccessedAPITypeReasons</key>
+			<array>
+				<string>C617.1</string>
+			</array>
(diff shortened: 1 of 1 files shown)


tonidero:
Hmm interesting... From what I see, it seems that it's detecting we need that permission because of the `AppleReceipt` and `AppleReceiptBuilder` containing the string `creationDate`, even though that does not come from a file... I think we don't really need this permission right now IMO, but maybe someone else can correct me if I'm wrong? Not sure how to disable that warning though, so maybe this is the easiest way...

vegaro:
I see what you mean. Maybe we don't need to add it yes.. I believe we do need the phc one though

aboedo:
+1, I believe that's a false positive from the detection script, but we do need the privacy manifest for PHC

vegaro:
Closing this for now

Issue #3871, opened 2024-05-06
macOS App hanging for at least 2000 ms because of check for macOS app running on the main thread.


jb-apps:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

1. Environment
   1. Platform: macOS
   2. SDK version: 4.40.1
   3. StoreKit version:
      - [ ] StoreKit 1
      - [x] StoreKit 2 (_enabled with `usesStoreKit2IfAvailable(true)`_)
   4. OS version: Sonoma 14.1.1 
   5. Xcode version: 15.3 
   6. Device and/or simulator:
      - [x] Device
      - [ ] Simulator
   7. Environment:
      - [ ] Sandbox
      - [ ] TestFight
      - [x] Production
   8. How widespread is the issue. Percentage of devices affected. 100% users.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue. **Complete logs with `Purchases.logLevel = .verbose` will help us debug this issue.**

```
All good in debug, it only happens in production and in the App Store.
```

3. Steps to reproduce, with a description of expected vs. actual behavior

App hangs at [status = SecStaticCodeCheckValidity](https://github.com/RevenueCat/purchases-ios/blob/5b20b57cfa5236ae97c888a9b9d54214c0f4ede5/Sources/Misc/MapAppStoreDetector.swift#L61)
<img width="951" alt="Screenshot 2024-05-06 at 22 28 15" src="https://github.com/RevenueCat/purchases-ios/assets/7238946/d936bcaa-d96f-4667-aad0-0848f2fb3f96">

It throws the following warning when running locally:
<img width="971" alt="Screenshot 2024-04-22 at 15 37 56" src="https://github.com/RevenueCat/purchases-ios/assets/7238946/e54d35ea-af97-41a6-82d4-007c8fe022fe">

[...]

PR #3875, opened 2024-05-07
fix mac os sandbox check slowness

aboedo:
Addresses #3871 

Solves the issue by ensuring that the calls to `isSandbox` are performed on a worker thread, and that we're pre-loading the values. 

I'm not a huge fan of this solution since it isn't technically perfectly safe, meaning that it's possible (albeit very unlikely) for us to still check sandbox values in the main thread. 

Additionally I don't love how hard it is to actually write tests for it. 

The ideal solutions would be: 
- change the check so that it loads instantly, which would fix the issue without further changes. But... I don't know of a faster way to check on macOS, sadly. 
- update `isSandbox` to make it asynchronous... which would be a BIG refactor, affecting a large chunk of the codebase. It doesn't feel worth the trouble.

### DIFF
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -99,23 +99,26 @@ class CustomerInfoManager {
     func fetchAndCacheCustomerInfoIfStale(appUserID: String,
                                           isAppBackgrounded: Bool,
                                           completion: CustomerInfoCompletion?) {
-        let isCacheStale = self.withData {
-            $0.deviceCache.isCustomerInfoCacheStale(appUserID: appUserID, isAppBackgrounded: isAppBackgrounded)
-        }
-
-        guard !isCacheStale, let customerInfo = self.cachedCustomerInfo(appUserID: appUserID) else {
-            Logger.debug(isAppBackgrounded
-                            ? Strings.customerInfo.customerinfo_stale_updating_in_background
-                            : Strings.customerInfo.customerinfo_stale_updating_in_foreground)
-            self.fetchAndCacheCustomerInfo(appUserID: appUserID,
-                                           isAppBackgrounded: isAppBackgrounded,
-                                           completion: completion)
-            return
-        }
-
-        if let completion = completion {
-            self.operationDispatcher.dispatchOnMainActor {
-                completion(.success(customerInfo))
+        self.operationDispatcher.dispatchOnWorkerThread {
+            
+            let isCacheStale = self.withData {
+                $0.deviceCache.isCustomerInfoCacheStale(appUserID: appUserID, isAppBackgrounded: isAppBackgrounded)
+            }
+            
+            guard !isCacheStale, let customerInfo = self.cachedCustomerInfo(appUserID: appUserID) else {
+                Logger.debug(isAppBackgrounded
+                             ? Strings.customerInfo.customerinfo_stale_updating_in_background
+                             : Strings.customerInfo.customerinfo_stale_updating_in_foreground)
+                self.fetchAndCacheCustomerInfo(appUserID: appUserID,
+                                               isAppBackgrounded: isAppBackgrounded,
+                                               completion: completion)
+                return
+            }
+            
+            if let completion = completion {
+                self.operationDispatcher.dispatchOnMainActor {
+                    completion(.success(customerInfo))
+                }
             }
         }
     }
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -100,11 +100,11 @@ class CustomerInfoManager {
                                           isAppBackgrounded: Bool,
                                           completion: CustomerInfoCompletion?) {
         self.operationDispatcher.dispatchOnWorkerThread {
-            
+
             let isCacheStale = self.withData {
                 $0.deviceCache.isCustomerInfoCacheStale(appUserID: appUserID, isAppBackgrounded: isAppBackgrounded)
             }
-            
+
             guard !isCacheStale, let customerInfo = self.cachedCustomerInfo(appUserID: appUserID) else {
                 Logger.debug(isAppBackgrounded
                              ? Strings.customerInfo.customerinfo_stale_updating_in_background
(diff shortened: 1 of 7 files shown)


aboedo:
tagging @MarkVillacampa here too since he's worked on these checks quite a bit

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -99,23 +99,26 @@ class CustomerInfoManager {
     func fetchAndCacheCustomerInfoIfStale(appUserID: String,
                                           isAppBackgrounded: Bool,
                                           completion: CustomerInfoCompletion?) {
```
aboedo, line 101: Note: I'd recommend setting the diff to hide whitespace, since it makes it much clearer that all this change does is to add a single line to run the code in a worker thread

Sources/Misc/MapAppStoreDetector.swift
```diff
@@ -32,6 +32,8 @@ final class DefaultMacAppStoreDetector: MacAppStoreDetector {
     /// it checked the common name but was changed to an extension check to make it more
     /// future-proof.
     ///
+    /// - warning: this method may take a long time to run on macOS. It should not be called on the main thread.
```
aboedo, line 35: these warnings should ideally be fixed by the solutions described in the PR description, but I couldn't find a great way to make them work, sadly

Sources/Misc/SystemInfo.swift
```diff
@@ -144,6 +144,11 @@ class SystemInfo {
         self.responseVerificationMode = responseVerificationMode
         self.dangerousSettings = dangerousSettings ?? DangerousSettings()
         self.clock = clock
+        // eager-load the isSandbox value from a worker thread so that's immediately available
+        // this is useful on macOS where it may take long for the value to compute
+        self.operationDispatcher.dispatchOnWorkerThread {
+            _ = self._isSandbox
+        }
```
aboedo, line 151: this pre-loading makes the changes quite safe. After all, for the most part we'll only be loading once, as soon as SystemInfo is created, and from a worker thread. 
So other accesses will likely not hit this from a different thread. Ideally we'd modify the property to be asynchronous though so that we're completely covered.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1743,21 +1743,23 @@ private extension Purchases {
         // Note: it's important that we observe "will enter foreground" instead of
         // "did become active" so that we don't trigger cache updates in the middle
         // of purchases due to pop-ups stealing focus from the app.
-        self.updateAllCachesIfNeeded(isAppBackgrounded: false)
-        self.dispatchSyncSubscriberAttributes()
+        self.operationDispatcher.dispatchOnWorkerThread {
```
aboedo, line 1746: I really dislike how indirect this is, but it's not a bad idea to dispatch this work on the worker thread in any case

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
```diff
@@ -208,6 +208,20 @@ class CustomerInfoManagerTests: BaseCustomerInfoManagerTests {
...
+        mockDeviceCache.stubbedIsCustomerInfoCacheStale = false
+
+        var invokedCompletion = false
+        customerInfoManager.fetchAndCacheCustomerInfoIfStale(appUserID: Self.appUserID,
+                                                             isAppBackgrounded: false) { _ in
+            invokedCompletion = true
+        }
+
+        expect(invokedCompletion).toEventually(beTrue())
+        expect(self.mockOperationDispatcher.invokedDispatchOnWorkerThread) == true
```
aboedo, line 222: Not a fan of this since it's not a very accurate check, other things could be invoking this method. 
I don't know of a much better way to test when using this implementation, though.

aboedo:
Note: this was only an issue in production, because [of this check](https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Misc/SandboxEnvironmentDetector.swift#L53)

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
vegaro: throwing something crazy in here, and not sure it will work, but maybe modifying `dispatchOnWorkerThread` to this:

```
func dispatchOnWorkerThread(delay: Delay = .none, caller: String = #function, file: String = #file, line: Int = #line, block: @escaping @Sendable () -> Void) {
```

would let you check what has called `dispatchOnWorkerThread` in the mock?

jamesrb1 (approved):

Sources/Misc/SystemInfo.swift
tonidero: Hmm I imagine we could potentially be calculating the `isSandbox` property multiple times at the same time right? One here on initialization in a worker thread, then another in the main thread, if that property gets accessed from the main thread (Or even more if accessed from multiple threads). I think this might be ok, but I wanted to mention it.

I also see that we are accessing the `sandboxEnvironmentDetector.isSandbox` in multiple places, like the `DeviceCache` or the `StoreKit1Wrapper`. Do we need to handle those as well? (That can happen in different PRs though)

Sources/Purchasing/Purchases/Purchases.swift
tonidero: I'm slightly concerned about any unknown side-effects... but nothing comes to mind right now. I think this should be safe yeah.

aboedo: I believe it should be okay, all this code is already meant to be working on its own without user interaction (other than the user opening the SDK, that is)

Tests/UnitTests/Identity/CustomerInfoManagerTests.swift
aboedo: definitely an interesting approach, I'll take a look

aboedo:
@tonidero yeah I definitely still don't love this solution. Maybe it makes sense to limit its effects to macOS for the sake of minimizing impact... Or to give the asynchronous isSandbox approach another try, even though that one feels a lot more impactful in terms of the changes

aboedo:
closing in favor of https://github.com/RevenueCat/purchases-ios/pull/3875

PR #3876, opened 2024-05-07
Cleanup: Remove test storekit configuration files when importing through SPM. 

aboedo:
This PR removes the storekit configuration files we use for testing when you import the SDK through SPM. 

This cleans up apps, since those files were somewhat "leaked" from our tests, but not meant for consumption by a customer. 

| Before | After |
| :-: | :-: |
| Screenshot 2024-05-07 at 4 23 39 PM | Screenshot 2024-05-07 at 4 34 20 PM |

## Implementation
This PR does two things: 
- Removes all test targets from SPM. We don't really use SPM for testing directly, so they didn't really add value there. 
- Adds a new `Package.swift` file at the root of the `Tests` folder, which in turn tells SPM to treat that folder as a separate package, and thus the files in there don't get picked up automatically when importing RevenueCat.

### DIFF
--- a/Tests/Package.swift
+++ b/Tests/Package.swift
@@ -1,0 +1,11 @@
+// swift-tools-version:5.5
+//
+//  Package.swift
+//
+//
+//  Created by Andrés Boedo on 5/7/24.
+//
+
+import PackageDescription
+
+let package = Package()
--- a/Tests/Package.swift
+++ b/Tests/Package.swift
@@ -5,6 +5,10 @@
 //
 //  Created by Andrés Boedo on 5/7/24.
 //
+// This empty file prevents Xcode from picking up the files in the Tests library
+// when importing through SPM. This is helpful so that users of the SDK
+// don't inadvertently get StoreKit Configuration files added to the project, which
+// can lead to confusion.
 
 import PackageDescription
 
--- a/Package.swift
+++ b/Package.swift
@@ -9,11 +9,7 @@ import class Foundation.ProcessInfo
 let environmentVariables = ProcessInfo.processInfo.environment
 let shouldIncludeDocCPlugin = environmentVariables["INCLUDE_DOCC_PLUGIN"] == "true"
 
-var dependencies: [Package.Dependency] = [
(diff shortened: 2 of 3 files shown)


aboedo:
apparently we did run tests from SPM for UI tests, but I'm not sure why yet, I'll take a look. Feels odd to have only one suite run from SPM.

Tests/Package.swift
```diff
@@ -0,0 +1,15 @@
...
+//  Created by Andrés Boedo on 5/7/24.
+//
+// This empty file prevents Xcode from picking up the files in the Tests library
+// when importing through SPM. This is helpful so that users of the SDK
+// don't inadvertently get StoreKit Configuration files added to the project, which
+// can lead to confusion.
+
+import PackageDescription
+
+let package = Package()
```
vegaro, line 15: just wondering but what is this file for?

vegaro (approved):

Tests/Package.swift
vegaro: oh just saw in the description:

> Adds a new Package.swift file at the root of the Tests folder, which in turn tells SPM to treat that folder as a separate package, and thus the files in there don't get picked up automatically when importing RevenueCat.

aboedo: haha yeah, exactly. Kinda reminds me of the `Dummy.swift` file we have in some of our repos (although it's for a different purpose of course)

aboedo:
started looking into fixing the SPM test, and I fell into quite the rabbit hole:
- the tests are set up to use the SPM package
- RevenueCatUI is set up solely as an SPM package
- Making it an Xcode target instead (with a corresponding test target) fixes that
- ... but then when doing that, I get a _lot_ of warnings and issues for sendable and ABI stability. Not sure why we're not getting those when compiling as an SPM package, whether that's that those aren't checked with SPM or that there's a config setting in Xcode that looks different

aboedo:
I'm gonna try a simpler approach with only the package.swift file, we'll see how that goes

aboedo:
closing in favor of #3878

PR #3878, opened 2024-05-08
Cleanup: Remove test storekit configuration files when importing through SPM

aboedo:
Alternative approach to #3876, where in this one we _only_ add the Package.swift file. 
That was the most important part in any case. 

I would still love to clean up the project structure, but it's a much bigger change.

### DIFF
--- a/Tests/Package.swift
+++ b/Tests/Package.swift
@@ -1,0 +1,15 @@
+// swift-tools-version:5.5
+//
+//  Package.swift
+//
+//
+//  Created by Andrés Boedo on 5/7/24.
+//
+// This empty file prevents Xcode from picking up the files in the Tests library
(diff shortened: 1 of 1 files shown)


aboedo:
can confirm that it still works: 

image


Now we just need to see whether tests still run correctly

aboedo:
This has been bugging me for so long it's kind of crazy that the solution was this simple

jamesrb1 (approved):
Works like a charm.

tonidero (approved):
🙌

vegaro:
I've noticed that after this change, the `Tests` folder stopped appearing in xcode when opening the SPM project, which is the only way I know to work on `RevenueCatUI` tests. The xcodeproj doesn't have the RevenueCatUI test files (nor the source files)

PR #3879, opened 2024-05-08
fix mac os sandbox check slowness

aboedo:
Alternative approach to #3875, also meant to address https://github.com/RevenueCat/purchases-ios/issues/3871

We're currently looking at a field called Environment in the local receipt to check whether the app is running in sandbox on macOS. 

However, since it's an undocumented field, we can't trust that it will always be there, so if the value is production then we check the bundle's signature, which can be slow to run. 

This PR changes that behavior so that we only perform that check if the value for that variable isn't available, which would cause a receipt parsing failure since it's marked as non-optional. 

Still need to write tests for this, but it should be significantly easier than for the other PR.

### DIFF
--- a/Sources/Misc/SandboxEnvironmentDetector.swift
+++ b/Sources/Misc/SandboxEnvironmentDetector.swift
@@ -50,7 +50,14 @@ final class BundleSandboxEnvironmentDetector: SandboxEnvironmentDetector {
         }
 
         #if os(macOS) || targetEnvironment(macCatalyst)
-            return !self.isProductionReceipt || !self.isMacAppStore
+        // this relies on an undocumented field in the receipt that provides the Environment.
+        // if it's not present, we go to a secondary check.
+        if let isProductionReceipt = self.isProductionReceipt {
+            return !isProductionReceipt
+        } else {
+            return !self.isMacAppStore
+        }
+
         #else
             return path.contains("sandboxReceipt")
         #endif
@@ -73,12 +80,12 @@ extension BundleSandboxEnvironmentDetector: Sendable {}
 
 private extension BundleSandboxEnvironmentDetector {
 
-    var isProductionReceipt: Bool {
+    var isProductionReceipt: Bool? {
         do {
             return try self.receiptFetcher.fetchAndParseLocalReceipt().environment == .production
         } catch {
             Logger.error(Strings.receipt.parse_receipt_locally_error(error: error))
-            return false
+            return nil
         }
     }
 
--- a/Sources/Misc/SandboxEnvironmentDetector.swift
+++ b/Sources/Misc/SandboxEnvironmentDetector.swift
@@ -82,7 +82,9 @@ private extension BundleSandboxEnvironmentDetector {
 
     var isProductionReceipt: Bool? {
         do {
-            return try self.receiptFetcher.fetchAndParseLocalReceipt().environment == .production
+            let receiptEnvironment = try self.receiptFetcher.fetchAndParseLocalReceipt().environment
+            guard receiptEnvironment != .unknown else { return nil } // don't make assumptions if we're not sure
+            return receiptEnvironment == .production
         } catch {
             Logger.error(Strings.receipt.parse_receipt_locally_error(error: error))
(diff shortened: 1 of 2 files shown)


Sources/Misc/SandboxEnvironmentDetector.swift
```diff
@@ -73,12 +80,12 @@ extension BundleSandboxEnvironmentDetector: Sendable {}
 
 private extension BundleSandboxEnvironmentDetector {
 
-    var isProductionReceipt: Bool {
+    var isProductionReceipt: Bool? {
         do {
             return try self.receiptFetcher.fetchAndParseLocalReceipt().environment == .production
```
tonidero, line 85: Should we also check if the `environment` is `.unknown`? Seems that the catch will happen with any parsing error, not specifically about the environment so I wonder if we will be doing extra operations in those cases... Having said that, parsing errors should be fairly uncommon, so I think it should be ok?

MarkVillacampa (approved):

Sources/Misc/SandboxEnvironmentDetector.swift
aboedo: Only 3 months late to respond to the comment 😅 
Good call, I'll add an extra check for that

aboedo: can't imagine it would happen often that we can correctly parse the receipt and it's production but it doesn't have the environment variable, but I can't discard it either since it's an undocumented field

Tests/UnitTests/Misc/SandboxEnvironmentDetectorTests.swift
```diff
@@ -45,7 +45,7 @@ class SandboxEnvironmentDetectorTests: TestCase {
 // `macOS` sandbox detection does not rely on receipt path
 class SandboxEnvironmentDetectorTests: TestCase {
 
-    func testIsNotSandboxIfReceiptIsProductionAndMAS() throws {
+    func testIsNotSandboxIfReceiptIsProduction() throws {
```
aboedo, line 48: we no longer look at MAS if we know what the receipt environment is

Tests/UnitTests/Misc/SandboxEnvironmentDetectorTests.swift
```diff
@@ -54,41 +54,85 @@ class SandboxEnvironmentDetectorTests: TestCase {
...
         ) == true
     }
 
-    func testIsSandboxIfReceiptIsNotProductionAndNotMAS() throws {
-        expect(
-            SystemInfo.with(
-                macAppStore: false,
-                receiptEnvironment: .sandbox
-            ).isSandbox
-        ) == true
```
aboedo, line 72: this test and the next one are technically still there, but now in a more complete test that also verifies that we indeed use MAS when needed and skip it when not needed

Tests/UnitTests/Misc/SandboxEnvironmentDetectorTests.swift
```diff
@@ -151,12 +196,17 @@ private final class MockLocalReceiptFetcher: LocalReceiptFetcherType {
...
 
     init(isMacAppStore: Bool) {
-        self.isMacAppStore = isMacAppStore
+        self.isMacAppStoreValue = isMacAppStore
     }
 
+    var isMacAppStore: Bool {
+        isMacAppStoreCalled = true
+        return isMacAppStoreValue
+    }
```
aboedo, line 211: changes here are to make it possible to check whether it was called

aboedo:
@tonidero @MarkVillacampa ready for another pass, hopefully it takes me less than 3 months this time to iterate on PR comments this time

aboedo:
Tested on device (by forcing the environment value in the receipt) and was able to verify that I got the warning before the changes and did not get it after. 

Still could be optimized further by caching the value (it won't change in a single app run) but then I started fighting Sendable and figured to live to fight another day

MarkVillacampa:
Looking goor!

tonidero (approved):
LGTM!

PR #3883, opened 2024-05-10
Improved Error Handling

jamesrb1:
Adds missing user-facing error alerts, makes the error messages more readable and uses `Error` rather than `NSError` as the type to use when passing errors for presentation.
 
resolves PWL-458


aboedo (approved):

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/ManagePaywallButton.swift
```diff
@@ -12,6 +12,9 @@ import WatchKit
 
 struct ManagePaywallButton: View {
 
+    @State
+    var error: Error?
```
NachoSoto, line 16: State properties should be private.

PR #3885, opened 2024-05-10
Paywalls Tester: Enable Example Paywalls in Release Mode

jamesrb1:
The paywalls examples weren't showing up when compiled in release mode. Enabling it meant making a bunch of more internal things public via the preprocessing script. `TestData` is no longer `#ifdef DEBUG`ed out, but most of its data is still `#ifdef DEBUG`, and when it's compiled with an `internal` modifier and thus nothing uses it, I'd expect what remains to be stripped away during linking.


tonidero (approved):
LGTM!

PR #3886, opened 2024-05-11
Hide decorative Paywall images from accessibility.

shiftingsand:
### Motivation

The paywall template I'm using (five) displays a decorative image but the image isn't marked as decorative to accessibility. In my particular test case that meant that VoiceOver could access it, but there was no audible description of what the image actually was. 

Normally I would change Image() to Image(decorative: ), but since that wasn't possible I used .accessibilityHidden(true).

### Description

Added .accessibilityHidden(true) to the two images in RemoteImage. I initially was hesitant to change RemoteImage itself, but went ahead and did so since it seems to be only used by the Paywall templates.


tonidero (approved):
Looks great @shiftingsand! We will get this merged. Thanks for the contribution!

shiftingsand:
Thank you so much.  Happy to help!

PR #3902, opened 2024-05-15
Retry backend integration tests

joshdholtz:
This is "ready for review" but it requires https://github.com/RevenueCat/purchases-ios/pull/3833 to be merged into it first to fully pass

### Motivation

Prevent the need to fully rerun all backend tests job when only a few flaky tests fail

### Description

1. Runs backend integration tests
    - Parses junit.xml for failed tests
    - Writes them to a file
2. Runs new step in CircleCI
    - Looks at file for tests to re-run and only runs those tests
    - Merges new junit.xml into original junit.xml

The retry happens 4 times right now and will fail (raise an error) if there are any left over tests that never ended up passing


fastlane/Fastfile
```diff
@@ -605,16 +605,153 @@ platform :ios do
...
+        scheme: "BackendIntegrationTests",
+        derived_data_path: "scan_derived_data",
+        output_types: 'junit',
+        output_style: 'raw',
+        result_bundle: true,
+        testplan: options[:test_plan],
+        configuration: 'Debug',
+        output_directory: "fastlane/test_output/xctest/ios",
+
+        fail_build: false # TODO: Need this for retry but maybe find a way to fail if we don't want this
```
tonidero, line 620: Hmm maybe make it a parameter like in the retry lane but default it to false?

fastlane/Fastfile
```diff
@@ -605,16 +605,153 @@ platform :ios do
...
+
+      moved_report_path = File.join(artifact_path, "report.junit.retry.#{retry_attempt}")
+      FileUtils.mv(report_path, moved_report_path)
+
+      save_failed_tests(path: moved_report_path)
+      merge_and_replace_junit(retry_path: moved_report_path)
+    end
+  end
+
+  lane :merge_and_replace_junit do |options|
```
tonidero, line 692: Hmm does ios use junit?

fastlane/Fastfile
```diff
@@ -605,16 +605,153 @@ platform :ios do
...
+    retry_report = options[:retry_path]
+
+    original_report = File.absolute_path('./test_output/xctest/ios/report.junit')
+
+    merged_report = merge_reports(original_report, retry_report)
+    save_report(merged_report, original_report)
+
+    FileUtils.rm_rf('./test_output/xctest/ios/retry')
+
+    puts "Merged report saved to #{original_report}"
```
tonidero, line 702: Should this use fastlane log tools like `UI.message`?

tonidero (approved):
Some small comments but LGTM!

fastlane/Fastfile
```diff
@@ -605,16 +605,153 @@ platform :ios do
...
+
+      scan(
+        scheme: "BackendIntegrationTests",
+        derived_data_path: "scan_derived_data",
+        output_types: 'junit',
+        output_style: 'raw',
+        result_bundle: true,
+        configuration: 'Debug',
+        output_directory: File.join('fastlane', report_dir),
+
```
fire-at-will, line 679: Nit: remove blank line

```suggestion
```

fire-at-will (commented):
Everything here looks good to me besides @tonidero 's comments!

fastlane/Fastfile
fire-at-will: Somehow I think we're using junit to keep track of the passed/failed reports

joshdholtz: Yup yup! Well... _fastlane_ takes the `xcresult` and converts it to a `junit` because that is what CircleCI likes

fastlane/Fastfile
joshdholtz: Ooooops! Yes yes yes 👍

PR #3906, opened 2024-05-16
RemoteImage Low Res Image support

jamesrb1:
`RemoteImage` now uses two images: a high res image (as before), and a new low res image which will be used if it is available before the high res image has been loaded.

To enable this, it now uses two `ImageLoader` instances, one for to fetch the low res image, a second for high res image.

It assumes that the low res image url is the same as the high res image url, with a suffix appended (default `_low_res`) to the file name, eg `/images/myImage.jpg` is assumed to have a low res equivalent at `/images/myImage_low_res.jpg`.

It requests the high and low res images concurrently, and displays an error when both requests fail. A low res image will never replace a high res image even if it is received last.

Here is a video of it loading a "low res" image, following by high res image. For this video the low-res image was intentionally set to a completely different image to make the change easy to see. If you move the slider manually you can control the speed.

https://github.com/RevenueCat/purchases-ios/assets/109382862/97521803-5c12-43d0-92a8-596b23cdcff6


aboedo (approved):
Love this! cc @joshdholtz and @RevenueCat/coresdk

aboedo:
why is the linter so angry? it's pointing at stuff that isn't changed in this PR

RevenueCatUI/Views/RemoteImage.swift
```diff
@@ -19,42 +19,80 @@ struct RemoteImage: View {
     let url: URL
     let aspectRatio: CGFloat?
     let maxWidth: CGFloat?
+    let lowResSuffix: String?
```
tonidero, line 22: I've been wondering... Would it make more sense to have the backend provide the name of the low res image on a different field if available, instead of relying in a convention like this? So basically RemoteImage would have 2 URLs, one for the high res and one for the low res version, and we need to pass both (if available) from the backend data. Wdyt?

RevenueCatUI/Views/RemoteImage.swift
```diff
@@ -19,42 +19,80 @@ struct RemoteImage: View {
...
-                        .accessibilityHidden(true)
-
+            if case let .success(image) = highResLoader.result {
+                displayImage(image)
+            } else if case let .success(image) = lowResLoader.result {
+                displayImage(image)
+            } else if case let .failure(highResError) = highResLoader.result {
+                if !fetchLowRes {
+                    emptyView(error: highResError)
+                } else if case .failure = lowResLoader.result {
```
tonidero, line 52: I think we're ignoring the low res error. Should we log something in this case?

tonidero (commented):
I suggested a different architecture to rely on the backend giving us the proper name instead of relying on a convention, but I think the current solution can also work! Happy to go with this for now, we can always change it later if needed.

RevenueCatUI/Views/RemoteImage.swift
jamesrb1: Ah yes, I think it would!

RevenueCatUI/Views/RemoteImage.swift
jamesrb1: Good point, but ImageLoader logs when it hits an error (below), so I think we are ok here.

```
    @Published
    private(set) var result: Value? {
        didSet {
            if let result {
                Logger.verbose(Strings.image_result(result.map { _ in () }))
            }
        }
    }
    ```

Comment (tonidero) [reply RevenueCatUI/Views/RemoteImage.swift:52]: Ahh then this is fine, thanks!

File: RevenueCatUI/Views/RemoteImage.swift (hunk 1 above)
Comment (jamesrb1) [reply RevenueCatUI/Views/RemoteImage.swift]: Implemented here, can you take a look? https://github.com/RevenueCat/purchases-ios/pull/3906/commits/16c7473a42a94195be279ab85a3c13e2df9f7b49

**Review by aboedo — APPROVED** [review]
looks good, although FWIW I think the prefix was fine too, I don't imagine it would have really backed us into a corner or created any issues

**Review by tonidero — APPROVED** [review]
LGTM!

And yeah, I think the suffix approach was fine too, but this is more versatile. Also, one small extra is that we know beforehand (without trying to fetch it) if there is a low res version of the image.

File: RevenueCatUI/Views/RemoteImage.swift [hunk 3]
```diff
@@ -17,44 +17,76 @@ import SwiftUI
 struct RemoteImage: View {
 
     let url: URL
+    let lowResUrl: URL?
     let aspectRatio: CGFloat?
     let maxWidth: CGFloat?
 
+    var fetchLowRes: Bool {
```
NachoSoto, line 24: This could be private

RevenueCatUI/Views/RemoteImage.swift
```diff
@@ -17,44 +17,76 @@ import SwiftUI
...
         .transition(Self.transition)
         .task(id: self.url) { // This cancels the previous task when the URL changes.
-            await self.loader.load(url: self.url)
+            await loadImages()
+        }
+    }
+
+    private func displayImage(_ image: Image) -> some View {
+        if let aspectRatio {
+            return AnyView(
```
NachoSoto, line 67: These don't need to be `AnyView` if you make this property a `@ViewBuilder`.

RevenueCatUI/Views/RemoteImage.swift
```diff
@@ -17,44 +17,76 @@ import SwiftUI
...
                 }
-
-            case let .failure(error):
-                self.emptyView(error: error)
+            } else {
+                emptyView(error: nil)
             }
         }
         .transition(Self.transition)
         .task(id: self.url) { // This cancels the previous task when the URL changes.
```
NachoSoto, line 60: This should probably be 2 separate tasks. This is unnecessarily cancelling loading `url` if the other UEL changes.

RevenueCatUI/Views/RemoteImage.swift
```diff
@@ -17,44 +17,76 @@ import SwiftUI
...
+        }
+    }
+
+    private func displayImage(_ image: Image) -> some View {
+        if let aspectRatio {
+            return AnyView(
+                image
+                    .fitToAspect(aspectRatio, contentMode: .fill)
+                    .frame(maxWidth: self.maxWidth)
+                    .accessibilityHidden(true)
```
NachoSoto, line 71: This here and help could be defined only once at the top level of `RemoteImage`.

RevenueCatUI/Views/RemoteImage.swift
NachoSoto: Actually this is defining an (unnecessary IMO) duplicate source of truth.

RevenueCatUI/Views/RemoteImage.swift
aboedo: Kinda feel like that might be more of a feature than a bug? Like, if one of the image URLs changes, it's more than likely than both would... And if they didn't, it might not be a bad idea to still reload both just in case the URL remains the same but not the image. 

PS: Hi Nacho 👋

RevenueCatUI/Views/RemoteImage.swift
aboedo: good point

RevenueCatUI/Views/RemoteImage.swift
jamesrb1: Hi Nacho! Thanks for reviewing the PR! I'm sorry we never got to work together!

As implemented, I didn't consider this a duplicate source of truth because it's a computed property, and this property makes other code more readable because the semantic meaning is clearly expressed rather than being derived inline whenever needed. Thoughts?

It could be private, and so could the four properties that precede it. I don't really have a strong opinion here but would be interested to hear any further thoughts you might have.

RevenueCatUI/Views/RemoteImage.swift
jamesrb1: Yes! Thank you!

jamesrb1:
I've opened a new draft PR here: https://github.com/RevenueCat/purchases-ios/pull/3937

RevenueCatUI/Views/RemoteImage.swift
jamesrb1: I kind of like the idea...

```swift
        .task(id: self.url) {
            await loadHighResImage()
        }
        .task(id: self.lowResUrl) {
            await loadLowResImage()
        }
```
calling
```swift
    private func loadHighResImage() async {
        await highResLoader.load(url: url)
    }

    private func loadLowResImage() async {
        guard let lowRes = lowResUrl else { return }
        await lowResLoader.load(url: lowRes)
    }
```
which seems more elegant than the logic currently inside `loadImages()`:

```swift
    private func loadImages() async {
        if fetchLowRes, let lowResLoc = lowResUrl {
            async let lowResLoad: Void = lowResLoader.load(url: lowResLoc)
            async let highResLoad: Void = highResLoader.load(url: url)
            _ = await (lowResLoad, highResLoad)

        } else {
            await highResLoader.load(url: url)
        }
    }
```

BUT keeping `.url` as the sole identifier I feel makes some sense too.

NachoSoto: Yes that implementation makes sense 👍🏻

PR #3909, opened 2024-05-17
[WIP] Fix swiftlint install on CI

joshdholtz:
### Motivation

_Coming soon_

### Description

_Coming soon_


aboedo (approved):

.circleci/config.yml
```diff
@@ -136,12 +139,22 @@ commands:
...
+          dependency_name: 'mint'
+      - when:
+          condition:
+            not:
+              equal: ["", << parameters.swiftlint_version >>]
+          steps:
+            - run:
+                name: Install swiftlint@<< parameters.swiftlint_version >> using Mint 
+                command: |
+                  mint install realm/swiftlint@<< parameters.swiftlint_version >>
```
tonidero, line 151: Hmm I'm wondering, should we remove installing swiftlint from `install-dependencies` and just install it for the linter job? I think that we don't really need to install it for all jobs? That way we can just keep installing the latest version only, I don't see that much value in running swiftlint multiple times or anything.

Tests/APITesters/CustomEntitlementComputationSwiftAPITester/SwiftAPITester/PurchasesAPI.swift
```diff
@@ -101,7 +101,7 @@ private func checkTypealiases(
                                                   userCancelled: userCancelled)
 
     // swiftlint:disable:next line_length
-    let purchaseCompletedBlock: PurchaseCompletedBlock = { (_: StoreTransaction?, _: CustomerInfo?, _: Error?, _: Bool) -> Void in }
+    let purchaseCompletedBlock: PurchaseCompletedBlock = { (_: StoreTransaction?, _: CustomerInfo?, _: Error?, _: Bool) in }
```
aboedo, line 104: I kinda disagree with swiftlint on this one, the `-> Void` makes it much clearer that you're looking at a block with void return type IMO, at first glance this seems like a tuple otherwise

.circleci/config.yml
aboedo: We _shouldn't_ need it for all jobs but it's part of a build phase or something IIRC, we'd need to remove it from there too. Otherwise a bunch of jobs will fail.

tonidero: Ohhh right... I guess we could skip that build phase if swiftlint is not installed, but yeah, I think this makes sense for now then 👍

tonidero (approved):

PR #3911, opened 2024-05-17
Jc/sample app swiftui

nyeu:
### Motivation
Apps were not showcasing the latest features and were outdated. This not only helps our customers to easily understand how RevenueCat features work but it's also helpful for when the DSE team has to debug something related to paywalls.

### Description

I've run the app both through a simulator and a real device using my own app projects, subscribing and seeing the expected result.


Examples/MagicWeatherSwiftUI/Magic Weather SwiftUI.xcodeproj/project.pbxproj
```diff
@@ -439,6 +436,30 @@
 			defaultConfigurationName = Release;
 		};
 /* End XCConfigurationList section */
+
+/* Begin XCRemoteSwiftPackageReference section */
+		4E3BE8642BF77F420045B504 /* XCRemoteSwiftPackageReference "purchases-ios" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/RevenueCat/purchases-ios";
```
tonidero, line 443: I think we want to change to use `https://github.com/RevenueCat/purchases-ios-spm` instead, since it's much faster. cc @jamesrb1

tonidero (commented):
A first view looks good! It would be great if you can also post some pictures/short video with the changes, but looks good! Will hold on approving until it's ready

Examples/MagicWeatherSwiftUI/Magic Weather SwiftUI.xcodeproj/project.pbxproj
nyeu: updated in the latest commit

PR #3917, opened 2024-05-24
Paywalls manually handle purchases, finishTransactions/ObserverMode -> PurchasesAreCompletedBy

joshdholtz:
### Motivation

- Allow developers who have their own purchase logic with StoreKit to use our paywalls.
- Adds a new API `purchasesAreCompletedBy: PurchasesAreCompletedBy` to `Purchases`, which replaces `finishTransactions`, and `observerMode` which have been marked as deprecated (public APIs only).
- Internal references to `finishTransactions` and `observerMode` will be done as a separate PR.

#### Todo
- [x] Update any doc on new method to make sure its SUPER CLEAR on what its doing and that it will only work if `finishTransactions` (trying to not call it observer mode) is disabled
- [x] Maybe add some debug logs so developer knows which path paywalls purchase buttons are taking if finish transaction is disabled
- [x] Fix broken unit/integration tests
- [x] Add new tests
- [x] Add new API tests
- [x] Also add a similar method for `handleRestore()`
 
### Description

Adds new `. handlePurchaseAndRestore` modifier that can be used to complete purchases initiated via a paywall:

```swift
.handlePurchaseAndRestore(
    performPurchase: { storeProduct, purchaseResultReporter in
        // make purchase for `storeProduct`
        
        // report result to RevenueCat
        purchaseResultReporter.reportResult(userCancelled: false, error: nil)
    }, performRestore: { restoreResultReporter in
        // restore purchases

        // report result to RevenueCat
        restoreResultReporter.reportResult(success: true, error: nil)
    })
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -338,7 +338,6 @@ PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
   arm64-darwin-23
-  ruby
   x86_64-darwin-22
   x86_64-linux
 
--- a/RevenueCatUI/Data/Strings.swift
+++ b/RevenueCatUI/Data/Strings.swift
@@ -105,8 +105,9 @@ extension Strings: CustomStringConvertible {
             return "Setting restored customer info"
 
         case .executing_external_purchase_logic:
-            return "Will execute custom StoreKit purchase logic provided by the SDK adopter. " +
-            "No StoreKit purchasing logic will be performed by RevenueCat."
+            return "Will execute custom StoreKit purchase logic provided by yourapp. " +
+            "No StoreKit purchasing logic will be performed by RevenueCat. " +
+            "You must use `.handlePurchase` on your `PaywallView`"
 
         case .executing_purchase_logic:
             return "Will execute purchase logic provided by RevenueCat."
@@ -115,8 +116,9 @@ extension Strings: CustomStringConvertible {
             return "Will execute restore purchases logic provided by RevenueCat."
 
         case .executing_external_restore_logic:
-            return "Will execute custom StoreKit restore purchases logic provided by the SDK adopter. " +
-            "No StoreKit restore purchases logic will be performed by RevenueCat."
+            return "Will execute custom StoreKit restore purchases logic provided by your app. " +
+            "No StoreKit restore purchases logic will be performed by RevenueCat. " +
+            "You must use `.handleRestorePurchases` on your `PaywallView`"
         }
     }
 
--- a/RevenueCatUI/Data/Strings.swift
+++ b/RevenueCatUI/Data/Strings.swift
@@ -107,7 +107,7 @@ extension Strings: CustomStringConvertible {
         case .executing_external_purchase_logic:
             return "Will execute custom StoreKit purchase logic provided by your app. " +
             "No StoreKit purchasing logic will be performed by RevenueCat. " +
-            "You must use `.handlePurchase` on your `PaywallView`."
+            "You must use `.handlePurchaseAndRestore` on your `PaywallView`."
 
         case .executing_purchase_logic:
             return "Will execute purchase logic provided by RevenueCat."
@@ -118,7 +118,7 @@ extension Strings: CustomStringConvertible {
         case .executing_external_restore_logic:
             return "Will execute custom StoreKit restore purchases logic provided by your app. " +
             "No StoreKit restore purchases logic will be performed by RevenueCat. " +
-            "You must use `.handleRestorePurchases` on your `PaywallView`."
+            "You must use `.handlePurchaseAndRestore` on your `PaywallView`."
         }
     }
 
--- a/RevenueCatUI/Data/Strings.swift
+++ b/RevenueCatUI/Data/Strings.swift
@@ -105,7 +105,7 @@ extension Strings: CustomStringConvertible {
             return "Setting restored customer info"
 
         case .executing_external_purchase_logic:
-            return "Will execute custom StoreKit purchase logic provided by yourapp. " +
+            return "Will execute custom StoreKit purchase logic provided by your app. " +
             "No StoreKit purchasing logic will be performed by RevenueCat. " +
             "You must use `.handlePurchase` on your `PaywallView`."
 
--- a/RevenueCatUI/Data/Strings.swift
+++ b/RevenueCatUI/Data/Strings.swift
@@ -40,7 +40,11 @@ enum Strings {
     case restore_purchases_with_empty_result
     case setting_restored_customer_info
 
+    case executing_purchase_logic
     case executing_external_purchase_logic
+    case executing_restore_logic
+    case executing_external_restore_logic
+
 
 }
 
@@ -104,6 +108,16 @@ extension Strings: CustomStringConvertible {
         case .executing_external_purchase_logic:
             return "Will execute custom StoreKit purchase logic provided by the SDK adopter. " +
             "No StoreKit purchasing logic will be performed by RevenueCat."
+
+        case .executing_purchase_logic:
+            return "Will execute purchase logic provided by RevenueCat."
+
+        case .executing_restore_logic:
+            return "Will execute restore purchases logic provided by RevenueCat."
+
+        case .executing_external_restore_logic:
+            return "Will execute custom StoreKit restore purchases logic provided by the SDK adopter. " +
+            "No StoreKit restore purchases logic will be performed by RevenueCat."
         }
     }
 
--- a/RevenueCatUI/Data/Strings.swift
+++ b/RevenueCatUI/Data/Strings.swift
@@ -107,7 +107,7 @@ extension Strings: CustomStringConvertible {
         case .executing_external_purchase_logic:
             return "Will execute custom StoreKit purchase logic provided by yourapp. " +
             "No StoreKit purchasing logic will be performed by RevenueCat. " +
-            "You must use `.handlePurchase` on your `PaywallView`"
+            "You must use `.handlePurchase` on your `PaywallView`."
 
         case .executing_purchase_logic:
             return "Will execute purchase logic provided by RevenueCat."
@@ -118,7 +118,7 @@ extension Strings: CustomStringConvertible {
         case .executing_external_restore_logic:
             return "Will execute custom StoreKit restore purchases logic provided by your app. " +
             "No StoreKit restore purchases logic will be performed by RevenueCat. " +
-            "You must use `.handleRestorePurchases` on your `PaywallView`"
+            "You must use `.handleRestorePurchases` on your `PaywallView`."
         }
     }
 
--- a/RevenueCatUI/Data/Strings.swift
+++ b/RevenueCatUI/Data/Strings.swift
@@ -45,7 +45,6 @@ enum Strings {
     case executing_restore_logic
     case executing_external_restore_logic
 
-
 }
 
 @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.0, *)
--- a/RevenueCatUI/Purchasing/PaywallPurchasesType.swift
+++ b/RevenueCatUI/Purchasing/PaywallPurchasesType.swift
@@ -17,7 +17,7 @@ import RevenueCat
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 protocol PaywallPurchasesType: Sendable {
 
-    var purchasesAreCompletedBy: Bool { get set }
+    var purchasesAreCompletedBy: PurchasesAreCompletedBy { get set }
 
     @Sendable
     func purchase(package: Package) async throws -> PurchaseResultData
--- a/RevenueCatUI/Purchasing/PaywallPurchasesType.swift
+++ b/RevenueCatUI/Purchasing/PaywallPurchasesType.swift
@@ -17,7 +17,7 @@ import RevenueCat
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 protocol PaywallPurchasesType: Sendable {
 
-    var finishTransactions: Bool { get set }
+    var purchasesAreCompletedBy: Bool { get set }
 
     @Sendable
     func purchase(package: Package) async throws -> PurchaseResultData
--- a/RevenueCatUI/Purchasing/PurchaseHandler.swift
+++ b/RevenueCatUI/Purchasing/PurchaseHandler.swift
@@ -115,7 +115,7 @@ extension PurchaseHandler {
 
     @MainActor
     func purchase(package: Package) async throws -> PurchaseResultData {
-        if self.purchases.purchasesAreCompletedBy {
+        if self.purchases.purchasesAreCompletedBy == .revenueCat {
             return try await performPurchase(package: package)
         } else {
             return try await performExternalPurchaseLogic(package: package)
@@ -189,7 +189,7 @@ extension PurchaseHandler {
     // MARK: - Restore
 
     func restorePurchases() async throws -> (info: CustomerInfo, success: Bool) {
-        if self.purchases.purchasesAreCompletedBy {
+        if self.purchases.purchasesAreCompletedBy == .revenueCat {
             return try await performRestorePurchases()
         } else {
             return try await performExternalRestoreLogic()
@@ -352,8 +352,8 @@ private extension PurchaseHandler {
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 private final class NotConfiguredPurchases: PaywallPurchasesType {
 
-    var purchasesAreCompletedBy: Bool {
-        get { return false }
+    var purchasesAreCompletedBy: PurchasesAreCompletedBy {
+        get { return .myApp }
         set { _ = newValue }
     }
 
--- a/RevenueCatUI/Purchasing/PurchaseHandler.swift
+++ b/RevenueCatUI/Purchasing/PurchaseHandler.swift
@@ -34,7 +34,6 @@ class HandleRestoreCallbackContainer: Equatable {
 
 }
 
-
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 // @PublicForExternalTesting
 final class PurchaseHandler: ObservableObject {
@@ -168,8 +167,6 @@ extension PurchaseHandler {
 
         return PurchaseResultData(nil, try await self.purchases.customerInfo(), false)
     }
-
-
 
     @MainActor
     func completeExternalHandlePurchase(_ userCancelled: Bool, _ error: Error?) {
--- a/RevenueCatUI/Purchasing/PurchaseHandler.swift
+++ b/RevenueCatUI/Purchasing/PurchaseHandler.swift
@@ -61,6 +61,7 @@ public class RestoreResultReporter: Equatable {
         reportRestoreResultCallback(success, error)
     }
 
+    /// Returns true if objects are the same object (same memory address); false otherwise.
     public static func == (lhs: RestoreResultReporter, rhs: RestoreResultReporter) -> Bool {
         return lhs === rhs
     }
--- a/RevenueCatUI/Purchasing/PurchaseHandler.swift
+++ b/RevenueCatUI/Purchasing/PurchaseHandler.swift
@@ -135,8 +135,8 @@ final class PurchaseHandler: ObservableObject {
 
     /// Returns a new instance of `PurchaseHandler` using `Purchases.shared` if `Purchases`
     /// has been configured, and using a PurchaseHandler that cannot be used for purchases otherwise.
-    static func `default`() -> Self {
     // @PublicForExternalTesting
+    static func `default`() -> Self {
         return Purchases.isConfigured ? .init() : Self.notConfigured()
     }
 
--- a/RevenueCatUI/Purchasing/PurchaseHandler.swift
+++ b/RevenueCatUI/Purchasing/PurchaseHandler.swift
@@ -25,8 +25,8 @@ class HandleRestoreCallbackContainer: Equatable {
     let handleRestoreCallback: (_ userCancelled: Bool, _ error: Error?) -> Void
 
     init(callback: @escaping (Bool, Error?) -> Void) {
-            self.handleRestoreCallback = callback
-        }
+        self.handleRestoreCallback = callback
+    }
 
     static func == (lhs: HandleRestoreCallbackContainer, rhs: HandleRestoreCallbackContainer) -> Bool {
         return lhs === rhs
@@ -112,12 +112,14 @@ final class PurchaseHandler: ObservableObject {
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 extension PurchaseHandler {
 
+    // MARK: - Purchase
+
     @MainActor
     func purchase(package: Package) async throws -> PurchaseResultData {
         if self.purchases.finishTransactions {
             return try await performPurchase(package: package)
         } else {
-            return try await performDeveloperPurchaseLogic(package: package)
+            return try await performExternalPurchaseLogic(package: package)
         }
     }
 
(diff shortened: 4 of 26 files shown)


RevenueCatUI/Purchasing/PaywallPurchasesType.swift
```diff
@@ -17,12 +17,17 @@ import RevenueCat
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 protocol PaywallPurchasesType: Sendable {
 
+    var finishTransactions: Bool { get set }
```
jamesrb1, line 20: I find this name a bit ambiguous, and personally I'd want to read the docs to make sure I understood what it meant when it was true or false. Additionally Joop pointed out that it has the opposite logic of Observer Mode, which makes it more likely customers who are migrating from one to the other may not negate the value they are passing, resulting in bugs. Cesar also noted that the concept of "transactions" is iOS-specific.

I have not thought of a perfect alternative, but maybe `customPurchaseImplementation` or `externalPurchaseImplementation`?

Thoughts?

jamesrb1: https://revenuecat.slack.com/archives/C075CNL6YAV/p1717057280915299

jamesrb1: I've updated it to `purchasesAreCompletedBy` of enum type `PurchasesAreCompletedBy`, and deprecated the publicly facing instances of `finishTransactions`. `finishTransactions` is still used internally, unsure if I should update those too....

joshdholtz: @jamesrb1 Should we change this base branch from `main` to `5.0-dev` if we including this enum change in there since its a breaking change? 🤔

jamesrb1: I don't think this should break anything, but you will get deprecation warnings if you use the `.finishTransactions` property on `Purchases`. The config builder still refers to Observer Mode - is that change to be part of this PR too....?

Gemfile.lock
```diff
@@ -338,6 +338,7 @@ PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
   arm64-darwin-23
+  ruby
```
joshdholtz, line 341: This feels like an odd add 😅 Do you know what caused this?

RevenueCatUI/Data/Strings.swift
```diff
@@ -98,6 +103,22 @@ extension Strings: CustomStringConvertible {
 
         case .setting_restored_customer_info:
             return "Setting restored customer info"
+
+        case .executing_external_purchase_logic:
+            return "Will execute custom StoreKit purchase logic provided by yourapp. " +
```
joshdholtz, line 108:
```suggestion
            return "Will execute custom StoreKit purchase logic provided by your app. " +
```

RevenueCatUI/Data/Strings.swift
```diff
@@ -98,6 +103,22 @@ extension Strings: CustomStringConvertible {
...
+            "No StoreKit purchasing logic will be performed by RevenueCat. " +
+            "You must use `.handlePurchase` on your `PaywallView`."
```
joshdholtz, line 110: Are we allowing `handlePurchase` and `handlePurchase` to be used in our SwiftUI view modifiers and in UIKit? If so, we maybe shouldn't referencing `PaywallView` since it wouldn't work in all contexts? 🤔

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -37,8 +37,17 @@ public protocol PurchasesType: AnyObject {
...
      */
+    @available(*, deprecated, message: "Use purchasesAreCompletedBy instead.")
     var finishTransactions: Bool { get set }
 
+    /** Whether transactions should be finished automatically. `.revenueCat` by default.
+     * - Warning: Setting this value to `.myApp` will prevent the SDK from finishing transactions.
+     * In this case, you *must* finish transactions in your app, otherwise they will remain in the queue and
+     * will turn up every time the app is opened.
+     * More information on finishing transactions manually [is available here](https://rev.cat/finish-transactions).
+     */
```
joshdholtz, line 48: For clarity, should we maybe word it like "Purchases will be made and transactions will be finished by RevenueCat"? 🤔

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -40,6 +59,12 @@ final class PurchaseHandler: ObservableObject {
     @Published
     fileprivate(set) var purchaseResult: PurchaseResultData?
 
+    @Published
+    fileprivate(set) var handlePurchase: HandlePurchaseData?
+
+    @Published
+    fileprivate(set) var handleRestore: HandleRestoreCallbackContainer?
+
```
joshdholtz, line 67: Mentioned in Slack but it would be neat if we could combine these in one view modifier because both need to be implemented

Gemfile.lock
jamesrb1: no.... 😬

RevenueCatUI/Data/Strings.swift
jamesrb1: Andy said to only worry about SwiftUI for V1, then decide what to do next...

Gemfile.lock
jamesrb1: Deleted, let's see what happens 😅

Sources/Purchasing/Purchases/PurchasesType.swift
jamesrb1: Yes!

RevenueCatUI/View+PurchaseRestoreCompleted.swift
```diff
@@ -332,6 +385,50 @@ private struct OnPurchaseCancelledModifier: ViewModifier {
...
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+private struct HandleRestoreModifier: ViewModifier {
+
+    let handler: PerformRestore
+
+    func body(content: Content) -> some View {
+        content
+            .onPreferenceChange(HandleRestorePreferenceKey.self) { restoreResultReporter in
+                if let restoreResultReporter {
+                    self.handler(restoreResultReporter)
```
jamesrb1, line 425: The restore method, which is called here, does not provide any information, just a callback which the user can report if it succeeded or not, plus an error. An implementation looks like this:
```swift
}, performRestore: { restoreResultReporter in
    // restore purchases

    // report result to RevenueCat
    restoreResultReporter.reportResult(success: true, error: nil)
})
```
I'm unsure if this is correct, and if we are able to or need to create a correct "CustomerInfo" object after this?

RevenueCatUI/Purchasing/PaywallPurchasesType.swift
JayShortway: Just FYI: we merged to `main` in purchases-android, since it's not breaking indeed, just deprecating.

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -143,6 +263,45 @@ extension PurchaseHandler {
...
+            // after the continuation is set below
+            self.performRestore = RestoreResultReporter(callback: self.reportExternalRestoreResult)
+        }
+
+        self.startAction()
+
+        defer {
+            self.restoreInProgress = false
+            self.actionInProgress = false
+        }
```
joshdholtz, line 285: Nit but can we move defer up to the top of the function so that it gets set if any issues happens in any of the other things about it currently?

RevenueCatUI/View+PurchaseRestoreCompleted.swift
```diff
@@ -332,6 +385,50 @@ private struct OnPurchaseCancelledModifier: ViewModifier {
...
+                }
```
joshdholtz, line 426: Do we need to add a log or something if `restoreResultReporter` is `nil`? Or is that already handled somewhere else? 🤔

joshdholtz (commented):
This looks sooooooo good! Just a few nits and then this will be shipshape 🚢

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -84,8 +148,20 @@ final class PurchaseHandler: ObservableObject {
...
+    // MARK: - Purchase
+
     @MainActor
     func purchase(package: Package) async throws -> PurchaseResultData {
+        if self.purchases.purchasesAreCompletedBy == .revenueCat {
+            return try await performPurchase(package: package)
+        } else {
+            return try await performExternalPurchaseLogic(package: package)
+        }
+    }
```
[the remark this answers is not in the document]
jamesrb1: Yes! Done!

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -70,8 +133,9 @@ final class PurchaseHandler: ObservableObject {
         self.purchases = purchases
     }
 
-    // @PublicForExternalTesting
-    static func `default`() -> Self {
+    /// Returns a new instance of `PurchaseHandler` using `Purchases.shared` if `Purchases`
+    /// has been configured, and using a PurchaseHandler that cannot be used for purchases otherwise.
+    public static func `default`() -> Self {
```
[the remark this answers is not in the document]
jamesrb1: Hmmm it appears not... 🤷‍♂️😅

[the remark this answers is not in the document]
jamesrb1: Actually... I wonder what changed...

image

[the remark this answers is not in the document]
jamesrb1: Ohhh.... this is private _except_ for paywalls tester. So ya it can stay private.

RevenueCatUI/View+PurchaseRestoreCompleted.swift
jamesrb1: This method gets triggered when PurchaseHandler.performRestore changes. It starts off as `nil`, and then during `performExternalRestoreLogic` it gets assigned an object and triggers this callback. We don't assign that variable anywhere else - the only we we could get here with a `nil` value is if we later assigned it back to `nil`, which as it stands now would be a programming error on our part (because the only time we change this is when we want to call the external callback, and we shouldn't be doing that if we don't have a callback to send to them...). (In theory we should be able to force-unwrap it here :yolo:.)

I [added a log message](https://github.com/RevenueCat/purchases-ios/pull/3917/commits/ce8b8e05520479867eae8b5e106c3d671a80b81c) to catch any future changes that may be in error.

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -115,13 +191,57 @@ extension PurchaseHandler {
...
+        return PurchaseResultData(nil, try await self.purchases.customerInfo(), false)
+    }
+
+    @MainActor
+    func reportExternalPurchaseResult(_ userCancelled: Bool, _ error: Error?) {
+        self.actionInProgress = false
+        self.performPurchase = nil
+
+        if let error {
+            self.purchaseError = error
```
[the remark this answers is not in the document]
jamesrb1: Do we currently do any paywalls analytics?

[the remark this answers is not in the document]
jamesrb1: Ah, I see just saw `purchases.track(paywallEvent: event)`

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -40,6 +93,14 @@ final class PurchaseHandler: ObservableObject {
     @Published
     fileprivate(set) var purchaseResult: PurchaseResultData?
 
+    /// Information used to perform a purchase in the app (rather than in RevenueCat)
```
joshdholtz, line 96: Nit on the comment wording 🙈 

```suggestion
    /// Information used to perform a purchase by the app (rather than by RevenueCat)
```

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -40,6 +93,14 @@ final class PurchaseHandler: ObservableObject {
...
+    @Published
+    fileprivate(set) var performPurchase: PurchaseResultReporter?
+
+    /// Information used to perform restoring a purchase in the app (rather than in RevenueCat)
```
joshdholtz, line 100:
```suggestion
    /// Information used to perform restoring a purchase by the app (rather than by RevenueCat)
```

RevenueCatUI/View+PurchaseRestoreCompleted.swift
```diff
@@ -270,6 +285,44 @@ extension View {
         self.environment(\.onRequestedDismissal, action)
     }
 
+    /// Use this method if you wish to execute your own StoreKit purchase and restore logic,
+    /// skipping RevenueCat's. This method is **only** called if `Purchases` is
+    /// confiugured with `purchasesAreCompletedBy` set to `.myApp`. This is typically used
+    /// when migrating from a direct StoreKit implementation to RevenueCat in stages, or if integrating
+    /// RevenueCat for experiments and growth tools only.
+    ///
+    /// After executing your StoreKit purchae code, you **must** communicate the result of your purchase
```
joshdholtz, line 294:
```suggestion
    /// After executing your StoreKit purchase code, you **must** communicate the result of your purchase
```

RevenueCatUI/View+PurchaseRestoreCompleted.swift
joshdholtz: I _think_ this is fine! But @aboedo would know for sure 😊

Sources/Purchasing/Configuration.swift
```diff
@@ -143,12 +151,26 @@ import Foundation
          * RevenueCat's backend. Default is `false`.
          *
          * - Warning: This assumes your IAP implementation uses StoreKit 1.
-         * Observer mode is not compatible with StoreKit 2.
+         * `.myApp` is not compatible with StoreKit 2.
```
joshdholtz, line 154: I think this will end up changing when this gets rebased by `5.0-dev` branch, right? 🤔  cc: @fire-at-will @MarkVillacampa

joshdholtz (commented):
Looking real good! A few really small nits but this is look ship shape 💪 🚢

RevenueCatUI/Purchasing/PurchaseHandler.swift
jamesrb1: Nit away!!

RevenueCatUI/View+PurchaseRestoreCompleted.swift
aboedo: I'm a bit lost on the context, but ideally `restore` gives you the latest customer info when it's done

jamesrb1: Here it's the app developer performing the restore operation (because `purchasesAreCompletedBy == .myApp`), but all they give us back is if the restore succeeded or not (and any associated error). Is that sufficient?

Sources/Purchasing/Configuration.swift
MarkVillacampa: Yep we'll change this when it gets rebased.

You can keep it, that way we'll surely notice when we get a conflict 😄

RevenueCatUI/View+PurchaseRestoreCompleted.swift
jamesrb1: It's the code block in my first comment that is the focus here (but this does appear in the PR because that would be customer code, so I commented on the code that triggers it instead, which is probably confusing!).

RevenueCatUI/Data/Strings.swift
```diff
@@ -98,6 +103,22 @@ extension Strings: CustomStringConvertible {
 
         case .setting_restored_customer_info:
             return "Setting restored customer info"
+
+        case .executing_external_purchase_logic:
+            return "Will execute custom StoreKit purchase logic provided by your app. " +
+            "No StoreKit purchasing logic will be performed by RevenueCat. " +
```
aboedo, line 109: This seems a bit misleading to me - a _lot_ of StoreKit purchasing logic will be performed by RevenueCat regardless. We still fetch products, post receipts, etc, right? 
The only real change is that we won't finish transactions. I realize that that might not be a concept that a user grasps all that well, but then again, we're assuming that this developer has a full StoreKit implementation of their own, so it feels reasonable that they could figure out what finishing transactions is?

RevenueCatUI/Data/Strings.swift
```diff
@@ -98,6 +103,22 @@ extension Strings: CustomStringConvertible {
...
+
+        case .executing_purchase_logic:
+            return "Will execute purchase logic provided by RevenueCat."
+
+        case .executing_restore_logic:
+            return "Will execute restore purchases logic provided by RevenueCat."
+
+        case .executing_external_restore_logic:
+            return "Will execute custom StoreKit restore purchases logic provided by your app. " +
+            "No StoreKit restore purchases logic will be performed by RevenueCat. " +
```
aboedo, line 120: > Will execute custom StoreKit restore purchases logic provided by your app. No StoreKit restore purchases logic will be performed by RevenueCat.

Doesn't quite roll off the tongue, right? It took me a couple of reads to understand this. Maybe we can simplify it a bit? 

how about something like: 
```suggestion
        case .executing_external_restore_logic:
            return "Initiating a restore. Note that the restore logic will be provided by your app and not RevenueCat, since the SDK was configured to blah blah blah...
```

RevenueCatUI/Data/Strings.swift
aboedo: How about something like

> Initiating a purchase for product <product_id>. Note that the StoreKit purchasing logic will be provided by your app and not RevenueCat, since the SDK was configured to blah blah blah.

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -115,13 +193,58 @@ extension PurchaseHandler {
...
+
+    // MARK: - Restore
+
+    func restorePurchases() async throws -> (info: CustomerInfo, success: Bool) {
+        switch self.purchases.purchasesAreCompletedBy {
+        case .revenueCat:
+            return try await performRestorePurchases()
+        case .myApp:
+            return try await performExternalRestoreLogic()
+        }
```
aboedo, line 237: curious about how this works - is the SDK still getting configured in this scenario, and observing queue / delegates?

RevenueCatUI/View+PurchaseRestoreCompleted.swift
aboedo: ohh I see, now I understand the initial question, then. Yes, we'd be able to create a CustomerInfo - assuming the SDK is configured (which I'm guessing it is, otherwise it would crash 😅). We would have to internally call `syncPurchases` however if we want CustomerInfo to reflect stuff that the user might have done

jamesrb1: Ah OK then I will need to add that! And we should do the same after a purchase then too?

aboedo: It should be only for restores - a purchase should still trigger activity in the SK1 queue, which we'll observe, and in SK2 we check purchases after the app comes back into foreground, which should also work.

aboedo: restores are a bit of a special case because nothing actually happens in the queue - when you restore using RC, we just fetch the receipt and send it to the backend

aboedo: restores are a bit of a special case because nothing actually happens in the queue - when you restore using RC, we just fetch the receipt and send it to the backend

jamesrb1: Perfect, thank you!! 🙏

RevenueCatUI/Purchasing/PurchaseHandler.swift
jamesrb1: When `purchasesAreCompletedBy` equals `.revenueCat`, there is no change in behavior. When `purchasesAreCompletedBy` equals `.myApp`, all behavior is as it was when `observerMode` was set to `true` (so I assume this means the answer to your questions about configuration/queues/delegates is "yes"). The change here, which is an addition, is that when someone taps restore on one of our paywalls, the app developer will get a callback via a view modifier that allows them to know this happened and execute their own restore code.

Does that answer everything?

aboedo: makes sense. What happens if I set it up as "purchases completed by my app" but I don't set up the view modifiers?

RevenueCatUI/Data/Strings.swift
jamesrb1: I think my confusion here is regarding exactly what StoreKit logic is outsourced to our customer's app.

This:
> Will execute custom StoreKit purchase logic provided by your app. No StoreKit purchasing logic will be performed by RevenueCat.

sounds the same to me as:
> the StoreKit purchasing logic will be provided by your app and not RevenueCat

The only real change is that we won't finish transactions, but when `purchasesAreCompletedBy` is `.myApp`, we don't call into `PurchasesOrchestrator.purchase(product:package:completion:)` - is this part of finishing transactions?

I think what matters most is that someone who has written or is going to write their own StoreKit interfacing code understands exactly what they are now responsible for. Should we list the StoreKit API calls we no longer call? I'm a bit concerned that unless we're crystal clear, there may be confusion on the customer's side too.

RevenueCatUI/Data/Strings.swift
jamesrb1: But even "Requesting" - requesting to whom? It needs to be clear it's of the app (not StoreKit). 🤔

jamesrb1: "The user has requested a restore. Note that the restore logic will be provided by your app and not RevenueCat, since the SDK was configured to blah blah blah..."  ?

RevenueCatUI/Data/Strings.swift
aboedo: Yeah, that was my specific concern as well - we need to make it clear that you need to finish transactions. 
That's really the only important part. 

Finishing transactions happens specifically when we _detect_ a purchase, then post it to the backend, and receive a response from the backend. 

So even if you don't go through the purchase method, we can still detect the purchase, and that's when we finish transactions (unless this mode is enabled). 

It'd be [finishTransaction](https://developer.apple.com/documentation/storekit/skpaymentqueue/1506003-finishtransaction#) in SK1, [transaction.finish()](https://developer.apple.com/documentation/storekit/transaction) in SK2

RevenueCatUI/Data/Strings.swift
aboedo: ohh I like that

RevenueCatUI/Purchasing/PurchaseHandler.swift
jamesrb1: You'll see a log message in the console that an external purchase/restore has started (we are discussing the exact wording needed in this PR), but if you haven't set up the view modifiers, there will be no error. I tried to think of a good way of detecting this but couldn't, short of setting up a hacky singleton. Discussed somewhere briefly with Josh, did not come to any solution for this. I'll look at this again, it would be a lot better if it actually could identify the bad state....

aboedo: honestly it feels like the way that this is _usually_ solved is that instead of "purchases completed by my app", you provide a delegate. 
If delegate exists, we call it. If not, we use our own logic. That way you can't possibly set it up wrong, right?

jamesrb1: Had a huddle about this with Andy, we are considering moving from view modifiers to a new initializer, to be continued next week. I'm going to separate out the API change so that these discussions don't hold up other projects.

joshdholtz: That makes sense to me! 💪

PR #3865, opened 2024-05-03
WIP: support workflow

aboedo:
https://www.loom.com/share/cd4d9305fb6d43338ee5905f095655a4

Adds the ability to open up a support workflow with a single line of code, `SupportView`. 

This is set up as a new folder within `RevenueCatUI`, and it reuses a lot of stuff from RevenueCat itself. 

It's currently very hacky, but all the functionality that is there works correctly. 

Includes: 
- Open up the support workflow
- If you have an active subscription
    - if subscription is from apple, open up manage subscriptions screen
    - if not, inform them that they need to manage it elsewhere, provide a button to open up management screen of the store
- Ability to restore with a nice-looking flow
- Ability to contact support (although we need to make the actual link configurable through a backend endpoint)
- Ability to begin a refund request through apple (although we'd also want to provide extra UI there with a winback offer)
- Ability to see your current subscription, expiration, renew status, date and price

PR #3919, opened 2024-05-28
[Customer Center] Create `CustomerCenterView`

vegaro:
Base branch is `integration/customer_support_workflow` so we don't merge into `main` yet

Borrows a lot from #3865 

Creates a new `CustomerCenterView` that can be used as a customer support workflow starting point.

All details can be found in https://linear.app/revenuecat/project/sdk-support-workflow-cf7f6a1d5340/overview


RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
```diff
@@ -0,0 +1,92 @@
+//
+//  File.swift
+//  
+//
+//  Created by Cesar de la Vega on 28/5/24.
+//
+
+import Foundation
+
+struct CustomerCenterData: Decodable {
```
vegaro, line 10: Haven't tested decoding this yet

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
```diff
@@ -0,0 +1,92 @@
...
+        let title: LocalizedString
+        let promotionalOffer: PromotionalOffer?
+
+    }
+
+    struct Appearance: Decodable {
+
+        let mode: String
+        let light: String
+        let dark: String
```
vegaro, line 74: This should use something similar to `PaywallColor`

RevenueCatUI/CustomerCenter/URLUtilities.swift
```diff
@@ -0,0 +1,22 @@
...
+
+    static func createMailURL() -> URL? {
+        let subject = "Support Request"
+        let body = "Please describe your issue or question."
+        let encodedSubject = subject.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) ?? ""
+        let encodedBody = body.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) ?? ""
+
+        let urlString = "mailto:support@revenuecat.com?subject=\(encodedSubject)&body=\(encodedBody)"
+        return URL(string: urlString)
+    }
```
vegaro, line 20: We need to make this configurable, but this works for now

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,103 @@
...
+import RevenueCat
+
+@available(iOS 15.0, *)
+class ManageSubscriptionsViewModel: ObservableObject {
+
+    enum State {
+        case notLoaded
+        case success
+        case error(Error)
+    }
```
vegaro: not really doing much with these states yet

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,103 @@
...
+        case .year: return "year"
+        default: return "Unknown"
+        }
+    }
+
+    func periodTitle() -> String {
+        let periodString = "\(self.value) \(self.durationTitle)"
+        let pluralized = self.value > 1 ?  periodString + "s" : periodString
+        return pluralized
+    }
```
vegaro, line 105: We have this copied in `PaywallViewController` as well, maybe we can extract it to a utilities?

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -0,0 +1,79 @@
...
+//  Created by Andrés Boedo on 5/3/24.
+//
+
+import RevenueCat
+import SwiftUI
+
+@available(iOS 15.0, *)
+struct CustomerCenterView: View {
+
+    @StateObject private var viewModel = CustomerCenterViewModel()
```
vegaro, line 14: is a viewmodel overkill?

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -0,0 +1,183 @@
...
+            }
+        }
+    }
+
+}
+
+@available(iOS 15.0, *)
+struct HeaderView: View {
+    var body: some View {
+        Text("How can we help?")
```
vegaro, line 63: All these strings will be localized in the future

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -0,0 +1,183 @@
...
+    var viewModel: ManageSubscriptionsViewModel
+    let openURL: OpenURLAction
+    @State
+    private var showRestoreAlert: Bool = false
+
+    var body: some View {
+        VStack(spacing: 16) {
+            if let configuration = viewModel.configuration {
+                ForEach(configuration.paths, id: \.id) { path in
+                    Button(path.title.en_US) {
```
vegaro, line 117: We will handle localization later

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
```diff
@@ -0,0 +1,92 @@
...
+struct CustomerCenterData: Decodable {
+
+    let id: String
+    let paths: [HelpPath]
+    let title: LocalizedString
+    let supportEmail: String
+    let appearance: Appearance
+
+    enum HelpPathType: String, Decodable {
+        case missingPurchase = "MISSING_PURCHASE"
```
JayShortway, line 19: Should this maybe be `missingEntitlement`? I'm assuming this is the case where the purchase has been made, e.g. money has been transferred, but nothing is received in return. In that case `missingPurchase` doesn't fully describe the situation imo.

RevenueCatUI/CustomerCenter/Data/CustomerCenterTestData.swift
```diff
@@ -0,0 +1,91 @@
...
+                            id: "jargnapocps",
+                            title: CustomerCenterData.LocalizedString(en_US: "Bought by mistake"),
+                            promotionalOffer: nil
+                        )
+                    ]
+                )
+            )
+        ],
+        title: CustomerCenterData.LocalizedString(en_US: "How can we help?"),
+        supportEmail: "ryan@revenuecat.com",
```
JayShortway, line 73: 😄

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
JayShortway: Maybe, seems pretty light on logic. But I'm not opposed to it, even less if it's consistent with other screens.

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
```diff
@@ -0,0 +1,92 @@
...
+        case unknown
+    }
+
+    struct HelpPath: Decodable {
+
+        let id: String
+        let title: LocalizedString
+        let type: HelpPathType
+        let promotionalOffer: PromotionalOffer?
+        let feedbackSurvey: FeedbackSurvey?
```
JayShortway, line 32: Is it always going to be either a `promotionalOffer` OR a `feedbackSurvey`, never both? If so, is there a way to enforce that at compile time? In Kotlin I would suggest a `sealed class/interface`. Can an enum with associated values do something like that in Swift?

JayShortway (commented):
Nice ship! Had some minor comments. I will let someone with more iOS knowhow approve though 😄

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
jamesrb1: I think it's helpful.

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -0,0 +1,184 @@
...
+
+    @Environment(\.openURL)
+    var openURL
+
+    @StateObject
+    private var viewModel = ManageSubscriptionsViewModel()
+
+    init() { }
+
+    init(viewModel: ManageSubscriptionsViewModel) {
```
jamesrb1, line 22: Is this initializer solely for enabling previews and possibly mocks when testing? We could make this method `fileprivate` to keep its interface cleaner. (This may apply to other classes too.)

```suggestion
    fileprivate init(viewModel: ManageSubscriptionsViewModel) {
```

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -0,0 +1,184 @@
...
+            Button("Contact support") {
+                Task {
+                    openURL(URLUtilities.createMailURL()!)
+                }
+            }
+            .padding()
+        }
+    }
+
+    private func handleAction(for path: CustomerCenterData.HelpPath) {
```
jamesrb1, line 134: Should this method be in the view model? It looks like it to me, other than `showRestoreAlert` would need moving in there, unsure on that.

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -0,0 +1,184 @@
...
+            }
+        }
+    }
+}
+
+@available(iOS 15.0, *)
+struct ManageSubscriptionsButtonsView: View {
+
+    @ObservedObject
+    var viewModel: ManageSubscriptionsViewModel
```
jamesrb1, line 108:
```suggestion
    private(set) var viewModel: ManageSubscriptionsViewModel
```

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
```diff
@@ -0,0 +1,106 @@
...
+    private var store: Store?
+
+    @Environment(\.openURL)
+    private var openURL
+
+    init() {
+        self._store = State(initialValue: nil)
+    }
+
+    init(store: Store) {
```
jamesrb1, line 25: If only used for previews/tests.
```suggestion
    fileprivate init(store: Store) {
```

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
```diff
@@ -0,0 +1,106 @@
...
+struct WrongPlatformView: View {
+
+    @State
+    private var store: Store?
+
+    @Environment(\.openURL)
+    private var openURL
+
+    init() {
+        self._store = State(initialValue: nil)
```
jamesrb1, line 22: I don't think this is necessary?
```suggestion
```

jamesrb1 (approved):
I've only run the views via previews - is there a better way?
I wasn't able to see the Figma file, but have requested access.

To get it to compile I had to change `LocalizedString` from `let enUS: String` to `let en_US: String` (I think the linter made this change, but we should ignore its rule in this case), and `CustomerCenterData.FeedbackSurvey.Option` to `CustomerCenterData.FeedbackSurveyOption`.

I think this is going to be a great and widely used new feature.

RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+
+import Foundation
+
+struct SubscriptionInformation {
+
+    let title: String
+    let duration: String
+    let price: String
+    let nextRenewal: String
```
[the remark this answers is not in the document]
vegaro: You're correct, I am going to test deserializing and making these types more semantically correct. Will probably open another PR to keep this one simple and more focus on the view

RevenueCatUI/CustomerCenter/Data/CustomerCenterTestData.swift
vegaro: lol will update this to prevent spam lol

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
vegaro: I will just keep it since it's already done haha. It is the first one I did and then I was like meh, could have been a `@State`. I am not that proficient with SwiftUI and wasn't sure, we should probably write up somewhere some guidelines on this :)

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
vegaro: yes, only for previews and tests

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
vegaro: I will make these changes in another PR so I keep this lean and easy to review. Next PR will take of calling the API and deserialization

vegaro:
>  I've only run the views via previews - is there a better way?

I've been doing that as well, since making a purchase with PaywallsTester has been complicated. I was planning on looking into that today, so I can test deserialization, etc

> To get it to compile I had to change LocalizedString from let enUS: String to let en_US: String (I think the linter made this change, but we should ignore its rule in this case), and CustomerCenterData.FeedbackSurvey.Option to CustomerCenterData.FeedbackSurveyOption.

Oops yes, that's on me. I fixed the linting issues and broke the compilation. Sorry about that

cc @jamesrb1

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
```diff
@@ -0,0 +1,93 @@
+//
+//  File.swift
```
fire-at-will, line 2:
```suggestion
// CustomerCenterData.swift
```

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,107 @@
...
+        case .week: return "week"
+        case .month: return "month"
+        case .year: return "year"
+        default: return "Unknown"
+        }
+    }
+
+    func periodTitle() -> String {
+        let periodString = "\(self.value) \(self.durationTitle)"
+        let pluralized = self.value > 1 ?  periodString + "s" : periodString
```
fire-at-will, line 103: You might want to consider setting this up in a way that can use SwiftUI's automatic grammar agreement functionality so that you don't have to handle pluralization here. The feature isn't super well known but it works like a charm!

You can read a little more on it here: https://www.swiftjectivec.com/morphology-in-ios-with-automatic-grammar-agreement/

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -0,0 +1,79 @@
...
+
+    init() { }
```
fire-at-will, line 16: Nit: This can be removed
```suggestion
```

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -0,0 +1,79 @@
...
+        NavigationView {
+            NavigationLink(destination: destinationView()) {
+                Text("Billing and subscription help")
+                    .padding()
+                    .background(Color.blue)
+                    .foregroundColor(.white)
+                    .cornerRadius(10)
+            }
+        }
+        .onAppear {
```
fire-at-will, line 32: Nit: use `.task` here instead of .onAppear and then you can make `checkAndLoadSubscriptions` async!

https://developer.apple.com/documentation/swiftui/view/task(priority:_:)

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -0,0 +1,184 @@
...
+                SubscriptionDetailsView(subscriptionInformation: subscriptionInformation,
+                                        refundRequestStatus: viewModel.refundRequestStatus)
+            }
+
+            Spacer()
+
+            ManageSubscriptionsButtonsView(viewModel: viewModel,
+                                           openURL: openURL)
+        }
+        .onAppear {
```
fire-at-will, line 40: Nit: this can also be made a .task

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    @Environment(\.dismiss) var dismiss
+    @State private var showRestoreAlert: Bool = false
+
+    var body: some View {
+        VStack {
+            Text("No Subscriptions found")
+                .font(.title)
+                .padding()
+            Text("We can try checking your Apple account for any previously purchased products")
```
fire-at-will, line 22: Nit: a more user-friendly string here might be something like "We can try checking your Apple account for any previous purchases" - I don't think most of my non-tech friends would know what a product means in this context

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
```diff
@@ -0,0 +1,106 @@
...
+                    if let customerInfo = try? await Purchases.shared.customerInfo(),
+                       let firstEntitlement = customerInfo.entitlements.active.first {
+                        self.store = firstEntitlement.value.store
+                    }
+                }
+            }
+        }
+    }
+
+    private func humanReadablePlatformName(store: Store) -> String {
```
fire-at-will, line 69: This is okay for this PR but we should look into finding a different name here for some of the platforms. Most end users won't know what "Stripe", "RCBilling", "External", or even "Promotional" are.

fire-at-will (commented):
This is going to be _so cool_! 🤩

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,107 @@
...
+        state = .success
+    }
+
+    func loadSubscriptionInformation() async {
+        do {
+            let customerInfo = try await Purchases.shared.customerInfo()
+            guard let currentEntitlementDict = customerInfo.entitlements.active.first,
+                  let subscribedProductID = customerInfo.activeSubscriptions.first,
+                  let subscribedProduct = await Purchases.shared.products([subscribedProductID]).first else {
+                return
```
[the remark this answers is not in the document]
vegaro: Created a `CustomerCenterError.couldNotFindSubscriptionInformation`

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,107 @@
...
+            }
+            let currentEntitlement = currentEntitlementDict.value
+
+            self.subscriptionInformation = SubscriptionInformation(
+                title: subscribedProduct.localizedTitle,
+                duration: subscribedProduct.subscriptionPeriod?.durationTitle ?? "",
+                price: subscribedProduct.localizedPriceString,
+                nextRenewal: "\(String(describing: currentEntitlement.expirationDate!))",
```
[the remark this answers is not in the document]
vegaro: Lifetime access will not be part of the `activeSubscriptions`, because that would be a non-subscription product. We should add support for non-subscriptions though. I am going to create a task for that. For now, I think this code is fine

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
vegaro: I check for them in https://github.com/RevenueCat/purchases-ios/pull/3919/files#diff-d35a1e787947e08d6465b6c369913eea0f5cae5f1b29428539e35162a4092ec3R33 so we don't display these weird unrecognizable names. But was also wondering if we could find something like "Alternative Store" in case someone calls this function in the future from another spot. Any ideas?

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
vegaro: Talked with the backend folks and they said let's keep as `MISSING_PURCHASE` for now since it's internal, will bring it back up whenever there's a PR up for the endpoint

JayShortway: Alright, thanks for checking! Maybe we can still think about the user-facing text, but not necessarily now.

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
aboedo: Thoughts on enforcing that on compilation in Swift:

- you could have both types conform to the same protocol and then have a property of that protocol, i.e.: 
```swift
protocol Foo: Decodable {}

struct PromotionalOffer: Foo {
}

struct FeedbackSurvey: Foo {
}

...
    let offerOrSurvey: Foo

```

- alternatively, you could have them be enums with associated values, i.e.: 
```swift
enum Foo {
        case promotionalOffer(PromotionalOffer)
        case feedbackSurvey(FeedbackSurvey)
}
```

I think I'd probably go with the latter but I'm not really a Swift expert either

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
```diff
@@ -0,0 +1,92 @@
...
+
+extension CustomerCenterData {
+
+    static func decode(_ json: String) -> CustomerCenterData {
+        let data = Data(json.utf8)
+        let decoder = JSONDecoder()
+        do {
+            return try decoder.decode(CustomerCenterData.self, from: data)
+        } catch {
+            fatalError("Failed to decode JSON: \(error)")
```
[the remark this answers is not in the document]
aboedo: +1

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
aboedo: the good thing about SwiftUI is that it'll pick up localization automatically here, we won't have to update lines like these with string literals as long as we have the values in String Catalogs

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
aboedo: I think we should just unify them under a single "Web subscription" hat, as an end user that's all you care about. The link will point you in the right direction.

Well, "Promotional" we could rename to something like "free subscription" or something to allude to the fact that you didn't directly pay for it

aboedo (commented):
looking great!!

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
jamesrb1: Maybe we could
`typealias RCColor = PaywallColor`
or change the name of the actual class and reverse the typealias for backward compatibility.

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
```diff
@@ -0,0 +1,93 @@
...
+        let en_US: String
+
+    }
+
+    struct PromotionalOffer: Decodable {
+
+        let iosOfferId: String
+        let eligibility: Eligibility
+
+    }
```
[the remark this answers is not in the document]
vegaro: most likely yes. We had a conversation about having the backend determine if the user was eligible. The backend hasn't been started yet, so maybe we can just ship without that part first, for the sake of shipping

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
vegaro: ohhh nice

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
vegaro: done in 553edba494998359e5add7abcf013f8dbc16143d

RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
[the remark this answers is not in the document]
vegaro: actually, updating them now, since in the other PR, I've decided to create a response object. thanks for pointing this out!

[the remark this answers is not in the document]
vegaro: oh lol this is SubscriptionInformation, which doesn't come from the backend lol, will definitely update these

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
```diff
@@ -0,0 +1,93 @@
...
+        let type: HelpPathType
+        let promotionalOffer: PromotionalOffer?
+        let feedbackSurvey: FeedbackSurvey?
+
+    }
+
+    struct LocalizedString: Decodable {
+        
+        // swiftlint:disable:next identifier_name
+        let en_US: String
```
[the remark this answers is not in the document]
vegaro: leaving this for now since this is most likely going to change. We will probably be passing the locale as a parameter and the API will only return stuff in that locale

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,141 @@
...
+
+@available(iOS 15.0, *)
+class ManageSubscriptionsViewModel: ObservableObject {
+
+    var isLoaded: Bool {
+        if case .notLoaded = state {
+            return false
+        }
+        return true
+    }
```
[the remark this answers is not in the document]
vegaro: It can, but `State` needs to implement `Equatable`

RevenueCatUI/CustomerCenter/Data/CustomerCenterData.swift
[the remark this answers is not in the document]
vegaro: going to remove all eligibility stuff for now

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -0,0 +1,70 @@
...
+        do {
+            let customerInfo = try await Purchases.shared.customerInfo()
+            self.hasSubscriptions = customerInfo.activeSubscriptions.count > 0
+            guard let firstActiveEntitlementStore = customerInfo.entitlements.active.first?.value.store else {
+                self.subscriptionsAreFromApple = false
+                return
+            }
+
+            self.subscriptionsAreFromApple =
+            firstActiveEntitlementStore == .appStore || firstActiveEntitlementStore == .macAppStore
```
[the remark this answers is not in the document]
vegaro: refactored this to something that I believe is more correct. We still need to support non-consumables, but need to think of an optimal way of detecting an active entitlement from a non-consumable 🤔

RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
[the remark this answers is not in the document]
vegaro: ok, gave it another thought, since this is only used as a data class to hold information that is created in the view model I think it's better to leave it as is. I renamed some properties so they better represent what they hold.

vegaro:
It looks like I still have to fix some availability checks, but this is good for another pass I think :)

RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigData.swift
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    struct Appearance {
+
+        let mode: Mode
+        let light: RCColor
+        let dark: RCColor
+
+        enum Mode: String {
+            case system = "SYSTEM"
```
tonidero, line 83: Should this also have light/dark modes?

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -0,0 +1,177 @@
...
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+struct ManageSubscriptionsButtonsView: View {
+
+    @ObservedObject
+    private(set) var viewModel: ManageSubscriptionsViewModel
+
+    let openURL: OpenURLAction
```
tonidero, line 133: Hmm this seems unused? (Maybe you're planning to use it later though)

RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigData.swift
vegaro: going to remove this for now, took this from the draft API and it is meant to be used later as an indication that it should be using the accent color from the app. Since we haven't defined the API completely and I am not using it I will remove it

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
vegaro: Oh yes, I had a contact support button, that I removed because it's not possible to configure it, and this is leftovers. Removing!

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -0,0 +1,90 @@
...
+//
+
+import Foundation
+import RevenueCat
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+class CustomerCenterViewModel: ObservableObject {
```
NachoSoto, line 23: Strongly suggest making these `@MainActor` 😇

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,185 @@
...
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+@MainActor
+class ManageSubscriptionsViewModel: ObservableObject {
+
+    @Published
+    var subscriptionInformation: SubscriptionInformation?
+    @Published
+    var refundRequestStatus: String?
```
tonidero, line 31: Nitpicking but should we rename to `refundRequestStatusMessage` or something like that since it has the actual message? Otherwise, we should change to some sort of enum I think

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,185 @@
...
+                guard let subscriptionInformation = self.subscriptionInformation else { return }
+                let productId = subscriptionInformation.productIdentifier
+                let status = try await purchasesProvider.beginRefundRequest(forProduct: productId)
+                switch status {
+                case .error:
+                    self.refundRequestStatus = "Error when requesting refund, try again"
+                case .success:
+                    self.refundRequestStatus = "Refund granted successfully!"
+                case .userCancelled:
+                    self.refundRequestStatus = "Refund canceled"
```
tonidero, line 127: We probably will need to localize these? I'm seeing that their usage is combined with another string so we would need to localize this (with `String(localized: "")` and the other text where it's used (in order to avoid localizing all the combinations of these

RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigTestData.swift
```diff
@@ -0,0 +1,79 @@
...
+//  CustomerCenterConfigTestData.swift
+//
+//
+//  Created by Cesar de la Vega on 28/5/24.
+//
+
+import Foundation
+import RevenueCat
+
+enum CustomerCenterConfigTestData {
```
tonidero, line 19: Should we compile this only on `DEBUG`?

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,185 @@
...
+@available(watchOS, unavailable)
+private extension SubscriptionPeriod {
+
+    var durationTitle: String {
+        switch self.unit {
+        case .day: return "day"
+        case .week: return "week"
+        case .month: return "month"
+        case .year: return "year"
+        default: return "Unknown"
```
tonidero, line 179: Same for these, probably need to be localized? Separately to avoid having to localize all the combinations I think.

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
tonidero: Separately, but I don't really like how these are localized with SwiftUI, since we don't have any compilation-time check that all the localizations are there... Maybe we should extract all these to strings enums so we have it better organized...

vegaro: right, I agree with the localization... I am surprised xcode doesn't warn or anything. We should come back to this because if I am not wrong it's the same mechanism in Paywalls right?

vegaro:
I ended up removing `LocalizedString` because localization will most likely be handled in a different way and this was making the response more complex [`c0a38a1` (#3919)](https://github.com/RevenueCat/purchases-ios/pull/3919/commits/c0a38a1fcd0f6fdee66c556ced86884fd9b6bd76)

vegaro:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3919?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3919** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3919?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3966** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3966?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,185 @@
...
+        do {
+            try await loadSubscriptionInformation()
+            await loadCustomerCenterConfig()
+            self.state = .success
+        } catch {
+            self.state = .error(error)
+        }
+    }
+
+    func loadSubscriptionInformation() async throws {
```
jamesrb1, line 82:
```suggestion
    private func loadSubscriptionInformation() async throws {
```

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,185 @@
...
+            active: currentEntitlement.isActive
+        )
+    }
+
+    func loadCustomerCenterConfig() async {
+        self.configuration = CustomerCenterConfigTestData.customerCenterData
+    }
+
+    #if os(iOS) || targetEnvironment(macCatalyst)
+    func handleAction(for path: CustomerCenterConfigData.HelpPath) {
```
jamesrb1, line 112: This method looks like it should be `async`, and use the `Task { }` blocks? This way the caller can present a spinner or some other "processing" UI if needed. Then to conveniently call it from a `Button`, you can use our custom `AsyncButton`:
```swift
                    AsyncButton {
                        await self.viewModel.handleAction(for: path)
                    } label: {
                        Text(path.title)
                    }
```
Disclaimer: have not tried running this code.

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,185 @@
...
+        switch path.type {
+        case .missingPurchase:
+            self.showRestoreAlert = true
+        case .refundRequest:
+            Task {
+                guard let subscriptionInformation = self.subscriptionInformation else { return }
+                let productId = subscriptionInformation.productIdentifier
+                let status = try await purchasesProvider.beginRefundRequest(forProduct: productId)
```
jamesrb1, line 120: Currently, any errors thrown inside the task will disappear - they should be caught within the task. (If the method is made async and the task blocks removed, then the compiler will complain.)

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -0,0 +1,185 @@
...
+        try await Purchases.shared.showManageSubscriptions()
+    }
+
+    func customerInfo() async throws -> RevenueCat.CustomerInfo {
+        return try await Purchases.shared.customerInfo()
+    }
+
+    func products(_ productIdentifiers: [String]) async -> [StoreProduct] {
+        return await Purchases.shared.products(productIdentifiers)
+    }
```
jamesrb1, line 163: I personally like omitting the `return` in single-line functions (as is done in the first function here), but either way these should probably be consistent.

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -0,0 +1,192 @@
...
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+@available(visionOS, unavailable)
+private extension ManageSubscriptionsView {
+
+    func checkAndLoadInformation() async {
```
jamesrb1, line 72: Maybe clearer name based on [Apple's similar situation](https://developer.apple.com/documentation/uikit/uiviewcontroller/1621446-loadviewifneeded).
```suggestion
    func loadInformationIfNeeded() async {
```

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -0,0 +1,99 @@
...
+@MainActor class CustomerCenterViewModel: ObservableObject {
+
+    typealias CustomerInfoFetcher = @Sendable () async throws -> CustomerInfo
+
+    @Published
+    var hasSubscriptions: Bool = false
+    @Published
+    var subscriptionsAreFromApple: Bool = false
+    @Published
+    var state: CustomerCenterViewState {
```
jamesrb1, line 32: This could be private(set) too, other than for the tests.

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -0,0 +1,99 @@
...
+            // swiftlint:disable:next todo
+            // TODO: support non-consumables
+            let customerInfo = try await self.customerInfoFetcher()
+            let hasSubscriptions = customerInfo.activeSubscriptions.count > 0
+
+            let subscriptionsAreFromApple = customerInfo.entitlements.active.first(where: {
+                $0.value.store == .appStore || $0.value.store == .macAppStore
+            }).map { entitlement in
+                customerInfo.activeSubscriptions.contains(entitlement.value.productIdentifier)
+            } ?? false
```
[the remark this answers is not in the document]
vegaro: Hmmm, yes I think if the first active entitlement comes from a consumable, but the second active entitlement is from a subscription, this will be false. I think I need to combine this. Good catch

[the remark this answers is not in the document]
vegaro: I think I addressed it in [`9c89313` (#3919)](https://github.com/RevenueCat/purchases-ios/pull/3919/commits/9c89313c18df5b2db0dabed33de6c8855268bca6)

RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigData.swift
```diff
@@ -1,8 +1,63 @@
...
+
+    let id: String
+    let paths: [HelpPath]
+    let title: String
+
+    enum HelpPathType: String {
+        case missingPurchase = "MISSING_PURCHASE"
+        case refundRequest = "REFUND_REQUEST"
+        case changePlans = "CHANGE_PLANS"
+        case cancel = "CANCEL"
```
tonidero, line 29: Oh I was using lowercase for these in the backend... I'm ok changing them to uppercase though 👍

RevenueCatUI/CustomerCenter/URLUtilities.swift
```diff
@@ -0,0 +1,32 @@
...
+
+    static func createMailURL() -> URL? {
+        let subject = "Support Request"
+        let body = "Please describe your issue or question."
+        let encodedSubject = subject.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) ?? ""
+        let encodedBody = body.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) ?? ""
+
+        // swiftlint:disable:next todo
+        // TODO: make configurable
+        let urlString = "mailto:support@revenuecat.com?subject=\(encodedSubject)&body=\(encodedBody)"
```
tonidero, line 28: We will be sending these from the backend soon. But we can address this todo in a separate PR 👍

tonidero (approved):
We probably still need to handle localization and it would be great to have someone more familiar with SwiftUI take a look, but everything makes sense to me!

vegaro:
Thank you @tonidero! Most of the hardcoded strings will go away since we are going to try to localize everything in the backend. So that's for future PRs.

vegaro:
Going to merge it since the PRs are accumulating. @jamesrb1 feel free to make more comments that I can address in another PR if needed

PR #3920, opened 2024-05-29
Send AppTransactions POC

fire-at-will:
This is a POC to see if we can accurately determine `originalAppVersion` and `originalPurchaseDate` with the [AppTransaction](https://developer.apple.com/documentation/storekit/apptransaction) struct provided by StoreKit 2.

### Changes Made
- Send the `AppTransaction` JWS to the POST /receipt endpoint when available in the `app_transaction` optional body param.
- Bump required OS version to use StoreKit 2 from iOS15 to iOS16.

### TODO
- Potentially send `AppTransaction` to backend if there are no purchases on the device. This will be done in a follow-up PR

### DIFF
--- a/Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
+++ b/Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
@@ -31,6 +31,12 @@ protocol StoreKit2TransactionFetcherType: Sendable {
     @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
     var firstVerifiedTransaction: StoreTransaction? { get async }
 
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    var appTransactionJWS: String? { get async }
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    func appTransactionJWS(_ completionHandler: @escaping (String?) -> Void)
+
 }
 
 final class StoreKit2TransactionFetcher: StoreKit2TransactionFetcherType {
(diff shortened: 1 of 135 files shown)


Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -31,6 +31,12 @@ protocol StoreKit2TransactionFetcherType: Sendable {
     @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
     var firstVerifiedTransaction: StoreTransaction? { get async }
 
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    var appTransactionJWS: String? { get async }
+
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    func appTransactionJWS(_ completionHandler: @escaping (String?) -> Void)
```
MarkVillacampa, line 38: When are we using this? It seems we might be fine with just the async var?

fire-at-will: It's used in the TransactionPoster [here](https://github.com/RevenueCat/purchases-ios/blob/bc02c9afa2540df6c4ba8d4c39b5e438935ca03a/Sources/Purchasing/Purchases/TransactionPoster.swift#L110), in a spot that uses completion handlers in a non-async function.

MarkVillacampa: sorry had not seen that! in that case, would it make sense to remove the @available from `appTransactionJWS`  and move the iOS 16 availability check inside it, and just return nil for earlier versions?

fire-at-will: Love this! I'll do that and then the code in TransactionPoster will only need to call `postReceipt` once :D

fire-at-will:
Got a verbal 👍 from @MarkVillacampa via a DM to merge this into the `5.0-fields-poc` branch, where we're centralizing all of the AppTransaction POC work.

PR #3925, opened 2024-05-30
Only POST AppTransaction in syncPurchases if No Transactions Present

fire-at-will:
This PR adjusts the behavior in `syncPurchases()` so that if no purchases are present, we still POST the `AppTransaction` to POST /receipt instead of just fetching the CustomerInfo.

## Description
- If no transactions are present when `syncPurchases()` is called and SK2 is used, then we POST the `AppTransaction` to the POST /receipt endpoint instead of just fetching the CustomerInfo.
- This PR introduces the ability to make a call to POST /receipt without a `fetch_token`.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -1114,24 +1114,28 @@ private extension PurchasesOrchestrator {
         self.attribution.unsyncedAdServicesToken { adServicesToken in
             _ = Task<Void, Never> {
                 let transaction = await self.transactionFetcher.firstVerifiedTransaction
-                guard let transaction = transaction, let jwsRepresentation = transaction.jwsRepresentation  else {
-                    self.customerInfoManager.customerInfo(appUserID: currentAppUserID,
-                                                          fetchPolicy: .cachedOrFetched) { result in
-                        self.operationDispatcher.dispatchOnMainThread {
-                            completion?(result.mapError(\.asPurchasesError))
-                        }
+                let appTransactionJWS = await self.transactionFetcher.appTransactionJWS
+
+                guard let transaction = transaction, let jwsRepresentation = transaction.jwsRepresentation else {
+                    // No receipt is found, just post the AppTransaction.
+                    self.backend.post(receipt: .empty,
+                                      productData: nil,
+                                      transactionData: .init(appUserID: self.appUserID,
+                                                             source: .init(isRestore: true, 
+                                                                           initiationSource: .restore)),
+                                      observerMode: self.observerMode,
+                                      appTransaction: appTransactionJWS) { result in
+
+                        self.handleReceiptPost(result: result,
+                                               transactionData: nil,
+                                               subscriberAttributes: unsyncedAttributes,
+                                               adServicesToken: adServicesToken,
+                                               completion: completion)
                     }
                     return
                 }
 
                 let receipt = await self.encodedReceipt(transaction: transaction, jwsRepresentation: jwsRepresentation)
-
-                let appTransactionJWS: String?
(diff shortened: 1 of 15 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1114,25 +1114,29 @@ private extension PurchasesOrchestrator {
...
-                        }
+                let appTransactionJWS = await self.transactionFetcher.appTransactionJWS
+
+                guard let transaction = transaction, let jwsRepresentation = transaction.jwsRepresentation else {
+                    // No receipt is found, just post the AppTransaction.
+                    self.backend.post(receipt: .empty,
+                                      productData: nil,
+                                      transactionData: .init(appUserID: currentAppUserID,
+                                                             source: .init(isRestore: true,
+                                                                           initiationSource: .restore)),
```
MarkVillacampa, line 1125: Should we propagate the values in the method parameters instead of hardcoding them? Values can be different depending on whether the developer called `restorePurchases` or `syncPurchases`.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1114,25 +1114,29 @@ private extension PurchasesOrchestrator {
...
                 }
 
                 let receipt = await self.encodedReceipt(transaction: transaction, jwsRepresentation: jwsRepresentation)
 
-                let appTransactionJWS: String?
-                if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
-                    appTransactionJWS = await self.transactionFetcher.appTransactionJWS
-                } else {
-                    appTransactionJWS = nil
-                }
```
MarkVillacampa, line 1134: Noice!

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
```diff
@@ -55,8 +55,6 @@ protocol PurchasesOrchestratorTests {
 
     func testSyncPurchasesPostsReceipt() async throws
 
-    func testSyncPurchasesDoesntPostReceiptAndReturnsCustomerInfoIfNoTransaction() async throws
```
MarkVillacampa, line 58: There is a method with this name in the `PurchasesOrchestratorSK1Tests.swift` file, which made me realize, we have an optimization in SK1 where we are not sending the receipt if:

- There are no transactions
- We have a cached `CustomerInfo`
- the cached CustomerInfo has an `originalPurchaseDate`

That logic is here: https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift#L1061

And we even have an SK1-specific test for that case: `testSyncPurchasesSK1PostsReceiptIfNoTransactionsAndEmptyOriginalPurchaseDate`

Would it make sense to add the same optimization to the SK2 path?

In that case, we could add back the `testSyncPurchasesDoesntPostReceiptAndReturnsCustomerInfoIfNoTransaction` test in SK2, and add a  common `testSyncPurchasesPostsReceiptIfNoTransactionsAndEmptyOriginalPurchaseDate` to the tests protocol.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
fire-at-will: Absolutely! Will get that updated 👨‍🔧

Tests/StoreKitUnitTests/PurchasesOrchestratorTests.swift
fire-at-will: Oh, yes, that's a great optimization! Since originalPurchaseDate/originalApplicationVersion should never change, there should be no reason to push up the `AppTransaction`. I'll add in this optimization to the SK2 path! :D

MarkVillacampa (approved):

PR #3926, opened 2024-05-30
James/clone to spm in fastlane

jamesrb1:
### Checklist

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1160,12 +1160,12 @@ jobs:
 
   deploy-to-spm-fastlane:
     docker:
-        - image: cimg/base:stable
-      resource_class: small
-      steps:
-        - run:
-            name: Clone purchases-ios and push to purchases-ios-spm
-            command: bundle exec fastlane deploy_to_spm
+      - image: cimg/base:stable
+    resource_class: small
+    steps:
+      - run:
+          name: Clone purchases-ios and push to purchases-ios-spm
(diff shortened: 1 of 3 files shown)


fastlane/Fastfile
```diff
@@ -876,6 +876,16 @@ platform :ios do
     update_snapshots_commit(new_hash)
   end
 
+  lane :deploy_to_spm do
+    source_repo = ENV["CIRCLE_REPOSITORY_URL"]
```
tonidero, line 880: It's fine to use an environment variable for this, but we could also build the URL here. We have the `repo_name` in this file already so we could just build the URL like `"https://github.com/revenuecat/#{repo_name}"` I think, this way we avoid needing extra environment variables. We could also use the environment variable if it exists but fallback to that, as a way to allow overriding. But as I said, not a blocker.

tonidero (approved):
Just a suggestion but looks good!

.circleci/config.yml
```diff
@@ -1142,22 +1142,14 @@ jobs:
               }
 
   deploy-to-spm:
-    docker:
-      - image: cimg/base:stable
-    resource_class: small
+    <<: *base-job
```
vegaro, line 1145: The base job is a Mac machine. Can we use something that's faster/cheaper? Maybe a ruby docker?

jamesrb1: Done!

PR #3928, opened 2024-05-31
Add SwiftLint as a build tool plugin for the Swift Package project

vegaro:
Following instructions in https://github.com/realm/SwiftLint#swift-package-build-tool-plugins

We had Swiftlint enabled for the Xcode project, but not for the Swift Package project. The Xcode project doesn't have `RevenueCatUI` code, so there was no way to get swiftlint errors in xcode for RevenueCatUI files.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,6 +1,24 @@
 {
   "pins" : [
     {
+      "identity" : "collectionconcurrencykit",
+      "kind" : "remoteSourceControl",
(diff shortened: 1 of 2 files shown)


jamesrb1 (approved):
LGTM!

vegaro:
I think the problem with this approach is that it will bring it for our consumers 🤔 @jamesrb1

vegaro:
going to close this for now, because I think the reorg of the project makes more sense

PR #3870, opened 2024-05-06
Update to use xcode 15.3 in CI

vegaro:
Follow up on https://github.com/RevenueCat/purchases-ios/pull/3858#discussion_r1585087632

Updates the default xcode to 15.3 and cleans up jobs that should use the latest version

PR #3929, opened 2024-05-31
Revert to use xcode 14 to fix deploys

vegaro:
After #3870 some jobs started failing when deploying in #3923 

This PR reverts some of the changes back to xcode 14, so we can still ship and we can upgrade in another set of changes


.circleci/config.yml
```diff
@@ -1278,6 +1275,7 @@ workflows:
         equal: [ scheduled_pipeline, << pipeline.trigger_source >> ]
     jobs:
       - release-checks:
+          xcode_version: '14.3.0'
```
vegaro, line 1278: This was failing due to visionos not being available in intel machines. Also carthage doesn't work with M1 machines, so we can't just upgrade to M1 and have visionOS and run carthage

We need to figure out what we do with carthage, because Intel machines are going away June 28

tonidero (approved):

PR #3945, opened 2024-06-07
[VERY MINOR] Fixes a typo in the bug_report issue template

JayShortway:
Came across this, admittedly great, typo when setting up issue templates for KMP. Kinda love it though 😄

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -29,7 +29,7 @@ A clear and concise description of what the bug is. The more detail you can prov
       - [ ] Simulator
    7. Environment:
       - [ ] Sandbox
-      - [ ] TestFight
+      - [ ] TestFlight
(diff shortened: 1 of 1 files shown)


JayShortway:
Those tests 🥲

vegaro:
@JayShortway just saw this PR which is still valid. You should try rebasing, since tests should have improved by now

JayShortway:
> @JayShortway just saw this PR which is still valid. You should try rebasing, since tests should have improved by now

Thanks, that worked (after another retry haha)!

vegaro (approved):

tonidero:
Looks good! We should just change the PR name so the automations don't pickup the `very minor` note in the CHANGELOG :P

JayShortway:
> Looks good! We should just change the PR name so the automations don't pickup the `very minor` note in the CHANGELOG :P

That's a good point! But it's very important that you all know how minor this is 😛 (changed)

PR #3954, opened 2024-06-12
[WIP] Paywalls manually handle purchases init based

jamesrb1:
This is a WIP making the paywall's ability to handle purchases and restore logic directly via trailing closures in its initializer rather than as a view modifier because it is easier for API consumers to use correctly. Note its base is https://github.com/RevenueCat/purchases-ios/pull/3917 where it is done via view modifiers.


jamesrb1:
Closed in favour of https://github.com/RevenueCat/purchases-ios/pull/3973

PR #3957, opened 2024-06-13
Fix onChangeOf to be internal instead of public

joshdholtz:
### Motivation

Prevent `onChangeOf` from being leaked as a public method

### Description

Remove `public` from `onChangeOf` which will make it `internal`

⚠️ Not marking as deprecated but will put something in the release notes about this


tonidero (approved):
Should we make this change in v5 instead? While people shouldn't be using it, this is still a breaking change...

joshdholtz:
This was already done for v5 in #3788 😅 Closing

PR #3958, opened 2024-06-13
Switch tests for iOS 12 and 13 to M1

tonidero:
### Description
Based on the changes in #3948, this makes it so the iOS 12 and 13 jobs run correctly on M1 machines.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -82,7 +82,7 @@ commands:
       #     command: | # Print all available simulators and install required one
       #         xcodes runtimes
       #         sudo xcodes runtimes install "<< parameters.runtime-name >>"
-      # Xcodes got broken with latest iOS 18 betas. This is a temporary workaround.
+      # Xcodes got broken with latest iOS 18 betas. This is a temporary workaround. https://github.com/XcodesOrg/xcodes/issues/368
       - run:
           name: Install fixed xcodes version
           command: |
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -703,12 +711,14 @@ jobs:
 
   run-test-ios-13:
     <<: *base-job
-    # M1 unsupported
-    resource_class: macos.x86.medium.gen2
```
tonidero, line 707: 👋

tonidero:
You can see tests passing for iOS 12 and 13 a couple commits up

.circleci/config.yml
```diff
@@ -703,12 +711,14 @@ jobs:
...
     steps:
       - checkout
       - install-dependencies:
+          install_xcbeautify: false
+          install_mint: true
           install_swiftlint: << parameters.install_swiftlint >>
+      - macos/install-rosetta
```
vegaro, line 720: Oh it's part of the orb sweet

vegaro (approved):

JayShortway (approved):
Nice job getting this fixed!

.circleci/config.yml
tonidero: Oh sorry, missed that! I think that would make sense if we executed that in fast lane, but as long as we are executing this in circleci, I think this is fine IMO

PR #3960, opened 2024-06-15
Xcode project with RevenueCatUI + Tests

jamesrb1:
### RevenueCatUI and its tests are now accessible in the files navigator:

image

### It add two new schemes:

image

Use RevenueCatUIDev to build RevenueCatUI, and RevenueCatUITests to run its tests.

### Linting works:

image

### Individual tests can be run from the UI:

image

## N.B.

- Tests must be run via the RevenueCatUITest scheme.
- Xcode sometimes has difficulties switching schemes. If you see the errors below, **clean the build folder** and try again and it will work
image
- Building with the RevenueCatUIDev scheme creates some warnings related to switches. Adding in an `@unknown default:` resolves the warning but causes an error in other configurations ("Default will never be executed").
image
- Because the root Package file was not altered, this PR won't have any affect on how our package works when integrated via SPM.


RevenueCatUI/Data/PaywallViewMode+Extensions.swift
```diff
@@ -18,31 +18,40 @@ extension PaywallViewMode {
...
+#endif
         }
     }
 
     var displayAllPlansButton: Bool {
         switch self {
         case .fullScreen: return false
+#if !os(watchOS)
         case .footer: return false
+
```
vegaro, line 33:
```suggestion
```
I think there's an extra line here

vegaro (approved):

aboedo (approved):
awesome

PR #3973, opened 2024-06-19
Paywalls with custom purchase and restore logic handlers

jamesrb1:
## tl;dr

PaywallView has a new constructor that takes `performPurchase` and `performRestore` blocks, which are called to preform purchasing/restore directly by the customer's app when the purchase/restore buttons are tapped on the paywall. This makes it possible to use RC Paywalls when `Purchases` has been configured `.with(purchasesAreCompletedBy: .myApp)`.

Example usage:
```swift
PaywallView(performPurchase: {
    var userCancelled = false
    var error: Error?
    
    // use StoreKit to perform purchase

    return (userCancelled: userCancelled, error: error)
}, performRestore: {
    var success = false
    var error: Error?

    // use StoreKit to perform restore

    return (success: success, error: error)
})
```

## Description

When a `PaywallView` is constructed, a new `PurchaseHandler` is created. The `PurchaseHandler` (an internal RevenueCatUI class) is owned by the `PaywallView`, and it is responsible for executing new purchases and restores. 

When a `PaywallView` is constructed **without** `performPurchase` and `performRestore` blocks, the `PaywallView` creates a `PurchaseHandler` capable of preforming purchases using RevenueCat. When a `PaywallView` is constructed with  `performPurchase` and `performRestore` blocks, it can also make purchases using the customer-supplied closures.

The `PurchaseHandler` is invoked when the user taps the `PurchaseButton`, calling `purchaseHandler.purchase(package: self.selectedPackage.content)`, which branches to either the internal or external purchase code, as defined by `purchasesAreCompletedBy`:

```swift
@MainActor
func purchase(package: Package) async throws {
    switch self.purchases.purchasesAreCompletedBy {
    case .revenueCat:
        try await performPurchase(package: package)
    case .myApp:
        try await performExternalPurchaseLogic(package: package)
    }
}
```

Purchase and Restore blocks can also be assigned for paywall footers:

```swift
MyAppDefinedPaywall()
    .paywallFooter(myAppPurchaseLogic: MyAppPurchaseLogic(performPurchase: { packageToPurchase in
        var userCancelled = false
        var error: Error?

        // use StoreKit to perform purchase

        return (userCancelled: userCancelled, error: error)
    }, performRestore: {
        var success = false
        var error: Error?

        // use StoreKit to perform restore

        return (success: success, error: error)
    }))
 ```

and via `.presentPaywallIfNeeded`:

```swift
MyAppDefinedPaywal()
    .presentPaywallIfNeeded(requiredEntitlementIdentifier: "test", myAppPurchaseLogic: MyAppPurchaseLogic(performPurchase: { packageToPurchase in
        return (userCancelled: false, error: nil)
    }, performRestore: {
        return (success: true, error: nil)
    }))
```

### Testing

A lot needs to be mocked, the purchase/restore blocks that are passed in need to be assigned directly to the `PurchaseHandler`, which is then passed in as part of a `PaywallConfiguration`, both of which are normally constructed internally in the PaywallView constructor.

```swift
    func testHandleExternalRestoreWithPurchaaseHandlers() throws {
        var completed = false
        var customRestoreCodeExecuted = false

        let purchasHandler = Self.externalPurchaseHandler { _ in
            return (userCancelled: true, error: nil)
        } performRestore: {
            customRestoreCodeExecuted = true
            return (success: true, error: nil)
        }

        let config = PaywallViewConfiguration(purchaseHandler: purchasHandler)

        try PaywallView(configuration: config).addToHierarchy()

        Task {
            _ = try await purchasHandler.restorePurchases()
            completed = true
        }

        expect(completed).toEventually(beTrue())
        expect(customRestoreCodeExecuted) == true
    }
```

### Error Handling

For the external code blocks to be called, `purchasesAreCompletedBy` needs to be set to `.myApp` AND the blocks need to be defined. So we need to handle cases when these are not consistent:

1. If someone configures purchasesAreCompletedBy to `.myApp`, and then displays a PaywallView **without** purchase/restore handlers, the SDK will:
* Log an error when the PaywallView is constructed
* When the PaywallView is displayed, replace it with a big red banner when in debug mode, fatalError() in release mode.
* If you purchase directly via the purchase handler (which is an internal class but this is done for testing), it will throw.

2. If someone configures purchases to use `.revenueCat`, and then displays a PaywallView **with** purchase/restore handler, the SDK will:

* Log a warning when the PaywallView is constructed

Rationale:
In case 1, there could be no (acceptable) way to make a purchase, so this is very bad and it needs to be dealt with.

In case 2, they've over-specified the paywall, and while this might be confusing (why isn't my code being called??), it's not as problematic, and we want to make it easy for people to switch from using their purchase logic to our purchase logic.

These checks are made in PaywallView.swift, method `checkForConfigurationConsitency()`.

### UIKit

Not yet supported, will add in a follow-up PR, unlikely to be complicated.

### Paywall Footer and Paywall If Needed
The `performPurchase` and `performRestore` parameters for the paywall footer are contained in a struct rather than as loose parameters of closure types, because doing the latter would create a very poor experience with regards to code completion, where Xcode will always offer complete your code as a trailing closure, but will always use the first closure where the function signature matches.

The following illustrates the problem of using loose closure parameters:

The trailing closure that Xcode auto-creates here does _not_ get called for `performPurchase`, but rather `purchaseStarted`, because it also has a `package` as its input parameter, and comes first in the parameter list 😱.

https://github.com/RevenueCat/purchases-ios/assets/109382862/1f809e7c-b661-4844-89e3-fd846a029531

This is the problematic function signature:

```swift
public func paywallFooter(
        offering: Offering,
        condensed: Bool = false,
        purchaseStarted: PurchaseOfPackageStartedHandler? = nil,
        purchaseCompleted: PurchaseOrRestoreCompletedHandler? = nil,
        purchaseCancelled: PurchaseCancelledHandler? = nil,
        restoreStarted: RestoreStartedHandler? = nil,
        restoreCompleted: PurchaseOrRestoreCompletedHandler? = nil,
        purchaseFailure: PurchaseFailureHandler? = nil,
        restoreFailure: PurchaseFailureHandler? = nil,
        performPurchase: PerformPurchase? = nil,
        performRestore: PerformRestore? = nil
    ) -> some View
 ```

To fix this, get rid of the two new parameters: 
```swift
        performPurchase: PerformPurchase? = nil,
        performRestore: PerformRestore? = nil
```
 
and embed them in a struct:

```swift
public struct MyAppPurchaseLogic {
    public let performPurchase: PerformPurchase
    public let performRestore: PerformRestore

    public init(performPurchase: @escaping PerformPurchase, performRestore: @escaping PerformRestore) {
        self.performPurchase = performPurchase
        self.performRestore = performRestore
    }
}
```

which we use as a last parameter:

```swift
        myAppPurchaseLogic: MyAppPurchaseLogic? = nil
```

 Now the code completes as so:

https://github.com/RevenueCat/purchases-ios/assets/109382862/332ac42d-ccba-42bc-bfc3-702d7870c99c

The `MyAppPurchaseLogic` doesn't complete in perfectly, but once you initialize one of those it goes nicely the rest of the way, including the error messages that instruct you on what you need to return.


RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -58,44 +74,87 @@ final class PurchaseHandler: ObservableObject {
...
 
 }
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 extension PurchaseHandler {
 
+    // MARK: - Purchase
+
     @MainActor
-    func purchase(package: Package) async throws -> PurchaseResultData {
```
jamesrb1, line 88: This was never used anywhere, and we can't create one from the `performExternalPurchaseLogic`, so forget it?

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -108,20 +167,70 @@ extension PurchaseHandler {
...
+    @MainActor
+    func performExternalPurchaseLogic(package: Package) async throws  {
+        Logger.debug(Strings.executing_external_purchase_logic)
+
+        guard let externalPurchaseMethod = self.performPurchase else {
+            throw PaywallError.performPurchaseAndRestoreHandlersNotDefined
+        }
+
+        self.packageBeingPurchased = package
+        self.purchaseResult = nil
```
jamesrb1, line 186: We never set this here because we can't construct a `PurchaseResultData`.

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -143,6 +252,34 @@ extension PurchaseHandler {
...
+            throw PaywallError.performPurchaseAndRestoreHandlersNotDefined
+        }
+
+        defer {
+            self.restoreInProgress = false
+            self.actionInProgress = false
+        }
+
+        self.restoreInProgress = true
+        self.restoredCustomerInfo = nil
```
jamesrb1, line 269: In original restorePurchase logic, `self.restoredCustomerInfo` was never set after the restore had executed.

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -382,7 +382,9 @@ private struct PaywallContainerView: View {
     let onSizeChange: (CGSize) -> Void
 
     var body: some View {
-        PaywallView(configuration: self.configuration)
+        PaywallView(configuration: self.configuration,
+        performPurchase: nil,
+        performRestore: nil)
```
jamesrb1, line 387: No support for external purchase via UIKit....

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/CustomPaywall.swift
```diff
@@ -36,7 +36,7 @@ struct CustomPaywall: View {
                            condensed: self.condensed,
                            fonts: DefaultPaywallFontProvider(),
                            introEligibility: self.introEligibility ?? .default(),
-                           purchaseHandler: self.purchaseHandler ?? .default()
+                           purchaseHandler: self.purchaseHandler ?? .default(performPurchase: nil, performRestore: nil)
```
joshdholtz, line 39: NIt: Left this in another comment but can we make `.default()` with default `nil` for these arguments?

RevenueCatUI/View+PresentPaywall.swift
```diff
@@ -413,7 +426,8 @@ private struct PresentingPaywallModifier: ViewModifier {
         self.fontProvider = fontProvider
         self.customerInfoFetcher = customerInfoFetcher
         self.introEligibility = introEligibility
-        self._purchaseHandler = .init(wrappedValue: purchaseHandler ?? .default())
+        self._purchaseHandler = .init(wrappedValue: purchaseHandler ??
+                                                    .default(performPurchase: nil, performRestore: nil))
```
joshdholtz, line 430: Same here with not needing by making these default to `nil`

RevenueCatUI/Purchasing/PurchaseHandler.swift
jamesrb1: The entire `PurchaseHandler` is internal, so nothing here is publicly accessible.

RevenueCatUI/Purchasing/PurchaseHandler.swift
jamesrb1: No longer the case!!

RevenueCatUI/Purchasing/PurchaseHandler.swift
jamesrb1: Apparently it was done in a Button's handler (?), so `self.setRestored(customerInfo)` has now been added.

RevenueCatUI/Purchasing/PurchaseHandler.swift
jamesrb1: I changed things so we do construct one, although Transaction is `nil` when we make an external purchase, unsure if that's unexpected. But as I wrote earlier, nothing uses this....

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
```diff
@@ -0,0 +1,391 @@
...
+#if !os(watchOS) && !os(macOS)
+
+enum TestError: Error {
+    case error
+}
+
+// swiftlint:disable type_body_length
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@MainActor
+class ZZExternalPurchaseAndRestoreTests: TestCase {
```
jamesrb1, line 30: When this class is named `ExternalPurchaseAndRestoreTests`, the test `testSamplePaywallWithLoadingEligibility` in `Template1ViewTests` reliably fails (although it passes when run on my machine).

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -109,19 +177,78 @@ extension PurchaseHandler {
...
+        }
+
+        if let error = result.error {
+            self.purchaseError = error
+            throw error
+        }
+
+        let resultInfo: PurchaseResultData = (transaction: nil,
+                                             customerInfo: try await self.purchases.customerInfo(),
+                                            userCancelled: result.userCancelled)
```
jamesrb1, line 219: When we do an external purchase, is transaction `nil`? Is the customer info retrieved from purchases correct? Could use some experienced opinions here!

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -143,6 +270,40 @@ extension PurchaseHandler {
...
+
+        if let error = result.error {
+            self.restoreError = error
+            throw error
+        }
+
+        let customerInfo = try await self.purchases.customerInfo()
+
+        // This is done by `RestorePurchasesButton` whend using RevenueCat logic.
+        self.setRestored(customerInfo)
```
jamesrb1, line 302: Could use an opinion on the validity of this code.

RevenueCatUI/Data/Errors/PaywallError.swift
```diff
@@ -25,17 +25,25 @@ enum PaywallError: Error {
     /// The selected offering was not found.
     case offeringNotFound(identifier: String)
 
+    /// The PaywallView must be initizlied with performPurchase and performRestore
```
vegaro, line 28:
```suggestion
    /// The PaywallView must be initialized with ``performPurchase`` and ``performRestore``
```

RevenueCatUI/Data/PaywallViewConfiguration.swift
```diff
@@ -70,16 +70,18 @@ extension PaywallViewConfiguration {
         fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
         displayCloseButton: Bool = false,
         introEligibility: TrialOrIntroEligibilityChecker? = nil,
-        purchaseHandler: PurchaseHandler? = nil
+        purchaseHandler: PurchaseHandler? = PurchaseHandler.default()
     ) {
+        let handler = purchaseHandler ?? PurchaseHandler.default()
```
vegaro, line 75: is this needed now that there's a default value in the parameters?

vegaro (commented):
still going through it, but have a couple of tiny comments

RevenueCatUI/Purchasing/PurchaseHandler.swift
joshdholtz: I _think_ this the right thing to do!

RevenueCatUI/Views/DebugErrorView.swift
```diff
@@ -78,7 +78,6 @@ struct DebugErrorView<Content: View>: View {
         Text(self.description)
             .unredacted()
             .padding()
-            .fixedSize(horizontal: true, vertical: false)
```
joshdholtz, line 81: I _think_ I've used this before for making sure that there are no weird cutoff overflow things with the text 🤔 was this actually not needed?

joshdholtz (commented):
This looks so good 😍 Just a few small comments I think...

Also, do we have an API Tester that needs to be updated for `PaywallView` and the view modifiers? We are adding public API changes so want to make sure have those API Tested somewhere 😇

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
joshdholtz: I would maybe add this in a comment in this file until we figure out _exactly_ what is happening 🤷‍♂️

RevenueCatUI/PaywallView.swift
```diff
@@ -33,19 +33,20 @@ public struct PaywallView: View {
     @Environment(\.locale)
     private var locale
 
-    @StateObject
-    private var purchaseHandler: PurchaseHandler
+    let purchaseHandler: PurchaseHandler
 
-    @StateObject
-    private var introEligibility: TrialOrIntroEligibilityChecker
+    let introEligibility: TrialOrIntroEligibilityChecker
```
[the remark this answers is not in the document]
jamesrb1: Hmmm the reasoning was that this class doesn't observe any changes but I see that it passes it down to subview that do, so that needs to be changed back. I also see that a subview redeclares `introEligibility` as @StateObject again, when it should be using @ObservedObject at this point, so I'll go over this and make sure it's all correct. Thank you!

RevenueCatUI/Views/DebugErrorView.swift
jamesrb1: When this was here a long error message was constrained to a single line, so it would get cut off at the beginning and end as the whole thing remained centred. Removing this makes it wrap to multiple lines.

RevenueCatUI/Data/PaywallViewConfiguration.swift
jamesrb1: No, it's not! The function parameter doesn't need to be optional anymore either.

RevenueCatUI/PaywallView.swift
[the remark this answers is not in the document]
jamesrb1: This is now resolved. `purchaseHandler` has a couple of different possible owners, the `PaywallView` or the `PresentingPaywallModifier`. For this reason the `PaywallView` methods reference the a new property `@ObservedObject var purchaseHandler: PurchaseHandler`, where the object is either owned by the `PaywallView`'s `@StateObject var purchaseHandler: PurchaseHandler` OR the `PresentingPaywallModifier`'s `@StateObject var purchaseHandler: PurchaseHandler`. If `PaywallView` doesn't own the handler, then the @StateObject is set to a dummy handler that should never be used.

RevenueCatUI/PaywallView.swift
```diff
@@ -33,19 +33,29 @@ public struct PaywallView: View {
...
+    // object as `purchaseHandler`.
     @StateObject
+    private var doNotAccessPurchaseHandler: PurchaseHandler
+
+    @ObservedObject
     private var purchaseHandler: PurchaseHandler
 
     @StateObject
-    private var introEligibility: TrialOrIntroEligibilityChecker
+    var introEligibility: TrialOrIntroEligibilityChecker
```
joshdholtz, line 47: Can this go back to `private` or `internal`? 🤔

RevenueCatUI/PaywallView.swift
```diff
@@ -80,31 +94,73 @@ public struct PaywallView: View {
...
                 fonts: fonts,
-                displayCloseButton: displayCloseButton
+                displayCloseButton: displayCloseButton,
+                purchaseHandler: purchaseHandler
             )
         )
     }
 
+    // Notes: Should this be internal, and public only for dev? Should configuration be totally private, yes?
+    // And these purchase blocks should ONLY be defined on the purchase handler?
```
joshdholtz, line 113: Should we remove this note before merging?

RevenueCatUI/PaywallView.swift
```diff
@@ -33,19 +33,29 @@ public struct PaywallView: View {
     @Environment(\.locale)
     private var locale
 
+    // Do NOT reference this object directly, always
+    // use `purchaseHandler`. It is here for lifecycle
+    // purposes only and may or may not be the same
+    // object as `purchaseHandler`.
```
joshdholtz, line 39: Could you explain in this comment what is meant by lifecycle purposes? I know I will totally forget what this means in like a week 😅

joshdholtz (changes requested):
Few thoughts/questions on latest commit but sooooo close! 😁

RevenueCatUI/PaywallView.swift
jamesrb1: Oops yes I never should have checked that in!

RevenueCatUI/PaywallView.swift
jamesrb1: Update the comment to:
```
    // Do NOT reference this object directly, always use `purchaseHandler`.
    // `doNotAccessPurchaseHandler` is here to own the `PurchaseHandler` when
    //  it is owned by this view. When it is owned by an external object, this
    //  references a "dummy" purchase handler.
```

aboedo:
the attention to detail in code completion is 🥇

RevenueCatUI/Purchasing/PurchaseHandler.swift
aboedo: from an external purchase: 
- customer info should be correct after we get the purchase in our backend, which will depend basically on our purchase observing code. So depending on when you're checking, maybe not. If we wanted to guarantee correctness, then... it's tricky. We _could_ check if the result is success, invalidate customerInfo cache, and refresh customerInfo, but that might create a redundant backend request if the customerInfo had already been refreshed (🐼). Otherwise we'd need some other hook to prevent that redundancy, nothing comes to mind for how to do it though. 
- transaction data... we could require it as part of the external purchase method result? WDYT? Although presumably if you already made the purchase yourself, you already have the transaction in the first place.

Tests/RevenueCatUITests/PaywallFooterTests.swift
```diff
@@ -151,11 +151,76 @@ class PaywallFooterTests: TestCase {
...
+                purchaseHandler: handler
+            )
+            .addToHierarchy()
+
+        Task {
+            _ = try? await handler.purchase(package: TestData.packageWithIntroOffer)
+            completed = true
+        }
+
+        expect(completed).toEventually(beTrue())
```
aboedo, line 208: if we make the test method async, we can use await here, right? 
asking because `toEventually` does polling which can be slower, especially when the expectation is not met, so avoiding it might save some test execution time

aboedo: same for all other instances of toEventually

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
aboedo: LOL that must have been a fun conclusion to reach. I would have thrown my computer out the window.

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
```diff
@@ -0,0 +1,393 @@
...
+            callbackOrder.append("onRestoreStarted")
+        })
+        .onRestoreCompleted({ info in
+            callbackOrder.append("onRestoreCompleted")
+            customerInfo = info
+        })
+        .onRestoreFailure({ _ in
+            callbackOrder.append("onRestoreFailure")
+        })
+        .addToHierarchy()
```
aboedo, line 224: I don't mind duplicate code in tests usually but it feels like this is an easy extract?

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
```diff
@@ -0,0 +1,393 @@
...
+
+        Task {
+            _ = try await purchasHandler.restorePurchases()
+            completed = true
+        }
+
+        expect(completed).toEventually(beTrue())
```
aboedo, line 231: we can simplify this a lot with async in the test signature

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
```diff
@@ -0,0 +1,393 @@
...
+        expect(customerInfo).toEventually(be(TestData.customerInfo))
```
aboedo, line 232: also unrelated note: once the  first `toEventually` goes through, the rest can be regular expectations without polling, since Nimble will stop at the line until the first `toEventually` evaluates to `true`, and only then continue on. 

this means slightly faster tests, but also helps us ensure that we only expect stuff to be async once, not more than once

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
```diff
@@ -0,0 +1,393 @@
...
+
+        Task {
+            _ = try await purchasHandler.purchase(package: Self.package)
+        }
+
+        expect(callbackOrder).toEventually(equal(["onPurchaseStarted", "performPurchase", "onPurchaseFailure"]))
+        expect(passedError).toEventually(matchError(TestError.error))
+    }
+
+    func testHandleExternalRestore() throws {
```
aboedo, line 184: nitpick: for all of these test names, I'd usually try to include the expectation in the test name, so that when it fails, the broken part jumps out. 

for example for this test, if it fails, does it mean that we just can't handle restores? does it mean that within the handling, we're not calling restore methods in the right order? all of the above? 
If we're testing more than one thing in a single test case it might not be a bad idea to extract the code a bit and split it into multiple separate tests, so that you can for example have 
`testExternalRestoreReturnsCorrectCustomerInfo`, `testExternalRestoreCallsMethodsInCorrectOrder`, etc

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
```diff
@@ -0,0 +1,393 @@
...
+
+        Task {
+            do {
+                _ = try await purchasHandler.purchase(package: Self.package)
+            } catch {
+                errorThrown = true
+            }
+        }
+
+        expect(errorThrown).toEventually(beTrue())
```
aboedo, line 302: This can be simplified with just async / await, but there are also some concurrency goodies in Nimble now that we could explore

https://github.com/Quick/Nimble/blob/6ba9e680a1962a86510e9ea2c5c79fd43c5eeff4/Sources/Nimble/Nimble.docc/Guides/Concurrency.md

aboedo (commented):
great work on this!

aboedo:
it'd be amazing to be able to resolve the inconsistent configuration at an API level, but I have a hard time coming up with great ways to do it. 

One thought would be to wrap the configuration methods from RevenueCat in RevenueCatUI so that we could also add one that has a purchaseHandler (or force you to set up a purchaseHandler when you set purchasesAreFinishedBy .myapp). 

Another would be to use Swift macros to check that the methods are called consistently. I think that's probably a lot of work, but not impossible if understand it correctly. Not something we need to take care of right away, though

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
jamesrb1: 😢 Ya it was quite annoying!

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
jamesrb1: Ugh I know, I almost mentioned this in the PR writeup. I tried to extract it a few ways but it's not easy, or at least I didn't find a way, I was being blocked because `callbackOrder.append("")` is a mutating function and these modifiers are all escaping closures and it is not allowed to do this. I tried to work around it but came up blank and decided for testing code it wasn't worth continuing to work on. IF there is a good way to extract it though I'd love to use it!!

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
jamesrb1: I thought so too and had looked into it, but it ends up being a bit complicated. For example the existing test:

```swift
    func testOnPurchaseCompleted() throws {
        var customerInfo: CustomerInfo?

        try PaywallView(
            offering: Self.offering.withLocalImages,
            customerInfo: TestData.customerInfo,
            introEligibility: .producing(eligibility: .eligible),
            purchaseHandler: Self.purchaseHandler
        )
            .onPurchaseCompleted {
                customerInfo = $0
            }
            .addToHierarchy()

        Task {
            _ = try await Self.purchaseHandler.purchase(package: Self.package)
        }

        expect(customerInfo).toEventually(be(TestData.customerInfo))
    }
```
If you make it async as below, it fails:
```swift
    func testOnPurchaseCompleted() async throws {
        var customerInfo: CustomerInfo?

        try PaywallView(
            offering: Self.offering.withLocalImages,
            customerInfo: TestData.customerInfo,
            introEligibility: .producing(eligibility: .eligible),
            purchaseHandler: Self.purchaseHandler
        )
            .onPurchaseCompleted {
                customerInfo = $0
            }
            .addToHierarchy()

        _ = try await Self.purchaseHandler.purchase(package: Self.package)


        expect(customerInfo).to(be(TestData.customerInfo))
    }
 ```
 The `onPurchaseCompleted` is called via a preference key change, and something about the way it's being called makes it come after the await continues. Which seems possibly incorrect, but I'll need more time to look into this, I'll come back to it if there is time to get this merged with it, but if not, maybe another PR?

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
jamesrb1: Great, done!

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
jamesrb1: Good point, done!

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
jamesrb1: We're using Nimble 10, a lot of this stuff came in in Nimble 11, current version is Nimble 13. Definitely time for an update, but maybe as part of a separate PR?

RevenueCatUI/Purchasing/PurchaseHandler.swift
jamesrb1: Looking at this again, the only thing that uses `purchaseResult` property (which is declared as `fileprivate`) is our tests. They never look at `transaction:`, and for `customerInfo:`, they just check if it's the same as the mock that was passed in. So I think this is all fine as is.

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
jamesrb1: I started looking at this, and I don't think there are any complicated changes but there are a lot of them (they've described their own time interval type, NimbleTimeInterval), so I think this would be better done as a separate PR.

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
jamesrb1: So I'm back to trying this again, and now the async method above is working. Why now, and why not before, I don't know...

jamesrb1: I think that using preference keys for triggering the modifier closures is not the way to do it if we want to guarantee they've all been executed by any specific time. These are tied to the SwiftUI view update system and when they execute aren't really under our control, so when `Self.purchaseHandler.purchase` returns, it may have set things in motion for `onPurchaseCompleted` to be called, but it may or may not have been called by any specific time. It's also why polling works (because it happens quickly), but checking immediately after the await generally doesn't.

So, if it's possible to change how the modifiers are called so that the ordering of these calls comes as one would expect, then that would be a good idea, but doing so touches quite a bit and should be a different PR.

Tests/RevenueCatUITests/PaywallFooterTests.swift
jamesrb1: These work and I have made the change. They work because we aren't relying on preference key changes, so the order is guaranteed.

Tests/RevenueCatUITests/Templates/ExternalPurchaseAndRestoreTests.swift
jamesrb1: I have gone through the new tests and found the ones that don't rely on preference key timing and made those ones async.

jamesrb1:
I believe I've addressed all the comments, save for a couple that have been left as to-dos for future PRs:
1. Update Nimble to take advantage of asynchronous expectations.
2. Investigate if we can call custom view modifier closures from within the modified view without using Preferences, so that we can run the closures before the triggering call returns.\
\
For example, in the code below, `.onPurchaseCompleted {` may be called after `await Self.purchaseHandler.purchase(package:` returns:
```swift
    func testOnPurchaseCompleted() async throws {
        var customerInfo: CustomerInfo?

        try PaywallView(
            offering: Self.offering.withLocalImages,
            customerInfo: TestData.customerInfo,
            introEligibility: .producing(eligibility: .eligible),
            purchaseHandler: Self.purchaseHandler
        )
        .onPurchaseCompleted {
            customerInfo = $0
        }
        .addToHierarchy()

        _ = try await Self.purchaseHandler.purchase(package: Self.package)

        expect(customerInfo).to(be(TestData.customerInfo))
    }
    ```

RevenueCatUI/Purchasing/PurchaseHandler.swift
jamesrb1: Done!

joshdholtz (approved):
I _tthhiiinnkkk_ this is good to ship! 🚢 💪

joshdholtz (approved):
I _tthhiiinnkkk_ this is good to ship! 🚢 💪

joshdholtz (approved):
I _tthhiiinnkkk_ this is good to ship! 🚢 💪

aboedo (approved):
:shipit:

PR #3978, opened 2024-06-20
Fix Carthage

jamesrb1:
It deletes all non-RevenueCat schemes which I assume are not needed by Carthage, so this should save a bit of time over what it was doing before too because it won't spend time on them.

https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/19899/workflows/a895e06e-305d-439d-9f99-3047a0375594

image

### DIFF
--- a/Tests/InstallationTests/CarthageInstallation/fastlane/Fastfile
+++ b/Tests/InstallationTests/CarthageInstallation/fastlane/Fastfile
@@ -16,7 +16,14 @@ lane :installation_tests do
     # doesn't try to build the other installation tests
     sh "carthage", "update", "--no-build"
     sh "rm", "-rf", "Carthage/Checkouts/purchases-root/Tests/InstallationTests/"
-    sh "rm", "-rf", "Carthage/Checkouts/purchases-root/RevenueCat.xcodeproj/xcshareddata/xcschemes/RevenueCatUIDev.xcscheme"
+
+    # Carthage builds all schemes including ones we don't need, so let's nuke them before proceeding
+    schemes_directory = "Carthage/Checkouts/purchases-root/RevenueCat.xcodeproj/xcshareddata/xcschemes"
+    Dir.glob("#{schemes_directory}/*.xcscheme").each do |file_path|
+      next if File.basename(file_path) == "RevenueCat.xcscheme"
+      sh "rm", "-f", file_path
+    end
+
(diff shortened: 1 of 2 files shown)


Tests/InstallationTests/CarthageInstallation/fastlane/Fastfile
```diff
@@ -16,6 +16,14 @@ lane :installation_tests do
     # doesn't try to build the other installation tests
     sh "carthage", "update", "--no-build"
     sh "rm", "-rf", "Carthage/Checkouts/purchases-root/Tests/InstallationTests/"
+
+    # Carthage builds all schemes including ones we don't need, so let's nuke them before proceeding
+    schemes_directory = "Carthage/Checkouts/purchases-root/RevenueCat.xcodeproj/xcshareddata/xcschemes"
+    Dir.glob("#{schemes_directory}/*.xcscheme").each do |file_path|
+      next if File.basename(file_path) == "RevenueCat.xcscheme"
```
aboedo, line 23: does this mean that we're skipping the test for RevenueCatUI?

aboedo (approved):
Looks good, but maybe we can allow-list revenueCatUI too?

Tests/InstallationTests/CarthageInstallation/fastlane/Fastfile
jamesrb1: RevenueCatUI scheme is only defined in the Package file, which is ignored by Carthage.

jamesrb1: Oh wait that might be wrong, it's there but not shown? Anyway, this is the directory it searches, so it shouldn't affect RevenueCatUI.
```
jamesborthwick@Jamess-MacBook-Pro-2 xcschemes % ls -l 
total 48
-rw-r--r--@ 1 jamesborthwick  staff  4978 13 Apr 14:02 BackendIntegrationTests.xcscheme
-rw-r--r--@ 1 jamesborthwick  staff  3591 13 Apr 14:02 ReceiptParser.xcscheme
-rw-r--r--@ 1 jamesborthwick  staff  7220 14 Jun 14:47 RevenueCat.xcscheme
-rw-r--r--@ 1 jamesborthwick  staff  2609 20 Jun 10:21 RevenueCatUIDev.xcscheme
```

PR #3968, opened 2024-06-17
[Customer Center] Promotional Offers support

vegaro:
Adds displaying a Promotional Offer if the path has an offer id configured

PR #3967, opened 2024-06-17
[Customer Center] Initial implementation

vegaro:
Alpha of the Customer Center. In this branch, it's been disabled behind a compilation flag.

PR #3959, opened 2024-06-14
[Customer Center] Build feedback survey from JSON

vegaro:
Based off #3933 

https://github.com/RevenueCat/purchases-ios/assets/664544/11eae984-294a-4e14-8c40-7c2d50994c09

Can probably use some animations, but tuning that up will come up later

It will open a feedback survey for an option if there's one

PR #3933, opened 2024-05-31
[Customer Center] Add API call to get Customer Center config

vegaro:
Adds `Purchases.shared.loadCustomerCenter()` that calls a new backend endpoint that returns the customer center configuration

This API call doesn't exist yet and it will change. This PR is the ground work so that we don't have to wait for the backend to add this API and we can already pretend the API is there.

PR #3983, opened 2024-06-21
[Customer Center] Add colors support

vegaro:
Builds buttons from the Appearance object in the JSON

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigTestData.swift
+++ b/RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigTestData.swift
@@ -92,12 +92,12 @@ enum CustomerCenterConfigTestData {
         ],
         appearance: .init(
             // swiftlint:disable force_try
-            mode: .custom(accentColor: try! RCColor(light: RCColor(stringRepresentation: "#ffffff"),
-                                                    dark: try! RCColor(stringRepresentation: "#000000")),
-                         backgroundColor: try! RCColor(light: RCColor(stringRepresentation: "#000000"),
-                                                       dark: try! RCColor(stringRepresentation: "#ffffff")),
-                         textColor: try! RCColor(light: RCColor(stringRepresentation: "#000000"),
-                                                 dark: try! RCColor(stringRepresentation: "#ffffff")))
+            mode: .custom(accentColor: try! .init(light: RCColor(stringRepresentation: "#ffffff"),
+                                                  dark: try! RCColor(stringRepresentation: "#000000")),
+                          backgroundColor: try! .init(light: RCColor(stringRepresentation: "#000000"),
+                                                      dark: try! RCColor(stringRepresentation: "#ffffff")),
+                         textColor: try! .init(light: RCColor(stringRepresentation: "#000000"),
+                                               dark: try! RCColor(stringRepresentation: "#ffffff")))
             // swiftlint:enable force_try
         ),
         localization: .init(
--- a/RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigTestData.swift
+++ b/RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigTestData.swift
@@ -91,17 +91,14 @@ enum CustomerCenterConfigTestData {
                   )
         ],
         appearance: .init(
-            mode: .custom,
-            light: .init(
-                accentColor: "#ffffff",
-                backgroundColor: "#000000",
-                textColor: "#000000"
-            ),
-            dark: .init(
-                accentColor: "#000000",
-                backgroundColor: "#ffffff",
-                textColor: "#ffffff"
-            )
+            // swiftlint:disable force_try
+            mode: .custom(accentColor: try! RCColor(light: RCColor(stringRepresentation: "#ffffff"),
+                                                    dark: try! RCColor(stringRepresentation: "#000000")),
+                         backgroundColor: try! RCColor(light: RCColor(stringRepresentation: "#000000"),
+                                                       dark: try! RCColor(stringRepresentation: "#ffffff")),
+                         textColor: try! RCColor(light: RCColor(stringRepresentation: "#000000"),
+                                                 dark: try! RCColor(stringRepresentation: "#ffffff")))
+            // swiftlint:enable force_try
         ),
         localization: .init(
             locale: "en_US",
--- a/RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigTestData.swift
+++ b/RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigTestData.swift
@@ -92,12 +92,9 @@ enum CustomerCenterConfigTestData {
         ],
         appearance: .init(
             // swiftlint:disable force_try
-            mode: .custom(accentColor: try! .init(light: RCColor(stringRepresentation: "#ffffff"),
-                                                  dark: try! RCColor(stringRepresentation: "#000000")),
-                          backgroundColor: try! .init(light: RCColor(stringRepresentation: "#000000"),
-                                                      dark: try! RCColor(stringRepresentation: "#ffffff")),
-                          textColor: try! .init(light: RCColor(stringRepresentation: "#000000"),
-                                                dark: try! RCColor(stringRepresentation: "#ffffff")))
+            mode: .custom(accentColor: try! .init(light: "#ffffff", dark: "#000000"),
+                          backgroundColor: try! .init(light: "#000000", dark: "#ffffff"),
+                          textColor: try! .init(light: "#000000", dark: "#ffffff"))
             // swiftlint:enable force_try
         ),
         localization: .init(
--- a/RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigTestData.swift
+++ b/RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigTestData.swift
@@ -96,8 +96,8 @@ enum CustomerCenterConfigTestData {
                                                   dark: try! RCColor(stringRepresentation: "#000000")),
                           backgroundColor: try! .init(light: RCColor(stringRepresentation: "#000000"),
                                                       dark: try! RCColor(stringRepresentation: "#ffffff")),
-                         textColor: try! .init(light: RCColor(stringRepresentation: "#000000"),
-                                               dark: try! RCColor(stringRepresentation: "#ffffff")))
+                          textColor: try! .init(light: RCColor(stringRepresentation: "#000000"),
+                                                dark: try! RCColor(stringRepresentation: "#ffffff")))
             // swiftlint:enable force_try
         ),
         localization: .init(
--- a/RevenueCatUI/CustomerCenter/ManageSubscriptionsButtonStyle.swift
+++ b/RevenueCatUI/CustomerCenter/ManageSubscriptionsButtonStyle.swift
@@ -23,10 +23,8 @@ import SwiftUI
 @available(watchOS, unavailable)
 struct ManageSubscriptionsButtonStyle: ButtonStyle {
 
-    var appearance: CustomerCenterConfigData.Appearance
-
-    @Environment(\.colorScheme)
-    private var colorScheme
+    @Environment(\.appearance) private var appearance: CustomerCenterConfigData.Appearance
+    @Environment(\.colorScheme) private var colorScheme
 
     func makeBody(configuration: ButtonStyleConfiguration) -> some View {
         configuration.label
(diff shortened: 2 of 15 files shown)


vegaro:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3983?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#3983** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3983?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#3968** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3968?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3959** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3959?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3919** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3919?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3967** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3967?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`

<h2></h2>

RevenueCatUI/CustomerCenter/ManageSubscriptionsButtonStyle.swift
```diff
@@ -22,12 +23,17 @@ import SwiftUI
...
+    private var colorScheme
+
+    func makeBody(configuration: ButtonStyleConfiguration) -> some View {
         configuration.label
             .padding()
             .frame(width: 300)
-            .background(Color.accentColor)
-            .foregroundColor(.white)
+            .background(color(from: appearance))
+            .foregroundColor(colorScheme == .dark ? Color.black : Color.white)
```
JayShortway, line 36: Should this also come from the `Appearance`, maybe something like `onAccentColor`? 

If that's out of scope, maybe we can use an auto-adapting `Color` such as `primary`, instead of reading the `colorScheme`? 
```swift
.foregroundColor(Color.primary)
```

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -31,8 +31,10 @@ struct ManageSubscriptionsView: View {
     @StateObject
     private var viewModel: ManageSubscriptionsViewModel
 
-    init(screen: CustomerCenterConfigData.Screen, localization: CustomerCenterConfigData.Localization) {
-        let viewModel = ManageSubscriptionsViewModel(screen: screen, localization: localization)
+    init(screen: CustomerCenterConfigData.Screen,
+    localization: CustomerCenterConfigData.Localization,
+    appearance: CustomerCenterConfigData.Appearance) {
```
tonidero, line 36: Indentation seems wrong?

tonidero: Also, I wonder if we should pass the whole config data object, instead of each element separately...

RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
```diff
@@ -26,6 +26,8 @@ import SwiftUI
 @available(visionOS, unavailable)
 struct PromotionalOfferView: View {
 
+    private let appearance: CustomerCenterConfigData.Appearance
```
tonidero, line 29: I'm wondering if it would also make sense to include this object in the environment so it's easier to access through the view tree, instead of having to pass it around... Wdyt?

tonidero (approved):
Just some suggestions but LGTM! It might also be good to try to get some iOS dev to double check, but I think this should be ok for now.

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -26,6 +26,7 @@ import RevenueCat
 class ManageSubscriptionsViewModel: ObservableObject {
 
     let screen: CustomerCenterConfigData.Screen
+    let appearance: CustomerCenterConfigData.Appearance
```
JayShortway, line 29: I'm curious what the ViewModel is going to use the `Appearance` for? At first glance it seems like a strictly UI concern (after it is initially fetched).

JayShortway (commented):
Nice nice! Just a question and a minor comment.

RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
JayShortway: Would adding it to the environment not make it harder to test and/or preview? I have no evidence of that specific to SwiftUI haha, but that's generally a downside of implicit dependencies.

tonidero: It's a good point, however, I feel it's pretty easy to mock objects from the environment in SwiftUI for tests. If you make it an environment object, you would add it to the environment as a view modifier `.environmentObject(objectToAddToTheEnvironment)`.

When testing for snapshot tests, you can just inject a custom version of that object using the same modifier, instead of passing it as a parameter. Since I think we will likely only do screenshot tests on the views, I think that would be enough? Lmk what you think!

vegaro: I am thinking of including the whole configuration in the environment. What do you thnk?

RevenueCatUI/CustomerCenter/ManageSubscriptionsButtonStyle.swift
vegaro: I am going to change it to `Color.primary` for now

vegaro: oh, actually... `Color.primary` is the text color. So for light mode it is `black` and for dark mode `white`. For buttons we actually want the opposite

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
vegaro: so it probably should live in the View right?

JayShortway: Yeah that would be my intuition. If the ViewModel were the one fetching and providing it to the View, you could argue it's "data", but I don't think that's what's happening here?

RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
vegaro: would you make a new `Appearance` class that conforms to `ObservableObject` and leave this struct? Or make the struct a class?

JayShortway: Right, but you'd have to remember that you need to provide this mocked environment object, because there's nothing forcing you to do so (as opposed to e.g. constructor params). 

It's definitely not a blocker for me though! (I have yet to run into this 😛)

vegaro: yeah, it's kinda like mixing dependency injection and "singletons"

tonidero: > you'd have to remember that you need to provide this mocked environment object, because there's nothing forcing you to do so

That's true, it's the good thing about explicit dependencies... I do agree with you for dependencies that might only be used on specific places, but for "generic" objects like these (`Localization`, `Appearance`) that need to be accessed in a lot of places, I feel it's cleaner. Happy to discuss it more though!

> would you make a new Appearance class that conforms to ObservableObject and leave this struct? Or make the struct a class?

Hmm right.... Is it possible to pass it as a normal `environment` (not an EnvironmentObject)?

RevenueCatUI/CustomerCenter/Views/TintedProgressView.swift
```diff
@@ -0,0 +1,47 @@
...
+import Foundation
+import RevenueCat
+import SwiftUI
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+@available(visionOS, unavailable)
+struct TintedProgressView: View {
```
JayShortway, line 23: Nice!

JayShortway (commented):
Looking pretty fancy! Cool to get this in! I'll leave approval to someone with a bit more context.

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -28,24 +28,31 @@ public struct CustomerCenterView: View {
 
     @StateObject private var viewModel: CustomerCenterViewModel
 
+    private var appearance: CustomerCenterConfigData.Appearance
+
     /// Create a view to handle common customer support tasks
-    public init(customerCenterActionHandler: CustomerCenterActionHandler? = nil) {
+    public init(customerCenterActionHandler: CustomerCenterActionHandler? = nil,
+                appearance: CustomerCenterConfigData.Appearance = .default) {
```
tonidero, line 35: Same here as in the promotional offers PR. I wonder if we need a default `Appearance` or if we can just pass `nil` to the environment... However, I think this is better in this case since default to `.system` makes more sense than defaulting to a specific language.

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -15,6 +15,9 @@
 
 import Foundation
 
+// swiftlint:disable missing_docs
+public typealias RCColor = PaywallColor
```
tonidero, line 19: Hmm I wonder if we should make this a more global typealias (outside the `CustomerCenter` folder)... Even better, we might want to have the main type be `RCColor` and change everything to that and just keep `PaywallColor` as a typealias instead? But that can come on future PRs as well.

tonidero (approved):
LGTM!

RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
vegaro: Made it a `Environment`, it's quite easy to mock, and it cleans up the code a lot so leaving it like this for now

RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigTestData.swift
```diff
@@ -92,12 +92,12 @@ enum CustomerCenterConfigTestData {
...
-                                                    dark: try! RCColor(stringRepresentation: "#000000")),
-                         backgroundColor: try! RCColor(light: RCColor(stringRepresentation: "#000000"),
-                                                       dark: try! RCColor(stringRepresentation: "#ffffff")),
-                         textColor: try! RCColor(light: RCColor(stringRepresentation: "#000000"),
-                                                 dark: try! RCColor(stringRepresentation: "#ffffff")))
+            mode: .custom(accentColor: try! .init(light: RCColor(stringRepresentation: "#ffffff"),
+                                                  dark: try! RCColor(stringRepresentation: "#000000")),
+                          backgroundColor: try! .init(light: RCColor(stringRepresentation: "#000000"),
+                                                      dark: try! RCColor(stringRepresentation: "#ffffff")),
+                         textColor: try! .init(light: RCColor(stringRepresentation: "#000000"),
```
tonidero, line 99: Do we also need to indent this one?

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -250,13 +269,17 @@ extension CustomerCenterConfigData.Appearance.AppearanceMode {
         case .system:
             self = .system
         case .custom:
+            func parseColorInformation(from response: CustomerCenterConfigResponse.Appearance) throws -> CustomerCenterConfigData.Appearance.ColorInformation {
+                return CustomerCenterConfigData.Appearance.ColorInformation(
+                    light: try RCColor(stringRepresentation: response.light.accentColor),
+                    dark: try RCColor(stringRepresentation: response.dark.accentColor)
```
tonidero, line 272: This is always using the `accentColor` but the callers are also expecting background/text colors right?

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
```diff
@@ -44,10 +44,11 @@ struct NoSubscriptionsView: View {
 
     var body: some View {
         VStack {
-            Text(localization.commonLocalizedString(for: .noSubscriptionsFound))
+            Text(self.configuration.screens[.noActive]?.title ?? "No Subscriptions found")
```
tonidero, line 47: We should come back to this since we shouldn't need to fallback here. Also, we need to build the buttons below from the paths provided by the backend.

tonidero (approved):

PR #3993, opened 2024-06-25
Remove AdClient framework related code

vegaro:
We had a report of a developer that is seeing that users (who are using iOS 18 beta) are reporting app crashes on launch with 55 unique devices so far. They are using using iOS SDK 4.17.7.

The line that is crashing is:

```
client = clientClass.perform(#selector(FakeAfficheClient.sharedClient)).takeUnretainedValue()
```

We just couldn’t have a single variable named `adClient` without an app getting rejected, even if it wasn’t using it at all. We built the `AfficheClient` so prevent app rejections.

It turns out Apple has removed documentation for `AdClient`, and by looking into https://developer.apple.com/documentation/iad/setting-up-apple-search-ads-attribution#overview

> After February 7, 2023, all requests made to the Apple Search Ads iAd Attribution API will return with a value of "iad-attribution" = false, or errors. See ADClient/requestAttributionDetails(_:). Use the AdServices framework for current attribution integration with the Apple Search Ads Campaign Management API for devices using iOS 14.3 and later. Attribution isn't available for downloads and redownloads from devices using iOS 14.2 or earlier.

So this wasn't even working anyways. I removed everything related but left the public API in place to not break anyone.


Sources/Misc/Deprecations.swift
```diff
@@ -130,7 +130,10 @@ public extension Purchases {
     /**
      * Enable automatic collection of Apple Search Ads attribution. Defaults to `false`.
      */
-    @available(*, deprecated, message: "Use Purchases.shared.attribution.enableAdServicesAttributionTokenCollection() instead")
+    @available(*, deprecated, message: """
+    Use Purchases.shared.attribution.enableAdServicesAttributionTokenCollection() instead.
+    AdClient doesn't work after February 7, 2023  so this boolean doesn't have any effect.
+    """)
     @objc static var automaticAppleSearchAdsAttributionCollection: Bool = false
```
aboedo, line 137: looks good, we should probably entirely obsolete for v5 since it doesn't work right now (cc @fire-at-will)

aboedo (approved):

Sources/Misc/Deprecations.swift
vegaro: addressed in #3994

PR #4005, opened 2024-06-28
Throw error for invalid Apple In App Subscription Key in debug POST /receipt requests

fire-at-will:
### Motivation
Making purchases with Store Kit 2 requires developers to upload an In App Purchase Key to RevenueCat. Currently, we log a warning when a developer tries to make a purchase with an invalid In App Purchase Key, but the problem can easily be missed because the purchase still completes, and if offline entitlements are enabled, entitlements can be unlocked, but the RevenueCat backend will never be able to validate and register the purchase because we can't verify the purchase with Apple.

Ultimately, this could lead to developers upgrading from SDK version 4.X to 5.X, not realize that they need to upload their in app purchase key, and ship a build to production that can't properly make purchases.

### Description
This PR modifies the `purchase()` function to throw an error if the app is running in a DEBUG build and the RevenueCat backend returns an error response because the developer has an invalid or missing In App Purchase Key. When it throws an exception, the following error message is logged with the `warn` level:

> Failed to post the transaction to RevenueCat's backend because your Apple In-App Purchase Key is invalid or not present. This error is thrown only in debug builds; in production, it will fail silently. You must configure an In-App Purchase Key. Please see https://rev.cat/in-app-purchase-key-configuration for more info.

The PR accomplishes this behavior by modifying the `CustomerInfoResponseHandler` class. To ensure that the error is only thrown for `POST /receipt` calls and not `GET /subscriber` calls, a new initializer parameter `failIfInvalidSubscriptionKeyDetectedInDebug` has been introduced to control the throwing behavior when the error is received from the backend.

### DIFF
--- a/Sources/Error Handling/BackendError.swift
+++ b/Sources/Error Handling/BackendError.swift
@@ -24,6 +24,7 @@ enum BackendError: Error, Equatable {
     case missingReceiptFile(URL?, Source)
     case missingTransactionProductIdentifier(Source)
     case missingCachedCustomerInfo(Source)
+    case invalidAppleSubscriptionKey(Source)
     case unexpectedBackendResponse(UnexpectedBackendResponseError, extraContext: String?, Source)
 
 }
@@ -116,6 +117,13 @@ extension BackendError: PurchasesErrorConvertible {
                                                         fileName: source.file,
                                                         functionName: source.function,
                                                         line: source.line)
+        case let .invalidAppleSubscriptionKey(source):
+            return ErrorUtils.unknownError(
+                message: ErrorCode.invalidAppleSubscriptionKeyError.description,
+                fileName: source.file,
+                functionName: source.function,
+                line: source.line
+            )
         }
     }
 
@@ -149,6 +157,7 @@ extension BackendError {
         case .missingAppUserID,
              .emptySubscriberAttributes,
              .missingReceiptFile,
+             .invalidAppleSubscriptionKey,
              .missingTransactionProductIdentifier,
              .missingCachedCustomerInfo,
              .unexpectedBackendResponse:
@@ -168,6 +177,7 @@ extension BackendError {
         case .missingAppUserID,
                 .emptySubscriberAttributes,
                 .missingReceiptFile,
+                .invalidAppleSubscriptionKey,
                 .missingTransactionProductIdentifier,
                 .missingCachedCustomerInfo:
             return nil
--- a/Sources/Error Handling/BackendError.swift
+++ b/Sources/Error Handling/BackendError.swift
@@ -118,7 +118,7 @@ extension BackendError: PurchasesErrorConvertible {
                                                         functionName: source.function,
                                                         line: source.line)
         case let .invalidAppleSubscriptionKey(source):
-            return ErrorUtils.unknownError(
+            return ErrorUtils.configurationError(
                 message: ErrorCode.invalidAppleSubscriptionKeyError.description,
                 fileName: source.file,
                 functionName: source.function,
--- a/Sources/Logging/Strings/ConfigureStrings.swift
+++ b/Sources/Logging/Strings/ConfigureStrings.swift
@@ -73,11 +73,11 @@ enum ConfigureStrings {
 
     case sk2_required_for_swiftui_paywalls
 
-    case handle_transaction_observer_mode_required
+    case record_purchase_requires_purchases_made_by_my_app
 
     case sk2_required
 
-    case fail_post_receipt_due_to_invalid_in_app_purchase_key_configured_in_debug
+    case sk2_invalid_inapp_purchase_key
 }
 
 extension ConfigureStrings: LogMessage {
@@ -189,7 +189,7 @@ extension ConfigureStrings: LogMessage {
             "transactions coming from SwiftUI paywalls. You must use `.with(storeKitVersion: .storeKit2)` " +
             "when configuring the SDK."
 
-        case .handle_transaction_observer_mode_required:
+        case .record_purchase_requires_purchases_made_by_my_app:
             return "Attempted to manually handle transactions with purchasesAreCompletedBy not set to .myApp. " +
             "You must use `.with(purchasesAreCompletedBy: .myApp, storeKitVersion: .storeKit2)` " +
             "when configuring the SDK."
@@ -198,7 +198,7 @@ extension ConfigureStrings: LogMessage {
             return "StoreKit 2 must be enabled. You must use `.with(storeKitVersion: .storeKit2)` " +
             "when configuring the SDK."
 
-        case .fail_post_receipt_due_to_invalid_in_app_purchase_key_configured_in_debug:
+        case .sk2_invalid_inapp_purchase_key:
             return "Failed to post the transaction to RevenueCat's backend because your Apple In-App Purchase Key is " +
             "invalid or not present. This error is thrown only in debug builds; in production, it will fail " +
             "silently. You must configure an In-App Purchase Key. Please see " +
--- a/Sources/Logging/Strings/ConfigureStrings.swift
+++ b/Sources/Logging/Strings/ConfigureStrings.swift
@@ -77,6 +77,7 @@ enum ConfigureStrings {
 
     case sk2_required
 
+    case fail_post_receipt_due_to_invalid_in_app_purchase_key_configured_in_debug
 }
 
 extension ConfigureStrings: LogMessage {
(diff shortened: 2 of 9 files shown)


Sources/Error Handling/BackendError.swift
```diff
@@ -116,6 +117,13 @@ extension BackendError: PurchasesErrorConvertible {
                                                         fileName: source.file,
                                                         functionName: source.function,
                                                         line: source.line)
+        case let .invalidAppleSubscriptionKey(source):
+            return ErrorUtils.unknownError(
```
MarkVillacampa, line 121: Would it make sense to use `.configurationError` here?

fire-at-will: Definitely! Will switch it

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -17,11 +17,26 @@ class CustomerInfoResponseHandler {
...
+    private let failIfInvalidSubscriptionKeyDetectedInDebug: Bool
+    private let isDebug: () -> Bool
 
     /// - Parameter offlineCreator: can be `nil` if offline ``CustomerInfo`` shouldn't or can't be computed.
-    init(offlineCreator: OfflineCustomerInfoCreator?, userID: String) {
+    init(
+        offlineCreator: OfflineCustomerInfoCreator?,
+        userID: String,
+        failIfInvalidSubscriptionKeyDetectedInDebug: Bool,
+        isDebug: @escaping () -> Bool = {
```
fire-at-will, line 28: Added this in to make it possible to inject the debug value during unit tests

Sources/Logging/Strings/ConfigureStrings.swift
```diff
@@ -196,6 +197,12 @@ extension ConfigureStrings: LogMessage {
         case .sk2_required:
             return "StoreKit 2 must be enabled. You must use `.with(storeKitVersion: .storeKit2)` " +
             "when configuring the SDK."
+
+        case .fail_post_receipt_due_to_invalid_in_app_purchase_key_configured_in_debug:
```
MarkVillacampa, line 201: I'd call this something that indicates it's SK2, and a bit shorter, since the return string should be enough explanation: `sk2_invalid_inapp_purchase_key`

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -17,11 +17,26 @@ class CustomerInfoResponseHandler {
 
     private let offlineCreator: OfflineCustomerInfoCreator?
     private let userID: String
+    private let failIfInvalidSubscriptionKeyDetectedInDebug: Bool
+    private let isDebug: () -> Bool
```
MarkVillacampa, line 21: should this store the boolean directly instead of the lambda?

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -55,7 +70,27 @@ class CustomerInfoResponseHandler {
 
         _ = Task<Void, Never> {
             do {
-                completion(.success(try await offlineCreator.create(for: self.userID)))
+                switch result {
+                case .success:
+                    completion(.success(try await offlineCreator.create(for: self.userID)))
+                case .failure(let failure):
+                    let failIfInvalidSubscriptionKeyDetectedInDebug = self.failIfInvalidSubscriptionKeyDetectedInDebug
```
MarkVillacampa, line 77: VERY NIT, but since we're not reusing this variable, can we just call it directly before?

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
```diff
@@ -55,7 +70,27 @@ class CustomerInfoResponseHandler {
...
+
+                    if isDebug() && failIfInvalidSubscriptionKeyDetectedInDebug,
+                        case let .networkError(networkError) = failure,
+                        case let .errorResponse(errorResponse, _, _) = networkError,
+                        errorResponse.code == .invalidAppleSubscriptionKey {
+
+                        Logger.warn(
+                            Strings.configure.fail_post_receipt_due_to_invalid_in_app_purchase_key_configured_in_debug
```
MarkVillacampa, line 85: Should we log this error in non-debug builds too? not sure how often that would happen but can't hurt, right?

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
MarkVillacampa: I'm thinking we could even make this an auto-executable lambda:

```swift
isDebug: Bool = {
            #if DEBUG
            return true
            #else
            return false
            #endif
        }()
```

Tests/UnitTests/OfflineEntitlements/CustomerInfoResponseHandlerTests.swift
```diff
@@ -235,7 +235,112 @@ class OfflineCustomerInfoResponseHandlerTests: BaseCustomerInfoResponseHandlerTe
             level: .error
         )
     }
+}
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+// swiftlint:disable:next type_name
+class CustomerInfoResponseHandlerThrowsErrorWithInvalidInAppPurchaseKeyTests: BaseCustomerInfoResponseHandlerTests {
```
MarkVillacampa, line 242: 🎉

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
fire-at-will: I'm down with the idea of making it an autoExecutable lambda! Making it a variable that can be overridden later feels risky, in the unlikely scenario that it gets modified in an unintended way at runtime 🤷‍♂️

Will make the change!

Sources/Logging/Strings/ConfigureStrings.swift
fire-at-will: Yeah, the changes James did only addressed the _public_ API changes with observer mode, not the internal ones. I'll go ahead and update this one here to leave it a little better than I found it! 😄

fire-at-will: Renamed to `record_purchase_requires_purchases_made_by_my_app`!

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
fire-at-will: Yeah, I've made this a bool and switched the constructor argument to use the autocompleting lambda!

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
fire-at-will: Shockingly not! 🤯

When we call it directly, we get a Swift 6 language mode error at compile time about an implicit capture of `self`:

```swift
Implicit capture of 'self' requires that 'CustomerInfoResponseHandler' conforms to `Sendable
```

I don't _fully_ understand it, but directly referencing it like this apparently gets around the self capture because it's only capturing one property on self, not the entire self.

Another option would be to make `CustomerInfoResponseHandler` conform to `Sendable`, but several of its dependencies don't confirm to `Sendable`, and I felt that a big Sendable refactor was a little out of scope for this PR.

Swift 6 Language Mode is hard™️ 😂

Sources/Networking/Operations/Handling/CustomerInfoResponseHandler.swift
fire-at-will: I really don't think this is necessary since we're also logging [this](https://github.com/RevenueCat/purchases-ios/blob/1b0fd5c802226e403d01794671cf550613d7a6d9/Sources/Error%20Handling/ErrorCode.swift#L129) when an invalid in app purchase key is detected in both debug and non-debug builds. 

While this means we're logging two things in debug, this message clarifies that the thrown error is only thrown in debug builds. If we'd like to consolidate these messages into one message, I can easily change the original log message to say that it's an error in debug builds and remove the log message in this PR.

fire-at-will:
@MarkVillacampa thanks for the review! 🙌 I've made the following changes:
- Renamed the case names in `ConfigureStrings.swift` as mentioned in your review notes
- In `CustomerInfoHandler.swift`:
   - Changed the `private let isDebug` property from `() -> Void` to `Bool`
   - Made the closure argument `isDebug` in the constructor a `Bool` and made the default closure value auto complete

MarkVillacampa (approved):

PR #4008, opened 2024-07-01
Add `X-Preferred-Locales` header

vegaro:
Added a new `X-Supported-Locales` header to be used in the new customer center feature

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -102,7 +102,7 @@ class HTTPClient {
 
     // Visible for tests
     var defaultHeaders: RequestHeaders {
-        let supportedLocales = self.systemInfo.preferredLanguages.prefix(3).map {
+        let preferredLanguages = self.systemInfo.preferredLanguages.prefix(3).map {
             $0.replacingOccurrences(of: "-", with: "_")
         }.joined(separator: ",")
         var headers: RequestHeaders = [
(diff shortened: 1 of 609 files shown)


Sources/Misc/Locale/PreferredLocalesProviderType.swift
tonidero, line 1: I would rename the file to the implementation, so `PreferredLocalesProvider`.

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -102,6 +102,9 @@ class HTTPClient {
 
     // Visible for tests
     var defaultHeaders: RequestHeaders {
+        let supportedLocales = self.systemInfo.preferredLanguages.map {
+            $0.replacingOccurrences(of: "-", with: "_")
+        }.joined(separator: ",")
```
tonidero, line 107: I would add a maximum here... Maybe first 3? Just so this header can't grow indefinitely.

tonidero (commented):
Just a couple comments but looks good!

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -102,6 +102,9 @@ class HTTPClient {
 
     // Visible for tests
     var defaultHeaders: RequestHeaders {
+        let supportedLocales = self.systemInfo.preferredLanguages.prefix(3).map {
```
tonidero, line 105: Maybe rename to `preferredLocales`?

tonidero (approved):

PR #4009, opened 2024-07-01
Obsolete all observer mode configure functions

fire-at-will:
I found two observer mode configure functions this morning that were deprecated but not obsoleted. This PR obsoletes them to match the obsoletion of the other observerMode configure functions.


aboedo (approved):
good catch

PR #4012, opened 2024-07-02
Fixed error with target/platform when generating docs

joshdholtz:
### Motivation

Doc generation is broken on 5.0 deploy with SDK target/platform

### Description

Using new `--sdk` option when building doc symbols


fastlane/Fastfile
```diff
@@ -811,7 +815,8 @@ platform :ios do
            hosting_base_path,
            "--transform-for-static-hosting",
            "--enable-inherited-docs",
-           "--additional-symbol-graph-dir", ".build")
+           "--additional-symbol-graph-dir", ".build",
```
joshdholtz, line 818: ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ ⚠️ 

❗ This is broken right now. Its errors with error: Symbol with identifier 's:8Dispatch0A12TimeIntervalO' couldn't be found` and not really sure how to track it down

Temporary fix would be comment this ou but I don't know the side effect 🤷‍♂️

joshdholtz:
Closing for now since our fix is using Xcode 14 in #4014

PR #4013, opened 2024-07-02
Use available resource class for backend-integration-tests-offline-job

fire-at-will:
### Description
The `5.0-dev` branch was using the `macos.x86.medium.gen2` resource class for the `backend-integration-tests-offline-job`, but that resource class was removed from CircleCI on June 28, 2024 ([source](https://discuss.circleci.com/t/macos-intel-support-deprecation-in-january-2024/48718)). Since the resource class is now unavailable, this job currently fails 100% of the time on `main`.

This PR brings the job onto the same M1 resource as most other jobs. While the tests may be more flaky on M1 machines, the job retry mechanism we have in place now should help counter this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1271,7 +1271,7 @@ workflows:
               ignore: /pull\/[0-9]+/
       - backend-integration-tests-offline:
           # These tests are flaky due to FB13133387 so we don't want the noise in every PR
-          <<: *release-branches-and-main
+          # <<: *release-branches-and-main
       - backend-integration-tests-custom-entitlements:
           filters:
             branches:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1270,7 +1270,7 @@ workflows:
               # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
               ignore: /pull\/[0-9]+/
       - backend-integration-tests-offline:
-          xcode_version: "15.2"
+          xcode_version: "15.3"
           install_swiftlint: false
           # These tests are flaky due to FB13133387 so we don't want the noise in every PR
           # <<: *release-branches-and-main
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -823,8 +823,6 @@ jobs:
 
   backend-integration-tests-offline:
     <<: *base-job
-    # These tests are even flakier running on M1
-    resource_class: macos.x86.medium.gen2
```
tonidero, line 827: Just to check but should we make sure it works before merging? (maybe you have already). We can just add it to the workflow executed on every commit temporarily, make sure they pass, then revert that commit.

tonidero (approved):
Changes look good! Just a comment.

.circleci/config.yml
fire-at-will: I'll check and verify!

.circleci/config.yml
```diff
@@ -1272,8 +1270,10 @@ workflows:
               # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
               ignore: /pull\/[0-9]+/
       - backend-integration-tests-offline:
+          xcode_version: "15.2"
```
vegaro, line 1273: The base job uses 15.3 and iOS 17.4 which is why the simulator can't be found in https://circleci.com/gh/RevenueCat/purchases-ios/231400

fire-at-will: THANK YOU!!! I was just trying to figure that out 😅

.circleci/config.yml
fire-at-will: @tonidero The job passed! I'll revert the changes I made to run it in the CI here and then merge it into main.

tonidero: Perfect! Thanks! 🎉

PR #4014, opened 2024-07-02
CI Build Docs Improvements

fire-at-will:
### Motivation
The doc building process can be fragile at times, and since we only build the docs when we make a deployment, we can be unaware when changes in individual PRs can break our docs' build. This PR aims to make changes that break the docs building process visible earlier in the development cycle to help avoid tricky failures during deployments.

### Description
This PR makes two main changes to how docs are built in the CI process:
- Fixes the existing docs job in the deploy pipeline by having it run on Xcode 14.3.0. This fix comes from an [earlier commit](https://github.com/RevenueCat/purchases-ios/commit/f05f0cf14aaf061d7d8f17d2f4f6ac1fec9200a2) by @vegaro that likely got overridden when `5.0-dev` was merged into `main` recently.
- Introduces a job to build the docs on each PR instead of only during the deploy pipeline
   - It splits the existing `docs-build` into two separate jobs:
      - `docs-build`: Only builds the docs, and is run on each build.
      - `docs-deploy`: Builds and deploys the docs. Replaces the old `docs-build` job and is only run in the deploy pipeline.
  - In Fastlane, we've split the existing `generate_docs` lane into two separate lanes and one private lane:
     - `generate_docs`: This now only builds the docs by running the `generate_docs_shared` lane
     - `generate_and_publish_docs`: Builds the docs by running the `generate_docs_shared` lane and then publishing the docs

### Testing
I've been able to test the `docs-build` job in this PR, but haven't been able to fully test the `generate_and_publish_docs` job since we haven't executed another deployment since 5.0.0. If anyone has any ideas to test `generate_and_publish_docs`, let me know! 😄

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -774,13 +774,13 @@ platform :ios do
   end
 
   desc "Generate docs without publishing"
-  lane :generate_docs do
-    generate_docs_shared
+  lane :build_docs do
+    build_docs_shared
   end
 
   desc "Generate & publish docs"
-  lane :generate_and_publish_docs do
-    generate_docs_shared
+  lane :build_and_publish_docs do
+    build_docs_shared
 
     version_number = current_version_number
     docs_repo_base_url = ENV["DOCS_REPO_BASE_URL"]
@@ -817,7 +817,7 @@ platform :ios do
   end
 
   desc "Builds the docs"
-  private_lane :generate_docs_shared do
+  private_lane :build_docs_shared do
     ENV["INCLUDE_DOCC_PLUGIN"] = "true"
 
     version_number = current_version_number
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -780,7 +780,7 @@ platform :ios do
 
   desc "Generate & publish docs"
   lane :build_and_publish_docs do
-    build_docs_shared
+    docs_generation_folder = build_docs_shared
 
     version_number = current_version_number
     docs_repo_base_url = ENV["DOCS_REPO_BASE_URL"]
@@ -800,7 +800,7 @@ platform :ios do
           # copy docs generated in the previous step into the docs folder
           # and push the changes
           docs_destination_folder = "docs/#{version_number}"
-          FileUtils.cp_r @docs_generation_folder + "/.", docs_destination_folder
+          FileUtils.cp_r docs_generation_folder + "/.", docs_destination_folder
           docs_files.each do |file|
             FileUtils.cp file, "docs/"
           end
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -773,22 +773,63 @@ platform :ios do
...
+          sh("git", "add", "docs/*.html")
+          sh("git", "commit", "-m", "Update documentation for #{version_number}")
+          sh("git", "push")
+        end
+      end
+    end
+  end
+
+  desc "Builds the docs"
+  private_lane :generate_docs_shared do
```
aboedo, line 820: nitpick: I'd call this one "build docs" and the other one "build_and_deploy" to reuse the language from other lanes

Also, having `generate_docs` be the one that both builds and deploys was not my brightest naming moment, feel free to rename

fastlane/Fastfile
```diff
@@ -773,22 +773,63 @@ platform :ios do
...
+    # clone docs repo
+    Dir.mktmpdir do |docs_repo_clone_dir|
+      Dir.chdir(docs_repo_clone_dir) do
+        sh("git", "clone", docs_repo_url)
+        Dir.chdir(docs_repo_name) do
+          # copy docs generated in the previous step into the docs folder
+          # and push the changes
+          docs_destination_folder = "docs/#{version_number}"
+          FileUtils.cp_r @docs_generation_folder + "/.", docs_destination_folder
+          docs_files.each do |file|
```
aboedo, line 804: I'd forgotten about this particular rube goldberg machine

aboedo (approved):

fastlane/Fastfile
fire-at-will: Done! We now have the public lanes `build_docs` and `build_and_publish_docs`, and a private lane `build_docs_shared`.

fastlane/Fastfile
fire-at-will: This is definitely one where ChatGPT helped me understand it 100% times faster :D

fastlane/Fastfile
```diff
@@ -773,22 +773,63 @@ platform :ios do
...
+    version_number = current_version_number
+    docs_repo_name = ENV["DOCS_REPO_NAME"]
+    ios_version = ENV['DOCS_IOS_VERSION'] || "16.1"
+
+    UI.user_error!("Missing environment variable: DOCS_REPO_NAME") unless docs_repo_name
+
     hosting_base_path = File.join(docs_repo_name, version_number)
 
     Dir.mktmpdir do |docs_generation_folder|
+      @docs_generation_folder = docs_generation_folder
```
joshdholtz, line 832: I'd prefer not to use a local variable here if possible 😇 I'd just return this value out of all of these blocks

```ruby
private_lane :build_docs_shared do
  docs_path = Dir.mktmpdir do |docs_generation_folder|
    # do the stuff

    docs_generation_folder # returning path from block
  end

  docs_path # returning path from lane
end

lane :build_and_publish_docs do
    docs_generation_folder = build_docs_shared
    # do stuff
end
```

joshdholtz (changes requested):
Looks good! Just some improvement on some Ruby return values

fastlane/Fastfile
fire-at-will: Love this, done! :D

joshdholtz (approved):
This looks good! :shipit:

PR #4021, opened 2024-07-05
Return any cached CustomerInfo if we dont have an AppTransaction

MarkVillacampa:
To avoid making a request to /receipt without any transaction or app_transaction, return a cached customer info object if we don't have an app_transaction, even if the originalAppVersion and originalPurchaseDate are empty.

### Checklist

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -1129,6 +1129,13 @@ private extension PurchasesOrchestrator {
                         return
                     }
 
+                    // We dont have an AppTransactionJWS, return the cached CustomerInfo
+                    // even if it does not have originalPurchaseDate and originalApplicationVersion
+                    if let cachedCustomerInfo, appTransactionJWS == nil {
+                        completion?(.success(cachedCustomerInfo))
+                        return
+                    }
+
                     self.backend.post(receipt: .empty,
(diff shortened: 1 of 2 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1129,6 +1129,13 @@ private extension PurchasesOrchestrator {
                         return
                     }
 
+                    // We dont have an AppTransactionJWS, return the cached CustomerInfo
+                    // even if it does not have originalPurchaseDate and originalApplicationVersion
+                    if let cachedCustomerInfo, appTransactionJWS == nil {
+                        completion?(.success(cachedCustomerInfo))
+                        return
+                    }
+
```
aboedo, line 1138: this is kind of like a rough equivalent to the missing receipt error from SK1, maybe we should return an error instead?

aboedo: from slack convo: this should only happen in sandbox and when there are no purchases to restore in any case, so returning cached customer info would work from a dev perspective. We should add a log or something to inform the dev that the receipt wasn't actually posted

aboedo (approved):

aboedo:
had conversation about this offline, we want to retain the ability for the backend to get information about how often this happens and the patterns in production, so we want to remove the early return. 

Closing this for now.

PR #4025, opened 2024-07-05
Reduce CI jobs

aboedo:
Will reduce the number of CI jobs we're running so that we:

- run all jobs on release PRs
- have some trigger to manually run all non-essential jobs based on something (could be a comment, a tag, a webhook)
- then on regular commits run only essential jobs

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1198,28 +1198,20 @@ workflows:
       - run-test-ios-16:
           xcode_version: '14.3.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - run-test-ios-15:
           xcode_version: '14.3.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - run-test-ios-14:
           xcode_version: '14.2.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - run-test-ios-13:
           xcode_version: '14.2.0'
           install_swiftlint: false
       - run-test-macos:
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
 
   generate_revenuecatui_snapshots:
     when: << pipeline.parameters.generate_revenuecatui_snapshots >>
@@ -1227,15 +1219,11 @@ workflows:
       - spm-revenuecat-ui-ios-15:
           xcode_version: '14.3.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - spm-revenuecat-ui-ios-16:
           xcode_version: '14.3.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - spm-revenuecat-ui-ios-17:
           xcode_version: "15.2"
 
@@ -1251,128 +1239,76 @@ workflows:
       - docs-build:
           xcode_version: '14.3.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - spm-release-build:
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - spm-custom-entitlement-computation-build:
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - spm-receipt-parser:
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - spm-release-build-xcode-14:
           xcode_version: '14.3.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - spm-release-build-xcode-15:
           xcode_version: '15.2'
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - spm-xcode-14-1:
           xcode_version: '14.1.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - spm-revenuecat-ui-ios-15:
           xcode_version: '14.3.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - spm-revenuecat-ui-ios-16:
           xcode_version: '14.3.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - spm-revenuecat-ui-ios-17:
           xcode_version: "15.2"
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - spm-revenuecat-ui-watchos:
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - run-test-macos:
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - run-test-ios-17:
           xcode_version: "15.2"
       - run-test-ios-16:
           xcode_version: '14.3.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - run-test-ios-15:
           xcode_version: '14.3.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - run-test-watchos:
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - run-test-tvos:
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       # Pre-iOS 15 requires macOS 12 which requires Xcode 14.2
       # See https://circleci.com/docs/using-macos/#supported-xcode-versions
       - run-test-ios-14:
           xcode_version: '14.2.0'
           install_swiftlint: false
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - run-test-ios-13:
           xcode_version: '14.2.0'
           install_swiftlint: false
       - build-tv-watch-and-macos:
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - build-visionos:
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - backend-integration-tests-SK1:
-          filters:
-            branches:
-              <<: *release-branches
-              ignore: *forked-pull-requests
+          <<: *release-branches
       - backend-integration-tests-SK2:
-          filters:
-            branches:
-              <<: *release-branches
-              ignore: *forked-pull-requests
+          <<: *release-branches
       - backend-integration-tests-other:
-          filters:
-            branches:
-              <<: *release-branches
-              ignore: *forked-pull-requests
+          <<: *release-branches
       - backend-integration-tests-offline:
-          filters:
-            branches:
-              <<: *release-branches
+          <<: *release-branches
       - backend-integration-tests-custom-entitlements:
-          filters:
-            branches:
-              <<: *release-branches
-              ignore: *forked-pull-requests
+          <<: *release-branches
 
   deploy:
     when:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,6 +69,12 @@ aliases:
       branches:
         only: /^release\/.*\.0$/
 
+  # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
+  forked-pull-requests: &forked-pull-requests
+    filters:
+      tags:
+        ignore: /pull\/[0-9]+/
+
 commands:
   install-runtime:
     parameters:
@@ -1192,16 +1198,28 @@ workflows:
       - run-test-ios-16:
           xcode_version: '14.3.0'
           install_swiftlint: false
+          filters:
+            branches:
+              <<: *release-branches
       - run-test-ios-15:
           xcode_version: '14.3.0'
           install_swiftlint: false
+          filters:
+            branches:
+              <<: *release-branches
       - run-test-ios-14:
           xcode_version: '14.2.0'
           install_swiftlint: false
+          filters:
+            branches:
+              <<: *release-branches
       - run-test-ios-13:
           xcode_version: '14.2.0'
           install_swiftlint: false
-      - run-test-macos
+      - run-test-macos:
+          filters:
+            branches:
+              <<: *release-branches
 
   generate_revenuecatui_snapshots:
     when: << pipeline.parameters.generate_revenuecatui_snapshots >>
@@ -1209,9 +1227,15 @@ workflows:
       - spm-revenuecat-ui-ios-15:
           xcode_version: '14.3.0'
           install_swiftlint: false
+          filters:
+            branches:
+              <<: *release-branches
       - spm-revenuecat-ui-ios-16:
           xcode_version: '14.3.0'
           install_swiftlint: false
+          filters:
+            branches:
+              <<: *release-branches
       - spm-revenuecat-ui-ios-17:
           xcode_version: "15.2"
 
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -69,6 +69,12 @@ aliases:
       branches:
         only: /^release\/.*\.0$/
 
+  # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
```
aboedo, line 72: I'm not currently using this because the other filter covers it too so it'd be redundant, but this feels like knowledge that is worth preserving in case it's useful in the future.

.circleci/config.yml
```diff
@@ -1385,3 +1407,30 @@ workflows:
       equal: [bump, << pipeline.parameters.action >>]
     jobs:
       - release-train
+
+  manual-test-workflow:
+    when:
+      equal: ["run-manual-tests", << pipeline.parameters.action >>]
```
aboedo, line 1415: to trigger, just go into circleCI, select a branch and click trigger pipeline, then enter the value: 
image

aboedo: we can get fancier by adding a github actions hook from a comment or tag that triggers this programmatically. Not sure it's worth it though since this isn't exactly hard to do. Thoughts?

aboedo: FWIW I did test triggering this manually and it works

aboedo:
another obvious benefit: much faster testing time for regular commits

image

.circleci/config.yml
aboedo: it'd also be quite easy to go beyond this and separate out backend integration tests, tests for iOS versions, etc in case we ever wanted to specifically run those, but it feels like this gets us 90% of the way there with 10% of the work

joshdholtz (approved):
🚢 💪

.circleci/config.yml
tonidero: Hmm I thought we weren't running tests on forks... If we are, we should be careful with the usage.

.circleci/config.yml
```diff
@@ -1192,26 +1198,33 @@ workflows:
       - run-test-ios-16:
           xcode_version: '14.3.0'
           install_swiftlint: false
+          <<: *release-branches
```
tonidero, line 1201: Hmm maybe I'm wrong but the `generate-snapshot` and `geenerate_revenuecatui_snapshots` are not run on every commit but on demand, in order to update the backend requests snapshots and UI snapshots respectively.

I think we shouldn't filter these to only run on release jobs in case we want to update the test snapshots in a normal PR? (Do lmk if I missed something)

tonidero (approved):
LGTM! Left a comment over Slack but I think we can ship this

vegaro (approved):

.circleci/config.yml
aboedo: what do we currently do for `bump` on tags? This mimics the way that the bump job was set up

aboedo: why would you need to run these tests on a tag, though? wouldn't we usually run tests _before_ tagging?

.circleci/config.yml
```diff
@@ -1223,75 +1236,84 @@ workflows:
         - not: << pipeline.parameters.generate_snapshots >>
         - not: << pipeline.parameters.generate_revenuecatui_snapshots >>
     jobs:
+      # mandatory jobs
```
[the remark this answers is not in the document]
aboedo: yeah, that's waaaay better, I'll do that

.circleci/config.yml
```diff
@@ -1385,3 +1410,32 @@ workflows:
       equal: [bump, << pipeline.parameters.action >>]
     jobs:
       - release-train
+
+  # to trigger tests manually, log into circleCI, select the project, a branch, and then click "Trigger Pipeline"
+  # and add a parameter called "run-manual-tests"
+  manual-test-workflow:
+    when:
+      equal: ["run-manual-tests", << pipeline.parameters.action >>]
```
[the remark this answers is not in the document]
aboedo: definitely, on it

.circleci/config.yml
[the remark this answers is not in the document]
aboedo: still cleaning up, but I split it into two workflows and it's much nicer

.circleci/config.yml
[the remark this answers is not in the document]
aboedo: yeah, splitting workflows and mixing in the rules is a loooot better, thanks!

.circleci/config.yml
aboedo: you're absolutely right, that was a pretty terrible oversight from me

.circleci/config.yml
aboedo: removing

.circleci/config.yml
```diff
@@ -389,6 +389,14 @@ commands:
 jobs:
   spm-release-build-xcode-14:
     <<: *base-job
+    parameters:
+      xcode_version:
+        type: string
+        default: '14.3.0'
+      install_swiftlint:
+        type: boolean
+        default: false
```
aboedo, line 398: we were repeating the code versions every time we called the jobs. This could make sense for some jobs if you want to call it again with a different xcode version, but I couldn't find any where we seemed to be doing that intentionally. 

Also some of them (like this one) are even named after their xcode version, so it makes no sense to have it be an option. 

We could remove jobs that are exactly the same but only have different xcode versions and use the param instead, but that's for another day. Right now I'm trying to preserve logic as much as possible. 

So I just moved the xcode version to the jobs themselves.

.circleci/config.yml
```diff
@@ -418,6 +439,15 @@ jobs:
           command: swift build -c release --target RevenueCatUI
           no_output_timeout: 30m
 
+  pod-lib-lint:
```
aboedo, line 442: new job, just does pod lib lint, but that in turn does a bunch of checks for compilation in different platforms and OS versions, so I've found it to be super useful in the past at catching missing availability checks

.circleci/config.yml
```diff
@@ -1177,33 +1286,19 @@ workflows:
...
-          xcode_version: '14.2.0'
-          install_swiftlint: false
-      - run-test-ios-13:
-          xcode_version: '14.2.0'
-          install_swiftlint: false
+      - run-test-ios-17
+      - run-test-ios-16
+      - run-test-ios-15
+      - run-test-ios-14
+      - run-test-ios-13
```
aboedo, line 1293: this is the result of moving the xcode versions up, the workflows are now very clean

.circleci/config.yml
```diff
@@ -1214,74 +1309,8 @@ workflows:
         - not: << pipeline.parameters.generate_revenuecatui_snapshots >>
     jobs:
       - lint
```
aboedo, line 1311: build-test job now only does lint, unit tests and pod lib lint

.circleci/config.yml
```diff
@@ -1375,3 +1406,50 @@ workflows:
       equal: [bump, << pipeline.parameters.action >>]
     jobs:
       - release-train
+
+  # to trigger tests manually, log into circleCI, select the project, a branch, and then click "Trigger Pipeline"
+  # and add a parameter called "run-manual-tests"
```
aboedo, line 1411: will add a github comment trigger for this as a follow-up PR

.circleci/config.yml
```diff
@@ -1375,3 +1406,50 @@ workflows:
...
+            equal: [scheduled_pipeline, << pipeline.trigger_source >>]
+        - not: << pipeline.parameters.generate_snapshots >>
+        - not: << pipeline.parameters.generate_revenuecatui_snapshots >>
+        - or:
+            - matches: 
+                pattern: "^release/.*$"
+                value: << pipeline.git.branch >>
+            - equal:
+                - "manual_tests"
+                - << pipeline.trigger_source >>
```
aboedo, line 1425: doesn't trigger for scheduled pipelines or snapshot pipelines, and will be triggered by release branches and manual trigger

.circleci/config.yml
```diff
@@ -1375,3 +1406,50 @@ workflows:
...
+      - spm-revenuecat-ui-ios-16
+      - spm-revenuecat-ui-ios-17
+      - spm-revenuecat-ui-watchos
+      - pod-lib-lint
+      - installation-tests-cocoapods
+      - installation-tests-swift-package-manager
+      - installation-tests-custom-entitlement-computation-swift-package-manager
+      - installation-tests-carthage
+      - installation-tests-xcode-direct-integration
+      - installation-tests-receipt-parser
```
aboedo, line 1455: these weren't included here but they were included in the deploy scripts, so you could have had that fail when releasing which is super annoying and waaay too late to find out that something's not working

fastlane/Fastfile
```diff
@@ -241,7 +241,6 @@ platform :ios do
 
   desc "build tvOS, watchOS, macOS"
   lane :build_tv_watch_mac do |options|
-    check_pods
```
aboedo, line 244: since we have an explicit job for this now, we don't need to run it again here

fastlane/Fastfile
```diff
@@ -250,7 +249,6 @@ platform :ios do
 
   desc "macOS build"
   lane :build_mac do |options|
-    check_pods
```
aboedo, line 253: and in turn these tests go faster and don't re-test the same thing independently

aboedo:
I'm going to try running the rest of the jobs manually to ensure that the system is working as expected, but at least regular jobs are running just fine

aboedo:
ok, all jobs other than docs-build work correctly and @fire-at-will mentioned earlier today that that one was broken, so I imagine it's not related to the changes here. 

I haven't gotten the branch filters to work well yet, though, so I'll figure that out and then ask for re-review

fire-at-will:
@aboedo I'm seeing `docs-build` passing in my PRs right now, it's docs-deploy that failed in the `5.1.0` release pipeline yesterday

aboedo:
@fire-at-will thanks for the heads up! I'll check it out then

aboedo:
@fire-at-will I had the wrong xcode version on it! I'm glad you brought it up. I had to go one by one on the jobs and copy the xcode version, I guess I had a copy/paste issue on that one

aboedo:
Okay, I think this is good to re-review now. A lot has changed since the original approach, but I did test that: 
- the jobs will correctly trigger from a release PR (see #4050) 
- the jobs will correctly trigger from a manual action in circleci
- they run correctly once they do trigger, meaning that the refactors for xcode versions are correct

.circleci/config.yml
```diff
@@ -1214,109 +1309,26 @@ workflows:
...
-          <<: *release-branches
-      - installation-tests-custom-entitlement-computation-swift-package-manager:
-          <<: *release-branches
-      - installation-tests-receipt-parser:
-          <<: *release-branches
-      - installation-tests-carthage:
-          <<: *release-branches
-      - installation-tests-xcode-direct-integration:
-          <<: *release-branches
+      - release-checks
```
aboedo, line 1324: All of these will run in release branches anyway from the all-tests job

aboedo: plus quite a few more

.circleci/config.yml
```diff
@@ -1214,109 +1309,26 @@ workflows:
...
           type: approval
           requires:
             - release-checks
-            - installation-tests-cocoapods
-            - installation-tests-swift-package-manager
-            - installation-tests-custom-entitlement-computation-swift-package-manager
-            - installation-tests-carthage
-            - installation-tests-xcode-direct-integration
-            - installation-tests-receipt-parser
-          <<: *release-branches
```
aboedo, line 1315: Deploy now only requires `release-checks`, but all of the tests will trigger anyway. 
I debated with myself on whether to require all the jobs anyway, but given how flaky they've been, I think it might not be a bad idea to let us approve and deploy even if a test isn't passing, and leave it up to reviewer discretion. 

Before we were only requiring installation tests to pass, so it's not _that_ different now

.circleci/config.yml
```diff
@@ -1214,109 +1309,26 @@ workflows:
...
       - tag-release-branch:
           requires:
             - hold
-          <<: *release-branches
```
aboedo, line 1319: redundant because of the when filter

.circleci/config.yml
```diff
@@ -1334,24 +1346,26 @@ workflows:
           <<: *release-tags
       - docs-deploy:
           <<: *release-tags
-          xcode_version: '14.3.0'
```
aboedo, line 1337: redundant, it's now in the job definition

Tests/TestingApps/PaywallsTester/PaywallsTester/Config/LocalConfigItems.swift
```diff
@@ -23,7 +23,7 @@ import Foundation
 // Then add local keys:
 // extension ConfigItem {
 //     static var apiKey: String {
-//         "appl_FOObar"
+//         "REVENUECAT_API_KEY"
```
aboedo, line 26: the changes from the #4026 PR ported here

.circleci/config.yml
```diff
@@ -1375,3 +1389,50 @@ workflows:
...
+
+  # to trigger tests manually, log into circleCI, select the project, a branch, and then click "Trigger Pipeline"
+  # and add a parameter called "run-manual-tests"
+  all-tests:
+    when:
+      and:
+        - not:
+            equal: [scheduled_pipeline, << pipeline.trigger_source >>]
+        - not: << pipeline.parameters.generate_snapshots >>
+        - not: << pipeline.parameters.generate_revenuecatui_snapshots >>
```
tonidero, line 1401: I was wondering, do we need the above conditions? Or would the ones in the `or` be enough?

.circleci/config.yml
tonidero: Not sure how you were planning to implement that but I was thinking maybe we could have something in our CircleCI orb that allows us to establish certain comment to workflow triggers, that way we can reuse it in the other repos in the future as needed. Not needed for this PR of course

tonidero (approved):
Love how clean it is now!

.circleci/config.yml
```diff
@@ -517,6 +563,14 @@ jobs:
 
   spm-revenuecat-ui-ios-17:
     <<: *base-job
+    parameters:
+      xcode_version:
+        type: string
+        default: '15.2'
```
vegaro, line 569: nit but we have this as `15.2.0` in line 652

.circleci/config.yml
```diff
@@ -853,6 +948,13 @@ jobs:
 
   docs-build:
     <<: *base-job
+    parameters:
+      xcode_version:
+        type: string
+        default: '14.3'
```
vegaro, line 954: Same here, we have 14.3.0 in other places

.circleci/config.yml
```diff
@@ -1375,3 +1389,50 @@ workflows:
...
+      - spm-custom-entitlement-computation-build
+      - spm-receipt-parser
+      - spm-release-build
+      - spm-release-build-xcode-14
+      - spm-release-build-xcode-15
+      - spm-revenuecat-ui-ios-15
+      - spm-revenuecat-ui-ios-16
+      - spm-revenuecat-ui-ios-17
+      - spm-revenuecat-ui-watchos
+      - pod-lib-lint
```
vegaro, line 1432: this one is ran already in `build-test`, I see `run-test-ios-17` is not part of these jobs, so I figured you only wanted to add "the other" jobs here

.circleci/config.yml
aboedo: yeah I noticed it but didn't want to modify the version numbers since they weren't from this PR, but maybe I just got lazy. updating

.circleci/config.yml
aboedo: good call

.circleci/config.yml
aboedo: good point, I didn't want to mess with existing conditions but this would never happen if it's a separate workflow

.circleci/config.yml
aboedo: great catch

aboedo:
> Not sure how you were planning to implement that but I was thinking maybe we could have something in our CircleCI orb that allows us to establish certain comment to workflow triggers, that way we can reuse it in the other repos in the future as needed. Not needed for this PR of course

Yeah, that'd be great. I've kinda struggled with it a bit because it's hard to test and it gets triggered directly from the main branch, but I'd be happy to update it it afterwards and make it generic to all repos if we find a good use for it @tonidero

aboedo:
@tonidero @vegaro @joshdholtz comments addressed! Thanks for the thorough reviews!

PR #4027, opened 2024-07-05
split ci files

aboedo:
CircleCI [doesn't have direct support for splitting a config file into multiple](https://support.circleci.com/hc/en-us/articles/360056463852-Can-I-split-a-config-into-multiple-files), so this adds support for it a bit more manually: 

It adds logic to our pre-commit hook to concatenate files in a configuration folder. 
This only runs if files in that folder have changed. 

This is the basic file structure: 
image

So all of the files in that folder, in order, comprise the config.yml file that looks pretty much identical to the one we had before. 

This approach does take two assumptions: 
1) developers will understand that they shouldn't edit the config.yml file directly (there's a warning in the file itself, and we could get fancier and detect this with git I suppose?)
2) developers will use the pre-commit hook (which... we _should_ in any case, there's a reason we have it)


aboedo:
@joshdholtz tagged you for early thoughts since you had had a stab at this in the past

.circleci/config.yml
```diff
@@ -1439,3 +1445,6 @@ workflows:
       - spm-revenuecat-ui-ios-16
       - spm-revenuecat-ui-ios-17
       - spm-revenuecat-ui-watchos
+
+
+
```
aboedo, line 1450: lol guess I had a ton of line breaks at the end of the other files. 

But the idea is for this one to get generated from the others

.circleci/configuration/config.yml
```diff
@@ -0,0 +1,1194 @@
+orbs:
```
aboedo, line 1: this is an exact copy of the other config file minus the workflows part

.circleci/configuration/workflows.yml
```diff
@@ -0,0 +1,252 @@
+workflows:
```
aboedo, line 1: and here is just the workflows part

scripts/pre-commit.sh
```diff
@@ -82,3 +82,34 @@ if [ $SHOULD_FAIL_PRECOMMIT -ne 0 ]; then
 else
   verify_no_included_apikeys
 fi
+
+
+# Function to concatenate CircleCI configuration files
```
aboedo, line 87: I'm generally allergic to bash, I might study porting all the logic into ruby at some point

.circleci/configuration/aliases.yml
```diff
@@ -0,0 +1,56 @@
+aliases:
```
aboedo, line 1: this is just cut/paste from what used to be only in config.yml

scripts/pre-commit.sh
```diff
@@ -82,3 +82,50 @@ if [ $SHOULD_FAIL_PRECOMMIT -ne 0 ]; then
 else
   verify_no_included_apikeys
 fi
+
+# CircleCI doesn't let us split a config file into multiple, so we have a pre-commit hook that does it for us
+# so that the file doesn't become unmanageable. 
+concatenate_circleci_configs() {
```
aboedo, line 88: this is the only file with real changes aside from the original config.yml changes, so those are the only two ones worth really reviewing

aboedo:
the diff makes this look big but there really are only 2 small files worth reviewing in here, the rest are mostly ignorable in they're cut / paste from the original config.yml file

.circleci/config.yml
```diff
@@ -1,11 +1,20 @@
+# This file is generated automatically by concatenating the files in the .circleci/configuration/ directory.
+# DO NOT EDIT this file directly. Instead, edit the files in the .circleci/configuration/ directory and rely on the pre-commit hook to update this file.
```
tonidero, line 2: I was wondering if we should add a test job to make sure this file matches the output from the individual configuration files, to make sure it's not outdated/it's changed accidentally.

scripts/pre-commit.sh
tonidero, line 88: Might be worth extracting this to a separate script and just call it from the `pre-commit.sh`? Specially if we follow my other suggestion about adding a test job to make sure the combined file matches the individual files.

JayShortway, line 88: Could we use a CircleCI setup job to generate the final CircleCI config file instead? This would avoid the need for a pre-commit hook, and keep the CircleCI-related logic self contained (and running on CI instead of locally). 

[Here's an example for exactly our use case.](https://circleci.com/docs/using-dynamic-configuration/#a-basic-example)

>In this example, a script is used to generate a YAML configuration file. The continuation orb is used to continue the pipeline using the generated configuration.

JayShortway (commented):
Major improvement! Had 1 suggestion.

PR #4023, opened 2024-07-05
Add contact support button

vegaro:
### Checklist
- [ ] If applicable, unit tests
- [ ] If applicable, create follow-up issues for `purchases-android` and hybrids

PR #4028, opened 2024-07-06
[Customer Center] Create a completion handler

vegaro:
### Checklist


vegaro:
> [!WARNING]
> <b>This pull request is not mergeable via GitHub because a downstack PR is open. Once all requirements are satisfied, merge this PR as a stack <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/4028?utm_source=stack-comment-downstack-mergeability-warning" >on Graphite</a>.</b>

* **#4028** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/4028?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* **#4023** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/4023?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3983** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3983?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3968** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3968?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3959** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3959?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3967** <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3967?utm_source=stack-comment-icon" target="_blank">Graphite</a>: 1 other dependent PR ([#3919](https://github.com/RevenueCat/purchases-ios/pull/3919) <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/3919?utm_source=stack-comment-icon" target="_blank">Graphite</a>)
* `main`

<h2></h2>

RevenueCatUI/CustomerCenter/Views/CustomerCenterCompletionHandler.swift
```diff
@@ -0,0 +1,57 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  CustomerCenterCompletionHandler.swift
+//
+//  Created by Cesar de la Vega on 6/7/24.
+
+import Foundation
+import SwiftUI
+
+final class CustomerCenterCompletionHandler: ObservableObject {
```
tonidero, line 17: I've been wondering whether it would make more sense to call this `CustomerCenterActionHandler`. Then, we would notify whenever an action is performed.

My main concern with this approach is that it moves away from what we have in paywalls, where we have individual callbacks for many actions like purchase/restore started/failed/completed... But I feel since there are going to be more actions here, it makes more sense to do something like that. Wdyt?

RevenueCatUI/CustomerCenter/Views/CustomerCenterCompletionHandler.swift
```diff
@@ -0,0 +1,57 @@
...
+        return .init()
+    }
+
+    func supportContacted() {
+        self.customerCenterResult = CustomerCenterResult(status: .contactSupport)
+    }
+
+}
+
+struct CustomerCenterResult: Equatable {
```
tonidero, line 32: If we follow my previous suggestion, I would move away from using the `Result` naming, since this may happen at any point, and even multiple times. I would only keep this naming if we want to return a single result after the customer center closes, but I think I prefer my suggested option. Lmk what you think!

RevenueCatUI/CustomerCenter/Views/CustomerCenterCompletionHandler.swift
vegaro: That's a good point... Right, I think at some point we will have actions that are no completion actions like `feedbackSurveyStarted` so we want this to be an action handler. I honestly prefer just having this handler with a bunch of different events that can be handled in a switch statement than a lot of functions like we do in paywalls

It is true that adding a new event to a sealed class is technically a breaking change, but I think it's assumable. It is also true that it's not possible to modify an existing case in an enum (and add more properties) without creating a new event, but that's also an issue with independent functions.

In summary I find it cleaner to use sealed class when we are going to have a lot of potential events.

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -26,13 +26,17 @@ import SwiftUI
 @available(visionOS, unavailable)
 public struct CustomerCenterView: View {
 
-    @StateObject private var viewModel = CustomerCenterViewModel()
+    @StateObject
+    private var viewModel = CustomerCenterViewModel()
+    @StateObject
+    private var completionHandler: CustomerCenterCompletionHandler = .default()
```
tonidero, line 32: Would it be possible to move this inside the view model? That way we centralize the logic there.

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -200,6 +209,7 @@ struct ManageSubscriptionsButtonsView: View {
                 Task {
                     openURL(URLUtilities.createMailURL()!)
                 }
+                completionHandler.supportContacted()
```
tonidero, line 212: Ideally this calls the view model, which internally calls whatever it needs I think.

PR #4033, opened 2024-07-09
Multi-tier paywall - offer overrides

joshdholtz:
### Motivation

Allow setting a custom name, details, and badge per package 👇 

Screenshot 2024-07-09 at 8 05 35 PM

### Description

- `ProcessedLocalizedConfiguration` now handles all the badge logic
    - Badge content will first look at offer overrides (which will process variables)
    - Will default to relative discount if there is one (this will work for existing templates)


RevenueCatUI/Data/ProcessedLocalizedConfiguration.swift
```diff
@@ -34,18 +35,38 @@ struct ProcessedLocalizedConfiguration: PaywallLocalizedConfiguration {
...
+        let offerDetails = offerOverrides?.offerDetails ?? configuration.offerDetails
+        let offerDetailsWithIntroOffer = offerOverrides?.offerDetailsWithIntroOffer
+            ?? configuration.offerDetailsWithIntroOffer
+        let offerName = offerOverrides?.offerName ?? configuration.offerName
+
+        let offerBadge: String?
+        if let offerOverrides {
+            offerBadge = offerOverrides.offerBadge
+        } else if let discount = context.discountRelativeToMostExpensivePerMonth {
+            offerBadge = dataProvider.localizedRelativeDiscount(discount, locale)
```
joshdholtz, line 50: This defaults the badge to the relative discount if not using offer overrides

RevenueCatUI/Templates/Template4View.swift
```diff
@@ -365,7 +365,7 @@ private struct PackageButton: View {
         }
     }
 
-    private func discountOverlay(_ discount: Double) -> some View {
+    private func discountOverlay(_ badge: String) -> some View {
```
vegaro, line 368: should this be badgeOverlay now?

RevenueCatUI/Templates/Template4View.swift
```diff
@@ -374,7 +374,7 @@ private struct PackageButton: View {
                     : self.configuration.colors.accent2Color
                 )
 
-            Text(Localization.localized(discount: discount, locale: self.locale))
+            Text(badge)
                 .textCase(.uppercase)
                 .foregroundColor(
                     self.selected
```
vegaro, line 380: <i class='graphite__hidden'>[Re: line 389]<br/><br/></i>

Same here, maybe this should be `discountLabelHeight`<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/RevenueCat/purchases-ios/4033?utm_source=unchanged-line-comment">Graphite</a>.</span>

Sources/Paywalls/PaywallData.swift
```diff
@@ -71,7 +71,6 @@ public protocol PaywallLocalizedConfiguration {
     var features: [PaywallData.LocalizedConfiguration.Feature] { get }
     /// An optional name representing the ``PaywallData/Tier``.
     var tierName: String? { get }
-
```
vegaro, line 74: nitpick but I think we still want this line according to the Swift style

RevenueCatUI/Data/TestData.swift
```diff
@@ -572,6 +572,20 @@ internal enum TestData {
...
+                    offerOverrides: [
+                        TestData.threeMonthPackage.identifier: .init(
+                            offerDetails: "Details",
+                            offerDetailsWithIntroOffer: "",
+                            offerName: "OVERRIDE Three Month",
+                            offerBadge: ""
+                        ),
+                        TestData.lifetimePackage.identifier: .init(
+                            offerDetails: "Details",
+                            offerDetailsWithIntroOffer: "",
```
vegaro, line 584: I think this is nullable, should we pass nil instead of ""?

vegaro (approved):

RevenueCatUI/Templates/Template4View.swift
vegaro: Same for line 389, maybe it should be `badgeLabelHeight`

RevenueCatUI/Templates/Template4View.swift
```diff
@@ -284,10 +284,10 @@ private struct PackageButton: View {
                     .frame(maxHeight: .infinity)
             }
             .background { // Discount overlay
-                if let discount = self.package.discountRelativeToMostExpensivePerMonth {
-                    self.discountOverlay(discount)
+                if let badge = self.package.localization.offerBadge, !badge.isEmpty {
```
fire-at-will, line 287: This is unrelated to the changes in this PR, but I think this would be a lot more readable if we only used one `.background` modifier and then rendered the entire background in one subview extracted out to a local `@ViewBuilder` function

fire-at-will (approved):
Everything here looks good to me!

PR #4035, opened 2024-07-10
Prevent paywall PurchaseHandler from being cleared on rerender

### LINKED ISSUE #4034: paywallFooter handlers like purchaseCompleted or purchaseCancelled are not called with the release 5.0.0
Hello,

I just updated my app to use the release 5.0.0 but I notice that the different paywallFooter handlers are never called: `purchaseCompleted`, `purchaseCancelled`, etc. This breaks my app because I have additional logic in these handlers.

It's working fine with the release 4.43.2. I rolled back my App Store version to use this release while this bug is fixed.

Are you aware of this issue?
Thanks,
Axel

joshdholtz:
### Motivation

Fixes #4034

`PurchaseHandler` was being cleared/reset when a paywall was being rendered (ex: if an overlay was applied to it)

### Description

- Restores `PurchaseHandler` back to being a `@StateObject` from #3187
- `checkForConfigurationConsistency()` has a few changes
  - Takes a `PurchaseHandler` as a parameter so it doesn't try to access the `StateObject`
      - This was fixes the purple Xcode warning of being access wrongly by it being a `StateObject` again
  - Removes `mutating` keyword and returns an `NSError?` which it now sets in the constructor

#### Using RevenueCat

https://github.com/RevenueCat/purchases-ios/assets/401294/7e6430ca-6243-49ad-b7c0-8a90d90894f8

#### Using MyApp

https://github.com/RevenueCat/purchases-ios/assets/401294/829f12c3-a11a-4952-883c-7ae7addd9c1d

### DIFF
--- a/RevenueCatUI/Data/PaywallViewConfiguration.swift
+++ b/RevenueCatUI/Data/PaywallViewConfiguration.swift
@@ -20,7 +20,7 @@ struct PaywallViewConfiguration {
     var fonts: PaywallFontProvider
     var displayCloseButton: Bool
     var introEligibility: TrialOrIntroEligibilityChecker?
-    var purchaseHandler: PurchaseHandler
+    var purchaseHandler: PurchaseHandler?
 
     init(
         content: Content,
@@ -29,7 +29,7 @@ struct PaywallViewConfiguration {
         fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
         displayCloseButton: Bool = false,
         introEligibility: TrialOrIntroEligibilityChecker? = nil,
-        purchaseHandler: PurchaseHandler
+        purchaseHandler: PurchaseHandler? = nil
     ) {
         self.content = content
         self.customerInfo = customerInfo
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -103,7 +103,7 @@ public struct PaywallView: View {
     }
 
     // @PublicForExternalTesting
-    init(configuration: PaywallViewConfiguration, paywallViewOwnsPurchaseHandler: Bool = true) {
+    init(configuration: PaywallViewConfiguration) {
         let purchaseHandler = configuration.purchaseHandler ?? .default()
         self._purchaseHandler = .init(wrappedValue: purchaseHandler)
 
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -121,17 +121,17 @@ public struct PaywallView: View {
         self.fonts = configuration.fonts
         self.displayCloseButton = configuration.displayCloseButton
 
-        self.initializationError = checkForConfigurationConsistency(purchaseHandler: purchaseHandler)
+        self.initializationError = Self.checkForConfigurationConsistency(purchaseHandler: purchaseHandler)
     }
 
-    private func checkForConfigurationConsistency(purchaseHandler: PurchaseHandler) -> NSError? {
+    private static func checkForConfigurationConsistency(purchaseHandler: PurchaseHandler) -> NSError? {
         switch purchaseHandler.purchasesAreCompletedBy {
         case .myApp:
-            if self.purchaseHandler.performPurchase == nil || self.purchaseHandler.performRestore == nil {
+            if purchaseHandler.performPurchase == nil || purchaseHandler.performRestore == nil {
                 let missingBlocks: String
-                if self.purchaseHandler.performPurchase == nil && self.purchaseHandler.performRestore == nil {
+                if purchaseHandler.performPurchase == nil && purchaseHandler.performRestore == nil {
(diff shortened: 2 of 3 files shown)


RevenueCatUI/PaywallView.swift
```diff
@@ -111,15 +104,8 @@ public struct PaywallView: View {
 
     // @PublicForExternalTesting
     init(configuration: PaywallViewConfiguration, paywallViewOwnsPurchaseHandler: Bool = true) {
```
vegaro, line 106: I think this `paywallViewOwnsPurchaseHandler` can be removed since it's not being checked anymore?

RevenueCatUI/Data/PaywallViewConfiguration.swift
```diff
@@ -29,7 +29,7 @@ struct PaywallViewConfiguration {
         fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
         displayCloseButton: Bool = false,
         introEligibility: TrialOrIntroEligibilityChecker? = nil,
-        purchaseHandler: PurchaseHandler
+        purchaseHandler: PurchaseHandler? = nil
```
vegaro, line 32: Instead of this being nullable can we initialize it to `default`? Also, why is it nullable now? We are not passing nil anywhere and this is internal

RevenueCatUI/PaywallView.swift
```diff
@@ -111,15 +104,8 @@ public struct PaywallView: View {
...
-        if paywallViewOwnsPurchaseHandler {
-            self._doNotAccessPurchaseHandler = .init(wrappedValue: configuration.purchaseHandler)
-        } else {
-            // this is unused and is only present to fulfill the need to have an object assigned
-            // to a @StateObject
-            self._doNotAccessPurchaseHandler = .init(wrappedValue: PurchaseHandler.default())
-        }
-
-        self.purchaseHandler = configuration.purchaseHandler
```
vegaro, line 122: was it being cleared here?

RevenueCatUI/PaywallView.swift
```diff
@@ -111,15 +104,8 @@ public struct PaywallView: View {
 
     // @PublicForExternalTesting
     init(configuration: PaywallViewConfiguration, paywallViewOwnsPurchaseHandler: Bool = true) {
-        if paywallViewOwnsPurchaseHandler {
-            self._doNotAccessPurchaseHandler = .init(wrappedValue: configuration.purchaseHandler)
```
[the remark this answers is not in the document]
joshdholtz: @vegaro I _think_ this was here to hold on to something a s reference at an attempt for the error that this PR is fixing... but didn't work as intended 🤔 

The `mutating` function that was being called from the constructor isn't allowed to access the `StateObject` so we were attempting to access the normal property.

However, the normal property was being reset on rerender where the `StateObject` was keeping the _actual_ `PurchaseHandler` callbacks that we needed 🤷‍♂️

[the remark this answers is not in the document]
vegaro: Oh I see now that  `purchaseHandler` is an `ObservedObject`, I was reading it as both being `StateObject` so I didn't understandthe difference. Thanks for the explanation!!

RevenueCatUI/PaywallView.swift
joshdholtz: Yeah yeah, this is where it was being cleared unintentionally during a rerender 😬

RevenueCatUI/Data/PaywallViewConfiguration.swift
joshdholtz: This is reverted back into the state that it was originally 🤷‍♂️  It does kind of seem like it should be `.default` but I should play with it a bit more before making that change to figure out why it wasn't set as this

RevenueCatUI/PaywallView.swift
```diff
@@ -135,11 +121,11 @@ public struct PaywallView: View {
...
-        checkForConfigurationConsistency()
+        self.initializationError = checkForConfigurationConsistency(purchaseHandler: purchaseHandler)
     }
 
-    private mutating func checkForConfigurationConsistency() {
-        switch self.purchaseHandler.purchasesAreCompletedBy {
+    private func checkForConfigurationConsistency(purchaseHandler: PurchaseHandler) -> NSError? {
+        switch purchaseHandler.purchasesAreCompletedBy {
         case .myApp:
             if self.purchaseHandler.performPurchase == nil || self.purchaseHandler.performRestore == nil {
```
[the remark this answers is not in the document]
joshdholtz: They do! I removed the `self.` but... maybe I didn't commit it 😱 I'm going to just rename the variable so it doesn't accidentally compile 😅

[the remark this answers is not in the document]
vegaro: Yes I was going to suggest that lol

aboedo (approved):

vegaro (approved):

PR #4046, opened 2024-07-12
[Customer Center] Select the right subscription to show

JayShortway:
## In this PR
- Refactored the test-fixture-building code in `ManageSubscriptionsViewModelTests`, to more easily fake scenarios of having multiple subscriptions. 
    - Moved this code from the `ManageSubscriptionsViewModelTests` extension to a separate `Fixtures` class to avoid errors and warnings for calling static methods from outside its actor context. 
- Translated the subscription selection requirements to tests in `ManageSubscriptionsViewModelTests`.
- Updated the logic in `ManageSubscriptionsViewModel` to pick the right subscription based on the requirements specified in the tests. 

## Not in this PR
- Handling of lifetime subscriptions (let's discuss in the sync).
- Any UI changes (will do a follow-up PR).

## Request
Please let me know any conventions I broke, and any tips you have for more idiomatic Swift! :pray:


Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
```diff
@@ -71,27 +71,187 @@ class ManageSubscriptionsViewModelTests: TestCase {
...
 
         expect(viewModel.subscriptionInformation?.title) == "title"
         expect(viewModel.subscriptionInformation?.durationTitle) == "month"
         expect(viewModel.subscriptionInformation?.price) == "$2.99"
-        expect(viewModel.subscriptionInformation?.nextRenewalString) == "Apr 12, 2062"
-        expect(viewModel.subscriptionInformation?.productIdentifier) == "com.revenuecat.product"
+        
+        let formatter = DateFormatter()
+        formatter.dateStyle = .medium
+        expect(viewModel.subscriptionInformation?.nextRenewalString) == formatter.string(from: ISO8601DateFormatter().date(from: expirationDate)!)
```
JayShortway, line 109: This fixed this test running on my local machine, where my locale would not have formatted the date as `"Apr 12, 2062"`.

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -84,13 +84,23 @@ class ManageSubscriptionsViewModel: ObservableObject {
...
+        
+        // Pick the soonest expiring iOS App Store entitlement and accompanying product.
+        guard let currentEntitlement = customerInfo.entitlements
+            .active
+            .values
+            .lazy
+            .filter({ entitlement in entitlement.store == .appStore })
+            .sorted(by: { lhs, rhs in
+                let lhsDateSeconds = lhs.expirationDate?.timeIntervalSince1970 ?? TimeInterval.greatestFiniteMagnitude
+                let rhsDateSeconds = rhs.expirationDate?.timeIntervalSince1970 ?? TimeInterval.greatestFiniteMagnitude
```
JayShortway, line 96: Lifetime subscriptions are last in line, but this current implementation is not great because the order is undefined for multiple lifetime subscriptions. We can discuss options in the sync (see **Not in this PR** in the PR description),  or feel free to suggest any quick wins here.

aboedo: we usually go with the one that's expiring latest into the future (that's how we compute customerInfo for example), so that if you're subscribed to a monthly that's about to expire and have moved into a yearly or something, we show you the one that'll matter most - the one that gives you access for the longest time

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
```diff
@@ -71,27 +73,256 @@ class ManageSubscriptionsViewModelTests: TestCase {
...
+
+        // Act
+        await viewModel.loadScreen()
+
+        // Assert
+        expect(viewModel.subscriptionInformation).toNot(beNil())
+        expect(viewModel.screen).toNot(beNil())
+        expect(viewModel.state) == .success
+
+        expect(viewModel.subscriptionInformation?.title) == "yearly"
```
aboedo, line 176: If subscriptionInformation is nil, all these 3 will fail but the error message won't be that clean. You can use `XCTUnwrap` to unwrap the variable and have a neat error message if there's no value there, i.e.: 
```swift
let subscriptionInformation = try XCTUnwrap(viewModel.subscriptionInformation)
```
(you'll have to add a `throws` in the method's signature, but in tests that's fine and it'll still give you a clean test failure)

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
```diff
@@ -71,27 +73,256 @@ class ManageSubscriptionsViewModelTests: TestCase {
...
         expect(viewModel.subscriptionInformation?.title) == "title"
         expect(viewModel.subscriptionInformation?.durationTitle) == "month"
         expect(viewModel.subscriptionInformation?.price) == "$2.99"
-        expect(viewModel.subscriptionInformation?.nextRenewalString) == "Apr 12, 2062"
-        expect(viewModel.subscriptionInformation?.productIdentifier) == "com.revenuecat.product"
+
+        let formatter = DateFormatter()
+        formatter.dateStyle = .medium
+        expect(viewModel.subscriptionInformation?.nextRenewalString) ==
+            formatter.string(from: ISO8601DateFormatter().date(from: expirationDate)!)
```
aboedo, line 122: we're doing `nextRenewalString` == string from date from string, maybe there's a shortcut we can use here? How are we formatting that renewalString in the first place?

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
```diff
@@ -71,27 +73,256 @@ class ManageSubscriptionsViewModelTests: TestCase {
...
+                    purchaseDate: purchaseDate,
+                    expirationDate: expirationDateSecond
+                )
+            ].shuffled()
+        )
+
+        let viewModel = ManageSubscriptionsViewModel(screen: ManageSubscriptionsViewModelTests.screen,
+                                                     purchasesProvider: MockManageSubscriptionsPurchases(
+                                                        customerInfo: customerInfo,
+                                                        products: products
```
aboedo, line 299: haven't looked in detail that the class, but this is the first thing that really jumps out at me as not very swiftUI-ish, the viewModel and the fact that it holds a reference to the screen. 

Why do we hold that reference? Typically you'd have the viewModel own the properties, and have those be published, then a View (or many) observes them

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
JayShortway: Yea that makes a lot of sense to me, but I was going off of this requirement on Linear:
>If there are more than once active subscription active, we pick the iOS subscription with the soonest non-null expiration date.

@vegaro Could you clarify/confirm which is the one we want?

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
JayShortway: Totally agree. In MVVM the ViewModel should be completely decoupled from the UI layer and just publish state, not caring about who listens. 

I did not change this ViewModel, but I can take a look how it can be improved, maybe in a follow-up PR. 

I see that in this case, the `Screen` class is not a SwiftUI view, but a `struct` holding some input data. Also, it is unused. @vegaro Are there any future plans for this, or can it be removed?

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
JayShortway: Ah thanks, TIL! A few lines up we do have this: 

```swift
expect(viewModel.subscriptionInformation).toNot(beNil())
```

Is `XCTUnwrap` the convention over `expect(*).toNot(beNil())`?

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
JayShortway: Yes we are basically re-formatting a ISO-8601 string representation of a date. I'm not sure if there's a shortcut here, but it can definitely be extracted to a separate method, which is what I did here: https://github.com/RevenueCat/purchases-ios/pull/4046/commits/41b8cc054928544833b8f7aaa0cb1c155d3e3136

The formatting of the `nextRenewalString` is unchanged and is as follows:
```swift
let dateFormatter = DateFormatter()
dateFormatter.dateStyle = .medium
let nextRenewalString = currentEntitlement.expirationDate.map { dateFormatter.string(from: $0) }
```

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
vegaro: Yes, we should use `XCTUnwrap` instead of `expect(*).toNot(beNil())`. We don't have linting errors for that so reviews/remembering is the only way. Those three up there are my fault, if you can update them it would be amazing.

Good thing about `XCTUnwrap` as well is that it removes the need to do `?` in subsequent checks

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
vegaro: Oh this is confusing, but screen is the data of the screen that is being sent down in the JSON, so the screen properties like screen title, feedback survey options, etc, it doesn't have anything to do with view classes. It's been used in upward branches so it can't be removed, for example https://github.com/RevenueCat/purchases-ios/pull/3968/files#diff-121dcc7927012032acdbbf026716bf9daf66d91cee530ab12eae052d06526aa0R187

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
vegaro: My thinking for this one was that let's say someone has one expiring tomorrow and another one further in the future, it is more relevant to display there's a subscription expiring soon. But from a length of access perspective, it's true the latest is more relevant. 

@aboedo when you say "that's how we compute customerInfo for example", what part of customer info you refer to? The expiration date of an entitlement?

I have been thinking and I think that what @aboedo is saying makes sense, since I want to see until when I have access to this (the one that gives you access for the longest time), but I also think that not displaying the soonest one (one that might charge the user tomorrow for example) might provide false information and cause more support issues than this view would solve. If I am a user and see that one of my subscriptions expires in a year, and I get charged tomorrow for another sub I would get mad and for sure contact support.

But it's true that showing soon-expiring subscriptions might trigger more cancellations.

The ideal solution is to display all entitlements I think, but for this first iteration, I think we have to pick what to display. What do you think about changing the perspective of this to be length of access (I think that's what we do in the dashboard) and instead of displaying "Expires on 12/30/2024" we display "Access until 12/30/2024" with the one with longest expiration?

Thoughts? I might be overthinking this...

vegaro: Let's go with that for now and get feedback on it. We can always change the logic. And yes, I think at some point 
 we should add another screen with the whole purchase history

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
JayShortway: Yea sure, no worries! Good to be consistent 💪 Updated: https://github.com/RevenueCat/purchases-ios/pull/4046/commits/f10ea101ead44f475214f97d9034ebccc38dab5a

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
```diff
@@ -112,27 +327,36 @@ class ManageSubscriptionsViewModelTests: TestCase {
...
+        let formatter = DateFormatter()
+        formatter.dateStyle = .medium
+        return formatter.string(from: ISO8601DateFormatter().date(from: ISO8601Date)!)
+    }
+
 }
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 final class MockManageSubscriptionsPurchases: ManageSubscriptionsPurchaseType {
-
```
vegaro, line 119: This empty line should be here to match our SwiftStyleGuide https://github.com/RevenueCat/purchases-ios/blob/dbcfe0ceb55879c98d7069c2c788af90b76b27d7/Contributing/SwiftStyleGuide.swift#L18

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
```diff
@@ -171,19 +390,76 @@ final class MockManageSubscriptionsPurchases: ManageSubscriptionsPurchaseType {
 }
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
-private extension ManageSubscriptionsViewModelTests {
+private class Fixtures {
+    private init() {}
```
vegaro, line 394: Same here about the empty lines at the beginning and the bottom in this class and subclasses

by the way, this Fixtures class is neat!

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
JayShortway: Maybe there's a better name for the struct, to avoid the confusion. Something like `Destination` maybe? 

Also I don't see it being used still, but maybe I'm looking in the wrong place haha. It looks like it's only being set, never read, from what I can tell.

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
JayShortway: Good to know, thanks! https://github.com/RevenueCat/purchases-ios/pull/4046/commits/89d4a8d61b50b0a04c8ff82c97b3d8d7b9d5526d

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
JayShortway: Thanks 😄 

Should be fixed: https://github.com/RevenueCat/purchases-ios/pull/4046/commits/89d4a8d61b50b0a04c8ff82c97b3d8d7b9d5526d

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
JayShortway: Alright, since that's a UI change I'll do that in a follow-up PR with more UI stuff.

vegaro (approved):

PR #4047, opened 2024-07-12
Use SK2 to get Storefront to prevent potential hang

joshdholtz:
### Motivation

In `5.0.0`, there were some observed hangs/freezes that were due to attempting to get the Storefront.

The Storefront was being accessed via a SK1 API because it needed to get be a non-async call to set the value for the `X-Storefront` header. However, this seems to sometimes hang when using SK2 by default. This doesn't happen all the time and has been observed in the simulator.

This will change to use the SK2 API for Storefront if available.

### Description

Asynchronously sets and caches the storefront in `SystemInfo`


joshdholtz:
Closing since not high priority ATM

PR #4048, opened 2024-07-12
Retry Requests with HTTP Status 429

fire-at-will:
### Motivation
In the past, the SDK has ignored HTTP Status 429s. This PR introduces a retry ability with an exponential backoff policy to the SDK's HTTPClient so that 429s can be retried.

#### Configuration
The implementation allows consumers of the HTTP client to specify which HTTP status codes to retry for. The exponential backoff policy will retry requests up to three times, with the following backoff times:
- First retry: 0sec
- Second retry: 0.75sec
- Third retry: 3sec

#### Server-Defined Backoff Durations
If the server returns the `Retry-After` header with an integer value, the SDK will use that integer as the number of milliseconds to wait before retrying again instead of computing a backoff duration.

#### New `X-Retry-After` Header For Outbound Network Requests
This PR introduces a new `X-Retry-After` header for all outbound network requests, whose value is an integer. For the initial request (with no retries), the value is 0. For the first retry, the value is 1. For the second retry, the value is 2, and so on.

#### Etag Management
To keep our etag behavior consistent, the backoff duration for a first retry will always be 0 seconds.

#### Logging
A debug message in the format `"Queued request \(httpMethod) \(path) for retry number \(retryNumber) in \(backoffInterval) seconds."` is logged when a request is queued for retry. If a request fails all retries, a message with the format `"Request \(httpMethod) \(path) failed all \(retryCount) retries."` will be logged.

#### Testing
A good number of unit tests have been written, for both the individual parts of the HTTP client introduced in this PR and for testing the entire request flow through the HTTPClient.

### DIFF
--- a/Sources/Logging/Strings/NetworkStrings.swift
+++ b/Sources/Logging/Strings/NetworkStrings.swift
@@ -28,7 +28,7 @@ enum NetworkStrings {
                             error: NetworkError,
                             metadata: HTTPClient.ResponseMetadata?)
     case api_request_failed_status_code(HTTPStatusCode)
-    case api_request_queued_for_retry(httpMethod: String, path: String, backoffInterval: TimeInterval)
+    case api_request_queued_for_retry(httpMethod: String, retryNumber: UInt, path: String, backoffInterval: TimeInterval)
     case api_request_failed_all_retries(httpMethod: String, path: String, retryCount: UInt)
     case reusing_existing_request_for_operation(CacheableNetworkOperation.Type, String)
     case enqueing_operation(CacheableNetworkOperation.Type, cacheKey: String)
@@ -48,7 +48,6 @@ enum NetworkStrings {
 
     #if DEBUG
     case api_request_forcing_server_error(HTTPRequest)
-    case api_request_forcing_network_error(HTTPRequest, NetworkError)
     case api_request_forcing_signature_failure(HTTPRequest)
     case api_request_disabling_header_parameter_signature_verification(HTTPRequest)
     #endif
@@ -134,8 +133,8 @@ extension NetworkStrings: LogMessage {
         case let .request_handled_by_load_shedder(path):
             return "Request was handled by load shedder: \(path.relativePath)"
 
-        case let .api_request_queued_for_retry(httpMethod, path, backoffInterval):
-            return "Queued request \(httpMethod) \(path) for retry in \(backoffInterval) seconds."
+        case let .api_request_queued_for_retry(httpMethod, retryNumber, path, backoffInterval):
+            return "Queued request \(httpMethod) \(path) for retry number \(retryNumber) in \(backoffInterval) seconds."
 
         case let .api_request_failed_all_retries(httpMethod, path, retryCount):
             return "Request \(httpMethod) \(path) failed all \(retryCount) retries."
@@ -143,15 +142,6 @@ extension NetworkStrings: LogMessage {
         #if DEBUG
         case let .api_request_forcing_server_error(request):
             return "Returning fake HTTP 500 error for \(request.description)"
-
-        case let .api_request_forcing_network_error(request, networkError):
-            var simulatedHTTPStatusCode: HTTPStatusCode
-            if case .errorResponse(_, let statusCode, _) = networkError {
-                simulatedHTTPStatusCode = statusCode
-            } else {
-                simulatedHTTPStatusCode = .internalServerError
-            }
-            return "Returning fake HTTP \(simulatedHTTPStatusCode.rawValue) error for \(request.description)"
 
         case let .api_request_forcing_signature_failure(request):
             return "Returning fake signature verification failure for '\(request.description)'"
--- a/Sources/Logging/Strings/NetworkStrings.swift
+++ b/Sources/Logging/Strings/NetworkStrings.swift
@@ -29,6 +29,7 @@ enum NetworkStrings {
                             metadata: HTTPClient.ResponseMetadata?)
     case api_request_failed_status_code(HTTPStatusCode)
     case api_request_queued_for_retry(httpMethod: String, path: String, backoffInterval: TimeInterval)
+    case api_request_failed_all_retries(httpMethod: String, path: String, retryCount: UInt)
     case reusing_existing_request_for_operation(CacheableNetworkOperation.Type, String)
     case enqueing_operation(CacheableNetworkOperation.Type, cacheKey: String)
     case creating_json_error(error: String)
@@ -133,7 +134,10 @@ extension NetworkStrings: LogMessage {
             return "Request was handled by load shedder: \(path.relativePath)"
 
         case let .api_request_queued_for_retry(httpMethod, path, backoffInterval):
-            return "Queued request \(httpMethod) to \(path) for retry in \(backoffInterval) seconds."
+            return "Queued request \(httpMethod) \(path) for retry in \(backoffInterval) seconds."
+
+        case let .api_request_failed_all_retries(httpMethod, path, retryCount):
+            return "Request \(httpMethod) \(path) failed all \(retryCount) retries."
 
         #if DEBUG
         case let .api_request_forcing_server_error(request):
--- a/Sources/Logging/Strings/NetworkStrings.swift
+++ b/Sources/Logging/Strings/NetworkStrings.swift
@@ -48,6 +48,7 @@ enum NetworkStrings {
 
     #if DEBUG
     case api_request_forcing_server_error(HTTPRequest)
+    case api_request_forcing_network_error(HTTPRequest, NetworkError)
     case api_request_forcing_signature_failure(HTTPRequest)
     case api_request_disabling_header_parameter_signature_verification(HTTPRequest)
     #endif
@@ -142,6 +143,15 @@ extension NetworkStrings: LogMessage {
         #if DEBUG
         case let .api_request_forcing_server_error(request):
             return "Returning fake HTTP 500 error for \(request.description)"
+
+        case let .api_request_forcing_network_error(request, networkError):
+            var simulatedHTTPStatusCode: HTTPStatusCode
+            if case .errorResponse(_, let statusCode, _) = networkError {
+                simulatedHTTPStatusCode = statusCode
+            } else {
+                simulatedHTTPStatusCode = .internalServerError
+            }
+            return "Returning fake HTTP \(simulatedHTTPStatusCode.rawValue) error for \(request.description)"
 
         case let .api_request_forcing_signature_failure(request):
             return "Returning fake signature verification failure for '\(request.description)'"
--- a/Sources/Logging/Strings/NetworkStrings.swift
+++ b/Sources/Logging/Strings/NetworkStrings.swift
@@ -28,7 +28,7 @@ enum NetworkStrings {
                             error: NetworkError,
                             metadata: HTTPClient.ResponseMetadata?)
     case api_request_failed_status_code(HTTPStatusCode)
-    case api_request_queued_for_retry(httpMethod: String, 
+    case api_request_queued_for_retry(httpMethod: String,
                                       retryNumber: UInt,
                                       path: String,
                                       backoffInterval: TimeInterval)
--- a/Sources/Logging/Strings/NetworkStrings.swift
+++ b/Sources/Logging/Strings/NetworkStrings.swift
@@ -28,7 +28,10 @@ enum NetworkStrings {
                             error: NetworkError,
                             metadata: HTTPClient.ResponseMetadata?)
     case api_request_failed_status_code(HTTPStatusCode)
-    case api_request_queued_for_retry(httpMethod: String, retryNumber: UInt, path: String, backoffInterval: TimeInterval)
+    case api_request_queued_for_retry(httpMethod: String, 
+                                      retryNumber: UInt,
+                                      path: String,
+                                      backoffInterval: TimeInterval)
     case api_request_failed_all_retries(httpMethod: String, path: String, retryCount: UInt)
     case reusing_existing_request_for_operation(CacheableNetworkOperation.Type, String)
     case enqueing_operation(CacheableNetworkOperation.Type, cacheKey: String)
--- a/Sources/Logging/Strings/NetworkStrings.swift
+++ b/Sources/Logging/Strings/NetworkStrings.swift
@@ -28,6 +28,7 @@ enum NetworkStrings {
                             error: NetworkError,
                             metadata: HTTPClient.ResponseMetadata?)
     case api_request_failed_status_code(HTTPStatusCode)
+    case api_request_queued_for_retry(httpMethod: String, path: String, backoffInterval: TimeInterval)
     case reusing_existing_request_for_operation(CacheableNetworkOperation.Type, String)
     case enqueing_operation(CacheableNetworkOperation.Type, cacheKey: String)
     case creating_json_error(error: String)
@@ -130,6 +131,9 @@ extension NetworkStrings: LogMessage {
 
         case let .request_handled_by_load_shedder(path):
             return "Request was handled by load shedder: \(path.relativePath)"
+
+        case let .api_request_queued_for_retry(httpMethod, path, backoffInterval):
+            return "Queued request \(httpMethod) to \(path) for retry in \(backoffInterval) seconds."
 
         #if DEBUG
         case let .api_request_forcing_server_error(request):
--- a/Sources/Misc/Concurrency/OperationDispatcher.swift
+++ b/Sources/Misc/Concurrency/OperationDispatcher.swift
@@ -57,7 +57,7 @@ class OperationDispatcher {
         Self.dispatchOnMainActor(block)
     }
 
-    func dispatchOnWorkerThread(delay: JitterableDelay = .none, block: @escaping @Sendable () -> Void) {
+    func dispatchOnWorkerThread(jitterableDelay delay: JitterableDelay = .none, block: @escaping @Sendable () -> Void) {
         if delay.hasDelay {
             self.workerQueue.asyncAfter(deadline: .now() + delay.random(), execute: block)
         } else {
@@ -65,11 +65,7 @@ class OperationDispatcher {
         }
     }
 
-    func dispatchOnWorkerThread(after timeInterval: TimeInterval, block: @escaping @Sendable () -> Void) {
-        self.workerQueue.asyncAfter(deadline: .now() + timeInterval, execute: block)
-    }
-
-    func dispatchOnWorkerThread(delay: JitterableDelay = .none, block: @escaping @Sendable () async -> Void) {
+    func dispatchOnWorkerThread(jitterableDelay delay: JitterableDelay = .none, block: @escaping @Sendable () async -> Void) {
         Task.detached(priority: .background) {
             if delay.hasDelay {
                 try? await Task.sleep(nanoseconds: DispatchTimeInterval(delay.random()).nanoseconds)
@@ -77,6 +73,10 @@ class OperationDispatcher {
 
             await block()
         }
+    }
+
+    func dispatchOnWorkerThread(after timeInterval: TimeInterval, block: @escaping @Sendable () -> Void) {
+        self.workerQueue.asyncAfter(deadline: .now() + timeInterval, execute: block)
     }
 
 }
--- a/Sources/Misc/Concurrency/OperationDispatcher.swift
+++ b/Sources/Misc/Concurrency/OperationDispatcher.swift
@@ -57,7 +57,7 @@ class OperationDispatcher {
         Self.dispatchOnMainActor(block)
     }
 
-    func dispatchOnWorkerThread(jitterableDelay delay: JitterableDelay = .none, block: @escaping @Sendable () -> Void) {
+    func dispatchOnWorkerThread(delay: JitterableDelay = .none, block: @escaping @Sendable () -> Void) {
         if delay.hasDelay {
             self.workerQueue.asyncAfter(deadline: .now() + delay.random(), execute: block)
         } else {
--- a/Sources/Misc/Concurrency/OperationDispatcher.swift
+++ b/Sources/Misc/Concurrency/OperationDispatcher.swift
@@ -23,6 +23,7 @@ enum Delay {
     case none
     case `default`
     case long
+    case timeInterval(TimeInterval)
 
     static func `default`(forBackgroundedApp inBackground: Bool) -> Self {
         return inBackground ? .default : .none
@@ -108,6 +109,8 @@ private extension Delay {
         case .none: return 0
         case .`default`: return 0
         case .long: return Self.maxJitter
+        case .timeInterval(let timeInterval):
+            return max(timeInterval - 0.1, 0)
         }
     }
 
@@ -116,6 +119,8 @@ private extension Delay {
         case .none: return 0
         case .`default`: return Self.maxJitter
         case .long: return Self.maxJitter * 2
+        case .timeInterval(let timeInterval):
+            return max(timeInterval + 0.1, 0)
         }
     }
 
--- a/Sources/Misc/Concurrency/OperationDispatcher.swift
+++ b/Sources/Misc/Concurrency/OperationDispatcher.swift
@@ -18,7 +18,7 @@ import Foundation
 ///
 /// These delays prevent DDOS if a notification leads to many users opening an app at the same time,
 /// by spreading asynchronous operations over time.
-enum Delay {
+enum Delay: Equatable {
 
     case none
     case `default`
--- a/Sources/Misc/Concurrency/OperationDispatcher.swift
+++ b/Sources/Misc/Concurrency/OperationDispatcher.swift
@@ -110,7 +110,7 @@ private extension Delay {
         case .`default`: return 0
         case .long: return Self.maxJitter
         case .timeInterval(let timeInterval):
-            return max(timeInterval - 0.1, 0)
+            return max(timeInterval, 0)
         }
     }
 
@@ -120,7 +120,7 @@ private extension Delay {
         case .`default`: return Self.maxJitter
         case .long: return Self.maxJitter * 2
         case .timeInterval(let timeInterval):
-            return max(timeInterval + 0.1, 0)
+            return max(timeInterval, 0)
         }
     }
 
--- a/Sources/Misc/Concurrency/OperationDispatcher.swift
+++ b/Sources/Misc/Concurrency/OperationDispatcher.swift
@@ -18,7 +18,7 @@ import Foundation
 ///
 /// These delays prevent DDOS if a notification leads to many users opening an app at the same time,
 /// by spreading asynchronous operations over time.
-enum Delay: Equatable {
+enum JitterableDelay: Equatable {
 
     case none
     case `default`
@@ -57,7 +57,7 @@ class OperationDispatcher {
         Self.dispatchOnMainActor(block)
     }
 
-    func dispatchOnWorkerThread(delay: Delay = .none, block: @escaping @Sendable () -> Void) {
+    func dispatchOnWorkerThread(delay: JitterableDelay = .none, block: @escaping @Sendable () -> Void) {
         if delay.hasDelay {
             self.workerQueue.asyncAfter(deadline: .now() + delay.random(), execute: block)
         } else {
@@ -65,7 +65,11 @@ class OperationDispatcher {
         }
     }
 
-    func dispatchOnWorkerThread(delay: Delay = .none, block: @escaping @Sendable () async -> Void) {
+    func dispatchOnWorkerThread(after timeInterval: TimeInterval, block: @escaping @Sendable () -> Void) {
+        self.workerQueue.asyncAfter(deadline: .now() + timeInterval, execute: block)
+    }
+
+    func dispatchOnWorkerThread(delay: JitterableDelay = .none, block: @escaping @Sendable () async -> Void) {
         Task.detached(priority: .background) {
             if delay.hasDelay {
                 try? await Task.sleep(nanoseconds: DispatchTimeInterval(delay.random()).nanoseconds)
@@ -90,7 +94,7 @@ extension OperationDispatcher {
 // MARK: -
 
 /// Visible for testing
-extension Delay {
+extension JitterableDelay {
 
     var hasDelay: Bool {
         return self.maximum > 0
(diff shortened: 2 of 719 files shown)


Sources/Misc/Concurrency/OperationDispatcher.swift
```diff
@@ -108,6 +109,8 @@ private extension Delay {
         case .none: return 0
         case .`default`: return 0
         case .long: return Self.maxJitter
+        case .timeInterval(let timeInterval):
+            return max(timeInterval - 0.1, 0)
```
fire-at-will, line 113: Including a +/- 0.1 here to allow for some jittering

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -207,7 +214,7 @@ extension HTTPClient: @unchecked Sendable {}
 
 // MARK: - Private
 
-private extension HTTPClient {
+internal extension HTTPClient {
```
fire-at-will, line 217: Modified the access modified here so I could do some testing

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -222,7 +229,14 @@ private extension HTTPClient {
...
         var completionHandler: HTTPClient.Completion<Data>?
-        var retried: Bool = false
+
+        /// Whether the request has been retried.
+        var retried: Bool {
+            return self.retryCount > 0
+        }
+
+        /// The number of times that we have retried the request
+        var retryCount: UInt = 0
```
fire-at-will, line 239: I'm using a `UInt` instead of an `Int` here so there won't be any negative values 😎

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -269,6 +283,32 @@ private extension HTTPClient {
...
+            maxBackoff=\(maxBackoff.debugDescription)
+            maxNumberOfRetries=\(maxNumberOfRetries)
+            """
+        }
+    }
+
+}
+
+internal extension HTTPClient.ExponentialRetryOptions {
+    static let `default` = HTTPClient.ExponentialRetryOptions(
```
fire-at-will, line 306: I'm totally open to changing these default values to something else if we want - this is just what came to the top of my head first!

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -582,6 +624,87 @@ private extension HTTPClient {
 
 }
 
+// MARK: - Request Retry Logic
+extension HTTPClient {
+    internal func retryRequestIfNeeded(
```
fire-at-will, line 629: This is the only function I wasn't able to directly test - let me know if y'all have any ideas! 😄

aboedo:
For end-to-end testing, I'd suggest using some proxy server ([like this one](https://github.com/RevenueCat/khepri-mitm-proxy/blob/main/index.js) that you can control and set stuff like "give me a 429 for the next request only"

Sources/Misc/Concurrency/OperationDispatcher.swift
aboedo: we should turn that into a comment

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -196,6 +202,7 @@ extension HTTPClient {
         case isLoadShedder = "X-RevenueCat-Fortress"
         case requestID = "X-Request-ID"
         case amazonTraceID = "X-Amzn-Trace-ID"
+        case retryAfter = "Retry-After"
```
aboedo, line 205: todo: confirm header name with backend devs

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -582,6 +624,87 @@ private extension HTTPClient {
...
 
+// MARK: - Request Retry Logic
+extension HTTPClient {
+    internal func retryRequestIfNeeded(
+        request: HTTPClient.Request,
+        httpURLResponse: HTTPURLResponse?
+    ) {
+        // retryCount is incremented before the retry is executed, so don't stop retries if
+        // retryCount == self.retryOptions.maxNumberOfRetries
+        guard request.retryCount > self.retryOptions.maxNumberOfRetries else { return }
```
aboedo, line 635: shouldn't this be the other way around?

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -582,6 +624,87 @@ private extension HTTPClient {
...
+        guard let httpURLResponse = httpURLResponse else { return }
+        guard shouldRetryRequest(withStatusCode: httpURLResponse.httpStatusCode) else { return }
```
aboedo, line 637: I'd probably make this the first check since conceptually it seems more important than the others

aboedo: we're always returning on these, maybe we could just chain them with `&&` and `,`?

aboedo: then again we may wanna log, especially for the case where we've reached max retries on 429

Sources/Misc/Concurrency/OperationDispatcher.swift
aboedo: do we need it though? the 429's shouldn't be triggered by something like a push notification, so presumably we wouldn't be sending out these requests en masse

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -582,6 +624,87 @@ private extension HTTPClient {
...
+    internal func shouldRetryRequest(withStatusCode statusCode: HTTPStatusCode) -> Bool {
+        return self.retryOptions.retriableStatusCodes.contains(statusCode)
+    }
+
+    internal func calculateRetryBackoffTime(
+        forResponse httpURLResponse: HTTPURLResponse,
+        retryCount: UInt
+    ) -> TimeInterval {
+        // If this is our first retry and there's no eTag header, ensure there's no delay
+        if httpURLResponse.allHeaderFields[ResponseHeader.eTag] == nil && retryCount == 1 {
```
aboedo, line 670: if we just make the min backoff time zero we can skip this bit

aboedo: I'd imagine that most of the time, if you hit a 429 and retry immediately, by the time you hit the backend again it'll already be solved

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -582,6 +624,87 @@ private extension HTTPClient {
...
+            // retryCount 2: backoff = baseBackoff * 2^1 = 0.5
+            // retryCount 3: backoff = baseBackoff * 2^2 = 1.0
+            return min(baseBackoff * pow(2.0, Double(retryCount - 1)), self.retryOptions.maxBackoff)
+        } else {
+            // If baseBackoff is greater than or equal to 1, just use baseBackoff^retryCount:
+            // retryCount 1: backoff = baseBackoff^1
+            // retryCount 2: backoff = baseBackoff^2
+            // ...
+            return min(pow(baseBackoff, Double(retryCount)), self.retryOptions.maxBackoff)
+        }
```
aboedo, line 704: we only have 3 possible values in that we only retry up to three times, we might as well just explicitly set the values

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -1664,6 +1665,145 @@ final class HTTPClientTests: BaseHTTPClientTests<MockETagManager> {
...
+        expect(
+            self.client.calculateDefaultExponentialBackoffTimeInterval(withRetryCount: 2)
+        ).to(equal(0.5))
+
+        expect(
+            self.client.calculateDefaultExponentialBackoffTimeInterval(withRetryCount: 3)
+        ).to(equal(1))
+    }
+
+}
```
aboedo, line 1805: You can also use OHTTPStubs to fake responses from the backend and test. 

The files named `somethingBackendTests` have plenty of examples you can mostly reuse, especially `BackendPostReceiptDataTests`

aboedo:
great work on this!

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -269,6 +283,32 @@ private extension HTTPClient {
         }
     }
 
+    struct ExponentialRetryOptions: CustomStringConvertible {
+
+        let retriableStatusCodes: Set<HTTPStatusCode>
+        let baseBackoff: TimeInterval
+        let maxBackoff: TimeInterval
+        let maxNumberOfRetries: UInt
```
tonidero, line 291: I was wondering, do we need both a `maxBackoff` and `maxNumberOfRetries`? One or the other could be computed of the other + baseBackoff right?

Sources/Misc/Concurrency/OperationDispatcher.swift
fire-at-will: Fair enough, I'll remove the jittering here!

Sources/Networking/HTTPClient/HTTPClient.swift
fire-at-will: Great catch, the correct version should be:

```swift
guard request.retryCount <= self.retryOptions.maxNumberOfRetries else { return }
```

Fixed!

Sources/Networking/HTTPClient/HTTPClient.swift
tonidero: I wonder if it would make sense to have different defaults per endpoint type... As in, maybe we don't need to retry some non-critical requests like diagnostics, product entitlement mapping,...

Sources/Networking/HTTPClient/HTTPClient.swift
fire-at-will: Logging here when we've reached the max retry count is a great idea - I'll add one in and combine the other two guard statements.

Sources/Networking/HTTPClient/HTTPClient.swift
fire-at-will: Nice, removed!

Sources/Networking/HTTPClient/HTTPClient.swift
fire-at-will: Confirmed!

fire-at-will:
@aboedo I've updated the PR with updates from your comments! Specifically, I've made the following changes:

- Removed the jitter from the `timeInterval` delay
- Fixed the logic that determines if if the request has exceeded the maximum number of retries.
- Refactored some guard statements that could be combined
- Added a log message to log when a request has used up all of its retries and still failed
- Hard coded the retry backoff intervals. As a part of this change, the `ExponentialRetryOptions` become redundant and was removed. Now, the `HTTPClient`'s constructor only takes in a set of HTTP status codes to retry for, which defaults to only 429s
- Fixed a bug so that a HTTPClient.Request's callback is only executed after the request succeeds or uses up all of its retries
- Wrote a few tests using OhHTTPStubs to test the entire retry flow in the HTTPClient. Thanks for pointing out that library to me! 😄

Sources/Networking/HTTPClient/HTTPClient.swift
fire-at-will: In my updates to some of the other PR comments, I've hardcoded the backoff intervals. After that change was done, most of this struct was redundant so it's been removed!

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -582,6 +606,95 @@ private extension HTTPClient {
...
+    ///
+    /// This function calculates the backoff time interval for a given retry count based on a predefined
+    /// list of backoff intervals. The `retryCount` is not 0-indexed and should start with 1.
+    ///
+    /// - Parameter retryCount: The count of the retry attempt, starting from 1.
+    /// - Returns: The backoff time interval for the given retry count. If the retry count exceeds
+    ///   the predefined list of backoff intervals, it returns 0.
+    internal func defaultExponentialBackoffTimeInterval(withRetryCount retryCount: UInt) -> TimeInterval {
+        return retryCount - 1 < self.retryBackoffIntervals.count
+        ? self.retryBackoffIntervals[Int(max(retryCount - 1, 0))]
```
MarkVillacampa, line 693: Extremely very nit, but we could extract the max() for more clarity:

```
var backoffIntervalIndex = Int(max(retryCount - 1, 0))
return backoffInternalIndex < self.retryBackoffIntervals.count ? self.retryBackoffIntervals[backoffInternalIndex] : 0
```

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -582,6 +606,95 @@ private extension HTTPClient {
...
+        return self.retriableStatusCodes.contains(statusCode)
+    }
+
+    internal func calculateRetryBackoffTime(
+        forResponse httpURLResponse: HTTPURLResponse,
+        retryCount: UInt
+    ) -> TimeInterval {
+        // Use the retry after value from the backend if present
+        if let retryAfterHeaderValue = httpURLResponse.allHeaderFields[ResponseHeader.retryAfter.rawValue] as? String {
+            if let retryAfterMS = Double(retryAfterHeaderValue) {
```
MarkVillacampa, line 674: could this be in the same `if`?
```
if let retryAfterHeaderValue = httpURLResponse.allHeaderFields[ResponseHeader.retryAfter.rawValue] as? String, let retryAfterMS = Double(retryAfterHeaderValue) {
```

Sources/Networking/HTTPClient/HTTPClient.swift
fire-at-will: Love this! Will make the change.

Sources/Networking/HTTPClient/HTTPClient.swift
fire-at-will: Nice catch! Always nice to have fewer nested conditions 😄

Sources/Misc/Concurrency/OperationDispatcher.swift
```diff
@@ -108,6 +109,8 @@ private extension Delay {
         case .none: return 0
         case .`default`: return 0
         case .long: return Self.maxJitter
+        case .timeInterval(let timeInterval):
```
aboedo, line 112: if we don't need the jitter, we might as well make another method in operation dispatcher that just takes an exact value as `TimeInterval` rather as an overload to the one we're using instead of passing in a min and max value that are the same

Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -1664,6 +1666,210 @@ final class HTTPClientTests: BaseHTTPClientTests<MockETagManager> {
...
+            .init(code: .unknownError,
+                  originalCode: 0,
+                  message: nil),
+            .tooManyRequests
+        )
+        expect(error.isServerDown) == false
+        self.logger.verifyMessageWasLogged("Queued request GET /v1/subscribers/identify for retry in 0.0 seconds.")
+        self.logger.verifyMessageWasLogged("Queued request GET /v1/subscribers/identify for retry in 0.25 seconds.")
+        self.logger.verifyMessageWasLogged("Queued request GET /v1/subscribers/identify for retry in 0.5 seconds.")
+        self.logger.verifyMessageWasLogged("Request GET /v1/subscribers/identify failed all 3 retries.")
```
aboedo, line 1833: these are quite dangerous in that the minute we change the logs we mess up tests. That'd be fine if we were trying to specifically test that we're logging, but here we're testing "Performs all retries if always gets retryable status code", so we should instead make sure to test for that. 

We test for request count to hit 4, maybe we could also mock the operation dispatcher as a separate test and ensure that it's getting the right delays in?

Tests/UnitTests/Networking/HTTPClientTests.swift
aboedo: feels like we should get at least one backend post receipt test in with the exact case we want to cover (first request is a 429, second goes through).

We can also use our BackendIntegrationTests suite to actually force a 429, remove the forced 429, post something to the backend, ensure it went through. 

`OfflineStoreKitIntegrationTests` has some examples of doing something similar

Sources/Misc/Concurrency/OperationDispatcher.swift
```diff
@@ -108,6 +109,8 @@ private extension Delay {
...
+            return max(timeInterval, 0)
```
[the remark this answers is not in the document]
fire-at-will: I can't think of any instances where the timeInterval would be <0 in the scope of this PR, but this is just defensive programming in case someone ever tries to pass in a negative `TimeInterval` as a delay in the future

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -33,15 +33,25 @@ class HTTPClient {
     private let signing: SigningType
     private let diagnosticsTracker: DiagnosticsTrackerType?
     private let dateProvider: DateProvider
+    private let retriableStatusCodes: Set<HTTPStatusCode>
+    private let operationDispatcher: OperationDispatcher
+
+    private let retryBackoffIntervals: [TimeInterval] = [
+        TimeInterval(0),
+        TimeInterval(0.25),
+        TimeInterval(0.5)
```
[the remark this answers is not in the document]
fire-at-will: That works for me! I'll make the adjustments to use 0sec, 0.75sec, and 3sec. 👍

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -446,9 +466,13 @@ private extension HTTPClient {
...
                     Logger.debug(Strings.network.request_handled_by_load_shedder(request.httpRequest.path))
                 }
+
+                requestRetryScheduled = self.retryRequestIfNeeded(request: request, httpURLResponse: httpURLResponse)
             }
 
-            request.completionHandler?(response)
+            if !requestRetryScheduled {
+                request.completionHandler?(response)
+            }
```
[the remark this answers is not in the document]
fire-at-will: Not yet, but this is an excellent thing to add. I'll look into adding the retry attempt number to the headers!

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -582,6 +606,94 @@ private extension HTTPClient {
...
+
+    /// Returns the exponential backoff time interval for a given retry count.
+    ///
+    /// This function calculates the backoff time interval for a given retry count based on a predefined
+    /// list of backoff intervals. The `retryCount` is not 0-indexed and should start with 1.
+    ///
+    /// - Parameter retryCount: The count of the retry attempt, starting from 1.
+    /// - Returns: The backoff time interval for the given retry count. If the retry count exceeds
+    ///   the predefined list of backoff intervals, it returns 0.
+    internal func defaultExponentialBackoffTimeInterval(withRetryCount retryCount: UInt) -> TimeInterval {
```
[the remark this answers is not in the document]
fire-at-will: removed the function!

Sources/Misc/Concurrency/OperationDispatcher.swift
fire-at-will: I think the argument can be made either way. If we were to make that change, we'd then have two similar dispatch functions that happen after a period of time: 
- `func dispatchOnWorkerThread(delay: Delay)`
- `func dispatchOnWorkerThread(after timeInterval: TimeInterval)`

When I read these function signatures, it isn't clear which one I should choose because the function signatures signify the same thing. 

If we were to go through with this change, I'd recommend renaming `Delay` to something like `JitterableDelay` to indicate that one of these APIs includes a jitter and the other doesn't. I'm down to make this change - any objections to that renaming?

Tests/UnitTests/Networking/HTTPClientTests.swift
fire-at-will: Love this! I've moved the log assertions to their own tests and created separate test to verify that the correct delay times are being sent to the OperationDispatcher.

fire-at-will:
Hey @aboedo, I've made several changes to this PR from your last review! Sorry for the huge increase in PR size, it's mostly from the HTTP snapshots 😅

Since you've last looked at it, I've made the following changes:

### Retry Logic
- Updated the retry backoff time intervals to 0sec, 0.75sec, and 3sec for consecutive retries

### New `X-Retry-Count` Header
- Added the `X-Retry-Count` header to all outbound requests for debugging purposes. Its value starts at 0 for a request that hasn't yet been retried, 1 for the first retry, and so on.
-  Updated the network request snapshots to include the header.

### OperationDispatcher
- Added a new function `func dispatchOnWorkerThread(after timeInterval: TimeInterval)` to accept a `TimeInterval` with no jittering.
- Removed `.timeInterval` from `Delay` and renamed `Delay` to `JitterableDelay` to make the APIs more clear.
- Wrote a test to verify that the correct delays are being sent to the OperationDispatcher for retries.

### Testing
- Introduced the ability for us to instruct the HTTPClient to fail requests to certain paths with a predefined error. This only executes in debug builds and when running one of our test suits.
   - Developers writing a test can force these failures on specific endpoints with the new `forcedServerErrors` property on the `InternalDangerousSettingsType`. 
   - A warning in the format `"Returning fake HTTP \(simulatedHTTPStatusCode) error for \(request)"` is logged when this happens.
- Wrote two integration tests :
   - One to verify that `purchase()` function calls still succeed as expected if a retry completes successfully
   - One to verify that `purchase()` function calls throw an error if all retry attempts fail
- Removed the logging asserts from the tests and moved them to their own tests

Let me know what you think!

Sources/Networking/HTTPClient/HTTPClient.swift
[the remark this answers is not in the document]
tonidero: Right, we are already tracking http requests response codes: https://github.com/RevenueCat/purchases-ios/blob/373abdd6629ff3eaa3472bd3f442397c11be3f2c/Sources/Diagnostics/DiagnosticsTracker.swift#L26.

We could add the retry count there. I'm a bit reticent to add another tag to the metric in Grafana (we are being mindful of increasing the cardinality of the data), but it might be useful especially if/when we move to using snowflake or similar for this data.
If we want to do this, we just need to append the new property as a valid property for the http_request_performed diagnostic event in the backend and we can start sending it from the SDK. Lmk if you want me to tackle this!

Sources/Misc/DangerousSettings.swift
```diff
@@ -124,6 +127,8 @@ internal protocol InternalDangerousSettingsType: Sendable {
     /// Whether `HTTPClient` will fake server errors
     var forceServerErrors: Bool { get }
 
+    var forcedServerErrors: [String: [NetworkError]]? { get }
```
tonidero, line 130: I wonder with this if we should obsolete the `forceServerErrors` property instead, since you could potentially do the same thing with this I guess... but maybe it's ok to leave as is for now.

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -33,15 +33,29 @@ class HTTPClient {
     private let signing: SigningType
     private let diagnosticsTracker: DiagnosticsTrackerType?
     private let dateProvider: DateProvider
+    private let retriableStatusCodes: Set<HTTPStatusCode>
+    private let operationDispatcher: OperationDispatcher
+
+    #if DEBUG
+    private var pathRequestCounts: [String: Int] = [:]
```
tonidero, line 40: Hmm I think we're not resetting these accross multiple requests? Probably ok for now since it's only used on tests for now though.

fire-at-will:
@MarkVillacampa pointed out that the forcedError stuff in the integration tests can probably be accomplished much more easily and safely with OhHTTPStubs, so I'm going to look at doing that! :D

Sources/Logging/Strings/NetworkStrings.swift
```diff
@@ -131,10 +134,25 @@ extension NetworkStrings: LogMessage {
         case let .request_handled_by_load_shedder(path):
             return "Request was handled by load shedder: \(path.relativePath)"
 
+        case let .api_request_queued_for_retry(httpMethod, path, backoffInterval):
+            return "Queued request \(httpMethod) \(path) for retry in \(backoffInterval) seconds."
```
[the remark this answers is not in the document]
fire-at-will: Love this, will add it in!

tonidero (approved):
LGTM! But might be good to get some other checks as well

fire-at-will:
@MarkVillacampa I've made the following changes from your review:
- Replaced the `forcedServerErrors` mechanism with OhHTTPStubs in the integration tests and removed the associated code from `HTTPClient` (YAY!!) 🙌
- Added the retry count number to the log message where we state that we've queued the request for a retry

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -932,6 +935,51 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
...
+
+        let product = try await self.monthlyPackage.storeProduct
+        let customerInfo = try await self.purchases.purchase(product: product).customerInfo
+        try await self.verifyEntitlementWentThrough(customerInfo)
+        expect(customerInfo.allPurchasedProductIdentifiers) == [
+            product.productIdentifier
+        ]
+    }
+
+    func testVerifyPurchaseDoesntGrantEntitlementsAfter429RetriesExhausted() async throws {
```
MarkVillacampa, line 967: Is there a way we can test that exactly 3 retries were made here?

fire-at-will: Discussed this with Mark in Slack - we added in verifications for the number of times that the request was failed before eventually succeeding with the backend

Sources/Misc/Concurrency/OperationDispatcher.swift
```diff
@@ -56,15 +57,19 @@ class OperationDispatcher {
...
+    func dispatchOnWorkerThread(delay: JitterableDelay = .none, block: @escaping @Sendable () -> Void) {
         if delay.hasDelay {
             self.workerQueue.asyncAfter(deadline: .now() + delay.random(), execute: block)
         } else {
             self.workerQueue.async(execute: block)
         }
     }
 
-    func dispatchOnWorkerThread(delay: Delay = .none, block: @escaping @Sendable () async -> Void) {
+    func dispatchOnWorkerThread(after timeInterval: TimeInterval, block: @escaping @Sendable () -> Void) {
```
joshdholtz, line 68: This might be a small Josh nit but `delay` and `after` seem like very similar argument names and I would probably be confused by trying to figure out the difference between these 😅

Not a thing that _has_ to change in the PR but just a thought 🤷‍♂️

Sources/Networking/HTTPClient/HTTPClient.swift
joshdholtz: Would using the `@Testable import RevenueCat` thing have made this visible for testing? I can't _exactly_ remember the correc usecase for it but figured I'd ask 😇

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -579,7 +606,84 @@ private extension HTTPClient {
...
+            let retriedRequest = request.retriedRequest()
+            self.state.modify {
+                $0.queuedRequests.insert(retriedRequest, at: 0)
+            }
+            self.beginNextRequest()
+        }
+        return true
+    }
+
+    internal func shouldRetryRequest(withStatusCode statusCode: HTTPStatusCode) -> Bool {
```
joshdholtz, line 668: Nit: I would have assumed that by this method name this it would do a few different checks (like status code and also number of retries already performed)

I would maybe rename this to be more explicity about status code or add in the retry count logic in here as well 🤷‍♂️

But also, not necessary to make the changes! Just a thought

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -14,6 +14,9 @@ import StoreKitTest
 import UniformTypeIdentifiers
 import XCTest
 
+import OHHTTPStubs
+import OHHTTPStubsSwift
```
joshdholtz, line 18: Nit: I know very little about what our standard is for organizing import but seems like these could be grouped without a new line inbetween 🤷‍♂️

joshdholtz (approved):
A few small nits/thought but otherwise looks good 👍

Sources/Networking/HTTPClient/HTTPClient.swift
fire-at-will: I wish! Unfortunately `@testable` only grants you access to internal stuff, not private

Sources/Misc/Concurrency/OperationDispatcher.swift
fire-at-will: I totally agree with you! The only difference is that `delay: JitterableDelay` doesn't let you specify the exact delay and includes a random jitter, while `after timeInterval: TimeInterval` allows you to specify the exact waiting time without a random jitter

Sources/Networking/HTTPClient/HTTPClient.swift
fire-at-will: Honestly I think I'll just get rid of the function since all it does is check that `retriableStatusCodes` set 😅

fire-at-will: Nice point! I'll rename it to `isStatusCodeRetryable()`

Sources/Networking/HTTPClient/HTTPClient.swift
aboedo: honestly any time we're making private stuff internal for testing, it's a sign that it should likely be extracted into a separate internal entity instead. Not only for easier testing, but also to reduce the amount of stuff a single object (in this case, HTTPClient) is doing

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -579,7 +606,84 @@ private extension HTTPClient {
             }
         }
     }
+}
+
+// MARK: - Request Retry Logic
+extension HTTPClient {
```
aboedo, line 612: This extension has some logic that might be better suited to its own file, such that the only stuff that's remaining in HTTPClient is `retryRequestIfNeeded`

aboedo: Doesn't need to be addressed in this PR, but would be a good follow-up

aboedo (approved):

PR #4050, opened 2024-07-12
[do not merge] test ci job triggers

aboedo:
just testing ci job triggers from #4025, I'd expect that this fires all tests and release checks since this branch format starts with `release`


aboedo:
note: `release-checks` is expected to fail because I didn't run `bump`, so it'll find a git tag that's already pushed since the version number hasn't changed

PR #4053, opened 2024-07-16
[CustomerCenter] Add `presentCustomerCenter` modifier

tonidero:
### Description
This adds a modifier, `presentCustomerCenter` that can be used to more simply display the customer center. The API looks like:
```
.presentCustomerCenter(isPresented: self.$presentingCustomerCenter) {
    self.presentingCustomerCenter = false
}
```

### DIFF
--- a/RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
+++ b/RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
@@ -53,17 +53,15 @@ extension View {
     /// - Parameter onDismiss: Callback executed when the customer center wants to be dismissed.
     /// Make sure you stop presenting the customer center when this is called
     /// - Parameter presentationMode: The desired presentation mode of the customer center. Defaults to `.sheet`.
-    /// - Parameter myAppPurchaseLogic: Use this if you are handling purchases/restores on your own instead of through RevenueCat.
     public func presentCustomerCenter(
         isPresented: Binding<Bool>,
         onDismiss: @escaping () -> Void,
-        presentationMode: CustomerCenterPresentationMode = .default,
-        myAppPurchaseLogic: MyAppPurchaseLogic? = nil
+        presentationMode: CustomerCenterPresentationMode = .default
     ) -> some View {
         return self.modifier(PresentingCustomerCenterModifier(
             isPresented: isPresented,
             onDismiss: onDismiss,
-            myAppPurchaseLogic: myAppPurchaseLogic,
+            myAppPurchaseLogic: nil,
             presentationMode: presentationMode
         ))
     }
--- a/RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
+++ b/RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
@@ -49,7 +49,6 @@ extension View {
     ///      .presentCustomerCenter()
     /// }
     /// ```
-
     /// - Parameter isPresented: Binding indicating whether the customer center should be displayed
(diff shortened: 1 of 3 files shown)


RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
```diff
@@ -0,0 +1,124 @@
...
+        myAppPurchaseLogic: MyAppPurchaseLogic?,
+        presentationMode: CustomerCenterPresentationMode,
+        purchaseHandler: PurchaseHandler? = nil
+    ) {
+        self._isPresented = isPresented
+        self.presentationMode = presentationMode
+        self.onDismiss = onDismiss
+        self._purchaseHandler = .init(wrappedValue: purchaseHandler ??
+                                      PurchaseHandler.default(performPurchase: myAppPurchaseLogic?.performPurchase,
+                                                              performRestore: myAppPurchaseLogic?.performRestore))
```
tonidero, line 91: I added this from paywalls, but removed it from the public API for now, since we still don't allow passing it to the `CustomerCenterView`

RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
```diff
@@ -0,0 +1,124 @@
...
+                content
+                    .fullScreenCover(isPresented: self.$isPresented, onDismiss: self.onDismiss) {
+                        self.customerCenterView()
+                    }
+            }
+        }
+    }
+
+    private func customerCenterView() -> some View {
+        CustomerCenterView()
```
tonidero, line 118: We need to allow passing in the `purchaseHandler` (and a lot of other things to the `CustomerCenterView`). We can add those in separate PRs.

RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
```diff
@@ -0,0 +1,126 @@
...
+//
+//  Created by Toni Rico Diez on 2024-07-15.
+
+import RevenueCat
+import SwiftUI
+
+#if !os(macOS) && !os(tvOS)
+
+/// Presentation options to use with the [presentCustomerCenter](x-source-tag://presentCustomerCenter) View modifiers.
+public enum CustomerCenterPresentationMode {
```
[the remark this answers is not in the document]
JayShortway: Yea I think this is fine too. `PresentationMode` would be very generic indeed, and we cannot be sure that all our future UI features will support all modes, which such a name would kind of imply.

JayShortway (commented):
It all looks good to me! I'll leave approval to someone with a bit more context on this modifier. 🙂

RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
```diff
@@ -0,0 +1,127 @@
...
+
+    /// Presents the ``CustomerCenter``.
+    /// Example:
+    /// ```swift
+    /// var body: some View {
+    ///    YourApp()
+    ///      .presentCustomerCenter()
+    /// }
+    /// ```
+
```
vegaro, line 52: Is this missing a `///` or will this block be associated with `presentCustomerCenter(` just fine?

vegaro (approved):

RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
tonidero: Ohh that was unintended. Didn't check, but I would think it could bring issues. Will fix. Nice catch!

PR #4054, opened 2024-07-16
[StoreKit 2] Raise error if no AppTransaction and no Transaction present

MarkVillacampa:
When we don't have an AppTransaction, and the list of transactions is empty as well, this is likely an store issue, so we return an error and let the developer retry the operation.

### Checklist

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -1129,6 +1129,14 @@ private extension PurchasesOrchestrator {
                         return
                     }
 
+                    // We dont have an AppTransactionJWS, and no transaction, return error
+                    if appTransactionJWS == nil {
+                        completion?(.failure(ErrorUtils.storeProblemError(
+                            withMessage: Strings.storeKit.sk2_app_transaction_unavailable.description
+                        )))
+                        return
+                    }
+
                     self.backend.post(receipt: .empty,
                                       productData: nil,
                                       transactionData: .init(appUserID: currentAppUserID,
--- a/Tests/StoreKitUnitTests/PurchasesOrchestratorSK2Tests.swift
+++ b/Tests/StoreKitUnitTests/PurchasesOrchestratorSK2Tests.swift
@@ -726,6 +726,27 @@ class PurchasesOrchestratorSK2Tests: BasePurchasesOrchestratorTests, PurchasesOr
         expect(customerInfo) == mockCustomerInfo
     }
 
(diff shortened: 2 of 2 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1129,6 +1129,14 @@ private extension PurchasesOrchestrator {
                         return
                     }
 
+                    // We dont have an AppTransactionJWS, and no transaction, return error
+                    if appTransactionJWS == nil {
+                        completion?(.failure(ErrorUtils.storeProblemError(
```
fire-at-will, line 1134: Should we make sure that this is dispatched on the main thread?

Tests/StoreKitUnitTests/PurchasesOrchestratorSK2Tests.swift
```diff
@@ -726,6 +726,27 @@ class PurchasesOrchestratorSK2Tests: BasePurchasesOrchestratorTests, PurchasesOr
...
+        self.mockTransactionFetcher.stubbedAppTransactionJWS = nil
+        self.customerInfoManager.stubbedCachedCustomerInfoResult = CustomerInfo.missingOriginalApplicationVersion
+
+        do {
+            _ = try await self.orchestrator.syncPurchases(receiptRefreshPolicy: .always,
+                                                          isRestore: true,
+                                                          initiationSource: .restore)
+            fail("Expected error")
+        } catch {
+            expect(error).to(matchError(ErrorCode.storeProblemError))
```
fire-at-will, line 741: Is there a way to make sure that this error is thrown from the main thread?

fire-at-will (commented):
I think this PR covers the case where AppTransaction is missing and there are no transactions present for the `syncPurchases` flow, but I think we need to do something to address the case of this happening in the purchase flow as well to be safe.

Perhaps adding something like this to `CustomerAPI.post`, around line 91?

```swift
if receipt == .empty && appTransaction == nil {
    // Call completion with error
}
```

Tests/StoreKitUnitTests/PurchasesOrchestratorSK2Tests.swift
MarkVillacampa: not here afaik, we'll figure out a different way to test this

fire-at-will: I was just doing some similar stuff on the retry PR, one thing that you might be able to do to help is to mock out the `OperationDispatcher` and then verify that `invokedDispatchOnMainThread` is true. While you're not directly testing the thread, it's close in that it's verifying how the OperationDispatcher is executing the work.

fire-at-will (approved):

PR #4055, opened 2024-07-16
[Customer Center] Adds the Customer Center to PaywallsTester

JayShortway:
As the title says. There's a new section called "Customer Center".

### DIFF
--- a/Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
+++ b/Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
@@ -141,19 +141,19 @@ struct SamplePaywallsList: View {
                 } label: {
                     TemplateLabel(name: "Unrecognized paywall", icon: "exclamationmark.triangle")
                 }
-
-                #if os(iOS)
-                Button {
-                    self.presentingCustomerCenter = true
-                } label: {
-                    TemplateLabel(name: "Customer center (sheet)", icon: "person.fill")
-                }
-                #endif
-                
+            }
+            
+            Section("Customer Center") {
                 Button {
                     self.display = .customerCenter
                 } label: {
-                    TemplateLabel(name: "Customer Center (SwiftUI)", icon: "person.fill.questionmark")
+                    TemplateLabel(name: "SwiftUI", icon: "person.fill.questionmark")
+                }
+                
+                Button {
+                    self.presentingCustomerCenter = true
+                } label: {
+                    TemplateLabel(name: "Sheet", icon: "person.fill")
                 }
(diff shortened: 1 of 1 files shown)


Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
```diff
@@ -84,6 +84,8 @@ struct SamplePaywallsList: View {
                     introEligibility: Self.introEligibility
                 )
             )
+        case .customerCenter:
+            CustomerCenterView()
```
tonidero, line 88: I believe this is only available in iOS, so we need to wrap all this code into `#if os(iOS)` macros. We can probably add support for other systems soon, but it's a limitation for now.

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
```diff
@@ -136,6 +138,12 @@ struct SamplePaywallsList: View {
                 } label: {
                     TemplateLabel(name: "Unrecognized paywall", icon: "exclamationmark.triangle")
                 }
+                
+                Button {
+                    self.display = .customerCenter
+                } label: {
+                    TemplateLabel(name: "Customer Center", icon: "person.fill.questionmark")
```
tonidero, line 145: I was actually adding a similar button in my [PR](https://github.com/RevenueCat/purchases-ios/pull/4053). I'm going to update the label to say `Customer Center (sheet)`, so it's clearer. Should we change the label here to `Customer Center (SwiftUI)` so it's more distinct? Maybe it's ok to leave this as is and just change it in other buttons.

tonidero (approved):
A couple comments but LGTM!

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
JayShortway: Ah good call! Will update the label as you suggested. Should we align on the icon as well?

tonidero: I feel we should create another section if we're going to have multiple buttons. As for the icon, we got a close one already :P. I think it's best for it to be slightly different... But honestly doesn't matter too much IMO.

JayShortway: A section sounds good! I'll rebase this onto your branch and add both of our buttons to the section.

PR #4057, opened 2024-07-16
[CustomerCenter] Add action handler

tonidero:
### Description
This PR provides a possible approach to implementing an "action" handler. This allows developer to respond to events that happen during the customer support flow.

The current approach consists of making `CustomerCenterActionHandler`, a lambda that receives an action that can be passed in by the developer. Then calling that with the appropriate action from the customer center. This PR also moves some code to the view model for simplicity and moving logic away from the view layer.


RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
```diff
@@ -54,17 +56,8 @@ struct RestorePurchasesAlert: ViewModifier {
...
-                                    return
-                                }
-                                let hasEntitlements = customerInfo.entitlements.active.count > 0
-                                if hasEntitlements {
-                                    self.setAlertType(.purchasesRecovered)
-                                } else {
-                                    self.setAlertType(.purchasesNotFound)
-                                }
+                                let alertType = await self.customerCenterViewModel.performRestore() 
+                                self.setAlertType(alertType)
```
tonidero, line 60: I tried to move all this logic to the VM since I think it makes much more sense for it to be there. We could potentially create a different view model, but it's easier to pass the `CustomerCenterViewModel` around... Lmk what you think!

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionHandler.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+import RevenueCat
+
+@objc
+public protocol CustomerCenterActionHandler {
+    @objc optional func purchaseCompleted(_ customerInfo: CustomerInfo)
+    @objc optional func restoreStarted()
+    @objc optional func restoreCompleted(_ customerInfo: CustomerInfo)
```
tonidero, line 8: I also thought about having a single method here and a enum for the different events, but this seems simpler and less prone to breaking changes IMO. Lmk what you think!

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionHandler.swift
```diff
@@ -0,0 +1,44 @@
+import Foundation
+import RevenueCat
+
+/// Allows to be notified of certain events the customer may perform during the Customer Center flow
+public protocol CustomerCenterActionHandler {
```
tonidero, line 5: I've been wondering whether to rename this to `CustomerCenterActionDelegate` instead, since it might be more common in iOS... Wdyt?

RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
```diff
@@ -31,6 +31,8 @@ struct RestorePurchasesAlert: ViewModifier {
     @Environment(\.openURL)
     var openURL
 
+    @EnvironmentObject private var customerCenterViewModel: CustomerCenterViewModel
```
tonidero, line 34: Made this an environment object for ease of access.

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionHandler.swift
```diff
@@ -0,0 +1,44 @@
...
+    /// This will be called right before displaying the manage subscription page,
+    /// which may happen for cancellations or changing plans
+    func onShowManageSubscriptions()
+
+    /// This will be called right before starting the refund request flow
+    func onRefundRequestStarted(_ productId: String)
+
+    /// This will be called right after the refund request flow, indicating the result of said flow
+    func onRefundRequestCompleted(_ refundRequestStatus: RefundRequestStatus)
+}
```
vegaro, line 28: Nit: empty line

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -107,4 +117,19 @@ import RevenueCat
         }
     }
 
+    func performRestore() async -> RestorePurchasesAlert.AlertType {
+        self.customerCenterActionHandler?.onRestoreStarted()
+        do {
+            let customerInfo = try await Purchases.shared.restorePurchases()
```
vegaro, line 123: In order to be able to mock this easily when we start writing more tests I think we can have a `purchasesProvider` like the one we have in `ManageSubscriptionsViewModel` that has the functions that are being used in this view model. Something like:

```
protocol CustomerCenterPurchasesType: Sendable {

    @Sendable
    func customerInfo() async throws -> CustomerInfo

    @Sendable
    func restorePurchases() async throws -> CustomerInfo

}
```

If we do that, we can remove the `private var customerInfoFetcher: CustomerInfoFetcher` in this class and mock the return of `customerInfo()` using a `MockCustomerCenterPurchasesType`

By the way the `PurchasesType` comes from a `PaywallPurchasesType` in `Paywalls`. That's where I got it from.

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -49,27 +49,34 @@ class ManageSubscriptionsViewModel: ObservableObject {
     @Published
     private(set) var refundRequestStatusMessage: String?
 
-    private var purchasesProvider: ManageSubscriptionsPurchaseType
+    private let purchasesProvider: ManageSubscriptionsPurchaseType
```
vegaro, line 52: Oopsies, I just realized this should be `ManageSubscriptionsPurchasesType`. I can fix it in another PR

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionHandler.swift
vegaro: Yes, I think that should be a `Delegate`. But maybe @aboedo can seal the deal since he has more iOS experience

RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
vegaro: ohhh nice

RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
vegaro: fantastic

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionHandler.swift
tonidero: Done in https://github.com/RevenueCat/purchases-ios/pull/4057/commits/da03b7c1b275eb29081bd312dde155194f36e5af

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
tonidero: Talked about this and we are going to unify the different abstractions into a single one for CustomerCenter, since we will need a few, and it was getting complex. However, will do this on a separate PR.

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionHandler.swift
tonidero: Hmm actually swiftlint is complaining with the new name: `warning: Class Delegate Protocol Violation: Delegate protocols should be class-only so they can be weakly referenced (class_delegate_protocol)`. I guess this means we should mark it as `@objc` which would make it so we can't implement it on structs like this... I'm wondering whether to just revert the rename or follow the convention specified by swiftlint...

tonidero:
Changed to a new approach where there is a single lambda parameter for the `CustomerCenterView` that receives an action  parameter which is an enum. Lmk what you think!

RevenueCatUI/CustomerCenter/Data/CustomerCenterAction.swift
```diff
@@ -0,0 +1,22 @@
...
+    case restoreFailed(_ error: Error)
+    /// Restore completed successfully
+    case restoreCompleted(_ customerInfo: CustomerInfo)
+    /// Going to display manage subscription page, whether for cancellation or changing plans.
+    case showingManageSubscriptions
+    /// Starting refund request process
+    case refundRequestStarted(_ productId: String)
+    /// Refund request process finished, with result provided.
+    case refundRequestCompleted(_ refundRequestStatus: RefundRequestStatus)
+}
```
vegaro, line 22: missing some lines at the top and at the bottom

```suggestion
public enum CustomerCenterAction {

    /// Purchase of a promotional offer is completed.
    case purchaseCompleted(_ customerInfo: CustomerInfo)
    /// Starting the restoration process
    case restoreStarted
    /// Restore errored out
    case restoreFailed(_ error: Error)
    /// Restore completed successfully
    case restoreCompleted(_ customerInfo: CustomerInfo)
    /// Going to display manage subscription page, whether for cancellation or changing plans.
    case showingManageSubscriptions
    /// Starting refund request process
    case refundRequestStarted(_ productId: String)
    /// Refund request process finished, with result provided.
    case refundRequestCompleted(_ refundRequestStatus: RefundRequestStatus)

}
```

RevenueCatUI/CustomerCenter/Data/CustomerCenterAction.swift
```diff
@@ -0,0 +1,22 @@
+import RevenueCat
+
+/// Typealias for handler for Customer center actions
+public typealias CustomerCenterActionHandler = @MainActor @Sendable (CustomerCenterAction) -> Void
+
+/// Represents an event the customer may perform during the Customer Center flow
+public enum CustomerCenterAction {
+    /// Purchase of a promotional offer is completed.
+    case purchaseCompleted(_ customerInfo: CustomerInfo)
```
vegaro, line 9: should we be more explicit here about it being a promo offer? Maybe we can add data about the product that was purchased and the offer?

vegaro: Since this event is not being triggered, feel free to leave this one out of the PR if you want, I feel we might need to define events around feedback surveys and promo offers a bit better

vegaro (approved):

RevenueCatUI/CustomerCenter/Data/CustomerCenterAction.swift
tonidero: Yeah I think I will remove it for now 👍. As for it being a promo offer/data of what's purchased, I think we can add that as well once we add support for that event.

PR #4058, opened 2024-07-16
Fix certain completion blocks not being dispatched on the main thread

MarkVillacampa:
### Checklist


Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
```diff
@@ -60,9 +60,15 @@ class TrialOrIntroPriceEligibilityChecker: TrialOrIntroPriceEligibilityCheckerTy
             return
         }
 
+        let completionBlock: ReceiveIntroEligibilityBlock = { result in
+            self.operationDispatcher.dispatchOnMainActor {
+                completion(result)
+            }
+        }
```
MarkVillacampa, line 67: Extracting and wrapping the completion block from the async call was the only way I found not to have to mark `ReceiveIntroEligibilityBlock` as `@Sendable` up to the public API thus making a breaking change.

Sources/Error Handling/Assertions.swift
```diff
@@ -44,3 +44,21 @@ func RCTestAssertNotMainThread(
...
 }
+
+@inline(__always)
+func RCTestAssertMainThread(
+    function: StaticString = #function,
+    file: StaticString = #file,
+    line: UInt = #line
+) {
+    #if DEBUG
+    if ProcessInfo.isRunningIntegrationTests {
```
MarkVillacampa, line 55: Only running the assertions when running integration tests, because on unit tests a lot of the dependencies are mocked, and mocks do not always execute completions on the main thread causing false positives.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -269,12 +269,16 @@ final class PurchasesOrchestrator {
...
+                completion(.failure(ErrorUtils.productDiscountMissingIdentifierError()))
+            }
             return
         }
 
         guard let subscriptionGroupIdentifier = product.subscriptionGroupIdentifier else {
-            completion(.failure(ErrorUtils.productDiscountMissingSubscriptionGroupIdentifierError()))
+            self.operationDispatcher.dispatchOnMainActor {
+                completion(.failure(ErrorUtils.productDiscountMissingSubscriptionGroupIdentifierError()))
+            }
```
aboedo, line 281: guess these were relying on the fact that no async operations have taken place yet so it'd have just called completion from the calling thread

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -284,14 +288,18 @@ final class PurchasesOrchestrator {
...
             }
         } else {
                 self.sk1PromotionalOffer(forProductDiscount: productDiscount,
                                          discountIdentifier: discountIdentifier,
                                          product: product,
                                          subscriptionGroupIdentifier: subscriptionGroupIdentifier) { result in
-                    completion(result)
+                    self.operationDispatcher.dispatchOnMainActor {
+                        completion(result)
+                    }
```
aboedo, line 302: We're dispatching on main actor for some and on main thread on the other, is there a difference / preference?

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1504,11 +1514,13 @@ extension PurchasesOrchestrator {
...
                        initiationSource: ProductRequestData.InitiationSource) async throws -> CustomerInfo {
-        return try await Async.call { completion in
+        return try await withCheckedThrowingContinuation { continuation in
             self.syncPurchases(receiptRefreshPolicy: receiptRefreshPolicy,
                                isRestore: isRestore,
-                               initiationSource: initiationSource,
-                               completion: completion)
+                               initiationSource: initiationSource) { result in
+                RCTestAssertMainThread()
+                continuation.resume(with: result)
```
MarkVillacampa, line 1522: This method is only used in integration tests. Modified it to be able to call the assertion. Will revert back once tests pass.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
aboedo: not sure there are any guarantees for order of operation between operation added from queue vs code in actor but I'd imagine not

Tests/UnitTests/Mocks/MockOperationDispatcher.swift
```diff
@@ -12,7 +12,7 @@ class MockOperationDispatcher: OperationDispatcher {
     var invokedDispatchOnMainThread = false
     var invokedDispatchOnMainThreadCount = 0
     var shouldInvokeDispatchOnMainThreadBlock = true
-    var forwardToOriginalDispatchOnMainThread = false
+    var forwardToOriginalDispatchOnMainThread = true
```
MarkVillacampa, line 15: Changed to the MockOperationDispatcher would actually dispatch to the main thread in tests.

Will revert back.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1504,11 +1514,13 @@ extension PurchasesOrchestrator {
     func syncPurchases(receiptRefreshPolicy: ReceiptRefreshPolicy,
                        isRestore: Bool,
                        initiationSource: ProductRequestData.InitiationSource) async throws -> CustomerInfo {
-        return try await Async.call { completion in
+        return try await withCheckedThrowingContinuation { continuation in
```
aboedo, line 1517: this seems simpler. Wonder if we should have an alternative to withCheckedThrowingContinuation that always resumes with an assert on main thread

aboedo (approved):

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
MarkVillacampa: Yep exactly

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
MarkVillacampa: No reason to use GCD methods anymore, converted them to actor.

I guess technically there is no difference since we're just dispatching to the main thread in both cases.

MarkVillacampa: Longer term that's the way to go: get completely rid of all the GCD stuff, switch to modern Swift Concurrency, and also adopt the Swift 6 concurrency model.

fire-at-will (approved):

PR #4072, opened 2024-07-17
[Customer Center] Improves the UI of the current subscription

JayShortway:
## Summary
This PR improves the presentation of the current subscription.
- Adds hierarchy to the various texts.
- Uses padding to indicate which texts belong together and which do not.
- Adds some descriptive icons.
- Adds a short explanation as to which subscription we're showing. 
- Describes the date as the "next billing date" (or date of expiry, if applicable). 
- Wraps it in a card.

This PR can be seen as a proposal. Feel free to add any suggestions! 

## Before

image

## After

image

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
+++ b/RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
@@ -20,11 +20,11 @@ struct SubscriptionInformation {
     let title: String
     let durationTitle: String
     let price: String
-    let nextRenewalString: String?
+    let expirationDateString: String?
     let productIdentifier: String
 
-    var renewalString: String {
-        return active ? (willRenew ? "Renews" : "Expires") : "Expired"
+    var expirationString: String {
+        return active ? (willRenew ? "Next billing date" : "Expires") : "Expired"
     }
 
     private let willRenew: Bool
@@ -33,7 +33,7 @@ struct SubscriptionInformation {
     init(title: String,
          durationTitle: String,
          price: String,
-         nextRenewalString: String?,
+         expirationDateString: String?,
          willRenew: Bool,
          productIdentifier: String,
          active: Bool
@@ -41,7 +41,7 @@ struct SubscriptionInformation {
         self.title = title
         self.durationTitle = durationTitle
         self.price = price
-        self.nextRenewalString = nextRenewalString
+        self.expirationDateString = expirationDateString
         self.productIdentifier = productIdentifier
         self.willRenew = willRenew
         self.active = active
--- a/RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
+++ b/RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
@@ -30,8 +30,8 @@ struct SubscriptionInformation {
     var explanation: String {
         return active ? (
             willRenew ?
-                "This is your subscription with the soonest billing date." :
-                "This is your subscription with the soonest expiration date."
+                "This is your subscription with the earliest billing date." :
+                "This is your subscription with the earliest expiration date."
         ) : "This subscription has expired."
     }
 
--- a/RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
+++ b/RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
@@ -27,6 +27,14 @@ struct SubscriptionInformation {
         return active ? (willRenew ? "Next billing date" : "Expires") : "Expired"
     }
 
+    var explanation: String {
+        return active ? (
+            willRenew ?
+                "This is your subscription with the soonest billing date." :
+                "This is your subscription with the soonest expiration date."
+        ) : "This subscription has expired."
+    }
+
     private let willRenew: Bool
     private let active: Bool
 
--- a/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
@@ -243,10 +243,18 @@ struct ManageSubscriptionsButtonsView: View {
 struct ManageSubscriptionsView_Previews: PreviewProvider {
 
     static var previews: some View {
-        let viewModel = ManageSubscriptionsViewModel(
+        let viewModelMonthlyRenewing = ManageSubscriptionsViewModel(
             screen: CustomerCenterConfigTestData.customerCenterData.screens[.management]!,
-            subscriptionInformation: CustomerCenterConfigTestData.subscriptionInformation)
-        ManageSubscriptionsView(viewModel: viewModel)
+            subscriptionInformation: CustomerCenterConfigTestData.subscriptionInformationMonthlyRenewing)
+        ManageSubscriptionsView(viewModel: viewModelMonthlyRenewing)
+            .previewDisplayName("Monthly renewing")
+
+        let viewModelYearlyExpiring = ManageSubscriptionsViewModel(
+            screen: CustomerCenterConfigTestData.customerCenterData.screens[.management]!,
+            subscriptionInformation: CustomerCenterConfigTestData.subscriptionInformationYearlyExpiring)
+
+        ManageSubscriptionsView(viewModel: viewModelYearlyExpiring)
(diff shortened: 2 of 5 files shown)


RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
```diff
@@ -20,11 +20,19 @@ struct SubscriptionInformation {
...
-        return active ? (willRenew ? "Renews" : "Expires") : "Expired"
+    var expirationString: String {
+        return active ? (willRenew ? "Next billing date" : "Expires") : "Expired"
+    }
+
+    var explanation: String {
+        return active ? (
+            willRenew ?
+                "This is your subscription with the soonest billing date." :
+                "This is your subscription with the soonest expiration date."
```
tonidero, line 34: Maybe double-check but I think these should be `earliest`?
```suggestion
                "This is your subscription with the earliest billing date." :
                "This is your subscription with the earliest expiration date."
```

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -142,20 +142,51 @@ struct SubscriptionDetailsView: View {
...
-                .foregroundColor(Color.gray)
-                .padding(.horizontal)
-
-            if let nextRenewal =  subscriptionInformation.nextRenewalString {
-                Text("\(subscriptionInformation.renewalString): \(String(describing: nextRenewal))")
+            VStack(alignment: .leading) {
+                Text("\(subscriptionInformation.title)")
+                    .font(.headline)
+                Text("\(subscriptionInformation.explanation)")
+                    .frame(maxWidth: 200, alignment: .leading)
```
tonidero, line 149: Hmm is setting a `maxWidth` necessary? Not sure if it would be better to set a percentage max width or something like that if it's necessary

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -142,20 +142,51 @@ struct SubscriptionDetailsView: View {
...
                     .font(.caption)
                     .foregroundColor(Color.gray)
-                    .padding([.horizontal, .bottom])
+            }.padding([.bottom], 10)
+
+            HStack(alignment: .center) {
+                Image(systemName: "coloncurrencysign.arrow.circlepath")
+                    .frame(width: 22)
```
tonidero, line 156: Since this is repeated, I would extract to a constant that can be reused.

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -142,20 +142,51 @@ struct SubscriptionDetailsView: View {
...
+                    .font(.headline)
+                Text("\(subscriptionInformation.explanation)")
+                    .frame(maxWidth: 200, alignment: .leading)
                     .font(.caption)
                     .foregroundColor(Color.gray)
-                    .padding([.horizontal, .bottom])
+            }.padding([.bottom], 10)
+
+            HStack(alignment: .center) {
+                Image(systemName: "coloncurrencysign.arrow.circlepath")
```
tonidero, line 155: I guess we should do a design review later, but I think these icons look great 🙌

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -142,20 +142,51 @@ struct SubscriptionDetailsView: View {
...
+                    .frame(width: 22)
+                VStack(alignment: .leading) {
+                    Text("Billing cycle")
```
tonidero, line 158: I guess we will need to extract all these strings so they are provided by the backend and used here, but that can come in a separate PR.

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
tonidero: We might want to hide these images from accessibility readers with `.accessibilityHidden(true)`. But then again, we might want to address accessibility later.

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -165,7 +196,11 @@ struct SubscriptionDetailsView: View {
                     .foregroundColor(Color.gray)
                     .padding([.horizontal, .bottom])
             }
-        }
+        }.padding()
+            .padding(.horizontal)
```
tonidero, line 200: Hmm this looks weird to me. Are we adding `padding` twice?

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -165,7 +196,11 @@ struct SubscriptionDetailsView: View {
...
+            .background(Color.white)
```
tonidero, line 201: Just checking but does this work well in dark mode?

tonidero (commented):
Looks amazing! Just some comments but nothing major

RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
JayShortway: Yes! I already didn't like "soonest" haha, and also "the earliest" is [more common](https://books.google.com/ngrams/graph?content=the+soonest%2Cthe+earliest&year_start=1900&year_end=2019&corpus=en-2019&smoothing=20).

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
JayShortway: Without this `maxWidth`, this text would cause the outer card to span the entire width of the (phone) screen, which I think doesn't look great. 

I'd be okay setting it to a relative size, but I haven't found an easy way to do so. Some suggest wrapping it all in a `GeometryReader`, or using a custom `Layout`. Let me know if you know another way! 

I did check `maxWidth: 200` on an iPhone SE, and that still looks fine.

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
JayShortway: We are adding a bit more padding horizontally than we do vertically. (The first `padding()` adds padding to all sides.) Afaik this is common in layouts to have a bit more breathing room on the sides. 

The reason this is not done in 1 go is because I don't know what the default padding is that's added by `padding()`.  The docs just say "a default amount of padding". 🤷

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
JayShortway: Good one! Done: https://github.com/RevenueCat/purchases-ios/pull/4072/commits/65ecb16c3cf097ab304d031c6eda0febf0d1a9a2

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
JayShortway: No this is a very good point! I'll do it now otherwise it'll risk being forgotten: https://github.com/RevenueCat/purchases-ios/pull/4072/commits/1ad91a0dacaeb529de1ce5e981ffa96ec5db48af

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
JayShortway: Good point, it doesn't haha. I changed them to `UIColor`s which adapt to the color scheme: https://github.com/RevenueCat/purchases-ios/pull/4072/commits/461a28b59b69b304f50a52c019474fe157144eac

tonidero (approved):
LGTM!

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
JayShortway: As discussed on Slack, this will be done in a separate PR.

PR #4073, opened 2024-07-17
Only checks staged files for leftover API keys.

JayShortway:
## What
As the title says. It does a rudimentary intersect of the existing `FILES_TO_CHECK` and the staged files, and checks _those_ files for leftover API keys. 

## Why
This allows you to set API keys in any of the `FILES_TO_CHECK` files, as long as you're not trying to commit them. Otherwise you'd have to revert your local API key every time you'd want to commit anything.


tonidero (approved):
Nice improvement! 🙌

PR #4078, opened 2024-07-17
Use newer Alert API for showing customer restored alert

MarkVillacampa:
This was causing issues for some developers, because the old, deprecated Alert API would fail silently if there was a view higher up in the view hierarchy which already had an Alert modifier

This in turn caused the `onRestoreCompleted` callback not to be called.

https://community.revenuecat.com/sdks-51/restore-completion-handlers-not-firing-in-presentpaywallifneeded-4674

The newer Alert API seems to not suffer from the same issue.

### Checklist

### DIFF
--- a/RevenueCatUI/Views/FooterView.swift
+++ b/RevenueCatUI/Views/FooterView.swift
@@ -184,6 +184,9 @@ private struct RestorePurchasesButton: View {
     @State
     private var restoredCustomerInfo: CustomerInfo?
 
+    @State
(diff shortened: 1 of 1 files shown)


aboedo (approved):

vegaro (approved):

FireLord:
its till a bug. i cant directly switch to v5 as of now. tried 4.42.0 still cant get the restore thing to work.

PR #4079, opened 2024-07-17
Fix release train deploy workflow

aboedo:
The deploy workflow didn't run correctly, because of a regression introduced in #4025 . 

The problem is that we replaced the tag filters with a `when` clause, but the job itself was relying on us using the filter and running the second part of the deploy workflow from the tag itself and not the branch. 

I.e.: it wasn't entirely a chain of jobs, rather, there were two separate chains, one for releases and one for tags, running in the same job. 

This PR splits the jobs into two parts: one that creates the tag, and one that fires when a new tag is pushed, to create a release for it

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1353,24 +1353,23 @@ workflows:
           requires:
             - hold
       - make-release:
-          <<: *release-tags
+          requires:
+            - hold
       - push-revenuecat-pod:
           requires:
             - make-release
-          <<: *release-tags
       - push-revenuecatui-pod:
           requires:
-            - make-release
             - push-revenuecat-pod
-          <<: *release-tags
       - deploy-to-spm:
           requires:
             - make-release
-          <<: *release-tags
       - docs-deploy:
-          <<: *release-tags
+          requires:
+            - make-release
       - deploy-purchase-tester:
-          <<: *release-tags
+          requires:
+            - make-release
       - notify-on-non-patch-release-branches:
           requires:
             - make-release
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1335,7 +1335,7 @@ workflows:
       - run-test-ios-17
       - pod-lib-lint
 
-  deploy:
+  create-tag:
     when:
       and:
         - not:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1353,24 +1353,23 @@ workflows:
           requires:
             - hold
       - make-release:
-          <<: *release-tags
```
aboedo, line 1356: removing the filters and adding the requirement should mean that this runs when the hold step finishes

.circleci/config.yml
```diff
@@ -1353,24 +1353,23 @@ workflows:
...
+          requires:
+            - hold
       - push-revenuecat-pod:
           requires:
             - make-release
-          <<: *release-tags
```
aboedo, line 1360: from here on it's pretty much the same thing, just removing the tag filter since we're already filtering at the `when` level

.version
```diff
@@ -1 +1 @@
-5.1.0
+5.2.0
```
aboedo: what, reverting

CHANGELOG.latest.md
```diff
@@ -1,14 +1,9 @@
 ### New Features
-* Paywalls with custom purchase and restore logic handlers (#3973) via James Borthwick (@jamesrb1)
```
aboedo: what

.circleci/config.yml
```diff
@@ -1335,7 +1335,7 @@ workflows:
       - run-test-ios-17
       - pod-lib-lint
 
-  deploy:
```
aboedo, line 1338: probably easier to read in a side-by-side diff

.circleci/config.yml
```diff
@@ -1345,32 +1345,44 @@ workflows:
...
+      and:
+        - equal: [push, << pipeline.trigger_source >>]
+        - exists: << pipeline.git.tag >>
+        - not:
+          matches:
+            pattern: ".*-SNAPSHOT$"
+            value: << pipeline.git.tag >>
+    jobs:
+      - hold-for-release:
+          type: approval
```
aboedo, line 1367: is having a second hold here too much?

aboedo: Kinda feel like we have to have _something_, otherwise any accidental push of a tag in a branch that has a PR triggers a release

joshdholtz (approved):
I think this looks good! 👍 Ship it ship it

PR #4080, opened 2024-07-17
Fix release release trains part 2

aboedo:
Follow-up to https://github.com/RevenueCat/purchases-ios/pull/4079, removes the `when` and re-introduces filters as the way to trigger the job since `when` doesn't seem to work for tags


joshdholtz (approved):
:shipit:

PR #4081, opened 2024-07-18
Fix docs deploy and add manual trigger on CI

joshdholtz:
### Motivation

Fix failing docs

### Description

When `Dir.mktmpdir` uses a block, the temp dir is deleted at the end of the block. Not using the block syntax anymore so the temp dir lives the entire process.

Also adds `run-manual-docs-deploy` action to trigger docs manually for easy testing since Xcode 14 can't be run on macOS Sonoma.

#### Proof

Docs build and published in https://github.com/RevenueCat/purchases-ios-docs/commit/56b09af83637a6e40450062e805849a04ade6043

Screenshot 2024-07-17 at 8 14 05 PM

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -9,7 +9,7 @@ version: 2.1
 parameters:
   action:
     type: enum
-    enum: [default, bump, run-manual-tests]
+    enum: [default, bump, run-manual-tests, run-manual-docs-deploy]
     default: default
   generate_snapshots:
     default: false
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -9,7 +9,7 @@ version: 2.1
 parameters:
   action:
     type: enum
-    enum: [default, bump, run-manual-tests]
+    enum: [default, bump, run-manual-tests, run-manual-docs-deploy]
```
aboedo, line 12: ❤️

aboedo (approved):

PR #4082, opened 2024-07-18
[Customer Center] Updates the presentation of the refund status

JayShortway:
As the title says. 

## Before
Scherm­afbeelding 2024-07-18 om 15 49 00

## After
Scherm­afbeelding 2024-07-18 om 15 50 56

### DIFF
--- a/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
@@ -66,10 +66,12 @@ class ManageSubscriptionsViewModel: ObservableObject {
     }
 
     init(screen: CustomerCenterConfigData.Screen,
-         subscriptionInformation: SubscriptionInformation) {
+         subscriptionInformation: SubscriptionInformation,
+         refundRequestStatusMessage: String? = nil) {
         self.screen = screen
         self.subscriptionInformation = subscriptionInformation
         self.purchasesProvider = ManageSubscriptionPurchases()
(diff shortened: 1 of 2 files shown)


RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -66,10 +66,12 @@ class ManageSubscriptionsViewModel: ObservableObject {
     }
 
     init(screen: CustomerCenterConfigData.Screen,
-         subscriptionInformation: SubscriptionInformation) {
+         subscriptionInformation: SubscriptionInformation,
+         refundRequestStatusMessage: String? = nil) {
```
JayShortway, line 70: I wasn't entirely sure about this, but from what I can tell this constructor is only used in previews so I thought it'd be okay. Happy to revert of course.

tonidero (approved):

vegaro (approved):

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
vegaro: yeah it's fine, these constructors for previews are always very not-ideal

tonidero: Though this just now but we could wrap this constructor in a `#if DEBUG` macro, so we make sure it's not used in release builds... NABD though.

PR #4083, opened 2024-07-18
[Customer Center] SubscriptionDetailsView gets its strings from CommonLocalizedString.

JayShortway:
As the title says. 

- The `SubscriptionDetailsView` gets its strings from `CommonLocalizedString`.
- Moves the strings out of `SubscriptionInformation`. 
    - This avoids `SubscriptionInformation` having to care about localization, and having to add `@available` annotations, since it's only available on iOS 15 and up. 
    - What (text) to display for which state is arguably more of a UI/presentation concern.


RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -148,7 +148,16 @@ struct SubscriptionDetailsView: View {
...
                 Text("\(subscriptionInformation.title)")
                     .font(.headline)
-                Text("\(subscriptionInformation.explanation)")
+
+                var explanation: String {
+                    return subscriptionInformation.active ? (
+                        subscriptionInformation.willRenew ?
+                            String(localized: "This is your subscription with the earliest billing date.") :
+                            String(localized: "This is your subscription with the earliest expiration date.")
+                    ) : String(localized: "This subscription has expired.")
```
tonidero, line 157: Afaik, the plan is to get all the strings we can from the backend, to make sure everything is using the same locale (considering the backend will only send a single locale). So I would probably move these strings to be obtained from the `Localization` struct in the customer center data instead. Happy to discuss it more over Slack though!

JayShortway: Ah yes, that makes sense. I was definitely counting on this possibility. Will refactor.

JayShortway: This is now done: https://github.com/RevenueCat/purchases-ios/pull/4083/commits/142ab1db2a822940a37d31366e495ed5b747ab80.

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -145,14 +146,24 @@ struct SubscriptionDetailsView: View {
...
                     .font(.headline)
-                Text("\(subscriptionInformation.explanation)")
+
+                var explanation: String {
+                    return subscriptionInformation.active ? (
+                        subscriptionInformation.willRenew ?
+                        localization.commonLocalizedString(for: .subEarliestRenewal) :
+                            localization.commonLocalizedString(for: .subEarliestExpiration)
+                    ) : localization.commonLocalizedString(for: .subExpired)
+                }
```
tonidero, line 164: NABD but any reason to keep this as a computed property? Or could we simplify it like this
```suggestion
                var explanation = subscriptionInformation.active ? (
                     subscriptionInformation.willRenew ?
                        localization.commonLocalizedString(for: .subEarliestRenewal) :
                            localization.commonLocalizedString(for: .subEarliestExpiration)
                    ) : localization.commonLocalizedString(for: .subExpired)
```

tonidero (approved):
Looks great!

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
JayShortway: No reason! I even asked ChatGPT if this was idiomatic SwiftUI haha. Thanks for the suggestion! Please keep em coming.

I'll do this here and for `expirationString` below.

JayShortway: This is now done: https://github.com/RevenueCat/purchases-ios/pull/4083/commits/2fc4bb667ed0daf5a80f0cab0b72161ed0999b6c

PR #4084, opened 2024-07-19
Fix workflow permission check logic

tonidero:
### Description
This improves the permission checking logic for running github actions. We also have it setup to only run actions by organization members, but this acts as an additional check.

### DIFF
--- a/.github/workflows/trigger_all_tests.yml
+++ b/.github/workflows/trigger_all_tests.yml
@@ -9,7 +9,8 @@ jobs:
     runs-on: ubuntu-latest
     if: |
       ${{ github.event.issue.pull_request }} &&
-      github.event.comment.body == '@RCGitBot please test'
+      github.event.comment.body == '@RCGitBot please test' &&
+      github.repository == 'RevenueCat/purchases-ios'
 
     steps:
       - name: Check membership in RevenueCat Org
--- a/.github/workflows/trigger_all_tests.yml
+++ b/.github/workflows/trigger_all_tests.yml
@@ -29,6 +29,6 @@ jobs:
 
         uses: CircleCI-Public/trigger-circleci-pipeline-action@v1.2.0
         with:
-          GHA_Action: "run-from-github-comments"
+          GHA_Meta: "run-from-github-comments"
         env:
           CCI_TOKEN: ${{ secrets.CIRCLECI_TOKEN }}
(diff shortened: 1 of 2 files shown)


tonidero:
@RCGitBot please test

tonidero:
@RCGitBot please test

tonidero:
@RCGitBot please test

.github/workflows/trigger_all_tests.yml
```diff
@@ -13,11 +13,13 @@ jobs:
```
JayShortway, line 13: Maybe we can add another `if` (right above this line 😅) to make sure this doesn't work on forks? 

```yaml
if: github.repository == 'RevenueCat/purchases-ios'
```

vegaro: I think that check could also be removed in the fork though? but yes, another extra check I guess

JayShortway: Yea I think so too 🤔 But I see this being done, e.g. here: https://github.com/cashapp/multiplatform-paging/blob/main-3.3.0-alpha02/.github/workflows/release.yaml#L25

tonidero: yeah... will add it since I think it shouldn't hurt in any case 👍

JayShortway (approved):
LGTM!

tonidero:
@RCGitBot please test

tonidero:
Will merge this for now, we can figure any other issues in followup PRs

.github/workflows/trigger_all_tests.yml
```diff
@@ -9,15 +9,18 @@ jobs:
...
     steps:
       - name: Check membership in RevenueCat Org
+        env:
+          READ_ORG_GITHUB_TOKEN: ${{ secrets.READ_ORG_GITHUB_TOKEN }}
         id: verify
         # ensure that only RevenueCat members can trigger this
         run: |
-          RESPONSE=$(curl https://api.github.com/orgs/RevenueCat/members/${{ github.event.comment.user.login }})
-          if [[ "$RESPONSE" == *"Not Found"* ]]; then
+          RESPONSE=$(curl -s -o /dev/null --head -w "%{http_code}" -H "Authorization: Bearer $READ_ORG_GITHUB_TOKEN" https://api.github.com/orgs/RevenueCat/members/${{ github.event.comment.user.login }})
```
aboedo, line 22: we should probably add a comment and maybe use `\` to split this up into lines for clarity

aboedo (commented):
thanks for taking care of it!

vegarotest:
@RCGitBot please test

PR #4085, opened 2024-07-19
Fix prepare next version job

tonidero:
### Description
I noticed the `prepare-next-version` job was failing due to missing github credentials. This should fix that.


JayShortway (approved):
🫡

PR #4022, opened 2024-07-05
Added new paywall template to support multiple tiered subscriptions

joshdholtz:
### Motivation

Template 5 but with multiple tiers 😊

### Description

Enabling Template 7 by changing `7_disabled` to `7`

- Changed a `header.jpg` to `header.heic` to try and fix issue with weirdly failing snapshot UI test
- Changed visibility on `TestData` from `public` to `internal`
- Hides tier selector if only one tier
- Changed `color_overrides_by_tier` to `colors_by_tier` and `image_overrides_by_tier` to `images_by_tier`
  - The original implementation was going to default to existing color and images configurations and then override by tier
  - However, this turned out to be harder to implement on the configuration side
- `Colors` got four new optional values
  - `tier_control_background`
  - `tier_control_foreground`
  - `tier_control_selected_background`
  - `tier_control_selected_foreground`

#### Full Screen

https://github.com/RevenueCat/purchases-ios/assets/401294/e5317a53-428c-415e-a1a1-b4231d32a776

### Footer

https://github.com/RevenueCat/purchases-ios/assets/401294/8905b55e-0fa2-49f0-aa1c-d4c1e694fb5b

### Footer Condensed

https://github.com/RevenueCat/purchases-ios/assets/401294/311d78fb-bdeb-43cd-b601-9c60a8a54b6f

PR #4087, opened 2024-07-19
Add missing @PublicForExternalTesting that broke PaywallsTester

vegaro:
It looks like some annotations `PublicForExternalTesting` got removed in https://github.com/RevenueCat/purchases-ios/pull/4022/files#diff-c458d88e52633024e0da56833bdfd34e2790a17f15a9b93ff75dfd7aec02a168, which prevented `PaywallsTester` from archiving

It was causing these errors https://github.com/RevenueCat/purchases-ios/runs/27655512796

### DIFF
--- a/Tests/TestingApps/PaywallsTester/PaywallsTester/Config/SamplePaywalls.swift
+++ b/Tests/TestingApps/PaywallsTester/PaywallsTester/Config/SamplePaywalls.swift
@@ -9,7 +9,7 @@
 
 import Foundation
 import RevenueCat
-@testable import RevenueCatUI
+import RevenueCatUI
 
 import UIKit
 
--- a/RevenueCatUI/Data/TestData.swift
+++ b/RevenueCatUI/Data/TestData.swift
@@ -15,6 +15,7 @@ import SwiftUI
 // @PublicForExternalTesting
 enum TestData {
 
+    // @PublicForExternalTesting
     static let customerInfo: CustomerInfo = {
         return .decode(
         """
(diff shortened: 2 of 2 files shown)


Tests/TestingApps/PaywallsTester/PaywallsTester/Config/SamplePaywalls.swift
```diff
@@ -9,7 +9,7 @@
 
 import Foundation
 import RevenueCat
-@testable import RevenueCatUI
```
tonidero, line 12: Hmm I guess if we're adding/removing the `public` on every build this won't be needed yeah... And it would break uploads to app store connect I imagine.

tonidero (approved):
I believe

Tests/TestingApps/PaywallsTester/PaywallsTester/Config/SamplePaywalls.swift
vegaro: yes, that's what was breaking the upload

vegaro:
@tonidero this is still not fixed, because if you notice, most of the stuff in `TestData` is conditionally compiled on `DEBUG`... I believe the only solution at the moment is to remove the check for pretty much everything in `TestData`

cc: @joshdholtz since this is related to the multi tier PR

joshdholtz:
> @tonidero this is still not fixed, because if you notice, most of the stuff in `TestData` is conditionally compiled on `DEBUG`... I believe the only solution at the moment is to remove the check for pretty much everything in `TestData`
> 
> cc: @joshdholtz since this is related to the multi tier PR

@vegaro @tonidero Oh no! I can take a look at fixing it since I broke it 💪

PR #4089, opened 2024-07-19
Move No subscriptions page strings to Localization

tonidero:
### Description
This is based on the changes in #3968 

This PR:
- Moves the strings currently hardcode into the `CustomerCenterConfigData.Localization` object.
- Fixes an issue compiling paywall tester introduced in a previous PR
- Modifies how we read backend strings to not account for the `common_` prefix, which is already removed by the backend.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
@@ -32,6 +32,9 @@ struct NoSubscriptionsView: View {
     @Environment(\.dismiss)
     var dismiss
 
+    @Environment(\.localization)
+    private var localization: CustomerCenterConfigData.Localization
+
     @State
     private var showRestoreAlert: Bool = false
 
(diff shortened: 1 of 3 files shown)


Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -42,18 +42,30 @@ public struct CustomerCenterConfigData {
...
+                case .cancel:
+                    return "Cancel"
                 }
             }
 
         }
 
         public func commonLocalizedString(for key: CommonLocalizedString) -> String {
-            return self.localizedStrings["common_\(key.rawValue)"] ?? key.defaultValue
+            return self.localizedStrings[key.rawValue] ?? key.defaultValue
```
tonidero, line 68: This prefix is removed by the backend in the response, so we don't need to use it here.

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
```diff
@@ -215,8 +215,6 @@ extension SamplePaywallsList {
         switch action {
         case .restoreCompleted(_):
             print("CustomerCenter: restoreCompleted")
-        case .purchaseCompleted(_):
-            print("CustomerCenter: purchaseCompleted")
```
tonidero, line 219: This was a leftover from a last minute change in #4057 which was causing paywall tester not to build.

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
```diff
@@ -41,25 +44,26 @@ struct NoSubscriptionsView: View {
...
-                .padding()
-            Text("We can try checking your Apple account for any previous purchases")
-                .font(.body)
-                .padding()
-
-            Spacer()
-
-            Button("Restore purchases") {
-                showRestoreAlert = true
+            if let localization {
```
tonidero, line 47: At this point `localization` should never be null... We could potentially inject a non-nullable parameter in #3968 @vegaro so we don't need to do this here, since at this point, the configuration from the backend is always loaded. Wdyt?

vegaro (approved):

PR #4092, opened 2024-07-19
Fix PaywallsTester by changing TestData DEBUG checks

vegaro:
Round 2 on https://github.com/RevenueCat/purchases-ios/pull/4087. There were still some stuff not being available to `PaywallsTester` on archive


joshdholtz (approved):
I think this makes sense!

PR #4095, opened 2024-07-22
Add Is-Retryable Header

fire-at-will:
This PR adds the ability for the server to disable retries on specific requests by setting the `Is-Retryable` response header on a request to `false`. If the header is missing or is `true`, the SDK will retry normally.

If the server sends `Is-Retryable: "true"` for a response code that the SDK isn't configured to retry, the SDK will not retry the request.

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -668,7 +668,7 @@ extension HTTPClient {
 
     internal func isRetryable(_ urlResponse: HTTPURLResponse) -> Bool {
         let isStatusCodeRetryable = self.retriableStatusCodes.contains(urlResponse.httpStatusCode)
-        let isRetryableString = urlResponse.allHeaderFields[ResponseHeader.isRetryable.rawValue] as? String
+        let isRetryableString = urlResponse.value(forHTTPHeaderField: ResponseHeader.isRetryable.rawValue)
         let isRetryable: Bool
         if let isRetryableString {
             isRetryable = Bool(isRetryableString.lowercased()) ?? true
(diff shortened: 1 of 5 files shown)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -665,8 +666,17 @@ extension HTTPClient {
         return true
     }
 
-    internal func isStatusCodeRetryable(_ statusCode: HTTPStatusCode) -> Bool {
-        return self.retriableStatusCodes.contains(statusCode)
+    internal func isRetryable(_ urlResponse: HTTPURLResponse) -> Bool {
+        let isStatusCodeRetryable = self.retriableStatusCodes.contains(urlResponse.httpStatusCode)
+        let isRetryableString = urlResponse.allHeaderFields[ResponseHeader.isRetryable.rawValue] as? String
```
fire-at-will, line 671: I cast this as a `String` because HTTPURLRequests' initializer only accepts strings as the header value:

```swift
public init?(url: URL, statusCode: Int, httpVersion HTTPVersion: String?, headerFields: [String : String]?)
```

Let me know if there's a better way to handle this!

MarkVillacampa: Maybe we could use:

https://developer.apple.com/documentation/foundation/httpurlresponse/3240613-value:

MarkVillacampa (approved):
Looking food!

aboedo (approved):
makes sense. 
This also got me thinking: 
- could we repurpose the existing retry after header, but set it to a negative, null, or really high value? 

which in turn got me thinking: 
- are we actually being safe enough retry after? I'm not sure we added a limit, but what if I set the server to retry after 10 years?

fire-at-will:
> are we actually being safe enough retry after? I'm not sure we added a limit, but what if I set the server to retry after 10 years?

@aboedo That definitely wouldn't be good! 😵‍💫 I'll make a tiny follow-up PR to cap the server-provided retry time at 1 hour.

fire-at-will:
@aboedo Follow up PR to cap the server-defined retry interval in the `Retry-After` header is here: https://github.com/RevenueCat/purchases-ios/pull/4096 💪

PR #4096, opened 2024-07-22
Server Provided Backoff Interval Validation

fire-at-will:
This PR performs some validations on the server-provided retry backoff period. Specifically, it:
- Defaults to a backoff interval of 0ms if `Retry-After` is negative
- Caps the backoff interval from the `Retry-After` header at a maximum of 1 hour.
- Adds tests to verify the behavior

### DIFF
--- a/Sources/Misc/Concurrency/OperationDispatcher.swift
+++ b/Sources/Misc/Concurrency/OperationDispatcher.swift
@@ -57,7 +57,7 @@ class OperationDispatcher {
         Self.dispatchOnMainActor(block)
     }
 
-    func dispatchOnWorkerThread(jitterableDelay delay: JitterableDelay = .none, block: @escaping @Sendable () -> Void) {
+    func dispatchOnWorkerThread(delay: JitterableDelay = .none, block: @escaping @Sendable () -> Void) {
         if delay.hasDelay {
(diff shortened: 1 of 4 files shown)


Sources/Misc/Concurrency/OperationDispatcher.swift
```diff
@@ -57,7 +57,7 @@ class OperationDispatcher {
         Self.dispatchOnMainActor(block)
     }
 
-    func dispatchOnWorkerThread(jitterableDelay delay: JitterableDelay = .none, block: @escaping @Sendable () -> Void) {
+    func dispatchOnWorkerThread(delay: JitterableDelay = .none, block: @escaping @Sendable () -> Void) {
```
fire-at-will, line 60: This accidentally got added in and was causing some issues, will address it in a follow-up PR

fire-at-will, line 60: I'm not sure what's going on with GitHub's diff rendering here, the `jitterableDelay` isn't present in the `retry-429` branch: https://github.com/RevenueCat/purchases-ios/blob/3b2018698e046295fc045c343ffb6021b88baa05/Sources/Misc/Concurrency/OperationDispatcher.swift#L60

PR #4098, opened 2024-07-22
Server Provided Backoff Interval Validation

fire-at-will:
This PR performs some validations on the server-provided retry backoff period. Specifically, it:

- Defaults to a backoff interval of 0ms if Retry-After is negative
- Caps the backoff interval from the Retry-After header at a maximum of 1 hour.
- Adds tests to verify the behavior


aboedo (approved):
looks good, but that's even more logic moved into httpclient now, we should take the time to extract it before we forget about it

PR #4099, opened 2024-07-22
Call AppTransaction.refresh() in certain scenarios if AppTransaction.shared is invalid

MarkVillacampa:
When `AppTransaction.shared` is empty or invalid, Apple recommends calling `AppTransaction.refresh().

https://developer.apple.com/documentation/storekit/apptransaction/4020517-refresh

Calling `refresh()` will always show an authentication prompt so we want to make sure we only call it in response to a user action

- After a purchase.
- When calling `restorePurchases`

### Checklist

### Motivation

`AppTransaction.shared` seems to be empty more often than expected. We're hoping refreshing it where applicable will improve the situation.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -223,14 +223,14 @@ final class PurchasesOrchestrator {
     }
 
     func restorePurchases(completion: (@Sendable (Result<CustomerInfo, PurchasesError>) -> Void)?) {
-        self.syncPurchases(receiptRefreshPolicy: .always,
+        self.syncPurchases(receiptRefreshAllowed: true,
                            isRestore: true,
                            initiationSource: .restore,
                            completion: completion)
     }
 
     func syncPurchases(completion: (@Sendable (Result<CustomerInfo, PurchasesError>) -> Void)? = nil) {
-        self.syncPurchases(receiptRefreshPolicy: .never,
+        self.syncPurchases(receiptRefreshAllowed: false,
                            isRestore: allowSharingAppStoreAccount,
                            initiationSource: .restore,
                            completion: completion)
@@ -1022,7 +1022,7 @@ private extension PurchasesOrchestrator {
         }
     }
 
-    func syncPurchases(receiptRefreshPolicy: ReceiptRefreshPolicy,
+    func syncPurchases(receiptRefreshAllowed: Bool,
                        isRestore: Bool,
                        initiationSource: ProductRequestData.InitiationSource,
                        completion: (@Sendable (Result<CustomerInfo, PurchasesError>) -> Void)?) {
@@ -1034,11 +1034,12 @@ private extension PurchasesOrchestrator {
 
         if self.systemInfo.storeKitVersion.isStoreKit2EnabledAndAvailable,
            #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) {
-            self.syncPurchasesSK2(isRestore: isRestore,
+            self.syncPurchasesSK2(refreshPolicy: receiptRefreshAllowed ? .onlyIfEmpty : .never,
+                                  isRestore: isRestore,
                                   initiationSource: initiationSource,
                                   completion: completion)
         } else {
-            self.syncPurchasesSK1(receiptRefreshPolicy: receiptRefreshPolicy,
+            self.syncPurchasesSK1(receiptRefreshPolicy: receiptRefreshAllowed ? .always : .never,
                                   isRestore: isRestore,
                                   initiationSource: initiationSource,
                                   completion: completion)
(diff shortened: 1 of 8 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1022,7 +1022,7 @@ private extension PurchasesOrchestrator {
         }
     }
 
-    func syncPurchases(receiptRefreshPolicy: ReceiptRefreshPolicy,
+    func syncPurchases(receiptRefreshAllowed: Bool,
```
MarkVillacampa, line 1025: Since `ReceiptRefreshPolicy` is very tied to SK1, I changed the parameter to a more SK-agnostic one and decide in the method body what policy to use for each SK version.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1139,6 +1140,9 @@ private extension PurchasesOrchestrator {
                         return
                     }
 
+                    let appTransactionJWS = await self.transactionFetcher.appTransactionJWS(
+                        refreshPolicy: refreshPolicy)
```
MarkVillacampa, line 1144: Moved fetching the `AppTransaction` to after the check to see if we have transactions, originalPurchaseDate and originalApplicationVersion.

This way, if we don't need to post the receipt, we wouldn't be fetching the AppTransaction unnecessarily, potentially showing the authentication prompt.

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -106,7 +106,10 @@ final class TransactionPoster: TransactionPosterType {
             switch result {
             case .success(let encodedReceipt):
                 self.product(with: productIdentifier) { product in
-                    self.transactionFetcher.appTransactionJWS { appTransaction in
+                    // Only allow refreshing the AppTransaction if the source was a user action (e.g. purchase)
+                    let refreshPolicy: AppTransactionRefreshPolicy =
+                        data.source.initiationSource == .purchase ? .onlyIfEmpty : .never
+                    self.transactionFetcher.appTransactionJWS(refreshPolicy: refreshPolicy) { appTransaction in
```
MarkVillacampa, line 112: We dont want to potentially show the login prompt for transactions coming from the queue

Sources/Purchasing/StoreKit2/SK2AppTransaction.swift
```diff
@@ -18,13 +18,15 @@ import StoreKit
 internal struct SK2AppTransaction {
 
     @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
-    init(appTransaction: AppTransaction) {
+    init(appTransaction: AppTransaction, jwsRepresentation: String) {
         self.bundleId = appTransaction.bundleID
         self.originalApplicationVersion = appTransaction.originalAppVersion
         self.originalPurchaseDate = appTransaction.originalPurchaseDate
         self.environment = .init(environment: appTransaction.environment)
+        self.jwsRepresentation = jwsRepresentation
```
MarkVillacampa, line 26: This is similar to what we already so with the `SK2StoreTransaction`. The JWS token is part of the `VerificationResult`, not the `AppTansaction`, but we want to preserve it.

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -279,22 +288,42 @@ extension StoreKit2TransactionFetcher {
         }
     }
 
-    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
-    private var appTransaction: SK2AppTransaction? {
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    private var refreshedAppTransaction: SK2AppTransaction? {
```
MarkVillacampa, line 292: The unified diff of this is a bit weird.

`appTransaction` was modified into the `appTransaction(refreshPolicy:)` method below.

and `refreshedAppTransaction` was added as a helper to fetch an `AppTransaction` via `StoreKit.AppTransaction.refresh()`

Tests/UnitTests/Mocks/MockStoreKit2TransactionFetcher.swift
```diff
@@ -55,6 +56,11 @@ final class MockStoreKit2TransactionFetcher: StoreKit2TransactionFetcherType {
         }
     }
 
+    var receivedRefreshPolicy: AppTransactionRefreshPolicy? {
+        get { return self._receivedRefreshPolicy.value }
+        set { self._receivedRefreshPolicy.value = newValue }
+    }
```
MarkVillacampa, line 62: Added a little helper in the mock so we can check the received refresh policy in different tests.

fire-at-will (approved):
👏👏

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -279,22 +288,42 @@ extension StoreKit2TransactionFetcher {
...
+                    return transaction
+                } else {
+                    return await refreshedAppTransaction
+                }
+            } catch {
+                switch refreshPolicy {
+                case .onlyIfEmpty:
+                    return await refreshedAppTransaction
+                case .never:
+                    Logger.warn(Strings.storeKit.sk2_error_fetching_app_transaction(error))
```
jamesrb1, line 319: Would this qualify for a a `Logger.appleError`?

jamesrb1 (approved):
I don't have context for this change, so you'll likely want approval from someone else too.

aboedo:
@MarkVillacampa @fire-at-will is this something we still want to do?

PR #4100, opened 2024-07-22
Treat Retry-After Header As Seconds

fire-at-will:
### Motivation
The RevenueCat REST API v2's `Retry-After` header uses seconds instead of milliseconds, so we've decided to treat it as seconds to maintain consistency.

### Description
Adjusts the SDK to treat the `Retry-After` header's value as seconds instead of milliseconds.

### DIFF
--- a/Tests/UnitTests/Networking/HTTPClientTests.swift
+++ b/Tests/UnitTests/Networking/HTTPClientTests.swift
@@ -1869,7 +1869,7 @@ extension HTTPClientTests {
         )!
 
         let backoffPeriod = self.client.calculateRetryBackoffTime(forResponse: httpURLResponse, retryCount: 2)
-        expect(backoffPeriod).to(equal(TimeInterval(retryAfterSeconds)))  // 0.1s == 100ms
+        expect(backoffPeriod).to(equal(TimeInterval(retryAfterSeconds)))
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/Networking/HTTPClientTests.swift
```diff
@@ -1857,19 +1857,32 @@ extension HTTPClientTests {
...
             httpVersion: nil,
             headerFields: [
-                HTTPClient.ResponseHeader.retryAfter.rawValue: "\(retryAfterMilliseconds)"
+                HTTPClient.ResponseHeader.retryAfter.rawValue: "\(retryAfterSeconds)"
             ]
         )!
 
-        let backoffPeriod = self.client.calculateRetryBackoffTime(forResponse: httpURLResponse, retryCount: 2)
-        expect(backoffPeriod).to(equal(TimeInterval(0.1)))  // 0.1s == 100ms
+        var backoffPeriod = self.client.calculateRetryBackoffTime(forResponse: httpURLResponse, retryCount: 2)
```
fire-at-will, line 1871: Since we're dealing with seconds now, I figured it'd be a good idea to make sure that we properly handle decimals as well as integers.

PR #4107, opened 2024-07-24
Only Retry POST Receipt Paths for 429

fire-at-will:
### Description
This PR modifies the HTTPClient's retry logic to only perform retries on requests to the `/v1/receipts` URL path that result in a 429 status code.

#### Testing
- Updated existing tests to set the `isRetryable` field correctly
- Added unit tests to `HTTPClient.Request` to ensure that the `isRetryable` field is being set properly
- Wrote new unit & integration tests to ensure that we don't retry requests to unsupported endpoints (like `/v1/subscribers/identify`)

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -671,9 +671,11 @@ extension HTTPClient {
     internal func isResponseRetryable(_ urlResponse: HTTPURLResponse) -> Bool {
         let isStatusCodeRetryable = self.retriableStatusCodes.contains(urlResponse.httpStatusCode)
         let doesServerAllowRetryString = urlResponse.value(forHTTPHeaderField: ResponseHeader.isRetryable.rawValue)
-        var doesServerAllowRetry: Bool = true
+        let doesServerAllowRetry: Bool
         if let doesServerAllowRetryString = doesServerAllowRetryString {
             doesServerAllowRetry = Bool(doesServerAllowRetryString.lowercased()) ?? true
+        } else {
+            doesServerAllowRetry = true
         }
 
         return isStatusCodeRetryable && doesServerAllowRetry
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -34,7 +34,6 @@ class HTTPClient {
     private let diagnosticsTracker: DiagnosticsTrackerType?
     private let dateProvider: DateProvider
     private let retriableStatusCodes: Set<HTTPStatusCode>
-    private let retriablePaths: Set<String>
     private let operationDispatcher: OperationDispatcher
 
     private let retryBackoffIntervals: [TimeInterval] = [
@@ -50,7 +49,6 @@ class HTTPClient {
          diagnosticsTracker: DiagnosticsTrackerType?,
          dnsChecker: DNSCheckerType.Type = DNSChecker.self,
          retriableStatusCodes: Set<HTTPStatusCode> = Set([.tooManyRequests]),
-         retriablePaths: Set<String> = Set(["/v1/receipts"]),
          requestTimeout: TimeInterval = Configuration.networkTimeoutDefault,
          dateProvider: DateProvider = DateProvider(),
          operationDispatcher: OperationDispatcher) {
@@ -67,7 +65,6 @@ class HTTPClient {
         self.signing = signing
         self.diagnosticsTracker = diagnosticsTracker
         self.dnsChecker = dnsChecker
-        self.retriablePaths = retriablePaths
         self.retriableStatusCodes = retriableStatusCodes
         self.timeout = requestTimeout
         self.apiKey = apiKey
@@ -628,8 +625,10 @@ extension HTTPClient {
         request: HTTPClient.Request,
         httpURLResponse: HTTPURLResponse?
     ) -> Bool {
+
         guard let httpURLResponse = httpURLResponse,
-              isRetryable(httpURLResponse) else { return false }
+              request.httpRequest.isRetryable,
+              isResponseRetryable(httpURLResponse) else { return false }
 
         // At this point, retryCount hasn't been incremented yet, so we'll need to do it early here
         // to determine if another retry is appropriate.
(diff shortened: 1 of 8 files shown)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -666,17 +668,15 @@ extension HTTPClient {
...
-    internal func isRetryable(_ urlResponse: HTTPURLResponse) -> Bool {
+    internal func isResponseRetryable(_ urlResponse: HTTPURLResponse) -> Bool {
         let isStatusCodeRetryable = self.retriableStatusCodes.contains(urlResponse.httpStatusCode)
-        let isRetryableString = urlResponse.value(forHTTPHeaderField: ResponseHeader.isRetryable.rawValue)
-        let isRetryable: Bool
-        if let isRetryableString {
-            isRetryable = Bool(isRetryableString.lowercased()) ?? true
-        } else {
-            isRetryable = true
+        let doesServerAllowRetryString = urlResponse.value(forHTTPHeaderField: ResponseHeader.isRetryable.rawValue)
```
fire-at-will, line 673: No change in the logic, just renaming `isRetryable` to `doesServerAllowRetry` to be more clear

fire-at-will:
@RCGitBot please test

fire-at-will:
Some of the test checks look a little funny right now because it took me a moment to figure out how to run all of the tests, I'll make sure they're all good to go before merging 💪

Tests/BackendIntegrationTests/OtherIntegrationTests.swift
```diff
@@ -173,7 +175,7 @@ class OtherIntegrationTests: BaseBackendIntegrationTests {
         try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
 
         let result = try await self.purchases.productEntitlementMapping().entitlementsByProduct
-        expect(result).to(haveCount(18))
+        expect(result).to(haveCount(21))
```
fire-at-will, line 178: I'm not sure when (the audit logs aren't loading for me), but we now have 21 iOS products in the StoreKit Integration Tests app instead of 18

fire-at-will: I'm looking into this a little more to confirm why this has changed

fire-at-will: The most recently added product to the project was on June 25, so this isn't an issue with a new product being present. Continuing to dig 👀

fire-at-will: The test is currently failing in main with the value `18`, but passed during yesterday's release pipeline 🧐


https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/20914/workflows/68cd311d-9aec-47a8-8f55-889bef8bdabc

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -666,17 +668,15 @@ extension HTTPClient {
...
+        var doesServerAllowRetry: Bool = true
+        if let doesServerAllowRetryString = doesServerAllowRetryString {
+            doesServerAllowRetry = Bool(doesServerAllowRetryString.lowercased()) ?? true
```
aboedo, line 676: I'm sure this version is more clever, but it was kinda clearer before the ternary IMO

Sources/Networking/HTTPClient/HTTPRequest.swift
```diff
@@ -22,26 +22,49 @@ struct HTTPRequest {
     var path: HTTPRequestPath
     /// If present, this will be used by the server to compute a checksum of the response signed with a private key.
     var nonce: Data?
-
-    init(method: Method, path: HTTPRequest.Path, nonce: Data? = nil) {
-        self.init(method: method, requestPath: path, nonce: nonce)
+    /// Whether or not this reqeust should be retried by the HTTPClient for certain status codes.
```
aboedo, line 25: typo `request`

Tests/BackendIntegrationTests/OtherIntegrationTests.swift
aboedo: @RevenueCat/coresdk thoughts?

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
```diff
@@ -935,6 +935,7 @@ class StoreKit1IntegrationTests: BaseStoreKitIntegrationTests {
     }
 
     func testVerifyPurchaseGrantsEntitlementsThroughOnRetryAfter429() async throws {
+        defer { HTTPStubs.removeAllStubs() }
```
aboedo, line 938: should this just be a setup or teardown?

aboedo (commented):
looking good so far

Sources/Networking/HTTPClient/HTTPClient.swift
fire-at-will: I'm okay reverting it! I'll keep the name `doesServerAllowRetry` over `isRetryable` though 😅

Tests/BackendIntegrationTests/StoreKitIntegrationTests.swift
fire-at-will: Moved it to the `tearDown` functions 😎

fire-at-will:
Hey @aboedo! Since your last review, I've made the following changes:

### Code Changes
- Reverted the syntax when computing `doesServerAllowRetry`
- Fixed the "request" typo in `HTTPRequest.swift`
- Moved the `HTTPStubs.removeAllStubs()` calls in the tests from a `defer` statements to the `tearDown` functions

### Test Fixes/Clarifications
- Fixed the configuration issues that were causing the `CustomEntitlementsComputationIntegrationTests` to fail
- Dug into why `OtherIntegrationTests.testProductEntitlementMapping` was returning 21 mappings instead of 18. Josh and I found that Khepri was returning 18 mappings on Tuesday when the test passed, but is now returning 21 in the network calls to `/v1/product_entitlement_mapping`. We now believe that 21 is the correct number and are following up with the backend team to investigate what caused the change to take so long to appear.

This should be ready for another look!

Tests/BackendIntegrationTests/OtherIntegrationTests.swift
fire-at-will: Josh and I took a look at some network logs and found that the backend was returning 18 mappings on Tuesday when the test passed, but is now returning 21 in the network calls to `/v1/product_entitlement_mapping`. This is happening on this branch, main, and the 4.32.2 release. We now believe that 21 is the correct number and are following up with the backend team to investigate what caused the change to take so long to appear.

aboedo (approved):

PR #4111, opened 2024-07-24
Clean up XCWorkspace and testing apps

aboedo:
## Motivation

Having to remember which Xcode project to open to work on every specific feature drives me insane. 

This just adds the Testing Apps to the main RevenueCat.workspace so we can launch them from there and I can stop pulling my hair out. 

It also removes the old UIKit PurchaseTester since we haven't been using it for years. 

The Backend Integration tests app and the PurchaseTesterSwiftUI app were using a storekit config file from the old PurchaseTester for some reason, so I cleaned that up too - added it as a separate file per app. 

| Before | After |
| :-: | :-: |
| Screenshot 2024-07-24 at 5 54 55 PM | image |

# Changes summary: 
- Add all Testing Apps to `RevenueCat.xcworkspace`
- Kill PurchaseTester UIKit
- Move StoreKit config file from PurchaseTester UIKit to BackendIntegrationTests
- Create new StoreKit Config file for PurchaseTester SwiftUI
- Replace SPM dependency from PurchaseTester -> RevenueCat with a local dependency


aboedo:
This does add noise to the overall schemes in Xcode, let me know if that feels like a bit too much

tonidero (approved):
Might be a good idea to get approval from some iOS devs, but this looks great to me! This is indeed painful right now 🙏

aboedo:
The diff probably makes this look awful, but all I did was to literally drag and drop the folder into Xcode

aboedo:
requested review from iOS folks too

fire-at-will (approved):
Looks good to me! We should definitely revisit our schemes at some point, but I don't think that should hold this back. It'll be SO NICE not to have to open a separate project each time for the test apps

fire-at-will:
I tested opening up the workspace on my machine from this branch and didn't have any issues building the SDK 👍

JayShortway:
THANK YOU! 🙇

aboedo:
most of the test failures are just flaky tests, but the carthage installation tests one is broken by this PR I believe, probably because carthage is trying to build all of the schemes in the workspace now and struggling

aboedo:
to clarify, the problem is in the test itself and the fact that carthage just blindly tries to build every single scheme it finds, not in the changes

aboedo:
okay, installation tests now pass. 

The PurchaseTester app had a SPM dependency on our SDK, which in turn created a duplicate package. it also made it super annoying to develop new things for that app because you also had to remove the dependency and re-add it as local. 

So I did that - added it as a local dependency and now we're good.

aboedo:
This turned into a bigger thing than I thought, since for some reason our backend tests were using a storekit configuration file from purchaseTester, and both purchase testers were pointing to the same storekit config file so I cleaned that up too

RevenueCat.xcodeproj/xcshareddata/xcschemes/BackendIntegrationTests.xcscheme
```diff
@@ -89,7 +89,7 @@
          </BuildableReference>
       </BuildableProductRunnable>
       <StoreKitConfigurationFileReference
-         identifier = "../Tests/TestingApps/PurchaseTester/RevenueCat_IntegrationPurchaseTesterConfiguration.storekit">
+         identifier = "../Tests/BackendIntegrationTests/RevenueCat_IntegrationPurchaseTesterConfiguration.storekit">
```
aboedo, line 92: the BackendIntegrationTests app was using the StoreKit config file from PurchaseTester UIKit for some reason. I moved it into this app and killed PurchaseTester UIKit

Tests/TestingApps/PaywallsTester/PaywallsTester.xcodeproj/xcshareddata/xcschemes/PaywallsTester - SK config.xcscheme
```diff
@@ -86,7 +86,7 @@
          </BuildableReference>
       </BuildableProductRunnable>
       <StoreKitConfigurationFileReference
-         identifier = "../PaywallsTester/Products.storekit">
+         identifier = "../Tests/TestingApps/PaywallsTester/PaywallsTester/Products.storekit">
```
aboedo, line 89: Cleans up the relative directory so it picks up correctly when opening the workspace, which is the only thing we should be using from now on

Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcodeproj/xcshareddata/xcschemes/PurchaseTester.xcscheme
```diff
@@ -51,7 +51,7 @@
          </BuildableReference>
       </BuildableProductRunnable>
       <StoreKitConfigurationFileReference
-         identifier = "../../PurchaseTester/RevenueCat_IntegrationPurchaseTesterConfiguration.storekit">
+         identifier = "../Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTesterStoreKitConfiguration.storekit">
```
aboedo, line 54: Created a new StoreKit Config file for PurchaseTesterSwiftUI, synced to app store connect

aboedo:
The changes _look_ like a lot, but it's really just dragging and dropping files + updating some references to storekit configuration files

tonidero (approved):
Love it! One question though, PurchaseTester for macOS was failing to build for me because of some certificate issues. I fixed that by opening the purchase tester workspace (instead of just selecting the project in the root project) but then, it didn't work to build from there (it couldn't find the packages). Seems like we broke building from that workspace? Maybe it's ok and we should just always build from the root workspace from now on though.

aboedo:
> Do we still need the PurchaseTester.xcworkspace in PurchaseTesterSwiftUI?

Uhhh... _probably_ not! I'll try removing it as a separate PR

aboedo:
> Love it! One question though, PurchaseTester for macOS was failing to build for me because of some certificate issues. I fixed that by opening the purchase tester workspace (instead of just selecting the project in the root project) but then, it didn't work to build from there (it couldn't find the packages). Seems like we broke building from that workspace? Maybe it's ok and we should just always build from the root workspace from now on though.

re: certs - we're currently set up to use fastlane, but honestly I think Xcode's automatic management has gotten good enough that it might not be worth it for apps that we don't intend to build in CI? I'll try moving to auto management in a separate PR. 

re: building from other workspaces - my goal is to have the One True workspace and always open things from there so we don't have to fish around for where to open projects

PR #4113, opened 2024-07-24
Re-added the RevenueCatUI tests job on every commit

aboedo:
I realized that we weren't running the suite of tests from RevenueCatUI in every commit. 

I missed this one because the name made it seem like it was a test specific to SPM, but it's just that it could only be run through SPM at the time. 

It can now be run from Xcode, but that's a separate story, so for now this just re-adds it and clarifies the naming a bit

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -569,7 +569,7 @@ jobs:
           name: RevenueCatUI API Tests
           command: bundle exec fastlane build_revenuecatui_api_tester
 
-  spm-revenuecat-ui-ios-17:
+  run-revenuecat-ui-ios-17:
     <<: *base-job
     parameters:
       xcode_version:
@@ -1269,7 +1269,7 @@ workflows:
     jobs:
       - spm-revenuecat-ui-ios-15
       - spm-revenuecat-ui-ios-16
-      - spm-revenuecat-ui-ios-17
+      - run-revenuecat-ui-ios-17
 
   build-test:
     when:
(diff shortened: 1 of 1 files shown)


aboedo:
Oh, I guess re-doing the names sadly means that I'd have to also update the name in the required jobs list. 
I'll leave it as is for now until folks share their thoughts on the naming

.circleci/config.yml
```diff
@@ -1279,9 +1279,14 @@ workflows:
...
-      - run-test-ios-17
-      - pod-lib-lint
+      - lint:
+          name: "swiftlint"
+      - run-test-ios-17:
+          name: "Unit tests"
+      - pod-lib-lint:
+          name: "Pod lib lint"
+      - spm-revenuecat-ui-ios-17:
+          name: "RevenueCatUI tests"
```
aboedo, line 1289: let me know if the name feels like it's not adding much and / or making it harder to map the jobs as they show up in github to the circleCI config. I just did this because the name `spm-revenuecat-ui-ios-17` seemed so specific that it didn't seem like a vital check to have until I realized what it was

tonidero: Hmm how about renaming the job to something like `run-revenuecat-ui-ios-17` instead? I feel changing the name like this might make it harder to try to find what job is failing later.

aboedo:
@RCGitBot please test

.circleci/config.yml
aboedo: done!

.circleci/config.yml
```diff
@@ -1269,7 +1269,7 @@ workflows:
     jobs:
       - spm-revenuecat-ui-ios-15
       - spm-revenuecat-ui-ios-16
```
tonidero, line 1271: I do wonder if we should also rename these, if they are doing the same thing but on different iOS versions...

tonidero (approved):
Just a comment but looks good!

PR #4116, opened 2024-07-26
Allow user to set the paywall content max width

MojtabaHs:
## Motivation

Sometimes, it's not ideal to use the entire width of the screen for the paywall.

## ↔️ Default Look
For example, a fullscreen paywall on an iPad looks like this by default:
old
<sub>☝️ This is a 10-inch iPad, but imagine a 13-inch!</sub>

This extremely wide content looks off, doesn't it? Who needs a button that's `2752 px` (or `1376 pt`) wide?! 😱

---

## ✅ Customized with the New API
By limiting it to a more readable size, it looks like this:
new

Much better, right?

### 💡 Usage
```swift
PaywallView()
    .environment(\.paywallContentMaxWidth, UIViewController().view.readableContentGuide.layoutFrame.size.width)
```
<sub>☝️ I used the native `readableContentGuide` as Apple suggested for the device here, but the user can pass any value that suits them.</sub>

### DIFF
--- a/RevenueCatUI/Templates/TemplateViewType.swift
+++ b/RevenueCatUI/Templates/TemplateViewType.swift
@@ -142,8 +142,10 @@ struct BackgroundedTemplateView: ViewModifier {
     let configuration: TemplateViewConfiguration
 
     func body(content: Content) -> some View {
-        content
-            .background(configuration.backgroundView)
+        ZStack {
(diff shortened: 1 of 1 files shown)


MojtabaHs:
Rebased on the `main` branch.

MojtabaHs:
Any Updates here?

PR #4120, opened 2024-07-26
[Customer Center] Make cancel subscription less prominent

vegaro:
Made the cancel subscription more destructive by removing the filling of the button and making it always red text

| Header | Header |
|--------|--------|
| simulator_screenshot_870B1F8E-5DC7-415A-AF46-B1B8A6767841 | simulator_screenshot_719D9495-A004-404E-84AC-1FF245457DBC |

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
@@ -303,7 +303,9 @@ struct ManageSubscriptionButton: View {
                                  product: promotionalOfferData.product,
                                  promoOfferDetails: promotionalOfferData.promoOfferDetails)
         })
-        .buttonStyle(ManageSubscriptionsButtonStyle())
+        .applyIf(self.path.type != .cancel) {
+            $0.buttonStyle(ManageSubscriptionsButtonStyle())
+        }
         .disabled(self.viewModel.loadingPath != nil)
     }
 }
--- a/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
@@ -306,6 +306,9 @@ struct ManageSubscriptionButton: View {
         .applyIf(self.path.type != .cancel) {
             $0.buttonStyle(ManageSubscriptionsButtonStyle())
         }
+        .applyIf(self.path.type == .cancel) {
+            $0.foregroundColor(Color.red)
+        }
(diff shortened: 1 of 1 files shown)


RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -303,7 +303,12 @@ struct ManageSubscriptionButton: View {
                                  product: promotionalOfferData.product,
                                  promoOfferDetails: promotionalOfferData.promoOfferDetails)
         })
-        .buttonStyle(ManageSubscriptionsButtonStyle())
+        .applyIf(self.path.type != .cancel) {
+            $0.buttonStyle(ManageSubscriptionsButtonStyle())
+        }
+        .applyIf(self.path.type == .cancel) {
+            $0.foregroundColor(Color.red)
```
vegaro, line 310: made it always red because I think it should have good contrast in light and dark modes. But maybe this should be configurable?

aboedo: I think if we allow for configuration of the background colors we should probably allow for configuration of this one because it might have bad contrast with your choice of a background

aboedo:
🤔 honestly not sure whether this is all that good, I feel like having that big red button would almost feel like that's how you dismiss the modal and I would press it right away, but it might just be me

vegaro:
I agree on the color being configurable. We need to fix the JSON of the colors to make them nullable so they are individually optional and add more colors to the JSON to be able to customize things like this.

Regarding changing the color to red, this was feedback from @codykerns and I've been thinking of ways of making the cancel less prominent, so it feels less than the other options. He actually sent me this screenshot, which I think is a good example:

image

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
tonidero: Hmm, this could look weird if the cancel button is not the last button in the paths configured in the backend right?

vegaro: Good point, we can make it so it is always the last one? We can guarantee that from the frontend maybe adding the cancel button as a checkbox instead of part of the configurable buttons. Or we could also leave it as part of the devs responsibilities. 

Let's talk about this in the sync, because I am noticing now this might make it more complex than we want at least for the first iterations

vegaro:
Going to close this for now, since we discussed it's out of scope for the MVP

PR #4121, opened 2024-07-26
tests triggrer: add target-branch parameter to trigger from the right branch

aboedo:
Fix the `@RCGitBot please test` issue where tests would always get triggered from the `main` branch. 

This solution is based on this comment: https://github.com/CircleCI-Public/trigger-circleci-pipeline-action/issues/61#issuecomment-1662021882

But you can also see in the code that this is indeed how the branch name is determined, and that that [logic looks the same as the one for `GHA_action` and `target-slug`](https://github.com/CircleCI-Public/trigger-circleci-pipeline-action/blob/a81cd720792a6088debd7f182b552845abb86f1b/src/lib/CircleCIPipelineTrigger.ts#L66), so it's just undocumented. 

We're using a fixed version, 1.2.0, so it shouldn't just randomly break.

### DIFF
--- a/.github/workflows/trigger_all_tests.yml
+++ b/.github/workflows/trigger_all_tests.yml
@@ -32,7 +32,7 @@ jobs:
           echo "User is a member of the organization"
 
 
-      - name: 'Get github branch'
+      - name: 'Get GitHub branch'
         run: echo ::set-output name=branch::$(gh pr view $PR_NO --repo $REPO --json headRefName --jq '.headRefName')
         env:
           REPO: ${{ github.repository }}
--- a/.github/workflows/trigger_all_tests.yml
+++ b/.github/workflows/trigger_all_tests.yml
@@ -31,12 +31,21 @@ jobs:
           fi
           echo "User is a member of the organization"
 
+
+      - name: 'Get github branch'
+        run: echo ::set-output name=branch::$(gh pr view $PR_NO --repo $REPO --json headRefName --jq '.headRefName')
+        env:
+          REPO: ${{ github.repository }}
+          PR_NO: ${{ github.event.issue.number }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
       - name: Trigger CircleCI workflow
         id: trigger_circleci_workflow
         if: success()
 
         uses: CircleCI-Public/trigger-circleci-pipeline-action@v1.2.0
         with:
+          target-branch: ${{ steps.get-branch.outputs.branch }}
           GHA_Meta: "run-from-github-comments"
         env:
           CCI_TOKEN: ${{ secrets.CIRCLECI_TOKEN }}
--- a/.github/workflows/trigger_all_tests.yml
+++ b/.github/workflows/trigger_all_tests.yml
@@ -32,6 +32,12 @@ jobs:
           echo "User is a member of the organization"
 
 
+      # Note: actions based on issue comments always trigger from the main branch. So we need to tell CircleCI to use a specific branch. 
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger_all_tests.yml
```diff
@@ -31,12 +31,27 @@ jobs:
...
 
+
+      # Note: actions based on issue comments always trigger from the main branch. So we need to tell CircleCI to use a specific branch. 
+      # We can do that by passing in the target-branch. 
+      # This solution is based on this comment: https://github.com/CircleCI-Public/trigger-circleci-pipeline-action/issues/61#issuecomment-1662021882
+      # But we can also see that the following code reads from this parameter: https://github.com/CircleCI-Public/trigger-circleci-pipeline-action/blob/a81cd720792a6088debd7f182b552845abb86f1b/src/lib/CircleCIPipelineTrigger.ts#L66
+      # Even though it seems to be undocumented as of writing. 
+      
+      - name: 'Get GitHub branch'
+        run: echo ::set-output name=branch::$(gh pr view $PR_NO --repo $REPO --json headRefName --jq '.headRefName')
```
tonidero, line 42: I guess the `gh` CLI will be installed right? I would imagine so, being a github action 😅

tonidero (approved):

.github/workflows/trigger_all_tests.yml
```diff
@@ -31,12 +31,27 @@ jobs:
...
           echo "User is a member of the organization"
 
+
+      # Note: actions based on issue comments always trigger from the main branch. So we need to tell CircleCI to use a specific branch. 
+      # We can do that by passing in the target-branch. 
+      # This solution is based on this comment: https://github.com/CircleCI-Public/trigger-circleci-pipeline-action/issues/61#issuecomment-1662021882
+      # But we can also see that the following code reads from this parameter: https://github.com/CircleCI-Public/trigger-circleci-pipeline-action/blob/a81cd720792a6088debd7f182b552845abb86f1b/src/lib/CircleCIPipelineTrigger.ts#L66
+      # Even though it seems to be undocumented as of writing. 
+      
+      - name: 'Get GitHub branch'
```
JayShortway, line 41: Nitpick: maybe this step can be named "Get PR branch"? "PR" is the relevant bit imo.

.github/workflows/trigger_all_tests.yml
```diff
@@ -31,12 +31,27 @@ jobs:
           fi
           echo "User is a member of the organization"
 
+
```
vegaro, line 34: super nit but there's an extra line here

vegaro:
```suggestion
```

JayShortway (approved):
Nice job getting this fixed! 🙌

vegaro (approved):

.github/workflows/trigger_all_tests.yml
aboedo: it is, listed here: 
https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md#cli-tools

I'll add a comment that this assumes that it's installed, though, just in case it randomly gets removed one day, although I can't imagine that would ever be the case

.github/workflows/trigger_all_tests.yml
aboedo: updated!

.github/workflows/trigger_all_tests.yml
aboedo: great call, updated!

PR #4123, opened 2024-07-26
[Customer Center] Progress view on restores

vegaro:
I couldn't find an easy way of adding a progress view in an Alert, so decided to display the progress in the button that triggers the restore instead.


RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
```diff
@@ -56,7 +58,11 @@ struct RestorePurchasesAlert: ViewModifier {
                                     """),
                         primaryButton: .default(Text("Check past purchases"), action: {
                             Task {
+                                isRestoring = true
                                 let alertType = await self.customerCenterViewModel.performRestore()
+                                // Simulate a delay to ensure the progress view is shown for at least one second
+                                await Task.sleep(seconds: 1)
```
vegaro, line 64: Without this, the text flashes. I figured it was actually not too bad, but maybe there's a better way to guarantee the progress view is displayed for a bit, or at least it doesn't flash?

aboedo: 🤔  maybe we can introduce an animation that has a delay, I think we've done it before (or I'm thinking about a personal app 😅 )

aboedo: Some chatGPT suggestions, but it's pretty much what I was thinking about

```swift
if showProgressIndicator {
    ProgressView()
        .transition(.opacity)
}

// and then when you want to show it: 
DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) {
    if isLoading {
        withAnimation {
            showProgressIndicator = true
        }
    }
}
```

JayShortway: I might be wrong, but doesn't this now add a second after however long `performRestore()` took? I deduced that from [this blog](https://www.avanderlee.com/swift/async-let-asynchronous-functions-in-parallel/). I think running it in parallel with `performRestore()` might be nicer? What do you think?

vegaro: Oh good points. The animation could be the right approach to this, but if that doesn't work I will try parallelising. Thanks!

jamesrb1 (commented):
Looks good however the loading indicator is very hard to see:

image

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
```diff
@@ -64,10 +66,16 @@ struct NoSubscriptionsView: View {
 
                 Spacer()
 
-                Button(localization.commonLocalizedString(for: .restorePurchases)) {
+                Button(action: {
                     showRestoreAlert = true
-                }
-                .restorePurchasesAlert(isPresented: $showRestoreAlert)
+                }, label: {
+                    if self.isRestoring {
```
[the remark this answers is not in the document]
vegaro: If I understand correctly our Swift style guide, we use explicit `self` to make it clear https://github.com/RevenueCat/purchases-ios/blob/d0dbbdce9eb00eaa3e4afc14b949e80853bd008e/Contributing/SwiftStyleGuide.swift#L73

vegaro:
putting this back to draft since https://github.com/RevenueCat/purchases-ios/pull/4190/files#diff-a409abe556c69c3eb7c7fdeb0b5ba0e46c4fb9e2f78e288772b2e748fabd1347 will change it

vegaro:
going to close this for now since the UI has changed a lot

PR #4131, opened 2024-07-29
More project structure cleanup

aboedo:
## More project structure cleanup: 

- Removes redundant workspace and project files for PurchaseTester
- Cleans up bad references
- Cleans up certificates
- Re-does the process to deploy PurchaseTester, since it was designed to run from the xcworkspace, which doesn't work anymore. 
- adds a `dry-run` option to deploying PurchaseTester and adds a job for it in the all-tests workflow. It tries building it and archiving but doesn't actually publish to TestFlight. We didn't really have a test for this, so it's another thing that could have broken in production.


aboedo:
@RCGitBot please test

aboedo:
@RCGitBot please test

aboedo:
@RCGitBot please test

.circleci/config.yml
```diff
@@ -1231,9 +1231,20 @@ jobs:
       - update-spm-installation-commit
       - run:
           name: Submit Purchase Tester
-          working_directory: "Tests/TestingApps/PurchaseTesterSwiftUI"
```
aboedo, line 1234: no longer needed or desired, we want the jobs to run from the root since that's where the framework is generated

.circleci/config.yml
```diff
@@ -1325,12 +1336,10 @@ workflows:
...
-      # temporarily disabled since this won't work when building directly from its workspace.
-      # there's another PR to re-enable it
-      # - deploy-purchase-tester:
-      #     requires:
-      #       - make-release
-      #     <<: *release-tags
+      - deploy-purchase-tester:
+          requires:
+            - make-release
+          <<: *release-tags
```
aboedo, line 1342: we can re-enable this now

.circleci/config.yml
```diff
@@ -1421,3 +1430,4 @@ workflows:
       - installation-tests-carthage
       - installation-tests-xcode-direct-integration
       - installation-tests-receipt-parser
+      - deploy-purchase-tester-dry-run
```
aboedo, line 1433: dry run job in all-tests so that we know if we're breaking purchase tester (I didn't realize that I had broken it until Toni mentioned it in a PR comment)

RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -1200,14 +1200,12 @@
 		2DD500E22C519EB4009C19B7 /* PurchaseTesterApp.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = PurchaseTesterApp.swift; sourceTree = "<group>"; };
 		2DD500E32C519EB4009C19B7 /* ReceiptInspector.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ReceiptInspector.swift; sourceTree = "<group>"; };
 		2DD500E42C519EB4009C19B7 /* ReceiptVerifier.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ReceiptVerifier.swift; sourceTree = "<group>"; };
-		2DD500E52C519EB4009C19B7 /* RevenueCat_SwiftUIConfiguration.storekit */ = {isa = PBXFileReference; lastKnownFileType = text; path = RevenueCat_SwiftUIConfiguration.storekit; sourceTree = "<group>"; };
```
aboedo, line 1203: cleanup, this no longer exists

RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -1200,14 +1200,12 @@
...
 		2DD500E82C519EB4009C19B7 /* AppIcon.xcassets */ = {isa = PBXFileReference; lastKnownFileType = folder.assetcatalog; path = AppIcon.xcassets; sourceTree = "<group>"; };
 		2DD500E92C519EB4009C19B7 /* PrivacyInfo.xcprivacy */ = {isa = PBXFileReference; lastKnownFileType = text.xml; path = PrivacyInfo.xcprivacy; sourceTree = "<group>"; };
 		2DD500EA2C519EB4009C19B7 /* PurchaseTester-Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = "PurchaseTester-Info.plist"; sourceTree = "<group>"; };
 		2DD500EB2C519EB4009C19B7 /* PurchaseTester.entitlements */ = {isa = PBXFileReference; lastKnownFileType = text.plist.entitlements; path = PurchaseTester.entitlements; sourceTree = "<group>"; };
 		2DD500EC2C519EB4009C19B7 /* PurchaseTester.xcodeproj */ = {isa = PBXFileReference; lastKnownFileType = "wrapper.pb-project"; path = PurchaseTester.xcodeproj; sourceTree = "<group>"; };
-		2DD500ED2C519EB4009C19B7 /* PurchaseTester.xcworkspace */ = {isa = PBXFileReference; lastKnownFileType = wrapper.workspace; path = PurchaseTester.xcworkspace; sourceTree = "<group>"; };
```
aboedo, line 1210: removed the workspace since it doesn't add any more value

RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -1200,14 +1200,12 @@
...
 		2DD500E62C519EB4009C19B7 /* Windows.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Windows.swift; sourceTree = "<group>"; };
 		2DD500E82C519EB4009C19B7 /* AppIcon.xcassets */ = {isa = PBXFileReference; lastKnownFileType = folder.assetcatalog; path = AppIcon.xcassets; sourceTree = "<group>"; };
 		2DD500E92C519EB4009C19B7 /* PrivacyInfo.xcprivacy */ = {isa = PBXFileReference; lastKnownFileType = text.xml; path = PrivacyInfo.xcprivacy; sourceTree = "<group>"; };
 		2DD500EA2C519EB4009C19B7 /* PurchaseTester-Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = "PurchaseTester-Info.plist"; sourceTree = "<group>"; };
 		2DD500EB2C519EB4009C19B7 /* PurchaseTester.entitlements */ = {isa = PBXFileReference; lastKnownFileType = text.plist.entitlements; path = PurchaseTester.entitlements; sourceTree = "<group>"; };
 		2DD500EC2C519EB4009C19B7 /* PurchaseTester.xcodeproj */ = {isa = PBXFileReference; lastKnownFileType = "wrapper.pb-project"; path = PurchaseTester.xcodeproj; sourceTree = "<group>"; };
```
aboedo, line 1208: I considered killing the project file, but I found it to be helpful in that we can keep the targets that are specific to PurchaseTester in that xcodeproj file and then it's a little less noisy

Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcodeproj/project.pbxproj
```diff
@@ -720,15 +716,14 @@
 				CLANG_MODULES_AUTOLINK = YES;
 				CODE_SIGN_ENTITLEMENTS = PurchaseTester.entitlements;
 				CODE_SIGN_IDENTITY = "Apple Development";
-				CODE_SIGN_STYLE = Manual;
+				CODE_SIGN_STYLE = Automatic;
```
aboedo, line 719: makes local testing soooo much easier, this was kind of broken when running macOS locally

Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcodeproj/project.pbxproj
```diff
@@ -740,9 +735,6 @@
 				PRODUCT_BUNDLE_IDENTIFIER = com.revenuecat.sampleapp;
 				PRODUCT_NAME = PurchaseTester;
 				PROVISIONING_PROFILE_SPECIFIER = "";
-				"PROVISIONING_PROFILE_SPECIFIER[sdk=iphoneos*]" = "match Development com.revenuecat.sampleapp";
-				"PROVISIONING_PROFILE_SPECIFIER[sdk=macosx*]" = "match Development com.revenuecat.sampleapp macos";
-				"PROVISIONING_PROFILE_SPECIFIER[sdk=watchos*]" = "match Development com.revenuecat.sampleapp";
```
aboedo, line 745: we're still using match for release schemes, but local testing is now Xcode automatic

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/RevenueCat_SwiftUIConfiguration.storekit
```diff
@@ -1,238 +0,0 @@
-{
```
aboedo, line 1: unused, left over from the previous cleanup PR

fastlane/Fastfile
```diff
@@ -701,7 +701,6 @@ platform :ios do
       './Tests/InstallationTests/ReceiptParserInstallation/ReceiptParserInstallation.xcodeproj/project.pbxproj',
       './Tests/APITesters/CustomEntitlementComputationSwiftAPITester/CustomEntitlementComputationSwiftAPITester.xcodeproj/project.pbxproj',
       './Tests/APITesters/RevenueCatUIAPITester/RevenueCatUISwiftAPITester.xcodeproj/project.pbxproj',
-      './Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcodeproj/project.pbxproj',
```
aboedo, line 704: I had missed this but we no longer have a hardcoded commit in that file since we're now using a local dependency instead, so this is unnecessary

fastlane/Fastfile
```diff
@@ -866,49 +865,85 @@ platform :ios do
   end
 
   desc "Build and deploy PurchaseTesterSwiftUI"
-  lane :deploy_purchase_tester do
-    latest_changelog = File.read("#{File.dirname(__FILE__)}/../#{changelog_latest_path}")
```
aboedo, line 870: quite a few changes here, but they ultimately boil down to updating things so that we're running the lane from the repo root rather than PurchaseTester root. 
This is because we do need to get the `RevenueCat` SDK that's generated at the top level. It wasn't needed before since we were using an SPM dependency on a hardcoded commit instead, but the local dependency is much easier to work with on a daily basis.

fastlane/Fastfile
```diff
@@ -866,49 +865,85 @@ platform :ios do
...
-        app_platform: 'osx'
+        app_platform: 'osx',
+        app_identifier: "com.revenuecat.sampleapp",
       )
+    else
+        UI.message("Dry run mode enabled. Skipping upload to TestFlight")
     end
-
-    ios_build(changelog)
-    macos_build(changelog)
```
aboedo, line 911: I also removed the local functions because I wanted to reuse a couple of the parameters without having to pass them in over and over. I can re-add them, but I don't usually find local functions to be easier to read. Thoughts?

fastlane/Fastfile
```diff
@@ -866,49 +865,85 @@ platform :ios do
...
+      scheme: "PurchaseTester",
+      destination: "generic/platform=ios",
+      xcodebuild_formatter: '',
+      export_method: 'app-store',
+      export_options: {
+        manageAppVersionAndBuildNumber: true,
+        provisioningProfiles: {
+          "com.revenuecat.sampleapp" => "match AppStore com.revenuecat.sampleapp",
+          "com.revenuecat.sampleapp.watchkitapp" => "match AppStore com.revenuecat.sampleapp.watchkitapp"
+        }
```
aboedo, line 900: I ended up passing the profiles explicitly here. I'm not entirely sure what I'm missing, but it seems like fastlane was reading the values from the Matchfile previously. I don't think there's a way for fastlane to read the Matchfile from a different directory. 

Maybe @joshdholtz knows? Maybe I could just import it?

scripts/pre-commit.sh
```diff
@@ -38,7 +38,6 @@ verify_no_included_apikeys() {
     "${SCRIPT_DIR}/../Examples/MagicWeather/MagicWeather/Constants.swift"
     "${SCRIPT_DIR}/../Examples/MagicWeatherSwiftUI/Shared/Constants.swift"
     "${SCRIPT_DIR}/../Tests/TestingApps/PurchaseTesterSwiftUI/Core/Constants.swift"
-    "${SCRIPT_DIR}/../Tests/TestingApps/PurchaseTester/PurchaseTester/Constants.swift"
```
aboedo, line 41: This was for the old purchaseTester UIKit, no longer exists so no longer needed

.circleci/config.yml
```diff
@@ -1231,9 +1231,20 @@ jobs:
...
+    <<: *base-job
+    steps:
+      - checkout
+      - setup-git-credentials
+      - trust-github-key
+      - install-dependencies
+      - update-spm-installation-commit
+      - run:
+          name: Submit Purchase Tester
+          command: bundle exec fastlane deploy_purchase_tester dry_run:true
```
tonidero, line 1246: Hmm I guess the `dry_run` could be a parameter of the job... But maybe this is better so it's more delimited, I'm ok keeping it as is.

fastlane/Fastfile
tonidero: I'm ok with that. I think it only could make some sense in some situations where the code needs to be executed multiple times, so I agree with the change 👍

tonidero (approved):
I tested it out for a bit and looks good to me!

.circleci/config.yml
vegaro: I think it should be a parameter, main reason is that if we update the `deploy-purchase-tester` job, we will probably forget about updating this other job. Having a single job with a parameter will make sure we only have to update one thing

Tests/TestingApps/PurchaseTesterSwiftUI/PurchaseTester.xcodeproj/project.pbxproj
vegaro: just curious, but what does this change? What's easier?

vegaro (approved):

.circleci/config.yml
aboedo: That's a great point. It's slightly tricky because of how we re-use jobs, but it's still very easy. I'll open up a separate PR for it since this one is approved and so that we can have a specific conversation about it

PR #4132, opened 2024-07-30
Price rounding logic

jamesrb1:
## Rounds prices for certain locales.

The server now returns a list of storefront country codes where paywalls should not display cents. This part of paywall data, in `zero_decimal_place_countries`.

Prices are only truncated if the price ends in .00.

Examples with and without rounding, for .00 and .99 prices (Thailand):

| With Rounding ON | With rounding OFF |
|:-------:|:-------:|
| image | image |
| image | image |


RevenueCatUI/Helpers/Package+VariableDataProvider.swift
```diff
@@ -15,6 +15,34 @@ import RevenueCat
...
+        }
+
+        return [
+            "TWN": Locale(identifier: "zh_Hant_TW"), // Taiwan
+            "KAZ": Locale(identifier: "kk_Cyrl_KZ"), // Kazakhstan
+            "MEX": Locale(identifier: "es_MX"), // Mexico
+            "PHL": Locale(identifier: "fil_PH"), // Philippines
+            "THA": Locale(identifier: "th_TH"), // Thailand
+        ]
+    }
```
joshdholtz, line 44: Will that API need to give both the strong front and locale identifier to the SDK for this? 👀 

And note that the countries/storefronts that will need to be formatted will be coming in from `PaywallData` so this might need to be passed into a function somewhere

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
```diff
@@ -24,7 +52,71 @@ extension Package: VariableDataProvider {
...
+        }
+
+        guard let priceToRound = formatter.number(from: self.storeProduct.localizedPriceString) else {
+            return self.storeProduct.localizedPriceString
+        }
+
+        let originalMaximumFractionalDigits = formatter.maximumFractionDigits
+
+        defer { formatter.maximumFractionDigits = originalMaximumFractionalDigits }
+        formatter.maximumFractionDigits = 0
```
joshdholtz, line 82: Can we create a new formatter instead of modifying the existing one? Modifying state scares me a little bit 😅

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
jamesrb1: > Will that API need to give both the strong front and locale identifier to the SDK for this? 👀

Ya, I'll see if this can be avoided. If not, a mapping will need to exist _somewhere_...

> And note that the countries/storefronts that will need to be formatted will be coming in from PaywallData so this might need to be passed into a function somewhere

Ya I knew this was temporary, just as far as I'd gotten at this point. I'll progress this to look in PaywallData.

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
jamesrb1: Good call, yes!

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
jamesrb1: Looks unlikely we'll be able to fish out a locale in this situation as per [code/comments here.](https://github.com/RevenueCat/purchases-ios/blob/781151df49f570b070c02961ccdfc1147707be5d/Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift#L91-L122).

joshdholtz: Oh nooooo... I guess we'll have to include that in the API response 😬

RevenueCatUI/Data/TestData.swift
```diff
@@ -103,6 +103,18 @@ enum TestData {
...
+        price: 59.99,
+        localizedPriceString: "$59.99",
+        productIdentifier: "com.revenuecat.product_3",
+        productType: .autoRenewableSubscription,
+        localizedDescription: "PRO annual",
+        subscriptionGroupIdentifier: "group",
+        subscriptionPeriod: .init(value: 1, unit: .year),
+        introductoryDiscount: Self.intro(14, .day, priceString: "$2.99"),
+        locale: Self.locale
+    )
```
jamesrb1, line 117: I added this because calculated prices (eg "price per month" for a product that is priced in cost per year, so the price per month is calculated as (cost per year)/12) are *not* rounded *except* if they end in .99 or .00. So 59.99/12 = 4.999 is used for some tests.

Sources/Paywalls/PaywallData.swift
```diff
@@ -37,6 +68,13 @@ public struct PaywallData {
         set { self._revision = newValue }
     }
 
+    /// The storefront country IDs that should not display cents in prices.
+    public var zeroDecimalPlaceCountries: ZeroDecimalPlaceCountries {
```
tonidero, line 72: Need to add to API tests?

Sources/Paywalls/PaywallData.swift
```diff
@@ -15,6 +15,37 @@
...
 
+/// Represents countries where currencies typically have zero decimal places
+public struct ZeroDecimalPlaceCountries: Codable, Sendable, Hashable, Equatable {
+
+    /// Storefront IDs that should typically display zero decimal places
+    public var apple: Set<String> {
+        _apple
+    }
+
+    private var google: Set<String> {
```
tonidero, line 26: Do we need to parse Google's countries at all? We can probably just ignore the key in the response I think? (Or even not send it from the backend when using an iOS API key if possible)

Sources/Paywalls/PaywallData.swift
```diff
@@ -15,6 +15,37 @@
 
 import Foundation
 
+/// Represents countries where currencies typically have zero decimal places
+public struct ZeroDecimalPlaceCountries: Codable, Sendable, Hashable, Equatable {
+
+    /// Storefront IDs that should typically display zero decimal places
+    public var apple: Set<String> {
```
tonidero, line 22: Probably a bit late to this, but a couple questions:
- This list of countries would be static right? And it probably depends on the currency, not the country, I would think? Just wondering if this info would be something that is ok to hardcode in the SDK...
- In case the above is wrong, I believe we send the storefront as a header in all requests? We could probably have the backend tell us whether to do this, instead of sending a whole list of countries... But we can continue with this for now if the backend has already been implemented.

tonidero (commented):
Just some questions about the architecture, but looking good.

Sources/Paywalls/PaywallData.swift
joshdholtz: > This list of countries would be static right? And it probably depends on the currency, not the country, I would think? Just wondering if this info would be something that is ok to hardcode in the SDK.

Well doing this on the currency _is_ correct, the storefront also determines the currency 🤷‍♂️ I guess `StoreProduct` does have `currencyCode` but it like there could be some discrepancy between what the underlying SK1 and SK2 return there while storefront is pretty well documented as `ISO 3166-1 Alpha-3`

joshdholtz (changes requested):

Sources/Paywalls/PaywallData.swift
jamesrb1: Done!

Sources/Paywalls/PaywallData.swift
jamesrb1: I think country is better than currency here, because it could be customary to write prices as "$7" in some countries that use dollars while maybe "$7.00" could be customary in others. Using country gives more granular control than using currency. But what could be missing here is the ability to round for non-computed values **only** if it's close to a whole number. But I suppose that is another issue....?

RevenueCatUI/Data/Strings.swift
```diff
@@ -121,6 +126,21 @@ extension Strings: CustomStringConvertible {
...
             "You must have initialized your `PaywallView` appropriately."
+
+        case .no_price_format_priceFormatter_unavailable:
+            return "Could not determine price format because priceFormatter is unavailable."
+
+        case .no_price_format_priceString_incompatible:
+            return " Could not determine price format because priceString is incompatible."
+
+        case .no_price_round_priceFormatter_nil:
+            return " Could not round price because priceFormatter is nil."
```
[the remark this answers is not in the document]
jamesrb1: ha, yes, thanks!

Sources/Paywalls/PaywallData.swift
```diff
@@ -15,6 +15,37 @@
...
+    /// Storefront IDs that should typically display zero decimal places on Apple and Google devices, respectively
+    public init(apple: Set<String>, google: Set<String>) {
+        self._apple = apple
+        self._google = google
+    }
+
+    private enum CodingKeys: String, CodingKey {
+        case _apple = "apple"
+        case _google = "google"
+    }
```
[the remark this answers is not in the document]
jamesrb1: This seems to be a limitation of the property wrapper, that it can't work on public types. The code you proposed generates an error "Property cannot be declared public because its property wrapper type uses an internal type". Based on the existing code below, I figured this was how it should be used.

```
    @DefaultDecodable.Zero
    internal private(set) var _revision: Int = 0

    /// The revision identifier for this paywall.
    public var revision: Int {
        get { return self._revision }
        set { self._revision = newValue }
    }
```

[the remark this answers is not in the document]
joshdholtz: Ohhhhhh ewww 😬 I guess that makes sense then 🤷‍♂️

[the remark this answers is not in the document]
joshdholtz: @jamesrb1 That property wrapper might not even be necessary since we are doing some optional chaining and nil coalescing here - https://github.com/RevenueCat/purchases-ios/pull/4132/files#diff-8ea63d6187042a18f8814d96bafd8472d618f95d1bf67def849fae6b041b28e0R63-R65

I'd opt for just making it `public var apple: [String]` without the property wrapper I think? 🤔

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -235,6 +235,13 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
         set { self.systemInfo.finishTransactions = newValue.finishTransactions }
     }
 
+    /// The three-letter code representing the country or region
+    /// associated with the App Store storefront.
+    /// - Note: This property uses the ISO 3166-1 Alpha-3 country code representation.
+    @objc public var storeFrontCountryCode: String? {
+        systemInfo.storefront?.countryCode
+    }
```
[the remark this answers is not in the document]
jamesrb1: Done!

Sources/Paywalls/PaywallData.swift
jamesrb1: [We do for PaywallData](https://github.com/RevenueCat/purchases-ios/blob/main/Tests/APITesters/AllAPITests/SwiftAPITester/PaywallAPI.swift)!

Sources/Paywalls/PaywallData.swift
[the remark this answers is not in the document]
jamesrb1: You are totally right, and done!

Sources/Paywalls/PaywallData.swift
tonidero: Well, I think it should be ok to use countries. Wdyt about my other suggestions? That is, hardcoding the countries in the SDK instead of getting them from the backend?

JayShortway: For currencies lacking minor units, wouldn't "7.00" be incorrect in any country using that currency? We'd be inventing "cents" which don't exist in that currency. 

I'd be inclined to vote for using currencies too because of this reason, but feel free to ignore me, especially if it's too late to change now.

joshdholtz: > That is, hardcoding the countries in the SDK instead of getting them from the backend

Trying to avoid hardcoding as many things in the SDK for paywalls as possible 😅 Activity trying to pull out the localizations we have hardcoded (for restore, terms, privacy, durations, etc) because there have been so many requested updates and its hard to say which version things are/aren't fixed/supported in (especially across multiple SDKs).

And the API trip is already being made so its mostly free to do this and I'd prefer to have more control if we need to add or remove any 🤷‍♂️ 

> I'd be inclined to vote for using currencies too because of this reason, but feel free to ignore me, especially if it's too late to change now.

FWIW, our backend also has a list of countries with zero decimal places by country 🤷‍♂️  I'm not an super expert on currencies/country formatting though

JayShortway: >Trying to avoid hardcoding as many things in the SDK for paywalls as possible
 
Currencies, time zone offsets, these things are [not](https://en.wikipedia.org/wiki/Croatia_and_the_euro#:~:text=Croatia%20adopted%20the%20euro%20as,member%20state%20of%20the%20eurozone.) [set](https://commission.europa.eu/news/commission-assesses-countries-progress-towards-adopting-euro-2024-06-26_en) in [stone](https://en.wikipedia.org/wiki/Bitcoin_in_El_Salvador). So would indeed be good to be able to update from the backend. 
  
Regarding currency/country formatting, my understanding is that lacking minor units is an attribute of the currency. Formatting rules on the other hand are country (locale) dependent. E.g. 7 US dollars is `$7.00` in `enUS` and `US$ 7,00` in `nlNL`. So whether a currency has minor units (decimals) and how to format a currency are different (but related) things. Again, I don't want to be the person needlessly dragging this on. So if the current approach works, it works 😄

aboedo: My $.02:
I don’t have a strong feeling either way. 

**Storefront vs currency:**
My intuition is that it’d be more tied to currency than country, but I haven’t done the research. 
Storefront is more granular, though, right? Because a currency can be in multiple storefronts but not the other way around, so using storefront seems more flexible. So the cost when going with storefront instead of currency is a _little_ bit more duplicated information but the gain is extra flexibility for us to fix things if we've made a wrong call. 

**Backend-driven vs hardcoded:**
Backend-driven seems like a good idea, since it makes things way easier to fix if we make wrong assumptions, which as we learned from localization stuff, we tend to do. It's been really painful for us to fix our bad localizations in the SDK, and even more painful for customers to deal with them and to have to update the SDK to get fixes.

So I'd imagine that storefront + backend-driven is the winning combination.

jamesrb1: This actually started because in Taiwan, their currency does have cents, but those are typically only seen written in prices at gas stations and dumpling shops. Including the cent component on the paywalls looked weird.

You could make the argument that all locales using the same currency have the same customs when it comes to including cents in written prices or not, but I don't know if that's true, so as Andy wrote above, storefronts is the flexible/safe bet.

JayShortway: Ah nice, TIL, thanks! It's always the exceptions 😅

jamesrb1:
Converting back to draft because I want to double-check some stuff after doing the Android version.

jamesrb1:
I've updated this so that it *only* rounds if prices end in .00.

jamesrb1:
@RCGitBot test please

jamesrb1:
@RCGitBot please test

RevenueCatUI/Data/Strings.swift
```diff
@@ -128,6 +134,21 @@ extension Strings: CustomStringConvertible {
...
+            return "Could not determine price format because priceFormatter is unavailable."
+
+        case .no_price_format_priceString_incompatible:
+            return "Could not determine price format because priceString is incompatible."
+
+        case .no_price_round_priceFormatter_nil:
+            return "Could not round price because priceFormatter is nil."
+
+        case .no_price_round_priceString_incompatible:
+            return "Could not round price because priceString is incompatible."
```
joshdholtz, line 147: Nit: I'd remove the camel case variable styling naming from the enum and the message 

```suggestion
        case .no_price_format_price_formatter_unavailable:
            return "Could not determine price format because price formatter is unavailable."

        case .no_price_format_price_string_incompatible:
            return "Could not determine price format because price string is incompatible."

        case .no_price_round_price_formatter_nil:
            return "Could not round price because price formatter is nil."

        case .no_price_round_price_string_incompatible:
            return "Could not round price because priceString is incompatible."
```

RevenueCatUI/Data/Variables.swift
```diff
@@ -62,6 +63,12 @@ enum VariableHandler {
     struct Context {
 
         var discountRelativeToMostExpensivePerMonth: Double?
+        var showZeroDecimalPlacePrices: Bool = false
+
+//        init(discountRelativeToMostExpensivePerMonth: Double? = nil, showZeroDecimalPlacePrices: Bool = false) {
+//            self.discountRelativeToMostExpensivePerMonth = discountRelativeToMostExpensivePerMonth
+//            self.showZeroDecimalPlacePrices = showZeroDecimalPlacePrices
+//        }
```
joshdholtz, line 71: Can this be deleted? 🤔

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
```diff
@@ -142,6 +166,46 @@ extension Package: VariableDataProvider {
 
 // MARK: - Private
 
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
+private extension Package {
+
+    func priceEndsIn00Cents(_ priceString: String) -> Bool {
```
joshdholtz, line 172: Its a little more swifty to start a function that returns a boolean with something like `is` 🤷‍♂️ 

```suggestion
    func isPriceEndingIn00Cents(_ priceString: String) -> Bool {
```

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
```diff
@@ -142,6 +166,46 @@ extension Package: VariableDataProvider {
...
+        guard let price = formatter.number(from: priceString)?.doubleValue else {
+            Logger.warning(Strings.no_price_format_priceString_incompatible)
+            return false
+        }
+
+        let roundedPrice = round(price * 100) / 100.0
+        return roundedPrice.truncatingRemainder(dividingBy: 1) == 0
+    }
+
+    func roundAndTruncatePrice(_ priceString: String) -> String {
```
joshdholtz, line 187: Maybe we could make the naming of this method more clear? 🤔 I _think_ truncate here means removing the `.00` but "truncate" can also have a different meaning like "round" where they almost contradict each other. 

Could we name this like `formatAsZeroDecimalPlaces`?

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
```diff
@@ -94,41 +115,44 @@ extension Package: VariableDataProvider {
...
         if !self.isSubscription || self.isMonthly {
-            return self.localizedPricePerPeriodFull(locale)
+            return self.localizedPricePerPeriodFull(locale, showZeroDecimalPlacePrices: showZeroDecimalPlacePrices)
         } else {
             let unit = Localization.unitLocalizedString(for: .init(value: 1, unit: .month),
                                                         locale: locale)
-            return "\(self.localizedPricePerPeriodFull(locale)) (\(self.localizedPricePerMonth)/\(unit))"
+
+            // swiftlint:disable:next line_length
+            return "\(self.localizedPricePerPeriodFull(locale, showZeroDecimalPlacePrices: showZeroDecimalPlacePrices)) (\(self.localizedPricePerMonth(showZeroDecimalPlacePrices: showZeroDecimalPlacePrices))/\(unit))"
```
joshdholtz, line 155: I think we can find a way to format this that doesn't require `// swiftlint:disable:next line_length`! 

Something like the following is easier to read, IMO 🤷‍♂️ 

```swift
let fullPrice = self.localizedPricePerPeriodFull(locale, showZeroDecimalPlacePrices: showZeroDecimalPlacePrices)
let monthlyPrice = self.localizedPricePerMonth(showZeroDecimalPlacePrices: showZeroDecimalPlacePrices)

return "\(fullPrice) (\(monthlyPrice)/\(unit))"
```

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
```diff
@@ -94,41 +115,44 @@ extension Package: VariableDataProvider {
...
+    func localizedPriceAndPerMonth(_ locale: Locale, showZeroDecimalPlacePrices: Bool = false) -> String {
         if !self.isSubscription || self.isMonthly {
-            return self.localizedPricePerPeriod(locale)
+            return self.localizedPricePerPeriod(locale, showZeroDecimalPlacePrices: showZeroDecimalPlacePrices)
         } else {
             let unit = Localization.abbreviatedUnitLocalizedString(for: .init(value: 1, unit: .month),
                                                                    locale: locale)
-            return "\(self.localizedPricePerPeriod(locale)) (\(self.localizedPricePerMonth)/\(unit))"
+            // swiftlint:disable:next line_length
+            return "\(self.localizedPricePerPeriod(locale, showZeroDecimalPlacePrices: showZeroDecimalPlacePrices)) (\(self.localizedPricePerMonth(showZeroDecimalPlacePrices: showZeroDecimalPlacePrices))/\(unit))"
```
joshdholtz, line 143: See below for how to write this without disabling swiftlint 👇

Sources/Paywalls/PaywallData.swift
```diff
@@ -15,6 +15,25 @@
...
+
+    // swiftlint:disable:next missing_docs
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.apple = try container.decodeIfPresent(Set<String>.self, forKey: .apple) ?? []
+    }
+
+}
+
```
joshdholtz, line 36: Can this struct live inside of `PaywallData` like other structs do? We don't necessarily want `ZeroDecimalPlaceCountries` to be a thing easily available at the top of our `RevenueCat` module namespace 🤷‍♂️ 

```swift
extension PaywallData {
    public struct ZeroDecimalPlaceCountries {
        // All the good stuff here
    }
}

File: Sources/Paywalls/PaywallData.swift [hunk 17]
```diff
@@ -15,6 +15,25 @@
 
 import Foundation
 
+/// Represents countries where currencies typically have zero decimal places
+public struct ZeroDecimalPlaceCountries: Codable, Sendable, Hashable, Equatable {
+
+    /// Storefront country codes that should typically display zero decimal places
+    public var apple: Set<String>
```
joshdholtz, line 22: Can this just be 👇 

```suggestion
    public var apple: [String]
```

This would allow us to not have a custom decoder 🤷‍♂️

joshdholtz (changes requested):
Looking good! Just some small code improvements/tweaks/suggestions but we are close 💪

jamesrb1:
@RCGitBot please test

Sources/Paywalls/PaywallData.swift
jamesrb1: `Set<String>` actually works without the custom decoder - the reason for the custom decoder is so that if the key `zero_decimal_place_countries` is there, but it doesn't contain `apple`, it will still decode `apple` to an empty Set rather than fail to decode and throw. Is defensively coding for a missing `apple` worthwhile? Or is there a better way? Thoughts?

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
jamesrb1: I made the change, but the use of "is" here feels off to me, I think because it's operating on the passed in object rather than the object that it's being called on.

RevenueCatUI/Data/Variables.swift
jamesrb1: 🙈

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
jamesrb1: I like it! Done!

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
jamesrb1: Done!

RevenueCatUI/Helpers/Package+VariableDataProvider.swift
jamesrb1: Done!

Sources/Paywalls/PaywallData.swift
jamesrb1: Done!

RevenueCatUI/Data/Strings.swift
jamesrb1: I made the change! Not a big deal, but I find using camel case for these "tokens" makes the message clearer because it directs your attention more precisely to the cause by referring to the variable by name. (The downside being it's now more tightly coupled to the code.)

jamesrb1:
@RCGitBot please test

jamesrb1:
@RCGitBot please test

aboedo:
looks like we'll have to update snapshots

joshdholtz:
> looks like we'll have to update snapshots

@jamesrb1 Not sure if you've done this before but you can update snapshots through fastlane 👇 

```sh
CIRCLE_TOKEN=your_personal_circle_ci_token bundle exec fastlane generate_snapshots_RC 
```

This will kick off some CI jobs that will open up PRs with updated snapshots if there are any changes

jamesrb1:
@RCGitBot please test

jamesrb1:
@RCGitBot please test

jamesrb1:
@RCGitBot please test

jamesrb1:
@RCGitBot please test

jamesrb1:
@RCGitBot please test

jamesrb1:
@RCGitBot please test

jamesrb1:
@RCGitBot please test

jamesrb1:
@RCGitBot please test

Sources/Paywalls/PaywallData.swift
jamesrb1: Went to `[String]` because order is important for test validation (json snapshot comparing). Realistically there won't be any performance difference.

joshdholtz (approved):
:shipit:

PR #4134, opened 2024-07-30
[Customer Center] Make colors nullable

vegaro:
- Adapt changes from https://github.com/RevenueCat/khepri/pull/9707/ which makes colors nullable and adds new colors for buttons
- Change background and text color in all screens
- Modified buttons to use `.buttonStyle(.borderedProminent)` so they look native by default
- Applied accent color to back button in navigation view

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/CustomerCenterEnvironment.swift
+++ b/RevenueCatUI/CustomerCenter/Data/CustomerCenterEnvironment.swift
@@ -43,7 +43,13 @@ extension CustomerCenterConfigData.Localization {
 extension CustomerCenterConfigData.Appearance {
 
     /// Default ``CustomerCenterConfigData.Appearance`` value for Environment usage
-    public static let `default` = CustomerCenterConfigData.Appearance(mode: .system)
+    public static let `default` = CustomerCenterConfigData.Appearance(
+        accentColor: .init(light: "#ffffff", dark: "#000000"),
+        textColor: .init(light: "#000000", dark: "#ffffff"),
+        backgroundColor: .init(light: "#000000", dark: "#ffffff"),
+        buttonTextColor: .init(light: "#000000", dark: "#ffffff"),
+        buttonBackgroundColor: .init(light: "#000000", dark: "#ffffff")
+    )
 
 }
 
--- a/RevenueCatUI/CustomerCenter/Data/CustomerCenterEnvironment.swift
+++ b/RevenueCatUI/CustomerCenter/Data/CustomerCenterEnvironment.swift
@@ -44,11 +44,11 @@ extension CustomerCenterConfigData.Appearance {
 
     /// Default ``CustomerCenterConfigData.Appearance`` value for Environment usage
     public static let `default` = CustomerCenterConfigData.Appearance(
-        accentColor: .init(light: "#ffffff", dark: "#000000"),
-        textColor: .init(light: "#000000", dark: "#ffffff"),
-        backgroundColor: .init(light: "#000000", dark: "#ffffff"),
-        buttonTextColor: .init(light: "#000000", dark: "#ffffff"),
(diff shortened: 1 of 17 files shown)


RevenueCatUI/CustomerCenter/ManageSubscriptionsButtonStyle.swift
```diff
@@ -66,7 +52,15 @@ struct ManageSubscriptionsButtonStyle_Previews: PreviewProvider {
     static var previews: some View {
         Button("Didn't receive purchase") {}
             .buttonStyle(ManageSubscriptionsButtonStyle())
-            .environment(\.appearance, CustomerCenterConfigData.Appearance(mode: .system))
+            .environment(\.appearance, CustomerCenterConfigData.Appearance(
```
fire-at-will, line 55: Nit: It'd be nice to extract this `CustomerCenterConfigData.Appearance` to a constant somewhere so that it can be reused across previews

fire-at-will (approved):
The changes here look good to me!

RevenueCatUI/CustomerCenter/Data/CustomerCenterEnvironment.swift
```diff
@@ -43,7 +43,13 @@ extension CustomerCenterConfigData.Localization {
 extension CustomerCenterConfigData.Appearance {
 
     /// Default ``CustomerCenterConfigData.Appearance`` value for Environment usage
-    public static let `default` = CustomerCenterConfigData.Appearance(mode: .system)
+    public static let `default` = CustomerCenterConfigData.Appearance(
+        accentColor: .init(light: "#ffffff", dark: "#000000"),
+        textColor: .init(light: "#000000", dark: "#ffffff"),
+        backgroundColor: .init(light: "#000000", dark: "#ffffff"),
+        buttonTextColor: .init(light: "#000000", dark: "#ffffff"),
+        buttonBackgroundColor: .init(light: "#000000", dark: "#ffffff")
```
tonidero, line 51: Should the defaults be all null? So we fallback to the system colors?

RevenueCatUI/CustomerCenter/ManageSubscriptionsButtonStyle.swift
```diff
@@ -17,42 +17,28 @@ import Foundation
...
-        case .custom(accentColor: let accentColor, backgroundColor: _, textColor: _):
-            return colorScheme == .dark ? accentColor.dark.underlyingColor : accentColor.light.underlyingColor
-        }
+        Button(action: { configuration.trigger() }, label: {
+            configuration.label.frame(width: 300)
+        })
+        .buttonStyle(.borderedProminent)
+        .controlSize(.large)
+        .applyIf(background != nil, apply: { $0.tint(background) })
+        .applyIf(textColor != nil, apply: { $0.foregroundColor(textColor) })
```
tonidero, line 41: I see we're just not applying custom colors if they are null... But I'm wondering if we should customize the default style of these to be the accentColor or something like that... But I'm ok going with just the defaults for now.

tonidero (approved):
Not super-familiar with swift UI but changes LGTM!

RevenueCatUI/CustomerCenter/Data/CustomerCenterEnvironment.swift
vegaro: right right, not sure what I was thinking!

RevenueCatUI/CustomerCenter/ManageSubscriptionsButtonStyle.swift
vegaro: That's actually done by `.boderedProminent` and it's the reason why I am using that style by default, otherwise it would just look like text. So I think we are covered

PR #4140, opened 2024-07-31
Deploy purchaserTester: clean up dry-run parameter

aboedo:
Cleans up the deploy purchase tester job's treatment of `dry-run`: it's now a parameter of the job instead of a separate job entirely. 

This is based off of @tonidero and @vegaro's feedback in #4131

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,9 +38,6 @@ executors:
       xcode_version:
         type: string
         default: '15.3'
-      install_swiftlint:
-        type: boolean
-        default: true
     macos:
       xcode: << parameters.xcode_version >>
     resource_class: macos.m1.medium.gen1
@@ -405,7 +402,6 @@ jobs:
     executor:
       name: macos-executor
       xcode_version: '14.3.0'
-      install_swiftlint: false
 
     steps:
       - checkout
@@ -430,7 +426,6 @@ jobs:
     executor:
       name: macos-executor
       xcode_version: '14.1.0'
-      install_swiftlint: false
 
     steps:
       - checkout
@@ -491,7 +486,6 @@ jobs:
     executor:
       name: macos-executor
       xcode_version: '14.3.0'
-      install_swiftlint: false
 
     steps:
       - checkout
@@ -521,7 +515,6 @@ jobs:
     executor:
       name: macos-executor
       xcode_version: '14.3.0'
-      install_swiftlint: false
 
     steps:
       - checkout
@@ -555,7 +548,6 @@ jobs:
     executor:
       name: macos-executor
       xcode_version: '15.2.0'
-      install_swiftlint: false
 
     steps:
       - checkout
@@ -636,7 +628,6 @@ jobs:
     executor:
       name: macos-executor
       xcode_version: '15.2.0'
-      install_swiftlint: false
 
     steps:
       - checkout
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -478,7 +478,8 @@ jobs:
           command: bundle exec fastlane run_api_tests
 
   spm-receipt-parser:
-    <<: *base-job
+    executor:
+      name: macos-executor
     steps:
       - checkout
       - run:
@@ -487,19 +488,15 @@ jobs:
           no_output_timeout: 30m
 
   spm-revenuecat-ui-ios-15:
-    <<: *base-job
-    parameters:
-      xcode_version:
-        type: string
-        default: '14.3.0'
-      install_swiftlint:
-        type: boolean
-        default: false
+    executor:
+      name: macos-executor
+      xcode_version: '14.3.0'
+      install_swiftlint: false
 
     steps:
       - checkout
       - install-dependencies:
-          install_swiftlint: << parameters.install_swiftlint >>
+          install_swiftlint: false
       - update-spm-installation-commit
       - run:
           name: SPM RevenueCatUI Tests
@@ -521,19 +518,15 @@ jobs:
           destination: scan-test-output
 
   spm-revenuecat-ui-ios-16:
-    <<: *base-job
-    parameters:
-      xcode_version:
-        type: string
-        default: '14.3.0'
-      install_swiftlint:
-        type: boolean
-        default: false
+    executor:
+      name: macos-executor
+      xcode_version: '14.3.0'
+      install_swiftlint: false
 
     steps:
       - checkout
       - install-dependencies:
-          install_swiftlint: << parameters.install_swiftlint >>
+          install_swiftlint: false
       - update-spm-installation-commit
       - run:
           name: SPM RevenueCatUI Release Build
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -44,6 +44,9 @@ aliases:
       install_swiftlint:
         type: boolean
         default: true
+      dry_run:
+        type: boolean
+        default: false
     environment:
```
aboedo, line 50: this bit I don't _love_: I had to make it a parameter of _base job_ rather than the concrete deploy purchase tester job. 

It doesn't actually affect anything other than the fact that all jobs get this param now. Maybe that's deceiving in that if you run the other deploy stuff with `dry_run: true`, it's not an actual dry run, and that's a bit concerning. 

But I couldn't find a better way of keeping it specific to the job: the thing is that we inherit all the stuff from the base job, and that includes `parameters`. If I declare it again, it breaks the config. 

I could also not inherit from base job and just copy/paste all of its things, but I feel like that's worse than what we had before. 

I'm open to other ideas, especially since overall this doesn't entirely feel better to me than having two almost identical jobs. I'm worried that a future dev could think that `dry_run` on all jobs actually means dry run.

tonidero: Right, I certainly don't love it either and it can be confusing... Not aware of any other options off the top of my head, so I agree with you we might as well leave it as 2 separate jobs.

vegaro: I think the base job should actually be an executor not an alias:

```
executors:
  macos-executor:
    parameters:
      xcode_version:
        type: string
        default: '15.3'
      install_swiftlint:
        type: boolean
        default: true
    macos:
      xcode: << parameters.xcode_version >>
    resource_class: macos.m1.medium.gen1
    environment:
      CIRCLECI_TESTS_GENERATE_SNAPSHOTS: << pipeline.parameters.generate_snapshots >>
      CIRCLECI_TESTS_GENERATE_REVENUECAT_UI_SNAPSHOTS: << pipeline.parameters.generate_revenuecatui_snapshots >>
    working_directory: ~/purchases-ios
    shell: /bin/bash --login -o pipefail
```

We can keep the alias so we don't have to update all jobs now:

```
aliases:
  base-job: &base-job
    executor:
      name: macos-executor
```

Then you can have custom parameters for the job

```
 deploy-purchase-tester:
    executor:
      name: macos-executor
    parameters:
      dry_run:
        type: boolean
        default: false
```

aboedo: ohh that's great, I'll give it a shot. Thanks!

aboedo: Updated with this approach, looks soooo much better! I hope that it does work. If it does I'll do another follow-up to remove the old base job alias

aboedo:
@RCGitBot please test

.circleci/config.yml
aboedo: I ran into some issues with failing tests when mixing and matching this approach and the alias... And this approach is just better than the alias, so I went ahead and replaced it altogether

.circleci/config.yml
```diff
@@ -32,23 +32,22 @@ parameters:
     type: string
     default: ""
 
-aliases:
-  base-job: &base-job
-    resource_class: macos.m1.medium.gen1
-    macos:
-      xcode: << parameters.xcode_version >>
+executors:
```
aboedo, line 35: entirely replaces the base job alias with the executor

.circleci/config.yml
```diff
@@ -32,23 +32,22 @@ parameters:
...
+  macos-executor:
     parameters:
       xcode_version:
         type: string
         default: '15.3'
-      install_swiftlint:
-        type: boolean
-        default: true
```
aboedo, line 46: This was honestly a pretty deceiving one - you would have to re-define it as a param, but then also remember to pass it in, which was kinda weird (you'll see what I mean in other comments). 
And it's not super useful, so I removed it so we can use it at the job level instead

.circleci/config.yml
```diff
@@ -396,14 +395,9 @@ commands:
...
-    parameters:
-      xcode_version:
-        type: string
-        default: '14.3.0'
-      install_swiftlint:
-        type: boolean
-        default: false
+    executor:
+      name: macos-executor
+      xcode_version: '14.3.0'
```
aboedo, line 400: regular usage is just so much cleaner

.circleci/config.yml
```diff
@@ -427,14 +419,9 @@ jobs:
...
 
   spm-xcode-14-1:
-    <<: *base-job
-    parameters:
-      xcode_version:
-        type: string
-        default: '14.1.0'
-      install_swiftlint:
-        type: boolean
-        default: false
```
aboedo, line 437: these jobs didn't actually install swiftlint, even if you pass in true here, since that happens if you call install dependencies, so you could have thought that it should work and gone crazy trying to figure it out

.circleci/config.yml
```diff
@@ -488,19 +479,14 @@ jobs:
...
-    parameters:
-      xcode_version:
-        type: string
-        default: '14.3.0'
-      install_swiftlint:
-        type: boolean
-        default: false
+    executor:
+      name: macos-executor
+      xcode_version: '14.3.0'
```
aboedo, line 484: again, this just looks nicer, and you don't have to redefine the parameter every time including the type

.circleci/config.yml
```diff
@@ -488,19 +479,14 @@ jobs:
...
 
     steps:
       - checkout
       - install-dependencies:
-          install_swiftlint: << parameters.install_swiftlint >>
+          install_swiftlint: false
```
aboedo, line 489: here I just use the param directly, I feel like this is just much clearer and less error-prone since you don't have to pass in the param

.circleci/config.yml
```diff
@@ -1233,7 +1193,7 @@ jobs:
       - update-spm-installation-commit
       - run:
           name: Submit Purchase Tester
-          command: bundle exec fastlane deploy_purchase_tester dry_run:true
+          command: bundle exec fastlane deploy_purchase_tester dry_run:<< parameters.dry_run >>
```
aboedo, line 1196: and this is the actual meat of the PR: being able to define a parameter at the job level easily and then use it

.circleci/config.yml
```diff
@@ -1419,5 +1380,7 @@ workflows:
       - installation-tests-carthage
       - installation-tests-xcode-direct-integration
       - installation-tests-receipt-parser
-      - deploy-purchase-tester-dry-run
       - api-tests
+      - deploy-purchase-tester:
+          dry_run: true
```
aboedo, line 1385: added the param and then moved it to the bottom so it looked a liiiitle nicer

aboedo:
@RCGitBot please tet

aboedo:
@RCGitBot please test

aboedo:
oops in an earlier commit I didn't pass in the dry run param for deploy purchase tester so I did send out a purchase tester build to TF unintended

aboedo:
@RevenueCat/coresdk this is ready for another pass

tonidero (approved):
Very nice!!

.circleci/config.yml
JayShortway: Yea, I think it's not a bad idea to use aliases only sparingly, because of the inflexibility they often introduce.

JayShortway (approved):
Nice improvement!

PR #4141, opened 2024-07-31
Clean up API Testers

aboedo:
# API Testers cleanup

- Moves all API Testers into main RevenueCat workspace
- ensure that they all use the local build, not a hardcoded commit
- kill all the separate projects, create a single project with a framework per test (objc api tests, swift api tests, etc, etc)
- creates a single scheme that builds all of those targets in one pass
- update relevant CI (to do)

Here's what this actually looks like: a single scheme in Xcode, which you can test using command + U (or just build, since it effectively works too)

Screenshot 2024-07-31 at 6 09 33 PM

Screenshot 2024-07-31 at 6 09 44 PM

*Note:*
Okay, I'd directly recommend _not_ reviewing the diff, since it's kind of insane, and instead checking out the branch to see if things make sense.


aboedo:
Note: I haven't fixed the tests yet, so if we ask gitbot to test this it's going to go a bit nuts

aboedo:
@RCGitBot please test

aboedo:
I'm particularly curious about the carthage installation tests, since those are the ones that do look at all the schemes, so that's where the bigger change would be

aboedo:
haha called it, carthage failed. I'll mark this as draft again while I figure that out

aboedo:
@RCGitBot please test

aboedo:
Carthage installation tests also thankfully raised a number of places where we weren't doing platform checks correctly in CustomEntitlementsComputationAPITests and a couple of other places, and also they made me notice a bunch of places where the deployment targets weren't correctly set, so I also cleaned that up

tonidero (approved):
Nice cleanup! Had to update Xcode since the target for the tests is iOS 17.5, but that should be ok :)

aboedo:
@tonidero good looking out, I hadn't noticed that. I'll update the tests's deployment targets in a follow up

PR #4142, opened 2024-08-02
[Customer Center] Fix for disabled promo offer button

vegaro:
I noticed the purchase button for the promo offers is always disabled. Changing the order of the modifiers fixes it

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
@@ -310,6 +310,8 @@ struct ManageSubscriptionButton: View {
                 Text(path.title)
             }
         })
+        .buttonStyle(ManageSubscriptionsButtonStyle())
+        .disabled(self.viewModel.loadingPath != nil)
         .restorePurchasesAlert(isPresented: self.$viewModel.showRestoreAlert)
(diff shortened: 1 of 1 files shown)


RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -322,8 +324,6 @@ struct ManageSubscriptionButton: View {
                                  product: promotionalOfferData.product,
                                  promoOfferDetails: promotionalOfferData.promoOfferDetails)
         })
-        .buttonStyle(ManageSubscriptionsButtonStyle())
-        .disabled(self.viewModel.loadingPath != nil)
```
tonidero, line 326: Hmm I'm not sure I understand why doing this before actually fixes this... But I'm ok with the fix 👍

tonidero (approved):

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
vegaro: It's because it applies it to the sheet as well, and the sheet is the one presenting the promo offer, so basically everything in the sheet is disabled

tonidero: Ahh makes sense. Thanks for explaining!

PR #4144, opened 2024-08-05
[Customer Center] Fixes an extra blank screen after popping FeedbackSurveyView on iOS 18.

JayShortway:
## Cause
The issue in the [ticket](https://linear.app/revenuecat/issue/SDK-3519/extra-blank-screen-after-back-button-on-cancel-subscription) is caused by the use of `.constant` and `.onDisappear`, which probably resulted in a timing issue where the `feedbackSurveyData` is not set to `nil` quickly enough after pressing the Back button. 

## Fix
The fix in this PR is to bind directly to the `.feedbackSurveyData` property. I've added an `.isNotNil` Binding extension for this, as it seemed useful to reuse in other parts of RevenueCatUI. Let me know if the `Binding+Extensions.swift` file should be placed somewhere else instead.  

---
Closes SDK-3519

### DIFF
--- a/RevenueCatUI/Binding+Extensions.swift
+++ b/RevenueCatUI/Binding+Extensions.swift
@@ -1,0 +1,23 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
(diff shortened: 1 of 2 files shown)


tonidero (approved):
Might be good to get an iOS dev eye on this, but LGTM!

MarkVillacampa (approved):

PR #4145, opened 2024-08-05
Update deployment targets for tests

aboedo:
Updates the deployment targets for a couple of test suites to match their corresponding frameworks. 

Specifically updates for `RevenueCatUITests` and for `AllAPITestsTests` (yeah, I don't love the name either, but that's the _test suite_ tied to the AllAPITests framework)


aboedo:
@RCGitBot please test

aboedo:
@RevenueCat/coresdk bump

PR #4147, opened 2024-08-06
[Customer Center] UI polish for empty content

codykerns:
Attempting some UI polish

- Implements `ContentUnavailableView` for the views that have 'empty' content (with a visually similar backwards compatible version)
- Improves instructions for each platform for wrong-platform subscription management

| New           | Old           |
|:-------------:|:-------------:|
| New Version | Old Version |

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/BackwardsCompatibleContentUnavailableView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/BackwardsCompatibleContentUnavailableView.swift
@@ -28,7 +28,7 @@ struct BackwardsCompatibleContentUnavailableView: View {
 
     var body: some View {
 
-        if #available(iOS 17.0, *) {
+        if #available(iOS 17.0, macOS 14.0, tvOS 17.0, watchOS 10.0, *) {
             ContentUnavailableView(
                 title,
                 systemImage: systemImage,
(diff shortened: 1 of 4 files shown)


aboedo:
The code is :chefskiss: and I love how the view looks! Thanks for doing this Cody!

We'll have to update snapshots for this (@RevenueCat/coresdk could you remind me how that's done?). 

I think the only thing that we'll have to figure out is how to make the strings used here localizable, do we already have any of those coming from the backend that we can reuse? (Also question for CoreSDK)

JayShortway (commented):
Love it! 🤩 

Regarding localized strings, I believe the idea is to place them in `CustomerCenterConfigData.Localization`. 
Not sure how to update the snapshots though. 
(cc @aboedo)

tonidero:
This looks nice!
> We'll have to update snapshots for this

It's been a while but I believe you should be able to trigger it through either fastlane (`bundle exec fastlane generate_snapshots_RCUI`) or CircleCI (selecting this branch in CircleCI, then triggering a new job with a booleean `generate_revenuecatui_snapshots` parameter set to true). Lmk if I can help with this!

As for localization, what @JayShortway mentioned is correct! The idea is to have them as common strings which can be overriden by the backend response, but still have some default in the SDK.

codykerns:
Thanks all!

@aboedo since this PR just refactors the already hard-coded strings, should the localization changes be made in a separate PR?

tonidero (approved):

aboedo:
My bad, I'd missed that the strings already existed and that we don't even have snapshots for customer center. 
This is why we shouldn't let me review things, feel free to merge once tests pass! Looks like CI doesn't like something in the availability checks mentioning `visionOS`?

codykerns:
Since the failing test was related to snapshots (that I ran manually), and affects all of customer center currently and not just this PR, going to merge this. Standard tests are succeeding 👍

PR #4148, opened 2024-08-06
iOS15/SK2/ObserverMode Improvements

fire-at-will:
### Checklist


fire-at-will:
@RCGitBot please test

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -685,6 +682,30 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
         }
     }
 
+    private static func buildPaymentQueueWrapper(
+        systemInfo: SystemInfo,
+        operationDispatcher: OperationDispatcher
+    ) -> EitherPaymentQueueWrapper {
+        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *),
```
fire-at-will, line 689: If we like this approach, we should extract this logic out to a common location, perhaps in `StoreKitVersion`

PR #4149, opened 2024-08-06
Customer Center compilation flag

aboedo:
We have a long-running integration branch for Customer Center, long running integration branches usually get very tricky to merge as time passes on. 

This PR attempts to solve it by creating a single custom compiler flag that enables the Customer Center feature. 

Here is what the compilation flag looks like: 
image

Then all we need to do to integrate into main is to delete that flag. And re-add it for development. Easy peasy. 

And when we want to ship Customer Center for real, we just delete all usages of that flag. 

If the flag isn't defined, then the code evaluates to false and is simply skipped by the compiler. If it evaluates to true, then the code runs.


RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -6742,6 +6742,7 @@
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = iphoneos;
 				SUPPORTS_MACCATALYST = YES;
+				SWIFT_ACTIVE_COMPILATION_CONDITIONS = CUSTOMER_CENTER_ENABLED;
```
aboedo, line 6745: in order to merge to `main`, we just make a new PR that kills this line. And then we bring it back in the integration branch

RevenueCatUI/CustomerCenter/Abstractions/CustomerCenterPurchasesType.swift
```diff
@@ -11,6 +11,8 @@
 //
 //  Created by Cesar de la Vega on 18/7/24.
 
+#if CUSTOMER_CENTER_ENABLED
+
 import Foundation
```
aboedo, line 16: okay, I'll admit that it's annoying to have this in 30 files, but it legitimately took me just a couple of mins to add

aboedo: and we'll only need to remove it when we're ready to ship the feature

RevenueCat.xcodeproj/project.pbxproj
aboedo: Note that since the flag is only defined for the Xcode project, SPM and CocoaPods builds will already evaluate it to false and skip Customer Center. Carthage won't, however, since it reads from Xcode, so we should remove it to merge into `main`

aboedo:
If you wanna test out the branch, I'd recommend running things as usual once, then removing the flag from the project settings (but keeping it in the code), then running again to verify that that actually removes all of CustomerCenter. 

You might need to clean builds if Xcode doesn't realize that the flag affects more than one project automatically.

RevenueCatUI/CustomerCenter/Abstractions/CustomerCenterPurchasesType.swift
tonidero: I bet we can probably leave it out in some of these files... But better to be sure 👍

tonidero (approved):
Love this! And yeah, with this we can merge the integration branch back to main much more easily 👍

RevenueCatUI/CustomerCenter/Abstractions/CustomerCenterPurchasesType.swift
JayShortway: No idea if this is in the realm of possibilities, but would it be possible to have the the flag only around public entry points of the Customer Center feature? E.g. `Purchases.loadCustomerCenter()`, `CustomerCenterView()` and the `presentCustomerCenter()` modifier?

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1189,6 +1189,8 @@ public extension Purchases {
         await self.paywallEventsManager?.track(paywallEvent: paywallEvent)
     }
 
+    #if CUSTOMER_CENTER_ENABLED
+
     /// Used by `RevenueCatUI` to download customer center data
     func loadCustomerCenter() async throws -> CustomerCenterConfigData {
```
tonidero, line 1195: Ahh just noticed we're not wrapping the `CustomerCenterConfigData` into these flags right? We probably should to avoid exposing that API

RevenueCatUI/CustomerCenter/Abstractions/CustomerCenterPurchasesType.swift
tonidero: While that should be fine for `RevenueCat` (we would just need the flag for `loadCustomerCenter` and `CustomerCenterConfigData`), since `CustomerCenterConfigData` is used throughout `RevenueCatUI`, we would need to wrap many files there.

JayShortway: What if.. we only wrap the constructor of `CustomerCenterConfigData` (or make the constructor `public` if the flag exists, `internal` otherwise). We'd be exposing some extra types, but no dev would be able to actually do anything with them because they can't construct anything. 
(Just spitballing, lmk if this is dumb.)

tonidero: There are a bunch of internal types inside `CustomerCenterConfigData` so we would need to wrap a bunch of constructors... It would however, remove the need to wrap the code in `RevenueCatUI`... 

Having said that, we would still be exposing new types which can potentially change, even if they are unusable... So I don't think it's ideal 😓

JayShortway: Hmm yea, not ideal indeed. 🤔

JayShortway: Yeah I think you're making very good points. Not worth the extra effort to do fine tuning imo. Also, we don't have a good idea what fine tuning would even look like yet.

Issue #4150, opened 2024-08-07
iOS 18 Beta 5 SwiftUICore/GraphHost.swift:215: Fatal error: no current graph host


rocxteady:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
I am receiving this weird crash after launching the app while showing the paywall. It happens right after the paywall is seen.

1. Environment
   1. Platform: iOS
   2. SDK version: (main branch from [spm](https://github.com/RevenueCat/purchases-ios-spm), commit: [348258a](https://github.com/rocxteady/purchases-ios-spm/commit/348258a41f9de0edbe8eb4d70ed90784f15445b1)
   3. StoreKit version:
      - [] StoreKit 1
      - [x] StoreKit 2 (_enabled with `usesStoreKit2IfAvailable(true)`_)
   4. OS version: 18 Beta 5
   5. Xcode version: Xcode 15.4, Xcode 16 Beta 4, Beta 5
   6. Device and/or simulator:
      - [x] Device
      - [x] Simulator
   7. Environment:
      - [x] Sandbox
      - [x] TestFlight
      - [] Production
   8. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue. **Complete logs with `Purchases.logLevel = .verbose` will help us debug this issue.**

Console prints this: 
```SwiftUICore/GraphHost.swift:215: Fatal error: no current graph host```
Xcode says this:
```
PressReaderStaging`static PressReaderApp.$main():
    0x100269614 <+0>:  stp    x20, x19, [sp, #-0x20]!
    0x100269618 <+4>:  stp    x29, x30, [sp, #0x10]
    0x10026961c <+8>:  add    x29, sp, #0x10
    0x100269620 <+12>: mov    x0, #0x0
    0x100269624 <+16>: bl     0x1002676ac               ; type metadata accessor for PressReaderStaging.PressReaderApp at <compiler-generated>
    0x100269628 <+20>: mov    x20, x0
[...]

PR #4151, opened 2024-08-07
Fix crash on iOS 18 beta 5

MarkVillacampa:
Potential fix for https://github.com/RevenueCat/purchases-ios/issues/4150

The extra `self` seems unnecessary.


RevenueCatUI/Modifiers/ViewExtensions.swift
```diff
@@ -90,8 +90,6 @@ extension View {
         if enabled {
             if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
                 ViewThatFits(in: axis.scrollViewAxis) {
-                    self
-
```
joshdholtz, line 94: This... makes me nervous to remove even thought it does look weird that `self` is in here twice 😬 

I feel like it had to provide some value so I'm going to go back into the PRs to see if there is any knowledge there

PR #4158, opened 2024-08-08
[Paywalls] Improve locale consistency

joshdholtz:
⚠️ Waiting for #4188 to be merged first (because needs all the tests to pass)

## Motivation

Fix cases where paywall was showing multiple languages at once and random languages.

## Description

👉  For a demo of the problem and solution, look in Slack in the paywalls or SDK channel for the latest Loom that I posted

### Flow of paywall localization logic (since it can be confusing)

1. [PaywallView](https://github.com/RevenueCat/purchases-ios/pull/4158/files#diff-ea75fe7280803e15b975d702a16c6851ffc2bbed01b7b196bf0fb80d75b4b332) starts with `Locale.current` but then calls `offering.validatedPaywall()` which will return the new `displayedLocale: Locale` variable
2. [PaywallData+Validation](https://github.com/RevenueCat/purchases-ios/pull/4158/files#diff-452790cf5287a44f75a4b2260254db2281a066faeea2b3e441a007ec60522b6b) returns `displayedLocale` if its a valid paywall, otherwise it return `Locale.current` for fallback paywall
3. [PaywallData+Localization](https://github.com/RevenueCat/purchases-ios/pull/4158/files#diff-e620baead94f7acad56d931c485d7691c4f909be3fe625e104528a2702cca2a0) has a lot of the new logic
    - New `locale: Locale` computer property that it determines the locale from the `localizedConfiguration` or `localizedConfigurationsByTier`
    -  New ` defaultLocalizedConfiguration` that uses the developer set `defaultLocale` (from `PaywallData`) to find the default localization
    - Updated `localizedConfiguration<Value>()` to take a new `defaultLocalization` to put in its priority (and log it if its used). If for some reason that's not found, it will use the legacy `fallbackLocalization` (which is the first localization it finds or english)
4. After all of that, the `displayedLocale` is passed into the actually SwiftUI `LoadedOfferingPaywallView` to show this locale everywhere

### Problem 1 - Inconsistent languages displayed at once

The user defined localizations and the pre-localized content in the SDK (like links in the footer and package duration names) could should different localizations. The choice of which locale was happening in two different spots.

Since the SDK has all possible localizations that the dashboard offers, the locale SDK displays from the dashboard will be chosen for the pre-localized content.

### Problem 2 - Showing random languages if not supported

We decided early on to not use a `default_locale` but that has hurt us. The solution to use was use `localizations.first` to show the first locale in the list and the intention that this was the first locale the user did in the paywall. That was not the case, however. The order of the localizations list was indeterministic so it should should a random language (even to the point where different random languages would show the the same user).

We are now introducing a `default_locale` that the developer can set in the paywall. If the locale the user is requesting is not translated by the developer, it will use the `default_locale`

### Problem 3 - Footer not using correct localization from the bundle

There were times where some parts of the footer itself (mainly the restore button) would be a different language than the terms of service and privacy policy buttons. This was due to a code issue where we were using `Bundle.module`. Instead, we needed to specify a localization bundle specifically that we wanted to use.

### DIFF
--- a/RevenueCatUI/Data/Localization.swift
+++ b/RevenueCatUI/Data/Localization.swift
@@ -62,9 +62,11 @@ enum Localization {
         return formatter.string(from: subscriptionPeriod.components) ?? ""
     }
 
+    // TODO: Josh
     static func localized(
         packageType: PackageType,
-        locale: Locale = .current
+        locale: Locale = .current,
+        defaultLocale: Locale = Self.fallbackLocale
     ) -> String? {
         guard let key = packageType.localizationKey else { return nil }
 
@@ -80,7 +82,7 @@ enum Localization {
         return value(
             locale: locale,
             // Or defaults to english
-            default: value(locale: Self.defaultLocale, default: nil)
+            default: value(locale: defaultLocale, default: nil)
         )
     }
 
@@ -194,9 +196,9 @@ private extension Localization {
     static let unitAbbreviationLengthPriorities = [ 2, 3 ]
 
     /// For falling back in case language isn't localized.
-    static let defaultLocale: Locale = .init(identifier: Self.defaultLocaleIdentifier)
+    static let fallbackLocale: Locale = .init(identifier: Self.fallbackLocaleIdentifier)
 
-    private static let defaultLocaleIdentifier: String = Locale.preferredLanguages.first ?? "en_US"
+    private static let fallbackLocaleIdentifier: String = Locale.preferredLanguages.first ?? "en_US"
 
 }
 
--- a/RevenueCatUI/Data/Localization.swift
+++ b/RevenueCatUI/Data/Localization.swift
@@ -64,16 +64,16 @@ enum Localization {
 
     static func localized(
         packageType: PackageType,
-        locale: Locale = .current,
-        defaultLocale: Locale = Self.fallbackLocale
+        locale: Locale,
+        fallbackLocale: Locale = Self.fallbackLocale
     ) -> String? {
         guard let key = packageType.localizationKey else { return nil }
 
-        func value(locale: Locale, default: String?) -> String {
+        func value(locale: Locale, fallback: String?) -> String {
             Self
                 .localizedBundle(locale)
                 .localizedString(forKey: key,
-                                 value: `default`,
+                                 value: fallback,
                                  table: nil)
         }
 
@@ -81,7 +81,7 @@ enum Localization {
         return value(
             locale: locale,
             // Or defaults to english
-            default: value(locale: defaultLocale, default: nil)
+            fallback: value(locale: fallbackLocale, fallback: nil)
         )
     }
 
--- a/RevenueCatUI/Data/Localization.swift
+++ b/RevenueCatUI/Data/Localization.swift
@@ -80,7 +80,7 @@ enum Localization {
         // Returns the localized string
         return value(
             locale: locale,
-            // Or falls back to first localization or english
+            // Or falls back to first localization, and if unavailable, english
             fallback: value(locale: fallbackLocale, fallback: nil)
         )
     }
--- a/RevenueCatUI/Data/Localization.swift
+++ b/RevenueCatUI/Data/Localization.swift
@@ -80,7 +80,7 @@ enum Localization {
         // Returns the localized string
         return value(
             locale: locale,
-            // Or defaults to english
+            // Or falls back to first localization or english
             fallback: value(locale: fallbackLocale, fallback: nil)
         )
     }
--- a/RevenueCatUI/Data/Localization.swift
+++ b/RevenueCatUI/Data/Localization.swift
@@ -62,7 +62,6 @@ enum Localization {
         return formatter.string(from: subscriptionPeriod.components) ?? ""
     }
 
-    // TODO: Josh
     static func localized(
         packageType: PackageType,
         locale: Locale = .current,
--- a/RevenueCatUI/Data/PaywallData+Validation.swift
+++ b/RevenueCatUI/Data/PaywallData+Validation.swift
@@ -45,7 +45,7 @@ extension Offering {
 extension Offering {
 
     /// - Returns: a validated paywall suitable to be displayed, and any associated error.
-    func validatedPaywall(locale: Locale = .current) -> ValidationResult {
+    func validatedPaywall(locale: Locale) -> ValidationResult {
         if let paywall = self.paywall {
             switch paywall.validate() {
             case let .success(template):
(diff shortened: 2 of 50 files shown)


RevenueCatUI/Data/Localization.swift
```diff
@@ -64,23 +64,24 @@ enum Localization {
 
     static func localized(
         packageType: PackageType,
-        locale: Locale = .current
+        locale: Locale,
```
joshdholtz, line 67: 1. Defaulting to `.current` is hard to traceback for decision making and debugging to getting rid of it

RevenueCatUI/Data/PaywallData+Validation.swift
```diff
@@ -45,24 +46,26 @@ extension Offering {
 extension Offering {
 
     /// - Returns: a validated paywall suitable to be displayed, and any associated error.
-    func validatedPaywall(locale: Locale = .current) -> ValidationResult {
+    func validatedPaywall(locale: Locale) -> ValidationResult {
         if let paywall = self.paywall {
             switch paywall.validate() {
             case let .success(template):
-                return (paywall, template, nil)
+                return (paywall, paywall.locale ?? locale, template, nil)
```
joshdholtz, line 53: `paywall.locale` SHOULD always exist at this point but needed to do something like this to make it happy 🤷‍♂️

RevenueCatUI/Views/FooterView.swift
```diff
@@ -114,13 +125,13 @@ struct FooterView: View {
...
 
-    private static func allPlansButton(_ binding: Binding<Bool>) -> some View {
+    private static func allPlansButton(_ binding: Binding<Bool>, bundle: Bundle) -> some View {
         Button {
             withAnimation(Constants.toggleAllPlansAnimation) {
                 binding.wrappedValue.toggle()
             }
         } label: {
-            Text("All subscriptions", bundle: .module)
+            Text("All subscriptions", bundle: bundle)
```
joshdholtz, line 134: Things like this were the cause of not putting correct localizations and so needing to "add" the localizations to Xcode

Sources/Paywalls/PaywallData+Localization.swift
```diff
@@ -15,43 +15,72 @@ import Foundation
 
 public extension PaywallData {
 
+    /// - Returns: the `Locale` being used with ``PaywallData/LocalizedConfiguration-swift.struct``
+    var locale: Locale? {
+        let singleTier = self.localizedConfiguration(for: Self.localesOrderedByPriority)?.0
+        let multiTier = self.localizedConfigurationByTier(for: Self.localesOrderedByPriority)?.0
+
+        return singleTier ?? multiTier
+    }
```
joshdholtz, line 24: This is _slightly gross_ but it gets the job done 🤷‍♂️

Sources/Paywalls/PaywallData+Localization.swift
joshdholtz, line 1: One thing I don't love here is how all of these variables are computed 😬 If they are called multiple times, it reruns all of this logic again. I don't really feel comfortable changing all of that in this PR yet

Sources/Paywalls/PaywallData+Localization.swift
```diff
@@ -15,43 +15,72 @@ import Foundation
...
     var localizedConfigurationByTier: [String: LocalizedConfiguration]? {
-        return self.localizedConfigurationByTier(for: Self.localesOrderedByPriority)
+        return self.localizedConfigurationByTier(for: Self.localesOrderedByPriority)?.1
     }
 
     // Visible for testing
-    internal func localizedConfiguration(for preferredLocales: [Locale]) -> LocalizedConfiguration? {
+    internal func localizedConfiguration(
+        for preferredLocales: [Locale]
+    ) -> (Locale, LocalizedConfiguration)? {
```
jamesrb1, line 43: Consider labeling the tuple parameters so that we can refer to them by name rather than order (ie instead of `?.0`, ``?.locale`).
```suggestion
    ) -> (locale: Locale, localizedConfiguration: LocalizedConfiguration)? {
```

Sources/Paywalls/PaywallData.swift
```diff
@@ -37,6 +37,9 @@ public struct PaywallData {
         set { self._revision = newValue }
     }
 
+    /// The default localue identifier for this paywall.
```
jamesrb1, line 40: typo
```suggestion
    /// The default locale identifier for this paywall.
```

Sources/Paywalls/PaywallData+Localization.swift
jamesrb1: I guess the computed properties might support changing localizations on the fly. Not sure how often that happens or if it's even a use case worth supporting though.

RevenueCatUI/Views/LoadingPaywallView.swift
```diff
@@ -44,7 +44,8 @@ struct LoadingPaywallView: View {
             fonts: DefaultPaywallFontProvider(),
             displayCloseButton: self.displayCloseButton,
             introEligibility: Self.introEligibility,
-            purchaseHandler: Self.purchaseHandler
+            purchaseHandler: Self.purchaseHandler,
+            locale: Locale.current
```
jamesrb1, line 48: Should we be defaulting to `Locale.preferredLocales.first ?? Locale.current`? (access restrictions aside.)

jamesrb1 (approved):
Very nice!!

tonidero (approved):
Not an expert, but this makes sense. Great fixes!

RevenueCatUI/Data/PaywallViewConfiguration.swift
```diff
@@ -29,7 +30,8 @@ struct PaywallViewConfiguration {
         fonts: PaywallFontProvider = DefaultPaywallFontProvider(),
         displayCloseButton: Bool = false,
         introEligibility: TrialOrIntroEligibilityChecker? = nil,
-        purchaseHandler: PurchaseHandler
+        purchaseHandler: PurchaseHandler,
+        locale: Locale = .current
```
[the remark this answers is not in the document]
joshdholtz: That's a fair point! I removed it 😊

RevenueCatUI/Views/LoadingPaywallView.swift
joshdholtz: Ah, probably! The loading view is all skeleton stuff anyway so there isn't _actually_ any text anywhere but that does feel better 😇

joshdholtz:
@RCGitBot please test

joshdholtz:
@RCGitBot please test

joshdholtz:
@RCGitBot please test

joshdholtz:
@RCGitBot please test

joshdholtz:
@RCGitBot please test

joshdholtz:
@RCGitBot please test

joshdholtz:
@RCGitBot please test

joshdholtz:
@RCGitBot please test

joshdholtz:
@RCGitBot please test

PR #4161, opened 2024-08-08
[Customer Center] Use system navigation titles

codykerns:
- Uses system navigation titles instead of text views for screen titles
- Display subtitle for the manage subscriptions screen
- improves preview test data with some standard default colors
- fixes progress view tint color

| New           | Old           |
|:-------------:|:-------------:|
| New Version | Old Version |


jamesrb1 (approved):
Look like good changes to me.

PR #4163, opened 2024-08-09
[Paywalls] Make iOS version calculation lazy

MarkVillacampa:
Since iOS version cannot change at runtime, we can make it a stored variable so we only compute it once.

### DIFF
--- a/RevenueCatUI/Modifiers/ViewExtensions.swift
+++ b/RevenueCatUI/Modifiers/ViewExtensions.swift
@@ -30,7 +30,7 @@ extension View {
 }
 
 private var isIOSVersionWithCrash: Bool = {
-    // There is a bug in iOS 18 beta 5 (as of writing, future versions uncofirmed) that causes a crash.
+    // There is a bug in iOS 18 beta 5 (as of writing, future versions unconfirmed) that causes a crash.
     // This has been reporte to Apple as FB14699941.
     // Until this is fixed, we're rolling back to pre-iOS 16 behavior for this view.
     // More information and discussion here: https://github.com/RevenueCat/purchases-ios/issues/4150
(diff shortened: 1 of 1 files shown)


RevenueCatUI/Modifiers/ViewExtensions.swift
```diff
@@ -29,6 +29,15 @@ extension View {
 
 }
 
+private var isIOSVersionWithCrash: Bool = {
+    // There is a bug in iOS 18 beta 5 (as of writing, future versions uncofirmed) that causes a crash.
```
tonidero, line 33:
```suggestion
    // There is a bug in iOS 18 beta 5 (as of writing, future versions unconfirmed) that causes a crash.
```

RevenueCatUI/Modifiers/ViewExtensions.swift
```diff
@@ -29,6 +29,15 @@ extension View {
...
+    // This has been reporte to Apple as FB14699941.
```
tonidero, line 34:
```suggestion
    // This has been reported to Apple as FB14699941.
```

tonidero (approved):
LGTM!

RevenueCatUI/Modifiers/ViewExtensions.swift
aboedo: lol that's what happens when you let me write things

RevenueCatUI/Modifiers/ViewExtensions.swift
aboedo: oops

aboedo (approved):

jamesrb1 (approved):

RevenueCatUI/Modifiers/ViewExtensions.swift
```diff
@@ -29,6 +29,15 @@ extension View {
 
 }
 
+private var isIOSVersionWithCrash: Bool = {
```
NachoSoto, line 32: This doesn't need to be a variable 😇 it'll be an error with Swift 6

PR #4165, opened 2024-08-09
[Customer Center] Improving customer center buttons

codykerns:
Various improvements around buttons for Customer Center. Several changes for the case where the customer center is presented full screen but wasn't dismiss-able.

**NoSubscriptionsView**
- Uses 'dismiss' localized string instead of 'cancel' for dismiss button

**Button styles**
- Renames existing `ManageSubscriptionButtonStyle` to `ProminentButtonStyle` as it's used elsewhere
- Adds a new reusable `SubtleButtonStyle` button style (standard, borderless style)
- Slightly thickens button fonts to more closely match buttons in system modals
- Slightly increases corner radius to more closely match buttons in system modals

| New | Old |
|-----|-----|
|  |  |

**WrongPlatformView**
- Adds an explicit 'dismiss' button

| New | Old |
|-----|-----|
|  |  |

**ManageSubscriptionsView**
- Fixes SDK-3537. Adds a toolbar 'dismiss' button to make sure it's dismiss-able when presented in full screen (or in scenarios where a swipe to dismiss isn't supported)

| New | Old |
|-----|-----|
|  |  |


RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
```diff
@@ -70,12 +70,13 @@ struct NoSubscriptionsView: View {
                     showRestoreAlert = true
                 }
                 .restorePurchasesAlert(isPresented: $showRestoreAlert)
-                .buttonStyle(ManageSubscriptionsButtonStyle())
+                .buttonStyle(ProminentButtonStyle())
 
-                Button(localization.commonLocalizedString(for: .cancel)) {
+                Button(localization.commonLocalizedString(for: .dismiss)) {
```
codykerns, line 75: I changed this to use `dismiss` instead of cancel, as it's a dismiss button

RevenueCatUI/CustomerCenter/ButtonStyles.swift
```diff
@@ -37,24 +37,56 @@ struct ManageSubscriptionsButtonStyle: PrimitiveButtonStyle {
...
+    func makeBody(configuration: PrimitiveButtonStyleConfiguration) -> some View {
+        let background = color(from: appearance.buttonBackgroundColor, for: colorScheme)
+
+        Button(action: { configuration.trigger() }, label: {
+            configuration.label.frame(maxWidth: .infinity)
+        })
+        .font(.body.weight(.medium))
+        .controlSize(.large)
+        .padding(.vertical)
+        .applyIf(background != nil, apply: { $0.foregroundColor(background) })
```
codykerns, line 68: this isn't using the button text color, because the subtle style is inverted (transparent background, with accented text instead of accented background and customizable text). since I think the intention is for the buttons to be displayed alongside each other with a similar theme, I thought this made the most sense

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -107,6 +110,16 @@ struct ManageSubscriptionsView: View {
         .task {
             await loadInformationIfNeeded()
         }
+        .toolbar {
```
codykerns, line 113: using `.topBarLeading` is pretty subjective. not sure if the placement, visibility, etc. needs to be customizable outside of the dismiss text

jamesrb1:
Looks good, will give this a thoughtful review Monday morning.

RevenueCatUI/CustomerCenter/ButtonStyles.swift
```diff
@@ -37,24 +37,56 @@ struct ManageSubscriptionsButtonStyle: PrimitiveButtonStyle {
...
+
+    func makeBody(configuration: PrimitiveButtonStyleConfiguration) -> some View {
+        let background = color(from: appearance.buttonBackgroundColor, for: colorScheme)
+
+        Button(action: { configuration.trigger() }, label: {
+            configuration.label.frame(maxWidth: .infinity)
+        })
+        .font(.body.weight(.medium))
+        .controlSize(.large)
+        .padding(.vertical)
```
jamesrb1, line 67: Probably better for padding to be controlled outside the button style? Or if not, should it also be horizontally padded?

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
jamesrb1: I think this makes sense.

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
```diff
@@ -58,6 +63,15 @@ struct WrongPlatformView: View {
...
                 )
 
+                Spacer()
+
+                Button {
+                    dismiss()
+                } label: {
+                    Text(localization.commonLocalizedString(for: .dismiss))
+                }
+                .buttonStyle(SubtleButtonStyle())
```
jamesrb1, line 73: imo this should also be in a leading toolbar position.

jamesrb1 (approved):
I think these are good changes, nice attention to detail. Pre-approving, but please see comments.

Side comment: I think this title area needs some UI design changes - it looks a bit cramped (localization may be hard), and having the dismiss button the same size and on the same as the screen's title makes the visual hierarchy not great.

image

RevenueCatUI/CustomerCenter/ButtonStyles.swift
codykerns: great catch, fixed 👍

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
```diff
@@ -70,17 +71,31 @@ struct NoSubscriptionsView: View {
...
+        if #available(iOS 16.0, *) {
+            NavigationStack {
+                content
+            }
+        } else {
+            NavigationView {
+                content
+            }
+        }
     }
```
jamesrb1, line 99: This view will _always_ be presented modally in some manner? If so this is fine, but if someone were to add this to an existing navigation stack, weird things could happen.

Additionally, I see we've used this pattern twice, if we use it a third time, should consider creating a component we can reuse for this.

jamesrb1 (approved):
Looks good!

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
codykerns: The modal question is one we'll need to answer, good call 👍 

Agreed on the reusable component - will pull this out into a new component in a separate PR.

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -382,7 +387,6 @@ struct SubscriptionDetailsView_Previews: PreviewProvider {
         )
         .previewDisplayName("Subscription Details - Monthly")
         .padding()
-
```
vegaro, line 385: According to our swift style guide we keep a line at the top and bottom of types

vegaro (approved):

PR #4166, opened 2024-08-10
Customer center improvements

jamesrb1:
I started going through some of the great new Customer Center feature. Just a few minor changes, I may submit more in the coming days/weeks.

Changes here:

- Avoid sorting an entire array to find the smallest value
- Use URLComponents to construct a URL
- Made a couple of properties `private(set)`


RevenueCatUI/CustomerCenter/URLUtilities.swift
```diff
@@ -27,17 +27,29 @@ enum URLUtilities {
...
+        var components = URLComponents()
+        components.scheme = "mailto"
+        components.path = email
 
-        if let url = URL(string: urlString),
-           UIApplication.shared.canOpenURL(url) {
-            return url
+        var queryItems: [URLQueryItem] = []
+
+        if !subject.isEmpty {
```
JayShortway, line 38: Is there an easy way to check whether a string is blank? Couldn't find anything built-in so far. If not, I think `isEmpty` is fine.

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -35,15 +35,15 @@ import RevenueCat
...
+    private(set) var state: CustomerCenterViewState {
         didSet {
             if case let .error(stateError) = state {
                 self.error = stateError
             }
         }
     }
     @Published
-    var configuration: CustomerCenterConfigData?
+    private(set) var configuration: CustomerCenterConfigData?
```
JayShortway, line 46: In `CustomerCenterViewModelTests.testIsLoaded()` the `configuration` field is set. That apparently still works even if it's `private`, judging from the fact that CI passed in this PR. Is that true? If so, why is that haha?

JayShortway (approved):
Thanks a lot for doing this! 🙏

tonidero (approved):
Nice changes! Maybe rename the PR to `Customer center improvements`? So it looks better in the changelog? NABD though

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
JayShortway: Ah no, I guess Customer Center tests are not being run because of the absence of the compiler flag? 🤔

RevenueCatUI/CustomerCenter/URLUtilities.swift
jamesrb1: I think this is the easy way! Does Kotlin have better syntax or functionality?

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
jamesrb1: Ah, you're totally right. I am now missing the mockk framework!

I [made a change](https://github.com/RevenueCat/purchases-ios/pull/4166/commits/83f12ce5e75cf2c18111103448c94c3ef1a38248) to allow this, but am unsure if it's "worth it" - the interface is already internal, so it's just for us in that sense. Thoughts?

JayShortway: I think the root cause of the problem is that we're using static references to `Purchases.shared`. This means we have an implicit dependency that we cannot "invert".

I'd propose to declare a dependency on either
- something like `CustomerCenterConfigFetcher` (we do the same for `CustomerInfoFetcher`, and [I'm adding `CurrentVersionFetcher`](https://github.com/RevenueCat/purchases-ios/pull/4169/files#diff-12a1214b23dae4d7d95973b6abf435214281404b49610c9539094248d1d60a6aR30)), or 
- the existing `CustomerCenterPurchasesType` / a new `PurchasesType` 

That way we can easily provide a fake dependency in tests. What do you think?

tonidero: Yeah, I agree with @JayShortway that using `CustomerCenterPurchasesType` might be best...

jamesrb1: Yes, I agree, but I'd like to finish other work before doing this. Should I merge without changing the access modifiers?

RevenueCatUI/CustomerCenter/CustomerInfo+CurrentEntitlement.swift
```diff
@@ -26,12 +26,11 @@ extension CustomerInfo {
             .values
             .lazy
             .filter { $0.store == .appStore }
-            .sorted { lhs, rhs in
+            .min { lhs, rhs in
```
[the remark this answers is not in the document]
jamesrb1: In Swift you'd have to implement `isBlank` yourself:

`myString.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty`

But now I feel like we're treading into email address validation, which is a minefield, and should probably be handled elsewhere or via some centralized code.

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
jamesrb1: I reverted my changes so that this is as it was before.

JayShortway: Yes, works for me!

RevenueCatUI/CustomerCenter/CustomerInfo+CurrentEntitlement.swift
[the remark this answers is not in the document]
JayShortway: (I suspect the above reply was meant for [this comment](https://github.com/RevenueCat/purchases-ios/pull/4166#discussion_r1714995931) instead.)

Was also mostly concerned with `subject` and `body`. Although it should work for `email` as well. If there's no `@`, it's not an email address. (That's as far as I'd dare to go for client-side email validation.)

But let's just use `isEmpty`. Sufficient, and more performant I'd imagine than first `trimmingCharacters`.

RevenueCatUI/CustomerCenter/URLUtilities.swift
```diff
@@ -27,17 +27,29 @@ enum URLUtilities {
...
     @available(watchOS, unavailable)
     static func createMailURLIfPossible(email: String, subject: String, body: String) -> URL? {
-        let encodedSubject = subject.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) ?? ""
-        let encodedBody = body.addingPercentEncoding(withAllowedCharacters: .urlQueryAllowed) ?? ""
+        guard !email.isEmpty else { return nil }
 
-        let urlString = "mailto:\(email)?subject=\(encodedSubject)&body=\(encodedBody)"
+        var components = URLComponents()
+        components.scheme = "mailto"
+        components.path = email
```
vegaro, line 34: thank you! the other way was horrible lol

vegaro (approved):

joshdholtz (approved):

PR #2811, opened 2023-07-13
`Paywalls`: changed variable handling to use Swift `Regex`

aboedo:
Had to join in the fun!

This PR updates the regex logic for variables to use RegexBuilder.

PR #4169, opened 2024-08-12
[Customer Center] CustomerCenterViewModel checks whether the app is the latest version

JayShortway:
## In this PR
- Logic is added to `CustomerCenterViewModel` allowing it to check if the app is the latest version. 
- A `SemanticVersion` struct is added which encapsulates version parsing and comparison logic.
- The constructors of `CustomerCenterViewModel` are consolidated somewhat with the use of default arguments, for simplification and to be able to more easily add fake values for the current app version in tests. 

## Not in this PR
Any UI changes. Will be in a follow-up.


JayShortway:
@RCGitBot please test

JayShortway:
@RCGitBot please test

JayShortway:
@RCGitBot please test

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -131,7 +152,21 @@ import RevenueCat
             return .purchasesNotFound
         }
     }
+}
 
+fileprivate extension String {
+    /// Takes the first characters of this string, if they conform to Major.Minor.Patch. Returns nil otherwise.
+    /// Note that Minor and Patch are optional. So if this string starts with a single number, that number is returned.
+    func takeVersion() -> String? {
```
jamesrb1, line 160: Maybe this function could be called something like `versionString()`? `takeVersion` doesn't read for me. Could also consider making this a computed property

```swift
var versionString: String? {
...
}

**Review by jamesrb1 — COMMENTED** [review]
Looks good!! A few minor items for your consideration.

(Swift has a RegexBuilder framework, which would be nice to use here, but is iOS 16+ :/)

File: RevenueCatUI/CustomerCenter/Data/SemanticVersion.swift [hunk 2]
```diff
@@ -0,0 +1,69 @@
...
+//  SemanticVersion.swift
+//
+//  Created by JayShortway on 08/08/2024.
+
+import Foundation
+
+struct SemanticVersion: Comparable {
+    let major: Int
+    let minor: Int
+    let patch: Int
```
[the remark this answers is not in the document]
JayShortway: Way better than a nit! This completely avoids a class of bugs at compile time! Shifting left and all that 😄 (https://github.com/RevenueCat/purchases-ios/pull/4169/commits/50d710289b757b1a9c197f890d2aa26061086e32)

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -27,11 +27,16 @@ import RevenueCat
...
+    typealias CurrentVersionFetcher = () -> String?
+
+    private lazy var currentAppVersion: String? = currentVersionFetcher()
 
     @Published
     private(set) var hasSubscriptions: Bool = false
     @Published
     private(set) var subscriptionsAreFromApple: Bool = false
+    @Published
+    private(set) var appIsLatestVersion: Bool = false
```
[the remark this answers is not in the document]
JayShortway: Ah good catch! This is an oversight. I had intended for it to default to `true`. I have moved `defaultAppIsLatestVersion` out of the `didSet` block to the top of the file, because I'd like to define this in one place. I don't know if this is how it's normally done? (In Kotlin I would define it as a `const` in a `companion object`.) See https://github.com/RevenueCat/purchases-ios/pull/4169/commits/54c6ff6b55802eeca4f04269d56ef26433eef8f0

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -43,50 +48,66 @@ import RevenueCat
...
+            guard Purchases.isConfigured else { throw PaywallError.purchasesNotConfigured }
             return try await Purchases.shared.customerInfo()
-        })
-    }
-
-    init(customerCenterActionHandler: CustomerCenterActionHandler?,
-         customerInfoFetcher: @escaping CustomerInfoFetcher) {
+        },
+        currentVersionFetcher: @escaping CurrentVersionFetcher = {
+            Bundle.main.infoDictionary?["CFBundleVersion"] as? String
```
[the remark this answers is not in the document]
JayShortway: These have been confusing me haha. 

[The docs for `CFBundleVersion`](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleversion) say: 
>This key is a machine-readable string composed of one to three period-separated integers, such as 10.14.1. The string can only contain numeric characters (0-9) and periods.

Which appears to be exactly what we need? 

Then again, [the docs for `CFBundleShortVersionString`](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundleshortversionstring) say: 
>This key is a user-visible string for the version of the bundle. The required format is three period-separated integers, such as 10.14.1. The string can only contain numeric characters (0-9) and periods.

So 🤷 

I had initially chosen `CFBundleVersion` because the docs claim it's intended to be machine readable.

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
JayShortway: Right, I can imagine that. `takeVersion()` was supposed to be analogous to Kotlin's use of the terms [_take_ and _drop_](https://kotlinlang.org/docs/collection-parts.html#take-and-drop) which apply to collections and strings.

`String.take(Int)` takes the first `n` characters from a string. We're kinda doing the same, but instead of `n` we take the first characters conforming to a version string.

I renamed it to `versionString()`, but kept it as a function. I felt that a function better signifies that we're doing some work on the string, and `versionString` is not really a property of a string. Weakly held opinion though.

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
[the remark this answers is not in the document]
jamesrb1: It is confusing, and the docs are... wrong. Give it a test! Also note our existing code: https://github.com/RevenueCat/purchases-ios/blob/9eec64d82ec3052d07979f012115f80c024d8b3e/Sources/Misc/SystemInfo.swift#L85-L90

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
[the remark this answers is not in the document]
jamesrb1: I don't think declaring private variables in the global namespace is common practice. Would declaring it as a static variable within the class achieve what you're looking for? (The value will be associated with the type rather than an instance.)

```swift
class CustomerCenterViewModel {
  private static let defaultAppIsLatestVersion = true
```

This can then be accessed as `CustomerCenterViewModel.defaultAppIsLatestVersion` or `Self.defaultAppIsLatestVersion`.

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
jamesrb1: Sounds good!

RevenueCatUI/CustomerCenter/Data/SemanticVersion.swift
```diff
@@ -0,0 +1,73 @@
...
+
+    init(major: UInt, minor: UInt, patch: UInt) {
+        self.major = major
+        self.minor = minor
+        self.patch = patch
+    }
+
+    init(_ version: String) throws {
+        let pattern = #"^(\d+)(?:\.(\d+))?(?:\.(\d+))?$"#
+        guard let regex = try? NSRegularExpression(pattern: pattern),
```
aboedo, line 31: my first thought was that we can use the new [RegexBuilder](https://developer.apple.com/documentation/RegexBuilder), but that's iOS 16+ and we're 15+... So close

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -43,50 +51,63 @@ import RevenueCat
...
+                return
+            }
+
+            do {
+                let currentVersion = try SemanticVersion(currentVersionString)
+                let latestVersion = try SemanticVersion(latestVersionString)
+                self.appIsLatestVersion = currentVersion >= latestVersion
+            } catch {
+                self.appIsLatestVersion = defaultAppIsLatestVersion
+            }
```
aboedo, line 71: You can combine all of these by chaining the conditions like

```swift
didSet {
    guard
        let currentVersionString = currentAppVersion?.versionString(),
        let latestVersionString = configuration?.lastPublishedAppVersion?.versionString(),
        let currentVersion = try? SemanticVersion(currentVersionString),
        let latestVersion = try? SemanticVersion(latestVersionString)
    else {
        self.appIsLatestVersion = defaultAppIsLatestVersion
        return
    }

    self.appIsLatestVersion = currentVersion >= latestVersion
}
```

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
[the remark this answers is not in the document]
aboedo: I also have some recollection of this behaving stupidly different in macOS vs native iOS for some reason, this stuff is beyond confusing

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -128,6 +128,56 @@ class CustomerCenterViewModelTests: TestCase {
...
+            (currentVersion: "10.2", latestVersion: "2", expectedAppIsLatestVersion: true),
+            // We default to true if we fail to parse any of the two versions.
+            (currentVersion: "not-a-number", latestVersion: "not-a-number-either", expectedAppIsLatestVersion: true),
+            (currentVersion: "not-a-number", latestVersion: "1.2.3", expectedAppIsLatestVersion: true),
+            (currentVersion: "1.2.3", latestVersion: "not-a-number", expectedAppIsLatestVersion: true),
+            (currentVersion: nil, latestVersion: nil, expectedAppIsLatestVersion: true),
+            (currentVersion: "1.2.3", latestVersion: nil, expectedAppIsLatestVersion: true),
+            (currentVersion: nil, latestVersion: "1.2.3", expectedAppIsLatestVersion: true)
+        ]
+        for (currentVersion, latestVersion, expectedAppIsLatestVersion) in testCases {
```
aboedo, line 161: Can't wait until we can use parameterized tests from Swift Testing
https://github.com/swiftlang/swift-testing?tab=readme-ov-file#scalable-coverage-and-execution

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -128,6 +128,56 @@ class CustomerCenterViewModelTests: TestCase {
...
+            (currentVersion: "2.0.0.2.3.4", latestVersion: "2.0.0.3.4.5", expectedAppIsLatestVersion: true),
+            (currentVersion: "1.0.0.2.3.4", latestVersion: "2.0.0.3.4.5", expectedAppIsLatestVersion: false),
+            (currentVersion: "1.2", latestVersion: "2", expectedAppIsLatestVersion: false),
+            (currentVersion: "1.2", latestVersion: "1", expectedAppIsLatestVersion: true),
+            (currentVersion: "2", latestVersion: "1", expectedAppIsLatestVersion: true),
+            (currentVersion: "0", latestVersion: "1", expectedAppIsLatestVersion: false),
+            (currentVersion: "10.2", latestVersion: "2", expectedAppIsLatestVersion: true),
+            // We default to true if we fail to parse any of the two versions.
+            (currentVersion: "not-a-number", latestVersion: "not-a-number-either", expectedAppIsLatestVersion: true),
+            (currentVersion: "not-a-number", latestVersion: "1.2.3", expectedAppIsLatestVersion: true),
```
aboedo, line 155: how about `not.a.number`?

(it'd pass, but I mean just in case we accidentally break it in the future so that we look for words split by `.`)

aboedo (approved):
Looking great!

aboedo:
> Swift has a RegexBuilder framework, which would be nice to use here, but is iOS 16+ :/)

lol @jamesrb1 I just read this comment but I had the exact same thought. FWIW a [very early early build of RevenueCatUI did use RegexBuilder](https://github.com/RevenueCat/purchases-ios/pull/2811/files), but we had to drop it when we extended compatibility to iOS 15

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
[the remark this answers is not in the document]
jamesrb1: That rings a bell...

jamesrb1:
> lol @jamesrb1 I just read this comment but I had the exact same thought. FWIW a [very early early build of RevenueCatUI did use RegexBuilder](https://github.com/RevenueCat/purchases-ios/pull/2811/files), but we had to drop it when we extended compatibility to iOS 15

One day... but probably many days from now...

RevenueCatUI/CustomerCenter/Data/SemanticVersion.swift
JayShortway: Yea that seems to be much better!

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
JayShortway: Very nice, thanks! [19b7881](https://github.com/RevenueCat/purchases-ios/pull/4169/commits/19b78817d6675482ef93a1a8b48a0c6874985a53)

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
JayShortway: Good call, added! Also added checks for empty strings. https://github.com/RevenueCat/purchases-ios/pull/4169/commits/14732226881de97bc2a7e108e5d5666cb2d7fa3f

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
JayShortway: Yea somehow I often default to parameterized tests. Easy to cover many bases. 

Swift Testing is Swift 6 only? We probably can't use that right away after it's released, right?

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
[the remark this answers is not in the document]
JayShortway: `private static let` is the thing I was looking for, thanks!

JayShortway:
@RCGitBot please test

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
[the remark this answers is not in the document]
JayShortway: Did not consider docs being wrong. 😅 Thanks for being vigilant! 
 
Would be nice to use the existing `SystemInfo`, but somehow that is not accessible in the scope. Not entirely sure why. I'm using `CFBundleShortVersionString` directly now.

jamesrb1 (approved):
Looks great!

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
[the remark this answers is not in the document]
JayShortway: Ah right, of course. Oh well, this does the trick.

JayShortway:
Changed the label, because I'd imagine we don't want this under "New Features" in the changelog until it's properly released.

PR #4170, opened 2024-08-12
re-enable customer center tests in main

aboedo:
We weren't running tests for Customer Center in PRs, since the `CUSTOMER_CENTER_ENABLED` flag is disabled. 

This could be dangerous, so this PR enables the flag in the absolutely dumbest way I could think of: just replace the instances where the flag is used with `true`, before running tests, i.e.: 
```swift
// before
#if CUSTOMER_CENTER_ENABLED

// after
#if true
```

It's easy to extend to other test suites. It could also have been its own test suite but I felt like it wouldn't have added much, we would have ended up most tests twice for no real gain.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -132,6 +132,10 @@ platform :ios do
 
   desc "Runs all the iOS tests"
   lane :test_ios do |options|
+
+    # todo: remove once Customer Center is released
+    enable_customer_center
+
     generate_snapshots = ENV["CIRCLECI_TESTS_GENERATE_SNAPSHOTS"] == "true"
 
     # For platforms that fail to run StoreKit tests.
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -472,6 +479,28 @@ platform :ios do
...
+    Find.find(project_root) do |path|
+      # process only Swift files
+      next unless File.file?(path) && File.extname(path) == ".swift"
+
+      content = File.read(path)
+
+      updated_content = content.gsub('CUSTOMER_CENTER_ENABLED', 'true')
+
+      File.write(path, updated_content) if content != updated_content
+    end
```
aboedo, line 499: would have been shorter as a find + sed, but this is much easier to read IMO

JayShortway (approved):
😚👌

fastlane/Fastfile
tonidero: Hmm I guess the fastlane commands to run tests are only run through CI normally, so this is probably ok... We could also make a sort of "block" so we restore it once tests finish, but I don't think it's worth it 👍

fastlane/Fastfile
```diff
@@ -472,6 +479,28 @@ platform :ios do
     }
   end
 
+  desc "Enable Customer Center"
+  lane :enable_customer_center do
```
tonidero, line 483: Nitpick but should we make this a private lane?
```suggestion
  private_lane :enable_customer_center do
```

tonidero (approved):
Looks good!

aboedo:
closing in favor of #4171

Issue #4168, opened 2024-08-12
RevenueCatUI missing from RevenueCat.xcframework?


hansemannn:
Is it expected that RevenueCatUI is missing from the RevenueCat.xcframework? We are using RevenueCat for a long time and very extensively, but haven't seen a way to use this one so far. It's also not built via Carthage, so it's kind of a blocker for us, as we need a pre-packaged version of it.

Thanks in advance!

aboedo:
We don't have a pre-built xcframework for it, we'd have to vend it separately. 

Let me see if I can set that up quickly

aboedo:
It wasn't quick 😅 but the XCFramework is now officially a part of releases 🎉  so I'll tentatively mark this one as closed. Feel free to reopen if needed!

aboedo:
It wasn't quick 😅 but the XCFramework is now officially a part of releases 🎉  so I'll tentatively mark this one as closed. Feel free to reopen if needed!

hansemannn:
@aboedo There seemed to be a signing issue with the attached binary, but I'll try a later release to see if it changes something. Just curious to see if that's a known issue atm.

<img width="683" alt="Screenshot 2024-09-06 at 9 37 57 PM" src="https://github.com/user-attachments/assets/6df53c56-d442-4830-b62d-a8351b16e458">

Issue #4146, opened 2024-08-05
XCFramework signing error


XericDesign:
I have migrated from the framework to the xcframework in a macOS project. Everything was working before.

After cleaning the project, during compile I get:

#import <RevenueCat/RevenueCat.h>. "'RevenueCat/RevenueCat.h' file not found"

Building a second time seems to get past this but then I get an OS-generated window saying "RevenueCat.framework" cannot be opened because the developer cannot be verified.

In my Framework search Paths, I have the parent folder where the RevenueCat.xcframework is stored in my project folder.

My project is set to automatic signing, the build phase has a Copy Files section where RevenueCat.xcframework is copied and signed on copy. In the General Build Tab, RevenueCat.xcframework is set to Embed & Sign.

Xcode 15.4

How can I fix this?

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/44209) and notified the team. Thank you for reporting, we're checking this out!

Issue #4056, opened 2024-07-16
Feature Request: Split the RevenueCat XCFramework into Separate UI and Core Frameworks


jesus-mg-ios:
## Description

Dear RevenueCat Team,

I am writing to request an enhancement to the RevenueCat SDK. Specifically, I propose splitting the existing XCFramework into two distinct frameworks:

1. **RevenueCatCore.xcframework**: This framework would include all core functionalities related to in-app purchases and subscription management, excluding any UI components.
2. **RevenueCatUI.xcframework**: This framework would exclusively contain the UI components, providing a clear separation from the core logic.

## Benefits

- **Modularity**: Developers who do not require UI components can integrate only the core functionalities, leading to a more modular and lightweight integration.
- **Flexibility**: Allows developers to implement custom UI solutions while still leveraging RevenueCat's robust backend.
- **Performance**: Reduces the size of the binary for projects that only need core functionalities, potentially improving app performance.

## Implementation Considerations

- Ensure that the split maintains backward compatibility or provide a clear migration path.
- Update documentation to reflect the new structure and guide developers on how to integrate both frameworks if needed.

Thank you for considering this request. I believe this enhancement would greatly benefit the developer community by providing more flexibility and efficiency in integrating RevenueCat into their projects.

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/43393) and notified the team. Thank you for reporting, we're checking this out!

aboedo:
This is definitely something that we intend to do. 

Note that our current xcframework actually doesn't include the `RevenueCatUI.framework`, so we're already _kinda_ doing it 😅 

Meaning that `RevenueCatUI` is currently (and sadly) not included in any way in the .xcframework we include in releases

jesus-mg-ios:
Thanks for your comment @aboedo, so this is weird because if I get the xcframework and get the binary passing through the strings command in Mac I get things like "PaywallColor", or "https://api-paywalls.revenuecat.com" "@"UIColor"16@?0@"UITraitCollection"8" "PaywallViewMode" that for the name should be UI. Could you double check it?

aboedo:
Yeah, I can see that being very confusing. 

Those entities are all a part of the main RevenueCat SDK instead of RevenueCat UI. 

Our thinking at the time was to have the split be: 
- RevenueCat: contains all of the logic, networking, entities, etc to power RevenueCat
[...]

PR #4172, opened 2024-08-12
export RevenueCatUI xcframework

aboedo:
We don't currently export RevenueCatUI as an xcframework, which can be a source of confusion and a dealbreaker for customers who need a pre-built framework. 
This adds RevenueCatUI's xcframework into the `export_xcframeworks` lane, and includes the output in the `github_release` lane. 

Should help address: 
- https://github.com/RevenueCat/purchases-ios/issues/4168
- https://github.com/RevenueCat/purchases-ios/issues/4146
- https://github.com/RevenueCat/purchases-ios/issues/4056


aboedo:
@RCGitBot please test

hansemannn:
@aboedo Thank you for this? Is there a chance to grab a binary from this PR already? It's quite a show stopper for us this week. Many thanks!

aboedo:
@hansemannn apologies, the approach I tested sadly didn't quite work, but I'm actively trying to figure out why the output format isn't quite correct

RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -6457,7 +6457,7 @@
 				PRODUCT_NAME = "$(TARGET_NAME:c99extidentifier)";
 				RUN_DOCUMENTATION_COMPILER = YES;
 				SDKROOT = "";
-				SKIP_INSTALL = YES;
+				SKIP_INSTALL = NO;
```
aboedo, line 6460: not strictly needed to have these set correctly since the fastlane lane does set them, but it's also a good idea to have them so that if we want to archive manually it works correctly. 

`SKIP_INSTALL` should be `NO` for frameworks, and `BUILD_LIBRARY_FOR_DISTRIBUTION` should be `YES`

RevenueCatUI/Modifiers/ViewExtensions.swift
```diff
@@ -40,10 +40,13 @@ extension View {
...
         #if swift(>=5.9)
+        // wrapping with AnyView to type erase is needed because when archiving an xcframework,
+        // the compiler gets confused between the types returned
+        // by the different implementations of self.onChange(of:value).
         if #available(iOS 17.0, macOS 14.0, tvOS 17.0, watchOS 10.0, *) {
-            self.onChange(of: value) { _, newValue in action(newValue) }
+            AnyView(self.onChange(of: value) { _, newValue in action(newValue) })
         } else {
-            self.onChange(of: value) { newValue in action(newValue) }
+            AnyView(self.onChange(of: value) { newValue in action(newValue) })
```
aboedo, line 49: this one drove me _nuts_ and did not have a nice error message, fun times

RevenueCatUI/Views/IconView.swift
```diff
@@ -35,7 +35,7 @@ struct IconView<S: ShapeStyle>: View {
 /// An icon to be displayed by `IconView`.
 enum PaywallIcon: String, CaseIterable {
 
-    case add
+    case plus
```
aboedo, line 38: talk about driving me NUTS. having it named `add` creates a collision with a UIKit name, but the collision happens when Swift auto-generates intermediate files, and has a terrible error message.

fastlane/Fastfile
```diff
@@ -536,22 +536,29 @@ platform :ios do
     }
   end
 
-  desc "Export XCFramework"
-  lane :export_xcframework do |options|
+  lane :export_xcframeworks do |options|
+    create_and_sign_xcframework(scheme: 'RevenueCat', product_name: 'RevenueCat')
+    create_and_sign_xcframework(scheme: 'RevenueCatUIDev', product_name: 'RevenueCatUI')
```
aboedo, line 541: only note here: I had to use the `RevenueCatUIDev` scheme, not `RevenueCatUI` because that one doesn't archive correctly for some reason. That one is tied to SPM. I intend to remove that scheme altogether and rename `RevenueCatUIDev` -> `RevenueCatUI` to prevent confusion, but one thing at a time

fastlane/Fastfile
```diff
@@ -536,22 +536,29 @@ platform :ios do
...
+  end
+
+  private_lane :create_and_sign_xcframework do |options|
```
aboedo, line 544: this part is just extracting common logic

fastlane/Fastfile
aboedo: also the product name determines the actual xcframework name that gets sent out

aboedo:
tested using the xcframework that this generates and it works correctly

RevenueCat.xcodeproj/project.pbxproj
aboedo: if you want to sanity check how these are set, sadly this is a terrible way to try to view it but you can go into xcode and look for the settings under build settings, and check them in the targets, i.e.: 

image
image

aboedo:
welp, the whole archive is pretty huge (130mb) so I can't just upload as a comment, but I'll get this merged and add it to the latest release

hansemannn:
Thanks a lot! Dropbox / iCloud Drive / Google Drive would also fine, but I am trying to wait 😅

fastlane/Fastfile
```diff
@@ -536,22 +536,39 @@ platform :ios do
...
+      'watchOS',
+    ])
+  end
+
+  private_lane :create_and_sign_xcframework do |options|
+    scheme = options[:scheme]
+    product_name = options[:product_name]
+    platforms = options[:platforms]
+
+    output_directory = "build/xcframeworks/#{product_name}"
```
JayShortway, line 566: Might be totally not applicable, but should we upload this as an artifact on CircleCI?

fastlane/README.md
```diff
@@ -242,6 +242,14 @@ Creates a new PR after new snapshot files were generated
...
 Creates a new PR on purchases-ios-snapshots after new snapshot files were generated
 
+### ios enable_customer_center
+
+```sh
+[bundle exec] fastlane ios enable_customer_center
+```
+
+Enable Customer Center
+
```
JayShortway, line 252: Does this lane exist? I can't seem to find it in the Fastfile.

JayShortway (approved):
Approved to unblock you, but would be good for someone with more context to have a final look!

aboedo:
@hansemannn we'll get it into the next release which should come out today or tomorrow, but here are the download links in the meantime:
- [RevenueCat.xcframework](https://drive.google.com/file/d/1H1Vs5jPTweroMds1iOvEi3TnpwHRFH6Y/view?usp=drive_link)
- [RevenueCatUI.xcframework](https://drive.google.com/file/d/1cpQNGvxKI0uWorRuELpaXf9nL1Yzok-S/view?usp=drive_link)

fastlane/README.md
aboedo: oops, no, not sure why it got added, I'll remove

fastlane/Fastfile
aboedo: not a bad idea, might be useful for the release checks lane. I'll do it as a separate thing, though, since we haven't done it before

RevenueCatUI/Views/IconView.swift
joshdholtz, line 38: Pretty sure we need to make this have the value of "add" so it codables correctly with the value we use in the JSON response for paywalls 🤔 `plus` isn't a known icon name 

```suggestion
    case plus = "add"
```

joshdholtz (changes requested):

fastlane/Fastfile
JayShortway: Sounds good!

RevenueCatUI/Views/IconView.swift
aboedo: ooff right I thought it was just referencing a local file, didn't realize it needed to match a backend name. That's a bit of a problem, given the UIKit naming conflict. 
Maybe we can solve it by decoupling the local property name from the key from the backend in the codable side, let me see

aboedo: lol just saw that you had actually already left a suggestion for the change 🤦 I'll try to repro with a paywall before and after to make sure all things work

jamesrb1: Be careful when you make the change that the ["+" image](https://github.com/RevenueCat/purchases-ios/blob/main/RevenueCatUI/Resources/icons.xcassets/add.imageset/add.svg) still shows up in the live paywall if you select it as the list tick. I ran into this issue during an aborted project reorg attempts.

aboedo: yeah, the image won't show up, gonna need to do something extra here, will post here again when it's solved

aboedo: Solved now. I don't _love_ the solution, but the local assets are named to match the backend exactly, and the backend sends names that might conflict with UIKit, so the solution decouples the two sides. 

Simulator Screenshot - iPhone 15 - 2024-08-15 at 13 18 20

aboedo: Thanks for flagging this!

jamesrb1: Good solution!

jamesrb1 (approved):

joshdholtz (approved):
:shipit:

aboedo:
Merged! @hansemannn thanks for bringing this up and for the patience with it!

hansemannn:
Thanks a lot for the fast solution here - I followed it every day :) We are unblocked already by the built and Paywalls rock! Thanks to the whole team!

fastlane/Fastfile
aboedo: actually it's a brilliant idea, not sure why I said "not bad" 😅 I'm tackling it here: https://github.com/RevenueCat/purchases-ios/pull/4189

JayShortway: Haha, it's the greatest midwestern compliment, apparently. Or was that "I don't hate it"? 😄 Anyway, awesome!

PR #4183, opened 2024-08-14
replace `color(from colorInformation:)` global with extension

aboedo:
Small refactor that replaces the color(from colorInformation:) global it with internal extension. 

The idea is to keep the global namespace cleaner. Also it feels a bit neater to be able to find it through Color.


jamesrb1 (commented):
💯

PR #4184, opened 2024-08-14
Speed up carthage installation tests

aboedo:
Speeds up carthage installation tests by reducing the amount of work that they do. 

It almost halves the execution time:

| Before | After |
| :-: | :-: |
| Screenshot 2024-08-14 at 12 05 07 PM | Screenshot 2024-08-14 at 12 30 54 PM |


jamesrb1 (approved):
Excellent!!

PR #4190, opened 2024-08-15
[Customer Center] Migrate to List style

codykerns:
- Migrates our custom Customer Center UI to use Lists instead of stacks of buttons
- Implements a reusable navigation stack component with backwards compatibility
- Pulls out SubscriptionDetailsView into a separate file

**ManageSubscriptionsView**

| New | Old |
|-----|-----|
|  |  |

**FeedbackSurveyView**

| New | Old |
|-----|-----|
|  |  |


RevenueCatUI/CustomerCenter/Views/CompatibilityNavigationStack.swift
```diff
@@ -0,0 +1,46 @@
...
+        if #available(iOS 16.0, *) {
+            NavigationStack {
+                content
+            }
+        } else {
+            NavigationView {
+                content
+            }
+        }
+    }
```
vegaro, line 40: Oh nice cleanup

RevenueCatUI/CustomerCenter/Views/FeedbackSurveyView.swift
```diff
@@ -83,33 +81,20 @@ struct FeedbackSurveyButtonsView: View {
...
+        ForEach(options, id: \.id) { option in
+            AsyncButton(action: {
+                await self.onOptionSelected(option)
+            }, label: {
+                if self.loadingState == option.id {
+                    TintedProgressView()
+                } else {
+                    Text(option.title)
+                }
+            })
```
vegaro, line 93: I know this was there before but I got this same suggestion in one of my PRs

```suggestion
            AsyncButton {
                await self.onOptionSelected(option)
            } label: {
                if self.loadingState == option.id {
                    TintedProgressView()
                } else {
                    Text(option.title)
                }
            }
```

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -84,27 +78,28 @@ struct ManageSubscriptionsView: View {
...
             }
 
-            ScrollView {
-                VStack {
-                    if self.viewModel.isLoaded {
-                        SubtitleTextView(subtitle: self.viewModel.screen.subtitle)
+            if self.viewModel.isLoaded {
+                List {
+                    if let subtitle = self.viewModel.screen.subtitle {
+                        SubtitleTextView(subtitle: subtitle)
```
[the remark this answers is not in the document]
vegaro: this is making me wonder if we need the subtitle?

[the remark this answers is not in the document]
vegaro: What about as header of the buttons section? I think it is fine if the system handles the style since we are setting the color in `SubtitleTextView` to be `textColor`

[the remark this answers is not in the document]
codykerns: @vegaro Yeah, I think that would be great! I don't want to lose the option to add an additional, optional message, and trying to stuff it above the subscription details feels weird. Will update!

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -84,27 +78,30 @@ struct ManageSubscriptionsView: View {
...
+                    Section {
                         ManageSubscriptionsButtonsView(viewModel: self.viewModel,
                                                        loadingPath: self.$viewModel.loadingPath)
-                    } else {
-                        ProgressView()
-                            .progressViewStyle(CircularProgressViewStyle())
+                    } header: {
+                        if let subtitle = self.viewModel.screen.subtitle {
+                            Text(subtitle)
+                                .textCase(nil)
```
codykerns, line 99: by default, lists apply an uppercase style for section headers - this disables the uppercase style to make it use the exact subtitle text that's configured (which could be made uppercase by the developer)

vegaro: we would need to apply the text color here though as we were doing in `SubtitleTextView` right?

codykerns: resolving this since we talked about it separately 👍

PR #4191, opened 2024-08-16
Create lane to enable customer center

vegaro:
This lane will do the cherry pick command that enables the customer center, without the need to check the commit hash

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -998,6 +998,32 @@ platform :ios do
     end
   end
 
+  desc "Enable customer center development by cherry-picking a specific commit"
+  lane :enable_customer_center do |options|
+    commit_hash = options[:commit_hash] || "517ad0bf5"
+
+    UI.message("Preparing to enable customer center development...")
+    UI.message("Checking git status...")
+    
+    ensure_git_status_clean
+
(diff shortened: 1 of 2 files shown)


fastlane/Fastfile
```diff
@@ -998,6 +998,33 @@ platform :ios do
...
   end
 
+  desc "Enable customer center development by cherry-picking a specific commit"
+  lane :enable_customer_center do |options|
+    commit_hash = options[:commit_hash] || "517ad0bf5"
+
+    UI.message("Preparing to enable customer center development...")
+    UI.message("Checking git status...")
+    
+    ensure_git_status_clean
```
JayShortway, line 1008: How required is this? Would be nice if it's possible to go without it, unless we can't easily abort/undo in the event of conflicts.

JayShortway (approved):
I love a good DX improvement! Just a single question, not a blocker.

aboedo:
great idea

fastlane/Fastfile
vegaro: I added this in case there's a conflict with the cherry pick, which would make it fail. Just to prevent those failures. But I can see how this might be more annoying than help so I will remove it

JayShortway: Alright! If it turns out to be annoying, we can always add this back.

PR #4188, opened 2024-08-15
Fix failing `all-tests` and retry more flaky tests

joshdholtz:
## Motivation

- Compile errors in Xcode 15 with iOS 16
- Flaky tests

### iOS 16 fixes

The following things aren't available:

- `@available(visionOS)` 
- `.topBarTrailing` (need to use `.navigationBarTrailing`)
- Fix issues with `Locale(identifier: "")` now returning `en` but ONLY on iOS 16 in Xcode 15
  - iOS 17 returns the expected `nil` in Xcode 15
- There was a test using `#if swift` that was now being run by iOS 16 tests that needed an additional `AvailabilityChecks.iOS17APIAvailableOrSkipTest()`

### Flaky tests fixes

Using new `scan_with_flaky_test_retries` action from in `fastlane-plugin-revenuecat_internal` 

- Replaces the `retry_failed_tests` lane that did a similar thing
- Removed a bunch of duplicated steps in the `.circle/config.yml` that needed to run the `retry_failed_tests`
- `scan_with_flaky_test_retries` is now used in:
  - `test_ios`
  - `test_tvos`
  - `test_watchos`
  - `backend_integration_tests`

### Carthage tests timing out

The carthage installation tests timed out pretty regularly. They will now use the large resource class (where everything else uses the medium)

PR #4193, opened 2024-08-16
[Customer Center] Shows a warning when the app is not the latest version

JayShortway:
## In this PR
- A new `AppUpdateWarningView` is added, which is heavily inspired by `WrongPlatformView`. 
- `CustomerCenterView` shows the `AppUpdateWarningView` when required. 

Simulator Screenshot - iPhone 15 Pro

## Not in this PR
- Actually opening the App Store when the user presses "Update". Will do that in a follow-up. Have asked for input on Slack.


Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -73,6 +73,10 @@ public struct CustomerCenterConfigData {
             case defaultBody = "default_body"
             case defaultSubject = "default_subject"
             case dismiss = "dismiss"
+            case updateWarningTitle = "update_warning_title"
+            case updateWarningDescription = "update_warning_description"
+            case updateWarningUpdate = "update_warning_update"
+            case updateWarningIgnore = "update_warning_ignore"
```
JayShortway, line 79: Not sure if we want to make these strings specific to the app update warning screen?

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -118,6 +122,17 @@ public struct CustomerCenterConfigData {
                     return "Support Request"
                 case .dismiss:
                     return "Dismiss"
+                case .updateWarningTitle:
+                    return "Update the app to fix your problem"
+                case .updateWarningDescription:
+                    return """
+                    A new update is available. Downloading the latest version of the app can help you solve your \
+                    problem.
+                    """
```
JayShortway, line 131: This is almost exactly Photoroom's copy.

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
```diff
@@ -0,0 +1,116 @@
...
+                }
+                .buttonStyle(ProminentButtonStyle())
+                .padding(.bottom)
+
+                Button(localization.commonLocalizedString(for: .updateWarningIgnore)) {
+                    onContinueAnywayClick()
+                }
+            }
+            .padding(.horizontal)
+            .applyIf(textColor != nil, apply: { $0.foregroundColor(textColor) })
```
JayShortway, line 65: This line makes the entire screen invisible for me, on PaywallsTester. Not entirely sure what's happening. I included it, because the same line exists in `WrongPlatformView`. (The same problem occurs on `WrongPlatformView` for me.)

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
```diff
@@ -0,0 +1,116 @@
...
+        }
+        .toolbar {
+            ToolbarItem(placement: .navigationBarTrailing) {
```
jamesrb1, line 68: New compatibility view [incoming](https://github.com/RevenueCat/purchases-ios/pull/4188/files#diff-8ac72e331918e7a7c60cdf485ea66bbd4e8324a2320e99afd6be38552efd55d8), but it's not merged yet.

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
jamesrb1: Odd, I am not seeing that.

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -91,8 +92,15 @@ private extension CustomerCenterView {
...
-                                        customerCenterActionHandler: viewModel.customerCenterActionHandler)
+                if ignoreAppUpdateWarning || viewModel.appIsLatestVersion {
+                    ManageSubscriptionsView(screen: screen,
+                                            customerCenterActionHandler: viewModel.customerCenterActionHandler)
+                } else {
+                    AppUpdateWarningView(
+                        onUpdateAppClick: { viewModel.onAppUpdateClick() },
+                        onContinueAnywayClick: { ignoreAppUpdateWarning = true }
+                    )
+                }
```
jamesrb1, line 103: When you click continue the transition is instant which feels a bit abrupt. This can be animated by putting the entire if statement in a ZStack and applying an animation modifier:

```swift
ZStack {
// if { ... } else { ... }
}.animation(.easeInOut, value: ignoreAppUpdateWarning)
```

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -118,6 +122,17 @@ public struct CustomerCenterConfigData {
                     return "Support Request"
                 case .dismiss:
                     return "Dismiss"
+                case .updateWarningTitle:
+                    return "Update the app to fix your problem"
```
jamesrb1, line 126: Does this need a period? I would also consider referring it to as "the" problem rather than "their" problem, which could sound a bit like they were somehow responsible for the problem.
```suggestion
                    return "Update the app to fix the problem."
```

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -118,6 +122,17 @@ public struct CustomerCenterConfigData {
...
+                case .updateWarningDescription:
+                    return """
+                    A new update is available. Downloading the latest version of the app can help you solve your \
+                    problem.
```
jamesrb1, line 130: I'd consider this slight change in wording.
```suggestion
                    A new update is available. Downloading the latest version of the app may help solve the \
                    problem.
```

jamesrb1 (approved):
Looks good! A few minor items for you to consider.

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
jamesrb1: Actually without this, I think `spm-revenuecat-ui-ios-15` will fail.

Sources/CustomerCenter/CustomerCenterConfigData.swift
JayShortway: Ah I didn't add the period because it's used as a title, but I see we have a period in `WrongPlatformView` too. 
  
Regarding "the" vs "your": I agree! I wasn't sure how much we wanted it to be like Photoroom's screen. (See screenshot in [SDK-3522](https://linear.app/revenuecat/issue/SDK-3522/[customercenter][sdk]-display-an-alert-when-user-tries-to-contact).)

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
JayShortway: Ah nice, using that cause it's merged now!

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
JayShortway: Much better, thanks for the suggestion! https://github.com/RevenueCat/purchases-ios/pull/4193/commits/eac6e3e00c9e465b29f2b4533011f1ba390ef834

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
```diff
@@ -0,0 +1,116 @@
...
+    @Environment(\.colorScheme)
+    private var colorScheme
+
+    @ViewBuilder
+    var content: some View {
+        ZStack {
+            if let background = Color.from(colorInformation: appearance.backgroundColor, for: colorScheme) {
+                background.edgesIgnoringSafeArea(.all)
+            }
+            let textColor = Color.from(colorInformation: appearance.textColor, for: colorScheme)
```
JayShortway, line 45: Note to self: make this not customizable as discussed in the sync.

JayShortway:
I updated this PR after the redesign of the existing "no content" screens in #4204. Screenshot in the OP is updated. Would be great to get a final look 🙏   
  
Commit: https://github.com/RevenueCat/purchases-ios/pull/4193/commits/84cc70b7f922c2306f54967a4704e909037a9fc3

JayShortway:
Note to self, update the dismiss button: https://github.com/RevenueCat/purchases-ios/pull/4208

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
```diff
@@ -0,0 +1,126 @@
...
+    private var colorScheme
+
+    @ViewBuilder
+    var content: some View {
+        ZStack {
+            List {
+                Section {
+                    CompatibilityContentUnavailableView(
+                        localization.commonLocalizedString(for: .updateWarningTitle),
+                        systemImage: "arrowshape.up.circle.fill",
```
jamesrb1, line 43: This image is iOS 17+ only. You can check in Xcode with this dialog (command-shift-L):

image

jamesrb1 (changes requested):

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
JayShortway: Ah TIL, thanks! Updated here: https://github.com/RevenueCat/purchases-ios/pull/4193/commits/5e3bebeae7cf6dc0a77752dd7b8bc8d3f25f4361.

PR #4196, opened 2024-08-16
Set Paywalls Tester deployment target to iOS 15

jamesrb1:
Updates the target of Paywalls Tester to iOS 15 so that we can test RevenueCatUI easily.

### DIFF
--- a/Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/CompatibilityContentUnavailableView.swift
+++ b/Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/CompatibilityContentUnavailableView.swift
@@ -1,0 +1,85 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  CompatibilityContentUnavailableView.swift
+//
+//
+//  Created by Cody Kerns on 8/6/24.
+//
+
+import Foundation
+import SwiftUI
+
+#if os(iOS)
+
+/// A SwiftUI view for displaying a message about unavailable content
+@available(iOS 15.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+struct CompatibilityContentUnavailableView: View {
+    let title: String
+    let systemImage: String
+    let description: Text?
+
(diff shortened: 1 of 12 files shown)


jamesrb1:
@RCGitBot please test

jamesrb1:
@RCGitBot please test

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/Login/LoginWall.swift
```diff
@@ -21,8 +21,10 @@ struct LoginWall<ContentView: View>: View {
         case .unknown:
             ProgressView()
                 .displayError(self.$error)
-                .task { @MainActor in
-                    await reload()
+                .onAppear {
+                    Task {
+                        await reload()
+                    }
```
jamesrb1, line 27: For some reason the .task caused an error (and subsequent error dialog) to pop up when run on iOS 15 (worked as expected on 16 and 17).

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/CompatibilityContentUnavailableView.swift
jamesrb1, line 1: Taken from Customer Center, which I assume will remain internal so unavailable to access from Paywalls Tester.

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/Login/LoginWall.swift
joshdholtz: Can you leave a comment in there for this? 😅  This is something that one of us (probably me) would change in the future and not realize why we did this 😛

joshdholtz (approved):
This looks good to me! Just one small nit about moving something to a comment in the code but other than that let's :shipit:

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/Login/LoginWall.swift
jamesrb1: Good idea, will do!

JayShortway (approved):
Nice! I ran into the iOS 15 incompatibility a while back. Thanks for taking care of it!

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/Login/LoginWall.swift
jamesrb1: Done!

jamesrb1:
@RCGitBot please test

PR #4197, opened 2024-08-17
Compatibility content unavailable improvements

jamesrb1:
A couple small updates to this view (which I was looking to use in Paywalls Tester).

- @State variables are not needed because they never change
- Initializer now matches `ContentUnavailableView`, which one day will make moving to the system one easier.


jamesrb1:
@RCGitBot please test

codykerns (approved):
👏

tonidero (approved):

PR #4199, opened 2024-08-19
[CustomerCenter] Open App Store when the user wants to update their app

JayShortway:
`AppUpdateWarningView` opens the App Store when the user clicks Update.


RevenueCatUI/CustomerCenter/Views/OpenAppStoreViewController.swift
```diff
@@ -0,0 +1,66 @@
...
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    /// Opens the App Store for the provided productId in-app using SKStoreProductViewController. Falls back to
+    /// redirecting to the App Store if this fails.
+    func openAppStore(productId: UInt) {
```
JayShortway, line 42: The [documentation](https://developer.apple.com/documentation/storekit/skstoreproductparameteritunesitemidentifier) says that the `productId`, or _iTunes item identifier_, is an instance of `NSNumber`. I opted to take a `UInt` as input, so we only accept positive integers, as that is the only type of ID I've seen in the wild. Please let me know if this is not true. 

The conformity to `NSNumber` leads me to assume that IDs cannot begin with a `0`, as I think `NSNumber` would deem that `0` insignificant and strip it. Please also correct me if I'm wrong.

RevenueCatUI/CustomerCenter/Views/OpenAppStoreViewController.swift
```diff
@@ -0,0 +1,66 @@
...
+        let storeProductViewController = SKStoreProductViewController()
+        storeProductViewController.delegate = self
+        let parameters = [SKStoreProductParameterITunesItemIdentifier: productId]
+
+        storeProductViewController.loadProduct(withParameters: parameters) { _, error in
+            guard error == nil,
+                  let windowScene = UIApplication.shared.connectedScenes.first as? UIWindowScene,
+                  let rootViewController = windowScene.windows.first?.rootViewController else {
+
+                let appStoreUrl = URL(string: "https://itunes.apple.com/app/id\(productId)")!
```
JayShortway, line 52: If `productId` can only ever exist of positive integers, I believe it is always safe to construct a URL from it. Unless I'm overlooking something?

Sources/Networking/Responses/CustomerCenterConfigResponse.swift
```diff
@@ -21,6 +21,7 @@ struct CustomerCenterConfigResponse {
 
     let customerCenter: CustomerCenter
     let lastPublishedAppVersion: String?
+    let productId: UInt
```
JayShortway, line 24: Field name TBD.

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -146,11 +146,6 @@ import RevenueCat
             return .purchasesNotFound
         }
     }
-
-    func onAppUpdateClick() {
-        // swiftlint:disable:next todo
-        // TODO: implement opening the App Store
-    }
```
JayShortway, line 153: Since this is done using a ViewController, I decided that it's best for the view layer to handle this.

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -94,12 +94,12 @@ private extension CustomerCenterView {
...
             if viewModel.subscriptionsAreFromApple,
                let screen = configuration.screens[.management] {
-                if ignoreAppUpdateWarning || viewModel.appIsLatestVersion {
+                if ignoreAppUpdateWarning || viewModel.appIsLatestVersion || configuration.productId == nil {
                     ManageSubscriptionsView(screen: screen,
                                             customerCenterActionHandler: viewModel.customerCenterActionHandler)
                 } else {
                     AppUpdateWarningView(
-                        onUpdateAppClick: { viewModel.onAppUpdateClick() },
+                        productId: configuration.productId!,
```
jamesrb1, line 102: This is currently safe given the code above, but changes above could render it un-safe. I would recommend turning the else into an else if let statement which can be used to safely unwrap `productId`, and then have an else block that displays maybe content unavailable and logs an error.

jamesrb1 (commented):
Looks nice! Just one minor item to consider.

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
JayShortway: Yes that's a good point. I was a bit conflicted about this as I didn't want to add an entire view just because the compiler isn't able to figure out that `productId` is actually nil here. 

I ended up inverting the `if` statement, so the `productId` is safely unwrapped in the initial condition. This avoids a useless view and keeps the compiler happy. See https://github.com/RevenueCat/purchases-ios/pull/4199/commits/a55224efbe047f7c197fd58a64da74cf8adef4d2.

tonidero (approved):
LGTM!

vegaro (approved):

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
JayShortway: Yea [smart casts](https://kotlinlang.org/docs/typecasts.html#smart-casts) are amazing, and they [got even better](https://kotlinlang.org/docs/whatsnew20.html#smart-cast-improvements) in Kotlin 2 (which we're not using anywhere yet though).

PR #4201, opened 2024-08-20
Adds Paywalls Components Compiler Feature Flag

jamesrb1:
Adds a feature flag that allows us to merge the paywalls component to main but keep it out of any compiled packages. Avoids the pains of a long-running feature branch.


Sources/Paywalls/PaywallData.swift
joshdholtz, line 1: In one of the next PRs, we will remove these `#if PAYWALL_COMPONENTS` in this file so we don't want to touch it 😇 

It all paywall components stuff will go into that new `PaywallComponentsData` struct. Nothing to do here but just wanted to not that for documentation purposes 😁

joshdholtz (approved):
:shipit: 💯

PR #4203, opened 2024-08-21
Fixes paywalls on iOS 15 when the close button is set to be displayed.

jamesrb1:


joshdholtz:
@RCGitBot please test

joshdholtz (commented):
Thanks for fixing this! I started a running of all tests on this to make sure that the RCUI tests still pass for iOS 15 or if we need to regenerate them 🤷‍♂️

joshdholtz (approved):
This looks good to me! There are still probably more improvements we could do but this is real good start since what out there is pretty broken 😬 

:shipit:

jamesrb1:
Converting back to draft - this needs a bit more work to correct a knock on effect.

PR #4204, opened 2024-08-21
[Customer Center] Clean up colors in WrongPlatformView and NoSubscriptionsView

vegaro:
Changed the `WrongPlatformView` and `NoSubscriptionsView` to use lists, so they don't use the prominent buttons and look more native

| WrongPlatformView | NoSubscriptionsView |
|--------|--------|
| Simulator Screenshot - iPhone 15 Pro - 2024-08-21 at 14 59 48 | Simulator Screenshot - iPhone 15 Pro - 2024-08-21 at 14 51 17
 |

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
@@ -47,34 +47,29 @@ struct NoSubscriptionsView: View {
     }
 
     var body: some View {
-        let background = Color.from(colorInformation: appearance.backgroundColor, for: colorScheme)
-        let textColor = Color.from(colorInformation: appearance.textColor, for: colorScheme)
-
         let fallbackDescription = "We can try checking your Apple account for any previous purchases"
 
-        ZStack {
-            if background != nil {
-                background.edgesIgnoringSafeArea(.all)
-            }
-            VStack {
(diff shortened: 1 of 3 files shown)


RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
```diff
@@ -47,34 +47,29 @@ struct NoSubscriptionsView: View {
...
                     systemImage: "exclamationmark.triangle.fill",
                     description:
                         Text(self.configuration.screens[.noActive]?.subtitle ?? fallbackDescription)
-
                 )
+            }
 
-                Spacer()
-
+            Section {
```
tonidero, line 62: Not for this PR! But we still need to make the SDK respond to changes in the `noActive` screen paths right? I believe right now that's mostly ignored....

Separately, I'm also wondering if the `WrongPlatformView` should also be a separate "screen"... but we can work on that later if needed as well.

tonidero (approved):
LGTM!

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
vegaro: oh right, forgot about that 😮‍💨

PR #4216, opened 2024-08-23
Pin swift-docc-plugin to 1.3.0

jamesrb1:
Pins swift-docc-plugin for the time being as 1.4.0 and 1.4.1 are failing to compile (https://revenuecat.slack.com/archives/C03HPN9UE80/p1724438146087079).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -15,6 +15,8 @@ var dependencies: [Package.Dependency] = [
     .package(url: "git@github.com:pointfreeco/swift-snapshot-testing.git", .upToNextMinor(from: "1.12.0"))
 ]
 if shouldIncludeDocCPlugin {
+    // Versions 1.4.0 and 1.4.1 are failing to compile, so we are pinning it to 1.3.0 for now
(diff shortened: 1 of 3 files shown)


jamesrb1:
@RCGitBot please test

Package.swift
```diff
@@ -15,7 +15,7 @@ var dependencies: [Package.Dependency] = [
     .package(url: "git@github.com:pointfreeco/swift-snapshot-testing.git", .upToNextMinor(from: "1.12.0"))
 ]
 if shouldIncludeDocCPlugin {
-    dependencies.append(.package(url: "https://github.com/apple/swift-docc-plugin", from: "1.0.0"))
+    dependencies.append(.package(url: "https://github.com/apple/swift-docc-plugin", .exact("1.3.0")))
```
aboedo, line 18: let's add a comment explaining why we're pinning to this specific version so future us don't hate present us

aboedo (approved):

Package.swift
jamesrb1: Done!

PR #4220, opened 2024-08-26
[CustomerCenter] Move sheet and restore alert creation to `ManageSubscriptionsView`

vegaro:
The sheet was being created multiple times and we realized it was due to it being created from the buttons instead of from the view `ManageSubscriptionsView` itself

We moved the creation of the sheet and the restore alert to the `ManageSubscriptionsView` to fix it


jamesrb1 (approved):
👍

JayShortway (approved):
Nice find!

tonidero (approved):

PR #4229, opened 2024-08-29
Set a maximum duration for iOS 15 tests

vegaro:
We are not sure why but iOS 15 tests get sometimes stuck re-running forever (see https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/21864/workflows/7af8697c-aad3-4e1f-92ae-f55186a72f27/jobs/254961). It looks like rerunning fixes it.

Until we find out what's up, we are going to set a max time for this job. CircleCI recommends setting a sleep and calling an API to cancel the job when that finishes https://support.circleci.com/hc/en-us/articles/14114124583195-How-to-set-a-custom-maximum-job-duration

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -94,18 +94,18 @@ aliases:
 commands:
   set-maximum-duration:
     parameters:
-      time:
+      seconds:
         type: integer
         default: 600
     steps:
       - run:
-          name: Set maximum job duration to << parameters.time >> minutes
+          name: Set maximum job duration to << parameters.seconds >> seconds
           background: true
           command: |
-            sleep << parameters.time >>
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -92,6 +92,20 @@ aliases:
         only: main
 
 commands:
+  set-maximum-duration:
+    parameters:
+      seconds:
+        type: integer
+        default: 600
```
joshdholtz, line 99: There are some jobs that will take like 20 to 25 minutes I think? Should we make this 30 minutes? 🤔

joshdholtz (approved):
This looks good to me! 👍

PR #4230, opened 2024-08-30
Paywalls Components Viewmodels + partial localization support

jamesrb1:
This PR makes all paywall component views rely on intermediate view models for better separation of concerns. The Component is decoded from the server response, the ComponentViewModel organizes the component's data into a view-friendly format, including translating "raw" types into SwiftUI types when needed, and the ComponentView renders the component.

[ComponentView] --owns---> [ComponentViewModel]  --owns---> [Component].

There has been file re-organization, including grouping all of our model type extensions and the code that transforms them into SwiftUI equivalents into their own files. Colors are now decoded without relying on template paywall code.


RevenueCatUI/Templates/Components/Spacer/SpacerComponentViewModel.swift
```diff
@@ -0,0 +1,33 @@
...
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+public class SpacerComponentViewModel: ObservableObject {
+
+    let locale: Locale
+    let component: PaywallComponent.SpacerComponent
+
+    init(locale: Locale, component: PaywallComponent.SpacerComponent) {
+        self.locale = locale
+        self.component = component
+    }
```
joshdholtz, line 31: `locale` can probably be removed from these too!

RevenueCatUI/Templates/Components/Stack/StackComponentView.swift
```diff
@@ -19,11 +19,20 @@ import SwiftUI
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 struct StackComponentView: View {
 
-    let component: PaywallComponent.StackComponent
+    let viewModel: StackComponentViewModel
+
+    var locale: Locale {
+        return viewModel.locale
+    }
```
joshdholtz, line 26: `locale` can probably be removed from here too!

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
```diff
@@ -12,20 +12,36 @@ import SwiftUI
...
+            // TOOD: This logic is so wrong
+            let localizations = paywallComponentsData.componentsLocalizations
+            let localization = localizations[locale.identifier] ?? localizations.values.first ?? [String:String]()
+
+            // Step 3 - Validate all variables are supported in localization
+
+            // Step 3.5 - Validate all packages needed exist (????)
+
+            // Step 4 - Make the view models
+            return component.toViewModel(offering: offering, locale: locale, localization: localization)
```
joshdholtz: Note to any other reviewers: This is where we will do a validation and these comments are essentially TODO so we can keep that in mind when coming back to this area in near future 😊

RevenueCatUI/Templates/Components/Text/TextComponentView.swift
```diff
@@ -0,0 +1,41 @@
...
+#if PAYWALL_COMPONENTS
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+struct TextComponentView: View {
+
+    let viewModel: TextComponentViewModel
+
+    var locale: Locale {
+        return viewModel.locale
+    }
```
joshdholtz, line 27: Can probably remove `locale` from here too?

RevenueCatUI/Templates/Components/Text/TextComponentViewModel.swift
```diff
@@ -0,0 +1,77 @@
...
+    var text: String {
+        // TODO: Replace variables like "{{ }}"
+        // TODO: Add logs?
+        if let textLid = component.textLid {
+            if let localizedText = localization[textLid] {
+                return localizedText
+            }
+            else {
+                return component.text.value.first?.value as? String ?? "missing localized text for \(textLid)"
+            }
```
joshdholtz: Note to any other reviewers: This is here to back support existing templates we were building and testing with.

All localizations _will_ come from the `localization` dict eventually 👍

Sources/Paywalls/Components/PaywallComponentTypes.swift
joshdholtz, line 1: Nit: Can we rename this file to `PaywallComponentPropertyTypes.swift`? 🤔

Sources/Paywalls/Components/PaywallTextComopnent.swift
```diff
@@ -25,6 +26,7 @@ public extension PaywallComponent {
 
         public init(
             text: DisplayString,
```
joshdholtz, line 28: Note: We will need to get rid of this _eventually_ when all of our test paywalls are using the localized strings in their new place/format

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/APIKeyDashboardList.swift
```diff
@@ -142,7 +142,8 @@ struct APIKeyDashboardList: View {
         .sheet(item: self.$presentedPaywall) { paywall in
             #if PAYWALL_COMPONENTS
             if let componentData = paywall.offering.paywallComponentsData {
-                TemplateComponentsView(paywallComponentsData: componentData)
+                // TODO: Get locale
```
joshdholtz, line 145: I think this todo can be removed! I think I added this 😅

Getting locale will happen internally in this component

joshdholtz (changes requested):
A little bit of cleanup but otherwise this is looking good! This will be much nicer to build on 😊

RevenueCatUI/Data/Strings.swift
```diff
@@ -149,6 +151,9 @@ extension Strings: CustomStringConvertible {
         case .no_price_round_formatter_failed:
             return "Could not round price because formatter failed to round price."
 
+        case .invalid_color_string(let colorString):
+            return "Invalid hex colour string: \(colorString)"
```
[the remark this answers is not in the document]
jamesrb1: Oops! I usually catch myself with that one.

RevenueCatUI/Templates/Components/Image/ImageComponentViewModel.swift
```diff
@@ -0,0 +1,51 @@
...
+import RevenueCat
+import SwiftUI
+
+#if PAYWALL_COMPONENTS
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+public class ImageComponentViewModel: ObservableObject {
+
+    let locale: Locale
+    @Published private(set) var component: PaywallComponent.ImageComponent
```
joshdholtz, line 26: I thought I asked this the first time but I can't remember 😅 

What's the benefit of this being an `ObservableObject` with a `@Published` property? 🤔 If we process are everything upfront, this view model _shouldn't_ change at all, right?

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
```diff
@@ -10,22 +10,41 @@ import SwiftUI
...
+
+            // STEP 1 - Get list of preferred locales (and default)
+
+            // STEP 2 - Get localization for one of preferred locales in order
+            let chosenLocale = Locale.current
+            // TOOD: This logic is so wrong
+            let localizations = paywallComponentsData.componentsLocalizations
+            let localization = localizations[chosenLocale.identifier] ??
+                                  localizations.values.first ??
+                                  [String: String]()
```
joshdholtz, line 32: I think this is probably addressed in the next PR but we should/will be erroring out and not defaulting to an empty dictionary if no locales found 😇

RevenueCatUI/Templates/Components/Text/TextComponentViewModel.swift
```diff
@@ -0,0 +1,78 @@
...
+        // TODO: Add logs?
+        if let textLid = component.textLid {
+            if let localizedText = localization[textLid] {
+                return localizedText
+            } else {
+                return component.text.value.first?.value as? String ?? "missing localized text for \(textLid)"
+            }
+        } else {
+            return component.text.value.first?.value as? String ?? "missing localized text"
+        }
```
joshdholtz, line 46: For anybody else reviewing, the way we get and validate locales are changing in the next PR

joshdholtz (approved):
This looks goooooood to me! Super clean and easy to follow 💪

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
jamesrb1: In the next PR it logs an error and returns an empty dictionary. It will then error out when if it tries to localize a string on any component. I guess we can discuss in the next PR!

RevenueCatUI/Templates/Components/Image/ImageComponentViewModel.swift
jamesrb1: I don't think we discussed this - I wrote it like that because I wanted to create a live paywall (for my loom and for customers in the future) where it would fetch the model every second or so, and have the view update automatically when the model changed. Somewhere this wasn't working and rather than spend more time debugging it I moved no, but left this as-is for the future. If we removed it now it would have no effect, so I guess maybe we should until if/when functionality is available.

joshdholtz: > I wrote it like that because I wanted to create a live paywall

Oh neat! That sounds kind of a neat way to debug/test some designs with the dashboard 😁 

> If we removed it now it would have no effect, so I guess maybe we should until if/when functionality is available.

I think it _might_ be best to remove for now if easy enough! Just to make the state/flow a little easier with these being immutable

jamesrb1: Done! https://github.com/RevenueCat/purchases-ios/pull/4230/commits/56b52cb276cc627d43a913aecfa013fc58ee8c2c

jamesrb1:
@RCGitBot please test

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
```diff
@@ -4,28 +4,48 @@
...
         self.paywallComponentsData = paywallComponentsData
+
+        let components = paywallComponentsData.componentsConfig.components
+        self.componentViewModels = components.map { component in
+
+            // STEP 1 - Get list of preferred locales (and default)
+
+            // STEP 2 - Get localization for one of preferred locales in order
+            let chosenLocale = Locale.current
+            // TOOD: This logic is so wrong
```
aboedo, line 29: note: 
This comment is very ambiguous. It doesn't explain what's actually wrong with it, why, or how to fix it. 
And now it's a part of our `main` code, so anyone cloning the repo sees this. If I'm checking out this repo and find this, I run. 

In my first job, on my first day, I saw a comment that I'll never forget. It was something like:
```
// TODO: THIS IS VERY VERY WRONG. FIX BEFORE TRIP. 
```

And I checked the history and the comment had been there for > 7 years. Nobody knew what was actually wrong, so you can't really remove it. What trip? I'll never know. The author had left the company years before I joined.

tonidero:
Changing labels from `pr:feat` to `pr:refactor` to make it appear correctly in the changelog.

PR #4234, opened 2024-08-30
[Customer Center] Build `WrongPlatformView` from JSON

vegaro:
Build `WrongPlatformView` using Strings from the JSON

Cleaned up the view as well so the logic is simpler and looks cleaner

Simulator Screenshot - iPhone 15 Pro - 2024-08-30 at 18 43 15

### DIFF
--- a/Sources/CustomerCenter/CustomerCenterConfigData.swift
+++ b/Sources/CustomerCenter/CustomerCenterConfigData.swift
@@ -143,11 +143,11 @@ public struct CustomerCenterConfigData {
                 case .pleaseContactSupportToManage:
                     return "Please contact support to manage your subscription."
                 case .appleSubscriptionManage:
-                    return "You can manage your subscription via the App Store app on an Apple device."
+                    return "You can manage your subscription by using the App Store app on an Apple device."
                 case .googleSubscriptionManage:
-                    return "You can manage your subscription via the Play Store app on an Android device"
+                    return "You can manage your subscription by using the Play Store app on an Android device"
                 case .amazonSubscriptionManage:
-                    return "You can manage your subscription via the Amazon Appstore app."
+                    return "You can manage your subscription in the Amazon Appstore app on an Amazon device."
                 }
             }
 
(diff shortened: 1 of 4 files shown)


Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -133,6 +138,16 @@ public struct CustomerCenterConfigData {
                     return "Update"
                 case .updateWarningIgnore:
                     return "Continue"
+                case .platformMismatch:
+                    return "Platform mismatch"
```
vegaro, line 142: Happy to talk about this. I went through different ideas, even not showing a title at all. Let me know what you think

image

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -133,6 +138,16 @@ public struct CustomerCenterConfigData {
...
+                case .pleaseContactSupportToManage:
+                    return "Please contact support to manage your subscription."
```
jamesrb1, line 144: Maybe not for this PR, but would be great if there were a link/button that would take them directly to contacting support from this page.

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -133,6 +138,16 @@ public struct CustomerCenterConfigData {
...
+                case .appleSubscriptionManage:
+                    return "You can manage your subscription via the App Store app on an Apple device."
```
jamesrb1, line 146: I would probably replace "via" with "by using" here, but just an opinion. But actually, can you with the app store app? I think this needs to be managed in Settings or via your AppleID somehow.

Sources/CustomerCenter/CustomerCenterConfigData.swift
tonidero: That button is probably something you can configure from the JSON :)

tonidero (approved):
LGTM!

Sources/CustomerCenter/CustomerCenterConfigData.swift
vegaro: Thanks for the feedback, I will update the text. You can actually manage the subscriptions in the app store by clicking the profile pic

jamesrb1: > You can actually manage the subscriptions in the app store by clicking the profile pic

TIL!

jamesrb1 (approved):

PR #4236, opened 2024-09-03
[Diagnostics] Add `backend_error_code` property

tonidero:
### Checklist

### Description
This adds the `backend_error_code` property to diagnostics `http_request_performed` events.

### DIFF
--- a/Sources/Networking/HTTPClient/HTTPClient.swift
+++ b/Sources/Networking/HTTPClient/HTTPClient.swift
@@ -590,17 +590,21 @@ private extension HTTPClient {
                                                                        responseTime: responseTime,
                                                                        wasSuccessful: true,
                                                                        responseCode: httpStatusCode,
+                                                                       backendErrorCode: nil,
                                                                        resultOrigin: response.origin,
                                                                        verificationResult: verificationResult)
                 case let .failure(error):
                     var responseCode = -1
-                    if case let .errorResponse(_, code, _) = error {
+                    var backendErrorCode: Int?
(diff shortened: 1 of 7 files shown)


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -590,17 +590,21 @@ private extension HTTPClient {
                                                                        responseTime: responseTime,
                                                                        wasSuccessful: true,
                                                                        responseCode: httpStatusCode,
+                                                                       backendErrorCode: nil,
                                                                        resultOrigin: response.origin,
                                                                        verificationResult: verificationResult)
                 case let .failure(error):
                     var responseCode = -1
-                    if case let .errorResponse(_, code, _) = error {
+                    var backendErrorCode: Int?
```
jamesrb1, line 598: I like using `nil` to represent the absence of a value (as this new code does), but we now seem to be mixing conventions now for "empty": `-1` for `responseCode` and `nil` for `backendErrorCode`. Not necessarily recommending a change here, but just surfacing this in case it hadn't been considered.

jamesrb1 (approved):
Looks good!

Sources/Networking/HTTPClient/HTTPClient.swift
tonidero: That's a good point... And yeah, I also prefer to send nil for these cases... I think it's something that can be changed separately to avoid changing the data we send though. Good callout!

PR #4237, opened 2024-09-03
Paywall Components Localized Strings

jamesrb1:
- Adds localized string support to paywall components
  -This is primarily done in `TemplateComponentsView`'s static function `chooseLocalization`.

- Validates that promised localized strings are actually available.
  -This is done as component view models are created (see `TextComponentViewModel`), by throwing if an expected key is not present:
`self.text = try localizedStrings.string(key: component.textLid)`
  -This pattern should be followed for every property that represents a localized string id, which ensures a non-nil value for every value.

https://github.com/user-attachments/assets/b4aaf221-e88f-42e0-a9c2-31683f771d1b


RevenueCatUI/Data/Strings.swift
```diff
@@ -52,6 +52,9 @@ enum Strings {
     case no_price_round_formatter_failed
 
     case invalid_color_string(String)
+    case paywall_view_model_construction_failed(Error)
+    case paywall_contains_no_localization_data
+    case paywall_could_not_find_localization_for(String)
```
joshdholtz, line 57: Nit: remove `for`
```suggestion
    case paywall_could_not_find_localization(String)
```

Sources/Paywalls/Components/PaywallComponentLocalization.swift
```diff
@@ -0,0 +1,42 @@
...
+//  Created by James Borthwick on 2024-09-03.
+// swiftlint:disable missing_docs
+
+import Foundation
+
+#if PAYWALL_COMPONENTS
+
+public typealias LocaleID = String
+public typealias LocalizationDictionary = [String: String]
+public typealias LocalizationKey = String
```
joshdholtz, line 21: Worried about having non-namespaced typealiases here 🤔 Anybody who does `import RevenueCatUI` will get these typealises and could be general enough where they clash 😬 

Not sure if we have an opinions on these (even though I do love typeliases 😇) 

Would like some thoughts from @aboedo and other @RevenueCat/coresdk people on this 🙏

Sources/Paywalls/Components/PaywallComponentPropertyTypes.swift
```diff
@@ -11,11 +11,14 @@
...
 //  Created by James Borthwick on 2024-08-29.
 
-// swiftlint:disable missing_docs todo
+// swiftlint:disable missing_docs
 import Foundation
 
 #if PAYWALL_COMPONENTS
 
+public typealias TierId = String
+public typealias ColorHex = String
```
joshdholtz, line 20: Same here with public typealiases... worried about people getting these since not namespaced

joshdholtz (changes requested):
This is looking good! Very easy to follow 😁 

I think my biggest concern is addressing `public typealias` 🤷‍♂️ Not sure if my concern is valid or not but its a lot of things that could get accidentally used since not namespaced inside of some paywall struct or enum

Sources/Paywalls/Components/PaywallComponentLocalization.swift
jamesrb1: These are defined in the RevenueCat module. If there is a collision, they can be resolved, but it would be a bit annoying. A few options:

1. Use weirder names
2. Prefix `RC` onto the type names, Obj-C style
3. Put these in an empty enum (with an unusual name) purely for name spacing purposes.

I'd opt for option three above, so for example:
```swift
public enum PaywallComponentType {
    public typealias LocaleID = String
}
```
Would then be accessed as
```swift
let foo: PaywallComponentType.TestSymbol = "Hello"
```
or to make it less verbose, you could choose to type alias the longer type:
```swift
// assumes no collision
public typealias TestSymbol = PaywallComponentType.TestSymbol
let foo: TestSymbol = "Hello"

// if there were a collision
public typealias RCTestSymbol = PaywallComponentType.TestSymbol
let foo: RCTestSymbol = "Hello"
```

joshdholtz: @jamesrb1 I like putting the typealias in an enum if we are able to do that!

jamesrb1: I just edited my response with another idea, to re-used a related public struct. Thoughts? @joshdholtz

joshdholtz: Putting in `PaywallComponentsData` sounds good to me 👍

aboedo: > Put these in the PaywallComponentsData struct (which is public).

FWIW I think this is the way that SK2 has been handling it

aboedo: cc @MarkVillacampa thoughts?

jamesrb1: I [moved them](3bc2c7d0bcfde2938389d652874c42da538eb31f) into the `PaywallComponent` struct, which seemed like the most "generic" place to put them for namespacing purposes.

_Aside: Swift needs an equivalent to C#'s "using" statement._

Sources/Paywalls/Components/PaywallComponentPropertyTypes.swift
jamesrb1: Discussion in [above comment](https://github.com/RevenueCat/purchases-ios/pull/4237#discussion_r1744112744), resolving this one.

Tests/TestingApps/PaywallsTester/PaywallsTester/Extensions/ArrayExtensions.swift
```diff
@@ -16,3 +17,42 @@ public extension Sequence {
     }
 
 }
+
+#if PAYWALL_COMPONENTS
+public extension Array where Element == PaywallComponent {
```
tonidero, line 22: NItpick but should this be internal? (doesn't matter much since it's for the test app)

Sources/Paywalls/Components/PaywallComponentPropertyTypes.swift
```diff
@@ -10,13 +10,15 @@
...
 //  Created by James Borthwick on 2024-08-29.
-
 // swiftlint:disable missing_docs
 
 import Foundation
 
 #if PAYWALL_COMPONENTS
 
+public typealias TierId = String
+public typealias ColorHex = String
```
tonidero, line 20: I see these were added as an extension to `PaywallComponent` above, so maybe we can remove these?

tonidero (commented):
LGTM! Just some small nitpicks but will leave to others to approve.

Tests/TestingApps/PaywallsTester/PaywallsTester/Extensions/ArrayExtensions.swift
jamesrb1: Done, and moved these to their own file (because dictionary isn't an array, and because these extensions are so specific, not generic array extensions).

Sources/Paywalls/Components/PaywallComponentPropertyTypes.swift
jamesrb1: Good catch! I also deleted `TierId` entirely because it's not used at this point.

jamesrb1:
@RCGitBot please test

jamesrb1:
@RCGitBot please test

RevenueCatUI/Templates/Components/TemplateComponentsView+extensions.swift
```diff
@@ -0,0 +1,115 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  FallbackPaywall.swift
```
joshdholtz, line 10: Nit: The comment doesn't match the file name 😇

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
```diff
@@ -51,14 +50,34 @@ public struct TemplateComponentsView: View {
...
+        guard !componentsData.componentsLocalizations.isEmpty else {
+            Logger.error(Strings.paywall_contains_no_localization_data)
+            return (Locale.current, PaywallComponent.LocalizationDictionary())
+        }
+
+        // STEP 1: Get available paywall locales
+        let paywallLocales = componentsData.componentsLocalizations.keys.map { Locale(identifier: $0) }
 
-    return displayString.value.values.first!
+        let fallbackLocale = Locale(identifier: componentsData.defaultLocale)
```
joshdholtz, line 65: Nit: Can we call this `defaultLocale` 😇

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
```diff
@@ -51,14 +50,34 @@ public struct TemplateComponentsView: View {
...
+
+        // STEP 2: choose best locale based on device's list of preferred locales.
+        let chosenLocale = Self.preferredLocale(from: paywallLocales) ?? fallbackLocale
+
+        // STEP 3: Get localization for one of preferred locales in order
+        if let localizedStrings = componentsData.componentsLocalizations[chosenLocale.identifier] {
+            return (chosenLocale, localizedStrings)
+        } else if let localizedStrings = componentsData.componentsLocalizations[fallbackLocale.identifier] {
+            Logger.error(Strings.paywall_could_not_find_localization("\(chosenLocale)"))
```
joshdholtz, line 74: Need to log the correct locale (and I made the suggestion `defaultLocale` like my suggestion above ☝️ 

```suggestion
            Logger.error(Strings.paywall_could_not_find_localization("\(defaultLocale)"))
```

RevenueCatUI/Templates/Components/Text/TextComponentViewModel.swift
```diff
@@ -21,26 +21,14 @@ import SwiftUI
...
-        if let textLid = component.textLid {
-            if let localizedText = localization[textLid] {
-                return localizedText
-            } else {
-                return component.text.value.first?.value as? String ?? "missing localized text for \(textLid)"
-            }
-        } else {
-            return component.text.value.first?.value as? String ?? "missing localized text"
-        }
+        self.text = try localizedStrings.string(key: component.textLid)
```
joshdholtz, line 31: :chefs_kiss:

Sources/Paywalls/Components/PaywallComponentBase.swift
```diff
@@ -37,6 +31,13 @@ public enum PaywallComponent: PaywallComponentBase {
 
 }
 
+public extension PaywallComponent {
+    typealias LocaleID = String
+    typealias LocalizationDictionary = [String: String]
+    typealias LocalizationKey = String
+    typealias ColorHex = String
+}
```
joshdholtz, line 39: Thank you! 🫶

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
```diff
@@ -89,7 +89,7 @@ struct SamplePaywallsList: View {
             )
         case .customerCenter:
             #if CUSTOMER_CENTER_ENABLED
-            CustomerCenterView(customerCenterActionHandler: self.handleCustomerCenterAction)
```
joshdholtz, line 92: Is this intended? 🤔

joshdholtz (changes requested):
A few _small_ things but real close! This localization handling looks 🔥

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
jamesrb1: We can - I chose fallback because I felt it is a more precise description of its function here, where "default" is what is chosen if you make no choice, and "fallback" is what you get when your choice is unavailable. However "default" is the user-facing term, so it makes sense to keep things consistent.

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
jamesrb1: Ooof, no!

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
joshdholtz: We also used "fallback" as the "first language" if no default was selected in Paywalls V1 so it kind of has a different meaning in my mind too 🫠 

We don't need that same behavior here but I'm already broken haha

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
jamesrb1: I think it's correct as written? It will only enter this block if it failed to find `chosenLocale.identifier` in the first if let statement. At this point, it has found the localization for `defaultLocale`.

This would be easier to read if it used successive if statements rather than using elses, so I can make that change.

joshdholtz: Oh yeaaahhh hahaha... I read it wrong 😅  Nevermind!

Sources/Paywalls/Components/PaywallComponentBase.swift
jamesrb1:
```
RevenueCat.Borthwick.James.reply("You're Welcome")
RevenueCat.Borthwick.James.reply("Thanks for the suggestion")
```

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
jamesrb1: 🫠 
Changed to default!

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
jamesrb1: Done!

joshdholtz (approved):
Looks good to me! 👍

tonidero:
Changing labels from `pr:feat` to `pr:refactor` for now, so it's assigned correctly in the changelog.

PR #4245, opened 2024-09-05
[SK2] send unsynced attributes when syncing purchases

MarkVillacampa:
### Checklist

### Motivation
When calling syncPurchases with StoreKit 2, we aren't including any unsynced subscriber attributes.

### Description
This PR modifies the StoreKit 2 syncPurchases network request to include unsynced subscriber attributes, the storefront, and a nil `presentedOfferingContext`.


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1142,6 +1142,9 @@ private extension PurchasesOrchestrator {
                     self.backend.post(receipt: .empty,
                                       productData: nil,
                                       transactionData: .init(appUserID: currentAppUserID,
+                                                             presentedOfferingContext: nil,
+                                                             unsyncedAttributes: unsyncedAttributes,
+                                                             storefront: await Storefront.currentStorefront,
```
MarkVillacampa, line 1147: we dont really need the storefront here I think since there are no purchases

fire-at-will, line 1147: Do you have any concerns about the additional `await` here adding extra time to this function? I feel like I recall `awaiting` on something SK2-related like this when doing network requests added latency in certain scenarios, but I can't recall exactly what it was off the top of my head right now.

aboedo: we're already awaiting sk2 twice in this codepath already though, right? first verified transaction + app transaction jws?

aboedo: nitpick: maybe it's worth splitting this into its own line, like we do in the else code path, that explicitly does `createProductRequestData`

aboedo (approved):
looks good, but I wouldn't oppose skipping the storefront if it doesn't add value either

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
fire-at-will: @aboedo Done! We're now creating `transactionData` on it's own line 😎

PR #4247, opened 2024-09-06
[Diagnostics] Add `apple_products_request` event

tonidero:
### Description
This adds a new diagnostics event `apple_products_request` that tracks the results when fetching products from Apple.

### DIFF
--- a/Sources/Purchasing/ProductsManager.swift
+++ b/Sources/Purchasing/ProductsManager.swift
@@ -144,13 +144,13 @@ private extension ProductsManager {
             let errorMessage = (error?.userInfo[NSUnderlyingErrorKey] as? Error)?.localizedDescription
                 ?? error?.localizedDescription
             let errorCode = error?.errorCode
-            Async.call(with: {}, asyncMethod: {
+            Task(priority: .background) {
                 await diagnosticsTracker.trackProductsRequest(wasSuccessful: error == nil,
                                                               storeKitVersion: storeKitVersion,
                                                               errorMessage: errorMessage,
                                                               errorCode: errorCode,
                                                               responseTime: responseTime)
-            })
+            }
         }
     }
 
--- a/Sources/Purchasing/ProductsManager.swift
+++ b/Sources/Purchasing/ProductsManager.swift
@@ -87,12 +87,12 @@ class ProductsManager: NSObject, ProductsManagerType {
         if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *),
            self.systemInfo.storeKitVersion.isStoreKit2EnabledAndAvailable {
             self.sk2Products(withIdentifiers: identifiers) { result in
-                self.trackProductRequestIfNeeded(startTime, storeKitVersion: .storeKit2, error: result.error)
+                self.trackProductsRequestIfNeeded(startTime, storeKitVersion: .storeKit2, error: result.error)
                 completion(result.map { Set($0.map(StoreProduct.from(product:))) })
             }
         } else {
             self.sk1Products(withIdentifiers: identifiers) { result in
-                self.trackProductRequestIfNeeded(startTime, storeKitVersion: .storeKit1, error: result.error)
+                self.trackProductsRequestIfNeeded(startTime, storeKitVersion: .storeKit1, error: result.error)
                 completion(result.map { Set($0.map(StoreProduct.from(product:))) })
             }
         }
@@ -135,12 +135,12 @@ private extension ProductsManager {
         return self.productsFetcherSK1.products(withIdentifiers: identifiers, completion: completion)
     }
 
-    func trackProductRequestIfNeeded(_ startTime: Date,
+    func trackProductsRequestIfNeeded(_ startTime: Date,
                                      storeKitVersion: StoreKitVersion,
                                      error: PurchasesError?) {
         if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *),
            let diagnosticsTracker = self.diagnosticsTracker {
-            let requestTime = self.dateProvider.now().timeIntervalSince(startTime)
+            let responseTime = self.dateProvider.now().timeIntervalSince(startTime)
             let errorMessage = (error?.userInfo[NSUnderlyingErrorKey] as? Error)?.localizedDescription
                 ?? error?.localizedDescription
             let errorCode = error?.errorCode
@@ -149,7 +149,7 @@ private extension ProductsManager {
                                                               storeKitVersion: storeKitVersion,
                                                               errorMessage: errorMessage,
                                                               errorCode: errorCode,
-                                                              responseTime: requestTime)
+                                                              responseTime: responseTime)
             })
         }
     }
--- a/Sources/Purchasing/ProductsManager.swift
+++ b/Sources/Purchasing/ProductsManager.swift
@@ -144,14 +144,7 @@ private extension ProductsManager {
             let errorMessage = (error?.userInfo[NSUnderlyingErrorKey] as? Error)?.localizedDescription
                 ?? error?.localizedDescription
             let errorCode = error?.errorCode
-            let storeKitErrorDescription: String?
-            if let skError = error?.userInfo[NSUnderlyingErrorKey] as? SKError {
-                storeKitErrorDescription = skError.code.trackingDescription
-            } else if let storeKitError = error?.userInfo[NSUnderlyingErrorKey] as? StoreKitError {
-                storeKitErrorDescription = storeKitError.trackingDescription
-            } else {
-                storeKitErrorDescription = nil
-            }
+            let storeKitErrorDescription = StoreKitErrorUtils.extractStoreKitErrorDescription(from: error)
             Task(priority: .background) {
                 await diagnosticsTracker.trackProductsRequest(wasSuccessful: error == nil,
                                                               storeKitVersion: storeKitVersion,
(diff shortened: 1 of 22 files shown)


Tests/StoreKitUnitTests/ProductsManagerTests.swift
```diff
@@ -126,3 +128,75 @@ class ProductsManagerTests: StoreKitConfigTestCase {
...
+        let params = self.mockDiagnosticsTracker.trackedProductsRequestParams.first
+        expect(params?.wasSuccessful) == true
+        expect(params?.storeKitVersion) == .storeKit2
+        expect(params?.errorMessage) == nil
+        expect(params?.errorCode) == nil
+    }
+
+    #if swift(>=5.9)
+    @available(iOS 17.0, tvOS 17.0, macOS 14.0, watchOS 10.0, *)
+    func testFetchProductsWithIdentifiersSK2ErrorTracksCorrectly() async throws {
```
tonidero, line 183: I only added an error test case for SK2 since `setSimulatedError` didn't seem to work well for SK1 (the test got stuck). Do lmk if I missed something!

tonidero:
This needs to wait until the backend has been deployed but should be ready for review. Please lmk if you think we should track other info in this event. We can process it in the backend, if we need to discretize it.

Sources/Purchasing/ProductsManager.swift
```diff
@@ -126,6 +135,25 @@ private extension ProductsManager {
         return self.productsFetcherSK1.products(withIdentifiers: identifiers, completion: completion)
     }
 
+    func trackProductRequestIfNeeded(_ startTime: Date,
```
vegaro, line 138: I think this should be:

```suggestion
    func trackProductsRequestIfNeeded(_ startTime: Date,
```

vegaro (approved):

Sources/Purchasing/ProductsManager.swift
```diff
@@ -126,6 +135,25 @@ private extension ProductsManager {
...
+            let requestTime = self.dateProvider.now().timeIntervalSince(startTime)
+            let errorMessage = (error?.userInfo[NSUnderlyingErrorKey] as? Error)?.localizedDescription
+                ?? error?.localizedDescription
+            let errorCode = error?.errorCode
+            Async.call(with: {}, asyncMethod: {
+                await diagnosticsTracker.trackProductsRequest(wasSuccessful: error == nil,
+                                                              storeKitVersion: storeKitVersion,
+                                                              errorMessage: errorMessage,
+                                                              errorCode: errorCode,
+                                                              responseTime: requestTime)
```
vegaro, line 152: Imo, either `responseTime` or `requestTime` should be renamed. What you're passing is actually the time the request took to come back right?

tonidero: Yeah that's true... but I was trying to reuse the same property we use for backend http request response times... I think I will rename the property to `responseTime` if that's ok, since I think it's ok to keep it as the same property (and so, reuse the validators in the backend)

Sources/Purchasing/ProductsManager.swift
tonidero: Addressed in https://github.com/RevenueCat/purchases-ios/pull/4247/commits/ea41bf72dc1b301b66a7cbdf02e1be8c01e3ccc8

Sources/Purchasing/ProductsManager.swift
tonidero: Addressed in https://github.com/RevenueCat/purchases-ios/pull/4247/commits/ea41bf72dc1b301b66a7cbdf02e1be8c01e3ccc8

Tests/StoreKitUnitTests/ProductsManagerTests.swift
```diff
@@ -126,3 +128,75 @@ class ProductsManagerTests: StoreKitConfigTestCase {
...
+        _ = waitUntilValue(timeout: Self.requestDispatchTimeout) { completed in
+            manager.products(withIdentifiers: Set([identifier]), completion: completed)
+        }
+
+        expect(self.mockDiagnosticsTracker.trackedProductsRequestParams.count) == 1
+        let params = self.mockDiagnosticsTracker.trackedProductsRequestParams.first
+        expect(params?.wasSuccessful) == true
+        expect(params?.storeKitVersion) == .storeKit2
+        expect(params?.errorMessage) == nil
+        expect(params?.errorCode) == nil
```
vegaro, line 178: I think these should be `to(beNil))`

tonidero: Good catch! Changed in https://github.com/RevenueCat/purchases-ios/pull/4247/commits/e81cd1c79681b2229f9d6abc2caad44329368982

Tests/StoreKitUnitTests/ProductsManagerTests.swift
```diff
@@ -126,3 +128,75 @@ class ProductsManagerTests: StoreKitConfigTestCase {
...
+        try AvailabilityChecks.iOS17APIAvailableOrSkipTest()
+
+        try await self.testSession.setSimulatedError(.generic(.unknown), forAPI: .loadProducts)
+        let manager = self.createManager(storeKitVersion: .storeKit2,
+                                         diagnosticsTracker: self.mockDiagnosticsTracker)
+
+        let identifier = "com.revenuecat.monthly_4.99.1_week_intro"
+        _ = try? await manager.products(withIdentifiers: Set([identifier]))
+
+        expect(self.mockDiagnosticsTracker.trackedProductsRequestParams.count) == 1
```
vegaro, line 193: one question... because I am facing the same issue. Since you're posting in a task inside the `ProductsManager` https://github.com/RevenueCat/purchases-ios/pull/4247/files#diff-b080ecaed1fab9b6af95a6a40653aa15720509e51c3faa1aa71263f99a4ff1cbR147 this check is going to be flaky, because you can't guarantee the task has executed, right?

Sources/Purchasing/ProductsManager.swift
```diff
@@ -126,6 +135,25 @@ private extension ProductsManager {
         return self.productsFetcherSK1.products(withIdentifiers: identifiers, completion: completion)
     }
 
+    func trackProductsRequestIfNeeded(_ startTime: Date,
+                                     storeKitVersion: StoreKitVersion,
+                                     error: PurchasesError?) {
+        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *),
+           let diagnosticsTracker = self.diagnosticsTracker {
+            let responseTime = self.dateProvider.now().timeIntervalSince(startTime)
+            let errorMessage = (error?.userInfo[NSUnderlyingErrorKey] as? Error)?.localizedDescription
```
vegaro, line 144: Does this error contain more information if casted to `SKError` or `StoreKitError`? I have been doing this in #4253 

```
if let skError = error?.userInfo[NSUnderlyingErrorKey] as? SKError {
    storeKitErrorDescription = skError.code.trackingDescription
} else if let storeKitError = error?.userInfo[NSUnderlyingErrorKey] as? StoreKitError {
    storeKitErrorDescription = storeKitError.trackingDescription
} else {
    storeKitErrorDescription = nil
}
```

Sources/Purchasing/ProductsManager.swift
```diff
@@ -126,6 +135,25 @@ private extension ProductsManager {
...
+                ?? error?.localizedDescription
+            let errorCode = error?.errorCode
+            Task(priority: .background) {
+                await diagnosticsTracker.trackProductsRequest(wasSuccessful: error == nil,
```
aboedo, line 148: wonder whether the diagnostics tracker should be the one to know that its stuff is all background, instead of every consumer needing to know that

aboedo (approved):

Sources/Purchasing/ProductsManager.swift
tonidero: That's a valid point... And yeah, you might be right. I will change this one and the other ones we're sending, so it's encapsulated. Thanks!

Sources/Purchasing/ProductsManager.swift
tonidero: Hmm that's a good question... Will try that out, thanks!

Sources/Purchasing/ProductsManager.swift
tonidero: I did this in https://github.com/RevenueCat/purchases-ios/pull/4270

tonidero:
Note that I refactored my PR over Césars #4253, to be able to include a new property that tracks the storekit error type if available. See commit: https://github.com/RevenueCat/purchases-ios/pull/4247/commits/818c18a8cd53049cd4df2a29538a349a44012e14.

Since that PR still needs the backend part and this property also requires a backend change, making this draft until then

Sources/Purchasing/ProductsManager.swift
vegaro: Do we want to add the storekit error `trackingDescription` now that we have that available?

tonidero: Isn't that handled by the [extractStoreKitErrorDescription](https://github.com/RevenueCat/purchases-ios/pull/4247/files#diff-b080ecaed1fab9b6af95a6a40653aa15720509e51c3faa1aa71263f99a4ff1cbR147) a few lines below?

vegaro: oh yes, my bad. I looked for `trackingDescription` 😮‍💨

PR #4251, opened 2024-09-08
[POC] Components have selected version of themselves

jamesrb1:
POC

### DIFF
--- a/RevenueCatUI/Templates/Components/Text/TextComponentView.swift
+++ b/RevenueCatUI/Templates/Components/Text/TextComponentView.swift
@@ -20,16 +20,18 @@ import SwiftUI
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 struct TextComponentView: View {
 
+    @Environment(\.selectionState) var selectionState
+
     let viewModel: TextComponentViewModel
 
     var body: some View {
         Text(viewModel.text)
-            .font(viewModel.textStyle)
-            .fontWeight(viewModel.fontWeight)
-            .multilineTextAlignment(viewModel.horizontalAlignment)
-            .foregroundStyle(viewModel.color)
-            .padding(viewModel.padding)
-            .background(viewModel.backgroundColor)
+            .font(viewModel.textStyle(for: selectionState))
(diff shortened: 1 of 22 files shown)


RevenueCatUI/Templates/Components/Text/TextComponentView.swift
```diff
@@ -20,16 +20,18 @@ import SwiftUI
...
-            .multilineTextAlignment(viewModel.horizontalAlignment)
-            .foregroundStyle(viewModel.color)
-            .padding(viewModel.padding)
-            .background(viewModel.backgroundColor)
+            .font(viewModel.textStyle(for: selectionState))
+            .fontWeight(viewModel.fontWeight(for: selectionState))
+            .multilineTextAlignment(viewModel.horizontalAlignment(for: selectionState))
+            .foregroundStyle(viewModel.color(for: selectionState))
+            .padding(viewModel.padding(for: selectionState))
+            .background(viewModel.backgroundColor(for: selectionState))
```
aboedo, line 34: feels like a lot of repetition here, should the viewModel itself know its own selection state so that it can figure this out on its own?

jamesrb1: Good comment, the repetition definitely looks clunky, but it's a tradeoff. Adding selection state to the view model would make it stateful for selection status, which here is more a concern of the view layer than of the view data modelling layer. SwiftUI makes it natural to handle presentation states like selection within views directly, without involving the VM. For example, the state of if a sheet is presented is typically held only by the view. The VM is then left to focus on prepping business data for presentation (or multi-view states, other more complicated state management).

In this particular case, where the state in question is passed down via the environment, I think is definitely one where the state should stay out of the VM if possible, because the state is owned elsewhere, and syncing an environment variable to a viewmodel is a relatively manual process, and feels like it's going against SwiftUI's declarative nature.

This is all definitely different from traditional MVVM where you want the view to be purely a reflection of state held in the VM. But SwiftUI seems to be designed this way, where these transient presentation states are managed by the views directly.

I've thought about other ways to get rid of this repetition, but haven't come up with one that I think is worth the tradeoffs yet. Will keep thinking.

Thoughts on the above?

PR #4253, opened 2024-09-09
[Diagnostics] Add `apple_purchase_attempt ` event

vegaro:
Create a new `applePurchaseAttempt`

Still as draft as it is missing tests and backend


Sources/Diagnostics/DiagnosticsTracker.swift
```diff
@@ -91,6 +98,24 @@ final class DiagnosticsTracker: DiagnosticsTrackerType {
...
+                              errorCode: Int?,
+                              storeKitErrorDescription: String?) async {
+        await track(
+            DiagnosticsEvent(eventType: .applePurchaseAttempt,
+                             properties: [
+                                .successfulKey: AnyEncodable(wasSuccessful),
+                                .storeKitVersion: AnyEncodable("store_kit_\(storeKitVersion.debugDescription)"),
+                                .errorMessageKey: AnyEncodable(errorMessage),
+                                .errorCodeKey: AnyEncodable(errorCode),
+                                .skErrorDescriptionKey: AnyEncodable(storeKitErrorDescription)
```
tonidero, line 113: ❤️ I think this could be useful! We probably should add it to the event I added in #4253 if we can.

tonidero (commented):
Looks great!

Sources/Diagnostics/DiagnosticsTracker.swift
vegaro: yes! I think it will be super useful. I thought about using an int code, but SK2 errors don't have an int representation. So I decided to go with the name of the event. It will make it easier to analyse as well.

Sources/Diagnostics/DiagnosticsTracker.swift
```diff
@@ -31,6 +31,13 @@ protocol DiagnosticsTrackerType {
                                    resultOrigin: HTTPResponseOrigin?,
                                    verificationResult: VerificationResult) async
 
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    func trackPurchaseRequest(wasSuccessful: Bool,
+                              storeKitVersion: StoreKitVersion,
+                              errorMessage: String?,
+                              errorCode: Int?,
+                              storeKitErrorDescription: String?) async
```
tonidero, line 39: Should we also track the request time? It could be useful to have a graph of how long these requests take.

vegaro: The issue is that the request time would include the time the purchase dialog is displayed, making it useless. I don't think we have any other way to track this? We could track post receipts only?

tonidero: Yeah that's true ignore me. We should already been tracking the post receipt calls in the `http_request_performed` event so not needed!

Sources/Error Handling/SKError+Extensions.swift
```diff
@@ -91,6 +91,57 @@ extension SKError: PurchasesErrorConvertible {
...
+        case .overlayTimeout:
+            return "overlay_timeout"
+        case .ineligibleForOffer:
+            return "ineligible_for_offer"
+        case .unsupportedPlatform:
+            return "unsupported_platform"
+        case .overlayPresentedInBackgroundScene:
+            return "overlay_presented_in_background_scene"
+        @unknown default:
+            return "unknown_future_error"
```
tonidero, line 140: Maybe `unknown_store_kit_error`?

Sources/Error Handling/StoreKitError+Extensions.swift
```diff
@@ -47,6 +47,29 @@ extension StoreKitError: PurchasesErrorConvertible {
...
+        case .notAvailableInStorefront:
+            return "not_available_in_storefront"
+        case .notEntitled:
+            if #available(iOS 15.4, macOS 12.3, tvOS 15.4, watchOS 8.5, visionOS 1.0, *) {
+                return "not_entitled"
+            } else {
+                return "unknown"
+            }
+        @unknown default:
+            return "unknown_future_error"
```
tonidero, line 69: Same here

tonidero (approved):
LGTM!

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -872,6 +889,36 @@ private extension PurchasesOrchestrator {
...
+
+                if let skError = underlyingError as? SKError {
+                    storeKitErrorDescription = skError.code.trackingDescription
+                } else if let storeKitError = underlyingError as? StoreKitError {
+                    storeKitErrorDescription = storeKitError.trackingDescription
+                } else if let storeKitError = underlyingError as? StoreKit.Product.PurchaseError {
+                    storeKitErrorDescription = storeKitError.trackingDescription
+                } else {
+                    storeKitErrorDescription = nil
+                }
```
vegaro, line 911: I would like this to be a function we can use from other places, but not really sure where to move it to? It would be an extension on Dictionary, which feels weird? Maybe we can just leave it here. cc: @tonidero

tonidero: Hmm maybe some kind of `StoreKitErrorHelper` class to abstract this logic? I think it would be ok even if this is a static method... Wdyt? 

I think this will very likely be used in other diagnostics events, so it would be ideal to avoid duplication.

vegaro:
@RCGitBot please test

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -413,8 +421,11 @@ final class PurchasesOrchestrator {
 
         let addPayment: Bool = self.addPurchaseCompletedCallback(
             productIdentifier: productIdentifier,
-            completion: { transaction, customerInfo, error, cancelled in
+            completion: { [weak self] transaction, customerInfo, error, cancelled in
+                guard let self = self else { return }
```
vegaro, line 425: without this, we were leaking `PurchasesOrchestator` and `DeviceCache` 

cc: @tonidero

vegaro:
@RCGitBot please test

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
tonidero: Hmm interesting... I'm not sure I understand the reason...  I think this should be mostly safe, except maybe we're not ensuring the completion blocks to be called if changing the purchases orchestrator instance for any reason... Which should normally be ok I think.

In any case, would it be possible to do this inside the `trackPurchaseEventIfNeeded` if that was the reason for the leak to avoid changing any other code?

vegaro: well the reason is that the completion block holds a reference to `self`, and we are adding the completion block to the `purchaseCompletedCallbacks`, which I think are not getting triggered when testing, so even if I add it to `trackPurchaseEventIfNeeded`, we still need to call `self. trackPurchaseEventIfNeeded`, and the completion would still be holding the reference to `self`. Am I misunderstanding something?

tonidero: Ohh ok right. Make sense. Thanks for the explanation!

tonidero (approved):
LGTM!

PR #4258, opened 2024-09-10
Pin SwiftLint

JayShortway:
Installs SwiftLint through SPM, and runs it as 
- [x] part of the build ([current problem](https://github.com/RevenueCat/purchases-ios/pull/4258#discussion_r1751968222)),
- [x] in the pre-commit hook, and
- [ ] on CI ([current problem](https://github.com/RevenueCat/purchases-ios/pull/4258#discussion_r1751969452)).

This is a draft because of the problems linked above.

### DIFF
--- a/RevenueCat.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/RevenueCat.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -44,6 +44,15 @@
         "revision" : "26ed3a2b4a2df47917ca9b790a57f91285b923fb",
         "version" : "1.12.0"
       }
+    },
+    {
+      "identity" : "swiftlintplugins",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/SimplyDanny/SwiftLintPlugins",
+      "state" : {
+        "revision" : "7c80ce6f142164b0201871e580b021d1b2c69804",
+        "version" : "0.57.0"
+      }
     }
   ],
   "version" : 2


scripts/swiftlint.sh
```diff
@@ -1,31 +1,3 @@
...
-  
-  pushd "${source_path}"
-  # Run swiftlint but filter out "Linting ..." to clean up output
-  $lint_command 2>&1 | grep -v 'Linting '
-  popd
-else
-  echo "Warning: SwiftLint not installed in ${HOMEBREW_BINARY_DESTINATION}, download from https://github.com/realm/SwiftLint"
-fi
-
+swift package plugin swiftlint lint --quiet
```
JayShortway, line 3: Running this as-is doesn't work in a Run Script Build Phase, not sure why yet. I suspect it has to do with environment variables. This is the error I get: 

> warning: using sysroot for 'iPhoneSimulator' but targeting 'MacOSX'
> error: unable to load standard library for target 'arm64-apple-macosx13.0'

.circleci/config.yml
```diff
@@ -1106,8 +1089,14 @@ jobs:
...
-            bundle exec fastlane run swiftlint raise_if_swiftlint_error:true strict:true \
-            reporter:junit output_file:fastlane/test_output/swiftlint/junit.xml
+            # FIXME This doesn't work because somehow swift package plugins run for every target/module/scheme?
+            #  Only the last of those runs will save the output file, causing issues to be lost.
+            swift package plugin \
+              --allow-writing-to-directory fastlane/test_output/swiftlint \
+              swiftlint \
+              --strict \
+              --reporter junit \
+              --output fastlane/test_output/swiftlint/junit.xml
```
JayShortway, line 1099: This doesn't work because somehow swift package plugins run for every target/module/scheme? Only the last of those runs will save the output file, causing issues to be lost.

scripts/swiftlint.sh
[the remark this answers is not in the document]
JayShortway: Awesome, thanks! Is there an advantage of using the absolute path to `xcrun`? Isn't `/usr/bin` usually part of the `$PATH`?

.circleci/config.yml
JayShortway: Running it with verbose output tells me it's running for each module, which map to the targets specified in our `Package.swift`: 
```shell
info: Finished running in module 'RevenueCat_CustomEntitlementComputation'
info: Finished running in module 'RevenueCatUITests'
info: Finished running in module 'RevenueCatUI'
info: Finished running in module 'RevenueCat'
info: Finished running in module 'ReceiptParserTests'
info: Finished running in module 'ReceiptParser'
```

I think this is `swift package` that's running the `plugin` command for each target. It seems to be happening outside of `swiftlint`'s control. Not sure yet how to make `swift package` confine itself to a single target.

scripts/swiftlint.sh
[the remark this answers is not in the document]
jamesrb1: No advantage if it's already available via `$PATH`.

.circleci/config.yml
JayShortway: There's a [convention](https://github.com/swiftlang/swift-package-manager/blob/main/Documentation/Plugins.md#implementing-the-command-plugin-script) to support a `--target` argument.  SwiftLint's command plugin supports it partly. It passes the `--target` argument on to the actual `swiftlint` program, which doesn't recognize it and errors out. 

There’s [a fork](https://github.com/dshikulin-mwb/SwiftLintPlugins) that fixes this issue. I’ve [asked](https://github.com/SimplyDanny/SwiftLintPlugins/pull/1#issuecomment-2346144183) if they can propose their fix upstream.

JayShortway: 0.57.1 includes a fix for this.

JayShortway:
Superseded by https://github.com/RevenueCat/sdks-circleci-orb/pull/23.

PR #4261, opened 2024-09-11
[Paywall Components] Match properties and behaviors from dashboard

joshdholtz:
## Motivation

Matched properties between frontend/backend and SDK for displaying components end-to-end (used in the first demo)

### `ImageLoader`
- Returning a `CGSize` with the `Image` to be used for manual max height. SwiftUI wasn't displaying images correctly with max height so manually doing it based off of the actual image size.
 
### `.roundedCorner` view modifier
- Only round corners if radius is greater than zero

### `ImageComponentView`
- Fix issue with `.fill` content mode where it needs to render the image in a `Color.clear.overlay`
  - Rendering in overlay prevents the image from pushing the width of parent views (so it properly clips)
- Apply `maxHeight` if given
- Apply rounded corners

### `ImageComponentViewModel`
- Load image object (with light/dark mode and all the image types) from localizations

### `PaywallComponentTypeTransformers`
- Changed `stackAlignment` from `SwiftUI.HorizontalAlignment` to `SwiftUI.Alignment`
  - This was needed for a rework in `StackComponentView` to stretch `VStack`'s full width

### `StackComponentView`
- Moved all common properties for `HStack`, `VStack`, and `HStack` in a `Group` so only one place where view modifiers are being applied
  - Added `.roundedCorner`
- Give all of `VStack`'s `ComponentsView` a max width of `.infinity`
  - Width needs to be full width so that horizontal alignment can move to center and trailing

### `TextComponentView`
- Added a `.fixedSize(horizontal: false, vertical: true)` to fix some lines being cut off
- Added `.padding()` for margin

### `RemoteImage`
- Also return `CGSize` in the content view builder

### `OfferingsResponse`
- `LocalizationData` is no longer just `[String: String]`... it is now `[String: Any]` because full image data is stored under one localization identifier
  - There is a new manual `Codable` encoder/decoder to return localization data `.string` or `.image` to be used by the view models
- Made `paywallComponents: PaywallComponentsData?` optional so that other paywalls can render (since this key doesn't return on existing paywalls)


RevenueCatUI/Templates/Components/Image/ImageComponentViewModel.swift
```diff
@@ -21,20 +21,25 @@ import SwiftUI
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 public class ImageComponentViewModel {
 
+    let localizedStrings: PaywallComponent.LocalizationDictionary
     private let component: PaywallComponent.ImageComponent
 
-    init(component: PaywallComponent.ImageComponent) {
+    let imageInfo: PaywallComponent.ThemeImageUrls
```
vegaro, line 27: can this and `localizedStrings` be private?

RevenueCatUI/Templates/Components/Stack/StackComponentView.swift
```diff
@@ -22,28 +22,66 @@ struct StackComponentView: View {
...
-            .background(viewModel.backgroundColor)
-            .padding(viewModel.padding)
+        } else {
+            content
         }
     }
+}
 
+extension View {
+    func width(_ width: PaywallComponent.WidthSize? = nil) -> some View {
```
vegaro, line 82: nitpick: add line above and after the function to match our Swift style guide. Same for `WidthModifier`, adding lines both at the beginning and at the end

RevenueCatUI/Templates/Components/Text/TextComponentView.swift
```diff
@@ -26,10 +26,13 @@ struct TextComponentView: View {
         Text(viewModel.text)
             .font(viewModel.textStyle)
             .fontWeight(viewModel.fontWeight)
+            .lineLimit(nil)
```
vegaro, line 29: I believe this is the default? https://developer.apple.com/documentation/swiftui/environmentvalues/linelimit

Sources/Paywalls/Components/PaywallImageComponent.swift
```diff
@@ -15,24 +15,24 @@ public extension PaywallComponent {
     struct ImageComponent: PaywallComponentBase {
 
         let type: ComponentType
-        public let url: URL
-        public let cornerRadius: Double
-        public let gradientColors: [ColorHex]
+        public let urlsLid: LocalizationKey
```
vegaro, line 18: What does `lid` stand for?

Sources/Paywalls/Components/PaywallComponentPropertyTypes.swift
```diff
@@ -81,7 +130,7 @@ public extension PaywallComponent {
 
     enum FontWeight: String, Codable, Sendable, Hashable, Equatable {
 
-        case ultraLight
+        case ultraLight = "ultra_light"
```
vegaro, line 133: just curious, but isn't this the default?

vegaro (approved):

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
```diff
@@ -43,9 +43,10 @@ public struct TemplateComponentsView: View {
 
     public var body: some View {
         VStack(spacing: 0) {
-                ComponentsView(
-                    componentViewModels: self.componentViewModels
-                )
+            ComponentsView(
+                componentViewModels: self.componentViewModels
+            )
+            Spacer()
```
jamesrb1, line 49: I assume this is here to expand and make things flow down from top to bottom rather than centre-out, but as the paywall fills up, this could add extra unwanted spacing at the bottom.

Does replacing it with a `.frame(maxHeight: .infinity, alignment: .top)` on `ComponentView(` have the desired effect, without adding a possibly problematic spacer at the bottom?

RevenueCatUI/Templates/Components/Text/TextComponentView.swift
```diff
@@ -26,10 +26,13 @@ struct TextComponentView: View {
...
+            .fixedSize(horizontal: false, vertical: true)
```
jamesrb1, line 30: This makes me a little uneasy, but maybe it's what CSS does automatically?

Sources/Networking/Responses/OfferingsResponse.swift
```diff
@@ -28,6 +28,35 @@ public struct PaywallComponentsData: Codable, Equatable, Sendable {
 
     }
 
+    public enum LocalizationData: Codable, Equatable, Sendable {
+        case string(String), image(PaywallComponent.ThemeImageUrls)
```
jamesrb1, line 32: I think it's our style to put these in two case statements / two lines

Sources/Paywalls/Components/PaywallComponentPropertyTypes.swift
```diff
@@ -18,6 +18,17 @@ import Foundation
 
 public extension PaywallComponent {
 
+    struct ThemeImageUrls: Codable, Sendable, Hashable, Equatable {
+        public let light: ImageUrls
+        public let dark: ImageUrls?
+    }
+
+    struct ImageUrls: Codable, Sendable, Hashable, Equatable {
+        public let original: URL
```
jamesrb1, line 27: Curious on the need for this?

Sources/Paywalls/Components/PaywallComponentPropertyTypes.swift
```diff
@@ -108,8 +157,8 @@ public extension PaywallComponent {
         case caption2
 
         // Swift 5.9 stuff
-        case extraLargeTitle
-        case extraLargeTitle2
+        case extraLargeTitle = "extra_large_title"
+        case extraLargeTitle2 = "extra-large_title_2"
```
jamesrb1, line 161: `extra-large` (hypen) is a typo?

Sources/Paywalls/Components/PaywallImageComponent.swift
jamesrb1: Localization ID.

Maybe we should adopt the suffix `LocalizationID`, which is clear.

Sources/Paywalls/Components/PaywallStackComponent.swift
```diff
@@ -21,23 +21,32 @@ public extension PaywallComponent {
...
 
         public init(components: [PaywallComponent],
                     dimension: Dimension = .vertical(.center),
+                    width: WidthSize?,
                     spacing: CGFloat?,
                     backgroundColor: ColorInfo?,
-                    padding: Padding
+                    padding: Padding,
+                    margin: Padding,
+                    cornerRadiuses: CornerRadiuses
```
jamesrb1, line 39: What do you think of giving these default values in the initializer? It makes creating paywall samples via code much quicker/easier, and I think there will be a continued need to do so as we expand the components' properties/capabilities.

jamesrb1 (commented):
Looks good.

This breaks the paywall samples created in code, but adding in default values for the new properties in the .init methods makes it much easier to write/update these samples (which are used when trying new things solely on the SDK-side). What do you think?

joshdholtz:
> Looks good.
> 
> This breaks the paywall samples created in code, but adding in default values for the new properties in the .init methods makes it much easier to write/update these samples (which are used when trying new things solely on the SDK-side). What do you think?

Oops, will fix the paywall samples! 💪

joshdholtz:
@RCGitBot please test

joshdholtz:
@RCGitBot please test

PR #4263, opened 2024-09-11
Add `$userAgent` tracking for attribution networks

tonidero:
### Description
This will set up a new property `$userAgent` set to `"true"` when using any attribution network method. When this is set, the backend will know to collect the user agent. 

This is a follow-up for the Kochava PR: #4238 

- [ ] Holding until backend has been deployed


tonidero:
Closing this in favor of #4274

PR #4270, opened 2024-09-12
[Diagnostics] Refactor diagnostics track methods to handle background work automatically

tonidero:
### Description
Before this, callers of the diagnostics `track` methods had to handle concurrency and offloading the work to a background thread. This moves the work inside the `DiagnosticsTracker`, which makes the methods non-async anymore.

### DIFF
--- a/Sources/Diagnostics/DiagnosticsTracker.swift
+++ b/Sources/Diagnostics/DiagnosticsTracker.swift
@@ -16,10 +16,10 @@ import Foundation
 protocol DiagnosticsTrackerType {
 
     @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
-    func track(_ event: DiagnosticsEvent) async
+    func track(_ event: DiagnosticsEvent)
 
     @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
-    func trackCustomerInfoVerificationResultIfNeeded(_ customerInfo: CustomerInfo) async
+    func trackCustomerInfoVerificationResultIfNeeded(_ customerInfo: CustomerInfo)
 
     @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
     // swiftlint:disable:next function_parameter_count
@@ -28,7 +28,7 @@ protocol DiagnosticsTrackerType {
                               errorMessage: String?,
                               errorCode: Int?,
                               storeKitErrorDescription: String?,
-                              responseTime: TimeInterval) async
+                              responseTime: TimeInterval)
 
     @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
     // swiftlint:disable:next function_parameter_count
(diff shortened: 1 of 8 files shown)


Tests/UnitTests/Mocks/MockOperationDispatcher.swift
```diff
@@ -90,9 +90,14 @@ class MockOperationDispatcher: OperationDispatcher {
...
             super.dispatchOnWorkerThread(jitterableDelay: delay, block: block)
         } else if self.shouldInvokeDispatchOnWorkerThreadBlock {
+            let semaphore = DispatchSemaphore(value: 0)
+
             Task<Void, Never> {
                 await block()
+                semaphore.signal()
             }
+
+            semaphore.wait()
```
tonidero, line 100: I had to do this for this to work as expected and wait for the completion of the block. It's very hacky, but I think it's what we wanted? Would appreciate feedback to make sure this is ok!

Sources/Diagnostics/DiagnosticsTracker.swift
```diff
@@ -36,30 +36,35 @@ protocol DiagnosticsTrackerType {
...
         self.diagnosticsFileHandler = diagnosticsFileHandler
+        self.diagnosticsDispatcher = diagnosticsDispatcher
         self.dateProvider = dateProvider
     }
 
-    func track(_ event: DiagnosticsEvent) async {
-        await self.clearDiagnosticsFileIfTooBig()
-        await self.diagnosticsFileHandler.appendEvent(diagnosticsEvent: event)
+    func track(_ event: DiagnosticsEvent) {
+        self.diagnosticsDispatcher.dispatchOnWorkerThread {
```
tonidero, line 59: This is the main change, now dispatching to a background thread happens internally.

Tests/UnitTests/Mocks/MockOperationDispatcher.swift
jamesrb1: Before it would **not** wait for `block()` to run before returning, but with your changes it does. This can be done by just calling, `await block()`, no need for the semaphone/Task.

Sources/Diagnostics/DiagnosticsTracker.swift
jamesrb1: I don't think you need a `diagnosticsDispatcher`? You could just create a Task here,

```swift
func track(_ event: DiagnosticsEvent) {
    Task {
        await self.clearDiagnosticsFileIfTooBig()
        await self.diagnosticsFileHandler.appendEvent(diagnosticsEvent: event)
    }
}
```

However, with this change, events submitted to `track()` aren't guaranteed to run sequentially as they are with the `OperationDispatcher`. But this guarantee won't have existed before, so unless you want to make that change, I think using `Task` is the way to go.

Tests/UnitTests/Mocks/MockOperationDispatcher.swift
tonidero: I believe we can't do `await block()` or `await Task` because we're inside a non-async function right? We would need to make this the operation dispatcher async, but the whole point of this object is to dispatch tasks in non-async contexts I believe.

Lmk if you want to chat more about this! I could have missed something! Note that this semaphore is only for tests too.

Sources/Diagnostics/DiagnosticsTracker.swift
tonidero: There are 2 main reasons I decided to go with the OperationDispatcher:
- First, as you said, this will make the operations run sequentially, which I believe would be safer, since there could be chances of changing contexts between tracking different events, and if we reach the file too big issue, we might be missing extra data. As you said, this problem would be present right now.
- Second, I wanted to be able to test this class. If I used `Task`, I don't have a good way to determine when the tracking finished to do the assertions. With OperationDispatcher, we can pass a MockOperationDispatcher in the tests that runs things synchronously (mostly), allowing us to do the assertions immediately.

Lmk if you would like to talk more about it!

Tests/UnitTests/Mocks/MockOperationDispatcher.swift
jamesrb1: Oh sheesh, I saw the `async` on the block's signature and misread it as the function's. Semaphore is the way to go!

Sources/Diagnostics/DiagnosticsTracker.swift
jamesrb1: OK, sounds good!

jamesrb1 (approved):

vegaro (approved):

JayShortway (approved):
Nice solution!

tonidero:
Going to merge this one now that the previous PRs have been merged. No significant changes have happened in this PR so we should be good

PR #4271, opened 2024-09-12
Stateobject instantiation fix

jamesrb1:
Just a very quick mini-fix I noticed, where we should avoid using any local variables when instantiating a @StateObject. This is because the state object's wrapped parameter is an `autoclosure` that is only executed once during the lifetime of the view, and so every time the view is recreated after the first time, these local variable assigned objects are created and then go unused.

Also added a missing case that was preventing paywalls tester from compiling.


tonidero (approved):
Ohh interesting! That makes sense, thanks for fixing it!

PR #4272, opened 2024-09-12
Update allowSharingAppStoreAccount deprecation message

fire-at-will:
### Description
Updates the deprecation message of `allowSharingAppStoreAccount` to tell people who are already using the field to continue using it and not remove it.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1548,7 +1548,7 @@ public extension Purchases {
     /**
      * Deprecated
      */
-    @available(*, deprecated, message: "Configure behavior through the RevenueCat dashboard instead. If have configured the "Legacy" restore behavior in the [RevenueCat Dashboard](app.revenuecat.com) and are currently setting this to `true`, keep this setting active.")
+    @available(*, deprecated, message: "Configure behavior through the RevenueCat dashboard instead. If you have configured the \"Legacy\" restore behavior in the [RevenueCat Dashboard](app.revenuecat.com) and are currently setting this to `true`, keep this setting active.")
     @objc var allowSharingAppStoreAccount: Bool {
         get { purchasesOrchestrator.allowSharingAppStoreAccount }
         set { purchasesOrchestrator.allowSharingAppStoreAccount = newValue }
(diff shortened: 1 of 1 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1548,7 +1548,7 @@ public extension Purchases {
     /**
      * Deprecated
      */
-    @available(*, deprecated, message: "Configure behavior through the RevenueCat dashboard instead")
+    @available(*, deprecated, message: "Configure behavior through the RevenueCat dashboard instead. If you are using the legacy restore behavior and are currently setting this to true, continue doing so.")
```
aboedo, line 1551:
```suggestion
    @available(*, deprecated, message: "Configure behavior through the RevenueCat dashboard instead. If have configured the "Legacy" restore behavior in the [RevenueCat Dashboard](app.revenuecat.com) and are currently setting this to `true`, keep this setting active.")
```

aboedo (approved):
thanks for taking care of this so fast!

Sources/Purchasing/Purchases/Purchases.swift
aboedo: we might wanna open up equivalents for other SDKs

jamesrb1 (approved):
Seems like an odd caveat, and worthy of more explanation somewhere? And customers will have to [live with the warning message](https://stackoverflow.com/questions/31540446/how-to-silence-a-warning-in-swift).

Approving so you are not blocked.

Sources/Purchasing/Purchases/Purchases.swift
fire-at-will: > we might wanna open up equivalents for other SDKs
Yeah, we should! I'll open up some PRs for it tomorrow. 👍

aboedo:
@jamesrb1 this was already true before this PR, it's actually been this way for > 4 years. 

So this PR is more surfacing a caveat that has implicitly been there for 4 years and making it explicit, rather than changing anything with regards to how things behave. 

It actually came up yesterday when a customer ran into this, so we figured that we should at least improve the error message and related docs. It's very uncommon for it to come up, since the related feature has been deprecated for a very long time, but it's not impossible, so this is meant as a small improvement over the current status

PR #4273, opened 2024-09-13
[Paywalls] Use store product for `{{ sub_period }}` duration

joshdholtz:
### Motivation

`{{ sub_period }} ` was returning package identifier when using custom package names

### Description

Look at the `StoreProduct` periods to determine the period length first 😁


aboedo (approved):
oh this is way smarter, nice!

Issue #4246, opened 2024-09-06
Rounding error in pricePerYear, pricePerMonth and similar variants


wilfredbtan:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
There seems to be a rounding issue for prices such as $19.99 and $1.99. Furthermore, there is inconsistent formatting between localizedPriceString and the localizedPricePerYear variants. 

1. Environment
   1. Platform: iOS
   2. SDK version: 5.3.2
   3. StoreKit version:
      - [] StoreKit 1
      - [x] StoreKit 2 (_enabled with `usesStoreKit2IfAvailable(true)`_)
   4. OS version: 17.6.1
   5. Xcode version: 15.1
   6. Device and/or simulator:
      - [x] Device
      - [ ] Simulator
   7. Environment:
      - [x] Sandbox
      - [ ] TestFlight
      - [ ] Production
   8. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue. **Complete logs with `Purchases.logLevel = .verbose` will help us debug this issue.**

```
Logs here
```

3. Steps to reproduce, with a description of expected vs. actual behavior

I have an annual subscription priced at $19.99/year, when using `localizedPriceString` (e.g. `yearlyPackage.storeProduct.localizedPriceString`), it is displayed as "**S $19.99**" (with the S infront). However, when I use `localizedPricePerYear`, the displayed string is "**$19.98**" (without the S and a rounding error in the last cent). 

The same issue occurs for my monthly subscription priced at $2.99. That is, using localizedPricePerYear on my monthly package is displayed as "$2.98" (with the last cent rounded wrongly)

I can't simply use localizedPriceString for all my products because I intend to show the localizedPricePerMonth for my annual subscription to aid the user in comparing my annual and monthly prices.

Thus it's a bit inconvenient if both are formatted differently. 
 
[...]

PR #4275, opened 2024-09-16
Round price per period calculations to nearest 2-decimal

tonidero:
### Description
We were rounding the price per period calculations to the floor of the second decimal of the price. This changes it to round to the closest number following the logic in: https://developer.apple.com/documentation/foundation/nsdecimalnumber/roundingmode

This might help with #4246


JayShortway (approved):
Nice find!

tonidero:
Need to update snapshots

PR #4285, opened 2024-09-17
created PoC code to handle deeplinks nicely

aboedo:
Quick-n-dirty PoC to open up paywalls from a deeplink. 

## Usage: 

In a SwiftUIView, just add a modifier: 
```swift
    .handleRevenueCatDeeplinks()
```

And we take it from there! You will have to set up deeplinks for your app, though, obviously. 
Then, once set up, you can deeplinks like: 

```
paywallsTester://paywall?offeringID=Andy_Test_2
paywallsTester://paywall?offeringID=Andy_Test_3
paywallsTester://paywall?offeringID=Andy_Test_4
```

Also added them to paywallsTester for easy testing.

### DIFF
--- a/RevenueCatUI/Views/RevenueCatDeeplinkHandler.swift
+++ b/RevenueCatUI/Views/RevenueCatDeeplinkHandler.swift
@@ -1,0 +1,107 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  RevenueCatDeeplinkHandler.swift
+//
+//  Created by Andrés Boedo on 9/17/24.
+
+import Foundation
+import RevenueCat
+import SwiftUI
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@available(macOS, unavailable, message: "RevenueCatUI does not support macOS yet")
+@available(tvOS, unavailable, message: "RevenueCatUI does not support tvOS yet")
+struct RevenueCatDeeplinkHandlerView<Content: View>: View {
+    @State private var isShowingPaywall: Bool = false
+    @State private var offeringID: String?
+
+    let content: Content
+
+    init(@ViewBuilder content: () -> Content) {
+        self.content = content()
+    }
+
+    var body: some View {
+        content
+            .onOpenURL { url in
+                if let extractedOfferingID = extractOfferingID(from: url) {
+                    offeringID = extractedOfferingID
+                    isShowingPaywall = true
+                }
+            }
+            .sheet(
+                isPresented: $isShowingPaywall,
+                onDismiss: {
+                    offeringID = nil
+                },
+                content: {
+                    if let offeringID = offeringID {
+                        OfferingLoaderView(offeringID: offeringID)
+                    } else {
+                        Text("Invalid offering ID.")
+                    }
+                })
+    }
+
+    private func extractOfferingID(from url: URL) -> String? {
+        let components = URLComponents(url: url, resolvingAgainstBaseURL: false)
+        return components?.queryItems?.first(where: { $0.name == "offeringID" })?.value
+    }
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@available(macOS, unavailable, message: "RevenueCatUI does not support macOS yet")
+@available(tvOS, unavailable, message: "RevenueCatUI does not support tvOS yet")
+extension View {
+    func handleRevenueCatDeeplinks() -> some View {
+        RevenueCatDeeplinkHandlerView {
+            self
+        }
+    }
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@available(macOS, unavailable, message: "RevenueCatUI does not support macOS yet")
+@available(tvOS, unavailable, message: "RevenueCatUI does not support tvOS yet")
+struct OfferingLoaderView: View {
+    let offeringID: String
+    @State private var offering: Offering?
+    @State private var isLoading = true
+
+    var body: some View {
+        Group {
+            if let offering = offering {
+                PaywallView(offering: offering)
+            } else if isLoading {
+                ProgressView()
+            } else {
+                Text("Could not load offering.")
+            }
+        }
+        .task {
+            await fetchOffering()
+        }
+    }
+
+    private func fetchOffering() async {
+        do {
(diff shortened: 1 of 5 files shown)


RevenueCatUI/Views/RevenueCatDeeplinkHandler.swift
```diff
@@ -0,0 +1,107 @@
...
+    let content: Content
+
+    init(@ViewBuilder content: () -> Content) {
+        self.content = content()
+    }
+
+    var body: some View {
+        content
+            .onOpenURL { url in
+                if let extractedOfferingID = extractOfferingID(from: url) {
```
tonidero, line 34: Hmm correct me if I'm wrong but this would actually open in any deep link that has the `offeringID` query parameter right? I wonder if we should verify the deep links path as well? As in, having `anything://paywall?offeringID=test-offering-id` and we verify the path is `/paywall` and the `offeringID` exists.

RevenueCatUI/Views/RevenueCatDeeplinkHandler.swift
```diff
@@ -0,0 +1,107 @@
...
+
+    let content: Content
+
+    init(@ViewBuilder content: () -> Content) {
+        self.content = content()
+    }
+
+    var body: some View {
+        content
+            .onOpenURL { url in
```
tonidero, line 33: Not sure if we would like to also support UIKit apps... but I guess that can come in a separate PR.

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/AppContentView.swift
```diff
@@ -44,6 +44,7 @@ struct AppContentView: View {
             }
             #endif
         }
+        .handleRevenueCatDeeplinks()
```
tonidero, line 47: Nothing to do here but I must say, I'm still not used to having these things as modifiers of a view, seems like a very weird API to me 😅

Tests/TestingApps/PaywallsTester/PaywallsTester/Products.storekit
```diff
@@ -1,4 +1,14 @@
...
+  "appPolicies" : {
+    "eula" : "",
+    "policies" : [
+      {
+        "locale" : "en_US",
+        "policyText" : "",
+        "policyURL" : ""
+      }
+    ]
+  },
```
aboedo, line 11: oh interesting, the stuff in this file just gets updated automatically when you open up Xcode 16. Will revert since it's not actually part of the PR

RevenueCatUI/Views/RevenueCatDeeplinkHandler.swift
```diff
@@ -0,0 +1,107 @@
...
+//  Created by Andrés Boedo on 9/17/24.
+
+import Foundation
+import RevenueCat
+import SwiftUI
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@available(macOS, unavailable, message: "RevenueCatUI does not support macOS yet")
+@available(tvOS, unavailable, message: "RevenueCatUI does not support tvOS yet")
+struct RevenueCatDeeplinkHandlerView<Content: View>: View {
```
aboedo, line 21: to get this through the finish line, we really need to: 
- settle on the exact format. I used `/paywall` in the examples but I'm not even enforcing it, maybe `revenuecatui` is more appropriate? 
- clean up error handling and logging
- add tests

RevenueCatUI/Views/RevenueCatDeeplinkHandler.swift
```diff
@@ -0,0 +1,107 @@
...
+@available(tvOS, unavailable, message: "RevenueCatUI does not support tvOS yet")
+struct OfferingLoaderView: View {
+    let offeringID: String
+    @State private var offering: Offering?
+    @State private var isLoading = true
+
+    var body: some View {
+        Group {
+            if let offering = offering {
+                PaywallView(offering: offering)
```
tonidero, line 81: Thinking about it, we might also want to offer a way to not display the paywall if the user has an entitlement (passing a `requiredEntitlementIdentifier`). But this could also be a separate PR I guess.

tonidero (commented):
Mostly I think we should validate the deep link path at least, to try to avoid all links opening the paywall inadvertently. Looks great otherwise!

RevenueCatUI/Views/RevenueCatDeeplinkHandler.swift
tonidero: Oh I see you mentioned this in another comment. Ignore me :P

RevenueCatUI/Views/RevenueCatDeeplinkHandler.swift
aboedo: yeah, that's a good point. It could be done as an extra query param in the deeplink too

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/AppContentView.swift
aboedo: 🤔 what would a more natural API look like to you as a dev? I'm open to anything as long as its easy to do. I know others do the setup in the app delegate, but the industry seems to be moving away from that

RevenueCatUI/Views/RevenueCatDeeplinkHandler.swift
aboedo: yeah, I figured for UIKit we'd basically do the appDelegate + present from root view controller path

RevenueCatUI/Views/RevenueCatDeeplinkHandler.swift
aboedo: haha yeah I'm not validating, this is as quick and dirty as it comes

RevenueCatUI/Views/RevenueCatDeeplinkHandler.swift
tonidero: Yeah, I think just adding a method that can be called by a developer in the app delegate method would work 👍

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/AppContentView.swift
tonidero: Yeah, I think this is the way it's supposed to be in SwiftUI (like the `onOpenURL` API we're using). Just seems very weird to tie this behavior to a view IMO... So yeah, I think this is the most SwiftUI solution for now.

JayShortway: Yes I feel you Toni haha, but indeed this seems to be the SwiftUI way. In Compose it would be more natural to add a route to a navigation graph. The downside of that is that there are multiple ways to build such a graph, as mentioned on Slack.

Issue #4177, opened 2024-08-13
Crash at launch Xcode 16 iOS 18 - EXC_BAD_ACCESS 0x100b63934 (0x100b638e0 + 84) specialized withCheckedContinuation<A>(isolation:function:_:)


joaopdcgarcia:
- [ ] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
Crashes at launch with a not so useful stacktrace.

RevenueCat 4.43.2 Xcode 16 beta 5 + iOS 18 crash 

1. Environment
   1. Platform: iOS
   2. SDK version: 4.43.2.
   3. StoreKit version:
      - [x] StoreKit 1
      - [ ] StoreKit 2 (_enabled with `usesStoreKit2IfAvailable(true)`_)
   4. OS version: 18
   5. Xcode version: 16 beta 5
   6. Device and/or simulator:
      - [x] Device
      - [ ] Simulator
   7. Environment:
      - [ ] Sandbox
      - [x] TestFlight
      - [ ] Production
   8. How widespread is the issue. Percentage of devices affected.
   100%
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue.

No logs, just this stacktrace.

3. Steps to reproduce, with a description of expected vs. actual behavior

Launch the app. 
 
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

```
<compiler-generated>
specialized withCheckedContinuation<A>(isolation:function:_:)
```

![image](https://github.com/user-attachments/assets/6774a67e-475c-4a05-ac0e-bd34e429dbe2)

![image](https://github.com/user-attachments/assets/bda520f0-a031-4480-80c2-598fd4ec35f8)

5. **Additional context**

We want to keep supporting iOS 12. *Is the v.4.x of RevenueCat going to be deprecated?*

RCGitBot:
[...]

PR #4286, opened 2024-09-18
Replace withCheckedThrowingContinuation Calls With withUnsafeThrowingContinuation

fire-at-will:
### Motivation
iOS 18 introduces a bug in Swift that causes calls to `withCheckedThrowingContinuation` to crash in certain scenarios. This PR provides some promise in addressing this issue, documented in https://github.com/RevenueCat/purchases-ios/issues/4177

### Description
This PR replaces all calls to `withCheckedThrowingContinuation` to `withUnsafeThrowingContinuation`, which we're unable to reproduce the issue with.

### Other Notes
`withUnsafeThrowingContinuation` doesn't provide any checking to ensure that the continuation isn't called multiple times, so this is something we'll need to be aware of going forward. Luckily, all existing usages of `withCheckedThrowingContinuation` were in just a few files and in small functions, so this is easy to check manually for now.

### DIFF
--- a/Sources/FoundationExtensions/AsyncExtensions.swift
+++ b/Sources/FoundationExtensions/AsyncExtensions.swift
@@ -101,7 +101,10 @@ internal enum Async {
     static func call<Value>(
         method: (@escaping @Sendable (Value) -> Void) -> Void
     ) async -> Value {
-        return await withCheckedContinuation { continuation in
+        // Note: We're using UnsafeContinuation instead of Checked because
+        // of a crash in iOS 18.0 devices when CheckedContinuations are used.
+        // See: https://github.com/RevenueCat/purchases-ios/pull/4286
+        return await withUnsafeContinuation { continuation in
             @Sendable
             func complete(_ value: Value) {
                 continuation.resume(with: .success(value))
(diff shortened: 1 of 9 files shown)


fire-at-will:
@RCGitBot please test

Sources/FoundationExtensions/AsyncExtensions.swift
```diff
@@ -80,7 +80,7 @@ internal enum Async {
     static func call<Value, Error: Swift.Error>(
         method: (@escaping @Sendable (Result<Value, Error>) -> Void) -> Void
     ) async throws -> Value {
-        return try await withCheckedThrowingContinuation { continuation in
+        return try await withUnsafeThrowingContinuation { continuation in
```
aboedo, line 83: let’s add a comment pointing to the issue for future reference

aboedo (approved):
We could probably follow up with a check to only use this version in iOS 18. Hopefully we wouldn’t have to live with it for too long in any case

aboedo:
@fire-at-will thanks for taking care of this!

kristofferbv:
We’ve tested the branch, and it resolves the crash on launch. However, when opening the paywall, it crashes again, this time in withCheckedContinuation:
Skjermbilde 2024-09-18 kl  10 40 03
Skjermbilde 2024-09-18 kl  10 38 54

kristofferbv:
> @kristofferbv thanks so much for reporting back! I've updated the branch to replace all instances of `withCheckedContinuation` -> `withUnsafeContinuation`, could you try it again and let us know if it does the trick?

Thanks! It works now with the latest changes. It would be really helpful if you could merge this branch into main.

aboedo:
@kristofferbv thanks for confirming! Merging to main now, will go out in the next release, likely today or tomorrow

PR #4293, opened 2024-09-23
DO NOT MERGE Add metadata to purchase

vlopezferrando:
### Motivation

The backend now supports (with the right feature flag) the addition of metadata when doing a purchase. This metadata is included in all generated events and `CustomerInfo`.

### Description

In this PR, we:

- Add an optional parameter `metadata` with default value `nil` to all `purchase` methods.
- Parse the new `metadata` field in the `CustomerInfo`.

### Tests

- Added calls to purchases methods in `PurchasesAPI.swift`.
- Added an integration test `testPurchaseMetadataIsInResponse()` that does a purchase with metadata, then gets the `customerInfo` and checks it includes the metadata.


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -53,6 +53,22 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
...
+
+    public func purchase(package: Package, completion: @escaping PurchaseCompletedBlock) {
+        purchasesOrchestrator.purchase(product: package.storeProduct, package: nil, transactionMetadata: nil, completion: completion)
+    }
+
+    public func purchase(product: StoreProduct, transactionMetadata: [String : String]?, completion: @escaping PurchaseCompletedBlock) {
+        purchasesOrchestrator.purchase(product: product, package: nil, transactionMetadata: transactionMetadata, completion: completion)
+    }
+
+    public func purchase(package: Package, transactionMetadata: [String : String]?, completion: @escaping PurchaseCompletedBlock) {
```
fire-at-will, line 68: Instead of making brand new purchase functions to make a purchase with metadata, I'd consider modifying the existing purchase functions and making the `transactionMetadata` parameter default to `nil` so that it isn't a breaking change.

There are quite a few purchase functions today across both `Purchases.swift` and `Purchases+async.swift`. Ideally folks would be able to upload transaction metadata no matter which function call they're already using without having to migrate to a new one.

fire-at-will (commented):
Looks like it's off to a good start! Let me know if you'd like to pair tomorrow on CI/writing tests/anything else and I'll be glad to hop on a call to help out 😄

vlopezferrando:
@MarkVillacampa @fire-at-will Could you take a look at this PR? It is ready for review.

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -112,6 +113,7 @@ final class TransactionPoster: TransactionPosterType {
                                          receipt: encodedReceipt,
                                          product: product,
                                          appTransaction: appTransaction,
+                                         metadata: data.metadata,
```
greenietea, line 116: Feel free to ignore since I don't know next to nothing about SDK. Why have a separate metadata parameter when data is already passed along in purchasedTransactionData?

vlopezferrando: I think you are right... in fact, I believe we are not even using that parameter and should remove it. I'll check if that's true.

vlopezferrando: Confirmed and removed unnecessary parameters. Great catch!!

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -951,22 +950,27 @@ public extension Purchases {
...
+    func purchase(product: StoreProduct,
+                  metadata: [String: String]? = nil,
+                  completion: @escaping PurchaseCompletedBlock) {
+        purchasesOrchestrator.purchase(product: product, package: nil, metadata: metadata, completion: completion)
     }
 
-    func purchase(product: StoreProduct) async throws -> PurchaseResultData {
-        return try await purchaseAsync(product: product)
+    func purchase(product: StoreProduct, metadata: [String: String]? = nil) async throws -> PurchaseResultData {
+        return try await purchaseAsync(product: product, metadata: metadata)
```
JayShortway, line 961: While adding a parameter with a default value is source compatible, it is not binary compatible. See [Library Evolution in Swift](https://www.swift.org/blog/library-evolution/): 

> ### Examples of non-resilient changes
> [...]
> - Adding a parameter to a function’s parameter list (even if a default value is provided)

This can cause crashes when other libraries depend on purchases-ios and encounter an incompatible version at runtime. 

Following [SemVer](https://semver.org/) strictly, this would require a major release.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -951,22 +950,27 @@ public extension Purchases {
         return await productsAsync(productIdentifiers)
     }
 
-    @objc(purchaseProduct:withCompletion:)
-    func purchase(product: StoreProduct, completion: @escaping PurchaseCompletedBlock) {
-        purchasesOrchestrator.purchase(product: product, package: nil, completion: completion)
+    @objc(purchaseProduct:metadata:withCompletion:)
+    func purchase(product: StoreProduct,
+                  metadata: [String: String]? = nil,
```
JayShortway, line 955: Just curious whether a deliberate decision has been made between defaulting to `nil` and empty?

Sources/Purchasing/Purchases/Purchases.swift
JayShortway: Reverting to your previous approach of adding new functions would be both source and binary compatible, but I definitely agree with @fire-at-will that that would result in too many purchase functions.

JayShortway (commented):
Nice job! Going in the right direction. Just a single comment on binary incompatibility.

Sources/Purchasing/Purchases/Purchases.swift
JayShortway: This is good to know, thanks! Adding new functions also keeps it binary compatible from a Swift perspective.

vlopezferrando:
Thanks, @JayShortway and @aboedo. We'll refactor the PR to use new methods instead of replacing the old ones. @MarkVillacampa will pick it up from here.

PR #4294, opened 2024-09-24
Cleanup of strings in Customer Center

vegaro:
Adds some strings that were not being localized


RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -236,13 +236,24 @@ private final class ManageSubscriptionPurchases: ManageSubscriptionsPurchaseType
...
+        case .day:
+            components.day = self.value
+        case .week:
+            components.weekOfMonth = self.value
+        case .month:
+            components.month = self.value
+        case .year:
+            components.year = self.value
+        default:
+            return NSLocalizedString("common_unknown", comment: "Unknown duration")
```
tonidero, line 253: Hmm I'm not sure I follow this, would we have a localized `common_unknown` key in the SDK? Ideally it should come from the backend, though we would need access to the localization data from here then...

tonidero (commented):
Just a question, looks good otherwise!

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
vegaro: this was just me mixing two things. This should just be `Unknown`. Good catch

vegaro:
@tonidero found a bunch of other things that needed to be localised so I pushed those fixes here

PR #4295, opened 2024-09-24
Run CI tests on iOS18/watchOS11 & Use Xcode 16

fire-at-will:
## Motivation
Apple released iOS 18.0 recently, so we need to update our CI pipeline to run tests on it! This is also necessary for testing iOS18 exclusive features like win-back offers.

### Supporting Xcode 16
- Updates the default Xcode version to Xcode 16.0
- The `pod-lib-lint` and `create-tag` workflows still use Xcode 15.x to avoid some Sendable warnings that come up when `pod lib lint` is run. These workflows can be changed to use Xcode 16 after the warnings themselves have been addressed. Here are the warnings we're seeing today:
Screenshot 2024-09-24 at 3 02 00 PM

### Running Tests on iOS 18
- Introduces new `run-revenuecat-ui-ios-18` and `run-test-ios-18` CI jobs. Both jobs are added to the `build-test` job so they are run during both partial and full CI pipeline executions.
- Updates the iOS simulators where required.
- Introduces new baseline snapshots for tests running on iOS 18.

### Running Tests on watchOS 11
- Updates the simulator/OS version used in the `run-test-watchos` and `spm-revenuecat-ui-watchos` jobs to run on watchOS11.
- Updates the network snapshots for watchOS - it looks like the ordering of the fields in the POST bodies has changed. Here's what the changes look like for each watchOS snapshot:

Screenshot 2024-09-26 at 9 30 20 AM

### CI Job Workflow Changes
- Updates the fastlane version in the CocoaPods Installation Tests to 2.222.0 to match the fastlane version used elsewhere in our CI pipelines.
- Makes two changes to avoid a fastlane issue:
   - Changes the CI machine to use Ruby 3.2.0 
   - Updates the cache key used for caching the CI machine dependencies to use the `v3` prefix. We suspect that the existing `v2` prefix was caching some Ruby dependencies that are not compatible with the new CircleCI image, and this was causing fastlane to file on the `replace_api_key_integration_tests` lane.
   - Thanks for the help in figuring this out @tonidero !

### Changes to Address Swift 6
- Fixes some `Function is unused` compilation errors in the revenuecat-ui test suite (these changes can be found in https://github.com/RevenueCat/purchases-ios/pull/4295/commits/e4a4c3acf1c4b4baa5296967c120a086c89ce818)

### Other Notes
I decided to leave us running iOS 17 & 18 tests in the `build-test` job to be run on both partial and full CI pipeline runs since iOS 18 is so new. We can remove the iOS 17 tests from the partial pipeline runs once more people have moved on to iOS 18.

### Known Issues
In some CI jobs, multiple setup steps are repeated during the flow. We should remove these duplicate executions, but they appear to be happening in `main` as well, so we didn't fix that here to keep the scope of the PR as small as possible.


fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

.circleci/config.yml
```diff
@@ -127,16 +127,16 @@ commands:
...
       - restore_cache:
           keys:
-            - v2-gem-cache-{{ checksum "Gemfile.lock" }}-{{ arch }}
+            - v3-gem-cache-{{ checksum "Gemfile.lock" }}-{{ arch }}
       # CircleCI images fail on Ruby 3.3
       # https://github.com/fastlane/fastlane/issues/21794#issuecomment-2021331335
       # Also, not using macos/switch-ruby because its broken
       - run:
-          name: Set Ruby 3.2.2
+          name: Set Ruby 3.2.0
```
vegaro, line 135: maybe add a note here why it's still 3.2.0

.circleci/config.yml
```diff
@@ -1383,9 +1460,11 @@ workflows:
       - spm-release-build
       - spm-release-build-xcode-14
       - spm-release-build-xcode-15
+      - spm-release-build-xcode-16-1
```
vegaro, line 1463: oh interesting, you're running on the beta on purpose? or should we run this on xcode 16 instead?

fire-at-will: I think @tonidero made this change - can you think of a reason not to do 16.0 @tonidero ?

tonidero: Oh I think the `spm-release-build` would run in xcode 16.0 with the other changes?

tonidero (approved):
I think everything looks good! Thanks for taking care of this! Might be good to wait for another check from someone more iOS savvy just in case :)

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendGetIntroEligibilityTests/watchOS-testPostsProductIdentifiers.1.json
```diff
@@ -1,7 +1,6 @@
 {
   "headers" : {
     "Authorization" : "Bearer asharedsecret",
-    "content-type" : "application/json",
```
aboedo, line 4: I'm guessing we don't really need the updates to the files that have the same content, right? only the addition of the new ones?

fastlane/Fastfile
```diff
@@ -1077,14 +1077,16 @@ lane :check_pods do
     verbose: true,
     podspec: 'RevenueCat.podspec',
     platforms: 'ios,osx,tvos,watchos,visionos',
-    fail_fast: true
+    fail_fast: true,
+    allow_warnings: true
   )
   pod_lib_lint(
```
aboedo, line 1083: in what circumstances do we actually use `pod_lib_lint` with Xcode 16? 
Asking because this change does worry me a bit - the whole point of pod lib lint is to find warnings that we otherwise miss. It might be more valuable to keep this running in Xcode 15 if it can still warn us about stuff that isn't this, or otherwise when running in Xcode 16 maybe update the concurrency mode

aboedo (commented):
maybe we can revert the json files that only contain the same keys but reordered to make reviewing a bit easier? 😅 

only concern is that I think we should see whether we can keep the warnings in pod lib lint, otherwise it loses a lot of its value

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendGetIntroEligibilityTests/watchOS-testPostsProductIdentifiers.1.json
fire-at-will: We could remove them for the sake of making the review easier. Even though the contents are functionally the same, since the order they appear in is different, the tests fail without these changes present.

fastlane/Fastfile
fire-at-will: I see your point above removing the warnings making things less safe, and I have to agree with you! We currently run `pod_lib_lint` in both the `pod-lib-lint` and `create-tag` (via the release-checks lane) workflows. I think it's probably fine to continue using Xcode 15 here - I'll try it out and see how it goes!

I did double check, and unfortunately `pod_lib_lint` provides no way to only ignore certain warnings 😅

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fastlane/Fastfile
fire-at-will: @aboedo I've updated the PR so that the `pod-lib-lint` and `create-tag` workflows use Xcode 15!

.circleci/config.yml
```diff
@@ -1383,9 +1465,11 @@ workflows:
       - spm-release-build
       - spm-release-build-xcode-14
       - spm-release-build-xcode-15
+      - spm-release-build-xcode-16
```
vegaro, line 1468: I believe this is the same as `spm-release-build` since that one runs on the default (16)

vegaro (approved):

.circleci/config.yml
```diff
@@ -1255,10 +1332,15 @@ workflows:
...
       - run-test-ios-17
+      - run-test-ios-18
       - pod-lib-lint
       - run-revenuecat-ui-ios-17
+      - run-revenuecat-ui-ios-18
 
   create-tag:
+    executor:
+      name: macos-executor
+      xcode_version: '15.2.0'
```
MarkVillacampa, line 1343: does it not work with Xcode 16?

.circleci/config.yml
```diff
@@ -442,6 +455,7 @@ jobs:
   pod-lib-lint:
     executor:
       name: macos-executor
+      xcode_version: '15.2.0'
```
MarkVillacampa, line 458: Does it not work with Xcode 16

.circleci/config.yml
fire-at-will: No :(

The fastlane `pod_lib_lint` job fails because of the Sendable warnings that we've been facing. `pod lib lint` doesn't provide a way to silence specific warnings, it's either all or nothing, so I decided to continue using Xcode 15 for now to avoid those warnings but still see any other warnings that might pop up.

Here are the warnings we're seeing:
image

We should definitely address them sooner or later, but this PR is already big enough so I didn't want to add them here.

fire-at-will: Same applies to the `pod-lib-lint` job too! https://github.com/RevenueCat/purchases-ios/pull/4295#discussion_r1782936624

fire-at-will: I also just added comments here explaining why they're using Xcode 15.2.0 :D

MarkVillacampa (approved):

fire-at-will:
@RCGitBot please test

PR #4296, opened 2024-09-24
Generating new test snapshots for `update-default-machine-xcode-16-will` - ios-18

RCGitBot:
Requested by @fire-at-will for [update-default-machine-xcode-16-will](https://github.com/RevenueCat/purchases-ios/tree/update-default-machine-xcode-16-will)


fire-at-will:
Ran this so that we can get baseline snapshots for iOS18 tests

PR #4321, opened 2024-09-27
Added scaffolding for paywall components, view models, and views

joshdholtz:
### Motivation

Added all the structs needed for packages

### Description

- `PackageGroupComponent`
- `PackageComponent`
- `PurchaseButtonComponent`

### DIFF
--- a/RevenueCat.xcodeproj/project.pbxproj
+++ b/RevenueCat.xcodeproj/project.pbxproj
@@ -16,6 +16,15 @@
 		1E99F81F2AC5917F0023E26E /* StoreMessagesHelperTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1E99F81D2AC5917F0023E26E /* StoreMessagesHelperTests.swift */; };
 		2C0B98CD2797070B00C5874F /* PromotionalOffer.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C0B98CC2797070B00C5874F /* PromotionalOffer.swift */; };
 		2C2AEB0F2CA64E0E00A50F38 /* Template1Preview.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C2AEB0E2CA64E0E00A50F38 /* Template1Preview.swift */; };
+		2C2AEB3B2CA7209F00A50F38 /* PaywallPackageComponent.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C2AEB3A2CA7209F00A50F38 /* PaywallPackageComponent.swift */; };
+		2C2AEB3D2CA720B700A50F38 /* PaywallPackageGroupComponent.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C2AEB3C2CA720B700A50F38 /* PaywallPackageGroupComponent.swift */; };
+		2C2AEB3F2CA7235300A50F38 /* PaywallPurchaseButtonComponent.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C2AEB3E2CA7235300A50F38 /* PaywallPurchaseButtonComponent.swift */; };
+		2C2AEB442CA72D9F00A50F38 /* PackageGroupComponentViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C2AEB432CA72D9F00A50F38 /* PackageGroupComponentViewModel.swift */; };
+		2C2AEB462CA7302400A50F38 /* PackageComponentViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C2AEB452CA7302400A50F38 /* PackageComponentViewModel.swift */; };
+		2C2AEB482CA7304900A50F38 /* PurchaseButtonComponentViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C2AEB472CA7304900A50F38 /* PurchaseButtonComponentViewModel.swift */; };
+		2C2AEB4A2CA7338A00A50F38 /* PackageGroupComponentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C2AEB492CA7338A00A50F38 /* PackageGroupComponentView.swift */; };
+		2C2AEB4C2CA7339200A50F38 /* PackageComponentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C2AEB4B2CA7339200A50F38 /* PackageComponentView.swift */; };
+		2C2AEB4E2CA7339E00A50F38 /* PurchaseButtonComponentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C2AEB4D2CA7339E00A50F38 /* PurchaseButtonComponentView.swift */; };
 		2C4C36132C6FBA8B00AE959B /* CompatibilityTopBarTrailing.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C4C36122C6FBA8B00AE959B /* CompatibilityTopBarTrailing.swift */; };
 		2C6CC1162B8D2B6900432E4D /* PurchasesSyncAttributesAndOfferingsIfNeededTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C6CC1152B8D2B6800432E4D /* PurchasesSyncAttributesAndOfferingsIfNeededTests.swift */; };
 		2C7F0AD32B8EEB4600381179 /* RateLimiter.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2C7F0AD22B8EEB4600381179 /* RateLimiter.swift */; };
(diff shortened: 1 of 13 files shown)


RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -2257,6 +2275,33 @@
...
+				2C2AEB452CA7302400A50F38 /* PackageComponentViewModel.swift */,
+			);
+			path = PackageGroup;
+			sourceTree = "<group>";
+		};
+		2C2AEB412CA72A3600A50F38 /* Package */ = {
+			isa = PBXGroup;
+			children = (
+				2C2AEB492CA7338A00A50F38 /* PackageGroupComponentView.swift */,
+				2C2AEB432CA72D9F00A50F38 /* PackageGroupComponentViewModel.swift */,
```
JayShortway, line 2291: Are these in the right place? Looks like `Package` contains `PackageGroup*` things, and vice versa.

Sources/Paywalls/Components/PaywallPackageComponent.swift
```diff
@@ -0,0 +1,41 @@
...
+        public let components: [PaywallComponent]
+
+        public init(type: ComponentType,
+                    packageID: String,
+                    isButton: Bool = true,
+                    components: [PaywallComponent]
+        ) {
+            self.type = .package
+            self.packageID = packageID
+            self.components = components
```
JayShortway, line 35: Just for my understanding: these child components are things like text, which can contain various details of the package such as price? 

Should they be somehow limited to certain types, so people don't add something like a PurchaseButton inside a Package? Or do we want to allow for maximum flexibility?

Sources/Paywalls/Components/PaywallPackageGroupComponent.swift
```diff
@@ -0,0 +1,37 @@
...
+    struct PackageGroupComponent: PaywallComponentBase {
+
+        let type: ComponentType
+        public let defaultSelectedPackageID: String
+        public let components: [PaywallComponent]
+
+        public init(type: ComponentType, defaultSelectedPackageID: String, components: [PaywallComponent]) {
+            self.type = .packageGroup
+            self.defaultSelectedPackageID = defaultSelectedPackageID
+            self.components = components
```
JayShortway, line 31: Similar question here: should these be limited to, say, only `PackageComponent`s?

Sources/Paywalls/Components/PaywallPurchaseButtonComponent.swift
```diff
@@ -14,17 +14,41 @@ public extension PaywallComponent {
 
     struct PurchaseButtonComponent: PaywallComponentBase {
 
-        let type: String
-        let cta: PaywallComponent.LocalizationKey
-        let ctaIntroOffer: PaywallComponent.LocalizationKey?
+        let type: ComponentType
+        public let ctaLid: LocalizationKey
+        public let ctaIntroOfferLid: LocalizationKey
```
JayShortway, line 19: Just a suggestion, not a blocker by any means, but imo these don't _need_ to have the `*Lid` suffix. The fact that it is typed as `LocalizationKey` already tells me it's a localization thing. 

```diff
-        public let ctaLid: LocalizationKey
-        public let ctaIntroOfferLid: LocalizationKey
+        public let cta: LocalizationKey
+        public let ctaIntroOffer: LocalizationKey
```

RevenueCat.xcodeproj/project.pbxproj
JayShortway: Ah no, I can see in the GitHub file tree that they're not in the right place 😄

JayShortway (changes requested):
Looks good to me! Just the folder name mixup is all 😄

RevenueCat.xcodeproj/project.pbxproj
JayShortway: I see it's fixed in https://github.com/RevenueCat/purchases-ios/pull/4324. Nvm!

JayShortway (approved):
Just a suggestion and some questions, but looks good to me!

Sources/Paywalls/Components/PaywallPackageComponent.swift
```diff
@@ -0,0 +1,41 @@
...
+
+    struct PackageComponent: PaywallComponentBase {
+
+        let type: ComponentType
+        public let packageID: String
+        public let components: [PaywallComponent]
+
+        public init(type: ComponentType,
+                    packageID: String,
+                    isButton: Bool = true,
```
tonidero, line 30: Hmm I'm guessing this will be used later on, and it's meant to make this "clickable"? If so, maybe rename to `isClickable` or `isTappable`?

Sources/Paywalls/Components/PaywallPackageComponent.swift
tonidero: Yeah I agree... Maybe we can define protocols for these different types and only accept those types of protocols? My main concern with that is we won't have compile safety inside when trying to distinguish between the different subtypes of the protocol...

Sources/Paywalls/Components/PaywallPurchaseButtonComponent.swift
```diff
@@ -14,17 +14,41 @@ public extension PaywallComponent {
...
+        public let fontFamily: String?
+        public let fontWeight: FontWeight
+        public let color: ColorInfo
+        public let textStyle: TextStyle
+        public let horizontalAlignment: HorizontalAlignment
+        public let backgroundColor: ColorInfo?
+        public let padding: Padding
+        public let margin: Padding
```
tonidero, line 27: I'm not up to date with the product specs so please ignore this comment unless I say something useful 😵.  I was wondering if we were planning to support some features like:
- Gradient colors
- Background images
- Shadow/elevation (in android)
- Leading/trailing icons
- ...

Maybe they are planned for later, which is fine, just asking.

RevenueCatUI/Templates/Components/PackageGroup/PackageComponentViewModel.swift
```diff
@@ -0,0 +1,35 @@
...
+
+import Foundation
+import RevenueCat
+
+// swiftlint:disable missing_docs
+
+#if PAYWALL_COMPONENTS
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+public class PackageComponentViewModel {
```
tonidero, line 22: You might have planned for this so FFTI me but I was thinking we are currently creating view models before we verify the package id is valid and the products are found in the store. In case the package/product isn't found, we would probably need to remove this view model and view from the tree somehow... 

I think we might want to have a processing step before creating these view models/views to make sure they are valid before adding them to the view tree.

Again, not a blocker for this PR, just a thought.

tonidero (commented):
Looking great! Left some rambling thoughts but please feel free to ignore them for now. Only thing is I think we should make the view models internal (unless I missed the reason to keep them public)

RevenueCatUI/Templates/Components/PackageGroup/PackageComponentViewModel.swift
tonidero: Oh I see in the other PRs that you are failing if the package can't be found... personally I feel we should just log and continue displaying the paywall (as long as there is at least a valid package). But lmk what you think!

Sources/Paywalls/Components/PaywallPackageComponent.swift
tonidero: I see this was removed later. Ignore.

Sources/Paywalls/Components/PaywallPackageComponent.swift
joshdholtz: > Should they be somehow limited to certain types, so people don't add something like a PurchaseButton inside a Package? Or do we want to allow for maximum flexibility?

This type of validation will happen in both the frontend and the backend! Eventually 😇 There is no validation there yet but as long as the frontend and backend have the schema with these rules, the SDK should be good 💪

Sources/Paywalls/Components/PaywallPackageGroupComponent.swift
joshdholtz: This change _does_ actually happen in #4332 😊 Everywhere else I think its fine to have `[PaywallComponent]` because it will be validated with schemas in the FE and BE but here, it is important for `Offering` validation that a `PackageGroupComponent` only has `[PackageComponent]`

Glad you caught this here though 😇

Sources/Paywalls/Components/PaywallPurchaseButtonComponent.swift
joshdholtz: I _do_ agree that here it is fine but I think we will still want to have `_lid` in the JSON otherwise its super unclear 🤷‍♂️ 

@JayShortway  We can make this change and then just map `cta_lid` to `cta` in this object 💪 Thoughts?

joshdholtz:
> Only thing is I think we should make the view models internal (unless I missed the reason to keep them public)

@tonidero I totally agree! I'm actually not sure why they were public to begin with 😬 I started making this change last night in #4346 when I realized that so I will continue doing that 💪

RevenueCatUI/Templates/Components/Package/PackageGroupComponentViewModel.swift
```diff
@@ -0,0 +1,35 @@
...
+
+import Foundation
+import RevenueCat
+
+// swiftlint:disable missing_docs
+
+#if PAYWALL_COMPONENTS
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+public class PackageGroupComponentViewModel {
```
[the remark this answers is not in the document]
fire-at-will: I think it can probably be public since it's behind the PAYWALL_COMPONENTS compiler flag. If the flag didn't exist, I'd say that it would need to be `internal` until the feature is complete, but devs won't see this unless they enable the `PAYWALL_COMPONENTS` flag

fire-at-will (approved):

RevenueCatUI/Templates/Components/Package/PackageGroupComponentViewModel.swift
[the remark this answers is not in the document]
tonidero: Right, but we should try to make it internal from the beginning (unless there is a reason to keep them public) so we don't forget :)

Sources/Paywalls/Components/PaywallPurchaseButtonComponent.swift
joshdholtz: Yeah yeah, lots of these are planned for later 😁

Sources/Paywalls/Components/PaywallPurchaseButtonComponent.swift
JayShortway: Oh right, I'm still not used to Swift's magic JSON (de)serialization. In that case I actually don't mind as much haha. Might be good to directly map the JSON with this object. (I agree that we need `_lid` in the JSON.)

PR #4324, opened 2024-09-30
Render packages, package, and purchase button views for paywall components

joshdholtz:
## Motivation

Get basic packages and package views rendering

## Description

Render views for:
- `PackageGroupComponent`
  - Renders as `StackViewComponent` but will have behavior for package selection
- `PackageComponent`
  - Renders as `StackViewComponent` but will have behavior for package selection
- `PurchaseButton`
  - Renders similar to `TextComponent` but as button and trigger purchase from select package ☝️  

### SwiftUI Previews

This PR adds render functionality for the view but there are updated SwiftUI previews in a PR further down the chain with updates. See https://github.com/RevenueCat/purchases-ios/pull/4332


RevenueCatUI/Templates/Components/Stackable+Extensions.swift
```diff
@@ -0,0 +1,44 @@
...
+
+import Foundation
+import RevenueCat
+
+#if PAYWALL_COMPONENTS
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+extension PaywallComponent.StackableComponent {
+
+    func toStackComponentViewModel(
```
JayShortway, line 22: Nice!

Sources/Paywalls/Components/Common/Dimension.swift
```diff
@@ -0,0 +1,89 @@
...
+
+#if PAYWALL_COMPONENTS
+
+public extension PaywallComponent {
+
+    enum Dimension: Codable, Sendable, Hashable {
+
+        case vertical(HorizontalAlignment)
+        case horizontal(VerticalAlignment)
+        case zlayer(TwoDimensionAlignment)
```
JayShortway, line 25: This could also be something like `depth`? Since this is public, feels slightly inconsistent to use a `z` here, but not use an `x` and `y` in the other cases.

Sources/Paywalls/Components/Common/PaywallComponentPropertyTypes.swift
```diff
@@ -55,6 +55,13 @@ public extension PaywallComponent {
 
     }
 
+    enum Shape: Codable, Sendable, Hashable, Equatable {
+
+        case rectangle
+        case pill
```
JayShortway, line 61: I see that `Shape` and `CornerRadiuses` are used together, but as separate things. I feel like corner radiuses are a property of the shape. For instance, a `pill` is a `rectangle` with 50% rounded corners.  

Maybe we can give `rectangle` a `CornerRadiuses` property, and remove `pill`? Or maybe we don't need `Shape` at all, if all shapes are variations in corner radiuses.

JayShortway (commented):
Good stuff! Happy to see some previews! Just some minor comments and suggestions.

Sources/Paywalls/Components/Common/PaywallComponentPropertyTypes.swift
JayShortway: Or the new `RadiusInfo` from https://github.com/RevenueCat/purchases-ios/pull/4328 😅

tonidero (commented):
Looking good! Just Joops coment about the shape and also same comment from my side about making view models in `RevenueCatUI` internal.

RevenueCatUI/Templates/Components/Packages/PackageGroup/PackageGroupComponentViewModel.swift
```diff
@@ -23,11 +23,21 @@ public class PackageGroupComponentViewModel {
...
     private let localizedStrings: PaywallComponent.LocalizationDictionary
     private let component: PaywallComponent.PackageGroupComponent
+    private let offering: Offering
+
+    let stackComponentViewModel: StackComponentViewModel
 
     init(localizedStrings: PaywallComponent.LocalizationDictionary,
-         component: PaywallComponent.PackageGroupComponent) throws {
+         component: PaywallComponent.PackageGroupComponent,
+         offering: Offering) throws {
```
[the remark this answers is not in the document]
joshdholtz: We will allow different groups soon! A multi-tier paywall would be two `PackageGroupComponent`s for the same offering but except they specify a different packages in them

Sources/Paywalls/Components/Common/Dimension.swift
joshdholtz: Ahhhh yes! I was really not liking the name of that one 😅 Depth could totally work 💪

RevenueCatUI/Templates/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
```diff
@@ -0,0 +1,142 @@
...
+
+    private let viewModel: PurchaseButtonComponentViewModel
+
+    internal init(viewModel: PurchaseButtonComponentViewModel) {
+        self.viewModel = viewModel
+    }
+
+    var body: some View {
+        Button {
+            // WIP: Need to perform purchase logic
```
fire-at-will, line 31: I know of a pretty cool open-source SDK that can help out with this 😉

RevenueCatUI/Templates/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
```diff
@@ -0,0 +1,142 @@
...
+        } label: {
+            // WIP: Need to add logic for intro offer
```
fire-at-will, line 33: Let's collab on this in the future to add support for win-back offers too!

RevenueCatUI/Templates/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
```diff
@@ -0,0 +1,142 @@
...
+                .fontWeight(viewModel.fontWeight)
+                .fixedSize(horizontal: false, vertical: true)
+                .multilineTextAlignment(viewModel.horizontalAlignment)
+                .foregroundStyle(viewModel.color)
+                .padding(viewModel.padding)
+                .background(viewModel.backgroundColor)
+                .shape(viewModel.clipShape)
+                .applyIfLet(viewModel.cornerRadiuses, apply: { view, value in
+                    view
+                    .roundedCorner(value.topLeading, corners: .topLeft)
```
fire-at-will, line 45: Nit: Instead of applying four view modifiers here, could we instead do

```swift
.clipShape(    // could still use .mask()
   UnevenRoundedRectangle(
      topLeadingRadius: viewModel.topLeading,
      bottomLeadingRadius: viewModel.bottomLeading,
      bottomTrailingRadius: viewModel.bottomTrailing,
      topTrailingRadius: viewModel.topTrailing
   )
)
```

The performance difference is probably negligible, but somehow it feels better than applying 4 separate view modifiers

RevenueCatUI/Templates/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
```diff
@@ -0,0 +1,142 @@
...
+        self.viewModel = viewModel
+    }
+
+    var body: some View {
+        Button {
+            // WIP: Need to perform purchase logic
+        } label: {
+            // WIP: Need to add logic for intro offer
+            Text(viewModel.cta)
+                .font(viewModel.textStyle)
```
fire-at-will, line 35: I'd consider making two changes here for code reusability:
- Extract all of these modifiers to a shared view style
- Make them be view modifiers on the button itself, not the text, so that it can be reused for other buttons as well if needed in the future.

I'd probably make this be a custom `ButtonStyle` named something like `CTAButtonStyle` so it can be applied to other buttons consistently, like what Paul Hudson does here: https://www.hackingwithswift.com/quick-start/swiftui/customizing-button-with-buttonstyle

Then, for future buttons, we could do something like:

```swift
Button {
    // Do something
} label: {
    Text("Look at me, I'm a CTA Button!")
}
.buttonStyle(CTAButtonStyle())
```

Then, to inject the style values from the view model, you could make `CTAButtonStyle` accept a struct that takes in all the values it needs:

```swift
struct CTAButtonStyleConfig {
    let font: Font
    let fontWeight: FontWight
    // ...

    init(
        font: Font = .body,
        fontWeight: FontWeight = .regular,
        // ...
    )
}

Button {
    // Do something
} label: {
    Text("Look at me, I'm a CTA Button!")
}
.buttonStyle(CTAButtonStyle(
    .init(
        font: viewModel.font,
        fontWeight: viewModel.fontWeight,
        // ...
    )
))
```

This might be over-abstracting a bit, but I know one of my biggest pain points in Pines has been getting consistent label/button styling across the app. It can be a real pain to copy/paste all of those view modifiers everywhere!

RevenueCatUI/Templates/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
```diff
@@ -0,0 +1,142 @@
...
+                    .roundedCorner(value.bottomLeading, corners: .bottomLeft)
+                    .roundedCorner(value.bottomTrailing, corners: .bottomRight)
+                })
+                .padding(viewModel.margin)
+        }
+    }
+
+}
+
+private struct ShapeModifier: ViewModifier {
```
fire-at-will, line 56: If you go with the `CTAButtonStyle` approach from above, this could probably be moved into that style!

RevenueCatUI/Templates/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
```diff
@@ -0,0 +1,142 @@
...
+}
+
+#if DEBUG
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+struct PurchaseButtonComponentView_Previews: PreviewProvider {
+
+    static var previews: some View {
+        // Pill
+        PurchaseButtonComponentView(
```
fire-at-will, line 85: Nit: If you make `PaywallComponent.Shape` conform to `CaseIterable`, you could do a `ForEach` over its cases and get rid of the duplicated code here

fire-at-will (approved):

RevenueCatUI/Templates/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
joshdholtz: I think I'll just use StoreKit 2 manually 😛

RevenueCatUI/Templates/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
joshdholtz: Ah yes! This is actually changed in #4328 😁

RevenueCatUI/Templates/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
joshdholtz: Ahhh, I always forget about `.buttonStyle()` 🙈 I need to use that in my own apps!

I don't know if this _exactly_ fits here because this is the only place where this is happening but... I could maybe do something similar since since this button and text and a few others might be setting this same group of properties. 

I might hold off on doing this until I see more of pattern but I really like how much cleaner that looks 😍

RevenueCatUI/Templates/Components/Packages/PackageGroup/PackageGroupComponentViewModel.swift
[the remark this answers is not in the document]
JayShortway: We might need a diagram 😅 Thanks for explaining!

PR #4328, opened 2024-09-30
Added new individual corner radius and border modifier

joshdholtz:
## Motivation

We need borders for package selection states 

## Description

Replaced `.roundedCorner()` view modifier (which was from Paywalls V1) with new `.borderCorner()` view modifier 
- Allows individual clip radius for each corner
- Allows individual border radius for each corner (border is an overlay)

Also added a bunch of SwiftUI previews (that will be used as snapshots once we enable Emerge)

### SwiftUI Previews

| Equal Radius - No Border | No Radius - Blue Border | Two Round - Blue Border | Equal Radius - Blue Border | Two Round - No Border |
|--------|--------|--------|--------|--------|
| Screenshot 2024-10-02 at 3 00 34 PM | Screenshot 2024-10-02 at 3 01 04 PM | Screenshot 2024-10-02 at 3 01 43 PM | Screenshot 2024-10-02 at 3 01 23 PM | Screenshot 2024-10-02 at 3 01 11 PM |


RevenueCatUI/Templates/Components/CornerBorder.swift
```diff
@@ -0,0 +1,301 @@
...
+        let width: CGFloat
+
+        init(color: Color, width: Double) {
+            self.color = color
+            self.width = width
+        }
+
+    }
+
+    struct RaidusInfo {
```
JayShortway, line 34: Typo 🙂 
```diff
-    struct RaidusInfo {
+    struct RadiusInfo {
```

RevenueCatUI/Templates/Components/CornerBorder.swift
```diff
@@ -0,0 +1,301 @@
...
+
+        let topLeft: CGFloat?
+        let topRight: CGFloat?
+        let bottomLeft: CGFloat?
+        let bottomRight: CGFloat?
```
JayShortway, line 39: How does this work with RTL languages? Not sure how SwiftUI handles that in general, but should we rename them like so? 

```diff
-        let topLeft: CGFloat?
-        let topRight: CGFloat?
-        let bottomLeft: CGFloat?
-        let bottomRight: CGFloat?
+        let topStart: CGFloat?
+        let topEnd: CGFloat?
+        let bottomStart: CGFloat?
+        let bottomEnd: CGFloat?
```

RevenueCatUI/Templates/Components/CornerBorder.swift
```diff
@@ -0,0 +1,301 @@
...
+
+        // Bottom edge and bottom-left corner
+        path.addLine(to: CGPoint(x: rect.minX + bottomLeft, y: rect.maxY))
+        path.addQuadCurve(to: CGPoint(x: rect.minX, y: rect.maxY - bottomLeft),
+                          control: CGPoint(x: rect.minX, y: rect.maxY))
+
+        // Left edge and top-left corner
+        path.addLine(to: CGPoint(x: rect.minX, y: rect.minY + topLeft))
+        path.addQuadCurve(to: CGPoint(x: rect.minX + topLeft, y: rect.minY),
+                          control: CGPoint(x: rect.minX, y: rect.minY))
```
JayShortway, line 185: Beautiful! 😚👌

RevenueCatUI/Templates/Components/CornerBorder.swift
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    // swiftlint:disable:next function_parameter_count
+    func conditionalOverlay(
+        color: Color?,
+        width: CGFloat?,
+        topLeft: CGFloat?,
+        topRight: CGFloat?,
+        bottomLeft: CGFloat?,
+        bottomRight: CGFloat?
```
JayShortway, line 111: Not sure if this makes sense, but we could also make these non-nil and default to 0? That way you can apply rounding to some corners, without having to specify all corners.

Sources/Paywalls/Components/Common/Border.swift
```diff
@@ -0,0 +1,30 @@
...
+//  Created by Josh Holtz on 9/30/24.
+// swiftlint:disable missing_docs
+
+import Foundation
+
+#if PAYWALL_COMPONENTS
+
+public extension PaywallComponent {
+
+    struct Border: Codable, Sendable, Hashable {
```
JayShortway, line 21: Just for my understanding: why is this an extension on `PaywallComponent`?

RevenueCatUI/Templates/Components/CornerBorder.swift
```diff
@@ -0,0 +1,301 @@
...
+            self.topLeft = topLeft.flatMap { CGFloat($0) }
+            self.topRight = topRight.flatMap { CGFloat($0) }
+            self.bottomLeft = bottomLeft.flatMap { CGFloat($0) }
+            self.bottomRight = bottomRight.flatMap { CGFloat($0) }
+        }
+
+    }
+
+    var border: BorderInfo?
+    var radiuses: RaidusInfo?
```
JayShortway, line 51: Is my understanding correct that we now have 2 ways to define corner radiuses, `RadiusInfo` and `CornerRadiuses`? Is that necessary, or can we consolidate the 2? I think it would be great if we could have a single concept of "corner radiuses".

JayShortway (commented):
Absolutely beautiful! Had a few questions and suggestions.

RevenueCatUI/Templates/Components/CornerBorder.swift
joshdholtz: This is how I leave my mark everywhere I go but... I'll change it just for you 🫶

RevenueCatUI/Templates/Components/CornerBorder.swift
joshdholtz: I should put a todo in here but we would have to manually flip it in this case for RTL languages since there isn't native support for this. Thanks for catching!

Sources/Paywalls/Components/Common/Border.swift
joshdholtz: Purely namespacing reasons 😇  If a developer does `import RevenueCatUI`, the `Border` would be directly available in the file and could cause some weird naming conflicts. So anything that is public and very generally named, we are trying to put under this namespace so its under `PaywallComponent.Border` 🤷‍♂️

RevenueCatUI/Templates/Components/CornerBorder.swift
joshdholtz: Yeaaaahhh, this is kind of confusing 🤔 I'll take another look to see if they can combined in a sensible way 💪

RevenueCatUI/Templates/Components/CornerBorder.swift
JayShortway: Haha I'm honored ❤️

Sources/Paywalls/Components/Common/Border.swift
JayShortway: Oh right, namespacing. Makes sense, thanks!

historicalstepscounter:
I cant add a package to my new project last 30 minutes. 5.6.0. It freezes on "Starting package resolution"

PR #4332, opened 2024-10-01
Select package and purchase

joshdholtz:
## Follow up PRs will be
- [ ] Reworking containers where every container is a "stack"
  - This will include an improvement for passing a `Package` into `PackageComponentViewModel`
- [ ] Selection state
- [ ] Introductory offers
- [ ] Interpolating variables 

## Motivation

Select a package and purchase 

⚠️ There is **no selection state** in the UI yet

## Description

- Adds new `PaywallState` observable object 
- Injects `PaywallState` as an environment object for:
  - `PackageGroupComponentView` to set the default package as selected package
  - `PackageComponentView` to select the package when tapped
  - `PurchaseButtonComponentView` to purchase

### SwiftUI Previews

| Package Group | Package | Purchase Button (Pill) | Purchase Button (Rectangle) |
|--------|--------|--------|--------|
| Screenshot 2024-10-02 at 2 38 25 PM | Screenshot 2024-10-02 at 2 47 22 PM | Screenshot 2024-10-02 at 2 49 06 PM | Screenshot 2024-10-02 at 2 49 11 PM |

### Demo

Here is a (not pretty) paywall demo with purchases. It is missing selection state (coming in a separate PR) . However, if you look at the logs, you can see the default select package is purchases first and then tapping on the packages chagnes the one being purchased.

Purchases do not work in the sample because its a hard coded paywall in the Paywalls Tester app 🤷‍♂️ 

https://github.com/user-attachments/assets/ec5f8328-7110-43ff-873d-449698a8de24


RevenueCatUI/Templates/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
```diff
@@ -20,15 +20,32 @@ import SwiftUI
...
+        AsyncButton {
+            guard !self.purchaseHandler.actionInProgress else { return }
+
+            // WIP: Need to log warning if currently subscribed
+            guard let selectedPackage = self.paywallState.selectedPackage
+//                    , selectedPackage.currentlySubscribed
+            else {
+                Logger.warning(Strings.product_already_subscribed)
+                return
+            }
```
joshdholtz, line 45: This is logic that was implemented in a very different way in Paywalls V1. This will be done in a new further PR 💪

RevenueCatUI/Templates/Components/Packages/PackageGroup/PackageGroupComponentView.swift
```diff
@@ -20,11 +20,16 @@ import SwiftUI
...
+    private var paywallState: PaywallState
+
     let viewModel: PackageGroupComponentViewModel
 
     var body: some View {
-        // WIP: Do something with default package id and selection
         StackComponentView(viewModel: self.viewModel.stackComponentViewModel)
+            .onAppear {
+                self.paywallState.select(package: self.viewModel.defaultPackage)
+            }
```
JayShortway, line 32: Just checking, but can we trust `onAppear`? E.g. what happens if the user selects a non-default package, clicks a link (e.g. to read the privacy policy), and comes back? Is the non-default package still selected, or does `onAppear` fire again?

RevenueCatUI/Templates/Components/Packages/PackageGroup/PackageGroupComponentViewModel.swift
```diff
@@ -34,12 +36,57 @@ public class PackageGroupComponentViewModel {
...
+                return (component: packageComponent, package: pkg)
+            } else {
+                return nil
+            }
+        }
+
+        // We need packages
+        guard let firstPackage = availablePackageInfos.first else {
+            Logger.error(Strings.paywall_could_not_find_any_packages)
+            throw PackageGroupValidationError.noAvailablePackages("No available packages found")
```
JayShortway, line 67: Is this an error scenario where the packages configured for this component are not available in the offering? If so, maybe the error message can reflect that? Something like: 

> Please make sure the packages configured for this package group are available in the Offering. 
  
(Can probably be improved with more context haha.)

RevenueCatUI/Templates/Components/Packages/Package/PackageComponentView.swift
```diff
@@ -20,11 +20,18 @@ import SwiftUI
...
+    @EnvironmentObject
+    private var paywallState: PaywallState
+
     let viewModel: PackageComponentViewModel
 
     var body: some View {
         // WIP: Do something with package id and selection
-        StackComponentView(viewModel: self.viewModel.stackComponentViewModel)
+        Button {
+            self.paywallState.select(package: self.viewModel.package)
```
JayShortway, line 31: I know SwiftUI has different recommended patterns from Jetpack Compose, but in Compose land the use of `CompositionLocal`s ( = `@EnvironmentObject`s) is discouraged, because it introduces hidden dependencies to the views. As an alternative, this view could take an `onPackageSelected` closure, which gets called by the button. 

Not a blocker, but just wanted to mention it. 😇

Sources/Paywalls/Components/PaywallPackageGroupComponent.swift
```diff
@@ -24,7 +24,11 @@ public extension PaywallComponent {
         let type: ComponentType
         public let defaultSelectedPackageID: String
 
-        public let components: [PaywallComponent]
+        // Most containers will accept an array of `PaywallComponent`s
+        // Package group specifically accepts `PackageComponent` for
+        // validation that packages exist
+        public let packages: [PaywallComponent.PackageComponent]
```
JayShortway, line 30: 👍

Tests/TestingApps/PaywallsTester/PaywallsTester.xcodeproj/project.pbxproj
```diff
@@ -637,7 +637,7 @@
 				"LD_RUNPATH_SEARCH_PATHS[sdk=macosx*]" = "@executable_path/../Frameworks";
 				MACOSX_DEPLOYMENT_TARGET = 13.0;
 				MARKETING_VERSION = 1.0;
-				OTHER_SWIFT_FLAGS = "";
+				OTHER_SWIFT_FLAGS = "-DPAYWALL_COMPONENTS";
```
JayShortway, line 640: Oopsie?

Tests/TestingApps/PaywallsTester/PaywallsTester.xcodeproj/project.pbxproj
```diff
@@ -687,7 +687,7 @@
...
```
JayShortway, line 690: Another oopsie? (Use the new `Local.xcconfig` and this never happens again. 😛)

JayShortway (commented):
Let's gooo! Few comments, suggestions and questions, but really close!

RevenueCatUI/Templates/Components/Packages/Package/PackageComponentViewModel.swift
```diff
@@ -34,12 +35,32 @@ public class PackageComponentViewModel {
...
+            components: self.component.components,
             localizedStrings: localizedStrings,
             offering: offering
         )
     }
 
+    static func findPackage(identifier: String, offering: Offering) throws -> Package {
+        guard let package = offering.package(identifier: identifier) else {
+            Logger.error(Strings.paywall_could_not_find_package(identifier))
+            throw PackageValidationError.missingPackage(
```
tonidero, line 50: I think this should never happen with the validations in the `PackageGroupComponentViewModel` right? Maybe we can pass the package insteead of the offering to the constructor here to avoid having to do this validation?

RevenueCatUI/Templates/Components/Packages/PackageGroup/PackageGroupComponentViewModel.swift
tonidero: It may also happen if the products are not found in Apple and so on... Basically the "No offerings found" error all over again 😅. But yeah, maybe we can have a link with more information about the reasons for this error.

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
```diff
@@ -4,18 +4,32 @@
...
+    @Published var selectedPackage: Package?
+
+    func select(package: Package) {
+        self.selectedPackage = package
+    }
+
+}
+
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 public struct TemplateComponentsView: View {
```
tonidero, line 25: Noticed this was public, since it's behind the `PAYWALL_COMPONENTS` flag, maybe we can also make it internal now?

RevenueCatUI/Templates/Components/Packages/PackageGroup/PackageGroupComponentViewModel.swift
JayShortway: Oh right, sad. 😢 Yea maybe for now we can link to Sharif's community post? And when we do the Offerings-are-empty project we make sure we cover this path?

tonidero: Yup I think we should do that 👍

RevenueCatUI/Templates/Components/Packages/Package/PackageComponentViewModel.swift
joshdholtz: Oh good point! I think I added the logic in `PackageGroupComponentViewModel` after this 😅 I will make that change 💪

RevenueCatUI/Templates/Components/Packages/PackageGroup/PackageGroupComponentViewModel.swift
joshdholtz: > It may also happen if the products are not found in Apple and so on

Yeah yeah, this is the main use case this is for ☝️ 😬

RevenueCatUI/Templates/Components/Packages/Package/PackageComponentViewModel.swift
joshdholtz: On second thought... going to save this to a follow up PR where we are reworking how every container is a stack because... this is not so easy to do in its current implementation 😬 Right now, view models are _pretty_ independent of each other but that will probably change in the follow up PR. So going to make a task for me to fix this later (later being tomorrow) 💪

RevenueCatUI/Templates/Components/Packages/PackageGroup/PackageGroupComponentView.swift
joshdholtz: The non-default should still be selected! `onAppear` only gets called one (its poorly named 😛)

RevenueCatUI/Templates/Components/Packages/Package/PackageComponentView.swift
joshdholtz: This is a completely fair question 😅 This `PaywallState` _could_ get passed in as both a `Binding<PaywallState>` and an `onPackageSelected` closure to the views. However, that kind of introduces a giant list of parameters that would need to get updated and passed into all these view recursively. I choose to to down the `@EnvironmentObject` route for the reason of simplifying the construction of all the views and any unknown callbacks or state that we know right now.

I _do_ agree that if this gets used too much and in a complex way that its unclear if that object will exist or not. However, in our case, the `PaywallState` gets injected _very_ much at the entry point of paywalls in `TemplateComponentsView` so any views are guaranteed to have it 🤷‍♂️

JayShortway: Alright, yea this seems like a reasonable trade-off indeed!

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
```diff
@@ -4,19 +4,33 @@
...
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
-public struct TemplateComponentsView: View {
+struct TemplateComponentsView: View {
 
     let paywallComponentsData: PaywallComponentsData
     let componentViewModels: [PaywallComponentViewModel]
     private let onDismiss: () -> Void
 
+    @StateObject
+    private var paywwallState = PaywallState()
```
tonidero, line 32: Typo
```suggestion
    private var paywallState = PaywallState()
```

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
```diff
@@ -51,6 +63,7 @@ public struct TemplateComponentsView: View {
         }
         .frame(maxHeight: .infinity, alignment: .topLeading)
         .edgesIgnoringSafeArea(.top)
+        .environmentObject(self.paywwallState)
```
tonidero, line 66:
```suggestion
        .environmentObject(self.paywallState)
```

tonidero (approved):
LGTM!

Sources/Paywalls/Components/Common/StackableComponent.swift
```diff
@@ -20,7 +20,6 @@ public extension PaywallComponent {
 
     protocol StackableComponent {
 
-        var components: [PaywallComponent] { get }
```
JayShortway, line 23: A `StackableComponent` without child `components` seems odd, but I guess the entire `StackableComponent` is gonna go, right?

JayShortway (approved):
This all looks good to me, nice job!

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
joshdholtz: I need you all to start letting me have a few of these typos... I try so hard to get them merged in 🤦‍♂️

Sources/Paywalls/Components/Common/StackableComponent.swift
joshdholtz: Yeah yeah yeah, is going to go in next PR 😇

PR #4333, opened 2024-10-01
Fixes double callbacks when using `PaywallViewController`

vegaro:
We got a report of duplicated callbacks in Flutter when using the Paywall view in iOS

We pinpointed the issue to the `update(with:` methods, that create a new controller. 

It turned out, the `didSet` function of `hostingController` was calling `loadView` when adding the `subView`, and the `PaywallViewController` ended up having two `PaywallContainerView`. Since each of these views was being set as delegates for the callbacks, Flutter ended up getting two callbacks for the `PaywallView` interactions.

After a lot of attempts, I ended up fixing it by replacing the `loadView` with `viewDidLoad` and ensuring we don't create a new controller there. `viewDidLoad` will be called from the `addSubView` but it won't recreate a new controller since it checks it already has one.

I've tested this both calling `update(with:)` and also calling `presentPaywall` and it looks like it's working.

### DIFF
--- a/RevenueCatUI/UIKit/PaywallViewController.swift
+++ b/RevenueCatUI/UIKit/PaywallViewController.swift
@@ -8,7 +8,7 @@
 //      https://opensource.org/licenses/MIT
 //
 //  PaywallViewController.swift
-//  
+//
 //  Created by Nacho Soto on 8/1/23.
 
 // swiftlint:disable file_length
@@ -137,10 +137,12 @@ public class PaywallViewController: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
-    public override func loadView() {
-        super.loadView()
+    public override func viewDidLoad() {
+        super.viewDidLoad()
 
-        self.hostingController = self.createHostingController()
+        if self.hostingController == nil {
+            self.hostingController = self.createHostingController()
+        }
     }
 
     public override func viewDidDisappear(_ animated: Bool) {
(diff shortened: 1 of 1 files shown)


RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -209,17 +211,20 @@ public class PaywallViewController: UIViewController {
...
+            oldController.view.removeFromSuperview()
+            oldController.removeFromParent()
         }
 
         didSet {
             guard let newController = self.hostingController else { return }
 
             self.addChild(newController)
+
+            self.view.subviews.forEach { $0.removeFromSuperview() }
```
tonidero, line 226: Just to check but this is still needed right? As in, it may still happen that addding the controller triggers a new set of the controller?

tonidero (approved):
Just a couple questions but I think it makes sense! As well, might be good to let an iOS dev double check

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -137,10 +137,12 @@ public class PaywallViewController: UIViewController {
...
 
-    public override func loadView() {
-        super.loadView()
+    public override func viewDidLoad() {
+        super.viewDidLoad()
 
-        self.hostingController = self.createHostingController()
+        if self.hostingController == nil {
+            self.hostingController = self.createHostingController()
+        }
```
joshdholtz, line 145: 👍

joshdholtz (approved):
This looks good to me!

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -137,10 +137,12 @@ public class PaywallViewController: UIViewController {
         fatalError("init(coder:) has not been implemented")
     }
 
-    public override func loadView() {
-        super.loadView()
+    public override func viewDidLoad() {
```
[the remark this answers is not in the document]
vegaro: tested it and it looks the same as before, I didn't notice any change

RevenueCatUI/UIKit/PaywallViewController.swift
vegaro: I added this just in case. To guarantee we are clearing before adding any subview. In any case, just this line didn't fix it, adding the subview was triggering the load view, and eventually calling `didSet`

- willSet (removes views)
- didSet
	- removes all views
	- start of addSubview
	    - loadView
			- willSet (removes views)
			- didSet
				- remove all views
				- addSubview
	- finish adding subview

At this point there are two views and the removal of all views in didSet didn't fix it.

tonidero: Makes sense thanks!

PR #4334, opened 2024-10-01
Fix Xcode 16 warnings

MarkVillacampa:
When compiling with Xcode 16, we started throwing a bunch of new warnings. Here's a list and explanation of each:

- The most common one was needing to explicitly restate the `@unchecked Sendable` conformance from a parent class.
- There were a couple switches where a new entry was added in tvOS but only for iOS 18, and needed to add it to make it exhaustive.
- There we a couple of places where we were adding a protocol conformance to a class from a different module e.g. `extension DispatchTimeInterval: Comparable`. This is considered unsafe now because the owners of the module could add the conformance later. Solved by removing the conformance which wasn't being used widely in the codebase, only in tests.
- Finally there was a class where we were capturing `SKRequest` and `SKProductsRequest` instances in a sendable block, but only to either log something, or use them as dictionary key, so I just made the classes Sendable (with the new `@retroactice` attribute because they belong to a different module)

### Checklist


fire-at-will (commented):
Looks good to me!

As a part of this PR, I'd also see if we can remove the Xcode 15.2.0 specification on the [pod-lib-lint](https://github.com/RevenueCat/purchases-ios/blob/d9e4258c65c71c190eec42cc8f1a0440d69490de/.circleci/config.yml#L455) and [create-tag](https://github.com/RevenueCat/purchases-ios/blob/d9e4258c65c71c190eec42cc8f1a0440d69490de/.circleci/config.yml#L1349) CI workflows

vegaro (approved):
thank you for looking into these!

Sources/FoundationExtensions/DispatchTimeInterval+Extensions.swift
```diff
@@ -98,15 +98,6 @@ func - (lhs: DispatchTimeInterval, rhs: DispatchTimeInterval) -> DispatchTimeInt
     return .milliseconds(lhs.milliseconds - rhs.milliseconds)
 }
 
-extension DispatchTimeInterval: Comparable {
```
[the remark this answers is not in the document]
MarkVillacampa: Nope, but was just used once outside of tests

RevenueCatUI/PaywallFontProvider.swift
```diff
@@ -49,6 +49,10 @@ open class DefaultPaywallFontProvider: PaywallFontProvider {
         case .caption: return .caption
         case .caption2: return .caption2
 
+        #if compiler(>=6.0) && os(tvOS)
+        case .caption3: return .caption3
+        #endif
```
MarkVillacampa, line 54: This got added in iOs 18 but only for tvOS

https://developer.apple.com/documentation/swiftui/font/textstyle/caption3

Also need to use `compiler()` (the version of the compiler) instead of `swift()` (the swift mode used), since we want to check we're compiling in Xcode 16, regardless of the swift mode.

Sources/Networking/Operations/PostAttributionDataOperation.swift
```diff
@@ -60,6 +60,9 @@ class PostAttributionDataOperation: NetworkOperation {
 
 }
 
+// Restating inherited @unchecked Sendable from Foundation's Operation
+extension PostAttributionDataOperation: @unchecked Sendable {}
```
NachoSoto, line 64: This will make you shoot yourself in the foot. These operations aren't actually Sendable?

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
```diff
@@ -255,3 +255,9 @@ private extension ProductsFetcherSK1 {
 // @unchecked because:
 // - It has mutable state, but it's made thread-safe through `queue`.
 extension ProductsFetcherSK1: @unchecked Sendable {}
+
+#if hasFeature(RetroactiveAttribute)
+// Conformance should be safe since it is only used as dictionary key
```
NachoSoto, line 260: I recommend creating an `UncheckedSendable<T>` wrapper so you only ignore warnings in places where it might actually be safe rather than everywhere.

PR #4336, opened 2024-10-02
Better logs for promotional offer view

vegaro:
- Improved debug logs for promotional offer view
- Updated SKConfig file to add a promotional offer
- Updated SKConfig file storefront to en_US

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
+++ b/RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
@@ -45,14 +45,14 @@ class LoadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType {
 
             guard let productIdentifier = customerInfo.earliestExpiringAppStoreEntitlement()?.productIdentifier,
                   let subscribedProduct = await self.purchasesProvider.products([productIdentifier]).first else {
-                Logger.warning(Strings.could_not_offer_for_active_subscriptions)
+                Logger.warning(Strings.could_not_offer_for_any_active_subscriptions)
                 return .failure(CustomerCenterError.couldNotFindSubscriptionInformation)
             }
 
             guard let discount = subscribedProduct.discounts.first(where: {
                 $0.offerIdentifier == promoOfferDetails.iosOfferId
             }) else {
-                Logger.warning(Strings.could_not_offer_for_active_subscriptions)
+                Logger.warning(Strings.could_not_offer_for_active_subscriptions(promoOfferDetails.iosOfferId, productIdentifier))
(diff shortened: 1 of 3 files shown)


RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
```diff
@@ -45,14 +45,14 @@ class LoadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType {
...
-                Logger.warning(Strings.could_not_offer_for_active_subscriptions)
+                Logger.warning(Strings.could_not_offer_for_any_active_subscriptions)
                 return .failure(CustomerCenterError.couldNotFindSubscriptionInformation)
             }
 
             guard let discount = subscribedProduct.discounts.first(where: {
                 $0.offerIdentifier == promoOfferDetails.iosOfferId
             }) else {
-                Logger.warning(Strings.could_not_offer_for_active_subscriptions)
+                Logger.warning(Strings.could_not_offer_for_active_subscriptions(promoOfferDetails.iosOfferId, productIdentifier))
```
tonidero, line 55: Hmm I'm not sure if these should be warnings or `debug`... seems like a case that may happen somewhat often if the customer is not eligible by apple? Not a strong opinion though

Tests/TestingApps/PaywallsTester/PaywallsTester/Products.storekit
```diff
@@ -37,8 +47,8 @@
   ],
   "settings" : {
     "_failTransactionsEnabled" : false,
-    "_locale" : "es_ES",
-    "_storefront" : "ESP",
+    "_locale" : "en_US",
+    "_storefront" : "USA",
```
tonidero, line 51: Just checking but any reason to change the storefront/locale? I'm ok with the change though.

tonidero (approved):

Tests/TestingApps/PaywallsTester/PaywallsTester/Products.storekit
vegaro: this is intentional, should've mention it in the description. This was causing the paywalls tester to have a mix of Spanish (subscription name) and English and figured we didn't want that.

RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
vegaro: > seems like a case that may happen somewhat often if the customer is not eligible by apple?

That will actually fail later, this will fail if the active product's discounts don't match with the promotional offers in the Customer Center config.

I actually think it could be a debug yes, I will change it

PR #4337, opened 2024-10-02
Feat/update swift sample apps

nyeu:
Motivation
Apps were not showcasing the latest features and were outdated. This not only helps our customers to easily understand how RevenueCat features work but it's also helpful for when the DSE team has to debug something related to paywalls.

Description
I've run the app both through a simulator and a real device using my own app projects, subscribing and seeing the expected result.


tonidero (approved):
I think it all makes sense, thanks for doing this!

PR #4339, opened 2024-10-02
Remove access to Localization env variable in ManageSubscriptionsViewModel

vegaro:
There was a warning caused by accessing the `Localization` environment variable from outside a View

```
Accessing Environment<Localization>'s value outside of being installed on a View. This will always read the default value and will not update.
```


tonidero (approved):
Ohh makes sense, thanks!

vegaro:
Ugh, I forgot to fix the test, fixing!

PR #4341, opened 2024-10-02
Allows enabling PAYWALL_COMPONENTS using a Local.xcconfig file.

JayShortway:
## What
This PR adds a `Local.xcconfig` file that can be used to enable various build settings, such as the `PAYWALL_COMPONENTS` flag or any other future flag. It's added to `.gitignore`, so these settings only affect local builds. 

This allows for a workflow similar to Gradle's `local.properties`. Personally I like this workflow, as it avoids having to find options in Xcode's GUI, which also result in changes in Xcode's project files (which we should be careful not to commit). 

Note that the original workflow of changing compiler flags in Xcode's GUI still works with this PR. It just adds another option.

## How
There's a new `Local.xcconfig.SAMPLE` file, in similar vein as the `fastlane/.env.SAMPLE` file. This file should be copied by the developer, removing the `.SAMPLE` extension. Then, any desired options can be enabled in the `Local.xcconfig` file. `Local.xcconfig` is gitignored, so enable options to your heart's content. 

I added a `Global.xcconfig` Configuration Settings file, which is referenced by the Debug and Release build configurations. Its only purpose is an optional include of `Local.xcconfig`. This allows builds to succeed if `Local.xcconfig` does not exist, such as on CI, or on developer machines who didn't create `Local.xcconfig` (yet). 

I also added a file reference to `Local.xcconfig` to the Xcode project, so it can easily be edited from within Xcode. The only downside is that it is displayed in red in the project navigator if the file doesn't exist (yet). Builds should still succeed regardless of the file existing. Let me know what you think. I can easily revert the commit that added this file reference. 

image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,3 +56,6 @@ Tests/InstallationTests/XcodeDirectInstallation/XcodeDirectInstallation.xcodepro
 # fastlane
 fastlane/.env
 .vscode/settings.json
+
+# Local build settings
+Local.xcconfig
--- a/Local.xcconfig.SAMPLE
+++ b/Local.xcconfig.SAMPLE
@@ -1,0 +1,27 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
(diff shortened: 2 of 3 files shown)


joshdholtz (approved):
This is so smart 😍

aboedo:
this is brilliant!! 💪

Issue #4335, opened 2024-10-01
CustomerCenter Dark mode colors looks off


unxavi:
- [x ] I have updated Purchases SDK to the latest version
- [ x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [ x] I have searched the [Community](https://community.revenuecat.com)
- [ x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

1. Environment
   1. Platform: iOS
   2. SDK version: 5.5.0
   3. StoreKit version:
      - [ ] StoreKit 1 (default on versions <5.0.0. Can be enabled in versions >=5.0.0 with `.with(storeKitVersion: .storeKit1)`)
      - [x] StoreKit 2 (default on versions >=5.0.0)
   4. OS version: 18.0
   5. Xcode version: 16.0
   6. Device and/or simulator:
      - [x ] Device
      - [ ] Simulator
   7. Environment:
      - [x ] Sandbox
      - [ ] TestFlight
      - [ ] Production
   8. How widespread is the issue. Percentage of devices affected.

3. Steps to reproduce, with a description of expected vs. actual behavior

Open Customer Center in dark mode, the text is next to the gray background, on light mode there are not 2 different colors for the background, so it does not look too bad

![IMG_AFD8DDBBF966-1](https://github.com/user-attachments/assets/0b431d03-eeec-412c-800f-9da999a5114f)

As additional feedback, when the user touches the offer button to accept the offer, there is a delay between the tap until something happens. First time I tested though that nothing was happening, then the offer in app purchase just keeps popping up all the time, kind of looking like a bug.

Maybe a loading should be implemented,  but for sure the button should be disable for touching multiple times, that way the in app purchase is not going to pop up multiple times.

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/46759) and notified the team. Thank you for reporting, we're checking this out!

vegaro:
Thank you for the feedback, those are all clear issues, we'll get them fixed

PR #4342, opened 2024-10-02
Improve promotional offer button when pressed

vegaro:
Reported in https://github.com/RevenueCat/purchases-ios/issues/4335

- Adds a loading to the button
- Disables the button when pressed to prevent multiple presses

https://github.com/user-attachments/assets/c0ae8189-9e5a-4e58-b209-3a1f9fdddf8b

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
@@ -35,7 +35,7 @@ struct PromotionalOfferView: View {
     private var appearance: CustomerCenterConfigData.Appearance
     @Environment(\.colorScheme)
     private var colorScheme
-    @State private var loadingState: Bool = false
+    @State private var isLoading: Bool = false
 
     init(promotionalOffer: PromotionalOffer,
          product: StoreProduct,
@@ -61,7 +61,7 @@ struct PromotionalOfferView: View {
 
                     Spacer()
 
-                    PromoOfferButtonView(loadingState: $loadingState,
+                    PromoOfferButtonView(isLoading: $isLoading,
                                          viewModel: self.viewModel,
                                          appearance: self.appearance)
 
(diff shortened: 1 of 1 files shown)


RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
```diff
@@ -131,24 +136,35 @@ struct PromoOfferButtonView: View {
...
             let localizedProductPricePerPeriod = product.localizedPricePerPeriod(.current)
 
-            Button(action: {
-                Task {
-                    await viewModel.purchasePromo()
+            AsyncButton {
+                withAnimation(.easeInOut(duration: 0.3)) {
+                    loadingState = true
+                }
+                await viewModel.purchasePromo()
```
tonidero, line 143: Just to make sure but this shouldn't trigger any errors right?

tonidero (approved):
Looks great! Thanks!

RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
```diff
@@ -35,6 +35,7 @@ struct PromotionalOfferView: View {
     private var appearance: CustomerCenterConfigData.Appearance
     @Environment(\.colorScheme)
     private var colorScheme
+    @State private var loadingState: Bool = false
```
JayShortway, line 38: Super nit suggestion for your consideration: 

```diff
-    @State private var loadingState: Bool = false
+    @State private var isLoading: Bool = false
```

RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
vegaro: we are catching the error, and not doing anything with it atm, but that's from before this change, so that improvement should come in a different PR

PR #4343, opened 2024-10-02
CAT-1726: Support Deferring Win-Back StoreKit Messages

fire-at-will:
### Motivation
iOS 18 introduces win-back offers to allow developers to extend offers to churned subscribers. In iOS 18+, StoreKit sends apps a [StoreKit Message](https://developer.apple.com/documentation/storekit/message) when a subscriber is eligible to redeem a win-back offer. We currently support showing this message automatically if the `showStoreMessagesAutomatically` configuration option is `true`, but don't yet support deferring displaying the message.

### Description
This PR adds win-back offers to the `StoreMessageType` enum so that we can support deferring StoreKit Messages where the reason is `winBackOffer`.

### Xcode Compatibility Issues
`StoreKit.Message.Reason.winBackOffer` is available on iOS 18.0+ and wasn't back ported to be available in Xcode versions <16.0. Thus, if we reference `StoreKit.Message.Reason.winBackOffer` without checking for the Xcode version, we'll break compilation on all versions of Xcode less than 16.0:

Screenshot 2024-10-02 at 1 13 04 PM

> Building the SDK when referencing `StoreKit.Message.Reason.winBackOffer` in Xcode 15.4 for a simulator running iOS 18.1 results in a build failure

Since Swift doesn't provide a way to do conditional compilation based on Xcode version, our next best option is to check the Swift compiler version, since the Swift compiler version is tied to Xcode versions. Since the Swift 6 compiler shipped with Xcode 16, this PR makes our handling of deferred win-back offer messages dependent on the Swift compiler version being >=6.0.

### Testing
- The changes were tested manually.


Sources/Support/StoreMessageType.swift
```diff
@@ -40,6 +47,14 @@ extension Message.Reason {
         case .priceIncreaseConsent: return .priceIncreaseConsent
         case .generic: return .generic
         default:
+            // winBackOffer message reason was added in iOS 18.0, but it's not recognized by xcode versions <16.0.
+            // https://developer.apple.com/documentation/xcode-release-notes/xcode-14_3-release-notes
+            #if swift(>=6.0)
```
fire-at-will, line 52: See the PR description for more details on why this is necessary

MarkVillacampa: the `swift` compilation condition defines the swift mode, whereas `compiler` defined the swift compiler version.

A lot of people will compile in Swift 5.X mode while using Xcode. Since what we actually want to do is detect the Xcode version, and Xcode 16 comes with the Swift 6.0 compiler, it's better to use the `compiler` condition instead of `swift`, as we know if ties directly the Xcode version.

https://docs.swift.org/swift-book/documentation/the-swift-programming-language/statements/#Conditional-Compilation-Block

fire-at-will: Woah, TIL! Will switch to using the `compiler` condition. Thank you!!

fire-at-will: Switched in to using the `compiler` condition in https://github.com/RevenueCat/purchases-ios/pull/4343/commits/086dce6a8a23877eeae163aeb0a6c44033409712 😎

tonidero (approved):
LGTM!

MarkVillacampa (approved):

PR #4344, opened 2024-10-02
Only Treat Deferred StoreKit Messages as Shown When They are Shown

fire-at-will:
### Motivation
Currently, when a developer attempts to show deferred StoreKit messages, all pending messages are removed from the list—regardless of their type—even if they haven't been displayed. This leads to a bug where, if a developer intends to display only a specific subset of message types, all messages are incorrectly marked as shown. As a result, subsequent calls to showStoreMessages() fail to display the messages that were initially skipped.

### Description
This PR addresses the issue by retaining any messages that don't match the specified types, allowing them to be displayed in future calls to showStoreMessages().

### Testing
Added a unit test to cover this scenario.

### DIFF
--- a/Sources/Support/StoreMessagesHelper.swift
+++ b/Sources/Support/StoreMessagesHelper.swift
@@ -68,10 +68,9 @@ actor StoreMessagesHelper: StoreMessagesHelperType {
                 }
             }
         }
-        self.deferredMessages.removeAll { message in
-            displayedMessages.contains { displayedMessage in
-                displayedMessage.reason == message.reason
-            }
+
+        for message in displayedMessages {
+            self.deferredMessages.removeAll(where: { $0.hashValue == message.hashValue })
         }
     }
 
@@ -98,6 +97,10 @@ protocol StoreMessage: Sendable {
 
     @available(iOS 16.0, *)
     var reason: Message.Reason { get }
+
+    @available(iOS 16.0, *)
+    // swiftlint:disable:next legacy_hashing
+    var hashValue: Int { get }
 
     @available(iOS 16.0, *)
     @MainActor
--- a/Sources/Support/StoreMessagesHelper.swift
+++ b/Sources/Support/StoreMessagesHelper.swift
@@ -57,16 +57,20 @@ actor StoreMessagesHelper: StoreMessagesHelperType {
     }
 
     func showStoreMessages(types: Set<StoreMessageType>) async {
+        var messagesToKeep: [StoreMessage] = []
         for message in self.deferredMessages {
             if let messageType = message.reason.messageType, types.contains(messageType) {
                 do {
                     try await message.display(in: self.systemInfo.currentWindowScene)
                 } catch {
                     Logger.error(Strings.storeKit.error_displaying_store_message(error))
+                    messagesToKeep.append(message)
                 }
+            } else {
+                messagesToKeep.append(message)
             }
         }
-        self.deferredMessages.removeAll()
+        self.deferredMessages = messagesToKeep
     }
 
     #endif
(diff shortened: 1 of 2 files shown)


Sources/Support/StoreMessagesHelper.swift
```diff
@@ -57,16 +57,20 @@ actor StoreMessagesHelper: StoreMessagesHelperType {
     }
 
     func showStoreMessages(types: Set<StoreMessageType>) async {
+        var messagesToKeep: [StoreMessage] = []
```
fire-at-will, line 60: There may be a more elegant solution here, but modifying `self.deferredMessages` while iterating over it could lead to unexpected behavior, so I opted for this approach to avoid potential issues.

MarkVillacampa: Good call. I'd probably do it the opposite way: store the messages we processed, and delete them from the array at the end. That way we don't loose any messages that might've been added while we were processing.

We can also make the variable `Atomic<>` to make sure we handle threading correctly.

fire-at-will: Is that case possible since `StoreMessagesHelper` is an actor, and `deferMessagesIfNeeded()` is called in it's own separate Task (created [here in PurchasesOrchestrator](https://github.com/RevenueCat/purchases-ios/blob/b5e05d6ad8c5a5cc87aecb75af61e8b28afd6569/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift#L169))?

MarkVillacampa: Good point, didn't realise it's an actor, we probably don't need the Atomic in that case.

However, there is an suspension point (await) inside the `deferredMessages`, and that could be an opportunity for the `storeMessagesProvider.messages` loop to process a new message, so I'd still store the messages we processed and delete them, rather than store the messages to keep.

fire-at-will: Makes sense - I'll make the change!

tonidero (approved):
Oh nice catch. Thanks for addressing!

Sources/Support/StoreMessagesHelper.swift
```diff
@@ -57,16 +57,22 @@ actor StoreMessagesHelper: StoreMessagesHelperType {
...
                     Logger.error(Strings.storeKit.error_displaying_store_message(error))
                 }
             }
         }
-        self.deferredMessages.removeAll()
+        self.deferredMessages.removeAll { message in
+            displayedMessages.contains { displayedMessage in
+                displayedMessage.reason == message.reason
+            }
+        }
```
JayShortway, line 75: Maybe `displayedMessages` could be a set (of `reason`s) to avoid having to iterate over it for every item in `deferredMessages`?

(Just for my understanding: why are we comparing on `reason`? I noticed that a `Message` also has a [`hashValue`](https://developer.apple.com/documentation/storekit/message/3954509-hashvalue).)

JayShortway (approved):
Nice one!

Sources/Support/StoreMessagesHelper.swift
fire-at-will: Great question! I was chatting with @MarkVillacampa about this on Friday. We were comparing on `reason` because the `StoreMessage` abstraction only exposed `reason`, and that was all we had available. After talking it through with @MarkVillacampa, we decided to add and compare a hashValue, so it'll only remove the exact instance that was shown. Great catch! :D

JayShortway: Nice, thanks for explaining! And I'm happy that it was a useful question haha.

PR #4348, opened 2024-10-03
Adds scaffolding for the ButtonComponent.

JayShortway:
As the title says. Just creates the files I _think_ are necessary, with very minimal content. How the `ButtonComponent` is going to interplay with the existing `LinkButtonComponent` and upcoming `PurchaseButtonComponent` can be seen later imo, but let me know your thoughts!


Sources/Paywalls/Components/PaywallButtonComponent.swift
```diff
@@ -0,0 +1,38 @@
...
+import Foundation
+
+#if PAYWALL_COMPONENTS
+
+public extension PaywallComponent {
+
+    struct ButtonComponent: PaywallComponentBase {
+
+        let type: ComponentType
+        public let text: PaywallComponent.TextComponent
```
joshdholtz, line 25: This looks similar to the `LinkButton` where it has a`TextComponent` inside of it (which makes sense) but I totally forgot to delete the `LinkComponent` and mention when why it was build that way 😅 It was built for very early demo purpose and just reused `TextComponent` because easy.

However, I _think_ we are going to be leaning towards more of an protocol style instead of composition for components that have similar things here. 

I started doing this with the `StackableComponent` in https://github.com/RevenueCat/purchases-ios/pull/4324/files#diff-ec1da30bf93d1075c4c1985847a4762e27386eb3a6cb81a81dee288d50ba2fcc

So here, we could create a new protocol called like `TextableComponent` (names are hard???) that have the properties that are similar between `TextComponent` and `ButtonComponent`.

My reason for this is:
- Keeps JSON flatter
- Keeps components isolated
  - Ex: if we had a very specific property to `TextComponent`, we might not want `ButtonComponent` to have it (which actually happens in https://github.com/RevenueCat/purchases-ios/pull/4346 with selection state)

RevenueCatUI/Templates/Components/Button/ButtonComponentView.swift
```diff
@@ -0,0 +1,65 @@
...
+struct ButtonComponentView: View {
+
+    private let viewModel: ButtonComponentViewModel
+
+    internal init(viewModel: ButtonComponentViewModel) {
+        self.viewModel = viewModel
+    }
+
+    var body: some View {
+        EmptyView()
```
joshdholtz, line 30: This can be deleted!

joshdholtz (changes requested):
Yay paywall components PR 🥳 

One change request on not composing with `TextComponent` (which I'll take credit that `LinkButton` was a bad example and I was going to delete it soon)

Sources/Paywalls/Components/PaywallButtonComponent.swift
JayShortway: That makes sense! However, are we going to allow people to give buttons arbitrary content, like SwiftUI and Jetpack Compose do? What about icons/images?

RevenueCatUI/Templates/Components/Button/ButtonComponentView.swift
JayShortway: Ah! That was a bit confusing indeed haha. It's gone: https://github.com/RevenueCat/purchases-ios/pull/4348/commits/857fad69a37086f3d4955eaee8ce9412f8840824

Sources/Paywalls/Components/PaywallButtonComponent.swift
joshdholtz: This is... a great question 🤔 In the current spec, the button only has text. However...

Maybe a `ButtonComponent` should really be a container/stack and hold a list of components... and then the dashboard component selector can give options for like `TextButtonComponent` or `IconButtonComponent` that adds a button with those sub components in it already as a template?

Does that make sense? If so... we should bring this up to the Dan and Barbara 😇

JayShortway: Yea that does make sense! And then things like "leading icon" vs "trailing icon" are just the same (horizontal) stack, but in a different order. I'll bring it up!

> Maybe a `ButtonComponent` should really be a container/stack

Yea, or it could have a single `StackComponent` child?

JayShortway: Maybe making buttons behave like stacks (in whatever implementation) might be good to do regardless, to not limit our future selves. 🤔

Sources/Paywalls/Components/PaywallButtonComponent.swift
```diff
@@ -0,0 +1,38 @@
...
+        public let stack: PaywallComponent.StackComponent
+
+        public init(
+            stack: PaywallComponent.StackComponent
+        ) {
+            self.type = .button
+            self.stack = stack
+        }
+
+    }
```
joshdholtz, line 34: So in my packages PRs, I have that `StackableComponent` protocol (see [here](https://github.com/RevenueCat/purchases-ios/pull/4324/files#diff-6fed0f670d26d3bdf79586129d096238c544e4022a76f86a096997b387e6764dR22-R34)) that does a similar thing to this 🤔 

But instead of the `ButtonComponent` _having_ a stack, it would make the `ButtonComponent` behave like a stack.

I know we don't `StackableComponent` in this branch yet but we should probably make sure we do the same approach for all the things 😇 

I do think this having a stack looks a lot cleaner in the code! 🫶  However, it does make the JSON data structure of the `ButtonComponent` more nesty...

```js
{
  "type": "button",
  # button properties
  "stack": {
    "type": "stack",
    "components": [
      # children components
    ]
    # all the stack properties
  }
}
```

vs


```js
{
  "type": "button",
  "components": [
      # children components
    ],
   # button properties
   # all the stack properties
}
```

Any thoughts? 😇

joshdholtz (commented):
Making the button behave like a stack looks good! Just one major question on the implementation of the stack here vs https://github.com/RevenueCat/purchases-ios/pull/4324/files#diff-6fed0f670d26d3bdf79586129d096238c544e4022a76f86a096997b387e6764dR22-R34

joshdholtz (approved):
This looks good! 😁 

Note: After some conversations, we will fix all container components (like a button) are stacks

PR #4353, opened 2024-10-04
Models the Action for the ButtonComponent

JayShortway:
As the title says. No implementation whatsoever. Will do that in follow-ups.


RevenueCatUI/Templates/Components/Button/ButtonComponentView.swift
```diff
@@ -27,7 +27,11 @@ struct ButtonComponentView: View {
     }
 
     var body: some View {
-        StackComponentView(viewModel: viewModel.stackViewModel)
+        EmptyView()
```
JayShortway, line 30: Note to self: remove this.

Sources/Paywalls/Components/PaywallButtonComponent.swift
```diff
@@ -21,13 +21,34 @@ public extension PaywallComponent {
...
+        public enum Destination: Codable, Sendable, Hashable, Equatable {
+            case customerCenter
+            case URL(url: URL, method: URLMethod)
+        }
+
+        public enum URLMethod: Codable, Sendable, Hashable, Equatable {
+            case inAppBrowser
+            case externalBrowser
+        }
+        // swiftlint:enable nesting
```
JayShortway, line 40: @joshdholtz This is definitely not final, but wanted to gather some early feedback. Specifically: 
- Do we have a scheme for this already? (If so, where can I find it?)
- I see SwiftLint doesn't like the nesting, but it's helpful for namespacing. Otherwise all these enums would need substantially longer names. What is our usual approach?

Sources/Paywalls/Components/PaywallButtonComponent.swift
```diff
@@ -21,13 +21,34 @@ public extension PaywallComponent {
...
+        // swiftlint:disable nesting
+        public enum Action: Codable, Sendable, Hashable, Equatable {
+            case restorePurchases
+            case navigateTo(destination: Destination)
+            case navigateBack
+        }
+
+        public enum Destination: Codable, Sendable, Hashable, Equatable {
+            case customerCenter
+            case URL(url: URL, method: URLMethod)
```
JayShortway, line 33: TODO: make the `url` localizable.

Sources/Paywalls/Components/PaywallButtonComponent.swift
joshdholtz: > Do we have a scheme for this already? (If so, where can I find it?)

We do not for this one! I had some Notion doc started where we were going to keep schemes and I'll link that to you in Slack

> I see SwiftLint doesn't like the nesting

It is a little agressive in its nesting requirements 😅  I usually put `// swiftlint:disable:next nesting` above each one so that I don't accidentally miss the "enable". Otherwise this looks fine!

JayShortway: Thanks so much! Will continue on this path and let you know once it's ready for a real review 😄

Sources/Paywalls/Components/PaywallButtonComponent.swift
```diff
@@ -21,13 +21,38 @@ public extension PaywallComponent {
...
+            case navigateTo(destination: Destination)
+            case navigateBack
+        }
+
+        // swiftlint:disable:next nesting
+        public enum Destination: Codable, Sendable, Hashable, Equatable {
+            case customerCenter
+            case URL(urlLid: String, method: URLMethod)
+            case privacyPolicy(urlLid: String, method: URLMethod)
+            case terms(urlLid: String, method: URLMethod)
```
JayShortway, line 36: The `privacyPolicy` and `terms` cases are the same as the `URL` case, but I included them since our model mirrors the dashboard. (Same reason why we have the `deepLink` case in the `URLMethod` enum, even though it is the same as `externalBrowser`.) Let me know if you have any thoughts on this.

JayShortway, line 36: These all have a `urLLid` property to follow our convention of passing "localization IDs". Let me know if this is not the way we should localize URLs.

joshdholtz: This looks good to me! 👍

joshdholtz: This makes sense to me! 👍

RevenueCatUI/Templates/Components/Button/ButtonComponentViewModel.swift
```diff
@@ -39,6 +39,11 @@ public class ButtonComponentViewModel {
         )
     }
 
+    func onClick() {
+        // swiftlint:disable:next todo
+        // TODO Handle the configured Action.
+    }
```
joshdholtz, line 45: I mentioned this in a PR further down the chain but I would probably name this to `performAction()`since it could be a click or a tap 😇 Just a nit of mine 😅

joshdholtz (approved):
A small nit on a rename but otherwise this looks good to me! Approving because I mentioned this on a PR further down the chain that I requested changes on

RevenueCatUI/Templates/Components/Button/ButtonComponentViewModel.swift
JayShortway: Oh yes, great point. I'll rename it in the other PR, since this function is moved in that PR as well.

PR #4359, opened 2024-10-07
Fix `hasFeature(RetroactiveAttribute)` check in iOS 14

vegaro:
iOS 14 tests are failing with:

```
❌  /Users/distiller/purchases-ios/Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift:259:5: unexpected platform condition (expected 'os', 'arch', or 'swift')

#if hasFeature(RetroactiveAttribute)
    ^
```

see https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/22627/workflows/1c541690-3b85-4aa2-ad9e-dc7f15ada779/jobs/262246/steps

I think adding a swift compiler check would fix this, but seeing this issues I realized we should maybe change the check to `#if $RetroactiveAttribute`
https://github.com/swiftlang/swift/issues/69643
https://forums.swift.org/t/hasattribute-only-supports-declaration-attributes/68202

### DIFF
--- a/Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
+++ b/Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
@@ -256,7 +256,8 @@ private extension ProductsFetcherSK1 {
 // - It has mutable state, but it's made thread-safe through `queue`.
 extension ProductsFetcherSK1: @unchecked Sendable {}
 
-#if $RetroactiveAttribute
+#if swift(>=5.8)
+#if hasFeature(RetroactiveAttribute)
 // Conformance should be safe since it is only used as dictionary key
 extension SKRequest: @unchecked @retroactive Sendable {}
 extension SKProductsRequest: @unchecked @retroactive Sendable {}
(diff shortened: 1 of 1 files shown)


vegaro:
@RCGitBot please test

Sources/Purchasing/StoreKit1/ProductsFetcherSK1.swift
```diff
@@ -256,7 +256,7 @@ private extension ProductsFetcherSK1 {
 // - It has mutable state, but it's made thread-safe through `queue`.
 extension ProductsFetcherSK1: @unchecked Sendable {}
 
-#if hasFeature(RetroactiveAttribute)
+#if $RetroactiveAttribute
```
aboedo, line 259: just when I thought I understood Swift

aboedo (approved):

MarkVillacampa:
Turns out `hasFeature` was introduced in Swift 5.8, which was first bundled with Xcode 14.3, so it's a pity we need to use Xcode 14.2 in order to run in macOS 12 because iOS sims only run there.

According to SE-0362 which introduced the `hasFeature` condition, variables with $ are still defined by the compiler to support older tools, however it is not recommended:
https://github.com/swiftlang/swift-evolution/blob/main/proposals/0362-piecemeal-future-features.md#x-instead-of-hasfeaturex

I think a more future-proof fix in this case would be to first check for the swift version, and then the feature. Something like:

```
#if swift(>=5.8)
#if hasFeature(ReroactiveAttribute)
XXXXXX
#endif
#endif
```

vegaro:
@RCGitBot please test

PR #4360, opened 2024-10-08
ButtonComponent can handle URL destinations

JayShortway:
As the title says.


RevenueCatUI/Templates/Components/Button/ButtonComponentView.swift
```diff
@@ -28,13 +31,74 @@ struct ButtonComponentView: View {
...
+            navigateToUrl(urlLid: urlLid, method: method)
+        }
+    }
+
+    private func navigateToUrl(urlLid: String, method: PaywallComponent.ButtonComponent.URLMethod) {
+        guard let urlString = try? viewModel.localizedStrings.string(key: urlLid),
+        let url = URL(string: urlString) else {
+            Logger.error(Strings.paywall_invalid_url(urlLid))
+            return
+        }
```
JayShortway, line 74: 2 comments for your consideration: 
- This means that any error in configuring the URLs will only surface once the button is clicked. Alternatively, we can parse the URL at construction time.
- This bit of logic in the view means it's no longer a "dumb view", as MVVM prescribes. Alternatively, it could be moved to the ViewModel.

Let me know what you think!

RevenueCatUI/Templates/Components/Button/ButtonComponentView.swift
```diff
@@ -28,11 +33,77 @@ struct ButtonComponentView: View {
...
+
+        switch method {
+        case .inAppBrowser:
+#if os(tvOS)
+            // There's no SafariServices on tvOS, so we're falling back to opening in an external browser.
+            openURL(url)
+#else
```
joshdholtz, line 83: Can we add a log that says this too? Don't think it will be a big deal but 🤷‍♂️

RevenueCatUI/Templates/Components/Button/ButtonComponentView.swift
joshdholtz: Instead of calling this in `navigateToUrl`, can we call this in the constructor of the `ButtonComponentViewModel`?

The constructor of the view models "throw" errors. So you could do something like how the `PurchaseButtonComponentViewModel` does 👇 

```swift
self.cta = try localizedStrings.string(key: component.cta)
self.ctaIntroOffer = try component.ctaIntroOffer.flatMap {
    try localizedStrings.string(key: $0)
}
```

You can try to grab and parse the URL (if it exists for the action) in the constructor and then throw an error to 💥 the paywall on initialization.

This _may_ require another type of view model enum or something but I think the best way to follow the standard error flow of paywalls to error as early as possible 😇 

Let me know if you have any questions on this! I'd be happy to pair 😊

joshdholtz (changes requested):
A few change requests for this one! Let me know if you have any questions or want to pair 💪 

This is looking good though!

RevenueCatUI/Templates/Components/Button/ButtonComponentViewModel.swift
```diff
@@ -20,8 +20,28 @@ import RevenueCat
...
+
+    /// A mirror of ButtonComponent.Destination, with any urlLid parameters changed to be of the actual URL type. This
+    /// allows us to verify the URLs exist and are properly formatted, before making them available to the view layer.
+    /// This way the view layer doesn't need to handle this error scenario.
+    internal enum Destination {
+        case customerCenter
+        case URL(url: URL, method: PaywallComponent.ButtonComponent.URLMethod)
+        case privacyPolicy(url: URL, method: PaywallComponent.ButtonComponent.URLMethod)
+        case terms(url: URL, method: PaywallComponent.ButtonComponent.URLMethod)
+    }
```
JayShortway, line 40: The goal here is for the `ButtonComponentView` to not have to worry about any "invalid config" scenarios. 

I also considered saving the `URL` in an optional field, which would always be non-nil in case we had a URL-like `Destination`. However, that would require the `ButtonComponentView` to trust the `ButtonComponentViewModel` on this (e.g. by force unwrapping). We would then need to write tests to ensure this "contract" never breaks through any refactorings. 

With the current approach, the `ButtonComponentView` reads a non-optional `URL` field, and thus it doesn't need to trust anyone (except the compiler), nor do we need to write a test for this case. 

Let me know if there's a better way of course!

RevenueCatUI/Templates/Components/Button/ButtonComponentView.swift
```diff
@@ -13,12 +13,17 @@
...
+#if canImport(SafariServices)
+import SafariServices
+#endif
 import SwiftUI
 
 #if PAYWALL_COMPONENTS
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 struct ButtonComponentView: View {
+    @Environment(\.openURL) var openURL
```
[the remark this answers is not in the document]
JayShortway: Ah yes, of course: https://github.com/RevenueCat/purchases-ios/pull/4360/commits/c3d8356deb907db3f43094edfe83b2e6f0fe45e1

RevenueCatUI/Templates/Components/Button/ButtonComponentView.swift
JayShortway: Yea this makes total sense! Thanks for the pointers too! I made the change here, and added some GitHub comments explaining my rationale: https://github.com/RevenueCat/purchases-ios/pull/4360/commits/63635a6190d858364200bffd28b2482e61fe7b0f.

RevenueCatUI/Templates/Components/Button/ButtonComponentViewModel.swift
```diff
@@ -31,17 +51,52 @@ public class ButtonComponentViewModel {
...
+                    destination: .privacyPolicy(url: try localizedStrings.urlFromLid(urlLid), method: method)
+                )
+            case .terms(let urlLid, let method):
+                self.action = .navigateTo(
+                    destination: .terms(url: try localizedStrings.urlFromLid(urlLid), method: method)
+                )
+            }
+        case .navigateBack:
+            self.action = .navigateBack
+        }
```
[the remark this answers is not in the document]
joshdholtz: I would write all my code in switches like this if I could 😅

RevenueCatUI/Templates/Components/Button/ButtonComponentViewModel.swift
joshdholtz: This looks like how I would have done it! 💪

joshdholtz (approved):
This looks good! :shipit:

RevenueCatUI/Templates/Components/Button/ButtonComponentViewModel.swift
JayShortway: A compliment of the highest regard 🙇

RevenueCatUI/Templates/Components/Button/ButtonComponentViewModel.swift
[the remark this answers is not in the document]
JayShortway: I feel so safe in the arms of the compiler 🤗

PR #4312, opened 2024-09-27
Adds an `onDismiss` callback to `ErrorDisplay`

JayShortway:
As the title says, and as discussed on Slack. This is needed for https://github.com/RevenueCat/purchases-kmp/issues/217.

PR #4365, opened 2024-10-09
ButtonComponent can dismiss the paywall

JayShortway:
As the title says. I decided to pass an `onDismiss` callback down, to avoid potential issues in hybrids where the paywall doesn't dismiss when we want it to. (See https://github.com/RevenueCat/purchases-kmp/issues/217 & https://github.com/RevenueCat/purchases-ios/pull/4312.)

Let me know if you think we should handle this differently!


joshdholtz (approved):
Looks good to me! 💪 :shipit:

PR #4367, opened 2024-10-10
Local.xcconfig is read by PurchaseTester and PaywallsTester

JayShortway:
As the title says. 

A follow up to #4341, some more things were needed: 
- Adds file references to the existing `Global.xcconfig` and `Local.xcconfig` files to the PurchaseTester and PaywallsTester projects. 
- PurchaseTester and PaywallsTester also read build settings from these files.

One piece of the puzzle left: SPM. This is tackled in https://github.com/RevenueCat/purchases-ios/pull/4368.


joshdholtz (approved):
YESSSSS 😍

PR #4368, opened 2024-10-10
Local.xcconfig is read by Package.swift

JayShortway:
As the title says.

A follow up to https://github.com/RevenueCat/purchases-ios/pull/4367. `Local.xcconfig` is parsed in `Package.swift` and any enabled compiler flags are extracted.

After this is merged, compiler flags can be fully managed in `Local.xcconfig`.


Package.swift
```diff
@@ -3,6 +3,27 @@
...
+    // We split the capture group by space and remove any special flags, such as $(inherited).
+    return config
+        .firstMatch(of: #/^SWIFT_ACTIVE_COMPILATION_CONDITIONS *= *(.*)$/#.anchorsMatchLineEndings())?
+        .output
+        .1
+        .split(whereSeparator: \.isWhitespace)
+        .filter { !$0.isEmpty && !$0.hasPrefix("$") }
+        .map { .define(String($0)) }
+        ?? []
+}()
```
joshdholtz, line 26: This change will go out to users of our SDK too, right? 🤔 So if they have a `Local.xcconfig` file, this logic will get executed?

If that's the case, should we end up naming our file something more like `RCLocal.xcconfig` so it has less likely chance t pick anything else up? 🤷‍♂️

joshdholtz (commented):
@JayShortway This is an amazing solution! One small concern about this logic being used by developers that use our SDK 😇

Package.swift
JayShortway: That's a good concern! But I think we should be good. The `#file` macro [returns the path of the file in which it is defined](https://developer.apple.com/documentation/swift/file()). So for library consumers that is `/path/to/purchases-ios-spm/Package.swift`. Unless they have a `Local.xcconfig` file in the `purchases-ios-spm` folder (unlikely imo), nothing should get picked up. Let me know what you think! 

(I changed it to the more future-proof `#filePath` in https://github.com/RevenueCat/purchases-ios/pull/4368/commits/030a01554010dcb3725673a3d52fb91769ed1242.)

joshdholtz (approved):
This looks good to me! Really love this solution 😍 

I'd probably like @aboedo's review on this as well since since its whole SDK thing 😇

Package.swift
```diff
@@ -3,6 +3,27 @@
...
```
aboedo, line 26: This seems like a great idea, but it also means that if you have a `Local.xcconfig` defined for your app, we're also going to read those, right? 
Maybe in turn, we should make sure that we don't have any risk of name conflicts? Like, either by updating the `Local.xcconfig` file name or the flags themselves with an RC prefix / suffix. 

Like: `RevenueCatLocal.xcconfig` or `REVENUECAT_PAYWALL_COMPONENTS`

aboedo (approved):

Package.swift
aboedo: Update: just read the other conversation right after asking about the same thing 😂 
Maybe we should document in a comment that it only applies if you have it defined in the exact folder, so that people using the library and looking at the code don't wonder about the same thing?

JayShortway: That's a good idea to clear any doubts! How about: https://github.com/RevenueCat/purchases-ios/pull/4368/commits/3f57bafae9cdf1718444e9784acb7e57067bb008.

aboedo: :chefskiss:

PR #4370, opened 2024-10-10
Support Redeeming Win-Back Offers with Streamlined Purchasing Disabled

fire-at-will:
This PR adds support for redeeming win-back offers with Streamlined Purchasing disabled.

### Feature Availability
This feature is available for developers under the following conditions:
- The app is run on:
   - iOS 18.0+
   - macOS 15.0+
- The app is compiled with Xcode 16.0+

### Observing the SK1 Payment Queue
Apple has a note in their [PurchaseIntent docs](https://developer.apple.com/documentation/storekit/purchaseintent) that reads:
> Important
>
> To enable promoted in-app purchases, your app must use either PurchaseIntent (starting in iOS 16.4) or [paymentQueue(_:shouldAddStorePayment:for:)](https://developer.apple.com/documentation/storekit/skpaymenttransactionobserver/2877502-paymentqueue) (starting in iOS 11). Don’t use both at the same time. If necessary, use conditional compilation to identify the OS version the app is running in. For more information, see [Running code on a specific platform or OS version](https://developer.apple.com/documentation/xcode/running-code-on-a-specific-version).

Of course, we can't perform conditional compilation based on the OS since version since that isn't known at compile time. Instead, this PR modifies the `PaymentQueueWrapper` class used when the SDK is running in SK2 mode so that it only observes the SK1 payment queue when the `PurchaseIntents` API is available (iOS >=16.4, macOS >=14.4, compiled on Xcode >=16.0).

### API Changes
There are no breaking API changes with this PR. When we receive a Purchase Intent with a win-back offer attached to it, we call the `PurchaseDelegate`'s `readyForPromotedProduct()` function, and when the developer is ready for the user to complete the win-back offer redemption flow, they call its completion handler.

### OS/Feature Availability
PurchaseIntents were introduced in iOS 16.4, so we need to be careful about whether we use PurchaseIntents or SK1, depending on the runtime environment. Here's a full matrix of what features/technologies are available on each OS version:
Screenshot 2024-10-24 at 1 50 41 PM

### Testing
- ✅ I've been able to successfully manually test that the promoted purchase flow works with these changes with both SK1 & SK2.
- ⚠️ Apple does not provide a way to test redeeming win-back offers with streamlined purchasing disabled, so this is currently untested.


Sources/Purchasing/StoreKit2/StoreKit2PurchaseIntentListener.swift
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    deinit {
+        self.taskHandle?.cancel()
+        self.taskHandle = nil
+    }
+}
+
+@available(iOS 16.4, tvOS 15.0, macOS 14.4, *)
+struct StorePurchaseIntent: Sendable, Equatable {
```
fire-at-will, line 105: Wrapper around `StoreKit.PurchaseIntent` so that we can test the PurchaseIntentListener. I'll add some comments here & possibly move it to its own file before this PR is ready for final review

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -486,6 +496,86 @@ final class PurchasesOrchestrator {
         }
     }
 
+    // Win-back offers weren't introduced until iOS 18 and Xcode 16, which shipped with version 6.0 of the Swift
+    // compiler. The win-back symbols won't be found if compiled on Xcode < 16.0, so we need to ensure that
+    // the Swift compiler 6.0 or higher is available.
+    #if compiler(>=6.0)
+    @available(iOS 18.0, macOS 15.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
```
MarkVillacampa, line 503: This method feels like it has a lot of duplication with the other purchase method that includes an optional `promotionalOffer` parameter. What do you think about adding `winBackOffer: Product.SubscriptionOffer` as an optional parameter there, and we can use the compiler flag check for just the usage of `.winBackOffer(winBackOffer)` inside? I believe that is the only symbol that requires the compiler check since `Product.SubscriptionOffer` is available since iOS 15.

```
#if compiler(>=6.0)
options.insert(.winBackOffer(winBackOffer))
#endif
```

Sources/Purchasing/StoreKit1/PaymentQueueWrapper.swift
```diff
@@ -57,14 +57,40 @@ class PaymentQueueWrapper: NSObject, PaymentQueueWrapperType {
 
     private let paymentQueue: SKPaymentQueue
 
+    private lazy var purchaseIntentsAPIAvailable: Bool = {
+        // PurchaseIntents was introduced in macOS with macOS 14.4, which was first shipped with Xcode 15.3,
+        // which shipped with version 5.10 of the Swift compiler. We need to check for the Swift compiler version
+        // because the PurchaseIntents framework isn't available on Xcode versions <15.3.
```
MarkVillacampa, line 63:
```suggestion
        // because the PurchaseIntents symbol isn't available on Xcode versions <15.3.
```

Sources/Purchasing/StoreKit1/PaymentQueueWrapper.swift
```diff
@@ -57,14 +57,40 @@ class PaymentQueueWrapper: NSObject, PaymentQueueWrapperType {
...
+        #endif
+    }()
+
     weak var delegate: PaymentQueueWrapperDelegate? {
         didSet {
             if self.delegate != nil {
                 self.paymentQueue.delegate = self
-                self.paymentQueue.add(self)
+
+                if !purchaseIntentsAPIAvailable {
```
MarkVillacampa, line 80: Should we check here `isStoreKit2EnabledAndAvailable` too?

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
fire-at-will: I think this makes a lot of sense. I've made those changes in https://github.com/RevenueCat/purchases-ios/pull/4370/commits/bbeb427c7690e6c2f95f73f2b3cbcf9d3b9e59ad. I think we'll just need to be careful not to pass in both a promotional offer and a win-back offer at the same time, but it's difficult to imagine a scenario where that'd happen, especially since this is an internal SDK function

Sources/Error Handling/ErrorUtils.swift
```diff
@@ -544,6 +544,19 @@ enum ErrorUtils {
                                 fileName: fileName, functionName: functionName, line: line)
     }
 
+    /**
+     * Constructs an Error with the ``ErrorCode/invalidSubscriptionOffer`` code.
+     */
+    static func invalidSubscriptionOfferError(
```
MarkVillacampa, line 550: I'm probably missing something but I can't find where this error is used 🤔

fire-at-will: Ahh. This was used in the `purchase(sk2Product:winbackOffer:)` function that we removed earlier in the review process. Good catch! 😄

I've removed this in 08ebe4ff8ac6fb5862c123489c0085fc7ea6621f. We might want to bring back something similar when implementing win-backs into the purchase params, but we can cross that bridge when we get there 😄

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -982,6 +1025,88 @@ extension PurchasesOrchestrator: StoreKit2TransactionListenerDelegate {
...
+    // swiftlint:disable:next function_body_length
+    func storeKit2PurchaseIntentListener(
+        _ listener: any StoreKit2PurchaseIntentListenerType,
+        purchaseIntent: StorePurchaseIntent
+    ) async {
+        // Making the extension unavailable on tvOS & watchOS doesn't
+        // stop the compiler from checking availability in the functions.
+        // We also need to ensure that we're on Xcode >= 15.3, since that is when
+        // PurchaseIntents were first made available on macOS.
+        #if !os(tvOS) && !os(watchOS) && compiler(>=5.10)
```
MarkVillacampa, line 1043: Does it work if we add the availability to the function instead?

MarkVillacampa (approved):

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

PR #4372, opened 2024-10-11
ButtonComponent can restore purchases

JayShortway:
As the title says. I'm using the `PurchaseHandler` to handle the restore action, as I saw a similar approach being taken in https://github.com/RevenueCat/purchases-ios/pull/4332 for the purchase action.


Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
```diff
@@ -95,7 +95,7 @@ struct SamplePaywallsList: View {
                 paywallComponentsData: data,
                 offering: Self.loader.offeringWithDefaultPaywall(),
                 onDismiss: { self.display = nil }
-            )
+            ).environmentObject(Self.purchaseHandler)
```
JayShortway, line 98: I don't think this is how we wanna do it (permanently)? It's a separate commit, so I can easily revert it.

joshdholtz: Ah yeah, I fixed this is one of my packages commits! So if you want to remove this commit so the merge is clean that would be 🔥 

But thanks for calling this out 🫶

joshdholtz (approved):
This looks good!

If you want to remove that commit for adding the environment object in the PaywallsTester that would be 💯 since I fixed that in one of my packages PRs 💪 

Other than that this is good to merge!

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
JayShortway: Totally! Revert is here: https://github.com/RevenueCat/purchases-ios/pull/4372/commits/985b78a4aaa8b65e876ed990fc711786818b8ddc.

JayShortway:
(@joshdholtz I see you reacted with a thumbs up, so I'm gonna go ahead and merge haha.)

PR #4376, opened 2024-10-17
Refactor Paywall events so it can be used for customer center

vegaro:
We want to reuse Paywall events for Customer Center, the issue is that the Paywall events system is very types and requires PaywallEvents to be passed around.

Customer Center events will have different properties than Paywalls events, so we need a way to "untype" the code so it stores and posts events of any type.

After some attempts, I figured that if we just stores `AnyEncodable` and if we post `AnyEncodable` we can keep storing and posting paywall events the same way, and also create new Customer Center events without being constrained by paywall events types.


Sources/Paywalls/Events/Networking/PaywallEventsRequest.swift
```diff
@@ -16,20 +16,39 @@ import Foundation
...
     init(events: [PaywallStoredEvent]) {
-        self.init(events: events.map { .init(storedEvent: $0) })
+        self.init(events: events.compactMap { storedEvent in
+            switch storedEvent.feature {
+            case .paywalls:
+                guard let event = PaywallEventRequest.Event(storedEvent: storedEvent) else {
+                    return nil
+                }
+                return AnyEncodable(event)
+        })
```
vegaro, line 34: there is going to be a case `.customerCenter` here, that will create a `CustomerCenterEventRequest.Event`

(In another PR)

Sources/Paywalls/Events/Networking/PaywallEventsRequest.swift
```diff
@@ -16,20 +16,39 @@ import Foundation
...
     }
 
 }
 
-extension PaywallEventsRequest {
+protocol FeatureEvent: Encodable {
```
vegaro, line 39: `CustomerCenterEventRequest.Event` will also conform to this

Sources/Paywalls/Events/Networking/PaywallEventsRequest.swift
```diff
@@ -16,20 +16,39 @@ import Foundation
...
+    var sessionID: String { get }
+
+}
+
+// This is a `struct` instead of `enum` so that
+// we can use make it conform to Encodable
+// swiftlint:disable:next convenience_type
+struct PaywallEventRequest {
 
     enum EventType: String {
```
vegaro, line 53: This used to be part of the `PaywallEventsRequest` but we now need to have different `EventType` for `PaywallEventRequest` and `CustomerCenterRequest`

Sources/Paywalls/Events/Networking/PaywallEventsRequest.swift
```diff
@@ -16,20 +16,39 @@ import Foundation
 /// The content of a request to the events endpoints.
 struct PaywallEventsRequest {
```
vegaro, line 17: Thought about renaming this, but we are still using the `api-paywalls` to post all type of events. So I am not sure. I thought about renaming to `FeatureEventsRequest`

Sources/Paywalls/Events/PaywallEventStore.swift
```diff
@@ -40,9 +40,14 @@ internal actor PaywallEventStore: PaywallEventStoreType {
 
     func store(_ storedEvent: PaywallStoredEvent) async {
         do {
-            Logger.verbose(PaywallEventStoreStrings.storing_event(storedEvent.event))
+            if let eventDescription = try? storedEvent.event.prettyPrintedJSON {
+                Logger.verbose(PaywallEventStoreStrings.storing_event(eventDescription))
```
vegaro, line 44: Improved the logging a bit so it's easier to see what's being stored

Sources/Paywalls/Events/Networking/PaywallEventsRequest.swift
tonidero: Hmm I think we should rename it or it would be very confusing down the line, and maybe just keep using the old endpoint for now, (or even alias to a new endpoint name in the backend and deprecate the old one... But that can come in a separate PR.

Sources/Paywalls/Events/Networking/PaywallEventsRequest.swift
```diff
@@ -57,17 +77,22 @@ extension PaywallEventsRequest {
...
+extension PaywallEventRequest.Event {
 
     @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
-    init(storedEvent: PaywallStoredEvent) {
-        let creationData = storedEvent.event.creationData
-        let data = storedEvent.event.data
+    init?(storedEvent: PaywallStoredEvent) {
+        guard let eventData = storedEvent.event.value as? [String: Any],
+              let paywallEvent: PaywallEvent = try? JSONDecoder.default.decode(dictionary: eventData) else {
+            return nil
```
tonidero, line 86: Should we log something in this case? It should ideally never happen.

Sources/Paywalls/Events/Networking/PaywallEventsRequest.swift
```diff
@@ -16,20 +16,40 @@ import Foundation
...
+    var version: Int { get }
+    var appUserID: String { get }
+    var sessionID: String { get }
+
+}
+
+// This is a `struct` instead of `enum` so that
+// we can use make it conform to Encodable
+// swiftlint:disable:next convenience_type
+struct PaywallEventRequest {
```
tonidero, line 52: Hmm I'm wondering... this is not actually a request (which is `PaywallEventsRequest`), but a single event... I was wondering whether to rename to `PaywallEvent.Event` or soemthing along those lines? Not sure... I think it's a bit confusing right now...

tonidero (commented):
I'm mostly concerned about the naming being confusing... not sure if we should perform some renames now or maybe delay it for later.

tonidero (approved):
All the decodable stuff still confuses me a bit but I think it makes sense. Just a few questions to make sure I didn't miss something!

Sources/Events/Networking/EventsRequest.swift
```diff
@@ -0,0 +1,49 @@
...
+struct EventsRequest {
+
+    var events: [AnyEncodable]
+
+    init(events: [AnyEncodable]) {
+        self.events = events
+    }
+
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+    init(events: [PaywallStoredEvent]) {
```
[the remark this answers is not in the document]
vegaro: oh right, good point

Sources/Paywalls/Events/PaywallStoredEvent.swift
```diff
@@ -31,7 +38,16 @@ extension PaywallStoredEvent: Codable {
...
+        case feature
+
+    }
+
+    init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
 
+        self.event = try container.decode(AnyEncodable.self, forKey: .event)
+        self.userID = try container.decode(String.self, forKey: .userID)
+        self.feature = try container.decodeIfPresent(Feature.self, forKey: .feature) ?? .paywalls
```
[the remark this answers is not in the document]
vegaro: wouldn't it go to `?? .paywalls`? Or maybe we need to `try?` yes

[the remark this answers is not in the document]
tonidero: Yeah I think we might need to  use “try?”

Tests/UnitTests/Paywalls/Events/PaywallEventsManagerTests.swift
```diff
@@ -196,22 +203,42 @@ class PaywallEventsManagerTests: TestCase {
         let event1 = await self.storeRandomEvent()
         _ = await self.storeRandomEvent()
 
-        let task1 = Task<Int, Error> { [manager = self.manager!] in try await manager.flushEvents(count: 1) }
```
vegaro, line 199: this test was flaky. It looks like it doesn't flake anymore

tonidero (approved):
🙌 Much cleaner now IMO. Thanks!!

tonidero:
cc @MarkVillacampa since you're also working with paywall events

vegaro:
just retested yet again that it works for previous stored events and it works perfectly, so merging

PR #4378, opened 2024-10-17
[WIP] new presentCodeSheet on iOS/macCatalyst

fire-at-will:
### Checklist

### DIFF
--- a/RevenueCat.xcworkspace/xcshareddata/swiftpm/Package.resolved
+++ b/RevenueCat.xcworkspace/xcshareddata/swiftpm/Package.resolved
@@ -1,0 +1,51 @@
+{
+  "originHash" : "501760807f1881d92609a594586edb2ceb01ad8cc54331a3bde63278db5971bf",
+  "pins" : [
+    {
+      "identity" : "cwlcatchexception",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/mattgallagher/CwlCatchException.git",
+      "state" : {
+        "revision" : "3b123999de19bf04905bc1dfdb76f817b0f2cc00",
+        "version" : "2.1.2"
+      }
+    },
+    {
+      "identity" : "cwlpreconditiontesting",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/mattgallagher/CwlPreconditionTesting.git",
+      "state" : {
+        "revision" : "a23ded2c91df9156628a6996ab4f347526f17b6b",
+        "version" : "2.1.2"
+      }
+    },
+    {
+      "identity" : "nimble",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/quick/nimble",
+      "state" : {
+        "revision" : "1f3bde57bde12f5e7b07909848c071e9b73d6edc",
+        "version" : "10.0.0"
+      }
+    },
+    {
+      "identity" : "ohhttpstubs",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/AliSoftware/OHHTTPStubs.git",
+      "state" : {
+        "revision" : "12f19662426d0434d6c330c6974d53e2eb10ecd9",
+        "version" : "9.1.0"
+      }
+    },
+    {
+      "identity" : "swift-snapshot-testing",
+      "kind" : "remoteSourceControl",
+      "location" : "https://github.com/pointfreeco/swift-snapshot-testing",
+      "state" : {
+        "revision" : "26ed3a2b4a2df47917ca9b790a57f91285b923fb",
+        "version" : "1.12.0"
+      }
+    }
+  ],
+  "version" : 3
+}
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -265,6 +265,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
     fileprivate let systemInfo: SystemInfo
     private let storeMessagesHelper: StoreMessagesHelperType?
     private var customerInfoObservationDisposable: (() -> Void)?
+    private let offerCodeRedemptionSheetPresenter: OfferCodeRedemptionSheetPresenterType
 
     private let syncAttributesAndOfferingsIfNeededRateLimiter = RateLimiter(maxCalls: 5, period: 60)
 
@@ -307,6 +308,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
 
         let purchasedProductsFetcher = OfflineCustomerInfoCreator.createPurchasedProductsFetcherIfAvailable()
         let transactionFetcher = StoreKit2TransactionFetcher()
+        let offerCodeRedemptionSheetPresenter = OfferCodeRedemptionSheetPresenter()
 
         let diagnosticsFileHandler: DiagnosticsFileHandlerType? = {
             guard diagnosticsEnabled, #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *) else { return nil }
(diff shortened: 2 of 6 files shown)


RevenueCat.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -0,0 +1,51 @@
+{
+  "originHash" : "501760807f1881d92609a594586edb2ceb01ad8cc54331a3bde63278db5971bf",
```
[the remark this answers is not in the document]
fire-at-will: Usually not - it was committed by mistake! Will remove :D

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1062,15 +1067,43 @@ public extension Purchases {
...
+    @available(tvOS, unavailable)
+    @available(macOS, unavailable)
+    internal func presentOfferCodeRedemptionSheet(uiWindowScene: UIWindowScene?) async throws {
+        var windowScene = uiWindowScene
+        if windowScene == nil {
+            windowScene = try await systemInfo.currentWindowScene
+        }
+
+        guard let windowScene else {
+            Logger.error(Strings.storeKit.error_displaying_offer_code_redemption_sheet_no_window_scene)
```
fire-at-will, line 1095: It might be a little awkward that this is possible, but I thought it was important to keep the provided window scene optional for two reasons:
1. So that developers using SwiftUI can use easily the API
2. To keep the new `Purchases.presentCodeRedemptionSheet` API as similar as possible to the deprecated one

Sources/Purchasing/StoreKitAbstractions/OfferCodeRedemptionSheetPresenter.swift
```diff
@@ -0,0 +1,102 @@
...
+            // to the SK1 implementation
+            self.sk1PresentCodeRedemptionSheet()
+            return
+        }
+
+        if #available(iOS 16.0, iOSApplicationExtension 16.0, *) {
+            try await AppStore.presentOfferCodeRedeemSheet(in: windowScene)
+        } else {
+            // This case should be covered by the above OS check, but we'll include here
+            // since it's a possible code case
```
aboedo, line 81: Are we not taking into account SK1 vs SK2 preference here? 

I imagine it shouldn't make too much of a difference, but given that we know the systems are _eventually_ consistent and even API availability isn't the same, they might actually be showing two different sheets that look the same but go to SK1 / SK2 first. 

If that's the case, we should probably make sure that we're only using one SK version at a time

aboedo (commented):
generally looks great! Left one question

Sources/Purchasing/StoreKitAbstractions/OfferCodeRedemptionSheetPresenter.swift
fire-at-will: Great point! I'll modify this to take the SK preference into account 😄

Sources/Logging/Strings/StoreKitStrings.swift
```diff
@@ -200,6 +208,22 @@ extension StoreKitStrings: LogMessage {
...
+            "determine the UIWindowScene to present the sheet over. Please try " +
+            "passing in the UIWindowScene that you'd like to present the sheet over."
+
+        case .error_displaying_offer_code_redemption_sheet_unavailable_in_app_extension:
+            return "Could not display the Offer Code redemption sheet: unavailable in iOS and Mac Catalyst" +
+            "App Extensions below iOS/Mac Catalyst 16."
+
+        case .not_displaying_offer_code_redemption_sheet_because_ios_app_on_macos:
+            return "Not displaying the Offer Code redemption sheet because this is an iOS app running on macOS," +
+            "and displaying an Offer Code redemption sheet as a 'Designed for iPad' app on macOS throws an error."
```
MarkVillacampa, line 226: I'd simplify a bit this message to remove the "Designed for iPad" part, since if the app only supports iPhone it would be "Designed for iPhone" (Xcode actually calls it this way).

Maybe:

```
Could not display the Offer Code redemption sheet: not supported in iOS apps running on macOS.
```

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -679,6 +679,28 @@ public protocol PurchasesType: AnyObject {
...
 
+    /**
+     * Displays a sheet that enables users to redeem subscription offer codes that you generated in App Store Connect.
+     *
+     * - Important: Even though the docs in `SKPaymentQueue.presentCodeRedemptionSheet`
+     * say that it's available on Catalyst 14.0, there is a note:
+     * "`This function doesn’t affect Mac apps built with Mac Catalyst.`"
+     * when, in fact, it crashes when called both from Catalyst and also when running as "Designed for iPad".
+     * This is why RevenueCat's SDK makes it unavailable in mac catalyst.
+     */
```
MarkVillacampa, line 690: Does this mean calling this function of the SDK has been crashing on Catalyst and iOS apps running on macOS?

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -679,6 +679,28 @@ public protocol PurchasesType: AnyObject {
 
     #if os(iOS) || VISION_OS
```
MarkVillacampa, line 680: Was the `presentCodeRedemptionSheet` function moved inside this conditional intentionally? I think this might be a breaking change since it changes where the function is compiled (even if unavailable) so you cannot use runtime availability checks with it.

Sources/Logging/Strings/StoreKitStrings.swift
fire-at-will: Love it! Updated in 409dcac82be0af39eb817e2b861877d3be6cace6

Sources/Logging/Strings/StoreKitStrings.swift
```diff
@@ -200,6 +208,22 @@ extension StoreKitStrings: LogMessage {
...
+            return "Error displaying Offer Code redemption sheet: '\(error)'"
+
+        case .error_displaying_offer_code_redemption_sheet_no_window_scene:
+            return "Could not display the Offer Code redemption sheet: could not " +
+            "determine the UIWindowScene to present the sheet over. Please try " +
+            "passing in the UIWindowScene that you'd like to present the sheet over."
+
+        case .error_displaying_offer_code_redemption_sheet_unavailable_in_app_extension:
+            return "Could not display the Offer Code redemption sheet: unavailable in iOS and Mac Catalyst" +
+            "App Extensions below iOS/Mac Catalyst 16."
```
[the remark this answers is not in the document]
fire-at-will: Great callout, updated in 6ce7f091ffb87966aada03a51567fe0f0ff0bb2c

Sources/Purchasing/Purchases/PurchasesType.swift
fire-at-will: For Catalyst, nope! Luckily the old version of `presentCodeRedemptionSheet()` was marked as unavailable on Catalyst (see the definition in the current main branch [here](https://github.com/RevenueCat/purchases-ios/blob/95aaf748887a6f5fdef1e45e3f9b2da67b2ad97c/Sources/Purchasing/Purchases/PurchasesType.swift#L753).

For "Designed for iPad" on macOS, I haven't tested it myself, but I wouldn't be surprised if the existing SDK function is crashing, since this PR introduces the `isiOSAppOnMac` check for the first time in the code redemption sheet flow

aboedo:
@fire-at-will @MarkVillacampa what's the status of this one? is it still needed?

fire-at-will:
@aboedo This PR totally dropped off my radar 😅 

The part of this PR description that talks about not using SK1's [paymentQueue(_:shouldAddStorePayment:for:)](https://developer.apple.com/documentation/storekit/skpaymenttransactionobserver/2877502-paymentqueue) and PurchaseIntents at the same time is outdated. We found that it was going to be a lot of extra work to completely remove the `PaymentQueueWrapper`, and instead decided to only add the `PaymentQueueWrapper` as a queue listener if the PurchaseIntents API is unavailable. The code documents this here: https://github.com/RevenueCat/purchases-ios/blob/973009bbb6976f27d67995afac5d302bdf641d9b/Sources/Purchasing/StoreKit1/PaymentQueueWrapper.swift#L89

As for the `presentCodeRedemptionSheet` bit, we should probably go forward with using `AppStore.presentOfferCodeRedeemSheet(in: UIWindowScene)` to move us off of using `SKPaymentQueue.presentCodeRedemptionSheet()` when the SDK isn't in SK1 mode. I'll see if I can spruce up this PR once I get done shipping some of the VC stuff I'm working on 👍

cc @MarkVillacampa

PR #4380, opened 2024-10-18
Paywall component containers are all stacks

joshdholtz:
ℹ️ I posted a bunch of comments in the PR explaining some decisions and things I want to change

## Motivation

Containers being `[PaywallComponent]` was not descriptive enough to how they should be rendered so now all containers have a stack.

## Description

1. Root paywall is now a stack (compared to a list of components)
2. `StackComponent` was changed to `GenericStackComponent`
     - Needed to allow different types of stack (ex: Package groups should only be a stack of packages)
3. `PackageGroupViewModel` is the only view model that directly creates its children view model (which is `PackageViewModel`)
    - This is because `PackageGroupViewModel` is the only view model that needs an extra parameter (which is `Package` in this case) so it can't be rendered in the `toViewModel` recursive function


Sources/Networking/Responses/OfferingsResponse.swift
```diff
@@ -18,12 +18,22 @@ import Foundation
...
 
+    public struct ComponentsConfigs: Codable, Equatable, Sendable {
+
+        public var base: ComponentsConfig
+
+        public init(base: ComponentsConfig) {
+            self.base = base
+        }
+
+    }
```
joshdholtz, line 29: Added another layer here for something we've been talking about but didn't have any support for. This new `ComponentsConfigs` will allow for multiple paywall configurations:
- Base (the default one and the only required one)
- Landscape
- iPad
- Watch
- Mac

This isn't final at all but wasn't to throw support for this in here because why not 🤷‍♂️

Sources/Networking/Responses/OfferingsResponse.swift
```diff
@@ -18,12 +18,22 @@ import Foundation
...
+
     public struct ComponentsConfig: Codable, Equatable, Sendable {
 
-        public var components: [PaywallComponent]
+        public var stack: PaywallComponent.StackComponent
 
-        public init(components: [PaywallComponent]) {
-            self.components = components
+        public init(stack: PaywallComponent.StackComponent) {
+            self.stack = stack
```
joshdholtz, line 36: This was the big change where a paywall _was_ just a list of components but now its a stack.

Sources/Paywalls/Components/PaywallStackComponent.swift
```diff
@@ -18,10 +18,16 @@ import Foundation
...
 public extension PaywallComponent {
 
-    struct StackComponent: PaywallComponentBase, StackableComponent {
+    // The most common type of stack that can contain any type of PaywallComponent
+    typealias StackComponent = GenericStackComponent<PaywallComponent>
+
+    // A specialized stack that can only hold PaywallComponent.PackageComponent
+    typealias PackagesStackComponent = GenericStackComponent<PaywallComponent.PackageComponent>
+
+    struct GenericStackComponent<T: StackableComponent>: PaywallComponentBase {
```
joshdholtz, line 27: This is _kind of_ the weirdest change 🤷‍♂️ 

**Problem:** Stacks can contain any type components but Package Group needs to only contain Package components.

**Solution:** The main thing is now called `GenericStackComponent` (ewwww). However, there are two typealiases. `StackComponent` is exactly the behavior it was previously which was a stack that took any `PaywallComponent`. The new one is `PackagesStackComponent` which is only used by the package group and it will only accept packages.

RevenueCatUI/Templates/Components/Packages/PackageGroup/PackageGroupComponentViewModel.swift
```diff
@@ -36,20 +36,42 @@ class PackageGroupComponentViewModel {
...
+                dimension: component.stack.dimension,
+                width: component.stack.width,
+                spacing: component.stack.spacing,
+                backgroundColor: component.stack.backgroundColor,
+                padding: component.stack.padding,
+                margin: component.stack.margin,
+                cornerRadiuses: component.stack.cornerRadiuses,
+                border: component.stack.border
+            ),
+            viewModels: componentViewModels
```
joshdholtz, line 60: I don't love this so I _might_ try to reworked this but...

Above this, available packages are mapped into an array of `PackageComponentViewModel` and then converted to an array of stack view models.

This needs to stay as an array of view models because the `PackageComponentViewModel` holds the `Package`. However, `StackComponentViewModel`s are usually made with the raw components (and the view models are created in the constructor.

But because we already have the view models, I added another initializer to `StackComponentViewModel` that takes the component AND the view models. But (another one), this expects a `StackComponent` (and not a `PackageStackComponent` so we need to map that and also pass in the view models.

I'm okay with the rest of this PR except for this. But I wanted to explain myself anyway before I try to change it since it _does_ work 😇

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
```diff
@@ -35,29 +35,28 @@ struct TemplateComponentsView: View {
         self.paywallComponentsData = paywallComponentsData
         self.onDismiss = onDismiss
 
+        // Step 0: Decide which ComponentsConfig to use (base is default)
+        let componentsConfig = paywallComponentsData.componentsConfigs.base
```
joshdholtz, line 39: Right now we are just always use that new base config thingy but... this would change depending on device type or orientation later on (however we decide its going to work 🤷‍♂️)

RevenueCatUI/Templates/Components/Packages/PackageGroup/PackageGroupComponentViewModel.swift
JayShortway: I agree that this is a bit confusing, especially this line makes me pause every time I read it haha: 
```swift
 components: [], // Empty on purpose because we are feeding this view models already
```

I know what's happening, but it's not obvious at first glance. Would be nice to improve this, but not a blocker for me.

RevenueCatUI/Templates/Components/PaywallComponentViewModel.swift
```diff
@@ -20,7 +20,7 @@ enum PaywallComponentViewModel {
     case linkButton(LinkButtonComponentViewModel)
     case button(ButtonComponentViewModel)
     case packageGroup(PackageGroupComponentViewModel)
-    case package(PackageComponentViewModel)
+    // Purposely leaving out a `case package` since `PackageGroupComponentViewModel` creates this model
```
JayShortway, line 23: 👍

RevenueCatUI/Templates/Components/Packages/PackageGroup/PackageGroupComponentViewModel.swift
```diff
@@ -36,20 +36,42 @@ class PackageGroupComponentViewModel {
...
         )
     }
 
     static func getPackages(
         component: PaywallComponent.PackageGroupComponent,
         offering: Offering
```
JayShortway, line 66: I know this wasn't changed in this PR, but maybe this name can be made a bit more precise? Something like `filterAvailablePackages()`? I'm also thinking that making it an extension on `PackageGroupComponent` could improve readability.

Sources/Networking/Responses/OfferingsResponse.swift
JayShortway: Like I said, clean!

Sources/Networking/Responses/OfferingsResponse.swift
JayShortway: Interesting! I guess these are all window dimensions? We're doing window size classes, e.g. defined viewport breakpoints?

Sources/Paywalls/Components/PaywallPackageComponent.swift
```diff
@@ -19,43 +19,18 @@ import Foundation
...
-        public let components: [PaywallComponent]
-        public let width: WidthSize?
-        public let spacing: CGFloat?
-        public let backgroundColor: ColorInfo?
-        public let dimension: Dimension
-        public let padding: Padding
-        public let margin: Padding
-        public let cornerRadiuses: CornerRadiuses?
-        public let border: Border?
+        public let stack: PaywallComponent.StackComponent
```
JayShortway, line 26: Look at those lines disappearing!

Sources/Paywalls/Components/PaywallStackComponent.swift
JayShortway: I actually think I kinda like it! Nice solution that allows us to keep benefiting from compiler type checks whenever we want.

Sources/Paywalls/Components/PaywallStackComponent.swift
```diff
@@ -54,6 +60,17 @@ public extension PaywallComponent {
...
 
+public extension PaywallComponent {
+
+    // Components that can be contained in a stack
+    protocol StackableComponent: PaywallComponentBase {}
+
+}
+
+extension PaywallComponent: PaywallComponent.StackableComponent {}
+extension PaywallComponent.PackageComponent: PaywallComponent.StackableComponent {}
```
JayShortway, line 74: For my understanding, could you explain what this is used for?

JayShortway (approved):
Very nice! And a net negative too (in terms of line count)! Had a few comments, but nothing major. So I'll approve so you can merge whenever you want.

Sources/Paywalls/Components/PaywallStackComponent.swift
joshdholtz: Of course! So, I wanted to be _very_ specific in the types that could be used by `struct GenericStackComponent<T: StackableComponent>`

So I created the new `StackableComponent` component protocol and then added it to `PaywallComponent` (the enum of all the types) and `PaywallComponent.PackageComponent` (the specific package type).

I _wanted_ to do a javascript thing like `PaywallComponent  | PaywallComponent.PackageComponent` but that doesn't really exist in Swift so this is the _very_ verbose way of doing it 😛

Sources/Networking/Responses/OfferingsResponse.swift
joshdholtz: Very much more long the lines of window size classes! We'll talk about this more at our next team sync 😇  Which means... I should probably extend that meeting time 😅

Sources/Paywalls/Components/PaywallStackComponent.swift
JayShortway: Oh nice! Make-your-own-union-type! 😄 In Kotlin my first try would be a sealed type, but this is an interesting approach!

Sources/Networking/Responses/OfferingsResponse.swift
JayShortway: Haha, sweet!

RevenueCatUI/Templates/Components/Packages/PackageGroup/PackageGroupComponentViewModel.swift
joshdholtz: > I know what's happening, but it's not obvious at first glance. Would be nice to improve this, but not a blocker for me.

I'll create a cleanup task for this for me! This isn't really affecting much else and I don't want this to slow things down so I'll sleep / take the weekend on this and see if sometime magical pops into my head 😇

JayShortway: That sounds like a fine plan!

PR #4381, opened 2024-10-18
[Experimental] Web purchase redemption

tonidero:
### Description
This is a PoC for supporting deep links to be able to redeem RCBilling purchases in the app using custom schema urls and using a new redemption endpoint. Equivalent of https://github.com/RevenueCat/purchases-android/pull/1889


Sources/DeepLink/DeepLink.swift
```diff
@@ -0,0 +1,41 @@
...
+//  DeepLink.swift
+//
+//  Created by Antonio Rico Diez on 2024-10-17.
+
+import Foundation
+
+@objc public extension Purchases {
+
+    /// Class representing a RevenueCat deep link that can be processed by the SDK.
+    class DeepLink: NSObject {
```
tonidero, line 19: Nested this inside `Purchases` to namespace it.

Sources/DeepLink/DeepLink.swift
```diff
@@ -0,0 +1,41 @@
...
+
+        private override init() {}
+
+        // swiftlint:disable nesting
+
+        /// Class representing a web redemption deep link that can be redeemed by the SDK.
+        /// - Seealso: ``Purchases/redeemWebPurchase(_:)``
+        @objc public final class WebPurchaseRedemption: DeepLink {
```
tonidero, line 27: We're not using this format to create "sealed classes" right now but I thought it has some advantages like:
- Keeping construction and access to the token internal
- Allows us to work with ObjC, which is difficult with enums that have properties

Having said that, please lmk if you don't like this and I'm happy to revisit.

Sources/DeepLink/WebPurchaseRedemptionResult.swift
```diff
@@ -0,0 +1,46 @@
...
+//
+//  WebPurchaseRedemptionResult.swift
+//
+//  Created by Antonio Rico Diez on 29/10/24.
+
+import Foundation
+
+/// Represents the result of a web purchase redemption
+/// - Seealso: ``Purchases/redeemWebPurchase(_:)``
+@objc public class WebPurchaseRedemptionResult: NSObject {
```
tonidero, line 18: Similar structure to the one used for `DeepLink`. I wanted to make this compatible with ObjC, and still seal the possible values so this was the best thing I found since I think we can't use enums with properties and keep it compatible with ObjC?

tonidero (commented):
This is missing tests but would appreciate a look at the API to make sure it makes sense and I'm not doing anything crazy. @RevenueCat/sdk

Sources/Misc/Concurrency/Purchases+async.swift
```diff
@@ -47,6 +47,15 @@ extension Purchases {
         }
     }
 
+    func redeemWebPurchaseAsync(_ deepLink: Purchases.DeepLink.WebPurchaseRedemption) async throws
+    -> WebPurchaseRedemptionResult {
```
MarkVillacampa, line 51: I think the return type looks a bit weird because if the result is a `WebPurchaseRedemptionResult.Error` it would never actually be returned because it would throw first. But if it is successful you still need to check if the type `WebPurchaseRedemptionResult.Success`.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -736,6 +742,13 @@ extension Purchases {
 
 public extension Purchases {
 
+    /// Parses a deep link URL to verify it's a RevenueCat supported DeepLink
+    /// and returns the RevenueCat deep link for that URL.
+    /// - Seealso: ``Purchases/redeemWebPurchase(_:)``
```
MarkVillacampa, line 747:
```suggestion
    /// - See also: ``Purchases/redeemWebPurchase(_:)``
```

Sources/DeepLink/WebPurchaseRedemptionResult.swift
MarkVillacampa: The typical way of doing this in Objc is to pass a separate success and error completion blocks, but what we've been doing instead is passing a single block that has multiple parameters, one of them being an error e.g. `@escaping (CustomerInfo?, PublicError?)`. I think that could work in this case.

MarkVillacampa (commented):
I think it would be helpful to add Swift and Objc API tests just to see how the API would look like.

Sources/DeepLink/WebPurchaseRedemptionResult.swift
tonidero: Main concern with that approach is that we want to add more error cases in the future, for example, if the token is expired or already redeemed. I was thinking of having those be a different subclass of the result... otherwise, they would need to be part of the error and we would need to return a special type of error.

After talking with Andy, what I was thinking was to create 2 APIs, one swift-only that uses an enum and is nicer and one for ObjC which could look similar to the one currently in this PR? Wdyt?

Sources/Misc/Concurrency/Purchases+async.swift
tonidero: Yeah... as I mentioned in the other comment, we're planning to add more result types so it would make more sense there... but yeah probably doesn't make sense for this to throw though since the error will be part of one of the result types. Will try to change that

Sources/Purchasing/Purchases/Purchases.swift
tonidero: Actually I think `seealso` is correct? At least the IDE seems to process that correctly.

MarkVillacampa: oh TIL it's a keyword for the IDE!

Sources/DeepLink/WebPurchaseRedemptionResult.swift
```diff
@@ -0,0 +1,25 @@
...
+import Foundation
+
+/// Represents the result of a web purchase redemption
+/// - Seealso: ``Purchases/redeemWebPurchase(_:)``
+public enum WebPurchaseRedemptionResult: Sendable {
+
+    /// Represents that the web purchase was redeemed successfully
+    case success(_ customerInfo: CustomerInfo)
+    /// Represents that the web purchase failed to redeem
+    case error(_ error: PublicError)
```
tonidero, line 23: More cases will be added in a follow-up PR, including things like `invalidToken`, `alreadyRedeemed` or `expiredToken`.

tonidero:
Closing in favor of #4439

PR #4382, opened 2024-10-18
Support custom URL paths in `ManageSubscriptionsView`

tonidero:
### Description
This allows to specify paths with custom URLs in the manage subscriptions view which can be opened either externally or in_app as a bottom sheet.

`IN_APP`:

https://github.com/user-attachments/assets/2767eb64-c645-4f52-882c-d88233702287

`EXTERNAL`: 

https://github.com/user-attachments/assets/8ecfa3c4-9fa7-4c3f-8fb7-61fd76c8a803

### DIFF
--- a/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
@@ -163,10 +163,13 @@ class ManageSubscriptionsViewModel: ObservableObject {
 }
 
 struct IdentifiableURL: Identifiable {
+
     var id: String {
         return url.absoluteString
     }
+
     let url: URL
+
 }
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
--- a/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
@@ -37,6 +37,8 @@ class ManageSubscriptionsViewModel: ObservableObject {
     @Published
     var promotionalOfferData: PromotionalOfferData?
     @Published
+    var inAppBrowserURL: URL?
+    @Published
     var state: CustomerCenterViewState {
         didSet {
             if case let .error(stateError) = state {
@@ -152,7 +154,20 @@ class ManageSubscriptionsViewModel: ObservableObject {
             self.loadingPath = nil
         }
     }
+
+    func handleInAppBrowserDismiss() {
+        self.inAppBrowserURL = nil
+    }
 #endif
+
+}
+
+extension URL: Identifiable {
+
+    // swiftlint:disable:next missing_docs
+    public var id: String {
+        return self.absoluteString
+    }
 
 }
 
@@ -163,6 +178,7 @@ class ManageSubscriptionsViewModel: ObservableObject {
 private extension ManageSubscriptionsViewModel {
 
 #if os(iOS) || targetEnvironment(macCatalyst)
+    // swiftlint:disable:next cyclomatic_complexity
     private func onPathSelected(path: CustomerCenterConfigData.HelpPath) async {
         switch path.type {
         case .missingPurchase:
@@ -185,6 +201,18 @@ private extension ManageSubscriptionsViewModel {
                 try await purchasesProvider.showManageSubscriptions()
             } catch {
                 self.state = .error(error)
+            }
+        case .customUrl:
+            guard let url = path.url,
+                let openMethod = path.openMethod else {
+                Logger.warning("Found a custom URL path without a URL or open method. Ignoring tap.")
+                return
+            }
+            switch openMethod {
+            case .external:
+                UIApplication.shared.open(url, options: [:], completionHandler: nil)
+            case .inApp:
+                self.inAppBrowserURL = url
             }
         default:
             break
--- a/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
@@ -155,7 +155,7 @@ class ManageSubscriptionsViewModel: ObservableObject {
         }
     }
 
-    func dismissInAppBrowser() {
+    func onDismissInAppBrowser() {
         self.inAppBrowserURL = nil
     }
 #endif
--- a/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
@@ -37,7 +37,7 @@ class ManageSubscriptionsViewModel: ObservableObject {
     @Published
     var promotionalOfferData: PromotionalOfferData?
     @Published
-    var inAppBrowserURL: URL?
+    var inAppBrowserURL: IdentifiableURL?
     @Published
     var state: CustomerCenterViewState {
         didSet {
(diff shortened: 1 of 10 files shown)


RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -186,6 +200,18 @@ private extension ManageSubscriptionsViewModel {
             } catch {
                 self.state = .error(error)
             }
+        case .customUrl:
```
tonidero, line 203: Note this is only currently available in the `ManageSubscriptionView`. It won't be available in the `NoSubscriptionsView` or `WrongPlatformView` since we're not using the paths for those views.

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -152,17 +154,29 @@ class ManageSubscriptionsViewModel: ObservableObject {
...
     }
+
+    func handleInAppBrowserDismiss() {
+        self.inAppBrowserURL = nil
+    }
 #endif
 
 }
 
+struct IdentifiableURL: Identifiable {
```
tonidero, line 165: Not sure if this is the best way but I couldn't make `URL` `Identifiable` directly without getting a warning.

RevenueCatUI/Views/SafariView.swift
```diff
@@ -0,0 +1,32 @@
...
+//  SafariView.swift
+//
+//  Created by Antonio Rico Diez on 2024-10-18.
+
+#if os(iOS)
+
+import SafariServices
+import SwiftUI
+
+struct SafariView: UIViewControllerRepresentable {
```
tonidero, line 19: Lmk what you think of this. Seems to work fine and seemed like an easy solution :)

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -425,10 +452,23 @@ extension CustomerCenterConfigData.Localization {
 
 extension CustomerCenterConfigData.HelpPath {
 
-    init(from response: CustomerCenterConfigResponse.HelpPath) {
+    init?(from response: CustomerCenterConfigResponse.HelpPath) {
```
tonidero, line 455: Note that I changed the constructor to be nullable. If we get a custom url path without a url or open method, we can filter it out and this was the easiest way to achieve that

JayShortway (commented):
This is gonna be good! The only thing is the duplicate `SafariView`, and had some other minor comments.

RevenueCatUI/Views/SafariView.swift
tonidero: Ohh I had missed that one! Thanks for the heads up! I can try to consolidate yeah

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -215,13 +222,33 @@ public struct CustomerCenterConfigData {
...
+                case "IN_APP":
+                    self = .inApp
+                case "EXTERNAL":
+                    self = .external
+                default:
+                    return nil
+                }
+            }
+
+        }
```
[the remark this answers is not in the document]
tonidero: Hmm right kinda same thinking as you here... It's internal APIs so I guess we can go either way... I would suggest keeping it separate for flexibility for now, and we can always refactor later if needed. For example, we are currently only adding `inApp` and `external` for customer center but we have a `deeplink` option for paywalls... Having said that, I think we could try to refactor everything into a "link handler" to be used by both paywalls and customer center since it's close to the same thing

Sources/CustomerCenter/CustomerCenterConfigData.swift
tonidero: Hmm so for customer center I would specially take a very "lax" approach. As in, if there is something misconfigured in the backend, I would prefer to try to display as much as possible instead of failing entirely. For example in this case, it basically means that if somehow a misconfiguration happened, we would just hide the path. I think we should do this since we're still changing the schema every now and then and we should be flexible.

Having said that, we could indeed throw in the constructor and handle it when creating the paths. The code becomes a bit less legible IMO, since we need to handle the catch though... I'm not sure TBH, since I know optional constructors are actually quite common in Swift (even if I don't like them as much 😅)... Wdyt @vegaro @aboedo ?

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -152,17 +154,29 @@ class ManageSubscriptionsViewModel: ObservableObject {
             self.loadingPath = nil
         }
     }
+
+    func handleInAppBrowserDismiss() {
```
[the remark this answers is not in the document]
aboedo: +1, I'd go with `onDismissInAppBrowser` before `dismissInAppBrowser` if we're handling something that already happened rather than doing it ourselves in the method, though

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
aboedo: interesting... It might be worth figuring out a way for those to work eventually since for those it might make even more sense - you can direct users to where you'd actually want them to manage

Sources/CustomerCenter/CustomerCenterConfigData.swift
aboedo: yeah, I think it also kinda maps better to what's happening underneath - you're using an optional constructor to try to build the URL in any case

aboedo (approved):
awesome1

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
tonidero: Yeah, I think we need to move some of the logic to a shared view for all 3 of them and handle paths in all 3. Right now the buttons are mostly hardcoded in those 2 other views. Probably not something for this PR though.

Sources/CustomerCenter/CustomerCenterConfigData.swift
JayShortway: Right, this is essentially validating JSON, which I agree should be lax. We could avoid handling the catch and just doing `try?`: 
```swift
self.paths = response.paths.compactMap { try? CustomerCenterConfigData.HelpPath(from: $0) }
```

I don't feel super strongly about this btw, especially since optional constructors are common in Swift, as you said.

tonidero: Hmm yeah that's not bad. I also don't feel too strongly about this TBH... But considering we don't do that many throws along the codebase, I'm thinking on leaving it as is... If you feel we should better throw an error, I'm ok changing to your suggestion.

JayShortway (approved):
Nice one!

tonidero:
@RCGitBot please test

RevenueCatUI/Templates/Components/Button/ButtonComponentView.swift
```diff
@@ -42,7 +39,7 @@ struct ButtonComponentView: View {
             action: { try await performAction() },
             label: { StackComponentView(viewModel: viewModel.stackViewModel, onDismiss: self.onDismiss) }
         )
-        #if canImport(SafariServices)
+        #if canImport(SafariServices) && canImport(UIKit)
```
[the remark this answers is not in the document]
JayShortway: Ooh, good to have this fixed!

Sources/CustomerCenter/CustomerCenterConfigData.swift
JayShortway: No it's minor imo. Feel free to leave as is!

vegaro: I would also leave this as nullable as well because of the optional constructor it is using underneath

Sources/Networking/Responses/CustomerCenterConfigResponse.swift
```diff
@@ -167,4 +179,10 @@ extension CustomerCenterConfigResponse.HelpPath.PathType: CodableEnumWithUnknown
 
 }
 
+extension CustomerCenterConfigResponse.HelpPath.OpenMethod: CodableEnumWithUnknownCase {
```
tonidero, line 182: @vegaro Note that I added this for the new enum

PR #4386, opened 2024-10-21
Add default values to enums in Customer Center config response

vegaro:
We added a new case in the backend for path types, and current code in the SDK fails to decode unknown values.

We need to handle unknown values and also make sure these new values are not sent down by the backend for versions that break with unknown values.


tonidero (approved):
Looks good! Thanks for fixing!

PR #4393, opened 2024-10-21
Adds compatibility for suffix offer identifiers

vegaro:
We ask folks to configure the same promotional offer ids for all products, and this works for products that don't belong to the same subscription group, but promotional offer ids have to be unique per product inside the same subscription group.

This PR adds support for promotional offer ids that can be configured using a suffix that can be set in the dashboard. This way promo offer ids can be set as "product_id_1_rc_refund_offer" and "product_id_2_rc_refund_offer", and the Customer Center will be able to load them and AppStore connect will allow to create them.


RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
```diff
@@ -49,8 +49,12 @@ class LoadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType {
...
             }
 
-            guard let discount = subscribedProduct.discounts.first(where: {
-                $0.offerIdentifier == promoOfferDetails.iosOfferId
+            guard let discount = subscribedProduct.discounts.first(where: { discount in
+                guard let offerIdentifier = discount.offerIdentifier else {
+                    return false
+                }
+                return offerIdentifier == promoOfferDetails.iosOfferId ||
+                       offerIdentifier.hasSuffix("_\(promoOfferDetails.iosOfferId)")
```
tonidero, line 57: Should we first search for the exact match and if it's not found, then find the suffix? I'm thinking that with the current implementation we might find a version with the suffix even if we have another one that is an exact match.

tonidero (commented):
Just a comment but looking good!

RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
vegaro: I think that's better yes. I updated it

tonidero (approved):
LGTM!

PR #4394, opened 2024-10-22
Fix iOS 15, 14 tests using wrong version of `swift-snapshot-testing` and API tests

vegaro:
```
❌ /Users/distiller/purchases-ios/Tests/UnitTests/Mocks/MockHTTPClient.swift:98:13: ambiguous use of 'assertSnapshot(matching:as:named:record:timeout:file:testName:line:)'
            assertSnapshot(matching: call,
            ^
```

We were getting that issue only in iOS 14 and iOS 15. @JayShortway noticed the version of `swift-snapshot-testing` might differ because of the how we set the version of the library in the Package.swift files


vegaro:
@RCGitBot please test

PR #4396, opened 2024-10-23
Fix integration tests simulator

vegaro:
We were using an unavailable simulator


tonidero (approved):
🥴 Thanks for fixing this!

PR #2358, opened 2023-03-21
`Offline Entitlements`: create `CustomerInfo` from offline entitlements

NachoSoto:
Based on #2352.
Depends on #2351 and #2365.

PR #4399, opened 2024-10-23
Fighting flakiness: no longer uses `beCloseToDate` in `CustomerInfoOfflineEntitlementsStoreKitTest.verifyEntitlement`

JayShortway:
I have seen that `beCloseToDate()` / `beCloseToNow()` is one of our sources of flakiness. For instance, see the last test [here](https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/22711/workflows/611f9df0-7cef-4abe-a622-c64cb5e97f97/jobs/263013/tests). A hiccup in the CI server, an extra background process, a little bit more heat in the CircleCI data center, and things can take a fraction longer, causing this to fail.

This PR makes the tests deterministic, by defining the exact date we expect. It only tackles `CustomerInfoOfflineEntitlementsStoreKitTest.verifyEntitlement()` for now. More to come in later PRs.

### DIFF
--- a/Sources/OfflineEntitlements/PurchasedSK2Product.swift
+++ b/Sources/OfflineEntitlements/PurchasedSK2Product.swift
@@ -47,7 +47,7 @@ extension PurchasedSK2Product {
         self.entitlement = .init(
             expiresDate: expiration,
             productIdentifier: transaction.productID,
-            purchaseDate: Date(),
+            purchaseDate: transaction.purchaseDate,
             rawData: (try? transaction.jsonRepresentation.asJSONDictionary()) ?? [:]
         )
     }
--- a/Tests/StoreKitUnitTests/OfflineEntitlements/CustomerInfoOfflineEntitlementsStoreKitTest.swift
+++ b/Tests/StoreKitUnitTests/OfflineEntitlements/CustomerInfoOfflineEntitlementsStoreKitTest.swift
@@ -57,7 +57,9 @@ class CustomerInfoOfflineEntitlementsStoreKitTest: StoreKitConfigTestCase {
                                    productID: transaction.productID,
                                    entitlementID: entitlementID,
                                    periodType: .trial,
-                                   expiration: transaction.expirationDate)
+                                   expiration: transaction.expirationDate,
+                                   originalPurchaseDate: transaction.originalPurchaseDate,
+                                   latestPurchaseDate: transaction.originalPurchaseDate)
     }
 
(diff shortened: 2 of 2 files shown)


Sources/OfflineEntitlements/PurchasedSK2Product.swift
```diff
@@ -47,7 +47,7 @@ extension PurchasedSK2Product {
         self.entitlement = .init(
             expiresDate: expiration,
             productIdentifier: transaction.productID,
-            purchaseDate: Date(),
+            purchaseDate: transaction.purchaseDate,
```
JayShortway, line 50: Note that this is an actual implementation change. However, using the transaction date might be more correct? This was introduced in https://github.com/RevenueCat/purchases-ios/pull/2358, but there are no details on this particular property. Let me know if this is actually not what we want.

JayShortway:
@RCGitBot please test

JayShortway:
@RCGitBot please test

Sources/OfflineEntitlements/PurchasedSK2Product.swift
tonidero: Seems this is only used for offline entitlements and conceptually it makes sense. Also checked on Android and seems this is the logic we use there so I think it's ok...

But I would like someone from @RevenueCat/catforms to double-check.

tonidero (approved):
This is great! Thanks for starting on this 🙏. Seems ok to me. I would like someone else from @RevenueCat/catforms to also confirm it's ok

Sources/OfflineEntitlements/PurchasedSK2Product.swift
JayShortway: Thanks Toni! Hi @RevenueCat/catforms, could you have a quick look? 🙏

aboedo: probably a remnant from the first drafts where we were just trying to make it work with whatever data was available, but I don't think there's any reason (or any wins) to make this `Date()`, transaction purchase date is more accurate

aboedo (approved):

MarkVillacampa (approved):
good catch!

Sources/OfflineEntitlements/PurchasedSK2Product.swift
JayShortway: That's good to know. Thanks for verifying!

PR #4400, opened 2024-10-24
Introduce PurchaseParams to allow passing extra configuration info when making a purchase

MarkVillacampa:
Current status:
- Functionality is gated behind the ENABLE_PURCHASE_PARAMS flag
- Does not work in StoreKit 1 mode
- Does not have Paywall purchase methods
- The transaction metadata is not read from the customer info response yet
- Added support for configuring adding metadata to a purchase in PurchaseTester.

Basically I focused on implementing the public API and making it possible to test the end-to-end flow using PurchaseTester.

The API looks like this:

```swift
 let params = PurchaseParams.Builder(package: package)
                            .with(metadata: ["key": "value"])
                            .with(promotionalOffer: promotionalOffer)
                            .build()
  Purchases.shared.purchase(package: package, params: params)
  ```

### Checklist

### CODE REVIEW

File: Sources/Logging/Strings/PurchaseStrings.swift [hunk 1]
```diff
@@ -202,19 +200,19 @@ extension PurchaseStrings: LogMessage {
...
+                message += " from package in Offering " +
+                "'\(package.presentedOfferingContext.offeringIdentifier)'"
+            }
+            if let discount = discount {
+                message += " with Offer '\(discount.identifier)'"
+            }
+            if let metadata = metadata {
+                message += " with metadata: \(metadata)"
+            }
+            return message
```
MarkVillacampa, line 215: Took the opportunity to simplify the log message to it can be conditionally composed depending on the optional parameters instead of creating a new message for every combination.

Sources/Networking/Responses/CustomerInfoResponse.swift
```diff
@@ -61,6 +61,9 @@ extension CustomerInfoResponse {
         @IgnoreDecodeErrors<PurchaseOwnershipType>
         var ownershipType: PurchaseOwnershipType
         var productPlanIdentifier: String?
+        #if ENABLE_PURCHASE_PARAMS
+        var metadata: [String: String]?
+        #endif
```
MarkVillacampa, line 66: Currently, the metadata information is only present in the `Subscription` object in the `customerInfo.subscriber.subscriptions` array. We'll expand this to expose it in other places.

Sources/Purchasing/Purchases/PurchaseParams.swift
```diff
@@ -0,0 +1,72 @@
...
+ *
+ * ```swift
+ * let params = PurchaseParams.Builder()
+ *                            .with(metadata: ["key": "value"])
+ *                            .with(promotionalOffer: promotionalOffer)
+ *                            .build()
+ *  Purchases.shared.purchase(package: package, params: params)
+ * ```
+ */
+@objc(RCPurchaseParams) public final class PurchaseParams: NSObject {
```
MarkVillacampa, line 32: Opted for using the builder pattern instead of using a Set like in StoreKit's [PurchaseOption](https://developer.apple.com/documentation/storekit/product/purchaseoption) as well as calling it `PurchaseParams` so it more closely resembles our Android APIs. Additionally, we already use the builder pattern in the `Configuration` class so it's consistent.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -315,43 +315,16 @@ final class PurchasesOrchestrator {
 
     func purchase(product: StoreProduct,
                   package: Package?,
-                  completion: @escaping PurchaseCompletedBlock) {
```
MarkVillacampa, line 318: took the opportunity to simplify a bit the purchase methods in `PurchaseOrchestrator` and removed the method to purchase a package without `promotionalOffer`, instead opting to make `promotionalOffer` and `metadata` optional in the method below and handling the routing to SK1 and SK2 appropriately.

MarkVillacampa:
@RCGitBot please test

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -962,13 +966,47 @@ public extension Purchases {
...
     }
 
     func purchase(package: Package) async throws -> PurchaseResultData {
         return try await purchaseAsync(package: package)
     }
 
+    #if ENABLE_PURCHASE_PARAMS
+
+    @objc(purchaseProduct:params:withCompletion:)
+    func purchase(product: StoreProduct, params: PurchaseParams, completion: @escaping PurchaseCompletedBlock) {
```
tonidero, line 983: I was wondering, should the `StoreProduct`/`Package` be part of the `PurchaseParams`? We can create multiple `Builder` constructors, one for each of these, with the mandatory parameters. FWIW, this is how it works in Android.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -962,13 +966,47 @@ public extension Purchases {
...
+                                       completion: completion)
+    }
+
+    func purchase(product: StoreProduct, params: PurchaseParams) async throws -> PurchaseResultData {
+        return try await purchaseAsync(product: product, params: params)
+    }
+
+    @objc(purchasePackage:params:withCompletion:)
+    func purchase(package: Package, params: PurchaseParams, completion: @escaping PurchaseCompletedBlock) {
+        purchasesOrchestrator.purchase(product: package.storeProduct,
```
fire-at-will, line 997: I suppose this would be the best place to switch to the different purchase() functions based on the present subscription offer once we support different offer types in the purchase params?

Just thinking ahead to when we support win-back offers :D

Sources/Purchasing/Purchases/Purchases.swift
MarkVillacampa: I like the idea of simplifying the purchase methods and including the `StoreProduct/Package` be part of the `PurchaseParams`. However I don't think there are any mandatory parameters beyond the product/package itself. All the rest are optional.

Sources/Purchasing/Purchases/PurchaseParams.swift
```diff
@@ -0,0 +1,72 @@
...
+
+    let promotionalOffer: PromotionalOffer?
```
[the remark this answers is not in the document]
MarkVillacampa: We already have [StoreProductDiscount](https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/StoreKitAbstractions/StoreProductDiscount.swift) which wraps StoreKit's [SubcriptionOffer](https://developer.apple.com/documentation/storekit/product/subscriptionoffer). We will probably have to add for the `winBack` type to it when we add manual `winBackOffer` purchasing.

Unfortunately, the `SubscriptionOffer` object does not contain the signed data required to purchase a promo offer. That's why we have [a method](https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/Purchases/PurchasesType.swift#L664) to create a `PromotionalOffer` from a `StoreProductDiscount` and we need a separate `.with(promotionalOffer: promotionalOffer)` method that accepts it.

It would be great if we could have just a `.with(storeProductDiscount: storeProductDiscount)` method that could work for all types of offers tho.

- For intro offers, it would be applied automatically, so we wouldn't need to do anything with it.
- For promo offers, we would generate the signature and perform the purchase, or throw an error if the  signature is not valid or the customer is not eligible.
- For wingback offers, we would have already validated eligibility via `eligibleWinBackOfferIDs`, or otherwise the purchase method will just return an error.

Another option could be just specifying the priority order in which the offers will be applied, so e.g. if you include a winback offer and a promo offer, the winback will be applied and the promo offer will be ignored.

Sources/Purchasing/Purchases/Purchases.swift
tonidero: Note that maybe we will need a third once we support win back offers? cc @fire-at-will. And mostly this allows us to have a single purchase entry point and the rest being implementation details inside the `PurchaseParams`. If there is a reason to want to keep different purchase methods I would be ok, but otherwise I feel that's a cleaner API IMO.

MarkVillacampa: I added an initialiser with product and another with package to `PurchaseParams.Builder` and removed it from the purchase method. Makes it much cleaner and can help us add better validation inside `PurchaseParams` 👍

Sources/Purchasing/Purchases/Purchases.swift
MarkVillacampa: I think we can just add an extra optional parameter to the existing purchase methods in PurchasesOrchestrator like I did here for metadata:

https://github.com/RevenueCat/purchases-ios/pull/4400/files#diff-c6164c2ee507c30293a823c9e66c5951f7dbed017e61a1115aaaf6e95a57d8a5R488

Sources/Networking/Operations/PostReceiptDataOperation.swift
```diff
@@ -271,6 +273,7 @@ extension PostReceiptDataOperation.PostData: Encodable {
         case paywall
         case testReceiptIdentifier = "test_receipt_identifier"
         case appTransaction = "app_transaction"
+        case metadata = "metadata"
```
tonidero, line 276: Hmm nitpick but is it needed to specify the key?
```suggestion
        case metadata
```

Sources/Networking/Responses/CustomerInfoResponse.swift
tonidero: Hmm right... How are we going to handle one time purchases?

Sources/Purchasing/Purchases/PurchaseParams.swift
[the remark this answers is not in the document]
tonidero: Yeah... I think probably logging some error and prioritizing to one of the offers should work... As for @fire-at-will's suggestion, could it work if we just made it a protocol, and have both `PromotionalOffer`, `WinBackOffer` and anything else implement that protocol? Then we would accept either in the `with` function, but only one.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
tonidero: ~~Hmm is this a problem for ObjC? Just making sure we don't add a breaking change there accidentally. But I think it makes sense!~~

Edit: Nvm, it doesn't break the API tests and I think it should be fine 👍

Sources/Purchasing/Purchases/TransactionPoster.swift
```diff
@@ -28,6 +28,7 @@ struct PurchasedTransactionData {
     var presentedOfferingContext: PresentedOfferingContext?
     var presentedPaywall: PaywallEvent?
     var unsyncedAttributes: SubscriberAttribute.Dictionary?
+    var metadata: [String: String]?
```
tonidero, line 31: I've been wondering whether it would make sense to make this a `[String: Any]` or something like that... but that might increase the scope quite a bit, so I think this is ok for now.

tonidero (approved):
LGTM!

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
MarkVillacampa: Yeah this method is private from PurchasesOrchestrator, so it's just an internal refactor.

vegaro (commented):
Even thought it's been merged already, just checked as well and didn't see anything besides Toni's comment on `[String: Any]`, which as discussed can come in a future PR

PR #4404, opened 2024-10-24
Post error test results to Slack

tonidero:
### Description
This adds a slack notification to all CircleCI jobs that error out. Currently posts to the #feed-circleci-ios-failures

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1237,7 +1237,6 @@ jobs:
       - run:
           name: Run Danger
           command: bundle exec danger --verbose
-      - slack-notify
   tag-release-branch:
     docker:
       - image: cimg/ruby:3.1.2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1387,25 +1387,25 @@ workflows:
     jobs:
       - lint:
           context:
-            - slack-secrets
+            - slack-secrets-ios
       - run-test-ios-17:
           context:
-            - slack-secrets
+            - slack-secrets-ios
       - run-test-ios-18:
           context:
-            - slack-secrets
+            - slack-secrets-ios
       - pod-lib-lint:
           context:
-            - slack-secrets
+            - slack-secrets-ios
       - run-revenuecat-ui-ios-17:
           context:
-            - slack-secrets
+            - slack-secrets-ios
       - run-revenuecat-ui-ios-18:
           context:
-            - slack-secrets
+            - slack-secrets-ios
       - emerge_purchases_ui_snapshot_tests:
           context:
-            - slack-secrets
+            - slack-secrets-ios
 
   create-tag:
     when:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -411,6 +417,7 @@ jobs:
           name: SPM Release Build
           command: swift build -c release --target RevenueCat
           no_output_timeout: 30m
+      - slack-notify
```
vegaro, line 420: I would generally prefer these to be part of fastlane, but I think in this case, since there's an orb that gets updates and automatically post failures (right?), I like it here

tonidero:
So I believe this might be quite noisy, but it's working for now so I thought we can merge and tweak as needed.

.circleci/config.yml
```diff
@@ -1200,7 +1236,7 @@ jobs:
       - run:
           name: Run Danger
           command: bundle exec danger --verbose
-
+      - slack-notify
```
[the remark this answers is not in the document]
JayShortway: +1 on `main` only

.circleci/config.yml
```diff
@@ -461,6 +472,7 @@ jobs:
       - run:
           name: Check pods and deployment targets
           command: bundle exec fastlane check_pods
+      - slack-notify
```
vegaro, line 475: This also runs on PRs, we don't need it here

.circleci/config.yml
```diff
@@ -623,6 +641,7 @@ jobs:
       - store_artifacts:
           path: fastlane/test_output
           destination: scan-test-output
+      - slack-notify
```
vegaro, line 644: same about running on every PR

.circleci/config.yml
```diff
@@ -698,7 +718,7 @@ jobs:
       - store_artifacts:
           path: fastlane/test_output/xctest
           destination: scan-test-output
-
+      - slack-notify
```
vegaro, line 721: this also runs on every PR

.circleci/config.yml
```diff
@@ -1056,6 +1086,7 @@ jobs:
           name: Deploy new version
           command: bundle exec fastlane push_revenuecat_pod
           no_output_timeout: 30m
+      - slack-notify
```
vegaro, line 1089: Also runs on PRs

.circleci/config.yml
```diff
@@ -1188,6 +1224,7 @@ jobs:
           path: fastlane/test_output
       - store_artifacts:
           path: fastlane/test_output
+      - slack-notify
```
vegaro, line 1227: we shouldn't run this on `lint`

.circleci/config.yml
```diff
@@ -1214,7 +1250,7 @@ jobs:
       - run:
           name: Tag branch
           command: bundle exec fastlane tag_current_branch
-
+      - slack-notify
```
vegaro, line 1253: same

vegaro (changes requested):

.circleci/config.yml
```diff
@@ -1313,6 +1353,7 @@ jobs:
             git remote set-url origin https://github.com/RevenueCat/purchases-ios-spm.git
             git push origin
             git push --tags
+      - slack-notify
```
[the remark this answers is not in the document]
tonidero: Sounds good, will remove those 👍

.circleci/config.yml
tonidero: Hmm pod-lib-lint is executed also in `main` right? I'm planning to only remove the ones for jobs that are not supposed to run on main never

.circleci/config.yml
tonidero: This actually runs on `main`. I think the chances of merging something breaking this would be low, but better to keep it. Lmk if you think otherwise!

JayShortway (commented):
Instead of adding a `slack-notify` step to every job, can we add a `slack-notify` **job** to certain workflows? That way we can more easily make it notify on `main` only.

We can make the job only run when the test job failed, for instance like so: 

```yaml
- slack-notify:
    requires:
       - lint: failed
```

https://circleci.com/docs/configuration-reference/#requires

tonidero:
Hmm that's interesting... I can try that, thanks! I guess we might lose the data of what jobs failed in Slack though, not sure if it's that important though.

tonidero:
So I was checking that alternative, however, I don't think Slack's orb officially supports that use case... The orb docs have a use case that looks like the current solution: https://circleci.com/developer/orbs/orb/circleci/slack#usage-notify_on_fail_with_template

I saw this: https://support.circleci.com/hc/en-us/articles/360047082992-How-to-send-a-slack-notification-at-end-of-workflow, but that would mean we need to add a job to run on failures as a post_step on each job. The only advantage of that would be that we could call this from the workflow instead of inside each job, but I don't know if it's worth going against the recommendations from the Slack orb for this. Wdyt? @JayShortway

tonidero:
An update here. We haven't had time to dedicate to this yet, so we're planning on merging the approach of a notification as a step of each job, instead of a separate job in the workflow for now. We can revisit this later, when we can.

.circleci/config.yml
```diff
@@ -92,6 +92,13 @@ aliases:
         only: main
 
 commands:
+  slack-notify:
```
JayShortway, line 95: Suggestion to name this `slack-notify-on-fail` to make it clearer when reading the steps list in a job that this is not always executed.

JayShortway (approved):
Great to have this! Just a naming suggestion.

vegaro (approved):

ajpallares (approved):

PR #4405, opened 2024-10-24
Add preview test target

noahsmartin:
### Checklist

### Motivation
Adds a preview snapshot test to locally generate and debug snapshots from previews. This can be used along with EmergeTools snapshot testing whenever you need to debug a snapshot locally.

### Description
Adds the previews snapshot swift package and a new test target that renders all Xcode previews in the app

This also caught a bug with a crashing preview, which I fixed in https://github.com/RevenueCat/purchases-ios/pull/4406


aboedo:
🫶 Emerge

PR #4409, opened 2024-10-25
Adds scaffolding for `StickyFooterComponent`

JayShortway:
## In this PR:
- There's a new component, `StickyFooterComponent`, which is an optional child of `ComponentsConfig`. I think conceptually it makes sense that `StickyFooterComponent` is not part of the root `StackComponent`, because (a) it visually isn't and (b) it would mean we'd have to loop over the root stack to find it, or have some sort of easily-broken convention that it's always the last item in the root stack.
- There's a new ViewModel+View combo, `RootViewModel`+`RootView`, which is the parent of the root `StackComponentView[Model]` and an optional `StickyFooterView[Model]`. It is not called `RootComponentView[Model]`, because there's no `RootComponent`. More on that below. 
  
## Please comment on:
- I initially had a new `RootComponent` as only child of `ComponentsConfig`, but I thought we might want to avoid the extra nesting in `ComponentsConfig`. So now we have a `RootViewModel` and `RootView` which are slightly different from the other View(Model)s in that they don't have a corresponding component. I removed the `RootComponent` in a separate commit, so I can easily add it back. Let me know your thoughts! 
- `StickyFooterComponent`'s name is not final. Open to suggestions. I wanted to avoid the use of "PaywallFooter" though, because that name already means something else.


Sources/Paywalls/Components/PaywallStickyFooterComponent.swift
```diff
@@ -0,0 +1,36 @@
...
+//
+// swiftlint:disable missing_docs
+
+import Foundation
+
+#if PAYWALL_COMPONENTS
+
+public extension PaywallComponent {
+
+    struct StickyFooterComponent: PaywallComponentBase {
```
JayShortway, line 22: I'm now starting to think whether we actually need this, or if we can get away with just using the existing `StackComponent`. Are there any properties that a sticky footer needs that a stack doesn't? Maybe a shadow? But one could argue that a regular stack can also have a shadow.

JayShortway: Oh! Maybe this can be a `StickyComponent` with and `edge` property? That way we can use it for a sticky header as well (and sticky sidepanel if we _ever_ want that).

joshdholtz: Ohhhh... this is a great question!

I think it might actually be almost entirely a stack but... It could be beneficial to keep as its own object if we do need to ever add anything specific (like translucent or a drawer type of functionality or something 🤷‍♂️

Thoughts?

JayShortway: Yea it's definitely safer to keep it separate like this, to be able to add any properties we might think of in the future. The JSON will get more nested, but I'm not super worried about that haha. Dunno 'bout you?

joshdholtz (approved):
This looks good to me 😁 :shipit:

PR #4410, opened 2024-10-25
Refactor `SubscriptionDetailsView` and better `WrongPlatformView`

vegaro:
| Header | Header |
|--------|--------|
| Play Store | Screenshot 2024-10-31 at 13 39 30 |
| RCBilling | Screenshot 2024-10-31 at 13 40 00 |
| Stripe |  Screenshot 2024-10-31 at 13 40 24|
| External | Screenshot 2024-10-31 at 13 40 48 |
| Promotional | Screenshot 2024-10-31 at 13 41 12 | 
| Promotional Lifetime | Screenshot 2024-11-05 at 16 08 18 |
| Amazon | Screenshot 2024-10-31 at 13 42 21 |


RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
```diff
@@ -14,33 +14,178 @@
...
 
-    init(title: String,
-         durationTitle: String,
-         price: String,
-         expirationDateString: String?,
+    let store: Store?
+
+    init(title: String?,
+         explanation: Explanation?,
+         durationTitle: String?,
```
JayShortway, line 36: What does it mean if these are all nil? Are there any constraints or is every combination of nil possible? (E.g. only the title is nil, both the title and explanation are nil, only the explanation is nil, etc.)

If there are constraints, maybe we can make constructors that force those constraints?

RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
```diff
@@ -14,33 +14,178 @@
...
         self.willRenew = willRenew
         self.active = active
+        self.store = store
+    }
+
+    init?(entitlement: EntitlementInfo,
+          subscribedProduct: StoreProduct? = nil) {
+        // swiftlint:disable:next todo
+        // TODO: support non-consumables
+        let dateFormatter = DateFormatter()
```
JayShortway, line 59: Maybe this can be an optional parameter (defaulted to `DateFormatter()`), or maybe the locale can be a parameter? I've noticed that some tests fail locally for me because my machine is not set to `en-US`. If we allow the test to specify a locale, the machine's locale has no influence on test results.

RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
```diff
@@ -14,33 +14,178 @@
...
+        case expired
+        case lifetime
+    }
+
+}
+// swiftlint:enable nesting
+
+fileprivate extension EntitlementInfo {
+
+    func priceBestGuess(product: StoreProduct?) -> SubscriptionInformation.PriceDetails {
```
JayShortway, line 108: Minor, maybe `BestGuess` can be `BestEffort`?

RevenueCatUI/CustomerCenter/Views/SubscriptionDetailsView.swift
```diff
@@ -24,114 +24,161 @@ import SwiftUI
...
+                        return nil
                     }
+                }()
+
+                if let price = priceValue {
+                    IconLabelView(
+                        iconName: "coloncurrencysign",
+                        label: localization.commonLocalizedString(for: .currentPrice),
+                        value: price
+                    )
```
JayShortway, line 65: I think this makes sense!

JayShortway (commented):
Did a quick initial pass as requested. Looking good imo! Added some questions and minor suggestions.

RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
vegaro: that's a good point. I will write previews and see if there are some common constraints

RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
vegaro: right good point

tonidero (approved):
LGTM! Great job on this!

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -160,11 +166,26 @@ public struct CustomerCenterConfigData {
                 case .pleaseContactSupportToManage:
                     return "Please contact support to manage your subscription."
                 case .appleSubscriptionManage:
-                    return "You can manage your subscription by using the App Store app on an Apple device."
+                    return "You have an active Apple App Store purchase. " +
```
dpannasch, line 169:
```suggestion
                    return "You have an active subscription from the Apple App Store. " +
```

dpannasch: Not sure if there's a case for a "purchase" being active that isn't a subscription that I'm not thinking of. "Active purchase" struck me funny because I don't know what "active" means in the context of a purchase. If "purchase" is necessary, I'd consider something more descriptive, like:

> You have a purchase from the Apple App Store that's still granting you access to this service.

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -160,11 +166,26 @@ public struct CustomerCenterConfigData {
...
+                    "You can manage your subscription by using the App Store app on an Apple device."
                 case .googleSubscriptionManage:
-                    return "You can manage your subscription by using the Play Store app on an Android device"
+                    return "You have an active Google Play Store purchase"
```
dpannasch, line 172:
```suggestion
                    return "You have an active subscription from the Google Play Store"
```

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -160,11 +166,26 @@ public struct CustomerCenterConfigData {
...
                 case .amazonSubscriptionManage:
-                    return "You can manage your subscription in the Amazon Appstore app on an Amazon device."
+                    return "You have an active Amazon Appstore purchase. " +
```
dpannasch, line 174:
```suggestion
                    return "You have an active subscription from the Amazon Appstore. " +
```

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -160,11 +166,26 @@ public struct CustomerCenterConfigData {
...
+                    "You can manage your subscription in the Amazon Appstore app."
+                case .webSubscriptionManage:
+                    return "You are subscribed to a web subscription." +
```
dpannasch, line 177:
```suggestion
                    return "You have an active subscription that was created on the web." +
```

Sources/CustomerCenter/CustomerCenterConfigData.swift
vegaro: I was using purchase to cover consumables/lifetimes, but I was wrong becasue this view won't open if there are no subscriptions, so this should be subscription

PR #4412, opened 2024-10-25
Adds long sample paywall with sticky footer to PaywallsTester

JayShortway:
As the title says. This actually surfaces 2 issues: 
- The sticky footer doesn't draw in the safe area. (See attached image.)
- The image component seems to reload every time it's scrolled back into view. 

These issues are addressed in these follow-up PRs:
- https://github.com/RevenueCat/purchases-ios/pull/4422
- https://github.com/RevenueCat/purchases-ios/pull/4423


joshdholtz (approved):
:shipit: :shipit: :shipit:

PR #4413, opened 2024-10-26
Remove PackageGroup

joshdholtz:
### Motivation

Remove unneeded package group from paywall components

### Description

`PackageGroupComponent` was not really needed. It was an extra type of component to think of that didn't provide much value over a `StackComponent`.

- Removed `PackageGroupComponent` (and its view and view model)
- `PackageComponent` now has boolean if its default selected
- `PackageComponentViewModel` now finds the `Package` from the package identifier
- New `PackageValidator` class
  - Injected into recursive view model creator and collects `PackageComponentViewModel`s all the package views
      - Previously this was done in `PackageGroupViewModel` however, `PackageComponent`s can live _anywhere_ in the paywall
      - Injecting this clas into the whole process is a way to collect all the view models without recursively finding them again later
  - Validate packages (that at least one exists)
  - Finds the default selected package


RevenueCatUI/Templates/Components/PaywallComponentViewModel.swift
```diff
@@ -29,6 +28,7 @@ enum PaywallComponentViewModel {
 extension PaywallComponent {
 
     func toViewModel(
+        packageValidator: PackageValidator,
```
tonidero, line 31: Hmm instead of passing the validator around the view models, would it be possible to perform an initial validation before starting to create views/viewmodels? I think it would be cleaner if after we can just assume it's valid. (There might be a reason why that's difficult not really sure, so FFTI).

tonidero (commented):
LGTM! just a suggestion... Also leaving it to someone with more context on the approach to approve

RevenueCatUI/Templates/Components/PaywallComponentViewModel.swift
joshdholtz: @tonidero This is what I _wanted_ to do however, it would required traversing the component tree twice (since there is no standard placement for packages)

So I _figured_ if we were already traversing/remapping the components to view models, might as well pull out the information we need at the same time?

It feels gross from a code aspect but felt better in performance? 🤷‍♂️

RevenueCatUI/Templates/Components/Packages/Package/PackageComponentViewModel.swift
```diff
@@ -23,19 +23,26 @@ class PackageComponentViewModel {
     private let component: PaywallComponent.PackageComponent
     private let offering: Offering
 
-    let package: Package
+    let isDefaultSelected: Bool
```
JayShortway, line 26: Could this be named `isSelectedByDefault`?

RevenueCatUI/Templates/Components/PackageValidator.swift
```diff
@@ -0,0 +1,59 @@
...
+import RevenueCat
+
+#if PAYWALL_COMPONENTS
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+class PackageValidator {
+
+    private var packageViewModels: [PackageComponentViewModel] = []
+
+    typealias PackageInfo = (package: Package, isDefaultSelected: Bool)
```
JayShortway, line 24: Same here, could this be:
```diff
-    typealias PackageInfo = (package: Package, isDefaultSelected: Bool)
+    typealias PackageInfo = (package: Package, isSelectedByDefault: Bool)
```

RevenueCatUI/Templates/Components/PackageValidator.swift
```diff
@@ -0,0 +1,59 @@
...
+            guard let package = packageViewModel.package else {
+                return nil
+            }
+            return (package: package, isDefaultSelected: packageViewModel.isDefaultSelected)
+        }
+    }
+
+    func add(_ packageViewModel: PackageComponentViewModel) {
+        self.packageViewModels.append(packageViewModel)
+    }
```
JayShortway, line 37: Instead of keeping hold of the entire ViewModel, can this function just take a `Package` and an `isSelectedByDefault: Bool`? We don't know how big the ViewModel can grow in the future, and all we need is some data (in the form of `Package` + `Bool`).

RevenueCatUI/Templates/Components/PaywallComponentViewModel.swift
JayShortway, line 31: This can get a default value of `PackageValidator()` so it _almost_ becomes an implementation detail of the `toViewModel()` function. This is a usual pattern I think with recursive functions, where the top level call gets a default argument, which it then passes down to every recursive call.

JayShortway: Maybe we can abstract this away somehow. What about a wrapper function, e.g. `buildInitialState()`, that wraps `toViewModel()` and calls `isValid` at the end, and returns the root VIewModel + `PaywallState`?

JayShortway (commented):
I like it! Had some comments, mostly around ease of understanding.

RevenueCatUI/Templates/Components/PaywallComponentViewModel.swift
joshdholtz: @JayShortway I _did_ think about this but I almost don't want to because it really needs to accept that single high level instance of `PackageValidator`. I want to make sure that its not accidentally skipped when somebody creates a new view model 😬 

I'm kind of keeping it here for somebody adds another container component and needs to pass `packageValidator` into it and it causes a compile error if its missing 🤷‍♂️ Hopefully that makes sense 😅

joshdholtz: > Maybe we can abstract this away somehow. What about a wrapper function, e.g. `buildInitialState()`, that wraps `toViewModel()` and calls `isValid` at the end, and returns the root VIewModel + `PaywallState`?

I dooooo like an idea of this! I think I'd like to come back to that a little later though 🤷‍♂️  Once we have things out and have a little more time to breath? Or until this absolutely feels horrible? 😛

JayShortway: Yea works for me! I can always get behind a lil more explicitness.

JayShortway: Yea it's not that much in the way, for sure! I'm okay with keeping it around for now.

JayShortway (approved):
Nice nice!

PR #4415, opened 2024-10-27
PurchaseButtonComponent is now just a container/stack like ButtonComponent

joshdholtz:
### Motivation

Make `PurchaseButtonComponent` more flexible by just letting it be a stack (like `ButtonComponent`)

### Description

Removed all properties and just gave it a `stack: StackComponent` property


RevenueCatUI/Templates/Components/Button/ButtonComponentView.swift
```diff
@@ -35,10 +35,12 @@ struct ButtonComponentView: View {
...
     var body: some View {
-        AsyncButton(
-            action: { try await performAction() },
-            label: { StackComponentView(viewModel: viewModel.stackViewModel, onDismiss: self.onDismiss) }
-        )
+        AsyncButton {
+            try await performAction()
+        } label: {
+            // Not passing an onDismiss - nothing in this stack should be able to dismiss
+            StackComponentView(viewModel: viewModel.stackViewModel, onDismiss: {})
```
JayShortway, line 42: I don't think this is true? There's a dismiss button action (`navigateBack`).

JayShortway (commented):
LGTM! 1 comment to double check.

RevenueCatUI/Templates/Components/Button/ButtonComponentView.swift
joshdholtz: Fixed!

JayShortway (approved):
Let's go!

PR #4417, opened 2024-10-27
Apply state and conditions ONLY for text component

joshdholtz:
## Motivation

Apply overridden properties for states (selected and intro) and conditions (landscape, tablet, and desktop)

## Description

⚠️ This ended up being a doozy of a change 😬 The API response part was already split up into #4414
⚠️ Some of this could end up being refactored and reused when applying this same change for Image and Stack

### State

The state is something based off of the data in the paywall. This is something that can allow overridden styles to appear.

- `selected`
  - The package the user selected and needs to have different highlighted colors
- `intro_offer`
  - If the package has an intro offer and needs to have different text

### Conditions

These are device/screen conditions that will make a paywall look different/better for that device. Example: A tablet or desktop might have bigger text sizes

The default paywall is meant for mobile. The other conditions are:
- Mobile landscape
- Tablet
- Tablet landscape
- Desktop

Desktop is applied on top of tablet landscape which is applied on top of tablet which is applied on top of tablet which is applied on top of mobile landscape which is applied on top of mobile. This means that desktop takes priority if there is anything overridden there BUT it can also reuse any other conditions below it.

### The Code

There is a new struct called `TextComponentStyle`(see [here](https://github.com/RevenueCat/purchases-ios/pull/4417/files?diff=unified&w=1#diff-16e747a589703d58e4b5475c49afc713f33e39ab8198bb591b6bc383fd0bbdb4R256-R291)) which is what gets applied to the `TextComponentView` in a new `style()` view builder function (see [here](https://github.com/RevenueCat/purchases-ios/pull/4417/files?diff=unified&w=1#diff-9315e31a6237830130bd9c111e918d2e13c737ae2431661f2e81285c0bb0ea04R36)).

`TextComponentStyle` has base of all the properties of `PaywallComponent.TextComponent` but then gets applied partials of `PaywallComponent.PartialTextComponent` on top of it for all the conditions and states that are applicable. 

```swift
// Small example
var body: some View {
    viewModel.styles(
        state: self.componentViewState,
        condition: self.screenCondition
    ) { style in
        Text(style.text)
            .font(style.textStyle)
            .fontWeight(style.fontWeight)
            .multilineTextAlignment(style.horizontalAlignment)
            .foregroundStyle(style.color)
    }
}
```

### Demo

The (ugly) example shows two texts. The first will change its text when switching between portrait and landscape. The second will disappear in landscape.

https://github.com/user-attachments/assets/e2d12509-158a-4410-a800-58dbd99b8326


RevenueCatUI/Templates/Components/Packages/Package/PackageComponentView.swift
```diff
@@ -35,6 +44,7 @@ struct PackageComponentView: View {
                     viewModel: self.viewModel.stackViewModel,
                     onDismiss: self.onDismiss
                 )
+                .environment(\.componentViewState, componentViewState)
```
joshdholtz, line 47: This is the important part. This is what tells all children views of the package that they are selected so they can be applied the new styles.

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
```diff
@@ -88,6 +94,7 @@ struct TemplateComponentsView: View {
         .frame(maxHeight: .infinity, alignment: .topLeading)
         .edgesIgnoringSafeArea(.top)
         .environmentObject(self.paywallState)
+        .environment(\.screenCondition, self.screenConditionObserver.conditionType)
```
joshdholtz, line 97: This is the important change here.

RevenueCatUI/Templates/Components/Text/TextComponentView.swift
```diff
@@ -20,22 +20,39 @@ import SwiftUI
...
-            .fixedSize(horizontal: false, vertical: true)
-            .multilineTextAlignment(viewModel.horizontalAlignment)
-            .foregroundStyle(viewModel.color)
-            .padding(viewModel.padding)
-            .background(viewModel.backgroundColor)
-            .padding(viewModel.margin)
+        viewModel.styles(
+            state: self.componentViewState,
+            condition: self.screenCondition
+        ) { style in
```
joshdholtz, line 39: This is a new `@ViewBuilder` on the view model that creates a `TextComponentStyle` object to use for styling the component 👇

Sources/Paywalls/Components/PaywallImageComponent.swift
joshdholtz, line 1: Nothing is being done with these yet but it felt right to all add the partial onto image component

Sources/Paywalls/Components/PaywallStackComponent.swift
joshdholtz, line 1: Nothing is being done with these yet but it felt right to all add the partial onto stack component to see how it looked.

Sources/Paywalls/Components/PaywallTextComponent.swift
```diff
@@ -64,6 +109,26 @@ extension PaywallComponent.TextComponent {
...
+        case visible
+        case text = "text_lid"
+        case fontFamily
+        case fontWeight
+        case color
+        case textStyle
+        case horizontalAlignment
+        case backgroundColor
+        case padding
+        case margin
```
joshdholtz, line 131: This was only necessary because we use `text` instead of `testLid` 🤷‍♂️

Sources/Paywalls/Components/PaywallImageComponent.swift
joshdholtz, line 1: Nothing is being done with these yet but it felt right to all add the partial onto image component to see how it looked.

RevenueCatUI/Templates/Components/LocalizedPartials.swift
```diff
@@ -0,0 +1,36 @@
...
+    let introOffer: T?
+
+}
+
+struct LocalizedConditions<T: LocalizedPartial> {
+
+    let mobileLandscape: T?
+    let tablet: T?
+    let tabletLandscape: T?
+    let desktop: T?
```
JayShortway, line 32: Same comments as https://github.com/RevenueCat/purchases-ios/pull/4414#discussion_r1819385459.

RevenueCatUI/Templates/Components/LocalizedPartials.swift
```diff
@@ -0,0 +1,36 @@
...
+import Foundation
+
+#if PAYWALL_COMPONENTS
+
+protocol LocalizedPartial {}
+
+struct LocalizedStates<T: LocalizedPartial> {
+
+    let selected: T?
+    let introOffer: T?
```
JayShortway, line 23: I don't think these are the same type of states 🤔 `introOffer` is really only ever applicable to packages, but `selected` can be applied to things like checkboxes and radiobuttons (which we don't have _now_).

I think `selected` state should be part of a more general system that allows any component to change their looks based upon it. To me, it falls into the same category as `hovered` (for any component) or `pressed` (for buttons). `introOffer` is not the same thing to me.

RevenueCatUI/Templates/Components/Packages/Package/ComponentViewState.swift
```diff
@@ -0,0 +1,39 @@
...
+//
+//  Created by Josh Holtz on 10/26/24.
+
+import Foundation
+import SwiftUI
+
+#if PAYWALL_COMPONENTS
+
+enum ComponentViewState {
+    case normal
```
JayShortway, line 20: `default`? Or is that a reserved keyword in swift?

JayShortway: Maybe `standard` then? `normal` is a bit vague to me haha.

RevenueCatUI/Templates/Components/ScreenConditionObserver.swift
```diff
@@ -0,0 +1,87 @@
...
+            deviceType = .phone
+        case .pad:
+            deviceType = .tablet
+        default:
+            deviceType = .phone
+        }
+        #else
+        // Set device type for macOS
+        deviceType = .desktop
+        #endif
```
JayShortway, line 48: Does `userInterfaceIdiom` take into account the window size? I suppose iOS apps on macOS are resizable? Also apps don't need to run in full screen on iPads if I'm not mistaken. The device and orientation is not enough to determine how big our window is.

RevenueCatUI/Templates/Components/ScreenConditionObserver.swift
```diff
@@ -0,0 +1,87 @@
...
+
+enum ScreenCondition {
+    case `default`, mobileLandscape, tablet, tabletLandscape, desktop
+}
+
+class ScreenConditionObserver: ObservableObject {
+
+    private enum Device {
+        case phone, tablet, desktop
+    }
```
JayShortway, line 27: Same comment as https://github.com/RevenueCat/purchases-ios/pull/4414#discussion_r1819385459. Just tagging it on here for tracking purposes.

RevenueCatUI/Templates/Components/TemplateComponentsView.swift
```diff
@@ -34,13 +34,19 @@ enum PackageGroupValidationError: Error {
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 struct TemplateComponentsView: View {
 
-    let paywallComponentsData: PaywallComponentsData
-    let componentViewModel: PaywallComponentViewModel
-    private let onDismiss: () -> Void
+    @Environment(\.userInterfaceIdiom)
+    private var userInterfaceIdiom
```
JayShortway, line 38: Just for my understanding, why does it need this? Wouldn't the `screenCondition` be enough?

RevenueCatUI/Templates/Components/Text/TextComponentView.swift
JayShortway: I love it!

RevenueCatUI/Templates/Components/Text/TextComponentViewModel.swift
```diff
@@ -17,51 +17,240 @@ import SwiftUI
...
+
     }
 
-    var fontWeight: Font.Weight {
-        component.fontWeight.fontWeight
+    @ViewBuilder
+    func styles(
+        state: ComponentViewState,
+        condition: ScreenCondition,
+        application: @escaping (TextComponentStyle) -> some View
```
JayShortway, line 77: Naming suggestion: `apply`. `application` makes me think of a computer program haha, _apps_ as the youth say.

Sources/Paywalls/Components/Common/PartialComponent.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    struct ComponentState<T: PartialComponent>: PaywallComponentBase {
+
+        public init(selected: T?, introOffer: T?) {
+            self.selected = selected
+            self.introOffer = introOffer
+        }
+
+        public let selected: T?
+        public let introOffer: T?
```
JayShortway, line 36: Same comment as on `LocalizedState`.

Sources/Paywalls/Components/Common/PartialComponent.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    }
+
+    enum ComponentConditionsType {
+        case mobileLandscape, tablet, tabletLandscape, desktop
+    }
```
JayShortway, line 42: Same comment on the devices as https://github.com/RevenueCat/purchases-ios/pull/4414#discussion_r1819385459.

JayShortway (commented):
Some great stuff in here! I really like the new ViewBuilder and the `TextComponentStyle` approach. Had a few comments and questions.

RevenueCatUI/Templates/Components/Text/TextComponentViewModel.swift
joshdholtz: Honestly... this was totally meant to be `apply` 😅 I think I had computer programs on my mind 🤷‍♂️

RevenueCatUI/Templates/Components/Text/TextComponentViewModel.swift
```diff
@@ -16,50 +16,226 @@ import SwiftUI
...
-    var padding: EdgeInsets {
-        component.padding.edgeInsets
+}
+
+extension PaywallComponent.PartialTextComponent {
+
+    var isEmpty: Bool {
+        return visible == nil && text == nil && fontFamily == nil && fontWeight == nil && color == nil &&
+               backgroundColor == nil && padding == nil && margin == nil &&
+               textStyle == nil && horizontalAlignment == nil
```
JayShortway, line 198: This might be hard to keep up to date if we ever add properties to `PartialTextComponent`. Maybe this should not be an extension but just a member of `PartialTextComponent`, so it's closer to the actual properties? For good measure, it might be good to also add a comment to `PartialTextComponent` instructing to update `isEmpty` if any properties are added.

JayShortway (commented):
Amazing! We're this 🤏 close imo. 3 minor unresolved comments left that I'd like to hear your thoughts on.

RevenueCatUI/Templates/Components/Text/TextComponentViewModel.swift
joshdholtz: Ahhh yes! Will see if we can move this and might also be able to add some unit tests to validate it gets update 🤷‍♂️ 

I did this with new properties on full components and partials and it caught the new `shadow` property 😁 👇 

Screenshot 2024-10-31 at 9 34 33 PM

JayShortway: Oh that's very nice!

JayShortway (approved):
Nice nice nice! 🎉 Thanks for making the changes 🙏

PR #4422, opened 2024-10-29
Fixes the sticky footer not drawing in the bottom safe area.

JayShortway:
As the title says. Here's a comparison:

| Before | After |
|--------|-------|
| Simulator Screenshot - before | Simulator Screenshot - after |


RevenueCatUI/Templates/Components/Root/RootView.swift
```diff
@@ -31,13 +33,61 @@ struct RootView: View {
...
+                            trailing: 0
+                        )
+                    )
+                }
+                // First we ensure our footer draws in the bottom safe area. Then we add additional padding, so its
+                // background shows in that same bottom safe area.
+                .ignoresSafeArea(edges: .bottom)
+                .onBottomSafeAreaPaddingChange { bottomPadding in
+                    self.additionalFooterPaddingBottom = bottomPadding
+                }
```
JayShortway, line 54: I thought about "filling up" the bottom safe area with the same background color (`stackViewModel.backgroundColor`) as set on the sticky footer using [`background(_:ignoressafeareaedges:)`](https://developer.apple.com/documentation/swiftui/view/background(_:ignoressafeareaedges:)), but  that won't work if we ever support background gradients or background images. So I went with the additional padding approach instead.

joshdholtz (changes requested):

PR #4423, opened 2024-10-29
Fixes reloading paywall images after they've been scrolled off screen

JayShortway:
As the title says. This showed up when testing the sticky footer component, but doesn't have anything to do with sticky footers. As you can see in the **before** video, the image keeps reloading, making it difficult for the user to scroll up.

This is fixed by the `ImageLoader` checking whether it has received a new URL before actually resetting its published state and loading an image. If it receives the same URL twice, it will ignore it the second time, keeping its published state the same. 

## Before
https://github.com/user-attachments/assets/1fe83961-6c28-454c-bc10-62d738958dfc

## After
https://github.com/user-attachments/assets/bb5712d3-edf3-44aa-9cc0-f4e4bf8832e9


joshdholtz (approved):
Ohhhhh, yeah, this is good! I ran into this in an image heavy Paywalls V2 example so thank you sooooo much for solving this 🫶

PR #4429, opened 2024-10-31
The StackComponent has an optional shadow

JayShortway:
As the title says. I was initially going to build this just for the sticky footer, but I noticed on Figma that any stack can have a shadow, hence this PR.

Applied to the sticky footer, it looks like this:

### DIFF
--- a/RevenueCatUI/Templates/Components/Stack/StackComponentView.swift
+++ b/RevenueCatUI/Templates/Components/Stack/StackComponentView.swift
@@ -56,6 +56,10 @@ struct StackComponentView: View {
         .background(viewModel.backgroundColor)
         .cornerBorder(border: viewModel.border,
                       radiuses: viewModel.cornerRadiuses)
+        .applyIfLet(viewModel.shadow) { view, shadow in
+            // Without compositingGroup(), the shadow is applied to the stack's children as well.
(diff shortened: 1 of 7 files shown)


RevenueCatUI/Templates/Components/Stack/StackComponentView.swift
```diff
@@ -56,6 +56,10 @@ struct StackComponentView: View {
         .background(viewModel.backgroundColor)
         .cornerBorder(border: viewModel.border,
                       radiuses: viewModel.cornerRadiuses)
+        .applyIfLet(viewModel.shadow) { view, shadow in
+            // Without compositingGroup(), the shadow is applied to the stack's children as well.
+            view.compositingGroup().shadow(shadow: shadow)
```
JayShortway, line 61: The reason for not adding `compositingGroup()` directly to the new `shadow()` modifier is that I wanted the latter to behave the same as the built-in one to avoid confusion. Let me know if you disagree!

joshdholtz: TIL about `compositingGroup()` 😅

joshdholtz (approved):
This looks so good 😍

PR #4431, opened 2024-10-31
Support fetching eligible win-back offers for a product

fire-at-will:
### Description
This PR adds the ability for developers to fetch the win-back offers that a subscriber is eligible for on a given product.

### New APIs
All API additions in this PR are marked as `internal` for now. They will be made public once we introduce the ability to redeem a win-back offer in a future PR.

The PR introduces a new WinBackOffer type to represent a win-back offer:
```swift
@available(iOS 18.0, macOS 15.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
@objc(RCWinBackOffer)
internal final class WinBackOffer: NSObject, Sendable {

    /// The ``StoreProductDiscount`` in this offer.
    @objc internal let discount: StoreProductDiscount

    init(discount: StoreProductDiscount) {
        self.discount = discount
    }

}
```

Additionally, the PR introduces two new functions on the `Purchases` singleton for fetching the win-back offers for a given product that a subscriber is eligible for. These functions are not available when the `ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION` compiler flag is enabled.
```swift
@available(iOS 18.0, macOS 15.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
 func eligibleWinBackOffers(
     forProduct product: StoreProduct
 ) async throws -> [WinBackOffer]

@available(iOS 18.0, macOS 15.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
 func eligibleWinBackOffers(
     forProduct product: StoreProduct,
     completion: @escaping @Sendable (Result<[WinBackOffer], PublicError>) -> Void
 )
```

### Note
Since we're keeping this functionality as internal for now, this PR is marked with the `pr:other` tag so that it doesn't get marked as a new feature in the auto-generated changelogs.


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1923,3 +1932,51 @@ private extension Purchases {
...
+     * Returns the win-back offers that the subscriber is eliglble for on the provided product.
+     *
+     * - Parameter product: The product to check for eliglble win-back offers.
+     * - Parameter completion: A completion block that is called with the eligible win-back
+     * offers for the provided product.
+     * - Important: Win-back offers are only supported when the SDK is running with StoreKit 2 enabled.
+     */
+    internal func eligibleWinBackOffers(
+        forProduct product: StoreProduct,
+        completion: @escaping @Sendable (Result<[WinBackOffer], PublicError>) -> Void
```
fire-at-will, line 1964: completion is marked as `@Sendable` to avoid warnings below when we dispatch the completion handler on the main actor

Tests/StoreKitUnitTests/TestHelpers/AvailabilityChecks.swift
```diff
@@ -44,7 +44,13 @@ enum AvailabilityChecks {
     }
 
     static func iOS17APIAvailableOrSkipTest() throws {
-        guard #available(iOS 17.0, tvOS 17.0, macOS 14.0, watchOS 10.0, *) else {
+        guard #available(iOS 17.0, tvOS 17.0, macOS 14.0, watchOS 10.0, visionOS 1.0, *) else {
```
fire-at-will, line 47: added visionOS check

Sources/Purchasing/StoreKit2/Win-Back Offers/WinBackOfferEligibilityCalculator.swift
```diff
@@ -0,0 +1,135 @@
...
+            switch $0.transaction {
+            case .unverified:
+                return false
+            case .verified(let transaction):
+                // Intentionally exclude transactions acquired through family sharing
+                return transaction.ownershipType == .purchased
+            }
+        })
+
+        let renewalInfos: [Product.SubscriptionInfo.RenewalInfo] = purchasedSubscriptionStatuses.compactMap({
```
MarkVillacampa, line 107: maybe we could add a little helper like we have for transactions:

https://github.com/RevenueCat/purchases-ios/blob/bcef02adff94f0aed2f72f4cb48f64d8a410b6cd/Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift#L156

Sources/Purchasing/StoreKit2/Win-Back Offers/WinBackOfferEligibilityCalculator.swift
```diff
@@ -0,0 +1,135 @@
...
+            return []
+        }
+
+        let purchasedSubscriptionStatuses = statuses.filter({
+            switch $0.transaction {
+            case .unverified:
+                return false
+            case .verified(let transaction):
+                // Intentionally exclude transactions acquired through family sharing
+                return transaction.ownershipType == .purchased
```
MarkVillacampa, line 103: Technically, you can only be subscribed to a product once, so there must be at most 1 renewalInfo instance that is NOT a family shared one. So maybe we could simplify the logic of the rest of the function by just finding the first status transaction with `transaction.ownershipType == .purchased`, like Apple does in their sample code:

```
let purchasedStatus = statuses.first {
    $0.transaction.unsafePayloadValue.ownershipType == .purchased
}
```

https://developer.apple.com/videos/play/wwdc2024/10110/

Sources/Purchasing/StoreKit2/Win-Back Offers/WinBackOfferEligibilityCalculator.swift
```diff
@@ -0,0 +1,135 @@
...
+            $0.eligibleWinBackOfferIDs
+        })
+
+        // Flatten the win-back offer IDs we've received for all of the renewalInfos while removing duplicates
+        let eligibleWinBackOfferIDs: [String] = {
+            var seen = Set<String>()
+            return eligibleWinBackOfferIDsPerRenewalInfo
+                .flatMap { $0 }
+                .filter { seen.insert($0).inserted }
+        }()
```
MarkVillacampa, line 127: By using just one status above, we can get rid of all the deduplication logic here.

MarkVillacampa (changes requested):
This is looking really really good!! Great work!

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1923,3 +1932,51 @@ private extension Purchases {
...
 
 }
+
+// MARK: - Win-Back Offers
+#if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+@available(iOS 18.0, macOS 15.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
+extension Purchases {
+
+    /**
+     * Returns the win-back offers that the subscriber is eliglble for on the provided product.
```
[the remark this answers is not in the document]
fire-at-will: AHHHH Mark you have no idea how many times I made this typo while working on this PR 🙈

Sources/Purchasing/StoreKit2/Win-Back Offers/WinBackOfferEligibilityCalculator.swift
fire-at-will: Love this, added in 1a5630e3c827d18e2250939ba8fb2c4237e5f9ec

Sources/Purchasing/StoreKit2/Win-Back Offers/WinBackOfferEligibilityCalculator.swift
fire-at-will: Yeah, this is where I thought I might be able to do a bit better than Apple's example in the event where there were multiple renewalInfos.

Is there any chance that there would be more than one renewalInfo for a single product ID if a user subscribes, churns, and then re-subscribed? I'm not seeing many details about this in the docs.

I've implemented this change [here](https://github.com/RevenueCat/purchases-ios/pull/4435) and will merge it in to this PR if we think we're safe in the case I mentioned above. 👍

MarkVillacampa: The documentation does not explicitly state that, but it says:

> The array can have more than one subscription status if your subscription supports Family Sharing.

Which I'd argue highly suggest the array can normally only have one status (the purchased one) and potentially more only if it supports family sharing.

Given that, plus Apple's sample code makes me very confident there will always be just one status with `ownershipType == .purchased`

https://developer.apple.com/documentation/storekit/product/subscriptioninfo/3822296-status

MarkVillacampa (approved):

PR #4434, opened 2024-11-01
CI: only store the final report file as test result.

JayShortway:
## Motivation
Sometimes CircleCI reports failed tests for a successful job. We deduced that this happens because all report files are uploaded as test results, including reports of test runs that were later retried. This causes CircleCI to parse that earlier report and show a failed test, even though the test eventually passed.

## Changes introduced
This PR changes the `store_test_results` command where applicable, to only pass the final report file. The entire `test_output` directory is still uploaded as an artifact, to help with debugging actual failures.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -263,11 +263,12 @@ commands:
           name: Run tests
           working_directory: << parameters.directory >>
           command: bundle exec fastlane scan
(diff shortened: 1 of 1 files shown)


JayShortway:
@RCGitBot please test

PR #4435, opened 2024-11-01
Only use first renewal info when calculating win-back eligibility

fire-at-will:
### Checklist


MarkVillacampa (approved):
🚀

PR #4436, opened 2024-11-01
Fixes for paywalls v2 renderer after testing some real life paywalls

joshdholtz:
## Motivation

Fixes for paywall tester for Paywalls V2

## Description

- Added `@frozen` to all the enums to get rid of Swift 6 warnings
- New `FlexHStack` to support horizontal `distribution`
- Fixed a width bug with `LazyVStack`
  - `LazyVStack` is used to make rendering a bit snappier feeling but it has issues with width sometimes
  - It was behaving as `maxWidth: .infinity` so added some naive logic to only use `LazyVStack` if more than 3 children. Ideally this should look recursively down  
- Temporarily removed the error in the paywall if there are no packages (because it makes initial testing of paywall designs hard)
  - Could maybe only show a warning when in debug
- Fixed a bunch of `Codable` things that were failing when testing real life paywalls

### Demo

| Blinkist | HabitKit |
|--------|--------|
| Simulator Screenshot - iPhone 16 Pro - 2024-11-01 at 15 56 46 | Simulator Screenshot - iPhone 16 Pro - 2024-11-01 at 17 46 22 |

### DIFF
--- a/Sources/Paywalls/Components/Common/Dimension.swift
+++ b/Sources/Paywalls/Components/Common/Dimension.swift
@@ -21,7 +21,7 @@ public extension PaywallComponent {
     enum Dimension: Codable, Sendable, Hashable {
 
         case vertical(HorizontalAlignment)
-        case horizontal(VerticalAlignment)
+        case horizontal(VerticalAlignment, FlexDistribution)
         case zlayer(TwoDimensionAlignment)
 
         public func encode(to encoder: any Encoder) throws {
@@ -31,9 +31,10 @@ public extension PaywallComponent {
             case .vertical(let alignment):
                 try container.encode(DimensionType.vertical.rawValue, forKey: .type)
                 try container.encode(alignment, forKey: .alignment)
-            case .horizontal(let alignment):
+            case .horizontal(let alignment, let distribution):
                 try container.encode(DimensionType.horizontal.rawValue, forKey: .type)
                 try container.encode(alignment, forKey: .alignment)
+                try container.encode(distribution, forKey: .distribution)
             case .zlayer(let alignment):
                 try container.encode(DimensionType.zlayer.rawValue, forKey: .type)
                 try container.encode(alignment.rawValue, forKey: .alignment)
@@ -50,7 +51,8 @@ public extension PaywallComponent {
                 self = .vertical(alignment)
             case .horizontal:
                 let alignment = try container.decode(VerticalAlignment.self, forKey: .alignment)
-                self = .horizontal(alignment)
+                let distribution = try container.decode(FlexDistribution.self, forKey: .distribution)
+                self = .horizontal(alignment, distribution)
             case .zlayer:
                 let alignment = try container.decode(TwoDimensionAlignment.self, forKey: .alignment)
                 self = .zlayer(alignment)
@@ -58,7 +60,7 @@ public extension PaywallComponent {
         }
 
         public static func horizontal() -> Dimension {
-            return .horizontal(.center)
+            return .horizontal(.center, .start)
         }
 
         public static func vertical() -> Dimension {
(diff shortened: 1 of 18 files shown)


Sources/Paywalls/Components/Common/Dimension.swift
```diff
@@ -18,10 +18,10 @@ import Foundation
 
 public extension PaywallComponent {
 
-    enum Dimension: Codable, Sendable, Hashable {
+    @frozen enum Dimension: Codable, Sendable, Hashable {
```
tonidero, line 21: I must say I'm not a big fan of this `@frozen` annotation... While it does make sense, it seems it might cause us some issues down the line. For this enum I don't foresee new values coming anytime soon but well... In any case, if this is the recommended approach, seems we just will need to be thinking forward a lot for these APIs.

Sources/Paywalls/Components/Common/PaywallComponentPropertyTypes.swift
```diff
@@ -109,7 +109,7 @@ public extension PaywallComponent {
 
     }
 
-    enum WidthSizeType: String, Codable, Sendable, Hashable, Equatable {
+    @frozen enum WidthSizeType: String, Codable, Sendable, Hashable, Equatable {
         case fit, fill, fixed
```
tonidero, line 113: Should we allow something like `fixedPercentage`?

Sources/Paywalls/Components/PaywallPackageComponent.swift
```diff
@@ -35,6 +35,18 @@ public extension PaywallComponent {
...
+    }
+
+}
+
+extension PaywallComponent.PackageComponent {
+
+    enum CodingKeys: String, CodingKey {
+        case type
+        case packageID = "packageId"
+        case isSelectedByDefault = "isDefaultSelected"
```
tonidero, line 48: Hmm strange these are using camelCase instead of snake_case? Same below

tonidero (approved):
Some questions but nothing blocking. Looking good!

Sources/Paywalls/Components/Common/PaywallComponentPropertyTypes.swift
joshdholtz: There has been A LOT of discussion about this 😅 We decided to leave this out (for now) but will probably introduce some type of "grid" component that has percentage like this in a more structured way

Sources/Paywalls/Components/Common/Dimension.swift
joshdholtz: I will just remove them for now 😇

Sources/Paywalls/Components/PaywallPackageComponent.swift
joshdholtz: Codables are a bit weird here 😅 

Keys will get deserialized from snake to camel automatically... however, `packageID` would get deserialized as `package_i_d` because of the `ID` both being capitals 🫠 

Some of the other codables are values that we _actualy_ are expecting to be snake case as well but `Codable` only auto converts for keys (not values)

tonidero: That's.... confusing 🥴. But yeah makes sense. Thanks for explaining!

joshdholtz: It honestly is 😬 Maybe I should document this is all the place? Because it will be more than just me supporting this and I don't want to make anybody else think about Swift quirks 😛

tonidero: Maybe a quick comment could help yeah. Having said that, I think as long as the decoding behavior is well tested, this should be caught with tests

PR #4438, opened 2024-11-05
[Paywalls] Send paywall events when the app is backgrounded and after a successful purchase

MarkVillacampa:
### Checklist

### Motivation

Right now we are only sending events when the app moves to the foregound. Sending events with backgrounding will help us record impression events that might happen when a user installs an app, opens it, proceeds to uninstall it or never use it again. Additionally, we try to sync events after a successful purchase.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -510,7 +510,8 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
                     storeMessagesHelper: storeMessagesHelper,
                     diagnosticsSynchronizer: diagnosticsSynchronizer,
                     diagnosticsTracker: diagnosticsTracker,
-                    winBackOfferEligibilityCalculator: winBackOfferEligibilityCalculator
+                    winBackOfferEligibilityCalculator: winBackOfferEligibilityCalculator,
+                    paywallEventsManager: paywallEventsManager
                 )
             } else {
                 return .init(
@@ -532,7 +533,8 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
                     manageSubscriptionsHelper: manageSubsHelper,
                     beginRefundRequestHelper: beginRefundRequestHelper,
                     storeMessagesHelper: storeMessagesHelper,
-                    winBackOfferEligibilityCalculator: winBackOfferEligibilityCalculator
+                    winBackOfferEligibilityCalculator: winBackOfferEligibilityCalculator,
+                    paywallEventsManager: paywallEventsManager
                 )
             }
         }()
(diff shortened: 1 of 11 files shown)


tonidero (approved):

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1913,13 +1918,4 @@ private extension Purchases {
...
     }
 
-    private func postPaywallEventsIfNeeded() {
-        guard #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *),
-              let manager = self.paywallEventsManager else { return }
-
-        self.operationDispatcher.dispatchOnWorkerThread(jitterableDelay: .long) {
-            _ = try? await manager.flushEvents(count: PaywallEventsManager.defaultEventFlushCount)
-        }
-    }
```
MarkVillacampa, line 1923: Moved this method to `PurchasesOrchestrator` so it can be called after a successful purchase.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -435,6 +440,8 @@ final class PurchasesOrchestrator {
                         Logger.rcPurchaseSuccess(Strings.purchase.purchased_product(
                             productIdentifier: productIdentifier
                         ))
+
+                        self.postPaywallEventsIfNeeded()
```
MarkVillacampa, line 444: This is called after a successful SK1 purchase. I think every SK1 purchase made using the SDK goes through this path, but I'd appreciate if there's somewhere else I missed.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -550,6 +557,8 @@ final class PurchasesOrchestrator {
 
         if let transaction = transaction {
             customerInfo = try await self.handlePurchasedTransaction(transaction, .purchase)
+
+            self.postPaywallEventsIfNeeded()
```
MarkVillacampa, line 561: Again I think this is called after a successful SK2 purchase. I think every SK2 purchase made using the SDK goes through this path, but let me know if there's somewhere else I missed.

vegaro: I think the only other place would be purchases that happen outside of the app, or basically other purchases in the queue.

I think the safest would be to sync in the same places we are syncing `unsyncedAttributes`. If you look at `storeKit2TransactionListener`, we add the unsynced attributes to `PurchasedTransactionData`, and post everything together. I think every single purchase goes through `storeKit2TransactionListener`

Since these are paywall events, I think those places are not that important. Something I was thinking is what happens if the transaction fails posting in a paywall? I think the event will be posted next time the app is opened and it should be safe, but maybe it's safer to just try to post events in `storeKit2TransactionListener`?

MarkVillacampa: I wanted to keep posting to only "newly purchased transactions via a paywall". If we sync in the listener, we'll be doing it for subscription updates, and purchases made outside the app which never generate paywall events... and yes those will be synced anyway when the app is backgrounded/foregrounded.

PR #4439, opened 2024-11-06
Support anonymous web purchase redemptions

tonidero:
### Description
Based on the work in #4381 

This adds support for web purchase redemption in iOS.

- [ ] Do not merge for now, just release a prerelease until the feature is out.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1232,6 +1232,7 @@ public extension Purchases {
         }
     }
 
+    /// Warning: This is currently experimental and subject to change.
     func redeemWebPurchase(
         webPurchaseRedemption: WebPurchaseRedemption,
         completion: @escaping (CustomerInfo?, PublicError?) -> Void
@@ -1240,6 +1241,7 @@ public extension Purchases {
                                                      completion: completion)
     }
 
+    /// Warning: This is currently experimental and subject to change.
     func redeemWebPurchase(_ webPurchaseRedemption: WebPurchaseRedemption) async -> WebPurchaseRedemptionResult {
         return await self.purchasesOrchestrator.redeemWebPurchase(webPurchaseRedemption)
     }
--- a/Sources/WebPurchaseRedemption/URL+WebPurchaseRedemption.swift
+++ b/Sources/WebPurchaseRedemption/URL+WebPurchaseRedemption.swift
@@ -16,7 +16,10 @@ import Foundation
 extension URL {
 
     /// Parses a URL and converts it to a ``WebPurchaseRedemption`` if possible that can be
-    /// redeemed using ``Purchases/redeemWebPurchase(_:)``
+    /// redeemed using ``Purchases/redeemWebPurchase(_:)`
+    ///
+    /// Warning: This is currently experimental and subject to change.
+    /// `
     /// - Seealso: ``Purchases/redeemWebPurchase(_:)``
     public var asWebPurchaseRedemption: WebPurchaseRedemption? {
         return Purchases.parseAsWebPurchaseRedemption(self)
--- a/Sources/WebPurchaseRedemption/WebPurchaseRedemption.swift
+++ b/Sources/WebPurchaseRedemption/WebPurchaseRedemption.swift
@@ -14,6 +14,8 @@
 import Foundation
 
 /// Class representing a web redemption deep link that can be redeemed by the SDK.
+/// Warning: This is currently experimental and subject to change.
+///
 /// - Seealso: ``Purchases/redeemWebPurchase(_:)``
 @objc(RCWebPurchaseRedemption) public final class WebPurchaseRedemption: NSObject {
 
--- a/Sources/WebPurchaseRedemption/WebPurchaseRedemptionResult.swift
+++ b/Sources/WebPurchaseRedemption/WebPurchaseRedemptionResult.swift
@@ -14,6 +14,8 @@
 import Foundation
 
 /// Represents the result of a web purchase redemption
+/// Warning: This is currently experimental and subject to change.
+/// 
 /// - Seealso: ``Purchases/redeemWebPurchase(_:)``
 public enum WebPurchaseRedemptionResult: Sendable {
 


Sources/Error Handling/ErrorCode.swift
```diff
@@ -180,7 +183,12 @@ extension ErrorCode: DescribableError {
             return "This feature is not available when utilizing the customEntitlementsComputation dangerousSetting."
         case .signatureVerificationFailed:
             return "Request failed signature verification. See https://rev.cat/trusted-entitlements for more info."
-
+        case .invalidWebPurchaseToken:
+            return "The link you provided does not contain a valid purchase token."
+        case .alreadyRedeemedWebPurchaseToken:
+            return "The link you provided has already been redeemed."
+        case .expiredWebPurchaseToken:
+            return "The link you provided has expired. A new one will be sent to the customer email."
```
MarkVillacampa, line 191: This might be more of a backend question but will the email always be sent automatically after trying to redeem an expired token? what if the user has already redeemed a different token for the same purchase?

Sources/Networking/HTTPClient/HTTPStatusCode.swift
```diff
@@ -60,6 +62,7 @@ extension HTTPStatusCode: RawRepresentable {
         case .temporaryRedirect: return 307
         case .invalidRequest: return 400
         case .unauthorized: return 401
+        case .forbidden: return 403
```
MarkVillacampa, line 65: surprised we didn't already have an endpoint returning 403 😄

Sources/Error Handling/ErrorCode.swift
tonidero: > will the email always be sent automatically after trying to redeem an expired token?

Most times, yes, except when an email has just been sent since it will be throttled, so we don't send continuous emails. cc @antoniobg to confirm.

> what if the user has already redeemed a different token for the same purchase?

The token is associated with a customer actually, also, if the token has already been redeemed we could be facing 2 situations:
- If it was redeemed to an anonymous user id, it will allow further redemptions, and a new one can be generated if expired.
- If it was redeemed to a non-anonymous user id, it will block further redemptions and return a `alreadyRedeemedWebPurchaseToken` so it won't send a new email.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -230,6 +235,43 @@ final class PurchasesOrchestrator {
...
+                let error = PurchasesError(error: .alreadyRedeemedWebPurchaseToken, userInfo: userInfo)
+                completion(nil, error.asPublicError)
+            case let .expired(obfuscatedEmail, wasEmailSent):
+                let userInfo: [NSError.UserInfoKey: Any] = [
+                    .obfuscatedEmail: obfuscatedEmail,
+                    .wasEmailSent: wasEmailSent
+                ]
+                let error = PurchasesError(error: .expiredWebPurchaseToken, userInfo: userInfo)
+                completion(nil, error.asPublicError)
+            }
```
MarkVillacampa, line 271: So for the Swift async API, we're exposing the result enum cases directly, which include success, generic error, and specific error cases, but for Objective-C we will expose every non-success case as as a `PurchasesError` right?

tonidero: That's correct! I think having specific errors makes for a better API, but that's tricky with ObjC... So we can just document how to handle the errors when using ObjC IMO. Lmk if you have other thoughts!

Sources/Error Handling/ErrorCode.swift
antoniobg: > Most times, yes, except when an email has just been sent since it will be throttled, so we don't send continuous emails. cc @antoniobg to confirm.

Yes, basically we only send the email when the token you tried to redeem is expired and matches the latest token we had in file. If you tried to redeem an older token, the email won't be sent. 
I think this can be improved and instead use a rate limit to ensure no more than X emails are sent in X minutes (cc @vicfergar)

joshdholtz (commented):
Overall this looks good to me! 😁

Sources/Logging/Strings/WebRedemptionStrings.swift
```diff
@@ -0,0 +1,41 @@
...
+
+import Foundation
+
+// swiftlint:disable identifier_name
+
+enum WebRedemptionStrings {
+
+    case redeeming_web_purchase
+    case redeemed_web_purchase
+    case error_redeeming_web_purchase(_ error: BackendError)
```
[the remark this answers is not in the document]
tonidero: Well, that's how we have the other strings files... Not sure if there is a reason, TBH 😅 , but I think it might be good to be consistent IMO.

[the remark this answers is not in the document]
joshdholtz: Oh is it? I didn't even realize that 😅 I'm fine with consistency then lol

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/PurchaseTesterApp.swift
```diff
@@ -62,6 +68,36 @@ struct PurchaseTesterApp: App {
             .navigationViewStyle(.automatic)
             .animation(.default, value: self.isConfigured)
             .transition(.opacity)
+            .onOpenURL { url in
+                if let webPurchaseRedemption = Purchases.parseAsWebPurchaseRedemption(url),
+                    Purchases.isConfigured {
+                    print("Redeeming web purchase.")
```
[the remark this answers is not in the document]
tonidero: Probably not, just had it for debugging issues, but will remove it 👍

Sources/Error Handling/ErrorCode.swift
aboedo: Thought on the copy: 
> The link you provided has expired. A new one will be sent to the customer email.

99/100 developers are going to just show exactly what we have in the error code message to the end user. It's not ideal, but that's typically what happens. So let's try to write this in a way that it would make sense to an end user and developer alike. 

> The link you provided has expired. A new one will be sent to the email used to make the purchase.

Sources/Networking/HTTPClient/HTTPStatusCode.swift
aboedo: +1

Sources/Networking/Operations/PostRedeemWebPurchaseOperation.swift
```diff
@@ -0,0 +1,146 @@
...
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  PostReceiptDataOperation.swift
+//
+//  Created by Joshua Liebowitz on 11/18/21.
```
aboedo, line 12: lol pretty sure this isn't right unless Josh was looking suuuper far ahead in 2021 and kept this file a secret for years

Sources/Networking/Operations/PostRedeemWebPurchaseOperation.swift
```diff
@@ -0,0 +1,146 @@
...
+        postData: PostData,
+        customerInfoResponseHandler: CustomerInfoResponseHandler,
+        customerInfoCallbackCache: CallbackCache<CustomerInfoCallback>
+    ) -> CacheableNetworkOperationFactory<PostRedeemWebPurchaseOperation> {
+        /// Cache key comprises of the following:
+        /// - `appUserID`
+        /// - `redemptionToken`
+        let cacheKey = "\(configuration.appUserID)-\(postData.redemptionToken)"
+
+        return .init({ cacheKey in
```
aboedo, line 51: never was a fan of `.init` personally but that's just me

Sources/Logging/Strings/WebRedemptionStrings.swift
```diff
@@ -0,0 +1,41 @@
...
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  ETagStrings.swift
+//
+//  Created by Nacho Soto on 6/15/23.
```
aboedo, line 12: lol Nacho's back

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -747,6 +753,12 @@ extension Purchases {
 
 public extension Purchases {
 
+    /// Parses a deep link URL to verify it's a RevenueCat web purchase redemption link
+    /// - Seealso: ``Purchases/redeemWebPurchase(_:)``
+    @objc static func parseAsWebPurchaseRedemption(_ url: URL) -> WebPurchaseRedemption? {
```
aboedo, line 758: this API kind of feels off to me, for two reasons: 
1) it feels like leaked abstractions, don't know, I just feel awkward having to ask an SDK to parse a url as a web purchase redemption. Shouldn't the SDK just be able to take the URL? 
2) it doesn't quite roll off the tongue, now does it?

aboedo: I mentioned this during CoreSDK weekly, I feel like we can keep the functionality by splitting the API into "is this link for a web purchase redemption" and "redeem", i.e.:

```swift
func isWebPurchaseRedemptionLink(_ url: URL) -> Bool
func redeemWebPurchase(_ webPurchaseRedemption: WebPurchaseRedemption) async -> WebPurchaseRedemptionResult {
```

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1217,6 +1229,18 @@ public extension Purchases {
...
+    func redeemWebPurchase(
+        webPurchaseRedemption: WebPurchaseRedemption,
+        completion: @escaping (CustomerInfo?, PublicError?) -> Void
+    ) {
+        self.purchasesOrchestrator.redeemWebPurchase(webPurchaseRedemption: webPurchaseRedemption,
+                                                     completion: completion)
+    }
+
+    func redeemWebPurchase(_ webPurchaseRedemption: WebPurchaseRedemption) async -> WebPurchaseRedemptionResult {
+        return await self.purchasesOrchestrator.redeemWebPurchase(webPurchaseRedemption)
```
aboedo, line 1242: should this be a throwing function instead, and have the errors be thrown instead of a valid result? That would feel more Swift-y to me, cc @MarkVillacampa for a swiftier brain

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
aboedo: oh this actually kind of gets to what I just asked in the other method, maybe we can have swift throw instead

aboedo (changes requested):
still reviewing, leaving some thoughts in the meantime

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
tonidero: Answered in https://github.com/RevenueCat/purchases-ios/pull/4439#discussion_r1833188288

Sources/Purchasing/Purchases/Purchases.swift
tonidero: From rereading, I think your preferred option would be something like:
```
func isWebPurchaseRedemptionLink(_ url: URL) -> Bool
func redeemWebPurchase(_ url: URL) async -> WebPurchaseRedemptionResult {
```
Where the `WebPurchaseRedemptionResult` has an `invalidLink` kind of result right? We can chat today to make sure we're in the same page.

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -907,6 +907,20 @@ public protocol PurchasesType: AnyObject {
     @available(iOS 13.0, macOS 10.15, tvOS 13.0, watchOS 6.2, *)
     func syncAttributesAndOfferingsIfNeeded() async throws -> Offerings?
 
+    /**
+     * Redeems a web purchase previously parsed from a deep link with ``Purchases/parseAsWebPurchaseRedemption(_:)``.
+     * We recommend using ``Purchases/redeemWebPurchase(_:)`` for a nicer API unless you're using ObjC.
```
aboedo: I don't think this is a thing we should even bring up - there are valid reasons to not use an async / await API, even now that Swift Concurrency is backported to iOS 13. 

Specifically, the Swift Concurrency model and the old concurrency model don't necessarily get along all that well, so if you're using completion blocks, it might be a good idea to stick with that. 

If folks are looking for async / await, they'll find it. And those are the APIs that we'll be advertising and showing first and foremost in our docs.

aboedo: Instead, we should add it as a seealso so that folks who find this find the other version too

Tests/TestingApps/PurchaseTesterSwiftUI/Shared/PurchaseTesterApp.swift
```diff
@@ -62,6 +68,35 @@ struct PurchaseTesterApp: App {
...
+                        }
+                        showAlert = true
+                    }
+                }
+            }
+            .alert("Redemption result", isPresented: $showAlert) {
+                Button("OK", role: .cancel) { }
+            } message: {
+                Text(alertMessage ?? "No result")
+            }
```
aboedo, line 99: ❤️

aboedo (approved):
LGTM, @MarkVillacampa I'd appreciate another pass since I made a few small updates

vegaro (approved):

vegaro (approved):

PR #4442, opened 2024-11-06
[Paywalls] Fix an issue where multi-tier paywalls might fail to render

aboedo:
Fixes an issue where multi-tier templates might fail to render if there's no value for `callToAction` at the paywall level

### DIFF
--- a/Sources/Paywalls/PaywallData.swift
+++ b/Sources/Paywalls/PaywallData.swift
@@ -108,8 +108,9 @@ extension PaywallData {
         // swiftlint:disable missing_docs
 
         public var title: String
-        public var callToAction: String
 
+        @DefaultDecodable.EmptyString
+        var _callToAction: String
         @NonEmptyStringDecodable
         var _subtitle: String?
         @NonEmptyStringDecodable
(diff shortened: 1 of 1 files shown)


aboedo:
@joshdholtz this fixes the issue in my testing but I won't pretend like I know what I'm doing anymore, could you give it a look and also tell me where I should be looking at for updating tests?

aboedo:
the full conversation is here: https://linear.app/revenuecat/issue/DENG-984/paywall-7-unstyled-in-certain-builds

From what I can tell, though, in Paywall 7 we don't really need a global CTA? Do we keep one just in case you move into a different template?

joshdholtz:
@aboedo Ahhhhh.... the issue is that `localized_strings` shouldn't actually have any values set at all for the multi-tier template. It should only have the values in `localized_strings_by_tier` 🤔 

So this fix in the iOS repo is not _ideal_ because its actually a data issue caused by the frontend. If Template 1 through 5 _is_ actually missing the CTA, we should go to the fallback instead of showing an empty string.

This fix needs to happen on the FE so that `localized_strings` is set as an empty string when using Template 7. We could also do something on the backend to make sure these strings are FOR SURE being sent to SDK as empty string (which would fix this for everyone without updating the SDK)

aboedo:
That makes sense. I would heavily favor a backend/frontend fix whenever possible since it won't require updating SDKs, just wasn't sure whether this was us mishandling the data SDK-side or us generating the wrong data backend-side. 

Closing in favor of backend-oriented fix

PR #4444, opened 2024-11-07
Close feedback survey after picking an option

vegaro:
Cleans up the flow a bit, so the feedback survey is closed after picking an option and doesn't stay there after cancelling a subscription

This is needed to be able to send the selected feedback survey option to the backend as event.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
@@ -18,7 +18,7 @@ import SwiftUI
 
 #if os(iOS)
 
-/// Warning: This is currently in beta and ubject to change.
+/// Warning: This is currently in beta and subject to change.
 ///
 /// A SwiftUI view for displaying a customer support common tasks
 @available(iOS 15.0, *)
--- a/RevenueCatUI/CustomerCenter/Views/FeedbackSurveyView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/FeedbackSurveyView.swift
@@ -33,19 +33,27 @@ struct FeedbackSurveyView: View {
     @Environment(\.colorScheme)
     private var colorScheme
 
+    @Binding
+    private var isPresented: Bool
+
     init(feedbackSurveyData: FeedbackSurveyData,
-         customerCenterActionHandler: CustomerCenterActionHandler?) {
+         customerCenterActionHandler: CustomerCenterActionHandler?,
+         isPresented: Binding<Bool>) {
         self._viewModel = StateObject(wrappedValue: FeedbackSurveyViewModel(
             feedbackSurveyData: feedbackSurveyData,
             customerCenterActionHandler: customerCenterActionHandler
         ))
+        self._isPresented = isPresented
     }
 
     var body: some View {
         ZStack {
             List {
                 FeedbackSurveyButtonsView(options: self.viewModel.feedbackSurveyData.configuration.options,
-                                          onOptionSelected: self.viewModel.handleAction(for:),
+                                          onOptionSelected: { option in
+                                              await self.viewModel.handleAction(for: option)
+                                              self.isPresented = false
+                                          },
                                           loadingState: self.$viewModel.loadingState)
             }
             .sheet(
--- a/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
@@ -55,7 +55,8 @@ struct ManageSubscriptionsView: View {
                 .navigationDestination(isPresented: .isNotNil(self.$viewModel.feedbackSurveyData)) {
                     if let feedbackSurveyData = self.viewModel.feedbackSurveyData {
                         FeedbackSurveyView(feedbackSurveyData: feedbackSurveyData,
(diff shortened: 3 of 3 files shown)


RevenueCatUI/CustomerCenter/Views/FeedbackSurveyView.swift
```diff
@@ -59,8 +85,21 @@ struct FeedbackSurveyView: View {
...
+            NavigationLink(destination: EmptyView(), isActive: $isOptionSelected) {
+                EmptyView()
+            }
+            .hidden()
+        )
     }
 
+    private func handleOptionSelected() {
+        if #available(iOS 16.0, *), let navigationPath = navigationPathBinding {
+            navigationPath.wrappedValue.removeLast()
```
vegaro, line 98: @fire-at-will this is not working, by the time it gets here, the stack is empty

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -64,6 +78,11 @@ public struct CustomerCenterView: View {
             await loadInformationIfNeeded()
         }
         .environmentObject(self.viewModel)
+        .onAppear {
```
fire-at-will, line 81: Nit: It might be simpler to do this initialization where `_navigationPathWrapper` is declared instead of here in an onAppear. Something like:

```swift
@State private var _navigationPathWrapper: Any? = {
    if _navigationPathWrapper == nil, #available(iOS 16.0, *) {
                return NavigationPathWrapper()
    } else {
        return nil
   }
}()
```

RevenueCatUI/CustomerCenter/Views/FeedbackSurveyView.swift
```diff
@@ -59,8 +85,21 @@ struct FeedbackSurveyView: View {
         }
         .navigationTitle(self.viewModel.feedbackSurveyData.configuration.title)
         .navigationBarTitleDisplayMode(.inline)
+        .background(
+            NavigationLink(destination: EmptyView(), isActive: $isOptionSelected) {
+                EmptyView()
+            }
+            .hidden()
+        )
```
fire-at-will, line 93: Why is this included?

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -50,26 +69,32 @@ struct ManageSubscriptionsView: View {
...
-            content
-                .navigationDestination(isPresented: .isNotNil(self.$viewModel.feedbackSurveyData)) {
-                    if let feedbackSurveyData = self.viewModel.feedbackSurveyData {
-                        FeedbackSurveyView(feedbackSurveyData: feedbackSurveyData,
-                                           customerCenterActionHandler: self.customerCenterActionHandler)
+        content
+            .background(
+                Group {
+                    if #available(iOS 16.0, *), let navigationPath = navigationPathBinding {
+                        content
```
fire-at-will, line 76: Is it necessary to re-render `content` in the background on `content`?

It might be more efficient to place these modifiers on something that we know won't have a lot of rendering complexity. I'll often use something like `Color.clear` for things like this, and set the opacity to 0 to be safe

vegaro:
@fire-at-will I was struggling so much to make this work with `NavigationStack`... I tried another simpler approach. I think we need to rethink the way we do navigation. We've already discussed that in the past, and I think it is going to be one of the next things to address. We probably need to build our own system

Anyways, I think this works for now and it's now ready for review

RevenueCatUI/CustomerCenter/Views/FeedbackSurveyView.swift
```diff
@@ -33,19 +33,27 @@ struct FeedbackSurveyView: View {
     @Environment(\.colorScheme)
     private var colorScheme
 
+    @Binding
+    private var isPresented: Bool
```
tonidero, line 37: Hmm I'm a bit confused, this seems to only be assigned but not read?

vegaro: this might be a bit confusing yes... I create a binding here https://github.com/RevenueCat/purchases-ios/pull/4444/files#diff-121dcc7927012032acdbbf026716bf9daf66d91cee530ab12eae052d06526aa0R68

```
.isNotNil(self.$viewModel.feedbackSurveyData)
```

So if I set that to false, it nullifies the `feedbackSurveyData` which hides the feedback survey, since it's only presented if `feedbackSurveyData` is not nil.

Another approach I can use is adding an `onDismiss` callback:

```
.navigationDestination(isPresented: .isNotNil(self.$viewModel.feedbackSurveyData)) {
      if let feedbackSurveyData = self.viewModel.feedbackSurveyData {
          FeedbackSurveyView(feedbackSurveyData: feedbackSurveyData,
                             customerCenterActionHandler: self.customerCenterActionHandler,
                             onDismiss: { self.viewModel.feedbackSurveyData = nil })
      }
  }
```

Maybe that's more clear?

tonidero: Ahhh got it! Well, I think it might be clearer, but I think it's ok as is. Thanks!

tonidero (approved):

PR #4446, opened 2024-11-07
Add new view modifier to redeem web purchases

tonidero:
### Description
Based on #4439 

This adds a new view modifier `onWebPurchaseRedemptionAttempt`. The usage looks like this:
```
Content()
   .onWebPurchaseRedemptionAttempt { result in 
       handleResult(result)
   } 
```
With this view modifier, we will listen to new URLs automatically, and if we detect a web purchase redemption link, perform the redemption automatically. 

If the developer needs more control over when the redemption should happen, we will still have the APIs introduced in #4439


tonidero:
@RCGitBot please test

Tests/TestingApps/PaywallsTester/PaywallsTester/Info.plist
```diff
@@ -2,6 +2,19 @@
...
+			<key>CFBundleTypeRole</key>
+			<string>Editor</string>
+			<key>CFBundleURLName</key>
+			<string>RevenueCat</string>
+			<key>CFBundleURLSchemes</key>
+			<array>
+				<string>rc-39976e83d0</string>
+			</array>
+		</dict>
+	</array>
```
vegaro, line 17: unsure what this is

vegaro (approved):

Tests/TestingApps/PaywallsTester/PaywallsTester/Info.plist
tonidero: Ah right, this is to support custom schema deep links in iOS. Basically the `rc-39976e83d0` is the code for our test app... We could potentially keep it empty, but I feel having a default might make it easier?

MarkVillacampa (approved):

PR #4467, opened 2024-11-11
[Paywalls V2] Update stack size spec

joshdholtz:
## Motivation

Update stack size per spec

## Description

All basic components (text, image, and stacks) gained a size property. It allows their `width` and `height` to be either `fit`, `fill`, or `fixed`

This is the implementation for `StackComponent` and `StackComponentView`
- Replaces `width` with `size`
- New `.size(Size)` view modifier that applies these width and height settings and can be used by other components
- Added new `FlexVStack` (similar to `FlexHStack`) which wasn't added before 🤷‍♂️ 

### Demo

https://github.com/user-attachments/assets/72652503-f7a6-4507-bd24-fce079c8ea85

### DIFF
--- a/RevenueCatUI/Templates/Components/SizeModifier.swift
+++ b/RevenueCatUI/Templates/Components/SizeModifier.swift
@@ -1,0 +1,71 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  SizeModifier.swift
+//
+//  Created by Josh Holtz on 11/11/24.
+
+import RevenueCat
+import SwiftUI
+
+#if PAYWALL_COMPONENTS
+
+struct SizeModifier: ViewModifier {
+
+    var size: PaywallComponent.Size
+
+    func body(content: Content) -> some View {
+        content
+            .applyWidth(size.width)
+            .applyHeight(size.height)
+    }
+
+}
+
+extension View {
+
+    @ViewBuilder
+    func applyWidth(_ sizeConstraint: PaywallComponent.SizeConstraint) -> some View {
+        switch sizeConstraint {
+        case .fit:
+            self
+        case .fill:
+            self
+                .frame(maxWidth: .infinity)
+        case .fixed(let value):
+            self
+                .frame(width: CGFloat(value))
+        }
+    }
+
+    @ViewBuilder
+    func applyHeight(_ sizeConstraint: PaywallComponent.SizeConstraint) -> some View {
+        switch sizeConstraint {
+        case .fit:
+            self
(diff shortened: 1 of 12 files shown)


RevenueCatUI/Templates/Components/SizeModifier.swift
```diff
@@ -0,0 +1,71 @@
...
+    func applyWidth(_ sizeConstraint: PaywallComponent.SizeConstraint) -> some View {
+        switch sizeConstraint {
+        case .fit:
+            self
+        case .fill:
+            self
+                .frame(maxWidth: .infinity)
+        case .fixed(let value):
+            self
+                .frame(width: CGFloat(value))
```
fire-at-will, line 43: Paranoid check: Since SizeConstraint.fixed's value is an `Int`, it might be good to make sure that value is >= 0:

```suggestion
                .frame(width: CGFloat(value >= 0 ? value : 0))
```

RevenueCatUI/Templates/Components/SizeModifier.swift
```diff
@@ -0,0 +1,71 @@
...
+    func applyHeight(_ sizeConstraint: PaywallComponent.SizeConstraint) -> some View {
+        switch sizeConstraint {
+        case .fit:
+            self
+        case .fill:
+            self
+                .frame(maxHeight: .infinity)
+        case .fixed(let value):
+            self
+                .frame(height: CGFloat(value))
```
fire-at-will, line 57:
```suggestion
                .frame(height: CGFloat(value >= 0 ? value : 0)))
```

RevenueCatUI/Templates/Components/Stack/FlexVStack.swift
```diff
@@ -0,0 +1,92 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  FlexHStack.swift
```
fire-at-will, line 10:
```suggestion
//  FlexVStack.swift
```

Sources/Paywalls/Components/Common/PaywallComponentPropertyTypes.swift
```diff
@@ -171,19 +171,69 @@ public extension PaywallComponent {
...
+        }
+
     }
 
-    struct WidthSize: Codable, Sendable, Hashable, Equatable {
+    enum SizeConstraint: Codable, Sendable, Hashable {
+
+        case fit
+        case fill
+        case fixed(Int)
```
fire-at-will, line 190: Nit: Maaaaybe consider making this a UInt to get rid of the possibility of negative values?

fire-at-will (approved):
Looks good! I might consider adding a few unit tests for the custom Codable implementations, but other than looks good!

joshdholtz:
@RCGitBot please test

PR #4469, opened 2024-11-11
[Paywalls V2] Update text spec

joshdholtz:
## Motivation

Update text to new platform common font size

## Description

| Text Settings |
|--------|
| Simulator Screenshot - iPhone 16 Pro - 2024-11-11 at 09 42 59 |


RevenueCatUI/Templates/Components/PaywallComponentTypeTransformers.swift
```diff
@@ -16,42 +16,44 @@ import SwiftUI
...
+        case .headingXS: fontSize = 16
+        case .bodyXL: fontSize = 18
+        case .bodyL: fontSize = 17
+        case .bodyM: fontSize = 15
+        case .bodyS: fontSize = 13
         }
+
+        // Create a UIFont and apply dynamic type scaling
+        let baseFont = UIFont.systemFont(ofSize: fontSize, weight: .regular)
+        return UIFontMetrics(forTextStyle: self.textStyle).scaledFont(for: baseFont)
```
joshdholtz, line 56: Did not want to rely on the common iOS `.title` and `.largeTitle` so that we have a little more control. This also creates all font as the `regular` weight so that its clear what the `font_weight` property on components is actually doing.

fire-at-will (approved):

PR #4470, opened 2024-11-11
Update PurchaseParam code sample

fire-at-will:
### Description
This PR updates the code sample in the `PurchaseParams` class-level DocC to accurately reflect the purchasing interface.

### DIFF
--- a/Sources/Purchasing/Purchases/PurchaseParams.swift
+++ b/Sources/Purchasing/Purchases/PurchaseParams.swift
@@ -26,7 +26,7 @@ import Foundation
  *                            .with(metadata: ["key": "value"])
  *                            .with(promotionalOffer: promotionalOffer)
  *                            .build()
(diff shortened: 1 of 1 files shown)


Sources/Purchasing/Purchases/PurchaseParams.swift
```diff
@@ -26,7 +26,7 @@ import Foundation
  *                            .with(metadata: ["key": "value"])
  *                            .with(promotionalOffer: promotionalOffer)
  *                            .build()
- *  Purchases.shared.purchase(package: package, params: params)
+ * Purchases.shared.purchase(params)
```
aboedo, line 29: looks good. Still not sure why we don't have these methods take in a Builder so that you don't have to remember to call `build()` yourself, that feels pretty pointless, but it's not related to this PR

Issue #4456, opened 2024-11-08
customer center accept offer doesn't dismiss sheet with offer


unxavi:
- [ ] I have updated Purchases SDK to the latest version
- [x ] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [ x] I have searched the [Community](https://community.revenuecat.com)
- [ x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [ x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
When displaying an offer for the cancel subscription and the user accepts the offer, then the sheet with the offer doesn't go away. The user stays with the screen displaying the offer and the only option to dismiss the sheet is to press "no thanks" and then is redirected to cancel the subscription

1. Environment
   1. Platform:
   2. SDK version:
   3. StoreKit version:
      - [ ] StoreKit 1 (default on versions <5.0.0. Can be enabled in versions >=5.0.0 with `.with(storeKitVersion: .storeKit1)`)
      - [x] StoreKit 2 (default on versions >=5.0.0)
   4. OS version: 18.0
   5. Xcode version: 16.1
   6. Device and/or simulator:
      - [ x] Device
      - [ ] Simulator
   7. Environment:
      - [x ] Sandbox
      - [x ] TestFlight
      - [x ] Production
   8. How widespread is the issue. Percentage of devices affected. 100%
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue. **Complete logs with `Purchases.logLevel = .verbose` will help us debug this issue.**

```
Logs here
```

3. Steps to reproduce, with a description of expected vs. actual behavior

Use customer center, press cancel subscription and accept the offer to continue
 
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

5. **Additional context**
The sheet should be dismiss and the user should have some feedback

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/48423) and notified the team. Thank you for reporting, we're checking this out!

unxavi:
Any ETA?

vegaro:
I might not have time to fix this before the weekend, but hoping it will go out in next week's release

fire-at-will:
[...]

PR #4475, opened 2024-11-12
Dismiss promotional offer sheet after successful purchase

fire-at-will:
### Description
This PR addresses https://github.com/RevenueCat/purchases-ios/issues/4456 by automatically dismissing the promotional offer sheet after successfully purchasing a promotional offer.

It does so by introducing back-flow logic into the `PromotionalOfferView` where presenting views can take actions based off of the result of the user visiting the `PromotionalOfferView`.

It introduces a new `PromotionalOfferViewAction` that tells presenters of the `PromotionalOfferView` what caused the `PromotionalOfferView` to be dismissed:

```swift
enum PromotionalOfferViewAction {
    /// The user clicked the "No Thanks" button and declined the offer
    case declinePromotionalOffer

    /// The user successfully redeemed the promotional offer
    case successfullyRedeemedPromotionalOffer(PurchaseResultData)

    // Promotional code redemption failed. Either the user attempted to redeem the promotional offer, and it failed,
    // or the promotional offer was not loaded successfully.
    case promotionalCodeRedemptionFailed(Error)
}
```

#### Refund Flow
   - Makes it so that successfully redeeming a promotional off on the promotional offer view dismisses the refund flow and goes back to the main customer center screen
   - Fixes a bug where an error in the promotional offer flow causes a blank sheet to be presented to the user. Instead, an error will return back to the normal refund flow.

#### Cancel/Survey Flow
- Fixed a bug where tapping an option always immediately calls `isPresented = false`, causing the feedback view screen to always dismiss, which then caused the promotional offer view to dismiss
- Makes it so that successfully redeeming a promotional off on the promotional offer view dismisses the cancel flow and goes back to the main customer center screen

#### Misc
- Abstracts out our calls to the Purchases SDK when purchasing a promotional offer

### Testing
Manually tested changes

#### Redeeming Promo Offer Through Refund Flow

https://github.com/user-attachments/assets/840898af-8136-407d-99eb-d5168bbff303

#### Redeeming Promo Offer Through Cancel/Survey Flow

https://github.com/user-attachments/assets/401c0fe3-37dd-425d-9334-4772f369c9e3


RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
```diff
@@ -83,8 +83,14 @@ struct PromotionalOfferView: View {
                 }
             }
         }
+        .onAppear {
+            self.viewModel.onPromotionalOfferSuccessfullyPurchased = self.onPromotionalOfferSuccessfullyPurchased
+        }
     }
 
+    private func onPromotionalOfferSuccessfullyPurchased() {
+        self.dismiss()
```
vegaro, line 92: Does this close the whole Customer Center? Or only the promotional offer?

fire-at-will: The `dismiss` environment variable is given to us by SwiftUI, and it dismisses the current view depending on how it's presented:
- If the current view is in a NavigationStack, it pops the view off of the stack
- If the current view is presented in a sheet, it dismisses the sheet

Since the `PromotionalOfferView` is always presented in a sheet on top of other Customer Center stuff, it'll only dismiss the sheet that the PromotionalOfferView is presented in

If you're curious, you can read more about the `dismiss` environment value here: https://developer.apple.com/documentation/swiftui/environmentvalues/dismiss

vegaro (approved):

PR #4476, opened 2024-11-12
[Paywalls V2] Update Stack to handle property overrides

joshdholtz:
## Motivation

Do what Stack version of #4417

## Description

Stacks can now use overrides based on:
- Intro offers
- State
  - Selected
- Conditions(size class)
  - Compact
  - Medium
  - Expanded

### Bonus

Implements a new system for handling UI component state overrides based on different 
presentation contexts (screen size, selection state, etc).

See #4477 for what this change looks like on when used on a new component (it was super clean when adding it to image)

Key changes:
- New `PresentedPartial` protocol for component combining behavior/properties
- Generic `PresentedOverrides<T>` struct to handle:
  - Intro offer overrides
  - Selection state overrides (selected/unselected)
  - Screen condition overrides (compact/medium/expanded)
- Helper methods for building partials based on current state/condition
- Conversion utilities to map between component types

The system follows a hierarchical override pattern where more specific states (like selection) take precedence over more general ones (like screen size). Screen conditions cascade (expanded includes medium + compact overrides).


RevenueCatUI/Templates/Components/Stack/StackComponentView.swift
```diff
@@ -32,79 +38,131 @@ struct StackComponentView: View {
...
+        .size(style.size)
+        .background(style.backgroundColor)
+        .shape(border: style.border,
+               shape: style.shape)
+        .applyIfLet(style.shadow) { view, shadow in
             // Without compositingGroup(), the shadow is applied to the stack's children as well.
             view.compositingGroup().shadow(shadow: shadow)
         }
-        .padding(viewModel.margin)
+        .padding(style.margin)
```
vegaro, line 85: how does this interact with line 75? Does it override it?

RevenueCatUI/Templates/Components/Stack/StackComponentViewModel.swift
```diff
@@ -16,23 +16,116 @@ import SwiftUI
...
+        margin: PaywallComponent.Padding,
+        shape: PaywallComponent.Shape?,
+        border: PaywallComponent.Border?,
+        shadow: PaywallComponent.Shadow?
+    ) {
+        self.visible = visible
+        self.dimension = dimension
+        self.size = size
+        self.spacing = spacing
+        self.backgroundColor = backgroundColor?.toDyanmicColor() ?? Color.clear
```
vegaro, line 123:
```suggestion
        self.backgroundColor = backgroundColor?.toDynamicColor() ?? Color.clear
```

Sources/Paywalls/Components/PaywallImageComponent.swift
```diff
@@ -54,7 +54,6 @@ public extension PaywallComponent {
         public let overrideSourceLid: LocalizationKey?
         public let maskShape: MaskShape?
         public let gradientColors: [ColorHex]?
-        public let maxHeight: CGFloat?
```
vegaro, line 57: this is public, is it okay?

vegaro (approved):

RevenueCatUI/Templates/Components/Stack/StackComponentView.swift
joshdholtz: @vegaro  This does look very unclear without context 😅  SwiftUI applies things on top of each other so...

1. Line 75 adds padding to the view to make it bigger
2. Line 78 adds background color (which is applied to the whole bigger view)
3. This line (85) adds `padding` again (but without background color) so it acts as margin

Sources/Paywalls/Components/PaywallImageComponent.swift
joshdholtz: Yup yup! Its behind the compiler flag still so nobody using 😇

PR #4482, opened 2024-11-13
Add `appSessionID` to `PaywallEventsManager` 

vegaro:
Added an `appSessionID` that can be shared between all events. Since at the moment `session_identifier` in the paywalls represent the paywall session, not the whole app session.

I just added the generation, a reset functionality and a fallback for now

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -820,6 +820,10 @@ public extension Purchases {
             self.systemInfo.isApplicationBackgrounded { isAppBackgrounded in
                 self.updateOfferingsCache(isAppBackgrounded: isAppBackgrounded)
             }
+
+            Task {
+                await self.paywallEventsManager?.resetAppSessionID()
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -820,6 +820,10 @@ public extension Purchases {
             self.systemInfo.isApplicationBackgrounded { isAppBackgrounded in
                 self.updateOfferingsCache(isAppBackgrounded: isAppBackgrounded)
             }
+
+            Task {
+                await self.paywallEventsManager?.resetAppSessionID()
+            }
```
vegaro, line 826: I am unsure about this, but `paywallEventsManager` is an actor

fire-at-will (commented):
It looks like we're generating an app session ID when login/logout are called, but does one get created if the developer doesn't call those functions?

I don't have the full context on this feature, so it might not be necessary, but just wanted to make sure!

vegaro:
@fire-at-will Yes, in the init of `PaywallEventsManager` https://github.com/RevenueCat/purchases-ios/pull/4482/files#diff-dcba7f0e5ca516087d45ef4b0bcd031d91ad329aa90d8e86b0928e2484ad07d3R48. Maybe it's not clear enough?

Btw, I am happy to change the approach :)

I tested that old Paywall Events would have an associated app session ID (on the SDK only not in the event sent yet)

PR #4489, opened 2024-11-14
Add support for `product_mapping` in promotional offers

vegaro:
There's a new `product_mapping` field in the `promotional_offer` JSON object. This map is more specific on which offer to load. This mapping will be configurable with the new UI, but we should still support the old `ios_offer_id` in case the developer has not configured a mapping yet.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
+++ b/RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
@@ -49,20 +49,32 @@ class LoadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType {
                 return .failure(CustomerCenterError.couldNotFindSubscriptionInformation)
             }
 
-            let exactMatch = subscribedProduct.discounts.first { discount in
-                discount.offerIdentifier == promoOfferDetails.iosOfferId
+            let exactMatch = if !promoOfferDetails.productMapping.isEmpty {
+                // Try to find match using product mapping first
+                subscribedProduct.discounts.first { discount in
+                    guard let offerIdentifier = discount.offerIdentifier else { return false }
+                    return promoOfferDetails.productMapping[productIdentifier] == offerIdentifier
+                }
+            } else {
+                // Fall back to iosOfferId if no mapping (in case of using legacy iosOfferId)
+                subscribedProduct.discounts.first { discount in
+                    discount.offerIdentifier == promoOfferDetails.iosOfferId
+                }
             }
 
             let discount: StoreProductDiscount?
             if let exactMatch = exactMatch {
                 discount = exactMatch
-            } else {
+            } else if promoOfferDetails.productMapping.isEmpty {
+                // Only try suffix matching if we're using iosOfferId (no mapping)
                 discount = subscribedProduct.discounts.first { discount in
                     guard let offerIdentifier = discount.offerIdentifier else {
                         return false
                     }
                     return offerIdentifier.hasSuffix("_\(promoOfferDetails.iosOfferId)")
                 }
+            } else {
+                discount = nil
             }
 
             guard let discount = discount else {
--- a/RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
+++ b/RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
@@ -50,7 +50,7 @@ class LoadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType {
             }
 
             let exactMatch = if !promoOfferDetails.productMapping.isEmpty {
-                // Try to find match using product mapping first
+                // Try to find match using product mapping if configured and ignore iosOfferId
                 subscribedProduct.discounts.first { discount in
                     guard let offerIdentifier = discount.offerIdentifier else { return false }
                     return promoOfferDetails.productMapping[productIdentifier] == offerIdentifier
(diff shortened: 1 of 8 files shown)


RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
```diff
@@ -43,47 +43,103 @@ class LoadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType {
...
         }
     }
 
+    private func getActiveSubscription(_ customerInfo: CustomerInfo) async throws -> (String, StoreProduct) {
+        guard let productIdentifier = customerInfo.earliestExpiringAppStoreEntitlement()?.productIdentifier,
+              let subscribedProduct = await self.purchasesProvider.products([productIdentifier]).first else {
+            Logger.warning(Strings.could_not_offer_for_any_active_subscriptions)
+            throw CustomerCenterError.couldNotFindSubscriptionInformation
+        }
+        return (productIdentifier, subscribedProduct)
```
aboedo, line 72: I'm slightly confused about the return type being a tuple here - shouldn't the product identifier be the one from the product we're returning anyway? why not just return the product instead of a tuple, and then use product.identifier later?

RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
```diff
@@ -43,47 +43,103 @@ class LoadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType {
...
+                product: subscribedProduct,
+                promoOfferDetails: promoOfferDetails
+            ))
         } catch {
             Logger.warning(Strings.error_fetching_promotional_offer(error))
             return .failure(CustomerCenterError.couldNotFindOfferForActiveProducts)
         }
     }
 
+    private func getActiveSubscription(_ customerInfo: CustomerInfo) async throws -> (String, StoreProduct) {
```
aboedo, line 66: I don't feel particularly strong about these things so FFTI but in our style guide we suggest moving [private functions to a private extension](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/SwiftStyleGuide.swift#L119)

RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
```diff
@@ -43,47 +43,103 @@ class LoadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType {
...
+        productIdentifier: String,
+        promoOfferDetails: CustomerCenterConfigData.HelpPath.PromotionalOffer
+    ) throws -> StoreProductDiscount {
+        let discount = if !promoOfferDetails.productMapping.isEmpty {
+            findMappedDiscount(for: product,
+                               productIdentifier: productIdentifier,
+                               promoOfferDetails: promoOfferDetails)
+        } else {
+            findLegacyDiscount(for: product, promoOfferDetails: promoOfferDetails)
+        }
```
aboedo, line 86: TIL let foo = if ... else works. I would have used a ternary but I feel like this might be more readable

RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
```diff
@@ -43,47 +43,103 @@ class LoadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType {
...
+
+    private func findMappedDiscount(
+        for product: StoreProduct,
+        productIdentifier: String,
+        promoOfferDetails: CustomerCenterConfigData.HelpPath.PromotionalOffer
+    ) -> StoreProductDiscount? {
+        product.discounts.first { discount in
+            guard let offerIdentifier = discount.offerIdentifier else { return false }
+            return promoOfferDetails.productMapping[productIdentifier] == offerIdentifier
+        }
```
aboedo, line 104: since ProductMapping is [String : String] (so no `nil` values), this could be 
```swift
    product.discounts.first { $0.offerIdentifier == promoOfferDetails.productMapping[productIdentifier] }
```
right? 

Asking because that would also match what we do in line 112, so maybe that's easier to read if it looks the same?

RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
```diff
@@ -43,47 +43,103 @@ class LoadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType {
...
+    }
+
+    private func findLegacyDiscount(
+        for product: StoreProduct,
+        promoOfferDetails: CustomerCenterConfigData.HelpPath.PromotionalOffer
+    ) -> StoreProductDiscount? {
+        // Try exact match first
+        if let exactMatch = product.discounts.first(where: {
+            $0.offerIdentifier == promoOfferDetails.iosOfferId
+        }) {
```
aboedo, line 114: FFTI, but you can also do it like this
```suggestion
        if let exactMatch = product.discounts.first {
            $0.offerIdentifier == promoOfferDetails.iosOfferId
        } {
```

Then again maybe the brackets add some clarity since it ends in a `{` too

RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
```diff
@@ -43,47 +43,103 @@ class LoadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType {
...
+            return exactMatch
+        }
+
+        // Fall back to suffix matching
+        return product.discounts.first { discount in
+            guard let offerIdentifier = discount.offerIdentifier else { return false }
+            return offerIdentifier.hasSuffix("_\(promoOfferDetails.iosOfferId)")
+        }
```
aboedo, line 122: another totally ignorable one since this is already easy to read

```suggestion
        return product.discounts.first { $0.offerIdentifier?.hasSuffix("_\(promoOfferDetails.iosOfferId)") == true }
```

aboedo (approved):
left a few swifty things but nothing that's actually important, looks good!

fire-at-will (approved):

RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
vegaro: It actually fails because of the  ending `{`

PR #4490, opened 2024-11-15
[Paywalls V2] Process variables in the text component

joshdholtz:
## Motivation

Process the same [Paywalls V1 variables](https://www.revenuecat.com/docs/tools/paywalls/creating-paywalls#variables) but in the Paywalls V2 text component

## Description

- Renamed `PaywallState` to `PackageContext`
  - Stored on the paywall to store the "selected package"
  - Renamed because this is now used by the `TextComponentView` to get the any type of package context
      - Could be because text is INSIDE `Package` component
      - Could be because text is OUTSIDE `Package` component (like a CTA) and needs to get the selected package

### Demo

It works 👇 

Screenshot 2024-11-14 at 11 40 12 PM

### DIFF
--- a/RevenueCatUI/Templates/Components/Text/TextComponentViewModel.swift
+++ b/RevenueCatUI/Templates/Components/Text/TextComponentViewModel.swift
@@ -40,6 +40,7 @@ class TextComponentViewModel {
     func styles(
         state: ComponentViewState,
         condition: ScreenCondition,
+        packageContext: PackageContext,
         apply: @escaping (TextComponentStyle) -> some View
     ) -> some View {
         let localizedPartial = LocalizedTextPartial.buildPartial(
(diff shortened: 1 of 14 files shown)


RevenueCatUI/Templates/Components/Text/TextComponentViewModel.swift
```diff
@@ -66,6 +69,37 @@ class TextComponentViewModel {
...
+        let discount = Self.discount(
+            from: package.storeProduct.pricePerMonth?.doubleValue,
+            relativeTo: packageContext.variableContext.mostExpensivePricePerMonth
+        )
+
+        let context: VariableHandler.Context = .init(
+            discountRelativeToMostExpensivePerMonth: discount,
+            showZeroDecimalPlacePrices: packageContext.variableContext.showZeroDecimalPlacePrices
+        )
+        let locale = Locale.current
```
joshdholtz, line 86: See #4491 which reworks some things to get the locale into the view models

MarkVillacampa (approved):

PR #4497, opened 2024-11-17
[Paywalls V2] File organization restructure

joshdholtz:
## Motivation

The project structure needed some cleanup 

## Description

Most of this is organizing of the project's file structure, file splitting, and file renaming 👇 

Any code changes will be called out and linked specifically.

Screenshot 2024-11-18 at 5 06 27 AM

### Smarter initialization of stuff in `PaywallsV2View`

ℹ️  `TemplateComponentsView` renamed to `PaywallsV2View` and is the main entry point into the paywall components implementation.

Previously, the initializer of `PaywallsV2View` would parse the recursive tree components and create the view model tree on every initialize. This _seems_ like what we want but SwiftUI will construct `PaywallsV2View` multiple times (like anytime things need to be redrawn). To prevent the heavy computation from happening, all of this logic has been moved into the initializing of a  [PaywallStateManager](https://github.com/RevenueCat/purchases-ios/pull/4497/files#diff-afbe82dc9d8fcd1986950a2b7f871aa9ce26fe552987bb9b9133c013af2dd43eR59-R70) `StateObject`.

A `StateObject`'s `wrappedValue` parameter is an `@autoclosure` that will only ever get executed once. We are leveraging this to ensure the heavy instantiation only happens once and remains in a consistent state.

The `PaywallV2View` is a light wrapper around `LoadedPaywallV2View` which then creates all the property UI and environment objects to pass into the rendered component tree.


RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -0,0 +1,286 @@
...
+
+    let viewModelFactory: ViewModelFactory
+    let packages: [Package]
+    let componentViewModel: PaywallComponentViewModel
+    let showZeroDecimalPlacePrices: Bool
+
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+struct PaywallsV2View: View {
```
JayShortway, line 34: Some related comments around naming: 
- The use of `V2` in the name. It's not terrible, but also not very descriptive. 
- "Paywalls" being plural is a bit confusing. It might seem that this is going to show multiple paywalls at once, but it's not.
- If I understand correctly, this is at the same level architecturally as `LoadedOfferingPaywallView`. Would be good to name it consistently to indicate that.

Maybe these comments can be resolved by naming changes such as: 
- `LoadedOfferingPaywallView` -> `LoadedLegacyPaywallView` or `LoadedTemplatePaywallView`
- `PaywallsV2View` -> `LoadedPaywallView`. If we move the creation of `PaywallState` one level up, we may not need the existing `LoadedPaywallsV2View`?

JayShortway (approved):
Very nice improvement! I'm assuming there were no code changes, because none were called out. Had just a single suggestion around naming. Will approve to unblock you.

PR #4500, opened 2024-11-19
[Paywalls] Use .frame(alignment:) to fix alignment in non-multiline text components

MarkVillacampa:
### Checklist


joshdholtz (approved):
Ohhh, this was a good catch! Thanks for fixing this 🙌 :shipit:

PR #4503, opened 2024-11-20
Add lifetime support to the Customer Center

vegaro:
| Header | Header |
|--------|--------|
| Lifetime (there's an entitlement and a non subscription) | simulator_screenshot_B11812CC-D3EA-4320-8E48-8AAD5F16E810 |
| Monthly + Lifetime: displays lifetime because the entitlement is assigned to the Lifetime. We should display the monthly information but we are missing CustomerInfo information | !simulator_screenshot_B11812CC-D3EA-4320-8E48-8AAD5F16E810 |
| Base case monthly assigned to an entitlement | simulator_screenshot_40CA9FE8-DB93-4591-9253-CFAAC015AB43 |

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
+++ b/RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
@@ -25,10 +25,6 @@ struct SubscriptionInformation {
     let price: PriceDetails
     let expirationOrRenewal: ExpirationOrRenewal?
     let productIdentifier: String
-
-    let willRenew: Bool
-    let active: Bool
-
     let store: Store
 
     init(title: String,
@@ -47,8 +43,6 @@ struct SubscriptionInformation {
         self.price = price
         self.expirationOrRenewal = expirationOrRenewal
         self.productIdentifier = productIdentifier
-        self.willRenew = willRenew
-        self.active = active
         self.store = store
     }
 
@@ -66,16 +60,12 @@ struct SubscriptionInformation {
         self.price = price
         self.expirationOrRenewal = expirationOrRenewal
         self.productIdentifier = productIdentifier
-        self.willRenew = willRenew
-        self.active = active
         self.store = store
     }
 
     init(entitlement: EntitlementInfo,
          subscribedProduct: StoreProduct? = nil,
          dateFormatter: DateFormatter = DateFormatter()) {
-        // swiftlint:disable:next todo
-        // TODO: support non-consumables
         dateFormatter.dateStyle = .medium
 
         self.title = subscribedProduct?.localizedTitle
@@ -83,10 +73,33 @@ struct SubscriptionInformation {
         self.durationTitle = subscribedProduct?.subscriptionPeriod?.durationTitle
         self.price = entitlement.priceBestEffort(product: subscribedProduct)
         self.expirationOrRenewal = entitlement.expirationOrRenewal(dateFormatter: dateFormatter)
-        self.willRenew = entitlement.willRenew
         self.productIdentifier = entitlement.productIdentifier
-        self.active = entitlement.isActive
         self.store = entitlement.store
+    }
+
+    init(product: StoreProduct,
+         expirationDate: Date?,
+         dateFormatter: DateFormatter = DateFormatter()) {
+        // We don't have enough information to determine if the subscription will renew or not because we
+        // are loading the information from the product without entitlement information.
+        // We also assume that the subscription is active.
+        // We also assume that the subscription will renew the earliest possible renewal date and this is the
+        // product with the earliest renewal date.
+        dateFormatter.dateStyle = .medium
+
+        self.title = product.localizedTitle
+        self.explanation = .earliestRenewal
+        self.durationTitle = product.subscriptionPeriod?.durationTitle
+        self.price = .paid(product.localizedPriceString)
+        if let dateString = expirationDate.map({ dateFormatter.string(from: $0) }) {
+            let date = SubscriptionInformation.ExpirationOrRenewal.Date.date(dateString)
+            self.expirationOrRenewal = SubscriptionInformation.ExpirationOrRenewal(label: .expires,
+                                                                                   date: date)
+        } else {
+            self.expirationOrRenewal = nil
+        }
+        self.productIdentifier = product.productIdentifier
+        self.store = .appStore
     }
 
     struct ExpirationOrRenewal {
--- a/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
@@ -32,9 +32,9 @@ import RevenueCat
     private lazy var currentAppVersion: String? = currentVersionFetcher()
 
     @Published
-    private(set) var hasSubscriptions: Bool = false
+    private(set) var hasActiveProducts: Bool = false
     @Published
-    private(set) var subscriptionsAreFromApple: Bool = false
+    private(set) var hasAppleEntitlement: Bool = false
     @Published
     private(set) var appIsLatestVersion: Bool = defaultAppIsLatestVersion
 
@@ -93,12 +93,12 @@ import RevenueCat
     #if DEBUG
 
     convenience init(
-        hasSubscriptions: Bool = false,
-        areSubscriptionsFromApple: Bool = false
+        hasActiveProducts: Bool = false,
+        hasAppleEntitlement: Bool = false
     ) {
         self.init(customerCenterActionHandler: nil)
-        self.hasSubscriptions = hasSubscriptions
-        self.subscriptionsAreFromApple = areSubscriptionsFromApple
+        self.hasActiveProducts = hasActiveProducts
+        self.hasAppleEntitlement = hasAppleEntitlement
         self.state = .success
     }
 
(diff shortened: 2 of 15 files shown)


RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -106,18 +106,30 @@ import RevenueCat
 
     func loadHasSubscriptions() async {
```
vegaro, line 107: I know this method is becoming crazy and I am planning on refactoring, but I need to do it step by step otherwise it becomes very complex. If we add subscription information to the CustomerInfo this will be easier

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -97,16 +97,47 @@ class ManageSubscriptionsViewModel: ObservableObject {
...
+            $0.productIdentifier == earliestProductId
+        }) {
+            let subscriptionInformation = SubscriptionInformation(entitlement: entitlement,
+                                                                  subscribedProduct: product)
+            self.subscriptionInformation = subscriptionInformation
+        } else {
+            let expirationDate = customerInfo.expirationDate(forProductIdentifier: product.productIdentifier)
+            let subscriptionInformation = SubscriptionInformation(product: product,
+                                                                  expirationDate: expirationDate)
+            self.subscriptionInformation = subscriptionInformation
```
vegaro, line 139: I am planning on directly loading SubscriptionInformation in `CustomerCenterViewModel` to make this easier to understand and less error prone. But that comes in a separate PR

RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
```diff
@@ -66,29 +60,48 @@ struct SubscriptionInformation {
...
     }
 
+    init(product: StoreProduct,
+         expirationDate: Date?,
+         dateFormatter: DateFormatter = DateFormatter()) {
+        // We don't have enough information to determine if the subscription will renew or not because we
+        // are loading the information from the product without entitlement information.
+        // We also assume that the subscription is active.
+        // We also assume that the subscription will renew the earliest possible renewal date and this is the
+        // product with the earliest renewal date.
```
vegaro, line 87: We have to add the Subscription information to the CustomerInfo. It's already being returned in the JSON, we just don't have it in the `CustomerInfo` class. 

Once we have that, we can fix this.

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -93,31 +93,25 @@ import RevenueCat
...
         do {
-            // swiftlint:disable:next todo
-            // TODO: support non-consumables
             let customerInfo = try await self.customerInfoFetcher()
-            let hasSubscriptions = customerInfo.activeSubscriptions.count > 0
-
-            let subscriptionsAreFromApple = customerInfo.entitlements.active.contains(where: { entitlement in
-                entitlement.value.store == .appStore || entitlement.value.store == .macAppStore &&
```
vegaro, line 115: the `macAppStore` product wouldn't really be able to be loaded so considering that a non compatible platform that should load the `WrongPlatformView`

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -97,16 +97,49 @@ class ManageSubscriptionsViewModel: ObservableObject {
     private func loadSubscriptionInformation() async throws {
         let customerInfo = try await purchasesProvider.customerInfo()
 
-        guard let currentEntitlement = customerInfo.earliestExpiringAppStoreEntitlement(),
-              let subscribedProduct = await purchasesProvider.products([currentEntitlement.productIdentifier]).first
+        // Find earliest expiring subscription
+        guard let earliestProductId =
```
vegaro, line 101: This code is a bit crazy, but I am planning on refactoring it by moving all the logic earlier in the flow. It would be helpful to get the Subscription information in the `CustomerInfo` so until then, we can leave it like this

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -98,14 +98,14 @@ class ManageSubscriptionsViewModel: ObservableObject {
         let customerInfo = try await purchasesProvider.customerInfo()
 
         guard let currentEntitlement = customerInfo.earliestExpiringAppStoreEntitlement(),
```
vegaro, line 100: this code is not handling correctly the case when both a subscription and a lifetime grant the same entitlement. The issue is that the entitlement will be associated to the lifetime since that's what the backend returns, even though we want to display the monthly information.

We don't have enough information in the CustomerInfo at the moment to filter which subscriptions are Apple subscriptions, so we can't have a better logic for this. When we have that information, we'll be able to filter those out and don't need the entitlement to determine which subscription to show.

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -93,31 +93,25 @@ import RevenueCat
...
-        self.hasSubscriptions = hasSubscriptions
-        self.subscriptionsAreFromApple = areSubscriptionsFromApple
+        self.hasActiveProducts = hasActiveProducts
+        self.hasAppleEntitlement = hasAppleEntitlement
         self.state = .success
     }
 
     #endif
 
     func loadHasSubscriptions() async {
```
fire-at-will, line 107: It might be a good idea to rename this function since the model no longer has a property named `hasSubscriptions` and it's also loading non-subscriptions now

```suggestion
    func loadHasActivePurchases() async {
```

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
```diff
@@ -196,6 +196,139 @@ class ManageSubscriptionsViewModelTests: TestCase {
...
+                    purchaseDate: purchaseDateMonthly,
+                    expirationDate: expirationDateMonthly
+                ),
+                CustomerInfoFixtures.Subscription(
+                    id: productIdYearly,
+                    store: "app_store",
+                    purchaseDate: purchaseDateYearly,
+                    expirationDate: expirationDateYearly
+                )
+            ].shuffled(),
```
fire-at-will, line 238: Instead of using `.shuffled()`, you might consider making this a parameterized test and passing in `[monthlyProduct, yearlyProduct]` and `[yearlyProduct, monthlyProduct]` separately.

This way, we always test both cases instead of randomly testing one case

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
```diff
@@ -196,6 +196,139 @@ class ManageSubscriptionsViewModelTests: TestCase {
...
+        let subscriptionInformation = try XCTUnwrap(viewModel.subscriptionInformation)
+        expect(subscriptionInformation.title) == "lifetime"
+        expect(subscriptionInformation.durationTitle).to(beNil())
+        expect(subscriptionInformation.price) == .paid("$29.99")
+        expect(subscriptionInformation.productIdentifier) == productIdLifetime
+
+        expect(subscriptionInformation.expirationOrRenewal?.date) == .never
+    }
+
+    func testShouldShowLifetim_whenUserHasLifetimeOneEntitlement() async throws {
```
fire-at-will, line 279:
```suggestion
    func testShouldShowLifetime_whenUserHasLifetimeOneEntitlement() async throws {
```

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
```diff
@@ -196,6 +196,139 @@ class ManageSubscriptionsViewModelTests: TestCase {
         expect(subscriptionInformation.productIdentifier) == productIdOne
     }
 
+    func testShouldShowEarliestExpiration_whenUserHasLifetimeAndSubscriptionsOneEntitlement() async throws {
```
fire-at-will, line 199: I think this test is ensuring that the lifetime purchase is shown, could the test be named this?

```suggestion
    func testShouldShowLifetime_whenUserHasLifetimeAndSubscriptionsOneEntitlement() async throws {
```

RevenueCatUI/CustomerCenter/Data/SubscriptionInformation.swift
```diff
@@ -25,10 +25,6 @@ struct SubscriptionInformation {
     let price: PriceDetails
     let expirationOrRenewal: ExpirationOrRenewal?
     let productIdentifier: String
-
-    let willRenew: Bool
```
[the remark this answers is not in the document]
vegaro: good point! Actually, in the update to the `CustomerInfo` I am working on I had another `SubscriptionInfo` so this change fits perfectly

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
vegaro: good idea. I did my own version of a parameterized test by making sure both cases get always tested. I don't think we have any other parameterized test in the source code and was spending too much time trying to build somehting generic lol

vegaro:
@RCGitBot please test

Tests/RevenueCatUITests/CustomerCenter/ManageSubscriptionsViewModelTests.swift
JayShortway: We have this existing parameterized test 😄 

[`CustomerCenterViewModelTests.testAppIsLatestVersion()`](https://github.com/RevenueCat/purchases-ios/blob/4a0156e1f63fb41c975d8df9d1e819a3ab3e7deb/Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift#L129)

PR #4508, opened 2024-11-21
Adds `subscriptions` to `CustomerInfo`

vegaro:
We had more information we were not exposing in the CustomerInfo JSON

This PR adds `CustomerInfo.subscriptions` which has some useful information for devs that don't use the entitlements system

### DIFF
--- a/Sources/Identity/CustomerInfo.swift
+++ b/Sources/Identity/CustomerInfo.swift
@@ -12,6 +12,7 @@
 //  Created by Madeline Beyl on 7/9/21.
 //
 
+// swiftlint:disable file_length
 import Foundation
 
 /**
--- a/Sources/Identity/CustomerInfo.swift
+++ b/Sources/Identity/CustomerInfo.swift
@@ -29,7 +29,9 @@ public typealias ProductIdentifier = String
     @objc public let entitlements: EntitlementInfos
 
     /// All *subscription* product identifiers with expiration dates in the future.
-    @objc public var activeSubscriptions: Set<ProductIdentifier> { self.activeKeys(dates: self.expirationDatesByProductId) }
+    @objc public var activeSubscriptions: Set<ProductIdentifier> {
+        self.activeKeys(dates: self.expirationDatesByProductId)
+    }
 
     /// All product identifiers purchases by the user regardless of expiration.
     @objc public let allPurchasedProductIdentifiers: Set<ProductIdentifier>
--- a/Sources/Identity/CustomerInfo.swift
+++ b/Sources/Identity/CustomerInfo.swift
@@ -225,7 +225,8 @@ import Foundation
                 ownershipType: subscriptionData.ownershipType,
                 periodType: subscriptionData.periodType,
                 refundedAt: subscriptionData.refundedAt,
-                storeTransactionId: subscriptionData.storeTransactionId
+                storeTransactionId: subscriptionData.storeTransactionId,
+                requestDate: response.requestDate
             )
         }
     }
--- a/Sources/Identity/CustomerInfo.swift
+++ b/Sources/Identity/CustomerInfo.swift
@@ -15,6 +15,11 @@
 import Foundation
 
 /**
+ An identifier used to identify a product.
+ */
+public typealias ProductIdentifier = String
+
+/**
  A container for the most recent customer info returned from `Purchases`.
  These objects are non-mutable and do not update automatically.
  */
@@ -24,10 +29,10 @@ import Foundation
     @objc public let entitlements: EntitlementInfos
 
     /// All *subscription* product identifiers with expiration dates in the future.
-    @objc public var activeSubscriptions: Set<String> { self.activeKeys(dates: self.expirationDatesByProductId) }
+    @objc public var activeSubscriptions: Set<ProductIdentifier> { self.activeKeys(dates: self.expirationDatesByProductId) }
 
     /// All product identifiers purchases by the user regardless of expiration.
-    @objc public let allPurchasedProductIdentifiers: Set<String>
+    @objc public let allPurchasedProductIdentifiers: Set<ProductIdentifier>
 
     /// Returns the latest expiration date of all products, nil if there are none.
     @objc public var latestExpirationDate: Date? {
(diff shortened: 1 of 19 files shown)


Sources/Identity/CustomerInfo.swift
```diff
@@ -88,6 +88,9 @@ import Foundation
      */
     @objc public let originalApplicationVersion: String?
 
+    /// Dictionary of all subscription product identifiers and their subscription info
+    @objc public let subscriptions: [String: SubscriptionInfo]
```
aboedo, line 92: my first thought was that it's a bit confusing that this is a dict but `nonSubscriptions` is an array. I get _why_, but naming-wise I'd maybe have expected both to be the same. Wonder if `subscriptionsByProductIdentifier` would have helped, but not sure. Or maybe even typealiasing `ProductIdentifier = String` so that then you have the type `[ProductIdentifier: SubscriptionInfo]`. 
Then again you could make the argument that we should have all APIs do that or none ¯\\_(ツ)_/¯

Sources/Purchasing/NonSubscriptionTransaction.swift
```diff
@@ -34,19 +34,24 @@ public final class NonSubscriptionTransaction: NSObject {
     /// The unique identifier for the transaction created by the Store.
     @objc public let storeTransactionIdentifier: String
 
+    /**
+     * The ``Store``tore where this transaction was performed from.
```
aboedo, line 38:
```suggestion
     * The ``Store`` where this transaction was performed.
```

aboedo (approved):
I think this is fine. 

But there's one thing that just feels a bit odd to me in the fact that we're using our own public REST API, but then we abstract it privately, then make it public again. 

It's not like it's a very internal REST API either - it's very much public and stable and we wouldn't just modify it out the blue since it would break older SDKs, which makes me feel like we're overcomplicating things a bit

aboedo:
or rather, I guess what I'm asking is "should we just make the `subscriber` object public instead"

Sources/Identity/CustomerInfo.swift
JayShortway: Instead of a type alias, a [value class](https://kotlinlang.org/docs/inline-classes.html) is also a great candidate for that `ProductIdentifier`. But you'd still have this argument, which I agree with: 
>Then again you could make the argument that we should have all APIs do that or none ¯\(ツ)/¯

Sources/Identity/SubscriptionInfo.swift
```diff
@@ -0,0 +1,134 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  SubscriptionInfo.swift
+//
+//  Created by Cesar de la Vega on 21/11/24.
+
+import Foundation
+
+/// Subscription purchases of the Customer
+@objc(RCSubscriptionInfo) public final class SubscriptionInfo: NSObject {
```
JayShortway, line 17: I think having this available is such a big improvement 🙌

Sources/Networking/Responses/CustomerInfoResponse.swift
```diff
@@ -49,7 +49,7 @@ extension CustomerInfoResponse {
 
         @IgnoreDecodeErrors<PeriodType>
         var periodType: PeriodType
-        var purchaseDate: Date?
+        var purchaseDate: Date
```
JayShortway, line 52: Just checking that we know this is always available, and having it optional was being too defensive?

JayShortway (approved):
I'm pretty excited about this! 😄 No comments other than Andy's.

Sources/Identity/SubscriptionInfo.swift
```diff
@@ -0,0 +1,134 @@
...
+    @objc public let periodType: PeriodType
+
+    /// Date when RevenueCat detected a refund of this subscription.
+    @objc public let refundedAt: Date?
+
+    /// The transaction id in the store of the subscription.
+    @objc public let storeTransactionId: String?
+
+    /// Whether the subscription is currently active.
+    @objc public let isActive: Bool
```
tonidero, line 70: Small concern but I'm wondering if we should have a `VerificationResult` (from trusted entitlements) here as well, or maybe move the existing one in `EntitlementInfos` up to the `CustomerInfo` entity... otherwise some devs might use this without checking trusted entitlements...

Sources/Identity/CustomerInfo.swift
tonidero: It's also a bit confusing with `activeSubscriptions` being only the subs ids and this being full objects... Maybe something like `allSubscriptionsInfo`? Or maybe too verbose not sure...

tonidero (approved):
Agreed this is great! Left some comments but I think this could work as is.

Sources/Identity/CustomerInfo.swift
JayShortway: Maybe we could deprecate `activeSubscriptions` as part of this change? Not a ton of value in that anymore, and very easy to recreate with `subscriptions`.

Sources/Networking/Responses/CustomerInfoResponse.swift
vegaro: I checked in the backend and in Android and it's not optional there

Sources/Identity/CustomerInfo.swift
vegaro: yes, @fire-at-will made the same comment that maybe `subscriptions` could be a `Set`. Imho I think a dictionary makes more sense since most people using this would want to get a subscription by product identifier, and making it a `Set` forces them to use a filter. Making it a dictionary offers accessing by product id and also offers a `.values` to access only the values.

Thoughts:
- `activeSubscriptions` should really be `activeSubscriptionProductIds` or something like that
- I think I would not want to deprecate `activeSubscriptions`, maybe in the next major, but I still see `activeSubscriptions` valuable since it saves folks from having to do logic on `subscriptions` to get the active ones. I would treat it as syntactic sugar.
- `allSubscriptionsInfo` also makes sense, but I don't think it cleans up the fact that `activeSubscriptions` is just the product ids and `allSubscriptionsInfo` classes. I think it would still missing the informatino that it is a dictionary
- `subscriptionsByProductIdentifier` makes a lot of sense. And the typealias `ProductIdentifier` can also be added. I think that would be the most informative and most correct. I think `byProductIdentifier` is important so we can replicate it in the rest of the SDKs that don't have type aliases

vegaro:
@tonidero @aboedo @JayShortway made some updates in [`0d388c5` (#4508)](https://github.com/RevenueCat/purchases-ios/pull/4508/commits/0d388c5012cd03870f1bd447289e4abbcd87dca0), let me know what you think

tonidero (approved):
:shipit:

PR #4509, opened 2024-11-21
[Paywalls] Add button and shortcut to refresh the presented live paywall

MarkVillacampa:
In order to make it faster to iterate when designing a paywall, this PR introduces a button in the `Sandbox Paywalls` which refreshes the offering list.

Additionally, the `shift + r` shortcut will refresh the offerings, and if a paywall is already presented, it will be reopened with the latest changes.

Limitation: because it's using `syncAttributesAndOfferingsIfNeeded` to force refresh the offerings, it is limited to 5 refreshes every 60 seconds.

Simulator_Screenshot_-_iPhone_16_Pro_-_2024-11-21_at_18_31_35


joshdholtz (approved):
:shipit: :shipit: :shipit:

PR #4511, opened 2024-11-21
[Paywalls] Fix stack alignment issues by applying frame alignment to the size modifier

MarkVillacampa:
So far, we were setting the alignment in the H/VStack initialiser, but that affects the alignment of the children items relative to each other. In order to affect their alignment with respect to their container, we need to apply the `.frame(alignment:)` modifier as well.


joshdholtz (approved):
:shipit: :shipit: :shipit: :shipit:

PR #4515, opened 2024-11-22
Refactors the creation of the subscription details in Customer Center

vegaro:
It's possible developers don't use the entitlements system. We can still get a lot of the information we need from CustomerInfo. This PR refactors the creation of the subscription details in Customer Center after the addition of `CustomerInfo.subscriptionsByProductIdentifier`, making it support more scenarios, like when entitlements are not present in the customer info.


RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -32,9 +32,7 @@ struct PurchaseInformation {
          explanation: Explanation,
          price: PriceDetails,
          expirationOrRenewal: ExpirationOrRenewal?,
-         willRenew: Bool,
          productIdentifier: String,
-         active: Bool,
```
vegaro, line 37: These were only used in this file and we don't need them anymore since the information is in the product

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -236,3 +220,36 @@ fileprivate extension String {
     }
 
 }
+
+protocol Transaction {
```
vegaro, line 224: This is probably a bad name, there are too many things named `Transaction` already

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -23,17 +23,14 @@ import RevenueCat
...
 
     typealias CurrentVersionFetcher = () -> String?
 
     private lazy var currentAppVersion: String? = currentVersionFetcher()
-
     @Published
-    private(set) var hasActiveProducts: Bool = false
-    @Published
-    private(set) var hasAppleEntitlement: Bool = false
+    private(set) var purchaseInformation: PurchaseInformation?
```
vegaro, line 33: Big improvement is holding the `purchaseInformation` here instead of in `ManageSubscriptionsViewModel`. That way most of the logic happens in this viewmodel and we don't need to get customer info anywhere else

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -157,21 +159,4 @@ private extension CustomerCenterView {
...
 
-#if DEBUG
-
-@available(iOS 15.0, *)
-@available(macOS, unavailable)
-@available(tvOS, unavailable)
-@available(watchOS, unavailable)
-struct CustomerCenterView_Previews: PreviewProvider {
-
-   static var previews: some View {
```
fire-at-will, line 168: It'd be nice to be able to keep this if it's possible - maybe we could mock out a `PurchaseInformation` object for the view model to use?

fire-at-will (commented):
Looking good so far! I _love_ that this removes the loading state entirely from the `ManageSubscriptionsView` 😎

I had a question or two and found a potential bug when `loadPurchaseInformation` is called multiple times, let me know what you think!

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
vegaro: good idea, I think I can mock the three different cases (Active, no active, and wrong platform)

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -46,60 +44,46 @@ struct PurchaseInformation {
...
+            self.expirationOrRenewal = entitlement.expirationOrRenewal(dateFormatter: dateFormatter)
+            self.productIdentifier = entitlement.productIdentifier
+            self.store = entitlement.store
+            self.price = entitlement.priceBestEffort(product: subscribedProduct)
         } else {
-            self.expirationOrRenewal = nil
+            switch transaction.type {
+            case .subscription(let isActive, let willRenew, let expiresDate):
+                self.explanation = expiresDate != nil
+                    ? (isActive ? (willRenew ? .earliestRenewal : .earliestExpiration) : .expired)
```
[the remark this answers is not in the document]
vegaro: oh so these are just little texts that appear at the top of the view. 

subEarliestRenewal: "This is your subscription with the earliest billing date."
subEarliestExpiration: "This is your subscription with the earliest expiration date."

So the logic is that if the sub is active and it will renew, it should say "billing date" and if it won't renew it should say "expiration date"

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -86,28 +83,28 @@ import RevenueCat
...
             let customerInfo = try await purchasesProvider.customerInfo()
-            self.hasActiveProducts = customerInfo.activeSubscriptions.count > 0 ||
-                                customerInfo.nonSubscriptions.count > 0
-            self.hasAppleEntitlement = customerInfo.entitlements.active.contains { entitlement in
-                entitlement.value.store == .appStore
+            let hasActiveProducts =
+            customerInfo.activeSubscriptions.count > 0 || customerInfo.nonSubscriptions.count > 0
+
+            if !hasActiveProducts {
+                self.state = .success
```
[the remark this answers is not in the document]
vegaro: I think that makes sense!

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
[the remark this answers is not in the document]
fire-at-will: Oohhhh, so it's the subscription with the earliest billing date in the _future_! I read it as "earliest billing date...in the past" and was confused 😅

Makes sense!

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
vegaro: I ended up adding just one preview for a monthly apple product, the other ones were becoming a bit of hell to support

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -83,19 +83,35 @@ import RevenueCat
         self.purchasesProvider = purchasesProvider
     }
 
+    #if DEBUG
```
fire-at-will, line 86: Nit: I might put this inside a fileprivate extension to separate it from the production code

fire-at-will (approved):

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
vegaro: It's used in CustomerCenterView as well :/

PR #4522, opened 2024-11-26
[Paywalls] Add support for gradient backgrounds

MarkVillacampa:
This PR adds support for gradient backgrounds.

Supported gradient types are Radial and Linear.

Screenshot 2024-11-26 at 11 35 52
Screenshot 2024-11-26 at 11 36 03

Additionally, gradients as text foreground style are now supported:

Screenshot 2024-11-27 at 01 17 02

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Text/TextComponentViewModel.swift
+++ b/RevenueCatUI/Templates/V2/Components/Text/TextComponentViewModel.swift
@@ -157,7 +157,7 @@ struct TextComponentStyle {
     let visible: Bool
     let text: String
     let fontWeight: Font.Weight
-    let color: Color
+    let color: PaywallComponent.ColorScheme
     let fontSize: Font
     let horizontalAlignment: Alignment
     let textAlignment: TextAlignment
@@ -182,7 +182,7 @@ struct TextComponentStyle {
         self.visible = visible
         self.text = text
         self.fontWeight = fontWeight.fontWeight
-        self.color = color.toDynamicColor()
+        self.color = color
 
         // WIP: Take into account the fontFamily mapping
         self.fontSize = fontSize.font
--- a/RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
+++ b/RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
@@ -20,7 +20,6 @@ enum BackgroundStyle {
 
     case color(PaywallComponent.ColorScheme)
     case image(PaywallComponent.ThemeImageUrls)
-    case gradient
 
 }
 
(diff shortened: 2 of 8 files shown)


RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
```diff
@@ -20,7 +20,6 @@ enum BackgroundStyle {
 
     case color(PaywallComponent.ColorScheme)
     case image(PaywallComponent.ThemeImageUrls)
-    case gradient
```
MarkVillacampa, line 23: Following the spec, gradient is not a different type of background but is encoded as part of the `ColorInfo` struct.

RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
```diff
@@ -47,7 +46,26 @@ fileprivate extension View {
     func apply(backgroundStyle: BackgroundStyle) -> some View {
         switch backgroundStyle {
         case .color(let color):
-            self.background(color.toDynamicColor())
+            switch color.light {
```
MarkVillacampa, line 49: My assumption is the light and dark colors must be of the same type e.g. both solid or both gradient.

MarkVillacampa: Modified to support mixed solid and gradient light/dark colors in ColorScheme

RevenueCatUI/Templates/V2/Components/Text/TextComponentViewModel.swift
```diff
@@ -157,7 +157,7 @@ struct TextComponentStyle {
     let visible: Bool
     let text: String
     let fontWeight: Font.Weight
-    let color: Color
+    let color: PaywallComponent.ColorScheme
```
MarkVillacampa, line 160: Storing the `ColorScheme` instead of the dynamic color now because it can potentially be a gradient.

RevenueCatUI/Views/GradientView.swift
```diff
@@ -0,0 +1,84 @@
...
+#if PAYWALL_COMPONENTS
+
+struct GradientView: View {
+    enum GradientStyle {
+        case linear(Int)
+        case radial
+    }
+
+    @Environment(\.colorScheme)
+    private var colorScheme
```
MarkVillacampa, line 25: For completeness, and in case we want to use use `GradientView` somewhere else,  I've kept support for light/dark mode here even though we're controlling it in `BackgroundStyle` and `ForegroundColorScheme`.

joshdholtz (approved):
This is sooooooo good! Thank you 😁

RevenueCatUI/Views/GradientView.swift
```diff
@@ -0,0 +1,84 @@
...
+    init(angle: Angle) {
+        // Convert the angle to radians
+        let radians = angle.radians
+
+        // Calculate the normalized x and y positions
+        let xPosition = cos(radians)
+        let yPosition = sin(radians)
+
+        // Determine the scaling factor to move the point to the edge of the enclosing square
+        let scaleFactor = max(abs(xPosition), abs(yPosition))
```
JayShortway, line 72: Thanks a lot for the inspiration here! This was very useful for Android too. 🙏 

Quick question: does this work for non-square shapes in SwiftUI, like rectangles? In Compose, I find that the angle isn't quite correct when using a rectangle, as it indeed seems to "pretend" it's a square and draws the angle based on that. Not sure how strict we are with the desired result. There's a bunch of discussion and suggested Compose solutions on [this StackOverflow question](https://stackoverflow.com/q/68218714).

Here's an example of what I mean. Both are set to an angle of 45 degrees, but only the square is actually 45 degrees.

RevenueCatUI/Views/GradientView.swift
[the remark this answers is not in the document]
MarkVillacampa: Added 👍

RevenueCatUI/Views/GradientView.swift
MarkVillacampa: Will handle in a new PR 👍

PR #4528, opened 2024-11-29
Add feedback survey option chosen event

vegaro:
Creates an event with this definition:

```
struct CustomerCenterEventSurveyOptionChosen {
    let id: String?
    var type: EventType // customer_center_survey_option_chosen
    var appUserID: String
    var appSessionID: String
    var timestamp: UInt64
    var path: String // CUSTOM_URL
    var url: String? // URL if CUSTOM_URL
    var surveyOptionID: String
    var additionalContext: String? // null for now until we support
    var darkMode: Bool
    var locale: String
    var isSandbox: Bool
    var displayMode: FULL_SCREEN | SHEET
    var revisionId: Int
}
```


RevenueCatUI/CustomerCenter/ViewModels/FeedbackSurveyViewModel.swift
```diff
@@ -57,10 +57,13 @@ class FeedbackSurveyViewModel: ObservableObject {
 
     func handleAction(
         for option: CustomerCenterConfigData.HelpPath.FeedbackSurvey.Option,
+        darkMode: Bool,
+        displayMode: CustomerCenterPresentationMode,
         dismissView: () -> Void
     ) async {
         if let customerCenterActionHandler = self.customerCenterActionHandler {
             customerCenterActionHandler(.feedbackSurveyCompleted(option.id))
+            trackSurveyOptionChosen(option: option, darkMode: darkMode, displayMode: displayMode)
```
tonidero, line 66: Hmm I'm wondering if we should do this before calling the handler so we make sure the event isn't lost in case of some error on the client handling code?

Sources/CustomerCenter/Events/CustomerCenterEvent.swift
```diff
@@ -13,70 +13,127 @@
...
+        self.id = id
+        self.date = date
     }
 
+}
+
+/// An event to be sent by the `RevenueCatUI` SDK.
+public enum CustomerCenterEvent: FeatureEvent, CustomerCenterEventType {
+
+    var eventDiscriminator: String? { "lifecycle" }
```
tonidero, line 45: Hmm I'm confused by this default event discriminator... Should this be something like `customer_center`? We could also make it a computed property based on the enum values I guess.

Sources/Events/Networking/EventsRequest.swift
```diff
@@ -32,10 +32,18 @@ struct EventsRequest {
                 }
                 return AnyEncodable(event)
             case .customerCenter:
-                guard let event = CustomerCenterEvent(storedEvent: storedEvent) else {
-                    return nil
+                switch storedEvent.eventDiscriminator {
+                case "survey_option_chosen":
```
tonidero, line 36: Hmm can we store these into some sort of enum? Would be better to avoid hardcoded strings

Sources/CustomerCenter/Events/CustomerCenterEvent.swift
```diff
@@ -13,70 +13,127 @@
...
+
     /// The Customer Center was displayed.
-    case impression(CreationData, Data)
+    case impression(CustomerCenterEventCreationData, Data)
+
+}
+
+/// An event to be sent by the `RevenueCatUI` SDK.
+public enum CustomerCenterSurveyOptionChosenEvent: FeatureEvent, CustomerCenterEventType {
```
tonidero, line 53: Hmm probably missing it but why does this need to be a different class like this? could it be just part of the `CustomerCenterEvent` enum?

tonidero (commented):
I'm probably missing why we need the `CustomerCenterSurveyOptionChosenEvent` instead of relying on `CustomerCenterEvent`

JayShortway (commented):
Looks good! Just a naming suggestion, besides Toni's comments.

Sources/CustomerCenter/Events/CustomerCenterEvent.swift
vegaro: The Customer Center events are more complex than the Paywall Events, and one single class can't represent all of them.

I should have linked the document with the definition but here it is https://docs.google.com/document/d/1uZ5-obuQXWDGZ-1zKh8qaVYyJOK8bDpEevyWPdXTODA/edit?usp=sharing 

As you can tell one single class can't really represent all of them. We could have created one with a bunch of nullables, but this way is cleaner. The only complication is differentiating them when deserializing them, and that's why I am using that `eventDiscriminator` to distinguish between the different types of Customer Center events when deserializing them from storage.

Sources/CustomerCenter/Events/CustomerCenterEvent.swift
```diff
@@ -13,70 +13,127 @@
...
-        public var isSandbox: Bool
-        public var displayMode: CustomerCenterPresentationMode
+        public var localeIdentifier: String { base.localeIdentifier }
+        public var darkMode: Bool { base.darkMode }
+        public var isSandbox: Bool { base.isSandbox }
+        public var displayMode: CustomerCenterPresentationMode { base.displayMode }
+        public let path: CustomerCenterConfigData.HelpPath.PathType
+        public let url: URL?
+        public let surveyOptionID: String
+        public let surveyOptionTitleKey: String
```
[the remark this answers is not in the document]
vegaro: Renamed. I will keep the event in the backend the same for the sake of shipping though

Sources/Events/Networking/EventsRequest.swift
vegaro: oh I need to do this

Tests/UnitTests/CustomerCenter/Events/CustomerCenterEventsRequestTests.swift
```diff
@@ -31,15 +31,16 @@ class CustomerCenterEventsRequestTests: TestCase {
         let storedEvent: StoredEvent = try XCTUnwrap(.init(event: event,
                                                            userID: Self.userID,
                                                            feature: .customerCenter,
-                                                           appSessionID: Self.appSessionID))
-        let requestEvent: EventsRequest.CustomerCenterEvent = try XCTUnwrap(.init(storedEvent: storedEvent))
+                                                           appSessionID: Self.appSessionID,
+                                                           eventDiscriminator: "lifecycle"))
```
tonidero, line 35: Can potentially use the new discriminator enum?

vegaro: yes. Thanks!

PR #4529, opened 2024-12-02
[Paywalls] Add Emerge Snapshot Tests

MarkVillacampa:
This PR:
- Adds a test target to `PaywallTester`.
- Adds Emerge's `SnapshotPreviews` package to the test target.
- Adds Emerge's Fastlane plugin and a new lane which uploads `PaywallsTester` to Emerge.
- Adds a CircleCI job to run the lane.

Shoutout to @noahsmartin for all the help!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1348,6 +1348,7 @@ workflows:
       - pod-lib-lint
       - run-revenuecat-ui-ios-17
       - run-revenuecat-ui-ios-18
+      - emerge_purchases_ui_snapshot_tests
 
   create-tag:
     when:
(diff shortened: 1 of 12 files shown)


joshdholtz (commented):
This looks so good! Just a few small naming suggestions 😊 

But also, is there a reason to not run this on every PR? I mean, not every commit is doing a UI change but we do run it on every PR for Android (see [here](https://github.com/RevenueCat/purchases-android/pull/1951)) and its kind of nice, IMO 😇 

Or if we can't do that... it might be nice to have another command that is simlar to the "test all" one but only for snapshots? 🤷‍♂️ Thoughts?

aboedo (approved):

PR #4530, opened 2024-12-02
Use `ManageSubscriptionsView` for users without active subscriptions

vegaro:
We let devs configure the screen that's displayed to users with NO active subscriptions, but the `NoSubscriptionsView` is not dynamically built. This PR changes the logic so `ManageSubscriptionsView` is also used with the data in the `NO_ACTIVE` screen of the configuration.

I had to change the view a little bit since the Missing purchase option is optional and it doesn't make sense it always displays the "We can try checking your Apple account for any previous purchases"

| Before | After |
|--------|--------|
| Screenshot 2024-12-02 at 17 54 02 | Screenshot 2024-12-02 at 17 54 02|

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
@@ -112,7 +112,13 @@ private extension CustomerCenterView {
                 WrongPlatformView()
             }
         } else {
-            NoSubscriptionsView(configuration: configuration)
+            if let screen = configuration.screens[.noActive] {
+                ManageSubscriptionsView(screen: screen,
+                                        customerCenterActionHandler: viewModel.customerCenterActionHandler)
+            } else {
+                // Fallback with a restore button
+                NoSubscriptionsView(configuration: configuration)
+            }
         }
     }
 
--- a/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsButtonsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsButtonsView.swift
@@ -1,0 +1,77 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  ManageSubscriptionsButtonsView.swift
+//
+//  Created by Cesar de la Vega on 2/12/24.
+
+import Foundation
+import RevenueCat
+import SwiftUI
+
+#if os(iOS)
+
+@available(iOS 15.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+struct ManageSubscriptionsButtonsView: View {
+
+    @ObservedObject
+    var viewModel: ManageSubscriptionsViewModel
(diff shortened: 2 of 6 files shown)


RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsButtonsView.swift
```diff
@@ -0,0 +1,77 @@
...
+    @Binding
+    var loadingPath: CustomerCenterConfigData.HelpPath?
+    @Environment(\.openURL)
+    var openURL
+
+    @Environment(\.localization)
+    private var localization: CustomerCenterConfigData.Localization
+
+    var body: some View {
+        let filteredPaths = self.viewModel.screen.paths.filter { path in
```
aboedo, line 37: kinda feel like this is something that the viewModel should be able to return, this is a lot of `.thing.otherthing.filter` of stuff that's not a part of the view

aboedo: (I realize that this is just moved over from before, so FFTI, but I hadn't seen that code yet)

aboedo (commented):
looks good! Left a couple of minor comments

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -112,7 +112,13 @@ private extension CustomerCenterView {
                 WrongPlatformView()
             }
         } else {
-            NoSubscriptionsView(configuration: configuration)
+            if let screen = configuration.screens[.noActive] {
+                ManageSubscriptionsView(screen: screen,
```
tonidero, line 116: We probably should now rename the `ManageSubscriptionsView` to something more generic... Can come in a cleanup PR if we're in a hurry though.

tonidero (commented):
LGTM! Just Andy's comments and a rename suggestion

tonidero (approved):

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -86,17 +86,33 @@ struct ManageSubscriptionsView: View {
...
+                                                           loadingPath: self.$viewModel.loadingPath)
+                        } header: {
+                            if let subtitle = self.viewModel.screen.subtitle {
+                                Text(subtitle)
+                                    .textCase(nil)
+                            }
+                        }
+                    } else {
+                        let fallbackDescription = "You currently have no active subscriptions. " +
+                        "We can try checking your Apple account for any previous purchases"
```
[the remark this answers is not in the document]
vegaro: Thank you for the suggestion, I like it. I wasn't convinced either and copy is clearly not my strength haha

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsButtonsView.swift
vegaro: but you have a point. When I was moving this over I stopped in this part of the code a bit to think what it was doing. I think we can move it there (viewmodel) at least

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
vegaro: yes, we have to. I am going to postpone it to prevent conflicts with all the PRs I have opened. But I promise I will do a cleanup on naming 🙏

PR #4531, opened 2024-12-03
Fix build for app extensions

vegaro:
Customer Center broke the build for app extensions, this PR brings an approach similar to https://github.com/RevenueCat/purchases-ios/blob/98e4a3322528e484c384fe535ad2d9ebba0529cf/Sources/Support/ManageSubscriptionsHelper.swift#L107-L109 to the Customer Center


tonidero (approved):
LGTM!

PR #4533, opened 2024-12-04
Customer Center deeplinks should always be opened externally

vegaro:
Deeplinks should always be configured as `external`, if someone configures a deeplink as inapp, we'll try to open that link with a Safari view, which crashes. This PR makes sure deeplinks are always opened correctly.

There should be an update in the UI as well to make sure deeplinks are configured correctly

### DIFF
--- a/RevenueCatUI/CustomerCenter/URLUtilities.swift
+++ b/RevenueCatUI/CustomerCenter/URLUtilities.swift
@@ -77,7 +77,12 @@ enum URLUtilities {
 extension URL {
 
     var isDeeplink: Bool {
-        !(scheme?.hasPrefix("http") ?? false)
+        switch scheme?.lowercased() {
+        case "http", "https":
+            return false
+        default:
+            return true
+        }
     }
 
 }
--- a/RevenueCatUI/CustomerCenter/URLUtilities.swift
+++ b/RevenueCatUI/CustomerCenter/URLUtilities.swift
@@ -74,6 +74,14 @@ enum URLUtilities {
 
 }
 
+extension URL {
+
+    var isDeeplink: Bool {
+        !(scheme?.hasPrefix("http") ?? false)
+    }
+
+}
+
 private extension URLUtilities {
 
     static var isAppExtension: Bool {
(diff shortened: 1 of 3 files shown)


RevenueCatUI/CustomerCenter/URLUtilities.swift
```diff
@@ -74,6 +74,14 @@ enum URLUtilities {
 
 }
 
+extension URL {
+
+    var isDeeplink: Bool {
+        !(scheme?.hasPrefix("http") ?? false)
```
tonidero, line 80: I wonder if we should match exactly `http` and `https`... Unlikely that someone would create a deep link schema like `httpmyschema` but 🤷

tonidero (approved):

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -226,10 +226,13 @@ private extension ManageSubscriptionsViewModel {
...
             }
             switch openMethod {
-            case .external:
+            case .external,
+                _ where url.isDeeplink:
                 URLUtilities.openURLIfNotAppExtension(url)
             case .inApp:
                 self.inAppBrowserURL = .init(url: url)
+            @unknown default:
+                URLUtilities.openURLIfNotAppExtension(url)
```
tonidero, line 235: Might be worth logging if entering here?

RevenueCatUI/CustomerCenter/URLUtilities.swift
vegaro: not that crazy for someone doing that. If it's an app that starts with http it could happen

aboedo (approved):

RevenueCatUI/CustomerCenter/URLUtilities.swift
```diff
@@ -74,6 +74,19 @@ enum URLUtilities {
 
 }
 
+extension URL {
+
+    var isDeeplink: Bool {
+        switch scheme?.lowercased() {
+        case "http", "https":
+            return false
```
JayShortway, line 82: On Android, clicking e.g. a Twitter link will deeplink into the Twitter app. That's an `https` URL. Is something like that possible on iOS? If so, we can't assume it's not a deeplink based on the scheme.

vegaro: the main issue is that `self.inAppBrowserURL = .init(url: url)` crashes if it's not `http` or `https`. But you have a point so maybe `isDeeplink` is not that accurate. I do think `isDeeplink` is easy to understand what it does tho

PR #4535, opened 2024-12-04
Renames isDeeplink to isWebLink

vegaro:
Followup on #4533 after @JayShortway comment on deeplinks also having http as prefix


JayShortway (approved):
Nice, thanks for the rename!

PR #4537, opened 2024-12-04
Adds missing revisionId to CustomerCenter impression event

vegaro:
We were missing the revision id for Customer Center impression events which the backend expects. I had that in another PR but not in `main`

It's hardcoded to 1 for now, since the config JSON doesn't have revision ids yet


fire-at-will (approved):

aboedo:
LGTM

PR #4538, opened 2024-12-04
[Paywalls] Add previews for different combinations of vertical/horizontal alignment and flex distributions

MarkVillacampa:
### Checklist

### DIFF
--- a/RevenueCatUI/Views/IconView.swift
+++ b/RevenueCatUI/Views/IconView.swift
@@ -110,8 +110,8 @@ struct IconView_Previews: PreviewProvider {
 
     static var previews: some View {
         LazyVGrid(columns: [GridItem(.adaptive(minimum: 50))]) {
-            ForEach(PaywallIcon.allCases, id: \.rawValue) { icon in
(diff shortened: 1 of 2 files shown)


RevenueCatUI/Views/IconView.swift
```diff
@@ -110,8 +110,8 @@ struct IconView_Previews: PreviewProvider {
 
     static var previews: some View {
         LazyVGrid(columns: [GridItem(.adaptive(minimum: 50))]) {
-            ForEach(PaywallIcon.allCases, id: \.rawValue) { icon in
-                Self.icon(icon, Self.colors.randomElement()!)
+            ForEach(Array(PaywallIcon.allCases.enumerated()), id: \.element.rawValue) { index, icon in
+                Self.icon(icon, Self.colors[index % Self.colors.count])
```
MarkVillacampa, line 114: If the colours are random, snapshot tests would fail on every run. We can avoid that by making the colors consistent across runs.

joshdholtz (approved):
:shipit: :shipit: :shipit:

PR #4554, opened 2024-12-05
[Customer Center] Slight improvement to the Customer Center Promotional Offer view

aboedo:
The promo offer view felt pretty barren. This is a very small update, but I feel like it looks significantly better this way. 

All I'm doing is reading the icon from the bundle if available. If the image cannot be loaded, it looks exactly the same as it used to. 

| Before | After |
| :-: | :-: |
|  Screenshot 2024-12-05 at 6 18 10 PM | Screenshot 2024-12-05 at 6 17 30 PM |
| Screenshot 2024-12-05 at 6 22 21 PM | image |

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
@@ -111,7 +111,9 @@ struct PromotionalOfferView: View {
                     .clipShape(RoundedRectangle(cornerRadius: 20, style: .continuous))
                     .shadow(radius: 10)
             } else {
-                EmptyView()
+                Color.clear
+                    // keep the same size as the image would have taken to ensure layout still looks good
+                    .frame(width: 100, height: 100)
             }
         }
     }
--- a/RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
@@ -102,6 +102,7 @@ struct PromotionalOfferView: View {
     }
 
     private struct AppIconView: View {
+
         var body: some View {
             if let appIcon = AppIconHelper.getAppIcon() {
                 Image(uiImage: appIcon)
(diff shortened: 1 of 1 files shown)


aboedo:
@vegaro genuinely not sure what the labels should be - it's customer center but not really a feature, went with "other"

RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
```diff
@@ -110,7 +143,6 @@ struct PromotionalOfferHeaderView: View {
     private(set) var viewModel: PromotionalOfferViewModel
 
     private let spacing: CGFloat = 30
-    private let topPadding: CGFloat = 150
```
aboedo, line 113: moved the padding from here to the icon view so it can be controlled from there

RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
```diff
@@ -94,6 +100,33 @@ struct PromotionalOfferView: View {
...
+
+    private enum AppIconHelper {
+        static func getAppIcon() -> UIImage? {
+            guard let iconsDictionary = Bundle.main.infoDictionary?["CFBundleIcons"] as? [String: Any],
+                  let primaryIcons = iconsDictionary["CFBundlePrimaryIcon"] as? [String: Any],
+                  let iconFiles = primaryIcons["CFBundleIconFiles"] as? [String],
+                  let lastIcon = iconFiles.last else {
+                return nil
+            }
+            return UIImage(named: lastIcon)
```
aboedo, line 127: just fetch the app icon, seems to work well in my testing

aboedo: 🤔 looks a bit blurry actually, maybe I can do something to ensure we're always getting the highest-res version

RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
```diff
@@ -62,6 +62,12 @@ struct PromotionalOfferView: View {
 
             VStack {
                 if self.viewModel.error == nil {
+
+                    AppIconView()
+                        .padding(.top, 100)
```
fire-at-will, line 67: Since this padding is always applied, the padding will still be present even if the AppIconView() is just an EmptyView(), which could make the layout look off if the app icon can't be found for some reason. I'd consider making a static property on AppIconView called `canDisplay` and then conditionally setting the top padding based on that value, like:

```swift
.padding(.top, AppIconView.canDisplay ? 100 : 0)
```

fire-at-will (approved):

RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
fire-at-will: Oh, I think you switched the missing case from being an EmptyView() to a Color.clear while I was reviewing. Ignore this comment since we want to preserve the space!

aboedo:
🤔 actually I think we might not wanna merge this and instead have the image in the backend unfortunately.

The image is blurred, since the only icon that is actually embedded in the app is the size that the app would need. So I just get a 60x60 icon, doesn't look great

RevenueCatUI/CustomerCenter/Views/PromotionalOfferView.swift
aboedo: yeah sorry for the switcheroo, I had uncommitted code. I guess I'm starting to forget how to do PRs properly

aboedo:
it's similar to what StoreKit Views do by default

aboedo:
ok, final version.
I updated the size to 70x70, which seems to roughly match StoreKit's SubscriptionStoreView sizing for their equivalent. The icon is technically 60x60, but at that size it doesn't look blurry or stretched to my eye

Here's what it looks like, when the icon can be picked up and when it can't

|  |  |
| :-: | :-: |
| Screenshot 2024-12-06 at 5 53 22 PM | Screenshot 2024-12-06 at 5 53 12 PM |
| Screenshot 2024-12-06 at 5 55 57 PM | Screenshot 2024-12-06 at 5 56 08 PM |

PR #4555, opened 2024-12-05
Support non-JSON object decodable values in `getMetadataValue`

codykerns:
### Checklist

### Motivation
When fetching an offering metadata value with `getMetadataValue(for:)` (without a default value), a primitive Decodable (like `String`) causes the JSON decoder to crash, as it assumes the value is a JSON object, but it isn't a valid top-level JSON type.

### Description
- Use JSONSerializer's `isValidJSONObject` method before attempting to serialize values
- If it's not a valid JSON object, attempt to return the metadata value as the provided decodable type

### DIFF
--- a/Sources/Purchasing/Offering.swift
+++ b/Sources/Purchasing/Offering.swift
@@ -310,8 +310,7 @@ extension Offering {
     }
 
     /// - Returns: The `metadata` value associated to `key` for the expected `Decodable` type,
-    /// or `nil` if not found.
-    /// - Throws: Error if the content couldn't be deserialized to the expected type.
+    /// or `nil` if not found or if the content couldn't be deserialized to the expected type..
     /// - Note: This decodes JSON using `JSONDecoder.KeyDecodingStrategy.convertFromSnakeCase`.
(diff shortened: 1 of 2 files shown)


Sources/Purchasing/Offering.swift
```diff
@@ -314,16 +314,22 @@ extension Offering {
...
+            do {
+                let data = try JSONSerialization.data(withJSONObject: value)
+                return try JSONDecoder.default.decode(
+                                T.self,
+                                jsonData: data,
+                                logErrors: true
+                            )
+            } catch {
+                return nil
+            }
```
aboedo, line 329: I think we can simplify this a bit with `try?` instead of `try`, right?

aboedo (commented):
looks great so far! We should add a test to ensure we don't break it again in the future and then ready to merge

Sources/Purchasing/Offering.swift
codykerns: totally, I tried to retain as much existing as possible but happy to simplify

aboedo (approved):
Thanks for taking care of this!!

PR #4567, opened 2024-12-10
Support not showing Contact Customer Support if the user isn't on the most recent app version

fire-at-will:
### Description
This PR updates the Customer Center to support the "Require users to update to the latest version of your app before contacting support" checkbox in the dashboard. The checkbox now controls:
- Whether users must be on the most recent app version to contact support
- Whether the `AppUpdateWarningView` is shown when the Customer Center is shown and the user is on an old app version

### Testing
- Tested manually
- Added unit tests for the new `CustomerCenterViewModel.appUpdateRequiredToContactSupport`

### DIFF
--- a/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
@@ -44,7 +44,7 @@ import RevenueCat
 
     /// Whether or not the user needs to update their app version to contact support.
     var appUpdateRequiredToContactSupport: Bool {
-        // For now, we're using the same flag as the app update warning.
+        // We're intentionally using the same flag as the app update warning.
         return self.shouldShowAppUpdateWarning
     }
 
--- a/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
@@ -38,9 +38,14 @@ import RevenueCat
     private(set) var appIsLatestVersion: Bool = defaultAppIsLatestVersion
     private(set) var purchasesProvider: CustomerCenterPurchasesType
 
+    var shouldShowAppUpdateWarning: Bool {
+        return !appIsLatestVersion && (configuration?.support.shouldWarnCustomerToUpdate ?? false)
+    }
+
     /// Whether or not the user needs to update their app version to contact support.
     var appUpdateRequiredToContactSupport: Bool {
-        return !appIsLatestVersion && (configuration?.support.shouldWarnCustomerToUpdate ?? false)
+        // For now, we're using the same flag as the app update warning.
+        return self.shouldShowAppUpdateWarning
     }
 
     // @PublicForExternalTesting
--- a/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
@@ -38,6 +38,11 @@ import RevenueCat
     private(set) var appIsLatestVersion: Bool = defaultAppIsLatestVersion
     private(set) var purchasesProvider: CustomerCenterPurchasesType
 
+    /// Whether or not the user needs to update their app version to contact support.
+    var appUpdateRequiredToContactSupport: Bool {
+        return !appIsLatestVersion && (configuration?.support.shouldWarnCustomerToUpdate ?? false)
+    }
+
     // @PublicForExternalTesting
     @Published
     var state: CustomerCenterViewState {
--- a/RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
+++ b/RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
@@ -83,7 +83,7 @@ struct RestorePurchasesAlert: ViewModifier {
 
                 case .purchasesNotFound:
                     let message = Text(localization.commonLocalizedString(for: .purchasesNotRecovered))
-                    if let url = supportURL {
+                    if let url = supportURL, !customerCenterViewModel.appUpdateRequiredToContactSupport {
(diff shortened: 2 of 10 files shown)


Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -550,6 +555,7 @@ extension CustomerCenterConfigData.Support {
 
     init(from response: CustomerCenterConfigResponse.Support) {
         self.email = response.email
+        self.shouldWarnCustomerToUpdate = response.shouldWarnCustomerToUpdate ?? false
```
fire-at-will, line 558: Defaults to false if it isn't included in the response from the backend since the checkbox's default value in the dashboard is false.

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -38,6 +38,16 @@ import RevenueCat
     private(set) var appIsLatestVersion: Bool = defaultAppIsLatestVersion
     private(set) var purchasesProvider: CustomerCenterPurchasesType
 
+    var shouldShowAppUpdateWarning: Bool {
+        return !appIsLatestVersion && (configuration?.support.shouldWarnCustomerToUpdate ?? false)
+    }
+
+    /// Whether or not the user needs to update their app version to contact support.
+    var appUpdateRequiredToContactSupport: Bool {
+        // For now, we're using the same flag as the app update warning.
```
aboedo, line 47: I'd remove the "for now", since it assumes some kind of temporal nature to it, and if we never change it later on it will look to a maintainer like this was temporary or hacky code

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -218,6 +219,55 @@ class CustomerCenterViewModelTests: TestCase {
...
+        )
+        viewModel.configuration = CustomerCenterConfigTestData.customerCenterData(
+            lastPublishedAppVersion: latestVersion,
+            shouldWarnCustomerToUpdate: true
+        )
+
+        expect(viewModel.appUpdateRequiredToContactSupport).to(beFalse())
+    }
+
+    func testAppUpdateRequiredToContactSupport_false_blocked_by_config() {
```
aboedo, line 255: the mix of casing here looks a little odd to my eye

aboedo: I would maybe go with something more like 
`testAppUpdateRequiredToContactSupportReturnsFalseIfBlockedByConfig `

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -38,6 +38,16 @@ import RevenueCat
     private(set) var appIsLatestVersion: Bool = defaultAppIsLatestVersion
     private(set) var purchasesProvider: CustomerCenterPurchasesType
 
+    var shouldShowAppUpdateWarning: Bool {
+        return !appIsLatestVersion && (configuration?.support.shouldWarnCustomerToUpdate ?? false)
+    }
+
+    /// Whether or not the user needs to update their app version to contact support.
+    var appUpdateRequiredToContactSupport: Bool {
```
aboedo, line 46: we don't entirely _require_ but rather _suggest_, right?

aboedo (commented):
looks great so far, left a question

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
fire-at-will: With this implementation, I built it the way the copy for the checkbox on the dashboard is worded and made it _required_ to contact support in the WrongPlatformView & the Restore Purchases flow.

The checkbox's copy reads:
> Require users to update to the latest version of your app before contacting support

That said, I don't think it's great to prevent people from contacting support and would suggest that we instead show a message or button in the alert urging the user to update in addition to letting them contact support.

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
fire-at-will: Ah, sometimes I mix them when the test names get super long in my apps - I'll change them to use the normal style here :)

Sources/Networking/Responses/CustomerCenterConfigResponse.swift
```diff
@@ -132,6 +132,7 @@ struct CustomerCenterConfigResponse {
     struct Support {
 
         let email: String
+        let shouldWarnCustomerToUpdate: Bool?
```
tonidero, line 135: Hmm I'm thinking... the original idea was that, if the backend sent the `lastPublishedAppVersion`, the SDK would consider it "enabled" implicitly. The backend shouldn't send that version if it's disabled. So this property would be unnecessary I think?

RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
```diff
@@ -83,7 +83,7 @@ struct RestorePurchasesAlert: ViewModifier {
 
                 case .purchasesNotFound:
                     let message = Text(localization.commonLocalizedString(for: .purchasesNotRecovered))
-                    if let url = supportURL {
+                    if let url = supportURL, !customerCenterViewModel.appUpdateRequiredToContactSupport {
```
tonidero, line 86: Hmm do we need to show the `AppUpdateWarningView` in the else case? As in, right now we won't be able to contact support if the app update is required right? Which I think we didn't want to do? Same for the `WrongPlatformView`.

tonidero (commented):
A couple comments... I believe we shouldn't force people to update in order to contact support, which I think is what we're doing now

fire-at-will:
Closing in favor of https://github.com/RevenueCat/purchases-ios/pull/4571

PR #4569, opened 2024-12-11
Fix hardcoded title in WrongPlatformView

vegaro:
We were hardcoding the title in `WrongPlatformView`

simulator_screenshot_11835AB6-6706-43D4-ADE0-51297B7C0247

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
@@ -125,6 +125,8 @@ private extension CustomerCenterView {
                     ManageSubscriptionsView(screen: screen,
                                             customerCenterActionHandler: viewModel.customerCenterActionHandler)
                 }
+            } else if let screen = configuration.screens[.management] {
+                WrongPlatformView(screen: screen)
             } else {
                 WrongPlatformView()
             }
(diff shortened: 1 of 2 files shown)


RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -125,6 +125,8 @@ private extension CustomerCenterView {
                     ManageSubscriptionsView(screen: screen,
                                             customerCenterActionHandler: viewModel.customerCenterActionHandler)
                 }
+            } else if let screen = configuration.screens[.management] {
+                WrongPlatformView(screen: screen)
```
tonidero, line 129: Hmm are we passing the management screen to the `WrongPlatformView`? Seems wrong to me... As in, it should be a separate screen right?

tonidero (commented):
LGTM other than a question

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
vegaro: We are not adding another screen to the frontend yet, so reusing that title which is for the "Users with an active subscription", which also applies to this screen.

We should build this screen dynamically as well, and I am thinking we should actually reuse the Management screen for it (filtering out things that don't make sense, like asking for a refund), what do you think? I would do that in another PR though

tonidero (approved):

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
vegaro: Yes it's mainly the custom URLs and the soon to come contact support one which is hardcoded in this screen atm. I don't know what to do with the manage subscriptions button, which is not a path either but doesn't make sense in the regular management screen

PR #4574, opened 2024-12-11
Add ErrorView to CustomerCenter

vegaro:
Added a simple error view when the Customer Center fails

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/ErrorView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/ErrorView.swift
@@ -21,6 +21,9 @@ import SwiftUI
 @available(watchOS, unavailable)
 struct ErrorView: View {
 
+    @Environment(\.locale)
+    private var locale
+
     var body: some View {
         VStack(spacing: 20) {
             let errorMessage: String = Localization.localizedBundle(self.locale)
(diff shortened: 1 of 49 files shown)


RevenueCatUI/CustomerCenter/Views/ErrorView.swift
```diff
@@ -0,0 +1,60 @@
...
+    @Environment(\.locale)
+    private var locale
+
+    var body: some View {
+        VStack(spacing: 20) {
+            let errorMessage: String = Localization.localizedBundle(self.locale)
+                .localizedString(forKey: "Something went wrong",
+                                 value: "Something went wrong",
+                                 table: nil)
+            CompatibilityContentUnavailableView(
```
tonidero, line 33: Hmm wondering if it would make sense to have a "retry" button... It might also be worth at least displaying network errors with a different message... But we can improve that in future PRs.

tonidero (approved):
LGTM! Just a comment but can be addressed in future PRs.

RevenueCatUI/CustomerCenter/Views/ErrorView.swift
vegaro: I wanted to do that, but it involves checking the type of network error, and thought it would complicate things a bit. Let's leave it for a future PR. I will add it to the list of things to do

PR #4575, opened 2024-12-11
Always reload customerInfo when Customer Center is loaded

fire-at-will:
### Description
This PR makes it so that we always reload the CustomerInfo from the backend when the Customer Center appears


RevenueCatUI/CustomerCenter/Abstractions/CustomerCenterPurchasesType.swift
```diff
@@ -23,7 +23,9 @@ protocol CustomerCenterPurchasesType: Sendable {
     var isSandbox: Bool { get }
 
     @Sendable
-    func customerInfo() async throws -> CustomerInfo
+    func customerInfo(
+        fetchPolicy: CacheFetchPolicy
```
fire-at-will, line 27: I wanted to make this default to `.default`, but Swift doesn't allow default values in protocols 😅

vegaro (approved):

PR #4587, opened 2024-12-13
Dont show refund cancelled message when user cancels refund

fire-at-will:
### Motivation
Currently, when the user cancels a refund, we show the following message on the SubscriptionDetailsView screen under the "Refund Status" message:

This message is a little confusing because it implies that a refund was created and was then cancelled, when one was never created in the first place. When I saw the message, my first thought was, "do I have the money, or does the developer?".

### Description

This PR modifies the behavior so that the "Refund Status" section doesn't display at all when the user dismisses the refund screen without creating a refund, since the user already knows that they cancelled it.

The PR doesn't remove the `refundCanceled` localized string in case we need it in the future here or on another platform.


vegaro (approved):
:ship-it:

PR #4592, opened 2024-12-16
[WEB-1757] Handle new backend error codes that may show in the redemption endpoint

tonidero:
### Description
Currently these error codes would display in the API as "UnknownBackendError". This performs some better processing of these error codes.

### DIFF
--- a/Sources/Error Handling/BackendErrorCode.swift
+++ b/Sources/Error Handling/BackendErrorCode.swift
@@ -39,9 +39,13 @@ enum BackendErrorCode: Int, Error {
     case playStoreGenericError = 7231
     case userIneligibleForPromoOffer = 7232
     case invalidAppleSubscriptionKey = 7234
+    case couldNotCreateAlias = 7255
+    case invalidAppUserId = 7256
     case subscriptionNotFoundForCustomer = 7259
     case invalidSubscriberAttributes = 7263
     case invalidSubscriberAttributesBody = 7264
+    case requestAlreadyInProgress = 7638
+    case subscriberAttributesAreBeingUpdated = 7629
     case purchasedProductMissingInAppleReceipt = 7712
     case invalidWebRedemptionToken = 7849
     case purchaseBelongsToOtherUser = 7852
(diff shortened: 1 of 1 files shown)


Sources/Error Handling/BackendErrorCode.swift
```diff
@@ -112,6 +117,11 @@ extension BackendErrorCode {
         case .invalidSubscriberAttributes,
              .invalidSubscriberAttributesBody:
             return .invalidSubscriberAttributesError
+        case .couldNotCreateAlias:
+            return .configurationError
+        case .requestAlreadyInProgress,
+             .subscriberAttributesAreBeingUpdated:
+            return .operationAlreadyInProgressForProductError
```
tonidero, line 124: Note that this error matches android's `OperationAlreadyInProgressError` without the `ForProduct` part... I think I prefer Android's naming but I didn't want to cause a breaking change due to this so I just reused the current naming.

Sources/Error Handling/BackendErrorCode.swift
```diff
@@ -112,6 +117,11 @@ extension BackendErrorCode {
         case .invalidSubscriberAttributes,
              .invalidSubscriberAttributesBody:
             return .invalidSubscriberAttributesError
+        case .couldNotCreateAlias:
+            return .configurationError
```
vegaro, line 121: looks weird that this is a configuration error?

vegaro (approved):

Sources/Error Handling/BackendErrorCode.swift
tonidero: A bit... The cases where this may happen are:
```
- 'Error creating subscriber alias. Alias limit reached.'
- 'Error creating subscriber alias. Subscriber belongs to another app.'
- 'Error creating subscriber alias. Subscriber already aliased.'
- 'Error creating subscriber alias. Subscriber cannot be aliased to another already aliased subscriber.'
```
I wasn't sure what error code to match it to... considering that these happen if you alias when you shouldn't, I thought `ConfigurationError` was the one that better matched, but open to suggestions... We can create a new error code if you think nothing matches.

tonidero:
Will merge this for now but we can change the relative error code if needed. It should be a pretty uncommon error code in any case.

Sources/Error Handling/BackendErrorCode.swift
vegaro: Yeah I guess that's not too bad. I can't find any other one that would suit better, and I'd rather not create more error codes.

PR #4594, opened 2024-12-16
[Paywalls] Fix issues with rounded corners and borders

MarkVillacampa:
- When using the `Capsule` shape for Pill style shapes, use the `Circular` `RoundedCornerStyle`.
- If available (iOS 16>) use `UnevenRoundedRectangle` for rounded corners and borders.
- Use `.strokeBorder` instead of `.stroke` where available, which first insets the shape by `lineWidth/2` so the whole border stays within the frame of the shape.
- Improve `BorderRoundedCornerShape` to make it conform to `InsettableShape`, so we can use use `.strokeBorder` on it.
- Use `addArc` instead of `addQuadCurve` in `BorderRoundedCornerShape` to match the circular corners in `SingleRoundedCornerShape`.

These changes help fix scenarios where the border would overflow outside of the Stack's frame:

|Before | After
|-------|--------
|Screenshot 2024-12-16 at 14 21 09 | Screenshot 2024-12-16 at 14 20 36

And where the roundness of the Stack's corners and of the border would differ, leaving some of the background color exposed:

Screenshot 2024-12-16 at 14 30 20

### DIFF
--- a/RevenueCatUI/Templates/V2/ViewHelpers/Shape.swift
+++ b/RevenueCatUI/Templates/V2/ViewHelpers/Shape.swift
@@ -216,11 +216,8 @@ private struct BorderRoundedCornerShape: InsettableShape {
     func path(in rect: CGRect) -> Path {
         var path = Path()
 
-        // Inset the bottom border to avoid overlapping with a sibling component
-        let newRect = CGRect(x: rect.minX, y: rect.minY, width: rect.width, height: rect.maxY - 1)
-
         // Adjust rect for insets
-        let insetRect = newRect.insetBy(dx: insetAmount, dy: insetAmount)
+        let insetRect = rect.insetBy(dx: insetAmount, dy: insetAmount)
 
         // Adjust corner radii for insets
         let adjustedTopLeft = max(0, topLeft - insetAmount)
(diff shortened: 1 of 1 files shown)


RevenueCatUI/Templates/V2/ViewHelpers/Shape.swift
```diff
@@ -115,19 +115,28 @@ extension View {
...
+                        $0.clipShape(SingleRoundedCornerShape(radius: topLeft, corners: [.topLeft]))
+                    }
+                    .applyIf(topRight > 0) {
+                        $0.clipShape(SingleRoundedCornerShape(radius: topRight, corners: [.topRight]))
+                    }
+                    .applyIf(bottomLeft > 0) {
+                        $0.clipShape(SingleRoundedCornerShape(radius: bottomLeft, corners: [.bottomLeft]))
+                    }
+                    .applyIf(bottomRight > 0) {
+                        $0.clipShape(SingleRoundedCornerShape(radius: bottomRight, corners: [.bottomRight]))
```
MarkVillacampa, line 137: There was a bug here where the radius was always set to `topLeft`, fixed now.

joshdholtz: Oh this was so on me 🫠

joshdholtz (approved):
Looks good to me! :shipit:

PR #4596, opened 2024-12-16
[Paywalls] Add Badge Modifier

MarkVillacampa:
This PR adds a `.stackBadge` modified to `Stacks` which is displayed in one of three ways: nested, edge_to_edge, or overlaid.

Check emerge for snapshots for all examples e.g.:

image

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Stack/StackComponentViewModel.swift
+++ b/RevenueCatUI/Templates/V2/Components/Stack/StackComponentViewModel.swift
@@ -24,8 +24,8 @@ class StackComponentViewModel {
     private let component: PaywallComponent.StackComponent
     let uiConfigProvider: UIConfigProvider
     private let presentedOverrides: PresentedOverrides<PresentedStackPartial>?
-    let badgeTextViewModel: TextComponentViewModel?
 
+    let badgeTextViewModel: TextComponentViewModel?
     let viewModels: [PaywallComponentViewModel]
 
     init(
@@ -156,7 +156,7 @@ struct StackComponentStyle {
         self.shape = shape?.shape
         self.border = border?.border(uiConfigProvider: uiConfigProvider)
         self.shadow = shadow?.shadow(uiConfigProvider: uiConfigProvider)
-        self.badge = badge?.badge(parentShape: self.shape)
+        self.badge = badge?.badge(stackShape: self.shape)
     }
 
     var vstackStrategy: StackStrategy {
@@ -240,7 +240,7 @@ private extension PaywallComponent.Shadow {
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 private extension PaywallComponent.Badge {
 
-    func badge(parentShape: ShapeModifier.Shape?) -> BadgeModifier.BadgeInfo? {
+    func badge(stackShape: ShapeModifier.Shape?) -> BadgeModifier.BadgeInfo? {
         BadgeModifier.BadgeInfo(
             style: self.style,
             alignment: self.alignment,
@@ -254,7 +254,7 @@ private extension PaywallComponent.Badge {
             horizontalAlignment: self.horizontalAlignment,
             color: self.color,
             backgroundColor: self.backgroundColor,
-            parentShape: parentShape
+            stackShape: stackShape
         )
     }
 
--- a/RevenueCatUI/Templates/V2/Components/Stack/StackComponentViewModel.swift
+++ b/RevenueCatUI/Templates/V2/Components/Stack/StackComponentViewModel.swift
@@ -24,17 +24,37 @@ class StackComponentViewModel {
     private let component: PaywallComponent.StackComponent
     let uiConfigProvider: UIConfigProvider
     private let presentedOverrides: PresentedOverrides<PresentedStackPartial>?
+    let badgeTextViewModel: TextComponentViewModel?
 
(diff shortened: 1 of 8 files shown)


RevenueCatUI/Templates/V2/Components/Stack/StackComponentViewModel.swift
```diff
@@ -23,16 +23,36 @@ class StackComponentViewModel {
...
         component: PaywallComponent.StackComponent,
-        viewModels: [PaywallComponentViewModel]
+        viewModels: [PaywallComponentViewModel],
+        localizationProvider: LocalizationProvider
     ) throws {
         self.component = component
         self.viewModels = viewModels
 
+        if let badge = component.badge {
+            badgeTextViewModel = try TextComponentViewModel(
```
MarkVillacampa, line 39: Initialising the `TextComponentViewModel` of the badge in the `StackComponentViewModel` constructor so we can throw if necessary.

RevenueCatUI/Templates/V2/ViewHelpers/Shape.swift
```diff
@@ -48,13 +48,6 @@ struct ShapeModifier: ViewModifier {
         let bottomLeft: CGFloat?
         let bottomRight: CGFloat?
 
-        init(topLeft: Double? = nil, topRight: Double? = nil, bottomLeft: Double? = nil, bottomRight: Double? = nil) {
-            self.topLeft = topLeft.flatMap { CGFloat($0) }
-            self.topRight = topRight.flatMap { CGFloat($0) }
-            self.bottomLeft = bottomLeft.flatMap { CGFloat($0) }
-            self.bottomRight = bottomRight.flatMap { CGFloat($0) }
-        }
-
```
MarkVillacampa, line 57: Apparently this initialiser was not needed because we're not using explicit Doubles anywhere

RevenueCatUI/Templates/V2/Components/Stack/StackComponentViewModel.swift
```diff
@@ -128,6 +151,7 @@ struct StackComponentStyle {
         self.shape = shape?.shape
         self.border = border?.border
         self.shadow = shadow?.shadow
+        self.badge = badge?.badge(stackShape: self.shape)
```
MarkVillacampa, line 154: The stack shape is only used to calculate the the corner radius of the edge-to-edge leading/trailing style badge, and make it match the parent stack.

RevenueCatUI/Templates/V2/ViewModelHelpers/ViewModelFactory.swift
```diff
@@ -90,10 +90,22 @@ struct ViewModelFactory {
                 )
             }
 
+            let badgeViewModels = try component.badge?.stack.value.components.map { component in
```
MarkVillacampa, line 93: Deserialising the badge view models here just like we do with the stack children models right above.

RevenueCatUI/Templates/V2/ViewHelpers/BadgeModifier.swift
```diff
@@ -0,0 +1,468 @@
...
+    @ViewBuilder
+    func apply(badge: BadgeModifier.BadgeInfo) -> some View {
+        switch badge.style {
+        case .edgeToEdge:
+            self.applyBadgeEdgeToEdge(badge: badge)
+        case .overlaid:
+            self.overlay(
+                VStack(alignment: .leading) {
+                    VStack {
+                        ComponentsView(componentViewModels: badge.badgeViewModels, onDismiss: {})
```
MarkVillacampa, line 56: I'm not entirely sure we need to pass the `onDismiss` block here? I assume it is only used for the NavigateBack action button and would only make sense if we want to include buttons in the children elements here?

joshdholtz: Yeah yeah, this if fine 😅

joshdholtz (approved):
:shipit: :shipit: :shipit: :shipit: :shipit: :shipit:

RevenueCatUI/Templates/V2/ViewHelpers/BadgeModifier.swift
```diff
@@ -0,0 +1,465 @@
...
+
+    // Helper to calculate the shape of the edge-to-edge badge in trailing/leading positions.
+    // swiftlint:disable:next cyclomatic_complexity function_body_length
+    private func effectiveShape(badge: BadgeModifier.BadgeInfo) -> ShapeModifier.Shape? {
+        switch badge.style {
+        case .edgeToEdge:
+            switch badge.stack.value.shape {
+            case .pill, .none:
+                // Edge-to-edge badge cannot have pill shape
+                return nil
```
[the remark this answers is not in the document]
JayShortway: The edge-to-edge badge itself cannot have a pill shape, but can a pill-shaped stack have an edge-to-edge badge? We can already support that in Android, and I think we should. 🤔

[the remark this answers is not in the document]
MarkVillacampa: The thing with Pill stack shape with a trailing/leading badge is we cannot easily programmatically calculate the corner radius of the stack, so we cannot make the badge  corner radius match.

There are probably ways to support it: use a mask of the parent shape to cur off the corner without actually knowing the radius, or calculate it based on half of the height/width of the stack, but not sure the increased complexity is worth it.

PR #4598, opened 2024-12-16
Bump timeout for deferred purchase tests for CI

fire-at-will:
While they run okay locally, the updated tests for https://github.com/RevenueCat/purchases-ios/pull/4584 have been flaky in CI in iOS <17. This PR increases the timeout interval to make them less flaky.


fire-at-will:
@RCGitBot please test

PR #4605, opened 2024-12-17
Use #file instead of #fileID to avoid including the full path in the compiled binary

MarkVillacampa:
These were the only non-test files where we were using #file instead of #fileID

### Checklist


JayShortway (approved):
🙌

PR #4607, opened 2024-12-17
Fixing text, image, and footer render issues

joshdholtz:
### Motivation

- Text was not rendering markdown
- Images where growing the width of its parent
- Footer was expanding whole height

### Description

- Text
  - Use the `.init` initializers
- Image
  - Use the `.fitToAspect` modifier and get the aspect ratio from the width and heigh property of an image
- Footer
  - Render with `.fixedSize` so it doesn't expand entire vertical view
  - But use `.overlay` because SwiftUI was locking up with `.safeAreaInset`

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
@@ -119,8 +119,8 @@ struct ImageComponentView_Previews: PreviewProvider {
                     component: .init(
                         source: .init(
                             light: .init(
-                                width: 1,
-                                height: 1,
+                                width: 750,
+                                height: 530,
                                 original: catUrl,
                                 heic: catUrl,
                                 heicLowRes: catUrl
@@ -147,8 +147,8 @@ struct ImageComponentView_Previews: PreviewProvider {
                     component: .init(
                         source: .init(
                             light: .init(
-                                width: 1,
-                                height: 1,
+                                width: 750,
+                                height: 530,
                                 original: catUrl,
                                 heic: catUrl,
                                 heicLowRes: catUrl
@@ -175,8 +175,8 @@ struct ImageComponentView_Previews: PreviewProvider {
                     component: .init(
                         source: .init(
                             light: .init(
-                                width: 1,
-                                height: 1,
+                                width: 750,
+                                height: 530,
                                 original: catUrl,
                                 heic: catUrl,
                                 heicLowRes: catUrl
@@ -206,8 +206,8 @@ struct ImageComponentView_Previews: PreviewProvider {
                     component: .init(
                         source: .init(
                             light: .init(
-                                width: 1,
-                                height: 1,
+                                width: 750,
+                                height: 530,
                                 original: catUrl,
                                 heic: catUrl,
                                 heicLowRes: catUrl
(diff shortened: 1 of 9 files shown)


RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
```diff
@@ -48,24 +51,48 @@ struct ImageComponentView: View {
                 darkUrl: style.darkUrl,
                 darkLowResUrl: style.darkLowResUrl
             ) { (image, size) in
-                renderImage(image, size, with: style)
+                self.renderImage(image, size, with: style)
             }
             .size(style.size)
```
tonidero, line 56: One thing we did in Android was that, if the size is `fit` in any axis, we change it to`Fixed` with the given backend sizes. Not sure if we want to do the same thing here?

RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
```diff
@@ -48,24 +51,48 @@ struct ImageComponentView: View {
...
         image
-            .resizable()
-            .aspectRatio(contentMode: style.contentMode)
+            .fitToAspect(
+                self.aspectRatio(style: style),
+                contentMode: style.contentMode,
+                containerContentMode: style.contentMode
+            )
+            .frame(maxWidth: .infinity)
+            .accessibilityHidden(true)
```
tonidero, line 85: Not needed for this PR! But we probably will need to change this with the accessibility string as well right? Should be fine to leave a comment I think.

RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
```diff
@@ -48,24 +51,48 @@ struct ImageComponentView: View {
...
+        @unknown default:
+            return (style.widthLight, style.heightLight)
+        }
+    }
+
     private func renderImage(_ image: Image, _ size: CGSize, with style: ImageComponentStyle) -> some View {
         image
-            .resizable()
-            .aspectRatio(contentMode: style.contentMode)
+            .fitToAspect(
```
tonidero, line 79: We might want to apply a similar thing to Android...

tonidero (approved):
LGTM!

RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
joshdholtz: OHHHH...  yeah yeah, good point 👍

RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
joshdholtz: @tonidero What if the parent/container is smaller than what the image is? Wouldn't the fixed make the image too big then? 🤔

JayShortway: That's a good thing to test, but I don't think so. We didn't use `requiredSize()` (which would ignore constraints set by the parent), so if the parent is smaller, the image will be sized down. (At least that's what I expect.)

tonidero: Yup, can confirm it will be constrained to the available space, but will add some previews so we have snapshot tests for that as well 👍

PR #1826, opened 2022-08-13
`Offering`/`StoreProductType`: `Sendable` conformance

NachoSoto:
The type is now immutable, so the compiler ensures it's thread-safe.

For [CSDK-379].
Extracted from #1795 to make it easier to review this in isolation.

[CSDK-379]: https://revenuecats.atlassian.net/browse/CSDK-379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #4612, opened 2024-12-19
Fix `@MainActor` reference in unsupported OS versions

aboedo:
https://github.com/RevenueCat/purchases-ios/pull/1826 introduced references to `@MainActor` that are utilized in OS versions that don't support Swift Concurrency - namely, iOS 11-12 and macOS 10.12-10.14. 

This PR addresses it by using a compilation flag.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -33,11 +33,18 @@ public typealias PurchaseResultData = (transaction: StoreTransaction?,
 /**
  Completion block for ``Purchases/purchase(product:completion:)``
  */
+#if swift(>=5.5) && canImport(_Concurrency)
 public typealias PurchaseCompletedBlock = @MainActor @Sendable (StoreTransaction?,
                                                                 CustomerInfo?,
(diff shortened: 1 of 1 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -33,11 +33,18 @@ public typealias PurchaseResultData = (transaction: StoreTransaction?,
 /**
  Completion block for ``Purchases/purchase(product:completion:)``
  */
+#if swift(>=5.5) && canImport(_Concurrency)
```
aboedo, line 36: I _think_ the swift concurrency import would work?

MarkVillacampa: I think since this is a compiler conditional, `canImport` only checks if the version of Xcode you're compiling with supports concurrency.

And since `@MainActor` is supposedly back ported to macOS 10/11, and the current version of Xcode supports concurrency, this means we would end up always compiling the first branch.

aboedo: verified that this approach wouldn't work, closing

PR #4599, opened 2024-12-16
Revert "Always call readyForPromotedProduct on the main actor"

fire-at-will:
### Description
Despite passing fine locally, these changes have been flaky in the CI for iOS 14-16. This PR reverts the changes in https://github.com/RevenueCat/purchases-ios/pull/4584 out of an abundance of caution

Issue #4582, opened 2024-12-12
Crash on launch in `Purchases.readyForPromotedProduct(_:purchase:)`


vrutberg:
- [ ] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

We've been seeing a crash that has been occurring since the launch of iOS 18. It hasn't generated any user feedback, and it doesn't seem to happen all that often, but it has been occurring since iOS 18 was released, and it's still occurring. According to Crashlytics, it seems to be occurring shortly after launch, but we haven't been able to reproduce this issue for ourselves.

Crashlytics reports 3 different variants of this crash, I'll post the most recent here:

```
          Crashed: com.apple.root.default-qos.cooperative
0  libdispatch.dylib              0x63b4 _dispatch_assert_queue_fail + 120
1  libdispatch.dylib              0x633c _dispatch_assert_queue_fail + 194
2  libswift_Concurrency.dylib     0x62b78 swift_task_isCurrentExecutorImpl(swift::SerialExecutorRef) + 284
3  Buddy                          0x4fcfa4 @objc PurchaseManager.purchases(_:readyForPromotedProduct:purchase:) + 4347793316 (<compiler-generated>:4347793316)
4  Buddy                          0x782148 @objc Purchases.readyForPromotedProduct(_:purchase:) + 1583 (Purchases.swift:1583)
5  Buddy                          0x7c28c8 specialized PurchasesOrchestrator.storeKit2PurchaseIntentListener(_:purchaseIntent:) + 1124 (PurchasesOrchestrator.swift:1124)
6  libswift_Concurrency.dylib     0x60f5c swift::runJobInEstablishedExecutorContext(swift::Job*) + 252
7  libswift_Concurrency.dylib     0x62514 swift_job_runImpl(swift::Job*, swift::SerialExecutorRef) + 144
8  libdispatch.dylib              0x15ec0 _dispatch_root_queue_drain + 392
9  libdispatch.dylib              0x166c4 _dispatch_worker_thread2 + 156
10 libsystem_pthread.dylib        0x3644 _pthread_wqthread + 228
[...]

PR #4613, opened 2024-12-19
Always call readyForPromotedProduct on main thread/actor

fire-at-will:
### Motivation
Potential fix for https://github.com/RevenueCat/purchases-ios/issues/4582. While I haven't been able to reproduce the crash, a user on the issue has reported that this fixes the crash.

### Description
https://github.com/RevenueCat/purchases-ios/issues/4582 describes an occasional crash that occurs when the SDK calls the PurchaseDelegate's readyForPromotedProduct function. The provided stacktraces show a threading issue occurring in both the SK1 & SK2 flows. Our working theory is that the calling code may expect `readyForPromotedProduct` to be called on the main thread/actor, and it is being called on a background thread.

This PR ensures that `readyForPromotedProduct` is always called on the Main Actor (for SK2) and main thread (for SK1).

A user on the GH Issue [has reported that this PR resolves the crash for them](https://github.com/RevenueCat/purchases-ios/issues/4582#issuecomment-2561603154).

### SK1 & SK2 Differences
https://github.com/RevenueCat/purchases-ios/pull/4584 attempted to address this issue by always calling `readyForPromotedProduct` on the main actor, but we found that this caused very consistent, but still flaky, test failures in SK1 on iOS 14-16. Out of an abundance of caution, this was reverted in https://github.com/RevenueCat/purchases-ios/pull/4599.

This PR calls `readyForPromotedProduct` on the main thread instead of the main actor for SK1, which has addressed these test failures.

### Testing
While I've been unable to reproduce the crash, I've tested the fix with both StoreKit 1 & 2 on iOS 17-18. I've been unable to test in iOS 15-16 because Xcode doesn't allow you to simulate Purchase Intents in simulators on iOS <17, and all of my physical devices are on iOS 18. That said, all automated tests pass on all iOS versions.


fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

.circleci/config.yml
```diff
@@ -1345,6 +1345,7 @@ workflows:
         - not: << pipeline.parameters.generate_revenuecatui_snapshots >>
     jobs:
       - lint
+      - run-test-ios-13
```
MarkVillacampa, line 1348: do we want to keep this?

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1597,7 +1597,21 @@ extension Purchases: PurchasesOrchestratorDelegate {
      */
     func readyForPromotedProduct(_ product: StoreProduct,
                                  purchase startPurchase: @escaping StartPurchaseBlock) {
-        self.delegate?.purchases?(self, readyForPromotedProduct: product, purchase: startPurchase)
+
+        switch self.systemInfo.storeKitVersion.effectiveVersion {
+        case .storeKit1:
+            // Calling the delegate method on the main actor causes test failures on iOS 14-16, so instead
+            // we dispatch to the main thread, which doesn't cause the failures.
```
MarkVillacampa, line 1604: yeah i think this is good practice anyway, specially if we're calling on the main actor in sk2

MarkVillacampa (approved):

.circleci/config.yml
fire-at-will: I had it there for testing - it's removed now 😄

PR #4620, opened 2024-12-27
[Paywalls V2] Fix analytics and dismiss

joshdholtz:
### Motivation

Paywalls V2 view was not tracking impression/close event, disabling while purchase, dismissing after purchase.

### Description

Brought in code that was being used in Paywalls V1 for this


JayShortway (approved):
📦 ➡️ 🚢

PR #4621, opened 2024-12-27
[Paywalls V2] Added fallback components

joshdholtz:
### Motivation

If the type is unknown, try to parse/render a fallback component

### Description

If the `type` is unknown for a component, look at `fallback`.
This should be a component a basic component that is known in initial release that all SDKs can render.

### DIFF
--- a/Tests/UnitTests/Paywalls/Components/ButtonComponentTests.swift
+++ b/Tests/UnitTests/Paywalls/Components/ButtonComponentTests.swift
@@ -11,7 +11,12 @@ class ButtonComponentCodableTests: TestCase {
         "type": "stack",
         "dimension": {
             "type": "vertical",
-            "alignment": "center"
+            "alignment": "center",
+            "distribution": "start"
+        },
+        "size": {
+            "width": { "type": "fill" },
+            "height": { "type": "fill" }
         },
         "padding": {
             "top": 0,
@@ -44,7 +49,11 @@ class ButtonComponentCodableTests: TestCase {
 
         let buttonComponent = PaywallComponent.ButtonComponent(
             action: .restorePurchases,
-            stack: .init(components: [])
+            stack: .init(
+                components: [],
+                dimension: .vertical(.center, .start),
+                size: .init(width: .fill, height: .fill)
+            )
         )
 
         XCTAssertEqual(decodedButton, buttonComponent)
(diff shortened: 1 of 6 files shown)


Tests/UnitTests/Paywalls/Components/PartialComponentTests.swift
joshdholtz, line 1: This test needed fixing 😇

Tests/UnitTests/Paywalls/Components/ButtonComponentTests.swift
```diff
@@ -11,7 +11,12 @@ class ButtonComponentCodableTests: TestCase {
         "type": "stack",
```
JayShortway, line 11: Are the changes in this file (below) related? If so, how? 😄

Tests/UnitTests/Paywalls/Components/FallbackComponentTests.swift
```diff
@@ -0,0 +1,117 @@
...
+        }
+        """
+        let jsonData = jsonString.data(using: .utf8)!
+
+        do {
+            _ = try JSONDecoder.default.decode(PaywallComponent.self, from: jsonData)
+            XCTFail("Should have failed to decode fallback property")
+        } catch DecodingError.dataCorrupted(let context) {
+            expect(context.debugDescription).to(
+                contain("Failed to decode fallback for unknown type \"super_new_type\"")
```
JayShortway, line 108: Shouldn't this say something like _"Failed to decode unknown type without fallback"_?

Tests/UnitTests/Paywalls/Components/ButtonComponentTests.swift
joshdholtz: @JayShortway Nope! Just tests that didn't run on CI because of the feature/compiler flag so didn't see they failed a while back 🫠

Tests/UnitTests/Paywalls/Components/FallbackComponentTests.swift
joshdholtz: @JayShortway Okay, I think i fixed this now in https://github.com/RevenueCat/purchases-ios/pull/4621/commits/1b418027a07cdf6b01dead8f1d7f3c91e06175a6 🤞 

I also adde a test for decoding, encoding, and redecoding 🙃

PR #4622, opened 2024-12-27
fixed navigation with new param

aboedo:
Fixes the navigation issues for Customer Center when presented through a NavigationStack. 

Solves the issue by adding an extra optional parameter that you use when showing `CustomerCenterView` in a `NavigationStack`: 

```swift
    CustomerCenterView(isInNavigationStack: true)
```

| Before | After |
| :-: | :-: |
| https://github.com/user-attachments/assets/68154d11-8a82-43e0-95ec-482c0c40446d | https://github.com/user-attachments/assets/034b1b88-a0ef-4e90-9d4c-cec4cddaa72b |

## The problem: 

When presenting the Customer Center in a navigation stack, if we add another one, we kinda break navigation. I reproduced this in a very, very simple SwiftUI app: 
image

What happens is that you go into customer center, it flashes but then automatically dismisses as soon as it loads, because the internal navigation stack and the external one kinda fight each other. If you try to tap again, it errors out. 

The solution is actually dead simple - if we're already in a `NavigationStack`, we don't create a new one. 

What I don't love about this solution is that we do kind of give the consumer the responsibility of passing in this parameter correctly, but I tried my best to find ways around it to auto-detect and couldn't come up with something reliable. 

And it doesn't feel that complicated of a thing to pass in. 

Still to be done: 

- API Tests
- Do we need to update the presentation modes? Do we want to reuse presentation modes? I kinda tried but it didn't fit all that well since it's meant for sheets, but maybe it can be fitted better. 
- Do we need to update events / metrics for this?

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/CompatibilityNavigationStack.swift
+++ b/RevenueCatUI/CustomerCenter/Views/CompatibilityNavigationStack.swift
@@ -24,9 +24,18 @@ import SwiftUI
 struct CompatibilityNavigationStack<Content: View>: View {
 
     @ViewBuilder var content: Content
+    let isInNavigationStack: Bool
+
+    init(isInNavigationStack: Bool = false,
+         @ViewBuilder content: () -> Content) {
+        self.isInNavigationStack = isInNavigationStack
+        self.content = content()
+    }
 
     var body: some View {
-        if #available(iOS 16.0, *) {
+        if isInNavigationStack {
+            content
+        } else if #available(iOS 16.0, *) {
             NavigationStack {
                 content
             }
--- a/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
@@ -42,30 +42,40 @@ public struct CustomerCenterView: View {
     private var colorScheme
 
     private let mode: CustomerCenterPresentationMode
+    private let isInNavigationStack: Bool
 
     /// Create a view to handle common customer support tasks
     /// - Parameters:
     ///   - customerCenterActionHandler: An optional `CustomerCenterActionHandler` to handle actions
     ///   from the Customer Center.
-    public init(customerCenterActionHandler: CustomerCenterActionHandler? = nil) {
-        self.init(customerCenterActionHandler: customerCenterActionHandler, mode: .default)
+    ///   - isInNavigationStack: Whether this view is already inside a navigation stack
+    public init(customerCenterActionHandler: CustomerCenterActionHandler? = nil,
+                isInNavigationStack: Bool = false) {
+        self.init(customerCenterActionHandler: customerCenterActionHandler,
+                  mode: .default,
+                  isInNavigationStack: isInNavigationStack)
     }
 
     /// Create a view to handle common customer support tasks
     /// - Parameters:
     ///   - customerCenterActionHandler: An optional `CustomerCenterActionHandler` to handle actions
     ///   from the Customer Center.
+    ///   - mode: The presentation mode for the Customer Center
+    ///   - isInNavigationStack: Whether this view is already inside a navigation stack
     init(customerCenterActionHandler: CustomerCenterActionHandler? = nil,
-         mode: CustomerCenterPresentationMode) {
+         mode: CustomerCenterPresentationMode,
+         isInNavigationStack: Bool = false) {
         self._viewModel = .init(wrappedValue:
                                     CustomerCenterViewModel(customerCenterActionHandler: customerCenterActionHandler))
         self.mode = mode
+        self.isInNavigationStack = isInNavigationStack
     }
 
     fileprivate init(viewModel: CustomerCenterViewModel,
                      mode: CustomerCenterPresentationMode = CustomerCenterPresentationMode.default) {
         self._viewModel = .init(wrappedValue: viewModel)
         self.mode = mode
+        self.isInNavigationStack = false
     }
 
     // swiftlint:disable:next missing_docs
@@ -157,7 +167,7 @@ private extension CustomerCenterView {
         let accentColor = Color.from(colorInformation: configuration.appearance.accentColor,
                                      for: self.colorScheme)
 
-        CompatibilityNavigationStack {
+        CompatibilityNavigationStack(isInNavigationStack: isInNavigationStack) {
             destinationContent(configuration: configuration)
         }
         .applyIf(accentColor != nil, apply: { $0.tint(accentColor) })


RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -42,30 +42,40 @@ public struct CustomerCenterView: View {
...
     /// Create a view to handle common customer support tasks
     /// - Parameters:
     ///   - customerCenterActionHandler: An optional `CustomerCenterActionHandler` to handle actions
     ///   from the Customer Center.
+    ///   - mode: The presentation mode for the Customer Center
+    ///   - isInNavigationStack: Whether this view is already inside a navigation stack
     init(customerCenterActionHandler: CustomerCenterActionHandler? = nil,
-         mode: CustomerCenterPresentationMode) {
+         mode: CustomerCenterPresentationMode,
+         isInNavigationStack: Bool = false) {
```
aboedo, line 67: maybe we don't need to add it to both, I was just shooting for speed 😅

aboedo:
@vegaro hope this is helpful! Can you take it from here to the finish line if it is?

RevenueCatUI/CustomerCenter/Views/CompatibilityNavigationStack.swift
```diff
@@ -24,9 +24,18 @@ import SwiftUI
...
+         @ViewBuilder content: () -> Content) {
+        self.isInNavigationStack = isInNavigationStack
+        self.content = content()
+    }
 
     var body: some View {
-        if #available(iOS 16.0, *) {
+        if isInNavigationStack {
+            content
+        } else if #available(iOS 16.0, *) {
```
aboedo, line 38: this is the heart of it

RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
```diff
@@ -164,7 +162,7 @@ struct RestorePurchasesAlert: ViewModifier {
 
     private func dismissAlert() {
         self.alertType = .restorePurchases
-        dismiss()
+        self.isPresented = false
```
aboedo, line 165: also found this when testing - if we call `dismiss` from SwiftUI, we'll dismiss the entire thing instead of just the alert. Feels like we just want to dismiss the alert, right?

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -42,30 +42,40 @@ public struct CustomerCenterView: View {
...
 
     /// Create a view to handle common customer support tasks
     /// - Parameters:
     ///   - customerCenterActionHandler: An optional `CustomerCenterActionHandler` to handle actions
     ///   from the Customer Center.
-    public init(customerCenterActionHandler: CustomerCenterActionHandler? = nil) {
-        self.init(customerCenterActionHandler: customerCenterActionHandler, mode: .default)
+    ///   - isInNavigationStack: Whether this view is already inside a navigation stack
+    public init(customerCenterActionHandler: CustomerCenterActionHandler? = nil,
+                isInNavigationStack: Bool = false) {
```
vegaro, line 53: Yes, I think if it's not possible to detect automatically, this boolean makes sense. The only thing I see is that `NavigationStack` is an iOS 16 concept, and before it was `NavigationView`

RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
vegaro: Oh interesting 🤔 

Yeah, we want to dismiss only the alert. It looks like a bug. I will test and probably open a separate PR with this only

RevenueCatUI/CustomerCenter/Views/CompatibilityNavigationStack.swift
facumenzella: I think this shouldn't belong to `CompatibilityNavigationStack`. It will only make internal components more complex.

We can handle it directly inside `destinationView`.

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
facumenzella: What about `shouldHandleNavigation` instead? Just to avoid mentioning the internals of it

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -157,7 +167,7 @@ private extension CustomerCenterView {
         let accentColor = Color.from(colorInformation: configuration.appearance.accentColor,
                                      for: self.colorScheme)
 
-        CompatibilityNavigationStack {
+        CompatibilityNavigationStack(isInNavigationStack: isInNavigationStack) {
             destinationContent(configuration: configuration)
         }
```
facumenzella, line 172:
```suggestion
        if isInNavigationStack { 
             CompatibilityNavigationStack(isInNavigationStack: isInNavigationStack) {
                 destinationContent(configuration: configuration)
             }
         } else {
             destinationContent(configuration: configuration)
         }
```

RevenueCatUI/CustomerCenter/Views/CompatibilityNavigationStack.swift
ajpallares: Yep. I agree with @facumenzella here. `CompatibilityNavigationStack` should just be an abstraction for both `NavigationStack` and `NavigationView`. Now the `isInNavigationStack` logic should probably be *outside* `CompatibilityNavigationStack`

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
ajpallares: I don't quite like this boolean, but after going back and forth with @facumenzella we haven't been able to find a more elegant solution. There's not avoiding this to work around the SwiftUI limitation of knowing if the view is inside a navigation view/stack 😓.

Regarding the naming, even though I understand @facumenzella's point, I think `isInNavigationStack` is more clear.

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
ajpallares: In fact, I don't think the `mode` makes sense at all if `isInNavigationStack` is `true`. I mean, if the view is going to be shown inside a navigation view/stack, then it won't be presented using any of the `CustomerCenterPresentationMode` cases (as `CustomerCenterPresentationMode` refers to modal presentation modes only).

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
ajpallares: Btw, since the suggested change already uses `isInNavigationStack` in the `if`, then `CompatibilityNavigationStack` init would not receive this `bool` anymore

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -43,29 +43,53 @@ public struct CustomerCenterView: View {
 
     private let mode: CustomerCenterPresentationMode
 
+    /// A flag indicating whether the view is already embedded in a navigation stack.
+    ///
+    /// - When set to `true`, the view must be part of an existing NavigationStack / NavigationView.
```
ajpallares, line 48: Small detail. I'd give NavigationStack and NavigationView code formatting:
* NavigationStack --> `NavigationStack`
* NavigationView --> `NavigationView`

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -43,29 +43,53 @@ public struct CustomerCenterView: View {
...
+    /// one internally.
+    private let isEmbeddedInNavigationStack: Bool
+
     /// Create a view to handle common customer support tasks
     /// - Parameters:
     ///   - customerCenterActionHandler: An optional `CustomerCenterActionHandler` to handle actions
     ///   from the Customer Center.
-    public init(customerCenterActionHandler: CustomerCenterActionHandler? = nil) {
-        self.init(customerCenterActionHandler: customerCenterActionHandler, mode: .default)
+    ///   - isEmbeddedInNavigationStack: Whether this view is already inside a NavigationStack / NavigationView
```
ajpallares, line 57: Same here: NavigationView --> `NavigationView`

ajpallares (approved):

RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
facumenzella: https://github.com/RevenueCat/purchases-ios/pull/4653

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -43,29 +43,53 @@ public struct CustomerCenterView: View {
...
         self.mode = mode
+        self.isEmbeddedInNavigationStack = isEmbeddedInNavigationStack
     }
 
-    fileprivate init(viewModel: CustomerCenterViewModel,
-                     mode: CustomerCenterPresentationMode = CustomerCenterPresentationMode.default) {
+    fileprivate init(
+        viewModel: CustomerCenterViewModel,
+        mode: CustomerCenterPresentationMode = CustomerCenterPresentationMode.default
+    ) {
```
ajpallares, line 89: Btw, do we know if there is some agreed standard on about this kind of formatting?

I see you've changed from 
```
init(param1: Type1,
     param2: Type2)
```

to 

```
init(
     param1: Type1,
     param2: Type2
)
```

Whereas I prefer the former, I don't have any problem with agreeing on the latter. But perhaps we should all stick to the same style?

hiddevdploeg:
Personally I'd prefer something like this:
 
```swift
NavigationLink("Customer Center") {
     CustomerCenterView()
}

ContentView()
.sheet(isPresented: $presentCustomerCenter) {
     CustomerCenterSheetView()
}

```

But I understand this would mean a everyone right now would have to refactor their current implementation so maybe `CustomerCenterNavigationView()` would make more sense?

Or we would create our own `NavigationLink` that looks something like this:

```swift
CustomerCenterNavigationLink {
     // Label goes here
}
```

Just my 2cts 😄

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
facumenzella: I've reverted this for now 👍

facumenzella:
@hiddevdploeg I'll open a follow-up PR after this one addressing your comments. I like it

tonidero:
Just a heads up that I'm adding the `pr:RevenueCatUI` label so the changelog gets generated correctly

PR #4625, opened 2024-12-31
Add text in default language to `feedbackSurveyCompleted` callback

vegaro:
With the changes we did to the localization UI, this callback is kinda broken since the ids are auto-generated, and mean nothing to the developer.

After this PR we'll be sending the English translation of the option as well so it's easier to understand what each option represent. It requires backend changes to send the English/default language translation to the callback as well.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/CustomerCenterAction.swift
+++ b/RevenueCatUI/CustomerCenter/Data/CustomerCenterAction.swift
@@ -23,7 +23,13 @@ public enum CustomerCenterAction {
     /// - Parameter refundRequestStatus: The status of the refund request
     case refundRequestCompleted(_ refundRequestStatus: RefundRequestStatus)
     /// An option of the feedback survey has been selected
-    /// - Parameter feedbackSurveyOptionId: The id of the feedback survey option selected
+    /// - Parameter feedbackSurveyOptionId: The id of the selected feedback survey option
     case feedbackSurveyCompleted(_ feedbackSurveyOptionId: String)
+    /// An option of the feedback survey has been selected
+    /// - Parameter feedbackSurveyOptionId: The id of the selected feedback survey option
+    /// - Parameter titleInDefaultLocale: The displayed text of the selected feedback survey
+    /// option in the default locale (English by default)
+    case feedbackSurveyCompletedWithTitleInDefaultLocale(_ feedbackSurveyOptionId: String,
+                                                         _ titleInDefaultLocale: String)
 
 }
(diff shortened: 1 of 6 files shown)


RevenueCatUI/CustomerCenter/Data/CustomerCenterAction.swift
```diff
@@ -23,7 +23,11 @@ public enum CustomerCenterAction {
     /// - Parameter refundRequestStatus: The status of the refund request
     case refundRequestCompleted(_ refundRequestStatus: RefundRequestStatus)
     /// An option of the feedback survey has been selected
-    /// - Parameter feedbackSurveyOptionId: The id of the feedback survey option selected
+    /// - Parameter feedbackSurveyOptionId: The id of the selected feedback survey option
     case feedbackSurveyCompleted(_ feedbackSurveyOptionId: String)
+    /// An option of the feedback survey has been selected
+    /// - Parameter feedbackSurveyOptionId: The id of the selected feedback survey option
+    /// - Parameter defaultLanguageText: The displayed text of the selected feedback survey option (in English)
+    case feedbackSurveyCompleted(_ feedbackSurveyOptionId: String, defaultLanguageText: String)
```
vegaro, line 31: unsure about the name of hte parameter

RevenueCatUI/CustomerCenter/Data/CustomerCenterAction.swift
```diff
@@ -23,7 +23,10 @@ public enum CustomerCenterAction {
     /// - Parameter refundRequestStatus: The status of the refund request
     case refundRequestCompleted(_ refundRequestStatus: RefundRequestStatus)
     /// An option of the feedback survey has been selected
-    /// - Parameter feedbackSurveyOptionId: The id of the feedback survey option selected
+    /// - Parameter feedbackSurveyOptionId: The id of the selected feedback survey option
     case feedbackSurveyCompleted(_ feedbackSurveyOptionId: String)
+    /// An option of the feedback survey has been selected
+    /// - Parameter option: The ``CustomerCenterConfigData.HelpPath.FeedbackSurvey.Option`` that was selected
+    case feedbackSurveyCompletedWithOption(_ option: CustomerCenterConfigData.HelpPath.FeedbackSurvey.Option)
```
vegaro, line 30: I ended up changing the name of the case because I coudln't find a way where the Swift compiler wouldn't get confused.

I tried adding another parameter and keeping the name and got this:

Screenshot 2025-01-21 at 12 42 55


I also tried returning the whole option (keeping one parameter only) and got this:
Screenshot 2025-01-21 at 12 58 59

I am also concerned because Switch must be exhaustive, meaning this is a breaking change:
Screenshot 2025-01-21 at 13 03 52

PR #4629, opened 2025-01-02
Expose parse as web purchase redemption APIs for hybrids

tonidero:
### Description
While working on supporting web2app flows in our KMP SDK, I noticed we're mostly not using PHC for many of the functionalities, but using the native SDKs directly. The problem with using the PHC `redeemWebPurchase` method was that that method returns a map for the `CustomerInfo` that needs to be mapped back to a `CustomerInfo` object in KMP. For existing methods, we are instead performing that step using the native types, and performing the mapping from there (which is much more type safe). If I wanted to use the PHC methods, I would have need to write a new mapper from a Map to a `CustomerInfo` object in KMP. I felt it was better to follow what we have and continue using the native types here.

However, I wasn't able to access the `parseAsWebPurchaseRedemption` nor the extension function for `URL` from KMP. I believe in order to access these, I need to make these functions public in the type, and not in a non-objc extension. Additionally, I added a method that takes a urlString instead of a `URL` for convenience.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -154,6 +154,21 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
      */
     @objc public static func canMakePayments() -> Bool { StoreKit1Wrapper.canMakePayments() }
 
+    /// Parses a deep link URL to verify it's a RevenueCat web purchase redemption link
+    /// - Seealso: ``Purchases/redeemWebPurchase(_:)``
+    @objc public static func parseAsWebPurchaseRedemption(_ url: URL) -> WebPurchaseRedemption? {
+        return DeepLinkParser.parseAsWebPurchaseRedemption(url)
+    }
+
+    /// Parses a deep link URL string to verify it's a RevenueCat web purchase redemption link
+    /// - Seealso: ``Purchases/redeemWebPurchase(_:)``
+    @objc public static func parseUrlStringAsWebPurchaseRedemption(_ urlString: String) -> WebPurchaseRedemption? {
+        guard let url = URL(string: urlString) else {
+            return nil
+        }
+        return DeepLinkParser.parseAsWebPurchaseRedemption(url)
+    }
+
     /**
      * Set a custom log handler for redirecting logs to your own logging system.
      *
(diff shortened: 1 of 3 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -154,6 +154,21 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
      */
     @objc public static func canMakePayments() -> Bool { StoreKit1Wrapper.canMakePayments() }
 
+    /// Parses a deep link URL to verify it's a RevenueCat web purchase redemption link
+    /// - Seealso: ``Purchases/redeemWebPurchase(_:)``
+    @objc public static func parseAsWebPurchaseRedemption(_ url: URL) -> WebPurchaseRedemption? {
```
JayShortway, line 159: I'm just curious, did you check what `URL` looks like from the Kotlin side? 😄

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -154,6 +154,21 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
...
+    }
+
+    /// Parses a deep link URL string to verify it's a RevenueCat web purchase redemption link
+    /// - Seealso: ``Purchases/redeemWebPurchase(_:)``
+    @objc public static func parseUrlStringAsWebPurchaseRedemption(_ urlString: String) -> WebPurchaseRedemption? {
+        guard let url = URL(string: urlString) else {
+            return nil
+        }
+        return DeepLinkParser.parseAsWebPurchaseRedemption(url)
+    }
```
JayShortway, line 170: >However, I wasn't able to access the parseAsWebPurchaseRedemption nor the extension function for URL from KMP. I believe in order to access these, I need to make these functions public in the type, and not in a non-objc extension.  
  
Did you check if they're now available in KMP? If not, we could move this function (with `urlString`) to PHC.

JayShortway (approved):
Nice one! Fully agree on not writing a `CustomerInfo` mapping function in KMP.

Sources/Purchasing/Purchases/Purchases.swift
tonidero: It's available as `NSURL`, which is the ObjC equivalent

Sources/Purchasing/Purchases/Purchases.swift
tonidero: That's a very good point... I thought to bypass PHC entirely, since we are using the redeem function from the native... but this is probably better there, and avoids exposing extra functions. Will close this and move it there, thanks!

tonidero: Moved it to PHC in https://github.com/RevenueCat/purchases-hybrid-common/pull/1004

JayShortway: Anything that's part of PHC's public API should be available from KMP. If PHC's public API uses a native iOS type, that type is part of the public API and becomes available from KMP. I'm just not sure how that works for static functions, so that's why I asked haha. Moving it to PHC definitely works!

PR #4630, opened 2025-01-02
[Paywalls] Use CALayer-backed shadows and refactor Shape.swift

MarkVillacampa:
### Motivation

By default shadows in SwiftUI are applied to children views as well, so we implemented a workaround by using `.compositingGroup()`.

However when the background color of a view with a shadow has less than 100% opacity, the `.compositingGroup()` workaround no longer works. Additionally, the shadow inherits the opacity of the view which is less than ideal.

### Description

In this PR, we replace SwiftUI shadows with a CALayer-backed solution. Because this solution requires knowing the shape of the view, I took the opportunity to refactor the Shape modifier code.

I didn't modify any current SwiftUI previews so make sure nothing broke. The only changes were minor rendering differences in the shadows which are not perceptible with your naked eye, and a change in how a shadow applied to a `Text` view renders, as it now applies to its bounding rectangle and not to the text itself. Fortunately, we don't currently support shadows in text.

Screenshot 2025-01-03 at 22 24 46

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
@@ -90,13 +90,11 @@ struct StackComponentView: View {
         }
         .padding(style.padding)
         .padding(additionalPadding)
-        .backgroundStyle(style.backgroundStyle, uiConfigProvider: self.viewModel.uiConfigProvider)
         .shape(border: style.border,
-               shape: style.shape)
-        .applyIfLet(style.shadow) { view, shadow in
-            // Without compositingGroup(), the shadow is applied to the stack's children as well.
-            view.compositingGroup().shadow(shadow: shadow)
-        }
+               shape: style.shape,
+               shadow: style.shadow,
+               background: style.backgroundStyle,
+               uiConfigProvider: self.viewModel.uiConfigProvider)
         .padding(style.margin)
     }
 
--- a/RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
+++ b/RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
@@ -30,10 +30,10 @@ struct BackgroundStyleModifier: ViewModifier {
     var colorScheme
 
     var backgroundStyle: BackgroundStyle?
-    var uiConfigProvider: UIConfigProvider
+    var uiConfigProvider: UIConfigProvider?
 
     func body(content: Content) -> some View {
-        if let backgroundStyle {
+        if let backgroundStyle, let uiConfigProvider {
             content
                 .apply(backgroundStyle: backgroundStyle, colorScheme: colorScheme, uiConfigProvider: uiConfigProvider)
         } else {
@@ -96,7 +96,7 @@ fileprivate extension View {
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 extension View {
 
-    func backgroundStyle(_ backgroundStyle: BackgroundStyle?, uiConfigProvider: UIConfigProvider) -> some View {
+    func backgroundStyle(_ backgroundStyle: BackgroundStyle?, uiConfigProvider: UIConfigProvider?) -> some View {
         self.modifier(BackgroundStyleModifier(backgroundStyle: backgroundStyle, uiConfigProvider: uiConfigProvider))
     }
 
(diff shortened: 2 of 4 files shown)


RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
```diff
@@ -90,13 +90,10 @@ struct StackComponentView: View {
...
-        .backgroundStyle(style.backgroundStyle)
         .shape(border: style.border,
-               shape: style.shape)
-        .applyIfLet(style.shadow) { view, shadow in
-            // Without compositingGroup(), the shadow is applied to the stack's children as well.
-            view.compositingGroup().shadow(shadow: shadow)
-        }
+               shape: style.shape,
+               shadow: style.shadow,
+               background: style.backgroundStyle)
```
MarkVillacampa, line 96: The `border` already depended on the `shape`, and now the `shadow` does as well, so we include it in the `shape` modifier.

Since the `shape` modifier already does shape clipping of the background, and requires the `backgroundStyle` modifier to be applied right before, it makes sense to move it inside the `shape` modifier as well so we can more easily track the interactions between all the pieces.

RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
```diff
@@ -16,7 +16,7 @@ import SwiftUI
 
 #if PAYWALL_COMPONENTS
 
-enum BackgroundStyle {
+enum BackgroundStyle: Hashable {
```
MarkVillacampa, line 19: Added a bunch of `Hashable`s to be able to programmatically create previews of all combinations using `ForEach`.

RevenueCatUI/Templates/V2/ViewHelpers/ShadowModifier.swift
```diff
@@ -49,9 +49,122 @@ struct ShadowModifier: ViewModifier {
...
+// Using the .shadow() modified to add a drop shadow in SwiftUI has multiple downsides:
+// - The shadow is applied to all children views (can be worked around with .compositingGroup())
+// - The `.compositingGroup()` workaround stops working if the view has less than 100% opacity.
+// - The shadow inherits the opacity of the view, making it impossible to have a translucent
+//   or transparent view with a 100% opacity shadow.
+//
+// ShadowUIView tries to work around these limitations by rendering the shadow in a backing UIView,
+// and using a Shape to mask off the inner part of the shadow.
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+private struct LayerShadowView: UIViewRepresentable {
```
MarkVillacampa, line 102: This is the main addition of the PR and the way we fix SwiftUI shadows by using a CALayer shadow instead. Documented in code as it feels like important context.

RevenueCatUI/Templates/V2/ViewHelpers/Shape.swift
```diff
@@ -59,35 +60,52 @@ struct ShapeModifier: ViewModifier {
...
-                .conditionalOverlay(color: self.border?.color,
-                                    width: self.border?.width,
-                                    topLeft: radiuses?.topLeft,
-                                    topRight: radiuses?.topRight,
-                                    bottomLeft: radiuses?.bottomLeft,
-                                    bottomRight: radiuses?.bottomRight)
+                .backgroundStyle(background)
+                // We want to clip only in case there is a non-Rectangle shape
+                // or if there's a border, otherwise we let the background color
+                // extend behind the safe areas
```
MarkVillacampa, line 85: I noticed this when it changed in an Emerge snapshot.

When the shape is a rectangle without rounded corners and there is no border, if we don't clip the background, it can extend behind the safe areas. This might be useful for the root view and for the sticky footer view so it felt important to keep the behavior and document it.

Screenshot 2025-01-03 at 22 08 19
Screenshot 2025-01-03 at 22 08 51

RevenueCatUI/Templates/V2/ViewHelpers/Shape.swift
```diff
@@ -290,7 +258,91 @@ extension View {
...
+        let shadows: [ShadowModifier.ShadowInfo?] = [
+            .init(color: .black, radius: 3, x: 2, y: 2),
+            nil
+        ]
+
+        ForEach(backgrounds, id: \.self) { background in
+            VStack {
+                ForEach(shapes, id: \.self) { shape in
+                    ForEach(borders, id: \.self) { border in
+                        ForEach(shadows, id: \.self) { shadow in
```
MarkVillacampa, line 329: Check the Emerge snapshots for all the new previews :)

joshdholtz (approved):
:shipit: :shipit: :shipit: :shipit: :shipit: :shipit:

PR #4632, opened 2025-01-06
[Paywalls V2] Add support for alias solid hex colors (not gradients)

joshdholtz:
### Motivation

Add support for aliased `hex` colors (for when dashboard supports them)

⚠️ Gradients will required a bigger rework (a Linear ticket has been added for this)

### Description

Passes a `UIConfigProvider` into the `toColor()` method which does the alias lookup

### DIFF
--- a/RevenueCatUI/Templates/V2/ViewHelpers/ForegroundColorScheme.swift
+++ b/RevenueCatUI/Templates/V2/ViewHelpers/ForegroundColorScheme.swift
@@ -23,27 +23,48 @@ struct ForegroundColorSchemeModifier: ViewModifier {
     var colorScheme
 
     var foregroundColorScheme: PaywallComponent.ColorScheme
+    var uiConfigProvider: UIConfigProvider
 
     func body(content: Content) -> some View {
-        content.foregroundColorScheme(foregroundColorScheme, colorScheme: colorScheme)
+        content.foregroundColorScheme(
+            self.foregroundColorScheme,
+            colorScheme: self.colorScheme,
+            uiConfigProvider: self.uiConfigProvider
+        )
     }
 
 }
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 extension View {
-    func foregroundColorScheme(_ colorScheme: PaywallComponent.ColorScheme) -> some View {
-        self.modifier(ForegroundColorSchemeModifier(foregroundColorScheme: colorScheme))
+    func foregroundColorScheme(
+        _ colorScheme: PaywallComponent.ColorScheme,
+        uiConfigProvider: UIConfigProvider
+    ) -> some View {
+        self.modifier(ForegroundColorSchemeModifier(foregroundColorScheme: colorScheme,
+                                                    uiConfigProvider: uiConfigProvider))
     }
 }
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 fileprivate extension View {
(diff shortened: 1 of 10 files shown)


RevenueCatUI/Templates/V2/ViewHelpers/ForegroundColorScheme.swift
```diff
@@ -23,27 +23,48 @@ struct ForegroundColorSchemeModifier: ViewModifier {
...
+        colorScheme: ColorScheme,
+        uiConfigProvider: UIConfigProvider
+    ) -> some View {
         switch color.effectiveColor(for: colorScheme) {
         case .hex, .alias:
-            self.foregroundColor(color.toDynamicColor())
+            let color = color.toDynamicColor(uiConfigProvider: uiConfigProvider)
+
+            // Only apply foreground color if not clear
+            // This is a way to
```
JayShortway, line 62: Are these lines switched around? 😄

RevenueCatUI/Templates/V2/ViewModelHelpers/PaywallComponentTypeTransformers.swift
```diff
@@ -239,13 +238,23 @@ extension PaywallComponent.FitMode {
...
+        case .alias(let alias):
+            guard let aliasColor = uiConfigProvider.getColor(for: alias) else {
+                Logger.warning("Aliased color '\(alias)' does not exist.")
+                return fallback
+            }
+
+            // Alias should never have an alias
+            // Using fallback so recursion doesn't happen
+            if case .alias = aliasColor {
+                return fallback
```
JayShortway, line 254: Should we log an error in this case too?

RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
```diff
@@ -24,6 +24,14 @@ struct UIConfigProvider {
         self.uiConfig = uiConfig
     }
 
+    func getColor(for name: String?) -> PaywallComponent.ColorInfo? {
+        guard let name else {
+            return nil
+        }
```
JayShortway, line 30: Hmm, why is `name` nullable here? I'd consider that invalid input (every alias color has a name), and so I would enforce that at compilation time instead of at runtime.

JayShortway (commented):
Looking good! 1 suggestion around the nullable `name` parameter.

RevenueCatUI/Templates/V2/ViewHelpers/ForegroundColorScheme.swift
MarkVillacampa: Curious why only apply if color is not clear? 🤔

joshdholtz: @JayShortway @MarkVillacampa 

Ooops, I probably feel asleep during this comment 😛

Clear is the fallback color that is used internally in this method and clear text is... really bad 😬 Clear background is fine?

To make sure this never happens, I'm going to eventually add validation to make sure the alias exists in:
1) Backend
2) Loading of paywall

RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
MarkVillacampa: ➕

MarkVillacampa (approved):

PR #4635, opened 2025-01-07
[EXTERNAL] Lock RateLimiter.shouldProceed() entirely to avoid race conditions

### LINKED ISSUE #4440: Bug: Data race in Ratelimiter.
https://github.com/RevenueCat/purchases-ios/blame/72b06f69438eff747fe0d6baa408b3a7745d4087/Sources/Misc/RateLimiter.swift#L41

https://github.com/RevenueCat/purchases-ios/blame/72b06f69438eff747fe0d6baa408b3a7745d4087/Sources/Misc/RateLimiter.swift#L42

```
internal class RateLimiter {
    private var timestamps: [Date?]
    private var index: Int = 0
    private let maxCallsInclusive: Int

    let maxCalls: Int
    let period: TimeInterval

    init(maxCalls: Int, period: TimeInterval) {
        self.maxCalls = maxCalls
        self.maxCallsInclusive = self.maxCalls + 1
        self.period = period

        self.timestamps = Array(repeating: nil, count: maxCallsInclusive)
    }

    func shouldProceed() -> Bool {
        let now = Date()
        let oldestIndex = (index + 1) % maxCallsInclusive
        let oldestTimestamp = timestamps[oldestIndex]

        // Check if the oldest timestamp is outside the rate limiting period or if it's nil
        if let oldestTimestamp = oldestTimestamp, now.timeIntervalSince(oldestTimestamp) <= period {
            return false
        } else {
            timestamps[index] = now
            index = oldestIndex
            return true
        }
    }
}
```

Some alternatives here:

* Make it an actor, so the shouldProceed signature changes to async
* Use atomics in all properties, maybe is too much
* Use a dispatchqueue as class property and use it in the whole shouldProceed method making it sync 
* Use a lock(.recursive)

nguyenhuy:
### Checklist

### Motivation
Introduced in #3709, `RateLimiter` is used exclusively by `Purchases.syncAttributesAndOfferingsIfNeeded(completion:)`. The [documentation of this method](https://www.revenuecat.com/docs/tools/targeting/custom-attributes#the-syncattributesandofferingsifneeded-method) doesn't mention any threading requirement, so it's reasonable to assume that it can be called on any thread. This means `RateLimiter` needs to be thread safe, otherwise clients will run into race conditions as reported in #4440.

### Description
Fortunately the `RateLimiter` class is quite simple.

Its contants don't need to be atomic/locked since they are thread safe by definition. Turning `timestamps` and `index` vars into atomics doesn't help much in this case because `shouldProceed()` needs multiple reads and writes in a "transaction".

Using a `Lock` to wrap the whole method is the simplest and self-contained approach so I picked it.  Android is doing the same thing ([here](https://github.com/RevenueCat/purchases-android/blob/eb3c8f4aa1a6993f12d6e2f93cc82d26063101a6/purchases/src/main/kotlin/com/revenuecat/purchases/utils/RateLimiter.kt#L10)).

The lock is a non-recursive lock because there is no need for reentrant. And in principle I avoid recursive locks as much as possible since one can easily lose control over the locks, especially in more complicated use cases/codebases. This is a big problem that we have had for a long time in AsyncDisplayKit/Texture.

Unit tests remain the same because testing for race conditions is rather difficult and unpredictable. Texture has a bunch of thrash tests that kind of work but not 100%. Since the subject under test in this case is rather simple and the fix is straightforward it should work. Otherwise, we can have follow up PRs for both iOS and [Android](https://github.com/RevenueCat/purchases-android/blob/eb3c8f4aa1a6993f12d6e2f93cc82d26063101a6/purchases/src/test/java/com/revenuecat/purchases/utils/RateLimiterTest.kt).

This should resolve #4440.

This PR also renames `RateLimiterRests.swift` to `RateLimiterTests.swift`.

### DIFF
--- a/Sources/Misc/RateLimiter.swift
+++ b/Sources/Misc/RateLimiter.swift
@@ -13,7 +13,7 @@
 
 import Foundation
 
-internal class RateLimiter {
+internal final class RateLimiter: @unchecked Sendable {
     private let lock = Lock()
     private var timestamps: [Date?]
     private var index: Int = 0


JayShortway:
Hi @nguyenhuy, thanks a lot for your contribution! It makes a lot of sense. I'll make sure it gets merged!

nguyenhuy:
@JayShortway: That was fast! ❤️

PR #4640, opened 2025-01-08
[Paywalls V2] Decode rectangle corners as optional

joshdholtz:
### Motivation

Decode rectangle corners as optional

### Description

Using `decodeIfPresent` and explicit optional typing to make things valid

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -24,7 +24,7 @@ var additionalCompilerFlags: [PackageDescription.SwiftSetting] = {
         .split(whereSeparator: \.isWhitespace)
         .filter { !$0.isEmpty && !$0.hasPrefix("$") }
         .map { .define(String($0)) }
-        ?? []
+    ?? [.define("PAYWALL_COMPONENTS")]
 }()
 
 var ciCompilerFlags: [PackageDescription.SwiftSetting] = [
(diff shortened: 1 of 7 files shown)


Package.swift
```diff
@@ -24,7 +24,7 @@ var additionalCompilerFlags: [PackageDescription.SwiftSetting] = {
         .split(whereSeparator: \.isWhitespace)
         .filter { !$0.isEmpty && !$0.hasPrefix("$") }
         .map { .define(String($0)) }
-        ?? []
+    ?? [.define("PAYWALL_COMPONENTS")]
```
tonidero, line 27: Probably not want to define this here?

Tests/TestingApps/PaywallsTester/PaywallsTester/Config/ConfigItem.swift
```diff
@@ -14,7 +14,7 @@ protocol AvailableConfigItems {
 
 // CI system adds keys here
 extension AvailableConfigItems {
-    static var apiKey: String { "" }
+    static var apiKey: String { "appl_fpIYNnFHeCcvJRZqibQfQOTUusd" }
```
tonidero, line 17: Also this should be reverted

tonidero (approved):

PR #4641, opened 2025-01-08
[Customer Center] Hack week: add contact support button to customer center

ekurutepe:
### Checklist

### Motivation
It seems like we have the ability to open system email client to contact support but we were not surfacing this for the users. See the discussion: https://revenuecat.slack.com/archives/C072SLPSTQS/p1736178563670389

### Description
Adds a user tappable ContactSupportButton to customer center at the top level for all purchase states.

### DIFF
--- a/RevenueCatUI/CustomerCenter/URLUtilities.swift
+++ b/RevenueCatUI/CustomerCenter/URLUtilities.swift
@@ -65,11 +65,15 @@ enum URLUtilities {
     }
 
     static func canOpenURL(_ url: URL) -> Bool {
+#if DEBUG
+        return true
+#else
         guard !Self.isAppExtension,
               let application = Self.sharedUIApplication else {
             return false
         }
         return application.canOpenURL(url)
+#endif
     }
 
 }
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -364,6 +364,7 @@ PLATFORMS
   arm64-darwin-21
   arm64-darwin-22
   arm64-darwin-23
+  arm64-darwin-24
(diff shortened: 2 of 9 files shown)


RevenueCatUI/CustomerCenter/URLUtilities.swift
```diff
@@ -65,11 +65,15 @@ enum URLUtilities {
...
     static func canOpenURL(_ url: URL) -> Bool {
+#if DEBUG
+        return true
+#else
         guard !Self.isAppExtension,
               let application = Self.sharedUIApplication else {
             return false
         }
         return application.canOpenURL(url)
+#endif
```
aboedo, line 76: curious about this part - shouldn't it still be false for app extensions even when in debug?

aboedo (commented):
Love it!! Left one question

vegaro (commented):
I do think we have to add the contact support button but with this changes, it will be hardcoded for everyone. 

Instead I think we should make it one of the paths in the dashboard (same as "Manage subscriptions", or "Request a refund"), so it's optional and dynamically added by the `ManageSubscriptionsButtonsView`

aboedo:
@ekurutepe I'm ridiculously late to this one but I think backend-driven is probably the longer-term solution. Are we good to close this one out until we get to build this backend-side?

ekurutepe:
> @ekurutepe I'm ridiculously late to this one but I think backend-driven is probably the longer-term solution. Are we good to close this one out until we get to build this backend-side?

oops, this is an abandoned PR. Sorry for wasting your time @aboedo.

PR #4485, opened 2024-11-13
Make PurchaseParams API & Win-Back Functions Public

fire-at-will:
### Description
This PR will ship the following functionality:
- PurchaseParams API
- Fetching & purchasing win-back offers on a developer's paywall

### Specific Changes
- Makes the `PurchaseParams` functionality publicly available by removing the `ENABLE_PURCHASE_PARAMS` compiler flag checks
- Makes the `PurchaseParams` functionality unavailable if the `ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION` flag is used
- Make the functions to fetch & redeem eligible win-back offers public and exposed to Objective-C
- Makes the new transaction metadata functionality public
- Adds the new functions to the purchase testers
- Hides the transaction metadata feature behind a new `ENABLE_TRANSACTION_METADATA` compiler flag
- Changes the function signature of the completion handler-based function to fetch eligible win-back offers from
```swift
@available(iOS 18.0, macOS 15.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
func eligibleWinBackOffers(
  forProduct product: StoreProduct,
  completion: @escaping @Sendable (Result<[WinBackOffer], PublicError>) -> Void
)
```
to
```swift
@available(iOS 18.0, macOS 15.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)
 func eligibleWinBackOffers(
     forProduct product: StoreProduct,
     completion: @escaping @Sendable ([WinBackOffer]?, PublicError?) -> Void
 )
```
So that the function can be called from Objective-C (and therefore the hybrid SDKs that rely on Objective-C)

PR #4645, opened 2025-01-09
Deprecate misnamed purchase(params) function in Obj-C

fire-at-will:
### Motivation
The `purchase(_ params: PurchaseParams, completion: @escaping PurchaseCompletedBlock)` was mistakenly exposed to Objective-C with the name `params` in https://github.com/RevenueCat/purchases-ios/pull/4485.

### Description
This PR deprecates that existing `params` name in Objective-C (`@objc(params:withCompletion:)`) and replaces it with `purchaseWithParams` (`@objc(purchaseWithParams:completion:)`).

### Implementation Details
Swift doesn't support deprecating only an Objective-C selector. You need to create a second Swift function with the correct `@objc` selector and ensure that the Swift function names don't conflict. This isn't ideal because it adds an extra deprecated function to the Swift API, but it gets the job done. 

So to accomplish this, this PR renames the existing Swift function with the incorrect `@objc` selector to `purchaseWithParams` and marks it as deprecated. 

#### Old Obj-C API
```objc
[p params:purchaseParams withCompletion:^(RCStoreTransaction *t, RCCustomerInfo *i, NSError *error, BOOL userCancelled) { }];
```

#### New Obj-C API
```objc
[p purchaseWithParams:purchaseParams completion:^(RCStoreTransaction *t, RCCustomerInfo *i, NSError *error, BOOL userCancelled) { }];
```

### DIFF
--- a/Sources/Misc/Deprecations.swift
+++ b/Sources/Misc/Deprecations.swift
@@ -259,6 +259,15 @@ public extension Purchases {
         self.attribution.setCreative(creative)
     }
 
+    @available(iOS, deprecated, renamed: "purchase(_:completion:)")
+    @available(tvOS, deprecated, renamed: "purchase(_:completion:)")
+    @available(watchOS, deprecated, renamed: "purchase(_:completion:)")
+    @available(macOS, deprecated, renamed: "purchase(_:completion:)")
+    @available(macCatalyst, deprecated, renamed: "purchase(_:completion:)")
+    @objc(params:withCompletion:)
+    func purchaseWithParams(_ params: PurchaseParams, completion: @escaping PurchaseCompletedBlock) {
+        self.purchase(params, completion: completion)
+    }
 }
 
 public extension StoreProduct {
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1028,7 +1028,7 @@ public extension Purchases {
 
     #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
 
-    @objc(params:withCompletion:)
+    @objc(purchaseWithParams:completion:)
     func purchase(_ params: PurchaseParams, completion: @escaping PurchaseCompletedBlock) {
         purchasesOrchestrator.purchase(params: params, completion: completion)
     }
--- a/Tests/APITesters/AllAPITests/ObjcAPITester/RCPurchasesAPI.m
+++ b/Tests/APITesters/AllAPITests/ObjcAPITester/RCPurchasesAPI.m
@@ -125,10 +125,8 @@ NSURL *url;
     [p setCreative: nil];
     [p setCreative: @""];
     #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
-    if (@available(iOS 18.0, macOS 15.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)) {
-        RCPurchaseParams *packageParams;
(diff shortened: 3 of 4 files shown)


fire-at-will:
@RCGitBot please test

Tests/APITesters/AllAPITests/ObjcAPITester/RCPurchasesAPI.m
```diff
@@ -124,6 +124,12 @@ + (void)checkAPI {
     [p setKeyword: @""];
     [p setCreative: nil];
     [p setCreative: @""];
+    #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+    if (@available(iOS 18.0, macOS 15.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)) {
+        RCPurchaseParams *packageParams;
+        [p params:packageParams withCompletion:^(RCStoreTransaction *t, RCCustomerInfo *i, NSError *error, BOOL userCancelled) { }];
```
fire-at-will, line 130: Moves the old call to the deprecated section

aboedo (approved):

Sources/Misc/Deprecations.swift
```diff
@@ -259,6 +259,15 @@ public extension Purchases {
         self.attribution.setCreative(creative)
     }
 
+    @available(iOS, deprecated, renamed: "purchase(_:completion:)")
+    @available(tvOS, deprecated, renamed: "purchase(_:completion:)")
+    @available(watchOS, deprecated, renamed: "purchase(_:completion:)")
+    @available(macOS, deprecated, renamed: "purchase(_:completion:)")
+    @available(macCatalyst, deprecated, renamed: "purchase(_:completion:)")
```
ajpallares, line 266: Just a small comment. We could save some lines with 
```
@available(*, deprecated, renamed: "purchase(_:completion:)")
```
But I see that previous deprecations are done the same (explicitly for each platform) so I guess that, for consistency, we should keep doing it that way.

Tests/APITesters/AllAPITests/ObjcAPITester/RCPurchasesAPI.m
```diff
@@ -124,6 +124,12 @@ + (void)checkAPI {
     [p setKeyword: @""];
     [p setCreative: nil];
     [p setCreative: @""];
+    #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+    if (@available(iOS 18.0, macOS 15.0, tvOS 18.0, watchOS 11.0, visionOS 2.0, *)) {
```
ajpallares, line 128: Is the `@available` `if` really needed here?

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1028,7 +1028,7 @@ public extension Purchases {
 
     #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
 
-    @objc(params:withCompletion:)
+    @objc(purchaseWithParams:completion:)
```
facumenzella, line 1031: I prefer this naming over the older one 🎖️

Tests/APITesters/AllAPITests/ObjcAPITester/RCPurchasesAPI.m
fire-at-will: Great callout! I think I'm too used to doing this for win-back offers 😅 Removed the check in a379af34f0dd7e4ead1b604a724253905826aa77 👍

Issue #4638, opened 2025-01-08
Customer center error title getting truncated w/ accessibility larger font


ronyfadel:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
Screenshot attached. note "Found" is getting truncated.
![IMG_9152](https://github.com/user-attachments/assets/48ebbcf7-e163-423f-9f1e-16507dabfceb)

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/51323) and notified the team. Thank you for reporting, we're checking this out!

facumenzella:
Just merged a fix. I'll close this now @ronyfadel ! Please let us know if you encounter any other issues 💪

PR #4647, opened 2025-01-10
fix: Use custom label for CompatibilityContentUnavailableView

facumenzella:
### Motivation
This PR tries to address https://github.com/RevenueCat/purchases-ios/issues/4638.

| Before | After |
|--------|--------|
| IMG_7180 | IMG_7179 |

### Description
For some reason, `ContentUnavailableView` truncates the label. I've tried using `lineLimit(nil)` over the `Label`, but that didn't work. This seems to do the trick. 

`fixedSize(vertical:true)` forces the view to grow vertically to the needed height.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/CompatibilityContentUnavailableView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/CompatibilityContentUnavailableView.swift
@@ -39,16 +39,38 @@ struct CompatibilityContentUnavailableView: View {
         if #available(iOS 17.0, *) {
             #if swift(>=5.9)
             if let description {
-                ContentUnavailableView(
-                    title,
-                    systemImage: systemImage,
-                    description: description
-                )
+                ContentUnavailableView {
+                    Label {
+                        Text(title)
+                            .font(.title2)
+                            .bold()
+                            .fixedSize(horizontal: false, vertical: true)
+                    } icon: {
+                        Image(systemName: systemImage)
+                            .resizable()
+                            .scaledToFill()
+                            .frame(width: 48, height: 48)
+                            .foregroundStyle(.secondary)
+                            .padding()
+                    }
+                }
+                description: { description }
             } else {
-                ContentUnavailableView(
-                    title,
-                    systemImage: systemImage
-                )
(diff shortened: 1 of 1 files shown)


RevenueCatUI/CustomerCenter/Views/CompatibilityContentUnavailableView.swift
```diff
@@ -80,6 +78,22 @@ struct CompatibilityContentUnavailableView: View {
         }
 
     }
+
```
aboedo, line 81: this is what the linter is complaining about, just extra spaces within this empty line
```suggestion

```

It's not exactly a big deal 😅 

I'd recommend setting this up in Xcode so you don't keep running into it in the future

Screenshot 2025-01-10 at 4 35 06 PM

cc @ajpallares since you might run into the same thing

aboedo: same in line 88

aboedo (approved):
Looks good to me! It does mean we have a liiitle more vertical padding now (see the screenshots) but it doesn't look wrong to my eye

aboedo:
(approved but we should fix the linting + approve the new snapshots before merging)

RevenueCatUI/CustomerCenter/Views/CompatibilityContentUnavailableView.swift
```diff
@@ -39,16 +39,22 @@ struct CompatibilityContentUnavailableView: View {
...
-                    title,
-                    systemImage: systemImage
-                )
+                ContentUnavailableView {
+                    Label {
+                        titleView
+                    } icon: {
+                        iconView
+                    }
+                }
```
ajpallares, line 57: I'm curious 🤔. Would it be possible to remove the `if` and keep only this?
```
ContentUnavailableView {
    Label {
        titleView
    } icon: {
        iconView
    }
}
description: { description }
```
I don't know if `nil` would have the same effect as `EmptyView()` (which is the default value for the `description` parameter when the parameter is not included)

ajpallares (approved):

PR #4650, opened 2025-01-11
[Paywalls V2] Updating UIConfig aliased colors to contain both light and dark

joshdholtz:
## Motivation

Aliased color will be both light and dark

## Description

This ended up being a whole thing...

TL;DR - Added a new `DisplayableColorScheme` object that has types of `hex`, `linear`, and `radial`... 

`PaywallComponent.ColorScheme -> `DisplayableColorScheme` -> `Color`/`Gradient`

### Issue

We have `.alias` as a value on `ColorInfo` but we need to alias an entire set of light/dark colors (aka `ColorScheme`).

Our logic right now directly converts `ColorInfo` into a SwiftUI/UI color (for foreground, background, etc). This works fine for `hex`, `linear`, and `radial` but will not work well for alias. We need to perform the alias lookup at the `ColorScheme` level (near the view model layer instead of at the render layer)

### Solution

Create a new `DisplayableColorScheme` and `DisplayableColorInfo` struct. These new data structures only have values for `hex`, `linear`, and `radial`. 

`ColorScheme` (and `ColorInfo`) get mapped to `DisplayableColorScheme` (and `DisplayableColorInfo`) and in this mapping is where the `alias` look up happens. This makes it safer/cleaner at the render level because everything has a specific color now.

The converting to `DisplayableColorScheme` currently will fallback to a clear color if there is an error and log an error. We will EVENTUALLY AND SOON (in a follow up PR) add this validation that the alias exists in the creation of the view models so that we can error out early if needed.

NOTE: We will also add big backend validation that the alias colors exist so the SDKs will never (🤞) have to encounter an alias color not existing.

### DIFF
--- a/RevenueCatUI/Templates/V2/ViewHelpers/DisplayableColor.swift
+++ b/RevenueCatUI/Templates/V2/ViewHelpers/DisplayableColor.swift
@@ -30,7 +30,6 @@ extension PaywallComponent.Background {
 
 }
 
-
 struct DisplayableColorScheme: Equatable, Hashable {
 
     init(light: DisplayableColorInfo, dark: DisplayableColorInfo? = nil) {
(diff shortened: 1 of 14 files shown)


RevenueCatUI/Templates/V2/ViewHelpers/DisplayableColor.swift
```diff
@@ -0,0 +1,146 @@
...
+                return .linear(degree, points)
+            case .radial(let points):
+                return .radial(points)
+
+            // Throwing error if alias has an alias
+            // This should NEVER happen though
+            case .alias(let name):
+                Logger.error("Aliased color '\(name)' has an aliased value which is not allowed.")
+                throw PaywallColorError.aliasedColorIsAliased(name)
+            }
```
JayShortway, line 126: Famous last words 😅 
This will move to the ViewModel validation/initialization too, right?

JayShortway (approved):
This makes a lot of sense! :shipit:

PR #4652, opened 2025-01-12
[Paywalls V2] Allow for app specific component overrides

joshdholtz:
### Motivation

There are occasions where we will want components to have overridden values PER app

Example:
- Different deep link schemes/urls
- Different terms/privacy policies (iOS vs Android)

### Description

Added a new optional `app` override property that will get dynamically set by the offferings response API for the app that requested it

⚠️  THE BACKEND IS NOT IMPLEMENTED YET


JayShortway (commented):
It's not entirely clear to me why the SDK needs to know about this? Why would the SDK ever receive values meant for a different app? I think handling this 100% on the backend has many benefits (saving bandwidth, seamless updates/bugfixes). Could you elaborate why we need this in the SDK? 🙏

joshdholtz:
> It's not entirely clear to me why the SDK needs to know about this? Why would the SDK ever receive values meant for a different app? I think handling this 100% on the backend has many benefits (saving bandwidth, seamless updates/bugfixes). Could you elaborate why we need this in the SDK? 🙏

@JayShortway Ahhhhh... I see what you are saying! 

So my thought was...
- Backend would give the SDK only the override it needed for its platform
- App would see that and then do its things

I thought this way because I _really_ didn't want the backend to have to go a modify/merge value _IN_ the component but... I see the way you were thinking now 🤔 

I'll go do a quick POC in the backend where this PR isn't needed at all 😇

joshdholtz:
Going to try to do this all server side

PR #4654, opened 2025-01-13
[Paywalls V2] Accept number as font size for text

joshdholtz:
### Motivation

More flexability in font size

### Description

Use `CGFloat` as `fontSize` on `TextComponent` but also back support `FontSize` for a while

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
@@ -114,7 +114,7 @@ struct TextComponentView_Previews: PreviewProvider {
                     component: .init(
                         text: "id_1",
                         color: .init(light: .hex("#000000")),
-                        fontSize: .headingXXL
+                        fontSize: 40
                     )
                 )
             )
@@ -137,7 +137,7 @@ struct TextComponentView_Previews: PreviewProvider {
                         text: "id_1",
                         fontName: "primary",
                         color: .init(light: .hex("#000000")),
-                        fontSize: .headingXXL
+                        fontSize: 40
                     )
                 )
             )
@@ -160,7 +160,7 @@ struct TextComponentView_Previews: PreviewProvider {
                         text: "id_1",
                         fontName: "This font name is not configured",
                         color: .init(light: .hex("#000000")),
-                        fontSize: .headingXXL
+                        fontSize: 40
                     )
                 )
             )
@@ -183,7 +183,7 @@ struct TextComponentView_Previews: PreviewProvider {
                         text: "id_1",
                         fontName: "primary",
                         color: .init(light: .hex("#000000")),
-                        fontSize: .headingXXL
+                        fontSize: 40
                     )
                 )
             )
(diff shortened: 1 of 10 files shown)


RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
```diff
@@ -114,7 +114,7 @@ struct TextComponentView_Previews: PreviewProvider {
                     component: .init(
                         text: "id_1",
                         color: .init(light: .hex("#000000")),
-                        fontSize: .headingXXL
+                        fontSize: 40
```
fire-at-will, line 117: Nit: It'd be nice if all of these could be hardcoded somewhere, like using the `PaywallComponent.FontSize.size` property defined further down in the PR, but since these are just previews, it's probably okay!

fire-at-will (approved):
Looks good! 🚀

Sources/Paywalls/Components/PaywallTextComponent.swift
```diff
@@ -54,6 +53,55 @@ public extension PaywallComponent {
...
+            try container.encode(color, forKey: .color)
+            try container.encode(horizontalAlignment, forKey: .horizontalAlignment)
+            try container.encodeIfPresent(backgroundColor, forKey: .backgroundColor)
+            try container.encode(size, forKey: .size)
+            try container.encode(padding, forKey: .padding)
+            try container.encode(margin, forKey: .margin)
+            try container.encodeIfPresent(overrides, forKey: .overrides)
+
+            // Encode fontSize as CGFloat
+            try container.encode(fontSize, forKey: .fontSize)
```
ajpallares, line 102: I know I'm late but (perhaps for a future PR), since this implementation of `encode(to:)` is the same the default implementation of `Encondable`, I think we can remove it. Or am I missing something?

Sources/Paywalls/Components/PaywallTextComponent.swift
```diff
@@ -37,7 +36,7 @@ public extension PaywallComponent {
             size: Size = .init(width: .fill, height: .fit),
             padding: Padding = .zero,
             margin: Padding = .zero,
-            fontSize: FontSize = .bodyM,
+            fontSize: CGFloat = 16,
```
JayShortway, line 39: @joshdholtz I noticed `.bodyM` is `15`, but you're setting the default value to `16` here. Why? Just checking as I'm doing this for Android now.

joshdholtz: Oh, this default in the constructor is _really_ just used for the SwiftUI preview so it doesn't _super_ matter 🤷‍♂️ I wasn't sure what it mapped to so I just picked something 🫠

JayShortway: Ah okay, thanks for explaining! In Android's case the default would be used when the JSON doesn't contain a `font_size`, so I'll keep it at `15` then.

joshdholtz: I think the `font_size` is technically required by the backend schema validation so hopefully you never run into that 😛

PR #4655, opened 2025-01-13
[Paywalls V2] Icon Component

joshdholtz:
### Motivation

New icon component

### Description

Screenshot 2025-01-13 at 3 33 13 PM

### DIFF
--- a/Sources/Paywalls/Components/PaywallIconComponent.swift
+++ b/Sources/Paywalls/Components/PaywallIconComponent.swift
@@ -1,8 +1,8 @@
 //
-//  PaywallImageComponent.swift
+//  PaywallIconComponent.swift
 //
 //
-//  Created by Josh Holtz on 6/12/24.
+//  Created by Josh Holtz on 1/12/24.
 //
 // swiftlint:disable missing_docs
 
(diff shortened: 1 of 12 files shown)


Sources/Paywalls/Components/PaywallIconComponent.swift
```diff
@@ -0,0 +1,132 @@
...
+
+        let type: ComponentType
+        public let baseUrl: String
+        public let iconName: String
+        public let formats: Formats
+        public let size: Size
+        public let padding: Padding?
+        public let margin: Padding?
+        public let color: ColorScheme
+        public let iconBackground: IconBackground?
```
tonidero, line 65: Hmm so the main differences with an `ImageComponent` are going to be the background + the iconName? (not sure if I missed any) I wonder if we should just reuse `ImageComponent`...

tonidero (approved):
I'm just wondering if we could unify `IconComponent` and `ImageComponent`... There are some differences so not totally sure TBH... But LGTM!

Sources/Paywalls/Components/PaywallIconComponent.swift
JayShortway: And if not for `ImageComponent`, maybe we can still get away with using `ImageComponentView` for icons?

joshdholtz: And no `overlayColor` and also no `maskShape` 🤷‍♂️ 

Could maybe combine them later but I didn't want to think that hard and rest breaking things in the state them I'm in right now 😛

PR #4658, opened 2025-01-14
[Paywalls V2] Prefetch low res images

joshdholtz:
### Motivation

Prefetch low res images of all paywalls v2 paywalls

### Description

Collects all the images in v2 paywalls for warming up cache
- Traverses the stack tree to get all of the low res images
- Collects all the images from the localizations

### DIFF
--- a/Sources/Paywalls/Components/PaywallV2CacheWarming.swift
+++ b/Sources/Paywalls/Components/PaywallV2CacheWarming.swift
@@ -1,0 +1,101 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  PaywallV2CacheWarming.swift
+//
+//  Created by Josh Holtz on 1/13/25.
+
+#if PAYWALL_COMPONENTS
+
+extension PaywallComponentsData {
+
+    var allImageURLs: [URL] {
+        var imageUrls = self.componentsConfig.base.allImageURLs
+
+        for (_, localeValues) in self.componentsLocalizations {
+            for (_, value) in localeValues {
+                switch value {
+                case .string:
+                    break
+                case .image(let image):
+                    imageUrls += image.imageUrls
+                }
+            }
+        }
+
+        return imageUrls
+    }
+
+}
+
+extension PaywallComponentsData.PaywallComponentsConfig {
+
+    var allImageURLs: [URL] {
+        let rootStackImageURLs = self.collectAllImageURLs(in: self.stack)
+        let stickFooterImageURLs = self.stickyFooter.flatMap { self.collectAllImageURLs(in: $0.stack) } ?? []
+
+        return rootStackImageURLs + stickFooterImageURLs
+    }
+
+    private func collectAllImageURLs(in stack: PaywallComponent.StackComponent) -> [URL] {
+
+        var urls: [URL] = []
+        for component in stack.components {
+            switch component {
+            case .text:
+                ()
+            case .image(let image):
+                urls += [
+                    image.source.light.heic,
+                    image.source.light.heicLowRes,
+                    image.source.dark?.heic,
+                    image.source.dark?.heicLowRes
+                ].compactMap { $0 }
+
+                if let overides = image.overrides {
+                    urls += [
+                        overides.introOffer?.source?.imageUrls ?? [],
+                        overides.states?.selected?.source?.imageUrls ?? [],
+                        overides.conditions?.compact?.source?.imageUrls ?? [],
+                        overides.conditions?.medium?.source?.imageUrls ?? [],
+                        overides.conditions?.expanded?.source?.imageUrls ?? []
+                    ].flatMap { $0 }
+                }
+            case .stack(let stack):
+                urls += self.collectAllImageURLs(in: stack)
+            case .button(let button):
+                urls += self.collectAllImageURLs(in: button.stack)
+            case .package(let package):
+                urls += self.collectAllImageURLs(in: package.stack)
+            case .purchaseButton(let purchaseButton):
+                urls += self.collectAllImageURLs(in: purchaseButton.stack)
+            case .stickyFooter(let stickyFooter):
+                urls += self.collectAllImageURLs(in: stickyFooter.stack)
+            }
+        }
+
+        return urls
+    }
+
+}
+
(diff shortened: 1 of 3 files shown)


Sources/Paywalls/PaywallCacheWarming.swift
```diff
@@ -160,6 +175,25 @@ private extension Offerings {
         )
     }
 
+    #if PAYWALL_COMPONENTS
+
+    private var allImagesInPaywallsV2: Set<URL> {
+        // Attempting to warm up all offerings for Paywalls V2.
```
JayShortway, line 181:
```suggestion
        // Attempting to warm up all images for Paywalls V2.
```

Sources/Paywalls/Components/PaywallV2CacheWarming.swift
```diff
@@ -0,0 +1,110 @@
...
+
+    var allImageURLs: [URL] {
+        let rootStackImageURLs = self.collectAllImageURLs(in: self.stack)
+        let stickFooterImageURLs = self.stickyFooter.flatMap { self.collectAllImageURLs(in: $0.stack) } ?? []
+
+        return rootStackImageURLs + stickFooterImageURLs
+    }
+
+    // swiftlint:disable:next cyclomatic_complexity
+    private func collectAllImageURLs(in stack: PaywallComponent.StackComponent) -> [URL] {
```
JayShortway, line 49: Traversing the tree of components is not the most performant, especially since we're already doing that to build the paywall. But I guess we need to do it in a separate traversal, because if we do it at the same time as building the paywall it's too late? (It wouldn't be **pre**fetch anymore ha.)

Sources/Paywalls/Components/PaywallV2CacheWarming.swift
```diff
@@ -0,0 +1,110 @@
...
+        for (_, localeValues) in self.componentsLocalizations {
+            for (_, value) in localeValues {
+                switch value {
+                case .string:
+                    break
+                case .image(let image):
+                    imageUrls += image.imageUrls
+                }
+            }
+        }
```
JayShortway, line 32: This made me think: should all images be in `componentsLocalizations`, including the "base" images?

JayShortway (approved):
Good one!

Sources/Paywalls/Components/PaywallV2CacheWarming.swift
joshdholtz: Yeaahhhh, I figured this could be improved in the future at some point if we wanted 🤷‍♂️ The thing this does _differently_ than paywalls is that it doesn't load the entire view models and very localization stuff. Its pretty specific in how/what its traversing.

But tbh... precaching a bunch of images is LESS performant than traversing this free but its leading to better performance of the paywall 😛 🥴

JayShortway: That's true, good perspective!

Sources/Paywalls/Components/PaywallV2CacheWarming.swift
joshdholtz: I had a reason against putting the base image in localization at one point and it made sense so I'm going to say... no. But now I need to remember that good argument that I had 😛 

But TBH, I did think the same thing as you while writing this code how that would be easier but... we'd have to do the same thing for icons anyway 🤷‍♂️

Sources/Paywalls/PaywallCacheWarming.swift
joshdholtz: Oh, this is actually not a typo but I could clarify...

Paywalls V1 only warmed up image for the default/current offering 😬 And I totally don't love that. But it had to be done because we didn't have:
1) Low res images
2) Explicit draft/published state

But with paywalls V2, we can load "all low res images for all offering" (so I'll change the comment to that to be more clear)

JayShortway: Ah, great! Thanks for clarifying!

nguyenhuy:
Nice attempt to prefetch images!

If you want (and have perf metrics in place to measure), you can push even further than this. For example, your backend knows which images it's about to include in the response body. So it can collect them all and put them in a seperate list at the top level of the JSON. This way, the client won't need to traverse the model tree to collect them.

Backend can even have a custom HTTP header just for this list. The client can then just grab the header and kick off the image requests. No need to wait for the body to be parsed or even fully transmitted (since headers are sent over first).

And this optimization works on both iOS and Android (and possibly web though I can't recall).

joshdholtz:
> Nice attempt to prefetch images!
> 
> If you want (and have perf metrics in place to measure), you can push even further than this. For example, your backend knows which images it's about to include in the response body. So it can collect them all and put them in a seperate list at the top level of the JSON. This way, the client won't need to traverse the model tree to collect them.
> 
> Backend can even have a custom HTTP header just for this list. The client can then just grab the header and kick off the image requests. No need to wait for the body to be parsed or even fully transmitted (since headers are sent over first).
> 
> And this optimization works on both iOS and Android (and possibly web though I can't recall).

Hey, thanks for this! 😊 

I did think of something similar a while back but definitely did not act on it 😅 But since you brought it up, it _might_ actually be something we look into! I'm not sure if we have the capacity/priority to also make the backend change for this _right now_ but... I think that's something we can add in soon ™️ and then just replace this logic.

What do you think? cc: @JayShortway @tonidero

JayShortway:
@joshdholtz @nguyenhuy I kinda love it! If we use a top-level dictionary, the image components in the tree can reference the keys from that dictionary. That way we can save a bit on bandwidth, as the JSON can already get quite big.

tonidero:
I like it! It should indeed simplify the logic in the SDKs. I'm not sure if it would save much bandwidth unless we have a bunch of repeated images (which might be the case) but still, it should simplify the parsing. I guess it might increase the complexity in the backend a bit, depending on how the images are stored? Not sure about this part... But well, I'm all for it!

nguyenhuy:
> we can control the prefetch logic a bit more on the backend.

Exactly. Backend can control which images clients should prefetch (only main ones for example) and which resolution(s).

> the image components in the tree can reference the keys from that dictionary

Yeap, that works too if network speed/consumption is a concern.

joshdholtz:
> A different but related question, I guess this will also need to support prefetching localized images if any? If we do it in the SDK, I guess that's more complexity that it needs to handle, that can be dynamically handled by the backend

Backend has support for locale filtering (we just aren't using it yet int he SDK) so it once the SDK support that, it would just really do looking at any localized images 🤷‍♂️ But yeah, doing it on backend does still decrease that complexity in the SDK by quite a bit 😇 

Will make some time to do that in the upcoming week 💪

PR #4659, opened 2025-01-14
feat: [CustomerCenter] Introduce CompatibilityLabeledContent

facumenzella:
### Motivation
This adds a simple wrapper over `LabeledContent` to work with iOS prior 16. It will be used in purchase history view

| Header |
|--------|
| image |

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/CompatibilityLabeledContent.swift
+++ b/RevenueCatUI/CustomerCenter/Views/CompatibilityLabeledContent.swift
@@ -17,7 +17,7 @@ import SwiftUI
 
 #if os(iOS)
 
-/// A SwiftUI view for displaying a message about unavailable content
+/// A SwiftUI view for displaying labeled content
 @available(iOS 15.0, *)
 @available(macOS, unavailable)
 @available(tvOS, unavailable)
--- a/RevenueCatUI/CustomerCenter/Views/CompatibilityLabeledContent.swift
+++ b/RevenueCatUI/CustomerCenter/Views/CompatibilityLabeledContent.swift
@@ -17,7 +17,49 @@ import SwiftUI
 
 #if os(iOS)
 
-/// A SwiftUI view for displaying labeled content
+/// A SwiftUI view that displays a label and content in a layout that adapts to the iOS version.
+///
+/// This view is designed to be compatible with both iOS 16.0+ and earlier versions.
+///
+/// - Parameters:
+///   - label: A closure that returns the label view, typically a `Text` or any other view.
+///   - content: A closure that returns the content view
+///
+/// ## Usage:
+/// ```swift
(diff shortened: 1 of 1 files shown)


RevenueCatUI/CustomerCenter/Views/CompatibilityLabeledContent.swift
```diff
@@ -0,0 +1,74 @@
...
+//
+//
+//  Created by Facundo Menzella on 14/1/25.
+//
+
+import SwiftUI
+
+#if os(iOS)
+
+/// A SwiftUI view for displaying a message about unavailable content
```
facumenzella, line 20:
```suggestion
/// A SwiftUI view for displaying labeled content
```

RevenueCatUI/CustomerCenter/Views/CompatibilityLabeledContent.swift
```diff
@@ -0,0 +1,74 @@
...
+
+/// A SwiftUI view for displaying labeled content
+@available(iOS 15.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+struct CompatibilityLabeledContent<Label: View, Content: View>: View {
+
+    let label: () -> Label
+    let content: (() -> Content)?
```
ajpallares, line 28: In `LabeledContent`, I don't think the `content` parameter can be `nil`. Does it have to be optional in `CompatibilityLabelContent`?

RevenueCatUI/CustomerCenter/Views/CompatibilityLabeledContent.swift
```diff
@@ -0,0 +1,74 @@
...
+#if os(iOS)
+
+/// A SwiftUI view for displaying labeled content
+@available(iOS 15.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+struct CompatibilityLabeledContent<Label: View, Content: View>: View {
+
+    let label: () -> Label
```
ajpallares, line 27: I think `label` and `content` should be marked with `@ViewBuilder`

RevenueCatUI/CustomerCenter/Views/CompatibilityLabeledContent.swift
facumenzella: Agreed. There's no point in having a labeled content without content 😅

RevenueCatUI/CustomerCenter/Views/CompatibilityLabeledContent.swift
facumenzella: I've added an init with `ViewBuilder` support 🚀

PR #4665, opened 2025-01-14
[Paywalls V2] Convert Codable structs to classes

joshdholtz:
### Motivation

We were getting random-ish `EXC_BAD_ACCESS` after the merge of the badge PR. It took some debuggin and it turned out..

When decoding 5-ish paywall v2 in an `OfferingsResponse`, the `struct` structure while in the encoder would just fail.

Note that this is in the ENCODER... and not the DECODER. So this was happening when we decode string response to Swift object and then encode back for storage.

### Description

After some debugging and experimentation with Mark, it was our conclusion that our OfferingsResponse struct with the Paywalls V2 struct were causing some sort of struct structure that was WAY too big causing some sort of buffer overflows or something.

We experiment with converting one struct to a class and it made the issue go away. So we continue this in the PR to convert all Paywalls V2 components that are codable from structs (value type) to classes (reference type) to mitigate this issue.

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Icon/IconComponentViewModel.swift
+++ b/RevenueCatUI/Templates/V2/Components/Icon/IconComponentViewModel.swift
@@ -73,18 +73,31 @@ class IconComponentViewModel {
 
 extension PresentedIconPartial: PresentedPartial {
 
-    static func combine(_ base: Self?, with other: Self?) -> Self {
+    static func combine(
+        _ base: PaywallComponent.PartialIconComponent?,
+        with other: PaywallComponent.PartialIconComponent?
+    ) -> Self {
+
+        let visible = other?.visible ?? base?.visible
+        let baseUrl = other?.baseUrl ?? base?.baseUrl
+        let iconName = other?.iconName ?? base?.iconName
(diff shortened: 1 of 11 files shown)


RevenueCatUI/Templates/V2/Components/Icon/IconComponentViewModel.swift
```diff
@@ -73,18 +73,31 @@ class IconComponentViewModel {
...
+
+        let visible = other?.visible ?? base?.visible
+        let baseUrl = other?.baseUrl ?? base?.baseUrl
+        let iconName = other?.iconName ?? base?.iconName
+        let formats = other?.formats ?? base?.formats
+        let size = other?.size ?? base?.size
+        let padding = other?.padding ?? base?.padding
+        let margin = other?.margin ?? base?.margin
+        let color = other?.color ?? base?.color
+        let iconBackground = other?.iconBackground ?? base?.iconBackground
```
joshdholtz, line 89: Had to do this because Swift compiler was timing out 🤷‍♂️

MarkVillacampa (approved):

PR #4667, opened 2025-01-15
[Paywalls V2] Use V1 default paywall when footers are used with V2 paywalls

joshdholtz:
### Motivation

We aren't planning to support footer paywalls in Paywalls V2

### Description

Uses the Paywalls V1 footer with the default paywalls when a footer is requested with a Paywalls V2 paywall

https://github.com/user-attachments/assets/affea726-41bc-4508-81f4-ea2e034a2f51

### DIFF
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -259,26 +259,39 @@ public struct PaywallView: View {
             // For fallback view or footer
             let paywall: PaywallData = .createDefault(with: offering.availablePackages,
                                                       locale: self.locale)
-            let paywallView = LoadedOfferingPaywallView(
-                offering: offering,
-                activelySubscribedProductIdentifiers: activelySubscribedProductIdentifiers,
-                paywall: paywall,
-                template: PaywallData.defaultTemplate,
-                mode: self.mode,
-                fonts: fonts,
-                displayCloseButton: self.displayCloseButton,
-                introEligibility: checker,
-                purchaseHandler: purchaseHandler,
-                locale: self.locale,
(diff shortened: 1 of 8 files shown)


RevenueCatUI/PaywallView.swift
```diff
@@ -254,21 +255,47 @@ public struct PaywallView: View {
 
         #if PAYWALL_COMPONENTS
         if let paywallComponents = offering.paywallComponents {
-            PaywallsV2View(
-                paywallComponents: paywallComponents,
+
+            // For fallback view or footer
+            let paywall: PaywallData = .createDefault(with: offering.availablePackages,
+                                                      locale: self.locale)
+            let paywallView = LoadedOfferingPaywallView(
```
tonidero, line 262: Hmm do we have to build the V1 paywallView before checking if it's actually needed? I guess we would need to validate the data before displaying it which might be quite a refactor... So I'm ok with this for now.

tonidero (approved):
Just a comment, but probably not enough time to act on it now. LGTM!

RevenueCatUI/PaywallView.swift
joshdholtz: @tonidero Just updated this if you want to have another look again! I moved the creation the view to inside `PaywallsV2View` by passing an enum with data (either data for the `LoadedOfferingPaywallView` for production or `AnyView` for tests)

tonidero: Ah nice! Yeah I think that's an improvement! I guess we do need to create the data for the default paywall before, but that's already a big improvement 👍

tonidero (approved):

PR #4669, opened 2025-01-15
fix: Set https urls for packages

facumenzella:
### Motivation
This is a follow-up of https://github.com/RevenueCat/purchases-ios/pull/4661
- The url keeps changing from ssh to https. 
- Tweak how nimble is declared as a dependency

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,6 +1,15 @@
 {
   "pins" : [
     {
+      "identity" : "accessibilitysnapshot",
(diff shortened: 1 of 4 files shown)


ajpallares (approved):
Looks good to me. However, let's have someone else take a look to make sure that they can properly resolve packages versions and doing so does not create diffs in `Package.resolved` for them

facumenzella:
@RCGitBot please test

PR #4671, opened 2025-01-15
[Trusted Entitlements] Do not clear CustomerInfo upon enabling Trusted Entitlements

tonidero:
### Description
Until now, when devs enabled Trusted Entitlements, we would clear the cached `CustomerInfo` and customers would need to refetch the latest value from the server, which could potentially cause delays and/or losing access temporarily.

With this PR, we won't be clearing the cached CustomerInfo anymore. Instead, we would return the cached customer info, which will have a `VerificationResult` of `NOT_REQUESTED`. 

iOS equivalent of https://github.com/RevenueCat/purchases-android/pull/2049

- [ ] Update docs

### DIFF
--- a/Tests/UnitTests/Identity/IdentityManagerTests.swift
+++ b/Tests/UnitTests/Identity/IdentityManagerTests.swift
@@ -116,10 +116,8 @@ class IdentityManagerTests: TestCase {
         expect(self.mockDeviceCache.invokedClearCachesForAppUserID) == false
         expect(self.mockBackend.invokedClearHTTPClientCaches) == true
         expect(self.mockBackend.invokedClearHTTPClientCachesCount) == 1
-        expect(self.mockDeviceCache.invokedClearCustomerInfoCache) == true
-        expect(self.mockDeviceCache.invokedClearCustomerInfoCacheParameters?.appUserID) == "nacho"
 
-        self.logger.verifyMessageWasLogged(Strings.identity.invalidating_cached_customer_info, level: .info)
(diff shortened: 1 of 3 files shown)


ajpallares (approved):

Tests/UnitTests/Identity/IdentityManagerTests.swift
```diff
@@ -116,10 +116,8 @@ class IdentityManagerTests: TestCase {
         expect(self.mockDeviceCache.invokedClearCachesForAppUserID) == false
         expect(self.mockBackend.invokedClearHTTPClientCaches) == true
         expect(self.mockBackend.invokedClearHTTPClientCachesCount) == 1
-        expect(self.mockDeviceCache.invokedClearCustomerInfoCache) == true
```
ajpallares, line 119: @tonidero, instead of removing this line, shouldn't we check here this? 
```
expect(self.mockDeviceCache.invokedClearCustomerInfoCache) == false
```
To make sure that the customer info cache was not cleared

tonidero: Good point! Yeah, that's better will do the change, thanks!

tonidero: Done in https://github.com/RevenueCat/purchases-ios/pull/4671/commits/496b035274038cd4eeaa524abe1f783b62dc64ee

ajpallares (approved):

PR #4672, opened 2025-01-15
[Trusted Entitlements] Enable Trusted Entitlements by default

tonidero:
### Description
This enables Trusted Entitlements by default. Behavior won't change for devs not check the verification result, but this will make it much easier to use in the future, as all SDK versions have it enabled.

This should be merged after #2049.


Tests/UnitTests/Mocks/MockSystemInfo.swift
```diff
@@ -22,6 +22,7 @@ class MockSystemInfo: SystemInfo {
                      finishTransactions: Bool,
                      customEntitlementsComputation: Bool = false,
                      storeKitVersion: StoreKitVersion = .default,
+                     responseVerificationMode: Signing.ResponseVerificationMode = .disabled,
```
tonidero, line 25: I've disabled this during tests by default... Reason being that a bunch of tests started failing because we had not mocked some things. I could fix those tests to cover that, but we already have tests specifically for having this enabled, so seemed easier to disable it by default during tests. I'm not sure if I should just take the effort to fix those tests... Thoughts?

JayShortway: Would the fix be to mock the verification behavior? Because if so I agree to keeping it `.disabled`. If we can "automatically" cover more ground by enabling this in tests, however, we could consider that. Can also be in a separate PR as far as I'm concerned.

JayShortway (approved):
LGTM!

Tests/UnitTests/Mocks/MockSystemInfo.swift
tonidero: > Would the fix be to mock the verification behavior?

That's correct. Yeah, I don't think it would be too difficult to do this mocking... But probably worth doing separately.

JayShortway: If we're mocking the verification behavior, that's equivalent to just disabling the verification altogether right? Or am I missing something? So in that case I would just leave it disabled.

tonidero: Well, it would mostly be mocking the "Signing" object, so it doesn't run the actual verification, so yeah it wouldn't run the actual signing behavior itself in most cases... We could try to not mock it and have it actually run with real data but that would be much much trickier 😅

PR #4675, opened 2025-01-15
[Paywalls V2] Move image mask after sizing

joshdholtz:
### Motivation

Image mask was being clipped

### Description

Moved image mask outside of view sizing and clipping


charliemchapman (approved):

RevenueCatUI/Templates/V2/ViewHelpers/Shape.swift
```diff
@@ -263,7 +263,7 @@ private struct ConvexShape: Shape {
 
     private var curveHeight: CGFloat {
         // Calculate the curve height as a percentage of the view's height
```
fire-at-will, line 265: What does diving the curve height in half here do?

PR #4677, opened 2025-01-16
feat: Add View extension based on CompatibilityNavigationStack

facumenzella:
### Motivation
To avoid having conditional navigation based on iOS versions everywhere, this adds a wrapper to navigate smoothly. Showing [here](https://github.com/RevenueCat/purchases-ios/pull/4662/files#diff-5b7c7f2707e8c72de42ed7acd513720b59fdaea1244f3d4dcc748ff813e6005cR86) how it is used


ajpallares (approved):
I think it's much cleaner now. Thanks @facumenzella!

PR #4680, opened 2025-01-17
Add root error info to public error

tonidero:
### Description
This is a possible approach to add root error info to public errors exposed by the SDK. Many times, we have nested errors, which makes discoverability of the root error difficult. This also hides actual StoreKit errors from being easily accessible, which some devs have asked about.

This PR adds a new property to the userInfo of the public error exposed by the SDK including relevant information about the root error. If the root error is a type of StoreKitError, we add some additional information related to that.

For example, when simulating an error getting products during the `getOfferings` call using StoreKit, a new property in userInfo `rc_root_error` will be added with the following structure:
```
{
    code = 4;
    domain = "StoreKit.StoreKitError";
    localizedDescription = "Item Unavailable";
    storeKitError =     {
        description = "not_available_in_storefront";
    };
}
```
or 
```
{
    code = 2;
    domain = "StoreKit.Product.PurchaseError";
    localizedDescription = "Unable to Complete Purchase";
    storeKitError =     {
        description = "purchase_not_allowed";
    };
}
```


tonidero:
I want to test this more and add some tests next week, but I wanted to get some feedback, specially from iOS folks, to make sure this makes sense and if there are other ideas to expose the root StoreKit error if any to developers 🙏

Sources/Error Handling/PurchasesError.swift
```diff
@@ -39,7 +40,29 @@ extension PurchasesError {
     /// let errorCode = error as? ErrorCode
     /// ```
     var asPublicError: PublicError {
```
tonidero, line 42: My main concern with this approach is whether we only expose PublicError everywhere and/or if we use this method somewhere else... Need to research it a bit more.

Sources/Error Handling/PurchasesError.swift
```diff
@@ -39,7 +40,29 @@ extension PurchasesError {
...
+            "domain": rootNSError.domain,
+            "localizedDescription": rootNSError.localizedDescription,
+        ] as [String : Any]
+        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, visionOS 1.0, *) {
+            if let storeKitErrorInfo = self.getStoreKitErrorInfoIfAny(error: rootError) {
+                rootErrorInfo = rootErrorInfo.merging(["storeKitError": storeKitErrorInfo])
+            }
+        }
+        let userInfoToUse = self.userInfo.merging([ErrorDetails.rootErrorKey: rootErrorInfo])
+        return NSError(domain: Self.errorDomain, code: self.errorCode, userInfo: userInfoToUse)
```
facumenzella, line 56: I am afraid we might lose some keys because of an unwanted overlap. I would just add a prefix like `rc_` to any new key we add to:
- be explicit on what we're adding
- avoid dropping useful data.

Sources/Error Handling/PurchasesError.swift
```diff
@@ -65,3 +88,42 @@ extension PurchasesError {
...
+                    "systemErrorDescription": systemError.localizedDescription
+                ])
+
+            @unknown default:
+                return resultMap
+            }
+        } else if let storeKitError = error as? StoreKit.Product.PurchaseError {
+            return ["description": storeKitError.trackingDescription]
+        } else {
+            return nil
```
facumenzella, line 125: I guess it would be useful to log what other error type we might be missing? 🤔

Sources/Error Handling/PurchasesError.swift
tonidero: Right, I added everything within a `rc_root_error` which I felt was enough as a namespace... I don't think it's worth adding to all nested properties then? Lmk if you think otherwise though!

Sources/Error Handling/PurchasesError.swift
tonidero: Right, in this case I'm assuming it's because the root error is not a store kit error (which may happen in several situations). I could add a log here, but I think the situation may happen relatively often... Note that this method is only to add extra data to the `userInfo` map in case it's actually a StoreKit error.

Sources/Error Handling/PurchasesError.swift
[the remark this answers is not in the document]
tonidero: Hmm so I included it in the docs for the `asPublicError`, but that's actually internal... I'm trying to figure out a better way to explain this, but maybe this is something we should add to the docs, for people that care to get the root error.

ajpallares (approved):
LGTM.
We should indeed add some documentation about the new user info keys if that's intended to be used by people, as the existing one is internal for the `asPublicError` variable.

tonidero:
Added new method to cast `PublicError` to `ErrorCode` as we discussed @ajpallares. Then, I will update the docs in the site about how to handle errors with this new API and how to obtain and process the root error.

aboedo:
what's the status for this one? it's looking good to go for me

Issue #4311, opened 2024-09-27
Cannot find CustomerCenterView importing RevenuecatUI


jesus-mg-ios:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**

We have been approved to use this alpha feature.

<img width="348" alt="image" src="https://github.com/user-attachments/assets/684554f1-be8f-4925-b968-64b4ce39b657">

<img width="889" alt="image" src="https://github.com/user-attachments/assets/be4e2f99-eaed-4ee5-abed-8237d2091e60">

1. Environment
   1. Platform:
   2. SDK version:
   3. StoreKit version:
      - [ ] StoreKit 1 (default on versions <5.0.0. Can be enabled in versions >=5.0.0 with `.with(storeKitVersion: .storeKit1)`)
      - [x] StoreKit 2 (default on versions >=5.0.0)
   4. OS version: any
   5. Xcode version: 15.2
   6. Device and/or simulator:
      - [x] Device
      - [x] Simulator
   7. Environment:
      - [x] Sandbox
      - [ ] TestFlight
      - [ ] Production
   8. How widespread is the issue. Percentage of devices affected. 100%
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue. **Complete logs with `Purchases.logLevel = .verbose` will help us debug this issue.**

```
Logs here
```

3. Steps to reproduce, with a description of expected vs. actual behavior

Download the latest revenucatui xcframework from tags. include it in your target, try to use CustomerCenterView. 
 
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

5. **Additional context**

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/46583) and notified the team. Thank you for reporting, we're checking this out!

vegaro:
@jesus-mg-ios we are just about to make a public beta that removes the directive and will solve this issues you're seeing

[...]

PR #4682, opened 2025-01-17
fix: [CustomerCenter] Introduce NavigationOptions for custom navigation

facumenzella:
### Motivation
This PR tackles https://github.com/RevenueCat/purchases-ios/issues/4311#issuecomment-2380571212, introducing `CustomerCenterNavigationOptions` to customize how navigation is handle within the `CustomerCenter`.

### Description
Instead of three separate variables I went for a struct, winning in future flexibility. New variables can be added without breaking the API.

It adds some view modifiers to simplify the use of it internally as well.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/CustomerCenterNavigationLink.swift
+++ b/RevenueCatUI/CustomerCenter/Views/CustomerCenterNavigationLink.swift
@@ -12,11 +12,10 @@ import SwiftUI
 
 /// A view that provides a navigation link to `CustomerCenterView` with a customizable label.
 ///
-/// This is the **preferred way** to integrate `CustomerCenterView` into your `NavigationStack`,
-/// ensuring proper navigation behavior by setting `isEmbededInNavigation` to `true`.
+/// This is the **preferred way** to integrate `CustomerCenterView` into your `NavigationView`,
+/// ensuring proper navigation behavior by pre-setting navigation options.
 ///
-///
-/// Example:
+/// ## Example Usage
 /// ```swift
 /// CustomerCenterNavigationLink {
 ///     HStack {
@@ -63,7 +62,11 @@ public struct CustomerCenterNavigationLink<Label: View>: View {
         NavigationLink {
             CustomerCenterView(
                 customerCenterActionHandler: customerCenterActionHandler,
-                isEmbeddedInNavigationStack: true)
+                navigationOptions: CustomerCenterNavigationOptions(
+                    usesNavigationStack: false,
+                    usesExistingNavigation: true,
+                    shouldShowCloseButton: false
+                ))
         } label: {
             label()
         }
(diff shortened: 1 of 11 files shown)


RevenueCatUI/CustomerCenter/Views/CustomerCenterNavigationLink.swift
```diff
@@ -0,0 +1,76 @@
...
+import SwiftUI
+
+#if os(iOS)
+
+/// A view that provides a navigation link to `CustomerCenterView` with a customizable label.
+///
+/// This is the **preferred way** to integrate `CustomerCenterView` into your `NavigationView`,
+/// ensuring proper navigation behavior by pre-setting navigation options.
+///
+/// ## Example Usage
```
aboedo, line 18: love the example

RevenueCatUI/CustomerCenter/Views/CustomerCenterNavigationLink.swift
```diff
@@ -0,0 +1,76 @@
...
+//
+
+import RevenueCat
+import SwiftUI
+
+#if os(iOS)
+
+/// A view that provides a navigation link to `CustomerCenterView` with a customizable label.
+///
+/// This is the **preferred way** to integrate `CustomerCenterView` into your `NavigationView`,
```
aboedo, line 15: nitpick: double-backticks so that these automatically turn into links:
```suggestion
/// A view that provides a navigation link to ``CustomerCenterView`` with a customizable label.
///
/// This is the **preferred way** to integrate ``CustomerCenterView`` into your `NavigationView`,
```

aboedo (commented):
looking great so far

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -43,29 +43,46 @@ public struct CustomerCenterView: View {
 
     private let mode: CustomerCenterPresentationMode
```
facumenzella, line 44: Ideally, I would love to deprecate this and move it inside `NavigationOptions` somehow, but that's definitely out of scope

ajpallares (changes requested):

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -82,7 +99,8 @@ public struct CustomerCenterView: View {
                         .environment(\.localization, configuration.localization)
                         .environment(\.appearance, configuration.appearance)
                         .environment(\.supportInformation, configuration.support)
-                        .environment(\.customerCenterPresentationMode, self.mode)
+//                        .environment(\.customerCenterPresentationMode, self.mode)
```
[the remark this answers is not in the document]
facumenzella: You're right. I just commented to see if I was able to replace it with the new struct, but I think we can leave that for further iterations.

ajpallares (approved):
LGTM!

PR #4683, opened 2025-01-17
v4: Fix main actor references in iOS 11-12

aboedo:
Having references to `@MainActor` crashes the SDK in runtime in iOS 11 and 12 in our SDK v4 when compiling with Xcode 16 (it's not an issue in Xcode 15). 

We've reported this to Apple as `FB16345033`. 

To solve this, we need to remove all references to `@MainActor` for iOS 11 and 12. Conceptually this makes sense - that should never have compiled in the first place, since iOS 11 and 12 do not support Swift Concurrency. 

But this is a fairly delicate open heart surgery, so we need to test it out thoroughly.

### DIFF
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -17,7 +17,7 @@ import Foundation
 
 class CustomerInfoManager {
 
-    // todo: remove main actor here
+    // todo: confirm removal of mainActor here
     typealias CustomerInfoCompletion = @Sendable (Result<CustomerInfo, BackendError>) -> Void
 
     private let offlineEntitlementsManager: OfflineEntitlementsManager
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -17,6 +17,7 @@ import Foundation
 
 class CustomerInfoManager {
 
+    // todo: remove main actor here
     typealias CustomerInfoCompletion = @MainActor @Sendable (Result<CustomerInfo, BackendError>) -> Void
 
     private let offlineEntitlementsManager: OfflineEntitlementsManager
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -18,7 +18,7 @@ import Foundation
 class CustomerInfoManager {
 
     // todo: remove main actor here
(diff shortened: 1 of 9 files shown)


Sources/Identity/CustomerInfoManager.swift
```diff
@@ -89,7 +90,7 @@ class CustomerInfoManager {
             }
 
             if let completion = completion {
-                self.operationDispatcher.dispatchOnMainActor {
+                self.operationDispatcher.dispatchAsyncOnMainThread {
```
aboedo, line 93: note: in this file I'm using `dispatchAsyncOnMainThread` and not `dispatchOnMainThread`. Not sure which one is best, haven't had time to dig in. But there is comment in line 307 where we're using `dispatchAsyncOnMainThread` saying that it **must** be async to prevent deadlocks, so since I didn't have time to dig in deeper I just went with that for the whole file

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -33,10 +33,11 @@ public typealias PurchaseResultData = (transaction: StoreTransaction?,
...
  */
-public typealias PurchaseCompletedBlock = @MainActor @Sendable (StoreTransaction?,
-                                                                CustomerInfo?,
-                                                                PublicError?,
-                                                                Bool) -> Void
+// todo: main actor here
+public typealias PurchaseCompletedBlock = @Sendable (StoreTransaction?,
+                                                     CustomerInfo?,
+                                                     PublicError?,
+                                                     Bool) -> Void
```
aboedo, line 40: this is the most concerning one, where we are effectively doing a public-facing change. But it's one that removes restrictions, rather than adding. I expect maybe it adds some Sendable warnings?
Then again... How in the world did we add this in a minor in the first place?

manuelescrig:
I'm using V4 of the SDK on iOS 12. I'm having this issue when compiling it with Xcode 16. Any idea if this PR will go trough at some point? Thanks! 😊

aboedo:
closing in favor of https://github.com/RevenueCat/purchases-ios/pull/4718, which does the same thing but more conservatively, introducing less changes that only target the issues when storing collections of `@MainActor` objects

aboedo:
@manuelescrig sorry I just saw your comment! We're working on testing the alternative PR and should get a fix out soon!

manuelescrig:
Thank you so much @aboedo!

PR #4687, opened 2025-01-20
fix: Do not lint deleted files

facumenzella:
### Motivation
While working on something else I noticed that the `pre-commit` hook is broken, and tries to lint deleted files. This fixes that by checking if the file was deleted

```bash
      status=$(git status --porcelain "$file" | cut -c1-2)
```

It also adds another validation checking if the variables exist before using them, to prevent another error:

```bash
.git/hooks/pre-commit: line 52: [: =: unary operator expected
.git/hooks/pre-commit: line 52: [: =: unary operator expected
.git/hooks/pre-commit: line 52: [: =: unary operator expected
.git/hooks/pre-commit: line 52: [: =: unary operator expected
.git/hooks/pre-commit: line 52: [: =: unary operator expected
``

### DIFF
--- a/scripts/pre-commit.sh
+++ b/scripts/pre-commit.sh
@@ -49,7 +49,7 @@ verify_no_included_apikeys() {
     for api_file in "${FILES_TO_CHECK[@]}"
     do
       absolute_api_file=$(realpath "$api_file")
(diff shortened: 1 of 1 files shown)


JayShortway (approved):
THANK YOU! 🙏

PR #4693, opened 2025-01-21
Add UI Preview Mode to SDK

ajpallares:
### Checklist

### Motivation
We need to add a specific mode for the SDK to allow Paywall Previews in the RevenueCat app

### Description
The UI Preview Mode of the SDK should limit all the functionality of the SDK to only enable implementing the Paywall Previews functionality in the RevenueCat, so that developers can properly test their paywalls on their own devices.

This is currently a WIP, as it still only contains the changes implemented by @aboedo.


Sources/Purchasing/OfferingsManager.swift
```diff
@@ -275,6 +276,58 @@ private extension OfferingsManager {
...
+            (type: "monthly", price: 5.99, period: SubscriptionPeriod(value: 1, unit: .month)),
+            (type: "yearly", price: 59.99, period: SubscriptionPeriod(value: 1, unit: .year)),
+            (type: "lifetime", price: 199.99, period: nil)
+        ]
+
+        let products = productIdentifiers.enumerated().map { index, identifier -> StoreProduct in
+            let productType = productTypes[index % productTypes.count]
+
+            let introductoryDiscount: TestStoreProductDiscount? = {
+                guard productType.period != nil && Bool.random() else { return nil }
```
JayShortway, line 300: I like the idea, but the randomness could add friction if you want to see a discount and all `random()`s happen to be false. (You'll need to go back and reopen the preview / the app.) 

What do you think about having a fixed (set of) product type(s) always having a discount, e.g. `monthly` and `yearly`?

Sources/Misc/DangerousSettings.swift
```diff
@@ -101,13 +107,24 @@ import Foundation
...
-    /// Designated initializer
+    /**
+     * Only use a Dangerous Setting if suggested by RevenueCat support team.
+     *
+     * - Parameter uiPreviewMode: if `true`, the SDK will return a set of mock products instead of the
+     * products obtained from StoreKit. This is useful for testing or preview purposes. 
+     */
+    public convenience init(uiPreviewMode: Bool) {
+        self.init(autoSyncPurchases: false, internalSettings: Internal.default, uiPreviewMode: uiPreviewMode)
+    }
```
JayShortway, line 118: Just a suggestion, but maybe we don't need a `convenience init` for something that we don't want people to use, as it's for "internal" use only?

JayShortway (approved):
Very clean and readable! 😗👌 Just 2 suggestions.

Sources/Misc/DangerousSettings.swift
ajpallares: I totally agree. The only point of having a `convenience init` is to make integration more convenience for integrators. But we don't need to make it convenient for ourselves since we know how to use it and we don't even want integrators to use it. So I'll remove it. Thanks!

Sources/Purchasing/OfferingsManager.swift
ajpallares: I totally see your point about the friction due to the randomness. I also believe that what we show in previews should be more deterministic.
On the other hand, I do imagine some clients wanting to show the exact information of their products on the previews. However, the introduced randomness does not solve this problem either.
WDYT @aboedo?

JayShortway: >I do imagine some clients wanting to show the exact information of their products on the previews.  
  
Just for some context: this is something I believe we will do at some point. We could on top of that allow users to configure the preview before/when showing it, e.g. with a setting that controls discounts. However these are not solutions for in this PR haha. 

This is not a blocker for me btw!

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -275,6 +276,58 @@ private extension OfferingsManager {
...
+    private func createMockProducts(productIdentifiers: Set<String>) -> Set<StoreProduct> {
+        let productTypes = [
+            (type: "weekly", price: 1.99, period: SubscriptionPeriod(value: 1, unit: .week)),
+            (type: "monthly", price: 5.99, period: SubscriptionPeriod(value: 1, unit: .month)),
+            (type: "yearly", price: 59.99, period: SubscriptionPeriod(value: 1, unit: .year)),
+            (type: "lifetime", price: 199.99, period: nil)
+        ]
+
+        let products = productIdentifiers.enumerated().map { index, identifier -> StoreProduct in
+            let productType = productTypes[index % productTypes.count]
```
tonidero, line 297: I wonder if we could act a bit smarter here and instead of assigning basically a random period to each product, we could try to get the [package type](https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/PackageType.swift#L23) provided by the offerings (which has the duration). In case the package type is `custom`, we could then do something "smart" like analyzing the product id for some keywords like `week`, `month`, `year`.... To assign the duration. And then if we don't find any, as a fallback, assign a random period.

tonidero (commented):
Looking good! Just another comment of possible improvements to make the previews more realistic/more deterministic

Sources/Purchasing/OfferingsManager.swift
aboedo: Makes sense, I can imagine the frustration. And the confusion if the paywall looks different between opens. 

Agreed with Joop that eventually we'll want to get realistic data in there, but in the meantime probably the easiest way to go would be to have a way to _tell_ the SDK what to show. 

That way we could do something like what the dashboard does, where it allows you to select whether to preview with intro offer eligible or not. 

Probably not something that needs to go into the MVP of this feature and can be added on later before shipping to prod. 

And it will set us up nicely for doing something similar for Customer Center, where we'll inevitably need to find a way to select whether to show the CC as a user with an active subscription would see it vs empty subs

ajpallares:
I'll open separate PRs for the new API for enabling the UI Preview mode and for the specific implementation details of this new mode

ajpallares:
See #4714 for the new API for the UI Preview mode

PR #4696, opened 2025-01-22
[Paywalls] Badge fixes

MarkVillacampa:
A few badge-related changes
- Position the Overlaid badge in the middle if the border instead of its outer edge.
- Position Nested badges inside the border of the stack by default.
- Clip trailing/leading edge-to-edge badges, as well as nested badges, to prevent them form rendering outside the bounds of the stack.
	- This requited us to extract the Shadow from the Shape modifier so they can be applied separatedly. This allows us to apply the badge modifier in the middle, and let it clip the stack before applying the Shadow.

### Checklist


RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
```diff
@@ -519,21 +519,32 @@ extension StackComponentViewModel {
...
+                localizationProvider: localizationProvider,
+                uiConfigProvider: uiConfigProvider
             )
         }
 
         try self.init(
             component: component,
             viewModels: viewModels,
-            badgeViewModels: [],
+            badgeViewModels: badgeViewModels ?? [],
```
MarkVillacampa, line 547: This is a convenience initialiser used only in previews. I ported the Badge previews to use real StackComponentView so I needed to actually parse the badge modifiers.

RevenueCatUI/Templates/V2/ViewHelpers/BadgeModifier.swift
```diff
@@ -118,6 +125,7 @@ fileprivate extension View {
                 VStack(alignment: .leading) {
                     VStack {
                         ComponentsView(componentViewModels: badge.badgeViewModels, onDismiss: {})
+                            .frame(maxWidth: .infinity)
```
MarkVillacampa, line 128: Something changed somewhere and the top/bottom edge-to-edge badges were not filling the whole width, this fixes that.

RevenueCatUI/Templates/V2/ViewHelpers/ShadowModifier.swift
```diff
@@ -38,58 +38,44 @@ struct ShadowModifier: ViewModifier {
...
-            spread: 0
+        self.modifier(ShadowModifier(
+            shadow: shadow,
+            shape: shape
         ))
     }
 }
 
-@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
-struct LayerShadowModifier: ViewModifier {
```
MarkVillacampa, line 67: The diff is a bit misleading here. `LayerShadowModifier` wasn't deleted, it was renamed to `ShadowModifier` and the original `ShadowModifier` which wasn't being used anymore was deleted.

RevenueCatUI/Templates/V2/ViewHelpers/BadgeModifier.swift
```diff
@@ -58,14 +59,17 @@ fileprivate extension View {
...
                             .backgroundStyle(badge.backgroundStyle)
                             .shape(border: nil, shape: effectiveShape(badge: badge))
                     }
                     .fixedSize()
                     .padding(effectiveMargin(badge: badge).edgeInsets)
                     .alignmentGuide(
                         effetiveVerticalAlinmentForOverlaidBadge(alignment: badge.alignment.stackAlignment),
-                        computeValue: { dim in dim[VerticalAlignment.center] })
+                        computeValue: { dim in
+                            dim[VerticalAlignment.center] + effetiveYTranslationForOverlaidBadge(badge: badge)
```
MarkVillacampa, line 71: This is the magic to position the overlaid badge in the middle of the stack's border

tonidero (approved):
I believe this looks good! the implementation differs a bit with Android, since there, we actually change the layouts so all the clipping is performed automatically... But the math here seems ok to me!

PR #4698, opened 2025-01-22
[Paywalls] Fix hex colors with alpha component

MarkVillacampa:
### Checklist


RevenueCatUI/Templates/V2/ViewModelHelpers/PaywallComponentTypeTransformers.swift
```diff
@@ -290,7 +290,7 @@ extension PaywallComponent.ColorHex {
 
         if scanner.scanHexInt64(&hexNumber) {
             // If Alpha channel is missing, it's a fully opaque color.
-            if hexNumber <= 0xffffff {
+            if hexColor.count == 6 {
```
MarkVillacampa, line 293: The condition assumed that if the hexNumber is less than or equal to `0xffffff` (indicating a 6-character hex string), it needs to append 0xff as the alpha channel. However the hex color `00ff0000` has an alpha component and it would still fall under this condition.

We can fix it by checking the length of the hex color instead.

marek-dabek (approved):

PR #4703, opened 2025-01-24
chore: Tweak default copy for dateWhenAppWasPurchased

facumenzella:
### Motivation
Tweak the default translation for `dateWhenAppWasPurchased`

### DIFF
--- a/Sources/CustomerCenter/CustomerCenterConfigData.swift
+++ b/Sources/CustomerCenter/CustomerCenterConfigData.swift
@@ -250,7 +250,7 @@ public struct CustomerCenterConfigData {
                 case .accountDetails:
                     return "Account Details"
                 case .dateWhenAppWasPurchased:
-                    return "Date when app was first purchased"
+                    return "Original Purchase Date"
                 case .userId:
                     return "User ID"
                 case .purchaseHistory:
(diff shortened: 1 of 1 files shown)


Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -250,7 +250,7 @@ public struct CustomerCenterConfigData {
                 case .accountDetails:
                     return "Account Details"
                 case .dateWhenAppWasPurchased:
-                    return "Date when app was first purchased"
+                    return "Original Purchase Date"
```
aboedo, line 253: can we make it "date when app was first downloaded"? I feel like this still doesn't make it less confusing, especially since most apps are free

facumenzella: This was actually a typo 🙃 
The copy in the backend is `"Original Download Date"`

facumenzella: Since this is in the backend already, I think we can do a quick round and revisit some of them; but I would merge it like this for consistency.

aboedo (approved):

PR #4705, opened 2025-01-25
[Paywalls V2] New overrides structure

joshdholtz:
## Motivation

Migrating `overrides` from a structured object to an array of override objects

### Previously

Overrides were a very structure object that had all of the different types of states/conditions that allowed base properties of a component to be overridden. However, this was not flexible when it came to multiple being activated as it was missing for cases when multiples types of conditions were being evaluated.

### Now

Overrides is an array that has an array of `conditions` and `properties`. It will get applied in order from first to last IF all of the `conditions` are true.

This allows the FE/BE to determine the orders of overrides from lowest to higher priority so all of the SDKs/renderers don't need to duplicate that logic.

### DIFF
--- a/RevenueCatUI/Templates/V2/ViewModelHelpers/PresentedPartials.swift
+++ b/RevenueCatUI/Templates/V2/ViewModelHelpers/PresentedPartials.swift
@@ -29,35 +29,14 @@ protocol PresentedPartial {
 
 }
 
-/// Structure holding override configurations for different presentation states
-struct PresentedOverrides<T: PresentedPartial> {
+/// Array holding override configurations for different presentation states
+typealias PresentedOverrides<T: PresentedPartial> = [PresentedOverride<T>]
 
-    /// Override for intro offer state
-    public let introOffer: T?
-    /// Override for different selection states
-    public let states: PresentedStates<T>?
-    /// Override for different screen size conditions
-    public let conditions: PresentedConditions<T>?
+/// Structure holding override configurations for a presentation state
+struct PresentedOverride<T: PresentedPartial> {
 
-}
-
-/// Structure defining states for selected/unselected components
-struct PresentedStates<T: PresentedPartial> {
-
-    /// Override for selected state
-    let selected: T?
-
-}
-
-/// Structure defining overrides for different screen size conditions
-struct PresentedConditions<T: PresentedPartial> {
-
-    /// Override for compact size
-    let compact: T?
-    /// Override for medium size
-    let medium: T?
-    /// Override for expanded size
-    let expanded: T?
+    let conditions: [PaywallComponent.Condition]
+    let properties: T?
 
 }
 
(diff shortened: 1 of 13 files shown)


RevenueCatUI/Templates/V2/ViewModelHelpers/PresentedPartials.swift
```diff
@@ -75,52 +54,58 @@ extension PresentedPartial {
...
-        case .selected:
-            conditionPartial = Self.combine(conditionPartial, with: presentedOverrides?.states?.selected)
+        var presentedPartial: Self?
+
+        for presentedOverride in presentedOverrides where self.shouldApply(
+            for: presentedOverride.conditions,
+            state: state,
+            activeCondition: condition,
+            isEligibleForIntroOffer: isEligibleForIntroOffer) {
+            presentedPartial = Self.combine(presentedPartial, with: presentedOverride.properties)
```
joshdholtz, line 68: This is where the overrides are iterated from lowest to higher priority (in the order they are given) and will only combine them if all the conditions are true

RevenueCatUI/Templates/V2/ViewModelHelpers/PresentedPartials.swift
```diff
@@ -75,52 +54,58 @@ extension PresentedPartial {
...
+            case .introOffer:
+                if !isEligibleForIntroOffer {
+                    return false
                 }
+            case .selected:
+                if state != .selected {
+                    return false
+                }
+            case .unsupported:
+                return false
```
joshdholtz, line 96: Logic for determining if all the conditions are true

Sources/Paywalls/Components/Common/ComponentOverrides.swift
```diff
@@ -21,48 +21,90 @@ public extension PaywallComponent {
...
+    enum Condition: String, Codable, Sendable, Hashable, Equatable {
+
+        case compact
+        case medium
+        case expanded
+        case introOffer = "intro_offer"
+        case selected
+
+        // For unknown cases
+        case unsupported
```
joshdholtz, line 47: I added a case for `unsupported` so that the decoding wouldn't fail for `multiple_intro_offers` or if any other new ones where added

RevenueCatUI/Templates/V2/ViewModelHelpers/PresentedPartials.swift
JayShortway: Clean!

RevenueCatUI/Templates/V2/ViewModelHelpers/PresentedPartials.swift
```diff
@@ -75,52 +54,58 @@ extension PresentedPartial {
...
-        return applicableConditions.reduce(nil) { partial, next in
-            Self.combine(partial, with: next)
         }
+
+        return true
     }
 
 }
 
 private extension ScreenCondition {
```
JayShortway, line 105: I'm not suggesting or requesting any changes, especially not in this PR, but do we still need `ScreenCondition`? Can we not make do with just `PaywallComponent.Condition`?

Sources/Paywalls/Components/Common/ComponentOverrides.swift
JayShortway: Good idea!

JayShortway (approved):
Very nice! 🙌

RevenueCatUI/Templates/V2/ViewModelHelpers/PresentedPartials.swift
joshdholtz: Ohhhhhh, this is a good point! I think we can totally merge the two in a follow up PR 👍

PR #4711, opened 2025-01-27
fix: Show close button in ErrorView

facumenzella:
### Motivation
I noticed the error view was not showing the close button. The navigation is not applied to all views, so there's nowhere to place the button.

This wraps all the views to the navigation stack if not provided

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
@@ -107,27 +107,30 @@ public struct CustomerCenterView: View {
         }
     }
 
-    // swiftlint:disable:next missing_docs
-    public var body: some View {
-        Group {
-            if navigationOptions.usesExistingNavigation {
+    @ViewBuilder
+    var navigationContent: some View {
+        if navigationOptions.usesExistingNavigation {
+            content
+        } else {
+            CompatibilityNavigationStack {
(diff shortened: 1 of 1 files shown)


RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -85,37 +85,52 @@ public struct CustomerCenterView: View {
...
 
-    // swiftlint:disable:next missing_docs
-    public var body: some View {
-        Group {
-            switch self.viewModel.state {
-            case .error:
-                ErrorView()
-            case .notLoaded:
+    @ViewBuilder
+    var content: some View {
```
ajpallares, line 89: This could be `private`

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -85,37 +85,52 @@ public struct CustomerCenterView: View {
...
-                    TintedProgressView()
-                }
             }
         }
-        .task {
-            await loadInformationIfNeeded()
+    }
+
+    @ViewBuilder
+    var navigationContent: some View {
```
ajpallares, line 111: This could also be `private`

ajpallares (approved):
LGTM 👍 
Just a couple of small suggestions

PR #4713, opened 2025-01-27
[Paywalls] Add Timeline component

MarkVillacampa:
### Checklist

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Icon/IconComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Icon/IconComponentView.swift
@@ -56,9 +56,8 @@ struct IconComponentView: View {
                    shadow: style.iconBackgroundShadow,
                    background: style.iconBackgroundStyle,
                    uiConfigProvider: self.viewModel.uiConfigProvider)
-            .padding(style.padding)
+            .padding(style.margin)
             .size(style.size)
-            .clipped()
         }
     }
 
@@ -153,7 +152,7 @@ struct IconComponentView_Previews: PreviewProvider {
                         ),
                         size: .init(width: .fixed(150), height: .fixed(150)),
                         padding: .init(top: 20, bottom: 20, leading: 20, trailing: 20),
-                        margin: .zero,
+                        margin: .init(top: 20, bottom: 20, leading: 20, trailing: 20),
                         color: PaywallComponent.ColorScheme(
                             light: .hex("#ff0000")
                         ),
--- a/RevenueCatUI/Templates/V2/Components/Timeline/TimelineComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Timeline/TimelineComponentView.swift
@@ -1,0 +1,331 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  TimelineComponentView.swift
+//
+//  Created by Mark Villacampa on 15/1/25.
+
+import Foundation
+import RevenueCat
+import SwiftUI
+
+#if PAYWALL_COMPONENTS
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+struct TimelineComponentView: View {
+    private let viewModel: TimelineComponentViewModel
+
+    @Environment(\.colorScheme)
+    private var colorScheme
+
+    internal init(viewModel: TimelineComponentViewModel) {
+        self.viewModel = viewModel
+    }
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: viewModel.component.itemSpacing ?? 0) {
+            ForEach(viewModel.items, id: \.component) { item in
+                timelineRow(item: item)
+            }
+        }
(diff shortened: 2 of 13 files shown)


RevenueCatUI/Templates/V2/Components/Icon/IconComponentView.swift
```diff
@@ -56,9 +56,8 @@ struct IconComponentView: View {
                    shadow: style.iconBackgroundShadow,
                    background: style.iconBackgroundStyle,
                    uiConfigProvider: self.viewModel.uiConfigProvider)
-            .padding(style.padding)
+            .padding(style.margin)
             .size(style.size)
-            .clipped()
```
MarkVillacampa, line 61: I can't see a reason why the icon needs to be clipped. This is likely to clip the shadow if present, which is not intended.

RevenueCatUI/Templates/V2/Previews/PreviewMock.swift
```diff
@@ -78,7 +78,6 @@ struct PreviewRequiredEnvironmentProperties: ViewModifier {
     let screenCondition: ScreenCondition
     let componentViewState: ComponentViewState
     let packageContext: PackageContext?
-    let colorScheme: ColorScheme
```
MarkVillacampa, line 81: `colorScheme` is always present in the environment as it's a default SwiftUI property. Hardcoding it in `previewRequiredEnvironmentProperties` was making it impossible to switch the color scheme in Xcode previews.

RevenueCatUI/Templates/V2/ViewHelpers/Fill.swift
```diff
@@ -0,0 +1,58 @@
...
+
+import RevenueCat
+import SwiftUI
+
+#if PAYWALL_COMPONENTS
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+extension Shape {
+
+    func fillColorScheme(
```
MarkVillacampa, line 22: We already have helpers to apply a foreground and background color scheme but I needed to `fill` a shape. 

At some point we need to refactor and unify the logic of all these methods.

RevenueCatUI/Templates/V2/Components/Icon/IconComponentView.swift
joshdholtz: Copy/paste from image 🫠

joshdholtz (changes requested):
A few minor comments but this is looking gooooood 🤩

Sources/Paywalls/Components/PaywallTimelineComponent.swift
```diff
@@ -0,0 +1,85 @@
...
+
+// swiftlint:disable missing_docs nesting
+public extension PaywallComponent {
+
+    struct TimelineComponent: PaywallComponentBase, Equatable, Hashable {
+        let type: ComponentType
+        public let iconAlignment: IconAlignment?
+        public let itemSpacing: CGFloat?
+        public let textSpacing: CGFloat?
+        public let columnGutter: CGFloat?
```
tonidero, line 26: Oh interesting... In android I was considering all these mandatory... I guess we could default to 0 in the SDK, but seems like something the backend should always provide? Wdyt?

RevenueCatUI/Templates/V2/Components/Timeline/TimelineComponentView.swift
```diff
@@ -0,0 +1,331 @@
...
+                        // Connect two items from center to center respecting margins
+                        if let connector = item.component.connector {
+                            let color = connector.color.asDisplayable(uiConfigProvider: viewModel.uiConfigProvider)
+                            Rectangle()
+                                .fillColorScheme(color, colorScheme: colorScheme)
+                                .frame(
+                                    width: connector.width,
+                                    height: proxy[to][.center].y - proxy[from][.center].y -
+                                    (item.component.connector?.margin.bottom ?? 0) -
+                                    (item.component.connector?.margin.top ?? 0)
```
tonidero, line 55: Hmm right... I think I will need to do some of this calculations in Android as well in cases where one item is bigger than the others 👍

tonidero (commented):
Nice! This lgtm! Just the existing comments

Sources/Paywalls/Components/PaywallTimelineComponent.swift
MarkVillacampa: I just made them optional out of abundance of caution 😅 I'd expect the backend to send a default value always yes.

Because of how Decodable works in Swift, I cannot specify a default value, I must define them as optional tho.

RevenueCatUI/Templates/V2/ViewModelHelpers/ViewModelFactory.swift
```diff
@@ -181,6 +182,39 @@ struct ViewModelFactory {
...
+        case .timeline(let component):
+            let models = try component.items.map { item in
+                var description: TextComponentViewModel?
+                if let descriptionComponent = item.description {
+                    description = try TextComponentViewModel(
+                        localizationProvider: localizationProvider,
+                        uiConfigProvider: uiConfigProvider,
+                        component: descriptionComponent
+                    )
+                }
```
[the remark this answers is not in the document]
MarkVillacampa: Not sure I follow what you mean here 🤔

joshdholtz (approved):
:shipit: :shipit: :shipit: :shipit: :shipit:

PR #4714, opened 2025-01-27
Add `uiPreviewMode` to `DangerousSettings`

ajpallares:
### Checklist

### Motivation
We need a way to activate the UI Preview mode in the SDK to enable the Paywall Previews functionality in the RC app without actually adding any new public APIs to the SDK

### Description
This PR adds a new property to `DangerousSettings`: `uiPreviewMode`. 
* This new public API is opt-in only, thanks to the [`@_spi` attribute](https://swiftpackageindex.com/swiftlang/swift-syntax/600.0.1/documentation/swiftsyntax/spi).
* The UI Preview Mode of the SDK will limit all the functionality of the SDK to only enable implementing the Paywall Previews functionality (and similar) in the RevenueCat app, so that developers can properly test their paywalls on their own devices.

### DIFF
--- a/Sources/Misc/DangerousSettings.swift
+++ b/Sources/Misc/DangerousSettings.swift
@@ -59,6 +59,12 @@ import Foundation
     @objc public let autoSyncPurchases: Bool
 
     /**
+     * if `true`, the SDK will return a set of mock products instead of the
+     * products obtained from StoreKit. This is useful for testing or preview purposes. 
+     */
+    @_spi(Internal) public let uiPreviewMode: Bool
+
+    /**
      * A property meant for apps that do their own entitlements computation, separated from RevenueCat.
      * It:
      *   - disables automatic CustomerInfo cache updates
@@ -101,13 +107,24 @@ import Foundation
 
     }
 
-    /// Designated initializer
+    /**
+     * Only use a Dangerous Setting if suggested by RevenueCat support team.
+     *
+     * - Parameter uiPreviewMode: if `true`, the SDK will return a set of mock products instead of the
+     * products obtained from StoreKit. This is useful for testing or preview purposes. 
+     */
+    @_spi(Internal) public init(uiPreviewMode: Bool) {
+        self.init(autoSyncPurchases: false, internalSettings: Internal.default, uiPreviewMode: uiPreviewMode)
+    }
+
     internal init(autoSyncPurchases: Bool,
                   customEntitlementComputation: Bool = false,
-                  internalSettings: InternalDangerousSettingsType) {
+                  internalSettings: InternalDangerousSettingsType,
+                  uiPreviewMode: Bool = false) {
         self.autoSyncPurchases = autoSyncPurchases
         self.internalSettings = internalSettings
         self.customEntitlementComputation = customEntitlementComputation
+        self.uiPreviewMode = uiPreviewMode
     }
 
 }
--- a/Sources/Misc/DangerousSettings.swift
+++ b/Sources/Misc/DangerousSettings.swift
@@ -108,7 +108,7 @@ import Foundation
     }
 
     /**
-     * Only use a Dangerous Setting if suggested by RevenueCat support team.
+     * Used to initialize the SDK in UI preview mode.
      *
      * - Parameter uiPreviewMode: if `true`, the SDK will return a set of mock products instead of the
      * products obtained from StoreKit. This is useful for testing or preview purposes. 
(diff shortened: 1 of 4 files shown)


ajpallares:
I have added the `uiPreviewMode` as a `DangerousSettings`:
1. This should make it double clear that it's not for public use (`DangerousSettings` + `@_spi` attribute). In fact, the `@_spi` attribute does prevent the use of this API for regular imports of the SDK.
2. In a future major release, we can move everything that currently lives under `DangerousSettings`, to use the `@_spi` attribute. So we can use this `uiPreviewMode` as a first step towards that goal.

Of course, I'm open to suggestions about adding the `uiPreviewMode` to a better-fitting place.

aboedo:
I triggered a re-run of the iOS 18 tests since the failures seemed unrelated

Sources/Misc/DangerousSettings.swift
```diff
@@ -58,6 +58,12 @@ import Foundation
      */
     @objc public let autoSyncPurchases: Bool
 
+    /**
+     * if `true`, the SDK will return a set of mock products instead of the
+     * products obtained from StoreKit. This is useful for testing or preview purposes. 
+     */
+    @_spi(Internal) public let uiPreviewMode: Bool
+
```
aboedo, line 66: maybe we can add the API tests to go along with it in this same PR? Those serve to also showcase the usage

Sources/Misc/DangerousSettings.swift
```diff
@@ -58,6 +58,12 @@ import Foundation
      */
     @objc public let autoSyncPurchases: Bool
 
+    /**
+     * if `true`, the SDK will return a set of mock products instead of the
+     * products obtained from StoreKit. This is useful for testing or preview purposes. 
+     */
+    @_spi(Internal) public let uiPreviewMode: Bool
```
aboedo, line 65: wonder if rather than internal, we actually want to call it something more explicit - like, I don't mind other RevenueCat clients using this, it's not necessarily _internal_, just as long as they know what they're doing and how it affects the app

aboedo: that being said I can't think of any names other than `@_api(UIPreviewMode) public let uiPreviewMode: Bool`, which admittedly feels redundant. 

Maybe it'd be less redundant once we add other related APIs, like the ability to state which products you'll get, fake errors, mock customer info, etc

aboedo (commented):
awesome!

Sources/Misc/DangerousSettings.swift
JayShortway: Imo "Internal" communicates this. It doesn't prohibit anyone from using it, but it's just us saying that we don't provide any guarantees wrt source and binary compatibility of this api going forward. So if they're using it, they're on their own.

aboedo: > Internal is immediately clear, where UIPreviewMode needs explanation.

This is already more than enough reason to go with `Internal`, I'm sold! Glad I'm not the one writing APIs anymore 😃

Sources/Misc/DangerousSettings.swift
```diff
@@ -101,13 +107,24 @@ import Foundation
 
     }
 
-    /// Designated initializer
+    /**
+     * Only use a Dangerous Setting if suggested by RevenueCat support team.
```
JayShortway, line 111: Just a nit, but we can also omit this line imo. It's very unlikely that the support team will suggest to use this API, and even if they would, they still can. This line kinda legitimizes the use of this API imo, but maybe that's just me haha.

JayShortway (approved):
`@_spi` all the things! 🙌 

I don't have a strong opinion on whether it should be in `DangerousSettings` or not. I'd also be okay to just move this to the regular `Configuration`. However, since it's internal we can always change this later.

Sources/Misc/DangerousSettings.swift
ajpallares: Yep. I agree. This line is already present in other places in `DangerousSettings`. But perhaps in this `init` it makes even less sense, because it's under the `@_spi(Internal)` attribute. I'll remove it. Thanks!

Sources/Misc/DangerousSettings.swift
ajpallares: Done in f406e82a16500e5e6ecb9b2921ba31d1f3107c30

ajpallares:
@RCGitBot please test

tonidero (approved):
LGTM! Nice knowing this 💪

PR #4715, opened 2025-01-27
Remove caption3 from paywalls font provider

joshdholtz:
### Motivation

`caption3` has been removed in latest Xcode 

### Description

Just removing `caption3` since its been deprecated since 2019


fire-at-will (approved):
🚀

PR #4718, opened 2025-01-27
v4: Fix crash in iOS 11-12 when using MainActor

MarkVillacampa:
Fix crash on iOS 11 and 12 when compiling with Xcode 16 and instantiating an array with type @MainActor lambda

`var foo: [@MainActor @Sendable () -> Void] = []`

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -33,10 +33,10 @@ public typealias PurchaseResultData = (transaction: StoreTransaction?,
 /**
  Completion block for ``Purchases/purchase(product:completion:)``
  */
-public typealias PurchaseCompletedBlock =  @MainActor @Sendable (StoreTransaction?,
-                                                      CustomerInfo?,
-                                                      PublicError?,
-                                                      Bool) -> Void
+public typealias PurchaseCompletedBlock = @MainActor @Sendable (StoreTransaction?,
+                                                                CustomerInfo?,
+                                                                PublicError?,
+                                                                Bool) -> Void
 
 /**
  Block for starting purchases in ``PurchasesDelegate/purchases(_:readyForPromotedProduct:purchase:)``
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -33,10 +33,10 @@ public typealias PurchaseResultData = (transaction: StoreTransaction?,
 /**
  Completion block for ``Purchases/purchase(product:completion:)``
  */
-public typealias PurchaseCompletedBlock = @MainActor @Sendable (StoreTransaction?,
-                                                                CustomerInfo?,
-                                                                PublicError?,
-                                                                Bool) -> Void
+public typealias PurchaseCompletedBlock =  @Sendable (StoreTransaction?,
+                                                      CustomerInfo?,
+                                                      PublicError?,
+                                                      Bool) -> Void
 
 /**
  Block for starting purchases in ``PurchasesDelegate/purchases(_:readyForPromotedProduct:purchase:)``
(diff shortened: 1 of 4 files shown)


Sources/Purchasing/StoreKit1/StoreKitRequestFetcher.swift
```diff
@@ -27,7 +27,7 @@ class StoreKitRequestFetcher: NSObject {
 
     private let requestFactory: ReceiptRefreshRequestFactory
     private var receiptRefreshRequest: SKRequest?
-    private var receiptRefreshCompletionHandlers: [@MainActor @Sendable () -> Void]
+    private var receiptRefreshCompletionHandlers: [@Sendable () -> Void]
```
facumenzella, line 30: Just to understand, we shouldn't be using `@MainActor`, right?

aboedo: exclusively in v4 of the iOS SDK. This is because it has a deployment target of iOS 11

aboedo: v5 is iOS 15+ so we can use it there

manuelescrig:
I'd love if this can get merged. I'm using Version 4 of SDK on iOS 12 and when compiling it with Xcode 16 I'm getting this crash so I'm still compiling it on Xcode 15.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -33,10 +33,10 @@ public typealias PurchaseResultData = (transaction: StoreTransaction?,
...
  Completion block for ``Purchases/purchase(product:completion:)``
  */
-public typealias PurchaseCompletedBlock = @MainActor @Sendable (StoreTransaction?,
-                                                                CustomerInfo?,
-                                                                PublicError?,
-                                                                Bool) -> Void
+public typealias PurchaseCompletedBlock =  @Sendable (StoreTransaction?,
+                                                      CustomerInfo?,
+                                                      PublicError?,
+                                                      Bool) -> Void
```
aboedo, line 39: I needed to remove this reference otherwise we'd still get a crash, since we have places where we store collections of these `@MainActor` references.

aboedo (approved):
tested and working on iOS 12, we should do a sanity pass to ensure we don't introduce any regression for newer devices and then ship!

aboedo:
guess we don't get off _that_ easy and I broke tests?

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -33,10 +33,10 @@ public typealias PurchaseResultData = (transaction: StoreTransaction?,
 /**
  Completion block for ``Purchases/purchase(product:completion:)``
  */
-public typealias PurchaseCompletedBlock = @MainActor @Sendable (StoreTransaction?,
-                                                                CustomerInfo?,
-                                                                PublicError?,
-                                                                Bool) -> Void
+public typealias PurchaseCompletedBlock =  @MainActor @Sendable (StoreTransaction?,
```
ajpallares, line 36: This one keeps the `@MainActor`. I'm guessing it's not used anymore in this v4. If so, perhaps we should remove it completely to prevent potential future errors if we need to make a new release of v4?

ajpallares (approved):
LGTM. Just one question

Sources/Purchasing/Purchases/Purchases.swift
aboedo: I think the concern around that one was about making further changes to the API - like, if your code relies on this being `@MainActor`, you might have broken compilation after upgrading. 

I mean, the original problem was to add `@MainActor` in a non-minor anyway, but kinda don't want to repeat the same issue twice

aboedo: might as well kill the extra space ` ` so it doesn't even show up in the diff tho 😅

aboedo (approved):

MarkVillacampa:
@manuelescrig we've released version 4.43.4 of the SDK which includes this fix, please try it and let us know if you have any issues!

https://github.com/RevenueCat/purchases-ios/releases/tag/4.43.4

PR #4719, opened 2025-01-27
[Paywalls V2] Fix paywalls badge rendering

joshdholtz:
## Motivation

Badge was not rendering on package and had some padding issues

## Description

- Badge is no longer using a box for codables
- Everything in `ViewModelFactory` is `toStackViewModel` for stacks
- `toStackViewModel` is doing badge view modeling
  - Models that whole stack (instead of the stacks components) which then includes the correct padding and margin on the stack

### Example

1) Badge was not showing before this (because badge wasn't being view model-ed on package's stack)
2) Badge didn't have any padding inside it (because view model was happening on its stacks components and needed to include stack)

👇 

Simulator Screenshot - iPhone 16 - 2025-01-27 at 17 42 58

### DIFF
--- a/Sources/Paywalls/Components/Common/PaywallComponentPropertyTypes.swift
+++ b/Sources/Paywalls/Components/Common/PaywallComponentPropertyTypes.swift
@@ -512,50 +512,31 @@ public extension PaywallComponent {
     enum BadgeStyle: String, Codable, Sendable, Hashable, Equatable {
 
         case edgeToEdge = "edge_to_edge"
-        case overlaid = "overlaid"
+        case overlaid = "overlay"
         case nested = "nested"
 
     }
 
-    struct Badge: Codable, Sendable, Hashable, Equatable {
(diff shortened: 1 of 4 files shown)


Sources/Paywalls/Components/Common/PaywallComponentPropertyTypes.swift
```diff
@@ -512,50 +512,31 @@ public extension PaywallComponent {
...
 
         public let style: BadgeStyle
         public let alignment: TwoDimensionAlignment
-        public let stack: CodableBox<StackComponent>
+        public let stack: StackComponent
 
-    }
-
-    // Holds a reference to a `Codable` value.
-    final class CodableBox<T: Codable>: Codable {
```
MarkVillacampa, line 529: 🔥

MarkVillacampa (approved):
All this makes sense!

PR #4720, opened 2025-01-28
[Paywalls V2] Deprecated `paywallFooter` for fwefafawefwe

joshdholtz:
### Motivation

We need to deprecate `paywallFooter` since it has a breaking behavior change for Paywalls V2

### Description

Deprecating `paywallFooter` in favor of `paywallFooterForV1`


RevenueCatUI/View+PresentPaywallFooter.swift
```diff
@@ -78,6 +81,12 @@ extension View {
     ///
     /// ### Related Articles
     /// [Documentation](https://rev.cat/paywalls)
+    @available(iOS, deprecated: 1, renamed: "paywallFooterForV1(condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    @available(tvOS, deprecated: 1, renamed: "paywallFooterForV1(condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    @available(watchOS, deprecated: 1, renamed: "paywallFooterForV1(condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    @available(macOS, deprecated: 1, renamed: "paywallFooterForV1(condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    @available(macCatalyst, deprecated: 1, renamed: "paywallFooterForV1(condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
```
ajpallares, line 88: If I'm not wrong, the `myAppPurchaseLogic:` parameter is missing in the `renamed:` attribute.

RevenueCatUI/View+PresentPaywallFooter.swift
```diff
@@ -168,6 +222,12 @@ extension View {
     ///
     /// ### Related Articles
     /// [Documentation](https://rev.cat/paywalls)
+    @available(iOS, deprecated: 1, renamed: "paywallFooterForV1(offering:condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    @available(tvOS, deprecated: 1, renamed: "paywallFooterForV1(offering:condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    @available(watchOS, deprecated: 1, renamed: "paywallFooterForV1(offering:condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    @available(macOS, deprecated: 1, renamed: "paywallFooterForV1(offering:condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    @available(macCatalyst, deprecated: 1, renamed: "paywallFooterForV1(offering:condensed:fonts:purchaseStarted:purchaseCompleted:purchaseCancelled:restoreStarted:restoreCompleted:purchaseFailure:restoreFailure:)")
+    // swiftlint:enable line_length
```
ajpallares, line 230: Same here. the `myAppPurchaseLogic:` parameter is missing in the `renamed:` attribute

RevenueCatUI/View+PresentPaywallFooter.swift
```diff
@@ -109,6 +118,50 @@ extension View {
...
+    /// ```swift
+    /// var body: some View {
+    ///    YourPaywall()
+    ///      .paywallFooterForV1()
+    /// }
+    /// ```
+    ///
+    /// ### Related Articles
+    /// [Documentation](https://rev.cat/paywalls)
+    public func paywallFooterForV1(
```
tonidero, line 134: How about, `paywallFooterForV1Template` or `paywallFooterNonCustomizableTemplate`? I'm honestly not sure either... Just wondering whether `V1` will actually mean something for developers...

tonidero (approved):
Just a possible rename, but I'm not in love with either TBH...

ajpallares (commented):
My thoughts about the rename:
* I'd argue that Paywalls v2 is a new feature. Thus having the existing `paywallFooter` method work differently for v2 paywalls is not a breaking change _per se_. It will still work the same for v1 paywalls. However, even though we could clarify this in the method documentation (as you've done in the new v1 methods), I totally see the potential for future confusion there.
* Now I see the advantage of making it clear via method name that the `paywallFooter` method should only be used for v1 paywalls. I'm not a huge fan of that naming either, but **if we're renaming it**, I think I would go with explicitly having the v1 in the method name.

Finally, about the implementation, since this is nothing but a method rename, perhaps it would be clearer (even for us) if the old `paywallFooter` methods would actually call their new `paywallFooterV1` counterparts, like this:
```
    public func paywallFooter(
        offering: Offering,
        condensed: Bool = false,
        myAppPurchaseLogic: MyAppPurchaseLogic? = nil,
        purchaseStarted: PurchaseOfPackageStartedHandler? = nil,
        purchaseCompleted: PurchaseOrRestoreCompletedHandler? = nil,
        purchaseCancelled: PurchaseCancelledHandler? = nil,
        restoreStarted: RestoreStartedHandler? = nil,
        restoreCompleted: PurchaseOrRestoreCompletedHandler? = nil,
        purchaseFailure: PurchaseFailureHandler? = nil,
        restoreFailure: PurchaseFailureHandler? = nil
    ) -> some View {
        return self.paywallFooterForV1(offering: offering,
                                       condensed: condensed,
                                       fonts: fonts,
                                       myAppPurchaseLogic: myAppPurchaseLogic,
                                       purchaseStarted: purchaseStarted,
                                       purchaseCompleted: purchaseCompleted,
                                       purchaseCancelled: purchaseCancelled,
                                       restoreStarted: restoreStarted,
                                       restoreCompleted: restoreCompleted,
                                       purchaseFailure: purchaseFailure,
                                       restoreFailure: restoreFailure)
    }
```

RevenueCatUI/View+PresentPaywallFooter.swift
JayShortway: I know we don't want to use the word "legacy" before V2 is GA, but... maybe something similar to "legacy"? 😅 Also "legacy" might be clearer than "V1": `legacyPaywallFooter`?

vegaro: I don't like the v1 either. There's no v2, and it actually defaults to a default footer if loading a v2 paywall.

I think it needs to be something that implies "stop using this". Throwing some bad ideas `paywallFooterOrDefaultIfUnsupported`

ajpallares (approved):

PR #4721, opened 2025-01-28
[Paywalls V2] Rename `paywallFooter` to `originalTemplatePaywallFooter`

joshdholtz:
### Motivation

We need to deprecate `paywallFooter` since it has a breaking behavior change for Paywalls V2

### Description

Deprecating `paywallFooter` in favor of `originalTemplatePaywallFooter`


ajpallares (approved):

tonidero (approved):
I think the naming makes more sense for now, since it will match what we have in the dashboard 👍

RevenueCatUI/View+PresentPaywallFooter.swift
```diff
@@ -89,10 +98,49 @@ extension View {
...
+            restoreStarted: restoreStarted,
+            restoreCompleted: restoreCompleted,
+            purchaseFailure: purchaseFailure,
+            restoreFailure: restoreFailure
+        )
+    }
+
+    /// Presents a ``PaywallFooterView`` at the bottom of a view that loads the `Offerings.current`.
+    /// If you are presenting a V1 paywall, this will show the footer of the template you selected.
+    /// If you are presenting a V2 paywall, this will show a default footer since V2 paywalls
```
JayShortway, line 118: We can keep the V1/V2 stuff here for now imo, but we should be mindful to update this when V2 becomes the default / gets to GA.

JayShortway (approved):
LGTM!

PR #4722, opened 2025-01-28
[Paywalls V2] Carousel component

joshdholtz:
### Motivation

Gotta carousel!

### Description

https://github.com/user-attachments/assets/bb996b63-a342-45cb-bf8c-569a5f6c39bb

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
@@ -10,17 +10,13 @@
 //  CarouselComponentView.swift
 //
 //  Created by Josh Holtz on 1/27/25.
+// swiftlint:disable file_length
 
 import Foundation
 import RevenueCat
 import SwiftUI
 
 #if PAYWALL_COMPONENTS
-
-// TODO: add default selected index
-// TODO: add colors and size to indicator
-// TODO: fix touching on animated
-// TODO: SUPER FUTURE - drag on indicator
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 struct CarouselComponentView: View {
@@ -137,9 +133,9 @@ private struct CarouselView<Content: View>: View {
     @GestureState private var translation: CGFloat = 0
 
     /// A timer for auto-play, if enabled.
-    @State private var autoTimer: Timer? = nil
+    @State private var autoTimer: Timer?
     @State private var isPaused: Bool = false
-    @State private var pauseEndDate: Date? = nil
+    @State private var pauseEndDate: Date?
 
     // MARK: - Init
 
@@ -359,16 +355,14 @@ private struct CarouselView<Content: View>: View {
 
         // If user is at least 2 copies in from the front, we can drop 1 copy from the front.
         while index >= 2 * originalCount,
-              data.count / originalCount > maxCopiesAllowed
-        {
+              data.count / originalCount > maxCopiesAllowed {
             data.removeFirst(originalCount)
             index -= originalCount
         }
 
         // If user is at least 2 copies from the end, we can drop 1 copy from the end.
         while index < data.count - 2 * originalCount,
-              data.count / originalCount > maxCopiesAllowed
-        {
+              data.count / originalCount > maxCopiesAllowed {
             data.removeLast(originalCount)
         }
     }
@@ -516,7 +510,10 @@ struct CarouselComponentView_Previews: PreviewProvider {
                         padding: PaywallComponent.Padding(top: 20, bottom: 20, leading: 20, trailing: 20),
                         margin: PaywallComponent.Padding(top: 20, bottom: 20, leading: 20, trailing: 20),
                         background: .color(.init(light: .hex("#ffcc00"))),
-                        shape: .rectangle(.init(topLeading: 20, topTrailing: 20, bottomLeading: 20, bottomTrailing: 20)),
+                        shape: .rectangle(.init(topLeading: 20,
+                                                topTrailing: 20,
+                                                bottomLeading: 20,
+                                                bottomTrailing: 20)),
                         pages: [
                             .init(
                                 components: [],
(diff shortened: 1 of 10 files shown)


Sources/Paywalls/Components/PaywallCarouselComponent.swift
```diff
@@ -0,0 +1,112 @@
...
+
+        let type: ComponentType
+        public let alignment: VerticalAlignment
+        public let size: Size
+        public let spacing: CGFloat?
+        public let padding: Padding
+        public let margin: Padding
+        public let shape: Shape?
+        public let border: Border?
+        public let shadow: Shadow?
```
tonidero, line 43: Should we also add a background color property? Also do we need to add the page indicator settings here?

tonidero: Another property that we might want is the initialPage to use

tonidero:
I merged a version of this in Android. The schema I used is in this PR: https://github.com/RevenueCat/purchases-android/pull/2092. Though we can perform any modifications if needed of course!

RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
```diff
@@ -0,0 +1,693 @@
...
+
+    // MARK: - Init
+
+    init(
+        width: CGFloat,
+        pages: [Content],
+        initialIndex: Int,
+        loop: Bool = false,
+        spacing: CGFloat = 16,
+        cardWidth: CGFloat = 300,
```
tonidero, line 149: Hmm seems weird to have these defaults? 🤔 Seems these should be in the previews if needed?

tonidero (commented):
LGTM! Lmk when it's ready for a final review and I can approve it!

RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
ajpallares: I was going to comment the same. If we're passing values anyways, we should avoid the defaults. I've been burned in the past by unnecessary default values

RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
```diff
@@ -0,0 +1,693 @@
...
+        } else {
+            // Non-loop clamp
+            index = max(0, min(index, data.count - 1))
+        }
+
+        // Pause auto-play for 10 seconds
+        pauseAutoPlay(for: 10)
+    }
+
+    private func pauseAutoPlay(for seconds: TimeInterval) {
```
ajpallares, line 326: Perhaps this method can do an early return if autoplay is off (to avoid unnecessary operations). Something like this
```suggestion
    private var autoPlayEnabled: Bool { 
        return self.msTimePerSlide != nil && self.msTransitionTime != nil 
    }
    
    private func pauseAutoPlay(for seconds: TimeInterval) {
        guard self.autoPlayEnabled else { return }
        
```

RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
```diff
@@ -0,0 +1,693 @@
...
+    }
+
+    var indicator: DisplayablePageControlIndicator {
+        pageControl.default
+    }
+
+    var body: some View {
+        if self.originalCount <= 1 {
+            EmptyView()
+        } else {
```
ajpallares, line 411: Perhaps we don't need the `EmptyView()`
```suggestion
        if self.originalCount > 1 {
```

RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
```diff
@@ -0,0 +1,693 @@
...
+                    Capsule()
+                        .fill(localCurrentIndex == index ? activeIndicator.color : indicator.color)
+                        .frame(
+                            width: localCurrentIndex == index ? activeIndicator.width : indicator.width,
+                            height: localCurrentIndex == index ? activeIndicator.height : indicator.height
+                        )
+                        .animation(.easeInOut, value: self.localCurrentIndex)
+                }
+            }
+            .padding(self.pageControl.padding)
```
ajpallares, line 423: Would it make sense here to ignore the horizontal padding? I say this to avoid these kinds of clipping of the cards on the edges. But I don't know if ignoring these paddings could have other undesired consequences
image

ajpallares (commented):
Some comments for now

RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
joshdholtz: Ooooops... this was here from my POC before I had an API figured out 😅

RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
```diff
@@ -0,0 +1,693 @@
...
+        if loop {
+            expandDataIfNeeded()
+            pruneDataIfNeeded()
+        } else {
+            // Non-loop clamp
+            index = max(0, min(index, data.count - 1))
+        }
+
+        // Pause auto-play for 10 seconds
+        pauseAutoPlay(for: 10)
```
tonidero, line 319: In android, I currently disable manual dragging if autoplay is enabled... I guess we want to allow this there? not sure if we might want it to be a setting though... (Not a blocker for this PR)

tonidero (approved):
LGTM!

RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
tonidero: Hmm it's a good point... But for now, I think we should just limit setting it in the dashboard.

ajpallares (approved):
Such an incredible PR 🤯

PR #4723, opened 2025-01-28
fix: Revisit environment values for CustomerCenter

facumenzella:
### Motivation
This PR revisits every `.environment` use inside CustomerCenter. The root cause is related to https://github.com/RevenueCat/purchases-ios/pull/4704

### Description
1. Order alphabetically environment declarations in each view
2. Set explicit environment values for each view
3. Reordered `ManageSubscriptionsView` moving all navigation modifiers to the same place

> I am still thinking what's the best way to test this, or write type-safe code.


ajpallares (approved):
LGTM

PR #4725, opened 2025-01-28
UI Preview Mode: app user ID

ajpallares:
### Checklist

### Motivation
The SDK uses a fixed, specific user ID when initialized in UI preview mode in order to prevent cluttering usage statistics for the developer

### Description
When using the UI preview mode introduced in #4714, the SDK will force the `appUserID` to be `"$RCPREVIEWMODE"`.

### DIFF
--- a/Sources/Identity/IdentityManager.swift
+++ b/Sources/Identity/IdentityManager.swift
@@ -36,6 +36,7 @@ class IdentityManager: CurrentUserProvider {
 
     init(
         deviceCache: DeviceCache,
+        systemInfo: SystemInfo,
         backend: Backend,
         customerInfoManager: CustomerInfoManager,
         attributeSyncing: AttributeSyncing,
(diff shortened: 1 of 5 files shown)


Sources/Identity/IdentityManager.swift
```diff
@@ -100,6 +105,7 @@ class IdentityManager: CurrentUserProvider {
         "$RCAnonymousID:\(UUID().uuidString.replacingOccurrences(of: "-", with: "").lowercased())"
     }
 
+    static let uiPreviewModeAppUserID: String = "$RCPREVIEWMODE"
```
tonidero, line 108: Hmm how about `$RC_PREVIEW_MODE_USER`?

Sources/Identity/IdentityManager.swift
```diff
@@ -50,10 +51,14 @@ class IdentityManager: CurrentUserProvider {
             Logger.warn(Strings.identity.logging_in_with_empty_appuserid)
         }
 
-        let appUserID = appUserID?.notEmptyOrWhitespaces
+        let appUserID = if systemInfo.dangerousSettings.uiPreviewMode {
+            Self.uiPreviewModeAppUserID
```
tonidero, line 55: Maybe we should add a debug log indicating that we're defaulting to this specific user ID (and ignoring any given user ID) just in case?

tonidero (approved):
Looks great! Just a couple of smallish comments

tonidero (approved):

PR #4730, opened 2025-01-29
[Paywalls V2] Fix footer spacing issues

joshdholtz:
### Motivation

Footer was showing empty space between footer and safe space on the bottom

### Description

- Ignoring edges on bottom instead of top (which I think was a typo)
- Added a spacer to the bottom of the footer to expand space to fill out the safe space

👇 This is what the `Spacer()` fixes

| Before | After |
|--------|--------|
| Simulator Screenshot - iPhone 16 - 2025-01-28 at 23 13 08 | Screenshot 2025-01-28 at 11 12 54 PM |

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Root/RootView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Root/RootView.swift
@@ -47,6 +47,11 @@ struct RootView: View {
                     )
                 )
                 .fixedSize(horizontal: false, vertical: true)
+
+                // This spacer is to prevent the bottom most view from expanding into the
+                // safe space. iOS will take the bottom view and expand it to fill the
+                // the safe space and we don't want anything there.
+                Spacer()
             }
         }
     }
--- a/RevenueCatUI/Templates/V2/PaywallsV2View.swift
+++ b/RevenueCatUI/Templates/V2/PaywallsV2View.swift
@@ -241,7 +241,7 @@ private struct LoadedPaywallsV2View: View {
             ComponentsView(
                 componentViewModels: [paywallState.componentViewModel],
                 onDismiss: self.onDismiss
-            )
+            )            
         }
         .environmentObject(self.selectedPackageContext)
         .frame(maxHeight: .infinity, alignment: .topLeading)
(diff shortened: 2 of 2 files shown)


RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -241,15 +241,15 @@ private struct LoadedPaywallsV2View: View {
             ComponentsView(
                 componentViewModels: [paywallState.componentViewModel],
                 onDismiss: self.onDismiss
-            )
+            )
```
ajpallares, line 244: This gives a lint warning

RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -241,15 +241,15 @@ private struct LoadedPaywallsV2View: View {
...
+            )            
         }
         .environmentObject(self.selectedPackageContext)
         .frame(maxHeight: .infinity, alignment: .topLeading)
         .backgroundStyle(
             self.paywallState.componentsConfig.background
                 .asDisplayable(uiConfigProvider: uiConfigProvider).backgroundStyle
         )
-        .edgesIgnoringSafeArea(.top)
+        .edgesIgnoringSafeArea(.bottom)
```
ajpallares, line 252: Yeah, this makes sense. Having `.edgesIgnoringSafeArea(.top)` could mean trouble when the paywall got presented in full cover

ajpallares (approved):

RevenueCatUI/Templates/V2/Components/Root/RootView.swift
```diff
@@ -47,6 +47,11 @@ struct RootView: View {
                     )
                 )
                 .fixedSize(horizontal: false, vertical: true)
+
+                // This spacer is to prevent the bottom most view from expanding into the
+                // safe space. iOS will take the bottom view and expand it to fill the
+                // the safe space and we don't want anything there.
+                Spacer()
```
tonidero, line 54: Hmm I'm wondering, should we remove the `edgesIgnoringSafeArea` completely so the view doesn't extend over the safe area? Would that allow us to remove this? (I'm not sure about this, I might be wrong/missing something).

tonidero (approved):
I just have a question... But code seems ok to me!

RevenueCatUI/Templates/V2/Components/Root/RootView.swift
joshdholtz: Its still need other this this happens 👇 🤷‍♂️ 
Screenshot 2025-01-29 at 5 17 17 AM

PR #4733, opened 2025-01-29
fix: Avoid the use of return switch

facumenzella:
### Motivation
This PR tries to avoid using `return switch` because it breaks the CI


vegaro (approved):
It got the broken one so let's go!

tonidero (approved):

vegaro:
@RCGitBot please test

PR #4735, opened 2025-01-29
UI Preview Mode: mock products

ajpallares:
### Checklist

### Motivation
The UI Preview Mode of the SDK should limit all the functionality of the SDK to only enable implementing the Paywall Previews functionality in the RevenueCat app.

### Description
Mock the `StoreProducts`, as we cannot fetch them from StoreKit

### DIFF
--- a/Sources/Purchasing/OfferingsManager.swift
+++ b/Sources/Purchasing/OfferingsManager.swift
@@ -192,7 +192,7 @@ private extension OfferingsManager {
             return
         }
 
-        self.productsManager.products(withIdentifiers: productIdentifiers) { result in
+        self.fetchProducts(withIdentifiers: productIdentifiers) { result in
             let products = result.value ?? []
 
             guard products.isEmpty == false else {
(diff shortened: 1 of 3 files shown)


Sources/Purchasing/OfferingsManager.swift
```diff
@@ -275,6 +275,95 @@ private extension OfferingsManager {
...
+                packageType = .twoMonth
+            } else if id.contains("month") {
+                packageType = .monthly
+            } else if id.contains("week") {
+                packageType = .weekly
+            } else {
+                packageType = .custom
+            }
+            return ProductTypeMock(packageType: packageType) ?? .default
+        }
```
ajpallares, line 365: This new logic tries to implement the ideas given by @tonidero [in this other PR](https://github.com/RevenueCat/purchases-ios/pull/4693#discussion_r1924919561). Was something like this what you had in mind?

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -275,6 +275,95 @@ private extension OfferingsManager {
...
+                guard productType.period?.unit == .year else { return nil }
+                return TestStoreProductDiscount(
+                    identifier: "intro",
+                    price: 0,
+                    localizedPriceString: "$0.00",
+                    paymentMode: .freeTrial,
+                    subscriptionPeriod: SubscriptionPeriod(value: 1, unit: .week),
+                    numberOfPeriods: 1,
+                    type: .introductory
+                )
```
ajpallares, line 313: Based on @JayShortway's comment [here](https://github.com/RevenueCat/purchases-ios/pull/4693#discussion_r1924834946), I'm only adding mocked introductory offers for yearly subscriptions

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -275,6 +275,95 @@ private extension OfferingsManager {
...
+            let mockProducts = self.createMockProducts(productIdentifiers: identifiers, fromResponse: response)
+            completion(.success(mockProducts))
+        } else {
+            self.productsManager.products(withIdentifiers: identifiers, completion: completion)
+        }
+    }
+
+    // MARK: - Mocks for UI Preview mode
+
+    private func createMockProducts(
```
tonidero, line 293: I'm wondering if it makes more sense to name this `createPreviewProducts` and the same for all these changes... Mostly so we don't have "Mocks" in release code, but just "preview" code. Wdyt?

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -275,6 +275,95 @@ private extension OfferingsManager {
...
+                )
+            }()
+
+            let testProduct = TestStoreProduct(
+                localizedTitle: "PRO \(productType.type)",
+                price: Decimal(productType.price),
+                localizedPriceString: String(format: "$%.2f", productType.price),
+                productIdentifier: identifier,
+                productType: productType.period == nil ? .nonConsumable : .autoRenewableSubscription,
+                localizedDescription: "\(productType.type) subscription",
```
tonidero, line 322: Hmm it might not be a subscription if it's a lifetime/consumable...  Maybe we can also determine the copy here according to the type?

tonidero (approved):
A couple of suggestions but LGTM!

Sources/Purchasing/OfferingsManager.swift
ajpallares: Good point! Yes, I like what you suggest

Sources/Purchasing/OfferingsManager.swift
ajpallares: You're right! I left only the `type` in case it's not a subscription (which can only be `"lifetime"` in the mocked/preview products)

tonidero (approved):
LGTM!

Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
```diff
@@ -12,7 +12,7 @@
 //  Created by Juanpe Catalán on 9/8/21.
 
 import Nimble
-@testable import RevenueCat
+@testable @_spi(Internal) import RevenueCat
```
facumenzella, line 15: neat 💅

PR #4740, opened 2025-01-29
[Paywalls V2] Support variable mapping

joshdholtz:
### Motivation

Support the backward compatibility variable mapping

### Description

- Unknown variables and functions will map existing ones that have a sensible match
- Super unknown (which shouldn't happen) just leave leave empty string for variable and no-op for functions

### DIFF
--- a/RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift
+++ b/RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift
@@ -49,7 +49,7 @@ struct VariableHandlerV2 {
             let variable = self.findVariable(variableRaw)
             let function = functionRaw.flatMap { self.findFunction($0) }
 
-            let processedVariable = variable?.process(
+            let processedVariable = variable.process(
                 package: package,
                 locale: locale,
                 localizations: localizations,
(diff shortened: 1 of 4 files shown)


RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift
```diff
@@ -39,25 +46,82 @@ struct VariableHandlerV2 {
...
+            return function?.process(processedVariable) ?? processedVariable
         }
 
         return whisker.render()
     }
 
+    private func findVariable(_ variableRaw: String) -> VariablesV2 {
+        guard let originalVariable = VariablesV2(rawValue: variableRaw) else {
+
+            let backSupportedVariableRaw = self.variableCompatibilityMap[variableRaw]
```
JayShortway, line 69: Wild thought: should this support recursion? E.g. `backSupportedVariableRaw` is also unsupported, and we go look in the `variableCompatibilityMap` again to see if there's a `backBackSupportedVariableRaw`. Probably not right? 😅

RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift
```diff
@@ -144,6 +208,8 @@ enum VariablesV2: String {
     case productRelativeDiscount = "product.relative_discount"
     case productStoreProductName = "product.store_product_name"
 
+    case unsupported = "__unsupported__"
```
JayShortway, line 211: I'd be okay if `findVariable` and `findFunction` returned `nil` if the variable / function is not found, as `nil` indicates absence of data, but not a blocker.

JayShortway (approved):
Good to have this in! 💪

PR #4744, opened 2025-01-30
[Paywalls V2] Ignore top safe area edges for image

joshdholtz:
### Motivation

Be smart about when and where to ignore safe area edges so header images expand all the way to the top edge.

### Description

1. In the `ViewModelFactory`, find if the first non-container type component is an image
    - If we find one, also pair it with its parent if its a ZStack
        - We need this so that we can apply safe area insets to the non-images in the ZStack so they stay pushed down
        - We add a new boolean into the `StackViewModel` of the containing ZStack that we need to apply safe area inset padding
2.  If we found that an image is first:
    - Apply `.edgesIgnoringSafeArea(.top)` to the entire paywall contents
    - Send the value of the safe area insets as an environment variable
3. When rendering the paywall, we apply the safe area insets into the ZStack components if needed so they stay pushed down

👉  See https://github.com/RevenueCat/purchases-ios/pull/4745 for the Paywall Tester changes

https://github.com/user-attachments/assets/f5b06e0b-051d-45d2-ac8c-302ba24d317d

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/ComponentsView.swift
+++ b/RevenueCatUI/Templates/V2/Components/ComponentsView.swift
@@ -75,9 +75,11 @@ struct ComponentsView: View {
                 case .tabControlToggle(let viewModel):
                     TabControlToggleComponentView(viewModel: viewModel, onDismiss: onDismiss)
                 }
-            }.applyIf(index > 0 && self.ignoreSafeArea) { view in
-                view
-                    .padding(.top, self.safeAreaInsets.top)
+            }
+            // Applies a top padding to mimmic safe area insets
+            // This was designed to be applied to the 
+            .applyIf(index > 0 && self.ignoreSafeArea) { view in
+                view.padding(.top, self.safeAreaInsets.top)
             }
         }
     }
--- a/RevenueCatUI/Templates/V2/Components/ComponentsView.swift
+++ b/RevenueCatUI/Templates/V2/Components/ComponentsView.swift
@@ -23,7 +23,7 @@ struct ComponentsView: View {
     private var safeAreaInsets
 
     let componentViewModels: [PaywallComponentViewModel]
-    let ignoreSafeArea: Bool
+    let applySafeAreaInsetForZStackChildren: Bool
     private let onDismiss: () -> Void
 
     init(
@@ -32,7 +32,7 @@ struct ComponentsView: View {
         onDismiss: @escaping () -> Void
     ) {
         self.componentViewModels = componentViewModels
-        self.ignoreSafeArea = ignoreSafeArea
+        self.applySafeAreaInsetForZStackChildren = ignoreSafeArea
         self.onDismiss = onDismiss
     }
 
@@ -77,10 +77,11 @@ struct ComponentsView: View {
                 }
             }
             // Applies a top padding to mimmic safe area insets
-            // This was designed to be applied to the 
-            .applyIf(index > 0 && self.ignoreSafeArea) { view in
-                view.padding(.top, self.safeAreaInsets.top)
-            }
+            // This was designed to be applied to for ZStacks when
+            // they have a full width header image and are the first
+            // component in the paywall. This will keep the other
+            // components from going into the safe area.
+            .padding(.top, (self.applySafeAreaInsetForZStackChildren && index > 0) ? self.safeAreaInsets.top : 0)
         }
     }
 
--- a/RevenueCatUI/Templates/V2/Components/ComponentsView.swift
+++ b/RevenueCatUI/Templates/V2/Components/ComponentsView.swift
@@ -26,7 +26,11 @@ struct ComponentsView: View {
     let ignoreSafeArea: Bool
     private let onDismiss: () -> Void
 
-    init(componentViewModels: [PaywallComponentViewModel], ignoreSafeArea: Bool = false, onDismiss: @escaping () -> Void) {
+    init(
+        componentViewModels: [PaywallComponentViewModel],
+        ignoreSafeArea: Bool = false,
+        onDismiss: @escaping () -> Void
+    ) {
         self.componentViewModels = componentViewModels
         self.ignoreSafeArea = ignoreSafeArea
         self.onDismiss = onDismiss
--- a/RevenueCatUI/Templates/V2/Components/ComponentsView.swift
+++ b/RevenueCatUI/Templates/V2/Components/ComponentsView.swift
@@ -19,11 +19,16 @@ import SwiftUI
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 struct ComponentsView: View {
 
+    @Environment(\.safeAreaInsets)
+    private var safeAreaInsets
+
     let componentViewModels: [PaywallComponentViewModel]
+    let ignoreSafeArea: Bool
     private let onDismiss: () -> Void
 
-    init(componentViewModels: [PaywallComponentViewModel], onDismiss: @escaping () -> Void) {
+    init(componentViewModels: [PaywallComponentViewModel], ignoreSafeArea: Bool = false, onDismiss: @escaping () -> Void) {
         self.componentViewModels = componentViewModels
+        self.ignoreSafeArea = ignoreSafeArea
         self.onDismiss = onDismiss
     }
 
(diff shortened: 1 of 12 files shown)


RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
joshdholtz, line 1: Ignoring edges on all views being used as the background

RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
```diff
@@ -91,7 +91,11 @@ struct StackComponentView: View {
                       verticalAlignment: verticalAlignment.frameAlignment)
             case .zlayer(let alignment):
                 ZStack(alignment: alignment.stackAlignment) {
-                    ComponentsView(componentViewModels: self.viewModel.viewModels, onDismiss: self.onDismiss)
+                    ComponentsView(
+                        componentViewModels: self.viewModel.viewModels,
+                        ignoreSafeArea: self.viewModel.shouldApplySafeAreaInset,
```
ajpallares, line 96: The parameter name `ignoreSafeArea` seems contradictory with sending `self.viewModel.shouldApplySafeAreaInset`

RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -237,23 +237,47 @@ private struct LoadedPaywallsV2View: View {
...
+            VStack(spacing: 0) {
+                ComponentsView(
+                    componentViewModels: [.root(paywallState.rootViewModel)],
+                    onDismiss: self.onDismiss
+                )
+            }
+            // If the first view in the first stack is an image,
+            // we will ignore safe area pass the safe area insets in to environment
+            // If the image is in a ZStack, the ZStack will push non-images
+            // down with the inset
```
tonidero, line 250: Talking about an edge case here but what happens if the parent of the image (being the first view within the first stack) is an HStack with an image + some text to the side? I imagine in that situation we might not want to ignore the safe area... So maybe this should only be done if the parent of the image is either a VStack or a ZStack?

RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -237,23 +237,47 @@ private struct LoadedPaywallsV2View: View {
...
+            .applyIf(paywallState.rootViewModel.firstImageInfo != nil, apply: { view in
+                view
+                    .environment(\.safeAreaInsets, proxy.safeAreaInsets)
+                    .edgesIgnoringSafeArea(.top)
+            })
+            // This is normal
```
ajpallares, line 256: What does this comment mean?

RevenueCatUI/Templates/V2/Components/ComponentsView.swift
```diff
@@ -34,36 +43,43 @@ struct ComponentsView: View {
...
+                case .tabControlButton(let viewModel):
+                    TabControlButtonComponentView(viewModel: viewModel, onDismiss: onDismiss)
+                case .tabControlToggle(let viewModel):
+                    TabControlToggleComponentView(viewModel: viewModel, onDismiss: onDismiss)
+                }
+            }
+            // Applies a top padding to mimmic safe area insets
+            // This was designed to be applied to the 
+            .applyIf(index > 0 && self.ignoreSafeArea) { view in
+                view.padding(.top, self.safeAreaInsets.top)
```
JayShortway, line 82: This comment feels unfinished 😅

ajpallares (approved):
Looks good to me! Just some very very small comments

RevenueCatUI/Templates/V2/Components/Root/RootViewModel.swift
```diff
@@ -19,15 +19,23 @@ import SwiftUI
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 class RootViewModel {
 
+    struct FirstImageInfo {
+        let imageComponent: PaywallComponent.ImageComponent
+        let parentZStack: PaywallComponent.StackComponent?
+    }
```
JayShortway, line 25: I think it would be helpful if this struct had some explanation in a comment around what the properties mean and how it's used. Doesn't have to be long imo, just something to rejig our memory when our future selves read this.

RevenueCatUI/Templates/V2/PaywallsV2View.swift
tonidero, line 250: I'm also wondering if we should make it so we only apply this if the image width is `Fill`, so it uses the whole space... In case there is something like this:
image

tonidero (approved):
LGTM! Just a couple cases I was thinking about... The edge case might be acceptable for now but I'm wondering if the design of having an image at the top of the paywall but not as a header would be affected with this...

RevenueCatUI/Templates/V2/ViewModelHelpers/ViewModelFactory.swift
```diff
@@ -293,21 +310,83 @@ struct ViewModelFactory {
...
             viewModels: viewModels,
             badgeViewModels: badgeViewModels,
+            shouldApplySafeAreaInset: shouldApplySafeAreaInset,
             uiConfigProvider: uiConfigProvider,
             localizationProvider: localizationProvider
         )
     }
 
+    // swiftlint:disable cyclomatic_complexity
+    private func findImageViewIfItsTheFirst(
```
JayShortway, line 335: Nice!

JayShortway (approved):
I think this is a great and pragmatic solution! 💪

RevenueCatUI/Templates/V2/PaywallsV2View.swift
ajpallares: Ok. I see you removed it in #4746 😄

RevenueCatUI/Templates/V2/Components/ComponentsView.swift
```diff
@@ -34,36 +43,43 @@ struct ComponentsView: View {
...
+                    TabControlComponentView(viewModel: viewModel, onDismiss: onDismiss)
+                case .tabControlButton(let viewModel):
+                    TabControlButtonComponentView(viewModel: viewModel, onDismiss: onDismiss)
+                case .tabControlToggle(let viewModel):
+                    TabControlToggleComponentView(viewModel: viewModel, onDismiss: onDismiss)
+                }
+            }
+            // Applies a top padding to mimmic safe area insets
+            // This was designed to be applied to the 
+            .applyIf(index > 0 && self.ignoreSafeArea) { view in
```
facumenzella, line 81: suuuuuper nit:
Whenever possible I prefer to avoid things like `applyIf` in favor of:

```swift
.padding(.top, (index > 0 && self.ignoreSafeArea) ? self.safeAreaInsets.top : 0)
```

	•	The view structure remains stable because .padding(.top, value) is always present.
	•	The padding is updated and SwiftUI handles efficiently.

This avoid unnecessary view hierarchy changes

RevenueCatUI/Templates/V2/PaywallsV2View.swift
joshdholtz: Oooops, you are right! Its supposed to only remove safe area inset if hte image has a width of `fill` 🤦‍♂️ WIll fix!

RevenueCatUI/Templates/V2/Components/ComponentsView.swift
joshdholtz: Ah yes! Will fix 😊 This used to be more complex in the `applyIf` but then it simplified to just one thing. Thanks for catching!

RevenueCatUI/Templates/V2/PaywallsV2View.swift
joshdholtz: Good point! I just changed the logic so that the image needs to have a `width` of `fill` for this to work so... I _think_ that should cover the `HStack` scenario 🤔

RevenueCatUI/Templates/V2/PaywallsV2View.swift
joshdholtz: Ha! This was a note to me when testing this but gets fixed in a follow up PR

ferrannp:
@joshdholtz any ideas on how to control the X (close button) position on top-right of a background image in this case? It seems to be placed differently in your video too. It seems hard to add it on the top-right corner for these devices:

- iPhone with notch or dynamic island
- iPhone SE
- Android

(to be in the same position).

PR #4755, opened 2025-02-03
feat: Dont show cancel path if lifetime subscription

facumenzella:
### Motivation
Display only the paths that are relevant to the subscription

### Description
Doing this in the backend requires some extra work, so for now we'll handle it in the client. `ManageSubscriptionsViewModel` has the relevant change. The other changes are just removing some unnecessary code, and re-organising stuff.

### DIFF
--- a/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
@@ -27,21 +27,33 @@ import SwiftUI
 class ManageSubscriptionsViewModel: ObservableObject {
 
     let screen: CustomerCenterConfigData.Screen
-    let paths: [CustomerCenterConfigData.HelpPath]
+
+    var relevantPathsForPurchase: [CustomerCenterConfigData.HelpPath] {
+        if purchaseInformation?.explanation == .lifetime {
+            return paths.filter { $0.type != .cancel }
+        } else {
+            return paths
+        }
+    }
 
     @Published
     var showRestoreAlert: Bool = false
+
     @Published
     var showPurchases: Bool = false
 
     @Published
     var feedbackSurveyData: FeedbackSurveyData?
+
     @Published
     var loadingPath: CustomerCenterConfigData.HelpPath?
+
     @Published
     var promotionalOfferData: PromotionalOfferData?
+
     @Published
     var inAppBrowserURL: IdentifiableURL?
+
     @Published
     var state: CustomerCenterViewState {
         didSet {
@@ -53,13 +65,15 @@ class ManageSubscriptionsViewModel: ObservableObject {
 
     @Published
     private(set) var purchaseInformation: PurchaseInformation?
+
     @Published
     private(set) var refundRequestStatus: RefundRequestStatus?
 
-    private var purchasesProvider: ManageSubscriptionsPurchaseType
-    private let loadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType
     private let customerCenterActionHandler: CustomerCenterActionHandler?
     private var error: Error?
+    private let loadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType
+    private let paths: [CustomerCenterConfigData.HelpPath]
+    private var purchasesProvider: ManageSubscriptionsPurchaseType
 
     init(screen: CustomerCenterConfigData.Screen,
          customerCenterActionHandler: CustomerCenterActionHandler?,
--- a/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsButtonsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsButtonsView.swift
@@ -25,16 +25,11 @@ struct ManageSubscriptionsButtonsView: View {
 
     @ObservedObject
     var viewModel: ManageSubscriptionsViewModel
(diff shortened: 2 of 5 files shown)


RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsButtonsView.swift
```diff
@@ -25,16 +25,11 @@ struct ManageSubscriptionsButtonsView: View {
...
     @ObservedObject
     var viewModel: ManageSubscriptionsViewModel
-    @Binding
-    var loadingPath: CustomerCenterConfigData.HelpPath?
-    @Environment(\.openURL)
-    var openURL
 
-    @Environment(\.localization)
-    private var localization: CustomerCenterConfigData.Localization
+    var loadingPath: CustomerCenterConfigData.HelpPath?
```
vegaro, line 29: Is this actually used?

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsButtonsView.swift
```diff
@@ -45,13 +40,10 @@ struct ManageSubscriptionsButtonsView: View {
...
 @available(watchOS, unavailable)
-struct ManageSubscriptionButton: View {
+private struct ManageSubscriptionButton: View {
 
     let path: CustomerCenterConfigData.HelpPath
-    @ObservedObject var viewModel: ManageSubscriptionsViewModel
-
-    @Environment(\.appearance)
-    private var appearance: CustomerCenterConfigData.Appearance
+    let viewModel: ManageSubscriptionsViewModel
```
vegaro, line 46: Will the button be updated if the `viewModel` changes?

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsButtonsView.swift
facumenzella: `ManageSubscriptionButton` uses it 🤔

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptionsViewModel.swift
```diff
@@ -27,21 +27,33 @@ import SwiftUI
 class ManageSubscriptionsViewModel: ObservableObject {
 
     let screen: CustomerCenterConfigData.Screen
-    let paths: [CustomerCenterConfigData.HelpPath]
+
+    var relevantPathsForPurchase: [CustomerCenterConfigData.HelpPath] {
+        if purchaseInformation?.explanation == .lifetime {
```
vegaro, line 32: I think this is a good proxy, but not bulletproof if we add another explanation that is also used for lifetime purchases. Maybe we can add another property to `purchaseInformation` that represents for real if it's a lifetime product

RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigTestData.swift
```diff
@@ -144,7 +144,8 @@ enum CustomerCenterConfigTestData {
         expirationOrRenewal: .init(label: .nextBillingDate,
                                    date: .date("June 1st, 2024")),
         productIdentifier: "product_id",
-        store: .appStore
+        store: .appStore,
+        isLifetimeSubscription: false
```
vegaro, line 148: lifetime purchases are not subscriptions, so it would be more correct to use `isLifetimeProduct`

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -80,9 +85,12 @@ struct PurchaseInformation {
                         : .expired
                     return ExpirationOrRenewal(label: label, date: .date(dateString))
                 }
+                self.isLifetimeSubscription = expiresDate == nil
```
vegaro, line 88: The transaction type is subscription, so this should always be false since it will never be a lifetime purchase

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -80,9 +85,12 @@ struct PurchaseInformation {
...
+
             case .nonSubscription:
                 self.explanation = .lifetime
                 self.expirationOrRenewal = nil
+                self.isLifetimeSubscription = false
```
vegaro, line 93: I think this should always be `true`, not `false`

vegaro: Technically we would also check the product type. 

By the way `subscribedProduct: StoreProduct? = nil` is not entirely correct since `subscribedProduct` doesn't have to be a subscription. We should rename it, but feel free to leave it like that since it's not part of this PR

vegaro (changes requested):

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
facumenzella: I'll apply this in another PR

vegaro (approved):
Another case I am thinking we shouldn't show the cancel path is if the product is a promotional. To detect a promotional, I've normally checked for the prefix of the product identifier. To test it, you can grant your user a promotional in the dashboard

Feel free to merge this and address that in another PR

facumenzella:
@vegaro I'll add another ticket and apply that too

PR #4756, opened 2025-02-03
Document Weak PurchasesDelegate Reference

fire-at-will:
### Motivation
`Purchases`'s `delegate` property is a computed property and isn't marked or documented as being weak, but the underlying `privateDelegate` that backs it is weak. This means that if a developer sets a PurchasesDelegate like this:

```swift
Purchases.shared.delegate = MyPurchasesDelegate()
```

the delegate will be deallocated. Since we only retain a _reference_ to the delegate, none of the log messages in the `Purchases.delegate.set` block will be logged, and future calls to the delegate will not be received. It's possible for this to happen without the developer's knowledge, so this PR documents the behavior.

### Description
This PR documents the fact that `Purchases.delegate` uses a weak reference.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -76,6 +76,13 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
     /// or one of is overloads.
     @objc public static var isConfigured: Bool { Self.purchases.value != nil }
 
+    /**
+     * The delegate for ``Purchases`` responsible for handling updating your app's state in response to updated
+     * customer info or promotional product purchases.
+     *
+     * - Warning: The delegate is not retained by `Purchases`, so your app must retain a reference to the delegate
+     * to prevent it from being unintentionally deallocated.
+     */
     @objc public var delegate: PurchasesDelegate? {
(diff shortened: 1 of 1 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -76,6 +76,13 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
     /// or one of is overloads.
     @objc public static var isConfigured: Bool { Self.purchases.value != nil }
 
+    /**
+     * The delegate for ``Purchases`` responsible for handling updating your app's state in response to updated
+     * customer info or promotional product purchases.
+     *
+     * - Warning: The delegate is not retained by `Purchases`, so your app must retain a reference to the delegate
```
ajpallares, line 83: Super nit: two single quotes on `Purchases` (like above)?

ajpallares (approved):
LGTM, just a very small comment

Sources/Purchasing/Purchases/Purchases.swift
fire-at-will: Love it, added in 6d46cff2a87fcc757bd7769128a8020a01af867f!

PR #4757, opened 2025-02-03
[CI]: Run iOS 13 tests in all-test CI workflow

fire-at-will:
### Description
This PR runs the iOS 13 tests when the `all-test` workflow is run since the SDK supports iOS 13+


fire-at-will:
@RCGitBot please test

facumenzella (approved):

PR #4761, opened 2025-02-05
Add Internal support for draft paywall previews

ajpallares:
### Checklist

### Motivation
We need a way to show paywalls in the draft state for the Paywall Previews functionality in the RC app

### Description
Adds some new  `@_spi`-protected public APIs regarding draft paywalls:
    * `draftPaywallComponents` property in `Offering`, to know if there's a draft paywall available.
    * A new `init` in `PaywallView` that adds the `useDraftPaywall` parameter to show the draft paywall instead of the published one.

### DIFF
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -11,7 +11,7 @@
 //
 //  Created by Nacho Soto.
 
-import RevenueCat
+@_spi(Internal) import RevenueCat
 import SwiftUI
 
 #if !os(macOS) && !os(tvOS)
@@ -31,6 +31,7 @@ public struct PaywallView: View {
     private let fonts: PaywallFontProvider
     private let displayCloseButton: Bool
     private let paywallViewOwnsPurchaseHandler: Bool
+    private let useDraftPaywall: Bool
 
     private var locale: Locale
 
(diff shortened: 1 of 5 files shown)


JayShortway (approved):
Makes sense to me!

RevenueCatUI/PaywallView.swift
```diff
@@ -104,13 +105,33 @@ public struct PaywallView: View {
...
+            offering: offering,
+            fonts: fonts,
+            displayCloseButton: displayCloseButton,
+            useDraftPaywall: false,
+            performPurchase: performPurchase,
+            performRestore: performRestore
+            )
+    }
+
+    // swiftlint:disable:next missing_docs
```
ajpallares, line 119: Idk if we should be documenting these public-but-not-public APIs 🤔  WDYT @RevenueCat/coresdk?

tonidero (approved):
This LGTM!

RevenueCatUI/PaywallView.swift
aboedo: I'd be more inclined to documenting since it's quick, and particularly adding a `- warn: this API is meant for internal use`

tonidero: Not sure... I would actually hide them from the reference docs if possible, but maybe we should have a note for people finding it through autocomplete?

tonidero (approved):
:shipit:

PR #4767, opened 2025-02-06
[Paywalls V2] Added `overflow` property to stack 

joshdholtz:
### Motivation

We need to allow stacks to have some overflow properties like scroll

### Description

Added `overflow` property to stack with values of:
- `none`
- `scroll`

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
@@ -34,6 +34,7 @@ struct StackComponentView: View {
     private var screenCondition
 
     private let viewModel: StackComponentViewModel
+    private let isScrollableByDefault: Bool
     private let onDismiss: () -> Void
     /// Used when this stack needs more padding than defined in the component, e.g. to avoid being drawn in the safe
     /// area when displayed as a sticky footer.
@@ -41,10 +42,12 @@ struct StackComponentView: View {
 
     init(
         viewModel: StackComponentViewModel,
+        isScrollableByDefault: Bool = false,
         onDismiss: @escaping () -> Void,
         additionalPadding: EdgeInsets? = nil
     ) {
         self.viewModel = viewModel
+        self.isScrollableByDefault = isScrollableByDefault
         self.onDismiss = onDismiss
         self.additionalPadding = additionalPadding ?? EdgeInsets(top: 0, leading: 0, bottom: 0, trailing: 0)
     }
@@ -75,6 +78,7 @@ struct StackComponentView: View {
                     viewModels: self.viewModel.viewModels,
                     onDismiss: self.onDismiss
                 )
+                .scrollableIfEnabled(.vertical, enabled: style.scrollable ?? self.isScrollableByDefault)
                 // This alignment positions the inner VStack horizontally and vertically
                 .size(style.size,
                       horizontalAlignment: horizontalAlignment.frameAlignment,
@@ -87,6 +91,7 @@ struct StackComponentView: View {
                     viewModels: self.viewModel.viewModels,
                     onDismiss: self.onDismiss
                 )
+                .scrollableIfEnabled(.horizontal, enabled: style.scrollable ?? self.isScrollableByDefault)
                 // This alignment positions the inner VStack horizontally and vertically
                 .size(style.size,
                       horizontalAlignment: distribution.horizontalFrameAlignment,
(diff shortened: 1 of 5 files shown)


MarkVillacampa:
Code looks good, just curious how this will behave with nested scrollviews 😅

joshdholtz:
> Code looks good, just curious how this will behave with nested scrollviews 😅

@MarkVillacampa Yeaaahhh, I think we prevent nesting in the frontend/backend validation... or at least warn about it 😛 

Initially we will only offer scrolling on horizontal stacks! And them maybe disabling of scrolling on the root stack 😇

RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
```diff
@@ -112,13 +115,52 @@ struct StackComponentView: View {
...
+    }
+
+}
+
+private extension Axis {
+
+    var scrollViewAxis: Axis.Set {
+        switch self {
+        case .horizontal: return .horizontal
+        case .vertical: return .vertical
```
tonidero, line 131: Just to make sure, we won’t allow scrolling in the opposite direction than the axis of the stack then? I think that can make sense for now (in the android PR I was assuming a direction would be given by the backend to make it more flexible… but we might not need that flexibility)

Sources/Paywalls/Components/PaywallStackComponent.swift
```diff
@@ -31,6 +36,7 @@ public extension PaywallComponent {
         public let border: Border?
         public let shadow: Shadow?
         public let badge: Badge?
+        public let overflow: Overflow?
```
tonidero, line 39: I guess this term/API is very iOS-like… but maybe that’s ok? I do wonder if this should be a more generic term for scrolling or not scrolling…

tonidero (commented):
Looks good! Just some questions

Sources/Paywalls/Components/PaywallStackComponent.swift
joshdholtz: @tonidero I got this term from the web and/or design actually 😅 I left this a bit more open/vague incase we ever decided to add a "wrapped" option 

So instead of a horizontal stack scrolling, it would overflow vertically 🤷‍♂️

tonidero: Ah ok sounds good... I can adapt the Android PR to this schema then 👍

tonidero: Adapted [Android PR](https://github.com/RevenueCat/purchases-android/pull/2108) to the new schema

JayShortway: Did we.. did we just create our own version of a stack overflow??? 🥁

RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
joshdholtz: @tonidero Correct correct! That might make it a bit more confusing to use 😅

RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
```diff
@@ -114,13 +117,52 @@ struct StackComponentView: View {
...
 }
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 fileprivate extension View {
 
+    @ViewBuilder
+    // @PublicForExternalTesting
+    func scrollableIfEnabled(
+        _ dimension: PaywallComponent.Dimension,
+        enabled: Bool = true
```
ajpallares, line 146: Perhaps this default value is not needed if we're always passing the parameter

RevenueCatUI/Templates/V2/Components/Stack/StackComponentViewModel.swift
```diff
@@ -167,6 +170,15 @@ struct StackComponentStyle {
                                   stackBorder: self.border,
                                   badgeViewModels: badgeViewModels,
                                   uiConfigProvider: uiConfigProvider)
+
+        self.scrollable = overflow.flatMap({ overflow in
```
ajpallares, line 174: This is just a huge nit, but for me it has been a TIL. I usually use the `map` method for this, and I didn't know the existance of `flatMap` in this context. It turns out that `flatMap` is meant to be used when the closure returns an `Optional`, in order to flatten the 2 optionals. In this case, since the closure returns a nonoptional `Bool`, `map` is more appropriate. Again, feel free to not change this, but I still think this is an interesting caveat 😊 (that I didn't know until now)

ajpallares (approved):
No objections here! 2 small comments

PR #4769, opened 2025-02-06
Fix period abbreviated when multiple days/weeks/months/years

joshdholtz:
⚠️ Dependent on a backend PR to be merged first with the new keys

### Motivation

The period abbreviation when multiple months was rendering as `months` or `mo` instead of `3 months` or `3mo`

### Description

Four new localization keys for abbreviated days/weeks/months/years

### DIFF
--- a/RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift
+++ b/RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift
@@ -389,7 +389,35 @@ extension VariablesV2 {
     }
 
     func productPeriodAbbreviated(package: Package, localizations: [String: String]) -> String {
-        return self.productPeriodAbbreviatedWithUnit(package: package, localizations: localizations)
+        guard let period = package.storeProduct.subscriptionPeriod else {
+            return ""
+        }
+
+        if period.value > 1 {
+            let localizedFormatKey: String
+            switch period.unit {
+            case .day:
+                localizedFormatKey = VariableLocalizationKey.numDaysShort.rawValue
+            case .week:
+                localizedFormatKey = VariableLocalizationKey.numWeeksShort.rawValue
+            case .month:
+                localizedFormatKey = VariableLocalizationKey.numMonthsShort.rawValue
+            case .year:
+                localizedFormatKey = VariableLocalizationKey.numYearsShort.rawValue
+            }
+
+            guard let localizedFormat = localizations[localizedFormatKey] else {
+                return ""
+            }
+            return String(format: localizedFormat, period.value)
+
+        } else {
+            guard let abbreviation = localizations[period.periodAbbreviatedLocalizationKey] else {
+                return ""
+            }
+
+            return abbreviation
+        }
     }
 
     func productPeriodInDays(package: Package) -> String {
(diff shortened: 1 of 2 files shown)


Tests/RevenueCatUITests/PaywallsV2/VariableHandlerV2Tests.swift
```diff
@@ -64,7 +64,58 @@ class VariableHandlerV2Test: TestCase {
             "num_year_two": "%d years",
             "num_year_few": "%d years",
             "num_year_many": "%d years",
-            "num_year_other": "%d years"
+            "num_year_other": "%d years",
+            "num_days_short": "%1$s%2$s",
+            "num_weeks_short": "%1$s%2$s",
+            "num_months_short": "%1$s%2$s",
+            "num_years_short": "%1$s%2$s"
```
joshdholtz, line 71: These are the new values coming from the backend

Even though the first is will be rendered as a integer (like "3 months"), we have to make it a string here because of a Swift bug that can cause a `EXC_BAD_ACCESS` if there are ordered types

I will link to the feedback when its created

Tests/RevenueCatUITests/PaywallsV2/VariableHandlerV2Tests.swift
```diff
@@ -64,7 +64,58 @@ class VariableHandlerV2Test: TestCase {
...
+            "num_year_zero": "%d year",
+            "num_year_one": "%d year",
+            "num_year_two": "%d years",
+            "num_year_few": "%d years",
+            "num_year_many": "%d years",
+            "num_year_other": "%d years",
+            "num_days_short": "%2$s%1$s",
+            "num_weeks_short": "%2$s%1$s",
+            "num_months_short": "%2$s%1$s",
+            "num_years_short": "%2$s%1$s"
```
joshdholtz, line 118: Notice here the `2$` is first

Tests/RevenueCatUITests/PaywallsV2/VariableHandlerV2Tests.swift
```diff
@@ -64,7 +64,58 @@ class VariableHandlerV2Test: TestCase {
...
             "num_year_few": "%d years",
             "num_year_many": "%d years",
-            "num_year_other": "%d years"
+            "num_year_other": "%d years",
+            "num_days_short": "%dd",
+            "num_weeks_short": "%dwk",
+            "num_months_short": "%dmo",
+            "num_years_short": "%dyr"
+        ],
+        "ar": [
```
tonidero, line 73: Hmm FFTI but should the values in this other locale be different than the `en_US` values? I guess this is mostly testing variable handling and not anything else, but might help making sure there isn't an issue during tests... NABD though, happy to leave it as is.

tonidero (approved):
Looks good!

RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift
```diff
@@ -372,15 +381,15 @@ extension VariablesV2 {
...
+        }
     }
 
     func productPeriodAbbreviated(package: Package, localizations: [String: String]) -> String {
-        guard let period = package.storeProduct.subscriptionPeriod else {
-            return ""
-        }
-
-        return localizations[period.periodAbbreviatedLocalizationKey] ?? ""
+        return self.productPeriodAbbreviatedWithUnit(package: package, localizations: localizations)
```
ajpallares, line 392: Just a nit: 
I see `productPeriodAbbreviatedWithUnit(package:,localization:)` is only called from here. Perhaps, to avoid the method jumping, the implementation of `productPeriodAbbreviatedWithUnit` could just go in this method `productPeriodAbbreviated`, saving the extra method `productPeriodAbbreviatedWithUnit`

ajpallares (approved):
LGTM! Just a very small comment

PR #4775, opened 2025-02-10
[Paywalls V2] Navigation bar component

joshdholtz:
## Motivation

_Coming soon_

## Description

_Coming soon_

### Demo

https://github.com/user-attachments/assets/4780aa4c-b46b-42e5-b59c-5fde3cf13492


RevenueCatUI/Templates/V2/Components/ComponentsView.swift
```diff
@@ -62,8 +62,6 @@ struct ComponentsView: View {
                     PackageComponentView(viewModel: viewModel, onDismiss: onDismiss)
                 case .purchaseButton(let viewModel):
                     PurchaseButtonComponentView(viewModel: viewModel)
-                case .stickyFooter(let viewModel):
-                    StickyFooterComponentView(viewModel: viewModel)
```
joshdholtz, line 66: Sticky footer is not a standalone component (not in the component tree) so removed it from here. I did this in this PR since the navigation bar fell in the same category 🤷‍♂️

RevenueCatUI/Templates/V2/ViewModelHelpers/PaywallComponentViewModel.swift
```diff
@@ -28,7 +28,6 @@ enum PaywallComponentViewModel {
     case button(ButtonComponentViewModel)
     case package(PackageComponentViewModel)
     case purchaseButton(PurchaseButtonComponentViewModel)
-    case stickyFooter(StickyFooterComponentViewModel)
```
joshdholtz, line 31: Same as other spot... removed here because sticky footer is never going to be in the normal view stack tree thingy

Sources/Paywalls/Components/PaywallNavigationBarComponent.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+//  Created by Jay Shortway on 24/10/2024.
+//
+// swiftlint:disable missing_docs
+
+import Foundation
+
+public extension PaywallComponent {
+
+    final class NavigationBarComponent: PaywallComponentBase {
```
JayShortway, line 20: Could we name this `StickyHeaderComponent`? Or maybe `HeaderComponent` if it's not sticky? This would be more consistent with our `StickyFooterComponent`, and also avoid confusion on the Android side. 

On Android, this would be a "Top App Bar". The "Navigation Bar" on Android is the bar at the bottom, controlled by the OS, containing the gesture handle thing or back/recents/home buttons.

Sources/Paywalls/Components/PaywallNavigationBarComponent.swift
```diff
@@ -0,0 +1,44 @@
...
+
+        public let leadingStack: PaywallComponent.StackComponent?
+        public let trailingStack: PaywallComponent.StackComponent?
```
JayShortway, line 23: This feels a bit like iOS implementation details are leaking. From a platform agnostic point of view, I would expect that this component is just a single (horizontal) stack.

joshdholtz:
Closing this because not ready anytime soon

PR #4777, opened 2025-02-10
Fix for survey answered event not being tracked when not setting a `customerCenterActionHandler`

vegaro:
If `customerCenterActionHandler` is not set, we are not tracking survey answered submitted events


vegaro:
Going to see if I can write a test for this in an easy way

facumenzella (approved):

facumenzella (approved):

PR #4781, opened 2025-02-12
[Paywalls V2] Add `visible` property to all components (and overrides to ones that were missing)

joshdholtz:
### Motivation

All (or almost all) need the new `visible` property

### Description

_Coming soon_


RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
```diff
@@ -45,21 +45,27 @@ struct ImageComponentView: View {
...
             )
         ) { style in
-            RemoteImage(
-                url: style.url,
-                lowResUrl: style.lowResUrl,
-                darkUrl: style.darkUrl,
-                darkLowResUrl: style.darkLowResUrl
-            ) { (image, size) in
-                self.renderImage(image, size, with: style)
+            Group {
```
tonidero, line 48: I see this `Group` being needed all over the place. Wdyt about creating a custom view that receives the visibility boolean + the swift UI view and is in charge of displaying according to the visibility parameter? Mostly so we don't need both the `Group` + the if-else everywhere.

RevenueCatUI/Templates/V2/Components/Timeline/TimelineComponentView.swift
```diff
@@ -29,14 +42,33 @@ struct TimelineComponentView: View {
...
+            state: self.componentViewState,
+            condition: self.screenCondition,
+            isEligibleForIntroOffer: self.introOfferEligibilityContext.isEligible(
+                package: self.packageContext.package
+            )
+        ) { style in
+            Group {
+                if style.visible {
+                    timeline(style: style)
+                }
```
tonidero, line 55: Hmm do we need to paint an `EmptyView` in this case like the other places? (Or maybe it's not needed, in which case, could we remove it from everywhere else?

tonidero (commented):
Looking good! Just a couple of cleanup things

RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
ajpallares: Even easier, I think, in order to remove the `Group`s, would be to add the `@ViewBuilder` attribute to the `apply` parameters in the view models. E.g., in `IconComponentViewModel.swift`:
```
@ViewBuilder
    func styles(
        state: ComponentViewState,
        condition: ScreenCondition,
        isEligibleForIntroOffer: Bool,
        @ViewBuilder apply: @escaping (IconComponentStyle) -> some View
    )
```

I think it's the same as we want, and the only difference is that without the `@ViewBuilder` attribute we're forced to return 1 `some View`, whereas `@ViewBuilder` allows to provide multiple views (the case of the `if-else` statement).

ajpallares: In fact, with the `@ViewBuilder` attribute, we can even remove the `else` case (removing `EmptyView()`). But that might have other consequences (I believe returning `EmptyView()` in the `else` does not have the same effect as not implementing the `else`).

RevenueCatUI/Templates/V2/ViewHelpers/VisibleIfNeeded.swift
```diff
@@ -0,0 +1,29 @@
...
+
+    let visible: Bool
+    @ViewBuilder var content: () -> Content
+
+    var body: some View {
+        Group {
+            if visible {
+                self.content()
+            }
+        }
```
ajpallares, line 26: Sorry for repeating myself @joshdholtz, but in case you missed my previous content. 
* The `group` is not needed here because `var body: some View` is a `@ViewBuilder`.
* Instead of introducing the `VisibleIfNeeded`, this could be easier solved adding the `@ViewBuilder` attribute to the `apply` parameters of the view models. E.g. 
```
@ViewBuilder
    func styles(
        state: ComponentViewState,
        condition: ScreenCondition,
        isEligibleForIntroOffer: Bool,
        @ViewBuilder apply: @escaping (IconComponentStyle) -> some View
```
I think that approach is cleaner and removes the need of the extra layer complexity layer of `VisibleIfNeeded`, which is what the `if` is for (in `@ViewBuilder`s)

Sources/Paywalls/Components/Common/PaywallComponentBase.swift
```diff
@@ -7,9 +7,9 @@
 // swiftlint:disable missing_docs
 import Foundation
 
-public protocol PaywallComponentBase: Codable, Sendable, Hashable, Equatable { }
+public protocol PaywallComponentBase: Codable, Sendable, Hashable, Equatable {}
 
-public enum PaywallComponent: PaywallComponentBase {
```
JayShortway, line 12: I'm curious about the motivation behind this change. (I have no opinion on it though haha.)

JayShortway (approved):
Looks good! Hiding whitespace on GitHub made reviewing this a breeze.

Sources/Paywalls/Components/Common/PaywallComponentBase.swift
joshdholtz: Ahhhh... at one point I had the `PaywallComponentBase` protocol required a `visible` property but then took it away 😅 

This could use some clean up work in the future anyway 🤷‍♂️

JayShortway: Haha okay okay, thanks for explaining!

PR #4785, opened 2025-02-13
Upgrade RevenueCat SDK to be able to test paywalls v2

nyeu:
Upgrading the SPM version of the sample app to minimum version 5.16.0 to support paywalls v2


ajpallares (approved):
Good. Thanks!

PR #4786, opened 2025-02-13
UI Preview Mode: mock `CustomerInfo`

ajpallares:
### Checklist

### Motivation
In UI Preview mode, the SDK should not make the requests to obtain the `CustomerInfo`.

### Description
When in UI Preview mode, the SDK:
* Avoids caching `CustomerInfo`
* Always returns mocked `CustomerInfo` (if requested either from the cache or from the backend)

### DIFF
--- a/Sources/Identity/CustomerInfoManager.swift
+++ b/Sources/Identity/CustomerInfoManager.swift
@@ -209,6 +209,10 @@ class CustomerInfoManager {
     }
 
     func cache(customerInfo: CustomerInfo, appUserID: String) {
+        guard !self.systemInfo.dangerousSettings.uiPreviewMode else {
+            return
+        }
+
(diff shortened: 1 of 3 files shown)


Sources/Identity/CustomerInfoManager.swift
```diff
@@ -189,6 +189,10 @@ class CustomerInfoManager {
     }
 
     func cachedCustomerInfo(appUserID: String) -> CustomerInfo? {
+        guard !self.systemInfo.dangerousSettings.uiPreviewMode else {
```
tonidero, line 192: Hmm I guess if we don't cache it, this would always be null and this might not be needed... But I guess I'm ok keeping it just in case.

tonidero (approved):
Makes sense! :shipit:

PR #4787, opened 2025-02-14
[Paywalls V2] Add purchase button activity indicator

joshdholtz:
### Motivation

Need a way to show when there is activity happening with a purchase button

### Description

https://github.com/user-attachments/assets/b4b682f6-63fd-4248-9564-f8a3ecc84838


RevenueCatUI/Templates/V2/ViewHelpers/ProgressViewModifier.swift
```diff
@@ -0,0 +1,64 @@
...
+    @ViewBuilder
+    private var progressView: some View {
+        switch backgroundStyle {
+        case .color(let displayableColorScheme):
+            let colorInfo = displayableColorScheme.effectiveColor(for: colorScheme)
+            ProgressView()
+                .progressViewStyle(CircularProgressViewStyle(tint: bestTint(for: colorInfo)))
+        case .image, .none:
+            ProgressView()
+                .progressViewStyle(CircularProgressViewStyle(tint: .white))
```
ajpallares, line 39: Not sure if the `tint` here should depend on the `colorScheme` and/or on something else. You know, to prevent showing the while progress view on a white background. But then, I don't even know if this case is even possible (looking at the Dashboard, it seems like Buttons only support color-based backgrounds)

ajpallares (approved):
Good! Just a couple of small comments

RevenueCatUI/Templates/V2/ViewHelpers/ProgressViewModifier.swift
JayShortway: Yea I think the best version would be to base the tint on the image pixels underneath the circular progress view. But that can also come in a next iteration imo.

JayShortway (approved):
Fancy!

RevenueCatUI/Templates/V2/ViewHelpers/ProgressViewModifier.swift
JayShortway: Or actually, @joshdholtz are there any plans on making the loading indicator have a configurable color on the dashboard? That would avoid the need for us to be smart about it.

ajpallares: If not, another way of trying to be smart about it could be to try and use the same color as the (first?) text component in the button. But I'm sure there are edge cases for this (e.g. no text or multiple text components). Just putting ideas out there

RevenueCatUI/Templates/V2/ViewHelpers/ProgressViewModifier.swift
```diff
@@ -0,0 +1,64 @@
...
+    private func bestTint(for colorInfo: DisplayableColorInfo) -> Color {
+        switch colorInfo {
+        case .hex:
+            return colorInfo.toColor(fallback: .black).brightness() > 0.6 ? .black : .white
+        case .linear, .radial:
+            let gradient = colorInfo.toGradient()
+            let averageBrightness = gradient.stops
+                .compactMap { $0.color.brightness() }
+                .reduce(0, +) / CGFloat(gradient.stops.count)
+            return averageBrightness > 0.6 ? .black : .white
```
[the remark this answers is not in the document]
joshdholtz: I would love to be smarter about this in the future 😅 It's not perfect but its something 🤷‍♂️

RevenueCatUI/Templates/V2/ViewHelpers/ProgressViewModifier.swift
joshdholtz: > are there any plans on making the loading indicator have a configurable color

@JayShortway I didn't _want_ to because its not something our users _should_ have to think about but I'm not opposed if we make it an optional style at some point

> Not sure if the tint here should depend on the colorScheme and/or on something else

@antoniobg The calculated tint does take the color scheme into effect so we should be good!

joshdholtz: Also, I added a blur to the background it so it should make it even easier for either tint color to be visible 🤞

RevenueCatUI/Templates/V2/ViewHelpers/ProgressViewModifier.swift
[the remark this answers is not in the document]
ajpallares: Totally! :shipit: 🚀

MortenGregersen:
Love it! 🎉

PR #4788, opened 2025-02-14
[Paywalls V2] Scroll fix for background/padding/border

joshdholtz:
### Motivation

A fix to scrolling stacks for background, border, and padding

### Description

I forgot to commit this with the previous PR (#4767) 🤦‍♂️


tonidero (approved):
So this makes it so only the content inside the border is scrollable correct? Makes sense 👍

PR #4789, opened 2025-02-14
[Paywalls] Use CSS linear-gradient spec to compute gradient start/end points

MarkVillacampa:
Getting linear gradient behaviour in line with dashboard and android.

Here's what Dashboard - iOS - Android look like after the changes from this PR:

Screenshot 2025-02-14 at 16 16 10 (1)

### DIFF
--- a/RevenueCatUI/Templates/V2/ViewHelpers/Fill.swift
+++ b/RevenueCatUI/Templates/V2/ViewHelpers/Fill.swift
@@ -56,4 +56,28 @@ extension Shape {
     }
 }
 
+extension UnitPoint {
+
+    init(angle: Angle) {
+        // Convert the angle to radians and negate to make clockwise
+        // Subtract π/2 (90 degrees) to place an angle of 0 degrees at the top
+        let radians = -angle.radians - (.pi / 2)
+
+        // Calculate the normalized x and y positions
+        let xPosition = cos(radians)
+        let yPosition = sin(radians)
+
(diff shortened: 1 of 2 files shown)


RevenueCatUI/Templates/V2/ViewHelpers/Fill.swift
```diff
@@ -56,4 +56,28 @@ extension Shape {
     }
 }
 
+extension UnitPoint {
+
+    init(angle: Angle) {
```
MarkVillacampa, line 61: We're still using the old way of calculating the gradient but only for the Timeline Component connectors because these use a different way of configuring the gradient using a Shape where we don't have the size of the rectangle precomputed. This shouldn't be a problem since connectors will usually have 0 or 90 degrees most of the time.

RevenueCatUI/Views/GradientView.swift
```diff
@@ -39,14 +39,56 @@ struct GradientView: View {
         }
     }
 
+    // Calculate the start and end points of the gradient following the CSS linear-gradient spec
+    // https://developer.mozilla.org/en-US/docs/Web/CSS/gradient/linear-gradient
+
+    // Heavily inspired by this blog post by Mukhtar Bimurat
+    // https://link.medium.com/LczMO6j8YQb
```
JayShortway, line 46: Awesome job using the math from that post! 🙌

JayShortway (approved):
Amazing! 💪

PR #4793, opened 2025-02-18
Fix paywall refreshable bug

ajpallares:
### Motivation
If presenting the `PaywallView` from a view that uses `refreshable`, then the `PaywallView` inherits the `refreshable` action, meaning that the refreshing indicator appears when pulling down on the first scroll view in the paywall's view hierarchy, executing the refresh action of the parent (🤯). This happens when presenting the paywall in any mode: `sheet`, `fullScreenCover` or navigation push.

### Description
This PR fixes this bug by disabling the `refreshable` action in the `PaywallView`

### How I tested
I added a new `refreshable` action [to the paywalls list in the PaywallsTester app](https://github.com/RevenueCat/purchases-ios/pull/4793/files#:~:text=.-,refreshable,-%7B). Without the fix, opening a paywall that has a scroll view in its view hierarchy would execute the `refreshable` action in the paywall.
Simulator Screen Recording - iPhone 16 Pro - 2025-02-18 at 10 09 59


Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/APIKeyDashboardList.swift
```diff
@@ -59,6 +59,9 @@ struct APIKeyDashboardList: View {
         .task {
             await fetchOfferings()
         }
+        .refreshable {
+            await fetchOfferings()
+        }
```
JayShortway, line 64: What do you think about a comment explaining why we have this?

JayShortway (approved):
Good find! 🙌

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/APIKeyDashboardList.swift
ajpallares: Oh sure! I didn't pay too much attention to the PaywallsTester app but yeah, it is useful to add some context 👍 Thanks!

JayShortway: Very nice, thanks!

JayShortway (approved):

RevenueCatUI/Modifiers/ViewExtensions.swift
```diff
@@ -343,6 +343,22 @@ private struct RoundedCorner: Shape {
...
+    @ViewBuilder
+    func refreshableDisabled() -> some View {
+        if let refreshKeyPath = \EnvironmentValues.refresh as? WritableKeyPath<EnvironmentValues, RefreshAction?> {
+            self.environment(refreshKeyPath, nil)
+        } else {
+            self
+        }
+    }
+
+}
```
facumenzella, line 360: I think this only works if the refreshable is applied before this, right?

ajpallares: Right. But that's the use case we need. To disable any `refreshable` that the `PaywallView` might inherit from the caller

PR #4794, opened 2025-02-18
fix: Remove force unwrapping from PurchaseHistoryView

facumenzella:
### Motivation
While debugging another thing, I noticed this was in production. This removes the force unwrap

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -10,7 +10,6 @@ excluded:
   - .build
 
 opt_in_rules:
-  - force_unwrapping
   - sorted_imports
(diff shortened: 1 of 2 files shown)


vegaro (approved):

vegaro:
I would probably introduce the rule in another PR, I see it's failing things

.swiftlint.yml
```diff
@@ -10,6 +10,7 @@ excluded:
   - .build
 
 opt_in_rules:
+  - force_unwrapping
```
facumenzella, line 13:
```
```

PR #4795, opened 2025-02-18
Use RC API key for local development from local.xcconfig

ajpallares:
### Motivation
When using the testing apps for developing the SDK, we usually need to add our own RevenueCat API key and/or Proxy URL. However, before this PR:
* Each testing app had a different place where the API key and Proxy URL needed to be added
* We often committed, by mistake, the changes that contained the API keys, which was not critical, but still undesired

### Description
This PR addresses the two pain points mentioned above when setting an API key for local development (or Proxy URL):
* Centralizes the place where the RevenueCat API key and Proxy URL are set for local development in the `local.xcconfig` file
* Because the `local.xcconfig` file is ignored by git in this repo, we can finally stop worrying about committing the API key by mistake

### Notes
* The implementation for this relies on adding the local API key and Proxy URL to the info.plist of the projects, to make them available for the code to access.
* The CI for PaywallsTester app still relies on the existing implementation of replacing the value of the `apiKey` `String` in `ConfigItem.swift`


Local.xcconfig.SAMPLE
```diff
@@ -27,5 +27,11 @@
 // as well as by Package.swift.
 // ======
 
-// Uncomment to enable the PAYWALL_COMPONENTS flag.
-// SWIFT_ACTIVE_COMPILATION_CONDITIONS = $(inherited) PAYWALL_COMPONENTS
+// Add new compiler flags like with:
```
JayShortway, line 30:
```suggestion
// Add new compiler flags like this:
```
(This doesn't always work unfortunately, probably due to SPM caching.)

Local.xcconfig.SAMPLE
```diff
@@ -27,5 +27,11 @@
...
+// SWIFT_ACTIVE_COMPILATION_CONDITIONS = $(inherited) NEW_FLAG
+
+// Set here your RevenueCat API key for local development with:
```
JayShortway, line 33:
```suggestion
// Set your RevenueCat API key for local development here:
```

Local.xcconfig.SAMPLE
```diff
@@ -27,5 +27,11 @@
...
+// REVENUECAT_API_KEY = your-api-key
+
+// Set here your RevenueCat Proxy URL for local development with:
```
JayShortway, line 36:
```suggestion
// Set your RevenueCat Proxy URL for local development here:
```

scripts/pre-commit.sh
```diff
@@ -38,7 +38,6 @@ verify_no_included_apikeys() {
     "${SCRIPT_DIR}/../Examples/MagicWeather/MagicWeather/Constants.swift"
     "${SCRIPT_DIR}/../Examples/MagicWeatherSwiftUI/Shared/Constants.swift"
     "${SCRIPT_DIR}/../Tests/TestingApps/PurchaseTesterSwiftUI/Core/Constants.swift"
-    "${SCRIPT_DIR}/../Tests/TestingApps/PaywallsTester/PaywallsTester/Config/LocalConfigItems.swift"
```
JayShortway, line 41: Can we not get rid of all of these files, so we can get rid of the entire `verify_no_included_apikeys()` check?

Tests/TestingApps/PaywallsTester/PaywallsTester/Config/ConfigItem.swift
```diff
@@ -7,17 +7,18 @@
...
-
-// CI system adds keys here
-extension AvailableConfigItems {
-    static var apiKey: String { "" }
-    static var proxyURL: String? { nil }
-}
-
-struct ConfigItem: AvailableConfigItems {
+// DO NOT MODIFY THIS FILE.
+// CI system adds the API key here.
```
JayShortway, line 11: What do you think about removing this file and making [the CI script](https://github.com/RevenueCat/purchases-ios/blob/main/Tests/TestingApps/PaywallsTester/ci_scripts/ci_pre_xcodebuild.sh) create a `Local.xcconfig` file instead?

Tests/TestingApps/PaywallsTester/PaywallsTester/Config/LocalConfigItems.swift
```diff
@@ -1,31 +0,0 @@
-//
```
JayShortway, line 1: (This comment has nothing to do with the line it's attached to 😅)

We also have a [`:replace_api_key_integration_tests`](https://github.com/RevenueCat/purchases-ios/blob/main/fastlane/Fastfile#L322) Fastlane lane. We _could_ refactor that to write values to `Local.xcconfig` as well, instead of to the various files listed in that lane. Just a suggestion and definitely not a blocker for this PR, as that is not really a pain point for local development.

JayShortway (approved):
Yessss omg thank you! This was one of my main motivations for adding the `Local.xcconfig` file, but I hadn't gotten around to it yet. ❤️ 

I have some suggestions to take it even further haha.

Tests/TestingApps/PaywallsTester/PaywallsTester/Config/ConfigItem.swift
ajpallares: Well, this file is still needed for the PaywallsTester app. I mean, not **needed** per se, but it's good to have a centralized place where we get the necessary variables from the info.plist.

Regarding having the CI script create a Local.xcconfig, I think that could totally work, yes. I'll open a separate PR for the CI changes related to this one 👍

Tests/TestingApps/PaywallsTester/PaywallsTester/Config/LocalConfigItems.swift
ajpallares: I see. But it would definitely be much cleaner. I'll do this in a separate PR for CI-related stuff. Thank you for the suggestions!!!

scripts/pre-commit.sh
ajpallares: Well I see the utility of having then in the sense that someone might forget about the `local.xcconfig` file thing and perhaps wants to do a quick change of the API key and they do it in the code directly. So it does not harm. That said, by that argument, I'd need to add the `ConfigItem.swift` file to it 😄

tonidero (approved):
This is great! 🫶

Tests/TestingApps/PaywallsTester/PaywallsTester/Config/ConfigItem.swift
```diff
@@ -7,17 +7,18 @@
...
+struct ConfigItem {
```
facumenzella, line 12: nit

```suggestion
enum ConfigItem {
```

facumenzella (approved):
lovely

PR #4796, opened 2025-02-18
Use RC API key in Examples from local.xcconfig for local development

ajpallares:
### Motivation
This continues the work in #4795, to have all the constants for local development centralized in `local.xcconfig`, which is not committed to git. The motivation is to take the same approach in the example projects (e.g. MagicWeather) while allowing integrators to quickly change the constants directly in code for convenience.

### Description
This PR adds the existing `Global.xcconfig` file and the reference to the same `Local.xcconfig` file to the projects in the `/Examples` directory, defining the variables:
* `EXAMPLE_APP_API_KEY` for the RC API key
* `EXAMPLE_APP_ENTITLEMENT_ID` for the Entitlement ID

### Notes
* I updated the Readme of MagicWeather accordingly.
* Before this PR, we had a `#error` directive to force everyone to add their own API keys in the `Constants.swift` file. Now, because it relies in the `Local.xcconfig` by default, I've added a `fatalError` if any of the required values is not set. Is not as nice as having an error in compile time, but at least the error will appear immediately in runtime (because the variables are defined as constants). I'm open to suggestions to improve this!
* Not very important, but it made sense to me to use a different environment variable for the API key in the examples (`EXAMPLE_APP_API_KEY`) with respect to the API key in the internal SDK projects like PaywallsTester (`REVENUECAT_API_KEY`).


Examples/MagicWeather/MagicWeather/Constants.swift
```diff
@@ -16,13 +16,23 @@ struct Constants {
     /*
      The API key for your app from the RevenueCat dashboard: https://app.revenuecat.com
      */
-    #error("Modify this property to reflect your app's API key, then comment this line out.")
-    static let apiKey = "REVENUECAT_API_KEY"
-    
+    static let apiKey: String = {
+        // return "your-api-key"
+        guard let key = Bundle.main.object(forInfoDictionaryKey: "EXAMPLE_APP_API_KEY") as? String, !key.isEmpty else {
```
tonidero, line 21: I believe there was some discussion about this already, but I wonder if it would be better for the sample app to have the key hardcoded in code instead of obtained from the bundle, to follow best practices... But as discussed, maybe it's ok since this is after all a "public" api key.

tonidero (approved):
Just a comment, but I think we can still go ahead with this in any case!

Examples/MagicWeather/MagicWeather/Constants.swift
ajpallares: Fair point, yes.
This is why I left the commented out code
```swift
// return "your-api-key"
```
And also why I separated this from #4795, because it's the same idea but with the caveat of being the example project for integrators of the SDK.
I really don't have a strong opinion about this one, but this is worth discussing, yes.

facumenzella (approved):

ajpallares:
After talking about this in a meeting, we decided not to merge this PR so that the examples remain with this
```swift
#error("Modify this property to reflect your app's API key, then comment this line out.")
static let apiKey = "REVENUECAT_API_KEY"
```
instead of 
```swift
static let apiKey: String = {
    // return "your-api-key"
    guard let key = Bundle.main.object(forInfoDictionaryKey: "EXAMPLE_APP_API_KEY") as? String, !key.isEmpty else {
        fatalError("Modify this property to reflect your app's API key")
    }
    return key
}()
```

The main arguments for this were:
* The public examples should follow the recommended and easier way to add the public API key: hardcode the public API key in code.
* It's much more straighforward to remove the `#error` and replace the `REVENUECAT_API_KEY` string with your own API key than doing so with the proposed implementation.
* The `#error` directive is clear and **fails at build time**, whereas the proposed approach used `fatalError`, which is triggered at runtime.

PR #4797, opened 2025-02-18
[Paywalls] Add support for shadows in image components

MarkVillacampa:


RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
```diff
@@ -58,6 +58,8 @@ struct ImageComponentView: View {
                 .clipped()
                 .shape(border: nil,
                        shape: style.shape)
+                .shadow(shadow: style.shadow,
+                        shape: style.shape?.toInsettableShape())
                 .padding(style.padding)
                 // WIP: Add border still
```
tonidero, line 64: Do we also need to add the border while at it? Though it can be a separate PR.

tonidero (approved):
Looks good! Just a comment about supporting borders as well.

RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
MarkVillacampa: Right! didnt know it was possible, will add it

JayShortway: Just FYI, here's how overlays and borders should look if both are set: [MON-905#comment-12116d55](https://linear.app/revenuecat/issue/MON-905/when-applying-overlay-and-border-to-an-image-dashboard-differs-from#comment-12116d55).

PR #4798, opened 2025-02-19
Config item rename

ajpallares:
### Motivation
As pointed out by @facumenzella, for consistency with other projects and repos, the `Constants` name should be used instead of `ConfigItem`. `ConfigItem` contains the code that reads the RC API key from the bundle in PaywallsTester (see #4795)

### Description
* Renames `ConfigItem` to `Constants`
* Bonus: adds a more detailed explanation to `refreshableDisabled()` modifier introduced in #4793 (again, thanks @facumenzella for pointing this out!)


facumenzella (approved):

tonidero (approved):
Makes sense :shipit:

PR #4799, opened 2025-02-19
[Diagnostics] Fix store kit error description tracking

tonidero:
### Description
Noticed that obtaining the store kit error description actually didn't try to get the store kit error when it was nested. I was able to reproduce in an existing test (by adding a new assertion), which gets fixed with the changes in this PR.

### DIFF
--- a/Sources/Error Handling/StoreKitErrorHelper.swift
+++ b/Sources/Error Handling/StoreKitErrorHelper.swift
@@ -27,9 +27,9 @@ enum StoreKitErrorUtils {
             return storeKitError.trackingDescription
         } else if let storeKitError = underlyingError as? StoreKit.Product.PurchaseError {
             return storeKitError.trackingDescription
+        } else {
+            return Self.extractStoreKitErrorDescription(from: underlyingError)
         }
-
-        return nil
     }
 
 }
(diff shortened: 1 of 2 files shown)


Sources/Error Handling/StoreKitErrorHelper.swift
```diff
@@ -27,9 +27,9 @@ enum StoreKitErrorUtils {
             return storeKitError.trackingDescription
         } else if let storeKitError = underlyingError as? StoreKit.Product.PurchaseError {
             return storeKitError.trackingDescription
+        } else {
+            return Self.extractStoreKitErrorDescription(from: underlyingError)
```
tonidero, line 31: In case the current underlying error is not a store kit error, we try to get the underlying error for the current error, if it exists. I don't think this should result in an infinite loop, as long as we don't have a circular reference in the underlying errors.

Tests/StoreKitUnitTests/ProductsManagerTests.swift
```diff
@@ -222,6 +223,7 @@ class SK2ProductsManagerDiagnosticsTrackingTests: ProductsManagerTests {
         expect(params?.storeKitVersion) == .storeKit2
         expect(params?.errorMessage) == "Products request error: Unable to Complete Request"
         expect(params?.errorCode) == 2
+        expect(params?.storeKitErrorDescription) == StoreKitError.unknown.trackingDescription
```
tonidero, line 226: This failed without adding the new code.

Sources/Error Handling/StoreKitErrorHelper.swift
ajpallares: > I don't think this should result in an infinite loop

Yeah, I agree. But still, this scares me a bit. However, I honestly can't think of a way to check this without overcomplicating things a lot 🤔

ajpallares (approved):
Makes sense!
Only one comment about the infinite loop

tonidero:
Will merge this for now, since I think we're safe for this change

PR #4800, opened 2025-02-19
UI Preview mode: avoid intro eligibility request

ajpallares:
### Checklist

### Motivation
In UI Preview mode, the SDK should only make the requests strictly required for the UI Preview mode functionality.

### Description
This PR just implements an early return with all `IntroEligibility` statuses set to `unknown`.

### DIFF
--- a/Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
+++ b/Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
@@ -54,6 +54,11 @@ class TrialOrIntroPriceEligibilityChecker: TrialOrIntroPriceEligibilityCheckerTy
 
     func checkEligibility(productIdentifiers: Set<String>,
                           completion: @escaping ReceiveIntroEligibilityBlock) {
+        guard !self.systemInfo.dangerousSettings.uiPreviewMode else {
+            completion([:])
+            return
+        }
+        
         guard !productIdentifiers.isEmpty else {
             Logger.warn(Strings.eligibility.check_eligibility_no_identifiers)
             completion([:])
--- a/Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
+++ b/Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
@@ -58,7 +58,7 @@ class TrialOrIntroPriceEligibilityChecker: TrialOrIntroPriceEligibilityCheckerTy
             completion([:])
             return
         }
-        
+
         guard !productIdentifiers.isEmpty else {
             Logger.warn(Strings.eligibility.check_eligibility_no_identifiers)
             completion([:])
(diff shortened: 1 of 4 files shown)


Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
```diff
@@ -54,6 +54,11 @@ class TrialOrIntroPriceEligibilityChecker: TrialOrIntroPriceEligibilityCheckerTy
 
     func checkEligibility(productIdentifiers: Set<String>,
                           completion: @escaping ReceiveIntroEligibilityBlock) {
+        guard !self.systemInfo.dangerousSettings.uiPreviewMode else {
+            completion([:])
```
tonidero, line 58: Hmm I wonder if it makes sense to return the requested product ids here with ineligible or something like that... I'm not sure if there would be any issues though.

tonidero (commented):
Just a comment about the behavior... Lmk what you think!

Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
JayShortway: I like this. I could also allow us to specify "fake" eligibility in the future. That way we can preview paywalls in different scenarios (eligible for intro / not eligible for intro).

ajpallares: Thank you both for your comments!

I _think_ I know what you both mean. 

My approach was to keep it as simple as possible to limit potential requests made to the backend in UI preview mode **by mistake**.

We will definitely need to add some way to "fake" eligibility in the future. I honestly don't know how yet. Probably the "fake" eligibility for the products will need to be injected somehow to the SDK exclusively for UI preview mode. So I don't know if that'll be done at this code location or elsewhere.

In my opinion, for this PR, we should not think about faking eligibility as it will not be used (and we don't really know how fake eligibility will look). The only purpose of this PR is to prevent the backend call, so I think this solution works for that and is simple enough.

That said, I'm probably missing something. So, happy to discuss this further!

tonidero: Right, but I think the API should normally return a dictionary with all the requested product ids and their eligibility status. Right now, this will return an empty dictionary, which basically means it's behaving differently. I think this is probably ok, but I think it would be good that the API behaves similarly in uiPreview mode in this case (if this API supported errors, this would definitely be an error I think but since that's not supported, we should get as close as a "normal" response as possible I think).

ajpallares: That's a good point, yes. So what you mean is that the default value in this case should not be an empty dictionary but rather a dictionary with all the products IDs with not eligible for intro. I get it now. Thanks!

Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
```diff
@@ -54,6 +54,14 @@ class TrialOrIntroPriceEligibilityChecker: TrialOrIntroPriceEligibilityCheckerTy
 
     func checkEligibility(productIdentifiers: Set<String>,
                           completion: @escaping ReceiveIntroEligibilityBlock) {
+        guard !self.systemInfo.dangerousSettings.uiPreviewMode else {
+            let result = productIdentifiers.reduce(into: [:]) { resultDict, productId in
+                resultDict[productId] = IntroEligibility(eligibilityStatus: IntroEligibilityStatus.unknown)
```
tonidero, line 59: Could be worth adding a comment... but NABD

tonidero (approved):
:shipit:

Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
ajpallares: Done!

PR #4801, opened 2025-02-19
[Paywalls V2] Fixes parsing generic fonts.

JayShortway:
## Bug
Generic fonts (`sans-serif`, `serif` and `monospace`) weren't displaying at all.

## Cause
This was caused by the fact that the font _display_ name was matched against the generic font, instead of the actual font name. 

## Fix
The fix is to look up the font name using its display name in `ui_config`, like we do for other custom fonts.

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Text/TextComponentViewModel.swift
+++ b/RevenueCatUI/Templates/V2/Components/Text/TextComponentViewModel.swift
@@ -286,16 +286,11 @@ extension TextComponentStyle {
             return GenericFont.sansSerif.makeFont(fontSize: fontSize)
         }
 
-        // Check if name is a generic font (serfic, sansserif, monospace)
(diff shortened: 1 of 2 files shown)


RevenueCatUI/Templates/V2/Components/Text/TextComponentViewModel.swift
```diff
@@ -305,6 +300,11 @@ extension TextComponentStyle {
             return nil
         }
 
+        // Check if the family name is a generic font (serif, sans-serif, monospace)
+        if let genericFont = GenericFont(rawValue: familyName) {
+            return genericFont.makeFont(fontSize: fontSize)
+        }
```
JayShortway, line 306: This is the fix.

ajpallares (approved):
Nice one!

tonidero (approved):
Thanks for fixing!!

PR #4804, opened 2025-02-20
UI preview mode: disable log in and log out

ajpallares:
### Checklist

### Motivation
Even though no calls to `logIn` or `logOut` should happen in UI preview mode, it's still appropriate to make sure that the SDK properly controls that scenario.

### Description
* With this PR, the SDK implements early returns in the `IdentityManager` for the `logIn` and `logOut` methods, preventing any logic in this scenario (including any log in request to the backend)
* Calling `logIn` using the app user ID reserved for UI preview mode also fails, **even if UI preview mode is not enabled**

### DIFF
--- a/Sources/Error Handling/BackendError.swift
+++ b/Sources/Error Handling/BackendError.swift
@@ -29,6 +29,7 @@ enum BackendError: Error, Equatable {
     case invalidWebRedemptionToken
     case purchaseBelongsToOtherUser
     case expiredWebRedemptionToken(obfuscatedEmail: String)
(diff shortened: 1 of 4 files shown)


tonidero (approved):
Makes sense! I guess we have a lot of APIs we might want to disable... But this is a good start!

ajpallares:
> Makes sense! I guess we have a lot of APIs we might want to disable... But this is a good start!

Exactly! I'm going one by one making sure that they are disabled in UI preview mode (even if they are not meant to be used while in UI preview mode). But #BetterSafeThanSorry 😄

PR #4805, opened 2025-02-20
feat: Don't show refund if free subscription

facumenzella:
### Motivation
We don't want to offer the refund option if the customer hasn't spend any money.

### Description
- Update snaps with free subscription
- Add test
- Add filter conditions


RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptions/ManageSubscriptionsViewModel.swift
```diff
@@ -263,10 +263,16 @@ private extension Array<CustomerCenterConfigData.HelpPath> {
...
+        return filter {
+            // if it's cancel, it cannot be a lifetime subscription
+            ($0.type != .cancel || !purchaseInformation.isLifetime) &&
+
+            // if it's refundRequest, it cannot be free
+            ($0.type != .refundRequest || purchaseInformation.price != .free) &&
+
+            // if it has a refundDuration, check it's still valid
+            ($0.refundWindowDuration?.isWithin(purchaseInformation) ?? true)
+        }
```
vegaro, line 275: Thanks for adding the comments. It's also much clearer checking the type first. I would have done positive checks instead of negative conditions, but it's a personal preference

vegaro (approved):

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptions/ManageSubscriptionsViewModel.swift
facumenzella: I think it might be a good idea to tweak it if this keeps growing. I'll keep it in mind :)

PR #4807, opened 2025-02-20
fix: [AppUpdateWarningView] Tweak buttons bottom alignment and padding

facumenzella:
### Motivation
While working on some padding, I went one step further  and aligned the buttons to the bottom.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
@@ -40,30 +40,43 @@ struct AppUpdateWarningView: View {
     }
 
     var body: some View {
-        List {
-            Section {
-                CompatibilityContentUnavailableView(
-                    localization[.updateWarningTitle],
-                    systemImage: "arrow.up.circle.fill",
-                    description: Text(localization[.updateWarningDescription])
-                )
-            }
+        Color(UIColor.systemGroupedBackground)
+            .ignoresSafeArea()
+            .overlay {
+                VStack(alignment: .center, content: {
+                    CompatibilityContentUnavailableView(
+                        localization[.updateWarningTitle],
+                        systemImage: "arrow.up.circle.fill",
+                        description: Text(localization[.updateWarningDescription])
+                    )
+                    .padding(.horizontal, 16)
+                    .padding(.vertical, 24)
+                    .fixedSize(horizontal: false, vertical: true)
+                    .background(
+                        RoundedRectangle(cornerRadius: 8)
+                            .fill(.white)
+                            .padding(.horizontal, 24)
+                            .padding(.top, 24)
+                    )
 
-            Section {
-                Button(localization[.updateWarningUpdate]) {
-                    onUpdateAppClick()
-                }
-                .buttonStyle(ProminentButtonStyle())
-                .padding(.top, 4)
+                    Spacer()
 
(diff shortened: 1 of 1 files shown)


ajpallares (approved):
Much better like this 👍

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
```diff
@@ -40,30 +40,47 @@ struct AppUpdateWarningView: View {
...
 
     var body: some View {
-        List {
-            Section {
-                CompatibilityContentUnavailableView(
-                    localization[.updateWarningTitle],
-                    systemImage: "arrow.up.circle.fill",
-                    description: Text(localization[.updateWarningDescription])
-                )
+        Color(colorScheme == .light ? UIColor.secondarySystemBackground : UIColor.systemBackground)
```
tonidero, line 43: Can this color be extracted since it's used in 2 different places? Also, I wonder if we should just default to default colors, but not sure how that would look like. I'm ok with this change!

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
```diff
@@ -40,30 +40,47 @@ struct AppUpdateWarningView: View {
...
+            .ignoresSafeArea()
```
tonidero, line 44: So I guess this works because we are adding the vertical padding right? Should we instead rely on the safe areas (and add any extra padding we need)?

tonidero (approved):
This makes sense! Just a couple questions since I'm not too familiarized with SwiftUI :)

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
facumenzella: I tried not doing this but couldn't find a combination that looked like the previous version. 🤔

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
facumenzella: `ignoresSafeArea` is for the color to go all the way to the background

```swift
                    CompatibilityContentUnavailableView(
                        localization[.updateWarningTitle],
                        systemImage: "arrow.up.circle.fill",
                        description: Text(localization[.updateWarningDescription])
                    )
                    .padding(.horizontal, 16)
                    .padding(.vertical, 24)
```

The extra padding was to match more or less what we had before. The list has some built in spaces that are lost now because there's no list anymore

tonidero: Ahhh good point! Thanks for explaining 👍

RevenueCatUI/CustomerCenter/Views/AppUpdateWarningView.swift
ajpallares: Yeah, I also thought about this. But note that the two places where it's used the colors are inverted. So it's not exactly the same.

My first though was that this felt unnecessary and that it should suffice using `UIColor.secondarySystemBackground` in one place and `UIColor.systemBackground` in the other one.

But then I checked the `Settings` app and it turns out that it uses the exact same combination of colors in light and dark mode, so this is the right way to go, IMO.

JayShortway (approved):
Love it!

PR #4808, opened 2025-02-20
Add no quotes hints to build settings in `Local.xcconfig.SAMPLE`

ajpallares:
Greatly improves #4795

### Motivation
Using double quotes for strings is set in stone for us, developers. But the strings in `xcconfig`'s files for Build Settings don't use quotes. This PR tries to break that habit only for the `Local.xcconfig` file.

### Description
Add some hints in `Local.xcconfig.SAMPLE` to prevent adding quotes in the Build settings specified in `Local.xcconfig`

### DIFF
--- a/Local.xcconfig.SAMPLE
+++ b/Local.xcconfig.SAMPLE
@@ -30,8 +30,8 @@
 // Add new compiler flags like this:
 // SWIFT_ACTIVE_COMPILATION_CONDITIONS = $(inherited) NEW_FLAG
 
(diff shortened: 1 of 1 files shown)


tonidero (approved):

aboedo (approved):
still laughing at myself for not understanding why it wouldn't work

facumenzella (approved):
Let's hope people read the docs now 😂😂

facumenzella (approved):

PR #4809, opened 2025-02-21
UI preview mode: disable cache updates

ajpallares:
### Checklist

### Motivation
In UI preview mode, there's no need to update the any caches, as the required data is always obtained on demand (e.g. Offerings).

### Description
This PR disables the caches updates that currently happen when:
* When the SDK is initialized while the app is in the foreground (`Purchases.updateCachesIfInForeground()`)
* The app enters in the foreground (`Purchases.applicationWillEnterForeground()`)


tonidero (approved):
Looks good!

PR #4810, opened 2025-02-21
fix: workaround to allow using proxy URL in `Local.xcconfig`

ajpallares:
Implements a workaround to allow specifying the proxy URL for the PaywallsTester app in the `Local.xcconfig` file (see also #4795)

### Motivation
Because two slashes `//` are interpreted as comments in xcconfig files, there's no clean way to specify a URL with the scheme in such file.

### Description
With this PR we define two different variables for the proxyURL scheme and host.
The idea is that the proxy URL must now be specified with
```
REVENUECAT_PROXY_URL_SCHEME = http
REVENUECAT_PROXY_URL_HOST = localhost:9000
```


facumenzella (approved):

vegaro (approved):
makes sense

PR #4811, opened 2025-02-21
[Paywalls] Fix badge background not using the new background field instead of ba…

MarkVillacampa:
…ckgroundColor

### Checklist


RevenueCatUI/Templates/V2/ViewHelpers/BadgeModifier.swift
```diff
@@ -34,6 +34,7 @@ struct BadgeModifier: ViewModifier {
 
         var backgroundStyle: BackgroundStyle? {
             stack.backgroundColor?.asDisplayable(uiConfigProvider: uiConfigProvider).backgroundStyle
+                ?? stack.background?.asDisplayable(uiConfigProvider: uiConfigProvider).backgroundStyle
```
tonidero, line 37: I believe this should be the other way around, first use the background property if it exists, if not, fallback to the backgroundColor

tonidero (commented):
Just one thing. Thanks for fixing!

tonidero (approved):
Just that comment but the situation of both properties appearing simultaneously would be unlikely (maybe for some older paywalls) so :shipit:

RevenueCatUI/Templates/V2/ViewHelpers/BadgeModifier.swift
MarkVillacampa: you're absolutely right!

ajpallares (approved):

Issue #4813, opened 2025-02-22
SubscriptionInfo collides with a new StoreKit object also named SubscriptionInfo (compiler error in Xcode 16.3 beta)


theoks:
In iOS 18.4 StoreKit introduced a new object named `SubscriptionInfo`. There is already a `SubscriptionInfo` object in the `RevenueCat` package, which caused a namespace collision when building with Xcode 16.3 (File `PurchaseInformation.swift`, line 327).

Using the package name explicitly, solves the issue.

**Package version: 5.17.0**

PS: It was an easy fix and I tried to create a PR but a) I couldn't push due to authorization issues, and b) when I tried to run the fastlane setup command it failed with error:

><internal:/opt/homebrew/Cellar/ruby/3.4.2/lib/ruby/3.4.0/rubygems/core_ext/kernel_require.rb>:136:in 'Kernel#require': cannot load such file -- nokogiri (LoadError)

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/53810) and notified the team. Thank you for reporting, we're checking this out!

fire-at-will:
Hi @theoks, thanks for reporting! 🙌 We're taking a look at it.

While we work on the fix, you should still be able to build the SDK with Xcode 16.2 in the meantime.

fire-at-will:
@theoks Version 5.18.0 has been released, which addresses the issue. I'm going to close this issue for now, but feel free to reopen it or open another issue if you see any other issues.

Thanks again for reporting! 🙌

https://github.com/RevenueCat/purchases-ios/releases/tag/5.18.0

PR #4814, opened 2025-02-22
Fix SDK Compilation on Xcode 16.3/iOS 18.4 Beta 1

fire-at-will:
### Motivation
This PR addresses GH issue https://github.com/RevenueCat/purchases-ios/issues/4813. Apple released iOS 18.4 beta 1 and Xcode 16.3, which introduces the [SubscriptionInfo](https://developer.apple.com/documentation/storekit/subscriptioninfo) typealias. This caused a name resolution conflict in `PurchaseInformation.swift` where the extension on `SubscriptionInfo` can't tell if it should resolve to `StoreKit.SubscriptionInfo` or `RevenueCat.SubscriptionInfo`.

Screenshot 2025-02-21 at 10 15 27 PM

### Description
This PR makes the extension's reference to `SubscriptionInfo` explicitly reference `RevenueCat.SubscriptionInfo`, which resolves the compilation issue.

### Note to the Reviewer
Our iOS 18 CI jobs aren't using Xcode 16.3 yet, so just because this PR passes CI doesn't necessarily mean it will work. I've tested and confirmed that `RevenueCatUI` on this branch builds for me on Xcode 16.3, but would appreciate it if you could try building as well to be safe.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
+++ b/RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
@@ -324,7 +324,7 @@ enum TransactionType {
 
 }
 
-extension SubscriptionInfo: Transaction {
+extension RevenueCat.SubscriptionInfo: Transaction {
 
     var type: TransactionType {
         .subscription(isActive: isActive,


fire-at-will:
@RCGitBot please test

ajpallares (approved):
Thanks for the quick fix @fire-at-will! I have checked and this works for me with Xcode 16.3 beta whereas it doesn't without the change 👍

PR #4817, opened 2025-02-24
[Paywalls V2] Adds a note on publishing to the missing paywall error.

JayShortway:
## Motivation
We've already had a few instances of feedback from people (including fellow cats) who forgot to publish their v2 paywall, and we're confused by the SDK error. This aims to clarify that.


tonidero (approved):

ajpallares (approved):
Thanks for this!

PR #4819, opened 2025-02-25
Adds `buildServer.json` to `.gitignore`

JayShortway:
## Motiviation
This helps those of us using/experimenting with [xcode-build-server](https://github.com/SolaWing/xcode-build-server). The `buildServer.json` file is machine-specific as it contains absolute paths so it's probably best to exclude it from source control.


tonidero (approved):
I think that's fine :)

PR #4824, opened 2025-02-26
feat: disable postAttributionData requests when in UI preview mode

ajpallares:
### Checklist

### Motivation
The SDK should minimize the requests it makes in UI preview mode. In particular, the post attribution data request should not happen.

### Description
This PR disables the `postAttributionData` request (`POST "subscribers/appUserID/attribution"`) when the SDK is initialized in UI preview mode


Sources/Attribution/AttributionPoster.swift
```diff
@@ -154,6 +154,10 @@ final class AttributionPoster {
         Self.postponedAttributionData = nil
     }
 
+    func clearPostponedAttributionData() {
+        Self.postponedAttributionData = nil
+    }
```
ajpallares, line 159: This is not strictly required, but I don't think it harms to clear the postponed attribution data when the SDK is initialized in UI preview mode.

ajpallares: You are right. Yes, I also think it's fine because UI preview mode is internal. But I'm not really sure about needing to have the `clearPostponedAttributionData()` method anyway, as in UI preview mode the `postponedAttributionData` should never populate with data 🤔

JayShortway (approved):
Nice!

PR #4826, opened 2025-02-26
Add default refundWindowDuration to HelpPath.init

fire-at-will:
### Description
`CustomerCenterConfigData.HelpPath.init()` has a non-optional `refundHelpWindow` parameter that would be a breaking change if released. This PR provides a default value of `.forever` to this initializer to prevent it from being a breaking change.

### DIFF
--- a/Sources/CustomerCenter/CustomerCenterConfigData.swift
+++ b/Sources/CustomerCenter/CustomerCenterConfigData.swift
@@ -343,7 +343,7 @@ public struct CustomerCenterConfigData: Equatable {
                     openMethod: OpenMethod? = nil,
                     type: PathType,
                     detail: PathDetail?,
-                    refundWindowDuration: RefundWindowDuration?) {
+                    refundWindowDuration: RefundWindowDuration? = .forever) {
             self.id = id
             self.title = title
             self.url = url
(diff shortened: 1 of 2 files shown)


ajpallares (approved):
Makes sense. One question though. Shouldn't this have been caught by the API tests? Or are they not executed in every PR?

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -343,7 +343,7 @@ public struct CustomerCenterConfigData: Equatable {
                     openMethod: OpenMethod? = nil,
                     type: PathType,
                     detail: PathDetail?,
-                    refundWindowDuration: RefundWindowDuration?) {
+                    refundWindowDuration: RefundWindowDuration? = .forever) {
```
facumenzella, line 346: I think this is correct. The refund window should be nil. It should only not be nil when it's a refund path, or at least I think that's how defined in the backend

facumenzella:
@RCGitBot please test

facumenzella:
@RCGitBot please test

facumenzella:
@fire-at-will @ajpallares just re-triggered all tests just to make sure nothing's broken.
I had to tweak one key from the backend response because it was incorrect. See https://github.com/RevenueCat/purchases-ios/pull/4826/commits/8a666e9b7940ce3da8d76aa4960c9b3585f8c7b6

This is not live yet (under a feature flag) so nothing to worry about

ajpallares (approved):

PR #4833, opened 2025-02-27
Allow choosing the locale for a `PaywallView`

ajpallares:
### Checklist

### Motivation
Currently, the `PaywallPreview` is displayed using a locale decided by the SDK (based on the device's locales preferences). There have been some requests to allow specifying the locale to use when presenting the paywall.

### Description
This PR introduces the `preferredLocale` in `PaywallView`. Initially, this parameter is only included under `@_spi(Internal)`, but the idea is to make this public in the short term.

#### Why `preferredLocale`?
The SDK can't guarantee that any locale specified by the caller will be available for a specific paywall. That's why the name `preferredLocale` was chosen. To highlight the fact that it will be used for the paywall **only when possible**. If that locale isn't available for the paywall, then the SDK will fall back to the current behavior of choosing the locale based on the device's locales preferences.


RevenueCatUI/Data/PaywallViewConfiguration.swift
```diff
@@ -22,7 +22,7 @@ struct PaywallViewConfiguration {
     let useDraftPaywall: Bool
     var introEligibility: TrialOrIntroEligibilityChecker?
     var purchaseHandler: PurchaseHandler
-    var locale: Locale
```
ajpallares, line 25: I realized that the `locale` of `PaywallViewConfiguration` was actually redundant. It was always holding `Locale.current`, as the `init` had `.current` as the default value for the `locale` parameter and this `locale` parameter was never used.
That's why I've replaced it with `preferredLocale`, which will now potentially come from the intregrator's side

ajpallares: In those places where a `locale` is non-optional (basically in v1 paywalls, if I'm not mistaken), I've replaced the use of `self.locale` to `self.preferredLocale ?? .current` to keep the existing behavior if `preferredLocale` is not specified

RevenueCatUI/Data/PaywallViewConfiguration.swift
```diff
@@ -22,7 +22,7 @@ struct PaywallViewConfiguration {
...
+    var preferredLocale: Locale?
```
JayShortway, line 25: Nice! We should probably add it to the `presentPaywallIfNeeded()` api too, as well as Android (and hybrids 🫠).

ferrannp:
Hello @ajpallares!

I came across this PR while looking into some related issues, and I found it very helpful—great work!

I’m reaching out from the React Native side, where we typically don’t rely on the device’s system language. Instead, most apps use an i18n package to manage localization dynamically. Because of this, when a user changes the app’s language, we update the locale at the JavaScript level, but the native iOS paywall still follows the device’s language.

It would be incredibly useful to have a way to pass the preferred locale from React Native to ensure that the paywall aligns with the app’s selected language.

Thank you!

ajpallares:
> I’m reaching out from the React Native side, where we typically don’t rely on the device’s system language. Instead, most apps use an i18n package to manage localization dynamically. Because of this, when a user changes the app’s language, we update the locale at the JavaScript level, but the native iOS paywall still follows the device’s language.

Thank you for the feedback @ferrannp! I honestly didn't know that fact about React Native and it makes this even more necessary. We don't have an exact timeline for it, but we definitely do intend to finish the work on this soon!

ferrannp:
Thanks @ajpallares ! FYI, I patched this library to make it work for now in my project. I was struggling with some stuff and I realized that here: https://github.com/RevenueCat/purchases-ios/blob/bfa73ed467a0e071fa7c0256c6ae046b4f9535ef/Sources/Networking/HTTPClient/HTTPClient.swift#L117 you download languages only based on the iOS system settings, so I had to patch that too. Just in case you want to handle this similar situation (e.g. I make the Paywall in Spanish but the user does not have Spanish in their iPhone settings).

ajpallares:
Closing in favor of https://github.com/RevenueCat/purchases-ios/pull/5288 and https://github.com/RevenueCat/purchases-ios/pull/5292

PR #4835, opened 2025-02-28
[Diagnostics] Add extra parameters to `applePurchaseAttempt`

ajpallares:
### Checklist

### Description
This adds requestedProductIdsKey and notFoundProductIdsKey properties to the appleProductsRequest. This won't be used by Grafana but will be in the future DW events.

### Notes
* A `Product.PurchaseResult` of `.pending`is interpreted as "successful until proven otherwise", so it is sent with `successful == true`.
* This PR also changes the `response_time_millis` parameter on the existing events `http_request_performed` and `apple_products_request` to send an `Int` instead of a `Double`, to make it equal to Android where `long` is being sent.


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -599,6 +616,14 @@ final class PurchasesOrchestrator {
                 throw ErrorUtils.purchaseCancelledError()
             }
 
+            self.trackPurchaseEventIfNeeded(startTime,
```
tonidero, line 619: Ah good catch!

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -608,43 +633,87 @@ final class PurchasesOrchestrator {
...
+                                            promotionalOfferId: promotionalOffer?.identifier,
+                                            winBackOfferApplied: winBackOfferApplied,
+                                            storeKitVersion: .storeKit2,
+                                            purchaseResult: nil,
+                                            error: error.asPublicError)
             throw error
         } catch {
             let purchasesError = ErrorUtils.purchasesError(withStoreKitError: error)
-            self.trackPurchaseEventIfNeeded(storeKitVersion: .storeKit2, error: purchasesError.asPublicError)
+            self.trackPurchaseEventIfNeeded(startTime,
```
tonidero, line 647: I'm wondering, with all these exceptions happening here, and now increasing the lines of code, is much less legible. Does it make sense to extract processing the error into a separate method?

tonidero (approved):
This looks great! Left some comments about the legibility in the `PurchasesOrchestrator`, but I think it all makes sense!

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
ajpallares: Yeah, I've thought about that (thanks to the linter). I'll try to rewrite it to make it clearer because I agree that the code is not very easy to follow anymore. Thanks!

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -551,100 +564,141 @@ final class PurchasesOrchestrator {
...
-                                                                           fetchPolicy: .cachedOrFetched)
-        }
+            guard !self.systemInfo.dangerousSettings.customEntitlementComputation else {
+                throw error
+            }
 
-        if !userCancelled {
-            self.trackPurchaseEventIfNeeded(storeKitVersion: .storeKit2, error: nil)
-        }
+            let purchasesError: PurchasesError = switch error {
```
tonidero, line 678: I believe this syntax (switch as an expression) is not supported in older xcode versions so we had to revert usages of it for now... We can probably use some `#if swift...` macros but might be better to write it without using those.

tonidero (commented):
Just a comment, but I like it!

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
ajpallares: Right! This is a feature in Swift 5.9 and Xcode 15 (minimum required to upload apps to the App Store) already comes with Swift 5.9. That said, I think it's possible for someone to still be using an older version of Swift even, so better to be safe than sorry. 
Thanks for the catch @tonidero!

PR #4837, opened 2025-02-28
chore: Remove unused key from customer center event

facumenzella:
### Motivation
We've realized this key is ignored, so better to remove it than to have it there for nothing.

### Description
As opposed to the Android implementation, the StoredEvent already stores the encoded event. This makes backwards compatibility so much easier, because once it's stored you can forget about future problems. Editing events is harmless.

There are also [integration tests](https://github.com/RevenueCat/purchases-ios/pull/4830) in place, so it's safe to say that this is harmless as well 🚀


Sources/CustomerCenter/Events/CustomerCenterEvent.swift
```diff
@@ -217,13 +214,13 @@ extension CustomerCenterBaseData: Equatable, Codable, Sendable {}
 
 extension CustomerCenterAnswerSubmittedEvent.Data: Equatable, Codable, Sendable {
 
+    // This keys are used for `StoredEvent` only
```
ajpallares, line 217:
```suggestion
    // These keys are used for `StoredEvent` only
```

Tests/UnitTests/CustomerCenter/Events/CustomerCenterAnswerSubmittedEventTests.swift
ajpallares, line 1: Just making sure that we really want to remove the whole test class and not only the `surveyOptionTitleKey` parameter as in the other files

ajpallares (approved):
👍  just a couple comments

facumenzella:
@RCGitBot please test

Tests/UnitTests/CustomerCenter/Events/CustomerCenterAnswerSubmittedEventTests.swift
facumenzella: Yes, I should have mentioned this in the PR description. I added the test on my first week haha, but now I've realised the test does not make sense because the encoding being tested is not relevant.

facumenzella:
@RCGitBot please test

PR #4839, opened 2025-02-28
Use array instead of sets in diagnostics events

ajpallares:
### Motivation
Some unit tests were flaky because of the use of `Set`s for Diagnostics events

### Description
This PR changes the `Set`s in DIagnostics events to be `Array`s, so that the order is predictable for tests. Note that the event parameters using `Set` were still being sent as `Array`s in the JSON anyways...

### Notes
Ideally, we should find a way to limit the use of `Set`s in `AnyEncodable`, otherwise we might introduce flaky tests in the future.

### DIFF
--- a/Sources/Diagnostics/DiagnosticsEvent.swift
+++ b/Sources/Diagnostics/DiagnosticsEvent.swift
@@ -13,7 +13,7 @@
 
 import Foundation
 
-struct DiagnosticsEvent: Codable, Equatable {
+struct DiagnosticsEvent: Codable {
 
     let version: Int = 1
     let eventType: DiagnosticsEvent.EventType
@@ -54,17 +54,6 @@ extension DiagnosticsEvent {
         case requestedProductIdsKey
         case notFoundProductIdsKey
 
-    }
-
-}
-
-extension DiagnosticsEvent {
-
-    static func == (lhs: DiagnosticsEvent, rhs: DiagnosticsEvent) -> Bool {
-        return lhs.version == rhs.version &&
-               lhs.eventType == rhs.eventType &&
-               lhs.properties == rhs.properties &&
-               lhs.timestamp == rhs.timestamp
     }
 
 }
--- a/Tests/StoreKitUnitTests/ProductsManagerTests.swift
+++ b/Tests/StoreKitUnitTests/ProductsManagerTests.swift
@@ -159,8 +159,8 @@ class SK1ProductsManagerDiagnosticsTrackingTests: ProductsManagerTests {
         let params = self.mockDiagnosticsTracker.trackedProductsRequestParams.value.first
         expect(params?.wasSuccessful) == true
         expect(params?.storeKitVersion) == .storeKit1
-        expect(params?.requestedProductIds) == Set([identifier, notFoundIdentifier])
-        expect(params?.notFoundProductIds) == Set([notFoundIdentifier])
+        expect(params?.requestedProductIds) == [identifier, notFoundIdentifier]
+        expect(params?.notFoundProductIds) == [notFoundIdentifier]
         expect(params?.errorMessage).to(beNil())
         expect(params?.errorCode).to(beNil())
     }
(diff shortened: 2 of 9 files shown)


Tests/UnitTests/TestHelpers/DiagnosticsEvent+Equatable.swift
```diff
@@ -0,0 +1,40 @@
...
+    public static func == (lhs: DiagnosticsEvent, rhs: DiagnosticsEvent) -> Bool {
+        return lhs.version == rhs.version &&
+               lhs.eventType == rhs.eventType &&
+               lhs.properties == rhs.properties &&
+               lhs.timestamp == rhs.timestamp
+    }
+
+}
+
+extension AnyEncodable: @retroactive Equatable {
```
tonidero, line 28: Hmm I wonder if this belongs in its own file... I think that's used in other places other than diagnostics?

Tests/StoreKitUnitTests/ProductsManagerTests.swift
```diff
@@ -156,13 +156,13 @@ class SK1ProductsManagerDiagnosticsTrackingTests: ProductsManagerTests {
...
-        expect(params?.storeKitVersion) == .storeKit1
-        expect(params?.requestedProductIds) == Set([identifier, notFoundIdentifier])
-        expect(params?.notFoundProductIds) == Set([notFoundIdentifier])
-        expect(params?.errorMessage).to(beNil())
-        expect(params?.errorCode).to(beNil())
+        let params = try XCTUnwrap(self.mockDiagnosticsTracker.trackedProductsRequestParams.value.first)
+        expect(params.wasSuccessful) == true
+        expect(params.storeKitVersion) == .storeKit1
+        expect(Set(params.requestedProductIds)) == [identifier, notFoundIdentifier]
+        expect(params.notFoundProductIds) == [notFoundIdentifier]
```
tonidero, line 163: Should this also be a `Set`?
```suggestion
        expect(Set(params.notFoundProductIds)) == [notFoundIdentifier]
```

Tests/StoreKitUnitTests/ProductsManagerTests.swift
```diff
@@ -195,14 +195,14 @@ class SK2ProductsManagerDiagnosticsTrackingTests: ProductsManagerTests {
...
-        expect(params?.requestedProductIds) == Set([identifier, notFoundIdentifier])
-        expect(params?.notFoundProductIds) == Set([notFoundIdentifier])
-        expect(params?.errorMessage).to(beNil())
-        expect(params?.errorCode).to(beNil())
-        expect(params?.storeKitErrorDescription).to(beNil())
+        let params = try XCTUnwrap(self.mockDiagnosticsTracker.trackedProductsRequestParams.value.first)
+        expect(params.wasSuccessful) == true
+        expect(params.storeKitVersion) == .storeKit2
+        expect(Set(params.requestedProductIds)) == [identifier, notFoundIdentifier]
+        expect(params.notFoundProductIds) == [notFoundIdentifier]
```
tonidero, line 202: Same here 
```suggestion
        expect(Set(params.notFoundProductIds)) == [notFoundIdentifier]
```

Sources/Diagnostics/DiagnosticsEvent.swift
```diff
@@ -13,7 +13,7 @@
 
 import Foundation
 
-struct DiagnosticsEvent: Codable, Equatable {
```
tonidero, line 16: I wonder if we should keep this here to avoid unnecessary changes... But then again, it does seem unused so I think it should be ok 👍

tonidero (commented):
Just some comments but looking great, thanks for fixing this! 🙇

Tests/StoreKitUnitTests/ProductsManagerTests.swift
tonidero: Shouldn't really matter with one element I guess, but to properly compare it I think it's best to have it :)

tonidero (approved):

Tests/StoreKitUnitTests/ProductsManagerTests.swift
ajpallares: Exactly. But you're right that, for consistency, better to have it like that! Thanks!

PR #4842, opened 2025-03-03
[Paywalls] Always use normal stack if the relevant axis size is fit

MarkVillacampa:
Having a `fit` size with a distribution of `spaceBetween`, `space_around` or `space_evenly` would add SwiftUI `Spacer` components, effectively making the size `fill`. With this change we prevent that by always using a `normal`, instead of a `flex` stack if the relevant axis is `fit`.

### Checklist


joshdholtz (approved):
This looks good to me! :shipit: 🫶

PR #4843, opened 2025-03-03
Create `DiagnosticsEvent.Properties` for type safe diagnostics

ajpallares:
### Checklist

### Motivation
The current implementation of the Diagnostics is not type safe and has certain limitations (e.g. does not allow the use of `Set`s). We intend to add new diagnostics properties + new diagnostics events in the SDK, and all properties will share the same meaning and value types across all diagnostics events.
To prevent errors while adding new diagnostics properties and events, we should make the properties type safe.

### Description
This PR adds the `DiagnosticsEvent.Properties` struct that contains all the properties that any diagnostics event needs, providing a type safe API for sending diagnostics events.

## Important notes
* With this PR, the entry lines in the local diagnostics file have the same JSON format as the entries in the diagnostics JSON sent to the server. I don't think this is a problem. In fact, it simplifies the code, but it's worth noting just in case.
* This PR (together with #4846) means that the existing cached diagnostics events would be discarded upon updating the SDK, because the `DiagnosticsEvents` object is now encoded into a slightly different JSON than before.


Sources/Diagnostics/DiagnosticsEvent.swift
```diff
@@ -13,59 +13,84 @@
...
+        case userCancelled = "user_cancelled"
+        case pending
     }
 
-    enum DiagnosticsPropertyKey: String, Codable {
+    struct Properties: Codable, Equatable {
+        let verificationResult: String?
+        let endpointName: String?
+        let responseTimeMillis: Int?
+        let storeKitVersion: String?
```
tonidero, line 43: I'm wondering if we should send the storeKitVersion for all events... but I think it's ok to keep it like this for now. That can be a future change if needed.

tonidero (approved):
This is great! Thank you so much for this improvement! It will help a lot with all the new events 🙏

Sources/Diagnostics/DiagnosticsEvent.swift
ajpallares: Does it makes sense in all events? E.g. `get_offerings_started` or `get_customer_info_started`.

PR #4844, opened 2025-03-03
Deprecates `CustomerCenterActionHandler` in favor of modifiers

vegaro:
We are moving away from public enums in our API since they limit our capabilities to make updates on them. This PR replaces the `CustomerCenterActionHandler` with view modifiers

### DIFF
--- a/RevenueCat.xcodeproj/project.pbxproj
+++ b/RevenueCat.xcodeproj/project.pbxproj
@@ -48,6 +48,7 @@
 		1E2F910B2CC8FE5600BDB016 /* ContactSupportUtilities.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1E2F910A2CC8FE5600BDB016 /* ContactSupportUtilities.swift */; };
 		1E2F911B2CC918ED00BDB016 /* ContactSupportUtilitiesTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1E2F911A2CC918ED00BDB016 /* ContactSupportUtilitiesTests.swift */; };
 		1E2F91722CCFA98C00BDB016 /* WebRedemptionStrings.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1E2F91712CCFA98C00BDB016 /* WebRedemptionStrings.swift */; };
+		1E42CC6C2D7F1A0500E0EE8D /* CacheStatus.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1E42CC6B2D7F1A0500E0EE8D /* CacheStatus.swift */; };
 		1E473B662AC42D34008B07F9 /* StoreMessagesHelper.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1E473B652AC42D34008B07F9 /* StoreMessagesHelper.swift */; };
 		1E473B682AC43254008B07F9 /* StoreMessageType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1E473B672AC43254008B07F9 /* StoreMessageType.swift */; };
 		1E568B512ACC6A8300D3C12F /* StoreMessageTypeTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1E568B502ACC6A8300D3C12F /* StoreMessageTypeTests.swift */; };
@@ -1342,6 +1343,7 @@
 		1E2F910A2CC8FE5600BDB016 /* ContactSupportUtilities.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ContactSupportUtilities.swift; sourceTree = "<group>"; };
 		1E2F911A2CC918ED00BDB016 /* ContactSupportUtilitiesTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ContactSupportUtilitiesTests.swift; sourceTree = "<group>"; };
 		1E2F91712CCFA98C00BDB016 /* WebRedemptionStrings.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = WebRedemptionStrings.swift; sourceTree = "<group>"; };
+		1E42CC6B2D7F1A0500E0EE8D /* CacheStatus.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CacheStatus.swift; sourceTree = "<group>"; };
 		1E473B652AC42D34008B07F9 /* StoreMessagesHelper.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = StoreMessagesHelper.swift; sourceTree = "<group>"; };
 		1E473B672AC43254008B07F9 /* StoreMessageType.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = StoreMessageType.swift; sourceTree = "<group>"; };
 		1E473B692AC46908008B07F9 /* MockStoreMessagesHelper.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MockStoreMessagesHelper.swift; sourceTree = "<group>"; };
@@ -5230,6 +5232,7 @@
 			children = (
 				B3B5FBC0269E17CE00104A0C /* DeviceCache.swift */,
 				B3B5FBBE269E081E00104A0C /* InMemoryCachedObject.swift */,
+				1E42CC6B2D7F1A0500E0EE8D /* CacheStatus.swift */,
 			);
 			path = Caching;
 			sourceTree = "<group>";
@@ -6119,6 +6122,7 @@
 				2D991ACA268BA56900085481 /* StoreKitRequestFetcher.swift in Sources */,
 				5723FBBE28EDE1360003BA16 /* InternalAPI.swift in Sources */,
 				A55D5D66282ECCC100FA7623 /* PostAdServicesTokenOperation.swift in Sources */,
+				1E42CC6C2D7F1A0500E0EE8D /* CacheStatus.swift in Sources */,
 				B3B5FBB4269CED4B00104A0C /* BackendErrorCode.swift in Sources */,
 				FDC892FE2CD157F1000AEB9F /* WinBackOffer.swift in Sources */,
 				2DDF41B624F6F387005BC22D /* ASN1ObjectIdentifierBuilder.swift in Sources */,
--- a/Sources/Caching/CacheStatus.swift
+++ b/Sources/Caching/CacheStatus.swift
@@ -1,0 +1,18 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  CacheStatus.swift
+//
+//  Created by Toni Rico Diez on 10/3/25.
+
+enum CacheStatus: String, Codable {
+    case stale = "STALE"
+    case notFound = "NOT_FOUND"
+    case valid = "VALID"
+}
--- a/Sources/Caching/DeviceCache.swift
+++ b/Sources/Caching/DeviceCache.swift
@@ -190,6 +190,16 @@ class DeviceCache {
         self.offeringsCachedObject.clearCacheTimestamp()
     }
 
+    func offeringsCacheStatus(isAppBackgrounded: Bool) -> CacheStatus {
+        if self.offeringsCachedObject.cachedInstance == nil {
+            return .notFound
+        } else if self.isOfferingsCacheStale(isAppBackgrounded: isAppBackgrounded) {
+            return .stale
+        } else {
+            return .valid
+        }
+    }
+
     // MARK: - subscriber attributes
 
     func store(subscriberAttribute: SubscriberAttribute, appUserID: String) {
--- a/Sources/Diagnostics/DiagnosticsEvent.swift
+++ b/Sources/Diagnostics/DiagnosticsEvent.swift
@@ -45,6 +45,8 @@ struct DiagnosticsEvent: Codable, Equatable {
         case errorEnteringOfflineEntitlementsMode = "error_entering_offline_entitlements_mode"
         case applePurchaseAttempt = "apple_purchase_attempt"
         case maxDiagnosticsSyncRetriesReached = "max_diagnostics_sync_retries_reached"
+        case getOfferingsStarted = "get_offerings_started"
+        case getOfferingsResult = "get_offerings_result"
     }
 
     enum PurchaseResult: String, Codable, Equatable {
@@ -79,6 +81,7 @@ struct DiagnosticsEvent: Codable, Equatable {
         let promotionalOfferId: String?
         let winBackOfferApplied: Bool?
         let purchaseResult: PurchaseResult?
+        let cacheStatus: CacheStatus?
         let isRetry: Bool?
 
         init(verificationResult: String? = nil,
@@ -99,6 +102,7 @@ struct DiagnosticsEvent: Codable, Equatable {
              promotionalOfferId: String? = nil,
              winBackOfferApplied: Bool? = nil,
              purchaseResult: PurchaseResult? = nil,
+             cacheStatus: CacheStatus? = nil,
              isRetry: Bool? = nil) {
             self.verificationResult = verificationResult
             self.endpointName = endpointName
@@ -118,6 +122,7 @@ struct DiagnosticsEvent: Codable, Equatable {
             self.promotionalOfferId = promotionalOfferId
             self.winBackOfferApplied = winBackOfferApplied
             self.purchaseResult = purchaseResult
+            self.cacheStatus = cacheStatus
             self.isRetry = isRetry
         }
 
--- a/Sources/Diagnostics/DiagnosticsTracker.swift
+++ b/Sources/Diagnostics/DiagnosticsTracker.swift
@@ -66,6 +66,18 @@ protocol DiagnosticsTrackerType {
     @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
     func trackErrorEnteringOfflineEntitlementsMode(reason: DiagnosticsEvent.OfflineEntitlementsModeErrorReason,
                                                    errorMessage: String)
+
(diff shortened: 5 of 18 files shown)


RevenueCatUI/CustomerCenter/Data/CustomerCenterActionPreferenceConnector.swift
```diff
@@ -0,0 +1,60 @@
...
+//  
+//  Created by Cesar de la Vega on 2024-06-17.
+
+import RevenueCat
+import SwiftUI
+
+#if os(iOS)
+
+/// A view modifier that connects CustomerCenterViewModel actions to the SwiftUI preference system
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
```
ajpallares, line 20:
```suggestion
@available(iOS 15.0, *)
```
As it's specified below that it's `unavailable` for other platforms. Could be confusing otherwise

RevenueCatUI/CustomerCenter/View+CustomerCenterActions.swift
ajpallares, line 1: A suggestion that could make improve the code clearer (by removing most `@available` clauses) and remove some warnings.
All these `structs` and `typealias` could be under the namespace of `CustomerCenterView`, inside an `extension CustomerCenterView`. This has many benefits:
* The code is clearer to read, as only the `extension CustomerCenterView` needs to be marked with the `@available` clause.
* Since all these public APIs are really meant for the CustomerCenter, I think it makes sense to use the `CustomerCenterView` namespace for their declarations.
* I think the autocomplete of `CustomerCen...` should be less overwhelming for clients.
* This should remove some linter warnings as the type names will be shortened: e.g.
  * `public typealias CustomerCenterRestoreStartedHandler` --> `RestoreStartedHandler`
  * `struct CustomerCenterRestoreStartedPreferenceKey` --> `RestoreStartedPreferenceKey`

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionBridge.swift
```diff
@@ -0,0 +1,69 @@
...
+//  
+//  Created by Cesar de la Vega on 2024-06-17.
+
+import RevenueCat
+import SwiftUI
+
+/// Helper class that bridges the CustomerCenterActionHandler to the new preference-based system
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@MainActor
+class CustomerCenterActionBridge {
```
ajpallares, line 20: This bridge is a good idea to never forget™️ calling the deprecated action handlers. However, I'd have it rely on some higher level object instead of on `CustomerCenterAction`, because we cannot add new cases to it without breaking the API. E.g. something in the vein of
```swift
internal enum CustomerCenterInternalAction {
    case `public`(CustomerCenterAction)
    case newCase1
    case newCase2
}
```

ajpallares (commented):
Looking good so far!
I have added some comments though

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionPreferenceConnector.swift
```diff
@@ -0,0 +1,67 @@
...
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+struct CustomerCenterActionPreferenceConnector: ViewModifier {
+
+    @ObservedObject var viewModel: CustomerCenterViewModel
```
facumenzella, line 26: nit: I think we can just inject the `actionBridge`.

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -98,6 +98,7 @@ public struct CustomerCenterView: View {
 #endif
                 self.trackImpression()
             }
+            .modifier(CustomerCenterActionPreferenceConnector(viewModel: viewModel))
```
facumenzella, line 101: I wonder if it's more precise to apply this just to `content` rather than at the body top level

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionBridge.swift
facumenzella, line 20:
```suggestion
final class CustomerCenterActionBridge {
```

facumenzella (commented):
One thing that I am not a particular fan of is the overall naming. We have:
- Bridge
- Connector 

I like the abstraction though.

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionBridge.swift
facumenzella: `legacy` or `deprecated` could work as well, I really like this

RevenueCatUI/CustomerCenter/View+CustomerCenterActions.swift
vegaro: love it

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -162,58 +158,57 @@ struct ManageSubscriptionsView: View {
...
-                .environment(\.localization, CustomerCenterConfigTestData.customerCenterData.localization)
-                .environment(\.appearance, CustomerCenterConfigTestData.customerCenterData.appearance)
-            }
-            .preferredColorScheme(colorScheme)
-            .previewDisplayName("Free subscription - \(colorScheme)")
-        }
-    }
-
-}
+// @available(iOS 15.0, *)
```
vegaro, line 161: this is not compiling and I don't know why

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptions/ManageSubscriptionsViewModel.swift
```diff
@@ -181,17 +182,19 @@ private extension ManageSubscriptionsViewModel {
             do {
                 guard let purchaseInformation = self.purchaseInformation else { return }
                 let productId = purchaseInformation.productIdentifier
-                self.customerCenterActionHandler?(.refundRequestStarted(productId))
+                self.actionBridge.handleAction(.public(.refundRequestStarted(productId)))
+
                 let status = try await self.purchasesProvider.beginRefundRequest(forProduct: productId)
                 self.refundRequestStatus = status
-                self.customerCenterActionHandler?(.refundRequestCompleted(status))
+                self.actionBridge.handleAction(.public(.refundRequestCompleted(status)))
```
vegaro, line 189: I am not convinced about the fact that legacy actions need to be wrapped under public now 🤔 @ajpallares

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionBridge.swift
```diff
@@ -0,0 +1,82 @@
...
+    init(customerCenterActionHandler: CustomerCenterActionHandler?) {
+        self.customerCenterActionHandler = customerCenterActionHandler
+    }
+
+    /// Main entry point for handling all actions
+    /// For legacy CustomerCenterAction, calls the legacy handler and triggers callbacks
+    /// For non-legacy actions, only triggers callbacks
+    func handleAction(_ internalAction: CustomerCenterInternalAction) {
+        // For public legacy actions, call the legacy handler
+        if case let .public(publicAction) = internalAction {
```
facumenzella, line 50: nit: I think I prefer `legacy` over `public` 💅

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionPreferenceConnector.swift
```diff
@@ -0,0 +1,67 @@
...
+import SwiftUI
+
+#if os(iOS)
+
+/// A view modifier that connects CustomerCenterViewModel actions to the SwiftUI preference system
+@available(iOS 15.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+struct CustomerCenterActionPreferenceConnector: ViewModifier {
```
facumenzella, line 24: I wonder if renaming this to `CustomerCenterActionPreferencesViewModifier` (or something similar) might make things a bit clearer. 

Sometimes these custom suffixes add clarity, but other times they just add noise when navigating the code.

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionBridge.swift
```diff
@@ -0,0 +1,82 @@
...
+/// Internal enum to represent both public CustomerCenterAction cases and internal-only actions
+internal enum CustomerCenterInternalAction {
+    case `public`(CustomerCenterAction)
+    // Example of a new internal-only action that doesn't exist in the public API
+    case buttonTapped(buttonId: String)
+}
+
+/// Helper class that bridges the CustomerCenterActionHandler to the new preference-based system
+@MainActor
+final class CustomerCenterActionBridge {
```
facumenzella, line 26: Same feeling about naming. I think I prefer `CustomerCenterActionHandlerWrapper` 🤔

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptions/ManageSubscriptionsViewModel.swift
ajpallares: Do you refer to the name `public` or to the fact that legacy actions are wrapped?

If it's about the naming, I agree that `public` is not the best name for this. As @facumenzella suggested, `legacy` is a better name.

vegaro: The thing is that they are not really legacy, they are still as valid as the new ones but these need to trigger an extra callback

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionBridge.swift
vegaro: take a look at [`0a85523` (#4844)](https://github.com/RevenueCat/purchases-ios/pull/4844/commits/0a85523271246fc7bcd690e7d216bf8634419c00)

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptions/ManageSubscriptionsViewModel.swift
vegaro: take a look at [`0a85523` (#4844)](https://github.com/RevenueCat/purchases-ios/pull/4844/commits/0a85523271246fc7bcd690e7d216bf8634419c00)

RevenueCatUI/CustomerCenter/Views/UIKit Compatibility/CustomerCenterViewController.swift
```diff
@@ -33,23 +33,176 @@ import SwiftUI
...
+
+    // MARK: - Initialization
+
     /// Create a view controller to handle common customer support tasks
     /// - Parameters:
     ///   - customerCenterActionHandler: An optional `CustomerCenterActionHandler` to handle actions
     ///   from the Customer Center.
+    @available(*, deprecated, message: "Use the initializer with individual action handlers instead")
     public init(
         customerCenterActionHandler: CustomerCenterActionHandler? = nil
```
ajpallares, line 59: A similar comment here.
```suggestion
        customerCenterActionHandler: CustomerCenterActionHandler?
```

To disambiguate the `init`s with all parameters set to its default values (`init()` in this case), I would remove the default parameter value of the `customerCenterActionHandler` so that 
1. Only customers explicitly using `init(customerCenterActionHandler: ...)` would get the deprecation warning
2. `init()` is not ambiguous.

RevenueCatUI/CustomerCenter/Views/UIKit Compatibility/CustomerCenterViewController.swift
```diff
@@ -33,23 +33,176 @@ import SwiftUI
...
+    private var restoreStartedHandler: CustomerCenterView.RestoreStartedHandler?
+    private var restoreCompletedHandler: CustomerCenterView.RestoreCompletedHandler?
+    private var restoreFailedHandler: CustomerCenterView.RestoreFailedHandler?
+    private var showingManageSubscriptionsHandler: CustomerCenterView.ShowingManageSubscriptionsHandler?
+    private var refundRequestStartedHandler: CustomerCenterView.RefundRequestStartedHandler?
+    private var refundRequestCompletedHandler: CustomerCenterView.RefundRequestCompletedHandler?
+    private var feedbackSurveyCompletedHandler: CustomerCenterView.FeedbackSurveyCompletedHandler?
+
+    // Legacy handler for backward compatibility
+    private var legacyHandler: DeprecatedCustomerCenterActionHandler?
```
ajpallares, line 49: Instead of storing both the legacy handler and the individual handlers, I believe we could map the legacy handler into the individual ones, similar to what's done in [RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift](https://github.com/RevenueCat/purchases-ios/blob/0dc7f065bf274f76b88d3ad6efce00a4a1f03301/RevenueCatUI/CustomerCenter/View%2BPresentCustomerCenter.swift#L73)

Not a big deal though. So FFTI

ajpallares (changes requested):
Looking great!!

I have some minor comments and 2 comments about ambiguous methods that I think we should fix.

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionPreferencesViewModifier.swift
```diff
@@ -38,17 +39,28 @@ struct CustomerCenterActionPreferencesViewModifier: ViewModifier {
...
-                actionWrapper.setRestoreFailed = { restoreFailed = $0 as NSError }
-                actionWrapper.setRestoreCompleted = { restoreCompleted = $0 }
+                actionWrapper.setRestoreStarted = { 
+                    // Increment counter to create a new unique value
+                    // This guarantees preference change detection
+                    restoreCounter += 1
+                }
+                
+                actionWrapper.setRestoreFailed = { error in
+                    restoreFailed = error as NSError
```
ajpallares, line 49: I believe the casting here is not necessary
```suggestion
                    restoreFailed = error
```

RevenueCatUI/CustomerCenter/CustomerCenterView+Actions.swift
```diff
@@ -0,0 +1,333 @@
...
+    // MARK: - View Modifiers
+
+    fileprivate struct OnRestoreStartedModifier: ViewModifier {
+        let handler: RestoreStartedHandler
+
+        func body(content: Content) -> some View {
+            content
+                .onPreferenceChange(RestoreStartedPreferenceKey.self) { inProgress in
+                    if inProgress {
+                        self.handler()
```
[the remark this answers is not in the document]
vegaro: let's keep the warning, like in Paywalls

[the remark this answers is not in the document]
ajpallares: Yep. I agree. We'll fix it once it becomes an error, because I think that my proposed fix only silences the warning but does not fixes the `MainActor` issue. So the warning is much better for now 👍

RevenueCatUI/CustomerCenter/CustomerCenterView+Actions.swift
```diff
@@ -0,0 +1,339 @@
...
+
+    /// Invokes the given closure when a refund request completes in the Customer Center.
+    /// Example:
+    /// ```swift
+    ///  var body: some View {
+    ///     ContentView()
+    ///         .sheet(isPresented: self.$displayCustomerCenter) {
+    ///             CustomerCenterView()
+    ///                 .onCustomerCenterRefundRequestCompleted { status in
+    ///                     print("Customer Center refund request completed with status: \(status)")
```
ajpallares, line 308:
```suggestion
    ///                 .onCustomerCenterRefundRequestCompleted { productId, status in
    ///                     print("Customer Center refund request for product \(productId) completed with status: \(status)")
```

ajpallares (approved):
Amazing job! 🚀 

Only one improvement in the documentation, as we've added the `productId` to the `RefundRequestCompletedHandler`

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionViewModifier.swift
```diff
@@ -0,0 +1,101 @@
...
+import RevenueCat
+import SwiftUI
+
+#if os(iOS)
+
+/// A wrapper that makes any value unique by including a UUID
+struct UniqueWrapper<T> {
+    let id = UUID()
+    let value: T
+}
```
facumenzella, line 23: Nice touch 💅

RevenueCatUI/CustomerCenter/Views/CustomerCenterNavigationLink.swift
```diff
@@ -49,24 +57,77 @@ public struct CustomerCenterNavigationLink<Label: View>: View {
...
     ///
-    /// - Parameter label: A closure that returns the view to display as the navigation link’s label.
+    /// - Parameter label: A closure that returns the view to display as the navigation link's label.
+    @available(*, deprecated, message: """
+    Use the view modifiers instead.
+    For example, use .onCustomerCenterRestoreStarted(), .onCustomerCenterRestoreCompleted(), etc.
+    """)
     public init(
-        customerCenterActionHandler: CustomerCenterActionHandler? = nil,
+        customerCenterActionHandler: CustomerCenterActionHandler?,
```
facumenzella, line 66: won't this break the public API? 👀

ajpallares: It won't because there's now the new init
```
public init(@ViewBuilder label: @escaping () -> Label)
```

So doing 
```
CustomerCenterNavigationLink {

}
```
now uses the new, nondeprecated `init`.

facumenzella: Nice

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionViewModifier.swift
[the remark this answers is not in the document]
vegaro: yes it could, let me update that

facumenzella:
Another thought, but I don't have a solution right now, is that we need a way to make sure this is working without actually testing it manually 🙃

facumenzella (approved):

PR #4845, opened 2025-03-03
Virtual Currency Balances

fire-at-will:
### Motivation
This PR allows developers to check the balances of their Virtual Currencies through the SDK's `CustomerInfo` object.

### Description
This PR allows developers to access the balances of their Virtual Currencies like so:
```swift
let coinBalance: Int? = customerInfo.virtualCurrencies["COIN"]?.balance
```

### New APIs
It accomplishes this with a few API additions:
```swift
@objc(RCCustomerInfo) public final class CustomerInfo: NSObject {
    #if ENABLE_VIRTUAL_CURRENCIES
    /// Dictionary of all virtual currency codes to their info
    @objc public let virtualCurrencies: [String: VirtualCurrencyInfo]
    #endif
}

@objc(RCVirtualCurrencyInfo)
public final class VirtualCurrencyInfo: NSObject {

    /// The current balance of the virtual currency.
    ///
    /// This property represents the amount of virtual currency currently available.
    /// The balance is represented as an integer value.
    @objc public let balance: Int
}
```

### ENABLE_VIRTUAL_CURRENCIES Compiler Flag
This PR introduces the `ENABLE_VIRTUAL_CURRENCIES` compiler flag to help us control the rollout of the new feature.

### Testing
- Adds the new APIs to the Swift & Obj-C API Testers
- Adds unit tests for parsing the virtual currency balances from JSON into the new objects on CustomerInfo


Sources/Networking/Responses/CustomerInfoResponse.swift
```diff
@@ -100,6 +105,12 @@ extension CustomerInfoResponse {
 
     }
 
+    #if ENABLE_VIRTUAL_CURRENCIES
+    internal struct VirtualCurrencyInfo {
+        let amount: Int
```
fire-at-will, line 110: This may change to `balance` in the near future, I'll update this here or in a follow-up PR, depending on when/if the changes are made

Sources/Purchasing/VirtualCurrencyInfo.swift
```diff
@@ -0,0 +1,40 @@
...
+/// - Warning: This feature is currently in beta and is subject to change.
+///
+@objc(RCVirtualCurrencyInfo)
+public final class VirtualCurrencyInfo: NSObject {
+
+    /// The current balance of the virtual currency.
+    ///
+    /// This property represents the amount of virtual currency currently available.
+    /// The balance is represented as an integer value.
+    @objc public let balance: Int
```
JayShortway, line 30: 2 separate comments:
- Just making sure it's a conscious decision not to support fractional amounts?
- Also, I understand an `Int` is 64-bit on 64-bit platforms. Do we still support any 32-bit devices? If so, would explicitly making this an `Int64` make it 64-bit on those platforms as well? Just asking questions for my understanding, and to make sure we support consistent amounts everywhere.

Sources/Identity/CustomerInfo.swift
```diff
@@ -99,6 +99,11 @@ public typealias ProductIdentifier = String
     /// Dictionary of all subscription product identifiers and their subscription info
     @objc public let subscriptionsByProductIdentifier: [ProductIdentifier: SubscriptionInfo]
 
+    #if ENABLE_VIRTUAL_CURRENCIES
+    /// Dictionary of all virtual currency codes to their info
+    @objc public let virtualCurrencies: [String: VirtualCurrencyInfo]
+    #endif
```
JayShortway, line 105: Instead of a compiler flag, we could also use `@_spi(Internal)` which we've been doing in a few places. (Or we could even think about using `@spi_(Experimental)` for this.) Just a suggestion.

JayShortway (approved):
I like it, and I like the API! Just some questions and suggestions. It all looks good to me, but since this is a new API maybe some more opinions are desired. 😄

Package.swift
```diff
@@ -85,7 +86,7 @@ let package = Package(
                 resources: [
                     .copy("../Sources/PrivacyInfo.xcprivacy")
                 ],
-                swiftSettings: [visionOSSetting] + ciCompilerFlags + additionalCompilerFlags),
+                swiftSettings: [visionOSSetting] + ciCompilerFlags + additionalCompilerFlags + [virtualCurrenciesFlag]),
```
JayShortway, line 89: Just checking: do we want to merge this?

MarkVillacampa: If I recall correctly, the idea is to have `virtual-currency-dev` as the development branch for this feature, which will *not* be merged with `main` yet, and has the flag disabled.

And then have a `virtual-currency-dev-flag-enabled` which is rebased constantly with `virtual-currency-dev`, and just has an extra commit enabling the flag.

That potential beta testers can use the `virtual-currency-dev-flag-enabled` branch directly and have the flag active, while we're free to merge the `virtual-currency-dev` branch with `main` eventually without fear of accidentally enabling the flag there.

@fire-at-will did I get it right?

MarkVillacampa: So maybe we want to add support for the flag but leave it disabled in this commit?

Sources/Identity/CustomerInfo.swift
MarkVillacampa: One downside of that is `@_spi` does not work for Obj-C :( it's worked fine with SwiftUI/Swift-only classes tho.

fire-at-will: Yeah, that's why I didn't go with `@_spi` here - I didn't want folks using Objective-C to start using the API before we're ready. `@_spi` would be _much_ easier to deal with than the compiler flags 😅

MarkVillacampa (approved):
Looks good! just lets double check if we want to enable the flag here or not

Sources/Purchasing/VirtualCurrencyInfo.swift
fire-at-will: Yep, it's a conscious decision to not support fractional balances! Balances are stored in the database as an integer in the backend.

I believe you're right - an `Int` on a 32bit device would default to a 32bit integer, and for apps with exceedingly high balances, that could cause inconsistencies across devices. I'll see if I can change this to be an Int64. Great callout! :D

Package.swift
fire-at-will: That's what we had discussed, but it started to feel like a bit much so I just put everything in here. 

That said, you're right, there's definitely some value in having a "ready-to-merge" branch, so I've removed the flag from this branch, and once this is merged into `virtual-currency-dev`, will recreate the `virtual-currency-dev-flag-enabled` branch to share with beta customers 👍

fire-at-will:
Thanks for the quick reviews @MarkVillacampa & @JayShortway! I've made two changes:
- Switched `balance` from being an `Int` to a `Int64`
- Removed the compiler flag from being enabled on this branch

Sources/Identity/CustomerInfo.swift
JayShortway: Oh TIL, thanks! That make sense indeed.

Sources/Purchasing/VirtualCurrencyInfo.swift
JayShortway: Alright, thanks for confirming! We could still (technically / in the future) support fractional balances with integers, by using micros. But that would mean adding another 6 zeros to every integer haha. Good thing we have 64 bit to play with 😛

Package.swift
JayShortway: Ahh okay, thanks for explaining!

JayShortway (approved):
Very nice!

PR #4846, opened 2025-03-03
Fix potential wrong count on diagnosis sync

ajpallares:
### Checklist

### Motivation
If for any reason, an invalid entry line gets written into the diagnostics cache file, then duplicated events will potentially be sent.

### Description
This PR makes sure that the number of lines deleted always matches the number of lines read, even if some lines couldn't be decoded to `DiagnosticsEvent`.

### DIFF
--- a/Tests/UnitTests/Diagnostics/DiagnosticsSynchronizerTests.swift
+++ b/Tests/UnitTests/Diagnostics/DiagnosticsSynchronizerTests.swift
@@ -177,6 +177,19 @@ class DiagnosticsSynchronizerTests: TestCase {
         expect(self.userDefaults.removeObjectForKeyCalledValues) == [cacheKey]
     }
 
+    func testSyncMultipleEventsWithInvalidEvent() async throws {
+        let event1 = await self.storeEvent()
+        await fileHandler.append(line: "Invalid entry line")
+        let event2 = await self.storeEvent(timestamp: Self.eventTimestamp2)
+
+        try await self.synchronizer.syncDiagnosticsIfNeeded()
+
+        expect(self.api.invokedPostDiagnosticsEvents) == true
+        expect(self.api.invokedPostDiagnosticsEventsParameters) == [[ event1, event2 ]]
+
+        await self.verifyEmptyStore()
+    }
+
 }
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
--- a/Sources/Diagnostics/DiagnosticsFileHandler.swift
+++ b/Sources/Diagnostics/DiagnosticsFileHandler.swift
@@ -16,7 +16,7 @@ import Foundation
 protocol DiagnosticsFileHandlerType: Sendable {
 
     @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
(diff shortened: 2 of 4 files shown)


Tests/UnitTests/Diagnostics/DiagnosticsSynchronizerTests.swift
```diff
@@ -177,6 +177,19 @@ class DiagnosticsSynchronizerTests: TestCase {
...
+        await fileHandler.append(line: "Invalid entry line")
+        let event2 = await self.storeEvent(timestamp: Self.eventTimestamp2)
+
+        try await self.synchronizer.syncDiagnosticsIfNeeded()
+
+        expect(self.api.invokedPostDiagnosticsEvents) == true
+        expect(self.api.invokedPostDiagnosticsEventsParameters) == [[ event1, event2 ]]
+
+        await self.verifyEmptyStore()
+    }
```
ajpallares, line 191: This test would not pass without the changes of the PR. Which means that invalid entries in the local diagnostics cache file could lead to duplicated events being sent

Tests/UnitTests/Diagnostics/DiagnosticsSynchronizerTests.swift
```diff
@@ -177,6 +177,19 @@ class DiagnosticsSynchronizerTests: TestCase {
         expect(self.userDefaults.removeObjectForKeyCalledValues) == [cacheKey]
     }
 
+    func testSyncMultipleEventsWithInvalidEvent() async throws {
```
tonidero, line 180: Should we add a test with invalid events but that leaves the expected amount in disk? Don't remember if that was easy to do though 😅

tonidero (approved):
Makes a lot of sense, thanks for fixing!! Just a comment about a possible test

Tests/UnitTests/Diagnostics/DiagnosticsSynchronizerTests.swift
ajpallares: Not really sure what you mean @tonidero 🤔

tonidero: Right sorry 😅, meant a test that has invalid events, syncs the events, but the events in the cache are still bigger than the number of events synced, so it leaves a few behind, possibly also some invalid events

ajpallares: If I'm not mistaken, when a sync happens, all the events in the file get synced.
So the situation you mention could only happen if a new event gets added in the middle of a sync (after the entries have been read and before the synced entries have been deleted). Is that the scenario you mean?

tonidero: Ohhh ok... In Android, we do have a maximum number of events that can be synced per request. But yeah you're right, we don't have that in iOS. So yeah, I think the test I was proposing doesn't make sense, sorry about the confussion 🙇

ajpallares: No problem. Thanks for the review!

PR #4848, opened 2025-03-04
Add activity indicator to restore purchases button behavior

joshdholtz:
### Motivation

Paywalls V2 "Restore Purchases" behavior didn't have any type of activity or progress indicator

### Description

- If purchase...
  - Purchase buttons are disabled and show activity indicator
  - Restore buttons are just disabled (and 0.35 opacity)
- If restore...
  - Restore buttons are disabled and show activity indicator
  - Purchase buttons are just disabled (and 0.35 opacity)

https://github.com/user-attachments/assets/1ef9b08a-4b80-4ba6-9354-20cd0a6519f5

### DIFF
--- a/RevenueCatUI/Templates/V2/ViewHelpers/ProgressViewModifier.swift
+++ b/RevenueCatUI/Templates/V2/ViewHelpers/ProgressViewModifier.swift
@@ -26,7 +26,9 @@ struct ProgressViewModifier: ViewModifier {
     func body(content: Content) -> some View {
         content
             #if !os(watchOS)
-            .background(.ultraThinMaterial)
+            .applyIfLet(self.backgroundStyle, apply: { view, _ in
+                view.background(.ultraThinMaterial)
+            })
             #endif
             .overlay(progressView)
     }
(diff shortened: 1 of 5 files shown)


RevenueCatUI/Templates/V2/ViewHelpers/ProgressViewModifier.swift
```diff
@@ -40,7 +42,7 @@ struct ProgressViewModifier: ViewModifier {
                 .progressViewStyle(CircularProgressViewStyle(tint: bestTint(for: colorInfo)))
         case .image, .none:
             ProgressView()
-                .progressViewStyle(CircularProgressViewStyle(tint: .white))
+                .progressViewStyle(CircularProgressViewStyle(tint: .gray))
```
ajpallares, line 45: Just curious: this change, is it to make it visible in both light and dark mode?

ajpallares (approved):
Looks great! 🚀

RevenueCatUI/Templates/V2/ViewHelpers/ProgressViewModifier.swift
joshdholtz: Yeah yeah, it would make it visible for both the default-ish feeling white and black backgrounds that can change for light and dark mode 😁 I noticed it on a few paywalls I was playing with. I think there are better odds are that defaulting to gray will show up better most of the time 🎲 😇

PR #4849, opened 2025-03-04
Run `all-tests` on `main` and notify Slack

vegaro:
- Run `all-tests` workflow on `main` and notify Slack if failures
- Cleaned up some duplicated jobs in between `all-tests` and `build-test`
- Invoked a job that we were not calling

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -418,6 +418,7 @@ jobs:
           name: SPM Release Build
           command: swift build -c release --target RevenueCat
           no_output_timeout: 30m
+      - slack-notify-on-fail
 
   spm-release-build-xcode-16:
     executor:
@@ -430,6 +431,7 @@ jobs:
           name: SPM Release Build
           command: swift build -c release --target RevenueCat
           no_output_timeout: 30m
+      - slack-notify-on-fail
 
   spm-release-build-xcode-15:
     executor:
@@ -442,6 +444,7 @@ jobs:
           name: SPM Release Build
           command: swift build -c release --target RevenueCat
           no_output_timeout: 30m
+      - slack-notify-on-fail
 
   spm-xcode-14-1:
     executor:
@@ -458,6 +461,7 @@ jobs:
           name: SPM RevenueCatUI Release Build
           command: swift build -c release --target RevenueCatUI
           no_output_timeout: 30m
+      - slack-notify-on-fail
 
   pod-lib-lint:
     executor:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -418,6 +418,7 @@ jobs:
           name: SPM Release Build
           command: swift build -c release --target RevenueCat
           no_output_timeout: 30m
+      - slack-notify-on-fail
```
vegaro, line 421: Wherever I added this is because it's a job invoked in `all-tests`

.circleci/config.yml
```diff
@@ -598,7 +606,6 @@ jobs:
       - store_artifacts:
           path: fastlane/test_output
           destination: scan-test-output
-      - slack-notify-on-fail
```
vegaro, line 601: If I removed it is because it's only invoked in `build-test` and that runs on PRs

.circleci/config.yml
```diff
@@ -1357,27 +1389,13 @@ workflows:
...
-      - emerge_purchases_ui_snapshot_tests:
-          context:
-            - slack-secrets-ios
+      - lint
+      - run-test-ios-17
+      - run-test-ios-18
+      - pod-lib-lint
+      - run-revenuecat-ui-ios-17
+      - run-revenuecat-ui-ios-18
+      - emerge_purchases_ui_snapshot_tests
```
vegaro, line 1398: This is `build-test` and it runs on PRs so we don't want a notification in that case

.circleci/config.yml
```diff
@@ -1478,45 +1496,97 @@ workflows:
...
-      - run-revenuecat-ui-ios-18
-      - spm-revenuecat-ui-watchos
-      - installation-tests-cocoapods
-      - installation-tests-swift-package-manager
-      - installation-tests-custom-entitlement-computation-swift-package-manager
-      - installation-tests-carthage
-      - installation-tests-xcode-direct-integration
-      - installation-tests-receipt-parser
-      - api-tests
+      - backend-integration-tests-SK1:
```
vegaro, line 1504: a bit hard to review but:
- I added the context to all
- Removed `run-revenuecat-ui-ios-17` and `run-revenuecat-ui-ios-18` and `emerge_purchases_ui_snapshot_tests` which are already run in the `build-test` workflow and should speed things up
- Added `backend-integration-tests-offline` which was missing

tonidero (approved):
Makes a lot of sense. Thanks for doing this 🙇. And sorry I misunderstood the intent with this 😩

vegaro:
It's all good @tonidero. I actually approved your PR so it's not on you 😁

PR #4850, opened 2025-03-04
feat: Add onClose handler support for CustomerCenter

facumenzella:
### Motivation
While working on kmp support for CustomerCenter, I realised there's no way to close programmatically the view. The close button does not work because it works with the environment dismiss

I'll need to update hybrid commons after this, and then kmp 🚀 

### Description
The PR adds `onCloseHandler` that overrides the environment dismiss behavior.
- Update `NavigationOptions`
- Update docs

### DIFF
--- a/RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
+++ b/RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
@@ -59,14 +59,12 @@ extension View {
         isPresented: Binding<Bool>,
         customerCenterActionHandler: CustomerCenterActionHandler? = nil,
         presentationMode: CustomerCenterPresentationMode = .default,
-        onDismiss: (() -> Void)? = nil,
-        onClose: (() -> Void)? = nil
+        onDismiss: (() -> Void)? = nil
     ) -> some View {
         self.modifier(
             PresentingCustomerCenterModifier(
                 isPresented: isPresented,
                 onDismiss: onDismiss,
-                onClose: onClose,
                 myAppPurchaseLogic: nil,
                 customerCenterActionHandler: customerCenterActionHandler,
                 presentationMode: presentationMode
@@ -87,13 +85,9 @@ private struct PresentingCustomerCenterModifier: ViewModifier {
     /// The closure to execute when dismissing the sheet / fullScreen present
     let onDismiss: (() -> Void)?
 
-    /// The closure to execute when tapping on the close button
-    let onClose: (() -> Void)?
-
     init(
         isPresented: Binding<Bool>,
         onDismiss: (() -> Void)?,
-        onClose: (() -> Void)?,
         myAppPurchaseLogic: MyAppPurchaseLogic?,
         customerCenterActionHandler: CustomerCenterActionHandler?,
         presentationMode: CustomerCenterPresentationMode,
@@ -102,7 +96,6 @@ private struct PresentingCustomerCenterModifier: ViewModifier {
         self._isPresented = isPresented
         self.presentationMode = presentationMode
         self.onDismiss = onDismiss
-        self.onClose = onClose
         self.customerCenterActionHandler = customerCenterActionHandler
         self._purchaseHandler = .init(wrappedValue: purchaseHandler ??
                                       PurchaseHandler.default(performPurchase: myAppPurchaseLogic?.performPurchase,
@@ -124,8 +117,7 @@ private struct PresentingCustomerCenterModifier: ViewModifier {
                         self.customerCenterView(
                             navigationOptions: CustomerCenterNavigationOptions(
                                 usesExistingNavigation: false,
-                                shouldShowCloseButton: true,
-                                onCloseHandler: onClose
+                                shouldShowCloseButton: true
                             )
                         )
                     }
@@ -136,8 +128,7 @@ private struct PresentingCustomerCenterModifier: ViewModifier {
                         self.customerCenterView(
                             navigationOptions: CustomerCenterNavigationOptions(
                                 usesExistingNavigation: false,
-                                shouldShowCloseButton: true,
-                                onCloseHandler: onClose
+                                shouldShowCloseButton: true
                             )
                         )
                     }
--- a/RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
+++ b/RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
@@ -52,7 +52,6 @@ extension View {
     ///   Defaults to `.default`.
     ///   - onDismiss: A callback triggered when either the sheet / fullscreen present is dismissed
     ///     Ensure you set `isPresented = false` when this is called.
-    ///   - onClose: A callback that replaces environment dismiss
     ///
     /// - Returns: A view modified to support presenting the Customer Center.
     public func presentCustomerCenter(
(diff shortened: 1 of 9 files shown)


RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
```diff
@@ -25,34 +25,54 @@ import SwiftUI
...
+    ///   - onClose: A callback that replaces environment dismiss
+    ///
+    /// - Returns: A view modified to support presenting the Customer Center.
     public func presentCustomerCenter(
         isPresented: Binding<Bool>,
         customerCenterActionHandler: CustomerCenterActionHandler? = nil,
         presentationMode: CustomerCenterPresentationMode = .default,
-        onDismiss: (() -> Void)? = nil
+        onDismiss: (() -> Void)? = nil,
+        onClose: (() -> Void)? = nil
```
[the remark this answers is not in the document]
JayShortway: For my understanding, what's the difference between the 2?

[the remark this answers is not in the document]
JayShortway: From the function docs, is `onDismiss` called _after_ it is dismissed, and `onClose` before?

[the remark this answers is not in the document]
facumenzella: As discussed in private, I've removed this because it's only needed if you use CustomerCenter directly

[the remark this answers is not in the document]
facumenzella: @JayShortway the difference is the context

- `onDismiss` is used in `sheet` and `fullscreen` view modifiers (swiftui native)
- `onClose` is a custom one to be used when tapping the `close` button (useful for when you don't use those modifiers, so the environment `dismiss` is not set ).

In any case, I've removed them from here because they are only valid when doing something custom, and you're using `CustomerCenterView` directly

RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
```diff
@@ -25,34 +25,52 @@ import SwiftUI
...
+    ///   - onDismiss: A callback triggered when either the sheet / fullscreen present is dismissed
+    ///     Ensure you set `isPresented = false` when this is called.
+    ///   - onClose: A callback that replaces environment dismiss
+    ///
+    /// - Returns: A view modified to support presenting the Customer Center.
     public func presentCustomerCenter(
         isPresented: Binding<Bool>,
         customerCenterActionHandler: CustomerCenterActionHandler? = nil,
         presentationMode: CustomerCenterPresentationMode = .default,
         onDismiss: (() -> Void)? = nil
```
JayShortway, line 62: Ignore this comment if this has already been released, but maybe this could be `onDismissed` to indicate that it has already been dismissed at this point (if I understand correctly)?

RevenueCatUI/CustomerCenter/Views/CustomerCenterNavigationOptions.swift
```diff
@@ -34,26 +34,52 @@ public struct CustomerCenterNavigationOptions {
...
+    ///
+    /// - Example Usage in SwiftUI:
+    /// ```swift
+    /// let options = CustomerCenterNavigationOptions(dismissHandler: {
+    ///     dismiss()
+    /// })
+    /// ```
+    ///
+    ///
+    public let onCloseHandler: (() -> Void)?
```
JayShortway, line 54: The docs seem to cal this `dismissHandler` instead of `onCloseHandler`. My suggestion would be to call it `onDismiss` as that is sufficient to indicate what it does, but it's not a blocker at all.

JayShortway (approved):
Nice one! Couple of suggestions here and there.

RevenueCatUI/CustomerCenter/Views/CustomerCenterNavigationOptions.swift
facumenzella: Damn, renamed the variable but not the docs. The thing is that onDismiss could be confusing, because it's using for native SwiftUI handlers like the ones in `.sheet` and `.fullscreen`

I started with that, and then renamed it. 
Let me re-do this and I'll ping you back

RevenueCatUI/CustomerCenter/View+PresentCustomerCenter.swift
facumenzella: Yeah, I agree. It has been released already 😢 

Apple documents it like:

```
onDismiss
The closure to execute when dismissing the modal view.
```

vegaro: yeah, I mentioned this to @facumenzella as well, `onDismissed` would have been more accurate tbh

RevenueCatUI/CustomerCenter/Views/CustomerCenterNavigationOptions.swift
facumenzella: I think that the only good renaming would be `onCloseButtonTapped`, but I don't have strong opinions on this. The docs are clear on how it's used.

RevenueCatUI/CustomerCenter/Views/CustomerCenterNavigationOptions.swift
```diff
@@ -34,26 +34,52 @@ public struct CustomerCenterNavigationOptions {
...
-    /// The default configuration for `CustomerCenterNavigationOptions`.
+    /// A custom handler to execute when closing the Customer Center from the close button in the navigation bar.
     ///
-    /// - `usesNavigationStack`: `true` (default to modern navigation).
-    /// - `usesExistingNavigation`: `false` (wraps content in a new navigation stack by default).
-    /// - `shouldShowCloseButton`: `true` (displays a close button by default).
-    public static let `default` = CustomerCenterNavigationOptions()
+    /// `onCloseHandler` allows developers to define a custom method for handling the dismissal of
+    /// the Customer Center. This is useful in cases where the default SwiftUI dismissal (`@Environment(\.dismiss)`)
+    /// is insufficient, such as when integrating with hybrid frameworks.
```
vegaro, line 41: this is a public API, not sure we want to mention this, it could be a bit confusing

RevenueCatUI/CustomerCenter/Views/CustomerCenterNavigationOptions.swift
```diff
@@ -34,26 +34,52 @@ public struct CustomerCenterNavigationOptions {
...
         self.usesNavigationStack = usesNavigationStack
         self.usesExistingNavigation = usesExistingNavigation
         self.shouldShowCloseButton = shouldShowCloseButton
+        self.onCloseHandler = onCloseHandler
     }
 }
+
+public extension CustomerCenterNavigationOptions {
+
+    /// The default configuration for `CustomerCenterNavigationOptions`.
```
vegaro, line 79: should we mention the `dismiss` env is called as default?

vegaro (approved):

RevenueCatUI/CustomerCenter/Views/CustomerCenterNavigationOptions.swift
facumenzella: Do you mean  ... ?

> This is useful in cases where the default SwiftUI dismissal (`@Environment(\.dismiss)`) is insufficient, such as when integrating with hybrid frameworks.

I thought it was useful, in case someone wants to build their own thing. But if it's not a common thing then I can remove it.

PR #4853, opened 2025-03-05
Trigger the RC app upload when releasing a SDK version

ajpallares:
### Motivation
We want to have a new version of the RC app built and uploaded whenever a new version of the purchases-ios SDK gets released

### Description
This PR adds a new CircleCI job that gets executed as part of the `deploy-tag` workflow.

Successful execution [here](https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/32673/workflows/5a8f6fd3-13dd-4be0-b87c-4b71d9e121f9/jobs/432141)

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -737,7 +737,7 @@ platform :ios do
     archive_path = build_binary_size_test_app(
       integration_method: integration_method
     )
-    
+
     send_to_emerge(
       archive_path: archive_path
     )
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -1119,6 +1119,46 @@ platform :ios do
...
+
+    project_slug = ENV["RC_APP_CIRCLE_CI_PROJECT_SLUG"]
+    UI.user_error!("Please set the RC_APP_CIRCLE_CI_PROJECT_SLUG environment variable") unless project_slug
+
+    # Make request
+    headers = {"Circle-Token": circle_token, "Content-Type": "application/json", "Accept": "application/json"}
+    data = {
+      definition_id: definition_id,
+      config: { branch: "main" },
+      checkout: { branch: "main" },
```
tonidero, line 1142: I guess we could make the branch a parameter... But I think it's ok to just have main for now and we could add it later when needed.

tonidero (approved):
Nice job!! 🚢

fastlane/Fastfile
ajpallares: Right. I actually thought about it, but "main" feels like static enough, and it's hardcoded everywhere else. But I agree, we can make it a parameter later if we need.
Thanks for the suggestion!

PR #4855, opened 2025-03-06
Remove resetting `appSessionId` for customer center + add `appSessionId` and `eventId` to diagnostics events

tonidero:
### Description
Main changes in this PR:
- We won't be resetting the `appSessionId` upon login/logout anymore. This will affect customer center events as well
- Reuse the same `appSessionId` for diagnostics events
- Add an `id` to identify each event uniquely.


Sources/Diagnostics/DiagnosticsEvent.swift
```diff
@@ -15,13 +15,27 @@ import Foundation
 
 struct DiagnosticsEvent: Codable {
 
+    let id: UUID
     let version: Int = 1
     let eventType: DiagnosticsEvent.EventType
     let properties: [DiagnosticsPropertyKey: AnyEncodable]
     let timestamp: Date
+    let appSessionId: UUID?
```
ajpallares, line 23: I'm guessing this is optional to make it backwards compatible with the existing cached.

If that's the case, if we're going to be discarding the existing cached events anyway due to #4843, then this can be nonoptional. Not a big deal though

ajpallares (approved):
Looks good. This will be "fun" to merge with #4843 🥲
One small comment only

Sources/Diagnostics/DiagnosticsEvent.swift
tonidero: Yup that was the idea... but yeah you're right, I can make it non-nullable. It will be better 👍 Thank you! And yeah, I will hold merging this until after your PR. I think that should be a bit easier 😅

Sources/Paywalls/Events/PaywallEventsManager.swift
```diff
@@ -41,18 +39,14 @@ actor PaywallEventsManager: PaywallEventsManagerType {
...
-        appSessionID: UUID = UUID()
+        appSessionID: UUID = SystemInfo.appSessionID
     ) {
         self.internalAPI = internalAPI
         self.userProvider = userProvider
         self.store = store
         self.appSessionID = appSessionID
     }
 
-    func resetAppSessionID() {
```
tonidero, line 52: @RevenueCat/customer-center I wanted to check with you that this change is ok since it will affect customer center events. We talked over Slack and I think it's ok to do this change but want to double-check

facumenzella (approved):

PR #4856, opened 2025-03-06
Have snapshot tests use same encoding as SDK

ajpallares:
### Motivation
The snapshot tests were using a different JSON encoding strategy than the SDK, in particular, for dates.
As an example, the `CustomerInfo` snapshot tests were mapping dates differently to what the SDK does (which maps them to its ISO8601 format) E.g. 
```JSON
"purchase_date" : -326323164
```
vs
```JSON
 "purchase_date" : "1990-08-30T02:40:36Z"
```

These tests should encode objects with a format as similar as possible to the actual encoding format used in the SDK.

### Description
With this PR:
* The snapshots now use the ISO8601 date format
* Encoding objects in unit tests now uses a `JSONEncoder` that copies the properties from the SDK's `JSONEncoder.prettyPrinted`

### DIFF
--- a/Tests/UnitTests/Networking/Responses/__Snapshots__/CustomerInfoDecodingTests/testEncoding.1.json
+++ b/Tests/UnitTests/Networking/Responses/__Snapshots__/CustomerInfoDecodingTests/testEncoding.1.json
@@ -1,28 +1,28 @@
 {
   "entitlement_verification" : 0,
-  "request_date" : 668454178,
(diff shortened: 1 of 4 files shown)


tonidero (approved):
I think this makes sense. Thanks for fixing all these 🙏

JayShortway (approved):
Amazing, this would also help a lot when debugging a failing test.

PR #4857, opened 2025-03-06
[Diagnostics] add missing parameter to `http_request_performed` event

ajpallares:
### Checklist

### Description
This PR adds the `is_retry` property to the `http_request_performed` event.

### Additional changes
This PR also changes the place where the event is triggered. The previous implementation would skip sending the event when receiving a network error


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -420,10 +420,6 @@ private extension HTTPClient {
             .asOptionalResult?
             .convertUnsuccessfulResponseToError()
 
-        self.trackHttpRequestPerformedIfNeeded(request: request,
-                                               requestStartTime: requestStartTime,
-                                               result: result)
-
```
ajpallares, line 426: The code would not reach this point when receiving a network error (due to some early returns in the `parse` method --> https://github.com/RevenueCat/purchases-ios/blob/1477e8ca9768e12027bcf5dd519e4157f7dbcb5b/Sources/Networking/HTTPClient/HTTPClient.swift#L334-L343)

tonidero: Ohhh true, nice catch!

tonidero (approved):
Thank you for that catch!

PR #4859, opened 2025-03-07
fix: diagnostics parameter key name

ajpallares:
### Description
This PR fixes the Diagnostics JSON field name for the `eTagHit` parameter. 

It was being sent as `e_tag_hit`, whereas it should be `etag_hit`.


tonidero (approved):
Ohh good catch, thank you!

PR #4860, opened 2025-03-07
Update `customerInfo` from an `AsyncStream` instead of the `PurchasesDelegate` in the SwiftUI sample app

polpielladev:
Now that we are going to allow users to download the SwiftUI sample app from Onboarding, we should make sure that it uses the most up-to-date APIs.

One area for improvement that we noticed was updating the `CustomerInfo` object from an `AsyncStream` instead of the delegate.

### Checklist

### DIFF
--- a/Examples/MagicWeatherSwiftUI/Shared/Sources/ViewModels/UserViewModel.swift
+++ b/Examples/MagicWeatherSwiftUI/Shared/Sources/ViewModels/UserViewModel.swift
@@ -13,7 +13,7 @@ import SwiftUI
 class UserViewModel: ObservableObject {
     static let shared = UserViewModel()
     
-    /* The latest CustomerInfo from RevenueCat. Updated by PurchasesDelegate whenever the Purchases SDK updates the cache */
+    /* The latest CustomerInfo from RevenueCat. Updated by the `customerInfoStream` in the initializer. */
     @Published var customerInfo: CustomerInfo? {
         didSet {
             subscriptionActive = customerInfo?.entitlements[Constants.entitlementID]?.isActive == true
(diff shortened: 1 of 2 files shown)


Examples/MagicWeatherSwiftUI/Shared/Sources/ViewModels/UserViewModel.swift
```diff
@@ -26,6 +26,23 @@ class UserViewModel: ObservableObject {
     /* Set from the didSet method of customerInfo above, based on the entitlement set in Constants.swift */
     @Published var subscriptionActive: Bool = false
     
+    /* Keep track of the task that listens for changes to the `customerInfoStream` so that it can be cancelled later on */
+    private var task = Task<Void, Never> {}
+    
+    private init() {
+        /* Listen to changes in the `customerInfo` object using an `AsyncStream` */
+        self.task = Task.detached {
```
ajpallares, line 34: Does it really need to be a `detached` `Task`? 

If I'm not mistaken, if it weren't a `detached` task, we wouldn't have to keep the reference to `task` and we wouldn't need to `cancel()` it on `deinit`. And it would still be canceled when `UserViewModel` got deallocated.

Not very important, but it would simplify the code a bit.

ajpallares (approved):
Looks good! Much better to use the `AsyncStream`, for sure. One suggestion though

Examples/MagicWeatherSwiftUI/Shared/Sources/ViewModels/UserViewModel.swift
polpielladev: I don't think it does tbf, let's go for simplicity for this. Let me update the code now 👀

PR #4861, opened 2025-03-07
[Diagnostics] add `max_diagnostics_sync_retries_reached` event

ajpallares:
### Checklist

### Description
This PR implements the `max_diagnostics_sync_retries_reached` diagnostics event

### DIFF
--- a/Sources/Diagnostics/Networking/DiagnosticsSynchronizer.swift
+++ b/Sources/Diagnostics/Networking/DiagnosticsSynchronizer.swift
@@ -25,6 +25,7 @@ actor DiagnosticsSynchronizer: DiagnosticsSynchronizerType {
 
     private let internalAPI: InternalAPI
     private let handler: DiagnosticsFileHandlerType
+    private let tracker: DiagnosticsTrackerType?
     private let userDefaults: SynchronizedUserDefaults
 
     private var syncInProgress = false
@@ -32,10 +33,12 @@ actor DiagnosticsSynchronizer: DiagnosticsSynchronizerType {
     init(
         internalAPI: InternalAPI,
         handler: DiagnosticsFileHandlerType,
+        tracker: DiagnosticsTrackerType?,
         userDefaults: SynchronizedUserDefaults
     ) {
         self.internalAPI = internalAPI
         self.handler = handler
+        self.tracker = tracker
         self.userDefaults = userDefaults
     }
 
@@ -79,6 +82,7 @@ actor DiagnosticsSynchronizer: DiagnosticsSynchronizerType {
                 if currentSyncRetries >= Self.maxSyncRetries {
                     Logger.error(Strings.diagnostics.failed_diagnostics_sync_more_than_max_retries)
                     await self.handler.emptyDiagnosticsFile()
+                    self.tracker?.trackMaxDiagnosticsSyncRetriesReached()
                     self.clearSyncRetries()
                 } else {
                     self.increaseSyncRetries(currentRetries: currentSyncRetries)
--- a/Sources/Diagnostics/DiagnosticsEvent.swift
+++ b/Sources/Diagnostics/DiagnosticsEvent.swift
@@ -41,6 +41,7 @@ struct DiagnosticsEvent: Codable, Equatable {
         case customerInfoVerificationResult = "customer_info_verification_result"
         case maxEventsStoredLimitReached = "max_events_stored_limit_reached"
         case applePurchaseAttempt = "apple_purchase_attempt"
+        case maxDiagnosticsSyncRetriesReached = "max_diagnostics_sync_retries_reached"
     }
 
     enum PurchaseResult: String, Codable, Equatable {
--- a/Sources/Diagnostics/DiagnosticsTracker.swift
+++ b/Sources/Diagnostics/DiagnosticsTracker.swift
@@ -54,6 +54,8 @@ protocol DiagnosticsTrackerType {
                               purchaseResult: DiagnosticsEvent.PurchaseResult?,
                               responseTime: TimeInterval)
 
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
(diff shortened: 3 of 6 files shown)


Sources/Diagnostics/Networking/DiagnosticsSynchronizer.swift
```diff
@@ -79,6 +82,7 @@ actor DiagnosticsSynchronizer: DiagnosticsSynchronizerType {
                 if currentSyncRetries >= Self.maxSyncRetries {
                     Logger.error(Strings.diagnostics.failed_diagnostics_sync_more_than_max_retries)
                     await self.handler.emptyDiagnosticsFile()
+                    self.tracker?.trackMaxDiagnosticsSyncRetriesReached()
```
ajpallares, line 85: Calling `trackMaxDiagnosticsSyncRetriesReached()` after `emptyDiagnosticsFile()` is crucial. Otherwise, the diagnostic file would be cleared right after adding the `max_diagnostics_sync_retries_reached` event.

However, I couldn't find an easy way to test this

Tried this approach, but it doesn't work because the `tracker` is a mock that does not actually calls the `DiagnosticsFileHandler` (like the actual `DiagnosticsTracker`).

```
func testMaxDiagnosticsSyncRetriesReachedEventRemainsIfMaxRetriesReached() async throws {
    _ = await self.storeEvent()

    let cacheKey = "com.revenuecat.diagnostics.number_sync_retries"

    let expectedError: NetworkError = .errorResponse(.defaultResponse, .internalServerError)

    self.api.stubbedPostDiagnosticsEventsCompletionResult = .networkError(expectedError)

    self.userDefaults.mockValues = [cacheKey: 3]

    try? await self.synchronizer.syncDiagnosticsIfNeeded()

    expect(self.tracker.trackedMaxDiagnosticsSyncRetriesReachedCalls.value) == 1
    let entries = (await self.handler.getEntries()).compactMap { $0 }
    expect(entries.count) == 1
    expect(entries[0].name) == DiagnosticsEvent.EventName.maxDiagnosticsSyncRetriesReached
}
```

@RevenueCat/coresdk, could you think of some way to add this test without having the `MockDiagnosticsTracker` reimplement the same behavior of `DiagnosticsTracker`?

tonidero: Hmm right this is tricky... I think we would need to either:
- Have both the clearing of the file + tracking the new event happen in the same object, so we can mock it and track the order of calls
- As you said, change to use the actual `DiagnosticsTracker` with some actual files that can test the behavior end to end, which might affect other tests...

Feels like we should go with the second option if possible... though that might indeed require some work...

ajpallares: Yes, that's what I thought. I agree that the 2nd option sounds better, but it's tricky because we might end up testing the mocks. 
I'll give it a thought to try and find a good solution. Thanks!

tonidero: I mean, it might make sense to make this a sort of "integration test" I guess, testing what happens with actual files... I'm ok if we want to do this in a separate PR/later though

JayShortway: Can we make `DiagnosticsTracker` receive a fake `DiagnosticsFileHandler`? That way the test can assert the "files" written by the fake `DiagnosticsFileHandler` (they can just be held in memory), while using the real `DiagnosticsTracker`. Maybe that's easier than checking actual files.

ajpallares: That could also be an approach, yes.

However, I've done some experiments with these tests and this statement of mine:

> Calling trackMaxDiagnosticsSyncRetriesReached() after emptyDiagnosticsFile() is crucial

is false. Currently, if we called `emptyDiagnosticsFile()` right after `trackMaxDiagnosticsSyncRetriesReached()`, the diagnostics file would still contain the `MaxDiagnosticsSyncRetriesReached` event. The reason is that while `emptyDiagnosticsFile` happens asynchronously (with an `await`), `trackMaxDiagnosticsSyncRetriesReached()` (and all tracking of events) is dispatched to a background thread. Which means that, regardless of the order, clearing the file will always happen **before** writing the `MaxDiagnosticsSyncRetriesReached` event to the file.

So, on the one hand, this proves the test of checking that "trackMaxDiagnosticsSyncRetriesReached() happens after emptyDiagnosticsFile()" useless and, on the other hand, we should even more have a E2E test that checks that, after this, the file contains the `MaxDiagnosticsSyncRetriesReached` event.

ajpallares: That could also be an approach, yes.

However, I've done some experiments with these tests and this statement of mine:

> Calling trackMaxDiagnosticsSyncRetriesReached() after emptyDiagnosticsFile() is crucial

is false. Currently, if we called `emptyDiagnosticsFile()` right after `trackMaxDiagnosticsSyncRetriesReached()`, the diagnostics file would still contain the `MaxDiagnosticsSyncRetriesReached` event. The reason is that while `emptyDiagnosticsFile` happens asynchronously (with an `await`), `trackMaxDiagnosticsSyncRetriesReached()` (and all tracking of events) is dispatched to a background thread. Which means that, regardless of the order, clearing the file will always happen **before** writing the `MaxDiagnosticsSyncRetriesReached` event to the file.

So, on the one hand, this proves the test of checking that "trackMaxDiagnosticsSyncRetriesReached() happens after emptyDiagnosticsFile()" useless and, on the other hand, we should even more have a E2E test that checks that, after this, the file contains the `MaxDiagnosticsSyncRetriesReached` event.

JayShortway: Ah yes, good finding! And indeed agreed on the need for more E2E tests!

PR #4862, opened 2025-03-10
chore: `EventsManagerIntegrationTests` working as expected

facumenzella:
### Motivation
We need to be sure events are working as expected =)

### Description
I've tweaked the test to just call Purchases api. The only caveat is that the tracking for customer center async, and the one for paywalls is not, so I had to add an arbitrary sleep. Once this is merged, I can go with https://github.com/RevenueCat/purchases-ios/pull/4837

This will enable further refactors for PaywallEventsManager, like renaming it to support `FeatureEvent`
Test is only ignored in the offline test plan

### DIFF
--- a/Tests/BackendIntegrationTests/EventsManagerIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/EventsManagerIntegrationTests.swift
@@ -23,10 +23,6 @@ import XCTest
 @MainActor
 final class EventsManagerIntegrationTests: BaseBackendIntegrationTests {
 
-    override var apiKey: String { return Constants.customEntitlementComputationApiKey }
-
-    var eventsManager: PaywallEventsManager!
-
     func testPostingPaywallsDoesNotFail() async throws {
         let events = [
             PaywallEvent.cancel(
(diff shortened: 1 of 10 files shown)


facumenzella:
@RCGitBot please test

facumenzella:
@RCGitBot please test

Tests/BackendIntegrationTests/EventsManagerIntegrationTests.swift
```diff
@@ -90,13 +77,22 @@ final class EventsManagerIntegrationTests: BaseBackendIntegrationTests {
...
 
-        _ = try await eventsManager.flushEvents(count: 2)
+        try await flushAndVerify(eventsCount: 2, trackingIsAsync: true)
+    }
+
+    private func flushAndVerify(eventsCount: Int, trackingIsAsync: Bool = false) async throws {
+        if trackingIsAsync {
+            // tracking customer center is async
+            try await Task.sleep(nanoseconds: 1_000_000_000)
+        }
```
JayShortway, line 87: This could add flakiness to our tests. Maybe we can reduce that risk by polling with a timeout instead? E.g. verifying the message was logged every X milliseconds, with a timeout of Y seconds?

JayShortway (commented):
More testing is more better! Just a comment on reducing the risk of flakiness.

Tests/BackendIntegrationTests/EventsManagerIntegrationTests.swift
facumenzella: You're right. I actually found a way to do it. Check 07bf34347

The tricky part is that we need a `Task.yield()` to let the background task run, but seems to work and is definitely less flaky

facumenzella:
@RCGitBot please test

JayShortway (approved):

Tests/BackendIntegrationTests/EventsManagerIntegrationTests.swift
JayShortway: Very nice!! 🙌

PR #4863, opened 2025-03-10
[Diagnostics] add `clearing_diagnostics_after_failed_sync` event

ajpallares:
### Checklist

### Description
This PR implements the `clearing_diagnostics_after_failed_sync` diagnostics event

### DIFF
--- a/Sources/Diagnostics/DiagnosticsEvent.swift
+++ b/Sources/Diagnostics/DiagnosticsEvent.swift
@@ -40,6 +40,7 @@ struct DiagnosticsEvent: Codable, Equatable {
         case appleProductsRequest = "apple_products_request"
         case customerInfoVerificationResult = "customer_info_verification_result"
         case maxEventsStoredLimitReached = "max_events_stored_limit_reached"
+        case clearingDiagnosticsAfterFailedSync = "clearing_diagnostics_after_failed_sync"
         case applePurchaseAttempt = "apple_purchase_attempt"
         case maxDiagnosticsSyncRetriesReached = "max_diagnostics_sync_retries_reached"
     }
--- a/Sources/Diagnostics/DiagnosticsTracker.swift
+++ b/Sources/Diagnostics/DiagnosticsTracker.swift
@@ -56,6 +56,9 @@ protocol DiagnosticsTrackerType {
 
     @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
     func trackMaxDiagnosticsSyncRetriesReached()
+
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    func trackClearingDiagnosticsAfterFailedSync()
 }
 
 @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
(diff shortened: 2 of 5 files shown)


ajpallares:
@RevenueCat/coresdk I believe the logic for whether a failure in the diagnostics events sync should retried or not is slightly different between iOS and Android. 

In Android, the logic is that it should be retried if the `error.code == PurchasesErrorCode.NetworkError` (see it [here](https://github.com/RevenueCat/purchases-android/blob/b54b9f67d2c58cde4029d312ba04bf63c6a2920e/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt#L465)).

In iOS, the logic is that it should be retried if the `NetworkError` is such that the `HTTPStatusCode` returns `true` for `isSuccessfullySynced` (see it [here](https://github.com/RevenueCat/purchases-ios/blob/33c2d066186cd65fd2b811fca819e4335e8c799e/Sources/Networking/HTTPClient/HTTPStatusCode.swift#L101)), which means the sync is only retried for 404 and 500...599 status codes.

Should we do something about this to make the sync retry logic the same?

JayShortway (approved):
Makes sense!

JayShortway:
@ajpallares Great find! I do think that we should make both platforms behave the same. @tonidero which implementation should we pick in your opinion?

tonidero:
Right good call! I think the iOS implementation is the most complete yeah, since it takes into account the server being down (I'm not sure about the 404 case though 🤔).

So I think this makes the most sense and we should update android. Could you leave a comment in this event in the doc and make the color "yellow or red" so we remember to fix this @ajpallares?

tonidero (approved):
Looks great! Thank you!

ajpallares:
Done @tonidero! Thanks!
 
> I'm not sure about the 404 case though

Yeah, in this case, a 404 response does not look to me like should trigger retries. The iOS SDK is using the  `isSuccessfullySynced` in `HTTPStatusCode`, which is used in other places as well.
I think that, perhaps, `isSuccessfullySynced` is being  "abused" in this case of diagnostics sync retries.

We could probably modify this once we tackle the Android part, to make them both the same and in a way that makes sense

ajpallares:
> In Android, the logic is that sync retries only happen if the `error.code == PurchasesErrorCode.NetworkError` (see it [here](https://github.com/RevenueCat/purchases-android/blob/b54b9f67d2c58cde4029d312ba04bf63c6a2920e/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt#L465)).

This comment of mine was not entirely true because in Android, there's also the logic where the sync are retried if the status code is a server error (e.g. >= 500) [here](https://github.com/RevenueCat/purchases-android/blob/b54b9f67d2c58cde4029d312ba04bf63c6a2920e/purchases/src/main/kotlin/com/revenuecat/purchases/common/Backend.kt#L477). Therefore, the logic is fairly complete in Android, with the only differences:
* iOS retries the sync when receiving a status code >= 500 or 404 . Android only for status codes >= 500. I think Android makes more sense here.
* Otherwise iOS retries the sync when the error is other than a `NetworkError` (e.g. an encoding error), while Android only retries the sync when the `PurchasesError`  `code == PurchasesErrorCode.NetworkError`. I don't know what is more correct, I'd lean towards Android, but I could see edge cases where iOS could make sense.

PR #4866, opened 2025-03-11
[Diagnostics] Add offerings start and result events

tonidero:
### Description
This adds 2 new diagnostics events:
- `get_offerings_started`
- `get_offerings_result`


Sources/Purchasing/OfferingsManager.swift
```diff
@@ -24,42 +24,77 @@ class OfferingsManager {
...
         self.productsManager = productsManager
+        self.diagnosticsTracker = diagnosticsTracker
+        self.dateProvider = dateProvider
     }
 
     func offerings(
         appUserID: String,
         fetchPolicy: FetchPolicy = .default,
         fetchCurrent: Bool = false,
+        trackDiagnostics: Bool = true,
```
tonidero, line 52: I had doubts about whether to add this... This method is also used upon storefront changes, so not when the developer calls any of the public APIs to get offerings. I didn't want to move the logic outside of this method though, since that would be much trickier and have more repetition, so I ended up adding a boolean parameter to disable tracking in that case.

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -24,42 +24,77 @@ class OfferingsManager {
...
-            return
-        }
+        self.trackPurchaseStartedIfNeeded(trackDiagnostics: trackDiagnostics)
+        let startTime = self.dateProvider.now()
 
-        guard let memoryCachedOfferings = self.cachedOfferings else {
-            self.fetchFromNetwork(appUserID: appUserID, fetchPolicy: fetchPolicy, completion: completion)
-            return
-        }
+        self.systemInfo.isApplicationBackgrounded { isAppBackgrounded in
```
tonidero, line 58: This will now cause a change in threads when fetching offerings... But that may also happen at any point if fetching offerings from network. So I think this is ok? It does mean we know check whether the app is backgrounded or not before fetching offerings in order to calculate the cache status, which might have some very slight behavior changes.

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -76,7 +111,7 @@ class OfferingsManager {
         appUserID: String,
         isAppBackgrounded: Bool,
         fetchPolicy: FetchPolicy = .default,
-        completion: (@MainActor @Sendable (Result<Offerings, Error>) -> Void)?
+        completion: (@MainActor @Sendable (Result<OfferingsResultData, Error>) -> Void)?
```
tonidero, line 114: This was also a bit tricky... I had to bubble up the requested product ids + not found product ids and this was the simplest... We could also just not add these to the event and just keep them in the store specific event (`apple_products_request` in this case). That might simplify this a bit, but might require more analysis in the data side to associate the events... I ended up just adding this for now. Lmk what you think!

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -24,42 +24,77 @@ class OfferingsManager {
...
 
-        self.systemInfo.isApplicationBackgrounded { isAppBackgrounded in
-            if self.deviceCache.isOfferingsCacheStale(isAppBackgrounded: isAppBackgrounded) {
+            Logger.debug(Strings.offering.vending_offerings_cache_from_memory)
+            self.trackPurchaseResultIfNeeded(trackDiagnostics: trackDiagnostics,
+                                             startTime: startTime,
+                                             cacheStatus: cacheStatus,
+                                             error: nil,
+                                             requestedProductIds: nil,
+                                             notFoundProductIds: nil)
```
tonidero, line 92: These properties do not make sense when fetching offerings from cache.

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -366,6 +405,39 @@ private extension OfferingsManager {
             return PreviewProductType(packageType: packageType) ?? .default
         }
     }
+
+    func trackPurchaseStartedIfNeeded(trackDiagnostics: Bool) {
+        if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *),
+            trackDiagnostics,
+           !self.systemInfo.dangerousSettings.uiPreviewMode,
```
ajpallares, line 412: This is a good one! `uiPreviewMode`. I honestly forgot about it in the other PRs.

However, do we want to force disable it when in UI preview mode? I think it makes sense, but since there's the `diagnosticsEnabled` parameter on the `init`, which is `false` by default, idk if we'll ever want to enable diagnostics for UI preview mode.

If we do want to disable diagnostics for UI preview mode, although it's totally ok to do it here to be safe, would it make sense to limit diagnostics completely for `uiPreviewMode` at initialization? I mean, doing it in 
https://github.com/RevenueCat/purchases-ios/blob/a54329ab56c84748390fc9114731503965230d7e/Sources/Purchasing/Purchases/Purchases.swift#L323

```swift
let diagnosticsFileHandler: DiagnosticsFileHandlerType? = {
    guard diagnosticsEnabled, dangerousSettings?.uiPreviewMode != true,
          #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
    else {
        return nil
    }
    return DiagnosticsFileHandler()
}()
```

Just to open the discussion, but we can do it in a separate PR if we want this

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -366,6 +405,39 @@ private extension OfferingsManager {
             return PreviewProductType(packageType: packageType) ?? .default
         }
     }
+
+    func trackPurchaseStartedIfNeeded(trackDiagnostics: Bool) {
+        if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *),
```
ajpallares, line 410: I don't think the `#available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)` check is needed here

Sources/Purchasing/OfferingsManager.swift
ajpallares: I see that the code is now executed inside a `Task { @MainActor `, which means that all this code now happens on the main thread. It shouldn't have any consequences, I think.

Perhaps the `dispatchCompletionOnMainThreadIfPossible` below could be simplified to call the `completion`, but since the completion in `isApplicationBackgrounded` is not marked with  `@MainActor`, it's safer this way (keeping `dispatchCompletionOnMainThreadIfPossible`).

Sources/Purchasing/OfferingsManager.swift
ajpallares: I get it, a new struct to wrap the three, but if it's useful to have this information in the event, I'd say we go with it 👍

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -366,6 +405,39 @@ private extension OfferingsManager {
...
+    }
+
+    // swiftlint:disable:next function_parameter_count
+    func trackPurchaseResultIfNeeded(trackDiagnostics: Bool,
+                                     startTime: Date,
+                                     cacheStatus: CacheStatus,
+                                     error: Error?,
+                                     requestedProductIds: Set<String>?,
+                                     notFoundProductIds: Set<String>?) {
+        if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *),
```
ajpallares, line 425: Same here. `#available` check is not needed here, I think

Tests/TestingApps/PaywallsTester/PaywallsTester/Config/Constants.swift
```diff
@@ -15,7 +15,7 @@ enum Constants {
      REVENUECAT_API_KEY = your-api-key
      */
     static let apiKey: String = {
-        Bundle.main.object(forInfoDictionaryKey: "REVENUECAT_API_KEY") as? String ?? ""
+        "appl_yBXmuJBTosTvPtfwIySkUzDkxAq"
```
ajpallares, line 18: 😱 🤣  trust me, `Local.xcconfig` works 😢

Tests/UnitTests/Diagnostics/DiagnosticsTrackerTests.swift
ajpallares, line 1: I forgot to add tests here for the other events I added 🤦‍♂️

Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
```diff
@@ -26,6 +26,7 @@ class OfferingsManagerTests: TestCase {
     var mockOfferings: MockOfferingsAPI!
     let mockOfferingsFactory = MockOfferingsFactory()
     var mockProductsManager: MockProductsManager!
+    var mockDiagnosticsTracker: DiagnosticsTrackerType!
```
ajpallares, line 29: I think we can have this be of `MockDiagnosticsTracker` type to avoid force castings below

```suggestion
    var mockDiagnosticsTracker: MockDiagnosticsTracker!
```

tonidero: I think we can't because `MockDiagnosticsTracker` is an iOS 15+ class, and this test class isn't... We could make this test class only support iOS 15+ but then we would be losing some coverage I think

ajpallares: Ohh right. My mistake! 🙏

Sources/Caching/CacheStatus.swift
```diff
@@ -0,0 +1,33 @@
...
+
+enum CacheStatus: String, Codable {
+    case stale
+    case notFound = "not_found"
+    case valid
+}
+
+extension CacheStatus {
+
+    var name: String {
```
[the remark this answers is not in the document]
tonidero: Hmm I was looking at the `VerificationResult` enum when implementing this, where we do implement it with uppercase... seems we have it a bit all over the place 😓... I was also thinking on android, where it might be easier to do it this way (thought we should just decide in one way and be consistent)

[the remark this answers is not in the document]
ajpallares: Yeah, we can just unify them. In any case, we can use the enum's raw values as we want. If we want them to be uppercase, we can just define the raw values in uppercase 👍

Sources/Diagnostics/DiagnosticsTracker.swift
```diff
@@ -62,6 +62,19 @@ protocol DiagnosticsTrackerType {
...
+    func trackOfferingsStarted()
+
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    func trackOfferingsResult(requestedProductIds: Set<String>?,
+                              notFoundProductIds: Set<String>?,
+                              errorMessage: String?,
+                              errorCode: Int?,
+                              verificationResult: VerificationResult?,
+                              cacheStatus: CacheStatus?,
+                              responseTime: TimeInterval)
```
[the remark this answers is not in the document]
tonidero: Right, I think that's mostly the reason... We could indeed create different types to ensure all properties that need to be sent in each case are sent with type safety... But seems more work than it's worth it right now? As in, we would need to either have different track methods for each different event situation, or some sort of enum class with the data for each type of situation... Certainly doable but feels like it increases the complexity a bit... Wdyt?

Sources/Diagnostics/DiagnosticsTracker.swift
```diff
@@ -207,6 +220,40 @@ final class DiagnosticsTracker: DiagnosticsTrackerType, Sendable {
...
+    }
+
+    func trackOfferingsResult(requestedProductIds: Set<String>?,
+                              notFoundProductIds: Set<String>?,
+                              errorMessage: String?,
+                              errorCode: Int?,
+                              verificationResult: VerificationResult?,
+                              cacheStatus: CacheStatus?,
+                              responseTime: TimeInterval) {
+        // WIP Add verification result property once we expose verification result in offerings object
```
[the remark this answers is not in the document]
tonidero: Totally right thanks!

Sources/Purchasing/OfferingsManager.swift
tonidero: Hmm yeah, I think you're totally right! Better to keep it centralized than having this check all over the place, thanks!!

Sources/Purchasing/OfferingsManager.swift
tonidero: Hmm I believe it is needed since `trackOfferingsStarted` is actually marked as available only in those versions?

Tests/TestingApps/PaywallsTester/PaywallsTester/Config/Constants.swift
tonidero: Whoops... I hadn't tried it yet, and went this route... I should learn 🥴  Thanks!

Sources/Diagnostics/DiagnosticsTracker.swift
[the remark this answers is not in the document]
ajpallares: No no, totally agree that it's more work than needed. But my proposal here is a simpler change. Just having `cacheStatus` be nonoptional in this method, as it's always going to be sent (and, perhaps, `verificationResult` in the future). Like this:
```suggestion
    func trackOfferingsResult(requestedProductIds: Set<String>?,
                              notFoundProductIds: Set<String>?,
                              errorMessage: String?,
                              errorCode: Int?,
                              verificationResult: VerificationResult?,
                              cacheStatus: CacheStatus,
                              responseTime: TimeInterval)
```

Sources/Purchasing/OfferingsManager.swift
tonidero: Done in https://github.com/RevenueCat/purchases-ios/pull/4866/commits/d6a9f442e6e8e2b70b18199f30307bdd53124634

Sources/Purchasing/OfferingsManager.swift
ajpallares: Ohhh you're right, again 😄. I didn't need this in my previous PRs because my where I added the events was already restricted to these versions. But yes, you're totally right. Sorry again 🙏

Sources/Diagnostics/DiagnosticsEvent.swift
```diff
@@ -108,6 +112,7 @@ struct DiagnosticsEvent: Codable, Equatable {
             self.promotionalOfferId = promotionalOfferId
             self.winBackOfferApplied = winBackOfferApplied
             self.purchaseResult = purchaseResult
+            self.cacheStatus = cacheStatus?.name
```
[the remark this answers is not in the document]
tonidero: Done in https://github.com/RevenueCat/purchases-ios/pull/4866/commits/c0e81370be4519eabdf9cfa1222f10e5d9f8a3d3

Sources/Diagnostics/DiagnosticsEvent.swift
```diff
@@ -71,6 +73,7 @@ struct DiagnosticsEvent: Codable, Equatable {
         let promotionalOfferId: String?
         let winBackOfferApplied: Bool?
         let purchaseResult: PurchaseResult?
+        let cacheStatus: String?
```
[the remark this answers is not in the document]
tonidero: Done in https://github.com/RevenueCat/purchases-ios/pull/4866/commits/c0e81370be4519eabdf9cfa1222f10e5d9f8a3d3

Tests/TestingApps/PaywallsTester/PaywallsTester/Config/Constants.swift
ajpallares: hahaha no problem! It was just a more fun & constructive way to point out the slipped API key 😆

Sources/Caching/CacheStatus.swift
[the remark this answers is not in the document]
tonidero: Removed in https://github.com/RevenueCat/purchases-ios/pull/4866/commits/c0e81370be4519eabdf9cfa1222f10e5d9f8a3d3

Sources/Caching/CacheStatus.swift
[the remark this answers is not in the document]
tonidero: Actually in this case it's actually used in the main logic, which reminds me I should add some more tests around that 😓. So I preferred to leave it on a separate file.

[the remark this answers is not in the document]
ajpallares: Yep. You're right. I wrote the comment at the beginning of the review and then I forgot to remove it when I saw it used elsewhere. In this case I agree that it's better for it to be separated 👍

Sources/Diagnostics/DiagnosticsTracker.swift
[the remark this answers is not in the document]
tonidero: Oh right, makes sense 👍 Changed!

Sources/Caching/CacheStatus.swift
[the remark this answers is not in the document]
tonidero: Addes some tests in https://github.com/RevenueCat/purchases-ios/pull/4866/commits/eeb82b4d6ab9dcd28509c0284c6537c688cdba98

Tests/UnitTests/Diagnostics/DiagnosticsTrackerTests.swift
ajpallares: Done in https://github.com/RevenueCat/purchases-ios/pull/4867/commits/5219b61a63a9713808cf7df92420b8d2dc200354 🙃

ajpallares: Done in https://github.com/RevenueCat/purchases-ios/pull/4867/commits/5219b61a63a9713808cf7df92420b8d2dc200354 🙃

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -366,6 +405,39 @@ private extension OfferingsManager {
             return PreviewProductType(packageType: packageType) ?? .default
         }
     }
+
+    func trackPurchaseStartedIfNeeded(trackDiagnostics: Bool) {
```
ajpallares, line 409: I believe the name of the method is incorrect
```suggestion
    func trackGetOfferingsStartedIfNeeded(trackDiagnostics: Bool) {
```

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -366,6 +405,39 @@ private extension OfferingsManager {
...
+        if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *),
+            trackDiagnostics,
+           let diagnosticsTracker = self.diagnosticsTracker {
+
+            let responseTime = self.dateProvider.now().timeIntervalSince(startTime)
+
+            diagnosticsTracker.trackOfferingsResult(requestedProductIds: requestedProductIds,
+                                                    notFoundProductIds: notFoundProductIds,
+                                                    errorMessage: error?.localizedDescription,
+                                                    errorCode: error?.errorCode,
```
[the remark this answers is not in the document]
tonidero: Right, I need to review this, but what I was thinking here was to send the error code we return to the developer. So one of these: https://github.com/RevenueCat/purchases-ios/blob/a02fc3f1bcb5966125fdd21934131ae070eec0d9/Sources/Error%20Handling/ErrorCode.swift#L24-L66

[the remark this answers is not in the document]
ajpallares: Yep. I thought about that right after leaving my comment. But I don't think the current implementation returns that `ErrorCode`.

The good think about the new `DiagnosticsEvent.Properties` is that we can have its `errorCode` property be of type `ErrorCode?` instead of `Int?`. This way, the compiler will force us to pass an actual `ErrorCode` 😄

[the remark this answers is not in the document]
ajpallares: @tonidero an easy way to get an `ErrorCode` would be
```
error?.asPurchasesError.errorCode
```
Perhaps that's the right way for us to do this

Sources/Purchasing/OfferingsManager.swift
tonidero: OMG I missed that, thank you! And fixed!

Sources/Purchasing/OfferingsManager.swift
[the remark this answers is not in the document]
tonidero: Right, I think it might be good to change the `DiagnosticsEvent.Properties`'s `errorCode` to use `ErrorCode`, that could be nice... But that would have other reaching changes so will leave it for future PRs. For now, I will use what you mentioned, which indeed seems the easiest. Thanks for catching this!!!

ajpallares (approved):
🚀

Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
```diff
@@ -232,7 +239,7 @@ extension OfferingsManagerTests {
         // then
         expect(result).to(beFailure())
 
-        let error = try XCTUnwrap(logger.messages.filter { $0.level == .error }.onlyElement)
+        let error = try XCTUnwrap(logger.messages.filter { $0.level == .error }.first)
```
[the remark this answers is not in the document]
ajpallares: Yes, it's something we can tackle later.

I wonder if `error.asPurchaseError` really has to log any message 🤔

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -24,42 +24,77 @@ class OfferingsManager {
...
-            return
-        }
+        self.systemInfo.isApplicationBackgrounded { isAppBackgrounded in
+            let cacheStatus = self.deviceCache.offeringsCacheStatus(isAppBackgrounded: isAppBackgrounded)
+
+            let completionFromNetworkWithTracking: @MainActor @Sendable (Result<OfferingsResultData, Error>) -> Void =
+            { result in
+                self.trackGetOfferingsResultIfNeeded(trackDiagnostics: trackDiagnostics,
+                                                     startTime: startTime,
+                                                     cacheStatus: .notFound,
```
ajpallares, line 65: @tonidero I'm starting to implement this in Android and I have my doubts about the value of the `cacheStatus`. According to this implementation, we're always sending `cacheStatus: .notFound` when we're fetching the offerings from the network. But I'm not sure if that should be the case. 

If `fetchCurrent == true`, then the offerings are fetched from the network regardless of the `cacheStatus`, but I it's isn't necessarily `.notFound`. Perhaps it would be more appropriate to send `nil` in the case where we fetch from network regardless of the `cacheStatus`.

It's very likely that this is too nitpicky, but I wanted to point it out just in case it could be important.

tonidero: Hmm you have a good point... It can be confusing to send `notFound` indeed... Maybe we could add another value `notChecked` for that scenario... Wdyt?

ajpallares: That could be a solution, yes. It actually reflects the truth in that scenario 👍

PR #4867, opened 2025-03-11
[Diagnostics] add `error_entering_offline_entitlements_mode` event

ajpallares:
### Checklist

### Description
This PR implements the `error_entering_offline_entitlements_mode` diagnostics event

### Notes

* The `offline_entitlement_error_reason` field is equal to what Android sends, plus a new `no_entitlement_mapping_available` that is not being sent in Android.
* The `error_message` is not going to match Android's, but I think it's OK in this case.


Sources/OfflineEntitlements/OfflineCustomerInfoCreator.swift
```diff
@@ -76,27 +76,51 @@ class OfflineCustomerInfoCreator {
...
+            ))
+
+            return offlineCustomerInfo
+        } catch {
+            let reason, errorMessage: String
+            switch error {
+            case let productsFetcherError as PurchasedProductsFetcher.Error:
+                switch productsFetcherError {
+                case .foundConsumablePurchase:
+                    reason = "one_time_purchase_found"
```
ajpallares, line 107: [In Android](https://github.com/RevenueCat/purchases-android/blob/66a17f0b22c918d08d27df5eec986a3ea60a8e16/purchases/src/main/kotlin/com/revenuecat/purchases/common/diagnostics/DiagnosticsTracker.kt#L238) we use this reason `"one_time_purchase_found"`. 
I really wasn't sure if I should respect this reason name or perhaps use `"consumable_purchase_found"`, which seems more appropriate in iOS. WDYT @RevenueCat/coresdk?

Sources/OfflineEntitlements/OfflineCustomerInfoCreator.swift
```diff
@@ -76,27 +76,51 @@ class OfflineCustomerInfoCreator {
...
+                    errorMessage = productsFetcherError.errorUserInfo[NSLocalizedDescriptionKey] as? String ?? ""
+                }
+            case let offlineCustomerInfoCreatorError as OfflineCustomerInfoCreator.Error:
+                switch offlineCustomerInfoCreatorError {
+                case .noEntitlementMappingAvailable:
+                    reason = "no_entitlement_mapping_available"
```
ajpallares, line 113: I created this new reason name, as I haven't seen a similar one in Android

Sources/OfflineEntitlements/OfflineCustomerInfoCreator.swift
tonidero: Hmm I believe this doesn't work either for non-consumables if I remember correctly? We could probably add support for this at some point but I think right now `one_time_purchase_found` is a better reflection of the reason?

ajpallares: If I'm following the code correctly, there are only two places that can cause the error.
* [Some lines above](https://github.com/RevenueCat/purchases-ios/blob/a0d8640c936f2b9a703be72128717d92f94ea332/Sources/OfflineEntitlements/OfflineCustomerInfoCreator.swift#L83), where `OfflineCustomerInfoCreator.Error.noEntitlementMappingAvailable` is thrown.
* [here](https://github.com/RevenueCat/purchases-ios/blob/a0d8640c936f2b9a703be72128717d92f94ea332/Sources/OfflineEntitlements/PurchasedProductsFetcher.swift#L94) where `PurchasedProductsFetcher.Error.foundConsumablePurchase` is thrown.
 
That's why I only considered these two error types. I don't really see anywhere else in code where an error is thrown when trying to compute the offline entitlements.

I'm ok with using the same for now, yes 👍 
Thanks @tonidero!

Tests/UnitTests/OfflineEntitlements/OfflineCustomerInfoCreatorTests.swift
```diff
@@ -67,6 +94,42 @@ class OfflineCustomerInfoCreatorTests: TestCase {
...
+
+        do {
+            _ = try await self.creator.create(for: "user")
+            fail("Expected error")
+        } catch {
+            expect(self.mockDiagnosticsTracker.trackedErrorEnteringOfflineEntitlementsModeCalls.value.count) == 1
+            let (reason, msg) = self.mockDiagnosticsTracker.trackedErrorEnteringOfflineEntitlementsModeCalls.value[0]
+            expect(reason) == "one_time_purchase_found"
+            expect(msg) == Strings.offlineEntitlements
+                .computing_offline_customer_info_for_consumable_product.description
```
ajpallares, line 113: I'm aware that with the current implementation, this check is a bit too strong in the sense that it depends, not only on the implementation of `OfflineCustomerInfoCreatorTests`, but also on the implementation of `PurchasedProductsFetcher.Error`. But I think it's worth it for now to always have a consistent diagnostics report

Sources/OfflineEntitlements/OfflineCustomerInfoCreator.swift
tonidero: Right that second error is not fully named correctly... Looking at the underlying implementation, it checks for non-subscriptions: https://github.com/RevenueCat/purchases-ios/blob/9d6cdd87e4bd1612346615809ee40c4638b168ed/Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift#L60

Sources/OfflineEntitlements/OfflineCustomerInfoCreator.swift
tonidero: Love this! We should add that to Android as well yeah... Can you add a comment in the android cell in the doc with maybe a link here and mark it as "yellow" so we don't forget?

tonidero, line 113: I'm thinking whether it's worth to add an enum here for this now that we have 3 values... I don't think it's a blocker for this PR though.

tonidero (approved):
Nice work!!

PR #4868, opened 2025-03-11
[Diagnostics] fix diagnostics sync retry logic

ajpallares:
### Checklist

### Motivation
The diagnostics sync retry logic is different between iOS and Android, being the logic in Android the one that makes more sense, as discussed [here](https://github.com/RevenueCat/purchases-ios/pull/4863#issuecomment-2714070723).

### Description
This PR equals the logic of diagnostics sync retry to that in Android, so that failed diagnostics syncs are only retried when
* Receiving a status code between 500 and 599
* Receiving an error of the case `NetworkError.networkError`, which should be similar to Android's check of `error.code == PurchasesErrorCode.NetworkError`.


JayShortway (commented):
I think this makes a lot of sense, thanks for taking care of it! 

(Seems that this PR should target `Diagnostics/error_entering_offline_entitlements_mode`, or wait for that one to be merged, to clean up the diff?)

ajpallares:
> (Seems that this PR should target `Diagnostics/error_entering_offline_entitlements_mode`, or wait for that one to be merged, to clean up the diff?)

Right @JayShortway! I  was faster than GH's automerge 😅. Sorry about that!

JayShortway (approved):

tonidero (approved):
Looks great, thanks for doing this!!

PR #4880, opened 2025-03-13
feat: Refresh customer center and purchases after restore

facumenzella:
### Motivation
We need to keep up-to-date customer center after a successful restore

### Description
This PR just reloads the Customer Center after the restore works. I think we should not leave the screen in an inconsistent state, so if the restore works, but the load screen fails it will error out.

> There's room for a better UI effect, but this does the trick for now


ajpallares (approved):
Makes sense!

vegaro (approved):

PR #4881, opened 2025-03-13
[Diagnostics] add `get_customer_info_started` and `get_customer_info_result` events

ajpallares:
### Checklist

### Description
This PR implements the get_customer_info_started` and `get_customer_info_result` diagnostics events

### DIFF
--- a/Sources/Diagnostics/DiagnosticsEvent.swift
+++ b/Sources/Diagnostics/DiagnosticsEvent.swift
@@ -47,6 +47,8 @@ struct DiagnosticsEvent: Codable, Equatable {
         case maxDiagnosticsSyncRetriesReached = "max_diagnostics_sync_retries_reached"
         case getOfferingsStarted = "get_offerings_started"
         case getOfferingsResult = "get_offerings_result"
+        case getProductsStarted = "get_products_started"
+        case getProductsResult = "get_products_result"
         case getCustomerInfoStarted = "get_customer_info_started"
         case getCustomerInfoResult = "get_customer_info_result"
     }
--- a/Sources/Diagnostics/DiagnosticsEvent.swift
+++ b/Sources/Diagnostics/DiagnosticsEvent.swift
@@ -86,7 +86,6 @@ struct DiagnosticsEvent: Codable, Equatable {
         let cacheStatus: CacheStatus?
         let cacheFetchPolicy: String?
         let hadUnsyncedPurchasesBefore: Bool?
-        let usedOfflineEntitlements: Bool?
         let isRetry: Bool?
 
         init(verificationResult: String? = nil,
@@ -110,7 +109,6 @@ struct DiagnosticsEvent: Codable, Equatable {
              cacheStatus: CacheStatus? = nil,
              cacheFetchPolicy: CacheFetchPolicy? = nil,
              hadUnsyncedPurchasesBefore: Bool? = nil,
-             usedOfflineEntitlements: Bool? = nil,
              isRetry: Bool? = nil) {
             self.verificationResult = verificationResult
             self.endpointName = endpointName
@@ -133,7 +131,6 @@ struct DiagnosticsEvent: Codable, Equatable {
             self.cacheStatus = cacheStatus
             self.cacheFetchPolicy = cacheFetchPolicy.map { $0.diagnosticsName }
             self.hadUnsyncedPurchasesBefore = hadUnsyncedPurchasesBefore
-            self.usedOfflineEntitlements = usedOfflineEntitlements
             self.isRetry = isRetry
         }
 
--- a/Sources/Diagnostics/DiagnosticsEvent.swift
+++ b/Sources/Diagnostics/DiagnosticsEvent.swift
@@ -59,7 +59,7 @@ struct DiagnosticsEvent: Codable, Equatable {
     }
 
     enum OfflineEntitlementsModeErrorReason: String, Codable, Equatable {
-        case oneTimePurchaseFound = "ONT_TIME_PURCHASE_FOUND"
+        case oneTimePurchaseFound = "ONE_TIME_PURCHASE_FOUND"
         case noEntitlementMappingAvailable = "NO_ENTITLEMENT_MAPPING_AVAILABLE"
         case unknown = "UNKNOWN"
     }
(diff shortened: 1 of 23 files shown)


Sources/Identity/CustomerInfoManager.swift
```diff
@@ -183,12 +217,13 @@ class CustomerInfoManager {
...
                                                    isAppBackgrounded: isAppBackgrounded,
                                                    completion: completion)
+                    // TODO: Track getCustomerInfoResult if !completionCalled
                 }
             }
         }
     }
 
-    func cachedCustomerInfo(appUserID: String) -> CustomerInfo? {
+    func cachedCustomerInfo(appUserID: String) throws -> CustomerInfo? {
```
ajpallares, line 226: I've changed this method to be throwing so that diagnostics event can gather more information about why fetching from the cache didn't work. 

I'm not sure if this is worth it instead of sending the same `BackendError.missingCachedCustomerInfo()` that we send to callers, as this will only be useful when using the `.fromCacheOnly` cache policy (as the other policies using the cache will fall back to fetching the latest one in case of error).

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -458,6 +580,58 @@ private extension CustomerInfoManager {
...
+                                                    result: Swift.Result<CustomerInfo, Error>) {
+        if #available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *), trackDiagnostics {
+
+            let error: PurchasesError?
+            switch result.error {
+            case let purchasesError as PurchasesError:
+                error = purchasesError
+            case let purchasesErrorConvertible as PurchasesErrorConvertible:
+                error = purchasesErrorConvertible.asPurchasesError
+            case let otherError:
```
tonidero, line 619: Hmm I'm wondering, will it ever enter here? Seems like for any `Swift.Error`, it will enter the previous case if I understand correctly?

Sources/Identity/CustomerInfoManager.swift
```diff
@@ -334,11 +380,13 @@ extension CustomerInfoManager {
 
     func customerInfo(
         appUserID: String,
+        trackDiagnostics: Bool,
```
tonidero, line 383: I wonder if we should default this to `false`, since seems to be the main use of this and just set it to `true` where needed... NABD

Sources/Identity/CustomerInfoManager.swift
tonidero: Hmm I see we were swallowing any exception before... I think it's fine to expose that to get more details of any failures yeah... I think this is ok!

tonidero (approved):
Looking great! Left some comments but nothing really blocking. Great job again!

Sources/Diagnostics/DiagnosticsEvent.swift
```diff
@@ -47,19 +47,21 @@ struct DiagnosticsEvent: Codable, Equatable {
...
+        case unverified = "UNVERIFIED"
+        case userCancelled = "USER_CANCELLED"
+        case pending = "PENDING"
     }
 
     enum OfflineEntitlementsModeErrorReason: String, Codable, Equatable {
-        case oneTimePurchaseFound = "one_time_purchase_found"
-        case noEntitlementMappingAvailable = "no_entitlement_mapping_available"
-        case unknown
+        case oneTimePurchaseFound = "ONT_TIME_PURCHASE_FOUND"
```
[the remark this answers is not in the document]
ajpallares: Great! And, btw, thanks to your comment I realized that the value of `oneTimePurchaseFound` has a misspelling 😱 `ONT_TIME_PURCHASE_FOUND`
Lucky us that you commented here! 😅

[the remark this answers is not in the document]
tonidero: Whoops haha nice catch!

Sources/Identity/CustomerInfoManager.swift
ajpallares: Not exactly. Because not all error types implement the `PurchasesErrorConvertible` protocol. E.g. a `DecodingError` in `func cachedCustomerInfo(appUserID: String) throws -> CustomerInfo?` will fall into this `otherError` category, because `DecodingError` does not implement our `PurchasesErrorConvertible` protocol.

In any case, we need to deal with the `default` case in the `switch` to make it exhaustive.

Sources/Identity/CustomerInfoManager.swift
ajpallares: Yes, I went back and forth with this one and I couldn't really make up my mind. So I agree that it's cleaner if it's going to be `false` most of the times anyway

Sources/Identity/CustomerInfoManager.swift
ajpallares: But to be clear, even with this method now `throwing`, we keep swallowing these errors in the SDK and the behavior for developer calls that go through this method won't change. 
You can see it [here](https://github.com/RevenueCat/purchases-ios/blob/09279498f4591b9fdecbb605a75837a6c71f300c/Sources/Identity/CustomerInfoManager.swift#L125)
```swift
                let result = Result { try self.cachedCustomerInfo(appUserID: appUserID) }

                // We want the specific error for diagnostics
                let resultForDiagnostics = Result(result.value as? CustomerInfo,
                                                  result.error ?? BackendError.missingCachedCustomerInfo())
                self.trackGetCustomerInfoResultIfNeeded(trackDiagnostics: trackDiagnostics,
                                                        startTime: startTime,
                                                        cacheFetchPolicy: fetchPolicy,
                                                        hadUnsyncedPurchasesBefore: nil,
                                                        usedOfflineEntitlements: false,
                                                        result: resultForDiagnostics)

                // But for callers we only pass `.missingCachedCustomerInfo()` error
                completion?(
                    Result(result.value as? CustomerInfo, .missingCachedCustomerInfo())
                )
```
where we only used the newly-thrown error for diagnostics (`resultForDiagnostics`) but for customers we keep the same behavior of throwing the `.missingCachedCustomerInfo()` error in case of `nil` `CustomerInfo`.

Does it make sense?

tonidero: Yup yup! We are now only "processing" errors for diagnostics, but otherwise keep swallowing those errors. Makes sense for this PR 👍

PR #4882, opened 2025-03-13
feat: Add basic test for CustomerCenter restoreStarted preferenceKey

facumenzella:
### Motivation
We've recently added preference keys to CustomerCenter. 

### Description
This adds an example test on how to test these from now on.


Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewPreferenceKeyTests.swift
```diff
@@ -0,0 +1,45 @@
...
+    func testActionWrapperAndModifiers() async {
+        let actionWrapper = await CustomerCenterActionWrapper()
+
+        var onCustomerCenterRestoreStarted = false
+        await confirmation { @MainActor confirmation1 in
+            let testView = Text("test")
+                .modifier(CustomerCenterActionViewModifier(actionWrapper: actionWrapper))
+                .onCustomerCenterRestoreStarted {
+                    onCustomerCenterRestoreStarted = true
+                    confirmation1()
```
JayShortway, line 32: I'm guessing this is what makes the test pass? How long will it wait for this? Is that timeout configurable? Just curious. 

Also, what's the `onCustomerCenterRestoreStarted` variable used for?

Package.swift
```diff
@@ -38,6 +38,7 @@ let shouldIncludeDocCPlugin = environmentVariables["INCLUDE_DOCC_PLUGIN"] == "tr
 
 var dependencies: [Package.Dependency] = [
     .package(url: "https://github.com/quick/nimble", revision: "1f3bde57bde12f5e7b07909848c071e9b73d6edc"),
+    .package(url: "https://github.com/swiftlang/swift-testing", revision: "18c42c19cac3fafd61cab1156d4088664b7424ae"),
```
JayShortway, line 41: I see we do this for nimble too, but can we not define a tag to depend on? A commit hash is so opaque haha.

JayShortway (commented):
I can't really comment on whether it is better, but it surely looks readable and straightforward! Had a few questions out of curiosity.

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewPreferenceKeyTests.swift
```diff
@@ -0,0 +1,45 @@
...
+                }
+
+            let view = UIHostingController(rootView: testView)
+            let window = UIWindow()
+            window.rootViewController = view
+            window.makeKeyAndVisible()
+            view.view.layoutIfNeeded()
+
+            actionWrapper.handleAction(.restoreStarted)
+            await Task.yield()
```
ajpallares, line 42: This is interesting. I've checked and this is required for the test to pass, but I'm not sure if a change in the implementation of `CustomerCenterView` could make this test wrongly fail because of this (I really don't know).

According to [`Task.yield()`s documentation](https://developer.apple.com/documentation/swift/task/yield()):
> Suspends the current task and allows other tasks to execute.

I'm guessing it suspends the current task until all other tasks finish execution?

If this is robust, I like it way more than the classic `waitForExpectations(timeout: )` because it means we don't need to specify a seemingly random timeout that could be flaky in other environments (e.g. slow CI machine)

ajpallares (approved):
Looks great! I'm all in for SwiftTesting if it covers all our needs. And we don't necessarily need to make a explicit migration, but rather start using it for new tests

Package.swift
aboedo: for nimble it was specifically because we needed something for compatibility and the next major didn't work and I can't recall the details, but for new things I would go with up to next major by default

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewPreferenceKeyTests.swift
facumenzella: > I'm guessing this is what makes the test pass?

Exactly, as the test is written it means "it will succeed unless the confirmation is not confirmed".

---

> How long will it wait for this? Is that timeout configurable?

As far as I understand, there's no explicit timeout because it relies on structured concurrency. The `await confirmation` will fail if it's not confirmed within the scope of the block.

[Task.yield](https://developer.apple.com/documentation/swift/task/yield()) suspends the confirmation block, and lets another task run. If we remove the yield, the confirmation just fails. I think we could rely on another confirmation block to timeout, but I need to dig into it.

---

> onCustomerCenterRestoreStarted

This is a leftover of other trials and errors. I'll remove it

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewPreferenceKeyTests.swift
```diff
@@ -0,0 +1,45 @@
...
+@testable import RevenueCatUI
+import SwiftUI
+import Testing
+
+@Suite
+struct CustomerCenterActionWrapperTests {
+
+    @Test
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, visionOS 1.0, *)
+    func testActionWrapperAndModifiers() async {
```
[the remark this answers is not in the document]
aboedo: nitpick: name the test after what was supposed to happen, i.e.: `actionWrapperHandlesRestoreActionFromModifier`, such that when you see the failing test name you instantly know what exactly you broke

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewPreferenceKeyTests.swift
facumenzella: That said, if it proves to be flaky we could write a handmade timeout mechanism at the end of the test.

Package.swift
JayShortway: Alright, thanks for the context both! Yea I think a version/tag would be more easily parseable by a human.

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewPreferenceKeyTests.swift
JayShortway: Ah got it, makes sense! Coroutines has similar mechanisms for testing, nice. It might couple the test to the implementation somewhat (breaks if the implementation no longer uses a `Task`?), but I think that's fine.

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewPreferenceKeyTests.swift
JayShortway: My understanding of a similar mechanism in Coroutines means that this could indeed break if the `CustomerCenterView` implementation no longer uses a `Task`. It's coupled to that implementation detail. I'm okay to live with that, but something to be aware of (if my understanding is correct).

ajpallares: That's an excellent point and it makes total sense. I think it's ok in this case, although it would probably be a headache for the one encountering that the test suddenly starts to fail. 
That said, I think the more natural direction for us is to move towards `Task` rather than away from it.

JayShortway: Yes, agreed!

Package@swift-5.9.swift
```diff
@@ -121,7 +121,7 @@ let package = Package(
                         .product(name: "Nimble", package: "nimble"),
                         .product(name: "SnapshotTesting", package: "swift-snapshot-testing")
                     ],
-                    exclude: ["Templates/__Snapshots__", "Data/__Snapshots__", "TestPlans"],
+                    exclude: ["Templates/__Snapshots__", "Data/__Snapshots__", "TestPlans", "**/Swift6/**"],
```
ajpallares, line 124: I'm curious. What does the `"**/Swift6/**"` do?

facumenzella:
I've rewritten this with XCTest https://github.com/RevenueCat/purchases-ios/pull/4916, so closing this one

PR #4886, opened 2025-03-14
[Diagnostics] add `purchase_started` and `purchase_result` events

ajpallares:
### Checklist

### Description
This PR implements the `purchase_started` and `purchase_result` diagnostics events


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -416,27 +417,44 @@ final class PurchasesOrchestrator {
                   promotionalOffer: PromotionalOffer.SignedData? = nil,
                   winBackOffer: WinBackOffer? = nil,
                   metadata: [String: String]? = nil,
+                  trackDiagnostics: Bool,
```
ajpallares, line 420: I'm not very convinced about the `trackDiagnostics` Boolean parameter here because it currently only affects the `purchase_started` and `purchase_result` event, whereas, more internally in the call tree there's the `apple_purchase_attempt` event which is not affected by it.

Instead, I'm thinking of adding a `CallerType` internal enum that specifies whether the method was called from the API or internally by the SDK, that way the `purchase_started` and `purchase_result` events would only be sent if the call happens from the API.

WDYT @RevenueCat/coresdk?

I'm aware that this perhaps overcomplicates things, but it'd be more consistent. Otherwise, I don't know if that boolean could lead to future errors 🤔

tonidero: Hmm I see what you mean... I didn't remember that the listener wrapper methods also use the `purchase` methods which seems a bit weird TBH... I would have assumed that purchases are only triggered by the public APIs...

I don't remember how that worked too well, going to take a look and will let you know.. But assuming a purchase can be triggered from outside the public APIs, I think what you mentioned would probably be the appropriate solution yeah...

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -416,27 +417,44 @@ final class PurchasesOrchestrator {
...
                   completion: @escaping PurchaseCompletedBlock) {
         Self.logPurchase(product: product, package: package, offer: promotionalOffer)
 
+        self.trackPurchaseStartedIfNeeded(trackDiagnostics: trackDiagnostics,
+                                          productId: product.productIdentifier,
+                                          productType: product.productType)
+        let startTime = self.dateProvider.now()
+
+        let completionWithTracking: PurchaseCompletedBlock =
+        { [weak self] transaction, customerInfo, error, userCancelled in
```
tonidero, line 430: Ohh I haven't done the `[weak self]` in my completion blocks 😬. Will need to do that, good thinking!

tonidero (commented):
Looking good but let me confirm how those purchase methods in the listeners worked 😅

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
tonidero: Ohh I see these are for `PurchaseIntents` and `PromotedPurchases`... Yeah, I guess it's not technically part of the public API, but just something we handle.

tonidero: Right, so I feel the solution you proposed is probably the correct solution for this... It might be a bigger refactor as you said, but probably good in the long run. Ideally we modify the other events that use a similar system to also use something like the `CallerType`, but I think that all can come in a separate PR. Not a blocker IMO.

We probably will also need to add another event for promoted purchases I think, which I had not added to the list.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
tonidero: It's probably not a huge deal since this is mostly a singleton... But we should avoid unnecessary leaks in case the SDK is configured multiple times.

ajpallares: Exactly. Not a big deal but I try to do it if it makes sense to avoid leaks, just in case

tonidero (approved):
I think this looks good! Regarding the change to `CallerType` I think it can be done separately if needed.

PR #4885, opened 2025-03-14
[Diagnostics] Add sync purchases tracking (part 1)

tonidero:
### Description
This PR:
- Adds the `sync_purchases_started` and `sync_purchases_result` events to `DiagnosticsTracker` (currently unused, will be in a follow-up PR).
- Refactors the code inside the `DiagnosticsTracker` to avoid some repetition

PR #4887, opened 2025-03-14
[Diagnostics] Add sync and restore purchases events

tonidero:
### Description
Follow-up to #4885.

 This PR adds 4 new events:
- `sync_purchases_started`
- `sync_purchases_result`
- `restore_purchases_started`
- `restore_purchases_result`

The cases of sync and restore are pretty similar right now in the code, so I shared most of the code paths. I decided against joining them in a single event since they are slightly different operations after all.

### DIFF
--- a/Tests/UnitTests/Purchasing/Purchases/PurchasesRestoreTests.swift
+++ b/Tests/UnitTests/Purchasing/Purchases/PurchasesRestoreTests.swift
@@ -274,3 +274,78 @@ class PurchasesRestoreNoSetupTests: BasePurchasesTests {
     }
 
 }
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+class PurchasesRestoreTrackingTests: BasePurchasesTests {
+
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+
+        self.setupPurchases()
+    }
+
+    func getMockDiagnosticsTracker() -> MockDiagnosticsTracker {
+        // swiftlint:disable:next force_cast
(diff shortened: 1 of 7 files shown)


Tests/UnitTests/Purchasing/Purchases/PurchasesRestoreTests.swift
```diff
@@ -274,3 +274,78 @@ class PurchasesRestoreNoSetupTests: BasePurchasesTests {
...
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+
+        self.setupPurchases()
+    }
+
+    func getMockDiagnosticsTracker() -> MockDiagnosticsTracker {
+        // swiftlint:disable:next force_cast
+        return self.diagnosticsTracker as! MockDiagnosticsTracker
+    }
```
ajpallares, line 290: Not a big deal, but to avoid force casting I'm doing
```suggestion
    private func getMockDiagnosticsTracker() throws -> MockDiagnosticsTracker {
        return try XCTUnwrap(self.mockDiagnosticsTracker as? MockDiagnosticsTracker)
    }
```
Although you have to call it with `try`.
This way a failure is more friendly than a crash

tonidero: That makes a lot of sense. Thank you for the suggestion! 🙌

tonidero: Done in https://github.com/RevenueCat/purchases-ios/pull/4887/commits/993aeee732ec7c333b55fb3c818b83e45ea32ee0

PR #4888, opened 2025-03-14
[Diagnostics] add `apple_purchase_intent_received` event

ajpallares:
### Checklist

### Description
This PR implements the apple_purchase_intent_received` diagnostics events

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -1206,7 +1206,6 @@ private extension PurchasesOrchestrator {
         }
     }
 
-
     @available(iOS 16.4, macOS 14.4, *)
     @available(tvOS, unavailable)
     @available(watchOS, unavailable)
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -1206,6 +1206,28 @@ private extension PurchasesOrchestrator {
         }
     }
 
+
+    @available(iOS 16.4, macOS 14.4, *)
+    @available(tvOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(visionOS, unavailable)
+    func trackApplePurchaseIntentReceivedIfNeeded(purchaseIntent: StoreKit.PurchaseIntent) {
+        let offerId: String?
+        let offerType: String?
+
+        if #available(iOSApplicationExtension 18.0, *) {
+            offerId = purchaseIntent.offer?.id
+            offerType = purchaseIntent.offer?.type.rawValue
+        } else {
+            offerId = nil
+            offerType = nil
+        }
+
+        self.diagnosticsTracker?.trackPurchaseIntentReceived(productId: purchaseIntent.product.id,
+                                                             offerId: offerId,
+                                                             offerType: offerType)
+    }
+
     /// - Parameter restored: whether the transaction state was `.restored` instead of `.purchased`.
     private func purchaseSource(
         for productIdentifier: String,
@@ -1300,6 +1322,8 @@ extension PurchasesOrchestrator: StoreKit2PurchaseIntentListenerDelegate {
 
         guard let purchaseIntent = purchaseIntent.purchaseIntent else { return }
         let storeProduct = StoreProduct(sk2Product: purchaseIntent.product)
+
+        self.trackApplePurchaseIntentReceivedIfNeeded(purchaseIntent: purchaseIntent)
 
         delegate?.readyForPromotedProduct(storeProduct) { completion in
 
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -1214,7 +1214,7 @@ private extension PurchasesOrchestrator {
         let offerId: String?
         let offerType: String?
 
-        if #available(iOSApplicationExtension 18.0, *) {
+        if #available(iOS 18.0, macOS 15.0, *) {
             offerId = purchaseIntent.offer?.id
             offerType = purchaseIntent.offer?.type.rawValue
         } else {
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -1210,7 +1210,7 @@ private extension PurchasesOrchestrator {
     @available(tvOS, unavailable)
     @available(watchOS, unavailable)
     @available(visionOS, unavailable)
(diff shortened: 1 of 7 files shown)


Sources/Purchasing/StoreKit2/StoreKit2PurchaseIntentListener.swift
```diff
@@ -139,6 +139,39 @@ struct StorePurchaseIntent: Sendable, Equatable {
...
+        return true
+    #endif
+    }
+}
+
+@available(iOS 16.4, macOS 14.4, *)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+@available(visionOS, unavailable)
+protocol StoreKit2PurchaseIntentType: Equatable, Identifiable, Sendable {
```
ajpallares, line 158: I introduced this protocol to be able to test that the delegate call in `PurchasesOrchestrator.swift` does indeed call the track diagnostics method. Another option would have been to use a wrapper with a generic, but I didn't want to overcomplicate things so much only to have type safety in unit tests

Tests/StoreKitUnitTests/PurchasesOrchestratorCommonTests.swift
```diff
@@ -495,4 +497,58 @@ class PurchasesOrchestratorTrackingTests: BasePurchasesOrchestratorTests {
...
+// MARK: - Purchase Intent Received events
+
+@available(iOS 16.4, macOS 14.4, *)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+@available(visionOS, unavailable)
+final class MockStoreKit2PurchaseIntent: StoreKit2PurchaseIntentType {
+
+    let product: StoreKit.Product
+    let offer: StoreKit.Product.SubscriptionOffer? = nil
```
ajpallares, line 514: I think there's not an easy way to mock `StoreKit.Product.SubscriptionOffer` without introducing more unnecessary complexity, so for now I have this be `nil` in mocks, as we only need this for tracking the `apple_purchase_intent_received` event

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1296,16 +1321,18 @@ extension PurchasesOrchestrator: StoreKit2PurchaseIntentListenerDelegate {
...
+
         delegate?.readyForPromotedProduct(storeProduct) { completion in
 
             var attemptedToPurchaseWithASubscriptionOffer = false
 
-            if #available(iOS 18.0, macOS 15.0, visionOS 2.0, *) {
+            if #available(iOS 18.0, macOS 15.0, *) {
                 #if compiler(>=6.0)
                 if let offer = purchaseIntent.offer {
                     switch offer.type {
```
JayShortway, line 1338: Curious about the `visionOS` changes here. Why were they needed?

Sources/Purchasing/StoreKit2/StoreKit2PurchaseIntentListener.swift
```diff
@@ -60,7 +60,7 @@ actor StoreKit2PurchaseIntentListener: StoreKit2PurchaseIntentListenerType {
 
     init(delegate: StoreKit2PurchaseIntentListenerDelegate? = nil) {
 
-        #if compiler(>=5.10)
+        #if compiler(>=5.10) && !os(tvOS) && !os(watchOS) && !os(visionOS)
```
JayShortway, line 63: Same curiosity about these added checks (and others in this file).

Tests/StoreKitUnitTests/PurchasesOrchestratorCommonTests.swift
```diff
@@ -495,4 +497,58 @@ class PurchasesOrchestratorTrackingTests: BasePurchasesOrchestratorTests {
...
+
 }
+
+@available(iOS 16.4, macOS 14.4, *)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+@available(visionOS, unavailable)
+extension PurchasesOrchestratorTrackingTests {
+
+    func test() async throws {
```
JayShortway, line 534: Could this have a more descriptive name? Something with "should" is always helpful in test names imo.

(Also, just curious: why is it in an extension?)

JayShortway (commented):
Looks good! Just have some questions and a rename request.

Tests/StoreKitUnitTests/PurchasesOrchestratorCommonTests.swift
ajpallares: Oh my! thanks for flagging 🙏 
Left the name by mistake

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
ajpallares: Good question!

I had to add the `StoreKit2PurchaseIntentType` protocol to be able to mock the object `StoreKit.PurchaseIntent` in unit tests. For reasons I don't understand, because this protocol is marked with `@available(tvOS, unavailable)` and `@available(visionOS, unavailable)`, the project was failing to build in tvOS and visionOS (you can check in commit https://github.com/RevenueCat/purchases-ios/pull/4888/commits/6927f0657991fb38cb1c6e85ffc5b490d97f9eda, that is prior to adding the compiler checks)

So I had to add the compile-time checks to fix the compilation issues in tvOS and visionOS. It makes sense anyway because `StoreKit.PurchaseIntent` is only available in iOS and macOS.

Tests/StoreKitUnitTests/PurchasesOrchestratorCommonTests.swift
ajpallares: > Also, just curious: why is it in an extension?

Not really required to be in an extension but it's more convenient for several reasons:
The main one in this case is that we can add the `availability` checks in the extension level (and avoid having to do per test) and, in this case, note that the availability check for the `PurchasesOrchestratorTrackingTests` extension
```
@available(iOS 16.4, macOS 14.4, *)
```
is different to that from the `PurchasesOrchestratorTrackingTests` class itself
```
```

Also, our linter defines a limit on the number of lines in a class, so doing the extension is a way to silence the warning (yeah, perhaps this is not the best argument 😄)

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
JayShortway: Ah yes, I see it's only available in iOS and macOS. Thanks for explaining!

Tests/StoreKitUnitTests/PurchasesOrchestratorCommonTests.swift
JayShortway: Haha, a lot of my tests start their life as `test()` too 😆 

Ah, TIL about the difference in `availability` checks. That's a cool trick!

JayShortway (approved):

ajpallares:
Hit the merge button too soon. Will reopen to merge after #4886 is merged

PR #4889, opened 2025-03-14
Fix for Chinese locale scripts that aren't supported in iOS 15

joshdholtz:
### Motivation

Tests were failing in iOS 15 when trying to get the `Locale`'s `scriptCode` (it was always returning nil)

### Description

Hardcoding script code for the locales that we know for iOS 15 back support

### DIFF
--- a/RevenueCatUI/Templates/V2/Localizations/LocaleExtensions.swift
+++ b/RevenueCatUI/Templates/V2/Localizations/LocaleExtensions.swift
@@ -44,10 +44,10 @@ extension Locale {
         if #available(macOS 13, iOS 16, tvOS 16, watchOS 9, visionOS 1.0, *) {
             return self.language.script?.identifier
         } else {
-            return self.scriptCode
+            return self.scriptCode ?? self.fallbackScriptForiOS15
         }
         #else
-        return self.scriptCode
+        return self.scriptCode ?? self.fallbackScriptForiOS15
         #endif
     }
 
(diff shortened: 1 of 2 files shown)


joshdholtz:
@RCGitBot please test

RevenueCatUI/Templates/V2/Localizations/LocaleExtensions.swift
```diff
@@ -100,4 +100,19 @@ extension Locale {
         return self.rc_languageCode.map(Locale.init(identifier:))
     }
 
+    /// iOS 15 returns a nil scrippCode for these locale identifiers so hardcoding fallbacks
+    private var fallbackScriptForiOS15: String? {
+        let map: [String: String] = [
+            "zh": "Hans",
```
JayShortway, line 106: I'm not entirely sure this is correct? `zh` provides no information regarding Traditional vs Simplified, can we assume Simplified in that case? Can I read about this somewhere? 😄

JayShortway (commented):
Looks good! Just some doubts about the assumptions for `zh`.

RevenueCatUI/Templates/V2/Localizations/LocaleExtensions.swift
JayShortway: FWIW In Android I only inferred the script from the region [here](https://github.com/RevenueCat/purchases-android/blob/958320fe4ced66a5dfc5ae46d3f8ff09f2101560/ui/revenuecatui/src/main/kotlin/com/revenuecat/purchases/ui/revenuecatui/components/ktx/Localization.kt#L125-L132).

```kotlin
/**
 * Scripts inferred from the region.
 */
private val scriptByRegion = mapOf(
    "CN" to "Hans",
    "SG" to "Hans",
    "MY" to "Hans",
    "TW" to "Hant",
    "HK" to "Hant",
    "MO" to "Hant",
)
```

JayShortway:
@RCGitBot please test

RevenueCatUI/Templates/V2/Localizations/LocaleExtensions.swift
joshdholtz: Ah yeah, probably safer to remove it!

joshdholtz:
@RCGitBot please test

vegaro (approved):

PR #4890, opened 2025-03-14
[Paywall] Center-align all Timeline component  icons by calculating and setting their max width

MarkVillacampa:
Icons with different widths are now center-aligned:

Screenshot 2025-03-14 at 14 37 21


joshdholtz (approved):
:shipit:

PR #4892, opened 2025-03-14
Fix possible memory leaks with diagnostics completion blocks

tonidero:
### Description
We were not doing `weak self` in the completion blocks for diagnostics. This fixes that. The chances of a leak here are small, considering these classes are usually singletons (unless the SDK is configured multiple times), but better to do it correctly.


ajpallares (approved):
Better safe than sorry 😄

PR #4893, opened 2025-03-17
[Diagnostics] add `apple_present_code_redemption_sheet_request` event

ajpallares:
### Checklist

### Description
This PR implements the `apple_present_code_redemption_sheet_request` diagnostics event

### DIFF
--- a/Tests/UnitTests/Mocks/MockDiagnosticsTracker.swift
+++ b/Tests/UnitTests/Mocks/MockDiagnosticsTracker.swift
@@ -292,4 +292,9 @@ final class MockDiagnosticsTracker: DiagnosticsTrackerType, Sendable {
             $0.append((errorMessage, errorCode, responseTime))
         }
     }
+
+    let trackedApplePresentCodeRedemptionSheetRequestCalls: Atomic<Int> = .init(0)
+    func trackApplePresentCodeRedemptionSheetRequest() {
+        self.trackedApplePresentCodeRedemptionSheetRequestCalls.modify { $0 += 1 }
+    }
 }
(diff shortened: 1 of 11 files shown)


Tests/UnitTests/Mocks/MockDiagnosticsTracker.swift
```diff
@@ -292,4 +292,9 @@ final class MockDiagnosticsTracker: DiagnosticsTrackerType, Sendable {
             $0.append((errorMessage, errorCode, responseTime))
         }
     }
+
+    let trackedApplePresentCodeRedemptionSheetRequestCalls: Atomic<Int> = .init(0)
```
JayShortway, line 296: This is not for this PR, and maybe not for any PR, but why doesn't this class keep track of the `DiagnosticsEvent`s in an array? That makes it easy to check which events were tracked, and avoids the need to add these `*Calls` and `*Params` fields for every new event. More of an idea for future improvements haha.

JayShortway (approved):
Very nice!

Tests/UnitTests/Mocks/MockDiagnosticsTracker.swift
ajpallares: Yeah, that's a good point.

The only limitation I see here is that we want to track not only the events sent but also the parameters sent on each event. Doing so in an array would mean having an heterogeneous array of event parameters.

That said, it can be useful to have it all centralized in an array so that we can check also the order of events. For example, in #4886, it would allow us to check that the `started` event was sent before the `result` event.

I think I could easily change the implementation to have this and also to avoid having to edit the `MockDiagnosticsTracker` for each new event added.

Thanks for pointing this out @JayShortway! 🙌

JayShortway: >would mean having an heterogeneous array of event parameters. 
  
Yea true, but in a test that might be fine. We know exactly what we expect to be there. 

>it would allow us to check that the started event was sent before the result event.

Yea indeed! 😄

PR #4894, opened 2025-03-17
[Diagnostics] add `apple_trial_or_intro_eligibility_request` event

ajpallares:
### Checklist

### Description
This PR implements the `apple_trial_or_intro_eligibility_request` diagnostics event


Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
```diff
@@ -267,3 +287,68 @@ extension TrialOrIntroPriceEligibilityChecker {
...
+            let purchasesError = purchasesErrorConvertible.asPurchasesError
+            errorCode = purchasesError.errorCode
+            errorMessage = purchasesError.localizedDescription
+        case let receiptParserError as PurchasesReceiptParser.Error:
+            errorCode = ErrorCode.invalidReceiptError.rawValue
+            errorMessage = receiptParserError.errorDescription ?? receiptParserError.localizedDescription
+        case let otherError:
+            errorCode = otherError != nil ? ErrorCode.unknownError.rawValue : nil
+            errorMessage = otherError?.localizedDescription
+        }
```
ajpallares, line 337: I'm not really sure about this implementation for `errorCode` and `errorMessage`, but based on the errors thrown in the implementations, I think that this could include useful information about the errors

Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
```diff
@@ -267,3 +287,68 @@ extension TrialOrIntroPriceEligibilityChecker {
...
+                case .unknown:
+                    $0.0 += 1
+                case .ineligible:
+                    $0.1 += 1
+                case .eligible:
+                    $0.2 += 1
+                case .noIntroOfferExists:
+                    $0.3 += 1
+                }
+            }
```
ajpallares, line 318: Note that this would mean sending `nil` counts only when `result` is empty. Because `requestedProductIds` is never empty at this point, this can only happen in case of error necessarily. 

However, not all errored cases will necessarily return an empty `result`, as there are some fallback checks when certain errors happen (e.g. if using SK1 and eligibility cannot be calculated from the local receipt, then the info is requested from the backend).

This means that there can be situations where `error != nil` but still we've returned the nonempty `result`.

Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
```diff
@@ -267,3 +287,68 @@ extension TrialOrIntroPriceEligibilityChecker {
...
+            errorCode = ErrorCode.invalidReceiptError.rawValue
+            errorMessage = receiptParserError.errorDescription ?? receiptParserError.localizedDescription
+        case let otherError:
+            errorCode = otherError != nil ? ErrorCode.unknownError.rawValue : nil
+            errorMessage = otherError?.localizedDescription
+        }
+
+        let responseTime = self.dateProvider.now().timeIntervalSince(startTime)
+
+        diagnosticsTracker.trackAppleTrialOrIntroEligibilityRequest(wasSuccessful: error == nil,
```
ajpallares, line 341: I have my doubts here. When should we consider it successful? when `error == nil` or when `result` is not empty?

Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
ajpallares: That's a very good point. I think we could do what you propose, to have a unified way of sending `errorMessage` and `errorCode` parameters in all diagnostics events.

I'll try to solve this thing in a separate PR. I think that, ideally, the same errors in different diagnostics events should be sent the same way.

Sources/Purchasing/TrialOrIntroPriceEligibilityChecker.swift
ajpallares: > Hmm regarding the tracking, I think we might even be able to just not send this property, since it does seem a bit superfluous with the error properties TBH... We do have the same in some other events, but seems a bit unnecessary...

Yep, in this case it seems unnecessary. Even more so if the meaning/behavior of the successful property is different between SK1 and SK2.

> Hmm I'm wondering, the public API never returns an error here, so we do have to do the mapping at some point even in SK1 right?

Exactly. The public API never returns an error. It only returns a dictionary of `productId` to `Eligibility`. In the case of failure to parse the receipt and failure to get the products, the `checkEligibility` method for SK1 returns an empty dictionary, while the errored case in SK2 returns a dictionary with all eligibilities set to `.unknown`.

I don't know if we could consider this to be a bug in the SK1 implementation. If that's the case we could do:
1. Fix the SK1 implementation to always return the eligibility for all the `productIdentifiers`
2. And remove the `wasSuccessful` parameter from the Diagnostics event, as it would be equivalent to checking if `eligibilityUnknownCount == 0` (I think).

WDYT @tonidero?

tonidero: > And remove the wasSuccessful parameter from the Diagnostics event, as it would be equivalent to checking if eligibilityUnknownCount == 0 (I think).

I completely agree with this!

> Fix the SK1 implementation to always return the eligibility for all the productIdentifiers

I think this might be worth a discussion over Slack to confirm with other folks. Since it's a change in behavior, I would like to make sure we're not missing anything.

PR #4895, opened 2025-03-17
[Diagnostics] add `apple_purchase_intent_received` event

ajpallares:
### Checklist

### Description
This PR implements the apple_purchase_intent_received` diagnostics events

### Notes
This is a duplicated PR of #4888 which was merged too soon by mistake

### DIFF
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -1206,7 +1206,6 @@ private extension PurchasesOrchestrator {
         }
     }
 
-
     @available(iOS 16.4, macOS 14.4, *)
     @available(tvOS, unavailable)
     @available(watchOS, unavailable)
--- a/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
+++ b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
@@ -1206,6 +1206,28 @@ private extension PurchasesOrchestrator {
         }
     }
 
+
+    @available(iOS 16.4, macOS 14.4, *)
+    @available(tvOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(visionOS, unavailable)
+    func trackApplePurchaseIntentReceivedIfNeeded(purchaseIntent: StoreKit.PurchaseIntent) {
+        let offerId: String?
+        let offerType: String?
+
+        if #available(iOSApplicationExtension 18.0, *) {
+            offerId = purchaseIntent.offer?.id
+            offerType = purchaseIntent.offer?.type.rawValue
+        } else {
+            offerId = nil
+            offerType = nil
+        }
+
+        self.diagnosticsTracker?.trackPurchaseIntentReceived(productId: purchaseIntent.product.id,
+                                                             offerId: offerId,
+                                                             offerType: offerType)
+    }
+
     /// - Parameter restored: whether the transaction state was `.restored` instead of `.purchased`.
     private func purchaseSource(
         for productIdentifier: String,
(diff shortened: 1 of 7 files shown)


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1206,6 +1206,31 @@ private extension PurchasesOrchestrator {
...
+
+    @available(iOS 16.4, macOS 14.4, *)
+    @available(tvOS, unavailable)
+    @available(watchOS, unavailable)
+    @available(visionOS, unavailable)
+    func trackApplePurchaseIntentReceivedIfNeeded(purchaseIntent: any StoreKit2PurchaseIntentType) {
+        let offerId: String?
+        let offerType: String?
+
+        if #available(iOS 18.0, macOS 15.0, *) {
```
tonidero, line 1219: I see we're also doing `#if compiler(>=6.0)` below when accessing `purchaseIntent.offer`. Do we need to do the same here?

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1296,16 +1321,18 @@ extension PurchasesOrchestrator: StoreKit2PurchaseIntentListenerDelegate {
         // stop the compiler from checking availability in the functions.
         // We also need to ensure that we're on Xcode >= 15.3, since that is when
         // PurchaseIntents were first made available on macOS.
-        #if !os(tvOS) && !os(watchOS) && compiler(>=5.10)
+        #if compiler(>=5.10) && !os(tvOS) && !os(watchOS) && !os(visionOS)
```
tonidero, line 1324: ~~Just checking but what's the reason we're not making this available in visionOS? If it's only for the tracking, we should probably just wrap that and avoid removing functionality?~~ Edit: Just saw the comment in the previous PR. I think it's ok then, just a bit nervous to change this in the tracking PR 😅.

tonidero (commented):
Would be great to get @fire-at-will confirm the changes in visibility, since he worked on winback offers and this is a bit related. But with the explanation from the previous PR, this LGTM!

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
ajpallares: Yeah... I don't like it much either but [there's an explanation](https://github.com/RevenueCat/purchases-ios/pull/4888#discussion_r1998802411) 😄

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
fire-at-will: I think probably so - [PurchaseIntent.offer](https://developer.apple.com/documentation/storekit/purchaseintent/offer) was introduced in iOS 18.0, which was released with Xcode 16 / Swift compiler 6.0.

@ajpallares I'd definitely recommend running all of the tests on this PR with `@RCGitBot please test` to make sure it still compiles on older Xcode versions. I had a pretty tough time when writing this code for the first time because of all of the device & OS compatibility requirements.

fire-at-will: @ajpallares In case you haven't seen it before, https://xcodereleases.com/ is the BEST resource out there for looking up what Swift versions shipped with what xcode & iOS versions. I don't know how I'd do my job without that website 😅

ajpallares: OMG THANK YOU! I didn't know it but it goes directly into my favorites! 🤩

ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

tonidero (approved):
:shipit:

PR #4896, opened 2025-03-17
[Diagnostics] Another approach to `DiagnosticsTracker` (open discussion)

ajpallares:
This is a POC PR to open a discussion about whether or not to apply the changes proposed in the PR

### Motivation
Originated from [this comment](https://github.com/RevenueCat/purchases-ios/pull/4893#discussion_r1998770899) in a different PR. The idea is to allow `MockDiagnosticsTracker` to have a unique events array, preventing the need to add a new array variable for each new track method in `DiagnosticsTrackerType`.

### Description
This PR basically replaces the track methods in the `DiagnosticsTrackerType` with cases in a new `DiagnosticsTracker.Event` enum.

#### Advantages & disadvantages

In terms of the SDK, both approaches are very similar. The new one uses the `DiagnosticsTracker.Event` enum whereas the old one used the different methods in the `DiagnosticsTrackerType` protocol. The main differences appear in unit tests:

* 👍  The `MockDiagnosticsTracker` does not need to be edited upon adding new events.
* 👍  The `MockDiagnosticsTracker` has a unified `trackedEvents` array for all the tracked events, which allows testing the order of some events (if that were important)
* 👎  Checking event parameters is more cumbersome now and the failures are less clear:
Before:
```swift
let params = try XCTUnwrap(self.mockDiagnosticsTracker.trackedProductsRequestParams.value.first)
expect(params.wasSuccessful) == true
expect(params.storeKitVersion) == .storeKit1
expect(Set(params.requestedProductIds)) == [identifier, notFoundIdentifier]
expect(Set(params.notFoundProductIds)) == [notFoundIdentifier]
expect(params.errorMessage).to(beNil())
expect(params.errorCode).to(beNil())
```

The failures happened in each line where there was a difference, which was much more clear to spot.

Now:
```swift
let event = try XCTUnwrap(self.mockDiagnosticsTracker.trackedEvents.value.first)
guard case .productsRequest(wasSuccessful: true,
                            storeKitVersion: .storeKit1,
                            errorMessage: nil,
                            errorCode: nil,
                            storeKitErrorDescription: _,
                            requestedProductIds: [identifier, notFoundIdentifier],
                            notFoundProductIds: [notFoundIdentifier],
                            responseTime: _) = event else {
    XCTFail("Event does not match")
}
```
The failure happens at `XCTFail`, which means the enum cases are not the same but it's not clear what failed. Different enum case? difference in an associated value?


ajpallares:
In my opinion, even though I love the `enum` approach for the implementation, the limitations when checking the expected event in unit tests make me feel like we should not implement this approach.

For me, the only missing piece would be to have a powerful way of comparing cases of enums with associated values in unit tests. Something like this:
```swift
r.trackedEvents.value.first)
expect(event).to(matchPattern(.productsRequest(wasSuccessful: true,
                                               storeKitVersion: .storeKit1,
                                               errorMessage: nil,
                                               errorCode: nil,
                                               storeKitErrorDescription: any(),
                                               requestedProductIds: [identifier, notFoundIdentifier],
                                               notFoundProductIds: [notFoundIdentifier],
                                               responseTime: any()))
```
In such a way that the failure clearly states when the case does not match, e.g.
> Expected `.productsRequest` case and got `purchaseRequest` case

and when one or more associated values do not match, e.g.

> Unexpected values in `.productsRequest` case:
> * Got `storeKitVersion: .storeKit1`, expected `storeKitVersion: .storeKit2`
> * Got `errorCode: nil`, expected `errorCode: 100`

tonidero:
It does look nice with the enum approach... But I do agree with your last message. It would be annoying to lose the ability to easily check what's different between each event during tests... 

I don't know if there is a way to easily compare 2 enum values like a `matchPattern` as you suggested... I'm guessing we would need some sort of equals method that first checks the type and then each value within the enum, and fail appropriately in each case right? But I think that might end up being a huge switch-case which might not be worth it... Might be worth asking for some more iOS devs for suggestions. Wdyt @MarkVillacampa @fire-at-will @joshdholtz?

JayShortway:
Ah, my comment wasn't meant to suggest changes to the `DiagnosticsTracker`, just the `MockDiagnosticsTracker`. Although I don't hate the enum. 🤔 

Couple of thoughts:
- Is there really no way to check equality on 2 enums? What about `XCTAssertEqual` or [`expectNoDifference`](https://github.com/pointfreeco/swift-custom-dump?tab=readme-ov-file#expectnodifference)?
- What if we keep everything untyped in a map in `MockDiagnosticsTracker`, or even `dump()`-ed to a string? We could have some utility assertion methods in `MockDiagnosticsTracker` to help ourselves, e.g. `assertEventTracked()`, `assertEventsTrackedInOrder()`. This can happen with or without the enum change in `DiagnosticsTracker`. 
- Philosophically it feels kinda wrong to have design decisions dictated by the assertion tool, but if that's what we're dealing with then so be it. If we can make this a quick-win improvement then let's do it, but otherwise I would not spend too much time on it.

ajpallares:
@tonidero 
> I don't know if there is a way to easily compare 2 enum values like a `matchPattern` as you suggested... I'm guessing we would need some sort of equals method that first checks the type and then each value within the enum, and fail appropriately in each case right? But I think that might end up being a huge switch-case which might not be worth it...

Exactly. We could probably implement that ourselves but, as you say, it wouldn't be worth it because it would be huge, growing with each new diagnostics event, and hard to maintain. Ideally there would exist some out-of-the-box solution for this. I believe it could be done using macros but we certainly don't want to get into that wormhole 😄 

@JayShortway 
> Ah, my comment wasn't meant to suggest changes to the `DiagnosticsTracker`, just the `MockDiagnosticsTracker`.

I know 😄, but I wanted to experiment with the enum approach. It allows to remove some boilerplate code in both the SDK and specially in the tests and I honestly like it better. But for me it's a no-go due to the tests thing.
Perhaps the right solution for this would be to move the enum approach (or even better, something like a Kotlin sealed class) to the `MockDiagnosticsTracker` to allow a single heterogeneous array.

> Is there really no way to check equality on 2 enums? What about XCTAssertEqual or [expectNoDifference](https://github.com/pointfreeco/swift-custom-dump?tab=readme-ov-file#expectnodifference)?

Yes, if the enum conforms to `Equatable` (as the one proposed in this PR does) then a simple `==` does it. **But** it has two deal-breaking limitations IMO: 
1. The failure message wouldn't clearly allow to discern what failed in the equal check
2. It wouldn't allow us to do partial checks on the associated values (e.g. `I expect `productId == "lifetime"` but I only want to check that `responseTime > 0`. With the equality in the test I would need to compare against an exact `responseTime`). This one is even more deal-breaking for me.

Btw, the link you shared [expectNoDifference](https://github.com/pointfreeco/swift-custom-dump?tab=readme-ov-file#expectnodifference) probably solves 1 but I don't think it solves 2. Good to know it exists though!

> What if we keep everything untyped in a map in MockDiagnosticsTracker, or even dump()-ed to a string? We could have some utility assertion methods in MockDiagnosticsTracker to help ourselves, e.g. assertEventTracked(), assertEventsTrackedInOrder(). This can happen with or without the enum change in DiagnosticsTracker.
Philosophically it feels kinda wrong to have design decisions dictated by the assertion tool, but if that's what we're dealing with then so be it. If we can make this a quick-win improvement then let's do it, but otherwise I would not spend too much time on it.

Yes, I 100% agree with you. I think your proposed approach is the one we should take. And I also agree that tests shouldn't lead design decisions although the truth is that it sometimes does (e.g. some `init`s that exist only for tests, or the existance of `DateProvider` itself). I wouldn't say "never have tests dictate the design decisions", but I do think that the gains in tests should be BIG for them to dictate some design decision, which is not the case in this PR.

Thank you both for your comments and thoughts!

JayShortway:
Just to clarify: testability should 100% influence design! Good design is easily testable. The _assertion tool_ limiting design is what feels wrong to me. However it doesn't seem to have an easy fix, so those are the constraints we have to work with. 

Thanks for opening this discussion, really useful! 😄

ajpallares:
Closing this one as we decided to take the alternative approach which was already deployed

PR #4897, opened 2025-03-18
Fix unit tests: make `presentCodeRedemptionSheet()` test only run on iOS

ajpallares:
### Motivation
Unit tests were failing in targets that are not iOS due to the unit test for `presentCodeRedemptionSheet()` added in #4893. 

### Description
This PR fixes the issue by making the unit test for `presentCodeRedemptionSheet()` only run on iOS

### DIFF
--- a/Tests/UnitTests/Purchasing/Purchases/PurchasesDiagnosticsTrackingTests.swift
+++ b/Tests/UnitTests/Purchasing/Purchases/PurchasesDiagnosticsTrackingTests.swift
@@ -24,7 +24,7 @@ class PurchasesDiagnosticsTrackingTests: BasePurchasesTests {
         self.setupPurchases()
     }
 
-    #if os(iOS)
+    #if os(iOS) || os(visionOS)
     @available(iOS 15.0, *)
     @available(watchOS, unavailable)
     @available(tvOS, unavailable)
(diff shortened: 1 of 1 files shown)


ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

Tests/UnitTests/Purchasing/Purchases/PurchasesDiagnosticsTrackingTests.swift
```diff
@@ -24,14 +24,18 @@ class PurchasesDiagnosticsTrackingTests: BasePurchasesTests {
         self.setupPurchases()
     }
 
+    #if os(iOS)
```
tonidero, line 27: Nitpick but should this be `#if os(iOS) || VISION_OS`? Seems to be the visibility for the original method I believe.

tonidero (approved):
:shipit:

Tests/UnitTests/Purchasing/Purchases/PurchasesDiagnosticsTrackingTests.swift
ajpallares: You're right. Thanks for pointing out!

ajpallares: Done in https://github.com/RevenueCat/purchases-ios/pull/4897/commits/a00530f04ac2dfbc7b92e8745328af44dde716e0

BTW, used `os(visionOS)` instead of `VISION_OS`, as we already use it like this in some places and it should be the way to go from now on, as it's the official way. Eventually I think we should remove `VISION_OS`

PR #4904, opened 2025-03-19
[Diagnostics] add `apple_transaction_update_received` event

ajpallares:
### Checklist

### Description
This PR implements the `apple_transaction_update_received` diagnostics events


Sources/Purchasing/StoreKit2/StoreKit2TransactionListener.swift
```diff
@@ -193,10 +200,51 @@ extension StoreKit2TransactionListener {
...
         }
     }
 }
+
+// MARK: - Diagnostics
+
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+private extension StoreKit2TransactionListener {
+
+    func trackTransactionUpdateReceivedIfNeeded(transaction: StoreTransaction, sk2Transaction: SK2Transaction) {
```
ajpallares, line 218: I know that having the two parameters `transaction` and `sk2Transaction` is not the cleanest solution as it's redundant. But I still think that it's the best compromise here because:
* We use `StoreTransaction` already-existing convenience getters to avoid some availability checks that we would need to get the properties from `sk2Transaction`
* We still need the `SK2Transaction` to get, at a minimum, the transaction ID as an integer (since `StoreTransaction` holds it a `String` and I'd like it less to have it converted back to `Int` from the `String`)
* I didn't want to add the properties that are exclusive for this event to `StoreTransaction` because they are only needed here.

tonidero (approved):
Nice! Looks great!

PR #4905, opened 2025-03-19
Fixing fit mode on background property

joshdholtz:
### Motivation

Fit mode was not being recognized on background image property

### Description

- Pass fit mode into the background view modifier and apply it
- Add optional alignment property for root paywall background to align background to top

| Fit | Fill |
|--------|--------|
| Screenshot 2025-03-19 at 3 37 41 PM | Screenshot 2025-03-19 at 3 38 08 PM |

### DIFF
--- a/RevenueCatUI/Templates/V2/PaywallsV2View.swift
+++ b/RevenueCatUI/Templates/V2/PaywallsV2View.swift
@@ -272,7 +272,8 @@ private struct LoadedPaywallsV2View: View {
             .frame(maxHeight: .infinity, alignment: .topLeading)
             .backgroundStyle(
                 self.paywallState.componentsConfig.background
-                    .asDisplayable(uiConfigProvider: uiConfigProvider).backgroundStyle
+                    .asDisplayable(uiConfigProvider: uiConfigProvider).backgroundStyle,
+                alignment: .top
             )
             .edgesIgnoringSafeArea(.bottom)
         }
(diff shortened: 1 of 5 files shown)


RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -272,7 +272,8 @@ private struct LoadedPaywallsV2View: View {
             .frame(maxHeight: .infinity, alignment: .topLeading)
             .backgroundStyle(
                 self.paywallState.componentsConfig.background
-                    .asDisplayable(uiConfigProvider: uiConfigProvider).backgroundStyle
+                    .asDisplayable(uiConfigProvider: uiConfigProvider).backgroundStyle,
+                alignment: .top
```
JayShortway, line 276: Is `.top` an arbitrary choice? Just checking.

JayShortway (approved):
Makes sense to me! Just a question.

RevenueCatUI/Templates/V2/PaywallsV2View.swift
joshdholtz: Kind of? 🤷‍♂️ It matched the behavior of background image on the root paywall on the dashboard which... I _think_ makes sense? 

Now I'm questioning it haha

JayShortway: Haha I think it's fine! The only other option I could think of would be `.center`.

joshdholtz: `.center` is what it _was_ and it was looking weird 😅 Probably _should_ be a configurable property 🤔

PR #4906, opened 2025-03-19
fix: Hide subscription detail if consumable

facumenzella:
### Motivation
When displaying a consumable, the renewal info does not make sense.

### Description
We'll hide it for now. The UI will change sooner than later.


ajpallares (approved):
It makes sense to me 👍

PR #4908, opened 2025-03-20
fix build issue in iOS 15

ajpallares:
### Motivation
I wanted to try and cover errors with `VerificationResult` similar to the conflict solved in #4814. But it looks like not only it's not needed but also it breaks compilation in iOS 15.

### Description
This PR removes the `RevenueCat` from `RevenueCat.VerificationInfo`


ajpallares:
@RCGitBot please test

PR #4909, opened 2025-03-20
Fix build issues and tests in older versions of Xcode

ajpallares:
### Motivation
The SDK is not building in Xcode 14 and some tests are failing in Xcode 15

### Description
This PR fixes these issues

### DIFF
--- a/Tests/StoreKitUnitTests/PurchasesOrchestratorCommonTests.swift
+++ b/Tests/StoreKitUnitTests/PurchasesOrchestratorCommonTests.swift
@@ -379,6 +379,8 @@ class PurchasesOrchestratorTrackingTests: BasePurchasesOrchestratorTests {
     func testTracksPurchaseSK2ProductSuccessWhenEnabledDiagnostics() async throws {
         try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
 
+        self.operationDispatcher.forwardToOriginalDispatchOnWorkerThread = true
+
         self.backend.stubbedPostReceiptResult = .success(mockCustomerInfo)
 
         let product = try await self.fetchSk2Product()
@@ -449,6 +451,8 @@ class PurchasesOrchestratorTrackingTests: BasePurchasesOrchestratorTests {
     func testDoesNotTrackPurchaseSK2ProductSuccessWhenDisabledDiagnostics() async throws {
         try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
 
+        self.operationDispatcher.forwardToOriginalDispatchOnWorkerThread = true
+        
         self.backend.stubbedPostReceiptResult = .success(mockCustomerInfo)
 
         let product = try await self.fetchSk2Product()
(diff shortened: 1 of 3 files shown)


ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

Tests/StoreKitUnitTests/PurchasesOrchestratorCommonTests.swift
```diff
@@ -379,12 +379,14 @@ class PurchasesOrchestratorTrackingTests: BasePurchasesOrchestratorTests {
     func testTracksPurchaseSK2ProductSuccessWhenEnabledDiagnostics() async throws {
         try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
 
+        self.operationDispatcher.forwardToOriginalDispatchOnWorkerThread = true
```
ajpallares, line 382: This unit test was consistently failing  in the CI for iOS 15.2 with Xcode 15.2. I couldn't reproduce it locally with Xcode 15.2 though. The test failure message was:
> Dispatch on worker thread timed out

This means that the failure was happening in this line of `MockOperationDispatcher.swift`
https://github.com/RevenueCat/purchases-ios/blob/2cbdf957baf9608fc7066fc2402b51d5f999d53b/Tests/UnitTests/Mocks/MockOperationDispatcher.swift#L111

I followed the stack trace in the unit test and the caller line is this one in `PurchasesOrchestrator.swift`
https://github.com/RevenueCat/purchases-ios/blob/2cbdf957baf9608fc7066fc2402b51d5f999d53b/Sources/Purchasing/Purchases/PurchasesOrchestrator.swift#L788

Note that in the test, the `manager` in this line is a `MockPaywallEventsManager` instance and the implementation couldn't be simpler:
https://github.com/RevenueCat/purchases-ios/blob/2cbdf957baf9608fc7066fc2402b51d5f999d53b/Tests/UnitTests/Mocks/MockPaywallEventsManager.swift#L29-L33
Perhaps, the fact that it's marked with `async` but not actually doing anything asynchronous is the problem here? I honestly don't know, but the fact is that this call seems to be producing a deadlock that causes the timeout in the `MockOperationDispatcher`.

So the solution has been to set `MockOperationDispatcher`'s `forwardToOriginalDispatchOnWorkerThread` to `true`, to avoid this locking behavior that we don't need for this test.

Tests/StoreKitUnitTests/PurchasesOrchestratorCommonTests.swift
```diff
@@ -449,12 +451,14 @@ class PurchasesOrchestratorTrackingTests: BasePurchasesOrchestratorTests {
     func testDoesNotTrackPurchaseSK2ProductSuccessWhenDisabledDiagnostics() async throws {
         try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
 
+        self.operationDispatcher.forwardToOriginalDispatchOnWorkerThread = true
```
ajpallares, line 454: Same explanation as above

Tests/StoreKitUnitTests/PurchasesOrchestratorCommonTests.swift
ajpallares: Two more clarifications:
* Note that this is not needed in unit tests that mock a failure since the execution doesn't reach this code
* Only the unit tests for SK2 products were failing. The same tests above for SK1 products were passing normally. The test implementations being the same (other than the SK2 vs SK1 difference) and the execution for SK1 products also go through the same lines in `MockOperationDispatcher` 🤷

tonidero (approved):

PR #4916, opened 2025-03-24
chore: Test action wrapper for customer center

facumenzella:
### Motivation
We've recently added preference keys to customer center without coverage.

### Description
This PR adds a basic check to verify that the preference keys + view modifier are working as expected.
Since https://github.com/RevenueCat/purchases-ios/pull/4882 needs SwiftTesting and Swift 6, we'll keep it in the XCTest world for now.


vegaro (approved):
thank you for taking care of this!

PR #4919, opened 2025-03-26
[Diagnostics]: add `NOT_CHECKED` value to `cache_status` parameter of `get_offerings_result` event

ajpallares:
This PR adds the `NOT_CHECKED` cache status value to the `get_offerings_result` diagnostics event when the offerings are fetch from the network regardless of the cache status


tonidero (approved):
Makes sense, thanks again!

PR #4924, opened 2025-03-27
[Diagnostics] Fix cache fetch policy key not matching specs

tonidero:
### Description
We were naming the key for the cache fetch policy as `cache_fetch_policy`, but we expected `fetch_policy`. This fixes the property name moving forward


ajpallares (approved):
Oh good catch! Totally missed this 😓
Thank you!

tonidero:
No problem! Easy to miss :)

PR #4929, opened 2025-03-28
[Diagnostics] Sync diagnostics if file reaches lower size limit

tonidero:
### Description
With all the new events we've been adding, we can potentially reach the file size limit much more easily, specially considering we only sync once per app open. This changes the logic, so we also sync if, after tracking an event, the file reaches a lower file size than the maximum. This should help us receiving events more frequently and avoid losing those events.

### DIFF
--- a/Sources/Diagnostics/DiagnosticsFileHandler.swift
+++ b/Sources/Diagnostics/DiagnosticsFileHandler.swift
@@ -15,6 +15,8 @@ import Foundation
 
 protocol DiagnosticsFileHandlerType: Sendable {
 
+    func updateDelegate(_ delegate: DiagnosticsFileHandlerDelegate?) async
+
     @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
     func getEntries() async -> [DiagnosticsEvent?]
 
@@ -35,8 +37,14 @@ protocol DiagnosticsFileHandlerType: Sendable {
 
 }
 
+protocol DiagnosticsFileHandlerDelegate: AnyObject, Sendable {
+    func onEventTracked() async
+}
+
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 actor DiagnosticsFileHandler: DiagnosticsFileHandlerType {
+
+    private weak var delegate: DiagnosticsFileHandlerDelegate?
 
     private let fileHandler: FileHandler
 
@@ -53,6 +61,10 @@ actor DiagnosticsFileHandler: DiagnosticsFileHandlerType {
         self.fileHandler = fileHandler
     }
 
+    func updateDelegate(_ delegate: DiagnosticsFileHandlerDelegate?) async {
+        self.delegate = delegate
+    }
+
     func appendEvent(diagnosticsEvent: DiagnosticsEvent) async {
         guard let jsonString = try? diagnosticsEvent.encodedJSON else {
             Logger.error("Failed to serialize diagnostics event to JSON")
@@ -60,6 +72,7 @@ actor DiagnosticsFileHandler: DiagnosticsFileHandlerType {
         }
 
         await self.fileHandler.append(line: jsonString)
+        await self.delegate?.onEventTracked()
     }
 
     func getEntries() async -> [DiagnosticsEvent?] {
--- a/Sources/Diagnostics/DiagnosticsFileHandler.swift
+++ b/Sources/Diagnostics/DiagnosticsFileHandler.swift
@@ -30,6 +30,9 @@ protocol DiagnosticsFileHandlerType: Sendable {
     @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
     func isDiagnosticsFileTooBig() async -> Bool
 
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+    func isDiagnosticsFileBigEnoughToSync() async -> Bool
+
 }
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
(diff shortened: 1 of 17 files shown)


Sources/Diagnostics/DiagnosticsTracker.swift
```diff
@@ -168,13 +174,15 @@ protocol DiagnosticsTrackerType: Sendable {
 }
 
 @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
-final class DiagnosticsTracker: DiagnosticsTrackerType, Sendable {
+final class DiagnosticsTracker: DiagnosticsTrackerType, @unchecked Sendable {
```
tonidero, line 177: TBH, I'm not sure this is the best way to do this... Would love some feedback on this @RevenueCat/coresdk.

ajpallares: So this `@unchecked` thing is basically the same as silencing the warning. But in reality, the potential race condition issues that `Sendable` helps avoid remains.

I went into the rabbit hole of trying to fix this and I could think of only two solutions:
1. Either apply some global actor (as `@MainActor`, but a custom one) to the `delegate` variable
2. Or make `DiagnosticsTracker` and `actor` (which is sendable) and adapting to the requirements. 

I currently don't know which approach would be cleaner. But I experimented with number 2 in this commit https://github.com/RevenueCat/purchases-ios/commit/cd26763b614cdcf8804c2665a73d597e934c2ff8

It's not a very clean solution. 2 wrapper methods I needed so just set the delegate! one for protocol conformance, and the second to actually set the `delegate`.

tonidero: Hmm ok will take a look on Monday but might be worth unifying the tracker and synchronizer classes... will study it on Monday. Thank you!!

ajpallares: Sure! Let me know if you need help!

tonidero: In the end, joining the synchronizer and tracker together felt wrong, and I don't fully like making the tracker an actor since it would force us to mark all tracking methods as nonisolated or making them async, which could cause issues in the call sites... I played with a different approach of making a delegate for the file handler, which is a common dependency for both the synchronizer and tracker. Wdyt? https://github.com/RevenueCat/purchases-ios/pull/4929/commits/22daa8ddd617264eb112687f48933f41f11f379d

ajpallares: That's a nice workaround! It still seems to me more appropriate to have it as a tracker delegate. In fact, the delegate method is called `onEventTracked()`. **However**, I totally understand that it's more convenient to have the delegate be in the file handler since:
1. It's already an `actor`.
2. All the `track` calls go through it anyway.

Honestly, I think this is fine for now, as a way to minimize changes 👍

tonidero: I tried to clean it up a bit by moving the size calculation into the file handler and changing the delegate name... I do agree it would feel better in the tracker, but this way it doesn't feel so out of place IMO.

Sources/Diagnostics/DiagnosticsFileHandler.swift
```diff
@@ -50,13 +58,21 @@ actor DiagnosticsFileHandler: DiagnosticsFileHandlerType {
...
         guard let jsonString = try? diagnosticsEvent.encodedJSON else {
             Logger.error("Failed to serialize diagnostics event to JSON")
             return
         }
 
         await self.fileHandler.append(line: jsonString)
+
+        if await self.isDiagnosticsFileBigEnoughToSync() {
+            await self.delegate?.onFileSizeIncreasedBeyondAutomaticSyncLimit()
+        }
```
ajpallares, line 75: This is a great idea! Even though the logic is basically the same, it is much more coherent now! 

Thinking out loud, and definitely not for this PR, but now it looks like the `isDiagnosticsFileTooBig()` logic could also be here. But it could easily get messy, since it's the tracker the one that should perhaps receive the event. Or perhaps, that logic could be all inside the file handler without any delegate call needed. Again, not a big deal and not for this PR, so let's keep it where it is 😄

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -485,17 +485,25 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
...
-                var diagnosticsSynchronizer: DiagnosticsSynchronizer?
+                let diagnosticsSynchronizer: DiagnosticsSynchronizer?
                 if diagnosticsEnabled {
                     if let diagnosticsFileHandler = diagnosticsFileHandler {
                         let synchronizedUserDefaults = SynchronizedUserDefaults(userDefaults: userDefaults)
                         diagnosticsSynchronizer = DiagnosticsSynchronizer(internalAPI: backend.internalAPI,
                                                                           handler: diagnosticsFileHandler,
                                                                           tracker: diagnosticsTracker,
                                                                           userDefaults: synchronizedUserDefaults)
+                        if let diagnosticsFileHandler = diagnosticsFileHandler as? DiagnosticsFileHandler {
```
ajpallares, line 496: I think that the casting is not necessary because the `updateDelegate` method is part of the `DiagnosticsFileHandlerType` protocol.
```suggestion
                        if let diagnosticsFileHandler = diagnosticsFileHandler {
```

ajpallares (approved):
Great improvement! Thanks for the iterations Toni. Looks great!

Sources/Diagnostics/DiagnosticsFileHandler.swift
```diff
@@ -127,6 +144,15 @@ private extension DiagnosticsFileHandler {
 
     }
 
+    private func isDiagnosticsFileBigEnoughToSync() async -> Bool {
+        do {
+            return try await self.fileHandler.fileSizeInKB() > Self.minFileSizeEnoughToSyncInKb
+        } catch {
+            Logger.error(Strings.diagnostics.failed_check_diagnostics_size(error: error))
+            return true
```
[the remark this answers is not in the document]
tonidero: Well, I thought that if there is a problem to check the file size, it might make sense to try to sync the file, and clear it to fix any issues happening... Lmk if you think otherwise!

Sources/Purchasing/Purchases/Purchases.swift
tonidero: Right! This is from a previous iteration, thanks! I belive I can remove the entire `if let`

Sources/Diagnostics/DiagnosticsFileHandler.swift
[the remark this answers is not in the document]
ajpallares: I didn't really have an opinion, but I was interested to know your reasoning to return `true`, and it makes total sense. So thanks for explaining!

Sources/Purchasing/Purchases/Purchases.swift
ajpallares: Yeah, much better even. Thanks!

PR #4930, opened 2025-03-28
Replace initial alert dialog with a progress view when restoring in Customer Center

vegaro:
Replaces the initial dialog in the restoring dialog with a loading dialog


RevenueCatUI/CustomerCenter/ViewModels/RestorePurchasesAlertViewModel.swift
```diff
@@ -0,0 +1,60 @@
...
+    func performRestore() async -> AlertType {
+        self.actionWrapper.handleAction(.restoreStarted)
+
+        do {
+            let customerInfo = try await purchasesProvider.restorePurchases()
+            self.actionWrapper.handleAction(.restoreCompleted(customerInfo))
+
+            let hasPurchases = !customerInfo.activeSubscriptions.isEmpty || !customerInfo.nonSubscriptions.isEmpty
+
+            self.state = .notLoaded
```
facumenzella, line 52: this is not needed either

RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
```diff
@@ -32,23 +32,50 @@ struct RestorePurchasesAlert: ViewModifier {
     var openURL
 
     @EnvironmentObject private var customerCenterViewModel: CustomerCenterViewModel
+    @StateObject private var viewModel: RestorePurchasesAlertViewModel
 
     @State
-    private var alertType: AlertType
+    private var alertType: RestorePurchasesAlertViewModel.AlertType
```
facumenzella, line 38: These `@State` and `@StateObject` don't make sense here. The variables are ok, but the property wrapper does not.

`alertType` should be passed a as `@Binding` like `isPresented`.  The view using the modifier is the one driving the State, not the ViewModifier which must be stateless.

facumenzella (changes requested):

ajpallares (commented):
I have some doubts about the implementation. Let me know if I can help!

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsView.swift
```diff
@@ -37,9 +38,13 @@ struct NoSubscriptionsView: View {
 
     @State
     private var showRestoreAlert: Bool = false
+    @State
+    private var showRestoreAlertType: RestorePurchasesAlertViewModel.AlertType = .loading
```
ajpallares, line 42: I think this can be removed now

RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
```diff
@@ -25,20 +26,38 @@ import SwiftUI
...
     @Environment(\.supportInformation)
     private var supportInformation: CustomerCenterConfigData.Support?
 
+    init(
+        isPresented: Binding<Bool>,
+        actionWrapper: CustomerCenterActionWrapper
+    ) {
+        self.init(isPresented: isPresented,
+                  viewModel: RestorePurchasesAlertViewModel(
+                    actionWrapper: actionWrapper
```
ajpallares, line 49: nit: I'm not sure this indentation is right

ajpallares (approved):
I think this looks great! Only a couple of minor things

RevenueCatUI/CustomerCenter/ViewModels/RestorePurchasesAlertViewModel.swift
```diff
@@ -0,0 +1,55 @@
...
+
+import Foundation
+import RevenueCat
+import SwiftUI
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+@MainActor class RestorePurchasesAlertViewModel: ObservableObject {
```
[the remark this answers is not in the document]
ajpallares: I see your point, but in this case, I'm not really sure whether `RestorePurchasesAlert` should even be a `ViewModifier`. It's very convenient to use as `ViewModifier`, for sure. But in this case it really needs to contain state and business logic (something that a `ViewModifier` probably shouldn't do either). The state is needed to determine what alert and buttons to show at any given time, and it contains the logic to perform the `restorePurchases()`.
How do you think we could improve this?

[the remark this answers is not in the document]
vegaro: I really didn't want to do more refactors but it looks like we keep going back and forth so I am just going to make it a View, it shouldn't be that hard I think right?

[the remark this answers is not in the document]
vegaro: take a look at [e29ed6b](https://github.com/RevenueCat/purchases-ios/pull/4930/commits/e29ed6b97d380ad26e7537bc03c42bd12dafd6d5)

ajpallares (approved):
Yep, using the overlay is a good idea!

I think this is an OK compromise as `RestorePurchasesAlert` is not really a simple alert without any logic. It contains some business logic and it needs to hold some state for that logic. I'm OK with this approach 👍 
Thanks for the iterations @vegaro!

facumenzella (approved):
Nice refactor!! 🚀

RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
```diff
@@ -166,32 +153,27 @@ struct RestorePurchasesAlert: ViewModifier {
...
-            return localization[.goingToCheckPurchases]
         }
     }
 
     private func dismissAlert() {
-        self.alertType = .restorePurchases
         self.isPresented = false
+        Task {
+            await self.customerCenterViewModel.loadScreen()
+        }
```
vegaro, line 163: Had to add this @facumenzella @ajpallares since I broke it with the refactor

Not sure it's entirely correct, but the loading needs to go through `customerCenterViewModel` since that's what we are reloading. I think the alternative is to add a `onDismiss()` to the `RestorePurchasesAlert`, add references to `customerCenterViewModel` in `NoSubscriptionsView` and call `self.customerCenterViewModel.loadScreen()` on the callers (`ManageSubscriptionsView` and `NoSubscriptionsView`). 

Wdyt?

ajpallares, line 163: I'm not 100% sure about this, but I _think_ that it might happen that because the `RestorePurchasesAlert` view gets dismissed with `self.isPresented = false`, the `Task` created here, which inherits the context from the parent, could get canceled on dismissal. If we wanted to create an independent task that gets executed regardless, perhaps we should do a `Task.detached`, like this
```suggestion
        Task.detached {
            await self.customerCenterViewModel.loadScreen()
        }
        self.isPresented = false
```
With detached, we guarantee that it does not get canceled even if the parent context gets canceled due to the dismissal.

ajpallares: @vegaro anyway, I think that a method in `CustomerCenterViewModel` to notify it of `onDismissRestorePurchasesAlert()` is a cleaner approach, since the `RestorePurchasesAlert` already has a reference to the view model, I think it makes total sense.
```suggestion
        self.onDismissRestorePurchasesAlert()
        self.isPresented = false
```
In the view model's `onDismissRestorePurchasesAlert()` method, a `Task` would be initiated. This way, we have the `Task` inherit the context of the view model and it's cleaner on the alert's side.

vegaro: right, I like that! Take a look at [8cc6d9b](https://github.com/RevenueCat/purchases-ios/pull/4930/commits/8cc6d9b0f6cb70a792c1ba49bbfa22cdd44e9542)

ajpallares: Nice! I like this much better 👍

PR #4933, opened 2025-03-31
Update strings for restore purchases alerts

vegaro:
Requires backend changes for the defaults, but we can still merge this:

If restore successful show alert:
- Title: Purchases restored
- Body: We restored your past purchases and applied them to your account. 
- Button: Done

If restore does not yield anything:
- Title: No past purchases
- Body: We could not find any purchases with your account. If you think this is an error, please contact support
- Default button: Contact Support…
- Button: Dismiss


ajpallares (approved):
Looks good 👍

PR #4936, opened 2025-03-31
Add diagnostics event when AppTransaction fails to be fetched

MarkVillacampa:
### Checklist

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -313,9 +313,6 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
         let userDefaults = userDefaults ?? UserDefaults.computeDefault()
         let deviceCache = DeviceCache(sandboxEnvironmentDetector: systemInfo, userDefaults: userDefaults)
 
-        let purchasedProductsFetcher = OfflineCustomerInfoCreator.createPurchasedProductsFetcherIfAvailable()
-        let transactionFetcher = StoreKit2TransactionFetcher()
-
         let diagnosticsFileHandler: DiagnosticsFileHandlerType? = {
             guard diagnosticsEnabled,
                   dangerousSettings?.uiPreviewMode != true,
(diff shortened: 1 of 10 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -334,6 +331,11 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
             return nil
         }()
 
+        let purchasedProductsFetcher = OfflineCustomerInfoCreator.createPurchasedProductsFetcherIfAvailable(
+            diagnosticsTracker: diagnosticsTracker
+        )
+        let transactionFetcher = StoreKit2TransactionFetcher(diagnosticsTracker: diagnosticsTracker)
+
```
ajpallares, line 338: Not really from this PR, but I'm curious. Since we're already creating the `StoreKit2TransactionFetcher` here, why not pass it to `OfflineCustomerInfoCreator.createPurchasedProductsFetcherIfAvailable` instead of creating another instance? Or do we want them to be two different instances for some reason?

Tests/StoreKitUnitTests/StoreKit2/StoreKit2TransactionFetcherTests.swift
ajpallares, line 1: I'm guessing it's not easy to add unit tests for these events because the implementation relies on `AppTransaction.shared`?

ajpallares (approved):
Nice to have this event! Just a couple of comments, but it looks great 👍

tonidero (approved):
Nice! This makes sense to me! It would be great to have tests I guess... Wonder if we should abstract the interaction with the storekit APIs so it can be mocked... Not a blocker though.

Tests/StoreKitUnitTests/StoreKit2/StoreKit2TransactionFetcherTests.swift
MarkVillacampa: Right, there is no built in way to make the `AppTransaction.shared` result return an error

PR #4938, opened 2025-04-01
Add `X-Is-Backgrounded` header

tonidero:
### Description
Adds a header to determine whether the request was performed in the background or foreground. This can be useful for debugging in some situations.


Sources/Misc/SystemInfo.swift
```diff
@@ -175,6 +183,10 @@ class SystemInfo {
         self.dangerousSettings = dangerousSettings ?? DangerousSettings()
         self.clock = clock
         self.preferredLocalesProvider = preferredLocalesProvider
+
+        self.isApplicationBackgrounded { isAppBackgrounded in
```
tonidero, line 187: There may be a short span of time where the value does not reflect the correct value until this is executed... That's the main reason I didn't want to start changing usages of the existing `isAppBackgrounded` API everywhere. I think `isAppBackgroundedState` may be used for less critical code paths. There are some places we could potentially change it, but I wanted to minimize the changes in this PR, which gets big with the snapshot changes already.

Sources/Misc/SystemInfo.swift
```diff
@@ -47,13 +47,19 @@ class SystemInfo {
         set { self._finishTransactions.value = newValue }
     }
 
+    var isAppBackgroundedState: Bool {
```
tonidero, line 50: Since I'm keeping `isApplicationBackgrounded` for now, I needed to dissambiguate with this new property somehow. Lmk what you think of the naming!

tonidero (commented):
Feel free to ignore all changes in the snapshot test files which are just adding the new header and are the bulk of the changes in this PR 🙏

tonidero:
@RCGitBot please test

Sources/Misc/SystemInfo.swift
JayShortway: At first glance it's a bit confusing. Could you explain the difference and why we need to have both?

JayShortway (approved):
Makes sense to me! Just a question on the new `isAppBackgroundedState`.

Sources/Misc/SystemInfo.swift
ajpallares: Yeah, I agree with @JayShortway.

I think the existing `isApplicationBackgrounded` variable is only used in the method
```
func isApplicationBackgrounded(completion: @escaping @Sendable (Bool) -> Void) {
        completion(self.isApplicationBackgrounded)
    }
}
```
so perhaps we could make it private (to avoid using it accidentally from outside `SystemInfo`) and change its name to something internal (perhaps event a `@MainActor private func getIsApplicationBackgrounded()` to make it more obvious and have the new variable be called `isApplicationBackgrounded`, which is more clear.

I understand that the two ways to check this are different and we don't need to change the existing, completion-based one for now.

JayShortway: (Ah I just saw the [other comment](https://github.com/RevenueCat/purchases-ios/pull/4938#discussion_r2023016534), apologies. I missed it because I was looking on the Files tab and the attached line is outdated.)

ajpallares: (same here! 😅)

tonidero: No worries! I should have updated that 😅. And I think that's valid suggestion! Will try to update that 👍

Sources/Misc/SystemInfo.swift
```diff
@@ -160,13 +166,15 @@ class SystemInfo {
...
          dangerousSettings: DangerousSettings? = nil,
+         isAppBackgrounded: Bool? = nil,
          clock: ClockType = Clock.default,
          preferredLocalesProvider: PreferredLocalesProviderType = PreferredLocalesProvider.default) {
         self.platformFlavor = platformInfo?.flavor ?? "native"
         self.platformFlavorVersion = platformInfo?.version
         self._bundle = .init(bundle)
 
         self._finishTransactions = .init(finishTransactions)
+        self._isAppBackgroundedState = .init(isAppBackgrounded ?? false)
```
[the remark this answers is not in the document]
tonidero: Yeah, it actually shouldn't really matter because we're initializing the value with the value obtained from the callback method a bit below and also start receiving updates immediately from `PurchasesOrchestrator`. 

As for this initial value, I thought it made the most sense in iOS, since usually initializing the SDK means the app is foregrounded. I think the exception may be extensions, but considering that should be the less common case, I preferred to go with `false` here (again, it shouldn't ultimately matter I think)

ajpallares (approved):
It's much cleaner now, I think. Thank you!

PR #4947, opened 2025-04-02
UI preview mode: enable customer center previews

ajpallares:
### Description
This PR makes some APIs `@_spi(Internal) public` to enable the Customer Center previews functionality of the RC app.

This is the minimum required `@_spi` APIs I could come up with for Customer Center previews.


RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -99,6 +102,11 @@ import RevenueCat
         self.customerCenterStoreKitUtilities = customerCenterStoreKitUtilities
     }
 
+    convenience init(uiPreviewPurchaseProvider: CustomerCenterPurchasesType) {
+        self.init(actionWrapper: CustomerCenterActionWrapper(legacyActionHandler: nil),
+                  purchasesProvider: uiPreviewPurchaseProvider)
+    }
```
ajpallares, line 108: This `init` is used by the new `@_spi(Internal) public init` of `CustomerCenterView` that receives a parameter `uiPreviewPurchaseProvider: CustomerCenterPurchasesType`

Sources/Purchasing/StoreKitAbstractions/Test Data/TestStoreProduct.swift
```diff
@@ -99,22 +99,25 @@ public struct TestStoreProduct {
 
 }
 
+// swiftlint:disable missing_docs
+
 // Ensure consistency with the internal type
-extension TestStoreProduct: StoreProductType {
+@_spi(Internal) extension TestStoreProduct: StoreProductType {
```
ajpallares, line 105: We need this because `TestStoreProduct` is public and now `StoreProductType` is `@_spi(Internal) public`

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
ajpallares, line 108: This `init` is used by the new `@_spi(Internal) public` init of `CustomerCenterView` that receives the `uiPreviewPurchaseProvider: CustomerCenterPurchasesType` parameter

Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift
```diff
@@ -94,8 +94,10 @@ public typealias SK2Product = StoreKit.Product
     // switflint:enable missing_docs
 }
 
+@_spi(Internal) extension StoreProduct: StoreProductType { }
```
ajpallares, line 97: Because `StoreProductType` is now `@_spi(Internal) public` but `StoreProduct` is `public`, the more restrictive conformance of `StoreProduct` to it must be done separately (this is enforced by the compiler)

Tests/RevenueCatUITests/CustomerCenter/Mocks/MockCustomerCenterPurchases.swift
```diff
@@ -13,7 +13,7 @@
 
 import Foundation
 import RevenueCat
-@testable import RevenueCatUI
+@_spi(Internal) @testable import RevenueCatUI
```
ajpallares, line 16: These imports are now required in test files that used any `@_spi(Internal) public` API

Tests/UnitTests/Purchasing/StoreKitAbstractions/SubscriptionPeriodTests.swift
```diff
@@ -141,7 +141,7 @@ class SubscriptionPeriodTests: TestCase {
 
         let expectations: [(
             inputValue: Int, inputUnit: SKProduct.PeriodUnit,
-            expectedValue: Int, expectedUnit: SubscriptionPeriod.Unit
+            expectedValue: Int, expectedUnit: RevenueCat.SubscriptionPeriod.Unit
```
ajpallares, line 144: This fixes a test build error about conflicting type with `StoreKit`s `SubscriptionPeriod`

ajpallares:
@RCGitBot please test

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1769,7 +1769,8 @@ extension Purchases: InternalPurchasesType {
     }
 
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
-    func productEntitlementMapping() async throws -> ProductEntitlementMapping {
+    // swiftlint:disable:next missing_docs
+    @_spi(Internal) public func productEntitlementMapping() async throws -> ProductEntitlementMapping {
```
tonidero, line 1773: Hmm I think some people might go the extra mile and opt in, in order to use this (there have been some requests before).

Just wondering, why do we need to expose this? Can we mock it in some way if we need it for the app?

Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift
```diff
@@ -94,8 +94,10 @@ public typealias SK2Product = StoreKit.Product
...
+
 /// Type that provides access to all of `StoreKit`'s product type's properties.
-internal protocol StoreProductType: Sendable {
+@_spi(Internal) public protocol StoreProductType: Sendable {
```
tonidero, line 100: Same question about why this now needs to be exposed, and whether we can get around just using TestStoreProduct. (Probably not but curious of the reason)

tonidero (commented):
Code changes seem good. We are exposing a few things with here that would be great if we could avoid to... I'm sure there is a reason though (haven't checked the app's PR yet). Will revisit this after reviewing that.

ajpallares:
Thank you very much for your comments @tonidero! Yeah, you're totally right. I tried to be too clever by mocking real products + entitlements from the customer's app and that approach led me to need to open all these up. 

Now I'm trying a different approach in which I mock all the information (purchases + entitlements) in the app itself, which will probably result in many (if not most) of these going away. 

So I'll make this a draft again until that all is resolved

Sources/Purchasing/StoreKitAbstractions/StoreProductDiscount.swift
```diff
@@ -114,6 +114,21 @@ public final class StoreProductDiscount: NSObject, StoreProductDiscountType {
     }
 
 }
+extension StoreProductDiscount {
+    // swiftlint:disable:next missing_docs
+    @_spi(Internal) public func promotionalOffer(withSignedDataIdentifier identifier: String,
```
tonidero, line 119: Hmm I'm wondering if it would make sense to provide a public initializer for `StoreProductDiscount` so people can write more tests... Or maybe the more Swifty approach would be to create a protocol and use that instead of the class, so it can be implemented as needed... Still, I think this is the best approach for now 👍

RevenueCatUI/CustomerCenter/Abstractions/CustomerCenterPurchasesType.swift
```diff
@@ -14,11 +14,13 @@
...
 import RevenueCat
 
+// swiftlint:disable missing_docs
+
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 @available(macOS, unavailable)
 @available(tvOS, unavailable)
 @available(watchOS, unavailable)
-protocol CustomerCenterPurchasesType: Sendable {
+@_spi(Internal) public protocol CustomerCenterPurchasesType: Sendable {
```
tonidero, line 23: Hmm I'm thinking we could potentially avoid exposing this, and in the `CustomerCenterView` constructor, just expose the parameters we want to override/mock as parameters right? Not sure if we do need to mock anything from our app instead.

tonidero (approved):
Just a couple questions, but I think this is ok!

Sources/Purchasing/StoreKitAbstractions/StoreProductDiscount.swift
ajpallares: Yeah, that's a good point. However, we do already expose the [`TestStoreProductDiscount`](https://github.com/RevenueCat/purchases-ios/blob/ac6da565f07d6048ddde257c21342add7cd00e5c/Sources/Purchasing/StoreKitAbstractions/Test%20Data/TestStoreProductDiscount.swift), which is (I think) exactly that, and can be mapped into a `StoreProductDiscount` object.
However, we do not have the same for `PromotionalOffer`, which I needed to create from the app. And because the `init` is `internal`, this way minimizes the APIs that need to be made public for my used case (which is creating the promotional offer associated to an existing `StoreProductDiscount` object)

RevenueCatUI/CustomerCenter/Abstractions/CustomerCenterPurchasesType.swift
ajpallares: > Not sure if we do need to mock anything from our app instead.

This is exactly why I opened up `CustomerCenterPurchasesType`. It does contain not only parameters, but also methods. In fact, it contains all the `Purchases` interfaces that the Customer Center needs to function. So as you say, the app needs to be intercepting these interfaces (parameters but also methods) to be able to mock data for Customer Center previews.

Does it make sense?

tonidero: Ahh right. Yeah makes total sense. Thanks for explaining!

PR #4948, opened 2025-04-02
[Diagnostics] Add `storefront` property to more events

tonidero:
### Description
This adds the `storefront` property to a few more events where it could be useful. Specifically:
- `apple_products_request`
- `apple_purchase_attempt`
- `apple_transaction_queue_received`
- `apple_trial_or_intro_eligibility_request`


Sources/Purchasing/ProductsManager.swift
```diff
@@ -162,6 +162,7 @@ private extension ProductsManager {
                                                     errorMessage: errorMessage,
                                                     errorCode: errorCode,
                                                     storeKitErrorDescription: storeKitErrorDescription,
+                                                    storefront: self.systemInfo.storefront?.countryCode,
```
tonidero, line 165: This uses SK1 to calculate the storefront... I believe it should be ok? Lmk if there are any concerns though!

Tests/StoreKitUnitTests/ProductsManagerTests.swift
```diff
@@ -114,15 +114,20 @@ class ProductsManagerTests: StoreKitConfigTestCase {
...
     fileprivate func createManager(storeKitVersion: StoreKitVersion,
+                                   storefront: StorefrontType? = nil,
                                    diagnosticsTracker: DiagnosticsTrackerType? = nil) -> ProductsManager {
         let platformInfo = Purchases.PlatformInfo(flavor: "xyz", version: "123")
+        let systemInfo = MockSystemInfo(
+            platformInfo: platformInfo,
+            finishTransactions: true,
+            storeKitVersion: storeKitVersion
+        )
+        if let storefront {
```
ajpallares, line 125: I think this `if let` is not needed since `stubbedStorefront` is already optional

ajpallares (approved):
Looks great! Thanks for adding this so fast! 🚀 
Just a minor comment

Sources/Purchasing/ProductsManager.swift
ajpallares: Yeah, I think it's fine. We could use SK2's `storefront` but it's an `async` API, so it would require deeper changes. It can be done in a future PR I think

PR #4951, opened 2025-04-02
Cc 10 make more clear that simulators dont have mail apps and

facumenzella:
### Motivation
Contact support is not visible on the simulator, and leads to unnecessary back and forth.

### Description
This extract the `supportURL` logic, and adds a disabled button view instead of not showing the button

### DIFF
--- a/RevenueCatUI/CustomerCenter/Extensions/CustomerCenterConfigDataSupport+URL.swift
+++ b/RevenueCatUI/CustomerCenter/Extensions/CustomerCenterConfigDataSupport+URL.swift
@@ -1,0 +1,31 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  CustomerCenterConfigDataSupport+Button.swift
+//
+//  Created by Facundo Menzella on 2/4/25.
+
+import RevenueCat
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+extension CustomerCenterConfigData.Support {
+
+    func supportURL(localization: CustomerCenterConfigData.Localization) -> URL? {
+        let subject = localization[.defaultSubject]
+        let body = calculateBody(localization)
+
+        return URLUtilities.createMailURLIfPossible(
+            email: email,
+            subject: subject,
+            body: body)
+    }
+}
--- a/RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
@@ -102,16 +102,18 @@ struct WrongPlatformView: View {
                     }
                 }
             } else if RuntimeUtils.isSimulator {
-                AsyncButton {
-                    // do nothing
-                } label: {
-                    Text(localization[.contactSupport])
+                AsyncButton { } label: {
+                    VStack(alignment: .leading) {
+                        Text(localization[.contactSupport])
+                            .frame(alignment: .leading)
+
+                        Text("Not available on simulator")
+                            .frame(alignment: .leading)
+                            .font(.caption)
+                            .foregroundStyle(.secondary)
+                    }
                 }
                 .disabled(true)
-
-                Text("Contact support not available on simulator")
-                    .font(.caption)
-                    .foregroundStyle(.secondary)
             }
 
         }
--- a/RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
@@ -52,15 +52,6 @@ struct WrongPlatformView: View {
     @Environment(\.openURL)
     private var openURL
 
-    private var supportURL: URL? {
-        guard let supportInformation = self.supportInformation else { return nil }
-        let subject = self.localization[.defaultSubject]
-        let body = supportInformation.calculateBody(self.localization)
-        return URLUtilities.createMailURLIfPossible(email: supportInformation.email,
-                                                    subject: subject,
-                                                    body: body)
-    }
-
     init(screen: CustomerCenterConfigData.Screen? = nil,
          purchaseInformation: PurchaseInformation) {
         self.screen = screen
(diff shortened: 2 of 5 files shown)


ajpallares (approved):
Looks good! 👍

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
```diff
@@ -92,15 +83,30 @@ struct WrongPlatformView: View {
...
+            } else if RuntimeUtils.isSimulator {
+                AsyncButton { } label: {
+                    VStack(alignment: .leading) {
+                        Text(localization[.contactSupport])
+                            .frame(alignment: .leading)
+
+                        Text("Not available on simulator")
+                            .frame(alignment: .leading)
+                            .font(.caption)
+                            .foregroundStyle(.secondary)
```
facumenzella, line 104: I am not sure if it's worth localizing this? What do you think @RevenueCat/coresdk ? I don't really want this to be shown in the customer center dashboard

vegaro: Yeah not worth it, it probably won't ever be seen by final users

RevenueCatUI/CustomerCenter/Extensions/CustomerCenterConfigDataSupport+URL.swift
```diff
@@ -0,0 +1,36 @@
...
+//  Created by Facundo Menzella on 2/4/25.
+
+import Foundation
+import RevenueCat
+
+#if os(iOS)
+
+extension CustomerCenterConfigData.Support {
+
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
```
ajpallares, line 21: Not a big deal, but I still wonder why we have
```
``` 
and then 
```
    @available(tvOS, unavailable)
```

I guess we'll need to make this consistent eventually 😄

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
```diff
@@ -92,13 +88,16 @@ struct WrongPlatformView: View {
                     }
                 }
             }
-            if let url = supportURL {
```
vegaro, line 95: removing this check will always show the button even if there's no `supportURL` configured right?

vegaro (changes requested):

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
```diff
@@ -113,10 +112,20 @@ struct WrongPlatformView: View {
...
+                dismissButton: .default(Text("Ok")))
+        })
     }
 
 }
 
+private struct NoEmailAppError: LocalizedError {
+
+}
+
```
ajpallares, line 128: I believe this is not used

ajpallares (approved):
I honestly like this approach better, so thank you!

Just one thing to remove that I believe it's not being used (`NoEmailAppError`)

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
ajpallares: Yeah, good point

facumenzella: Yes, but if we want to show the alert there's no other alternative right?

vegaro: We only want to show the alert if there's a `supportURL`and someone taps the Contact Support button using a simulator. If there's no `supportURL` , which it's pretty common, the Contact Support button shouldn't be shown at all

RevenueCatUI/CustomerCenter/Extensions/CustomerCenterConfigDataSupport+URL.swift
facumenzella: Oh no wait, I took this from `URLUtilities.createMailURLIfPossible`. It was throwing errors otherwise

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
facumenzella: Ok so I've extracted `canOpenURL` out of the creation of the url, and check on tap if it can be opened at runtime 👍

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
```diff
@@ -92,10 +88,15 @@ struct WrongPlatformView: View {
...
-            if let url = supportURL {
+
+            if let url = supportInformation?.supportURL(localization: localization) {
                 Section {
                     AsyncButton {
-                        openURL(url)
+                        if URLUtilities.canOpenURL(url) {
+                            openURL(url)
+                        } else {
+                            self.showSimulatorAlert = true
```
ajpallares, line 98: Sorry to get perhaps too nitpicky, but it looks like `URLUtilities.canOpenURL(url)` can return `false` in cases other than the simulator (e.g. extensions or if `application.canOpenURL(url)` return `false`)

facumenzella: I've brought back RuntimeUtils, and added that check as well. Only open if `canOpenURL` is true, and its not a sim 😅

facumenzella: Ok after discussing the whole approach again with @ajpallares , I think we've got to a solution 😅

RevenueCatUI/CustomerCenter/Views/WrongPlatformView.swift
```diff
@@ -92,10 +88,16 @@ struct WrongPlatformView: View {
                     }
                 }
             }
-            if let url = supportURL {
+
+            if let url = supportInformation?.supportURL(localization: localization),
+               URLUtilities.canOpenURL(url) || RuntimeUtils.isSimulator {
```
vegaro, line 93: Does `URLUtilities.canOpenURL(url)` return false in the simulator?

Maybe this should just be

```
if let url = supportInformation?.supportURL(localization: localization),
               URLUtilities.canOpenURL(url) {
```

ajpallares: I think some URL schemes that would work on a real device might return `false` in the iOS Simulator (e.g. a `tel://` URL). So with this check we make sure that for any valid URL, even on simulator, we show the contact support button

vegaro: I see, so if it's not a simulator, it will show if it can open it. If it's a simulator, it will always show. Got it! Thank you!!

vegaro (approved):

vegaro (approved):

PR #4957, opened 2025-04-03
[Customer center] unify and propagate Purchases provider

ajpallares:
### Motivation
The Customer Center view's depend on `Purchases.shared`. This dependency is abstracted and passed as the protocols `CustomerCenterPurchasesType` and `ManageSubscriptionsPurchaseType` for testing purposes. However:
* `CustomerCenterPurchasesType` and `ManageSubscriptionsPurchaseType` had some similar methods.
* The dependency was not consistently being propagated down the view hierarchy.

### Description
This PR unifies `CustomerCenterPurchasesType` and `ManageSubscriptionsPurchaseType` into `CustomerCenterPurchasesType` and propagates the same injected `CustomerCenterPurchasesType` instance from `CustomerCenterView` down to its subviews.
In particular, after this PR, all the views in the Customer Center view hierarchy use the injected instance of `CustomerCenterPurchasesType` instead of `Purchases.shared` directly.


RevenueCatUI/CustomerCenter/Abstractions/CustomerCenterPurchasesType.swift
```diff
@@ -44,4 +50,15 @@ protocol CustomerCenterPurchasesType: Sendable {
...
+
+    #if os(iOS) || os(macOS) || os(visionOS)
+    @Sendable
+    func showManageSubscriptions() async throws
+    #endif
+
+    #if os(iOS) || os(visionOS)
+    @Sendable
+    func beginRefundRequest(forProduct productID: String) async throws -> RefundRequestStatus
+    #endif
```
ajpallares, line 63: These compiler conditions are exactly the same as those in the corresponding methods in `Purchases.swift`

RevenueCatUI/CustomerCenter/ViewModels/PromotionalOfferViewModel.swift
```diff
@@ -40,12 +40,12 @@ class PromotionalOfferViewModel: ObservableObject {
 
     init(
         promotionalOfferData: PromotionalOfferData?,
-        purchasesProvider: CustomerCenterPurchasesType = CustomerCenterPurchases(),
+        purchasesProvider: CustomerCenterPurchasesType,
```
facumenzella, line 43: I kind of dislike to give up this convenience, but I guess the dependency injection pattern is a broader discussion, out of scope for sure 😭

RevenueCatUI/CustomerCenter/Views/FeedbackSurveyView.swift
```diff
@@ -44,11 +44,13 @@ struct FeedbackSurveyView: View {
 
     init(
         feedbackSurveyData: FeedbackSurveyData,
+        purchasesProvider: CustomerCenterPurchasesType = CustomerCenterPurchases(),
```
facumenzella, line 47: Where do we draw the line now? I mean, I see that in some places the default init has been removed 🤔

RevenueCatUI/CustomerCenter/ViewModels/PromotionalOfferViewModel.swift
ajpallares: Well, we can totally keep the default parameters for convenience. But at the same time, we need to make sure that we're passing the dependency when we have it. 
It seems to me that we only want the convenience for previews, and perhaps that shouldn't be the leading factor in this case 🤔  WDYT?

RevenueCatUI/CustomerCenter/Views/FeedbackSurveyView.swift
ajpallares: Yes, you're right I've not been consistent on this

RevenueCatUI/CustomerCenter/ViewModels/PromotionalOfferViewModel.swift
JayShortway, line 43: Instead of passing the entire `CustomerCenterPurchasesType`, could we pass a closure instead? This class for instance only needs to call `purchase()` but it gets the full `CustomerCenterPurchasesType` with a lot of things it doesn't need. What if this class received a `onPurchasePromoClick()` function instead? 

This pattern avoids passing entire dependencies down to every leaf node of our UI tree. It also helps creating the `PromotionalOfferViewModel`, `PromotionalOfferView`, and other clases that follow the pattern in isolation, e.g. in tests or previews, as you don't need to implement all of `CustomerCenterPurchasesType` in that case.

This is more of a suggestion / discussion starter. Not a full review haha.

ajpallares: Yes, that is a very good point. I think what you say makes total sense and we should probably make that change.
However, I think what you mention is a task for a separate PR (and an approach that we should take from now on), since what this PR does is only make sure that the existing dependencies are injected instead of newly instantiated every time. WDYT? cc @facumenzella

JayShortway: Yea agreed! It's not a blocker for this PR for me.

facumenzella: I love this. I also agree that's a bigger refactor that we can tackle later

facumenzella (approved):

ajpallares:
@RCGitBot please test

PR #4960, opened 2025-04-03
Fix `CustomerCenterViewController` view

ajpallares:
`CustomerCenterViewController` was actually showing 2 `CustomerCenterView`s

This PR fixes it.


facumenzella (approved):
I think we should deprecate this tbh, but let's fix it anyways

PR #4969, opened 2025-04-08
Remove preprocessor script for PaywallsTester

ajpallares:
This PR removes the Preprocessor and Postprocessor scripts that were used by the PaywallsTester app to allow previewing some paywalls.

Since we can preview any Paywall in the app, it does not make sense to have this overhead anymore that was sometimes causing some build issues for the team.

### Notes
* This PR also fixes the compilation of PaywallsTester for watchOS
* As a consequence of removing the Preprocessor script, now we have to use `@testable import` in PaywallsTesters to access the internal APIs of the RevenueCatUI it required. The solution has been to use `@testable import` in `DEBUG` only, adding many `#if DEBUG` checks to remove from release the parts that rely on the internal APIs. As as result, we have differences between the app in Debug vs Release
image


Tests/TestingApps/PaywallsTester/PaywallsTester/Config/SamplePaywalls.swift
```diff
@@ -1,705 +0,0 @@
...
-
-
-
-import Foundation
-import RevenueCat
-import RevenueCatUI
-
-import UIKit
-
-final class SamplePaywallLoader {
```
JayShortway, line 16: Was this no longer needed?

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/CustomPaywallContent.swift
```diff
@@ -18,7 +18,6 @@ struct CustomPaywallContent: View {
 
     var body: some View {
         self.content
-            .scrollableIfNecessary(.vertical)
```
JayShortway, line 21: What's the reason for this change?

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
```diff
@@ -24,72 +26,22 @@ struct SamplePaywallsList: View {
 
     var body: some View {
         NavigationView {
-            self.list(with: Self.loader)
-                .navigationTitle("Example Paywalls")
+            self.list
+                .navigationTitle("Examples")
```
JayShortway, line 30: So does this mean we won't have any sample paywalls anymore? Sometimes it can be useful to try out a paywall built in code, to avoid having to go to the dashboard to create one, or if you're building something that's not available on the dashboard yet. Is it possible to keep that functionality?

JayShortway (approved):
Thanks for doing this! Just some questions, most importantly about sample paywalls.

Tests/TestingApps/PaywallsTester/PaywallsTester/Config/SamplePaywalls.swift
ajpallares: Assuming we're removing the local paywalls, it's not

facumenzella (approved):
Love this!!

Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/SamplePaywallsList.swift
ajpallares: After internal discussions, we decided to leave the existing functionality only in `DEBUG` (where we can do `@testable import` to access internal APIs)

JayShortway (approved):
This is perfect! Thanks for finding a solution to keep local sample paywalls alive 😄

PR #4970, opened 2025-04-08
Use fallback API hosts when receiving server down response

ajpallares:
### Checklist

### Description
This PR implements a retry mechanism using fallback URL hosts for supported requests that fail with a server down error

### Notes
* Only some requests support a fallback API host. Currently only `getCustomerInfo` and `getProductEntitlementMapping` support it.
* `HTTPRequestPath`s `serverHostURLs` is now an array containing the list of hosts, in order of preference, to use for supported requests.
* When a request's response is a server error (5XX), the same request is immediately retried with the next host of the array.
* The logic is implemented per-request: that is, if a request fails with a server-down error for host 1 and then falls back and succeeds for host 2, then all future requests will use again host 1 initially. No state is held between different requests.
* No fallback retries happen when using a proxy URL

Android PR counterpart --> https://github.com/RevenueCat/purchases-android/pull/2368


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -463,26 +477,28 @@ private extension HTTPClient {
...
                     Logger.debug(Strings.network.request_handled_by_load_shedder(request.httpRequest.path))
                 }
 
-                requestRetryScheduled = self.retryRequestIfNeeded(request: request, httpURLResponse: httpURLResponse)
+                retryScheduled = self.retryRequestWithNextHostIfNeeded(request: request,
+                                                                       httpURLResponse: httpURLResponse)
+                if !retryScheduled {
+                    retryScheduled = self.retryRequestIfNeeded(request: request,
+                                                               httpURLResponse: httpURLResponse)
+                }
```
ajpallares, line 494: These lines are the actual changes in this method. The other changes in this method are to accommodate the linter (`function_body_length`).

Sources/Diagnostics/Networking/DiagnosticsHTTPRequestPath.swift
```diff
@@ -16,7 +16,7 @@ import Foundation
 extension HTTPRequest.DiagnosticsPath: HTTPRequestPath {
 
     // swiftlint:disable:next force_unwrapping
-    static let serverHostURL = URL(string: "https://api-diagnostics.revenuecat.com")!
+    static let serverHostURLs = [URL(string: "https://api-diagnostics.revenuecat.com")!]
```
ajpallares, line 19: TODO: add final hosts list

Sources/Networking/HTTPClient/HTTPRequestPath.swift
```diff
@@ -91,8 +93,14 @@ extension HTTPRequest {
 
 extension HTTPRequest.Path: HTTPRequestPath {
 
-    // swiftlint:disable:next force_unwrapping
-    static let serverHostURL = URL(string: "https://api.revenuecat.com")!
+    static let serverHostURLs = [
+        "https://api.revenuecat.com",
+        "https://api2.revenuecat.com", // TODO: Add real values
+        "https://api3.revenuecat.com"
```
ajpallares, line 99: TODO: add final hosts list

Sources/Paywalls/Events/Networking/PaywallHTTPRequestPath.swift
```diff
@@ -16,7 +16,7 @@ import Foundation
 extension HTTPRequest.PaywallPath: HTTPRequestPath {
 
     // swiftlint:disable:next force_unwrapping
-    static let serverHostURL = URL(string: "https://api-paywalls.revenuecat.com")!
+    static let serverHostURLs = [URL(string: "https://api-paywalls.revenuecat.com")!]
```
ajpallares, line 19: TODO: add final hosts list

Sources/Networking/HTTPClient/HTTPRequestPath.swift
```diff
@@ -94,6 +96,17 @@ extension HTTPRequest.Path: HTTPRequestPath {
     // swiftlint:disable:next force_unwrapping
     static let serverHostURL = URL(string: "https://api.revenuecat.com")!
 
+    var fallbackPaths: [HTTPRequestPath] {
+        switch self {
+        case .getOfferings:
+            return [HTTPRequest.FallbackPath.getOfferings]
+        case .getProductEntitlementMapping:
+            return [HTTPRequest.FallbackPath.getProductEntitlementMapping]
```
tonidero, line 104: So we're only adding fallbacks for these 2 endpoints correct?

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -623,6 +645,36 @@ private extension HTTPClient {
...
+              let nextRequest = request.requestWithNextFallbackPath() else {
+            return false
+        }
+
+        Logger.debug(Strings.network.retrying_request_with_fallback_path(
+            httpMethod: nextRequest.method.httpMethod,
+            path: nextRequest.path
+        ))
+        self.state.modify {
+            $0.queuedRequests.insert(nextRequest, at: 0)
```
tonidero, line 673: Hmm nothing to do for this PR since I believe this could actually happen already, but I wonder if we should dedupe any queued requests here... In android we do something like that, and just "hook" the callbacks to the existing queued/in progress request if it exists.

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -623,6 +645,36 @@ private extension HTTPClient {
...
+    ///
+    /// This function checks the HTTP response status code to determine if the request should be retried
+    /// with the next fallback path. If the retry conditions are met, it schedules the request immediately and
+    /// returns `true` to indicate that the request was retried.
+    ///
+    /// - Parameters:
+    ///   - request: The original `HTTPClient.Request` that may need to be retried.
+    ///   - httpURLResponse: An optional `HTTPURLResponse` that contains the status code of the response.
+    /// - Returns: A Boolean value indicating whether the request was retried.
+    internal func retryRequestWithNextFallbackPathIfNeeded(
```
tonidero, line 658: Again, nothing to do in this PR, but I was thinking it would be great to have a way to test this more easily. One thing that came to mind would be to add an option to our test proxy to block only these endpoints in the main host...

tonidero (approved):
LGTM!

Sources/Networking/HTTPClient/HTTPRequestPath.swift
ajpallares: That's right. Our fallback backend only supports these two endpoints at the moment.

Sources/Networking/HTTPClient/HTTPClient.swift
ajpallares: Oh I like that! In fact I'm trying to add these fallback endpoints to our backend tests and I'm really struggling so far 😅

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -279,25 +280,46 @@ internal extension HTTPClient {
...
         func retriedRequest() -> Self {
             var copy = self
             copy.retryCount += 1
             copy.headers[RequestHeader.retryCount.rawValue] = "\(copy.retryCount)"
+            return copy
+        }
 
+        func requestWithNextFallbackPath() -> Self? {
+            var copy = self
+            copy.fallbackPathIndex = self.fallbackPathIndex?.advanced(by: 1) ?? 0
```
JayShortway, line 298: Did you by any chance ask COIN whether we should randomize the fallback that is picked?

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -279,25 +280,46 @@ internal extension HTTPClient {
...
             """
         }
+
+        private func getCurrentPath() -> HTTPRequestPath? {
+            if let fallbackPathIndex = self.fallbackPathIndex {
+                return self.httpRequest.path.fallbackPaths[safe: fallbackPathIndex]
+            } else {
+                return self.httpRequest.path
+            }
+        }
```
JayShortway, line 322: So only the path changes in a fallback scenario? I'm a bit confused as this PR title mentions hosts (and the [Diagnostics PR](https://github.com/RevenueCat/purchases-ios/pull/4982) does too), but I don't see hosts being changed. Maybe I'm misunderstanding something.

JayShortway (approved):
Looks good! Just have some general questions about naming consistency.

Sources/Networking/HTTPClient/HTTPClient.swift
ajpallares: For now there's only going to be 1 fallback host, so no randomization is needed

Sources/Networking/HTTPClient/HTTPClient.swift
ajpallares: Sorry about that. Originally, the host would only change, but we realized that the fallback backend had different paths. In the end, after talking with COIN, they've added a mapping in the fallback backend which means that the original premise was correct: only the host will change. 
I'll put this back as a draft until I've made the changes back. Sorry again!

JayShortway: No worries at all! Thanks for explaining 😄

PR #4972, opened 2025-04-09
Add new APIs to CustomEntitlementsComputationMode SDK

tonidero:
### Description
This adds some new APIs to our `CustomEntitlementsComputationMode` SDK:
- A new `ConfigurationInCustomEntitlementsComputationMode` builder to configure the SDK, including the option to select storeKit versions
- A new `configureInCustomEntitlementsComputationMode` that takes a `ConfigurationInCustomEntitlementsComputationMode` parameter
- Exposes methods to obtain `PromotionalOffer` from a product/discount
- Exposes purchase method that takes a `PurchaseParams` that accepts a `PromotionalOffer`.

### DIFF
--- a/Sources/Misc/StoreKitVersion.swift
+++ b/Sources/Misc/StoreKitVersion.swift
@@ -35,6 +35,13 @@ public extension StoreKitVersion {
     /// Let RevenueCat use the most appropiate version of StoreKit
     static let `default` = Self.storeKit2
 
+    #if ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+
+    /// Let RevenueCat use the most appropiate version of StoreKit
+    static let defaultForCustomEntitlementComputation = Self.storeKit1
+
+    #endif
+
 }
 
 extension StoreKitVersion: CustomDebugStringConvertible {
--- a/Sources/Purchasing/ConfigurationInCustomEntitlementsComputation.swift
+++ b/Sources/Purchasing/ConfigurationInCustomEntitlementsComputation.swift
@@ -1,0 +1,229 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  Configuration.swift
+//
+//  Created by Joshua Liebowitz on 5/6/22.
+
+import Foundation
+
+#if ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+
+// swiftlint:disable type_name
+/**
+ * ``ConfigurationInCustomEntitlementsComputation`` can be used when configuring the ``Purchases`` instance.
+ * It is not required to be used, but highly recommended. This class follows a builder pattern.
+ *
+ * To configure your `Purchases` instance using this object, follow these steps.
+ *
+ * **Steps:**
+ * 1. Call ``ConfigurationInCustomEntitlementsComputation/builder(withAPIKey:)``
+ *  To obtain a ``ConfigurationInCustomEntitlementsComputation/Builder`` object.
+ * 2. Set this builder's properties using the "`with(`" functions.
+ * 3. Call ``ConfigurationInCustomEntitlementsComputation/Builder/build()`` to obtain the
+ *  `ConfigurationInCustomEntitlementsComputation` object.
+ * 4. Pass the `ConfigurationInCustomEntitlementsComputation` object into
+ *  ``Purchases/configureInCustomEntitlementsComputationMode(with:)``.
+ *
+ * ```swift
+ * let configuration = ConfigurationInCustomEntitlementsComputation.Builder(withAPIKey: "MyKey",
+ *                                                                          appUserID: "SomeAppUserID).build()
+ *  Purchases.configure(with: configuration)
+ * ```
+ */
+public final class ConfigurationInCustomEntitlementsComputation: NSObject {
+
+    static let storeKitRequestTimeoutDefault: TimeInterval = 30
+    static let networkTimeoutDefault: TimeInterval = 60
+
+    let apiKey: String
+    let appUserID: String?
+    internal let observerMode: Bool
+    internal let userDefaults: UserDefaults?
+    let storeKitVersion: StoreKitVersion
+    internal let dangerousSettings: DangerousSettings?
+    internal let networkTimeout: TimeInterval
+    internal let storeKit1Timeout: TimeInterval
+    internal let platformInfo: Purchases.PlatformInfo?
+    internal let responseVerificationMode: Signing.ResponseVerificationMode
+    let showStoreMessagesAutomatically: Bool
+    internal let diagnosticsEnabled: Bool
+
+    private init(with builder: Builder) {
+        Self.verify(apiKey: builder.apiKey)
+
+        self.apiKey = builder.apiKey
+        self.appUserID = builder.appUserID
+        self.observerMode = builder.observerMode
+        self.userDefaults = builder.userDefaults
+        self.storeKitVersion = builder.storeKitVersion
+        self.dangerousSettings = builder.dangerousSettings
+        self.storeKit1Timeout = builder.storeKit1Timeout
+        self.networkTimeout = builder.networkTimeout
+        self.platformInfo = builder.platformInfo
+        self.responseVerificationMode = builder.responseVerificationMode
+        self.showStoreMessagesAutomatically = builder.showStoreMessagesAutomatically
+        self.diagnosticsEnabled = builder.diagnosticsEnabled
+    }
+
+    /// Factory method for the ``Configuration/Builder`` object that is required to create a `Configuration`
+    @objc public static func builder(withAPIKey apiKey: String, appUserID: String) -> Builder {
+        return Builder(withAPIKey: apiKey, appUserID: appUserID)
+    }
+
+    /// The Builder for ```ConfigurationInCustomEntitlementsComputation```.
+    public class Builder: NSObject {
+
+        private static let minimumTimeout: TimeInterval = 5
+
+        private(set) var apiKey: String
+        private(set) var appUserID: String?
+        internal var observerMode: Bool {
+            switch purchasesAreCompletedBy {
+            case .revenueCat:
+                return false
+            case .myApp:
+                return true
+            }
+        }
+        private(set) internal var purchasesAreCompletedBy: PurchasesAreCompletedBy = .revenueCat
+        private(set) internal var userDefaults: UserDefaults?
+        private(set) internal var dangerousSettings: DangerousSettings? = DangerousSettings(
+            customEntitlementComputation: true
+        )
+        private(set) internal var networkTimeout = Configuration.networkTimeoutDefault
+        private(set) internal var storeKit1Timeout = Configuration.storeKitRequestTimeoutDefault
+        private(set) internal var platformInfo: Purchases.PlatformInfo?
+        private(set) internal var responseVerificationMode: Signing.ResponseVerificationMode = .default
+        private(set) var showStoreMessagesAutomatically: Bool = true
+        private(set) internal var diagnosticsEnabled: Bool = false
+        private(set) var storeKitVersion: StoreKitVersion = .defaultForCustomEntitlementComputation
+
+        /**
+         * Create a new builder with your API key.
+         * - Parameter apiKey: The API Key generated for your app from https://app.revenuecat.com/
+         */
+        @objc public init(withAPIKey apiKey: String, appUserID: String) {
+            self.apiKey = apiKey
+            self.appUserID = appUserID
+        }
+
+        /// Update your API key.
+        @objc public func with(apiKey: String) -> Builder {
+            self.apiKey = apiKey
+            return self
+        }
+
+        /**
+         * Set an `appUserID`.
+         * - Parameter appUserID: The unique app user id for this user. This user id will allow users to share their
+         * purchases and subscriptions across devices. Pass `nil` or an empty string if you want ``Purchases``
+         * to generate this for you.
+         *
+         * - Note: Best practice is to use a salted hash of your unique app user ids.
+         *
+         * - Important: Set this property if you have your own user identifiers that you manage.
+         */
+        @_disfavoredOverload
+        @objc public func with(appUserID: String?) -> Builder {
+            self.appUserID = appUserID
+            return self
+        }
+
+        @available(*, deprecated, message: """
+        The appUserID passed to configure is a constant string known at compile time.
+        This is likely a programmer error. This ID is used to identify the current user.
+        See https://docs.revenuecat.com/docs/user-ids for more information.
+        """)
+        // swiftlint:disable:next missing_docs
+        public func with(appUserID: StaticString) -> ConfigurationInCustomEntitlementsComputation.Builder {
+            Logger.warn(Strings.identity.logging_in_with_static_string)
+            return self.with(appUserID: "\(appUserID)")
+        }
+
+        /// Set `showStoreMessagesAutomatically`. Enabled by default.
+        /// If enabled, if the user has billing issues, has yet to accept a price increase consent, is eligible for a
+        /// win-back offer, or there are other messages from StoreKit, they will be displayed automatically when
+        /// the app is initialized.
+        ///
+        /// If you want to disable this behavior so that you can customize when these messages are shown, make sure
+        /// you configure the SDK as early as possible in the app's lifetime, otherwise messages will be displayed
+        /// automatically.
+        /// Then use the ``Purchases/showStoreMessages(for:)`` method to display the messages.
+        /// More information:  https://rev.cat/storekit-message
+        /// - Important: Set this property only if you're using Swift. If you're using ObjC, you won't be able to call
+        /// the related methods
+        @objc public func with(showStoreMessagesAutomatically: Bool) -> Builder {
+            self.showStoreMessagesAutomatically = showStoreMessagesAutomatically
+            return self
+        }
+
+        /// Set ``StoreKitVersion``.
+        ///
+        /// Defaults to ``StoreKitVersion/default`` which lets the SDK select
+        /// the most appropriate version of StoreKit. Currently defaults to StoreKit 2.
+        ///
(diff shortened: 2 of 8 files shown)


tonidero:
@RCGitBot please test

Sources/Misc/StoreKitVersion.swift
```diff
@@ -35,6 +35,13 @@ public extension StoreKitVersion {
     /// Let RevenueCat use the most appropiate version of StoreKit
     static let `default` = Self.storeKit2
 
+    #if ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+
+    /// Let RevenueCat use the most appropiate version of StoreKit
+    static let defaultForCustomEntitlementComputation = Self.storeKit1
```
ajpallares, line 41: I wonder, would it make more sense to have it like this?

```suggestion
    #if ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
  
    /// Let RevenueCat use the most appropiate version of StoreKit
    static let `default` = Self.storeKit1
    
    #else

    /// Let RevenueCat use the most appropiate version of StoreKit
    static let default = Self.storeKit2
    
    #endif
```

Sources/Purchasing/ConfigurationInCustomEntitlementsComputation.swift
```diff
@@ -0,0 +1,229 @@
...
+ * 4. Pass the `ConfigurationInCustomEntitlementsComputation` object into
+ *  ``Purchases/configureInCustomEntitlementsComputationMode(with:)``.
+ *
+ * ```swift
+ * let configuration = ConfigurationInCustomEntitlementsComputation.Builder(withAPIKey: "MyKey",
+ *                                                                          appUserID: "SomeAppUserID).build()
+ *  Purchases.configure(with: configuration)
+ * ```
+ */
+public final class ConfigurationInCustomEntitlementsComputation: NSObject {
```
ajpallares, line 40: Do we need the `@objc` attribute here as in `Configuration`? If not, then does it need to be `NSObject` subclass?

Sources/Purchasing/ConfigurationInCustomEntitlementsComputation.swift
```diff
@@ -0,0 +1,229 @@
...
+        self.storeKit1Timeout = builder.storeKit1Timeout
+        self.networkTimeout = builder.networkTimeout
+        self.platformInfo = builder.platformInfo
+        self.responseVerificationMode = builder.responseVerificationMode
+        self.showStoreMessagesAutomatically = builder.showStoreMessagesAutomatically
+        self.diagnosticsEnabled = builder.diagnosticsEnabled
+    }
+
+    /// Factory method for the ``Configuration/Builder`` object that is required to create a `Configuration`
+    @objc public static func builder(withAPIKey apiKey: String, appUserID: String) -> Builder {
```
ajpallares, line 76: Related to my previous comment. Is `@objc` needed here?

Sources/Purchasing/ConfigurationInCustomEntitlementsComputation.swift
```diff
@@ -0,0 +1,229 @@
...
+        self.dangerousSettings = builder.dangerousSettings
+        self.storeKit1Timeout = builder.storeKit1Timeout
+        self.networkTimeout = builder.networkTimeout
+        self.platformInfo = builder.platformInfo
+        self.responseVerificationMode = builder.responseVerificationMode
+        self.showStoreMessagesAutomatically = builder.showStoreMessagesAutomatically
+        self.diagnosticsEnabled = builder.diagnosticsEnabled
+    }
+
+    /// Factory method for the ``Configuration/Builder`` object that is required to create a `Configuration`
```
ajpallares, line 75:
```suggestion
    /// Factory method for the ``ConfigurationInCustomEntitlementsComputation/Builder`` object that is required to create a `ConfigurationInCustomEntitlementsComputation`
```

Sources/Purchasing/ConfigurationInCustomEntitlementsComputation.swift
ajpallares, line 1: I have my doubts about this one. I see that the code of `ConfigurationInCustomEntitlementsComputation` is basically the same as `Configuration`, only some builder methods are missing and thus the documentation on the class is different. 
Wouldn't it make more sense to use the existing `Configuration` object and making the distinctions in place with `#if ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION`. 
This way, we don't even need to declare a new `Purchases.configureInCustomEntitlementComputationMode` and opening up the existing `Purchases.configure` would suffice. WDYT?

It would be the same idea as what you did with `PurchaseParams`

ajpallares (commented):
This was quick! Thank you! I have some comments/suggestions, mostly about the `ConfigurationInCustomEntitlementsComputation`

Sources/Purchasing/ConfigurationInCustomEntitlementsComputation.swift
ajpallares: One potential reason to add `ConfigurationInCustomEntitlementsComputation` instead of reusing `Configuration` is that in the current version, the `Configuration` object and the `Purchases.configure(with configuration: Configuration)` is open to `ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION`. So making changes to `Configuration` would probably be a breaking change. 

That said, I believe that with these changes, the `Purchases.configure(with configuration: Configuration)` is now under `#if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION`, which means it's not available anymore if `ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION`. So that'd be a breaking change.

In summary, if we are to make some breaking changes of this style, I think we might as well only change the `Configuration` object using the `ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION` flag.

tonidero: Right, TBH, I also would have preferred to go with a single  Configuration object... I had 2 reasons to go this way:
- I wanted to change the `Builder` constructor to require an `appUserID` as well, which is why I decided to go with creating a separate Configuration object... But then again, we could make the different init object within the same class as well. 
- I wanted to avoid breaking changes for the CustomEntitlementComputationMode, since `Configuration` is already exposed there. However... It might be ok? I think few companies use this mode, so it might be ok to do this and just give a heads up.

Sources/Purchasing/ConfigurationInCustomEntitlementsComputation.swift
tonidero: Yeah I also had doubts about that... I don't think it's needed TBH, since this shouldn't be used by any hybrids and customers of this mode are using Swift... So yeah, I will remove NSObject here. Good catch!

Sources/Misc/StoreKitVersion.swift
tonidero: Hmm right, that probably makes more sense 😅. Thanks!

Sources/Purchasing/ConfigurationInCustomEntitlementsComputation.swift
tonidero: Yeah and as you said, I already made the breaking change by hiding this method. So yeah I will change this then 👍 Thanks!

Sources/Purchasing/Configuration.swift
```diff
@@ -70,11 +72,22 @@ import Foundation
         self.diagnosticsEnabled = builder.diagnosticsEnabled
     }
 
+    #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
```
tonidero, line 75: Also, as discussed, some of these are breaking changes, but we think it's ok since these methods shouldn't have been used by people in this mode.

Sources/Purchasing/ConfigurationInCustomEntitlementsComputation.swift
tonidero: Does not apply anymore after removing this class

Sources/Purchasing/ConfigurationInCustomEntitlementsComputation.swift
tonidero: Does not apply anymore after removing this class

Sources/Purchasing/ConfigurationInCustomEntitlementsComputation.swift
tonidero: Done in https://github.com/RevenueCat/purchases-ios/pull/4972/commits/6dc59ec6087c1160262a7efc70338f961956a68f

Sources/Purchasing/Configuration.swift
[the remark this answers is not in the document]
ajpallares: Yeah, I agree 👍

ajpallares: Agreed

Sources/Purchasing/Configuration.swift
```diff
@@ -211,6 +243,8 @@ import Foundation
             return self
         }
```
ajpallares, line 244: Just making sure that we still want this when `ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION`, or perhaps, `showStoreMessagesAutomatically` should be disabled by default in this mode.

Sources/Purchasing/Configuration.swift
JayShortway: We've also been told they're not being used, so should be good.

Sources/Purchasing/Purchases/PurchaseParams.swift
```diff
@@ -34,24 +32,39 @@ import Foundation
     let package: Package?
     let product: StoreProduct?
     let promotionalOffer: PromotionalOffer?
+
+    #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+
     let winBackOffer: WinBackOffer?
     let metadata: [String: String]?
 
+    #endif
```
JayShortway, line 41: Good call leaving these disabled!

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1371,6 +1367,8 @@ public extension Purchases {
         )
     }
 
+    #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+
```
ajpallares, line 1371: Just to confirm that we really want to remove this `configure(with builder: Configuration.Builder)` when `ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION`

Sources/Purchasing/Configuration.swift
```diff
@@ -247,6 +279,8 @@ import Foundation
...
 
+        #endif
+
         /// Set ``StoreKitVersion``.
         ///
         /// Defaults to ``StoreKitVersion/default`` which lets the SDK select
```
JayShortway, line 288: Nit, the line right below this one says "Currently defaults to StoreKit 2.", which is not entirely true for CEC mode. Not super important, but could be confusing.

JayShortway (commented):

Sources/Purchasing/Purchases/PurchaseParams.swift
```diff
@@ -34,24 +32,39 @@ import Foundation
     let package: Package?
     let product: StoreProduct?
     let promotionalOffer: PromotionalOffer?
+
+    #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+
     let winBackOffer: WinBackOffer?
```
ajpallares, line 38: Again, to confirm. Are we sure we don't want to enable `winBackOffers` when `ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION`

ajpallares (commented):
I think it looks great! I have some more comments, mostly some checks to make sure.

Sources/Purchasing/Configuration.swift
tonidero: Good question! I believe devs using this mode wanted to disable this so I left the API here, so they can choose any behavior they prefer... We can change the default, but I think it's better for them to have the option to choose.

Sources/Purchasing/Purchases/Purchases.swift
tonidero: TBH, this feels a bit wrong to me, considering we shouldn't be passing the builder object directly, unless we wanted to do additional changes, which we shouldn't... So, since I believe there was no need I hid it. Lmk if you think otherwise!

Sources/Purchasing/Purchases/PurchaseParams.swift
tonidero: From my understanding, I believe we were going to recommend promo offers for now, so I kept this disabled. However, it should be very easy to add it if needed.

Sources/Purchasing/Configuration.swift
ajpallares: Makes sense!

Sources/Purchasing/Purchases/Purchases.swift
ajpallares: Yeah, I think it makes sense. Thanks for the explanation!

Sources/Purchasing/Purchases/PurchaseParams.swift
JayShortway: Yea indeed, as they want to target users who are about to cancel, which winbacks cannot be used for.

ajpallares: Thank you both for the explanation!

Sources/Purchasing/Configuration.swift
tonidero: Just an update that I decided to revert the change changing the default for now. If we bring that change back, I will do this change 👍

Sources/Misc/Deprecations.swift
```diff
@@ -337,11 +337,15 @@ extension CustomerInfo {
...
 public extension Configuration.Builder {
 
+    #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+
     @available(*, deprecated, message: "Use .with(storeKitVersion:) to enable StoreKit 2")
     @objc func with(usesStoreKit2IfAvailable: Bool) -> Configuration.Builder {
         return self.with(storeKitVersion: usesStoreKit2IfAvailable ? .storeKit2 : .default)
     }
 
+    #endif
```
JayShortway, line 347: Just checking that this is a conscious decision to remove this in CEC mode, rather than deprecate?

JayShortway (approved):
Good stuff! Just 1 question.

Sources/Misc/Deprecations.swift
tonidero: Right, I mostly wanted to avoid confussion by having this setting from V4 and also the "false" case in this method, is the `default`, which in V5 we are leaving to SK2... so this method doesn't really do anything in V5 I think 😓 I don't really feel strongly about it though.

JayShortway: Yea, I think it makes sense to remove it to avoid any possible confusion!

ajpallares (approved):
Great!

Sources/Purchasing/Configuration.swift
aboedo: @fire-at-will I think this might have implications for streamlined purchases? As in, we _didn't_ want it enabled if it could cause you to initiate a purchase while not logged in? 
Or is that fine as long as devs are using the readyForPromotedProduct API?

fire-at-will: @aboedo Here's my understanding:

At least for win-backs, StoreKit messages won't trigger `readyForPromotedProduct`. According to Apple's docs:

> When customers redeem a win-back offer using the win-back offer sheet, StoreKit treats this transaction the same as a purchase that occurs outside of the app. Specifically, the app receives the transaction in the [updates](https://developer.apple.com/documentation/storekit/transaction/updates) sequence of [Transaction](https://developer.apple.com/documentation/storekit/transaction). Be sure your app creates a task to iterate through [updates](https://developer.apple.com/documentation/storekit/transaction/updates) as soon as it launches. For more information and sample code, see [updates](https://developer.apple.com/documentation/storekit/transaction/updates). [(source)](https://developer.apple.com/documentation/storekit/supporting-win-back-offers-in-your-app#Handle-offers-redeemed-through-the-win-back-offer-sheet)

That said, `showStoreMessagesAutomatically` enabled does still open a possibility for a purchase to be made when a user is signed out, even if the developer wants to force users to sign in before making a purchase.

Here's an example with win-back offers, but the same might apply in other scenarios like billing issues:
1. A user purchases a sub, cancels, signs out, and then waits a while and becomes eligible for a win-back offer
2. The user opens their app
3. `showStoreMessagesAutomatically` is enabled, and StoreKit shows a StoreKit message with the [win-back reason](https://developer.apple.com/documentation/storekit/message/reason-swift.struct/winbackoffer), allowing the user to redeem a win-back offer. This happens immediately on app open before the user signs back in
4. The user redeems the win-back offer and is re-subscribed

In this case, they've made a purchase without being signed in. We'd probably be able to match the receipts with their old user ID in the backend, but there are probably some edge cases to this as well.

Developers can delay showing the win-back offer by setting `showStoreMessagesAutomatically` to false and then calling `showStoreMessages(for : [.winBackOffer])`

As far as I can tell, Apple's docs don't say anything about disabling streamlined purchasing also disabling the StoreKit messages

It's all a bit complicated, let me know if y'all would like to huddle to chat about it 👍

aboedo: let's chat about it quickly, but I think we should be fine for custom entitlements mode at least - there isn't a concept of "sign out", rather just "switch user" so in that case the purchase would get associated with the previous user until they log in again

Sources/Purchasing/Purchases/PurchaseParams.swift
aboedo: let's enable them (it can be a follow-up). I don't see any issues with allowing for customers to try out winbacks. Maybe they're not great at solving the specific "about to cancel" use case but they can solve other cases.

aboedo: 🤔 maybe it's actually better for it to be a separate step: https://developer.apple.com/help/app-store-connect/manage-subscriptions/manage-streamlined-purchasing/

This mode would not be compatible with Streamlined Purchasing, but you can only disable it after you have an approved build that migrates to v5, which has the PurchaseIntent APIs.

JayShortway:
@RCGitBot please test

JayShortway:
@RCGitBot please test

PR #4974, opened 2025-04-09
feat: Remove Something went wrong from bundle

facumenzella:
### Motivation
While adding accessibility identifiers for maestro, I'll be slowly removing localized strings from the bundle.

### Description
This removes `Something went wrong`


facumenzella:
Closing for now

PR #4975, opened 2025-04-09
Make Paywalls v2 Text use verbatim

joshdholtz:
### Motivation

The `TextViewComponent` component could end up using localized values in the developer's app

### Description

Have the `TextViewComponent` use the `Text(verbatim:)` constructor so it doesn't attempt to localize anything

(This was _intended_ when developing this but was not done correctly)

| Before | After (with fix) |
|--------|--------|
| Simulator Screenshot - iPhone 16 Pro Max - 2025-04-09 at 09 47 34 | Simulator Screenshot - iPhone 16 Pro Max - 2025-04-09 at 09 47 01 |

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
@@ -50,7 +50,7 @@ struct TextComponentView: View {
             )
         ) { style in
             if style.visible {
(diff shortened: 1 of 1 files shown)


JayShortway (approved):
Good find!

tonidero (approved):
Oh interesting... Heads up as well @RevenueCat/customer-center, not sure if this could affect customer center as well

PR #4979, opened 2025-04-10
[DX-345] Update the `testSDKHealth` to use the new `/health_report` endpoint

polpielladev:
### Checklist

### Motivation

To guide RevenueCat users better while configuring the SDK, we would like to modify the `testSDKHealth` endpoint to use the new `/health_report` endpoint and, at the same time, improve the `Status` section of the Debug View to include better and more in-depth information about why the configuration is not valid.

https://github.com/user-attachments/assets/2e1de5a4-7d8f-4f7a-afcb-5fe2dccd674c

### Description

The new method and view do the following validations:
- Check if the API key is valid.
- Check if the SDK version is supported.
- Check if the Bundle ID set in the app matches the one configured in the RevenueCat dashboard.
- Check if there is at least one product that is 'OK' in App Store Connect. If not, we fail. We also log warnings for any other products that might not be okay.
- If there are no offerings we log a warning (offerings are optional but recommended). This will serve as a bit of education to let people know that they need offerings to make use of Paywalls.
- If there are offerings and the default offering has no packages, we fail.
- If there are offerings and there are no valid products in the default offering, we fail.
- If the default offering is okay but we spot issues in any other ones, we log a warning and we display it in the details view.


ajpallares (commented):
Looking good so far! Thank you! Some initial comments, very small things.

I need to finish reviewing it though, mostly the UI parts. 

The only thing that feels weird to me, is that I would think that in order to check the health status of the configuration, the `appUserID` shouldn't be needed. I saw the code comment where it's explained that it's needed for fetching the offerings, but, is it really really needed? I'm curious.

Sources/Networking/HTTPClient/HTTPRequestPath.swift
```diff
@@ -260,6 +268,9 @@ extension HTTPRequest.Path: HTTPRequestPath {
         case .postRedeemWebPurchase:
             return "post_redeem_web_purchase"
 
+        case .appHealthReport:
+            return "get_app_health_report"
+
```
aboedo, line 273: we can do it as a follow-up, but we might as well remove the other health API and instead bridge internally to this one

Sources/Networking/Operations/HealthReportOperation.swift
```diff
@@ -0,0 +1,73 @@
...
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  HealthReportOperation.swift
+//
+//  Created by Pol Piella on 4/8/25.
```
aboedo, line 12: one of us one of us 
one of us

Sources/Networking/Responses/HealthReportResponse.swift
```diff
@@ -0,0 +1,133 @@
...
+                .map({ .bundleId($0) })
+
+        case .products:
+            details = (try container.decodeIfPresent(ProductsCheckDetails.self, forKey: .details))
+                .map({ .products($0) })
+
+        default:
+            details = nil
+        }
+    }
```
aboedo, line 113: This bit is mostly so that if the value isn't present, the call doesn't entirely fail, right? Maybe in this case, we don't really need it? We are trying to find out what could go wrong after all. It might be the case where a high level catch works

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -1213,6 +1213,11 @@ internal protocol InternalPurchasesType: AnyObject {
     /// - Throws: `PublicError` if request failed.
     func healthRequest(signatureVerification: Bool) async throws
 
+    /// Asks the server for an in-depth report of the SDK's configuration
+    /// - Throws: A `BackendError` if the request fails for an invalid API key or for connectivity issues
+    /// - Returns: A health report with all checks performed in the server and their status
```
aboedo, line 1218: let's add a bit more detail about _when_ you want to use this. 
Or maybe even keep it internal but automatically call it in debug builds? 

Otherwise, this could easily get called in a loop or like on every screen shown in a mega app

Sources/Support/DebugUI/DebugContentViews.swift
```diff
@@ -414,6 +435,78 @@ private struct DebugPackageView: View {
...
+                                        offering.status.icon.imageScale(.small)
+                                        Text(offering.identifier)
+                                    }
+                                    Text(
+                                        offering.packages.isEmpty ?
+                                        "Offerings must have at least one package" :
+                                        """
+                                        Products \(offering.packages.map { "'\($0.productIdentifier)'" }.formatted()) \
+                                        in the offering have issues, check the products list above for more details.
+                                        """
```
aboedo, line 495: this is unreasonably exciting to me

Sources/Support/DebugUI/DebugViewModel.swift
```diff
@@ -83,8 +83,27 @@ extension DebugViewModel {
...
+        }
+    }
+
+    var diagnosticsExplainer: String? {
+        switch self.diagnosticsResult {
+        case let .loaded(healthStatus):
+            switch healthStatus {
+            case .healthy(warnings: let warnings):
+                return warnings.count > 0 ? """
+                The dashboard configuration is valid, however we encountered some potential issues \
```
aboedo, line 100: what's "the dashboard"? 
Feels like this copy bakes in the assumption that external developers think of the dashboard the same we do, but it's really an internal name. 

We can point to revenuecat.com, or mention "the revenuecat web app" or even "your revenuecat configuration", but dashboard might not mean anything to the reader.

Sources/Support/PurchasesDiagnostics.swift
```diff
@@ -44,6 +45,63 @@ public final class PurchasesDiagnostics: NSObject {
...
+    /// Enum representing the status of a product in the store
+    public enum ProductStatus {
+        /// Product is configured correctly in App Store Connect
+        case valid
+        /// There was a problem checking the product state in App Store Connect
+        case couldNotCheck
+        /// The product does not exist in App Store Connect
+        case notFound
+        /// The product is in a state that requires action from Apple or the Developer before being ready for production
+        case actionInProgress
```
aboedo, line 57: crying tears of joy

Sources/Support/PurchasesDiagnostics.swift
```diff
@@ -167,11 +263,41 @@ extension PurchasesDiagnostics.Error: CustomNSError {
 
     var localizedDescription: String {
         switch self {
+        case .notAuthorizedToMakePayments: return "The person is not authorized to make payments on this device."
         case let .unknown(error): return "Unknown error: \(error.localizedDescription)"
         case let .failedConnectingToAPI(error): return "Error connecting to API: \(error.localizedDescription)"
         case let .failedFetchingOfferings(error): return "Failed fetching offerings: \(error.localizedDescription)"
         case let .failedMakingSignedRequest(error): return "Failed making signed request: \(error.localizedDescription)"
         case .invalidAPIKey: return "API key is not valid"
+        case .invalidSDKVersion: return "SDK Version is not supported"
```
aboedo, line 272: when does this one happen?

Sources/Support/PurchasesDiagnostics.swift
```diff
@@ -167,11 +263,41 @@ extension PurchasesDiagnostics.Error: CustomNSError {
 
     var localizedDescription: String {
         switch self {
+        case .notAuthorizedToMakePayments: return "The person is not authorized to make payments on this device."
```
aboedo, line 266: :chefskiss:

Sources/Support/PurchasesDiagnostics.swift
```diff
@@ -59,36 +117,74 @@ extension PurchasesDiagnostics {
...
         }
     }
 
+    private var canMakePayments: Bool {
+        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, visionOS 1.0, *) {
+            return AppStore.canMakePayments
+        } else {
+            return SKPaymentQueue.canMakePayments()
+        }
+    }
```
aboedo, line 168: we already have our own API available through Purchases, let's use that instead

aboedo: also I'm now realizing that we haven't updated it for SK2 (cc @fire-at-will @MarkVillacampa) so we might as well add replace that method's implementation with this one lol 
https://github.com/RevenueCat/purchases-ios/blob/d9626ab8574c88f1bbde7e2610e2a7adcebfde48/Sources/Purchasing/Purchases/Purchases.swift#L162

(can be a separate PR, this isn't exactly urgent or blocking)

Sources/Support/PurchasesDiagnostics.swift
```diff
@@ -167,11 +263,41 @@ extension PurchasesDiagnostics.Error: CustomNSError {
...
+            """
+
+            return offendingOffering.packages.isEmpty ? noPackages : packagesNotReady
+        case let .invalidBundleId(payload):
+            guard let payload else {
+                return "Bundle ID in your app does not match the Bundle ID in the RevenueCat dashboard"
+            }
+            let sdkBundleId = payload.sdkBundleId
+            let appBundleId = payload.appBundleId
+            return "Bundle ID in your app '\(sdkBundleId)' does not match the RevenueCat app Bundle ID '\(appBundleId)'"
```
aboedo, line 294: I'm confused about these two, which one is which? 
does "app" here represent the configuration in the web app?

aboedo: would be killer to add a link to the app configuration in RC web so that we can put it in these and then you can easily go fix web side

Sources/Support/DebugUI/InvalidProductErrorPayload+StatusMessage.swift
```diff
@@ -0,0 +1,45 @@
...
+            return self.description
+        case .notFound:
+            return """
+                Product not found in App Store Connect. You need to create a product with identifier: \
+                '\(self.identifier)' in App Store Connect to use it for production purchases.
+                """
+        case .actionInProgress:
+            return """
+                Some process is ongoing and needs to be completed before using this product in production purchases, \
+                controlled either by the developer or by Apple (state: \(self.description))
```
aboedo, line 31: this is the only one that doesn't quite have remediation steps, what do I do if I run into this?

aboedo (commented):
this is amazing Pol!! I left a few comments but I absolutely love it

Sources/Networking/Backend.swift
```diff
@@ -158,6 +158,24 @@ extension Backend {
...
+    /// - Parameter appUserID: An `appUserID` that allows the Back End to fetch offerings
+    /// - Returns: A report will all validation checks along with their status
+    func healthReportRequest(appUserID: String) async throws -> HealthReport {
+        try await Async.call { (completion: @escaping (Result<HealthReport, BackendError>) -> Void) in
+            self.internalAPI.healthReportRequest(appUserID: appUserID) { result in
+                switch result {
+                case let .success(response): completion(.success(response))
+                case .failure(let error): completion(.failure(error))
+                }
+            }
```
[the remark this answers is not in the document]
polpielladev: Ah! Good point! Updated now 💪

Sources/Networking/HTTPClient/HTTPRequestPath.swift
polpielladev: Oh yeah 100%, this is the next PR that I will be doing, getting rid of the calls to existing health endpoint from the SDK 💪

polpielladev: Logged it as a ticket: https://linear.app/revenuecat/issue/DX-366/clean-up-the-old-health-endpoint-calls-from-the-sdk

Sources/Support/DebugUI/DebugContentViews.swift
polpielladev: Honestly! ❤️ I can't wait to ship thisssss 🚀💪

Sources/Purchasing/Purchases/PurchasesType.swift
polpielladev: 100% Let me work on the comment a bit more. Unfortunately, since we have to use it in the Sample App, we're going to have to mark it as public, but we can definitely inform users of when they would like to use it.

In terms of running automatically on DEBUG builds, that is for sure the next step, I would like to try validate the approach in a less risky way through the debug menu and while we build the sample app, as we might need some tweaks here and there, but once we are done, let's do it!! 💪

polpielladev: Looking at this type it looks like it is already internal, but I think we are both referring to the methods in the `PurchasesDiagnostics` type.

I have updated the comments now and added a couple of `Important` notes to the documentation:

Screenshot 2025-04-11 at 11 36 22

Sources/Support/PurchasesDiagnostics.swift
polpielladev: Oh that's a very good point! Let's definitely fix it in a follow-up PR 🚀 💪

Sources/Support/PurchasesDiagnostics.swift
polpielladev: Actually, by the looks of things, this is an Android-only error, let me remove it from here 👍

polpielladev: Updated! ✅

Sources/Support/PurchasesDiagnostics.swift
polpielladev: Yeah tbf I did struggle with the wording of this and reading it back it is incredibly confusing hahaha.

Also, that is a killer feature!! Let me tweak the BE and add a small button for it in the debug view model, be right back hahaha  👀

polpielladev: Added it!!

RocketSim_Screenshot_iPhone_16_Pro_6 3_2025-04-11_14 14 16

Sources/Support/DebugUI/DebugViewModel.swift
polpielladev: Good shout! I have updated the copy now 👍

polpielladev:
@ajpallares I seem to be getting some failures when compiling the unit test targets, any idea why these types can not be found? Could it be Objective-C interop related? Or have I missed something else? 👀 

https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/26584/workflows/e56b190c-0186-4477-b6be-e8b1d0019066/jobs/310082

ajpallares:
> @ajpallares I seem to be getting some failures when compiling the unit test targets, any idea why these types can not be found? Could it be Objective-C interop related? Or have I missed something else? 👀
> 
> https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/26584/workflows/e56b190c-0186-4477-b6be-e8b1d0019066/jobs/310082

All the tests use the `RevenueCat.xcworkspace`. I deduce you used the Package.swift to develop these changes, right?  That would explain why the new files you created are not referenced in the workspace. 

We definitely need to fix this so that the SDK can be developed directly using the Package.swift. Sorry about that!

Sources/Purchasing/Purchases/PurchasesType.swift
aboedo: maybe we could also only leave it defined for debug builds to prevent this getting shipped in a mega app

aboedo: i.e.: wrapping it in `#if DEBUG`

Then we wouldn't even need the second part of that warning - the compiler would prevent it (although it's good to keep it for clarity)

Sources/Support/PurchasesDiagnostics.swift
aboedo: 😻😻😻

polpielladev:
> All the tests use the RevenueCat.xcworkspace. I deduce you used the Package.swift to develop these changes, right? That would explain why the new files you created are not referenced in the workspace.

Currently, the "official" way to develop the SDK is using the Xcode workspace. We definitely need to fix this so that the SDK can be developed directly using the Package.swift. Sorry about that!

Ahhhhh!!! Gotcha!!! Thanks so much, I will fix it now

polpielladev:
> > All the tests use the RevenueCat.xcworkspace. I deduce you used the Package.swift to develop these changes, right? That would explain why the new files you created are not referenced in the workspace.
> 
> Currently, the "official" way to develop the SDK is using the Xcode workspace. We definitely need to fix this so that the SDK can be developed directly using the Package.swift. Sorry about that!

Ahhhhh!!! Gotcha!!! Thanks so much, I will fix it now

Sources/Purchasing/Purchases/PurchasesType.swift
polpielladev: Oh yeah good point, done it now. I will leave the `Important` block just in case it is not immediate at first glance if people get failures on prod builds 👍

Sources/Support/DebugUI/SDKHealthStatus+Icon.swift
```diff
@@ -0,0 +1,29 @@
...
+
+@available(iOS 16.0, macOS 13.0, *)
+extension PurchasesDiagnostics.SDKHealthStatus {
+    var icon: some View {
+        switch self {
+        case let .healthy(warnings):
+            Image(systemName: warnings.count > 0 ? "checkmark.circle.badge.questionmark.fill" : "checkmark.circle.fill")
+                .foregroundColor(.green)
+        case .unhealthy:
+            Image(systemName: "exclamationmark.triangle")
```
ajpallares, line 24: Not my expertise, but since an `unhealthy` status is actually meaning an error that must be fixed, shouldn't we use here `"xmark.circle.fill" (the same as `PurchasesDiagnostics.SDKHealthCheckStatus` is using for the `failed` case)

Sources/Support/PurchasesDiagnostics.swift
```diff
@@ -59,36 +132,107 @@ extension PurchasesDiagnostics {
...
 
-    /// Perform tests to ensure SDK is configured correctly.
-    /// - `Throws`: ``PurchasesDiagnostics/Error`` if any step fails
-    @objc(testSDKHealthWithCompletion:)
+    /// Status of the SDK Health report
+    public enum SDKHealthStatus: Sendable {
+        /// SDK configuration is valid but might have some non-blocking issues
+        case healthy(warnings: [PurchasesDiagnostics.Error])
+        /// SDK configuration is not valid and has issues that must be resolved
+        case unhealthy(PurchasesDiagnostics.Error)
```
ajpallares, line 192: Question here. I see that we show all the `healthy` issues together (using `[PurchasesDiagnostics.Error]`) but for the unhealthy ones, we only show them one at a time. Is there a reason for this? It looks to me like we could have `unhealthy([PurchasesDiagnostics.Error])` to show them all at once.

ajpallares (approved):
This is Awesome. Thanks so much Pol! It will be extremely users for our users (and for us 😄)! To me it looks great!

Sources/Support/PurchasesDiagnostics.swift
polpielladev: The reason why we only return one error, as opposed to the healthy with warnings case is that the BE just halts on the first error, so we should only highlight that one and there won't be any other errors at that time.

I wanted to leave warnings out of the `unhealthy` state, just so that people keep focus on the one error that's blocking 👍

ajpallares: Ohh makes sense. Thanks for the explanation!

Sources/Support/DebugUI/SDKHealthStatus+Icon.swift
polpielladev: 100%! Great catch, I ended up using what we had before, but having the x asset makes a lot more sense, on ittt!!

PR #4981, opened 2025-04-10
Add Paddle Store enum case

fire-at-will:
### Description
This PR adds the `Paddle` case to the `Store` enum. Its value is set to 9, which is the same as its value in the backend. It also adds the Paddle store to several required places in the Customer Center, including introducing a new `store_paddle` localized string in Customer Center. 

### Follow Up Items
- Add the new Paddle localized string to the backend for Customer Center
- Merge the corresponding Android PR https://github.com/RevenueCat/purchases-android/pull/2353
- Add the new Store case to the hybrid SDKs


fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

RevenueCatUI/Data/CustomerInfoFixtures.swift
```diff
@@ -14,7 +14,7 @@
 import RevenueCat
 
 // swiftlint:disable force_unwrapping
-
+// swiftlint:disable type_body_length
```
facumenzella: nit:
```suggestion
// swiftlint:disable force_unwrapping type_body_length 
```

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -141,6 +141,7 @@ public struct CustomerCenterConfigData: Equatable {
             case storeRCBilling = "store_web"
             case storeExternal = "store_external"
             case storeUnknownStore = "store_unknown"
+            case storePaddle = "store_paddle"
```
facumenzella, line 144: this is not localized in khepri right? can't find it

fire-at-will: It's not! A few of the other stores aren't localized either. I asked about this in the private Catforms channel and folks said that it should just be "Paddle" in other languages

RevenueCatUI/Data/CustomerInfoFixtures.swift
fire-at-will: For some reason GitHub wouldn't let me auto-apply this change, but I've done it in e74c2122cf3a5de228264253300703f81979a38c 👍

PR #4982, opened 2025-04-10
[Diagnostics] add `host` parameter to `http_request_performed` event

ajpallares:
### Checklist

### Description
Adds a new `host` parameter to the existing`http_request_performed` diagnostics event, which can be useful with the new logic for retry requests using fallback hosts when receiving a server-down response

### DIFF
--- a/Sources/Diagnostics/DiagnosticsEvent.swift
+++ b/Sources/Diagnostics/DiagnosticsEvent.swift
@@ -81,6 +81,7 @@ struct DiagnosticsEvent: Codable, Equatable {
     struct Properties: Codable, Equatable {
         let verificationResult: String?
         let endpointName: String?
+        let host: String?
         let responseTimeMillis: Int?
         let storeKitVersion: String?
         let successful: Bool?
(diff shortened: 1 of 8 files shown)


Sources/Diagnostics/DiagnosticsEvent.swift
```diff
@@ -81,6 +81,7 @@ struct DiagnosticsEvent: Codable, Equatable {
     struct Properties: Codable, Equatable {
         let verificationResult: String?
         let endpointName: String?
+        let host: String?
```
tonidero, line 84: We need to add this to the backend schema as well (should just be ignored until then). Lmk if you need any help with this!!

tonidero (commented):
LGTM! Will approve once it's ready for review

Sources/Diagnostics/DiagnosticsEvent.swift
ajpallares: Yep! Before doing the work, I wanted to make sure that this is the implementation (and parameter name) we wanted. Thanks for flagging!

JayShortway (commented):
Nice one!

tonidero (approved):
LGTM! Tests seem to be failing though

ajpallares:
> LGTM! Tests seem to be failing though

Oops! I forgot to update with latest changes in the fallback API host logic. Thank you!

PR #4988, opened 2025-04-14
Revert "Make Paywalls v2 Text use verbatim"

joshdholtz:
Reverts RevenueCat/purchases-ios#4975

Using `verbatim` doesn't apply markdown. Also need to figure out why snapshot tests didn't fail with this one.


fire-at-will (approved):
:shipit:

PR #4992, opened 2025-04-15
fix: Unify finding active transaction for customer center

facumenzella:
### Motivation
We've noticed that we have two different logics to find the active transaction. 
Entitlements are optional, so we shouldn't rely on those to find it. 

### Description
The correct way was in CustomerCenterViewModel, so I've extracted that into an extension.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Extensions/CustomerInfo+ActiveTransaction.swift
+++ b/RevenueCatUI/CustomerCenter/Extensions/CustomerInfo+ActiveTransaction.swift
@@ -1,0 +1,43 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  CustomerInfo+ActiveTransaction.swift
+//
+//  Created by Facundo Menzella on 14/4/25.
+
+import RevenueCat
+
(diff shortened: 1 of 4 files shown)


RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -168,7 +168,7 @@ private extension CustomerCenterViewModel {
             return
         }
 
-        guard let activeTransaction = findActiveTransaction(customerInfo: customerInfo) else {
+        guard let activeTransaction = customerInfo.findActiveTransaction() else {
```
vegaro, line 171: we can remove the `findActiveTransaction` in this file then?

RevenueCatUI/CustomerCenter/Extensions/CustomerInfo+ActiveTransaction.swift
```diff
@@ -0,0 +1,43 @@
...
+//
+//  CustomerInfo+ActiveTransaction.swift
+//
+//  Created by Facundo Menzella on 14/4/25.
+
+import RevenueCat
+
+extension CustomerInfo {
+
+    func findActiveTransaction() -> Transaction? {
```
vegaro, line 18: I wonder if we can find a better name, since this is a very particular implementation for the customer center

vegaro (approved):
I was just wondering if we can add a test since this is fixing a bug. Not blocking this PR though 😄

facumenzella:
@vegaro I'll work on a test asap

PR #4993, opened 2025-04-15
fix: Reload customer center on subscription cancelled

facumenzella:
### Motivation
We're not reloading customer center when cancelling a subscription.

### Description
The approach consists of:
1. Open the management subscription screen using the SwiftUI approach
2. Use our preference keys to open the management screen at the correct moment
3. `syncPurchases` after closing the management screen

This updates the purchase information accordingly.

> Note that the UI still does not reflect the cancellation, because we're working on a whole UI redesign.


RevenueCatUI/CustomerCenter/Actions/CustomerCenterView+Actions.swift
```diff
@@ -346,6 +361,12 @@ extension View {
     ) -> some View {
         return self.modifier(CustomerCenterView.OnPromotionalOfferSuccess(handler: handler))
     }
+
+    func onSubscriptionCancelled(
```
vegaro, line 365: I don't think adding this is what we should be doing because we won't trigger this when the subscription is cancelled, but when the button is pressed. We don't have a good way for detecting the subscription was actually cancelled. 

We are already triggering `onShowingManageSubscription` when the `Cancel subscription` button or the `Change plans` is selected, and we also have the `onButtonSelected` modifier as well. So I believe this is redundant informatino. 

I think we should instead just reload when that action happens, without adding any new modifier.

tonidero: Hmm I've been thinking... did cancelling trigger a new CustomerInfo in the CustomerInfo listener? If so, maybe we would want to listen to changes in that listener and just reload whenever a new version is received (just don't remember if cancellations show there immediately)

vegaro: Hmm cancellations by themselves won't trigger the listener, but reloading the customer info itself could, if the backend has had enough time to detect the cancellation.

facumenzella: I've found https://developer.apple.com/documentation/storekit/transaction/updates

I guess that when the cancelled starts, we could listen to this for a couple of minutes?

facumenzella: We could do something like:

```swift
    func listenToTransactionUpdates() async {
        for await update in StoreKit.Transaction.updates {
            await loadScreen()
        }
    }
```

1. We set it up once the customer center is loaded
2. If someone enters the cancel flow, we add a "waiting for cancellation" label (or something better) for a couple of seconds

facumenzella: Although I am not sure this will work, I see there are a couple of options:
- https://developer.apple.com/documentation/storekit/transaction/latest(for:)
- https://developer.apple.com/documentation/storekit/transaction/all

Since we know the productIdentifier, we could use `latest(for:)` to see if there are any updates, and maybe subscribe to either `.all` or `.updates`

I am not completely sure it will work, but I can test it out and see what happens when cancelling

skinandbones:
As a SDK user, I had similar issues dealing with showing updated CustomerInfo immediately after cancelation.   I resorted to force fetching customer info when app becomes active.   Obviously not an optimal solution.

It would be very convenient if the SDK would handle watching for updates on the StoreKit transaction queue and update CustomerInfo as needed (I assume my client code would get the updated CustomerInfo via the Purchases delegate).  That same feature could also be used to resolve the issue in the customer center.  This is similar to what is being proposed above except I'm advocating you should resolve this issue more generally.

RevenueCatUI/CustomerCenter/Actions/CustomerCenterView+Actions.swift
facumenzella: In the end this didn't work. I've taken a complete different approach, relying on `syncPurchases` right after the management screen is closed

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -117,8 +122,12 @@ import RevenueCat
 
     #endif
 
-    func loadScreen() async {
+    func loadScreen(shouldSync: Bool = false) async {
         do {
+            if shouldSync {
+                try await self.purchasesProvider.syncPurchases()
```
vegaro, line 128: `syncPurchases` returns a `CustomerInfo` and `loadPurchaseInformation` as well. Maybe you could add a `CustomerInfo` parameter to `loadPurchaseInformation` to prevent loading `CustomerInfo` twice unnecessarily.

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -157,10 +157,24 @@ private extension CustomerCenterView {
...
+        .onCustomerCenterFeedbackSurveyCompleted { _ in
+            viewModel.manageSubscriptionsSheet = true
+        }
         .onCustomerCenterPromotionalOfferSuccess {
             Task {
-                await viewModel.loadScreen()
+                await viewModel.loadScreen(shouldSync: true)
+            }
+        }
+        .onCustomerCenterShowingManageSubscriptions {
```
vegaro, line 170: If developers set `onCustomerCenterShowingManageSubscriptions` themselves on `CustomerCenterView`, will they also get their modifer executed? Or does setting the modifier here "swallow " the modifier. I think that's what you are testing in `testNestedActionWrappers` right? but that's only testing `onCustomerCenterPromotionalOfferSuccess`

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
facumenzella: Just refactored it, so we do only one call

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
facumenzella: The test I've added covers that case 🚀

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -117,9 +122,13 @@ import RevenueCat
 
     #endif
 
-    func loadScreen() async {
+    func loadScreen(shouldSync: Bool = false) async {
         do {
-            try await self.loadPurchaseInformation()
+            let customerInfo = shouldSync ?
+            try await self.purchasesProvider.syncPurchases() :
```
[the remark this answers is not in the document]
facumenzella: blink twice

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -157,10 +157,24 @@ private extension CustomerCenterView {
                 }
             }
         }
+        .manageSubscriptionsSheet(isPresented: $viewModel.manageSubscriptionsSheet)
         .modifier(CustomerCenterActionViewModifier(actionWrapper: viewModel.actionWrapper))
+        .onCustomerCenterFeedbackSurveyCompleted { _ in
+            viewModel.manageSubscriptionsSheet = true
```
vegaro, line 163: why is this setting the `manageSubscriptionsSheet`? the modifier is for `onCustomerCenterFeedbackSurveyCompleted`

vegaro: Also we are never setting if back to `false`

facumenzella: > why is this setting the manageSubscriptionsSheet? the modifier is for onCustomerCenterFeedbackSurveyCompleted

This gets called when you tap the feedback survey, after cancelling. Before we were using `showManageSubscriptions`, but I've replaced that with the SwiftUI approach.

It gets back to false by SwiftUI.

facumenzella:
@skinandbones yes!! I did some tests too and nothing really worked.

With the latest implementation we can identify the cancellation by syncing purchases, right after the manage subscriptions sheet is dismissed.

After merging this, I'll tweak the UI and we'll probably add a preference key to be able to act on it

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -15,6 +15,7 @@
 
 import Foundation
 import RevenueCat
+import StoreKit
```
vegaro, line 18: is this needed?

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
vegaro: ohh I see. The thing is that any path can have a feedback survey completed, not only the cancel or change. For example, a custom URL could also have a feedback survey attached to it. This means the manage subscription sheet would be displayed after completing ANY feedback survey

facumenzella: As discussed, I've removed this. I don't remember why I added it, but it's definitely not needed to cover this 😅

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
facumenzella: Not anymore 👍

PR #4990, opened 2025-04-14
Fix Paywalls v2 Text component to not localized from bundle but support markdown

joshdholtz:
### Motivation

Fixes #4986

### Description

- Manually attempts to parse text with `AttributedString(markdown:)` and sets that into `Text`
  - This prevents attempting to look up localized strings from bundle
- Falls back to using `Text(verbatim:)`
  - If markdown failed to parse, probably more important to not localize

| Markdown | Invalid Markdown |
|--------|--------|
| Screenshot 2025-04-14 at 11 51 27 AM | Screenshot 2025-04-14 at 11 51 34 AM |

PR #4995, opened 2025-04-15
Fix paywalls text component compilation for iOS 15

joshdholtz:
### Motivation

Recent paywalls text component changes did not compile on iOS 15 with Xcode 14 (from #4990)

### Description

Added API and Swift version checks

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
@@ -93,17 +93,20 @@ private struct NonLocalizedMarkdownText: View {
         #if swift(>=5.7)
         Group {
             if let markdownText = self.markdownText {
+                // Use markdown if we can successfully parse it
                 Text(markdownText)
                     .font(self.font)
                     .fontWeight(self.fontWeight)
             } else {
+                // Otherwise use verbatim to prevent using localized bundle strings
                 Text(verbatim: self.text)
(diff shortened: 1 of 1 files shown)


joshdholtz:
@RCGitBot please test

RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
```diff
@@ -74,30 +74,50 @@ private struct NonLocalizedMarkdownText: View {
...
+        #if swift(>=5.7)
+        if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
+            return try? AttributedString(
+                // AttributedString allegedly uses CommonMark hard line breaks
+                // which is two or more spaces before line break
+                markdown: self.text.replacingOccurrences(of: "\n", with: "  \n")
+            )
+        } else {
+            return nil
+        }
```
JayShortway, line 86: Does this mean we don't support markdown on iOS 15?

JayShortway (approved):
Good stuff!

RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
joshdholtz: Oh good point... I made the fix for this in on spot but not the other 🤦‍♂️ Updating now!

PR #4997, opened 2025-04-15
Add `getStorefront` APIs

tonidero:
### Description
We already had a `Purchases.storeFrontCountryCode` property. This basically used StoreKit 1 to get the storefront.

This PR adds some new APIs to obtain the `storefront` asynchronously, which is the real nature of this data. We first try with StoreKit 2 if possible. If not, we fallback to StoreKit 1.

iOS equivalent of https://github.com/RevenueCat/purchases-android/pull/2361


Sources/Misc/Concurrency/Purchases+async.swift
```diff
@@ -22,6 +22,14 @@ extension Purchases {
...
     #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
 
+    func getStorefrontAsync() async -> Storefront? {
+        return await withUnsafeContinuation { continuation in
+            getStorefront { storefrontCountryCode in
+                continuation.resume(returning: storefrontCountryCode)
+            }
+        }
+    }
+
```
ajpallares, line 32: I believe we don't need this with the existing `func getStorefront() async -> Storefront?` method 🤔

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -26,6 +26,14 @@ public protocol PurchasesType: AnyObject {
      */
     var appUserID: String { get }
 
+    /**
+     * The storefront country code currently used by the Apple account. This is always provided by StoreKit 1.
+     * #### Related articles
+     * - ``Purchases/getStorefront(completion:)``
+     * - ``Purchases/getStorefront()``
+     */
+    var storeFrontCountryCode: String? { get }
```
ajpallares, line 35: The documentation of this in `Purchases` is 
```
    /// The three-letter code representing the country or region
    /// associated with the App Store storefront.
    /// - Note: This property uses the ISO 3166-1 Alpha-3 country code representation.
```
I don't know if we should unify them both as customers will only see one of the two.

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -49,6 +57,28 @@ public protocol PurchasesType: AnyObject {
 
     #if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
 
+    /**
+     * Obtain the storefront currently used by the Apple account. This will use StoreKit 2 first,
+     * and if not possible, fallback to StoreKit 1. It will be null if we can't obtain Apple's storefront.
```
ajpallares, line 62: Nit
```suggestion
     * and if not possible, fallback to StoreKit 1. It will be `nil` if we can't obtain Apple's storefront.
```

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -49,6 +57,28 @@ public protocol PurchasesType: AnyObject {
...
+     *
+     * #### Related Articles
+     * - ``Purchases/storeFrontCountryCode``
+     * - ``Purchases/getStorefront()``
+     */
+    func getStorefront(completion: @escaping GetStorefrontBlock)
+
+    /**
+     * Obtain the storefront currently used by the Apple account. This will use StoreKit 2 first,
+     * and if not possible, fallback to StoreKit 1. It will be null if we can't obtain Apple's storefront.
```
ajpallares, line 74: Same nit
```suggestion
     * and if not possible, fallback to StoreKit 1. It will be `nil` if we can't obtain Apple's storefront.
```

ajpallares (approved):
Looking good, thank you! Just some nits on documentation and one method that (I think) is not needed.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -40,6 +40,11 @@ public typealias PurchaseCompletedBlock = @MainActor @Sendable (StoreTransaction
                                                                 PublicError?,
                                                                 Bool) -> Void
 
+/**
+ Completion block for ``Purchases/getStorefrontCountryCode(``
+ */
+public typealias GetStorefrontBlock = @MainActor @Sendable (Storefront?) -> Void
```
ajpallares, line 46: This is not a big deal and probably better to leave like this for consistency with other similar types, but "block" is how the callback concept is named in Objective-C while "closure" is the Swift counterpart.
Usually, these are called `Completion`, so `GetStorefrontCompletion` would probably be a more iOS-y name.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -40,6 +40,11 @@ public typealias PurchaseCompletedBlock = @MainActor @Sendable (StoreTransaction
                                                                 PublicError?,
                                                                 Bool) -> Void
 
+/**
+ Completion block for ``Purchases/getStorefrontCountryCode(``
```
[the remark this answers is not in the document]
tonidero: Oh yes! That was from a previous iteration which I forgot to update. Thank you!!

Sources/Purchasing/Purchases/Purchases.swift
tonidero: That's a good point... But yeah, I mostly wanted to be consistent, which I think it's better for now. Happy to deprecate these typealiases in favor of new ones in the future though!

ajpallares: Yeah, we can think of better namings for a next major / API redesign. Besides, the closure typealias names are the least important, IMO

PR #5001, opened 2025-04-16
[Paywalls v2] Fixes decoding `TabControlType`

JayShortway:
## Bug
Decoding the `TabControlType` gave errors such as
```
Paywall V2 Error: [
    "componentsConfig": RevenueCat.PaywallComponentsData.EquatableError(
        description: "typeMismatch(
            Swift.Dictionary<Swift.String, Any>, 
            Swift.DecodingError.Context(
                codingPath: [
                    CodingKeys(stringValue: \"offerings\", intValue: nil), 
                    _CodingKey(stringValue: \"Index 45\", intValue: 45), 
                    CodingKeys(stringValue: \"paywallComponents\", intValue: nil), 
                    CodingKeys(stringValue: \"componentsConfig\", intValue: nil), 
                    CodingKeys(stringValue: \"base\", intValue: nil), 
                    CodingKeys(stringValue: \"stack\", intValue: nil), 
                    CodingKeys(stringValue: \"components\", intValue: nil), 
                    _CodingKey(stringValue: \"Index 2\", intValue: 2), 
                    CodingKeys(stringValue: \"control\", intValue: nil), 
                    CodingKeys(stringValue: \"type\", intValue: nil)
                ], 
                debugDescription: \"Expected to decode Dictionary<String, Any> but found a string instead.\", 
                underlyingError: nil
            )
        )"
    )
]
```

## Fix
This PR fixes that by manually decoding the `TabControlType` to a string.

### DIFF
--- a/Sources/Paywalls/Components/PaywallTabsComponent.swift
+++ b/Sources/Paywalls/Components/PaywallTabsComponent.swift
@@ -121,6 +121,28 @@ public extension PaywallComponent {
             public enum TabControlType: Codable, Sendable, Hashable, Equatable {
                 case buttons
                 case toggle
+
+                public init(from decoder: Decoder) throws {
+                    let container = try decoder.singleValueContainer()
+                    let value = try container.decode(String.self)
+
+                    switch value {
+                    case "buttons": self = .buttons
+                    case "toggle": self = .toggle
+                    default: throw DecodingError.dataCorruptedError(
+                        in: container,
+                        debugDescription: "Invalid TabControlType: \(value)"
+                    )
+                    }
+                }
+
+                public func encode(to encoder: Encoder) throws {
+                    var container = encoder.singleValueContainer()
+                    switch self {
+                    case .buttons: try container.encode("buttons")
+                    case .toggle: try container.encode("toggle")
+                    }
+                }
             }
 
             public let type: TabControlType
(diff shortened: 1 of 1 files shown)


ajpallares (approved):
Nice catch! My first thought was to add `String` conformance to the `TabControlType` enum. I think that would work as well. However, since it's a public enum, that would be an API change (breaking perhaps?). WDYT? 
Other than that, this seems fine to me 👍

JayShortway:
@ajpallares Actually, that's even better! Thanks for the suggestion, I always forget Swift enums can do that haha. https://github.com/RevenueCat/purchases-ios/pull/5001/commits/f6add67a17b145f85979ffba11acd500d71f4ab5

I think it's fine. This type is not _really_ meant to be public, and the Tabs component hasn't been publicly released yet either, so no one will have a Tabs component in their Paywall's JSON. As I understand it, this will mostly _add_ to the public API, e.g. the `rawValue` field. Also, there's precedent for using String conformance for Paywall enums, such as `HorizontalAlignment` (and many more). 

(@joshdholtz feel free to disagree, and I'll revert to my earlier implementation.)

ajpallares:
> This type is not really meant to be public

Why is it public then? (My mind always goes to `@_spi` 😄)

JayShortway:
I think it should ideally use `@_spi` indeed! All of these JSON models should, imo. @joshdholtz what do you think?

JayShortway:
Maybe we can make 'em all `@_spi` (in the next major)?

CreativeStudio16 (commented):
Excellent ⭐️⭐️⭐️⭐️⭐️

CreativeStudio16 (commented):
Renewed

Issue #4954, opened 2025-04-02
RevenueCat Offerings Not Loading on iOS 18.4


cem256:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [x] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
Offerings are not loading on iOS 18.4. They work fine on earlier versions like 18.1, 18.2, and 18.3.1. This is happening in all of my apps that use RevenueCat. Nothing has changed in my setup, and everything still works perfectly on prior iOS versions.

I tested on real devices and the simulator using a sandbox account. I also created a brand new test app and a new product in both App Store Connect and RevenueCat just to make sure the issue is not project related, but offerings still fail to load on 18.4. Since app reviewers are also on iOS 18.4, this is blocking app updates and affecting users on that version who can’t see any paywall content.

I’ve checked everything in [this]( https://community.revenuecat.com/sdks-51/why-are-offerings-or-products-empty-124) guide multiple times. Everything is set up exactly as described there, and again everything works fine on prior iOS versions.

1. Environment
   1. Platform: iOS
   2. SDK version: 5.18.0, 5.20.2
   3. StoreKit version:
      - [ ] StoreKit 1 (default on versions <5.0.0. Can be enabled in versions >=5.0.0 with `.with(storeKitVersion: .storeKit1)`)
      - [x] StoreKit 2 (default on versions >=5.0.0)
   4. OS version: MacOS 15.4, iOS 18.4
   5. Xcode version: 16.3
   6. Device and/or simulator:
      - [x] Device
      - [x] Simulator
   7. Environment:
      - [x] Sandbox
      - [x] TestFlight
      - [x] Production
   8. How widespread is the issue. Percentage of devices affected: Users in iOS 18.4 affected by this

2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue. 

I have anonymized identifiers
```
DEBUG: ℹ️ Configuring SDK using RevenueCat's UserDefaults suite.
DEBUG: 👤 Identifying App User ID
DEBUG: ℹ️ Debug logging enabled
[...]

PR #5002, opened 2025-04-16
Log a warning to iOS 18.4 Simulator users with empty offerings

cperriam-rc:
### Motivation
 - The iOS 18.4 Simulator has a known issue which prevents RevenueCat Offerings from loading
 - Relates to #4954 

### Description
 - This change provides affected developers with a helpful message and a link to the GitHub Issue
 - The change has been manually tested with the iOS 18.4 Simulator


Sources/Logging/Strings/OfferingStrings.swift
```diff
@@ -155,6 +156,12 @@ extension OfferingStrings: LogMessage {
 
         case let .overriding_package(old, new):
             return "Package: \(old) already exists, overwriting with: \(new)"
+
+        case .known_issue_ios_18_4_simulator:
+            return "None of the products registered in the RevenueCat dashboard could be fetched from App Store " +
+            "Connect (or the StoreKit Configuration file if one is being used)." +
+            "\nThis issue is widely reported by iOS 18.4 simulator users. Try using a different simulator version." +
+            "\nMore information: https://github.com/RevenueCat/purchases-ios/issues/4954"
```
aboedo, line 164: might be a good idea to put a rev.cat link (we have credentials in 1pass for the link shortener) so that if we have a new, better place to point to in the future we can change that dynamically in the future

Sources/Misc/SystemInfo.swift
```diff
@@ -218,6 +218,22 @@ class SystemInfo {
...
+    /// Checks for exposure to https://github.com/RevenueCat/purchases-ios/issues/4954
+    func isSubjectToKnownIssue_18_4_sim() -> Bool {
+        let firstOSVersionWithBug = OperatingSystemVersion(majorVersion: 18,
+                                                           minorVersion: 4,
+                                                           patchVersion: 0)
+
+        // Conservative estimate. No iOS fix version currently known (as at 2025-04-15).
+        let firstOSVersionWithFix = OperatingSystemVersion(majorVersion: 19,
+                                                           minorVersion: 0,
+                                                           patchVersion: 0)
```
aboedo, line 230: let's keep the cap at 18.5, which I realize is optimistic, but then again we wanna be conservative with how much we show this

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -236,7 +236,9 @@ private extension OfferingsManager {
             let products = result.value ?? []
 
             guard products.isEmpty == false else {
-                completion(.failure(Self.createErrorForEmptyResult(result.error)))
+                let isSubjectToSimIssue = self.systemInfo.isSubjectToKnownIssue_18_4_sim()
```
aboedo, line 239: as a maintainer, just looking at this I wouldn't have a way of knowing what "sim issue" this is referring to other than "a thing that happened in 18.4". 

Let's add a comment here explaining _why_. Doesn't have to be long, but it'd be helpful to read something like "There is a bug in simulators in iOS 18.4 where HTTP requests don't always finish" or something

aboedo (commented):
looking great so far! Left a few comments

Sources/Logging/Strings/OfferingStrings.swift
```diff
@@ -43,6 +43,7 @@ enum OfferingStrings {
     case unknown_package_type(Package)
     case custom_package_type(Package)
     case overriding_package(old: String, new: String)
+    case known_issue_ios_18_4_simulator
```
fire-at-will, line 46: Nit: I'd name this something like `known_issue_empty_products_ios_18_4_simulator` on the off chance that we discover more issues with the iOS 18.4 sim 😅

Sources/Logging/Strings/OfferingStrings.swift
```diff
@@ -155,6 +156,12 @@ extension OfferingStrings: LogMessage {
 
         case let .overriding_package(old, new):
             return "Package: \(old) already exists, overwriting with: \(new)"
+
+        case .known_issue_ios_18_4_simulator:
+            return "None of the products registered in the RevenueCat dashboard could be fetched from App Store " +
+            "Connect (or the StoreKit Configuration file if one is being used)." +
+            "\nThis issue is widely reported by iOS 18.4 simulator users. Try using a different simulator version." +
```
fire-at-will, line 163:
```suggestion
            "\nThis issue is widely reported by iOS 18.4 simulator users. Try using a different iOS version with your simulator." +
```

Nit: Technically the issue depends on the OS version in the simulator, not the simulator version. For example, the bug presents in Xcode 16.3 simulators running iOS 18.4, but not in Xcode 16.3 simulators running iOS 18.3

Sources/Misc/SystemInfo.swift
cperriam-rc: Sounds good. Changed in 7f09f8520a63626f2238f9759f3291fc2c036656 to show this for versions up to (but not including) 18.5.0.

Sources/Purchasing/OfferingsManager.swift
cperriam-rc: Added inline comments in 7f09f8520a63626f2238f9759f3291fc2c036656 to give context to the issue.

I also changed the naming from `isSubjectToSimIssue` to `showSimulatorWarning` with the aim of conveying what we're using this information for. I attempted naming similar to @fire-at-will's suggestion, but didn't feel the added line breaks (due to 120 char width limit) were worth it.

Sources/Logging/Strings/OfferingStrings.swift
cperriam-rc: Good call. I adopted this in 7f09f8520a63626f2238f9759f3291fc2c036656, but used the naming of `known_issue_ios_18_4_simulator_empty_products` so that any other known issues with the iOS18.4 sim can share the same prefix.

Sources/Logging/Strings/OfferingStrings.swift
cperriam-rc: Good point, resolved in 7f09f8520a63626f2238f9759f3291fc2c036656

Sources/Misc/SystemInfo.swift
```diff
@@ -218,6 +218,22 @@ class SystemInfo {
         return ProcessInfo.processInfo.isOperatingSystemAtLeast(version)
     }
 
+    /// Checks for exposure to https://github.com/RevenueCat/purchases-ios/issues/4954
+    func isSubjectToKnownIssue_18_4_sim() -> Bool {
```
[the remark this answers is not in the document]
cperriam-rc: I considered this, but I don't believe it's possible to mock `targetEnvironment(simulator)`.

ajpallares (approved):
Yeah, looks good! 👌

PR #5005, opened 2025-04-18
[Diagnostics] Add extra fields when tracking AppTransaction errors

MarkVillacampa:


Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -307,11 +307,22 @@ extension StoreKit2TransactionFetcher {
                     return nil
                 }
             } catch {
-                self.diagnosticsTracker?.trackAppleAppTransactionError(errorMessage: error.localizedDescription)
-                Logger.warn(Strings.storeKit.sk2_error_fetching_app_transaction(error))
+                trackAppleAppTransactionError(error)
```
tonidero, line 310: Should we add a `self` here?
```suggestion
                self.trackAppleAppTransactionError(error)
```

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
```diff
@@ -307,11 +307,22 @@ extension StoreKit2TransactionFetcher {
...
                 return nil
             }
         }
     }
 
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    private func trackAppleAppTransactionError(_ error: Error) {
```
tonidero, line 317: I wonder if it would be helpful to know from which of the 2 callers the error happened... but can also be added later if needed

tonidero (approved):
Looks great! 🙌

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
MarkVillacampa: Added explicit selfs 👍

Sources/Purchasing/StoreKit2/StoreKit2TransactionFetcher.swift
MarkVillacampa: hmmm not super useful I think, we just care about the overall amount and types of errors returned by AppTransaction.shared

PR #5008, opened 2025-04-18
Maybe fix some flaky PaywallViewEventsTests in CI

joshdholtz:
### Motivation

Fix flaky tests in `PaywallViewEventsTests`

### Description

- Remove randomness with scheme
- Wait after adding to heirarchy
- Wait more for close event

### DIFF
--- a/Tests/RevenueCatUITests/PaywallViewEventsTests.swift
+++ b/Tests/RevenueCatUITests/PaywallViewEventsTests.swift
@@ -27,7 +27,7 @@ class PaywallViewEventsTests: TestCase {
     private var handler: PurchaseHandler!
 
     private let mode: PaywallViewMode = .random
-    private let scheme: ColorScheme = Bool.random() ? .dark : .light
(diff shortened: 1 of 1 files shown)


Tests/RevenueCatUITests/PaywallViewEventsTests.swift
```diff
@@ -27,7 +27,7 @@ class PaywallViewEventsTests: TestCase {
     private var handler: PurchaseHandler!
 
     private let mode: PaywallViewMode = .random
-    private let scheme: ColorScheme = Bool.random() ? .dark : .light
+    private let scheme: ColorScheme = .light
```
fire-at-will, line 30: Nit: instead of only running the test for the light ColorScheme, should we instead create two tests, one for each color scheme?

joshdholtz: ON IT

fire-at-will (approved):

PR #5011, opened 2025-04-22
[Customer center] missing inject of Purchases provider

ajpallares:
### Description

Injects the `CustomerCenterPurchasesType` into `RestorePurchasesAlertViewModel`, which was still missing after #4957


facumenzella (approved):
I'll set up a slack message so we can at least get notified when this happens :)

PR #5015, opened 2025-04-23
Add Virtual Currencies to Customer Center

fire-at-will:
### DO NOT MERGE
Even once this PR is approved, it should not be merged until the corresponding backend/frontend work is completed.

### Description
This PR adds support for displaying virtual currency balances to the Customer Center.

Upon loading the Customer Center, if it is configured to display virtual currencies, a list of the user's current virtual currency balances will appear on the home screen. If the user has more than 3 VCs, the three with the highest balances will be displayed, and a "See all in-app currencies" button will appear that takes the user to a screen that lists all of their balances.

Currently, we're displaying the VC's code with the balances - in the future we'll change this to use the virtual currency's name and potentially icon as well.

#### ManageSubscriptionsView
This is the "home" screen that users see when they have an in-app purchase.
Manage Subscriptions

#### VirtualCurrencyBalancesScreen
This is the screen that users see when they tap the "See all in-app currencies" button.
Virtual Currency Balance Screen

#### NoSubscriptionsView
This is the "home" screen that users see if they have no in-app purchases. We've added the virtual currencies section here since it's possible for a developer to grant virtual currencies to users with no purchases.
No Subs View

### New Localized Strings
This PR introduces several new localized strings:
| Enum Value                     | Key                              | Purpose                                                                              | Default Value                                                |
|--------------------------------|----------------------------------|--------------------------------------------------------------------------------------|--------------------------------------------------------------|
| `seeAllVirtualCurrencies`        | `see_all_virtual_currencies`       | Button that you tap to view all VC balances                                          | See all in-app currencies                                    |
| `virtualCurrencyBalancesScreenHeader`  | `virtual_currency_balances_screen_header` | The title that appears in the VC section header and on the VC balances screen        | In-App Currencies                                            |
| `noVirtualCurrencyBalancesFound` | `no_virtual_currency_balances_found`    | Displayed on the VC balances screen if no virtual currencies are found for the user. | It doesn't look like you've purchased any in-app currencies. |

These strings will need to be added to the backend and the dashboard in follow-up PRs.

### Next Steps
Upon approval of this PR, we won't immediately merge this into the `virtual-currency-dev` branch, as there's some backend work we also need to do. Specifically, we need to:
- Add the new localized strings to the backend/frontend
- Add a new checkbox to the dashboard to allow developers to decide whether or not they want to display virtual currencies in the Customer Center.
Once those items are complete, we can merge this into the `virtual-currency-dev` branch and make a second beta release.

Later on, we'll also update this to use the virtual currency's name and icon instead of the virtual currency's code.


Package.swift
```diff
@@ -56,6 +56,7 @@ if shouldIncludeDocCPlugin {
 // See https://github.com/RevenueCat/purchases-ios/pull/2989
 // #if os(visionOS) can't really be used in Xcode 13, so we use this instead.
 let visionOSSetting: SwiftSetting = .define("VISION_OS", .when(platforms: [.visionOS]))
+let virtualCurrenciesSetting: SwiftSetting = .define("ENABLE_VIRTUAL_CURRENCIES")
```
fire-at-will, line 59: We needed to enable the `ENABLE_VIRTUAL_CURRENCIES` compiler flag to access the `VirtualCurrencyInfo` struct. This will be removed when we eventually merge into main

RevenueCat.podspec
```diff
@@ -24,6 +24,7 @@ Pod::Spec.new do |s|
   
   s.pod_target_xcconfig = {
     'DEFINES_MODULE' => 'YES',
+    'SWIFT_ACTIVE_COMPILATION_CONDITIONS' => 'ENABLE_VIRTUAL_CURRENCIES',
```
fire-at-will, line 27: We needed to enable the `ENABLE_VIRTUAL_CURRENCIES` compiler flag to access the `VirtualCurrencyInfo` struct. This will be removed when we eventually merge into main

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrenciesListSection.swift
```diff
@@ -0,0 +1,120 @@
...
+        let sortedCurrencies = virtualCurrencies.map {
+            VirtualCurrencyBalanceListRow.RowData(
+                virtualCurrencyCode: $0.key,
+                balance: $0.value.balance
+            )
+        }
+            .sorted(by: { $0.balance > $1.balance })
+
+        self.virtualCurrencies = Array(sortedCurrencies.prefix(3))
+        self.displayShowAllButton = sortedCurrencies.count > 3
```
ajpallares, line 56: Perhaps these `3` should be a constant (which also helps document what it means)

RevenueCatUI/CustomerCenter/ViewModels/VirtualCurrencyBalancesScreenViewModel.swift
```diff
@@ -0,0 +1,111 @@
...
+final class VirtualCurrencyBalancesScreenViewModel: ObservableObject {
+
+    /// The current state of the virtual currencies screen.
+    @Published var viewState: VirtualCurrencyBalancesScreen.ViewState
+
+    /// A flag indicating whether the view model is running in a SwiftUI preview.
+    ///
+    /// When true, data loading is skipped to prevent unnecessary network calls
+    /// during preview rendering.
+    private let isRunningInSwiftUIPreview: Bool
```
facumenzella, line 36: I would rather just inject a mock for `CustomerCenterPurchasesType` than adding this one, or just injecting a function that can easily be mocked 🤓

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptions/ManageSubscriptionsViewModel.swift
```diff
@@ -41,6 +41,8 @@ final class ManageSubscriptionsViewModel: ObservableObject {
     @Published
     var showPurchases: Bool = false
 
+    @Published var showBalances: Bool = false
```
facumenzella, line 44: Is this being used? Can't find where 🤔

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptions/ManageSubscriptionsViewModel.swift
```diff
@@ -53,6 +55,9 @@ final class ManageSubscriptionsViewModel: ObservableObject {
     @Published
     var inAppBrowserURL: IdentifiableURL?
 
+    @Published
```
facumenzella, line 58: nit: can we skip this `Published`? It's not getting reloaded anywhere

facumenzella (approved):
Looks good 🚀

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrenciesListSection.swift
```diff
@@ -0,0 +1,120 @@
...
+        virtualCurrencies: [String: RevenueCat.VirtualCurrencyInfo],
+        purchasesProvider: CustomerCenterPurchasesType
+    ) {
+        let sortedCurrencies = virtualCurrencies.map {
+            VirtualCurrencyBalanceListRow.RowData(
+                virtualCurrencyCode: $0.key,
+                balance: $0.value.balance
+            )
+        }
+            .sorted(by: { $0.balance > $1.balance })
```
ajpallares, line 53: I'm not sure about this sorting. I mean, it makes sense for now, but I could see some customers wanting their own sorting order logic. Is this something that would make sense to have defined in the dashboard? (I'm thinking, for example, there's 1 VC that is much more valuable than any amount of any other VC)

ajpallares (commented):
Great job! It's looking great! I have some comments, mostly nits, some doubts and some suggestions.

RevenueCatUI/CustomerCenter/ViewModels/VirtualCurrencyBalancesScreenViewModel.swift
fire-at-will: Yeah, having a `isRunningInSwiftUIPreview` param isn't ideal. I didn't go with the mock route because `MockCustomerCenterPurchases` is only available in the `RevenueCatUITestsDev` target, not in the `RevenueCatUI` target. I guess we could move it into the `RevenueCatUI` target and keep it internal - do you have strong opinions either way?

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptions/ManageSubscriptionsViewModel.swift
fire-at-will: It isn't! This is leftover from an earlier approach during development. I've removed it in 039034f3f9b5f670c4a4dcec1e10f6f220ba623c. Good catch!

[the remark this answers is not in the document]
fire-at-will: It isn't! This is leftover from an earlier approach during development. I've removed it in 039034f3f9b5f670c4a4dcec1e10f6f220ba623c. Good catch!

RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptions/ManageSubscriptionsViewModel.swift
fire-at-will: Yes! Removed in b82e80fd04f8ba240d2e72a8385286116c7bd42f

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrencyBalancesScreen.swift
```diff
@@ -0,0 +1,219 @@
...
+        case (.loading, .loading):
+            return true
+        case (.error, .error):
+            return true
+        case let (.loaded(lhsData), .loaded(rhsData)):
+            return lhsData == rhsData
+        default:
+            return false
+        }
+    }
```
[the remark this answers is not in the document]
fire-at-will: Streamlined it in a84f38abc91251251f07e083e853116348c96780 - that's much better :)

At some point in development, the compiler was forcing me to write a custom equatable conformance, but this is much better :)

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrencyBalanceListRow.swift
```diff
@@ -0,0 +1,85 @@
...
+    ///
+    /// This struct contains the necessary information to display a virtual currency balance,
+    /// including the currency code and the current balance amount.
+    struct RowData: Identifiable, Hashable {
+        /// A unique identifier for the row.
+        let id = UUID()
+        /// The code representing the virtual currency (e.g., "GLD" for gold).
+        let virtualCurrencyCode: String
+        /// The current balance of the virtual currency.
+        let balance: Int
```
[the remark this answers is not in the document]
fire-at-will: Great question! The decision was made at the product level to only support integer values for VC balances. The balances are also represented as integers in the database :)

RevenueCatUI/CustomerCenter/ViewModels/VirtualCurrencyBalancesScreenViewModel.swift
```diff
@@ -0,0 +1,111 @@
...
+
+    init(
+        viewState: VirtualCurrencyBalancesScreen.ViewState = .loading,
+        purchasesProvider: CustomerCenterPurchasesType,
+        isRunningInSwiftUIPreview: Bool = false
```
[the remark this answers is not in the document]
fire-at-will: Good idea! I've implemented the change in 98af2c0d883a58ea7b24f1da8b8df67b10fcd13b, and even put the logic behind an `#if DEBUG` so it'll always be false in release builds.

RevenueCatUI/CustomerCenter/ViewModels/VirtualCurrencyBalancesScreenViewModel.swift
fire-at-will: @ajpallares Had a good idea to compute `isRunningInSwiftUIPreview ` at runtime using `ProcessInfo.processInfo.environment["XCODE_RUNNING_FOR_PREVIEWS"]`, which allows us to remove `isRunningInSwiftUIPreview` from the initializer 😄. I've implemented that in 98af2c0d883a58ea7b24f1da8b8df67b10fcd13b

RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -1125,9 +1125,14 @@
 		FD2E6C9F2C480FF000CB4BD7 /* OHHTTPStubs in Frameworks */ = {isa = PBXBuildFile; productRef = FD2E6C9E2C480FF000CB4BD7 /* OHHTTPStubs */; };
 		FD2E6CA12C48100900CB4BD7 /* OHHTTPStubsSwift in Frameworks */ = {isa = PBXBuildFile; productRef = FD2E6CA02C48100900CB4BD7 /* OHHTTPStubsSwift */; };
 		FD33CD4D2D034CBD000D13A4 /* CustomerCenterViewController.swift in Sources */ = {isa = PBXBuildFile; fileRef = FD33CD4C2D034CBD000D13A4 /* CustomerCenterViewController.swift */; };
+		FD3FB00F2DB6984800D4AEC7 /* VirtualCurrenciesScreenViewModel.swift in Sources */ = {isa = PBXBuildFile; fileRef = FD3FB00E2DB6984800D4AEC7 /* VirtualCurrenciesScreenViewModel.swift */; };
```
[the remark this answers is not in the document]
fire-at-will: Hmm. It looks like Xcode didn't rename this to `VirtualCurrencyBalancesScreenViewModel` when I renamed it 🤦‍♂️

Good catch, will fix!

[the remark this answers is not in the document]
fire-at-will: Fixed in 25f3a803c7fac55dd3486664a6543030bbfcd48f 👍

RevenueCatUI/CustomerCenter/ViewModels/VirtualCurrencyBalancesScreenViewModel.swift
```diff
@@ -0,0 +1,111 @@
...
+
+        await self.loadData()
+    }
+}
+
+@available(iOS 15.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+internal extension VirtualCurrencyBalancesScreenViewModel {
```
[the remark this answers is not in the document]
fire-at-will: Done in 2dd8a673e00b48112320e58ad90b0223d57ea052 👍

RevenueCatUI/CustomerCenter/ViewModels/VirtualCurrencyBalancesScreenViewModel.swift
```diff
@@ -0,0 +1,111 @@
...
+
+    /// Loads the virtual currency data from the RevenueCat SDK.
+    ///
+    /// This method updates the view state to reflect the loading process and handles
+    /// any errors that occur during data fetching. On success, it transforms the
+    /// customer information into a format suitable for display in the UI.
+    private func loadData() async {
+        await MainActor.run {
```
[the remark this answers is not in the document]
fire-at-will: Good idea! Done in a4c4becbe83ead49b1a49d5cc6d417dcc4b46941

RevenueCatUI/CustomerCenter/ViewModels/VirtualCurrencyBalancesScreenViewModel.swift
```diff
@@ -0,0 +1,111 @@
...
+        from customerInfo: RevenueCat.CustomerInfo) -> [VirtualCurrencyBalanceListRow.RowData] {
+            return customerInfo.virtualCurrencies
+                .map {
+                    VirtualCurrencyBalanceListRow.RowData(
+                        virtualCurrencyCode: $0.key,
+                        balance: $0.value.balance
+                    )
+                }
+                .sorted { $0.balance > $1.balance }
+    }
```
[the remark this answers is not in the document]
fire-at-will: Yeah, that is wonky. Fixed in 4fdf156643401ab211acae49d6d8cca627fded81!

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrencyBalancesScreen.swift
```diff
@@ -0,0 +1,219 @@
...
+            }
+
+        }
+        .animation(.default, value: viewModel.viewState)
+        .navigationTitle(localization[.virtualCurrencyBalancesScreenHeader])
+        .onAppear {
+            Task {
+                await self.viewModel.onAppear()
+            }
+        }
```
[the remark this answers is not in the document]
fire-at-will: Nice! I've made this change in 5ba1ace55fef3dcce49d6224f8ffea608e78c926

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrenciesListSection.swift
fire-at-will: I think that's a great idea! It feels a little out of scope for now, but it's definitely something that we could add if we get customer requests for additional sorting methods!

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrencyBalancesScreen.swift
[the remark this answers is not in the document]
fire-at-will: > Are we sure we want to refresh the VCs when showing this screen instead of taking them from the existing CustomerInfo?

My thinking was that yes, we want to refresh them, just in case the balances have changed for any reason while the customer center was loaded.

> If so, would it make sense to add pull to refresh?
I think that's a good idea! I'll consider implementing it in a follow-up PR since this one is already a little big

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrenciesListSection.swift
fire-at-will: This is a _fantastic_ suggestion! I'll work on getting it implemented.

fire-at-will: Implemented in 4092d7a1cf7a8e2b01bf5c4c2827feda3090c1d8 💪

Great idea!

fire-at-will:
@ajpallares This is ready for a second look! I implemented most of your feedback. :)

The biggest difference is that I implemented your suggestion that on the VC balances list:
- If there are 4 VCs or less, show them all
- If there are 5 VCs or more, show the first 3 VCs + the See all VCs button

Thanks! :D

ajpallares (approved):
Thanks for making the changes! I think this is good to go! 🚀

RevenueCatUI/CustomerCenter/ViewModels/VirtualCurrencyBalancesScreenViewModel.swift
facumenzella: I like this :)

facumenzella (approved):

PR #5016, opened 2025-04-24
[Paywalls v2] fixes crash when getting invalid URL from paywall components localization

ajpallares:
This code would produce a runtime crash when getting a `nil` url

```
func urlFromLid(_ urlLid: String) throws -> URL {
    let urlString = try string(key: urlLid)
    let url = URL(string: urlString)
    if url == nil {
        Logger.error(Strings.paywall_invalid_url(urlLid))
    }
    return url! // <-- Crash here
}
```


JayShortway (approved):

PR #5018, opened 2025-04-24
Add load shedder integration tests for consumable and non-consumable products

tonidero:
### Description
This adds some additional load shedder integration tests for consumables and non-consumables, since support for those was recently added.


ajpallares (approved):
Good one! 🚀

PR #5021, opened 2025-04-25
fix compilation issues in older versions of Xcode

ajpallares:
### Description

The builds are failing on CI when using Xcode 14. This is due to some code recently added that uses some newer features of Swift that are not available in Xcode 14.


ajpallares:
@RCGitBot please test

polpielladev (approved):
Great catch!

PR #5022, opened 2025-04-25
Fix Backwards compatibility errors introduced by new `testSDKHealthCheck` implementation

polpielladev:
We recently introduced breaking changes to the `PurchasesDiagnostics` API through the `testSDKHealthCheck` method. We have decided to instead make the changes backwards compatible and instead deprecate the existing method. 

### Checklist

### DIFF
--- a/Sources/Support/PurchasesDiagnostics.swift
+++ b/Sources/Support/PurchasesDiagnostics.swift
@@ -201,11 +201,11 @@ extension PurchasesDiagnostics {
         case unhealthy(PurchasesDiagnostics.SDKHealthError)
     }
 
+    /// Perform tests to ensure SDK is configured correctly.
+    /// - `Throws`: ``PurchasesDiagnostics/Error`` if any step fails
     @available(*, deprecated, message: """
     Use the `PurchasesDiagnostics.shared.checkSDKHealth()` method instead.
     """)
-    /// Perform tests to ensure SDK is configured correctly.
-    /// - `Throws`: ``PurchasesDiagnostics/Error`` if any step fails
     @objc(testSDKHealthWithCompletion:)
     public func testSDKHealth() async throws {
         do {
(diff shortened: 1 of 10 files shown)


Sources/Support/PurchasesDiagnostics.swift
```diff
@@ -187,17 +196,39 @@ extension PurchasesDiagnostics {
...
+            // Catch every other error to ensure that we only throw `Error`s from here.
+            throw Error.unknown(error)
+        }
     }
 
     #if DEBUG
     /// Performs a full SDK configuration health check and throws an error if the configuration is not valid.
     /// - Important: This method is intended solely for debugging configuration issues with the SDK implementation.
     /// It should not be invoked in production builds.
     /// - Throws: The specific configuration issue that needs to be solved.
```
ajpallares, line 230: I would keep this method documentation in the existing (now deprecated) `testSDKHealth` method. Mostly because that method can still be called outside of `DEBUG` builds. 
And I would simplify the documentation of the new `checkSDKHealth()` method. E.g., "It should not be invoked in production builds." could be removed as now supposedly "it cannot be invoked in production builds". WDYT?

Sources/Support/PurchasesDiagnostics.swift
```diff
@@ -132,6 +132,16 @@ extension PurchasesDiagnostics {
         /// Failure performing a signed request
         case failedMakingSignedRequest(Swift.Error)
 
+        /// Any other not identifier error. You can check the undelying error for details.
+        case unknown(Swift.Error)
+
+    }
+
+    /// An error that represents a problem in the SDK's configuration
+    public enum SDKHealthError: Swift.Error {
```
ajpallares, line 141: Could this type be added under the `DEBUG` compiler flag?

ajpallares (commented):
Thanks for making these changes so fast! 🚀 
This looks great, really. I have two small comments + since we're adding new public APIs, perhaps it would be good to write new API tests to cover these in the future.

Sources/Support/PurchasesDiagnostics.swift
polpielladev: I decided to just revert to what the documentation looked like before my changes, but I 100% agree, it's an improvement if we add this. Let me change this 👍

polpielladev: Updated!

Sources/Support/PurchasesDiagnostics.swift
polpielladev: I think so, I will move all of the health report types and entities under the `DEBUG` compiler flag

polpielladev: Updated!

Sources/Support/PurchasesDiagnostics.swift
```diff
@@ -187,17 +173,68 @@ extension PurchasesDiagnostics {
...
+        } catch let error {
+            // Catch every other error to ensure that we only throw `Error`s from here.
+            throw Error.unknown(error)
+        }
+    }
+
+    #if DEBUG
+    /// Performs a full SDK configuration health check and throws an error if the configuration is not valid.
+    /// - Important: This method can not be invoked in production builds.
+    /// - Throws: The specific configuration issue that needs to be solved.
```
ajpallares, line 236: Nit here. We should perhaps indicate the type of the error thrown
```suggestion
    /// - Throws: ``SDKHealthError`` indicating the specific configuration issue that needs to be solved.
```

ajpallares (approved):
Thanks for making these changes! Looks great now, only one suggestion

PR #5023, opened 2025-04-29
fix: Wrap viewmodel binding into another binding

facumenzella:
### Motivation
We're getting this warning:

```
\CustomerCenterViewModel.manageSubscriptionsSheet is isolated to the main actor. Accessing it via Binding from a different actor will cause undefined behaviors, and potential data races; This warning will become a runtime crash in a future version of SwiftUI.
Publishing changes from background threads is not allowed; make sure to publish values from the main thread (via operators like receive(on:)) on model updates.
Publishing changes from background threads is not allowed; make sure to publish values from the main thread (via operators like receive(on:)) on model updates.
```

### Description
I missed that the SwiftUI extension is marked as `non isolated`, and it's causing this warning when displaying the manage subscriptions view.  This wraps the binding that comes from the `Published` property, into another binding that dispatches the setter to the main queue.

#### Thoughts
I am thinking that most probably `CustomerCenterViewModel` shouldn't be wrapped as `MainActor`. However, I think this wrapper can't be avoided, because the `Published` property would produce a different warning.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
@@ -157,7 +157,12 @@ private extension CustomerCenterView {
                 }
             }
         }
-        .manageSubscriptionsSheet(isPresented: $viewModel.manageSubscriptionsSheet)
+        // This is needed because `CustomerCenterViewModel` is isolated to @MainActor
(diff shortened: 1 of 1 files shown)


facumenzella:
@ajpallares as discussed in private, this seems to be enough to remove the warning and keep the same functionality, so we're ok 🚀

PR #5024, opened 2025-04-29
[Paywalls v2] Fixes blank lines not showing up

JayShortway:
Quick follow-up to #4990 which seems to have resulted in ignoring blank lines.

## Bug
Blank lines are ignored.

## Cause
`AttributedString` seems to ignore `\n\n`, or even `  \n  \n`. 

## Fix
Using `InterpretedSyntax.inlineOnly`. I tried `InterpretedSyntax.inlineOnlyPreservingWhitespace` as well, and both seem to fix it. I went with the former as that seemed closer to what we had, but would be totally fine with the latter as well.


ajpallares (approved):
Thanks for the fix! Looks good to me

PR #5031, opened 2025-05-01
Add support for cross product promotional offers

vegaro:
We are adding support in the Customer Center for cross product promotional offers. This PR adds the code needed for that to work and make sure old style promotional offers continue to work in case we have to disable the cross product promotional offers in the backend.


RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
```diff
@@ -84,22 +84,58 @@ private extension LoadPromotionalOfferUseCase {
     }
 
     private func findDiscount(
-        for product: StoreProduct,
+        for activeProduct: StoreProduct,
         productIdentifier: String,
         promoOfferDetails: CustomerCenterConfigData.HelpPath.PromotionalOffer
```
facumenzella, line 89: Is there a case where the `activeProduct` is not the one defining the `productIdentifier`?

RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
```diff
@@ -84,22 +84,58 @@ private extension LoadPromotionalOfferUseCase {
     }
 
     private func findDiscount(
```
facumenzella, line 86: nit: What about renaming this function to `firstEligibleDiscount` and documenting the logic with a test?

facumenzella: To do that I would either create a extension on `PromotionalOffer` and inject `purchasesProvider`, or do the oposite and create yet another function in `CustomerCenterPurchasesType`

RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
vegaro: yes, the productIdentifier could be of the target product. So active product could be `monthly` and productIdentifier could be `annual`. I will change `productIdentifier` to `discountedProductIdentifier`

vegaro: no, it's always `activeProduct.productIdentifier`. I will remove it

RevenueCatUI/CustomerCenter/Data/LoadPromotionalOfferUseCase.swift
vegaro: It's being a bit complex to write those tests, but I am on it...

Sources/Networking/Responses/CustomerCenterConfigResponse.swift
```diff
@@ -76,6 +76,12 @@ struct CustomerCenterConfigResponse {
             let title: String
             let subtitle: String
             let productMapping: [String: String]
+            let crossProductPromotions: [String: CrossProductPromotion]?
+
+            struct CrossProductPromotion {
+                let storeOfferIdentifier: String
+                let targetProductId: String
+            }
```
facumenzella, line 84: I think I would define this at the same level of PromotionalOffer at least. This level of nesting is unnecessary

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -452,17 +452,33 @@ public struct CustomerCenterConfigData: Equatable {
...
+                public let targetProductId: String
+
+                public init(
+                    storeofferingidentifier: String,
+                    targetproductid: String
+                ) {
+                    self.storeOfferIdentifier = storeofferingidentifier
+                    self.targetProductId = targetproductid
+                }
+            }
```
facumenzella, line 468: nit: I think we should stop nesting types. It's absolutely annoying for a type that's internal. Ideally we should flatten all the types, and make them `spi`.

We can do that later, of course. End of rant

facumenzella (approved):
Nice work @vegaro !! 👑

Sources/CustomerCenter/CustomerCenterConfigData.swift
vegaro: yeah honestly there's a `nesting` linter rule for that reason that we are ignoring. I completely agree

PR #5034, opened 2025-05-01
chore: Make StoreProductDiscountType @__spi public

facumenzella:
### Motivation
While working on the new billing information for Customer Center subscriptions, I realised I need this protocol in `RevenueCatUI`

### Description
Internally expose `StoreProductDiscountType` using `@_spi`

### DIFF
--- a/RevenueCatUI/Helpers/StoreProductDiscount+Extensions.swift
+++ b/RevenueCatUI/Helpers/StoreProductDiscount+Extensions.swift
@@ -12,9 +12,30 @@
 //  Created by Cesar de la Vega on 25/9/24.
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
 
-extension StoreProductDiscount {
+@_spi(Internal) extension StoreProductDiscountType {
+
+    var discountedPeriodsWithUnit: String {
+        let formatter = DateComponentsFormatter()
+        formatter.unitsStyle = .full
+        var components = DateComponents()
+
+        switch self.subscriptionPeriod.unit {
+        case .day:
+            components.day = self.numberOfPeriods
+        case .week:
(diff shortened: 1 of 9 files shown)


RevenueCatUI/Helpers/StoreProductDiscount+Extensions.swift
```diff
@@ -12,9 +12,30 @@
 //  Created by Cesar de la Vega on 25/9/24.
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
 
-extension StoreProductDiscount {
+@_spi(Internal) extension StoreProductDiscountType {
+
+    var discountedPeriodsWithUnit: String {
```
ajpallares, line 19: Why can't this property keep being part of the `StoreProductDiscount` and has to be defined in the `StoreProductDiscountType` extension? Perhaps I'm missing something (or maybe the why will come in a future PR), but it looks to me that it could remain as a variable in `StoreProductDiscount`

ajpallares (approved):
I assume that this is needed for something that will come in a future PR, right?

Other than that, it looks great to me. And honestly, I think this is the perfect use case for `@_spi` in our SDK: APIs that need to be accessed by RevenueCatUI that don't have to be public. This prevent cluttering the public API for customers, so that's great!

RevenueCatUI/Helpers/StoreProductDiscount+Extensions.swift
facumenzella: Let me clarify.

This is indeed needed for an upcoming PR. We'll include `StoreProductDiscountType` in `PurchaseInformation` so we can work with discounts and display the most accurate billing information.

To be able to work with mocks I need the protocol, and all the available methods to use. In this case, the one I need is `localizedPricePerPeriodByPaymentMode`

PR #5035, opened 2025-05-02
fix: Hide cancel & refundRequest if subscription is already cancelled

facumenzella:
### Motivation
We shouldn't offer actions that are when the subscription is not eligible for it

### Description
1. Hide cancel path if it's already cancelled
2. Hide requestRefund if it's already cancelled

I've realized that the fact that `ManageSubscriptionsViewModel` is a `StateObject` was causing rendering issues. The fix works using `@Binding` +`onChangeOf`.
Having the `Binding` in the view let's you hook to `onChangeOf`, updating the `Published` property in the view model.

I am not happy with the solution, but I'll revisit the screens once this is out. Now there's a test to make sure it's working

### Video
https://github.com/user-attachments/assets/2098f9a1-a864-4bcf-a568-fb9b006ba21e


RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -54,6 +54,13 @@ struct PurchaseInformation {
     /// - `false` for purchases outside the trial period.
     let isTrial: Bool
 
+    /// Indicates wheter the purchased subscripcion is cancelled
```
ajpallares, line 57: 2 misspellings

```suggestion
    /// Indicates whether the purchased subscription is cancelled
```

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -109,6 +118,7 @@ struct PurchaseInformation {
             }
             self.isLifetime = entitlement.expirationDate == nil
             self.isTrial = entitlement.periodType == .trial
+            self.isCancelled = entitlement.unsubscribeDetectedAt != nil && !entitlement.willRenew
```
ajpallares, line 121: Huge nit, but the same way `isCancelled` is defined below in the extension of `SubscriptionInfo` that makes it conform to `Transaction`, it would perhaps be cleaner to define the `isCancelled` in `EntitlementInfo` as well (there's already an extension below). FFTI though

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
```diff
@@ -39,7 +39,7 @@ struct ManageSubscriptionsView: View {
     @Environment(\.navigationOptions)
     var navigationOptions
 
-    @StateObject
+    @ObservedObject
```
ajpallares, line 42: This is the correct approach. `@StateObject` should only be used when the view creates the object. If it's injected (as in this case), then it should be `@ObservedObject` 👍

ajpallares (approved):
Looks good!! A couple of comments only

RevenueCatUI/CustomerCenter/Views/ManageSubscriptionsView.swift
facumenzella: I've rewritten it using `Binding` + `onChangeOf` 🚀

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
facumenzella: ✅

PR #5036, opened 2025-05-03
[WIP] feat: Show active subscriptions list instead of the one with earliest expiration date

facumenzella:
### Motivation
Due to recent apple changes regarding IAPs in the US, we're prioritizing up this task.

### Description
Show the list of active subscriptions instead of the one with earliest expiration date.
Show a warning message that's managed via dashboard.

To be fast, I've copied `ManageSubscriptionsView` to `ManageSubscriptionView` (singular instead of plural)
The singular works exactly like the old predecessor, and the new one just displays the list. 

A caveat is that most of the code lives in the view model, and I didn't want to move that around just yet.


facumenzella:
closing for now

PR #5017, opened 2025-04-24
feat: Add missing pieces for maestro local execution

facumenzella:
### Motivation & Description
~Run maestro tests in CircleCI~

It's not possible to launch an application with an storekit config file using `xcrun simctl`, so our dreams of running the test suite on CircleCI will be delayed.

There's this [issue](https://github.com/mobile-dev-inc/Maestro/issues/945) that has a POC to do so, but we'll have to wait for now. 

In the meantime, this is still useful for local validation.

PR #5037, opened 2025-05-03
other: Use not synced storekit config to avoid unwanted changes

facumenzella:
### Motivation
As @ajpallares pointed out, the synced file generated unwanted local changes https://github.com/RevenueCat/purchases-ios/pull/5017#discussion_r2068357541

This intends to remove that.

### Description
Use a local .storekit config file instead of a synced one for maestro test app.
I've also updated the tests accordingly.


Examples/rc-maestro/Project.swift
```diff
@@ -46,7 +46,7 @@ let project = Project(
                 configuration: "Debug",
                 executable: "Maestro-Debug",
                 options: .options(
-                    storeKitConfigurationPath: "rc-maestro/Resources/StoreKit/StoreKitConfigDefault.storekit"
+                    storeKitConfigurationPath: "rc-maestro/Resources/StoreKit/StoreKitConfiguration.storekit"
```
ajpallares, line 49: This is the main change, right? I mean, the StoreKitConfiguration.storekit file is not synced while StoreKitConfigDefault.storekit is. Or am I missing something

facumenzella: I learned that you can't make a synced file non-synced, so I had to create a new one and create new subscriptions.
That's why I tweaked the tests accordingly.

But yeah, the main change is generating the project with a different file

ajpallares: Thanks for the explanation!

ajpallares (approved):
🚀

PR #5038, opened 2025-05-03
fix: Use ObservedObject instead of StateObject

facumenzella:
### Motivation
While working on https://github.com/RevenueCat/purchases-ios/pull/5035 I've noticed `ManageSubscriptionsView` wasn't updating accordingly when purchasesInformation was updated (cancelled)

### Description
This fixes the issue, and will refresh then screen whenever `PurchaseInformation` is updated.


ajpallares (commented):
In this case, I do think that `@StateObject` is the right approach. Why is the change needed? In theory, as this view is the one creating the object, it should be a `@StateObject`

PR #5043, opened 2025-05-05
Fix offerings not being returned in the `offerings` property of the SDK Health Report

polpielladev:
### Checklist

### Motivation

While working on the new sample app, we noticed that we were not getting offerings products back on the health report struct. This PR fixes the issue.


ajpallares (approved):
Good one!

PR #5046, opened 2025-05-05
Add support for fetching web billing products

tonidero:
### Description
TODO


Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift
```diff
@@ -367,3 +376,130 @@ private extension StoreProduct {
...
+        return 0
+    }
+
+    var localizedPriceString: String {
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .currency
+        if let currencyCode = self.currencyCode {
+            formatter.currencyCode = currencyCode
+        }
+        return formatter.string(from: self.price as NSDecimalNumber) ?? "\(self.price)"
```
ajpallares, line 439: I think you can use here the `priceFormatter` defined below
```suggestion
        self. priceFormatter.string(from: self.price as NSDecimalNumber) ?? "\(self.price)"
```

Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift
```diff
@@ -367,3 +376,130 @@ private extension StoreProduct {
...
+
+    var subscriptionPeriod: SubscriptionPeriod? {
+        guard let base = self.defaultPurchaseOption?.base,
+              let periodDuration = base.periodDuration else {
+            return nil
+        }
+
+        // Parse period duration string (e.g. "P1M" for 1 month)
+        // Format is ISO 8601 duration format
+        if periodDuration.hasPrefix("P") {
```
ajpallares, line 472: Just an idea: we could use `ISOPeriodFormatter` to convert into `ISODuration` and then, from it, convert it into a `SubscriptionPeriod`. It wouldn't be so clean because we would need to check something like this: 

```
if isoDuration.years > 0 {
    return SubscriptionPeriod(value: isoDuration.years, unit: .year)
} else if isoDuration.months > 0 {
    return SubscriptionPeriod(value: isoDuration.months, unit: .month)
...
```
and so on. I think it could be better because we would be centralizing the parsing of this kind of period strings. FFTI though, as it's not perfect either

ajpallares (commented):
I didn't fully check the test, but so far it looks good to me. Only some comments. Thank you!

Sources/Purchasing/OfferingsFactory.swift
```diff
@@ -85,14 +92,30 @@ class OfferingsFactory {
...
+        webProductsById: [String: StoreProduct],
         offeringIdentifier: String
     ) -> Package? {
-        guard let product = productsByID[data.platformProductIdentifier] else {
+        let webProductId = data.productIdByStoreType[.rcBilling]
+        var webProduct: StoreProduct?
+        if let webProductId {
+            webProduct = webProductsById[webProductId]
+        }
+        let iosProduct = productsByID[data.platformProductIdentifier]
```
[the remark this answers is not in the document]
tonidero: I think that's possible, however the iOS product would only appear in the `productsById` dict and the web product in the `webProductsById`. I think there wouldn't be a collision? Happy to chat if you have concerns though.

Sources/Purchasing/OfferingsManager.swift
```diff
@@ -219,54 +220,89 @@ private extension OfferingsManager {
...
-                // There is a widely reported bug in the iOS 18.4 Simulator affecting some HTTP requests
-                let showSimulatorWarning = self.systemInfo.isSubjectToKnownIssue_18_4_sim()
-                completion(.failure(Self.createErrorForEmptyResult(result.error,
-                                                                   showSimulatorWarning: showSimulatorWarning)))
-                return
+        self.fetchWebProductsInfo(appUserID: appUserID,
+                                  productIDs: webProductIdentifiers) { webProductsByID in
+            let missingWebProductIDs = self.getMissingProductIDs(productIDsFromStore: Set(webProductsByID.keys),
+                                                                 productIDsFromBackend: webProductIdentifiers)
+            if !missingWebProductIDs.isEmpty {
```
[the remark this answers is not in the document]
tonidero: I'm unsure about this one TBH... I think it can make sense if it's only used by the health SDKs so I might as well add it 👍

JayShortway (approved):
Very excited about this! Only minor stuff. Will approve to unblock.

Sources/Networking/Responses/OfferingsResponse.swift
```diff
@@ -68,6 +69,22 @@ extension OfferingsResponse {
...
+    var productIdentifiersByStoreType: [String: Set<String>] {
+        var result: [String: Set<String>] = [:]
+
+        for offering in self.offerings {
+            for package in offering.packages {
+                if let productIdByStoreType = package.productIdByStoreType {
+                    for (store, productId) in productIdByStoreType {
+                        result[store, default: Set()].insert(productId)
+                    }
+                }
```
JayShortway, line 81: Shouldn't the `package` in the response have a dictionary keyed the other way around? So `storeByProductId`? That would make the lookup here cheaper no?

Sources/Purchasing/Purchases/PurchaseParams.swift
```diff
@@ -109,6 +112,14 @@ import Foundation
         }
         #endif
 
+        /**
+         * 
+         */
+        @objc public func with(packageStoreProduct: StoreProduct?) -> Self {
+            self.packageStoreProduct = packageStoreProduct
+            return self
+        }
```
JayShortway, line 121: What about `productFromPackage`?

JayShortway (commented):
We're getting there! Really cool to see this. Just added some thoughts.

tonidero:
This is stale and was kinda done separately for the test store. We can still add support for this but would be easier to do it on top of the work done for the test store if so.

PR #5047, opened 2025-05-05
feat: Introduce fade transition for loop animated CarouselView

facumenzella:
### Motivation
We want to add support for these type of animations. 

### Description
In order to not break previous implementations, we're adding a few if / else here and there.

We can't use modern `withAnimation:complete` API so we had to go the old way. We animate the opacity, and simply don't animate the offset if the fade transition is turned on.

https://github.com/user-attachments/assets/04125683-beb3-4415-bb7b-de0693a7cbb2

Excellent team work @ajpallares @hiddevdploeg @polpielladev 🚀

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
@@ -64,7 +64,7 @@ struct CarouselComponentView: View {
                         pageControl: style.pageControl,
                         msTimePerSlide: style.autoAdvance?.msTimePerPage,
                         msTransitionTime: style.autoAdvance?.msTransitionTime,
-                        setUpAutoAdvanceAnimation: style.autoAdvance.map { _ in true } ?? false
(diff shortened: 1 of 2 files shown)


RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
```diff
@@ -190,10 +196,11 @@ private struct CarouselView<Content: View>: View {
...
                 PageControlView(
-                    originalCount: self.originalCount,
-                    pageControl: pageControl,
-                    currentIndex: self.$index
-                )
+                        originalCount: self.originalCount,
+                        pageControl: pageControl,
+                        currentIndex: self.$index,
+                        animationDuration: autoPlayTimerDuration.map { $0 / 5 }
+                    )
```
MarkVillacampa, line 203: indentation looks off here?

MarkVillacampa (approved):

aboedo (approved):

PR #5055, opened 2025-05-06
Add endpoint for fetching web billing products

tonidero:
### Description
Add endpoint to fetch web billing product information. Currently unused, to be used in future PRs.

### DIFF
--- a/Sources/Networking/Caching/WebProductsCallback.swift
+++ b/Sources/Networking/Caching/WebProductsCallback.swift
@@ -7,9 +7,9 @@
 //
 //      https://opensource.org/licenses/MIT
 //
-//  OfferingsCallback.swift
+//  WebProductsCallback.swift
 //
-//  Created by Joshua Liebowitz on 11/19/21.
+//  Created by Toni Rico on 5/6/25.
 
 import Foundation
 
--- a/Sources/Networking/Caching/WebProductsCallback.swift
+++ b/Sources/Networking/Caching/WebProductsCallback.swift
@@ -1,0 +1,21 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  OfferingsCallback.swift
+//
+//  Created by Joshua Liebowitz on 11/19/21.
+
+import Foundation
+
+struct WebProductsCallback: CacheKeyProviding {
+
+    let cacheKey: String
+    let completion: (Result<WebProductsResponse, BackendError>) -> Void
+
+}
--- a/Sources/Networking/Operations/GetWebProductsOperation.swift
+++ b/Sources/Networking/Operations/GetWebProductsOperation.swift
@@ -7,9 +7,9 @@
 //
 //      https://opensource.org/licenses/MIT
 //
-//  GetOfferingsOperation.swift
+//  GetWebProductsOperation.swift
 //
-//  Created by Joshua Liebowitz on 11/19/21.
+//  Created by Toni Rico on 5/6/25.
 
 import Foundation
 
--- a/Sources/Networking/Operations/GetWebProductsOperation.swift
+++ b/Sources/Networking/Operations/GetWebProductsOperation.swift
@@ -1,0 +1,89 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  GetOfferingsOperation.swift
+//
+//  Created by Joshua Liebowitz on 11/19/21.
+
+import Foundation
+
+final class GetWebProductsOperation: CacheableNetworkOperation {
+
+    private let webProductsCallbackCache: CallbackCache<WebProductsCallback>
+    private let configuration: AppUserConfiguration
+    private let productIds: Set<String>
+
+    static func createFactory(
+        configuration: UserSpecificConfiguration,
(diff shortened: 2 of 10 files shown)


Sources/Networking/Caching/WebProductsCallback.swift
```diff
@@ -0,0 +1,21 @@
...
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  OfferingsCallback.swift
+//
+//  Created by Joshua Liebowitz on 11/19/21.
```
JayShortway, line 12: Name change 😄

Sources/Networking/Operations/GetWebProductsOperation.swift
```diff
@@ -0,0 +1,88 @@
...
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  GetOfferingsOperation.swift
+//
+//  Created by Joshua Liebowitz on 11/19/21.
```
JayShortway, line 12: 😄

Sources/Networking/Responses/WebProductsResponse.swift
```diff
@@ -0,0 +1,61 @@
...
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  OfferingsResponse.swift
+//
+//  Created by Nacho Soto on 3/31/22.
```
JayShortway, line 12: 😄

Sources/Networking/Responses/WebProductsResponse.swift
```diff
@@ -0,0 +1,61 @@
...
+
+import Foundation
+
+struct WebProductsResponse {
+
+    struct Price {
+        let amountMicros: Int
```
JayShortway, line 19: Should this be 64 bits on all platforms, i.e.  `Int64`?

Sources/Networking/Responses/WebProductsResponse.swift
```diff
@@ -0,0 +1,61 @@
...
+        let currency: String
```
JayShortway, line 20: Is this the code or the symbol? Maybe we can clarify in the field name?

Sources/Networking/Responses/WebProductsResponse.swift
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    struct PricingPhase {
```
JayShortway, line 23: Is this supported on the backend already? Just curious.

Sources/Networking/WebBillingHTTPRequestPath.swift
```diff
@@ -0,0 +1,64 @@
...
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  DiagnosticsHTTPRequestPath.swift
+//
+//  Created by Cesar de la Vega on 8/4/24.
```
JayShortway, line 12: 😄

Sources/Networking/WebBillingHTTPRequestPath.swift
```diff
@@ -0,0 +1,64 @@
...
+
+import Foundation
+
+extension HTTPRequest.WebBillingPath: HTTPRequestPath {
+
+    // swiftlint:disable:next force_unwrapping
+    static let serverHostURL = URL(string: "https://api.revenuecat.com")!
```
JayShortway, line 19: Should this support the fallback URLs?

Sources/Networking/WebBillingHTTPRequestPath.swift
```diff
@@ -0,0 +1,64 @@
...
+        switch self {
+        case .getWebProducts:
+            return true
+        }
+    }
+
+    var shouldSendEtag: Bool {
+        switch self {
+        case .getWebProducts:
+            return false
```
JayShortway, line 31: Curious why not?

JayShortway (approved):
Good stuff! Just minor things. Approving to unblock you.

Sources/Networking/Responses/WebProductsResponse.swift
tonidero: Yes it is! We followed a similar basic model as Google's for this

Sources/Networking/WebBillingHTTPRequestPath.swift
tonidero: I don't think we do, and if we do, we can add it later I think

Sources/Networking/WebBillingHTTPRequestPath.swift
tonidero: Hmm probably can be set to Yes actually. Good call!

Sources/Networking/Responses/WebProductsResponse.swift
tonidero: Hmm I guess this means we would support a maximum of ~2k for whatever currency... Probably better to increase it yeah. Good call!

tonidero:
Will merge this since it should be safe (nothing is used just yet).

Sources/Networking/WebBillingHTTPRequestPath.swift
ajpallares: Yeah, currently a fallback URL is only supported for get customer info and for get product entitlement mapping

Sources/Networking/Responses/WebProductsResponse.swift
JayShortway: Works for me!

PR #5064, opened 2025-05-07
Fix sheet view in v2 paywall not covering bottom safe area

ajpallares:
### Description
As the title says


RevenueCatUI/Templates/V2/Components/Button/BottomSheetView.swift
```diff
@@ -100,13 +100,13 @@ struct BottomSheetOverlayModifier: ViewModifier {
                         viewModel: sheetViewModel.sheetStackViewModel,
                         onDismiss: {
                             self.sheetViewModel = nil
-                        }
+                        },
+                        additionalPadding: self.safeAreaInsets
```
JayShortway, line 104: Beautiful!

JayShortway (approved):
Does what it says on the tin!

PR #5069, opened 2025-05-08
Split `PurchaseInformation.price` into `pricePaid` and `renewalPrice`

vegaro:
I split `price` into `pricePaid` and `renewalPrice` so we can better represent the current renewal price and the price paid in the UI. This cleaned up the code a lot actually.

This makes the UI look like:

Renewing paid web subscription:
Screenshot 2025-05-07 at 18 10 17

Non-renewing paid web subscription:
Screenshot 2025-05-07 at 18 11 01

Renewing trial web subscription (we don't know the price yet until we call the products endpoint):
Screenshot 2025-05-07 at 18 13 03

Non-renewing trial web subscription:
Screenshot 2025-05-07 at 18 12 20


RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -307,20 +308,60 @@ extension PurchaseInformation: Identifiable {
     }
 }
 
-private extension EntitlementInfo {
+private extension Transaction {
 
-    var isCancelled: Bool {
-        unsubscribeDetectedAt != nil && !willRenew
-    }
+    func guessRenewalPrice() -> PurchaseInformation.PriceDetails {
```
JayShortway, line 313: Maybe `determineRenewalPrice`?

JayShortway (approved):
This makes a ton of sense!

Sources/Purchasing/NonSubscriptionTransaction.swift
```diff
@@ -34,11 +34,12 @@ public final class NonSubscriptionTransaction: NSObject {
...
     @objc public let storeTransactionIdentifier: String
 
-    /**
-     * The ``Store`` where this transaction was performed.
-     */
+    /// The ``Store`` where this transaction was performed.
     @objc public let store: Store
 
+    /// The display name of the product as configured in the RevenueCat dashboard.
+    @objc public let displayName: String?
```
facumenzella, line 41: I remember we had a conversation about this, and decided that this might not be something the developer wants to show to their customers 🤔 
Right?

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -54,36 +57,51 @@ struct PurchaseInformation {
     /// - `false` for purchases outside the trial period.
     let isTrial: Bool
 
-    /// Indicates whether the purchased subscription is cancelled
+    /// Indicates wheter the purchased subscripcion is cancelled
```
facumenzella, line 60: /// Indicates whether the purchased subscription is cancelled

Sources/Purchasing/NonSubscriptionTransaction.swift
vegaro: yeah, but the alternative is the product id. We would only get here for other store products. I think it's a good compromise.

I would say let's leave this and if someone complains we can change it

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
vegaro: I was missing an update in the branch

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -24,16 +24,19 @@ struct PurchaseInformation {
 
     /// The title of the storekit product, if applicable.
     /// - Note: See `StoreProduct.localizedTitle` for more details.
```
facumenzella, line 26: This one is outdated 🤓

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -465,7 +465,7 @@ final class CustomerCenterViewModelTests: TestCase {
         let purchaseInformation = try XCTUnwrap(viewModel.purchaseInformation)
         expect(purchaseInformation.title) == "lifetime"
         expect(purchaseInformation.durationTitle).to(beNil())
-        expect(purchaseInformation.price) == .paid("$29.99")
+        expect(purchaseInformation.pricePaid) == .paid("$29.99")
         expect(purchaseInformation.productIdentifier) == productIdLifetime
```
facumenzella, line 470: can we update these with `renewalPrice` as well?

facumenzella (approved):

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -263,7 +295,8 @@ final class CustomerCenterViewModelTests: TestCase {
         expect(purchaseInformation.title) == "title"
         expect(purchaseInformation.durationTitle) == "1 month"
 
-        expect(purchaseInformation.pricePaid) == .nonFree("$2.99")
+        expect(purchaseInformation.pricePaid) == .nonFree(formatted(price: 3.99, currencyCode: currency))
+        expect(purchaseInformation.renewalPrice).to(beNil())
```
vegaro, line 299: We can also mock a RenewalInfo here to have a `renewalPrice`?

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -346,7 +380,8 @@ final class CustomerCenterViewModelTests: TestCase {
             // Should always show yearly subscription since it expires first
             expect(purchaseInformation.title) == yearlyProduct.title
             expect(purchaseInformation.durationTitle) == yearlyProduct.duration
-            expect(purchaseInformation.pricePaid) == .nonFree(formatPrice(yearlyProduct.price))
+
+            expect(purchaseInformation.pricePaid) == .nonFree(formatted(price: 29.99))
```
vegaro, line 384: should we test the `renewalPrice` here as well?

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -462,7 +500,7 @@ final class CustomerCenterViewModelTests: TestCase {
                     entitlementId: "pro",
                     productId: productIdLifetime,
                     purchaseDate: purchaseDateLifetime,
-                    expirationDate: nil
+                    expirationDate: nil,
```
vegaro, line 503: I think this will break in older swift versions? I am unsure

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -706,11 +749,11 @@ final class CustomerCenterViewModelTests: TestCase {
         expect(viewModel.activePurchases.count) == 1
         expect(viewModel.activePurchases.first?.productIdentifier) == purchaseInformation.productIdentifier
 
-        expect(purchaseInformation.title).to(beNil())
+        expect(purchaseInformation.title) == "Weekly Scratched Sofa" // display name
         expect(purchaseInformation.durationTitle).to(beNil())
         expect(purchaseInformation.explanation) == .earliestRenewal
         expect(purchaseInformation.store) == .appStore
-        expect(purchaseInformation.pricePaid) == .unknown
+        expect(purchaseInformation.pricePaid) == .nonFree(formatted(price: 1.99)) // from transaction
```
vegaro, line 756: same here about `renewalInfo`

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
facumenzella: We can, I've done the two combinations in one test to prove it works. 

After fixing all of them, I realised there's a dedicated test for it already `testPurchaseInformationUsesInfoFromRenewalInfoWhenAvailable`. 😅

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
facumenzella: we need a formatter

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -285,57 +300,112 @@ extension PurchaseInformation {
...
+        }
 
-    func priceBestEffort(product: StoreProduct?) -> PurchaseInformation.PriceDetails {
-        if let product {
-            return .paid(product.localizedPriceString)
+        guard self.store == .rcBilling else {
+            // RCBilling does not support product price changes yet
+            // So it's the only store we can infer the renewal price from
+            // latest price paid
+            return nil
```
ajpallares, line 358: Should this be `nil` or `.unknown`?

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -285,57 +300,112 @@ extension PurchaseInformation {
...
-        if self.store == .promotional {
+
+        switch self.type {
+        case .subscription(isActive: let isActive,
+                           willRenew: let willRenew,
+                           expiresDate: let expiresDate,
+                           isTrial: let isTrial):
+            if !willRenew || isTrial {
+                // If it was a trial we can't infer the renewal price
+                return nil
```
ajpallares, line 368: Same. `nil` or `.unknown`?

RevenueCatUI/CustomerCenter/Data/Transaction.swift
```diff
@@ -56,4 +60,10 @@ extension NonSubscriptionTransaction: Transaction {
     var managementURL: URL? {
         nil
     }
+
+    var price: ProductPaidPrice? {
+        // We don't have that information in the CustomerInfo
+        nil
```
ajpallares, line 66: `.unknown`?

Sources/Identity/SubscriptionInfo.swift
```diff
@@ -78,6 +78,9 @@ import Foundation
     /// Management purchase URL
     @_spi(Internal) public let managementURL: URL?
 
+    /// The display name of the subscription as configured in the RevenueCat dashboard.
+    @objc public let displayName: String?
```
ajpallares, line 82: Just to confirm: do we need to make this property public or is it only supposed to be used by RevenueCatUI (e.g. `@_spi`)?

Sources/Purchasing/NonSubscriptionTransaction.swift
ajpallares: Oh this probably answer my previous comment. However, this property would not be localized, so could it mean trouble if developers rely on this property to be user-facing and contains a localization the don't expect? Just asking

ajpallares (approved):
It looks good to me in general. 

I have multiple comments, but they mostly refer to the `PriceDetails` enum, where I may be confusing things (when it is `nil` vs `.unknown`). 

Other than that, I think it's good to go!

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -214,7 +221,7 @@ struct PurchaseInformation {
 
     enum PriceDetails: Equatable, Hashable {
         case free
-        case paid(String)
+        case nonFree(String)
         case unknown
```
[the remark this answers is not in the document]
facumenzella: Absolutely valid point. I think the main issue is that we're using the same enum for both renewal and paid price.

In the context of renewal, `unknown` does not make sense. Either it's there, or it's not. 

In the context of paid price, `unknown` does make sense because there are some cases where we might not be able to tell the paid price (customer info does not provide info about consumables)

---

That said, I think that I kind of like having an `unknown` in `PriceDetails`. BUT we can use wrapper types like:

```swift
    enum PriceDetails: Equatable, Hashable {
        case free
        case nonFree(String)
        case unknown
    }
 
    // and a similar structure for `RenewalPrice`
    struct PaidPrice: Equatable, Hashable {
        let details: PriceDetails

        init?(_ price: PriceDetails) {
            self.price = price
        }
    }
```

Also, this line won't be a thing anymore. With the new specs, each price is shown in a different place so we don't need to treat them the same type.
```swift
let priceToUse = purchaseInformation.renewalPrice ?? purchaseInformation.pricePaid
```

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
facumenzella: same as https://github.com/RevenueCat/purchases-ios/pull/5069#discussion_r2089408695

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
facumenzella: same as https://github.com/RevenueCat/purchases-ios/pull/5069#discussion_r2089408695

Sources/Identity/SubscriptionInfo.swift
facumenzella: I'll revert this for now. It's definitely out of scope of the PR and deserves a broader conversation 👍

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
facumenzella: Same as above

RevenueCatUI/CustomerCenter/Data/Transaction.swift
facumenzella: This is tricky. `RevenueCatUI.Transaction` has `var price: ProductPaidPrice? { get }`.

`CustomerInfo` provides `ProductPaidPrice` only for `SubscriptionInfo`, so we're adding it for `NonSubscriptionTransaction` to simplify the handling inside `PurchaseInformation`.

I think that eventually we will get rid of `RevenueCatUI.Transaction` because it's just another layer of abstraction that's used ONLY in the initializer.

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -285,57 +300,112 @@ extension PurchaseInformation {
...
+                // If it was a trial we can't infer the renewal price
+                return nil
+            }
+        case .nonSubscription:
+            return nil
+        }
+
+        guard let price = self.price,
+              price.amount != 0 else {
+            return nil
```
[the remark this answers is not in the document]
facumenzella: This should be nil because it's `RenewalPrice` now

RevenueCatUI/CustomerCenter/Views/SubscriptionDetailsView.swift
```diff
@@ -47,11 +47,13 @@ struct SubscriptionDetailsView: View {
...
                 let priceValue: String? = {
-                    switch purchaseInformation.price {
+                    switch priceToUse {
                     case .free:
                         return localization[.free]
-                    case .paid(let localizedPrice):
+                    case .nonFree(let localizedPrice):
                         return localizedPrice
                     case .unknown:
                         return nil
```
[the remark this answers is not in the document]
facumenzella: I've updated this, and now it behaves as you described above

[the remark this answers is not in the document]
facumenzella: However, this won't be used once we ship the subscription list

PR #5074, opened 2025-05-12
Fix flakiness of `uuid` implementation

ajpallares:
### Description
Unit test `testStringDataAsUUID()` was failing regularly, mostly in watchOS, due to the flaky implementation of `Data's` `var uuid: UUID?` in the SDK.

### Explanation
The implementation of `var uuid: UUID?` in `Data` uses `NSUUID(uuidBytes: baseAddress)`, which expects exactly 16 bytes (i.e. `MemoryLayout<UUID>.size`). The unit test is checking `"sample string".asData`, which has 13 bytes, so `NSUUID(uuidBytes:)` reads 3 extra bytes from memory, which can contain:
* zeroes (--> test passes)
* non-zero garbage (--> test fails)

The fix consists of modifying the implementation of `var uuid: UUID?` in `Data` so that, if the `Data`'s length is shorter than `MemoryLayout<UUID>.size` (16 bytes), a padding of zeroes is added to make sure that `NSUUID(uuidBytes:)` is always called with at least 16 bytes.


ajpallares:
@RCGitBot please test

facumenzella (approved):
Nice catch!!

PR #5080, opened 2025-05-12
Add management URL to PurchaseInformation

facumenzella:
### Motivation
After reverting the subscription list, I'm re-applying it in pieces to cleanup and track down the issue.

### Description
Add `managementURL` to `PurchaseInformation` for non appstore transactions.
Cancel subscription will open the URL if available.


facumenzella:
@RCGitBot please test

RevenueCat.xcworkspace/xcshareddata/swiftpm/Package.resolved
vegaro, line 1: Is this intended? I am also getting these sort of changes, but I am unclear if we should commit them

vegaro (approved):
Just that question about the Package.resolved. The rest of the changes make sense since they had already been reviewed

PR #5082, opened 2025-05-12
Fix some flaky tests

ajpallares:
### Motivation
Flaky tests

### Description
This PR fixes the flakiness of 2 unit tests that fail regularly:
1. `StoreKit2TransactionListenerDiagnosticsTests`'s `testTracksDiagnosticsForRenewals()`: which was expecting exactly 1 transaction diagnostics event, but due to the `testSession.timeRate` being set to `.oneRenewalEveryTwoSeconds`, more than one renewal can happen during the test. I've relaxed the condition to expect at least 1 transaction diagnostics event with the expected parameters.
2. `PurchasesConfiguringTests`'s `testOfflineCustomerInfoDisabledForCustomEntitlementsComputation()` was crashing due sometimes due to the SDK being configured for Customer Entitlements Computation without an App User ID (see these lines https://github.com/RevenueCat/purchases-ios/blob/fd893d2a34460525c6c48891018fe76797acef89/Sources/Purchasing/Purchases/Purchases.swift#L657-L660).
As a result, this test was only passing if another test setting the `appUserId` had been executed before.

There are still many more flaky tests to fix, but... baby steps


vegaro (approved):
thank you thank you thank you

PR #5084, opened 2025-05-13
[DO NOT MERGE] Add log when detecting use of StoreKit Configuration file

ajpallares:
### Checklist

### Description
Adds an info log when the SDK detects that a StoreKit Configuration file is being used

### Notes
Because the implementation relies on `StoreKit.AppTransaction.shared`, which requires authentication in the App Store to be accessed. This means that it may prompt for the Apple credentials login.

### DIFF
--- a/Sources/Logging/Strings/StoreKitStrings.swift
+++ b/Sources/Logging/Strings/StoreKitStrings.swift
@@ -95,6 +95,8 @@ enum StoreKitStrings {
 
     case skunknown_purchase_result(String)
 
+    case skconfig_file_detected
+
 }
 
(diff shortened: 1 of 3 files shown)


ajpallares:
Based on my tests (with Xcode 16.3), I've found the following:
* In an iOS device with iOS 18.4.1 with an already-authenticated Apple account, I'm not getting prompted for the Apple account credentials in any case (either using SK Config file or not).
* In simulators, when using SK Config file, no prompt happens.
* In simulators, when not using a SK Config file, a prompt appears in iOS 18.3 and lower.
* In iOS 18.4 simulators, no prompts appear (bug or feature?).

PR #5085, opened 2025-05-13
Add expirationDate and renewalDate to PurchaseInformation

facumenzella:
### Motivation
After https://github.com/RevenueCat/purchases-ios/pull/5078, I am bringing back the subscription list.

### Description
Add `expirationDate` and `renewalDate` to `PurchaseInformation`. This will be used as part of the new billing information label for each subscription.

PR #5086, opened 2025-05-13
Add missing keys for billing information 

facumenzella:
### Motivation
After https://github.com/RevenueCat/purchases-ios/pull/5078, I am bringing back the subscription list.

### Description
I am adding missing keys for the billing information.

I am working on automating the generation of the localized string structure. This is an example of the generated one:
- sorted enum cases
- generated based on https://github.com/RevenueCat/khepri/blob/main/khepri/services/customer_center/domain/default_stored_customer_center_config.json


ajpallares (commented):
Two comments:
1. I believe the missing changes from #5085 about the `dateFormatter` are still missing here
2. Other than renaming the existing casesin the `Localization`, I see some new cases that are not being used (e.g. `discountedRecurringThenPrice`). Will this be used in a future PR?

facumenzella:
@ajpallares I've pushed the missing commit about using the correct formatter. 

About unused keys will be used for the billing information text. However, it's possible that there are some legacy keys from customer center v0 that should be removable. Let me double-check with @vegaro on how to do that, and I'll remove them.

In the meantime I think we're good as it is.

facumenzella:
Ok after discussing this in private, I'll hold on to this because the struct is already private. The way to go is to deprecate the old one, and start using the new one.

I'll come back to this later

PR #5088, opened 2025-05-13
Use dateFormatter inside PurchaseInformation

facumenzella:
### Description
Follow-up of https://github.com/RevenueCat/purchases-ios/pull/5085

Use the instance dateFormatter instead of default one.


ajpallares (approved):
Thank you!

PR #5090, opened 2025-05-13
Introduce PurchaseInformationCardView

facumenzella:
### Motivation
We want to show the whole list of active subscriptions, instead of just one purchase with the earliest expiry date.

### Description
Introduce PurchaseInformationCardView as a new preview for a subscription. Note that this view is not used yet, as it will be used in the new subscription list view

### Next
Introduce views, and integrate with CustomerCenter

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
@@ -1,0 +1,117 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  PurchaseInformationCardView.swift
+//
+//  Created by Facundo Menzella on 13/5/25.
+
+import RevenueCat
+import SwiftUI
+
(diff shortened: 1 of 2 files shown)


RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
```diff
@@ -0,0 +1,117 @@
...
+    private let title: String
+    private let subtitle: String
+    private let storeTitle: String
+    private let showChevron: Bool
+
+    init(
+        title: String,
+        subtitle: String,
+        storeTitle: String,
+        showChevron: Bool = true
```
ajpallares, line 34: I'm curious: when will we need to show/hide the chevron?

ajpallares (approved):
Looks good to me 👍

vegaro (approved):

RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
facumenzella: As for now, the detail page will also show this card (without the chevron).

There's a bigger refactor, that consists of removing "Purchase history", and displaying purchase details on the detail screen. Once that's done, we can delete this flag

ajpallares: Makes sense! Thanks for the explanation!

PR #5091, opened 2025-05-13
Introduce SubscriptionDetailViewModel & BaseManageSubscriptionViewModel

facumenzella:
### Motivation
We want to display a list of active subscriptions in Customer Center. 

### Description
To avoid opening a huge PR, I'm splitting it into smaller pieces. This PR contains:
- `BaseManageSubscriptionViewModel` that's pretty much the existing `ManageSubscriptionViewModel`
- `SubscriptionDetailViewModel` that will be the new subscription detail screen

### Upcoming
- `SubscriptionListViewModel` to handle the list, that will also inherit `BaseManageSubscriptionViewModel`

> I've duplicated the tests from `ManageSubscriptionViewModel` to `BaseManageSubscriptionViewModel` to make sure the logic works. Once finished, I'll delete the old code in one PR

### DIFF
--- a/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptions/ManageSubscriptionsViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptions/ManageSubscriptionsViewModel.swift
@@ -53,15 +53,6 @@ final class ManageSubscriptionsViewModel: ObservableObject {
     @Published
     var inAppBrowserURL: IdentifiableURL?
 
-    @Published
-    var state: CustomerCenterViewState {
-        didSet {
-            if case let .error(stateError) = state {
-                self.error = stateError
-            }
-        }
-    }
-
     let actionWrapper: CustomerCenterActionWrapper
 
     @Published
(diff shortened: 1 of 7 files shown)


RevenueCatUI/CustomerCenter/ViewModels/ManageSubscriptions/ManageSubscriptionsViewModel.swift
```diff
@@ -90,7 +81,6 @@ final class ManageSubscriptionsViewModel: ObservableObject {
             self.actionWrapper = actionWrapper
             self.loadPromotionalOfferUseCase = loadPromotionalOfferUseCase
             ?? LoadPromotionalOfferUseCase(purchasesProvider: purchasesProvider)
-            self.state = .success
```
vegaro, line 93: doesn't removing this break the view? maybe this should be done in another PR?

facumenzella: The state is not being used anywhere, besides in tests that just check the initial state 😅 
I think it was meant to be used, but ended up in lost in the ether. I can bring it back once the screen refactor is done.

vegaro (approved):

PR #5092, opened 2025-05-14
Allow custom url on purchase button

joshdholtz:
### Motivation

Allow custom url on paywalls v2 purchase button for web checkout

### Description

`PurchaseButtonComponent` needed a more complex structure than the `action: Action?` enum to work better with web checkout and custom web checkout url

- New `method: Method?` property added as a successor to `action: Action?` (essentially deprecated)
  - `inAppCheckout`
  - `webCheckout(WebCheckout)`
      - `autoDimiss: Bool?`
      - `openMethod: URLMethod?` 
  - `webProductSelection(WebCheckout)`
      - `autoDimiss: Bool?`
      - `openMethod: URLMethod?`
  - `customWebCheckout(CustomWebCheckout)`
      - `autoDimiss: Bool?`
      - `openMethod: URLMethod?`
      - `customUrl: CustomerURL`
          - `url: LocalizationKey`
          - `packageParam: String?`
- Added support for open method (external or in-app browser)
- Added unit tests for deserializing `PurchaseButtonComponent`
- Added more to a SwiftUI preview for testing out all these methods (which prints what it would do in a real app to the console)

### Demo

https://github.com/user-attachments/assets/3e0a1e4f-5f24-47b6-be6b-51932526b72b

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
@@ -214,7 +214,8 @@ struct PurchaseButtonComponentView_Previews: PreviewProvider {
                                            trailing: 30)
                         ))
                     ]),
-                    action: .inAppCheckout
+                    action: .inAppCheckout,
+                    method: .inAppCheckout
                 ),
                 localizationProvider: .init(
                     locale: Locale.current,
@@ -258,7 +259,8 @@ struct PurchaseButtonComponentView_Previews: PreviewProvider {
                                                 bottomLeading: 8,
                                                 bottomTrailing: 8))
                     ),
-                    action: .inAppCheckout
+                    action: .inAppCheckout,
+                    method: .inAppCheckout
                 ),
                 localizationProvider: .init(
                     locale: Locale.current,
@@ -298,7 +300,8 @@ fileprivate extension PurchaseButtonComponentViewModel {
             offering: offering
         )
 
-        self.init(
+        try self.init(
+            localizationProvider: localizationProvider,
             component: component,
             offering: offering,
             stackViewModel: stackViewModel
--- a/RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
@@ -108,7 +108,7 @@ struct PurchaseButtonComponentView: View {
     private func purchaseSelectedWebProduct() async throws {
         self.logIfInPreview(package: self.packageContext.package)
 
-        guard let webCheckoutUrl = self.packageContext.package?.webCheckoutUrl else {
+        guard let webCheckoutUrl = self.viewModel.urlForWebProduct(packageContext: self.packageContext) else {
             Logger.error(Strings.no_web_checkout_url_found)
             return
         }
@@ -129,10 +129,10 @@ struct PurchaseButtonComponentView: View {
 
     private func openWebPaywallLink(url: URL, method: PaywallComponent.ButtonComponent.URLMethod) {
         Purchases.shared.invalidateCustomerInfoCache()
-#if os(watchOS)
+        #if os(watchOS)
         // watchOS doesn't support openURL with a completion handler, so we're just opening the URL.
         openURL(url)
-#else
+        #else
         openURL(url) { success in
             if success {
                 Logger.debug(Strings.successfully_opened_url_external_browser(url.absoluteString))
@@ -140,8 +140,11 @@ struct PurchaseButtonComponentView: View {
                 Logger.error(Strings.failed_to_open_url_external_browser(url.absoluteString))
             }
         }
-#endif
-        onDismiss()
+        #endif
+
+        if self.viewModel.webAutoDimiss {
+            self.onDismiss()
+        }
     }
 
     /// Used to see purchasing information when using SwiftUI Previews
--- a/RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
@@ -119,7 +119,7 @@ struct PurchaseButtonComponentView: View {
             return
         }
 
-        self.openWebPaywallLink(url: webCheckoutUrl, method: .externalBrowser)
+        self.openWebPaywallLink(url: webCheckoutUrl.0, method: webCheckoutUrl.1)
     }
 
     private func openWebProductSelection() async throws {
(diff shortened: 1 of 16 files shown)


RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentViewModel.swift
ajpallares, line 1: I think the two methods `func urlForWebProduct(packageContext: PackageContext)` and `var offeringWebCheckoutUrl` could be unified into a single one:
```swift
func urlForWebCheckout(packageContext: PackageContext?) -> (URL, PaywallComponent.ButtonComponent.URLMethod)? {
    if let method = component.method {
        switch method {
        case .inAppCheckout, .unknown:
            return nil
        case .webCheckout(let webCheckout), .webProductSelection(let webCheckout):
            if let checkoutUrl = packageContext?.package?.webCheckoutUrl ?? offering.webCheckoutUrl {
                return (checkoutUrl, webCheckout.openMethod ?? .externalBrowser)
            } else {
                return nil
            }
        case .customWebCheckout(let customWebCheckout):
            if let customUrl = self.customWebCheckoutUrl {
                if let package = packageContext?.package,
                   let packageParam = customWebCheckout.customUrl.packageParam {
                    let url = customUrl.appending(name: packageParam, value: package.identifier)
                    return (url, customWebCheckout.openMethod ?? .externalBrowser)
                } else {
                    return (customUrl, customWebCheckout.openMethod ?? .externalBrowser)
                }
            } else {
                return nil
            }
        }
    } else if let url = packageContext?.package?.webCheckoutUrl ?? offering.webCheckoutUrl {
        return (url, .externalBrowser)
    } else {
        return nil
    }
}
```

where you would pass `packageContext: nil` to get the `offeringWebCheckoutUrl` (for web product selection). I believe the logic is exactly the same with the benefit in the `.webCheckout` and `.webProductSelection` cases that, if `package.webCheckoutUrl` were `nil` for any reason, then it still would fall back to the `offering.webCheckoutUrl` for web package selection.

Then on the caller site (`PurchaseButtonComponentView`) you would have
```swift
private func purchaseSelectedWebProduct() async throws {
    guard let webCheckoutUrl = self.viewModel.urlForWebCheckout(packageContext: self.packageContext) else {
        Logger.error(Strings.no_web_checkout_url_found)
        return
    }
    ...
}

private func openWebProductSelection() async throws {
    guard let webCheckoutUrl = self.viewModel.urlForWebCheckout(packageContext: nil) else {
        Logger.error(Strings.no_web_checkout_url_found)
        return
    }
    ...
}
```
WDYT? I don't think it adds complexity and it's less code. FFTI as perhaps makes things more complex

RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentViewModel.swift
```diff
@@ -24,22 +24,128 @@ class PurchaseButtonComponentViewModel {
...
+                return customWebCheckout.autoDismiss ?? Self.defaultWebAutoDismiss
+            case .inAppCheckout, .unknown:
+                break
+            }
+        } else if component.action != nil {
+            // Legacy action was previously always dismissing
+            return Self.defaultWebAutoDismiss
+        }
+
+        return Self.defaultWebAutoDismiss
```
ajpallares, line 97: Nit: 
```suggestion
        } else {
            return Self.defaultWebAutoDismiss
        }
```

Sources/Paywalls/Components/PaywallPurchaseButtonComponent.swift
```diff
@@ -16,6 +16,7 @@ public extension PaywallComponent {
         public let stack: PaywallComponent.StackComponent
 
         public let action: Action?
```
ajpallares, line 18: Should we mark it as `deprecated`? It's actually unused (literally, if you apply my nit above)
```suggestion
        @available(*, deprecated, message: "This API is unused and will be removed in a future version.")
        public let action: Action?
```

RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
ajpallares, line 1: In the `func purchase()` method, shouldn't we use the `Method` instead of the `Action`? I think it does not make any difference but since the rest of the implementation now relies in `Method`, I believe it would be more consistent.

This way, we could also get rid of the `Action` in code (we should only perhaps maintain it and deprecate it because it's a `public` API)

Sources/Paywalls/Components/PaywallPurchaseButtonComponent.swift
```diff
@@ -24,23 +25,134 @@ public extension PaywallComponent {
             case webProductSelection = "web_product_selection"
         }
 
+        public enum Method: Codable, Sendable, Hashable, Equatable {
```
ajpallares, line 28: This is, I think, the perfect place for `@_spi`. This way, it can be accessed from RevenueCatUI (which is needed), but it's not really a public API. That way, in the future, if we need to replace it (the same way `Method` replaces `Action`), we could just remove it instead of deprecating it. Not a big deal though, as every other API here is `public` already.

ajpallares (approved):
Thanks so much for implementing this! I think it makes total sense!

I have some comments, mostly nits, and some refactor (simplification?) suggestions. But looks good overall!

Sources/Paywalls/Components/PaywallPurchaseButtonComponent.swift
joshdholtz: Can't do this deprecation because the `pod lib lint` job will yell at us 😅

RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentViewModel.swift
```diff
@@ -24,22 +24,94 @@ class PurchaseButtonComponentViewModel {
...
+                return .webCheckout(.init(autoDismiss: true, openMethod: .externalBrowser))
+            case .webProductSelection:
+                return .webProductSelection(.init(autoDismiss: true, openMethod: .externalBrowser))
+            }
+        })
+    }
+
+    typealias LaunchWebCheckout = (url: URL, method: PaywallComponent.ButtonComponent.URLMethod, autoDismiss: Bool)
+
+    func urlForWebCheckout(packageContext: PackageContext?) -> LaunchWebCheckout? {
```
ajpallares, line 64: I think `PackageContext` can be nonoptional
```suggestion
    func urlForWebCheckout(packageContext: PackageContext) -> LaunchWebCheckout? {
```

RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentViewModel.swift
```diff
@@ -24,22 +24,94 @@ class PurchaseButtonComponentViewModel {
...
+            switch action {
+            case .inAppCheckout:
+                return .inAppCheckout
+            case .webCheckout:
+                return .webCheckout(.init(autoDismiss: true, openMethod: .externalBrowser))
+            case .webProductSelection:
+                return .webProductSelection(.init(autoDismiss: true, openMethod: .externalBrowser))
+            }
+        })
+    }
```
ajpallares, line 60: Nice mapping as safeguard! 👌  (at least we're giving a use to the existing `action` 😄)

ajpallares (approved):
I like it much more now that all the web checkout logic is unified into a single method. I believe it's more clean and clear. Thank you for doing it!

PR #5093, opened 2025-05-14
Add _spi(Internal) to StoreProductType

facumenzella:
### Motivation
While working on some tests for https://github.com/RevenueCat/purchases-ios/pull/5031 I realised I need to use `StoreProductType` instead of `StoreProduct` to be able to write some tests.
 
### Description
Make `StoreProductType` `_spi(Internal)` to be able to use it from `RevenueCaUI`


facumenzella:
Just found out about `TestStoreProduct`. Thanks @ajpallares

PR #5094, opened 2025-05-14
Add basic test for Cross Promotional Offers

facumenzella:
### Motivation
Sample test for cross promotional offers


vegaro:
Thank you! ❤️

PR #5095, opened 2025-05-14
PurchaseInformation conforms to Identifiable & Equatable

facumenzella:
### Motivation
I need `PurchaseInformation` conforming to these protocols

### Description
1. Identifiable: Use it inside a `ForEach`
2. Equatable: Use it with `onChangeOf` to update the view

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
+++ b/RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
@@ -17,7 +17,7 @@ import Foundation
 import RevenueCat
 import StoreKit
 
-// swiftlint:disable nesting
+// swiftlint:disable nesting file_length
 
 /// Information about a purchase.
 struct PurchaseInformation {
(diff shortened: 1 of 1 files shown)


RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -235,6 +235,16 @@ struct PurchaseInformation {
          return dateFormatter
      }()
 }
+
+extension PurchaseInformation: Identifiable {
+
+     var id: String {
+         return "\(productIdentifier)_\(Self.defaultDateFormatter.string(from: customerInfoRequestedDate))"
```
ajpallares, line 242: I guess this could be enough, but I'm not totally convinced. I think it would be better to make it conform to `Hashable`, and then hash more properties into the `Hasher`. This way, you basically get conformance to `Equatable` and `Identifiable` (using `typealias ID = Self`) for free.

In addition, I'm thinking, what about a customer who purchased two products from different stores (app + web) that have the same `productIdentifier`? I'm not 100% that this situation could happen, but anyway we can easily cover for these edge cases.

ajpallares (commented):
I get why we need this, but I'm not too convinced about the implementation. See my comment with my suggestion and lmk what you think!

ajpallares (approved):
Thanks for making the changes! I think it makes more sense like this!

PR #5096, opened 2025-05-14
Update Nimble dependency to v13.7.1

ajpallares:
### Motivation

The version of Nimble used before this PR did not support async/await. And using async/await could potentially help us fix some flaky tests that currently fail due to some expectation timeouts being exceeded.

### Description

Update Nimble to v13.7.1, the most recent release version (at the time or writing this), which is now compatible with async/await


ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

Tests/BackendIntegrationTests/Helpers/BaseStoreKitIntegrationTests+Verification.swift
```diff
@@ -26,7 +26,8 @@ extension BaseStoreKitIntegrationTests {
     @discardableResult
     func verifyEntitlementWentThrough(
         _ customerInfo: CustomerInfo,
-        file: FileString = #file,
+        file: FileString = #filePath,
+        filename: StaticString = #file,
```
ajpallares, line 30: Whereas this seems like duplication, the reason I kept these 2 parameters is to preserve the existing behavior.
Before: Nimble's `FileString` was a typealias for `StaticString`.
Now: v13.7.1 Nimble's `FileString` is a typealias for `String`, which does not match anymore with the `StaticString` that `XCTUnwrap` (in line 60) expects.

And I did `file: FileString = #filePath` instead of `file: FileString = #file` because that's the actual default value for `file` in Nimble now.
*

ajpallares, line 30: Whereas this seems like duplication, the reason I kept these 2 parameters is to preserve the existing behavior.
Before: Nimble's `FileString` was a typealias for `StaticString`.
Now: v13.7.1 Nimble's `FileString` is a typealias for `String`, which does not match anymore with the `StaticString` that `XCTUnwrap` (in line 60) expects.

And I did `file: FileString = #filePath` instead of `file: FileString = #file` because that's the actual default value for `file` in Nimble now.

Package.swift
```diff
@@ -37,7 +37,7 @@ let environmentVariables = ProcessInfo.processInfo.environment
 let shouldIncludeDocCPlugin = environmentVariables["INCLUDE_DOCC_PLUGIN"] == "true"
 
 var dependencies: [Package.Dependency] = [
-    .package(url: "https://github.com/quick/nimble", revision: "1f3bde57bde12f5e7b07909848c071e9b73d6edc"),
+    .package(url: "https://github.com/quick/nimble", exact: "13.7.1"),
```
facumenzella, line 40: I fixed it to commits because tags can be technically moved, but maybe it's too much  😅

ajpallares: Yeah, I'd say we should be fine fixing the exact version. It's more explicit and in this case it's for tests only. 
So potentially Nimble moving a tag would only be done to fix something that's broken

Tests/UnitTests/Purchasing/ProductsFetcherSK1Tests.swift
```diff
@@ -148,8 +148,9 @@ class ProductsFetcherSK1Tests: TestCase {
             completionCallCount += 1
         }
 
+        let timeout = (productsRequestResponseTime + .milliseconds(30)).nimbleInterval
```
facumenzella, line 151: I see that below you're doing `.seconds.nimbleInterval` when coming from `milliseconds`, just double-checking this is correct

facumenzella (approved):
Looks smoother than I expected

Tests/UnitTests/Purchasing/ProductsFetcherSK1Tests.swift
ajpallares: Actually this is a build error. Trying to fix lint I forgot to add the `seconds`. Good catch!

ajpallares:
@RCGitBot please test

PR #5103, opened 2025-05-16
Move RevenueCatUI CustomerCenter mocks from test target to RevenueCatUI

facumenzella:
### Motivation
Its not possible to mock `CustomerCenterPurchasesType` for previews, because the mock is in the test target.

### Description
This PR moves the code around from the test target to the main target, so we can mock previews.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Mocks/PurchaseInformationFixtures.swift
+++ b/RevenueCatUI/CustomerCenter/Mocks/PurchaseInformationFixtures.swift
@@ -13,7 +13,6 @@
 
 import Foundation
 @testable import RevenueCat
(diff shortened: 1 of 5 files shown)


RevenueCatUI/CustomerCenter/Mocks/PurchaseInformationFixtures.swift
```diff
@@ -13,7 +13,6 @@
 
 import Foundation
 @testable import RevenueCat
```
facumenzella, line 15:
```suggestion
import RevenueCat
```

ajpallares (approved):

PR #5104, opened 2025-05-16
Fix some flaky tests (Part 2)

ajpallares:
Continues the work of #5082

### Motivation
Flaky tests

### Description
Fixes some more flaky tests:

* In BackendPostDiagnosticsTests.swift, we use the async `expect` method from Nimble to avoid hitting the timeout (which happened sometimes in slow CI machines)
* `HTTPClientTests`' `testDiagnosticsHttpRequestPerformedTrackedOnError()` now stubs the request (as in the other tests), which should avoid hitting the timeout of `waitUntil`


tonidero (approved):
🫶 Thanks so much for fixing these!!

PR #5106, opened 2025-05-16
Remove ObservableObject from FeedbackSurveyData

facumenzella:
### Motivation
As part of cleaning up CustomerCenter, we've noticed that this shouldn't be a class nor observable object


ajpallares (approved):
That's right!

And I don't know if even the `var`s should be `let`s

PR #5108, opened 2025-05-16
Virtual Currency Support

fire-at-will:
## Description
This PR introduces virtual currencies to the iOS SDK. It's an exceptionally large PR, but all parts of it have been reviewed individually before.

It includes the following:

#### New Top-Level Functions
These functions allow you to fetch the virtual currencies for the current subscriber. The `invalidateVirtualCurrenciesCache` function allows developers to clear the cached virtual currencies, which will then force the refreshing of the virtual currencies from the backend the next time `virtualCurrencies()` is called.

```swift
class Purchases {
   @objc(virtualCurrenciesWithCompletion:)
    func getVirtualCurrencies(
        completion: @escaping @Sendable (VirtualCurrencies?, PublicError?) -> Void
    )

   func virtualCurrencies() async throws -> VirtualCurrencies

   @objc
   func invalidateVirtualCurrenciesCache()
}
```

#### New Objects
```swift
// A collection of virtual currencies returned by the backend
@objc(RCVirtualCurrencies) public final class VirtualCurrencies: NSObject {

    /// Dictionary of all VirtualCurrency(``VirtualCurrency``) objects keyed by virtual currency code.
    /// This dictionary can also be access through an index subscript on ``VirtualCurrencies``, e.g.
    /// `virtualCurrencies["VC_CODE"]`.
    @objc public let all: [String: VirtualCurrency]
}

@objc(RCVirtualCurrency)
public final class VirtualCurrency: NSObject {
    @objc public let balance: Int
    @objc public let name: String
    @objc public let code: String
    @objc public let serverDescription: String?
}
```

#### Example Usage
Swift:
```swift
purchases.invalidateVirtualCurrenciesCache()
let allVirtualCurrencies = try await purchases.virtualCurrencies().all
let gold = allVirtualCurrencies["GLD"]!
print("\(gold.name): \(gold.balance.description)"
```

Objective-C:
```objc
[p invalidateVirtualCurrenciesCache];

[purchases getVirtualCurrenciesWithCompletion: ^(RCVirtualCurrencies * _Nullable virtualCurrencies, NSError * _Nullable error) {
        RCVirtualCurrency *gold = virtualCurrencies[@"GLD"];
        NSLog(@"%@: %ld", gold.name, gold.balance);
}];
```

### Virtual Currency Support in Customer Center
If developers enable it in the RevenueCat dashboard, their virtual currency balances can be displayed in the Customer Center. Currently, it displays the Virtual Currency's code and balance, but in a future PR we'll update this to include the name as well.

No Subs View (1)


fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

RevenueCatUI/CustomerCenter/Views/RelevantPurchasesListView.swift
```diff
@@ -34,7 +35,7 @@ struct RelevantPurchasesListView: View {
     @Environment(\.navigationOptions)
     var navigationOptions
 
-    @StateObject
+    @ObservedObject
```
fire-at-will: This was required to fix a bug where the virtual currencies weren't being displayed in both RelevantPurchasesListView and SubscriptionDetailView after I made the changes in this PR.

Here’s what’s going on:

- In both RelevantPurchasesListView and SubscriptionDetailView, the views are initialized multiple times. The first time, virtualCurrencies haven’t been loaded yet, and are nil. The second time, virtualCurrencies have been loaded.
- However, the views are never updated with the newly passed in virtual currencies 😅
- This is because in the views’ initializers, we create a new view model instance, but the view model is stored as a @StateObject, which just stores one instance per scene. This meant that the second view model instance with the virtual currencies was being ignored by the view
- Changing this from @StateObject to @ObservedObject fixes this issue, since now the view will observe the second view model instead of the first one

We had some discussion on this in the original review here: https://github.com/RevenueCat/purchases-ios/pull/5151#discussion_r2140397852

Sources/Caching/DeviceCache.swift
```diff
@@ -378,8 +381,72 @@ class DeviceCache {
...
+        virtualCurrenciesLock.perform {
+            self.userDefaults.write {
+                Self.clearVirtualCurrenciesCacheLastUpdatedTimestamp($0, appUserID: appUserID)
+                $0.removeObject(forKey: CacheKey.virtualCurrencies(appUserID))
+            }
+        }
+    }
+
+    func clearVirtualCurrenciesCacheLastUpdatedTimestamp(appUserID: String) {
+        self.userDefaults.write {
```
tonidero, line 437: Should these also use virtualCurrenciesLock? I get that userDefaults already handles concurrency correctly, but I think there is a chance of this getting in between other operations that perform multiple userDefaults oprations like:
```
func cache(
    virtualCurrencies: Data,
    appUserID: String
)
```
I might be wrong if the `.write` lambda is performed atomically... But in that case, do we really need the `virtualCurrenciesLock` here?

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1297,6 +1308,43 @@ public extension Purchases {
...
+            } catch {
+                let publicError = NewErrorUtils.purchasesError(withUntypedError: error).asPublicError
+                OperationDispatcher.dispatchOnMainActor {
+                    completion(nil, publicError)
+                }
+            }
+        }
+    }
+
+    func virtualCurrencies() async throws -> VirtualCurrencies {
```
tonidero, line 1350: I've been wondering whether we might want to add a `CacheFetchPolicy` parameter to these methods, same as the `CustomerInfo` ones... Could be useful, but I'm ok adding that later as well.

Sources/Virtual Currencies/VirtualCurrencyManager.swift
```diff
@@ -0,0 +1,138 @@
...
+                    completion(result.mapError(\.asPurchasesError))
+                }
+            }
+
+            Logger.debug(Strings.virtualCurrencies.virtual_currencies_updated_from_network)
+            return VirtualCurrencies(from: virtualCurrenciesResponse)
+        } catch {
+            Logger.error(Strings.virtualCurrencies.virtual_currencies_updated_from_network_error(error))
+
+            throw error
```
tonidero, line 152: I see that if we fail fetching virtual currencies from backend, we fail the request, even if we have cached virtual currencies... I guess this differs from how it works by default for CustomerInfo in a few points but I'm wondering if we should still fallback to cache if it exists rather than failing. For example, this would fail for cases of no connectivity, and I think it might be good to use the currency we already have rather than failing... Wdyt?

Sources/Virtual Currencies/VirtualCurrencyManager.swift
```diff
@@ -0,0 +1,138 @@
...
+        backend: Backend,
+        systemInfo: SystemInfo
+    ) {
+        self.identityManager = identityManager
+        self.deviceCache = deviceCache
+        self.backend = backend
+        self.systemInfo = systemInfo
+    }
+
+    func virtualCurrencies() async throws -> VirtualCurrencies {
```
tonidero, line 43: I see we're not prefetching VC when configuring the SDK... I guess that might be ok and we can revisit if people need more immediate access to this data 👍

RevenueCatUI/CustomerCenter/ViewModels/VirtualCurrencyBalancesScreenViewModel.swift
```diff
@@ -0,0 +1,100 @@
...
+        purchasesProvider: CustomerCenterPurchasesType
+    ) {
+        self.viewState = viewState
+        self.purchasesProvider = purchasesProvider
+
+        #if DEBUG
+        self.isRunningInSwiftUIPreview = ProcessInfo.processInfo.environment["XCODE_RUNNING_FOR_PREVIEWS"] == "1"
+        #else
+        self.isRunningInSwiftUIPreview = false
+        #endif
```
facumenzella, line 49: nit: Maybe we could have an extension over `ProcessInfo`? I see that this is used in other places too

RevenueCatUI/CustomerCenter/ViewModels/VirtualCurrencyBalancesScreenViewModel.swift
```diff
@@ -0,0 +1,100 @@
...
+    ///
+    /// This method updates the view state to reflect the loading process and handles
+    /// any errors that occur during data fetching. On success, it transforms the
+    /// virtual currency data into a format suitable for displaying in the UI.
+    private func loadData() async {
+        self.viewState = .loading
+
+        do {
+            self.purchasesProvider.invalidateVirtualCurrenciesCache()
+            let virtualCurrencies = try await self.purchasesProvider.virtualCurrencies()
```
facumenzella, line 77: I see these methods are always called one after the other (CustomerCenterViewModel does the same). Why not having only one?

RevenueCatUI/CustomerCenter/Views/SubscriptionDetailView.swift
```diff
@@ -40,7 +41,7 @@ struct SubscriptionDetailView: View {
     @Environment(\.supportInformation)
     private var support
 
-    @StateObject
+    @ObservedObject
```
facumenzella: What's the reason for this one?

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrenciesScrollViewWithOSBackgroundSection.swift
```diff
@@ -0,0 +1,184 @@
...
+        virtualCurrencies: RevenueCat.VirtualCurrencies,
+        onSeeAllInAppCurrenciesButtonTapped: @escaping () -> Void
+    ) {
+        let sortedCurrencies = virtualCurrencies.all.map {
+            VirtualCurrencyBalanceListRow.RowData(
+                virtualCurrencyCode: $0.key,
+                balance: $0.value.balance
+            )
+        }
+            .sorted(by: { $0.balance > $1.balance })
```
facumenzella, line 93: This is also in `extractVirtualCurrencyBalanceData`, right? What about having an extension over `RevenueCat.VirtualCurrencies`

Sources/Networking/Operations/GetVirtualCurrenciesOperation.swift
```diff
@@ -0,0 +1,88 @@
...
+//
+//      https://opensource.org/licenses/MIT
+//
+//  GetVirtualCurrenciesOperation.swift
+//
+//  Created by Will Taylor on 6/9/25.
+
+import Foundation
+
+class GetVirtualCurrenciesOperation: CacheableNetworkOperation {
```
facumenzella:
```suggestion
final class GetVirtualCurrenciesOperation: CacheableNetworkOperation {
```

RevenueCatUI/CustomerCenter/ViewModels/VirtualCurrencyBalancesScreenViewModel.swift
fire-at-will: Same reason as the one [here](https://github.com/RevenueCat/purchases-ios/pull/5108#discussion_r2167663358)!

RevenueCatUI/CustomerCenter/Views/SubscriptionDetailView.swift
fire-at-will: Same reason as the one that we have [here](https://github.com/RevenueCat/purchases-ios/pull/5108#discussion_r2167663358)!

Sources/Caching/DeviceCache.swift
fire-at-will: You're right - it looks like `SynchronizedUserDefaults.write` performs atomically, so we don't really need the `virtualCurrenciesLock`. I've removed it in 96271d008070f6875b3052cf87303db788b917dd 👍

Sources/Purchasing/Purchases/Purchases.swift
fire-at-will: That was our thinking as well - if we find that we need it in the future, we can add it in! For the time being, we decided not to add it since the `CacheFetchPolicy` that we use for CustomerInfo is a bit complex.

RevenueCatUI/CustomerCenter/ViewModels/VirtualCurrencyBalancesScreenViewModel.swift
fire-at-will: When we're displaying virtual currencies in the Customer Center, we always want to make sure that the user sees their current balances, not a cached value, in case their balances have changed recently. 

We decided to omit a `Purchases.shared.virtualCurrencies() function that allows developers to force fetching for now for simplicity's sake, but we can definitely add one in the future if people request it!

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrenciesScrollViewWithOSBackgroundSection.swift
fire-at-will: We considered adding one there, along with other helpers (like currencies with a non-zero balance, currencies with a zero balance, etc.), but decided to hold off on adding them for now to keep the API's surface area as small as possible. We can certainly add them in later if folks ask for them!

If you'd like, I'd be happy to add this in as an internal extension in `RevenueCatUI` for us to use in Customer Center

Sources/Virtual Currencies/VirtualCurrencyManager.swift
fire-at-will: I think I could honestly go either way on it. 🤔

Here's the argument for the current behavior: If the SDK is making a backend request, that means that the cached virtual currencies are invalid for some reason (either stale or invalidated by the developer), in which case the developer could unknowingly display inaccurate VC balances to the user if we return cached VCs after a network failure.

This isn't a _huge_ deal for entitlements, as they don't typically change very often and are binary (you either do or don't have the entitlement). On the other hand, virtual currency balances can have many values >= 0, and can change much more often, like if:
- A developer grants currencies for each app open, or once per app open per day
- The developer could grant currencies each time the user watches an ad or performs an action in or out of the app
- The user could spend their currency multiple times an hour
In these cases, it could be problematic if the developer _thinks_ that they're showing the most recent balance when they're actually showing an old balance.

Happy to have a huddle to talk about this with you tomorrow if you'd like!

tonidero: Hmm I understand the reasoning, but feels to me that, since we don't provide a `CacheFetchPolicy` parameter, that means it would be impossible to use a cached value even if the developer wants to... An improvement we could do is to add "source" field in `VirtualCurrencies` that indicates whether the value comes from cache, stale cache, or network... That way the developer can decide what to do with it.

In any case, happy to discuss it more after yeah!

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1297,6 +1308,43 @@ public extension Purchases {
     }
 }
 
+// MARK: - Virtual Currencies
+#if !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+public extension Purchases {
+
+    @objc func virtualCurrencies(
```
tonidero: [Nothing to do here]

I've been wondering if it would make sense to add a listener of updates to VC in order to be consistent with CustomerInfo, but I don't think it's worth it considering we are not performing updates in the background ourselves. We can always revisit this in the future.

Tests/APITesters/AllAPITests/SwiftAPITester/VirtualCurrencyAPI.swift
```diff
@@ -0,0 +1,25 @@
...
+    let serverDescription: String? = virtualCurrency.serverDescription
+
+    // Ensure that we don't break the nullability of serverDescription.
+    // Just ensuring that it's a String? isn't enough since a String can be automatically casted
+    // to a String?, like so:
+    //
+    // let string1: String = ""
+    // let string2: String? = string1
+    var serverDescriptionNullabilityTest: String? = virtualCurrency.serverDescription
+    serverDescriptionNullabilityTest = nil
```
tonidero: Hmm I'm unsure how this test actually confirms that `virtualCurrency.serverDescription` still remains nullable... Can you give more details?

IMO, the best way to check this would be by creating an instance of a `VirtualCurrency` here... Wdyt?

Tests/UnitTests/Caching/DeviceCacheTests.swift
```diff
@@ -474,6 +483,292 @@ class DeviceCacheTests: TestCase {
         expect(self.deviceCache.isProductEntitlementMappingCacheStale) == true
     }
 
+    // MARK: - Virtual Currencies
```
tonidero, line 517: Not a blocker and can be done in the future! but I think we should try to split these huge test files into multiple files... These `MARK` sections are probably where I would start.

tonidero (commented):
Looking good! After we discuss the matter of using cached results on errors, I think this is ready!

Tests/UnitTests/Caching/DeviceCacheTests.swift
fire-at-will: Definitely! There are so many tests here 😅

Sources/Purchasing/Purchases/Purchases.swift
fire-at-will: Yeah, I think we'll probably add a listener in the future, but we haven't for now to ship a MVP. 

We'll either need to implement it with polling like we do for CustomerInfo today, or ideally, we could send a silent push notification from the backend to the SDK that instructs it to refresh the VCs so we can avoid polling.

Sources/Virtual Currencies/VirtualCurrencyManager.swift
fire-at-will: Yeah, it's definitely an optimization we could make in the future. Since a majority of apps won't use virtual currencies, we'll probably consider only prefetching if the developer has already called the `virtualCurrencies()` function so that we don't make the call for apps that aren't using it.

Tests/APITesters/AllAPITests/SwiftAPITester/VirtualCurrencyAPI.swift
fire-at-will: Yeah, I suppose that `virtualCurrency.serverDescription` could be a `String`, and that would still successfully cast to a `String`. 

Unfortunately, there's no public initializer for `VirtualCurrency` besides `init(from:)`, so I think I'll just remove this for now and keep the `let serverDescription: String? = virtualCurrency.serverDescription`

fire-at-will: Removed in 0dbdb876ddd87244003c80bd625cad220fa2e708

RevenueCatUI/CustomerCenter/Views/RelevantPurchasesListView.swift
facumenzella: I think the main risk here is loosing state (like alerts, loading, etc)
I am not completely sure this is the way to go, but I've went this path before and for sure fixes any reloading issues. What we've done before is to rely on `CustomerCenterViewModel` (used here as a `ObservedObject` but declared as a `StateObject` in the other view. In this case, we could just pass a `VirtualCurrenciesModel` declared in `CustomerCenterView` as a `StateObject`.

I don't mind moving forward with it, but it might introduce unwanted side-effects. 
I'll test manually to make sure nothing's broken 💪

fire-at-will:
@RCGitBot please test

RevenueCatUI/CustomerCenter/Views/RelevantPurchasesListView.swift
fire-at-will: I think that approach makes sense and sounds safer. I'll code that up in a separate PR going into this one and send it over to you for review 👍

fire-at-will: PR here: https://github.com/RevenueCat/purchases-ios/pull/5347

It actually wound up being easier to reuse the virtual currencies inside the `CustomerCenterViewModel`, so I did that instead of creating a `VirtualCurrenciesModel`

Sources/Virtual Currencies/VirtualCurrencyManager.swift
fire-at-will: Toni and I discussed this in Slack! We decided to move forward with adding a `cachedVirtualCurrencies: VirtualCurrencies?` property that developers can use to get the cached virtual currencies if they need it.

The code for that was implemented here and was merged into this PR: https://github.com/RevenueCat/purchases-ios/pull/5345

tonidero (approved):
Gave it a quick rereview and I think it's looking good! Let's :shipit:

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1311,6 +1321,47 @@ public extension Purchases {
...
+            } catch {
+                let publicError = NewErrorUtils.purchasesError(withUntypedError: error).asPublicError
+                OperationDispatcher.dispatchOnMainActor {
+                    completion(nil, publicError)
+                }
+            }
+        }
+    }
+
+    var cachedVirtualCurrencies: VirtualCurrencies? {
```
ajpallares: Should this be `@objc` as well?

fire-at-will:
@RCGitBot please test

Sources/Purchasing/Purchases/Purchases.swift
fire-at-will: Added `@objc` in 2af58f1fd7a14aac5df3907c750dae76ed5d11e3 👍

ajpallares: And, as a follow up, we should probably also add this `cachedVirtualCurrencies` to ObjCApiTester

fire-at-will:
@RCGitBot please test

Sources/Purchasing/Purchases/Purchases.swift
fire-at-will: Added it here! :D 

 https://github.com/RevenueCat/purchases-ios/blob/e6948eaa0214f607ed84da3d04fadda2933a2761/Tests/APITesters/AllAPITests/ObjcAPITester/RCPurchasesAPI.m#L212

ajpallares (approved):
Took one more look and I think this is ready! 🚀 
Amazing job Will!

fire-at-will:
@RCGitBot please test

PR #5114, opened 2025-05-19
fix auto dismiss of in-app browser web checkout, to happen only after…

ajpallares:
### Description

This PR fixes the auto dismiss logic for web purchase links in paywalls that use the in-app browser. Now, the auto dismiss only happens after the paywall itself dismisses.

# Important

Note that the **in-app browser web purchase links are currently disabled in RevenueCat**, since it has not been explicitly allowed by Apple by the time of writing this PR notes. But now, the iOS SDK is ready to support it for when/in case Apple allows this in the future

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
@@ -33,6 +33,7 @@ struct PurchaseButtonComponentView: View {
     private var purchaseHandler: PurchaseHandler
 
     @State private var inAppBrowserURL: URL?
+    @State private var inAppBrowserDidDisappearCompletion: (() -> Void)?
 
     private let viewModel: PurchaseButtonComponentViewModel
     private let onDismiss: () -> Void
@@ -79,6 +80,11 @@ struct PurchaseButtonComponentView: View {
         #if canImport(SafariServices) && canImport(UIKit)
         .sheet(isPresented: .isNotNil(self.$inAppBrowserURL)) {
             SafariView(url: self.inAppBrowserURL!)
+        }
+        .onChange(of: self.inAppBrowserURL) { inAppBrowserURL in
+            guard inAppBrowserURL == nil else { return }
+            inAppBrowserDidDisappearCompletion?()
+            inAppBrowserDidDisappearCompletion = nil
         }
         #endif
     }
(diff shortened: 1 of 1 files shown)


ajpallares:
In this other similar PR #5083, which I closed because the changes were required in `PurchaseButtonComponent` and not in `ButtonComponent`, I had added a more complex logic for the auto dismiss functionality. The idea was to only dismiss the paywall if auto dismiss was true and if the user had new active RC Billing entitlement. I think we don't need to complicate things for now as it may make the whole auto dismiss functionality more confusing for developers.

Let me know if you thing otherwise!

RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
```diff
@@ -139,7 +145,13 @@ struct PurchaseButtonComponentView: View {
                            inAppBrowserURL: self.$inAppBrowserURL)
 
         if launchWebCheckout.autoDismiss {
-            self.onDismiss()
+            if case .inAppBrowser = method {
+                self.inAppBrowserDidDisappearCompletion = {
+                    self.onDismiss()
```
tonidero, line 150: Hmm I wonder... in this case, would it make sense to do something like:
- Fetching new CustomerInfo from backend (display some loading state in the meantime).
- If it has a new purchase, then dismiss, otherwise keep the paywall displaying.

Mostly thinking that a purchase could have been cancelled here... Not sure if there is some nice way to communicate this info from the web view directly though... Also, this might be acceptable for now though... Just possibly could cause some weird behavior when closing the web

tonidero (approved):
I think this makes sense for now, but just a comment about a possible improvement that could be tackled separately.

RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
ajpallares: Thanks for your suggestion @tonidero!

Exactly! Actually, that is what I meant in [my comment above](https://github.com/RevenueCat/purchases-ios/pull/5114#issuecomment-2890009018). In fact, I did implement something similar in #5083 (now discarded).

Now, I agree that we could be smarter about this with your proposed approach. However, should that be the expected behavior for "Auto dismiss" (i.e. "Auto dismiss == auto dismiss when purchase successfully made")?

As you said, I think we should get this fix out as is (otherwise, in-app browser web checkout would be broken as the the whole paywall would get dismissed when presenting the in-app browser in this case). Then, we can discuss about whether or not we want to be more clever about when to auto dismiss.

cc @joshdholtz

tonidero: Ah right, that makes sense! I'm ok merging this for now yeah and we can modify  later if needed.

PR #5115, opened 2025-05-19
Show account details in active subscription list

facumenzella:
### Motivation
We are sunsetting purchase history in favor of one list with all the purchases.

### Description
1. Move account details to active list
2. Remove account details from purchase history

### DIFF
--- a/RevenueCatUI/CustomerCenter/ViewModels/ActiveSubscriptionsListViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/ActiveSubscriptionsListViewModel.swift
@@ -27,14 +27,22 @@ final class ActiveSubscriptionsListViewModel: BaseManageSubscriptionViewModel {
     @Published
     private(set) var activePurchases: [PurchaseInformation] = []
 
+    let originalAppUserId: String
+    let originalPurchaseDate: Date?
+
     init(
         screen: CustomerCenterConfigData.Screen,
         actionWrapper: CustomerCenterActionWrapper,
         activePurchases: [PurchaseInformation] = [],
+        originalAppUserId: String,
+        originalPurchaseDate: Date?,
         refundRequestStatus: RefundRequestStatus? = nil,
         purchasesProvider: CustomerCenterPurchasesType,
         loadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType? = nil) {
             self.activePurchases = activePurchases
+            self.originalAppUserId = originalAppUserId
+            self.originalPurchaseDate = originalPurchaseDate
+
             super.init(
                 screen: screen,
                 actionWrapper: actionWrapper,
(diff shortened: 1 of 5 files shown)


RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -107,6 +117,7 @@ import RevenueCat
         self.actionWrapper = actionWrapper
         self.purchasesProvider = purchasesProvider
         self.customerCenterStoreKitUtilities = customerCenterStoreKitUtilities
+        self.customerInfo = nil
```
tonidero, line 120: Hmm this is always set to nil right? Seems they are unused?

RevenueCatUI/CustomerCenter/ViewModels/ActiveSubscriptionsListViewModel.swift
```diff
@@ -48,12 +56,16 @@ final class ActiveSubscriptionsListViewModel: BaseManageSubscriptionViewModel {
     // Used for Previews
     convenience init(
         screen: CustomerCenterConfigData.Screen,
-        activePurchases: [PurchaseInformation] = []
+        activePurchases: [PurchaseInformation] = [],
+        originalAppUserId: String = UUID().uuidString,
+        originalPurchaseDate: Date? = Date()
```
tonidero, line 61: Hmm I wonder if these should have these defaults... I think it would be preferable to make them nullable and default to nil? Or make them mandatory parameters so we don't forget passing them.

tonidero (commented):
A couple questions but looking good!

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
facumenzella: I missed to push one last commit 🙃 
There it is, it's assigned at the beginning of `loadPurchaseInformation`

RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionsListView.swift
```diff
@@ -201,6 +249,13 @@ struct ActiveSubscriptionsListView: View {
             .disabled(self.viewModel.loadingPath != nil)
         }
     }
+
+    private var dateFormatter: DateFormatter {
+        let formatter = DateFormatter()
```
tonidero, line 254: I wonder if we should pass the chosen locale to the date formatter, to make sure it matches the locale used in the rest of CustomerCenter, in case the device locale is not supported...

tonidero (approved):
Looking good! Just a question, but can potentially be tackled in a follow-up PR if needed.

RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionsListView.swift
facumenzella: That's actually a great idea. 

We are not passing around the locale, but it's an easy tweak. In addittion, PurchaseInformation has an injected formatter, so I can use that one + locale to make sure it's correct 👍

PR #5118, opened 2025-05-19
Introduce purchase card badges

facumenzella:
### Motivation
The card will reflect some special states using badges

### Description
Highlight cancelled and free trial state.

### Upcoming
- Expired state


tonidero (approved):
LGTM!

PR #5119, opened 2025-05-19
Fix paywall auto dismiss from bottom sheet

ajpallares:
### Motivation
If the auto dismiss of a paywall happened from the paywall's bottom sheet, only the bottom sheet would get dismissed, and not the whole paywall.

### Description
This PR fixes this bug so that the whole paywall gets dismissed if it gets auto dismissed from the bottom sheet (i.e. both the bottom sheet and the paywall itself get dismissed)

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Button/BottomSheetView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Button/BottomSheetView.swift
@@ -44,6 +44,7 @@ struct SheetViewModel: Equatable {
 struct BottomSheetOverlayModifier: ViewModifier {
     @Binding var sheetViewModel: SheetViewModel?
     let safeAreaInsets: EdgeInsets
+    let onDismiss: (() -> Void)?
 
     @State private var parentHeight: CGFloat?
 
@@ -100,6 +101,7 @@ struct BottomSheetOverlayModifier: ViewModifier {
                         viewModel: sheetViewModel.sheetStackViewModel,
                         onDismiss: {
                             self.sheetViewModel = nil
+                            onDismiss?()
                         },
                         additionalPadding: self.safeAreaInsets
                     )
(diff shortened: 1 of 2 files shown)


tonidero (approved):
Ohhh makes sense. Thanks for fixing this!

RevenueCatUI/Templates/V2/Components/Button/BottomSheetView.swift
```diff
@@ -100,6 +101,7 @@ struct BottomSheetOverlayModifier: ViewModifier {
                         viewModel: sheetViewModel.sheetStackViewModel,
                         onDismiss: {
                             self.sheetViewModel = nil
+                            onDismiss?()
```
tonidero, line 104: Hmm actually... if someone adds a close button to the sheet content, we probably would only want to dismiss the sheet I guess? This seems like a behavior breaking change as well....

ajpallares: Great catch Toni! you are probably right!
It's confusing though, in the sense that close buttons in paywalls are supposed to close the paywall BUT I totally get that people would want to add the close button to close the sheet only 🤔

tonidero: Yeah... I'm not sure of the best way to fix this... We could differentiate between "close entire paywall" and "close this screen" events I guess, but seems a bit more work involved...

ajpallares: Yep. I'd also argue that this all should be transparent in the dashboard. I.e. we should implement this in the SDK: so that a close button inside a bottom sheet only closes the sheet (as it is currently) but the auto dismiss dismisses the whole paywall (the current auto dismiss logic in all places other than in the sheet).

PR #5121, opened 2025-05-19
Include PurchaseInformationCardView in SubscriptionDetail

facumenzella:
### Motivation
Replace old card used in the detail with the new one.

### Description
We will use `PurchaseInformationCardView` in the purchase detail. This removes a bunch of code and simplifies the UI a lot.
I am also adding `additionalInfo` to display the refund status

| Header | Header |
|--------|--------|
|  Screenshot 2025-05-19 at 17 29 08 | Screenshot 2025-05-19 at 17 29 00 |


RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionsListView.swift
```diff
@@ -287,7 +287,7 @@ struct ActiveSubscriptionsListView_Previews: PreviewProvider {
                 ActiveSubscriptionsListView(
                     viewModel: ActiveSubscriptionsListViewModel(
                         screen: warningOffMock.screens[.management]!,
-                        originalAppUserId: UUID().uuidString,
+                        originalAppUserId: "originalAppUserId",
```
ajpallares, line 290: There's another `UUID().uuidString` that could use the same fix original app user ID, I think

RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
```diff
@@ -85,58 +105,104 @@ struct PurchaseInformationCardView: View {
...
                         .foregroundStyle(.secondary)
-                        .font(Font.system(size: 12, weight: .bold))
+                        .padding(.bottom, 4)
+                        .frame(maxWidth: .infinity, alignment: .leading)
+                        .multilineTextAlignment(.leading)
                 }
+                .padding()
+                .background(Color(colorScheme == .light
+                                  ? UIColor.secondarySystemBackground
+                                  : UIColor.systemBackground))
```
ajpallares, line 183: Visually I think this is a bit weird in Dark mode

image

Because the background of the additional info view is the same color as the whole view background color, _I think_ it doesn't look great. And I think it's not visually clear either that the additional info refers to the  Product above. I'm not really sure on how to improve this.

Perhaps using `tertiarySystemBackground`? I'm not sure, because it would appear as the additional information is highlighted.

image

Another possibility would be to add a stroke in dark mode, something like this
image

This would be the code in `ActiveSubscriptionsListView`, although I'm not very convinced either.

```swift
PurchaseInformationCardView(
    purchaseInformation: purchase,
    localization: localization
)
                  ? UIColor.systemBackground
                  : UIColor.secondarySystemBackground))
.overlay {
    if colorScheme == .dark {
        RoundedRectangle(cornerRadius: 10)
            .stroke(Color(UIColor.tertiarySystemBackground))
    }
}
.cornerRadius(10)
.padding(.horizontal)
```

ajpallares (commented):
Many big improvements! Looking great! I have some small comments. Only one is I think big enough not to approve yet (about the additional information on the product card in dark mode)

RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
facumenzella: Oh nice catch. I think it's good enough for now. 
The specs are not clear in dark mode, so I tried to take the same approach we took in other places. We can move forward like this and iterate it in detail once the whole feature is ready.

ajpallares (approved):

RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
facumenzella: I've moved that code to the view itself. It will always be used in the same context, so makes sense to keep it there for now.

ajpallares: Nice! 💪

PR #5122, opened 2025-05-19
Add DEBUG check to SDK Health API tests

ajpallares:
### Motivation
Some of the new APIs for the new check SDK health APIs are `DEBUG` only. This is causing the API tests added in #5117 to fail because the `DEBUG` check is missing when testing these new APIs.

### Description
This PR adds the `#if DEBUG` check to the corresponding API tests.


ajpallares:
@RCGitBot please test

polpielladev (approved):

polpielladev:
Looks good to me! 👍  Sorry I missed this completely on my previous PR, thanks for fixing it! 🚀

PR #5123, opened 2025-05-19
Introduce ActiveSubscriptionButtonsView to use it inside a scrollview

facumenzella:
### Motivation
We're using `List` instead of `ScrollView`, and we want to do the opposite. 

### Description
Create a duplicate ButtonsView inspired in `ManageSubscriptionsButtonsView` so I can keep this PR as small as possible. 

After merging this, I can go with https://github.com/RevenueCat/purchases-ios/pull/5121 and use `PurchaseInformationCardView` in the same way its used in `ActiveSubscriptionsListView` (inside a scrollview)

I'll delete the old `ManageSubscriptionsButtonsView` once this is places everywhere.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionButtonsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionButtonsView.swift
@@ -35,7 +35,7 @@ struct ActiveSubscriptionButtonsView: View {
                 AsyncButton(action: {
                     await self.viewModel.handleHelpPath(
                         path,
-                        wihtActiveProductId: viewModel.purchaseInformation?.productIdentifier)
+                        withActiveProductId: viewModel.purchaseInformation?.productIdentifier)
                 }, label: {
                     if self.viewModel.loadingPath?.id == path.id {
                         TintedProgressView()
--- a/RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionButtonsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionButtonsView.swift
@@ -1,0 +1,63 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//      https://opensource.org/licenses/MIT
+//
+//  ActiveSubscriptionButtonsView.swift
+//
+//  Created by Facundo Menzella on 19/5/25.
+
+import Foundation
+import RevenueCat
+import SwiftUI
+
+#if os(iOS)
+
+@available(iOS 15.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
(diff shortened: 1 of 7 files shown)


RevenueCatUI/CustomerCenter/Views/CompatibilityLabeledContent.swift
```diff
@@ -99,6 +99,17 @@ extension CompatibilityLabeledContent where Label == Text {
     }
 }
 
+@available(iOS 15.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+extension CompatibilityLabeledContent where Label == Text, Content == EmptyView {
+    init(_ label: String) {
```
ajpallares, line 107: Why not just use `Text(label)` in these cases?

RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionButtonsView.swift
```diff
@@ -0,0 +1,63 @@
...
+    @ObservedObject
+    var viewModel: BaseManageSubscriptionViewModel
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: 0) {
+            ForEach(self.viewModel.relevantPathsForPurchase, id: \.id) { path in
+                AsyncButton(action: {
+                    await self.viewModel.handleHelpPath(
+                        path,
+                        wihtActiveProductId: viewModel.purchaseInformation?.productIdentifier)
```
ajpallares, line 38: Not from this PR, but we might as well correct now
```suggestion
                        withActiveProductId: viewModel.purchaseInformation?.productIdentifier)
```

ajpallares (approved):
Looks great! Just a couple of small comments

RevenueCatUI/CustomerCenter/Views/CompatibilityLabeledContent.swift
facumenzella: I find this super convenient, to make the SwiftUI code less messy. 

As an example, this now can be used like:
```swift
CompatibilityLabeledContent("Username") {
    Text("john@example.com")
}
```
instead of:
```swift
CompatibilityLabeledContent(label: { Text("Username") }, content: { Text("john@example.com") })
```

ajpallares: Right, but this case of the `init(_ label: String)` where Content is `EmptyView`, you'd be using 

```swift
CompatibilityLabeledContent("Something")
```

whereas, in reality, it's just a `Text("Something")`. So you might as well just use `Text` if you have no content anyway

facumenzella: I misunderstood your comment 🤓 

`CompatibilityLabeledContent` takes the whole width, whereas `Text` needs the `.frame(maxWidth: .infinity)`.

ajpallares: Oh right. Makes sense. Thank you for the clarification!

PR #5125, opened 2025-05-20
Fixed locale in RevenueCatUI test data

ajpallares:
### Motivation
Some tests don't pass locally because thy depend on `Locale.current`. This means that some tests expect prices in USD while running them locally result in EUR prices for some of us working from the EU.

### Description
* Use fixed en_US locale in RevenueCatUI test data for consistent localization in tests (`TestData.locale`).
* In addition, to remove duplicated test code, PaywallsTester now uses the same test packages and products in RevenueCatUI's `TestData`.


RevenueCatUI/Data/TestData.swift
```diff
@@ -825,12 +825,7 @@ enum TestData {
 
     #endif
 
-    #if os(watchOS)
-    // `Locale.current` in watchOS produces `en_001` when running tests
     static let locale: Locale = .init(identifier: "en_US")
-    #else
-    static let locale: Locale = .current
```
tonidero, line 832: I guess this might make it slightly harder to test other locales in paywall tester, which could be useful if having a conference in a different country... But TBH, I think this is totally fine 👍

tonidero (approved):
Nice! Makes a lot of sense to me, thanks for doing this!! 🫶

RevenueCatUI/Data/TestData.swift
ajpallares: Yeah, you're right. Perhaps we should add some kind of locale selection in the app? 

In any case, as it is right now, there were inconsistencies in the preconfigured v1 paywalls (e.g. showing a price in $ + the price per month in €).

That said, this change would not affect the paywalls downloaded from the server. It only affects the v1 paywalls in the first tab, so I _think_ it should be fine, yes.

PR #5126, opened 2025-05-20
Show other purchases in Purchases List

facumenzella:
### Motivation
This PR adds other type of purchases to the ActiveList for customer center.

### Description
1. Rename ActiveList to RelevantPurchasesList
2. Compute nonSubscriptions and show the first ones (what's relevant here might change in the future, but this is a good start)


RevenueCatUI/CustomerCenter/ViewModels/RelevantPurchasesListViewModel.swift
```diff
@@ -22,24 +22,33 @@ import SwiftUI
...
 @available(watchOS, unavailable)
 @MainActor
-final class ActiveSubscriptionsListViewModel: BaseManageSubscriptionViewModel {
+final class RelevantPurchasesListViewModel: BaseManageSubscriptionViewModel {
 
     @Published
     private(set) var activePurchases: [PurchaseInformation] = []
 
+    @Published
+    private(set) var nonSubscriptionPurchases: [PurchaseInformation] = []
```
ajpallares, line 31: Would it make sense to rename `activePurchases` to `activeSubscriptionPurchases`? 
Otherwise, it could be confusing as it might imply that it contain all `nonSubscriptionPurchases`.

Also, for consistency with naming in other files, should `nonSubscriptionPurchases` be called `activeNonSubscriptionPurchases`? Or perhaps it contains non-active nonsubscriptions?

RevenueCatUI/CustomerCenter/Views/RelevantPurchasesListView.swift
```diff
@@ -113,71 +122,104 @@ struct ActiveSubscriptionsListView: View {
...
+            LazyVStack(spacing: 0) {
+                if viewModel.isEmpty {
                     CompatibilityContentUnavailableView(
                         self.viewModel.screen.title,
                         systemImage: "exclamationmark.triangle.fill",
-                        description: Text(self.viewModel.screen.subtitle ?? fallbackDescription)
+                        description: Text(self.viewModel.screen.subtitle ?? localization[.tryCheckRestore])
+                    )
+                    .padding(.horizontal)
+                    .padding(.vertical)
```
ajpallares, line 134: I think this is the same as
```suggestion
                    .padding()
```
unless you need different values for horizontal vs vertical

RevenueCatUI/CustomerCenter/Views/ScrollViewSection.swift
```diff
@@ -0,0 +1,42 @@
...
+@available(iOS 15.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+struct ScrollViewSection<Content: View>: View {
+    @Environment(\.colorScheme)
+    private var colorScheme
+
+    let title: String
+    @ViewBuilder var content: () -> Content
```
ajpallares, line 27: I think this could be a `let`
```suggestion
    @ViewBuilder let content: () -> Content
```

ajpallares (approved):
Looks good I think! Some comments, mostly about naming

RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
```diff
@@ -236,7 +236,9 @@ private extension Array<CustomerCenterConfigData.HelpPath> {
         for purchaseInformation: PurchaseInformation?
     ) -> [CustomerCenterConfigData.HelpPath] {
         guard let purchaseInformation else {
-            return self
+            return filter {
+                $0.type == .missingPurchase
+            }
```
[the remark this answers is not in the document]
facumenzella: I noticed that if there's no active purchase we were showing all the actions, which is not correct. 
In the context of the list, we should only show `restore purchases`, and the detail should have all the relevant actions.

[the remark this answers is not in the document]
ajpallares: Okay. Thanks for explaining!

RevenueCatUI/CustomerCenter/ViewModels/RelevantPurchasesListViewModel.swift
facumenzella: Make sense 👍

PR #5127, opened 2025-05-20
Update margins and copies for SubscriptionList

facumenzella:
### Motivation
Doing the final testing round, I'm adjusting a few margins and using the final copies defined in Figma.

### Description
- New copies
- Tweak margins
- Update previews to show different scenarios.


ajpallares (approved):
Makes sense!

PR #5128, opened 2025-05-20
Show list if all purchases together are more than one

facumenzella:
### Motivation
We need to show the list if there's at least more than one purchase. 

### Description
The old code checks only for subscription. This fix now takes into account both, and adds tests for it.

### DIFF
--- a/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
@@ -77,6 +77,10 @@ import RevenueCat
         !activeSubscriptionPurchases.isEmpty || activePurchase != nil || !activeNonSubscriptionPurchases.isEmpty
     }
 
+    var shouldShowList: Bool {
+        activeSubscriptionPurchases.count + activeNonSubscriptionPurchases.count > 1
+    }
+
     var  originalAppUserId: String {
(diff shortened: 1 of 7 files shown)


RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -140,6 +144,18 @@ import RevenueCat
...
+        activeNonSubscriptionPurchases: [PurchaseInformation],
+        configuration: CustomerCenterConfigData
+    ) {
+        self.init(actionWrapper: CustomerCenterActionWrapper(legacyActionHandler: nil))
+        self.activeSubscriptionPurchases = activeSubscriptionPurchases
+        self.activeNonSubscriptionPurchases = activeNonSubscriptionPurchases
+        self.configuration = configuration
+        self.state = .success
+    }
+
```
ajpallares, line 158: Total nit: if this is only used in tests, perhaps it should be declared in the tests target?

ajpallares (approved):
Also makes sense! Just one nit!

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
facumenzella: Ok I kind of like that. I think there's a bunch of initializers that can be moved. I'll apply that in another PR to keep it clean.

PR #5131, opened 2025-05-20
Feat/addprice nonsubscription

facumenzella:
### Motivation
While testing non-consumables for CustomerCenter, I realized the price was missing.

### Description
Adds price paid for `Transaction` and `NonSubscriptionTransaction`, and showcases it in the PurchaseInformationCardView.


ajpallares (approved):
Makes sense!

PR #5133, opened 2025-05-21
Filter changePlans path for lifetime purchases in CustomerCenter

facumenzella:
### Motivation
While testing non-consumables for CustomerCenter, I realized change plans path makes no sense for lifetime.

### Description
Add filter for it, and basic test.


ajpallares (approved):
Makes total sense, yes!

PR #5134, opened 2025-05-21
Dont show `see all purchases` button if there's nothing else to show

facumenzella:
### Motivation
In the old and new version, the see all purchases was being shown even if there was no more new data. This tweaks that, and takes into account what's being shown.

### Description
Add extension to `CustomerInfo` to decide if the button should be shown.


ajpallares (approved):
Also makes sense 👍

PR #5135, opened 2025-05-21
Change to use new endpoint to fetch web product info

tonidero:
### Description
We ended up changing to use a new endpoint instead of relying on the existing offerings + products endpoints to fetch product info. This changes the implementation to use this new endpoint. This is currently unused, so it's not a problem to change it.

### DIFF
--- a/Sources/Networking/WebBillingHTTPRequestPath.swift
+++ b/Sources/Networking/WebBillingHTTPRequestPath.swift
@@ -48,9 +48,8 @@ extension HTTPRequest.WebBillingPath: HTTPRequestPath {
 
     var relativePath: String {
         switch self {
-        case let .getWebProducts(appUserId, productIds):
-            let productIdsQuery = productIds.map(\.trimmedAndEscaped).joined(separator: "&id=")
-            return "/rcbilling/v1/subscribers/\(appUserId.trimmedAndEscaped))/products?id=\(productIdsQuery)"
+        case let .getWebProducts(appUserID):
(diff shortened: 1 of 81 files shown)


tonidero:
Please do not be scare by the diff 🙏. Most of the changes are automated backend snapshot tests

Sources/Networking/WebBillingHTTPRequestPath.swift
```diff
@@ -48,9 +48,8 @@ extension HTTPRequest.WebBillingPath: HTTPRequestPath {
 
     var relativePath: String {
         switch self {
-        case let .getWebProducts(appUserId, productIds):
-            let productIdsQuery = productIds.map(\.trimmedAndEscaped).joined(separator: "&id=")
-            return "/rcbilling/v1/subscribers/\(appUserId.trimmedAndEscaped))/products?id=\(productIdsQuery)"
+        case let .getWebProducts(appUserID):
+            return "/rcbilling/v1/subscribers/\(appUserID.trimmedAndEscaped)/offering_products"
```
facumenzella, line 52: unrelated question, what's the reasoning for being `v1` or `v2`? 🤔

facumenzella (approved):
Looks good!

Sources/Networking/WebBillingHTTPRequestPath.swift
tonidero: Well, I believe all SDK endpoints currently point to V1... but we do have a V2 for public facing APIs mostly... In any case, it was done this way to support versioning, but we haven't really versioned this endpoints much until now (it's more painful so we've been trying to avoid it if possible)

PR #5143, opened 2025-05-21
Add ownership type to PurchaseInformation

facumenzella:
### Motivation
The long term goal is to delete `PurchaseInfo` from `PurchaseHistory`, in favor of using `PurchaseInformation`. This info is not available for PurchaseInformation right now.

### Description
- Add `ownershipType` to `PurchaseInformation`


tonidero (approved):
I think this looks good!

PR #5146, opened 2025-05-21
Set environmentObject for CustomerCenterViewModel when using an external navstack

facumenzella:
### Motivation
When using an external navigation stack, the customer center viewmodel is not present, so we need to explicitly set it.

### Description
Set the environment object for pushed views, so RestoreAlert has it available:

```swift
        .overlay {
            RestorePurchasesAlert(
                isPresented: $showRestoreAlert,
                actionWrapper: actionWrapper
            )
        }
```

```swift
struct RestorePurchasesAlert: View {
    @EnvironmentObject private var customerCenterViewModel: CustomerCenterViewModel
    // more things
}
``


ajpallares (approved):
Much cleaner 👌

PR #5149, opened 2025-05-21
Fix Customer Center VC Navigation Warnings

fire-at-will:
### Motivation
In the current implementation of the virtual currency feature in the Customer Center, SwiftUI logs the following warning in the console when the `VirtualCurrenciesListSection` and `VirtualCurrenciesScrollViewWithOSBackgroundSection` is displayed:

```
Do not put a navigation destination modifier inside a "lazy” container, like 
`List` or `LazyVStack`. These containers create child views only when 
needed to render on screen. Add the navigation destination modifier 
outside these containers so that the navigation stack can always see 
the destination. There's a misplaced 
`navigationDestination(isPresented:destination:)` modifier presenting 
`VirtualCurrencyBalancesScreen`. It will be ignored in a future release.
```

This is because the navigationDestination is inside the section view, which is then nested inside a `List` or `LazyVStack`, respectively. This PR prevents this warning from being displayed by extracting the navigation logic outside of the Virtual Currency Sections themselves and outside of the parent `List` or `LazyVStack`.

### Description
- Hoists navigation logic out of the virtual currency list sections above the `List` or `LazyVStack` to prevent the warning from being logged
- Adds an automated test
- Adds documentation so consumers of these views know that they're responsible for implementing the navigation logic.

### Testing
- Added an automated test
- Manually verified that the warning is no longer being logged

### DIFF
--- a/RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
@@ -57,6 +57,9 @@ class BaseManageSubscriptionViewModel: ObservableObject {
     var purchaseInformation: PurchaseInformation?
 
     @Published
+    var showAllInAppCurrenciesScreen: Bool = false
+
+    @Published
     private(set) var refundRequestStatus: RefundRequestStatus?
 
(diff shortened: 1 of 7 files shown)


facumenzella (approved):
Makes sense 👍

PR #5151, opened 2025-05-21
[Virtual Currencies] Break VC Balances out of CustomerInfo

fire-at-will:
### Motivation
We've decided to break virtual currencies out of CustomerInfo for scalability reasons. This PR removes `virtualCurrencies` from `CustomerInfo`, and will replace it with a new top-level `virtualCurrencies()` function on the `Purchases` object that uses a new dedicated backend endpoint.

### Overview
This PR:
- **Breaking**: Removes virtual currencies from `CustomerInfo`
- **Breaking**: Renames `VirtualCurrencyInfo` struct to `VirtualCurrency`
- Creates new top-level functions and objects to allow developers to fetch virtual currencies' balances and metadata
- Caches virtual currency network requests for up to 5 minutes
- Adds automated tests + API tests
- Updates the iOS Customer Center <> Virtual Currencies integration to use the new APIs

> Note: This PR represents a breaking change, but since the APIs that are being broken have only been released in beta SDK versions, we can merge + ship this without a major release.

#### New Top-Level Functions
These functions allow you to fetch the virtual currencies for the current subscriber. The `invalidateVirtualCurrenciesCache` function allows developers to clear the cached virtual currencies, which will then force the refreshing of the virtual currencies from the backend the next time `virtualCurrencies()` is called.

```swift
class Purchases {
   @objc(virtualCurrenciesWithCompletion:)
    func virtualCurrencies(
        completion: @escaping @Sendable (VirtualCurrencies?, PublicError?) -> Void
    )

   func virtualCurrencies() async throws -> VirtualCurrencies

   @objc
    func invalidateVirtualCurrenciesCache()
}
```

#### New Objects
```swift
// A collection of virtual currencies returned by the backend
@objc(RCVirtualCurrencies) public final class VirtualCurrencies: NSObject {

    /// Dictionary of all VirtualCurrency(``VirtualCurrency``) objects keyed by virtual currency code.
    /// This dictionary can also be access through an index subscript on ``VirtualCurrencies``, e.g.
    /// `virtualCurrencies["VC_CODE"]`.
    @objc public let all: [String: VirtualCurrency]
}

@objc(RCVirtualCurrency)
public final class VirtualCurrency: NSObject {
    @objc public let balance: Int
    @objc public let name: String
    @objc public let code: String
    @objc public let serverDescription: String?
}
```

#### Example Usage
Swift:
```swift
let allVirtualCurrencies = try await purchases.virtualCurrencies().all
let gold = allVirtualCurrencies["GLD"]!
print("\(gold.name): \(gold.balance.description)"
```

Objective-C:
```objc
[p invalidateVirtualCurrenciesCache];

[purchases virtualCurrenciesWithCompletion: ^(RCVirtualCurrencies * _Nullable virtualCurrencies, NSError * _Nullable error) {
        RCVirtualCurrency *gold = virtualCurrencies[@"GLD"];
        NSLog(@"%@: %ld", gold.name, gold.balance);
}];
```

### TODO
- Write integration tests. This will likely require some backend work to ensure that adding virtual currencies to the integration test app doesn't add VCs to the CustomerInfo response, which could break existing tests
- Add logging in a follow-up PR. This wasn't included in this PR to keep the size from getting any bigger. This work has already been done in the branch `add-vcmanager-log-messages`, and we'll create a PR for that after this is merged into the VC dev branch.
- Consider adding the VerificationResult to the `VirtualCurrencies` object.
- Add the ability for developers to be notified when the VC balances change without having to manually refresh the virtual currencies
- Caching optimizations:
   - Invalidate the cache after a purchase is made
   - Cache the virtual currencies in memory in addition to on the disk
- Add diagnostics events for fetching virtual currencies


RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigData+Mock.swift
```diff
@@ -146,65 +146,4 @@ extension CustomerCenterConfigData {
         buttonTextColor: .init(light: "#ffffff", dark: "#000000"),
         buttonBackgroundColor: .init(light: "#287aff", dark: "#287aff")
     )
-
-    static var fourVirtualCurrencies: [String: RevenueCat.VirtualCurrencyInfo] {
```
fire-at-will, line 150: Moved these to VirtualCurrenciesFixtures.swift

RevenueCatUI/CustomerCenter/Data/CustomerCenterPurchases.swift
```diff
@@ -82,6 +82,10 @@ final class CustomerCenterPurchases: CustomerCenterPurchasesType {
         try await Purchases.shared.syncPurchases()
     }
 
+    func virtualCurrencies() async throws -> VirtualCurrencies {
+        return try await Purchases.shared.virtualCurrencies(forceRefresh: true)
```
fire-at-will: We always want to hard refresh VCs in the Customer Center so that we don't accidentally show stale balances

RevenueCatUI/CustomerCenter/Views/RelevantPurchasesListView.swift
```diff
@@ -35,7 +35,7 @@ struct RelevantPurchasesListView: View {
     @Environment(\.navigationOptions)
     var navigationOptions
 
-    @StateObject
+    @ObservedObject
```
fire-at-will, line 38: This was required to fix a bug where the virtual currencies weren't being displayed in both RelevantPurchasesListView and SubscriptionDetailView after I made the changes in this PR.

Here’s what’s going on:
- In both RelevantPurchasesListView and SubscriptionDetailView,  the views are initialized multiple times. The first time, virtualCurrencies haven’t been loaded yet, and are nil. The second time, virtualCurrencies have been loaded.
- However, the views are never updated with the newly passed in virtual currencies 😅
- This is because in the views’ initializers, we create a new view model instance, but the view model is stored as a @StateObject, which just stores one instance per scene. This meant that the second view model instance with the virtual currencies was being ignored by the view 
- Changing this from @StateObject to @ObservedObject fixes this issue, since now the view will observe the second view model instead of the first one

fire-at-will:
@RCGitBot please test

.circleci/config.yml
```diff
@@ -959,7 +959,8 @@ jobs:
       - checkout
       - install-dependencies:
           install_xcbeautify: false
-          install_mint: true
+          # Mint 0.17.5 is required here as newer versions of Mint don't support macOS 12.
```
fire-at-will, line 969: We need to force installing mint 17.5 in the iOS 13 job, since mint 18+ don't support macOS 12, which is required for iOS 13/14. Needed to run the iOS 13 job to generate the new snapshots

See https://github.com/RevenueCat/purchases-ios/pull/5256 for more details

fire-at-will:
@RCGitBot please test

.circleci/config.yml
MarkVillacampa: Once we rebase main this will not be needed then?

Sources/Networking/HTTPClient/HTTPRequestPath.swift
```diff
@@ -183,6 +186,7 @@ extension HTTPRequest.Path: HTTPRequestPath {
                 .postAdServicesToken,
                 .postOfferForSigning,
                 .postRedeemWebPurchase,
+                .getVirtualCurrencies,
```
fire-at-will: TODO: We may want to enable signature verification in a future PR

.circleci/config.yml
fire-at-will: We should still keep this and merge it into main. PR #5256 fixes this for the iOS 14 job, but not iOS 13, since we're apparently not running the iOS 13 job today. It looks like I wrote a PR to enable that job, but it never got merged. I'll see if I can get it merged into main after we merge the VCs in with this fix: https://github.com/RevenueCat/purchases-ios/pull/4757

RevenueCatUI/CustomerCenter/Views/SubscriptionDetailView.swift
```diff
@@ -54,7 +54,7 @@ struct SubscriptionDetailView: View {
         customerInfoViewModel: CustomerCenterViewModel,
         screen: CustomerCenterConfigData.Screen,
         purchaseInformation: PurchaseInformation?,
-        virtualCurrencies: [String: VirtualCurrencyInfo]?,
+        virtualCurrencies: RevenueCat.VirtualCurrencies?,
```
MarkVillacampa, line 57: I see we're using `RevenueCat.` prefix for some types, was it required to disambiguate something?

fire-at-will: It's not required, but something I went with to prevent potential future conflicts in the unlikely event that StoreKit adds their own `VirtualCurrencies`/`VirtualCurrency` objects in the future and breaks SDK compilation, like they did with `SubscriptionInfo` in iOS 18.4. It's probably overkill, but I didn't want to have to deal with that again in the future 😅

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1297,6 +1308,50 @@ public extension Purchases {
...
+                    let publicError = NewErrorUtils.purchasesError(withUntypedError: error).asPublicError
+                    completion(nil, publicError)
+                }
+            }
+        }
+    }
+
+    func virtualCurrencies() async throws -> VirtualCurrencies {
+        return try await self.virtualCurrencies(forceRefresh: false)
+    }
```
MarkVillacampa, line 1340: isn't this functionally the same as having the method below have a default param value of false?

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -1233,6 +1265,31 @@ public protocol PurchasesSwiftType: AnyObject {
...
+     * bypassing the SDK's cache. Pass in `true` to force the SDK to fetch the most recent value
+     * from the RevenueCat backend.
+     * - Returns: The ``VirtualCurrencies`` object containing the virtual currencies for the subscriber.
+     *
+     * #### Related Articles
+     * -  [Virtual Currencies](https://www.revenuecat.com/docs/offerings/virtual-currency)
+     */
+    func virtualCurrencies(
+        forceRefresh: Bool
+    ) async throws -> VirtualCurrencies
```
MarkVillacampa: same here, any reason to have them as separate functions?

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -978,6 +978,38 @@ public protocol PurchasesType: AnyObject {
...
+     * - Parameter completion: The callback that is called when the request is complete with a ``VirtualCurrencies``
+     * object containing the subscriber's virtual currencies.
+     *
+     * #### Related Articles
+     * -  [Virtual Currencies](https://www.revenuecat.com/docs/offerings/virtual-currency)
+     */
+    @objc(virtualCurrenciesWithForceRefresh:completion:)
+    func virtualCurrencies(
+        forceRefresh: Bool,
+        completion: @escaping @Sendable (VirtualCurrencies?, PublicError?) -> Void
```
MarkVillacampa: I can see Objc needing two methods because it doesnt have default parameters. Which then makes be think, should we instead have a separate `invalidateVirtualCurrencyCache` method? would make it more explicit.

Sources/Virtual Currencies/VirtualCurrencies.swift
```diff
@@ -0,0 +1,72 @@
...
+    @objc public subscript(key: String) -> VirtualCurrency? {
+        return self.all[key]
+    }
+}
+
+public extension VirtualCurrencies {
+    /// Returns a dictionary containing only the virtual currencies that have a balance greater than zero.
+    /// - Returns: A dictionary of virtual currency codes to their corresponding info objects,
+    ///     filtered to only include those with non-zero balances.
+    var withNonZeroBalance: [String: VirtualCurrency] {
```
MarkVillacampa: was this needed for Customer Center or just potential nice public API to have?

Sources/Virtual Currencies/VirtualCurrency.swift
```diff
@@ -0,0 +1,72 @@
...
+extension VirtualCurrency: Codable {}
+extension VirtualCurrency: Sendable {}
+
+extension VirtualCurrency {
+    /// Compares this virtual currency with another one.
+    /// - Parameter object: The other object to compare with
+    /// - Returns: `true` if both objects are virtual currencies with the same balance, `false` otherwise
+    @objc public override func isEqual(_ object: Any?) -> Bool {
+        guard let other = object as? VirtualCurrency else { return false }
+        return self.balance == other.balance
```
MarkVillacampa, line 59: why only compare the balance? should compare at least code as well right?

Sources/Virtual Currencies/VirtualCurrencies.swift
fire-at-will: `withNonZeroBalance` and `withZeroBalance` aren't needed for CustomerCenter - CustomerCenter just displays `VirtualCurrencies.all`. I figured they'd be nice APIs to have!

Sources/Virtual Currencies/VirtualCurrency.swift
fire-at-will: Good catch! This is an oversight - I added in the `equals` function before the other fields were added in. I've added them in this commit: 270897dde4f888b0475de401ccbb6f448bcd3692 👍

Sources/Purchasing/Purchases/Purchases.swift
fire-at-will: Good point - I've removed this and added the default value of `false` in f7433107758dfffa31da562cd7739b9d5c569040

Sources/Purchasing/Purchases/PurchasesType.swift
fire-at-will: Consolidated these in f7433107758dfffa31da562cd7739b9d5c569040!

Sources/Purchasing/Purchases/PurchasesType.swift
fire-at-will: It looks like CustomerInfo actually does _both_ approaches 😅

You can call all of these:
```swift
Purchases.shared.cachedCustomerInfo
try! Purchases.shared.customerInfo()
try! Purchases.shared.customerInfo(fetchPolicy: .fetchCurrent)
Purchases.shared.invalidateCustomerInfoCache()
```

```objc
RCCustomerInfo * _Nullable __unused cachedCustomerInfo = purchases.cachedCustomerInfo;
[purchases getCustomerInfoWithCompletion:^(RCCustomerInfo *info, NSError *error) {}];
[purchases getCustomerInfoWithFetchPolicy:RCCacheFetchPolicyFetchCurrent completion:^(RCCustomerInfo *customerInfo,
                                                                                  NSError *error) {}];
[purchases invalidateCustomerInfoCache];
```

If you think we should add them in, I can add
`Purchases.shared.cachedCustomerInfo` and `Purchases.shared.invalidateVirtualCurrenciesCache()` in this PR or in a follow-up PR to help the API match the `CustomerInfo` APIs

fire-at-will:
@RCGitBot please test

Sources/Virtual Currencies/VirtualCurrencies.swift
MarkVillacampa: I don't hate it, was just thinking of minimising API surface area for now to make documentation and adoption easier, but this is small enough we can keep it

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1297,6 +1308,51 @@ public extension Purchases {
...
+public extension Purchases {
+
+    @objc func virtualCurrencies(
+        completion: @escaping @Sendable (VirtualCurrencies?, PublicError?) -> Void
+    ) {
+        OperationDispatcher.dispatchOnMainActor {
+            Task {
+                do {
+                    let virtualCurrencies = try await self.virtualCurrencies()
+                    completion(virtualCurrencies, nil)
```
MarkVillacampa, line 1322: I think we should remove the outer `dispatchOnMainActor` and wrap only the completion calls in it. For reference check how we're doing it in `eligibleWinBackOffers`

Sources/Purchasing/Purchases/PurchasesType.swift
```diff
@@ -1233,6 +1258,24 @@ public protocol PurchasesSwiftType: AnyObject {
...
+     * #### Related Articles
+     * -  [Virtual Currencies](https://www.revenuecat.com/docs/offerings/virtual-currency)
+     */
+    func virtualCurrencies() async throws -> VirtualCurrencies
+
+    /**
+     * Invalidates the cache for virtual currencies.
+     *
+     * This is useful for cases where a virtual currency's balance might have been updated
+     * outside of the app, like if virtual currency is granted through the RevenueCat dashboard.
```
MarkVillacampa: I think this is example might be a bit of a less common use case. The typical scenario where you'd use this method is when the balance has changed because you have either decreased it (user spending) or increased it (new grant) from your backend using server APIs.

Sources/Virtual Currencies/VirtualCurrencyManager.swift
```diff
@@ -0,0 +1,119 @@
...
+            appUserID: appUserID,
+            isAppBackgrounded: isAppBackgrounded
+        )
+
+        cacheVirtualCurrencies(virtualCurrencies, appUserID: appUserID)
+
+        return virtualCurrencies
+    }
+
+    func invalidateVirtualCurrenciesCache() async {
```
MarkVillacampa: I think this method does not need to be async, and that would bubble up to the public methods simplifying usage 😄

fire-at-will:
@RCGitBot please test

Sources/Virtual Currencies/VirtualCurrencies.swift
```diff
@@ -0,0 +1,56 @@
...
+//  VirtualCurrencies.swift
+//
+//  Created by Will Taylor on 5/21/25.
+
+import Foundation
+
+/// This class contains all the virtual currencies associated to the user.
+@objc(RCVirtualCurrencies) public final class VirtualCurrencies: NSObject {
+
+    /// Dictionary of all VirtualCurrency(``VirtualCurrency``) objects keyed by virtual currency code.
```
ajpallares: Why not just
```suggestion
    /// Dictionary of all ``VirtualCurrency`` objects keyed by virtual currency code.
```

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -192,7 +192,8 @@ import Foundation
             try await self.loadCustomerCenterConfig()
 
             if self.configuration?.support.displayVirtualCurrencies == true {
-                self.virtualCurrencies = customerInfo.virtualCurrencies
+                purchasesProvider.invalidateVirtualCurrenciesCache()
```
ajpallares, line 195: Do we really need this all the times or could we have another parameter to control when the VCs cache is invalidated?

I do see the point of always invalidating in `VirtualCurrencyBalancesScreenViewModel` though, in order to always have the latest information in that screen

RevenueCatUI/CustomerCenter/Views/RelevantPurchasesListView.swift
ajpallares: Ideally, I think this should be a `@StateObject` because the `RelevantPurchasesListView` is the owner/creator of the `ObservedObject`.

But I understand the problem (which, tbh, I didn't fully understand until today, also with the help of [this link](https://swiftui-lab.com/random-lessons/#data-10)). A potential workaround would be to have an `updateVirtualCurrencies` method in the `RelevantPurchasesListViewModel` object. But I think that the real underlying problem is the fact that we don't really have a unique source of truth for our data. And that's perhaps a much deeper problem into the whole CC UI, for sure not a topic to tackle in this PR. So I'm fine with this as a workaround 👍

Sources/Caching/DeviceCache.swift
```diff
@@ -378,8 +381,72 @@ class DeviceCache {
...
+        virtualCurrencies: Data,
+        appUserID: String
+    ) {
+        virtualCurrenciesLock.perform {
+            self.userDefaults.write {
+                $0.set(virtualCurrencies, forKey: CacheKey.virtualCurrencies(appUserID))
+                Self.setVirtualCurrenciesCacheLastUpdatedTimestampToNow($0, appUserID: appUserID)
+            }
+        }
+    }
```
ajpallares, line 397: Just to make sure. Do we want/would it be useful to have an in-memory cache of VCs (the same as for offerings)?

Sources/Caching/DeviceCache.swift
```diff
@@ -378,8 +381,72 @@ class DeviceCache {
...
+                    return true
+                }
+
+                let timeSinceLastCheck = cachesLastUpdated.timeIntervalSinceNow * -1
+                let cacheDurationInSeconds = self.cacheDurationInSeconds(
+                    isAppBackgrounded: isAppBackgrounded,
+                    isSandbox: self.sandboxEnvironmentDetector.isSandbox
+                )
+
+                return timeSinceLastCheck >= cacheDurationInSeconds
```
ajpallares, line 420: Huge nit: perhaps we could extract this into a method (which would also help when reading the code, I think). FFTI

Sources/Misc/SystemInfo.swift
```diff
@@ -208,6 +210,15 @@ class SystemInfo {
...
     }
 
+    /// Async/await API to check if an app is backgrounded at a specific moment.
+    func isApplicationBackgrounded() async -> Bool {
+        return await Async.call { completion in
+            self.isApplicationBackgrounded { isAppBackgrounded in
+                completion(isAppBackgrounded)
+            }
+        }
+    }
```
ajpallares: Just in case you didn't know, Toni [recently added](https://github.com/RevenueCat/purchases-ios/pull/4938) a [isAppBackgroundedState](https://github.com/RevenueCat/purchases-ios/blob/2e503c956ee33a8ff419560fb97bba0c8036328a/Sources/Misc/SystemInfo.swift#L50) property in `SystemInfo`, so checking this asynchronously should not be needed anymore.

Sources/Virtual Currencies/VirtualCurrencies.swift
```diff
@@ -0,0 +1,56 @@
...
+    /// This dictionary can also be access through an index subscript on ``VirtualCurrencies``, e.g.
```
ajpallares: If I'm not wrong
```suggestion
    /// This dictionary can also be accessed through an index subscript on ``VirtualCurrencies``, e.g.
```

Sources/Virtual Currencies/VirtualCurrencies.swift
```diff
@@ -0,0 +1,56 @@
...
+    @objc public subscript(key: String) -> VirtualCurrency? {
+        return self.all[key]
+    }
+}
+
+extension VirtualCurrencies: Sendable {}
+extension VirtualCurrencies: Codable {}
+
+extension VirtualCurrencies {
+    /// Compares two VirtualCurrencies objects for equality by comparing their underlying dictionaries.
```
ajpallares: Nit
```suggestion 
    /// Compares two ``VirtualCurrencies`` objects for equality by comparing their underlying dictionaries.
```

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -592,6 +593,13 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
             paywallCache = nil
         }
 
+        let virtualCurrencyManager = VirtualCurrencyManager(
```
ajpallares, line 596: Perhaps it would make sense to add some kind of diagnostics for Virtual Currencies? Of course, it would be a matter of a separate PR and discussing it internally with the team. But just something to have in mind

Tests/APITesters/AllAPITests/ObjcAPITester/RCVirtualCurrenciesAPI.m
```diff
@@ -0,0 +1,18 @@
...
+//
+
+@import RevenueCat;
+#import "RCVirtualCurrenciesAPI.h"
+
+@implementation RCVirtualCurrenciesAPI
+
++ (void)checkAPI {
+    RCVirtualCurrencies *virtualCurrencies = nil;
+    NSDictionary<NSString *, RCVirtualCurrency *> *all = virtualCurrencies.all;
```
ajpallares, line 15: What about checking here the subscripting as well or wouldn't this work in Objective-C?

ajpallares (commented):
Some early comments! Will get back to this to keep reviewing. But so far (i.e. based on the files I've actually reviewed) it's looking great! Amazing job!

RevenueCatUI/CustomerCenter/Mocks/MockCustomerCenterPurchases.swift
```diff
@@ -118,6 +118,27 @@ final class MockCustomerCenterPurchases: @unchecked Sendable, CustomerCenterPurc
...
+        guard let virtualCurrenciesResult = virtualCurrenciesResult else {
+            return VirtualCurrenciesFixtures.noVirtualCurrencies
+        }
+
+        switch virtualCurrenciesResult {
+        case .success(let virtualCurrencies):
+            return virtualCurrencies
+        case .failure(let error):
+            throw error
+        }
```
ajpallares: This all can be replaced by this one liner, but FFTI
```suggestion
        return try virtualCurrenciesResult?.get() ?? VirtualCurrenciesFixtures.noVirtualCurrencies
```

RevenueCatUI/Data/VirtualCurrenciesFixtures.swift
```diff
@@ -0,0 +1,115 @@
...
+
+        guard let data = jsonData else {
+            return Self.noVirtualCurrencies
+        }
+
+        guard let virtualCurrencies = try? JSONDecoder().decode(RevenueCat.VirtualCurrencies.self, from: data) else {
+            return Self.noVirtualCurrencies
+        }
+
+        return virtualCurrencies
```
ajpallares: nit: just to make it easier to catch potential errors, wouldn't it be better to do the same as in `noVirtualCurrencies`?
```suggestion
        return try! JSONDecoder().decode(RevenueCat.VirtualCurrencies.self, from: jsonData!)
```

RevenueCatUI/Data/VirtualCurrenciesFixtures.swift
```diff
@@ -0,0 +1,115 @@
...
```
ajpallares: same nit?
```suggestion
        return try! JSONDecoder().decode(RevenueCat.VirtualCurrencies.self, from: jsonData!)
```

Sources/Virtual Currencies/VirtualCurrencyManager.swift
```diff
@@ -0,0 +1,119 @@
...
+        deviceCache.clearVirtualCurrenciesCache(appUserID: appUserID)
+    }
+
+    private func cacheVirtualCurrencies(
+        _ virtualCurrencies: VirtualCurrencies,
+        appUserID: String
+    ) {
+        guard let virtualCurrenciesData = try? JSONEncoder().encode(virtualCurrencies) else {
+            return
+        }
```
ajpallares, line 73: Question: should we clear the cache at this point?
I guess this should never happen, but still, if we've just received a new virtual currency object, it means that the one currently in cache is not needed anymore.
```suggestion
        guard let virtualCurrenciesData = try? JSONEncoder().encode(virtualCurrencies) else {
            self.deviceCache.clearVirtualCurrenciesCache(appUserID: appUserID)
            return
        }
```

Tests/APITesters/AllAPITests/SwiftAPITester/VirtualCurrencyAPI.swift
```diff
@@ -0,0 +1,16 @@
...
+//
+
+import Foundation
+import RevenueCat
+
+func checkVirtualCurrency(virtualCurrency: VirtualCurrency) {
+    let balance: Int = virtualCurrency.balance
+    let name: String = virtualCurrency.name
+    let code: String = virtualCurrency.code
+    let serverDescription: String? = virtualCurrency.serverDescription
```
ajpallares, line 15: Probably not too important and for sure it's not something only in this API test, but I've just realized that doing this does not cover from us changing the `serverDescription` to nonnullable 🤔 
I guess we could add something like this:
```suggestion
    let serverDescription: String? = virtualCurrency.serverDescription
    var serverDescriptionAlt = virtualCurrency.serverDescription
    serverDescriptionAlt = ""
    serverDescriptionAlt = nil
```
If we where to change its type or nullability, one of the two lines would complain

Tests/UnitTests/Mocks/MockVirtualCurrencyManager.swift
```diff
@@ -0,0 +1,43 @@
...
+    func virtualCurrencies() async throws -> RevenueCat.VirtualCurrencies {
+        self.virtualCurrenciesCallCount += 1
+        self.virtualCurrenciesCalled = true
+
+        switch stubbedVirtualCurrenciesResult {
+        case .success(let virtualCurrencies):
+            return virtualCurrencies
+        case .failure(let error):
+            throw error
+        }
```
ajpallares: Same nit as above. This could be
```suggestion
        return try stubbedVirtualCurrenciesResult.get()
```

Tests/UnitTests/Networking/Backend/BackendGetVirtualCurrenciesTests.swift
```diff
@@ -0,0 +1,291 @@
...
+            appUserID: Self.userID,
+            isAppBackgrounded: false
+        ) { _ in }
+
+        self.virtualCurrenciesAPI.getVirtualCurrencies(
+            appUserID: Self.userID,
+            isAppBackgrounded: false
+        ) { _ in }
+
+        expect(self.httpClient.calls).toEventually(haveCount(1))
```
ajpallares, line 101: This "eventually", I'm not sure it'd cover against also having count of 2. Perhaps we could add a line that checks that `.toNever(haveCount(2))`?

Tests/UnitTests/Networking/Backend/BackendGetVirtualCurrenciesTests.swift
```diff
@@ -0,0 +1,291 @@
...
+                            delay: .milliseconds(10))
+        )
+        let result: Atomic<Result<VirtualCurrenciesResponse, BackendError>?> = nil
+
+        self.virtualCurrenciesAPI.getVirtualCurrencies(
+            appUserID: Self.userID,
+            isAppBackgrounded: false
+        ) { result.value = $0 }
+
+        expect(result.value).toEventuallyNot(beNil())
```
ajpallares: I think we should avoid the `toEventually` clauses (which can make the test flakier and slower). In this case, we can just wait for the result instead.
```suggestion
        let result = waitUntilValue { completed in
            self.virtualCurrenciesAPI.getVirtualCurrencies(
                appUserID: Self.userID,
                isAppBackgrounded: false,
                completion: completed
            )
        }
```

Line `expect(result.value).toEventuallyNot(beNil())` should not be needed now because there's already the `let response = try XCTUnwrap(result.value?.value)` check

Tests/UnitTests/Networking/Backend/BackendGetVirtualCurrenciesTests.swift
```diff
@@ -0,0 +1,291 @@
...
```
ajpallares: Same comment here to avoid the `toEventually`

Tests/UnitTests/Networking/Backend/BackendGetVirtualCurrenciesTests.swift
```diff
@@ -0,0 +1,291 @@
...
+        self.virtualCurrenciesAPI.getVirtualCurrencies(
+            appUserID: Self.userID,
+            isAppBackgrounded: false
+        ) { _ in }
+        self.virtualCurrenciesAPI.getVirtualCurrencies(
+            appUserID: Self.userID,
+            isAppBackgrounded: false
+        ) { _ in }
+
+        expect(self.httpClient.calls).toEventually(haveCount(1))
```
ajpallares, line 240: In this case we cannot avoid the `toEventually` so easily because we want the two requests to happen at the same time. So I think this is OK. Although we could potentially use Async/await and trigger both (with `async let`) and then `await` for the two results. But perhaps it's too much to avoid the `toEventually` 😅

Tests/UnitTests/Purchasing/Purchases/PurchasesVirtualCurrenciesTests.swift
```diff
@@ -0,0 +1,121 @@
...
+    func testVirtualCurrenciesCallbackCallsSuccessOnMainThread() async throws {
+        self.mockVirtualCurrencyManager.stubbedVirtualCurrenciesResult = .success(Self.mockVirtualCurrencies)
+
+        await waitUntil { completed in
+            self.purchases.virtualCurrencies { _, _ in
+                expect(Thread.isMainThread).to(beTrue())
+                completed()
+            }
+        }
+    }
```
ajpallares, line 100: Nice one!

ajpallares (approved):
Looking great! I got the change to look at everything and added some comments, nothing too important tbh, just small improvements and nits. Amazing work! 🙌

fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

PR #5152, opened 2025-05-21
Add isActive to PurchaseInformation for CustomerCenter

facumenzella:
### Motivation
We need to show `active` badge when the purchase is active

### Description
Add `isActive` to purchase information, to use it in the card.
Also remove the title from the additional info. We're not using it in the end


RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -61,6 +61,13 @@ struct PurchaseInformation {
     /// Note: `false` for non-subscriptions
     let isCancelled: Bool
 
+    /// Indicates whether the purchased subscription is active
+    /// - `true` if the subscription is active
+    /// - `false` if the subscription is not active
```
ajpallares, line 66: I would remove these two comment lines. Too obvious perhaps

RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
```diff
@@ -78,13 +75,7 @@ struct PurchaseInformationCardView: View {
...
         }
 
-        if let refundStatus, let message = refundStatus.subtitle(localization: localization) {
-            self.additionalInfoTitle = localization[.refundStatus]
-            self.additionalInfoSubtitle = message
-        } else {
-            self.additionalInfoTitle = nil
-            self.additionalInfoSubtitle = nil
-        }
+        self.additionalInfoSubtitle = refundStatus?.subtitle(localization: localization)
```
ajpallares, line 78: Nit: since the old title was removed, perhaps this could be renamed to `additionalInfoTitle`. Or even better, just `additionalInfo`

ajpallares (approved):
Great! Just two nits

PR #5153, opened 2025-05-22
Wait a max of 20 minutes for TestFlight processing

joshdholtz:
### Motivation

Waiting for builds to process on CI can sometimes go indefinitely and that is bad for environment and $$$$

### Description

Set a max of 20 minutes for waiting for build processing (if it's longer than 20 minutes, odds are its not going to happen)


tonidero (approved):
🛳️

PR #5154, opened 2025-05-22
Pass CustomerCenterViewModel as a ObservedObject to the detail screen

facumenzella:
### Motivation

We need the ability to respond to `CustomerCenterActions` from multiple parts of the UI, not just from the main view. This is relevant when navigating into detail screens (new UI) where actions may be triggered that the main view needs to observe (or vice versa).

### IMPORTANT
We will most likely remove the use of PreferenceKeys to notify about updates that are not SwiftUI related, like restoration, manage subscriptions, etc. 
That will be done in another PR, after releasing 5.25.0

### Description

This PR refactors the `CustomerCenterViewModel`:
- The CustomerCenterViewModel is now passed as an @ObservedObject.
  - Now we show StoreKit sheet from the Detail instead of the List.
- Refactored `CustomerCenterActionViewModifier` to use Combine instead of a single callback approach:
  - Actions are now exposed as `PassthroughSubject` publishers inside `CustomerCenterActionWrapper`.
  - The view modifier subscribes to those publishers and translates them into PreferenceKey values using @State and a UniqueWrapper to ensure distinct updates.
  - This makes it safe to apply the view modifier in multiple views without overriding or disconnecting handlers.


RevenueCatUI/CustomerCenter/Data/CustomerCenterActionWrapper.swift
```diff
@@ -62,57 +62,96 @@ internal enum CustomerCenterInternalAction {
...
 
+    private var restoreStartedCallbacks: [() -> Void] = []
+    private var restoreFailedCallbacks: [(NSError) -> Void] = []
+    private var restoreCompletedCallbacks: [(CustomerInfo) -> Void] = []
+    private var showingManageSubscriptionsCallbacks: [() -> Void] = []
+    private var refundRequestStartedCallbacks: [(String) -> Void] = []
+    private var refundRequestCompletedCallbacks: [(String, RefundRequestStatus) -> Void] = []
+    private var feedbackSurveyCompletedCallbacks: [(String) -> Void] = []
+    private var managementOptionSelectedCallbacks: [(CustomerCenterActionable) -> Void] = []
+    private var promotionalOfferSuccessCallbacks: [() -> Void] = []
```
facumenzella, line 75: This won't work because there's no way to unregistered the block 🤔 . I think I need to go with another approach.

RevenueCatUI/CustomerCenter/Views/SubscriptionDetailView.swift
```diff
@@ -40,34 +40,68 @@ struct SubscriptionDetailView: View {
...
+        customerInfoViewModel: CustomerCenterViewModel,
         viewModel: SubscriptionDetailViewModel
     ) {
+        self.customerInfoViewModel = customerInfoViewModel
         self._viewModel = .init(wrappedValue: viewModel)
     }
 
     var body: some View {
         content
+            .modifier(CustomerCenterActionViewModifier(actionWrapper: viewModel.actionWrapper))
```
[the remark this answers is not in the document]
ajpallares: This smells like Combine to me

ajpallares (commented):
I think it's looking better. Perhaps we could use combine for the action handlers thing?

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionWrapper.swift
facumenzella: Re-wrote this using combine 👍

RevenueCatUI/CustomerCenter/Views/SubscriptionDetailView.swift
```diff
@@ -297,8 +351,8 @@ struct SubscriptionDetailView_Previews: PreviewProvider {
         .environment(\.appearance, CustomerCenterConfigData.default.appearance)
     }
 
-}
+ }
 
-#endif
+ #endif
```
ajpallares, line 356: same?

RevenueCatUI/CustomerCenter/Views/RelevantPurchasesListView.swift
```diff
@@ -254,14 +244,14 @@ struct RelevantPurchasesListView: View {
...
-struct ActiveSubscriptionsListView_Previews: PreviewProvider {
+ #if DEBUG
+ @available(iOS 15.0, *)
+ @available(macOS, unavailable)
+ @available(tvOS, unavailable)
+ @available(watchOS, unavailable)
+ struct RelevantPurchasesListView_Previews: PreviewProvider {
 
-    // swiftlint:disable force_unwrapping
+     // swiftlint:disable force_unwrapping
```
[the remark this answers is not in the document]
facumenzella: I can pick this up in https://github.com/RevenueCat/purchases-ios/pull/5146

PR #5157, opened 2025-05-22
Add a push navigation to Customer Center maestro app

facumenzella:
### Motivation
To be able to automate testing for both cases, we need to navigate to it inside a NavStack.

### Description
- Add navigationStack to the sample app


ajpallares (approved):
👌

PR #5158, opened 2025-05-22
Introduce CustomerCenterButtonStyle to highlight CustomerCenter buttons

facumenzella:
### Motivation
"See all button" is not correctly highlighted when tapping on it. This adds a button style that takes into account the pressed state.

### Description
- Use button style for both buttons that are inside the scrollview 💅 
- Apply tint to feedback survey

### DIFF
--- a/RevenueCatUI/CustomerCenter/ButtonStyles.swift
+++ b/RevenueCatUI/CustomerCenter/ButtonStyles.swift
@@ -66,7 +66,7 @@ struct CustomerCenterButtonStyle: ButtonStyle {
 @available(tvOS, unavailable)
 @available(watchOS, unavailable)
 extension ButtonStyle where Self == CustomerCenterButtonStyle {
-    static func customerCenterScrollView(for colorScheme: ColorScheme) -> CustomerCenterButtonStyle {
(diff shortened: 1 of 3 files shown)


RevenueCatUI/CustomerCenter/ButtonStyles.swift
```diff
@@ -42,7 +42,40 @@ struct ProminentButtonStyle: PrimitiveButtonStyle {
...
+            .cornerRadius(10)
+    }
+}
+
+@available(iOS 15.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
+extension ButtonStyle where Self == CustomerCenterButtonStyle {
+    static func customerCenterScrollView(for colorScheme: ColorScheme) -> CustomerCenterButtonStyle {
```
ajpallares, line 69: I believe the name is wrong
```suggestion
    static func customerCenterButtonStyle(for colorScheme: ColorScheme) -> CustomerCenterButtonStyle {
```

ajpallares (approved):
Makes sense! Just a small naming issue

PR #5159, opened 2025-05-22
Minor UI tweaks for CustomerCenter 2.0

facumenzella:
### Motivation
We did a round of testing with Kike, and this came out.

### Description
- Filter .missingPurchase only if the detail is not the root. If the detail is the second screen, then don't show it.
- Add icon in the additionalInfo HStack
- Fix section titles for RelevantList
- Dont show consumable price next to the chevron (only below the title)

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
@@ -31,6 +31,7 @@ struct PurchaseInformationCardView: View {
 
     private let storeTitle: String
 
+    private let additionalIcon: Image?
     private let additionalInfo: String?
 
     private let paidPrice: String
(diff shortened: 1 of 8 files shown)


RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
```diff
@@ -153,14 +159,22 @@ struct PurchaseInformationCardView: View {
...
-            if let additionalInfo {
-                Text(additionalInfo)
-                    .font(.caption)
-                    .foregroundStyle(.primary)
-                    .padding(.bottom, 4)
-                    .frame(maxWidth: .infinity, alignment: .leading)
-                    .multilineTextAlignment(.leading)
-                    .padding()
+            if let additionalInfo, let additionalIcon {
+                HStack(alignment: .center, spacing: 12) {
```
ajpallares, line 163: Question: should this be a `CompatibilityLabeledContent`? I guess it's not a big deal

ajpallares (approved):
Makes sense! Only a small question, but NABD

RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
facumenzella: It's the opposite. The icon is on the left.

PR #5160, opened 2025-05-22
[Paywalls] Fix scrolling stack

MarkVillacampa:
### Checklist

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
@@ -744,7 +744,7 @@ func stackAlignmentAndDistributionPreviews(dimensions: [PaywallComponent.Dimensi
                     component: PaywallComponent.StackComponent(
                         components: innerStacks(dimension: dimension),
                         dimension: dimension,
-                        size: .init(width: .fill, height: .fixed(150)),
+                        size: .init(width: .fill, height: .fixed(170)),
                         spacing: 10,
                         backgroundColor: .init(light: .hex("#ff0000")),
                         padding: .init(top: 10, bottom: 10, leading: 10, trailing: 10),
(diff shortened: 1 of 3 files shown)


Sources/Paywalls/Components/PaywallStackComponent.swift
```diff
@@ -20,7 +20,7 @@ public extension PaywallComponent {
 
         public enum Overflow: String, PaywallComponentBase {
             case `default`
-            case scroll
+            case scroll = "scrollable"
```
MarkVillacampa, line 23: Changing the string representation of the scroll property to make it backwards-incompatible. That way we prevent older versions of the SDK with incorrect behaviour from using it. We'll need to change this in the schema & backend too.

RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
```diff
@@ -462,22 +477,22 @@ struct StackComponentView_Previews: PreviewProvider {
                         components: [
                             .stack(.init(
                                 components: [],
-                                size: .init(width: .fixed(300), height: .fixed(50)),
+                                size: .init(width: .fixed(50), height: .fixed(50)),
```
MarkVillacampa, line 480: Had to tweak the fixed height of some previews and increase them by the amount of padding in that axis so snapshots would not change.

RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
```diff
@@ -110,22 +124,23 @@ struct StackComponentView: View {
                         onDismiss: self.onDismiss
                     )
                 }
+                // Padding needs to be applied before scrollview so it applies to the inner stack
+                .padding(style.padding.extend(by: style.border?.width ?? 0))
+                .padding(additionalPadding)
```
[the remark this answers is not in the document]
JZDesign: The editor behaves such that you fix the size, and padding is additive to that

PR #5161, opened 2025-05-22
CAT-1975: Dont Post Receipt with No Receipt or AppTransaction when Syncing Purchases with SK2

fire-at-will:
### Motivation
We've identified what appears to be an SDK bug in `PurchasesOrchestrator.syncPurchasesSK2` where we call POST /receipt with no `fetch_token` and no `appTransaction` when:
- No transactions are found on the device
- No app transaction is able to be fetched from StoreKit
- One or more of the following is true:
   - There is no cached CustomerInfo
   - The cached CustomerInfo has no `originalPurchaseDate`
   -  The cached CustomerInfo has no `originalApplicationVersion`

When this occurs, the backend returns a 400: bad request response, and the SDK will return an error to the calling code.

To maintain the `syncPurchases()` contract, we should address this issue by not POSTing to /v1/receipt in this scenario and instead fetching and returning the CustomerInfo.

### Description
This PR modifies the syncPurchasesSK2 logic so that if the above scenario occurs, we log a message, and then hard refresh and return CustomerInfo.

### TODO
- Write a test or two

### DIFF
--- a/Sources/Logging/Strings/StoreKitStrings.swift
+++ b/Sources/Logging/Strings/StoreKitStrings.swift
@@ -95,6 +95,8 @@ enum StoreKitStrings {
 
     case skunknown_purchase_result(String)
 
+    case sk2_sync_purchases_no_transaction_or_apptransaction_found
+
 }
 
 extension StoreKitStrings: LogMessage {
@@ -225,6 +227,10 @@ extension StoreKitStrings: LogMessage {
 
         case let .skunknown_purchase_result(name):
             return "Unrecognized Product.PurchaseResult: \(name)"
+
+        case .sk2_sync_purchases_no_transaction_or_apptransaction_found:
+            return "syncPurchases was called, but we were unable to fetch any transactions or an AppTransaction " +
+            "from StoreKit 2. This seems to occur occassionally in sandbox."
         }
     }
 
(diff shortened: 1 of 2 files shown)


fire-at-will:
@RCGitBot please test

fire-at-will:
@RCGitBot please test

Sources/Logging/Strings/StoreKitStrings.swift
```diff
@@ -225,6 +227,10 @@ extension StoreKitStrings: LogMessage {
 
         case let .skunknown_purchase_result(name):
             return "Unrecognized Product.PurchaseResult: \(name)"
+
+        case .sk2_sync_purchases_no_transaction_or_apptransaction_found:
+            return "syncPurchases was called, but we were unable to fetch any transactions or an AppTransaction " +
+            "from StoreKit 2."
```
MarkVillacampa, line 233: The same code path is used by `restorePurchases` so we could make this error a bit more generic:

```suggestion
        case .sk2_sync_purchases_no_transaction_or_apptransaction_found:
            return "Couldn't find previous transactions or an AppTransaction."
```

MarkVillacampa (approved):

PR #5164, opened 2025-05-23
fix: Do not reload actions when selecting purchase

facumenzella:
### Motivation
With @ajpallares we noticed that the buttons were reloading on selection. This addressed it

### Description
While the detail computes it for every purchase selected, the list now just passes `nil` to it.

https://github.com/user-attachments/assets/36bb4194-2e41-4559-a58f-2eefe40324f2

### DIFF
--- a/RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
@@ -68,7 +68,7 @@ class BaseManageSubscriptionViewModel: ObservableObject {
 
     private var error: Error?
     private let loadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType
-    private(set) var paths: [CustomerCenterConfigData.HelpPath]
+    let paths: [CustomerCenterConfigData.HelpPath]
     private(set) var purchasesProvider: CustomerCenterPurchasesType
 
     init(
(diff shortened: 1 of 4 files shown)


RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
```diff
@@ -68,7 +68,7 @@ class BaseManageSubscriptionViewModel: ObservableObject {
 
     private var error: Error?
     private let loadPromotionalOfferUseCase: LoadPromotionalOfferUseCaseType
-    private let paths: [CustomerCenterConfigData.HelpPath]
+    private(set) var paths: [CustomerCenterConfigData.HelpPath]
```
ajpallares, line 71: I believe this is not being set anywhere, so it can still be a `let`

ajpallares (approved):
I think it makes sense. Just one small nit

RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
facumenzella: I did it so it can be accessed by the subclass. It's kind of a workaround, but it does the trick now. 
I'll most probably refactor this after the release in order to make `paths` private again, or even just compute paths differently.

ajpallares: If it only need to be read but not written, then this could just be
```suggestion
    let paths: [CustomerCenterConfigData.HelpPath]
```

PR #5166, opened 2025-05-23
reload customer center after re-syncing customer info

facumenzella:
### Motivation
While doing the last testing round, I noticed the detail wasn't reloading properly:
- The view uses it's own published property that needs to be reloaded

### Description
1. Reposition where we update customer info in CustomerCenterViewModel
2. Use combine to subscribe to updates
3. publisher method in customer center view model

https://github.com/user-attachments/assets/945d1873-2b63-44cf-a858-003dadb0df2e

### Discussion
I think the ultimate solution would be what we've discussed, and have a CustomerInfoViewModel that only handles:
- active subs
- other purchases
- selected purchase

With that, we can remove the `purchaseInformation` published property from any other StateObject, and rely only on `CustomerInfoViewModel` to handle refreshing

### DIFF
--- a/Examples/rc-maestro/Tuist/Package.swift
+++ b/Examples/rc-maestro/Tuist/Package.swift
@@ -18,7 +18,8 @@ let package = Package(
     dependencies: [
         .package(
             url: "https://github.com/RevenueCat/purchases-ios", 
-            branch: "main"
+            // branch: "main"
+            revision: "981075481"
         ),
     ]
 )
--- a/Examples/rc-maestro/Tuist/Package.swift
+++ b/Examples/rc-maestro/Tuist/Package.swift
@@ -18,8 +18,7 @@ let package = Package(
     dependencies: [
         .package(
             url: "https://github.com/RevenueCat/purchases-ios", 
-            // branch: "main"
-            revision: "981075481"
+            branch: "main"
         ),
     ]
 )
--- a/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
@@ -227,6 +227,8 @@ import RevenueCat
 private extension CustomerCenterViewModel {
 
     func loadPurchases(customerInfo: CustomerInfo) async throws {
+        self.customerInfo = customerInfo
+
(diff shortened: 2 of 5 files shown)


Examples/rc-maestro/Tuist/Package.swift
```diff
@@ -18,7 +18,8 @@ let package = Package(
     dependencies: [
         .package(
             url: "https://github.com/RevenueCat/purchases-ios", 
-            branch: "main"
+            // branch: "main"
+            revision: "981075481"
```
facumenzella, line 22:
```suggestion
            branch: "main"
```

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -212,22 +227,22 @@ import RevenueCat
...
-        self.customerInfo = customerInfo
-
         let hasActiveProducts =  !customerInfo.activeSubscriptions.isEmpty || !customerInfo.nonSubscriptions.isEmpty
 
         if !hasActiveProducts {
             self.activeSubscriptionPurchases = []
             self.activeNonSubscriptionPurchases = []
             self.activePurchase = nil
             self.state = .success
+            self.customerInfo = customerInfo
```
facumenzella, line 237: I think now this can be reverted

ajpallares (approved):
I'm not too convinced about the throttling, but I honestly can't think of a better approach right now. It's an improvement anyways, so we can keep iterating on this if needed 👍

RevenueCatUI/CustomerCenter/Views/SubscriptionDetailView.swift
```diff
@@ -99,9 +98,7 @@ struct SubscriptionDetailView: View {
             }
             .onChangeOf(customerInfoViewModel.manageSubscriptionsSheet) { manageSubscriptionsSheet in
                 if !manageSubscriptionsSheet {
-                    Task {
-                        await customerInfoViewModel.loadScreen(shouldSync: true)
-                    }
+                    viewModel.refreshPurchase()
                 }
             }
```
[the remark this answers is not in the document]
facumenzella: This is actually a great idea. I'll explore it in another one, but I like it

facumenzella:
Yeah, the throttling is not the most robust solution. I'll explore alternatives, I would love to get rid of the combine part for this

PR #5173, opened 2025-05-26
Refactor NoSubscriptionsView in CustomerCenter

facumenzella:
### Motivation
We're moving away from lists, and every screen should follow the same structure. 

### Description
- Use scrollview as other screens
- Remove the use of configuration, since this screen is a fallback
- Renamed screen to reflect the correct use of it + comment

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/FallbackNoSubscriptionsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/FallbackNoSubscriptionsView.swift
@@ -25,7 +25,6 @@ import SwiftUI
 @available(watchOS, unavailable)
 struct FallbackNoSubscriptionsView: View {
 
-    let configuration: CustomerCenterConfigData
     let actionWrapper: CustomerCenterActionWrapper
 
     @Environment(\.appearance)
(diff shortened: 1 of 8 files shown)


RevenueCatUI/CustomerCenter/Views/FallbackNoSubscriptionsView.swift
```diff
@@ -18,13 +18,13 @@ import SwiftUI
 
 #if os(iOS)
 
+/// If fetching the configuration fails (NO_ACTIVE screen is not present) we display this
```
JayShortway, line 21: This is helpful!

RevenueCatUI/CustomerCenter/Views/FallbackNoSubscriptionsView.swift
```diff
@@ -82,6 +85,22 @@ struct NoSubscriptionsView: View {
         }
     }
 
+    private var restorePurchasesView: some View {
```
JayShortway, line 88: Nit: why not "button"?
```suggestion
    private var restorePurchasesButton: some View {
```

JayShortway (approved):
Makes sense!

RevenueCatUI/CustomerCenter/Views/FallbackNoSubscriptionsView.swift
facumenzella: Works too! I'll go for it

PR #5174, opened 2025-05-26
return .free for CustomerCenterStoreKitUtilities if price.isZero

facumenzella:
### Motivation
I've noticed that the renewal price does not return free if it's zero.

### Description
- return free if `isZero` == true
- comparison done with `isZero` instead of `== 0`


JayShortway (approved):
Good catch!

PR #5176, opened 2025-05-26
Add accessibility identifier to PurchaseCardView

facumenzella:
### Motivation
While writing tests to verify the state of the card after a successful / unsuccessful promo offer, I need an id to validate multiple subs.

### Description
- Add identifier to card

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/ScrollViewSection.swift
+++ b/RevenueCatUI/CustomerCenter/Views/ScrollViewSection.swift
@@ -67,7 +67,6 @@ struct PurchasesInformationSection: View {
                     )
                     .cornerRadius(10)
                     .padding(.horizontal)
-                    .accessibilityIdentifier("title_\(id)")
                 }
                 .padding(.bottom, 16)
             }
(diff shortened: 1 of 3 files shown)


RevenueCatUI/CustomerCenter/Views/ScrollViewSection.swift
```diff
@@ -54,12 +54,16 @@ struct PurchasesInformationSection: View {
     var body: some View {
         ScrollViewSection(title: title) {
             ForEach(items) { purchase in
+                // only a few active subscriptions
```
ajpallares, line 57: Wouldn't it be possible to do `ForEach(items.enumated())` to get the index directly? 

I think that's going to be possible with Swift 6.2, but I may be wrong 🤔

ajpallares (approved):
Looks good! Thanks!
Just one small comment

RevenueCatUI/CustomerCenter/Views/ScrollViewSection.swift
ajpallares: I think it makes sense!

facumenzella:
Moving forward with this @ajpallares

PR #5178, opened 2025-05-26
Introduce NoSubscriptions card view for empty states

facumenzella:
### Motivation
Replace the old design with a new one, that matches the purchases card view style

### Description
- Add new card
- Introduce new card in `NoSubscriptionsView`


RevenueCatUI/CustomerCenter/Views/RelevantPurchasesListView.swift
```diff
@@ -120,47 +120,33 @@ struct RelevantPurchasesListView: View {
...
-                            .fill(
-                                Color(colorScheme == .light
-                                      ? UIColor.systemBackground
-                                      : UIColor.secondarySystemBackground)
-                            )
-                            .padding(.horizontal)
-                            .padding(.top)
-                    )
-                    .padding(.bottom, 32)
+                if !customerInfoViewModel.activeSubscriptionPurchases.isEmpty {
```
vegaro, line 123: This is changing from changing subs + non-subs to only subs. What's the reason?

facumenzella: The empty state is for subscriptions now, and since there are two sections we only put it if there are no subs

vegaro: I see, makes sense

vegaro (approved):

PR #5179, opened 2025-05-26
Pass productIdentifier for promo offer flow

facumenzella:
### Motivation
While working on maestro tests, I noticed that the id was being picked from customer info, instead of the context.

### Description
- Use view model product identifier to accept promo offer

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/FeedbackSurveyData.swift
+++ b/RevenueCatUI/CustomerCenter/Data/FeedbackSurveyData.swift
@@ -24,13 +24,19 @@ import RevenueCat
 @available(watchOS, unavailable)
 struct FeedbackSurveyData: Equatable {
 
+    let productIdentifier: String
     let configuration: CustomerCenterConfigData.HelpPath.FeedbackSurvey
     let path: CustomerCenterConfigData.HelpPath
     let onOptionSelected: (() -> Void)
 
-    init(configuration: CustomerCenterConfigData.HelpPath.FeedbackSurvey,
-         path: CustomerCenterConfigData.HelpPath,
-         onOptionSelected: @escaping (() -> Void)) {
+    init(
+        productIdentifier: String,
+        configuration: CustomerCenterConfigData.HelpPath.FeedbackSurvey,
+        path: CustomerCenterConfigData.HelpPath,
+        onOptionSelected: @escaping (() -> Void
+        )
+    ) {
+        self.productIdentifier = productIdentifier
         self.configuration = configuration
         self.path = path
         self.onOptionSelected = onOptionSelected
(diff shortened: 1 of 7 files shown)


RevenueCatUI/CustomerCenter/Data/FeedbackSurveyData.swift
```diff
@@ -24,21 +24,27 @@ import RevenueCat
...
+        self.productIdentifier = productIdentifier
         self.configuration = configuration
         self.path = path
         self.onOptionSelected = onOptionSelected
     }
 
     static func == (lhs: FeedbackSurveyData, rhs: FeedbackSurveyData) -> Bool {
         return lhs.configuration == rhs.configuration &&
-               lhs.path == rhs.path
+        lhs.path == rhs.path
```
ajpallares, line 47: Should we also compare the `productIdentifier` here for equality?

RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
```diff
@@ -100,6 +100,7 @@ class BaseManageSubscriptionViewModel: ObservableObject {
         switch path.detail {
         case let .feedbackSurvey(feedbackSurvey):
             self.feedbackSurveyData = FeedbackSurveyData(
+                productIdentifier: purchaseInformation?.productIdentifier ?? "-",
```
ajpallares, line 103: Instead of defaulting to `"-"`, would it make sense for `FeedbackSurveyData`'s `productIdentifier` to be `optional`? This way, in the changed lines below, we would treat the `nil` case as the ineligible case 
```swift
if promotionalOffer.eligible, let productIdentifier = feedbackSurveyData?.productIdentifier {
... // Would not enter here if `productIdentifier` were `nil`
}
```

and the same would apply in `FeedbackSurveyViewModel` around line 70
```swift
if let promotionalOffer = option.promotionalOffer,
           promotionalOffer.eligible,
           let productIdentifier = feedbackSurveyData.productIdentifier {
... // Would not enter here if `productIdentifier` were `nil`
}
```

RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
```diff
@@ -108,9 +109,12 @@ class BaseManageSubscriptionViewModel: ObservableObject {
                 }
 
         case let .promotionalOffer(promotionalOffer) where purchaseInformation?.store == .appStore:
-            if promotionalOffer.eligible {
+            if promotionalOffer.eligible, let productIdentifier = feedbackSurveyData?.productIdentifier {
```
ajpallares, line 112: Is it always the case the the promotionalOffer path comes from a feedback survey? Just to be sure

ajpallares (commented):
I mostly looks good to me. But I would like clarification of some aspects before approving

RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
facumenzella: Yes, you first need to explain the reason

ajpallares (approved):

PR #5184, opened 2025-05-27
Add missing fields to PurchaseInformation

facumenzella:
### Motivation
The goal is to re-write PurchaseHistory using `PurchaseInformation` instead of `PurchaseInfo` so we can reuse all the UI from `RelevantPurchasesListView`

### Description
- Add missing fields

### Discussion
The more I see `PurchaseInformation`, I think that we could go to something that just wraps either `SubscriptionInfo` o `NonSubscriptionTransaction`, and an optional `StoreProduct` + `Entitlement. That should reduce a lot of code.

I'll probably explore this once PurchaseHistory is refactored.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
+++ b/RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
@@ -81,12 +81,36 @@ struct PurchaseInformation {
     /// Note: `nil` for non-subscriptions, and for expiring subscriptions
      let renewalDate: Date?
 
+    /// The date an unsubscribe was detected.
+    ///
+    /// Note: `nil` for non-subscriptions, and for expiring subscriptions
+    let unsubscribeDetectedAt: Date?
+
+    /// Date when RevenueCat detected any billing issues with this subscription.
+    ///
+    /// Note: `nil` for non-subscriptions, and for expiring subscriptions
+    let billingIssuesDetectedAt: Date?
+
+    /// Date when any grace period for this subscription expires/expired.
+    /// nil if the customer has never been in a grace period.
+    ///
+    /// Note: `nil` for non-subscriptions, and for expiring subscriptions
+    let gracePeriodExpiresDate: Date?
+
+    let refundedAtDate: Date?
+
     /// Product specific management URL
     let managementURL: URL?
 
     let periodType: PeriodType
 
     let ownershipType: PurchaseOwnershipType?
+
+    /// The unique identifier for the transaction created by RevenueCat.
+    let transactionIdentifier: String?
+
+    /// The unique identifier for the transaction created by the Store.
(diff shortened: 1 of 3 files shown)


RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -77,13 +77,37 @@ struct PurchaseInformation {
...
+    /// Note: `nil` for non-subscriptions, and for expiring subscriptions
+    let billingIssuesDetectedAt: Date?
+
+    /// Date when any grace period for this subscription expires/expired.
+    /// nil if the customer has never been in a grace period.
+    ///
+    /// Note: `nil` for non-subscriptions, and for expiring subscriptions
+    let gracePeriodExpiresDate: Date?
+
+    let refundedAtDate: Date?
```
tonidero, line 96: I guess this is not public so it doesn't matter but any reason this has no docs when the others do?

RevenueCatUI/CustomerCenter/Data/Transaction.swift
```diff
@@ -24,6 +24,12 @@ protocol Transaction {
     var price: ProductPaidPrice? { get }
     var periodType: PeriodType { get }
     var purchaseDate: Date { get }
+    var unsubscribeDetectedAt: Date? { get }
+    var billingIssuesDetectedAt: Date? { get }
+    var gracePeriodExpiresDate: Date? { get }
+    var refundedAtDate: Date? { get }
+    var storeIdentifier: String? { get }
+    var identifier: String? { get }
```
tonidero, line 32: I have been wondering if we should move some of these improvements to the `CustomerInfo`... but I guess that can be a separate PR if needed.

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
facumenzella: https://github.com/RevenueCat/purchases-ios/pull/5184/commits/91338eabdf190ab04f52564e7f84bba57cc5e0af

RevenueCatUI/CustomerCenter/Data/Transaction.swift
facumenzella: I've been thinking on renaming this to something like `AnyPurchase` to disambiguate it from the existing `Transaction` struct. I can open a draft after I wrap this up and see how it looks

tonidero: No hurry, I just thought trying to join the different transaction types into one like this could be useful... But not a problem for this PR.

Also, personally I would prefer something more related to `Transaction` than to Purchase... Mostly because `Purchase` as a term already has a lot of meanings.

PR #5187, opened 2025-05-27
Use combine instead of view modifiers internally

facumenzella:
### Motivation
While having a debugging session with @joshdholtz and @ajpallares we agreed that using preference keys for things that are not SwiftUI 100% makes no sense, and it's flaky. This is an attempt to use combine internally instead of preference keys.

Removing preference keys for the public API is trickier, because the API is deeply tied to that.

### Description
- Add methods to `CustomerCenterActionWrapper` to subscribe to updates
- Make existing `PassthroughSubjects` private
- Move business logic to the view model, instead of the view.
- Remove preference keys in favor of a environment object that holds functions


RevenueCatUI/CustomerCenter/Data/CustomerCenterActionViewModifier.swift
```diff
@@ -48,14 +48,12 @@ struct CustomerCenterActionViewModifier: ViewModifier {
...
 
-    @State private var cancellables = Set<AnyCancellable>()
-
     func body(content: Content) -> some View {
         content
             .onAppear {
                 subscribeToActionWrapper()
             }
-            // Apply preferences based on state
+        // Apply preferences based on state
```
ajpallares: Maybe unintentional change?

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionWrapper.swift
```diff
@@ -151,3 +153,85 @@ extension CustomerCenterConfigData.HelpPath {
...
 
 }
+
+#if os(iOS)
+
+/// Internal Combine-based wiring for handling `CustomerCenterInternalAction` events.
+/// These methods attach handlers to internal publishers and are used internally
+/// by any ViewModel to bridge actions to business logic.
+///
+/// - Important: This API is intended **for internal use only** within the SDK.
```
ajpallares, line 175: Why this scare message? 😆

RevenueCatUI/CustomerCenter/ViewModels/SubscriptionDetailViewModel.swift
```diff
@@ -69,14 +69,32 @@ final class SubscriptionDetailViewModel: BaseManageSubscriptionViewModel {
...
+
+        actionWrapper.onCustomerCenterShowingManageSubscriptions { [weak self] in
+            self?.customerInfoViewModel.manageSubscriptionsSheet = true
+        }
+
+        customerInfoViewModel.$manageSubscriptionsSheet
+            .filter { !$0 }
+            .sink { [weak self] _ in
+                self?.refreshPurchase()
+            }
```
ajpallares: maybe confusing use of `filter` here. Why not just this?
```suggestion
            .sink { [weak self] _ in
                if !$0 {
                    self?.refreshPurchase()
                }
            }
```

RevenueCatUI/CustomerCenter/ViewModels/SubscriptionDetailViewModel.swift
```diff
@@ -69,14 +69,32 @@ final class SubscriptionDetailViewModel: BaseManageSubscriptionViewModel {
...
         }
 
+    func didAppear() {
+        actionWrapper.onCustomerCenterPromotionalOfferSuccess { [weak self] in
+            self?.refreshPurchase()
+        }
+
+        actionWrapper.onCustomerCenterShowingManageSubscriptions { [weak self] in
+            self?.customerInfoViewModel.manageSubscriptionsSheet = true
+        }
```
ajpallares: Maybe this wouldn't happen right now, but I'm afraid that, if we ever make a navigation/modal from this screen and `didAppear()` could be called again when back on this screen, we would have repeated callbacks, as the old ones are not cancelled (I think).

For that reason, I think the subscriber (in this case, this ViewModel), should be the one managing the `cancellable`, so that at the beginning of `didAppear()` we can cancel all the previous `cancellables` and re-add them

ajpallares: And the subjects can (and should) still be `private` as done in this PR. But we should create a type-erased publisher for each subject that is the one used by subscribers. Something like this

```swift
private let restoreStarted = PassthroughSubject<Void, Never>()
var onRestoreStarted: AnyPublisher<Void, Never> {
    restoreStarted.eraseToAnyPublisher()
}
```

This way, subscribers would register to the `onRestoreStarted` publisher and would be able to handle cancellation by themselves.

ajpallares (commented):
I have some initial comments I left in the code. In addition, I see that `CustomerCenterActionViewModifier` still seems to rely on the `.preferences` modifier. Probably I'm understanding it wrong, but isn't this PR supposed to remove the use of `.preferences`?

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionWrapper.swift
facumenzella: I am thinking on how to 100% delete the preference use. The tricky part is that for the public API, we rely on preference keys so I am not sure how to avoid it yet.

One idea I have is using environment functions instead of preference keys, but that's not great either.

ajpallares: Oh, right. I see what you mean... 🤔 
I really wouldn't know how to easily solve that, tbh. I do like the idea of environment functions **but** I'm fear/wonder that it could have the same limitations as `preference` keys...? I don't really know, but yeah, we should perhaps try that approach

RevenueCatUI/CustomerCenter/Actions/CustomerCenterView+Actions.swift
```diff
@@ -64,144 +64,100 @@ extension CustomerCenterView {
...
-
         func body(content: Content) -> some View {
-            content
-                .onPreferenceChange(ShowingManageSubscriptionsPreferenceKey.self) { wrappedIsShowing in
-                    if wrappedIsShowing != nil {
-                        self.handler()
-                    }
-                }
+            content.transformEnvironment(\.customerCenterActions) { actions in
+                actions.addShowingManageSubscriptions(handler)
```
facumenzella: I think this should be `set` instead of `add`

Examples/rc-maestro/rc-maestro/Sources/rc-maestro/ContentView.swift
```diff
@@ -38,12 +38,19 @@ public struct ContentView: View {
...
                     )
                 )
+                .onCustomerCenterRestoreStarted {
+                    print("🙌 Restore started")
+                }
+                .onCustomerCenterRestoreCompleted { customerInfo in
+                    print("🙌 Restore completed")
+                }
+                .onCustomerCenterCustomActionSelected { actionIdentifier, purchaseIdentifier in
+                    print("🙌 Custom Action")
```
ajpallares, line 48: Total nit, but it could be useful to add to the print the `actionIdentifier` and/or `purchaseIdentifier`? NABD, since this is from the maestro test project. Something like
```suggestion
                    print("🙌 Custom Action selected: \(actionIdentifier) for purchase \(purchaseIdentifier)")
```

RevenueCatUI/CustomerCenter/Actions/CustomerCenterView+Actions.swift
```diff
@@ -66,142 +66,105 @@ extension CustomerCenterView {
         let handler: RestoreStartedHandler
 
         func body(content: Content) -> some View {
-            content
-                .onPreferenceChange(RestoreStartedPreferenceKey.self) { wrappedStarted in
-                    if wrappedStarted != nil {
-                        self.handler()
-                    }
-                }
+            content.transformEnvironment(\.customerCenterActions) { actions in
```
ajpallares, line 69: Interesting way to use `transformEnvironment`. I didn't know it, but I think this makes sense

RevenueCatUI/CustomerCenter/Actions/CustomerCenterView+Actions.swift
```diff
@@ -360,12 +323,26 @@ extension View {
...
     }
 
+    /// Invokes the given closure when a promotional offer succeeds
+    /// ```swift
+    /// CustomerCenterView()
+    ///     .onCustomerCenterPromotionalOfferSuccess {
+    ///         // handle succees
+    ///     }
+    /// ```
     func onCustomerCenterPromotionalOfferSuccess(
```
ajpallares, line 333: I see you added the documentation there. Is it because this needs to be `public`?

RevenueCatUI/CustomerCenter/Actions/CustomerCenterView+Actions.swift
```diff
@@ -360,12 +323,26 @@ extension View {
...
     }
 
+    /// Invokes the given closure when change plan button is selected
+    /// ```swift
+    /// CustomerCenterView()
+    ///     .onCustomerCenterChangePlansSelected { optionId in
+    ///         // handle change plan
+    ///     }
+    /// ```
     func onCustomerCenterChangePlansSelected(
```
ajpallares, line 346: Same here. Should this be `public`?

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionWrapper.swift
```diff
@@ -166,3 +165,91 @@ extension CustomerCenterConfigData.HelpPath {
...
+        return showingChangePlans.sink(receiveValue: handler)
+    }
+
+    func onCustomerCenterCustomActionSelected(
+        _ handler: @escaping (String, String?) -> Void
+    ) -> AnyCancellable {
+        return customActionSelected
+            .sink { (identifier, purchaseId) in
+                handler(identifier, purchaseId)
+            }
```
ajpallares, line 251: For consistency, perhaps this?
```suggestion
            .sink(receiveValue: handler)
```

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionWrapper.swift
```diff
@@ -166,3 +165,91 @@ extension CustomerCenterConfigData.HelpPath {
...
+        return refundRequestStarted.sink(receiveValue: handler)
+    }
+
+    func onCustomerCenterRefundRequestCompleted(
+        _ handler: @escaping (String, RefundRequestStatus) -> Void
+    ) -> AnyCancellable {
+        return refundRequestCompleted
+            .sink { (productId, status) in
+                handler(productId, status)
+            }
```
ajpallares, line 218: Perhaps this as the others?
```suggestion
            .sink(receiveValue: handler)
```

RevenueCatUI/CustomerCenter/Data/CustomerCenterActionWrapper.swift
ajpallares, line 1: If I'm not mistaken, the usual implementation for this kind of thing (where you have the `private` `PassthroughSubject` and you need to expose a type-erased publisher is to do this:

```swift
private let _restoreStarted = PassthroughSubject<Void, Never>()
var restoreStarted: AnyPublisher<Void, Never> {
    return self._restoreStarted.eraseToAnyPublisher()
}
```

From the caller's side, they just need to subscribe to `restoreStarted`
```swift
restoreStarted.sink(receiveValue: handler).store(in: &cancellables)
```

One could argue that your approach with the wrapper method
```swift 
func onCustomerCenterRestoreStarted(
        _ handler: @escaping () -> Void
    ) -> AnyCancellable {
        return restoreStarted.sink(receiveValue: handler)
    }
```
serves as a way to erase the Combine implementation of this. But, it does not really, since the method is still returning `AnyCancellable` (i.e. from Combine). In addition, the advantage of having a `AnyPublisher` is that the caller can decide how they want to consume the publishers (simple `sink`? or perhaps more complex operations such as `zip`, `throttle`, etc.?)

RevenueCatUI/CustomerCenter/Data/CustomerCenterEnvironment.swift
```diff
@@ -92,3 +92,142 @@ extension EnvironmentValues {
...
+    var feedbackSurveyCompleted: @MainActor @Sendable (String) -> Void = { _ in }
+    var managementOptionSelected: @MainActor @Sendable (CustomerCenterActionable) -> Void = { _ in }
+    var promotionalOfferSuccess: @MainActor @Sendable () -> Void = {}
+    var changePlansSelected: @MainActor @Sendable (String) -> Void = { _ in }
+    var customActionSelected: @MainActor @Sendable (String, String?) -> Void = { _, _ in }
+
+    // Simple chaining helpers (no IDs)
+    func addRestoreStarted(_ handler: @escaping @MainActor @Sendable () -> Void) {
+        self.restoreStarted = handler
+    }
```
ajpallares, line 121: This implementation is dangerous, in the sense that multiple calls to `addRestoreStarted` only keeps the latest `handler` passed. However, from the `public` side, we have the API to be
```swift
/// Invokes the given closure when a restore begins in the Customer Center.
public func onCustomerCenterRestoreStarted(
    _ handler: @escaping CustomerCenterView.RestoreStartedHandler
) -> some View {
    return self.modifier(CustomerCenterView.OnRestoreStartedModifier(handler: handler))
}
```

So, actually, with the current implementation, **only the latest given closure** would be invoked.

In fact, the word `add` in the name method could be misleading in this sense. 

I _think_ we could solve this by making the properties arrays instead. Like this:

```swift
var restoreStarted: [@MainActor @Sendable () -> Void] = []
```

That would make sure that any callback added via `public` modifiers is indeed called (even if a modifier gets repeated more than once).

ajpallares: As a follow up of my previous comment, I would unify the methods' prefixes to be `add` (if we use the array approach) or to `set` if we're overriding the handlers.

ajpallares (commented):
Sorry I missed this PR for so long 😓

I have... comments 😅. Some of them I think are important, so I can retake the review once we address them!

RevenueCatUI/CustomerCenter/Actions/CustomerCenterView+Actions.swift
facumenzella: This will eventually be public, but not until we refactor the whole preference key handling

facumenzella:
Closing in favor of the new PR

PR #5198, opened 2025-05-28
Show most recent expired if no active subscriptions

facumenzella:
### Motivation
If there are no active subscriptions, we want to show the last expired one.

### Description
- Rename `activeSubscriptionPurchases` to `subscriptionsSection`
- Rename `activeNonSubscriptionPurchases` to `nonSubscriptionsSection`
- Add expired badge to `PurchaseInformationCardView`
- Removed some duplicated snapshots


RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -227,34 +227,62 @@ private extension CustomerCenterViewModel {
...
             )
-
             activeNonSubscriptionPurchases.append(purchaseInfo)
         }
-        self.activeNonSubscriptionPurchases = activeNonSubscriptionPurchases
+        self.nonSubscriptionsSection = activeNonSubscriptionPurchases
     }
 
-    func loadActiveSubscriptions(customerInfo: CustomerInfo) async {
+    func loadFirstExpirenTransaction(customerInfo: CustomerInfo) async {
```
vegaro:
```suggestion
    func loadFirstExpiredTransaction(customerInfo: CustomerInfo) async {
```

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -266,19 +294,20 @@ private extension CustomerCenterViewModel {
...
-                entitlement: entitlement,
                 customerInfo: customerInfo
             )
 
             activeSubscriptionPurchases.append(purchaseInfo)
         }
 
-        self.activeSubscriptionPurchases = activeSubscriptionPurchases
+        if activeSubscriptionPurchases.isEmpty {
+            // load last expired
```
vegaro: is it last or first? because the function says `first`. I assume it would be the last expired one? in that case we might want to rename the function

facumenzella: Just pushed it https://github.com/RevenueCat/purchases-ios/pull/5198/commits/d9f23826b0dd72242c9c55806ed3c5886b27d751

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -266,19 +294,20 @@ private extension CustomerCenterViewModel {
...
-                entitlement: entitlement,
                 customerInfo: customerInfo
             )
 
             activeSubscriptionPurchases.append(purchaseInfo)
         }
 
-        self.activeSubscriptionPurchases = activeSubscriptionPurchases
+        if activeSubscriptionPurchases.isEmpty {
+            // load most recent expired
```
ajpallares: nit: this comment is redundant I think, since it's perfectly explained by the method signature

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -292,6 +321,20 @@ private extension CustomerCenterViewModel {
...
+        return await createPurchaseInformation(
+            for: transaction,
+            entitlement: entitlement,
+            customerInfo: customerInfo
+        )
+    }
+
     func createPurchaseInformation(for transaction: RevenueCatUI.Transaction,
                                    entitlement: EntitlementInfo?,
                                    customerInfo: CustomerInfo) async -> PurchaseInformation {
```
ajpallares: One more nit. It seems the second `createPurchaseInformation` method is only called from the first one. Therefore, we could reduce this to only one, like this:

```suggestion
    func createPurchaseInformation(
        for transaction: RevenueCatUI.Transaction,
        customerInfo: CustomerInfo
    ) async -> PurchaseInformation {
        let entitlement = customerInfo.entitlements.all.values
            .first(where: { $0.productIdentifier == transaction.productIdentifier })
```
and then continue with the implementation of the second one.

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
vegaro: Agreed

facumenzella:
```suggestion
```

PR #5199, opened 2025-05-28
Add isSanbox to NonSubscriptionTransaction

facumenzella:
### Motivation
I've noticed `isSandbox` is not exposed to `NonSubscriptionTransaction`, but it's parsed in `Transaction`

### Description
- Add `isSandbox` to `NonSubscriptionTransaction`

### DIFF
--- a/Sources/Purchasing/NonSubscriptionTransaction.swift
+++ b/Sources/Purchasing/NonSubscriptionTransaction.swift
@@ -40,6 +40,9 @@ public final class NonSubscriptionTransaction: NSObject {
     /// Paid price for the subscription
     @objc public let price: ProductPaidPrice?
 
+    /// Whether or not the purchase was made in sandbox mode.
+    @objc public let isSandbox: Bool
+
     init?(with transaction: CustomerInfoResponse.Transaction, productID: String) {
         guard let transactionIdentifier = transaction.transactionIdentifier,
               let storeTransactionIdentifier = transaction.storeTransactionIdentifier else {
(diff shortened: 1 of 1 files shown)


Sources/Purchasing/NonSubscriptionTransaction.swift
```diff
@@ -40,6 +40,9 @@ public final class NonSubscriptionTransaction: NSObject {
     /// Paid price for the subscription
     @objc public let price: ProductPaidPrice?
 
+    /// Whether or not the purchase was made in sandbox mode.
+    @objc public let isSandbox: Bool
```
tonidero, line 44: Just to make sure, this won't be optional for non-subscription transactions right? Just making sure there won't be any issues parsing this.

tonidero: Also, might be good to add to some test.

tonidero (approved):
Just a couple comments but looking good!

Sources/Purchasing/NonSubscriptionTransaction.swift
facumenzella: I checked the https://www.revenuecat.com/docs/api-v1#tag/customers/operation/subscribers and it's there. Also, wasn't optional for my tests, but worth double-checking

facumenzella: For reference, we've checked the backend and it's there so we're good.

PR #5200, opened 2025-05-28
SampleCat: A new iOS Sample app that guides users through any configuration issues

polpielladev:
### Checklist

### Motivation
In the Developer Experience team, we have been working on a Sample app that would allow users to check their SDK configuration in a way that we can guide them.

To begin with, we have added three main components:
- A full-screen error view to flag issues such as invalid API key or mismatching bundle IDs.
- An Offerings view that shows users whether their configuration is correct or not.
- A products view that shows statuses for each product.

The app also has the ability to purchase both products and packages and shows which products have been purchased.

https://github.com/user-attachments/assets/25996b1f-5734-4e3a-b4d8-414cf0dde4ff


facumenzella:
@polpielladev have you considered introducing tuist for this? Just to avoid adding yet another project to the workspace.
I guess it depends on what will be the main use case. Is it expected for devs to run this directly? Or it's just a distributed app to test. If it's the latter, I think tuist makes a lot of sense

polpielladev:
> @polpielladev have you considered introducing tuist for this? Just to avoid adding yet another project to the workspace. I guess it depends on what will be the main use case. Is it expected for devs to run this directly? Or it's just a distributed app to test. If it's the latter, I think tuist makes a lot of sense
> 
> Definitely not a blocker 🤓

That is a very good point, the main use case will be that we vend this from the dashboard as a standalone project that people can just grab and run on their machines. The Back End will inject a bunch of data into the app:

- Bundle ID
- Api key for project
- Entitlement Identifier
- StoreKit configuration file

This should hopefully be the fastest way for users to check your configuration. I think to reduce complexity for users to get started, let's keep the Xcode project simple and not introduce another dependency that the user needs to have. I like the idea of tuist, but since we will vend the project stand-alone, I think it makes sense to keep it as simple as possible.

Examples/SampleCat/SampleCat/Constants.swift
```diff
@@ -0,0 +1,18 @@
...
+    /*
+     The API key for your app from the RevenueCat dashboard: https://app.revenuecat.com
+     */
+    #error("Modify this property to reflect your app's API key, then comment this line out.")
+    static let apiKey = "REVENUECAT_API_KEY"
+    
+    /*
+     The entitlement identifier from the RevenueCat dashboard that is activated upon successful in-app purchase for the duration of the purchase.
+     */
+    #error("Modify this property to reflect your app's entitlement identifier, then comment this line out. If you do not use entitlements, you can set this property to nil.")
```
tonidero: Nitpick but should we ask them to remove this?

```suggestion
    #error("Modify this property to reflect your app's API key, then remove this line out.")
    static let apiKey = "REVENUECAT_API_KEY"
    
    /*
     The entitlement identifier from the RevenueCat dashboard that is activated upon successful in-app purchase for the duration of the purchase.
     */
    #error("Modify this property to reflect your app's entitlement identifier, then remove this line out. If you do not use entitlements, you can set this property to nil.")
```

Examples/SampleCat/SampleCat/Screens/Entitlements/EntitlementsView.swift
```diff
@@ -0,0 +1,23 @@
...
+import SwiftUI
+
+struct EntitlementsView: View {
+    @Environment(UserViewModel.self) private var userViewModel
+
+    var body: some View {
+        NavigationStack {
+            Text("Entitlements")
+        }
+        .navigationTitle("Entitlements")
```
tonidero: I'm wondering, is this best practice in SwiftUI for localization? I wonder how complex we want to get on these things, if devs might take this as a base app to implement their own, ideally we follow best practices for all these things so just making sure.

Examples/SampleCat/SampleCat/Screens/Entitlements/EntitlementsView.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  EntitlementsView.swift
+//  SampleCat
+//
+//  Created by Hidde van der Ploeg on 7/4/25.
```
tonidero: Total nitpick but if this might be downloadable by users, should we remove app/developer names? Or I guess we can also generate them 🤷

Examples/SampleCat/SampleCat/Screens/Products/ProductCell.swift
```diff
@@ -0,0 +1,108 @@
...
+                .foregroundStyle(.secondary)
+            Text(product.description)
+                .font(.caption)
+
+            Button(action: {
+                guard let storeProduct = self.product.storeProduct else { return }
+
+                Task {
+                    isPurchasing = true
+                    await userViewModel.purchase(storeProduct)
```
tonidero: FFTI, total nitpick... but I wonder if this should be part of the `ProductViewModel`. Feels like a better place... but not too concerned about it if there is a reason to keep it here.

Examples/SampleCat/SampleCat/Screens/Offerings/OfferingCell.swift
```diff
@@ -0,0 +1,32 @@
...
+            }
+            .font(.headline)
+            .symbolRenderingMode(.hierarchical)
+
+            if offering.products.isEmpty {
+                Text("No packages")
+                    .font(.footnote)
+                    .foregroundStyle(.secondary)
+            } else {
+                Text("\(offering.products.count) package\(offering.products.count > 1 ? "s" : "")")
```
tonidero: Hmm I'm a bit confused here... should we be checking and counting the packages here, not the products? (which should normally match I guess)

Examples/SampleCat/SampleCat/Screens/Offerings/OfferingPackagesView.swift
```diff
@@ -0,0 +1,25 @@
...
+
+    var body: some View {
+        ScrollView {
+            ConceptIntroductionView(imageName: "visual-offerings",
+                                    title: "Packages",
+                                    description: "Packages are a representation of the products that you “offer” to customers on your paywall.")
+
+            VStack {
+                ForEach(offering.products) { product in
+                    ProductCell(product: product)
```
tonidero: Same here about using products instead of packages here

Examples/SampleCat/SampleCat/UserViewModel.swift
```diff
@@ -0,0 +1,82 @@
...
+            }
+        }
+    }
+
+    func purchase(_ product: StoreProduct) async {
+        isPurchasing = true
+        defer { isPurchasing = false }
+
+        do {
+            let (_, customerInfo, userCancelled) = try await Purchases.shared.purchase(product: product)
```
tonidero, line 42: Should we try to move devs to use `purchase(package:)` rather than this one? We potentially miss a lot of data if the developer doesn't get the product from an offering

Examples/SampleCat/SampleCat.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,381 @@
...
+		};
+/* End XCConfigurationList section */
+
+/* Begin XCRemoteSwiftPackageReference section */
+		53E9CF982DAD41F5000285C6 /* XCRemoteSwiftPackageReference "purchases-ios" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/RevenueCat/purchases-ios";
+			requirement = {
+				branch = main;
+				kind = branch;
```
ajpallares, line 510: Just to be sure, do we want this app to depend on `purchases-ios` `main` or perhaps it should depend on the latest release?

Examples/SampleCat/SampleCat/HealthViewModel.swift
```diff
@@ -0,0 +1,78 @@
...
+
+        func fetchHealthReport() async {
+            defer { isfetchingHealthReport = false }
+            isfetchingHealthReport = true
+
+            let report = await PurchasesDiagnostics.default.healthReport()
+            if case let .unhealthy(error) = report.status {
+                blockingError = error
+                return
+            }
```
ajpallares, line 51: Just to be sure. We're not doing anything with the warnings (i.e. not covering the `SDKHealthStatus.healthy` status). I guess this is ok for the first version?

Examples/SampleCat/SampleCat/Screens/Entitlements/EntitlementsView.swift
ajpallares: The thing is... this is how things are localized in SwiftUI. `"Entitlements"` here is actually a [`LocalizedStringKey`](https://developer.apple.com/documentation/swiftui/localizedstringkey)

Examples/SampleCat/SampleCat/Screens/Entitlements/EntitlementsView.swift
ajpallares, line 1: I think `EntitlementsView` is not being used. If it should be used, then its implementation seems incomplete.

ajpallares (commented):
Some initial comments

Examples/SampleCat/SampleCat/Screens/FullScreenErrorView.swift
```diff
@@ -0,0 +1,66 @@
...
+                    .frame(maxWidth: .infinity)
+                    .padding(.vertical, 8)
+            })
+            .padding(.horizontal)
+            .buttonStyle(.borderedProminent)
+            .tint(.accent)
+        }
+        .background(.red.opacity(0.1))
+        .animation(.default, value: healthViewModel.isfetchingHealthReport)
+        .scrollBounceBehavior(.basedOnSize)
```
ajpallares: `.scrollBounceBehavior(.basedOnSize)` also appears in line 37. Is this duplication intentional?

Examples/SampleCat/SampleCat/Screens/Offerings/OfferingCell.swift
ajpallares: I believe this is actually the packages. But the `ProductViewModel` is being used (abused perhaps?) to be valid for both a package and a product. Based on what I see in HealthViewModel.swift, there's
```swift
offerings = report.offerings.map { offering in
    OfferingViewModel(
        identifier: offering.identifier,
        status: offering.status,
        products: offering.packages.map { package in
            ProductViewModel(
                id: package.identifier,
                status: package.status,
                title: package.productIdentifier,
                description: package.description,
                storeProduct: storeProducts[package.productIdentifier]
            )
        }
    )
}
products = reportProducts.map { product in
    ProductViewModel(
        id: product.identifier,
        status: product.status,
        title: product.title,
        description: product.description,
        storeProduct: storeProducts[product.identifier]
    )
}
```

which means, in this case, the `OfferingViewModel.products` actually refers to packages. Perhaps `OfferingViewModel.products` should be called `OfferingViewModel.packages`? But then the `ProductViewModel` type would be confusing

Examples/SampleCat/SampleCat/Screens/Products/ProductCell.swift
```diff
@@ -0,0 +1,108 @@
+//
+//  ProductCell.swift
+//  SampleCat
+//
+//  Created by Hidde van der Ploeg on 15/4/25.
+//
```
ajpallares: As Toni mentioned above, and to have consistency with other files, perhaps it would be good to remove the header

Examples/SampleCat/SampleCat/Screens/Products/ProductCell.swift
```diff
@@ -0,0 +1,108 @@
...
+
+import SwiftUI
+
+enum ProductCellState {
+    case readyToPurchase
+    case purchasing
+    case purchased
+    case canNotPurchase
```
ajpallares: Huge nit?
```suggestion
    case cannotPurchase
```

Examples/SampleCat/SampleCat/Screens/Products/ProductCell.swift
```diff
@@ -0,0 +1,108 @@
...
+
+                Task {
+                    isPurchasing = true
+                    await userViewModel.purchase(storeProduct)
+                    isPurchasing = false
+                }
+            }) {
+                Group {
+                    if state == .purchased {
+                        Label(purchaseButtonTitle, systemImage: purchaseButtonIcon)
```
ajpallares: Nit: have the system image name here directly as it's only being used here
```suggestion
                        Label(purchaseButtonTitle, systemImage: "checkmark.circle.fill")
```

Examples/SampleCat/SampleCat/Screens/Products/ProductsView.swift
```diff
@@ -0,0 +1,53 @@
+//
+//  ProductsView.swift
+//  SampleCat
+//
+//  Created by Hidde van der Ploeg on 7/4/25.
+//
```
ajpallares: Same comment about headers

Examples/SampleCat/SampleCat/Utils/Extensions/AppHealth+Ext.swift
```diff
@@ -0,0 +1,47 @@
+//
+//  AppHealth+Ext.swift
+//  SampleCat
+//
+//  Created by Hidde van der Ploeg on 7/4/25.
+//
```
ajpallares: Extra header here

Examples/SampleCat/SampleCat/Utils/Views/ConceptIntroductionView.swift
```diff
@@ -0,0 +1,37 @@
+//
+//  ConceptIntroductionView.swift
+//  SampleCat
+//
+//  Created by Hidde van der Ploeg on 15/4/25.
+//
```
ajpallares: Extra header

Examples/SampleCat/SampleCat/Utils/Views/ContentBackgroundView.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  ContentBackgroundView.swift
+//  SampleCat
+//
+//  Created by Hidde van der Ploeg on 15/4/25.
+//
```
ajpallares: Header

ajpallares (commented):
Added some more comments!

Examples/SampleCat/SampleCat/Constants.swift
polpielladev: Good point! I have rephrased the error now 👍

Examples/SampleCat/SampleCat/Screens/Entitlements/EntitlementsView.swift
polpielladev: Yeah SwiftUI views are localized by default unless you use the `verbatim:` variable. This should be the best practice already as we could localize in the future to multiple languages and not have to change much in the views 🚀

tonidero: Sounds good... I've always found SwiftUI localization confusing... Some people might want to have their "default" languages be a different one for one... And having plain strings doesn't provide a lot of security on making sure everything is localized correctly I think... But I'm ok with this since I agree it seems to be best pracice.

Examples/SampleCat/SampleCat.xcodeproj/project.pbxproj
polpielladev: We should depend on `main` by default, and there is a [`fastlane` lane that updates this reference whenever we cut a new release to point to that commit hash](https://github.com/RevenueCat/purchases-ios/blob/main/fastlane/Fastfile#L676)

I have updated that lane to include this project too in this PR 👍

Examples/SampleCat/SampleCat/HealthViewModel.swift
polpielladev: Yeah all good for v1, this is the top level status, we do handle warnings at offering and product level, so it's all good 👍

Examples/SampleCat/SampleCat/Screens/Entitlements/EntitlementsView.swift
polpielladev: Good point, deleted it now ✂️

Examples/SampleCat/SampleCat/Screens/FullScreenErrorView.swift
polpielladev: Not intentional, I did change the order of the SwiftUI views, and this one lingered. Removed it now ✂️

Examples/SampleCat/SampleCat/Screens/Entitlements/EntitlementsView.swift
polpielladev: Very good point, just removed all headers now ✂️

Examples/SampleCat/SampleCat/Screens/Products/ProductCell.swift
polpielladev: Good point, fixed now ✅

Examples/SampleCat/SampleCat/Screens/Products/ProductCell.swift
polpielladev: Done! ✅

Examples/SampleCat/SampleCat/Screens/Products/ProductCell.swift
polpielladev: Removed ✂️

Examples/SampleCat/SampleCat.xcodeproj/project.pbxproj
ajpallares: Oh, that's great! And TIL 😅

ajpallares (approved):
There are some comments from Toni that are worth clarifying. But other than that, I think this is good to go. Great job! 🚀

polpielladev:
@tonidero @ajpallares I have updated the codebase a bit so that we actually purchase packages when we have offerings and we purchase raw products otherwise, as it will be clearer for the user. I have also furthered the divide between `UserViewModel` and `HealthViewModel` and explicitly added comments to make sure users know that they should follow `UserViewModel` for production purposes.

Let me know if you are happy for me to go ahead with this! 🚀

Examples/SampleCat/SampleCat/HealthViewModel.swift
```diff
@@ -0,0 +1,122 @@
...
+                        await self.userViewModel.purchase(storeProduct)
+                    }
+                )
+            }
+        }
+
+        private func buildOfferingViewModels(from reportOfferings: [PurchasesDiagnostics.OfferingDiagnosticsPayload]) async -> [OfferingViewModel] {
+            await userViewModel.fetchOfferings()
+
+            let sdkOfferings = userViewModel.offerings?.all
```
vegaro: nitpick, but can this just be `offerings`? Is the  intention of the `sdk` prefix to mean "what the SDK loaded"?

Maybe it could be `loadedOfferings`. Same for `sdkPackage`

Examples/SampleCat/SampleCat/Screens/ContentView.swift
```diff
@@ -0,0 +1,28 @@
...
+                }
+        }
+        .fullScreenCover(item: $healthViewModel.blockingError) { error in
+            FullScreenErrorView(error: error)
+        }
+    }
+}
+
+#Preview {
+    ContentView()
```
vegaro: These previews don't work if there's no API key. I think it would be cool to be able to use emerge in this app as well.

Examples/SampleCat/SampleCat/Screens/Views/PurchasableCell.swift
```diff
@@ -0,0 +1,91 @@
...
+                .padding(.horizontal, 16)
+                .background(Color.accent)
+                .foregroundStyle(.white)
+                .clipShape(Capsule())
+            }
+            .buttonStyle(.plain)
+            .padding(.top, 16)
+            .disabled(state != .readyToPurchase)
+
+            if purchasable.purchasable == nil {
```
vegaro: This reads a bit weird. Maybe it would be better to rename the underlying purchasable to `purchaseItem`? Or something along those lines

Examples/SampleCat/SampleCatStoreKitConfiguration.storekit
vegaro, line 1: just curious but how will this be used in the flow? I thought the point of the app would be to try to fetch the products from real StoreKit and try to detect issues?

vegaro (commented):
This is looking good! My comments are mainly nitpicks around variable naming. I guess my main concern would be that the previews don't work without an API key

Examples/SampleCat/SampleCat/HealthViewModel.swift
polpielladev: Yeah, there is a split between what we get from the health report and what we get back from the SDK, we do use it to inform the user about configuration issues.

I will rename to `loadedOfferings` 👍

Examples/SampleCat/SampleCat/Screens/ContentView.swift
polpielladev: That's a good point, I am going to remove the Previews for now, I don't think they will add much value for a Sample App. We can log a ticket and discuss this in a future iteration, to see if we can add them back if people ask for them.

Examples/SampleCat/SampleCat/Screens/Views/PurchasableCell.swift
polpielladev: Ah!! I completely agree! Let me fix that quickly 🏃

polpielladev: Done!

vegaro (approved):

Examples/SampleCat/SampleCatStoreKitConfiguration.storekit
polpielladev: It's a good point, we do get the status from the API for products, so we will still be able to let them know about configuration issues. But we also wanted to provide a way for users to copy/paste a StoreKitConfiguration that is valid for when they begin testing and that they can bring into their app.

Let's see how it goes after we ship, let's try and see if we get any reports, we could potentially remove the StoreKitConfig in the future if it provides confusion.

polpielladev:
@RCGitBot please test

polpielladev:
@RCGitBot please test

polpielladev:
@RCGitBot please test

polpielladev:
@RCGitBot please test

tonidero:
Sorry I'm late to this, but should we add a job in CircleCI to build this app automatically? Can be part of the `all-tests` workflow, if we want to avoid running it on every commit.

PR #5203, opened 2025-05-29
Update CustomerInfo sample with is_sandbox to make sure it works

facumenzella:
### Motivation
Follow-up of https://github.com/RevenueCat/purchases-ios/pull/5199

### Description
Update sample CustomerInfo for tests to make sure decoding is_sandbox works.


ajpallares (approved):
Makes sense!

PR #5206, opened 2025-05-30
add missing import Foundation in Tests

facumenzella:
### Motivation
I don't have all the context on why these are missing, but here they are.

### Description
Add missing imports

### DIFF
--- a/Tests/UnitTests/Mocks/MockSystemInfo.swift
+++ b/Tests/UnitTests/Mocks/MockSystemInfo.swift
@@ -78,7 +78,7 @@ class MockSystemInfo: SystemInfo {
 
 extension MockSystemInfo: @unchecked Sendable {}
 
-extension RevenueCat.OperatingSystemVersion: Swift.Comparable {
+extension OperatingSystemVersion: Swift.Comparable {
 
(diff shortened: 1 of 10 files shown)


Tests/UnitTests/Mocks/MockSystemInfo.swift
```diff
@@ -78,7 +78,7 @@ class MockSystemInfo: SystemInfo {
 
 extension MockSystemInfo: @unchecked Sendable {}
 
-extension RevenueCat.OperatingSystemVersion: Swift.Comparable {
```
ajpallares, line 81: I remember needing to do this change but I'm not sure if it'd break other things... However, there isn't any `OperatingSystemVersion` defined in RevenueCat, so this should have never been allowed in the first place. I don't understand 😅

ajpallares (approved):
Thanks for fixing this! I guess some extra imports shouldn't harm

Tests/UnitTests/Mocks/MockSystemInfo.swift
facumenzella: I think this file should not be included in the project, but let's tackle that later

PR #5208, opened 2025-05-30
Introduce PaywallFontManagerType to handle custom fonts in the paywall editor

facumenzella:
### Motivation
We want to download custom fonts included in Offerings

### Description
- Add `PaywallFontManagerType ` with its implementation
- Add tests
- Fonts get downloaded during warm up, and also when a paywall detects a missing one (fire and forget fashion)

### DIFF
--- a/Tests/UnitTests/Paywalls/PaywallCacheWarmingTests.swift
+++ b/Tests/UnitTests/Paywalls/PaywallCacheWarmingTests.swift
@@ -175,6 +175,8 @@ final class PaywallCacheWarmingTests: TestCase {
         expect(self.imageFetcher.imageDownloadRequestCount.value) == 3
     }
 
+#if !os(macOS) && !os(tvOS) // For Paywalls V2
+
     func testTriggerFontDownload_DeduplicatesConcurrentDownloads() async throws {
         let font = DownloadableFont(
             name: "MockFont",
@@ -212,6 +214,8 @@ final class PaywallCacheWarmingTests: TestCase {
             expectedCount: 1
         )
     }
+
+#endif
 
     func testDownloadFont_PerformsExpectedActions() async throws {
         let mockSession = MockSession()


RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -6096,7 +6099,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "scripts/swiftlint.sh\n";
+			shellScript = "exit 0\nscripts/swiftlint.sh\n";
```
facumenzella, line 6102:
```suggestion
			shellScript = "scripts/swiftlint.sh\n";
```

ajpallares (commented):
Some early comments!

Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
```diff
@@ -32,13 +32,29 @@ public struct UIConfig: Codable, Equatable, Sendable {
...
+    }
+
+    public enum FontWeight: String, Codable, Sendable {
+        case thin, extraLight = "extra_light", light, regular, medium, semiBold = "semi_bold"
+        case bold, extraBold = "extra_bold", black
+    }
 
+    public enum FontStyle: String, Codable, Sendable {
+        case normal
+        case italic
```
ajpallares: I think we don't need these

Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
```diff
@@ -32,13 +32,29 @@ public struct UIConfig: Codable, Equatable, Sendable {
     }
 
     public struct FontsConfig: Codable, Equatable, Sendable {
-
         public var ios: FontInfo
+        public var web: FontInfo?
+        public var family: String?
```
ajpallares: Will this be needed? (right now it's not, but I'm not sure we'll need it in the paywalls later)

Sources/Paywalls/PaywallCacheWarming.swift
```diff
@@ -224,3 +265,44 @@ private extension PaywallData.Configuration.Images {
     }
 
 }
+
+#if !os(macOS) && !os(tvOS) // For Paywalls V2
+
+private extension UIConfig.AppConfig {
+    var allFonts: [(String, URL)] {
```
ajpallares: I think this name could be misleading since `iOS` fonts that don't have a `web?.url` are still valid fonts to consider, and wouldn't be included in this array. IMO, we should rename this to something like this:
```suggestion
    var allDownloadableFonts: [(String, URL)] {
```

Sources/Paywalls/PaywallCacheWarming.swift
```diff
@@ -224,3 +265,44 @@ private extension PaywallData.Configuration.Images {
...
+        case let .name(name):
+            if let url = URL(string: name) {
+                return url
+            } else {
+                Logger.error(PaywallsStrings.error_prefetching_font_invalid_url(name))
+                return nil
+            }
+        }
+    }
+}
```
ajpallares: This is a nit, but I think we could write this so that we avoid the temptation of doing `ios.url` and `web.iosName`. Like this:
```suggestion
private extension UIConfig.FontsConfig {

    var iOSName: String? {
        switch self.ios {
        case .googleFonts:
            return nil
        case let .name(name):
            return name
        }
    }

    var url: URL? {
        switch self.web {
        case .googleFonts, nil:
            return nil
        case let .name(name):
            if let url = URL(string: name) {
                return url
            } else {
                Logger.error(PaywallsStrings.error_prefetching_font_invalid_url(name))
                return nil
            }
        }
    }
}
```

Sources/Paywalls/PaywallFontFetcherType.swift
```diff
@@ -0,0 +1,87 @@
...
+        if families.contains(familyName) {
+            return
+        }
+
+        // already downloaded
+        if fileManager.fileExists(atPath: destination.path) {
+            return
+        }
+
+        try fileManager.copyItem(at: url, to: destination)
```
ajpallares: This `url` is the download URL, and not the local URL after the download happens. I think we need to use the `Data` returned from `try await session.data(from: url)`

Sources/Paywalls/PaywallFontFetcherType.swift
```diff
@@ -0,0 +1,87 @@
...
+    func downloadFont(from url: URL, familyName: String) async throws {
+        _ = try await session.data(from: url)
+
+        let destination = tempDirectory.appendingPathComponent(familyName)
+
+        // WIP: Use the name to verify if it exists
+        let families = CTFontManagerCopyAvailableFontFamilyNames() as? [String] ?? []
+        if families.contains(familyName) {
+            return
+        }
```
ajpallares: I think that we need to do the check of whether the font exists locally before downloading the font with `try await session.data(from: url)`

Sources/Paywalls/PaywallFontFetcherType.swift
```diff
@@ -0,0 +1,87 @@
...
+    private let fileManager: FileManaging
+    private let session: FontDownloadSession
+    private let registrar: FontRegistrar
+    private let tempDirectory: URL
+
+    init(
+        fileManager: FileManaging = FileManager.default,
+        session: FontDownloadSession = URLSession.shared,
+        registrar: FontRegistrar = SystemFontRegistry(),
+        tempDirectory: URL = FileManager.default.temporaryDirectory
```
ajpallares: I'm not sure we should use the temporaryDirectory. Since we're registering the fonts every time the app gets launched (by using the `CTFontManagerScope.process`, which I think it's the correct way), we will presumably need to keep registering the required fonts every time we launch the app / configure the SDK.

Now, I'm not sure about the best approach for this. I see two options:
1. We implement the local cache logic ourselves: by saving the downloaded font into a specific local path we control (based on the download URL of the font)
2. We always trigger the request and we let the `NSURLRequestReturnCacheDataElseLoad` do its job. This would seem simpler, but we would need to deal with the fact that we're receiving `Data` and need to store it into a `URL` we control to pass it to `CTFontManagerRegisterFontsForURL`.

Since the `CTFontManagerRegisterFontsForURL` says: 
> Once fonts have been registered from a file, it shouldn't be moved or renamed.

I think we should be the ones managing the stored URLs for the downloaded fonts, solving the two problems at once (local cache + having a URL we control for font registration)

Sources/Paywalls/PaywallFontFetcherType.swift
```diff
@@ -0,0 +1,87 @@
...
+    ) {
+        self.fileManager = fileManager
+        self.session = session
+        self.registrar = registrar
+        self.tempDirectory = tempDirectory
+    }
+
+    @available(iOS 15.0, macOS 12.0, watchOS 8.0, tvOS 15.0, *)
+    func downloadFont(from url: URL, familyName: String) async throws {
+        _ = try await session.data(from: url)
```
ajpallares: TODO: We need to add the `hash` verification around here

Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
ajpallares: Also, we need to add the `hash: String?` property, which will be used to verify the downloaded font file

ajpallares (commented):
Some more comments!

Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
facumenzella: I kept it just to test which one will be the one used to check if it's already installed, but probably we don't need it. 🤔

facumenzella: Now that we have the cache, we dont need this

Tests/UnitTests/Paywalls/PaywallCacheWarmingTests.swift
```diff
@@ -175,6 +175,138 @@ class PaywallCacheWarmingTests: TestCase {
         expect(self.imageFetcher.imageDownloadRequestCount.value) == 3
     }
 
+    func testTriggerFontDownload_DeduplicatesConcurrentDownloads() async throws {
```
facumenzella, line 180: Added this @ajpallares

joshdholtz:
@RCGitBot please test

joshdholtz:
@RCGitBot please test

ajpallares:
@RCGitBot please test

joshdholtz:
@RCGitBot please test

ajpallares:
@RCGitBot please test

PR #5209, opened 2025-05-30
[Paywalls] Use tab id instead of tab index to select tab

MarkVillacampa:
### Checklist

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Tabs/TabControlToggleComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Tabs/TabControlToggleComponentView.swift
@@ -59,7 +59,8 @@ struct TabControlToggleComponentView: View {
             )
             .labelsHidden()
         .onChangeOf(self.isOn) { newValue in
-            self.tabControlContext.selectedIndex = newValue ? 1 : 0
+            self.tabControlContext.selectedTabId =
+                newValue ? self.tabControlContext.tabIds[1] : self.tabControlContext.tabIds[0]
         }
     }
 
--- a/RevenueCatUI/Templates/V2/Components/Tabs/TabsComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Tabs/TabsComponentView.swift
@@ -85,7 +85,7 @@ struct LoadedTabsComponentView: View {
 
     var activeTabViewModel: TabViewModel? {
         return self.viewModel.tabViewModels[self.tabControlContext.selectedTabId] ??
-            self.viewModel.tabViewModels.values.first { $0.defaultSelectedPackage != nil }
+            self.viewModel.tabViewModels.values.first
     }
 
     init(viewModel: TabsComponentViewModel,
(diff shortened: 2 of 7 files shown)


RevenueCatUI/Templates/V2/Components/Tabs/TabControlToggleComponentView.swift
```diff
@@ -59,7 +59,8 @@ struct TabControlToggleComponentView: View {
             )
             .labelsHidden()
         .onChangeOf(self.isOn) { newValue in
-            self.tabControlContext.selectedIndex = newValue ? 1 : 0
+            self.tabControlContext.selectedTabId =
+                newValue ? self.tabControlContext.tabIds[1] : self.tabControlContext.tabIds[0]
```
MarkVillacampa, line 63: I think the TabControlToggleComponent is not actually implemented yet so I just opted for the straightforward fix here

RevenueCatUI/Templates/V2/Components/Tabs/TabsComponentView.swift
```diff
@@ -22,15 +22,16 @@ import SwiftUI
...
-    let tabControlStackViewModels: [StackComponentViewModel]
+    let tabIds: [String]
 
     init(controlStackViewModel: StackComponentViewModel,
-         tabControlStackViewModels: [StackComponentViewModel]) {
+         tabIds: [String]) {
         self.controlStackViewModel = controlStackViewModel
-        self.tabControlStackViewModels = tabControlStackViewModels
+        self.tabIds = tabIds
+        self._selectedTabId = .init(initialValue: tabIds.first ?? "")
```
MarkVillacampa, line 34: Removed `tabControlStackViewModels` as it wasn't being used anywhere.

`tabIds` are stored here for two reasons: to initialise `selectedTabId` with the first tab, and to be able to toggle between first and second in a future toggle component.

RevenueCatUI/Templates/V2/Components/Tabs/TabsComponentView.swift
```diff
@@ -80,10 +81,11 @@ struct LoadedTabsComponentView: View {
...
 
     @State
-    private var tierPackageContexts: [PackageContext]
+    private var tierPackageContexts: [String: PackageContext]
 
-    var activeTabViewModel: TabViewModel {
-        return self.viewModel.tabViewModels[self.tabControlContext.selectedIndex]
+    var activeTabViewModel: TabViewModel? {
+        return self.viewModel.tabViewModels[self.tabControlContext.selectedTabId] ??
+            self.viewModel.tabViewModels.values.first
```
MarkVillacampa, line 88: Make `activeTabViewModel` optional so we never actually crash, and instead we just don't render the `LoadedTabComponentView` below if the model cannot be found.

RevenueCatUI/Templates/V2/Components/Tabs/TabControlToggleComponentView.swift
fire-at-will: Do we know for sure (from backend validation, etc.) that `self.tabControlContext.tabIds` will always have 2 or more elements? Otherwise, `self.tabControlContext.tabIds[1]` could cause a crash if there is only one tab present, and `self.tabControlContext.tabIds[0]` could crash if there are somehow no tabs

MarkVillacampa:
@fire-at-will can't reply directly in your comment for some reason, but the answer is: TabControlToggleComponent is not being used right now so I just did the minimal change there.

PR #5210, opened 2025-05-30
[Paywalls] Render top level tabs component stack properties

MarkVillacampa:
This is a PoC implementation to see if it would be possible to return a stack component embedded in the top-level tabs component, and render each tab inside it. Right now we're creating a stack component on the fly, and wrapping each tab stack inside a copy of it.

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Stack/StackComponentViewModel.swift
+++ b/RevenueCatUI/Templates/V2/Components/Stack/StackComponentViewModel.swift
@@ -34,8 +34,7 @@ class StackComponentViewModel {
         viewModels: [PaywallComponentViewModel],
         badgeViewModels: [PaywallComponentViewModel],
         shouldApplySafeAreaInset: Bool = false,
-        uiConfigProvider: UIConfigProvider,
-        localizationProvider: LocalizationProvider
+        uiConfigProvider: UIConfigProvider
     ) throws {
         self.component = component
         self.viewModels = viewModels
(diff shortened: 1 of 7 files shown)


RevenueCatUI/Templates/V2/Components/Stack/StackComponentViewModel.swift
```diff
@@ -45,6 +44,16 @@ class StackComponentViewModel {
...
 
+    func copy(withViewModels newViewModels: [PaywallComponentViewModel]) throws -> StackComponentViewModel {
+        return try StackComponentViewModel(
+            component: self.component,
+            viewModels: newViewModels,
+            badgeViewModels: self.badgeViewModels,
+            shouldApplySafeAreaInset: self.shouldApplySafeAreaInset,
+            uiConfigProvider: self.uiConfigProvider
+        )
+    }
```
MarkVillacampa, line 55: This is the key new method that allows to clone a stack view model changing its child view models.

RevenueCatUI/Templates/V2/ViewModelHelpers/ViewModelFactory.swift
```diff
@@ -238,6 +238,52 @@ struct ViewModelFactory {
...
+            )
+
+            let tabsStackViewModel = try toStackViewModel(
+                component: tabsStackComponent,
+                packageValidator: PackageValidator(),
+                firstImageInfo: firstImageInfo,
+                localizationProvider: localizationProvider,
+                uiConfigProvider: uiConfigProvider,
+                offering: offering
+            )
```
MarkVillacampa, line 285: Ideally we dont need to build a stack component on the fly because the backend returns it as a property of the TabsComponent.

RevenueCatUI/Templates/V2/ViewModelHelpers/ViewModelFactory.swift
```diff
@@ -257,7 +303,7 @@ struct ViewModelFactory {
 
                 return try .init(
                     tab: tab,
-                    stackViewModel: stackViewModel,
+                    stackViewModel: tabsStackViewModel.copy(withViewModels: [.stack(stackViewModel)]),
```
MarkVillacampa, line 306: And we just wrap each tab stack in a copy of the top-level stack.

PR #5212, opened 2025-06-02
Add @_spi(Internal) to ClockTests

facumenzella:
### Motivation
While trying to fix tests for the workspace, I noticed this one does not compile

### Description
The protocol is using `_spi`, so we need to import it using it as well to access it

### DIFF
--- a/Tests/UnitTests/Misc/ClockTests.swift
+++ b/Tests/UnitTests/Misc/ClockTests.swift
@@ -12,7 +12,7 @@
 //  Created by Nacho Soto on 8/16/23.
 
 import Nimble
-@testable import RevenueCat
+@_spi(Internal) @testable import RevenueCat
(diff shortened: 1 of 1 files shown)


facumenzella:
@RCGitBot please test

ajpallares (approved):
I guess these tests weren't being executed at all before. I think we can add them, but the statements of the type
```
expect(...).to(beCloseTo(...))
```
scare me a bit. Seems like the recipe for flakyness

facumenzella:
@ajpallares looks like `defaultDelta` is used to check "beCloseTo", which should be fine I guess? I'll be paying attention, if it starts to fail we can always tweak it.

```
public func defaultDelta<F: FloatingPoint>() -> F { 1/10000 /* 0.0001 */ }
```

PR #5213, opened 2025-06-02
Use effectiveVersion.debugDescription in StoreKitVersionTests

facumenzella:
### Motivation
While trying to fix tests for the workspace, I noticed this is not compiling.

### Description
- Use ` effectiveVersion.debugDescription` instead of `versionString` that does not exist.
- Add test to `UnitTests` testplan


facumenzella:
@RCGitBot please test

ajpallares (approved):
Thanks for adding these tests back!

PR #5214, opened 2025-06-02
Add hash validation of downloaded font

ajpallares:
Implements downloaded font hash validation.

### Considerations
* Using md5 (this can change potentially before merging)
* Changed to use `cachesDirectory()` instead of `applicationSupportDirectory()`, since I believe it's more appropriate (the fonts cache is reproducible data that can be purged by the system if needed).
* Implemented `withTaskGroup` in `warmUpPaywallFontsCache()` for concurrent font installation.
* Renamed `FontsFetcher` to `FontsManager`, since it contains more logic besides fetching the fonts.
* Renamed `DefaultPaywallFontsFetcher.UnknownError` to be `DefaultPaywallFontsManager.FontsManagerError` to also contain more detailed information about what failed.


facumenzella (approved):
🚀

PR #5220, opened 2025-06-03
Trigger download of font that failed to load in a paywall

ajpallares:
### Motivation
When a paywall fails to load a custom font, the font should be tried to be downloaded when applicable

### Description
When a paywall fails to load, `Purchases` gets notified and triggers the font download

### DIFF
--- a/RevenueCatUI/Templates/V2/ViewHelpers/DisplayableColor.swift
+++ b/RevenueCatUI/Templates/V2/ViewHelpers/DisplayableColor.swift
@@ -17,6 +17,7 @@ import SwiftUI
 
 #if !os(macOS) && !os(tvOS) // For Paywalls V2
 
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 extension PaywallComponent.Background {
 
     func asDisplayable(uiConfigProvider: UIConfigProvider) -> BackgroundStyle {
@@ -61,6 +62,7 @@ enum DisplayableColorInfo: Codable, Sendable, Hashable {
 
 }
 
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 extension DisplayableColorScheme {
 
     static func from(colorScheme: PaywallComponent.ColorScheme,
@@ -73,6 +75,7 @@ extension DisplayableColorScheme {
 
 }
 
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 extension PaywallComponent.ColorScheme {
 
     func asDisplayable(uiConfigProvider: UIConfigProvider) -> DisplayableColorScheme {
@@ -86,6 +89,7 @@ extension PaywallComponent.ColorScheme {
 
 }
 
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 extension PaywallComponent.ColorInfo {
 
     func asDisplayable(forLight: Bool, uiConfigProvider: UIConfigProvider) throws -> DisplayableColorInfo {
--- a/RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
+++ b/RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
@@ -18,7 +18,7 @@ import SwiftUI
 #if !os(macOS) && !os(tvOS) // For Paywalls V2
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
-final class UIConfigProvider: Sendable {
+struct UIConfigProvider {
     private let uiConfig: UIConfig
     private let purchases: PaywallPurchasesType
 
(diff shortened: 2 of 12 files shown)


RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
```diff
@@ -13,15 +13,18 @@
...
 import Foundation
 import RevenueCat
+import SwiftUI
 
 #if !os(macOS) && !os(tvOS) // For Paywalls V2
 
-struct UIConfigProvider {
-
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+final class UIConfigProvider: Sendable {
```
facumenzella: I think we can avoid going from `struct` to `final class` in this PR right?

Sources/Paywalls/PaywallCacheWarming.swift
```diff
@@ -23,6 +23,9 @@ protocol PaywallCacheWarmingType: Sendable {
 
     @available(iOS 15.0, macOS 12.0, watchOS 8.0, tvOS 15.0, *)
     func warmUpPaywallFontsCache(offerings: Offerings) async
+
+    @available(iOS 15.0, macOS 12.0, watchOS 8.0, tvOS 15.0, *)
+    func triggerFontDownloadIfNeeded(fontsConfig: UIConfig.FontsConfig) async
```
facumenzella, line 38: I wonder if we can remove the `async` here? This should be a `fire and forget` operation, so there's no need other wait

RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
ajpallares: Yes, I think we can. But since we're holding a reference to `PaywallPurchasesType` now, I thought that using a class would be less error prone for the future. But I don't have a very strong opinion for this and you're right, so I'll change it back!
Done in https://github.com/RevenueCat/purchases-ios/pull/5220/commits/ebe07c921a941d9e84583e51b0db7ac8e85c4d21

Sources/Paywalls/PaywallCacheWarming.swift
ajpallares: > I wonder if we can remove the async here?

Not really, because `PaywallCacheWarming` is an `actor`, so its methods are `async` by default. Unless we made them `nonisolated`. But I think, in this case, having it be `async` makes sense. This way, the caller can potentially `await` if needed. If not, the caller can still implement the "fire and forget" strategy, as done in `Purchases.swift` with
```swift
self.operationDispatcher.dispatchOnWorkerThread {
    await self.paywallCache?.triggerFontDownloadIfNeeded(fontsConfig: fontConfig)
}
```

facumenzella (approved):
Tests are failing, but looks good to me

RevenueCatUI/Templates/V2/ViewHelpers/DisplayableColor.swift
```diff
@@ -17,6 +17,7 @@ import SwiftUI
 
 #if !os(macOS) && !os(tvOS) // For Paywalls V2
 
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
```
facumenzella, line 20: Could this break anything? Technically its becoming more restrictive right? 🤔

RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
```diff
@@ -32,32 +35,49 @@ struct UIConfigProvider {
...
+            return genericFont.makeFont(fontSize: fontSize)
         }
 
-        return localizations
-    }
+        guard let customFont = UIFont(name: fontName, size: fontSize) else {
+            Logger.warning("Custom font '\(fontName)' could not be loaded. Falling back to system font.")
+            self.purchases.failedToLoadFontWithConfig(fontsConfig)
+            return nil
+        }
```
facumenzella, line 76: Should we restrict font resolution to the main thread? Using`UIFont` could cause issues, right?

Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
```diff
@@ -96,13 +99,16 @@ public struct UIConfig: Codable, Equatable, Sendable {
 
     public struct WebFontInfo: Codable, Sendable, Hashable {
         public var value: String
-        public var hash: String
-        public var type: String
 
-        public init(value: String, hash: String, type: String) {
+        /// MD5 hash of the font file.
+        ///
+        /// Should never be `nil`, but it is optional to prevent potential decoding errors if, for some reason,
```
facumenzella, line 105: 👍

facumenzella (approved):

RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
ajpallares: You're right it should only be called from the `@MainActor`, which is already happening before but was not being enforced before by any `@MainActor` mark. I can add the `@MainActor` mark, but that would cascade into some more `@MainActor`s being required up the caller tree

Done in https://github.com/RevenueCat/purchases-ios/pull/5220/commits/ff5ed0aed601fb69cfe1f3b4055fc60d131ba468

Issue #5219, opened 2025-06-03
Customer Center: contact customer support button crashes the app


salling:
- [x] I have updated Purchases SDK to the latest version
- [x] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [ ] I have searched the [Community](https://community.revenuecat.com)
- [ ] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

1. Environment
   1. Platform: iOS
   2. SDK version: 5.25.3
   3. StoreKit version:
      - [ ] StoreKit 1 (default on versions <5.0.0. Can be enabled in versions >=5.0.0 with `.with(storeKitVersion: .storeKit1)`)
      - [x] StoreKit 2 (default on versions >=5.0.0)
   4. OS version: iOS 17 -> iOS 18
   5. Xcode version: Xcode 16.4
   6. Device and/or simulator:
      - [x] Device
      - [ ] Simulator
   7. Environment:
      - [x] Sandbox
      - [x] TestFlight
      - [x] Production
   8. How widespread is the issue. Percentage of devices affected. 100%
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue. **Complete logs with `Purchases.logLevel = .verbose` will help us debug this issue.**

```
*** Terminating app due to uncaught exception 'NSInvalidArgumentException', reason: 'The specified URL has an unsupported scheme. Only HTTP and HTTPS URLs are supported.'
*** First throw call stack:
[...]

PR #5221, opened 2025-06-03
Use env openURL instead of SFSafariVC to open mailto

facumenzella:
### Motivation
https://github.com/RevenueCat/purchases-ios/issues/5219

### Description
Use environment `openURL` instead of SFSafariVC to open the mailto URL

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/SubscriptionDetailView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/SubscriptionDetailView.swift
@@ -34,6 +34,9 @@ struct SubscriptionDetailView: View {
     @Environment(\.navigationOptions)
     var navigationOptions
 
+    @Environment(\.openURL)
+    var openURL
+
     @Environment(\.supportInformation)
     private var support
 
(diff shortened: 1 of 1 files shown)


RevenueCatUI/CustomerCenter/Views/SubscriptionDetailView.swift
```diff
@@ -34,6 +34,9 @@ struct SubscriptionDetailView: View {
     @Environment(\.navigationOptions)
     var navigationOptions
 
+    @Environment(\.openURL)
+    var openURL
```
vegaro, line 38: I think this will never happen but could this cause issues in app extensions? I believe it will never try to open the url in that case since canOpenURL could add return false, but we did add an openURL if not app extension function and I wanted to ask, just in case.

https://github.com/RevenueCat/purchases-ios/pull/4531/files

vegaro (approved):

RevenueCatUI/CustomerCenter/Views/SubscriptionDetailView.swift
facumenzella: I think we're good. This has been here for a while

Issue #5223, opened 2025-06-03
Restore Purchases: on 2nd attempt -> crashes


salling:
- [x] I have updated Purchases SDK to the latest version
- [ ] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [x] I have searched the [Community](https://community.revenuecat.com)
- [ ] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [x] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
A clear and concise description of what the bug is. The more detail you can provide the faster our team will be able to triage and resolve the issue. **Do not remove any of the steps from the template below. If a step is not applicable to your issue, please leave that step empty.**

1. Environment
   1. Platform: iOS
   2. SDK version: 5.25.3
   3. StoreKit version:
      - [ ] StoreKit 1 (default on versions <5.0.0. Can be enabled in versions >=5.0.0 with `.with(storeKitVersion: .storeKit1)`)
      - [x] StoreKit 2 (default on versions >=5.0.0)
   4. OS version: iOS 17.7.6 (possibly some related wonk also in iOS 18, but crash and console output is exclusive to iOS 17)
   5. Xcode version: 16.4
   6. Device and/or simulator:
      - [X] Device
      - [X] Simulator
   7. Environment:
      - [X] Sandbox
      - [X] TestFlight
      - [X] Production
   8. How widespread is the issue. Percentage of devices affected: 100% of iOS 17.7.6
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue. **Complete logs with `Purchases.logLevel = .verbose` will help us debug this issue.**

```
Publishing changes from within view updates is not allowed, this will cause undefined behavior.
Publishing changes from within view updates is not allowed, this will cause undefined behavior.
```

3. Steps to reproduce, with a description of expected vs. actual behavior

- Unentitled customer
- Open Customer Center
- Tap Restore Purchases
- Dismiss the alert that communicates failure to restore any purchases
- Tap Restore Purchases (again)
- Note Console log output re: `Publishing changes from within view updates is not allowed, this will cause undefined behavior` followed by crash
 
4. **Other information** (e.g. stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow, etc.)

```
[...]

PR #5224, opened 2025-06-03
Reset alertType state for RestorePurchasesAlertViewModel before dismissing

facumenzella:
### Motivation
https://github.com/RevenueCat/purchases-ios/issues/5223

### Description
As the alert is showed using an overlay, we need to make sure the state is reset before showing it again.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
+++ b/RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
@@ -171,6 +171,7 @@ struct RestorePurchasesAlert: View {
     private func dismissAlert() {
         self.customerCenterViewModel.onDismissRestorePurchasesAlert()
         self.isPresented = false
+        self.viewModel.alertType = .loading
     }
 }
 


RevenueCatUI/CustomerCenter/Views/RestorePurchasesAlert.swift
```diff
@@ -171,6 +171,7 @@ struct RestorePurchasesAlert: View {
     private func dismissAlert() {
         self.customerCenterViewModel.onDismissRestorePurchasesAlert()
         self.isPresented = false
+        self.viewModel.alertType = .loading
```
vegaro, line 174: how does this fix the "Publishing changes from within view updates is not allowed, this will cause undefined behavior."?

facumenzella: When showing the alert for the second time, the initial state is not `loading`. The `.task` is getting re-triggered every time `isPresented` changes, so after the first time it will start in a dirty state (not loading)

This wasn't happening before we refactored this to be an overlay (a while ago). Once we reset it, there are no changes triggered within view updates.

Note that this does not happen on every iOS, but as @salling pointed out, the loading is lost on iOS18 due to this.

vegaro (approved):
I think it makes sense, but yes, it would be a good idea to refactor this in the future. I can see something like this biting us again at some point

PR #5226, opened 2025-06-03
Fix StoreKitVersionTests on iOS 15

fire-at-will:
### Description
We found in https://github.com/RevenueCat/purchases-ios/pull/5213 that `StoreKitVersionTests` weren't being run. They're now failing in main for iOS 15 only. They're failing because we treat StoreKit 2 as being unavailable on iOS 15. 

I believe that what happened is that the `StoreKitVersionTests` file was excluded from the `UnitTests` test plan before we made StoreKit 2 exclusive to iOS 16+, we didn't update these tests, and didn't notice that they were broken because the tests weren't being run.

This PR updates the tests in `StoreKitVersionTests` to reflect that we treat StoreKit 1 as unavailable on iOS 15.

### DIFF
--- a/Tests/StoreKitUnitTests/TestHelpers/AvailabilityChecks.swift
+++ b/Tests/StoreKitUnitTests/TestHelpers/AvailabilityChecks.swift
@@ -62,6 +62,13 @@ enum AvailabilityChecks {
         }
     }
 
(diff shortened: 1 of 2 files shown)


fire-at-will:
@RCGitBot please test

PR #5234, opened 2025-06-04
Add back ReceiptStringsTests to UnitTests

facumenzella:
### Motivation
Part of organizing the test targets / test plans, I'm adding back this test that looks forgotten in the ether

### Description
- Re add test
- Delete one test that looks outdated


facumenzella:
@RCGitBot please test

Tests/UnitTests/Purchasing/ReceiptStringsTests.swift
```diff
@@ -212,16 +212,6 @@ final class ReceiptStringsTests: TestCase {
...
 
-    func testReceiptRetryingMechanismNotAvailable() {
-        let subject = ReceiptStrings
-            .receipt_retrying_mechanism_not_available
-        let expectedDescription = "Receipt retrying mechanism is not available in iOS 12. " +
-        "Will only attempt to fetch once."
-
-        expect(subject.category).to(equal("receipt"))
-        expect(subject.description).to(equal(expectedDescription))
-    }
```
facumenzella, line 223: Looks like this was a thing in iOS 12, and we've dropped support for that already. See b3875a9

ajpallares: Yep. In fact it seems it was removed v5.0.0 (https://github.com/RevenueCat/purchases-ios/commit/2bcd0ce3287c391a292825228666d3147074b16c#diff-b90ac6f010c07896e177bbd44986b42d86117183a4cc0af00417ab0a00d8c361)

ajpallares (commented):
I really don't know if these tests are useful. Basically they check the `description` and `category` of the log messages... 
IMO, we should remove them.

In fact, the test `testRetryingReceiptFetchAfterDuration` should fail right now, as there's an extra space in the `expectedDescription` string (this proves how redundant they are)

WDYT?

Tests/UnitTests/Purchasing/ReceiptStringsTests.swift
facumenzella: Test strikes back

facumenzella:
I am open to delete them. Let's bring this up

facumenzella:
@ajpallares I think this might just be ignored by every test plan, that's why its not failing 🤔

ajpallares:
So, do we delete them then?

facumenzella:
Closing this one because it was tackled in the tuist PR

PR #5237, opened 2025-06-04
New schema for `UIConfig.FontInfo`

ajpallares:
Support new (backwards-compatible) JSON Format for `UIConfig.FontInfo`. Example:
```json
"font_id": {
    "android": {
        "family": "Good Sans",
        "hash": "1234567890qwertyuiop",
        "style": "normal",
        "type": "name",
        "url": "https://fonts.pawwalls.com/font_file.otf",
        "value": "GoodSans-Light",
        "weight": "300"
    },
    "ios": {
        "family": "Good Sans",
        "hash": "1234567890qwertyuiop",
        "style": "normal",
        "type": "name",
        "url": "https://fonts.pawwalls.com/font_file.otf",
        "value": "GoodSans-Light",
        "weight": "300"
    },
    "web": {
        "family": "Good Sans",
        "hash": "1234567890qwertyuiop",
        "style": "normal",
        "type": "name",
        "url": "https://fonts.pawwalls.com/font_file.otf",
        "weight": "300"
    }
},
```

### DIFF
--- a/Sources/Logging/Strings/PaywallsStrings.swift
+++ b/Sources/Logging/Strings/PaywallsStrings.swift
@@ -27,7 +27,6 @@ enum PaywallsStrings {
     case error_creating_fonts_directory(Error)
     case error_installing_font(URL, Error)
     case error_prefetching_font_invalid_url(name: String, invalidURLString: String)
-    case error_prefetching_font_missing_hash(name: String)
 
     case caching_presented_paywall
     case clearing_presented_paywall
@@ -90,10 +89,6 @@ extension PaywallsStrings: LogMessage {
 
         case let .error_prefetching_font_invalid_url(name, invalidURLString):
             return "Error installing font \(name). Malformed url: \(invalidURLString)"
-
-        case let .error_prefetching_font_missing_hash(name):
-            return "Font \(name) does not have a validation hash. Skipping download. " +
-            "Please try to re-upload the font in the RevenueCat dashboard."
 
         case .caching_presented_paywall:
             return "PurchasesOrchestrator: caching presented paywall"
--- a/Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
+++ b/Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
@@ -44,10 +44,10 @@ public struct UIConfig: Codable, Equatable, Sendable {
         @_spi(Internal) public let value: String
         let webFontInfo: WebFontInfo?
 
-        @_spi(Internal) public init(name: String) {
+        @_spi(Internal) public init(name: String, webFontInfo: WebFontInfo? = nil) {
             self.type = .name
             self.value = name
-            self.webFontInfo = nil
+            self.webFontInfo = webFontInfo
         }
 
         // swiftlint:disable:next nesting
@@ -89,6 +89,12 @@ public struct UIConfig: Codable, Equatable, Sendable {
         /// Should never be `nil`, but it is optional to prevent potential decoding errors if, for some reason,
         /// the hash is not provided from the server.
         internal let hash: String
+
+        @_spi(Internal) public init(url: String, hash: String) {
+            self.family = nil
+            self.url = url
+            self.hash = hash
+        }
     }
 
     public struct VariableConfig: Codable, Equatable, Sendable {
--- a/Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
+++ b/Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
@@ -50,6 +50,7 @@ public struct UIConfig: Codable, Equatable, Sendable {
             self.webFontInfo = nil
         }
 
+        // swiftlint:disable:next nesting
         enum CodingKeys: String, CodingKey {
             case type
             case value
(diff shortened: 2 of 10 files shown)


Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
```diff
@@ -32,84 +32,62 @@ public struct UIConfig: Codable, Equatable, Sendable {
...
         ///
         /// Should never be `nil`, but it is optional to prevent potential decoding errors if, for some reason,
         /// the hash is not provided from the server.
-        public var hash: String?
-
-        public init(value: String, hash: String?) {
-            self.value = value
-            self.hash = hash
-        }
+        internal let hash: String
```
ajpallares: These are intentionally `internal`, as RevenueCatUI does not need to be aware of these

ajpallares:
### Important note
Some previously-public APIs have now be marked as `@_spi(Internal) public`. I.e., they are not public anymore (e.g. `UIConfig.FontInfo`). They were never meant to be `public` but they were because they are declared in RevenueCat and they need to be visible by RevenueCatUI. This is what `@_spi` solves.

In fact, all the `UIConfig` type and subtypes could be made `@_spi public`, but baby steps

Sources/Paywalls/PaywallCacheWarming.swift
```diff
@@ -333,6 +332,7 @@ private extension PaywallData.Configuration.Images {
     }
 }
 
+/// Business logic object to easily manage the download of fonts.
```
facumenzella, line 334:
```suggestion
/// MARK: - Fonts
```

Sources/Logging/Strings/PaywallsStrings.swift
```diff
@@ -27,7 +27,6 @@ enum PaywallsStrings {
     case error_creating_fonts_directory(Error)
     case error_installing_font(URL, Error)
     case error_prefetching_font_invalid_url(name: String, invalidURLString: String)
-    case error_prefetching_font_missing_hash(name: String)
```
facumenzella, line 30: Is this one not used? 🤔

Sources/Paywalls/PaywallCacheWarming.swift
```diff
@@ -343,6 +343,11 @@ struct DownloadableFont: Hashable, Sendable {
 
     let url: URL
     let hash: String
+
+    func hash(into hasher: inout Hasher) {
+        // Purposedly only hash the url, as two fonts with the same remote URL should only downloaded once
+        hasher.combine(url)
+    }
```
facumenzella: Even if they hash to the same value, two `DownloadableFont` will not be equal, right? Isn't this "too tied" to ` ongoingFontDownloads: [DownloadableFont: Task<Void, Never>] = [:]` ? If we change the way we track these ongoing tasks, we might break this.

What do you think about changing `[DownloadableFont: Task<Void, Never>]` to `[URL: Task<Void, Never>]`, and remove the hash here?

ajpallares: Yep. You're totally right. I think it's much clearer as you propose. Thank you!

Sources/Paywalls/PaywallFontManagerType.swift
```diff
@@ -55,6 +55,7 @@ struct DefaultFontFileManager: FontsFileManaging {
         guard let url = fileManager.urls(for: .cachesDirectory, in: .userDomainMask).first else {
             throw CocoaError(.fileNoSuchFile)
         }
+        print("cache directory: \(url)")
```
facumenzella: 👀

facumenzella (approved):
My main concern is the hash thingy, and let's not forget to delete the `print` 💃

Sources/Paywalls/PaywallCacheWarming.swift
ajpallares: In addition, it was confusing having the `hash` property (md5 hash of the font file) and not using it in the `hash(into: )` method 😅 
Did the change in https://github.com/RevenueCat/purchases-ios/pull/5237/commits/4293cc0b94e65038770fc9f46439c7952e8d7884

Sources/Logging/Strings/PaywallsStrings.swift
ajpallares: Not anymore. It gets discarded at decoding time with https://github.com/RevenueCat/purchases-ios/blob/68cd62297a442f8b91805a74f64a7b51ab42812e/Sources/Networking/Responses/RevenueCatUI/UIConfig.swift#L60

Sources/Paywalls/PaywallCacheWarming.swift
ajpallares: Do you really think we should remove the documentation? It's an internal object, but I think it can be helpful for us? 🤔

Sources/Paywalls/PaywallFontManagerType.swift
ajpallares: 💨 https://github.com/RevenueCat/purchases-ios/pull/5237/commits/68cd62297a442f8b91805a74f64a7b51ab42812e

PR #5238, opened 2025-06-04
Adding deep link for testing in Paywalls Tester

joshdholtz:
### Motivation

Needed a way to visually test deep links

### Description

Shows an alert when `scheme://deeplinktest` is detected for the URL

https://github.com/user-attachments/assets/d20fb758-3c1e-4c00-a417-080e8e94e5f2

### DIFF
--- a/Tests/TestingApps/PaywallsTester/PaywallsTester/PaywallsPreview.swift
+++ b/Tests/TestingApps/PaywallsTester/PaywallsTester/PaywallsPreview.swift
@@ -88,9 +88,17 @@ struct PaywallsPreview: App {
 extension PaywallsPreview {
 
     func processURL(_ url: URL) {
-        // set to nil to trigger re-render if presenting same paywall with new data
-        paywallPreviewData = nil
-        paywallPreviewData = getPaywallDataFrom(incomingURL: url)
+        if isDeepLinkTest(url) {
+            showAlert(title: "Deep Link", message: url.absoluteString)
+        } else {
+            // set to nil to trigger re-render if presenting same paywall with new data
+            paywallPreviewData = nil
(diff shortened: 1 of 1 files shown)


Tests/TestingApps/PaywallsTester/PaywallsTester/PaywallsPreview.swift
```diff
@@ -88,9 +88,17 @@ struct PaywallsPreview: App {
 extension PaywallsPreview {
 
     func processURL(_ url: URL) {
-        // set to nil to trigger re-render if presenting same paywall with new data
-        paywallPreviewData = nil
-        paywallPreviewData = getPaywallDataFrom(incomingURL: url)
+        if isDeepLinkTest(url) {
+            showAlert(title: "Deep Link", message: url.absoluteString)
```
fire-at-will, line 92: Would it be a good idea if we gave the alert an action that dismisses the alert and then refreshes + displays the `paywallPreviewData`? That way we can test that the paywall can actually be displayed when loaded from a deep link.

Essentially calling this code in one of the alert's actions:
```swift
// set to nil to trigger re-render if presenting same paywall with new data
paywallPreviewData = nil
paywallPreviewData = getPaywallDataFrom(incomingURL: url)
```

joshdholtz: OHHHH, this wasn't actually for testing the paywall being displayed from a deep link.

This is for the button on a paywall having a deep link and us just testing if that deep link works 😇

fire-at-will: ooohhhh perfect then!!

PR #5240, opened 2025-06-05
Warm up caches in parallel

ajpallares:
### Motivation
Warming up the caches serially could stall other warm-ups if taking too long.

### Description
This PR makes the cache warm-ups in parallel


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -2106,9 +2106,13 @@ private extension Purchases {
...
-                    await cache.warmUpEligibilityCache(offerings: offerings)
-                    await cache.warmUpPaywallImagesCache(offerings: offerings)
+                await withTaskGroup(of: Void.self) { group in
+                    group.addTask {
+                        await cache.warmUpEligibilityCache(offerings: offerings)
+                    }
+                    group.addTask {
+                        await cache.warmUpPaywallImagesCache(offerings: offerings)
+                    }
                 }
```
facumenzella: I assume we'll add fonts here too, in parallel?

I wondering if it wouldn't be better to have one task per image, eligibility, or font, instead of parallelizing the method. I say this because inside fonts we're launching yet another task group

ajpallares: > I assume we'll add fonts here too, in parallel?

Exactly

> I wondering if it wouldn't be better to have one task per image, eligibility, or font, instead of parallelizing the method.

I think the system handles that automatically for us. 

> I say this because inside fonts we're launching yet another task group

In fact, I think the idea is to do the same with the images. We launch everything in parallel and let the system handle it efficiently. Take into account that we're using a `.background` thread in `dispatchOnWorkerThread`, i.e. the lowest `TaskPriority`, so it should be fine.

PR #5242, opened 2025-06-06
Remove dependency of Purchases.shared from UIConfigProvider

joshdholtz:
### Motivation

Fix failing cross platform validation screenshots

### Description

`UIConfigProvider` required that `Purchases.shared` be configured but some of the paywalls tests (like the screenshot taker) doesn't do that

Removed dependency of `Purchases.shared` from within Paywalls v2 and replaced it with a function that is passed into `PaywallsV2View`

### DIFF
--- a/RevenueCatUI/PaywallView.swift
+++ b/RevenueCatUI/PaywallView.swift
@@ -330,7 +330,12 @@ public struct PaywallView: View {
                         }
                         onRequestedDismissal()
                     },
-                    fallbackContent: .paywallV1View(dataForV1DefaultPaywall)
+                    fallbackContent: .paywallV1View(dataForV1DefaultPaywall),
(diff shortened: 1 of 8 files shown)


RevenueCatUI/PaywallView.swift
```diff
@@ -330,7 +330,12 @@ public struct PaywallView: View {
                         }
                         onRequestedDismissal()
                     },
-                    fallbackContent: .paywallV1View(dataForV1DefaultPaywall)
+                    fallbackContent: .paywallV1View(dataForV1DefaultPaywall),
+                    failedToLoadFont: { fontConfig in
+                        if Purchases.isConfigured {
```
ajpallares, line 335: Nice touch adding the `Purchases.isConfigured` check 👌

PR #5246, opened 2025-06-06
Remove submodules temporarily to fix SPM

tonidero:
### Description
We added a submodule to a private repo in order to perform validation of paywalls. However, this breaks SPM, which tries to clone all submodules, even if devs don't have access to it. This just removes the submodules temporarily, and makes jobs that use it fail

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1385,40 +1385,46 @@ jobs:
       xcode_version: "16.0"
 
     steps:
-      - checkout
-      - checkout-submodule:
-          path: upstream/paywall-preview-resources
-      - install-dependencies:
-          install_swiftlint: false
       - run:
-          name: Clone paywall-rendering-validation repository
-          command: git clone git@github.com:RevenueCat/paywall-rendering-validation.git ../paywall-rendering-validation
-      - revenuecat/setup-git-credentials
-      - run:
-          name: Set GitHub token environment variable
-          command: |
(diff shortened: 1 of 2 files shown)


tonidero:
@RCGitBot please test

ajpallares (approved):
👍

.circleci/config.yml
```diff
@@ -1385,40 +1385,46 @@ jobs:
...
+      #     name: Set GitHub token environment variable
+      #     command: |
+      #       echo "export GITHUB_TOKEN=$GH_TOKEN_PAYWALL_RENDERING_VALIDATION_PR_RW" >> $BASH_ENV
+      #       source $BASH_ENV
+      # - run:
+      #     name: Run tests
+      #     command: bundle exec fastlane record_and_push_paywall_template_screenshots target_repository_path:../paywall-rendering-validation
+      #     no_output_timeout: 15m
+      #     environment:
+      #       SCAN_DEVICE: iPhone 16
```
vegaro, line 1410: total nitpick but maybe instead of this (have a failing job), we can just comment the call in the workflow?

tonidero: I thought about it, but I wasn't sure what would happen since both of these jobs are the only ones within their respective workflow, and wasn't sure if we could have a workflow without any jobs. Didn't want to remove the workflows since that would also affect other repos and wanted to keep the changes contained here.

tonidero:
Failing tests not related to the changes in this PR so merging

PR #5250, opened 2025-06-06
Add Danger rule to avoid submodules

ajpallares:
### Motivation
Preventing adding git submodules in the future (which cause SPM not to be able to allow the dependency when the submodule is private)

### Description
Adds a danger rule that fails if it detects a submodule in the repository

### DIFF
--- a/Dangerfile
+++ b/Dangerfile
@@ -3,5 +3,5 @@ danger.import_dangerfile(github: 'RevenueCat/Dangerfile')
 # Check for submodules
 submodules = `git submodule status`.strip
 if !submodules.empty?
-  fail("This repository should not contain any submodules. Please remove them.")
+  fail("This repository should not contain any submodules. When using Swift Package Manager, developers will get a resolution error because SPM cannot access private submodules.")
 end
--- a/Dangerfile
+++ b/Dangerfile
@@ -1,1 +1,7 @@
 danger.import_dangerfile(github: 'RevenueCat/Dangerfile')
+
+# Check for submodules
+submodules = `git submodule status`.strip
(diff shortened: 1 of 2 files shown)


ajpallares:
Danger failed successfully for commit https://github.com/RevenueCat/purchases-ios/pull/5250/commits/0442ad8eea63218b6c8fce1b8330ab79525f7cf8 which adds a submodule

upstream/paywall-preview-resources
vegaro, line 1: 😰

ajpallares: hahaha I was testing it 😅

joshdholtz (approved):
:shipit: :shipit: :shipit: :shipit:

Dangerfile
```diff
@@ -1 +1,7 @@
 danger.import_dangerfile(github: 'RevenueCat/Dangerfile')
+
+# Check for submodules
+submodules = `git submodule status`.strip
+if !submodules.empty?
+  fail("This repository should not contain any submodules. Please remove them.")
```
vegaro: maybe explain why for the future devs that get this in 20 years

tonidero (commented):
Main concern is that this will also affect android, which has a submodule without issues...

Dangerfile
ajpallares: Good point 👍

ajpallares: Done in https://github.com/RevenueCat/purchases-ios/pull/5250/commits/b21b3ed95f89e4caf2dbb2a022ee962524c1bf96

PR #5252, opened 2025-06-09
Bring back paywall rendering validation (no submodules)

ajpallares:
Re-enable paywall rendering validation. 

Instead of depending on `paywall-preview-resources` as a submodule, there's a step that explicitly clones it.


fastlane/Fastfile
```diff
@@ -1152,6 +1150,23 @@ platform :ios do
     end
   end
 
+  desc "Clones or updates paywall-preview-resources repo"
+  lane :fetch_paywall_preview_resources do
+    Dir.chdir("../Tests") do
+      if !File.directory?(paywall_preview_resources_repo_name)
+        UI.important("Cloning paywall-preview-resources repo")
+        sh "git", "clone", "git@github.com:RevenueCat/#{paywall_preview_resources_repo_name}.git"
+      end
```
facumenzella: I wonder if we can do a shallow clone? The thing is that the commit is not necessarily HEAD right?

ajpallares: Yeah, this is a good point. But I don't know if we should be making many assumptions about this commit. It could potentially lead to future bugs. But on the other hand, doing a shallow clone on the resources repo seems like a proper thing to do. 

> the commit is not necessarily HEAD right?

That's correct, I think

fastlane/Fastfile
```diff
@@ -1152,6 +1150,23 @@ platform :ios do
     end
   end
 
+  desc "Clones or updates paywall-preview-resources repo"
+  lane :fetch_paywall_preview_resources do
```
facumenzella, line 1154: Is this something we also want for android?

ajpallares: Yeah, I think we will also implement it in Android

.gitignore
```diff
@@ -38,6 +38,12 @@ IntegrationTests/CarthageInstallation/Cartfile.resolved
 scan_derived_data/
 generated_docs/
 
+# Local clone of `paywall-preview-resources` repo to be available for tests.
+Tests/paywall-preview-resources
```
JayShortway, line 42: If we want to use these resources for SwiftUI previews, will we be able to access them from the `Tests` folder?

fastlane/Fastfile
JayShortway: Yea when this is confirmed working we should move it to our Fastlane plugin and get rid of the submodule in Android too.

JayShortway (approved):
Great job getting this fixed! Just a question.

.gitignore
ajpallares: For now, I think they would not be available because the contents of the folder are not referenced in the Xcode project. If we used the Swift Package directly, I think it would be possible. 
But I'm not 100% sure. I need to check this

JayShortway: Would it be easier to have the `paywall-preview-resources` folder be checked out somewhere else then, e.g. in `upstream/` like before?

ajpallares: 🤔  I don't see why that would make things easier. But I could probably be missing something

JayShortway: No in that case it's probably me who is missing something haha. In Android tests are in a `test` source set, and the dependency direction is such that the `test` source set depends on the `main` source set, but not the other way around. Previews would be in the `main` (or `debug`) source set, so they won't have access to any resources in the `test` source set. But it seems this does not apply to the `Tests` folder in iOS!

ajpallares: Yeah, in the iOS case, I think `Tests` is just a name for a folder. Could have been any other.

It would be different if it were using SPM, where folder names are sometimes important

PR #5256, opened 2025-06-09
Use Mint 0.17.5 for `run-test-ios-14` CI job

ajpallares:
### Motivation

CI for `run-test-ios-14` was failing when installing Mint
> error: 'mint-0.18.0': package 'mint-0.18.0' is using Swift tools version 5.9.0 but the installed version is 5.7.1
> Error: You are using macOS 12.
> We (and Apple) do not provide support for this old version.

### Description

Mint 0.17.5 is used in this CI job as newer versions of Mint don't support macOS 12

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -220,8 +220,7 @@ commands:
       - when:
           condition: << parameters.install_mint_0_17_5 >>
           steps:
-            - install-brew-dependency:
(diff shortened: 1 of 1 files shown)


ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

PR #5264, opened 2025-06-10
FIx image load in paywall rendering validation

ajpallares:
### Motivation

The transition animation that happens when loading paywall images was making paywall rendering validation tests to be flaky: sometimes the images were not appearing or they appeared with transparency (i.e. the transition was not finished at the time of taking the snapshot).

### Description

This PR removes the image load transition animation in paywalls when running in paywall rendering validation.


RevenueCatUI/Views/RemoteImage.swift
```diff
@@ -37,7 +37,16 @@ struct RemoteImage<Content: View>: View {
...
     }
 
-    private let transition: AnyTransition = .opacity.animation(Constants.defaultAnimation)
+    private var transition: AnyTransition {
+        #if DEBUG
+        if self.url.isFileURL {
+            // No transition for the load of the local image
+            // This is used for paywall screenshot validation
+            return .identity
+        }
```
JayShortway: Legacy paywalls may use local images for icons (there's a fixed icon set in the SDK), and the fallback paywall uses local images too. It's not the end of the world if those don't animate, but did you happen to check if it's noticeable? (My only concern is confused developers.)

JayShortway (approved):
Good to have this fixed! Just a question, but will approve to unblock.

RevenueCatUI/Views/RemoteImage.swift
ajpallares: Great point about legacy paywalls and fallback paywall. I need to check, but I think you're right. Although I'm not sure if it would really be noticeable as the fade animation would probably not be seen due to the paywall presentation animation.

In any case, I think I can make the condition more restrictive and also check if the code is being run in tests to be totally sure that this does not affect anyone in normal builds.

Thanks for the suggestion!

ajpallares: @JayShortway done in https://github.com/RevenueCat/purchases-ios/pull/5264/commits/44b5169d3abc69d4cd138d69d89530c1a61cc4d8
Please, lmk what you think! 😊

JayShortway: Oh this is very nice! Thanks for going the extra mile here 😄

PR #5266, opened 2025-06-11
Remove trust-github-key from installation tests

facumenzella:
### Motivation
We want to make sure that `installation` tests don't have access to private repositories

### Description
- Remove `trust-github-key`


facumenzella:
@RCGitBot please test

PR #5267, opened 2025-06-11
Enable markdown in paywalls in iOS 15 and watchOS 8

ajpallares:
### Motivation

We got a report of markdown in paywalls not being rendered correctly in iOS 15

### Description

This PR enables markdown in `TextComponentView` for iOS 15 and watchOS 8.

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
@@ -75,7 +75,7 @@ private struct NonLocalizedMarkdownText: View {
 
     var markdownText: AttributedString? {
         #if swift(>=5.7)
-        if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
+        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *) {
             return try? AttributedString(
                 markdown: self.text,
                 // We want to only process inline markdown, preserving line feeds in the original text.
--- a/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
@@ -75,15 +75,11 @@ private struct NonLocalizedMarkdownText: View {
 
     var markdownText: AttributedString? {
         #if swift(>=5.7)
-        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *) {
-            return try? AttributedString(
-                markdown: self.text,
-                // We want to only process inline markdown, preserving line feeds in the original text.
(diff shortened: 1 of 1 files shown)


ajpallares:
I wasn't sure if there was any special reason why it wasn't enabled, so I checked and found [this conversation](https://github.com/RevenueCat/purchases-ios/pull/4995#discussion_r2044353247). I think it was just a simple oversight in out part, so I think it's safe to enable it. I also tested it in iOS 15 and it works correctly.

RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
```diff
@@ -75,7 +75,7 @@ private struct NonLocalizedMarkdownText: View {
 
     var markdownText: AttributedString? {
         #if swift(>=5.7)
-        if #available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *) {
+        if #available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *) {
```
facumenzella: What about the macOS and tvOS changes? 👀

ajpallares: I changed the availability of macOS and tvOS to be consistent with the APIs used there, but actually, the `TextComponentView` is not available in macOS or tvOS
https://github.com/RevenueCat/purchases-ios/blob/84f0386d9dd59c955ae67a214fb2f02bed3a4f90/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift#L20

facumenzella (approved):
Looks good, but maybe just double-check with @joshdholtz in case we're missing something?

RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
MarkVillacampa: if the availability matches the one of the struct we can probably drop it altogether?

ajpallares:
Yeah, it would be good to have @joshdholtz's check here in case I'm missing something.

However, this is the situation on iOS 15

| Before this PR | After this PR |
| -------- | ------- |
| image | image |

So I think it's safe to merge so that it can be included in today's release.

Btw, I've also check that this builds and works properly on iOS 15 using Xcode 14. And the original PR https://github.com/RevenueCat/purchases-ios/pull/4995 purpose was to fix a compilation error on iOS 15 with Xcode 14, which this PR is not breaking.

RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
ajpallares: Yeah, realized the same and did it in https://github.com/RevenueCat/purchases-ios/pull/5267/commits/fdb46c4d2501c6a534c29e02d5ecb60cee435ca4
Thanks for pointing it out!

PR #5270, opened 2025-06-11
Add _spi(Internal) to Customer Center

facumenzella:
### Motivation
We want to add `spi` to currently `public` entities that shouldn't be public.

### Description
- Add missing `spi` to existing entities

### DIFF
--- a/Tests/UnitTests/Networking/Backend/BackendGetCustomerCenterConfigTests.swift
+++ b/Tests/UnitTests/Networking/Backend/BackendGetCustomerCenterConfigTests.swift
@@ -15,9 +15,9 @@ import Foundation
 import Nimble
 import XCTest
 
-@testable import RevenueCat
+@_spi(Internal) @testable import RevenueCat
 
-class BackendGetCustomerCenterConfigTests: BaseBackendTests {
+final class BackendGetCustomerCenterConfigTests: BaseBackendTests {
 
     override func createClient() -> MockHTTPClient {
         super.createClient(#file)


vegaro (approved):
Went through all the files and didnt' see anything to call out

RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigData+Mock.swift
```diff
@@ -135,7 +135,8 @@ extension CustomerCenterConfigData {
     }
 
     @available(iOS 14.0, *)
-    static let `default` = mock()
+    // swiftlint:disable missing_docs
+    @_spi(Internal) public static let `default` = mock()
```
ajpallares: Why is this one needed? It's part of RevenueCatUI, so I don't think it is 🤔

Tests/RevenueCatUITests/CustomerCenter/BaseManageSubscriptionViewModelTests.swift
```diff
@@ -17,7 +17,7 @@
 
 import Nimble
 @_spi(Internal) @testable import RevenueCat
-@testable import RevenueCatUI
+@_spi(Internal) @testable import RevenueCatUI
```
ajpallares, line 20: I think that RevenueCatUI shouldn't need to be imported as `@_spi(Internal)` in tests, as `@testable` should be enough

ajpallares (commented):
Love this! This will result in a much cleaner public API for developers! 🎉  
And thank you for taking over as the official `spi` guy 😄 

Before approving, I have some comments and a question: are all the `@_spi(Internal) import`s in RevenueCatUI actually needed?

RevenueCatUI/CustomerCenter/Data/CustomerCenterConfigData+Mock.swift
facumenzella: You're right!! 🙌 
The compiler was failing to find it, but looks like it was just a hiccup. Removing it

Tests/RevenueCatUITests/CustomerCenter/BaseManageSubscriptionViewModelTests.swift
facumenzella: Without this, `RevenueCatUITests` can't find `.mock()` and `.default`.
Both declared as extensions of the `spi` CustomerCenterConfigData 🤔

ajpallares: Oh, I understand now! You are right. I didn't know that the `_spi` was inherited for the `CustomerCenterConfigData` extensions in RevenueCatUI from RevenueCat. 
I thought that having the `@_spi(Internal)` in the `import RevenueCat` was enough. But it isn't 👍

ajpallares (approved):

PR #5275, opened 2025-06-12
Use `card_store_promotional` for RC Promos in card & history

facumenzella:
### Motivation
We're currently showing `Promotional` for RC promos and it's kind of confusing

### Description
- Use new key `via Support` 👍


Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -341,6 +342,8 @@ import Foundation
                     return "Promotional"
                 case .storeAmazon:
                     return "Amazon Store"
+                case .cardStorePromotional:
+                    return "via Support"
```
facumenzella:
```suggestion
                    return "Via Support"
```

vegaro (approved):

PR #5286, opened 2025-06-13
Remove CustomerCenterConfigDataAPI from API Tests

facumenzella:
### Motivation
I forgot to update those after https://github.com/RevenueCat/purchases-ios/pull/5270

### Description
- Remove test from ALL API tests


JayShortway (approved):
Makes sense!

PR #5288, opened 2025-06-13
Preferred locale override

ajpallares:
### Checklist

### Motivation
End goal: allowing developers to choose the locale for Paywalls

### Description
Prepares RevenueCat to support setting the preferred locale at configure time. A continuation PR will add new APIs to RevenueCatUI to allow setting the preferred locale.

### Notes
Don't get scared about the number of files changes in this PR. Most are small changes in tests (new parameter), but the actual implementation changes in this PR are only a few files


RevenueCatUI/Purchasing/Purchases+RevenueCatUI.swift
ajpallares, line 1: Note that this file is part of `RevenueCatUI`

ajpallares:
@RCGitBot please test

Tests/APITesters/AllAPITests/RevenueCatUISwiftAPITester/Purchases+RevenueCatUI_API.swift
```diff
@@ -0,0 +1,16 @@
...
+
+import RevenueCat
+import RevenueCatUI
+
+func checkPreferredUILocaleAPIs() {
+    Purchases.updatePreferredUILocale("de_DE")
+    let _: RevenueCat.Configuration.Builder = Configuration
+        .builder(withAPIKey: "")
+        .with(preferredUILocale: "de_DE")
+}
```
JayShortway: I'm confused by this haha, aren't these APIs only added by https://github.com/RevenueCat/purchases-ios/pull/5292?

Sources/Misc/Locale/PreferredLocalesProvider.swift
```diff
@@ -16,17 +16,22 @@ import Foundation
 /// A type that can determine the current list of preferred locales.
 protocol PreferredLocalesProviderType {
 
+    /// Developer-set preferred locale that takes precedence over the system preferred locales.
+    var preferredLocaleOverride: String? { get set }
+
     /// Returns a list of the user's preferred languages.
-    var preferredLanguages: [String] { get }
+    var preferredLocales: [String] { get }
```
JayShortway: I think having these 2 public fields might be a bit confusing. It's not immediately clear what the difference is. 

Do we need to have both separate or can `preferredLocales` contain the overridden one too? 

If so, what if the protocol has 
```swift
// Takes overridden locale into account
var preferredLocales: [String] { get }

// Updates preferredLocales to take the provided locale into account
func overrideLocale(locale: String?)
```

Sources/Misc/SystemInfo.swift
```diff
@@ -40,7 +40,7 @@ class SystemInfo {
     let responseVerificationMode: Signing.ResponseVerificationMode
     let dangerousSettings: DangerousSettings
     let clock: ClockType
-    let preferredLocalesProvider: PreferredLocalesProviderType
+    var preferredLocalesProvider: PreferredLocalesProviderType
```
JayShortway: Why was this needed?

JayShortway (commented):
Great stuff, thanks for doing this! Just a few questions and suggestions.

Tests/APITesters/AllAPITests/RevenueCatUISwiftAPITester/Purchases+RevenueCatUI_API.swift
ajpallares: ohh nice catch!
Will remove from here! (I artificially separated this into two PRs afterwards, so there are still some leftovers 😅)

ajpallares: Removed in https://github.com/RevenueCat/purchases-ios/pull/5288/commits/366475a46e1934edf9237da172f0cf91e3e840f1

Sources/Misc/SystemInfo.swift
ajpallares: It is needed but it's not very obvious!
Since `preferredLocalesProvider` only conforms to protocol `PreferredLocalesProviderType`, the compiler can't know if it's a class or a struct. Because setting the `preferredLocaleOverride` is a mutating operation, the compiler requires `preferredLocalesProvider` to be a `var` so that it can always be mutable (because a `let` struct cannot be mutated).

Sources/Misc/Locale/PreferredLocalesProvider.swift
ajpallares: That was my first approach but then I found out that I would need to reimplement/overcomplicate the implementation of `MockPreferredLocalesProvider` for tests.

But I just got a better idea right while answering to improve it so that we don't need the mock in tests and we can do what you propose for the implementation. 

I think I'm going to do this in a separate PR because this one is already big enough.

Thank you for the suggestion!

ajpallares: This is the PR https://github.com/RevenueCat/purchases-ios/pull/5299

Now, the only APIs of `PreferredLocalesProvider` are
```swift
var preferredLocales: [String]
func overridePreferredLocale(_ locale: String?)
```

Because of that change, I needed to adapt tests to avoid duplicating the `PreferredLocalesProvider` in the `MockPreferredLocalesProvider` object

JayShortway: This is great, thanks for doing this!

Sources/Misc/SystemInfo.swift
JayShortway: TIL! Thank you for explaining 😄

Issue #5285, opened 2025-06-12
RevenueCat iOS SDK 5.27.1 and later change paywall v2 configuration resulting in misaligned images


keremerkan:
- [X] I have updated Purchases SDK to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [X] I have searched the [Community](https://community.revenuecat.com)
- [X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [X] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
On iOS SDK 5.27.1 and later, v2 paywalls are shown with misaligned elements. See the images at the bottom for reference.

This bug only appears when the paywall is shown from a modally presented view controller. It is not apparent when the paywall is shown from the root view controller. It does not happen on SDK versions 5.26.0 and earlier. Also iPad doesn't seem to be affected as well.

1. Environment
   1. Platform: iOS
   2. SDK version: 5.27.1 and later
   3. StoreKit version:
      - [ ] StoreKit 1 (default on versions <5.0.0. Can be enabled in versions >=5.0.0 with `.with(storeKitVersion: .storeKit1)`)
      - [x] StoreKit 2 (default on versions >=5.0.0)
   4. OS version: 18.1 and later (including betas), perhaps also earlier
   5. Xcode version: 16.4
   6. Device and/or simulator:
      - [X] Device
      - [ ] Simulator
   7. Environment:
      - [X] Sandbox
      - [ ] TestFlight
      - [ ] Production
2. [Debug logs]

```
DEBUG: ℹ️ Configuring SDK using RevenueCat's UserDefaults suite.
DEBUG: 👤 Identifying App User ID
DEBUG: ℹ️ Debug logging enabled
DEBUG: ℹ️ SDK Version - 5.27.1
DEBUG: ℹ️ Bundle ID - xxx
DEBUG: ℹ️ System Version - Version 18.1 (Build 22B83)
DEBUG: ℹ️ Not using a simulator.
DEBUG: 👤 No initial App User ID
DEBUG: ℹ️ Purchases is configured with informational response verification
DEBUG: ℹ️ Observing StoreKit.Transaction.updates
DEBUG: ℹ️ Purchases is configured with StoreKit version 2
DEBUG: ℹ️ Observing StoreKit.PurchaseIntent.intents
[...]

PR #5291, opened 2025-06-14
[Paywalls] Update ImageComponent max width after initial layout if it changes.

MarkVillacampa:
Sometimes the size of the parent view can change after the initial layout. This can happen more often when using [PaywallViewController](https://github.com/RevenueCat/purchases-ios/blob/main/RevenueCatUI/UIKit/PaywallViewController.swift) because the `UIHostingController` can change size after being added to the parent VC and having its AutoLayout constrains set.

With this change, we still only show the image after we've received an initial max width, but we update it when it changes.

Should fix https://github.com/RevenueCat/purchases-ios/issues/5285


ajpallares (approved):
Nice trick! Hopefully this will get fixed once and for all! 🙏 
The code change makes a lot of sense to me! 👍

PR #5292, opened 2025-06-16
Preferred UI locale for UI components

ajpallares:
Completes #5288 

### Checklist

### Motivation
Allowing developers to choose the locale for Paywalls and Customer Center

### Description
Adds new APIs in RevenueCatUI to allow setting the preferred locale for the UI at SDK `configure` time

```swift
import RevenueCat
import RevenueCatUI

let builder = Configuration
        .builder(withAPIKey: "")
        .with(preferredUILocaleOverride: "de_DE")
```

or at any time (as long as `Purchases` is configured)

```swift
import RevenueCat
import RevenueCatUI

Purchases.shared.overridePreferredUILocale("de_DE")
```

### Notes
* Note how these APIs require `import RevenueCatUI` (and not only `import RevenueCat`). The reason is that this preferred locale setting only affects UI components, so it makes more sense to have them declared as part of `RevenueCatUI`.
* Note that no further changes are needed for Customer Center to use the preferred locale for UI. As of #5288, this is achieved automatically, thanks to the preferred locale being sent in the first position for the `"X-Preferred-Locales"` requests header. The Customer Center response already only contains the localizations for the preferred locale (if available)


ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

RevenueCatUI/Purchasing/Purchases+RevenueCatUI.swift
```diff
@@ -0,0 +1,44 @@
...
+@_spi(Internal) import RevenueCat
+
+extension RevenueCat.Purchases {
+
+    /// Updates the preferred locale for RevenueCatUI components.
+    /// - Parameter locale: The preferred locale string in the format "language_region" (e.g., "en_US").
+    /// Use `nil` to reset to the default user locale.
+    ///
+    /// Setting this will affect the display of RevenueCat UI components, such as the Paywalls.
+    public static func updatePreferredUILocale(_ locale: String?) {
```
JayShortway: In terms of naming, both this function and the config parameter, maybe it should somehow indicate that this is not something you normally call. I'm thinking about using words like "override" or "force". Just thinking out loud: 
- `setForcedUILocale()`
- `forceUILocale()`
- `forcePreferredUILocale()`
- `overrideUILocale()`
- `overridePreferredUILocale()`

RevenueCatUI/Purchasing/Purchases+RevenueCatUI.swift
```diff
@@ -0,0 +1,44 @@
...
+//  Created by Antonio Pallares on 13/6/25.
+
+import Foundation
+@_spi(Internal) import RevenueCat
+
+extension RevenueCat.Purchases {
+
+    /// Updates the preferred locale for RevenueCatUI components.
+    /// - Parameter locale: The preferred locale string in the format "language_region" (e.g., "en_US").
+    /// Use `nil` to reset to the default user locale.
```
JayShortway:
```suggestion
    /// Use `nil` to reset to the default user locale determined by the system.
```

RevenueCatUI/Purchasing/Purchases+RevenueCatUI.swift
ajpallares: That's an excellent point. I really like the `overridePreferredUILocale()` option 👌 
Thanks for the improvement!

JayShortway: I liked that one the most too 😄

ajpallares: Done in https://github.com/RevenueCat/purchases-ios/pull/5292/commits/d2dcc0789079c27183919d6c976b9e49e7cdb4c4

Btw @JayShortway, I changed the builder method to
```swift
public func with(preferredUILocaleOverride: String?)
```

Do you agree with this name?

JayShortway: Yes, makes sense!

ajpallares:
@RCGitBot please test

RevenueCatUI/Purchasing/Purchases+RevenueCatUI.swift
```diff
@@ -0,0 +1,45 @@
...
+
+extension RevenueCat.Purchases {
+
+    /// Overrides the preferred locale for RevenueCatUI components.
+    /// - Parameter locale: A locale string in the format "language_region" (e.g., "en_US").
+    /// Use `nil` to remove the override and use the default user locale determined by the system.
+    ///
+    /// Setting this will affect the display of RevenueCat UI components, such as the Paywalls.
+    /// - Important: This method only takes effect after `Purchases` has been configured.
+    public static func overridePreferredUILocale(_ locale: String?) {
```
facumenzella: This is probably out of scope of this PR, but I find counterintuitive that a `static` method has to be used after `configure`. I think the main reason is that this is a silent error. 

What do you think about using an assertion? So it crashes in DEBUG, but has no effect in RELEASE?

Sources/Misc/Locale/PreferredLocalesProvider.swift
```diff
@@ -38,7 +38,7 @@ final class PreferredLocalesProvider {
         self.systemPreferredLocalesGetter = systemPreferredLocalesGetter
     }
 
-    /// Returns the list of the user's preferred languages, with the preferred locale override at the top.
+    /// Returns the list of the user's preferred languages, including the preferred locale override at the top.
```
facumenzella:
```suggestion
    /// Returns the list of the user's preferred languages, including the preferred locale override as the first locale of the array.
```

RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -361,23 +365,38 @@ fileprivate extension PaywallsV2View {
             return .init(locale: Locale.current, localizedStrings: PaywallComponent.LocalizationDictionary())
         }
 
+        var notFoundLocales = [Locale]()
```
facumenzella, line 368: Won't this have at most one locale? 🤔 
What am I missing?

RevenueCatUI/Purchasing/Purchases+RevenueCatUI.swift
ajpallares: You have an **excellent** point and it absolutely is in the scope of this PR. I was checking the other `Purchases` APIs and we are basically crashing when calling `Purchases.shared` before `configure`
```swift
public static var shared: Purchases {
    guard let purchases = Self.purchases.value else {
        fatalError(Strings.purchase.purchases_nil.description)
    }

    return purchases
}
```
> but I find counterintuitive that a static method has to be used after configure

This is key and 100% accurate. Therefore, I think I'm going to change this to not be a `static` method, therefore, callers will need to do `Purchases.shared.overridePreferredUILocale()`, making it similar to all other APIs that require the `configure` to happen beforehand.

Thanks so much for the comment! 🙏

RevenueCatUI/Templates/V2/PaywallsV2View.swift
ajpallares: Good catch! I need to add both the `chosenLocale` and the `defaultLocale` to the array in the final `else` clause below, and I only added the `defaultLocale` 🤦 
Thank you!

RevenueCatUI/Purchasing/Purchases+RevenueCatUI.swift
ajpallares: Changed in https://github.com/RevenueCat/purchases-ios/pull/5292/commits/699c216e63c518bd2f0eb0d5c203ae29b550fc15

RevenueCatUI/Templates/V2/PaywallsV2View.swift
ajpallares: Fixed in https://github.com/RevenueCat/purchases-ios/pull/5292/commits/19834155e0333f1dc4688487e2dae9c72598845f

Sources/Misc/Locale/PreferredLocalesProvider.swift
ajpallares: Changed in https://github.com/RevenueCat/purchases-ios/pull/5292/commits/19834155e0333f1dc4688487e2dae9c72598845f

ajpallares:
@RCGitBot please test

facumenzella (approved):

PR #5296, opened 2025-06-17
Add promotional offers to paywalls

joshdholtz:
## Summary

This PR adds comprehensive promotional offer support to paywalls, enabling paywalls to display and handle promotional offers for eligible users.

### Key Features Added

- **Promotional Offer Cache System**: New `PaywallPromoOfferCache` actor-based system that manages promotional offer eligibility and caching using Combine publishers
- **Subscription History Tracking**: Integrated subscription history monitoring to determine promotional offer eligibility
- **Purchase Handler Integration**: Enhanced purchase flows to support promotional offers in both RevenueCat-managed and app-managed purchase scenarios
- **Component-Level Support**: Promotional offer integration in V2 paywall templates with component overrides for `promo_offer` conditions
- **Eligibility Logic**: Smart eligibility determination based on subscription history and signed promotional offers

### Recent Changes

The latest commits include significant architecture improvements:

- **Combine Publisher Integration**: Refactored `PaywallPromoOfferCache` to use Combine publishers for reactive promotional offer state management (f32458c67)
- **Simplified Component Views**: Updated all V2 component views to use the new cache system (223cb6cb0)
- **Enhanced Cache Logic**: Improved cache implementation with better error handling and state management (a6a142682)
- **Code Cleanup**: Removed unused types and logging to streamline the implementation (18c741a91, 1f918477a)

#### Cache Implementation
- `PaywallPromoOfferCache`: Actor-based cache using Combine publishers for reactive promotional offer state
- Thread-safe with proper concurrency handling using Swift actors
- Publishes promotional offer eligibility changes for real-time UI updates

#### Purchase Flow Integration
- Extended `PaywallPurchasesType` protocol to support promotional offer purchases
- Enhanced `PurchaseHandler` with promotional offer parameter handling
- Updated `MockPurchases` to support promotional offer testing scenarios
- Maintains backward compatibility with existing purchase flows

#### Component Integration
- All V2 template components now integrate with the promotional offer cache
- Component override conditions for promotional offer display logic
- Reactive UI updates based on promotional offer eligibility changes

#### Core Implementation
- `Sources/Paywalls/PaywallPromoOfferCache.swift` - Main cache implementation with Combine publishers
- `RevenueCatUI/Templates/V2/EnvironmentObjects/PaywallPromoOfferCache.swift` - Environment object wrapper
- `RevenueCatUI/Purchasing/PaywallPurchasesType.swift` - Protocol extension for promo offers
- `RevenueCatUI/Purchasing/PurchaseHandler.swift` - Purchase flow integration
- `RevenueCatUI/Purchasing/MockPurchases.swift` - Mock implementation for testing

#### UI Components
- `RevenueCatUI/Templates/V2/PaywallsV2View.swift` - V2 paywall integration with reactive cache
- All V2 component views updated to use the new cache system
- `Sources/Paywalls/Components/Common/ComponentOverrides.swift` - Override conditions

#### Configuration & Testing
- Project configuration updates for all targets
- Paywall tester integration for testing promotional offers
- Build configuration enhancements

### Testing Strategy

⚠️ **Test Coverage Gap**: The new `PaywallPromoOfferCache` classes currently have no test coverage. This should be addressed before merging.

Existing promotional offer tests cover:
- Core promotional offer data structures (`PromotionalOfferTests.swift`)
- Purchase handler flows (`PurchaseHandlerTests.swift`)
- Customer center promotional offer UI (`PromotionalOfferViewModelTests.swift`)

### Known Issues

1. **Missing Test Coverage**: No tests exist for the new cache implementations
2. **Mock Interface**: MockPurchases may need additional promotional offer purchase method implementation

### Migration Notes

This implementation is backward compatible. Existing paywalls without promotional offer configuration will continue to work unchanged. The promotional offer functionality is opt-in through paywall configuration.

### Next Steps

1. Add comprehensive test coverage for `PaywallPromoOfferCache` classes
2. Add integration tests for end-to-end promotional offer scenarios with Combine publishers
3. Consider performance testing for reactive cache operations

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Packages/Package/PackageComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Packages/Package/PackageComponentView.swift
@@ -88,7 +88,7 @@ struct PackageComponentView_Previews: PreviewProvider {
     )
     @MainActor
     static let packageContextSelected = PackageContext(
-        package: nil,
+        package: Self.package,
         variableContext: .init()
     )
 


RevenueCatUI/CustomerCenter/Data/CustomerCenterPurchases.swift
```diff
@@ -56,6 +56,14 @@ final class CustomerCenterPurchases: CustomerCenterPurchasesType {
                                                     product: product)
     }
 
+    func purchase(
+        product: StoreProduct
+    ) async throws -> PurchaseResultData {
+        try await Purchases.shared.purchase(
+            product: product
+        )
+    }
```
JayShortway: Why was this needed, in `CustomerCenterPurchases`?

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -191,7 +208,7 @@ extension PurchaseHandler {
     }
 
     @MainActor
-    func performExternalPurchaseLogic(package: Package) async throws {
+    func performExternalPurchaseLogic(package: Package, promotionalOffer: PromotionalOffer?) async throws {
```
JayShortway, line 218: This new parameter seems to be unused?

RevenueCatUI/Templates/V2/EnvironmentObjects/PromotionalOfferEligibilityContext.swift
```diff
@@ -0,0 +1,138 @@
...
+
+        let status = cache[package.storeProduct.productIdentifier] ?? .unknown
+
+        switch status {
+        case .unknown, .ineligible:
+            return false
+        case .unsignedEligible, .signedEligible:
+            return true
+        }
+    }
```
JayShortway: It seems like this could result in the paywall showing promo offer content while the user is actually not eligible, and thus the actual purchase sheet not reflecting that promo offer. Is that correct? That would be a suboptimal user experience imo. 

Is there no way we can wait for a successfully signed offer? Can we proactively sign all offers we find?

RevenueCatUI/Templates/V2/Components/Packages/Package/PackageComponentView.swift
```diff
@@ -144,6 +144,7 @@ struct PackageComponentView_Previews: PreviewProvider {
                 component: .init(
                     packageID: "weekly",
                     isSelectedByDefault: false,
+                    applePromoOfferProductCode: nil,
```
JayShortway, line 149: Is "Code" the same as "Identifier" here? If so, the latter would be more clear imo.

RevenueCatUI/Templates/V2/EnvironmentObjects/PromotionalOfferEligibilityContext.swift
```diff
@@ -0,0 +1,138 @@
...
+
+                do {
+                    let promoOffer = try await Purchases.shared.promotionalOffer(
+                        forProductDiscount: discount,
+                        product: storeProduct
+                    )
+                    cache[storeProduct.productIdentifier] = .signedEligible(promoOffer)
+                } catch {
+                    // Not eligible or signing failed — leave status unchanged
+                    print("Signed offer creation failed for \(storeProduct.productIdentifier): \(error)")
```
JayShortway: This should probably use our own Logger?

RevenueCatUI/Templates/V2/EnvironmentObjects/PromotionalOfferEligibilityContext.swift
```diff
@@ -0,0 +1,138 @@
...
+
+    enum Status: Equatable {
+        case unknown
+        case ineligible
+        case unsignedEligible
+        case signedEligible(PromotionalOffer)
+    }
+
+    @Published
+    private(set) var cache: [ProductID: Status] = [:]
```
JayShortway: Seems like we're not doing any cache invalidation? If someone is `signedEligible` for a promo offer and purchases it, both `get()` and `isMostLikelyEligible()` will still return `signedEligible`.

I understand we can call `computeEligibility()` before calling `get()` or `isMostLikelyEligible()`, but that's not enforced and so can easily be forgotten in the future, resulting in bugs.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -352,7 +352,7 @@ final class PurchasesOrchestrator {
     }
 
     func promotionalOffer(forProductDiscount productDiscount: StoreProductDiscountType,
-                          product: StoreProductType,
+                          product: StoreProduct,
```
JayShortway: Why was this needed? Could we not make the required changes to `StoreProductType`?

JayShortway (commented):
Awesome that we're building this! Just a few comments.

RevenueCatUI/CustomerCenter/Data/CustomerCenterPurchases.swift
joshdholtz: Uhhhh... I thought I deleted this 😅 Double checking!

joshdholtz: I know I deleted this... I blame Xcode! Deleted this for a second time now

RevenueCatUI/Purchasing/PurchaseHandler.swift
joshdholtz: Ah, it is still! I think I need to bubble this up to the external purchase logic 😬

RevenueCatUI/Templates/V2/Components/Packages/Package/PackageComponentView.swift
joshdholtz: It _is_ but in ASC and other doc its actually referred to as the code which is what we will put in paywalls editor 😬  So... at some place it will have to differ but not sure where that place would be I guess 🤷‍♂️

RevenueCatUI/Templates/V2/EnvironmentObjects/PromotionalOfferEligibilityContext.swift
joshdholtz: @JayShortway Yeah yeah, this _definitely_ is not its final form 😅 Right now, this context gets computed anytime a paywall is shown. Its not really acting as a cache _yet_.

In its current form, my goal was...
1) Using local SK2 (and non-RC backend), get eligible status only so that the paywall UI overrides don't change on the use
2) After that, then go get the signed promo offers from the RC backend (which is needed for the purchase)

I would ideally like to move step 1 outside of paywall to an actual cache so less async code needed for paywall:
1) On SDK initialize
2) On queue/transactions updated 

So in its current form, I don't think cache invalidation is actually needed but moving it outside of the paywall lifecycle is when we definitely would.

And I'm not sure what my goal with this PR is yet 😅  I _really_ don't want to make a whole complicated cache right now and make that a follow up PR 🤷‍♂️ 

Thoughts?

RevenueCatUI/Templates/V2/EnvironmentObjects/PromotionalOfferEligibilityContext.swift
joshdholtz: Soooooo....

> Is there no way we can wait for a successfully signed offer? Can we proactively sign all offers we find?

Proactively signing offers is not good for the pandas 🐼 ... It requires an API request and we do not want to do that for every instance of the SDK on every load (that would be a lot). So what I'm trying to do is...

The logic for Apple promo offers is actually pretty straight forward. You either must:
1) Have an existing subscription
2) Have an expired subscription

So that can pretty confidently be done locally with only StoreKit 2's `Transactions` API. We need to do this as efficiently and quickly as possibly (before paywall loads) since a bunch of components (text, images, stacks, etcs) have overrides for the `promo_offer` condition. So we don't want to do anything async because it after its done, it will trigger a UI refresh which could be jarring to the user.

So, IMO, the trade off of using SK2 locally to get the status upfront is worth saving the pandas and the potential UI change 🤷‍♂️

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
joshdholtz: Oh, I was experimenting 😅 I will revert this!

RevenueCatUI/Templates/V2/EnvironmentObjects/PromotionalOfferEligibilityContext.swift
JayShortway: Totally agree with the goals, and I think the trade off makes sense. I'm just afraid of the signing failing, and the bad UX that follows from that. What's your estimated probability of the signing failing? Maybe we should add diagnostics for it (in a separate PR)?

RevenueCatUI/Templates/V2/EnvironmentObjects/PromotionalOfferEligibilityContext.swift
JayShortway: Yes I think that all makes sense! `preWarmEligibilityCache` on SDK initialize haha. A follow-up PR sounds good to me, but maybe we should have both ready before merging? Or do you think the cache implementation will take very long and we should #abs? I'd really like to avoid this lingering, as I'm afraid its current form can lead to bugs in the future haha.

joshdholtz:
@RCGitBot please test

Sources/Paywalls/PaywallPromoOfferCache.swift
```diff
@@ -0,0 +1,180 @@
...
+//
+//  Created by Josh Holtz on 6/17/25.
+//
+// swiftlint:disable missing_docs
+
+import Combine
+import StoreKit
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+public actor SubscriptionHistoryTracker {
```
ajpallares: Is there any reason why `SubscriptionHistoryTracker` is `public`. It looks to me (based on its usage), that it could be `fileprivate`.

Sources/Paywalls/PaywallCacheWarming.swift
```diff
@@ -55,6 +58,7 @@ actor PaywallCacheWarming: PaywallCacheWarmingType {
     private let introEligibiltyChecker: TrialOrIntroPriceEligibilityCheckerType
     private let imageFetcher: PaywallImageFetcherType
     private let fontsManager: PaywallFontManagerType
+    internal let promoOfferCache: PaywallPromoOfferCacheType
```
ajpallares: Probably one of my nits but it looks like adding the `promoOfferCache` variable to this `PaywallCacheWarming` object doesn't really follow the `PaywallCacheWarming`'s purpose (which is a way to pre-load some caches). 
Based on this implementation, regarding `promoOfferCache`, it simply servers as a way to hold a reference to it.

Sources/Paywalls/PaywallPromoOfferCache.swift
```diff
@@ -0,0 +1,180 @@
...
+                    break
+                }
+            }
+        }
+
+        continuation?.yield(Update(hasAnySubscriptionHistory: found))
+    }
+}
+
+@_spi(Internal) public protocol PaywallPromoOfferCacheType: Sendable {
```
ajpallares: I think this protocol is unnecessary as it's empty and in `PaywallView` there's this casting `Purchases.shared.paywallPromoOfferCache as? PaywallPromoOfferCache`. So we might as well simply expose the `PaywallPromoOfferCache` object directly.

Alternatively, we could add the `hasAnySubscriptionHistory` variable to the protocol declaration and use the object in `PaywallView` without the cast

Sources/Paywalls/PaywallPromoOfferCache.swift
```diff
@@ -0,0 +1,180 @@
...
+
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@_spi(Internal) public actor PaywallPromoOfferCache: ObservableObject, PaywallPromoOfferCacheType {
```
ajpallares: I think `ObservableObject` is unnecessary since there are no `@Published` properties

Sources/Paywalls/PaywallPromoOfferCache.swift
```diff
@@ -0,0 +1,180 @@
...
+    private func listenToSubscriptionUpdates() async {
+        for await update in await subscriptionTracker.updates {
+            self.hasAnySubscriptionHistory = update.hasAnySubscriptionHistory
+        }
+    }
+
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+@_spi(Internal) public final class PaywallPromoOfferCacheV2: ObservableObject {
```
ajpallares: I _think_ this could be part of RevenueCatUI, as it's only being used there. This way we avoid the `@_spi public` clauses.

RevenueCatUI/PaywallView.swift
```diff
@@ -261,6 +264,14 @@ public struct PaywallView: View {
         }
     }
 
+    var paywallPromoOfferCache: PaywallPromoOfferCache {
+        if Purchases.isConfigured, let cache = Purchases.shared.paywallPromoOfferCache as? PaywallPromoOfferCache {
```
ajpallares: Here we should rely on the `purchases` instance of `self.purchaseHandler` instead of using `Purchases.shared` (we can also use `self.purchaseHandler.isConfigured` instead of `Purchases.isConfigured`). Trust me, the RC App will thank us for this 😅

Sources/Paywalls/PaywallPromoOfferCache.swift
ajpallares, line 1: Seeing the code, it looks to me that what we need is to offer a way for interested parties (e.g. `PaywallPromoOfferCacheV2`) to subscribe to the value of the `hasAnySubscriptionHistory` boolean in `PaywallPromoOfferCache`.

I _think_ a cleaner solution would be to have `hasAnySubscriptionHistory` be a Combine publisher that we can send around to interested parties so that they can subscribe to it.¡ and update accordingly. Or maybe an `AsyncStream`.

To me, this publisher would be the only `@_spi(Internal) public` _thing_  that RevenueCat (i.e. `Purchases`) should expose for RevenueCatUI (instead of exposing the whole `PaywallPromoOfferCache` as it's happening now). This way, we can get rid of the `PaywallPromoOfferCacheType` protocol and the `PaywallPromoOfferCache` would just be an `internal` implementation detail in RevenueCat which would only be responsible of publishing the latest `hasAnySubscriptionHistory` boolean value (as it already does).

In my mind, the `PaywallPromoOfferCacheV2` object would live in RevenueCatUI and it would receive the `hasAnySubscriptionHistory` publisher (or AsyncStream) in its `init`.

Perhaps having a publisher/AsyncStream is overengineering and we just need to expose `hasAnySubscriptionHistory` as a boolean that we pass to `PaywallPromoOfferCacheV2` at `init` time (as right now).

But, unless I'm missing something, I think that my whole point still stands, which is only exposing the `hasAnySubscriptionHistory` boolean from RevenueCat (instead of the whole object) and keep `PaywallPromoOfferCacheV2` in RevenueCatUI.

WDYT?

RevenueCatUI/Templates/V2/Components/Packages/Package/PackageComponentView.swift
tonidero: Hmm could be good to make it more generic in the dashboard to account for Google's "tags" or "base-plan-id" (not sure which one we should use on Android yet)

RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -98,11 +112,18 @@ struct PaywallsV2View: View {
     private let onDismiss: () -> Void
     private let fallbackContent: FallbackContent
 
-    init(
+    @ObservedObject
+    private var paywallPromoOfferCache: PaywallPromoOfferCache
+
+    @StateObject
+    private var myViewModel = MyViewModel()
```
tonidero: This name seems too generic 😅 Maybe something like `PaywallsV2ViewModel`?

RevenueCatUI/Templates/V2/Previews/TemplateComponentsViewPreviews/ButtonWithFooterPreview.swift
```diff
@@ -92,6 +92,7 @@ private enum ButtonWithSheetPreview {
     static let package = PaywallComponent.PackageComponent(
         packageID: "weekly",
         isSelectedByDefault: false,
+        applePromoOfferProductCode: nil,
```
tonidero, line 95: I guess that `promoOffer` is already an "Apple" concept so I wonder if we should remove the `apple` prefix... But then again, I'm not opposed to make it more explicit

RevenueCatUI/Templates/V2/ViewModelHelpers/PackageValidator.swift
```diff
@@ -19,12 +19,12 @@ import RevenueCat
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 class PackageValidator {
 
-    typealias PackageInfo = (package: Package, isSelectedByDefault: Bool)
+    typealias PackageInfo = (package: Package, isSelectedByDefault: Bool, promotionalOfferProductCode: String?)
```
tonidero, line 22: This is out of scope for this PR, but I've been thinking if we should support a list of promo offers codes, instead of a single one... So basically the first code would have the highest priority if it's eligibile, if not fall back to the next... Just thinking about making the system more flexible, but I think this is good for now :)

tonidero (commented):
Mostly just @ajpallares's comments, but excited to support this! 🙌

Sources/Paywalls/PaywallPromoOfferCache.swift
joshdholtz: Okay, this is now implemented! Can you review again? 😇

RevenueCatUI/Templates/V2/PaywallsV2View.swift
joshdholtz: This is gone now 😁

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -160,14 +164,24 @@ extension PurchaseHandler {
     func purchase(package: Package) async throws {
         switch self.purchases.purchasesAreCompletedBy {
         case .revenueCat:
-            try await performPurchase(package: package)
+            try await performPurchase(package: package, promotionalOffer: nil)
+        case .myApp:
+            try await performExternalPurchaseLogic(package: package, promotionalOffer: nil)
+        }
+    }
```
ajpallares: Nit: I would replace the whole method body with 
```swift
func purchase(package: Package) async throws {
    try await purchase(package: package, promotionalOffer: nil)
}
```

_Sent from the good old Objective-C days_

RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
```diff
@@ -13,19 +13,22 @@
 // swiftlint:disable file_length
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares: I think the `@_spi(Internal)` is not needed for now
```suggestion
import RevenueCat
```

RevenueCatUI/Templates/V2/Components/Icon/IconComponentView.swift
```diff
@@ -12,19 +12,22 @@
 //  Created by Josh Holtz on 6/11/24.
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares: Same here
```suggestion
import RevenueCat
```

RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
```diff
@@ -12,19 +12,22 @@
 //  Created by Josh Holtz on 6/11/24.
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares: Same here
```suggestion
import RevenueCat
```

RevenueCatUI/Templates/V2/Components/Packages/Package/PackageComponentView.swift
```diff
@@ -12,7 +12,7 @@
 //  Created by Josh Holtz on 9/27/24.
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares, line 15: This `@_spi(Internal)` would not be needed, unless we make `PaywallComponent.PackageComponent`'s `applePromoOfferProductCode` `@_spi(Internal)`

Sources/Paywalls/Components/PaywallPackageComponent.swift
```diff
@@ -22,30 +22,35 @@ public extension PaywallComponent {
         let type: ComponentType
         public let packageID: String
         public let isSelectedByDefault: Bool
+        public let applePromoOfferProductCode: String?
```
ajpallares: There's discussion about the name of this property. I _think_ it doesn't need to be public, so we can make it `@_spi(Internal)` to only access it from RevenueCatUI. That will allow us to change the name in the future if we want without technically breaking the API.
```suggestion
        @_spi(Internal) public let applePromoOfferProductCode: String?
```

RevenueCatUI/Templates/V2/Components/Packages/Package/PackageComponentView.swift
ajpallares: I think we can make `applePromoOfferProductCode` be `@_spi(Internal)` so that we can rename it when we need without breaking the API

RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
```diff
@@ -11,7 +11,7 @@
 //
 //  Created by James Borthwick on 2024-08-20.
 
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares: I think this is not needed

RevenueCatUI/Templates/V2/Components/Timeline/TimelineComponentView.swift
```diff
@@ -12,7 +12,7 @@
 //  Created by Mark Villacampa on 15/1/25.
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares:
```suggestion
import RevenueCat
```

RevenueCatUI/Templates/V2/EnvironmentObjects/PackageContext.swift
```diff
@@ -11,7 +11,7 @@
 //
 //  Created by Josh Holtz on 11/14/24.
 
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares:
```suggestion
import RevenueCat
```

RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -144,57 +156,66 @@ struct PaywallsV2View: View {
...
-                "Error decoding paywall response on: \(errorInfo.keys.joined(separator: ", "))"
-            )
-        } else {
-            switch self.paywallStateManager.state {
-            case .success(let paywallState):
-                LoadedPaywallsV2View(
-                    paywallState: paywallState,
-                    uiConfigProvider: self.uiConfigProvider,
-                    onDismiss: self.onDismiss
+        VStack(spacing: 0) {
```
ajpallares, line 159: Why do we need the `VStack` here? It looks to me like the views are the same as before, so if it wasn't needed before, it shouldn't be needed now?

RevenueCatUI/Templates/V2/Previews/TemplateComponentsViewPreviews/ButtonWithFooterPreview.swift
```diff
@@ -12,7 +12,7 @@
 //  Created by Josh Holtz on 9/26/24.
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares:
```suggestion
import RevenueCat
```

RevenueCatUI/Templates/V2/Previews/TemplateComponentsViewPreviews/FamilySharingTogglePreview.swift
```diff
@@ -13,7 +13,7 @@
 // swiftlint:disable file_length
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares:
```suggestion
import RevenueCat
```

RevenueCatUI/Templates/V2/Previews/TemplateComponentsViewPreviews/MultiTierPreview.swift
```diff
@@ -13,7 +13,7 @@
 // swiftlint:disable file_length
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares:
```suggestion
import RevenueCat
```

RevenueCatUI/Templates/V2/Previews/TemplateComponentsViewPreviews/PurchaseButtonInPackagePreview.swift
```diff
@@ -12,7 +12,7 @@
 //  Created by Josh Holtz on 9/26/24.
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares:
```suggestion
import RevenueCat
```

RevenueCatUI/Templates/V2/Previews/TemplateComponentsViewPreviews/Template1Preview.swift
```diff
@@ -12,7 +12,7 @@
 //  Created by Josh Holtz on 9/26/24.
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares:
```suggestion
import RevenueCat
```

Tests/UnitTests/Mocks/MockPaywallCacheWarming.swift
```diff
@@ -12,7 +12,7 @@
 //  Created by Nacho Soto on 8/7/23.
 
 import Foundation
-@testable import RevenueCat
+@_spi(Internal) @testable import RevenueCat
```
ajpallares:
```suggestion
@testable import RevenueCat
```

ajpallares (approved):
Looks great!! 🙌 
Only some minor comments and nits (`@_spi` policeman here)

Sources/Paywalls/Components/PaywallPackageComponent.swift
joshdholtz: All of these should be SPI-ed at some point... even thought it would be "breaking" 😅

PR #5298, opened 2025-06-17
Fix index and mgiration guides docs not being updated

MarkVillacampa:
### Checklist


ajpallares (approved):
Nice catch! 👏  Thank you!

PR #5300, opened 2025-06-17
Remove trailing text for PurchaseCardView, and simplify billing information. 

facumenzella:
### Motivation
The trailing text was showing duplicated data for most of the cases, so we're removing it

### Description
- Remove trailing price text
- Simplify billing information (subtitle)
- Use entitlement identifier if promotional.
- Add lifetime badge
- Tweak `expired` vs `expires` in billing information
- Mark it as lifetime if non-consumable

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/PurchaseInformation+Mock.swift
+++ b/RevenueCatUI/CustomerCenter/Data/PurchaseInformation+Mock.swift
@@ -93,7 +93,7 @@ extension PurchaseInformation {
         store: .appStore,
         isSubscription: true,
         isExpired: true,
-        expirationDate: nil,
+        expirationDate: Self.defaultExpirationDate,
         renewalDate: nil
     )
 


RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -236,7 +240,7 @@ struct PurchaseInformation {
                 self.isExpired = !isActive
 
             case .nonSubscription:
-                self.isLifetime = true
+                self.isLifetime = subscribedProduct?.productType == .nonConsumable
```
vegaro: fyi in android I changed this to `isSubscription`, because I was getting confused with lifetime promotionals

vegaro (approved):

RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
facumenzella: Replaces with `isSubscription`, and added a computed property `isLifetimeSubscription`

PR #5303, opened 2025-06-18
Add Mac and Mac Catalyst variants to a preview.

chrisvasselli:
I don't plan to actually merge this – just testing how Emerge handles Mac and Mac Catalyst.

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Text/TextComponentView.swift
@@ -139,6 +139,50 @@ struct TextComponentView_Previews: PreviewProvider {
         .previewRequiredEnvironmentProperties()
         .previewLayout(.sizeThatFits)
         .previewDisplayName("Default")
+        
+        // Default - Mac
+        TextComponentView(
+            // swiftlint:disable:next force_try
+            viewModel: try! .init(
+                localizationProvider: .init(
+                    locale: Locale.current,
+                    localizedStrings: [
+                        "id_1": .string("Hello, world")
+                    ]
+                ),
+                uiConfigProvider: .init(uiConfig: PreviewUIConfig.make()),
+                component: .init(
+                    text: "id_1",
+                    color: .init(light: .hex("#000000"))
+                )
+            )
+        )
+        .previewRequiredEnvironmentProperties()
+        .previewLayout(.sizeThatFits)
+        .previewDisplayName("Default - Mac")
+        .previewDevice("Mac")
+        
+        // Default - Mac Catalyst
+        TextComponentView(
+            // swiftlint:disable:next force_try
+            viewModel: try! .init(
+                localizationProvider: .init(
+                    locale: Locale.current,
+                    localizedStrings: [
+                        "id_1": .string("Hello, world")
+                    ]
+                ),
+                uiConfigProvider: .init(uiConfig: PreviewUIConfig.make()),
+                component: .init(
+                    text: "id_1",
+                    color: .init(light: .hex("#000000"))
+                )
+            )
+        )
+        .previewRequiredEnvironmentProperties()
+        .previewLayout(.sizeThatFits)
+        .previewDisplayName("Default-Mac Catalyst")
+        .previewDevice("Mac Catalyst")
 
         // Markdown
         TextComponentView(


clindsay3:
I think the emerge failures are just because the Catalyst versions are being uploaded as separate bundle ID's, so that they can be verified separately. I assume that once they are merged into main, this issue will stop happening (although I'm new to precisely how Emerge works).

clindsay3:
There are also a few failures in the Emerge screenshots, but I think that shouldn't block this commit, since this commit is just about getting the screenshots setup, not actually addressing the issues they discover yet.

Sources/LocalReceiptParsing/Helpers/ProcessInfo+Extensions.swift
```diff
@@ -59,6 +63,44 @@ extension ProcessInfo {
...
+            } else {
+                switch UIDevice.current.userInterfaceIdiom {
+                case .mac:
+                    return "Mac Catalyst Optimized for Mac"
+                case .pad:
+                    return "Mac Catalyst Scaled to iPad"
+                default:
+                    return "Unexpected Platform on Mac Catalyst"
+                }
+            }
```
joshdholtz, line 98: This is a crazy switch but it makes sense 😀

Tests/TestingApps/PaywallsTester/PaywallsTesterTests/PaywallsTesterTests.swift
```diff
@@ -9,5 +9,10 @@ import XCTest
 import SnapshottingTests
 
 final class PaywallsTesterTests: SnapshotTest {
-
+    override class func snapshotPreviews() -> [String]? {
+        // Gets around an issue that was causing StoreKit previews to be included when running in Catalyst mode.
+        // Should be fixed in EmergeTools v0.10.23+, and this won't be necessary anymore.
+        // PR: https://github.com/EmergeTools/SnapshotPreviews/pull/239
+        return ["PaywallsTester.*", "RevenueCatUI.*"]
+    }
```
joshdholtz: :chefs_kiss:

joshdholtz:
@clindsay3 FYI, I just rebased this from `main` so that the Emerge Snapshot build has latest previous commit to run 💪

Sources/LocalReceiptParsing/Helpers/ProcessInfo+Extensions.swift
```diff
@@ -59,6 +63,44 @@ extension ProcessInfo {
...
+        } else {
+            return "iOS"
```
ajpallares: Would it make sense to have the iOS iPadOS differentiation here?
```swift
switch UIDevice.current.userInterfaceIdiom {
case .phone:
    return "iOS"
case .pad:
    return "iPadOS"
default:
    return "Unexpected iOS app"
}
```

Tests/TestingApps/PaywallsTester/PaywallsTester.xcodeproj/project.pbxproj
```diff
@@ -791,6 +793,7 @@
 				BUNDLE_LOADER = "$(TEST_HOST)";
 				CODE_SIGN_STYLE = Automatic;
 				CURRENT_PROJECT_VERSION = 1;
+				DEVELOPMENT_TEAM = 8SXR2327BM;
```
ajpallares, line 796: I don't know why the dev team appeared here, but it's already present in this same file and in other files as well, so I think it's fine 👍

Tests/TestingApps/PaywallsTester/PaywallsTesterTests/PaywallsTesterTests.swift
ajpallares: Is this workaround related to the new file `emerge/emerge_config.yaml`? If so, I would append instructions to the sentence 
> and this won't be necessary anymore

so that when this is not necessary anymore, we also remember to delete the `emerge/emerge_config.yaml` file. Does it make sense?

fastlane/Fastfile
```diff
@@ -344,12 +344,135 @@ platform :ios do
 
   desc "Build the Paywalls Test app for Emerge Snapshots"
   lane :build_paywalls_tester_for_emerge do |options|
+    build_paywalls_tester_for_emerge_ios(options)
+    build_paywalls_tester_for_emerge_mac(platform: "mac-catalyst-scaled-to-match-ipad")
+    build_paywalls_tester_for_emerge_mac(platform: "mac-catalyst-optimized-for-mac")
+
+    # As of July 2025, Emerge Tools does not support iOS apps on Mac.
+    # build_paywalls_tester_for_emerge_mac(platform: "ipad-app-on-mac")
+    # build_paywalls_tester_for_emerge_mac(platform: "iphone-app-on-mac")
```
ajpallares: This comment seems to contradict the one below. If I'm understanding correctly, this will eventually be supported, right? Then maybe I'd say it like this:
```suggestion
    # As of July 2025, Emerge Tools does not support iOS apps on Mac yet.
    # build_paywalls_tester_for_emerge_mac(platform: "ipad-app-on-mac")
    # build_paywalls_tester_for_emerge_mac(platform: "iphone-app-on-mac")
```

ajpallares (approved):
Looks good to me! I just have some small questions, mostly about the wording of comments for future us 😄 
Great job!

Tests/TestingApps/PaywallsTester/PaywallsTesterTests/PaywallsTesterTests.swift
clindsay3: Yup, good call! Will do.

Basically, this is the solution for running locally, and the emerge_config.yaml file is the solution for running on Emerge's server. I'll add that to the comments on both as well.

Sources/LocalReceiptParsing/Helpers/ProcessInfo+Extensions.swift
clindsay3: I like it! Will do!

Tests/TestingApps/PaywallsTester/PaywallsTester.xcodeproj/project.pbxproj
clindsay3: I _think_ it's because of differences in codesigning requirements for Mac Catalyst apps (which run on the local Mac), and iOS apps which run in the simulator. But agreed that it seems pretty harmless, as none of our other CI builds fail with it.

Tests/TestingApps/PaywallsTester/PaywallsTesterTests/PaywallsTesterTests.swift
clindsay3: Done!

Sources/LocalReceiptParsing/Helpers/ProcessInfo+Extensions.swift
clindsay3: Done!

PR #5309, opened 2025-06-20
Fix paywall locale receiving UN M.49 standard localization

ajpallares:
### Motivation
Found a bug where displaying a paywall with the only Spanish localization being `"es_MX"` would not show some localization strings (device region: Spain, preferred languages: es, en_US)

### Description
In this case, the server's `ui_config.localization` contains `"es_419"` (UN M.49 standard for the Latin America and the Caribbean region) but no `"es_MX"`. With the existing implementation, the SDK was not able to match `"es_MX"` with `"es_419"`.

This PR adds an extra step when trying to find the matching locale from `UIConfig.localizations`. As a last resort, it tries to find a key in `ui_config.localization` with matching language and script.

### DIFF
--- a/RevenueCatUI/Templates/V2/Localizations/LocaleFinder.swift
+++ b/RevenueCatUI/Templates/V2/Localizations/LocaleFinder.swift
@@ -42,6 +42,14 @@ extension Dictionary where Key == String {
             return exactMatch
         }
 
+        // Last resort: matching language and script converting dictionary keys to language and script as well
+        // This covers cases where dictionary keys contain the UN M.49 standard. For example:
+        // If locale = "es_MX" and dictionary key = "es_419" (both have rc_languageAndScript = "es-Latn")
+        if let onlyLanguageAndScriptIdentifier = locale.rc_languageAndScript,
+           let match = self.valueForLanguageAndScript(onlyLanguageAndScriptIdentifier) {
+            return match
+        }
+
         return nil
     }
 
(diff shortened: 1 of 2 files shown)


MarkVillacampa (approved):

RevenueCatUI/Templates/V2/Localizations/LocaleFinder.swift
```diff
@@ -42,6 +42,14 @@ extension Dictionary where Key == String {
             return exactMatch
         }
 
+        // Last resort: matching language and script converting dictionary keys to language and script as well
+        // This covers cases where dictionary keys contain the UN M.49 standard. For example:
+        // If locale = "es_MX" and dictionary key = "es_419" (both have rc_languageAndScript = "es-Latn")
+        if let onlyLanguageAndScriptIdentifier = locale.rc_languageAndScript,
+           let match = self.valueForLanguageAndScript(onlyLanguageAndScriptIdentifier) {
```
tonidero: Not sure if it's possible/worth to add some tests for this?

tonidero (approved):
I think it makes sense! :shipit:

RevenueCatUI/Templates/V2/Localizations/LocaleFinder.swift
ajpallares: Indeed. Good point that I totally forgot about! Thank you for the suggestion!

ajpallares:
Actually, I was writing a test for this and the first one I wrote did not pass
```swift
func test_es_MX_Matches_es_419_ByLanguageAndScript() {
    let localizations = [
        "es_419": Self.expectedTranslations,
        "es": ["wrong": "this is es"]
    ]

    let locale = Locale(identifier: "es_MX")

    // This tests that "es_MX" matches "es_419" due to both sharing the same language and script ("es-Latn")
    let foundLocalizations = localizations.findLocale(locale)
    expect(foundLocalizations).to(equal(Self.expectedTranslations))
}
```

The reason is that I originally added the new check as a last resort (it was the last check). But that makes "es_MX" match "es" before "es_419", which is wrong.

Therefore, I think the new check should happen **before** checking only the language identifier (done https://github.com/RevenueCat/purchases-ios/pull/5309/commits/14d3c99fa875e0b279d589b3f8adf9cd2395f577)

Please lmk if you think otherwise!

tonidero (approved):
Yeei for tests! 🙌

ajpallares:
The fix in this PR is not correct (tests were failing randomly; thanks again @tonidero for suggesting the tests 🙏). I created a new PR with a whole different approach here https://github.com/RevenueCat/purchases-ios/pull/5316

PR #5311, opened 2025-06-20
Fix broken Customer Center strings

vegaro:
Found some strings that had the wrong key compared to the backend


facumenzella:
We need to generate this in the near future! 🙏

PR #5312, opened 2025-06-20
Consider Offerings cache stale when preferred locales change

ajpallares:
Similar idea to https://github.com/RevenueCat/purchases-android/pull/2438

### Checklist

### Motivation
Now that we allow for the developer to override the preferred locale, we need to make sure that the caching of Get Offerings considers the locales as well.

### Description
Up until now, the offerings cache would only implement time-based staleness. With this PR, the offerings cache will also be considered stale if the preferred locales change.


ajpallares:
I considered actively refreshing the offerings cache when receiving a different preferred language override here https://github.com/RevenueCat/purchases-ios/blob/e74be6ee753ded7ea7cf8d5678c63d136c764dba/Sources/Purchasing/Purchases/Purchases.swift#L1851-L1853

but I thought it would overcomplicate things and I don't really think we want to call get offerings whenever the preferred locale is overriden. WDYT @RevenueCat/coresdk?

tonidero:
> don't really think we want to call get offerings whenever the preferred locale is overriden

I guess it depends... I would think that's an uncommon enough operation that it would be fine either way. I guess I kinda like the current lazy approach in case a customer changes the languages multiple times, so we only fetch once, when required. So yeah, I think it's fine as is!

tonidero (approved):

JayShortway (approved):
I think this is a good improvement either way (include locale in "staleness" check), whether we proactively fetch offerings or not. 

We should consider Paywalls though. We want the Paywall to be properly localized and to appear immediately, so it might be good to proactively fetch offerings. I'd say that can be a separate PR. Happy to hear thoughts! @RevenueCat/coresdk

ajpallares:
> We should consider Paywalls though. We want the Paywall to be properly localized and to appear immediately, so it might be good to proactively fetch offerings

@JayShortway this was my thinking. But then I thought of @tonidero's argument for not doing it
> I guess I kinda like the current lazy approach in case a customer changes the languages multiple times, so we only fetch once, when required.

and I don't think we should because it could be abused as a workaround to force refreshing the offerings cache.

JayShortway:
@ajpallares Hmm, good point. We could use the same rate limiter we have on [`syncAttributesAndOfferingsIfNeeded`](https://github.com/RevenueCat/purchases-ios/blob/99eb52bc16f5e4d87b5537ffa8a355d5b361a76d/Sources/Purchasing/Purchases/Purchases.swift#L913)?

ajpallares:
> @ajpallares Hmm, good point. We could use the same rate limiter we have on [`syncAttributesAndOfferingsIfNeeded`](https://github.com/RevenueCat/purchases-ios/blob/99eb52bc16f5e4d87b5537ffa8a355d5b361a76d/Sources/Purchasing/Purchases/Purchases.swift#L913)?

I didn't know about that rate limiter! It seems like that would certainly solve the potential abuse of the locale API 🤔 
Ok then. We can discuss this topic and decide whether we want to do the proactive get offerings on locale change then. Thanks @JayShortway!

PR #5314, opened 2025-06-23
Fix issue where new UI tests were being added to the paywall screenshots test plan.

chrisvasselli:
### Motivation
Extra unnecessary tests were being run every time we generated validation screenshots.

### Description
Changed the test plan os that instead of excluding all the other tests, we now just include the one test we need. Tested that `fastlane record_paywall_screenshots` still worked properly after the change, will verify that the CI integration works before merging.


clindsay3:
@joshdholtz if there are others I should assign to review PR's, let me know!

PR #5315, opened 2025-06-23
[Draft] Remove the RevenueCat and RevenueCatUI targets in RevenueCat.xcodeproj.

clindsay3:
### Motivation
Broadly, I think these targets are probably causing more confusion than anything, since I suspect they aren't actually used.

More specifically, in order to run our paywall validation tests (`TakeScreenshotTests`) on macOS, we will need to move the screenshot generation to a target in the Xcode project, from the current `RevenueCatUI` test target defined in the Package.swift. This is because test targets in a Package.swift cannot be run inside a host application, and in order to interact with UI API's on macOS, your tests must run inside a host application.

So, we need to be able to _actually_ add a dependency on RevenueCat and RevenueCatUI frameworks in a target in the Xcode project, and currently that's broken by these duplicated targets (or that's my hypothesis). 

### Description
The RevenueCat and RevenueCatUI targets defined in RevenueCat.xcodeproj duplicate the RevenueCat and RevenueCatUI targets defined in Package.swift, and when adding them to other targets in RevenueCat.xcodeproj, it causes build issues related to multiple sources for build items.

My hunch is that these are not actually actively being used any more. The two targets where they were being referenced (BackendIntegrationTests and StoreKitUnitTests) did not actually embed them. I could be wrong though, so part of the goal with this PR is to see the behavior of all the automated tests, and see if anything breaks.


clindsay3:
Just to keep this PR up to date with the discussion in Slack:

Mostly got this working, but the biggest hangup is Carthage support. Carthage is using those existing RevenuCat and RevenueCatUI targets. I tried building a little wrapper framework that imports and then re-exports the SwiftPM version, but Carthage wasn't really liking it. I think there's a path here – maybe by [distributing for carthage using prebuilt frameworks](https://github.com/Carthage/Carthage?tab=readme-ov-file#archive-prebuilt-frameworks-into-zip-files), but I feel like it's a rabbit hole and has the potential to affect customers in subtle ways. It's a shame, because I'm so close to just having one framework definition to rule them all!

Sounds like there may be some appetite for dropping Carthage support, so that's another possible route here.

PR #5316, opened 2025-06-23
Improve resolution of locale in paywalls components localizations

ajpallares:
This is closely related to #5309, as the solution in that PR was not correct.

### Checklist

### DIFF
--- a/RevenueCatUI/Templates/V2/Localizations/LocaleFinder.swift
+++ b/RevenueCatUI/Templates/V2/Localizations/LocaleFinder.swift
@@ -15,60 +15,49 @@ import Foundation
 
 extension Dictionary where Key == String {
 
+    /// Finds the best matching value for the provided locale with the restriction that the key
+    /// must match the language of the provided locale.
     func findLocale(_ locale: Locale) -> Value? {
-        let localeIdentifier = locale.identifier
+        let preferredIdentifiers = Self.preferredMatchedLocalesIdentifiers(from: Array(self.keys),
+                                                                           preferredLanguages: [locale.identifier])
 
-        if let exactMatch = self.valueForLocaleString(localeIdentifier) {
-            return exactMatch
-        }
-
-        // For matching unknown locales with full identifier
-        // Ex: `zh_CN` will become `zh_Hans_CN`
-        if let maxIdentifier = locale.rc_lanuageMaxIdentifier {
-            if let exactMatch = self.valueForLocaleString(maxIdentifier) {
-                return exactMatch
+        for localeIdentifier in preferredIdentifiers {
+            if let value = self[localeIdentifier] {
+                return value
             }
-        }
-
-        // For matching language and script without region
-        if let onlyLanguageAndScriptIdentifier = locale.rc_languageAndScript {
-            if let exactMatch = self.valueForLocaleString(onlyLanguageAndScriptIdentifier) {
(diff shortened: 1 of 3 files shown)


ajpallares:
@RCGitBot please test

RevenueCatUI/Templates/V2/Localizations/LocaleFinder.swift
```diff
@@ -15,60 +15,49 @@ import Foundation
...
+    ///
+    /// This method takes into account the order of `preferredLanguages` to determine the best match.
+    /// E.g.
+    /// if `preferredLanguages` is `["en-US", "es-ES"]`, then the identifier `en-GB`
+    /// will be preferred over `es-ES`.
+    ///
+    /// - important: As specified in the documentation of `Bundle.preferredLocalizations(from:forPreferences:)`
+    /// "_This method doesn’t return all localizations in order of user preference. To get this information,
+    /// you can call this method repeatedly, each time removing the identifiers returned by the previous call._"
+    /// This means that not all matches will be returned, but only the best ones based on `preferredLanguages`.
```
JayShortway: I'm a bit confused by these 2 statements: 
> [...] sorted by best match.  
  
and
> [...] doesn’t return all localizations in order of user preference.
  
They seem to contradict each other, but I'm probably misunderstanding.

Tests/RevenueCatUITests/PaywallsV2/LocaleFinderTests.swift
```diff
@@ -132,16 +132,40 @@ class LocaleFinderTest: TestCase {
...
         let foundLocalizations = localizations.findLocale(locale)
         expect(foundLocalizations).to(equal(Self.expectedTranslations))
     }
+
+    func test_es_MX_Matches_es_before_es_ES() {
+        let localizations = [
+            "en": ["wrong": "this is en"],
+            "es_ES": ["wrong": "this is es_ES"],
+            "es": Self.expectedTranslations
+        ]
```
JayShortway, line 153: Could we add another test to check that `es_MX` matches `es_ES` if that's the only available Spanish language?

JayShortway (approved):
So much simpler! Just a request for 1 more test.

RevenueCatUI/Templates/V2/Localizations/LocaleFinder.swift
ajpallares: You're right that it's confusing. I'll rephrase that. 
Anyway, I'm struggling a lot when trying to explain it because what the method `Bundle.preferredLocalizations(from:forPreferences:)` returns is also very confusing! 😅

ajpallares: Simplified implementation and documentation in https://github.com/RevenueCat/purchases-ios/pull/5316/commits/d761e7f3538282936a5dc4b585dfedb56bd96986

Tests/RevenueCatUITests/PaywallsV2/LocaleFinderTests.swift
ajpallares: Makes sense! Done in https://github.com/RevenueCat/purchases-ios/pull/5316/commits/be52a72bfeac246a0b90e749d7d8bfa305c8ad18

PR #5320, opened 2025-06-24
Move the Paywall-Screenshots test plan into the RevenueCatUITestsDev scheme, and add a host application to run them.

chrisvasselli:
### Motivation
This is in preparation for having these run on Mac Catalyst and Mac Native platforms, which will require a host application.

### Description
* Create a new RevenueCatUITestsDev scheme that builds the existing RevenueCatUITestsDev target.
* Move the Paywall-Screenshots test plan into the RevenueCatUITestsDev scheme.
* Add a UnitTestsHostApp as the host app for RevenueCatUITestsDev
* Add RevenueCat and RevenueCatUI as direct dependencies for RevenueCatUITestsDev target.
* Update the Fastfile to use the new RevenueCatUITestsDev scheme.


clindsay3:
Ran all-tests twice, and the failed jobs were different each time except for spm-revenuecat-ui-ios-15, which seems like it's failing for everyone. The failures don't seem related to my changes as far as I can tell.

clindsay3:
Actually, looks like all-tests doesn't trigger the actual paywall screenshot job. Let me make sure that's working.

clindsay3:
OK this is now looking good and the paywall validation screenshots are being updated properly. I have sanity checked that it looks like it's verifying the exact same set of images it was generating before.

.circleci/config.yml
```diff
@@ -1601,6 +1602,14 @@ workflows:
           context:
             - paywall-rendering-validation-pr-rw
 
+  record-paywall-screenshots:
+    when:
+      equal: ["record_paywall_screenshots", << pipeline.parameters.action >>]
+    jobs:
+      - record-and-push-paywall-template-screenshots:
+          context:
+            - paywall-rendering-validation-pr-rw
```
ajpallares: Question: couldn't just the `equal: ["record_paywall_screenshots", << pipeline.parameters.action` condition be added as an `OR` condition in `record-and-push-paywall-template-screenshots`? I ask because, if I'm not wrong, both workflows do the same, the only difference being the trigger conditions

ajpallares (approved):
Looks great! Just one small comment

.circleci/config.yml
clindsay3: Good call! Will make that change.

clindsay3:
@ajpallares @joshdholtz looks like I don't have permission to merge, but I think this is ready to merge now.

PR #5325, opened 2025-06-25
Dont be explicit on store for paddle, web, rcBilling, external

facumenzella:
### Motivation
We've been discussing not exposing these as different stores.

### Description
- Use storeWeb instead of each store key.


vegaro (approved):

tonidero:
I wouldn't consider this a `pr:feat`. Going to remake #5332 but changing labels to make it a patch.

PR #5322, opened 2025-06-24
Add Intro Eligibility Functions to CEC Mode

fire-at-will:
### Description
This PR exposes the following functions to the SDK's Custom Entitlements Computation Mode:
```swift
func checkTrialOrIntroDiscountEligibility(productIdentifiers: [String], completion: @escaping ([String: IntroEligibility]) -> Void)
func checkTrialOrIntroDiscountEligibility(productIdentifiers: [String]) async -> [String: IntroEligibility]

func checkTrialOrIntroDiscountEligibility(packages: [Package]) async -> [Package: IntroEligibility]

func checkTrialOrIntroDiscountEligibility(product: StoreProduct, completion: @escaping (IntroEligibilityStatus) -> Void)
func checkTrialOrIntroDiscountEligibility(product: StoreProduct) async -> IntroEligibilityStatus
```

It also adds those functions to the CEC API tester, and adds a button to the CEC example app to test the functionality.

These changes should be source compatible and should not cause breaking changes.

PR #5328, opened 2025-06-25
[v4] Add Intro Eligibility Functions to CEC Mode

JayShortway:
## Description
Manual backport of https://github.com/RevenueCat/purchases-ios/pull/5322 to v4.

### DIFF
--- a/Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
+++ b/Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation/ContentView.swift
@@ -89,6 +89,19 @@ struct ContentView: View {
             .cornerRadius(20)
             .padding()
 
+            Button("Fetch product's intro/trial eligibility") {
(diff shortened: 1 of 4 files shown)


ajpallares (approved):
👍 thank you!

JayShortway:
What do we think about the failing `run-test-ios-14` job? Seems to need some extensive changes to even run. Mint complains with:
>Error: You are using macOS 12.
>We (and Apple) do not provide support for this old version.

ajpallares:
> What do we think about the failing `run-test-ios-14` job? Seems to need some extensive changes to even run. Mint complains with:
> 
> > Error: You are using macOS 12.
> > We (and Apple) do not provide support for this old version.

Maybe we can do the same as here? https://github.com/RevenueCat/purchases-ios/pull/5256

JayShortway:
@ajpallares Thank you! Cherry-picked that one. Let's see 🤞

PR #5334, opened 2025-06-25
[Virtual Currency] Initial support for Virtual Currency in Paywalls

MarkVillacampa:
### Checklist


RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift
```diff
@@ -306,6 +332,24 @@ extension VariablesV2 {
...
+            return self.productGrantAllAmountCodePerPeriodAbbreviated(package: package, localizations: localizations)
+        case .productGrantsAllAmountNamePerPeriodAbbreviated:
+            return self.productGrantAllAmountNamePerPeriodAbbreviated(package: package,
+                                                                      localizations: localizations,
+                                                                      virtualCurrencies: virtualCurrencies)
+        case .productGrantsAmount:
+            return self.productGrantAmount(package: package, virtualCurrencyCode: variableSubscript ?? "")
+        case .productGrantsName:
+            return self.productGrantName(package: package,
+                                         virtualCurrencyCode: variableSubscript ?? "",
```
fire-at-will, line 351: Nit: although the results would be the same, I might consider only calling `productGrantName` & `productGrantAmount` here if `variableSubscript` is non-nil to avoid the case where we try to look up a virtual currency that doesn't exist.

It doesn't really matter though since `productGrantAmount()` returns an empty string when the VC isn't found, so feel free to ignore 👍

RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift
```diff
@@ -702,6 +746,70 @@ private extension VariablesV2 {
...
+    }
+
+    func productGrantAllAmountCodePerPeriod(package: Package, localizations: [String: String]) -> String {
+        let periodAbbreviated = self.productPeriod(package: package, localizations: localizations)
+
+        var currencies: [String] = []
+        package.virtualCurrencyGrants.forEach { key, value in
+            currencies.append("\(value.amount) \(key)/\(periodAbbreviated)")
+        }
+        return currencies.joined(separator: " ")
```
fire-at-will, line 774: Instead of using `currencies.joined()`, would it be worth using [ListFormatter.localizedString()](https://developer.apple.com/documentation/foundation/listformatter) to try to get a localized list string here?

That way, instead of having "... ... ...", you could have "..., ..., and ..." in English, and the appropriate localized list for each other locale.

RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift
```diff
@@ -702,6 +746,70 @@ private extension VariablesV2 {
...
+    }
+
+    func productGrantAllAmountCodePerPeriodAbbreviated(package: Package, localizations: [String: String]) -> String {
+        let periodAbbreviated = self.productPeriodAbbreviated(package: package, localizations: localizations)
+
+        var currencies: [String] = []
+        package.virtualCurrencyGrants.forEach { key, value in
+            currencies.append("\(value.amount) \(key)/\(periodAbbreviated)")
+        }
+        return currencies.joined(separator: " ")
```
fire-at-will, line 784: Same question about `ListFormatter.localizedString()` here

RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift
```diff
@@ -702,6 +746,70 @@ private extension VariablesV2 {
...
+                                            localizations: [String: String],
+                                            virtualCurrencies: [String: VirtualCurrencyMetadata]) -> String {
+        let periodAbbreviated = self.productPeriod(package: package, localizations: localizations)
+
+        var currencies: [String] = []
+        package.virtualCurrencyGrants.forEach { key, value in
+            let virtualCurrencyName = virtualCurrencies[key]?.name ?? key
+            currencies.append("\(value.amount) \(virtualCurrencyName)/\(periodAbbreviated)")
+        }
+        return currencies.joined(separator: " ")
```
fire-at-will, line 797: Same question about `ListFormatter.localizedString()` here

RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift
```diff
@@ -702,6 +746,70 @@ private extension VariablesV2 {
...
+                                                       localizations: [String: String],
+                                                       virtualCurrencies: [String: VirtualCurrencyMetadata]) -> String {
+        let periodAbbreviated = self.productPeriodAbbreviated(package: package, localizations: localizations)
+
+        var currencies: [String] = []
+        package.virtualCurrencyGrants.forEach { key, value in
+            let virtualCurrencyName = virtualCurrencies[key]?.name ?? key
+            currencies.append("\(value.amount) \(virtualCurrencyName)/\(periodAbbreviated)")
+        }
+        return currencies.joined(separator: " ")
```
fire-at-will, line 810: Same question about `ListFormatter.localizedString()` here

Sources/Purchasing/VirtualCurrencyMetadata.swift
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * The name of the virtual currency.
+     */
+    @objc public let name: String
+
+    /**
+     * The description of the virtual currency.
+     */
+    @objc public let currencyDescription: String
```
fire-at-will, line 29: Since this class is going to be public, we might want to consider renaming this to `serverDescription` to match the naming convention used in `VirtualCurrency` from the balance fetching PR

PR #5344, opened 2025-06-30
Fix some tests snapshots after

ajpallares:
Fixes some tests failing after merging #5288 

* In `BackendSubscriberAttributesTests`, now the preferred locale is injected in the tests to be `en_US`, instead of taking it from the system's locale. In iOS 13 and 14, the system locale was being reported to `en` (which is different to `en_US`, thus making the JSON snapshots different)
* Some tests were failing due to JSON sorting. I sorted the JSON keys to make them equal to how they are expected to result in CI.
  * TIL: `JSONEncoder.OutputFormatting.sortedKeys` is locale dependent  which means the keys in the exported JSON can be printed in different order. We should probably fix this (ideally we would use https://github.com/swiftlang/swift-corelibs-foundation/issues/3676 but that does not seem to happen anytime soon).


ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

facumenzella (approved):

PR #5347, opened 2025-07-01
Use VCs in CustomerCenterViewModel

fire-at-will:
### Description
After a discussion in the https://github.com/RevenueCat/purchases-ios/pull/5108's review, we decided to revert the change of making the view models in `RelevantPurchasesListView` and `SubscriptionDetailView` from `StateObject` to `ObservedObject`. This PR accomplishes this by using the `VirtualCurrencies` object in the `CustomerCenterViewModel` in each of those views instead of injecting it directly in the views' constructors.

**Changes:**
- Makes `RelevantPurchasesListView.viewModel` a `@StateObject` again
- Makes `SubscriptionDetailView.viewModel` a `@StateObject` again
- In both `RelevantPurchasesListView` and `SubscriptionDetailView`, reference the virtual currencies in the `CustomerCenterViewModel`

### Testing
- Updated automated tests
- Manually tested each screen that the virtual currencies are displayed on


facumenzella (approved):
Thanks for taking care of this 🫶 !!

PR #5349, opened 2025-07-03
Fix google products display for customer center

facumenzella:
### Motivation
We got a report where some google products were not being shown in the customer center list.

### Description
- Ignore base plan to grab the subscription
- Add test to cover the issue

### Caveats
I'll go through the existing tests one by one because we cannot trust them anymore 😿

### DIFF
--- a/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
@@ -289,16 +289,21 @@ private extension CustomerCenterViewModel {
 
     func loadSubscriptionsSection(customerInfo: CustomerInfo) async {
         var activeSubscriptionPurchases: [PurchaseInformation] = []
-        for subscription in customerInfo.activeSubscriptions
-            .compactMap({ id in customerInfo.subscriptionsByProductIdentifier[id] })
+        let subscriptions = customerInfo.activeSubscriptions
+            .compactMap({ id in
+                // Do the opposite as CustomerInfo.extractProductIDAndBasePlan for non-apple products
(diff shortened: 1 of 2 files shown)


RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -289,16 +289,21 @@ private extension CustomerCenterViewModel {
 
     func loadSubscriptionsSection(customerInfo: CustomerInfo) async {
         var activeSubscriptionPurchases: [PurchaseInformation] = []
-        for subscription in customerInfo.activeSubscriptions
-            .compactMap({ id in customerInfo.subscriptionsByProductIdentifier[id] })
+        let subscriptions = customerInfo.activeSubscriptions
+            .compactMap({ id in
+                // Do the opposite as CustomerInfo.extractProductIDAndBasePlan for non-apple products
+                let idWithoutBasePlan = id.split(separator: ":").first.map { id in String(id) } ?? id
```
ajpallares, line 295: This doesn't affect any other store's product identifier, right?

ajpallares (approved):
Makes sense! Just a question

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
vegaro: we could filter for play store only to be on the safe side

facumenzella: I've added a fallback to the old way just to be on the safe side https://github.com/RevenueCat/purchases-ios/pull/5349/commits/97d6e39463986a682fcdb00bdfad1a6beb370432

Thanks @ajpallares @vegaro

PR #5351, opened 2025-07-03
Upload screenshots to Emerge in addition to pushing them to the paywall validation repo.

clindsay3:
### Motivation
This will give us a nice way to validate changes to the paywall validation screenshots using Emerge and quickly identify regressions. We are [already doing this for Android](https://www.emergetools.com/snapshot/bbf76110-14b0-4cf6-a097-2ba0688c7c9b).

### Description
After generating the screenshots and before pushing them to the paywall rendering validation repo, take the folder and upload it to Emerge using the emerge command line tool.

PR #5352, opened 2025-07-03
Use a more accurate method for generating a screenshot of a UIView

clindsay3:
### Motivation
Before this change, buttons (and possibly other controls) were simply not rendered in these tests on Mac Catalyst builds because they are not backed by CALayers.

### Description
Switch from CALayer.render(in:) to UIView.drawHierarchy(in:afterScreenUpdates:) for rendering paywall screenshots. This change should capture the screen render more accurately.

I'll wait until after https://github.com/RevenueCat/purchases-ios/pull/5351 is merged to merge this PR, so that we can get a clear picture of any potential iOS rendering differences that may have been introduced by this change.


joshdholtz (approved):
TIL 😅 Thank yoooouuuu!!!

clindsay3:
TIL too! Was very surprised when the screenshots were missing buttons haha.

PR #5353, opened 2025-07-03
Temporary revert of automatic health reporting in the SDK

polpielladev:
We have some failing BackEnd integration tests that are failing after we added this.

We also noticed some signature verification issues, so in the meantime let's revert this so that we can release.

### Checklist


polpielladev:
@RCGitBot please test

PR #5356, opened 2025-07-03
Fix one flaky unit test

ajpallares:
Test `testCustomerInfoIsOnlyFetchedOnceOnAppLaunch()` was flaky because this line
```swift
// 1. Make sure any existing customer info requests finish
_ = try? await purchases.customerInfo(fetchPolicy: .fromCacheOnly)
```
did not really cause the test to wait for the existing customer info requests finish. The reason is that the code path of `fetchPolicy: .fromCacheOnly` does not consider any existing requests; it just returns the customer info from the cache.

Instead, the test now does the following:
```swift
var customerInfoIterator = try? purchases.customerInfoStream.makeAsyncIterator()
_ = await customerInfoIterator?.next()
```
which makes sure that execution is only continued once a `CustomerInfo` valued is received (i.e. after the existing customer info request finishes)


ajpallares:
@RCGitBot please test

polpielladev (approved):
LGTM

PR #5360, opened 2025-07-04
[DX-457] Re-apply Health SDK logging on app launch

polpielladev:
We recently had to revert the changes as we noticed some signature verification warnings and some increased flakiness in the Backend Integration tests.

These issues have now been addressed:
- Changed the signature verification from the `health_report_availability` endpoint: https://github.com/RevenueCat/khepri/pull/13783
- The unit test flakiness was not caused by the Health SDK, but they were aggravated by introducing more asynchronous behaviour. @ajpallares recently fixed the flakiness for the existing unit test: https://github.com/RevenueCat/purchases-ios/pull/5356

### Checklist

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -274,6 +274,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
     fileprivate let systemInfo: SystemInfo
     private let storeMessagesHelper: StoreMessagesHelperType?
     private var customerInfoObservationDisposable: (() -> Void)?
+    private let healthManager: SDKHealthManager
 
     private let syncAttributesAndOfferingsIfNeededRateLimiter = RateLimiter(maxCalls: 5, period: 60)
     private let diagnosticsTracker: DiagnosticsTrackerType?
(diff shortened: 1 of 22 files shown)


polpielladev:
@RCGitBot please test

polpielladev:
@RCGitBot please test

polpielladev:
@RCGitBot please test

polpielladev:
@RCGitBot please test

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -2065,6 +2077,31 @@ private extension Purchases {
         }
     }
 
+    #if DEBUG && !ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION
+    func runHealthCheckIfInForeground() {
+        // This is a workaround and needs to be fixed at some point. Explanation:
```
polpielladev, line 2082: @ajpallares I ended up having to add a workaround to this so that it does not run with the integration tests. Unfortunately the StoreKit tests rely too much on time and adding an extra requests makes products expire while requests are in flight. We really need to make the StoreKit integration tests more resilient, but this is a workaround to not make them more flaky for the time being.

polpielladev: Let me know if you are okay with this, I have tried taking a stab at fixing the StoreKit tests flakiness and I got pretty far but wow they are incredibly fragile 😭

ajpallares: Ohhh I see. I didn't expect this having any effect on StoreKit integration tests but I totally understand. In fact, this probably gives us a clue on what could be making StoreKit integration tests be more flaky.

I think it makes sense for now, as it's only disabled for integration tests, and not for all the tests. So I'm ok with this 👍

PR #5364, opened 2025-07-07
Add build configurations for tuist workspace

facumenzella:
### Motivation
While testing the maestro app I realized the xcconfig files were not being picked.

### Description
- Add build configurations to project
- Extract configurations to be reused
- Reuse workspace xcconfig files for maestro app
- Use recommended project settings for those projects
- Add abstractions for target settings, project settings
- Remove run action from RevenueCat
- Add xctestplans to Workspace


Projects/PaywallTester/Project.swift
```diff
@@ -49,6 +46,9 @@ let project = Project(
             sources: [
                 "../../Tests/TestingApps/PaywallsTester/PaywallsTester/**/*.swift"
             ],
+            resources: [
+                "../../Local.xcconfig"
```
ajpallares: is this correct? what if the file doesn't exist?

Projects/PaywallTester/Project.swift
```diff
@@ -23,7 +23,7 @@ let allDeploymentTargets: DeploymentTargets = .multiplatform(
 var additionalFiles: [FileElement] = [
     .glob(pattern: "../../Global.xcconfig")
 ]
-if FileManager.default.fileExists(atPath: "../../Local.xcconfig") {
+if FileManager.default.fileExists(atPath: "Local.xcconfig") {
     additionalFiles.append(.glob(pattern: "../../Local.xcconfig"))
 }
```
ajpallares: Maybe not for this PR, but would it be possible/make sense to have this be reused from the main `Project.swift` instead of repeated in every `Project.swift` file?
Maybe something similar to `xcconfigFileConfigurations` below. E.g. `commonFiles`.

ajpallares (commented):
Looks good but I have one question before approving. Would it work without the `Local.xcconfig` file being present?

Projects/PaywallTester/Project.swift
facumenzella: You're right, this should be conditional 👍

Projects/PaywallTester/Project.swift
facumenzella: I think so, let me re-do this

Projects/PaywallTester/Project.swift
facumenzella: https://github.com/RevenueCat/purchases-ios/pull/5364/commits/b88ba1e7012662804d9c83b87d4269f6cb886aac

Projects/PaywallTester/Project.swift
facumenzella: I'll tackle this in another PR, but I think that since we're adding those files to the workspace it should be ok if we remove them from each project.

Projects/PaywallTester/Project.swift
```diff
@@ -49,6 +47,9 @@ let project = Project(
             sources: [
                 "../../Tests/TestingApps/PaywallsTester/PaywallsTester/**/*.swift"
             ],
+            resources: [
+               additionalFiles.count == 2 ? "../../Local.xcconfig" : ""
```
ajpallares: This looks dangerous. And I don't really understand why this is needed? I don't think xcconfig files need to be added to the targets

facumenzella: I've removed `additionalFiles`from the project definition. Now those are declared in the workspace.

Examples/PurchaseTester/Project.swift
```diff
@@ -30,7 +30,11 @@ let allDestinations = destinations + [.appleWatch]
 let project = Project(
     name: "PurchaseTester",
     organizationName: .revenueCatOrgName,
-    settings: .settings(base: [:].automaticCodeSigning(devTeam: .revenueCatTeamID)),
+    settings: .settings(
+        base: [:].automaticCodeSigning(devTeam: .revenueCatTeamID),
+        configurations: .xcconfigFileConfigurations,
+        defaultSettings: .essential
+    ),
```
facumenzella: this can probably be generalized even further (for another PR)

facumenzella:
@RCGitBot please test

Tuist/ProjectDescriptionHelpers/Configurations.swift
```diff
@@ -0,0 +1,12 @@
+import ProjectDescription
+
+extension Array<Configuration> {
+
+    /// Debug and Release configurations that use the Global.xcconfig file
+    public static var xcconfigFileConfigurations: [Configuration] {
+        [
+            .debug(name: "Debug", xcconfig: .relativeToRoot("Global.xcconfig")),
+            .release(name: "Release", xcconfig: .relativeToRoot("Global.xcconfig")),
```
facumenzella: I am not sure about this, I think it shouldn't point to the `Global.xcconfig` right @ajpallares ?

facumenzella:
@RCGitBot please test

Examples/rc-maestro/Project.swift
```diff
@@ -31,7 +19,9 @@ let project = Project(
                         "UIColorName": "",
                         "UIImageName": "",
                     ],
-                    "REVENUECAT_API_KEY": "$(REVENUECAT_API_KEY)"
+                    "REVENUECAT_API_KEY": "$(REVENUECAT_API_KEY)",
+                    "REVENUECAT_PROXY_URL_SCHEME": "$(REVENUECAT_PROXY_URL_SCHEME)",
+                    "REVENUECAT_PROXY_URL_HOST": "$(REVENUECAT_PROXY_URL_HOST)"
```
ajpallares, line 24: Question: wouldn't this be covered by the xcconfig?

ajpallares (approved):
Great job! We keep iterating on this 💪

Examples/rc-maestro/Project.swift
facumenzella: mmm good question, I'll try it out

facumenzella: Just tested it out, and we need it. Otherwise it's not picked by the bundle

PR #5365, opened 2025-07-07
Fixes API tests after changes to public API

polpielladev:
@fire-at-will realized that we made some breaking changes to the public API by mistake when adding health report logging.

The reason was that we had already shipped the new `PurchasesDiagnostics` APIs just under the `#if DEBUG` directive, so we ca not make it more restrictive now without breaking the public API.

The rest of changes (logging the health report on configure) remain under `DEBUG` and `CEC`.

This PR fixes these issues while keeping the new logging changes out of the CEC mode.

### Checklist


fire-at-will:
@RCGitBot please test

PR #5367, opened 2025-07-07
Add a few additional VC integration tests

fire-at-will:
### Description
Adds a few additional integration tests for virtual currencies. Specifically, we added tests to ensure that:
- Newly created users can access all virtual currencies + metadata, despite having a balance of 0 for all VCs
- The `cachedVirtualCurrencies` computed property + the `invalidateVirtualCurrenciesCache()` function work as expected


fire-at-will:
@RCGitBot please test

MarkVillacampa (approved):

PR #5370, opened 2025-07-09
Add an app to run paywall validation tests locally

clindsay3:
### Motivation
I found myself wanting to be able to run individual paywalls from the paywall validation set, interact with them, look at specific ones based on offering ID, and set breakpoints in them. There wasn't an easy way to do this, since previously they were just generating screenshots.

### Description
* Refactors the logic for loading paywalls from `paywall-validation-resources` out of the `TakeScreenshot` test, and into a separate file `PaywallPreviewResourcesLoader`.
* Create a new target `PaywallValidationTester`, that uses that shared logic to display the paywalls and has a Swift preview where you can specify an individual paywall to display.

I initially planned to add this functionality to the existing PaywallsTester app, rather than create a new separate app. But I found that the PaywallsTester app redefines a lot of the internal RevenueCat structures (`OfferingsResponse`, for example), and so it would have required some significant refactoring to fit it into that app, which seemed like more trouble than it was worth, and would if anything be a less exact representation of the behavior of these paywalls in a live scenario.

*Note: When reviewing this PR, it might be easier to look at the individual commits separately. The initial "refactor" commit should cause no change in behavior (and seemed to when I tested locally), and is the bulk of the lines of code changed in this PR. The rest is setting up the new app target that leverages the same logic.*


clindsay3:
Looks like adding the new target has caused Xcode 15.2 to be unable to read the project file. Working on it...

facumenzella:
@clindsay3 will this be for local development only? If that's the case, I think it would make a lot of sense to put it in the `tuist` project.

clindsay3:
@facumenzella I have migrated this target to be defined via Tuist instead of as a new target in the Xcode project!

Projects/PaywallValidationTester/Project.swift
```diff
@@ -0,0 +1,63 @@
...
+)
+
+let project = Project(
+    name: "PaywallValidationTester",
+    organizationName: .revenueCatOrgName,
+    settings: .settings(
+        base: [:].automaticCodeSigning(devTeam: .revenueCatTeamID),
+        configurations: .xcconfigFileConfigurations,
+        defaultSettings: .recommended
+    ),
```
ajpallares: We have the `.appProject` constant defined here https://github.com/RevenueCat/purchases-ios/blob/01c00c4ef86a91bda1f6460c6babe3d92d3091c7/Tuist/ProjectDescriptionHelpers/Settings.swift#L12-L18
```suggestion
    settings: .appProject,
```
(you need to merge `main` into your branch first)

Tests/TestingApps/PaywallValidationTester/PaywallValidationTesterView.swift
```diff
@@ -0,0 +1,68 @@
...
+@testable import RevenueCatUI
+
+struct PaywallValidationTesterView: View {
+    var offeringId: String? = ProcessInfo.processInfo.environment["OFFERING_ID"]
+    @State var offerings:[Offering] = []
+
+    var body: some View {
+        TabView {
+            ForEach(offerings, id: \.self) { offering in
+                if offering.id == offeringId || offeringId == nil {
```
ajpallares: I've tested running the app setting `offeringId` to `nil`. The UX is a bit confusing, since in the middle of the horizontal scrolling to see the next paywall, the UI flashes and sometimes it results in having scrolled two paywalls at once. Not a big deal, but it would be nice to fix.

In case it helps, it seems that all paywalls get re-rendered by SwiftUI when scrolling the `TabView`

clindsay3: I dug into this, and I think it's some weird SwiftUI quirkiness around TabViews and nested ForEach. I tried commenting out different pieces to figure out what's causing that flashing is it scrolls to the next page, and I isolated it to, bizarrely, the ForEach in `ComponentsView.layoutComponents`. If I replace the contents of that foreach with a simple Text, I still get the flickering. If I remove the ForEach and have just a single Text block, it doesn't happen.

There may be some other complicated interaction going on, but it's easier to just switch to using a ScrollView and LazyHStack, which is what I've done in the latest commit..

ajpallares: Ohh nice investigation! I would have never imagined that the culprit of this would be deeper 🤯 
Thanks for taking a look and thank you for fixing it! I think the `ScrollView` + `LazyHStack` is a great idea! And it makes it more efficient, as before all the paywalls were rendering at once and every time scrolling happened

PR #5371, opened 2025-07-09
Add Mac Catalyst and iPad/iPhone app on Mac Paywall Validation Screenshot Generation

clindsay3:
### Motivation
Adds validation screenshot generation for our newly supported Mac platforms, so we can have a clear path towards parity with other platforms, and then avoid regressions in the future.

I'm not sure if having two separate "iPhone App running on Mac" and "iPad App running on Mac" variations is truly necessary, but it felt like it couldn't hurt to include them separately. If it becomes burdensome in the future and we find that in practice there are no actual differences between the two, we could remove one of them in the future.

### Description
Updates our paywall validation screenshot generation so that it executes in 4 new contexts:

1. iPhone App running on Mac
2. iPad App running on Mac
3. Mac Catalyst with "Optimize for Mac" mode enabled
4. Mac Catalyst with "Scaled to Match iPad" mode enabled


clindsay3:
Failing iOS 17 tests seem unrelated to this change.

fastlane/Fastfile
```diff
@@ -178,43 +178,66 @@ platform :ios do
...
+      )
+    end
+
+    # Commit, push, and create a PR in the target repository
+    Dir.chdir(target_repository_path) do
+      commit_push_and_create_pr_if_necessary(
+        repo: target_repository_name,
+        branch_name: target_repository_branch,
+        commit_message: "[Automated] Updates template screenshots for all platforms (commit: #{this_commit_hash})",
+        title: "[AUTOMATIC] Updates template screenshots for all platforms",
```
JayShortway: Small suggestion: 
```suggestion
        commit_message: "[Automated] Updates template screenshots for all Apple platforms (commit: #{this_commit_hash})",
        title: "[AUTOMATIC] Updates template screenshots for all Apple platforms",
```

ajpallares: Came here to suggest exactly this ⬆️

clindsay3: Good call! Look at my Apple-centric brain, haha

fastlane/Fastfile
```diff
@@ -178,43 +178,66 @@ platform :ios do
     end
     UI.user_error!("Failed to determine target repository name") if target_repository_name.nil? || target_repository_name.empty?
 
-    # Loop over all PNG files in the tmp_dir, parse the offering ID from the file names, and move them into the target repository
-    output_dir = record_paywall_screenshots
+    # Loop over all platforms and generate screenshots for each
+    platforms = ["ios", 
+                # These platforms work locally but we have been unable to figure out how to get them to work on CI due to codesigning issues.
+                #  "iphone-app-on-mac",
+                #  "ipad-app-on-mac",
```
ajpallares, line 185: Are these supposed to be commented out? I ask because the PR title says that it also adds iPad/iPhone app on Mac validation

fastlane/Fastfile
```diff
@@ -325,20 +348,55 @@ platform :ios do
...
+      "mac-catalyst-optimized-for-mac" => "platform=macOS,arch=arm64,variant=Mac Catalyst",
+      "mac-catalyst-scaled-to-match-ipad" => "platform=macOS,arch=arm64,variant=Mac Catalyst"
+    }
+    
+    # Map platform to TARGETED_DEVICE_FAMILY build setting. This overrides the setting in the Xcode Project,
+    # if we ever change the list of supported platforms we'll need to make sure to update this as well.
+    targeted_device_families = {
+      "mac-catalyst-optimized-for-mac" => "1,2,3,4,6",
+      "mac-catalyst-scaled-to-match-ipad" => "1,2,3,4"
+    }
```
ajpallares, line 367: I'm seeing [here](https://developer.apple.com/documentation/xcode/build-settings-reference) that `3` is for Apple TV and `4` is for Apple watch. Why are these included in the targeted device families?

ajpallares (commented):
I think it looks good overall! I have a couple of questions though

fastlane/Fastfile
clindsay3: They work locally, but we have not been able to figure out how to get them to run on CI yet because of codesigning issues(@joshdholtz spent a bunch of time on this and wasn't able to get it working). I still think it's useful to include the logic in the commit, because I can run them locally to verify those platforms. But they are commented out so that the CI build doesn't fail on them.

fastlane/Fastfile
clindsay3: I included those platforms because they were included in the target before this change, and so I was trying not to mess with anything that was already in there.

Ideally the logic here would be to just add/remove that 6 from the list of targets that was included in the build settings already. But it seemed like a much bigger change to read the build settings here and modify them, than to just pass in a hard-coded override. So I felt like it was a better approach to simply include the full list as it was before this change.

clindsay3:
@ajpallares whoops I had set it to auto-merge when all the checks passed, so when I approved the Emerge screenshot diff it merged! But feel free to leave any requests for changes and I'll open a follow-up PR!

ajpallares:
@clindsay3 No problem! I was going to approve anyway after reading the comments. So it's all good! Thanks for the explanations!

PR #5376, opened 2025-07-10
When taking screenshots for validation, ignore safe area.

clindsay3:
### Motivation
This fixes an issue where screenshots were being cutoff on the bottom on Mac Catalyst, I believe because the view was being pushed down to insert a window titlebar.

### Description
* When taking screenshots for validation, ignore safe area.
* Bump minimum deployment target for RevenueCatUITestsDev target to iOS 14, since that's the minimum required for ignoresSafeArea(). Presuming this won't be an issue for this target, but if any jobs fail I'll come up with another solution.

We should be careful to separately verify that the presentation of paywalls is sized correctly on different platforms, but I think we can't verify that through these screenshots, we need to use the actual mechanisms for presenting paywalls in sheets, etc.


joshdholtz (approved):
Smart! 😁 :shipit:

PR #5377, opened 2025-07-10
Add a mechanism to test the presentIfNeeded API in the Paywalls tester.

clindsay3:
### Motivation
I want to test the behavior of presenting a paywall via this API vs. using PaywallView directly, for the purposes of implementing paywalls on Mac.

### Description
* Adds a new option to the context menu on items in the Sandbox Paywalls tab to present the paywall using the `presentIfNeeded` API.


Tests/TestingApps/PaywallsTester/PaywallsTester/UI/Views/OfferingList/APIKeyDashboardList.swift
```diff
@@ -200,6 +203,15 @@ struct APIKeyDashboardList: View {
                     }
                 }
         }
+        .overlay {
+            if offeringToPresent != nil {
+                VStack {}.presentPaywallIfNeeded(offering: offeringToPresent,
```
clindsay3: Is there a better way to get a paywall to conditionally appear? I was trying to get something working similar to the `sheet` modifier that would make the paywall appear whenever the offering was not nil, but couldn't come up with anything. This feels convoluted.

ajpallares: Found myself with the same question a couple of days ago. I think this API is meant to be as simple as possible make the "when" to show the paywall pretty static (i.e. is not meant to be used with a Binding as you need here).
That said, I totally agree that this PR is totally needed and the `presentPaywallIfNeeded` API should be added as an option to Paywall Tester.
I think a cleaner solution that could probably work would be to create a custom view modifier that depends on the `$offeringToPresent` binding, that does this
```
if let offeringToPresent {
    content.presentPaywallIfNeeded(offering: offeringToPresent,
                                   shouldDisplay: { _ in true },
                                   onDismiss: { offeringToPresent = nil })
} else {
    content
}
```

Then, you only need to use it like this:

```
.presentPaywallIfNeededModifier(offering: $offeringToPresent)
```
Lmk what you think!

clindsay3: That's a lot nicer, I'll make this change!

clindsay3: @ajpallares Done. Feels like something along the lines of this API might be a nice thing to expose in the public API to be honest!

ajpallares: Yep, that's right. I guess developers with this need use the `PaywallView` API but still I think it may be worth to have a `presentPaywall` API that receives a binding

ajpallares (approved):
Nice addition 👌 Thank you!

clindsay3:
There were a few emerge snapshots unrelated to any of this work that had diffs -- seems just like flakiness, so I approved.

PR #5378, opened 2025-07-10
Update default height of paywalls when using .presentIfNeeded on Mac Catalyst to something that is more reasonable.

clindsay3:
### Motivation
On Mac Catalyst, the default sheet size is very square-shaped, and looks terrible with our paywalls. It is significantly shorter than on iPad. 

### Description
* On Mac Catalyst, update the default height of paywalls presented using the .presentIfNeeded API to 667 points.


ajpallares (approved):
Yep! I think this makes sense 👌

PR #5381, opened 2025-07-15
Use SwiftUI instead of UIKit to present an alert in Paywall Tester app

clindsay3:
### Motivation
The existing mechanism for presenting an alert will not work on macOS.

### Description
Switch from using UIKit to present an alert, to using SwiftUI.

Follow-up on https://github.com/RevenueCat/purchases-ios/pull/5238

I haven't actually been able to test this, because I cannot figure out how to get this deep link testing code to activate. I tried creating a deep link button in the Paywalls web interface, and it didn't seem to like that. @joshdholtz any advice on how to test this?

Here's the error I got:

```
Failed to open URL myscheme://deeplinktest: Error Domain=NSOSStatusErrorDomain Code=-10814 "(null)" UserInfo={_LSLine=279, _LSFunction=-[_LSDOpenClient openURL:fileHandle:options:completionHandler:]}
ERROR: Failed to open URL as deep link: myscheme://deeplinktest
```

### DIFF
--- a/Tests/TestingApps/PaywallsTester/PaywallsTester/PaywallsPreview.swift
+++ b/Tests/TestingApps/PaywallsTester/PaywallsTester/PaywallsPreview.swift
@@ -35,6 +35,12 @@ struct PaywallsPreview: App {
     @State
     private var shouldShowWebPurchaseRedemptionResultAlert: Bool = false
 
+    @State
+    private var deepLinkAlertMessage: String?
+
+    @State
+    private var shouldShowDeepLinkAlert: Bool = false
+
     var body: some Scene {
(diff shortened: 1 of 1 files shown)


ajpallares (approved):
I think it makes sense! 

As it only affects the Paywall tester app, could you update the PR's title to make it clear? (e.g. "Use SwiftUI to present an alert in Paywall Tester")

clindsay3:
I tested triggering this alert manually just to sanity test it's displaying properly, and it is. Since the logic around triggering it is unchanged, even though I couldn't actually trigger it I'm going to go ahead and merge.

PR #5382, opened 2025-07-15
Increase simulated renewal time in some StoreKit integration tests

ajpallares:
### Motivation
We have [some flaky StoreKit integration tests](https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/29143/workflows/d1930d9e-1e13-475d-8341-32a9a0857552/jobs/355063/tests). This PR tries to make them less flaky.

### Description
Some unit tests were randomly failing because the receipt posting would happen after the purchase expiration, which would cause the product entitlement become inactive.

This PR changes the `SKTestSession`'s `timeRate` of some tests from `.oneRenewalEveryTwoSeconds.` to `.oneSecondIsOneDay`. With this change, the first renewal time of the affected tests changes from 2 seconds to 3 seconds (because these tests simulate the purchase of a product with a 3-day trial).

### DIFF
--- a/Tests/StoreKitUnitTests/TestHelpers/StoreKitTestHelpers.swift
+++ b/Tests/StoreKitUnitTests/TestHelpers/StoreKitTestHelpers.swift
@@ -36,6 +36,18 @@ extension XCTestCase {
         }
     }
 
+    // Some tests were randomly failing on CI when using `.oneRenewalEveryTwoSeconds` due to a race condition where the
+    // purchase would expire before the receipt was posted.
+    // This time rate is used to work around that issue by having a longer time rate.
+    func setOnSecondIsOneDayTimeRate(_ testSession: SKTestSession) {
+        // Using rawValue: 6 because the compiler shows this warning for `.oneSecondIsOneDay`:
+        // 'oneSecondIsOneDay' was deprecated in iOS 15.2: renamed to
+        // 'SKTestSession.TimeRate.monthlyRenewalEveryThirtySeconds'
+        // However, we've found that their behavior is not equivalent since using `monthlyRenewalEveryThirtySeconds`
+        // results in a crash in our tests.
+        testSession.timeRate = .init(rawValue: 6)! // == .oneSecondIsOneDay
+    }
+
     func verifyNoUnfinishedTransactions(file: FileString = #filePath, line: UInt = #line) async {
         let unfinished = await StoreKit.Transaction.unfinished.extractValues()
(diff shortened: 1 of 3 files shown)


ajpallares:
@RCGitBot please test

Tests/StoreKitUnitTests/TestHelpers/StoreKitTestHelpers.swift
```diff
@@ -36,6 +36,18 @@ extension XCTestCase {
...
+    // Some tests were randomly failing on CI when using `.oneRenewalEveryTwoSeconds` due to a race condition where the
+    // purchase would expire before the receipt was posted.
+    // This time rate is used to work around that issue by having a longer time rate.
+    func setOnSecondIsOneDayTimeRate(_ testSession: SKTestSession) {
+        // Using rawValue: 6 because the compiler shows this warning for `.oneSecondIsOneDay`:
+        // 'oneSecondIsOneDay' was deprecated in iOS 15.2: renamed to
+        // 'SKTestSession.TimeRate.monthlyRenewalEveryThirtySeconds'
+        // However, we've found that their behavior is not equivalent since using `monthlyRenewalEveryThirtySeconds`
+        // results in a crash in our tests.
+        testSession.timeRate = .init(rawValue: 6)! // == .oneSecondIsOneDay
```
ajpallares, line 60: Not a huge fan of this, but using `.oneSecondIsOneDay` directly creates a warning with a `fix me` button that changes it to `.monthlyRenewalEveryThirtySeconds`. And using `.monthlyRenewalEveryThirtySeconds` always results in a crash in my local machine.
I think this is acceptable since we are in unit tests and it is clearly (?) documented with comments.

Lmk if you think otherwise!

vegaro: could the `rawValue` change between versions of iOS?

ajpallares: Great question. I don't think so, as it would mean breaking the API. 

I wouldn't rely on this in the main code, but in unit tests I think it's fine (better this, with no warning and no fix me button that could lead to a crash.

For reference, this is the crash I'm getting locally when using `.monthlyRenewalEveryThirtySeconds` instead of `. oneSecondIsOneDay`.

https://forums.swift.org/t/sigabrt-freed-pointer-was-not-the-last-allocation-crash/78772

PR #5383, opened 2025-07-15
Display VC Name in Customer Center

fire-at-will:
### Description
This PR updates the iOS Customer Center to show both the virtual currency's name and code when displaying virtual currency balances so that the Customer Center's UI matches the designs.

TODO: Add Before/After screenshots

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrencyBalanceListRow.swift
+++ b/RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrencyBalanceListRow.swift
@@ -48,8 +48,13 @@ struct VirtualCurrencyBalanceListRow: View {
     struct RowData: Identifiable, Hashable {
         /// A unique identifier for the row.
         let id = UUID()
+
+        /// The virtual currency's name
+        let virtualCurrencyName: String
+
         /// The code representing the virtual currency (e.g., "GLD" for gold).
         let virtualCurrencyCode: String
+
         /// The current balance of the virtual currency.
         let balance: Int
     }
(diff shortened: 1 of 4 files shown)


fire-at-will:
@RCGitBot please test

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrencyBalanceListRow.swift
```diff
@@ -59,7 +64,7 @@ struct VirtualCurrencyBalanceListRow: View {
 
     var body: some View {
         CompatibilityLabeledContent {
-            Text(rowData.virtualCurrencyCode)
+            Text("\(rowData.virtualCurrencyName) (\(rowData.virtualCurrencyCode))")
```
vegaro, line 67: What happens if this is too long?

vegaro (approved):
Looks good after that question about the lenght

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrencyBalanceListRow.swift
fire-at-will: Good question. 

How about I add in:
```swift
.lineLimit(2)
.multilineTextAlignment(.leading)
```

so it can grow some, but not too much? Anything over 2 lines will be truncated with a `...`

The only other place in Customer Center that we're setting a line limit is in a title in `FeedbackSurveyView`, which sets the line limit to 2. It uses `.multilineTextAlignment(.center)`, which is appropriate for a title, but I don't think it's super fitting to center text in a list like this.

vegaro: yeah, I think that works! It will probably not be very common but I think 2 lines is better than elipsis

fire-at-will: Added in 2795fde2e8239a525ebd3bd1aa18e616c318b315 !

PR #5385, opened 2025-07-16
Add validation of Test Store API keys

ajpallares:
This is the first PR of a series to add support for the Test Store in the SDK.

### Checklist

### Motivation
The SDK has to identify API keys for the Test Store to modify its behavior accordingly

### Description
Adds the validation of the Test Store API keys under the `TEST_STORE` compiler flag

### DIFF
--- a/Sources/Purchasing/Configuration.swift
+++ b/Sources/Purchasing/Configuration.swift
@@ -376,11 +376,19 @@ extension Configuration {
 
     enum APIKeyValidationResult {
         case validApplePlatform
+        case testStore
         case otherPlatforms
         case legacy
     }
 
     static func validate(apiKey: String) -> APIKeyValidationResult {
+        #if TEST_STORE
+        if apiKey.hasPrefix(testStoreKeyPrefix) {
+            // Test Store key format: "test_CtDdmbdWBySmqJeeQUTyrNxETUVkajsJ"
+            return .testStore
+        }
+        #endif // TEST_STORE
+
         if applePlatformKeyPrefixes.contains(where: { prefix in apiKey.hasPrefix(prefix) }) {
             // Apple key format: "apple_CtDdmbdWBySmqJeeQUTyrNxETUVkajsJ"
             return .validApplePlatform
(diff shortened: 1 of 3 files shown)


Sources/Purchasing/Configuration.swift
```diff
@@ -376,11 +376,19 @@ extension Configuration {
...
         case legacy
     }
 
     static func validate(apiKey: String) -> APIKeyValidationResult {
+        #if TEST_STORE
+        if apiKey.hasPrefix(testStoreKeyPrefix) {
+            // Test Store key format: "test_CtDdmbdWBySmqJeeQUTyrNxETUVkajsJ"
+            return .testStore
+        }
+        #endif // TEST_STORE
```
ajpallares, line 390: Having only this under the `TEST_STORE` flag is enough, since the `APIKeyValidationResult` enum is an internal API

Sources/Purchasing/Configuration.swift
```diff
@@ -396,12 +404,14 @@ extension Configuration {
     fileprivate static func verify(apiKey: String) {
         switch self.validate(apiKey: apiKey) {
         case .validApplePlatform: break
+        case .testStore: Logger.warn(Strings.configure.testStoreAPIKey)
```
ajpallares, line 407: I wasn't sure about this being a `warn`ing or just a `debug` log. I think making it a warning doesn't harm and serves as a constant reminder that the Test Store API key will eventually need to change. Lmk if you think otherwise!

Sources/Purchasing/Configuration.swift
```diff
@@ -396,12 +404,14 @@ extension Configuration {
...
         case .legacy: Logger.debug(Strings.configure.legacyAPIKey)
         case .otherPlatforms: Logger.error(Strings.configure.invalidAPIKey)
         }
     }
 
     private static let applePlatformKeyPrefixes: Set<String> = ["appl_", "mac_"]
+    private static let testStoreKeyPrefix = "test_"
```
tonidero, line 414: I'm wondering should we make it a bit more explicit? Something like `rc_test_`? Not sure if we discussed this in the past though 😅

tonidero (approved):
Looks great! Just a small comment but not blocking

Sources/Purchasing/Configuration.swift
ajpallares: Yeah, that's a very good point. I don't have a strong opinion in favor or against it, but, for sure, it's something that we can change anytime **as long as we have the `TEST_STORE` compiler flag ON**

PR #5386, opened 2025-07-16
Store API key validation in memory

ajpallares:
### Motivation
The SDK now needs to remember the API Key validation result to adapt its behavior to the Test Store when using a Test Store API key.

### Description
This PR stores the API key validation result in `SystemInfo`, as it's an object shared across most layers of the SDK.

### DIFF
--- a/Sources/Purchasing/Configuration.swift
+++ b/Sources/Purchasing/Configuration.swift
@@ -388,7 +388,7 @@ extension Configuration {
     private static let applePlatformKeyPrefixes: Set<String> = ["appl_", "mac_"]
     private static let testStoreKeyPrefix = "test_"
 
-    static func validate(apiKey: String) -> APIKeyValidationResult {
+    private static func validate(apiKey: String) -> APIKeyValidationResult {
         #if TEST_STORE
         if apiKey.hasPrefix(testStoreKeyPrefix) {
             // Test Store key format: "test_CtDdmbdWBySmqJeeQUTyrNxETUVkajsJ"
--- a/Tests/UnitTests/Purchasing/ConfigurationTests.swift
+++ b/Tests/UnitTests/Purchasing/ConfigurationTests.swift
@@ -20,24 +20,24 @@ import XCTest
 class ConfigurationTests: TestCase {
 
     func testValidateAPIKeyWithApplPlatformSpecificKey() {
-        expect(Configuration.validate(apiKey: "appl_1a2b3c4d5e6f7h")) == .validApplePlatform
+        expect(Configuration.validateAndLog(apiKey: "appl_1a2b3c4d5e6f7h")) == .validApplePlatform
     }
 
     func testValidateAPIKeyWithMacPlatformSpecificKey() {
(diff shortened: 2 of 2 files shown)


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -304,11 +304,16 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
         let receiptRefreshRequestFactory = ReceiptRefreshRequestFactory()
         let fetcher = StoreKitRequestFetcher(requestFactory: receiptRefreshRequestFactory,
                                              operationDispatcher: operationDispatcher)
+
+        let apiKeyValidationResult = Configuration.validate(apiKey: apiKey)
+        apiKeyValidationResult.logIfNeeded()
```
ajpallares: Since we now need the API key validation result in the `init` of the `Purchases` instance, I thought it would make more sense to add the log here. From the developers perspective, it should be seamless. Lmk if you think otherwise

tonidero: Hmm right... I guess it's not entirely seamless since now we will only log once the SDK is configured, not when creating the `Configuration` object right? 

I wonder if we should instead store the verification result in the `Configuration` object as an `internal` property and access it here when initializing purchases... Not sure how much of a lift this would be though, so not a strong opinion.

ajpallares: > I guess it's not entirely seamless since now we will only log once the SDK is configured, not when creating the Configuration object right?

Yeah, that is a good point. I didn't think of the possibility of creating a `Configuration` that was not going to be used for `configure`ing the SDK, but you are 100% right in that it would be a behavior change in the logs.

That said, the existing logged strings are called `Strings.configure.invalidAPIKey` and `Strings.configure.legacyAPIKey`, which makes me think they were always meant to be logged at `configure` time... 

I checked the behavior on Android and it seems like these strings are logged at `configure` time [in the PurchasesFactory.createPurchases method](https://github.com/RevenueCat/purchases-android/blob/4fc17af98d10381a057e27a64b88bcc365ce062d/purchases/src/main/kotlin/com/revenuecat/purchases/PurchasesFactory.kt#L73), so it's probably fine (even better) to make this equal in iOS? WDYT?

> I wonder if we should instead store the verification result in the Configuration object as an internal property and access it here when initializing purchases.

If we want to keep logging these messages at `Configuration` init time, then yes, it would be a cleaner solution 👍.

tonidero: Yeah I'm honestly ok with this change :) 99% of the time, you would only create this object to configure the SDK, so it would be pretty much seamless 👍

ajpallares: Btw, I did a small refactor in https://github.com/RevenueCat/purchases-ios/pull/5386/commits/a2609f3ec34c265952556556c7fd205ef983b0a2 to make the implementation similar to what you've done in Android, which is cleaner

tonidero (approved):

PR #5393, opened 2025-07-17
Tuist: unify project settings

ajpallares:
This PR makes it so that every tuist project in the repository uses one of the `Settings` declared in the constants (`Tuist/ProjectDescriptionHelpers/Settings.swift`)


Projects/RevenueCat/Project.swift
```diff
@@ -30,11 +30,7 @@ let allDeploymentTargets: DeploymentTargets = .multiplatform(
 let project = Project(
     name: "RevenueCat",
     organizationName: .revenueCatOrgName,
-    settings: .settings(
-        base: [:].automaticCodeSigning(devTeam: .revenueCatTeamID),
-        configurations: .xcconfigFileConfigurations,
-        defaultSettings: .recommended
-    ),
+    settings: .framework,
```
ajpallares, line 33: I think the RevenueCat project does not need the code signing or the xcconfig files. Is that correct or am I missing something?

Projects/RevenueCatUI/Project.swift
```diff
@@ -12,11 +12,7 @@ let allDeploymentTargets: DeploymentTargets = .multiplatform(
 let project = Project(
     name: "RevenueCatUI",
     organizationName: .revenueCatOrgName,
-    settings: .settings(
-        base: [:].automaticCodeSigning(devTeam: .revenueCatTeamID),
-        configurations: .xcconfigFileConfigurations,
-        defaultSettings: .recommended
-    ),
+    settings: .framework,
```
ajpallares, line 15: Same as for RevenueCat

Tuist/ProjectDescriptionHelpers/Settings.swift
```diff
@@ -2,24 +2,17 @@ import ProjectDescription
...
-    /// 
+    /// Default settings configuration for app projects in the RevenueCat SDK repo.
+    ///
     /// This provides a standardized settings configuration that includes:
     /// - Base settings from `.projectBase`
     /// - Debug configuration with incremental compilation
     /// - Release configuration with whole module optimization
     /// - Recommended default settings
-    public static var project: Settings {
+    public static var app: Settings {
```
ajpallares: Lmk what you think of the name. I was also doubting with `appProject`.

Projects/RevenueCat/Project.swift
facumenzella: Technically the frameworks do not, BUT this project has a host app inside that might need it if someone wants to run it manually.

I see two options:
1. Set specific settings for the target
2. Create a new project only for that

I think I prefer 1 for now, and if we experience any downsides, we can always split it

ajpallares: I 100% agree. Thanks for the input! 🙌

ajpallares: I just tried out the two apps inside the project: `UnitTestsHostApp` and `BackendIntegrationTestsHostApp`. These are actually empty apps only used to run tests on CI. 
* Not having the xcconfig files is OK, since they don't really use any of the configs in them
* Not having the signing team doesn't seem a problem as these 2 apps can be totally run locally and I don't see the need of them run in actual devices (and it runs properly on Mac without signing anyway)
So I'd say we leave them like this for now. WDYT?

Tuist/ProjectDescriptionHelpers/Settings.swift
facumenzella: I think `appProject` is better, because the other one is called `target` which is pretty generic.
We cn also rename the other one to `framework` maybe?

ajpallares: Not exactly. We now have `.app`, `.target` and `.framework` already. `.framework` is used for packages/libraries and `.target` is used by the apps target. Perhaps we should have

`.appProject`, `.appTarget` and `.framework`. WDYT?

facumenzella: I like it!!

ajpallares: Done in https://github.com/RevenueCat/purchases-ios/pull/5393/commits/f802dcfd469d8b91c36aaceb494deba5398f8786

PR #5398, opened 2025-07-18
Tuist: prevent MagicWeather Xcode projects names clashing

ajpallares:
### Motivation
The command `tuist generate MagicWeather` generated an Xcode project named `MagicWeather.xcodeproj`, whereas that file name is already used by the MagicWeather example project meant for developers.

### Description
This PR renames the name of the Xcode project generated by `tuist generate MagicWeather` to `MagicWeatherApp.xcodeproj`.

Note that the generated `MagicWeatherApp.xcodeproj` is currently `.gitignore`d

### DIFF
--- a/Examples/MagicWeather/Project.swift
+++ b/Examples/MagicWeather/Project.swift
@@ -2,7 +2,7 @@ import ProjectDescription
 import ProjectDescriptionHelpers
 
 let project = Project(
-    name: "MagicWeather",
+    name: "MagicWeatherApp",
     organizationName: .revenueCatOrgName,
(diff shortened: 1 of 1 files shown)


facumenzella:
Thinking out loud, can we fully delete the project in favor of generating it when needed?

ajpallares:
> Thinking out loud, can we fully delete the project in favor of generating it when needed?

@facumenzella I don't think so as developers use MagicWeather as a sample app (see #5397). And we don't want to force them to use tuist to check it out. WDYT?

PR #5400, opened 2025-07-18
Tuist: fix some setup issues

ajpallares:
This PR fixes some issues we had with tuist-generated projects:
* `SWIFT_ACTIVE_COMPILATION_CONDITIONS` from xcconfig files were not being propagated to framework targets (e.g. RevenueCat and RevenueCatUI)
* `SWIFT_ACTIVE_COMPILATION_CONDITIONS` from xcconfig files were not being propagated to the visionOS platforms
* Running apps in visionOS would not work due to the minimum deployment version for visionOS not being correctly set

### DIFF
--- a/Tuist/ProjectDescriptionHelpers/Settings.swift
+++ b/Tuist/ProjectDescriptionHelpers/Settings.swift
@@ -42,9 +42,9 @@ extension Settings {
                 "CODE_SIGNING_ALLOWED": "NO",
                 "CODE_SIGNING_REQUIRED": "NO",
                 "CODE_SIGN_IDENTITY": "",
-                "SWIFT_ACTIVE_COMPILATION_CONDITIONS[sdk=xros*]": "VISION_OS",
-                "SWIFT_ACTIVE_COMPILATION_CONDITIONS[sdk=xrsimulator*]": "VISION_OS"
-            ]
+                "SWIFT_ACTIVE_COMPILATION_CONDITIONS[sdk=xros*]": "$(inherited) VISION_OS",
+                "SWIFT_ACTIVE_COMPILATION_CONDITIONS[sdk=xrsimulator*]": "$(inherited) VISION_OS"
+            ],
         )
     }
 }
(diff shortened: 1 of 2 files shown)


Tuist/ProjectDescriptionHelpers/Settings.swift
```diff
@@ -42,9 +42,10 @@ extension Settings {
                 "CODE_SIGNING_ALLOWED": "NO",
                 "CODE_SIGNING_REQUIRED": "NO",
                 "CODE_SIGN_IDENTITY": "",
-                "SWIFT_ACTIVE_COMPILATION_CONDITIONS[sdk=xros*]": "VISION_OS",
-                "SWIFT_ACTIVE_COMPILATION_CONDITIONS[sdk=xrsimulator*]": "VISION_OS"
-            ]
+                "SWIFT_ACTIVE_COMPILATION_CONDITIONS[sdk=xros*]": "$(inherited) VISION_OS",
+                "SWIFT_ACTIVE_COMPILATION_CONDITIONS[sdk=xrsimulator*]": "$(inherited) VISION_OS"
+            ],
+            configurations: .xcconfigFileConfigurations
```
JayShortway, line 48: Just checking if this works consistently? Because the other approach that Tuist unlocks is to [use env vars](https://docs.tuist.dev/en/guides/features/projects/dynamic-configuration) to set these when we generate the project, instead of dynamically in xcconfig files. We'd then only be using xcconfig files for API keys.

ajpallares: Hmm good point. For CI it's probably more convenient to use env vars, but for development I think config files (especially the local.xcconfig that is not commuted to git) is very convenient. In addition, it could get confusing having the xcconfig files only work for apps as it would break consistency (I can see future me struggling to understand (again) why a flag is working on the app code and not on the SDK code). 
That said, I'd say that if the two approaches are complementary, we could be supporting both. WDYT?

facumenzella: `$(inherited)` is doing the trick, right?

ajpallares: For visionOS yes + the `xcconfigFileConfigurationsFile` for all the platforms to see the settings from the xcconfig file in the SDKs

JayShortway: Nice, yea if this works consistently then let's do it! I remember that previously it would require cleaning and removing derived data before every part of our codebase would recognize that a flag had changed. It's very well possible that those problems were caused by the way our project was set up, and are gone with Tuist!

PR #5403, opened 2025-07-21
Add Test Store purchase UI

ajpallares:
This PR integrates the (fake) purchase UI for Test Store products (under the `TEST_STORE` compiler flag).

### Screenshots

<details><summary>iOS</summary>
<p>
image</p>
</details>
<details><summary>macOS</summary>
<p>
image 
</p>
</details>

<details><summary>tvOS</summary>
<p>
image 
</p>
</details>

<details><summary>visionOS</summary>
<p>
image
</p>
</details>

<details><summary>watchOS</summary>
<p>
image
</p>
</details>


Sources/Purchasing/TestStore/TestStorePurchaseHandler.swift
```diff
@@ -0,0 +1,225 @@
...
+    static let purchaseActionTitle = "Test Purchase"
+    static let cancelActionTitle = "Cancel"
+
+}
+
+fileprivate extension TestStoreProduct {
+
+    var purchaseAlertMessage: String {
+        var message = "⚠️ This is a test purchase and should be tested with real products using " +
+        "an Apple API key from RevenueCat.\n\n"
```
JayShortway: This is slightly confusing imo. How about something like: 
> This is a test purchase and should only be used during development. In production, use an Apple API key from RevenueCat.

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1944,6 +1938,56 @@ private extension PurchasesOrchestrator {
 
 }
 
+// MARK: - Test Store Purchases
+
+private extension PurchasesOrchestrator {
+
+    func handlePurchase(testStoreProduct: TestStoreProduct, completion: @escaping PurchaseCompletedBlock) {
+        #if TEST_STORE
```
JayShortway, line 1946: Just curious, this is our temporary compiler flag? Our will this exist when we release the feature as well?

Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift
```diff
@@ -323,8 +323,8 @@ extension StoreProduct {
         return (self.product as? SK2StoreProduct)?.underlyingSK2Product
     }
 
-    var isTestProduct: Bool {
-        return self.product is TestStoreProduct
+    internal var testStoreProduct: TestStoreProduct? {
+        return self.product as? TestStoreProduct
```
JayShortway, line 328: In Kotlin's stdlib there's a bit of a convention to name things like this `testStoreProductOrNull`, to be explicit about the optionality. Totally okay with me not to follow that convention here of course, but just mentioning it.

Sources/Purchasing/TestStore/TestStorePurchaseHandler.swift
```diff
@@ -0,0 +1,225 @@
...
+
+    @MainActor
+    private var purchaseInProgress: Bool = false
+
+    init(systemInfo: SystemInfo) {
+        self.systemInfo = systemInfo
+    }
+
+    #if TEST_STORE
+    /// - Throws: an `PurchasesError` if there's an error when trying to make the test purchase
```
JayShortway:
```suggestion
    /// - Throws: a `PurchasesError` if there's an error when trying to make the test purchase
```

JayShortway (approved):
Very nice!

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1944,6 +1938,56 @@ private extension PurchasesOrchestrator {
 
 }
 
+// MARK: - Test Store Purchases
+
+private extension PurchasesOrchestrator {
+
+    func handlePurchase(testStoreProduct: TestStoreProduct, completion: @escaping PurchaseCompletedBlock) {
```
tonidero, line 1945: I see we created a `TestStorePurchaseHandler`. Would it make sense to also move all this logic there? Mostly want to avoid increasing the size of the orchestrator too much... and this all seems like it could live there?

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -1944,6 +1938,56 @@ private extension PurchasesOrchestrator {
...
+
+    #if TEST_STORE
+    private func purchase(testStoreProduct: TestStoreProduct, completion: @escaping PurchaseCompletedBlock) {
+        self.operationDispatcher.dispatchOnMainActor {
+            do {
+                try self.testStorePurchaseHandler.purchase(product: testStoreProduct) { userConfirmed in
+                    if userConfirmed {
+                        // WIP: Implement actual test purchase completion logic
+                        // For now, we'll simulate a successful purchase
+                        completion(nil, nil, nil, false)
```
tonidero: Do we need to return the CustomerInfo for this? Though I'm ok adding that later once we add the actual implementation I think

tonidero (approved):
This looks great to me! Nice job!!

Sources/Purchasing/TestStore/TestStorePurchaseHandler.swift
ajpallares: I like it! It may be worth changing this copy also in react-native-purchases @tonidero?

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
ajpallares: This is our temporary compiler flag, yes. It will be completely removed when we release this

Sources/Purchasing/TestStore/TestStorePurchaseHandler.swift
tonidero: Yup! Will update my branch!

Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift
ajpallares: I see. Using the `orNull` suffix is not a convention in Swift AFAIK. In any case, I copied the naming style of the two properties in the lines above: 
```swift
/// Returns the `SKProduct` if this `StoreProduct` represents a `StoreKit.SKProduct`.
@objc public var sk1Product: SK1Product? {
    return (self.product as? SK1StoreProduct)?.underlyingSK1Product
}

/// Returns the `Product` if this `StoreProduct` represents a `StoreKit.Product`.
public var sk2Product: SK2Product? {
    return (self.product as? SK2StoreProduct)?.underlyingSK2Product
}
```

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
ajpallares: Yes, I think that's a good point. I intend to add the whole after-purchase logic in a separate PR so I'll take this into account to add all the logic in the `TestStorePurchaseHandler`. I agree it's better to keep `PurchasesOrchestrator` as dedicated as possible to the "real" purchases logic.
Thanks for the suggestion!

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
ajpallares: Yes, exactly. We need to return the `CustomerInfo` here. Right now, confirming the purchase crashes due to it not containing the `CustomerInfo`. I could return the same mocked `CustomerInfo` that I used for UI preview mode. It will only contain minimum hardcoded information, but I think it's better than a crash. WDYT?
It's not a big deal as this is all under the `TEST_STORE` flag, but perhaps it's better a hardcoded response than a crash anyway.

tonidero: Yeah... either that or just fetching the latest customer info... but that's probably harder, and not something we will need to do anyway, so I think a hardcoded version is ok for now!

ajpallares:
@RCGitBot please test

Sources/Purchasing/TestStore/TestStorePurchaseHandler.swift
```diff
@@ -0,0 +1,232 @@
...
+            } else {
+                window = windowScene.windows.first(where: { $0.isKeyWindow })
+            }
+        } else {
+            if #available(iOS 15.0, macCatalyst 15.0, *) {
+                window = nil
+            } else {
+                // Fallback to legacy approach on OSs where UIApplication's `windows` property is not deprecated
+                window = application.windows.first(where: { $0.isKeyWindow })
+            }
```
ajpallares, line 111: `UIApplication`'s `windows` property is deprecated since iOS 15, visionOS 1, so this was triggering a warning that was making CI fail. I've tested with visionOS 1, iOS 15 and tvOS and it works properly (the `window` is taken from `application.currentWindowScene` above

ajpallares:
I did some cleanup and added small changes based on the review comments and also to remove some deprecation warnings.
I will add unit tests in a future PR when I implement the complete Test Store Product purchase flow.

tonidero (approved):
Looks great!

Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift
JayShortway: Yes, that makes sense!

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
JayShortway: Great, thanks for explaining!

PR #5412, opened 2025-07-22
Introduce tuist-generate-workspace for CircleCI

facumenzella:
### Motivation
We want to start testing tuist in CircleCI, to slowly clear the existing workspace in favor of the generated one. It will be isolated for now, till it is stable.

### Description
- Add `tuist-generate-workspace` to generate the workspace
  - install deps
  - generate workspace
- Add manual trigger

### Next steps
- Test different modules
- As we remove target from the main project, move them here

### DIFF
--- a/Sources/Logging/Strings/VirtualCurrencyStrings.swift
+++ b/Sources/Logging/Strings/VirtualCurrencyStrings.swift
@@ -32,7 +32,7 @@ extension VirtualCurrencyStrings: LogMessage {
         case .invalidating_virtual_currencies_cache:
             return "Invalidating VirtualCurrencies cache."
         case .no_cached_virtual_currencies:
-            return "No cached VirtualCurrencies, fetching from network."
+            return "There are no cached VirtualCurrencies."
         case .virtual_currencies_stale_updating_from_network:
             return "VirtualCurrencies cache is stale, updating from network."
         case .vending_from_cache:


.circleci/config.yml
```diff
@@ -411,6 +412,16 @@ commands:
...
             bundle exec fastlane update_carthage_commit
 
+  tuist-generate-workspace:
+    description: "Install Tuist and generate workspace"
+    steps:
+      - revenuecat/install-tuist
+      - run:
+          name: Generate Tuist workspace
+          command: |
+            export PATH="$HOME/.local/share/mise/shims:$PATH"
```
ajpallares, line 422: Is this command necessary? 
I see it's already executed in the `revenuecat/install-tuist` command https://github.com/RevenueCat/sdks-circleci-orb/blob/9aa69e356bfa06564c230613e64b1ebadd544ed0/src/scripts/install-tuist.sh#L22

facumenzella: I haven't found another way yet, but yesterday I found this https://circleci.com/docs/set-environment-variable/

I need to test it out in the orb to see if it works

JayShortway: I think that should work. I used that pattern in the orb in [`set-ssh-key-for-git`](https://github.com/RevenueCat/sdks-circleci-orb/blob/main/src/scripts/set-ssh-key-for-git.sh).

PR #5413, opened 2025-07-22
Fix issue where previews running on Emerge servers were not following specialized codepaths for previews

clindsay3:
### Motivation
Certain screenshots in Emerge are flaky, and we suspect it might be because they are not using the codepaths setup explicitly for previews, to avoid doing things like network requests or accessing external resources.

### Description
Adds a new extension on `ProcessInfo` called `isRunningForPreviews` that checks for both `XCODE_RUNNING_FOR_PREVIEWS == "1"` (set when running previews in Xcode) and `EMERGE_IS_RUNNING_FOR_SNAPSHOTS == "1"` (set when running on Emerge's screenshot servers). Update all references to `XCODE_RUNNING_FOR_PREVIEWS` to this new shared ProcessInfo extension.


ajpallares (approved):
Nice one! Perfect use of `@_spi` 👨‍🍳 👌

PR #5414, opened 2025-07-23
Add missing RevenueCatUI test plans

facumenzella:
### Motivation
Just noticed that the `RevenueCatUI` test plans are not being referenced in the workspace.

### Description
- Add test plans


ajpallares (approved):
Great catch!

PR #5415, opened 2025-07-23
Dont show cancel if non-sub for customer center

facumenzella:
### Motivation
We've got a report that the OTPs were showing cancel. This fixes it.

### Next steps
We're going to move the PATHs to the backend, so we can address this type of issues more efficiently.

### DIFF
--- a/RevenueCatUI/CustomerCenter/Actions/CustomerCenterConfigData.HelpPath+PurchaseInformation.swift
+++ b/RevenueCatUI/CustomerCenter/Actions/CustomerCenterConfigData.HelpPath+PurchaseInformation.swift
@@ -34,11 +34,21 @@ extension Array<CustomerCenterConfigData.HelpPath> {
             let isNonAppStorePurchase = purchaseInformation.store != .appStore
             let isAppStoreOnlyPath = $0.type.isAppStoreOnly
 
-            let isCancel = $0.type == .cancel
+            // skip AppStore only paths if the purchase is not from App Store
+            if isNonAppStorePurchase && isAppStoreOnlyPath {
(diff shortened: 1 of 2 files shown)


RevenueCatUI/CustomerCenter/Actions/CustomerCenterConfigData.HelpPath+PurchaseInformation.swift
```diff
@@ -34,11 +34,21 @@ extension Array<CustomerCenterConfigData.HelpPath> {
...
-            let isEligibleCancel = !purchaseInformation.isSubscription
-                || (!purchaseInformation.isCancelled &&  !purchaseInformation.isLifetimeSubscription)
+            if $0.type == .cancel {
+                // don't show cancel if there's no URL
+                if isNonAppStorePurchase && purchaseInformation.managementURL == nil {
+                     return false
+                }
+
+                return purchaseInformation.isSubscription
+                    && !purchaseInformation.isLifetimeSubscription
```
vegaro: There's no way something `isSubscription == true` and `isLifetimeSubscription == true` right? so I think just checking `isSubscription` is enough? Or do promotionals break that?

vegaro (approved):
I think it makes sense

RevenueCatUI/CustomerCenter/Actions/CustomerCenterConfigData.HelpPath+PurchaseInformation.swift
facumenzella: I think you're right. I'll address this in another PR

PR #5417, opened 2025-07-23
Replace isLifetimeSubscription for isLifetime

facumenzella:
### Motivation
After discussing with @vegaro, we've realized that `isLifetimeSubscription` made no sense.

### Description
- Replace `isLifetimeSubscription` with `isLifetime`, which is true only for `nonConsumable` store products


RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -123,6 +123,9 @@ struct PurchaseInformation {
     /// Remote configured set of product ids to handle change plans flow.
     let changePlan: CustomerCenterConfigData.ChangePlan?
 
+    /// Indicates the product is lasts forever
```
vegaro:
```suggestion
    /// Indicates the product lasts forever
```

PR #5419, opened 2025-07-23
Add GetWebBillingProductsOperation

ajpallares:
### Description
Adds the `GetWebBillingProductsOperation` that will allow us to fetch the Products from the Test Store

## Note
Most of the changes in the diff are the new files generated by the new test cases for `GetWebBillingProductsOperation`


ajpallares:
@RCGitBot please test

Sources/Networking/OfferingsAPI.swift
```diff
@@ -74,6 +77,25 @@ class OfferingsAPI {
         )
     }
 
+    func getWebProducts(appUserID: String, productIds: Set<String>, completion: @escaping WebProductsResponseHandler) {
```
ajpallares, line 80: This is unused for now. Will be used by a future PR

Sources/Networking/Operations/GetWebProductsOperation.swift
```diff
@@ -0,0 +1,90 @@
...
+    ) -> CacheableNetworkOperationFactory<GetWebProductsOperation> {
+        return .init({ cacheKey in
+                    .init(
+                        configuration: configuration,
+                        webProductsCallbackCache: webProductsCallbackCache,
+                        productIds: productIds,
+                        cacheKey: cacheKey
+                    )
+            },
+            individualizedCacheKeyPart: configuration.appUserID + "\n" + productIds.sorted().joined(separator: "\n"))
```
ajpallares, line 35: I think this cache key make sense. LMK if you see some issues there

Sources/Networking/Responses/WebOfferingProductsResponse.swift
```diff
@@ -15,43 +15,10 @@ import Foundation
...
-    }
-
-    struct Product {
-        let identifier: String
-        let productType: String
-        let title: String
-        let description: String?
-        let defaultPurchaseOptionId: String?
-        let purchaseOptions: [String: PurchaseOption]
-    }
```
ajpallares, line 49: These are now part of `WebProductsResponse`, since the models are exactly the same

ajpallares:
@RCGitBot please test

Sources/Networking/OfferingsAPI.swift
```diff
@@ -74,6 +77,27 @@ class OfferingsAPI {
         )
     }
 
+    func getWebBillingProducts(
```
ajpallares: This method is the entry point for all the changes in this PR but it's unused for now. It will be used by a future PR

Sources/Networking/Operations/GetWebBillingProductsOperation.swift
```diff
@@ -0,0 +1,91 @@
...
+    ) -> CacheableNetworkOperationFactory<GetWebBillingProductsOperation> {
+        return .init({ cacheKey in
+                    .init(
+                        configuration: configuration,
+                        webBillingProductsCallbackCache: webBillingProductsCallbackCache,
+                        productIds: productIds,
+                        cacheKey: cacheKey
+                    )
+            },
+            individualizedCacheKeyPart: configuration.appUserID + "\n" + productIds.sorted().joined(separator: "\n"))
```
ajpallares, line 35: I think this cache key make sense. LMK if you see some issues there

ajpallares:
@RCGitBot please test

Sources/Networking/OfferingsAPI.swift
tonidero: I do wonder if this should be out of `OfferingsAPI` since it's not completely related to offerings... We can always move it later though, so I don't think it's a blocker.

Sources/Networking/WebBillingHTTPRequestPath.swift
```diff
@@ -50,12 +54,20 @@ extension HTTPRequest.WebBillingPath: HTTPRequestPath {
         switch self {
         case let .getWebOfferingProducts(appUserID):
             return "/rcbilling/v1/subscribers/\(appUserID.trimmedAndEscaped)/offering_products"
+        case let .getWebBillingProducts(userId, productIds):
+            let encodedUserId = userId.trimmedAndEscaped
+            let encodedProductIds = productIds.sorted().map { productId in
```
tonidero: Hmm I guess the sorting is to make tests deterministic? Is there some data structure that can be used as a "sorted set" of sorts? (kotlin does have a data structure for that FWIW). If not, I think this is probably ok for now

tonidero (approved):
This looks great to me!! Amazing job!

Sources/Networking/OfferingsAPI.swift
ajpallares: That is an excellent point. And I'd rather correct it now than leave it for later, so thank you for pointing it out!

Sources/Networking/WebBillingHTTPRequestPath.swift
ajpallares: Good catch!

I wasn't thinking about tests directly but, in reality, yes. I was trying to make the requests always exactly the same for the same `productIds`, which is indeed only useful for tests I guess.
There's [OrderedSet](https://swiftpackageindex.com/apple/swift-collections/1.2.0/documentation/orderedcollections/orderedset) in Swift, but it's part of the [Swift Collections Package](https://github.com/apple/swift-collections), which is official but we don't want to add external dependencies (even less only for this).
There's also `NSOrderedSet`, which comes from Objective-C, but I don't think it's worth using it.

Instead, if you agree, I think it's better to remove the `sorted()` method and deal with it in the tests side. Thank you!

ajpallares: Done in https://github.com/RevenueCat/purchases-ios/pull/5419/commits/09baec0c7768df3d2fa85084413c536a4e2108c9

Sources/Networking/OfferingsAPI.swift
ajpallares: Done in https://github.com/RevenueCat/purchases-ios/commit/24335549929aeb2cdcd2b4f067734e8077795d39

ajpallares:
@RCGitBot please test

Sources/Networking/WebBillingHTTPRequestPath.swift
ajpallares: And... fixed the flakiness (due to the product ids not being sorted anymore) in the tests in https://github.com/RevenueCat/purchases-ios/pull/5419/commits/d5e1a667b32f4720ac7a9203b5312881e6a6b1f5 😅

ajpallares:
@RCGitBot please test

PR #5434, opened 2025-07-30
Create StoreTransaction for Test Store purchases

ajpallares:
### Checklist

### Description

When a purchase of a Test Store product is successful, the SDK generates a fake `StoreTransaction` that is used to complete the (existing) successful purchase flow. The "fake" transaction:
* gets posted in the post receipt request.
* is also the transaction returned from the public `purchase` API.

In addition, the purchase of Test Store products return the Customer Info following this logic:
* Returns the cached customer info in case of simulated purchase failure (including the purchase cancelation).
* Returns the customer info from the result of the post receipt request in case of simulated successful purchase.

### Notes
All of this is implemented under the TEST_STORE compiler flag for now. Trying to purchase a `TestStoreProduct` still results in an error.

### DIFF
--- a/Sources/Purchasing/TestStore/TestStorePurchaseHandler.swift
+++ b/Sources/Purchasing/TestStore/TestStorePurchaseHandler.swift
@@ -88,7 +88,6 @@ actor TestStorePurchaseHandler: TestStorePurchaseHandlerType {
         let testStoreTransaction = TestStoreTransaction(productIdentifier: product.productIdentifier,
                                                         purchaseDate: purchaseDate,
                                                         transactionIdentifier: transactionId,
-                                                        quantity: 1,
                                                         storefront: storefront,
                                                         jwsRepresentation: nil)
         return StoreTransaction(testStoreTransaction)
--- a/Sources/Purchasing/TestStore/TestStoreTransaction.swift
+++ b/Sources/Purchasing/TestStore/TestStoreTransaction.swift
@@ -14,10 +14,9 @@ struct TestStoreTransaction: StoreTransactionType, Equatable {
     let purchaseDate: Date
     let transactionIdentifier: String
 
-    var hasKnownPurchaseDate: Bool { return true}
+    var hasKnownPurchaseDate: Bool { return true }
     var hasKnownTransactionIdentifier: Bool { return true }
-
-    let quantity: Int
+    var quantity: Int { return 1 }
 
     let storefront: Storefront?
 
--- a/Tests/UnitTests/TestStore/PurchasesOrchestratorTestStoreTests.swift
+++ b/Tests/UnitTests/TestStore/PurchasesOrchestratorTestStoreTests.swift
@@ -460,7 +460,6 @@ class PurchasesOrchestratorTestStoreTests: TestCase {
         return TestStoreTransaction(productIdentifier: "test_product_id",
                                     purchaseDate: self.eventTimestamp1,
                                     transactionIdentifier: "test_transaction_id",
-                                    quantity: 1,
                                     storefront: Storefront(mockStorefront),
                                     jwsRepresentation: nil)
     }


Sources/Purchasing/TestStore/TestStorePurchaseHandler.swift
```diff
@@ -18,64 +18,121 @@ import AppKit
...
+    ) {
+        Task {
+            let transaction = await self.createStoreTransaction(product: product)
+            completion(.success(transaction))
+        }
+    }
+
+    private func createStoreTransaction(product: TestStoreProduct) async -> StoreTransaction {
+        let transactionId = "test_order_" + UUID().uuidString
+        let storefront = await Storefront.currentStorefront
```
ajpallares, line 87: I see in the Android SDK the `Storefront` is centralized by `BillingAbstract`'s `getStorefront` method ([here](https://github.com/RevenueCat/purchases-android/blob/3a31fcc0f69c011cff2559415c9bb8b2fc143dfc/purchases/src/main/kotlin/com/revenuecat/purchases/common/BillingAbstract.kt#L124)). In iOS, we have `Storefront.currentStorefront` all over the place, which returns StoreKit 2's storefront (the user's Storefront):
https://github.com/RevenueCat/purchases-ios/blob/d68ec4df92732b58059c677ae25319158b524531/Sources/Purchasing/StoreKitAbstractions/Storefront.swift#L80
For now I think it's OK to use that same Storefront until we need to mock it (see https://github.com/RevenueCat/purchases-android/pull/2554#discussion_r2241987004)

Sources/Purchasing/TestStore/TestStorePurchaseHandler.swift
```diff
@@ -18,64 +18,121 @@ import AppKit
...
-protocol TestStorePurchaseHandlerType: AnyObject {
+protocol TestStorePurchaseHandlerType: AnyObject, Sendable {
 
     #if TEST_STORE
-    /// - Throws: a `PurchasesError` if there's an error when trying to make the test purchase
-    /// (e.g. there's already a purchase in progress).
+
     @MainActor
-    func purchase(product: TestStoreProduct, completion: @escaping (TestPurchaseResult) -> Void) throws
+    func purchase(product: TestStoreProduct) async -> TestPurchaseResult
```
ajpallares, line 22: I also used this PR to change this API from closure-based to async/await

Sources/Purchasing/TestStore/TestStorePurchaseHandler.swift
```diff
@@ -18,64 +18,121 @@ import AppKit
...
+        product: TestStoreProduct, completion: @escaping (TestPurchaseResult) -> Void
+    ) {
+        Task {
+            let transaction = await self.createStoreTransaction(product: product)
+            completion(.success(transaction))
+        }
+    }
+
+    private func createStoreTransaction(product: TestStoreProduct) async -> StoreTransaction {
+        let transactionId = "test_\(Date().millisecondsSince1970)_\(UUID().uuidString)"
```
ajpallares: This fake transaction identifier gets sent to our servers and should have the same format as the identifiers sent from Android and JS

Sources/Purchasing/TestStore/TestStorePurchaseUI.swift
ajpallares, line 1: I separated the whole fake UI logic for the purchase of test products into a new object `DefaultTestStorePurchaseUI` to enable injecting a mocked `TestStorePurchaseUI` in unit tests

Sources/Purchasing/StoreKitAbstractions/StoreTransaction.swift
```diff
@@ -156,6 +156,10 @@ extension StoreTransaction {
         return (self.transaction as? SK2StoreTransaction)?.underlyingSK2Transaction
     }
 
+    internal var testStoreTransaction: TestStoreTransaction? {
```
ajpallares, line 159: Needed in unit tests to check the expected transaction is returned

Sources/Purchasing/TestStore/TestStorePurchaseHandler.swift
tonidero: Right... I think it's ok to do this for now but ideally in the future we disconnect it from the store entirely, and can be selected by the backend or some type of configuration. Not a blocker though!

Sources/Purchasing/TestStore/TestStoreTransaction.swift
```diff
@@ -0,0 +1,33 @@
...
+
+import Foundation
+
+struct TestStoreTransaction: StoreTransactionType, Equatable {
+
+    let productIdentifier: String
+    let purchaseDate: Date
+    let transactionIdentifier: String
+
+    var hasKnownPurchaseDate: Bool { return true}
```
tonidero:
```suggestion
    var hasKnownPurchaseDate: Bool { return true }
```

Sources/Purchasing/TestStore/TestStoreTransaction.swift
```diff
@@ -0,0 +1,33 @@
...
+    var hasKnownTransactionIdentifier: Bool { return true }
+
+    let quantity: Int
```
tonidero: Hmm should we default this to 1? Not sure if we need to support multiquantity transactions right now 😅

tonidero (approved):
I think this all makes sense! Cool to have this working 🙌

Sources/Purchasing/TestStore/TestStoreTransaction.swift
ajpallares: You're totally right. I set the 1 in the parameters, but I could totally do it here. Thanks for the suggestion!

ajpallares: Done in https://github.com/RevenueCat/purchases-ios/pull/5434/commits/a02e8ca4863bba2ef81b018d906114c933b35eca

PR #5437, opened 2025-07-30
Add offer code redemption support to paywall buttons

joshdholtz:
## Summary
- Add new `offerCode` destination type to `PaywallButtonComponent` with proper encoding/decoding
- Implement navigation handling in `ButtonComponentView` to present the system offer code redemption sheet
- Add cache invalidation after offer code redemption to ensure customer info is refreshed

## Test plan
- [ ] Verify paywall buttons can be configured with `offer_code` destination in JSON
- [ ] Test that tapping a button with offer code destination presents the system redemption sheet
- [ ] Confirm customer info cache is invalidated after offer code redemption
- [ ] Ensure existing button functionality remains unaffected

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Button/ButtonComponentViewModel.swift
+++ b/RevenueCatUI/Templates/V2/Components/Button/ButtonComponentViewModel.swift
@@ -107,9 +107,9 @@ class ButtonComponentViewModel {
         case .navigateTo(destination: let destination):
             if case .offerCodeRedemptionSheet = destination {
 #if os(iOS) && !targetEnvironment(macCatalyst)
-                    return true
-#else
                     return false
+#else
+                    return true
 #endif
             } else {
                 return false


RevenueCatUI/Templates/V2/Components/Button/ButtonComponentView.swift
```diff
@@ -134,10 +137,14 @@ struct ButtonComponentView: View {
         case .unknown:
             break
         case .webPaywallLink(url: let url, method: let method):
-            openWebPaywallLink(url: url, method: method)
+            self.openWebPaywallLink(url: url, method: method)
         }
     }
 
+    private func openCodeREDemptionSheet() {
+        Purchases.shared.presentCodeRedemptionSheet()
```
tonidero, line 147: Not part of this PR, since I see we were already calling `Purchases.shared` when invalidating the customer info when opening a web paywall link... But I do think these methods should ideally belong to the view model, not to the view.

RevenueCatUI/View+PresentPaywall.swift
```diff
@@ -480,17 +482,30 @@ private struct PresentingPaywallModifier: ViewModifier {
             }
         }
         .task {
-            guard let info = try? await self.customerInfoFetcher() else { return }
+            await self.updateCustomerInfo()
+        }
+        .onChange(of: scenePhase) { newPhase in
+            // Used when Offer Code Redemption sheet dismisses
```
tonidero, line 488: Correct me if I'm wrong but would this be called every time the app becomes active even if it didn't go through the offer code redemption sheet, but something else? I guess it's not a huge deal... but we should be careful about invalidating customer info too often IMO, since it could lead to a user temporarily losing entitlements momentarily. If you can confirm this would only happen on offer code redemption sheet dismissal, I'm ok with this :)

tonidero (commented):
Just a question but looking great!!

RevenueCatUI/Templates/V2/Components/Button/ButtonComponentView.swift
ajpallares: Besides what Toni says, we should probably avoid calling `Purchases.shared` directly and go through the existing `PurchaseHandler` instance instead (the person working on the RC app will later thank us for this 😄).

RevenueCatUI/Templates/V2/Components/Button/ButtonComponentView.swift
```diff
@@ -134,10 +137,14 @@ struct ButtonComponentView: View {
         case .unknown:
             break
         case .webPaywallLink(url: let url, method: let method):
-            openWebPaywallLink(url: url, method: method)
+            self.openWebPaywallLink(url: url, method: method)
         }
     }
 
+    private func openCodeREDemptionSheet() {
```
ajpallares: Unless it was an intended (code RED) pun:
```suggestion
    private func openCodeRedemptionSheet() {
```

RevenueCatUI/View+PresentPaywall.swift
ajpallares: Good point from Toni here. The code currently only checks when the `scenePhase` passes to `active`. Perhaps we could restrict this to make it closer to the "when Offer Code Redemption sheet dismisses" event by using https://developer.apple.com/documentation/swiftui/view/onchange(of:initial:_:)-4psgg and only do `updateCustomerInfo()` `if newPhase == .active && oldPhase == .inactive` (this way we avoid calling it when coming from the `.background`)

ajpallares (commented):
I have a couple of more comments, building on Toni's

RevenueCatUI/View+PresentPaywall.swift
joshdholtz: Screenshot 2025-07-31 at 2 45 57 PM

🫠

joshdholtz: GOT IT

RevenueCatUI/Modifiers/ViewExtensions.swift
```diff
@@ -55,6 +55,42 @@ extension View {
...
+        perform action: @escaping (_ change: ChangeOf<V>) -> Void
+    ) -> some View where V: Equatable {
+        #if swift(>=5.9)
+        if #available(iOS 17.0, macOS 14.0, tvOS 17.0, watchOS 10.0, *) {
+            AnyView(self.onChange(of: value) { old, new in
+                if old == new {
+                    action(.new(new))
+                } else {
+                    action(.changed(old: old, new: new))
+                }
```
ajpallares: Total nit, and I understand that it could be misleading to say `changed` when `old == new`, but that would be the fault of the system API calling `onChange` when the value didn't change. This way, we don't lose the `oldValue` information (and we wouldn't be getting the Customer Info below if both `old` and `new` were `.active`
```suggestion
                action(.changed(old: old, new: new))
```

RevenueCatUI/Templates/V2/Components/Button/ButtonComponentView.swift
ajpallares: As talked with Josh privately, removing the `Purchases.shared` here is more intricate than it looks. So it'll be tackled in a future PR (thank you Swift protocols for not allowing @availability checks in your methods)

RevenueCatUI/Modifiers/ViewExtensions.swift
joshdholtz: Oh yeah, good point! Will change this 👍

PR #5440, opened 2025-07-31
Delete .swiftpm scheme folder to avoid scheme pollution

facumenzella:
### Motivation
We've noticed that when integrating RC, the workspace xcschemes get polluted with our stuff. This PR tries to address that.

image

This should clean only CustomEntitlements and ReceiptParser. RC and RCUI need to be tested through the package, to be able to test from iOS14...iOS18

### Description
- Move RC_CRC and ReceiptParser from`.swiftpm/xcode/xcshareddata/xcschemes` folder to RevenueCat.xcworkspace schemes
- Ignore those schemes


facumenzella:
@RCGitBot please test

JayShortway (commented):
Thanks for taking care of this! Seems that some of the SPM tests don't like it though haha. Also, can this break anything for anyone, either for people already using RC through SPM, or for new people? Just trying to be careful.

ajpallares:
> Also, can this break anything for anyone, either for people already using RC through SPM, or for new people? Just trying to be careful.

@JayShortway it shouldn't break anything. If fact Firebase also made a similar change in the past to achieve the same (see https://github.com/firebase/firebase-ios-sdk/pull/8060). That said, we should totally be extra careful about this and test it thoroughly. 

@facumenzella, perhaps, instead of just gitignoring `.swiftpm/xcode/xcshareddata/xcschemes`, we could gitignore the whole `xcshareddata` folder: `.swiftpm/xcode/xcshareddata`

JayShortway:
@ajpallares Thanks for explaining and the additional context!

facumenzella:
@RCGitBot please test

facumenzella:
@RCGitBot please test

facumenzella:
@RCGitBot please test

facumenzella:
@RCGitBot please test

.gitignore
```diff
@@ -88,3 +88,8 @@ Projects/**/*.xcworkspacedata
 Examples/**/.xcworkspacedata
 Examples/**/*.xcscheme
 Examples/**/xcshareddata/**
+
+# Avoid xcscheme pollution when integrating SDK with SPM
+# We can't avoid the whole folder yet because RC and RCUI need to be tested through the package, rather than the workspace
+.swiftpm/xcode/xcshareddata/xcschemes/ReceiptParser.xcscheme
+.swiftpm/xcode/xcshareddata/xcschemes/RevenueCat_CustomEntitlementComputation.xcscheme
```
ajpallares, line 95: Doesn't this affect the tests of RC_CEC?

facumenzella: Just double-checked, this still works because it uses a test plan: `BackendIntegrationTests-CustomEntitlements.xctestplan`. That works with the workspace, so it does not affect the `.swiftpm` folder

ajpallares: Thank you for checking!

ajpallares (approved):

facumenzella:
@RCGitBot please test

PR #5441, opened 2025-07-31
Simplification of `SubscriptionHistoryTracker`

ajpallares:
This is a slight simplification to merge the logic of both `SubscriptionHistoryObserver` and `SubscriptionHistoryTracker` into `SubscriptionHistoryTracker`.

In addition, the `SubscriptionHistoryTracker` now exposes the `Status` publisher, so that the `PaywallPromoOfferCache` can subscribe to the `Status` changes instead of only passing them at `PaywallV2View` initialization time. This ensures that the `hasAnySubscriptionHistory` is always up to date during the lifetime of the `PaywallPromoOfferCache` instance.


ajpallares:
@joshdholtz if we wanted to have the views react to the changes in `PaywallPromoOfferCache`'s `hasAnySubscriptionHistory` boolean, I think this might do the trick (not completely sure)

```swift
private var hasAnySubscriptionHistory: Bool = false {
    didSet {
        self.objectWillChange.send()
    }
}
```

PR #5443, opened 2025-07-31
Remove ReceiptParser from workspace

facumenzella:
### Motivation
After [this](https://revenuecat.slack.com/archives/CEU9GNWTW/p1753820816944279) thread, we're removing `ReceiptParser` from the workspace. 

It's still accessible via the `Package.swift`, and via `tuist generate ReceiptParser` 

### Description
- Delete workspace references for `ReceiptParser`


facumenzella:
@RCGitBot please test

facumenzella:
This is way harder than I thought 😓 
I'll come back to this later

PR #5444, opened 2025-07-31
Fix previews of paywall components not using Mac Catalyst button styling applied at top-level of Paywalls V2

clindsay3:
### Motivation
This is a follow-up PR to https://github.com/RevenueCat/purchases-ios/pull/5372.

That PR fixed the presentation of buttons in actual paywalls, but the Swift previews (and Emerge screenshots) did not reflect the newly applied button styling on Mac Catalyst.

### Description
* Adds a new `fixMacButtons()` modifier that applies the appropriate styling.
* Add `fixMacButtons()` to the `previewRequiredEnvironmentProperties()` modifier we already use on all of our Paywalls V2 components.
* Rename `previewRequiredEnvironmentProperties` to `previewRequiredPaywallsV2Properties`. This (1) reflects the fact that we're not only setting environment properties after this change, and (2) makes it clear that this modifier should only be used on paywall components. This is important, because there is an implicit assumption here that the `fixMacButtons()` modifier will be applied wherever these components are actually used (in practice, in `LoadedPaywallsV2View`). If this modifier were used on a non-paywall-components view, it might not accurately reflect how that view will be presented on Mac Catalyst.

I considered an alternative approach where I applied the button styling on each individual component, instead of at the top-level, which would also solve this problem. But I was concerned that it would be too easy for someone to add a new component and not know that they need to apply the `fixMacButtons()` property to it. So I felt like this approach was a little more future-proof (although not without its own downsides).


joshdholtz (approved):
:shipit: :shipit: :shipit: :shipit: :shipit:

PR #5449, opened 2025-08-01
Fix dynamic type scaling for generic fonts in Paywalls

clindsay3:
### Motivation
Dynamic Type was not working when a generic font was being used in Paywalls. According to @joshdholtz, dynamic type is supposed to be supported in Paywalls.

### Description
* Update `GenericFont.makeFont` to scale the font size value according to dynamic type.

Note: I tried adding a Swift preview that would test that the scaling is being applied, so we could verify this behavior. But it appears that `UIFontMetrics.default.scaledValue` doesn't work in Swift Previews. 😞 So I'm open to other ideas on how to add testing for this behavior.


vegaro:
Closing since this was fixed in https://github.com/RevenueCat/purchases-ios/pull/5931/

PR #5453, opened 2025-08-05
Crash on release when using a Test Store API key

ajpallares:
### Motivation

Using a Test Store API Key in production will result in apps being submitted to the App Store without proper in-app purchase configuration. This would lead to a broken user experience, no proceeds being made from IAPs and, very likely, to the rejection of the app. We want to ensure these builds are caught early and prevented from reaching review.

### Description

This PR introduces a crash at runtime if the app is running a Release build when configuring the SDK with a Test Store API Key. **Debug builds are unaffected**.

## Important: XCFramework support

When the SDK is distributed as an XCFramework (binary compiled with Release configuration), the client app’s `DEBUG` compiler flag isn’t propagated into the binary. As a result, any build—including those intended as “development” when integrated via XCFramework—will crash if a Test Store API Key is used. 

This means that **developing with a Test Store API key isn’t supported when adding the SDK dependency as an XCFramework**.


Sources/Purchasing/Configuration.swift
```diff
@@ -392,7 +392,14 @@ extension Configuration {
         #if TEST_STORE
         if apiKey.hasPrefix(testStoreKeyPrefix) {
             // Test Store key format: "test_CtDdmbdWBySmqJeeQUTyrNxETUVkajsJ"
+
+            #if DEBUG
             return .testStore
+            #else
+            // In release builds, we intentionally crash to prevent submitting an app with a Test Store API key
+            fatalError("[RevenueCat]: Test Store API key used in RELEASE build." +
+                       "Please configure a valid API key before releasing.")
```
JayShortway: Should we guide the dev a bit more? Maybe: 
```suggestion
                       "Please configure the App Store app on the RevenueCat dashboard and use its corresponding Apple API key before releasing.")
```

JayShortway (approved):
Good one!

Sources/Purchasing/Configuration.swift
ajpallares: I like it! yes, it's much clearer this way 👌 Thanks!!
Done in https://github.com/RevenueCat/purchases-ios/pull/5453/commits/9257137401bbf15d518fbed7eb9c01e82cdfa79f

PR #5454, opened 2025-08-06
Log Storefront country code at SDK configuration time

ajpallares:
This PR logs the device's Storefront country code at configuration time, like these
> DEBUG: ℹ️ Storefront country code: ESP

> DEBUG: ℹ️ Storefront country code: USA

This should be helpful when debugging


Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
```diff
@@ -208,6 +208,8 @@ final class PurchasesOrchestrator {
         Task {
             await syncDiagnosticsIfNeeded()
         }
+
+        Logger.debug(Strings.storeKit.sk_storefront_country_code(countryCode: systemInfo.storefront?.countryCode))
```
tonidero: In what cases would this be null? Just asking in case we need to wait to obtain it slightly...

tonidero (commented):
Just a question, but looks good!

Sources/Purchasing/Purchases/PurchasesOrchestrator.swift
ajpallares: Right now it's using SK1 (see [documentation](https://developer.apple.com/documentation/storekit/skpaymentqueue/storefront)), which is a synchronous API. That said, the documentation does recommend calling this API from a background thread, so I will modify it just in case.

I honestly don't know under what circumstances it returns nil. I haven't seen it documented.

I do intend to migrate `systemInfo.storefront` to SK2's Storefront which is indeed asynchronous. But I'll do that in a separate PR because it's likely to imply other changes.

Thanks for the observations!

ajpallares: Made this PR depend on https://github.com/RevenueCat/purchases-ios/pull/5455, where I implement the asynchronous SK2's `Storefront.current`. 

Now the log in this PR is asynchronous and uses the new implementation to fetch the Storefront.

PR #5455, opened 2025-08-07
Update storefront apis

ajpallares:
### Motivation
As stated in [SK1's `SKPaymentQueue`'s `storefront` property documentation](https://developer.apple.com/documentation/storekit/skpaymentqueue/storefront):
> [storefront](https://developer.apple.com/documentation/storekit/skpaymentqueue/storefront) is a synchronous API that may take significant time to return. Don’t use [storefront](https://developer.apple.com/documentation/storekit/skpaymentqueue/storefront) in a time-sensitive thread, such as during app launch. To get asynchronous behavior, dispatch it to a separate queue, or use the asynchronous [current](https://developer.apple.com/documentation/storekit/storefront/current) property of [Storefront](https://developer.apple.com/documentation/storekit/storefront) instead.

The iOS SDK is currently using this API as a synchronous API and not considering that it may be a blocking call.

In addition, this API has been deprecated in iOS 26.

### Description
The main changes in this PR is to make `SystemInfo`'s `storefront` an async getter. Under the hood:
* It uses SK2's `Storefront`'s [`current`](https://developer.apple.com/documentation/storekit/storefront/current) async property by default in iOS 15+.
* In iOS 13 and iOS 14, it still uses SK1's `SKPaymentQueue`'s `storefront`, but converted into an async operation with `Task.detached`, to avoid blocking (as the documentation recommends).

Most of the changes are a consequence of making it async: e.g. `HTTPClient`, diagnostics tracking methods that receive the storefront and unit tests.


ajpallares:
@RCGitBot please test

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -248,7 +248,7 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
     }
 
     @objc public var storeFrontCountryCode: String? {
```
tonidero, line 250: Should we deprecate this property and provide an async/callback API?

Sources/Purchasing/StoreKitAbstractions/Storefront.swift
```diff
@@ -74,19 +74,23 @@ extension Storefront: Sendable {}
...
-    private static var currentStorefrontType: StorefrontType? {
+    internal static var currentStorefrontType: StorefrontType? {
         get async {
             if #available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *) {
                 let sk2Storefront = await StoreKit.Storefront.current
                 return sk2Storefront.map(SK2Storefront.init)
             } else {
-                return Self.sk1CurrentStorefrontType
+                return await Task.detached {
+                    return Self.sk1CurrentStorefrontType
```
tonidero, line 84: Just to check, since my Swift concurrency has always been rusty 😅. Since this  is running on a detached task, does the returned value also runs on a low priority thread in this case? If so, It could be a behavior change unless we change the thread back to the main thread somewhere in the call chain (didn't check if we do). Otherwise, I think this is good!

Tests/UnitTests/Mocks/MockHTTPClient.swift
```diff
@@ -80,37 +80,41 @@ class MockHTTPClient: HTTPClient {
         with verificationMode: Signing.ResponseVerificationMode? = nil,
         completionHandler: Completion<Value>?
     ) {
+        Task {
```
tonidero: The contents inside this don't seem to be correctly indented?

Sources/Purchasing/Purchases/Purchases.swift
ajpallares: I thought about it, but wasn't really sure.

It can be done in a separate PR, with these considerations:
* There's one downside: the new API wouldn't be ObjC-compatible. We could create a ObjC-compatible method with a completion handler for it if we think it's needed.
* I can't think of a good name for it, as the most straightforward one is taken. We could "fix" the existing name by removing the capital case "F", having the new property be `storefrontCountryCode`, but I think it's too similar to the one we'd be deprecating.

Sources/Purchasing/StoreKitAbstractions/Storefront.swift
ajpallares: The only reason for this to be a `detached` `Task` is to make sure that it runs on a background thread. If we were to use `Task` and this code were to be run from the main thread, because `Self.sk1CurrentStorefrontType` is not marked as `async` (even though this coming from SK1' `storefront` getter means that it could be a blocking call), then `Self.sk1CurrentStorefrontType` would still be run on the main thread.

Now, because we're `await`ing for the `Task.detached` value, I think the code after the return value is also run in the same actor context. I.e., if called from the main actor, the code after getting the storefront would still run on the main actor.

> It could be a behavior change unless we change the thread back to the main thread somewhere in the call chain

I don't think so, only because this is not affecting any of the public APIs of the SDK. The existing `storeFrontCountryCode` public getter still relies on the sync (potentially thread-blocking) SK1's implementation with no async call in the middle

tonidero: Ahh cool, that all makes sense. Thanks for explaining!!

Sources/Purchasing/Purchases/Purchases.swift
tonidero: Yeah that's a good point... I think this is something we might want to consider breaking in the next major. Should probably take note in Linear so we don't forget for the next major, whenever that is.

Tests/UnitTests/Mocks/MockHTTPClient.swift
ajpallares: Good eye! Thank you!

ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -160,6 +160,24 @@ class HTTPClient {
...
+
+        Task {
+            storefront.value = await self.systemInfo.storefront
+            semaphore.signal()
+        }
+
+        semaphore.wait()
+
+        return storefront.value
+    }
```
ajpallares: Trying to introduce an `async` call in `HTTPClient` got more complicated than I expected. Introducing a `Task` to asynchronous get the Storefront for the `defaultHeaders` in this call
https://github.com/RevenueCat/purchases-ios/blob/85c853457ee12d232589726acc12b296cacca187/Sources/Networking/HTTPClient/HTTPClient.swift#L99-L104

cannot be done without breaking the expectation of requests happening serially, as tested by https://github.com/RevenueCat/purchases-ios/blob/85c853457ee12d232589726acc12b296cacca187/Tests/UnitTests/Networking/HTTPClientTests.swift#L1222

A deeper migration of `HTTPClient` to support async/await tasks would be needed, I think.

For now, I just went with this workaround to keep the storefront getter synchronous by using a Semaphore. Note that while this would block the current thread, this is OK as long as this blocking never happens in the main thread, which is true since all network operations are added to a background queue. See https://github.com/RevenueCat/purchases-ios/blob/85c853457ee12d232589726acc12b296cacca187/Sources/Networking/Backend.swift#L48-L54

The (only) advantage here is that we are now using SK2's storefront when available.

The alternative would be keep using the existing (but deprecated since iOS 26) SK1's storefront getter which is synchronous but also blocking as per its documentation. The advantage is that we wouldn't need this workaround.

@RevenueCat/coresdk please, let me know what you think!

ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

PR #5457, opened 2025-08-07
Show a subscribe button in customer center when there are no subscriptions

facumenzella:
### Motivation
We want to show a subscribe button whenever there's an empty state. This is configurable from the dashboard.

### Description
- Add subscribe button
- Integrate `ScreenOffering`

Thanks to @ajpallares for the help with the animation 💅 

https://github.com/user-attachments/assets/a04b502c-ceff-4d62-8a8b-d2670e9bd602

### DIFF
--- a/Sources/CustomerCenter/CustomerCenterConfigData.swift
+++ b/Sources/CustomerCenter/CustomerCenterConfigData.swift
@@ -777,7 +777,7 @@ extension CustomerCenterConfigData.Screen {
         self.paths = response.paths.compactMap { CustomerCenterConfigData.HelpPath(from: $0) }
         self.offering = response.offering.map { offering in
             switch offering.type {
-            case "SPECIFIC":
+            case CustomerCenterConfigData.ScreenOffering.OfferingType.specific.rawValue:
                 return CustomerCenterConfigData.ScreenOffering(
                     type: .specific,
                     offeringId: offering.offeringId


RevenueCatUI/CustomerCenter/Views/NoSubscriptionsCardView.swift
```diff
@@ -59,11 +83,66 @@ struct NoSubscriptionsCardView: View {
...
+                    displayCloseButton: false
+                )
+            )
+        })
+        .task(priority: .userInitiated) {
+            await refreshOffering()
+        }
+    }
+
+    private func refreshOffering() async {
```
vegaro: It feels like this should live in a viewModel, maybe you can pass the one we have in this view?

RevenueCatUI/CustomerCenter/ViewModels/NoSubscriptionsCardViewModel.swift
```diff
@@ -0,0 +1,75 @@
...
+@available(watchOS, unavailable)
+final class NoSubscriptionsCardViewModel: ObservableObject {
+
+    @Published var offering: Offering?
+    @Published var isLoadingOffering = true
+    @Published var showOffering = false
+
+    private let screenOffering: CustomerCenterConfigData.ScreenOffering?
+    private let purchasesProvider: CustomerCenterPurchasesType
+    init(
```
vegaro:
```suggestion
    private let screenOffering: CustomerCenterConfigData.ScreenOffering?
    private let purchasesProvider: CustomerCenterPurchasesType

    init(
```

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -174,6 +175,8 @@ import Foundation
 
             @_spi(Internal) public var defaultValue: String {
                 switch self {
+                case .subscribe:
+                    return "Subscribe"
```
vegaro, line 179: Same comment as in the Android PR, I wonder if we should use something else here

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -746,6 +775,21 @@ extension CustomerCenterConfigData.Screen {
         self.title = response.title
         self.subtitle = response.subtitle
         self.paths = response.paths.compactMap { CustomerCenterConfigData.HelpPath(from: $0) }
+        self.offering = response.offering.map { offering in
+            switch offering.type {
+            case "SPECIFIC":
```
vegaro: Can we reference the enum here to not hardcode the string?

Sources/CustomerCenter/CustomerCenterConfigData.swift
facumenzella: I think we can add a textfield in the dashboard to edit the copy, if more people start requesting it

PR #5459, opened 2025-08-08
Rename Test Store to Simulated Store internally

ajpallares:
### Motivation
The "test" term can be easily confused between the Test Store and the test code (e.g. unit tests).

### Description

Rename to conform with new internal naming (Simulated Store). 
Android counterpart: https://github.com/RevenueCat/purchases-android/pull/2591

## Important 
Public-facing logs, messages and APIs keep using the Test Store naming.

### DIFF
--- a/Examples/testCustomEntitlementsComputation/Project.swift
+++ b/Examples/testCustomEntitlementsComputation/Project.swift
@@ -38,7 +38,11 @@ let project = Project(
                 .revenueCat,
                 .revenueCatUI,
             ],
-            settings: .appTarget
+            settings: .appTarget(including: [
+                    "APPLICATION_EXTENSION_API_ONLY": "YES",
+                    "SWIFT_ACTIVE_COMPILATION_CONDITIONS": "$(inherited) ENABLE_CUSTOM_ENTITLEMENT_COMPUTATION"
+                ]
+            )
         )
     ],
 )
--- a/Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation.xcodeproj/project.pbxproj
+++ b/Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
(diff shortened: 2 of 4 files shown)


ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -141,7 +141,7 @@ import Foundation
             case storeAppStore = "app_store"
             case storeMacAppStore = "mac_app_store"
             case storePlayStore = "google_play_store"
-            case testStore = "test_store"
+            case simulatedStore = "simulated_store"
```
ajpallares, line 145: If I'm not mistaken, this has not been added in Android yet.

I don't know if it's useful to add this now as I don't think we're getting this information for Customer Center yet, but it can be good for future proofing I guess. Also, I think "simulated_store" would be the decided String for this. We could change this at any time if needed. WDYT @RevenueCat/coresdk?

JayShortway: `simulated_store` remains private right? If so, I think this makes sense indeed.

ajpallares: That's correct!

ajpallares:
bump @RevenueCat/coresdk

Sources/Purchasing/SimulatedStore/SimulatedStoreProduct.swift
```diff
@@ -0,0 +1,17 @@
...
+//
+//      https://opensource.org/licenses/MIT
+//
+//  SimulatedStoreProduct.swift
+//
+//  Created by Antonio Pallares on 6/8/25.
+
+import Foundation
+
+internal typealias SimulatedStoreProduct = TestStoreProduct
```
tonidero, line 16: Interesting... We could apply a similar typealias in Android to keep the internal code using the `Simulated` naming. I like it! 👍

PR #5461, opened 2025-08-08
Make TrialOrIntroEligibilityChecker @_spi public

ajpallares:
In order to allow previewing intro trials in Paywall Previews in the RC app, we need to allow the app to inject its own `TrialOrIntroEligibilityChecker`. That's why we're making `TrialOrIntroEligibilityChecker` `@_spi` public

### DIFF
--- a/RevenueCatUI/Data/IntroEligibility/TrialOrIntroEligibilityChecker.swift
+++ b/RevenueCatUI/Data/IntroEligibility/TrialOrIntroEligibilityChecker.swift
@@ -13,6 +13,7 @@
 
 import Foundation
 import RevenueCat
+import Combine
 
 @_spi(Internal) public final class TrialOrIntroEligibilityChecker: ObservableObject {
 
--- a/RevenueCatUI/Data/IntroEligibility/TrialOrIntroEligibilityChecker.swift
+++ b/RevenueCatUI/Data/IntroEligibility/TrialOrIntroEligibilityChecker.swift
@@ -14,9 +14,9 @@
 import Foundation
 import RevenueCat
 
-final class TrialOrIntroEligibilityChecker: ObservableObject {
+@_spi(Internal) public final class TrialOrIntroEligibilityChecker: ObservableObject {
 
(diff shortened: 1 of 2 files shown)


ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

RevenueCatUI/Data/IntroEligibility/TrialOrIntroEligibilityChecker.swift
```diff
@@ -11,12 +11,15 @@
 //
 //  Created by Nacho Soto on 7/13/23.
 
+import Combine
```
JayShortway, line 14: Curious why this was needed?

JayShortway (approved):
Makes sense! Just a question.

RevenueCatUI/Data/IntroEligibility/TrialOrIntroEligibilityChecker.swift
facumenzella: I think `ObservableObject` is part of Combine

JayShortway: Yea, but we were already using that haha, which is what made me curious.

ajpallares: It's as @facumenzella says. Because now this file uses `ObservableObject`, it needs to `import Combine`. Before, this file was not using any Combine APIs, so it didn't need to import it.

ajpallares: Adding the `@_spi(Internal) public` to `TrialOrIntroEligibilityChecker` without `import Combine` triggered this warning.

Image

JayShortway: Ahh, thanks for sharing that! So it was caused by making it public, interesting.

PR #5468, opened 2025-08-13
fix non-renewable appstore PATHs for customer center

facumenzella:
### Motivation
We've got a zendesk report for a non-renewable appstore transaction. We were showing CANCEL, even though it's not cancellable. This also applies for change plans.

### Description
- Store productType only for appstore products
- Update PATH logic

### DIFF
--- a/RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
+++ b/RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
@@ -40,8 +40,11 @@ struct PurchaseInformation {
     /// The store from which the purchase was made (e.g., App Store, Play Store).
     let store: Store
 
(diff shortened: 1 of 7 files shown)


RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
```diff
@@ -215,7 +220,10 @@ struct PurchaseInformation {
 
         self.customerInfoRequestedDate = customerInfoRequestedDate
         self.managementURL = managementURL
-        self.isSubscription = transaction.isSubscrition && transaction.store != .promotional
+        self.isSubscription = transaction.isSubscrition
```
vegaro: Just noticed there's a typo `subscrition`. Not for this PR though

vegaro (approved):

PR #5470, opened 2025-08-13
Extend appTarget settings to pass custom settings

facumenzella:
### Motivation
While going through the tuist set up once again, I noticed that `testCustomEntitlementsComputation` a project to test `RevenueCat_CustomEntitlementComputation` didn't have the settings.

### Description
- Extend `appTarget` to receive extra settingsDictionary
- Tweak `testCustomEntitlementsComputation` to include it for the only target it has

### DIFF
--- a/Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation.xcodeproj/project.pbxproj
+++ b/Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation.xcodeproj/project.pbxproj
@@ -3,7 +3,7 @@
 	archiveVersion = 1;
 	classes = {
 	};
-	objectVersion = 55;
+	objectVersion = 56;
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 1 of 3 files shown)


Examples/testCustomEntitlementsComputation/testCustomEntitlementsComputation.xcodeproj/project.pbxproj
ajpallares, line 1: I'm curious. Is this file being used in CI at all? If not, perhaps we could remove it and/or always have tuist generate it? 
Note that the .xcodeproj file is [already gitignored](https://github.com/RevenueCat/purchases-ios/blob/33bc578290782557ccd19e6f9087bc5d72560099/.gitignore#L83)

facumenzella: I think you're right. I can do that in another PR

PR #5472, opened 2025-08-13
Add NoSubscriptionsViewModel

facumenzella:
### Motivation
Adding a new file broke the old project.

### Description
This PR https://github.com/RevenueCat/purchases-ios/pull/5457

PR #5473, opened 2025-08-13
Test if new swift file throws a warning

facumenzella:
### Checklist


rickvdl:
@facumenzella This is great 😀 Do we want the PR check to fail if a file is missing since it might break installations?

facumenzella:
@rickvdl kind of. I wasn't sure about failing, mainly because I don't trust the script yet 😅 . If it works, I think we can make it more strict.

For context:

This https://github.com/RevenueCat/purchases-ios/pull/5472 broke the APITests.
Until we fully migrate to tuist, we will need to support both so this felt like an easy check. 

What do you think?

rickvdl:
@facumenzella Sounds good, I was also on the fence between that and making it fail right away, so let's do that 👍

RevenueCatUI/View+OnRedeemWebPurchaseAttempt copy.swift
```diff
@@ -0,0 +1,61 @@
+//
```
rickvdl: If this file was added for testing, should it be removed?

Dangerfile
```diff
@@ -5,3 +5,40 @@ submodules = `git submodule status`.strip
...
+def check_swift_files_in_project
+  project_file = 'RevenueCat.xcodeproj/project.pbxproj'
+  
+  unless File.exist?(project_file)
+    warn("RevenueCat.xcodeproj/project.pbxproj not found")
+    return
+  end
+  
+  # Get all added Swift files in this PR that belong to relevant directories
+  added_swift_files = git.added_files.select do |file|
```
JayShortway, line 19: Does `added_files` include renamed/moved files?

facumenzella: [Nope](https://github.com/danger/danger/blob/master/lib/danger/danger_core/plugins/dangerfile_git_plugin.rb#L54C1-L60C8), this only includes added files.

My reasoning is that new files can break part of the project because the project does not know they exist, but removed files will break it instantly because the project expects those to be there.

Also, handling renamed / removed will make this more complex. IF we need it, I can add it but this should catch most of the bugs

JayShortway: Yea, works for me! I was thinking we can grep for "added + renamed" files, but you're right that most bugs will be caused by (missing) added files. We can always revisit it later indeed.

PR #5062, opened 2025-05-07
[AUTOMATIC] Release/5.23.0

RCGitBot:
**This is an automatic release.**

### ✨ New Features
* Paywalls v2 sheet anysize (#5056) via Josh Holtz (@joshdholtz)
* Purchase web package (#5049) via Josh Holtz (@joshdholtz)

### ✨ New Features
* Add Bottom Sheet View (#5044) via Will Taylor (@fire-at-will)
#### ✨ New Features
* Fix new carousel transition (#5051) via Josh Holtz (@joshdholtz)
* [Paywalls v2] add stroke support in carousel dots (#5045) via Antonio Pallares (@ajpallares)
* feat: Introduce fade transition for loop animated CarouselView (#5047) via Facundo Menzella (@facumenzella)
#### 🐞 Bugfixes
* fix: Hide cancel & refundRequest if subscription is already cancelled (#5035) via Facundo Menzella (@facumenzella)

### 🔄 Other Changes
* use m1 pro large (#5060) via Mark Villacampa (@MarkVillacampa)
* use m2pro.large for all jobs (#5059) via Mark Villacampa (@MarkVillacampa)
* Add endpoint for fetching web billing products (#5055) via Toni Rico (@tonidero)
* Bump nokogiri from 1.18.5 to 1.18.8 in /Tests/InstallationTests/CocoapodsInstallation (#5054) via dependabot[bot] (@dependabot[bot])
* Bump fastlane from 2.225.0 to 2.227.2 (#5052) via dependabot[bot] (@dependabot[bot])
* Bump nokogiri from 1.18.7 to 1.18.8 (#5032) via dependabot[bot] (@dependabot[bot])
* [Paywalls] Remove url manipulation and alert from web paywall links (#5041) via Mark Villacampa (@MarkVillacampa)
* chore: Add right chevron to see all purchases (#5039) via Facundo Menzella (@facumenzella)
* other: Use not synced storekit config to avoid unwanted changes (#5037) via Facundo Menzella (@facumenzella)
* chore: Make StoreProductDiscountType @__spi public (#5034) via Facundo Menzella (@facumenzella)
* feat: Add missing pieces for maestro local execution (#5017) via Facundo Menzella (@facumenzella)

PR #5049, opened 2025-05-06
Purchase web package

joshdholtz:
### Motivation

Purchase web packages

### Description

- Adds `web_checkout_url` to package
- Adds `web_checkout_url` to offering
- `PurchaseButton` now has a few actions
  - purchase in-app
  - purchase web for selected product
  - purchase web and show product select on web thingy

PR #5474, opened 2025-08-13
Restore Old Package & Offering Initializer APIs

fire-at-will:
### Context
In release [5.23.0](https://github.com/RevenueCat/purchases-ios/pull/5062), we unintentionally broke the public API for the `Offering` and `Package` initializers by adding a non-optional `webCheckoutUrl: URL?` parameter.

The breaking changes were introduced in the these PRs/commits:
- https://github.com/RevenueCat/purchases-ios/pull/5049
- https://github.com/RevenueCat/purchases-ios/pull/5062/commits/5ff1c6fbf48b71c480ce18717750b780a1e30729

### Description
This PR restores the previous API by re-introducing new public initializers for these structs that omit the webCheckoutUrl parameter, matching the API contracts prior to 5.23.0. It also adds these re-introduced constructors to the API testers.

This PR introduces entirely new convenience initializers rather than adding a default value of nil to the existing ones, because initializers with default parameter values aren’t exposed to Objective-C. Simply adding a default would not restore the original Objective-C API.


fire-at-will:
@RCGitBot please test

Sources/Purchasing/Offering.swift
```diff
@@ -178,6 +179,25 @@ import Foundation
 
     // swiftlint:disable cyclomatic_complexity
 
+    /// Initialize an ``Offering`` given a list of ``Package``s.
+    @objc
+    public convenience init(
```
tonidero, line 184: I wonder, should we deprecate these constructors? Not sure if we want to push devs into just a single constructor... Maybe we should even try to make the "main" constructor internal to avoid these issues but we would need to keep the current ones public for backwards compatibility

Tests/APITesters/AllAPITests/SwiftAPITester/OfferingAPI.swift
```diff
@@ -46,19 +46,37 @@ func checkOfferingAPI() {
 }
 
 private func checkCreateOfferingAPI(package: Package) {
+    _ = Offering(
+        identifier: "",
+        serverDescription: "",
+        availablePackages: [package]
```
tonidero: The coloring always confuses me 😅. Very small nitpick but I wonder if we should rename to `rcPackage` to make it less confusing...

tonidero (approved):
This looks good! Just a couple of non-blocker comments.

Tests/APITesters/AllAPITests/SwiftAPITester/OfferingAPI.swift
fire-at-will: Hah, I'll make the change! The coloring is definitely distracting

fire-at-will: Updated in c0b38269b10ae605f00bc2aa307fbeaedd677c26

PR #5477, opened 2025-08-14
Create File Repository for use in upcoming feature work

JZDesign:
### Checklist

### Motivation

In the process of planning/writing the video component for paywalls, I noticed a need for a reusable cache for larger objects. This code is a lot on it's own, so I decided to break that work into smaller chunks so we can review this separately from the video component itself.

### Description

This introduces several new things to the project:

- A deferred value store: Essentially a way to share a task's execution  across call sites or over multiple invocations
- A file repository: A file cache that will fetch the data if the file is not found and then write the file returning the new address
- Test helpers for memory tracking and concurrency 

The goal of the repository is to have the ability to prewarm a URL and if that execution is not complete by the time a paywall is launched, that new call site will benefit from sharing the execution of that fetch task—so long as it's the same instance of the repository.

> [!NOTE] 
> This code supports swift 6, so there are some odd choices made with `@MainActor` for this reason

> [!NOTE] 
> This code is not used anywhere yet

### DIFF
--- a/Sources/Caching/FileRepository.swift
+++ b/Sources/Caching/FileRepository.swift
@@ -45,7 +45,7 @@ import Foundation
     @_spi(Internal) public func prefetch(urls: [InputURL]) {
         for url in urls {
             Task { [weak self] in
-                try await self?.getCachedURL(for: url)
+                try await self?.generateOrGetCachedFileURL(for: url)
             }
         }
     }
@@ -53,7 +53,7 @@ import Foundation
     /// Create and/or get the cached file url
     /// - Parameters:
     ///   - url: The url for the remote data to cache into a file
-    @_spi(Internal) public func getCachedURL(for url: InputURL) async throws -> OutputURL {
+    @_spi(Internal) public func generateOrGetCachedFileURL(for url: InputURL) async throws -> OutputURL {
         try await store.getOrPut(
             Task { [weak self] in
                 guard let self, let cachedUrl = self.fileManager.generateLocalFilesystemURL(forRemoteURL: url) else {
@@ -104,7 +104,7 @@ import Foundation
     /// Create and/or get the cached file url
     /// - Parameters:
     ///   - url: The url for the remote data to cache into a file
-    func getCachedURL(for url: InputURL) async throws -> OutputURL
+    func generateOrGetCachedFileURL(for url: InputURL) async throws -> OutputURL
 }
 
 /// The input URL is the URL that the repository will read remote data from
--- a/Tests/UnitTests/Caching/FileRepositoryTests.swift
+++ b/Tests/UnitTests/Caching/FileRepositoryTests.swift
@@ -21,7 +21,7 @@ class FileRepositoryTests: TestCase {
     func test_ifContentExists_networkServiceIsNotCalled() async throws {
         let sut = await makeSystemUnderTest()
         sut.cache.stubCachedContentExists(with: true)
-        let data = try await sut.fileRepository.getCachedURL(for: someURL)
+        let data = try await sut.fileRepository.generateOrGetCachedFileURL(for: someURL)
 
         XCTAssertNotNil(data)
         XCTAssertEqual(sut.networkService.invocations, [])
@@ -45,7 +45,7 @@ class FileRepositoryTests: TestCase {
     func test_whenCacheURLCannotBeAssembled_returnsNil() async throws {
         let sut = await makeSystemUnderTest(cacheDirectoryURL: nil)
         do {
-            _ = try await sut.fileRepository.getCachedURL(for: someURL)
+            _ = try await sut.fileRepository.generateOrGetCachedFileURL(for: someURL)
         } catch {
             switch error as? FileRepository.Error {
             case .failedToCreateCacheDirectory: break
@@ -63,7 +63,7 @@ class FileRepositoryTests: TestCase {
         sut.cache.stubCachedContentExists(with: false)
         sut.networkService.stubResponse(at: 0, result: .failure(SampleError()))
         do {
-            _ = try await sut.fileRepository.getCachedURL(for: someURL)
+            _ = try await sut.fileRepository.generateOrGetCachedFileURL(for: someURL)
             XCTFail(#function)
         } catch {
             switch error as? FileRepository.Error {
@@ -81,7 +81,7 @@ class FileRepositoryTests: TestCase {
         sut.cache.stubCachedContentExists(with: false)
         sut.cache.stubSaveData(with: .success(.init(data: data, url: someURL)))
         sut.networkService.stubResponse(at: 0, result: .success(data))
-        let result = try await sut.fileRepository.getCachedURL(for: someURL)
+        let result = try await sut.fileRepository.generateOrGetCachedFileURL(for: someURL)
 
         let expectedCachedURL = URL(string: "data:sample/someurl").unsafelyUnwrapped
 


Sources/Caching/KeyedDeferredValueStore.swift
```diff
@@ -0,0 +1,65 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  KeyedDeferredValueStore.swift
+//
+//  Created by Jacob Zivan Rakidzich on 8/12/25.
+
+import Foundation
+
+/// Some Task in which the value is Sendable and it can result in an error
+public typealias AnyTask<T: Sendable> = Task<T, Error>
```
facumenzella: I don't think this should be `public`. If we need this in `RevenueCatUI`, declare it as `@_spi(Internal) public`

facumenzella: Same for any other `public` that shouldn't be

Sources/Networking/HTTPClient/SimpleNetworkService.swift
```diff
@@ -0,0 +1,35 @@
...
+    /// - Returns: Data upon success
+    func data(from url: URL) async throws -> Data
+}
+
+extension URLSession: SimpleNetworkService {
+    /// Fetch data from the network
+    /// - Parameter url: The URL to fetch data from
+    /// - Returns: Data upon success
+    public func data(from url: URL) async throws -> Data {
+        let (data, response) = try await URLSession.shared.data(from: url)
```
facumenzella: I am confused with this one. This is already an instance method, but it still uses `URLSession.shared`. Shouldn't it use self?

Sources/Caching/FileRepository.swift
```diff
@@ -0,0 +1,110 @@
...
+    public init(networkService: SimpleNetworkService = URLSession.shared, fileManager: Caching = FileManager.default) {
+        self.networkService = networkService
+        self.fileManager = fileManager
+    }
+
+    /// Prefetch files at the given urls
+    /// - Parameter urls: An array of URL to fetch data from
+    public func prefetch(urls: [InputURL]) {
+        for url in urls {
+            Task(priority: .high) { [weak self] in
```
facumenzella: We don't have any `Task(priority: .high)` in the codebase (or at least I haven't found any). Shouldn't this be the default priority instead?

Sources/Caching/KeyedDeferredValueStore.swift
JZDesign: Thanks! I've addressed this feedback.

JZDesign: Screenshot 2025-08-14 at 10 40 54 AM
And now it's a lucky number

Sources/Caching/Caching.swift
```diff
@@ -0,0 +1,44 @@
...
+//
+//  Created by Jacob Zivan Rakidzich on 8/13/25.
+
+import Foundation
+
+/// An inteface representing a simple cache
+@_spi(Internal) public protocol Caching {
+
+    /// A URL for a cache directory if one is present
+    var cacheDirectory: URL? { get }
```
facumenzella: I wonder why we need the directory to be part of the protocol. This feels like implementation details to me.

Sources/Caching/FileRepository.swift
```diff
@@ -0,0 +1,113 @@
...
+import Foundation
+
+/// A file cache
+@_spi(Internal) public class FileRepository: @unchecked Sendable {
+    let networkService: SimpleNetworkService
+
+    private let store = KeyedDeferredValueStore<InputURL, OutputURL>()
+    private let fileManager: Caching
+
+    private lazy var cacheDirectory: URL? = fileManager.cacheDirectory
```
facumenzella: I think we can set this in the initializer, to avoid the laziness and make it inmutable, right?

facumenzella: Also, wouldn't making it inmutable make `FileRepository: Sendable`?

Sources/Networking/HTTPClient/SimpleNetworkService.swift
JZDesign: Copy paste during refactor issues… thanks for catching that

Sources/Caching/FileRepository.swift
JZDesign: Great question. That is really up for debate and I'm open to most things. What I wanted to make sure we did was prevent inheriting the priority from the app that is initializing/using the sdk. Prefetch is called after initialization, so if they decided to put it in as a priority low, then this would be a low priority too—if I understand swift concurrency anyway.

What are your thoughts on that?

Sources/Caching/Caching.swift
JZDesign: great point, I really should move the other function that is in the file repository that creates a cache url

Sources/Caching/FileRepository.swift
facumenzella: My main concern is that `high` is kind of aggressive, but I understand the need for this to run asap. I need to think about it. There are other "warmers" for paywalls, and I think they just run with the default priority so I guess we're good with that.

We can always increase it in the future.

What do you think @joshdholtz or @MarkVillacampa @fire-at-will ?

Sources/Caching/Caching.swift
```diff
@@ -0,0 +1,44 @@
...
+
+    /// Store data to a url
+    func saveData(_ data: Data, to url: URL) throws
+
+    /// Check if there is content cached at the given path
+    func cachedContentExists(at path: String) -> Bool
```
fire-at-will: Should `path` be a URL instead of a String to be consistent with the param type in the `saveData` function?

Sources/Caching/Caching.swift
```diff
@@ -0,0 +1,44 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  Caching.swift
+//
+//  Created by Jacob Zivan Rakidzich on 8/13/25.
+
+import Foundation
+
+/// An inteface representing a simple cache
+@_spi(Internal) public protocol Caching {
```
fire-at-will: Nit: We already have the `DeviceCache` class to store things in UserDefaults that doesn't conform to this protocol. Also, the naming convention in the repo for protocols is to usually name them in the format `name` + `Type` (like `DeviceCacheKeyType`, `StoreKit2PurchaseIntentListenerType`, `PurchasesType`, etc.). 

It's a bit clunky, but maybe we could name this protocol something like `DataCacheType`, `LargeDataCacheType`, or `FileCacheType`?

Sources/Networking/HTTPClient/SimpleNetworkService.swift
```diff
@@ -0,0 +1,35 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  SimpleNetworkService.swift
+//
+//  Created by Jacob Zivan Rakidzich on 8/12/25.
+
+import Foundation
+
+/// A protocol representing a simple network service
+@_spi(Internal) public protocol SimpleNetworkService {
```
fire-at-will: Nit: I'd consider naming this `SimpleNetworkServiceType` to be consistent with how other protocols in the codebase are named

Sources/Caching/FileRepository.swift
fire-at-will: I think I'd vote for using the default or even background priority level. At the end of the day, pre-fetching assets (and their URLs) is an optimization, not a mission-critical or user initiated task. If the app has a truly high priority task that it needs to execute, I don't think that prefetching our assets should consume CPU resources that could be allocated to the other high priority tasks.

Now, when we're actually trying to show the paywall, if the assets aren't already loaded, then it'd probably be appropriate to use a `.high` or `.userInitiaited` priority level

Sources/Caching/FileRepository.swift
```diff
@@ -0,0 +1,113 @@
...
+            let message = "Failed to save File to \(url.absoluteString): \(error)"
+            Logger.error(message)
+            throw Error.failedToSaveCachedFile(message)
+        }
+    }
+}
+
+extension FileRepository {
+    /// The input URL is the URL that the repository will read remote data from
+    @_spi(Internal) public typealias InputURL = URL
```
fire-at-will: LOVE using typealiases to ensure that we don't mix up input/output URLs 🤩👏🤩

Sources/Caching/FileRepository.swift
```diff
@@ -0,0 +1,113 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  FileRepository.swift
+//
+//  Created by Jacob Zivan Rakidzich on 8/13/25.
+
+import Foundation
+
+/// A file cache
+@_spi(Internal) public class FileRepository: @unchecked Sendable {
```
fire-at-will: Nit: I might consider making `FileRepository` conform to a `FileRepository` protocol so it's easier to mock out in tests in the future

Sources/Caching/FileRepository.swift
```diff
@@ -0,0 +1,113 @@
...
+            },
+            forKey: url
+        ).value
+    }
+
+    private func downloadFile(from url: URL) async throws -> Data {
+        do {
+            return try await networkService.data(from: url)
+        } catch {
+            let message = "Failed to download File from \(url.absoluteString): \(error)"
```
fire-at-will: Nit: 
```suggestion
            let message = "Failed to download file from \(url.absoluteString): \(error)"
```

fire-at-will: We usually try to extract log messages to a centralized location so they're easier to track down - check out the pattern in `Strings.swift`. I'd extract this out to a new category in there, or maybe add a new one or reuse an existing string from `NetworkStrings`

Sources/Caching/FileRepository.swift
```diff
@@ -0,0 +1,113 @@
...
+            Logger.error(message)
+            throw Error.failedToFetchFileFromRemoteSource(message)
+        }
+    }
+
+    private func saveCachedFile(url: URL, data: Data) throws {
+        do {
+            try fileManager.saveData(data, to: url)
+        } catch {
+            let message = "Failed to save File to \(url.absoluteString): \(error)"
```
fire-at-will: Same log comments from `downloadFile` apply here too

Sources/Caching/KeyedDeferredValueStore.swift
```diff
@@ -0,0 +1,74 @@
...
+            return wrapped
+        }
+        return result
+    }
+
+    /// Replaces a task in the store
+    /// - Parameter task: The new function to store
+    /// - Parameter key: The key to look up the result by
+    /// - Returns: The stored task
+    @_spi(Internal) public func replaceValue(
```
fire-at-will: Is this function required? I don't see it used outside of the tests yet

Sources/Caching/FileRepository.swift
JZDesign: We would have to make other things sendable… but yea that will help. I removed this url from this class though, it didn't belong

Sources/Caching/Caching.swift
JZDesign: Addressed

Sources/Caching/FileRepository.swift
```diff
@@ -0,0 +1,107 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  FileRepository.swift
+//
+//  Created by Jacob Zivan Rakidzich on 8/13/25.
+
+import Foundation
+
+/// A file cache
+@_spi(Internal) public class FileRepository: @unchecked Sendable {
```
JZDesign: Something to talk about here… The file repository relies on a `KeyedDeferredValueStore` which allows it to await the completion of a single Task from multiple call-sites, But that only works if it's the same instance of the `KeyedDeferredValueStore`

To ensure that this is setup correctly, we need to share either the file repository (and thus its instance of the deferred store), or the deferred value store itself.

I am leaning toward making a singleton of the `KeyedDeferredValueStore` and injecting it via the constructor to try and enforce that behavior when this is used in the sdk, but still allows for flexibility in the test suite. Or perhaps make an internal only constructor for testing that can inject the KeyedDeferredValueStore, and force the current constructor to use that shared instance.

What are your thoughts on that?

fire-at-will (commented):
Looking good!

One other question I have that didn't fit on a specific line - are we planning to use the cache in the core `RevenueCat` module anywhere, or only in paywalls? If we're only using it in paywalls/other UI products one day, it might make sense to move theses files to live in the `RevenueCatUI` module instead of the `RevenueCat` module.

Sources/Caching/KeyedDeferredValueStore.swift
JZDesign: It's not, but in personal projects it's used a bit… so I brought it in for thoroughness

Sources/Caching/FileRepository.swift
```diff
@@ -0,0 +1,107 @@
...
+    ///   - url: The url for the remote data to cache into a file
+    @_spi(Internal) public func getCachedURL(for url: InputURL) async throws -> OutputURL {
+        try await store.getOrPut(
+            Task(priority: .high) { [weak self] in
+                guard let self, let cachedUrl = self.fileManager.generateLocalFilesystemURL(forRemoteURL: url) else {
+                    Logger.error("Failed to create cache directory for \(url.absoluteString)")
+                    throw Error.failedToCreateCacheDirectory(url.absoluteString)
+                }
+
+                if fileManager.cachedContentExists(at: cachedUrl.path) {
```
fire-at-will: It might be worth adding in some functionality to mark an OutputURL on the disk as stale after a certain time period, and if we try to fetch the OutputURL for an OutputURL that is stale, then redownload the file and mark it as not being stale any longer. 

It's possible for a file to be replaced/updated on the server and still maintain the same URL. In this case, if we never invalidate cached files/local URLs, then I don't think the SDK would ever download the updated file.

JZDesign:
@fire-at-will 

> One other question I have that didn't fit on a specific line - are we planning to use the cache in the core RevenueCat module anywhere, or only in paywalls? If we're only using it in paywalls/other UI products one day, it might make sense to move theses files to live in the RevenueCatUI module instead of the RevenueCat module.

2 answers to your question:

1. There are paywall cache warming things in RevenueCat already, so this fits here by that pattern. Business logic separated from UI.

2. We need a better cache solution for images, we are currently using the URLCache and there are thoughts that this may be responsible for some visual defects and hangs that we are experiencing. Additionally we are doing filesystem—cache directory—storage for the imported fonts. So we could reuse this across all 3 of these use cases.

Sources/Caching/FileRepository.swift
JZDesign: That's a good point. I would rather pass nothing and let the call-site context apply than default to something low like `.background`

Sources/Caching/FileRepository.swift
JZDesign: @fire-at-will ?

Sources/Caching/FileRepository.swift
JZDesign: That's an interesting thought. I chose the cache directory so we don't have to do that right away. But we could have the server append a query parameter of last updated date to the url, that should allow us to store the Most recent file on device and not have to handle that logic.

Sources/Caching/FileRepository.swift
fire-at-will: To avoid using too many singletons, we usually store an instance of the object that could be a singleton on the `Purchases` object itself, since it's a singleton. Then, we pass down the object to any children of the `Purchases` class that need access to it. Since `FileRepository` is a class, it'll be passed down by reference and all children will be able to access the same instance of it. 

There are a few examples in `Purchases.swift`, lines 254-283 that should give you a good idea for the pattern!

Sources/Caching/KeyedDeferredValueStore.swift
```diff
@@ -0,0 +1,74 @@
...
+    ///   - task: The function that should execute if one is not found
+    ///   - key: The key to look up the result by
+    /// - Returns: The stored task
+    @_spi(Internal) public func getOrPut(
+        _ task: @escaping @Sendable @autoclosure () -> AnyTask<T>,
+        forKey key: H
+    ) -> AnyTask<T> {
+        guard let result = self.deferred[key] else {
+            let wrapped: AnyTask<T> = self.forgettingFailure(task, forKey: key)
+            self.deferred[key] = wrapped
```
fire-at-will, line 34: For long-running apps or scenarios with many unique tasks, it might be worth considering adding a cap on the size of the `deferred` dictionary to avoid this store from taking up too much memory. 

It looks like we remove entries for tasks that result in a failure, but since we never remove successful items, the growth potential of the deferred dictionary is uncapped. Once the size cap is reached, we could evict a previous entry, perhaps the least recently used entry?

JZDesign: Yes, I think that optimization can be included. I'll think on that.

JZDesign: I think—given the current use case—that I will hold off on implementing that for now while I stew on it more.

Sources/Caching/FileRepository.swift
JZDesign: I will follow that pattern when I wire things together. I will end up with 2 I think. One for fonts, and one for the rest.

Sources/Caching/FileRepository.swift
JZDesign: I've opened up a chat about this in the video component channel. I'm hoping to leverage the dashboard and the server implementation to handle this well without including that cache policy locally. But we will do what is feasible to support this once we nail that down. Thanks for bringing this up.

JZDesign: It sounds like we will move forward with a server side cache management pattern, we can talk details if you're interested.

Tests/UnitTests/Caching/FileRepositoryTests.swift
```diff
@@ -0,0 +1,127 @@
...
+
+        let expectedCachedURL = URL(string: "data:sample/someurl").unsafelyUnwrapped
+
+        XCTAssertEqual(sut.networkService.invocations, [someURL])
+        XCTAssertEqual(sut.cache.saveDataInvocations, [.init(data: data, url: expectedCachedURL)])
+        XCTAssertEqual(result, expectedCachedURL)
+        XCTAssertNotEqual(someURL, expectedCachedURL)
+    }
+
+    func makeSUT(
```
joshdholtz:
```suggestion
    func makeSystemUnderTest(
```

Small nit but just for clarity 😇 Odds are I will never remember what SUT means 🫠

Sources/Networking/HTTPClient/SimpleNetworkServiceType.swift
```diff
@@ -0,0 +1,35 @@
...
+
+/// A protocol representing a simple network service
+@_spi(Internal) public protocol SimpleNetworkServiceType {
+    /// Fetch data from the network
+    /// - Parameter url: The URL to fetch data from
+    /// - Returns: Data upon success
+    func data(from url: URL) async throws -> Data
+}
+
+@_spi(Internal) extension URLSession: SimpleNetworkServiceType {
```
joshdholtz: Do these need to be `@_spi`? 🤔 Or will these only be used internally only in `RevenueCat` for now? It _seems_ like these mainly be used by the `FileRepository` in the short term. If so, let's remove `@_spi` and then just added it back whenever its needed.

But let me know if I'm missing an immediate use case 😅

Sources/Caching/LargeItemCacheType.swift
```diff
@@ -0,0 +1,57 @@
...
+
+    /// Check if there is content cached at the given path
+    public func cachedContentExists(at url: URL) -> Bool {
+        (try? loadFile(at: url)) != nil
+    }
+
+    /// Generate a url for a location on disk based in the input URL
+    public func generateLocalFilesystemURL(forRemoteURL url: URL) -> URL? {
+        cacheDirectory?.appendingPathComponent(url.pathComponents.joined(separator: "/"))
+    }
```
joshdholtz: Do we want to log if `cacheDirectory` is `nil`? 🤔 

It seems bad/wrong if it is. We can probably assume it won't ever be but might be good to have a log to trace back just incase? 🤷‍♂️

Sources/Caching/KeyedDeferredValueStore.swift
```diff
@@ -0,0 +1,74 @@
...
+//
+//  Created by Jacob Zivan Rakidzich on 8/12/25.
+
+import Foundation
+
+/// Some Task in which the value is Sendable and it can result in an error
+@_spi(Internal) public typealias AnyTask<T: Sendable> = Task<T, Error>
+
+/// Holds onto ``AnyTask`` objects by key, autoclearing on failure
+@_spi(Internal) public actor KeyedDeferredValueStore<H: Hashable, T: Sendable> {
```
joshdholtz: Another question here about if all of this needs to be `@_spi`-ed or if this can just be private for now. It seems like `RevenueCatUI` would end up using these indirectly through a `FileRepository`, correct? If that's the case. I would try to limit what is `@_spi`-ed 😊

Sources/Caching/FileRepository.swift
```diff
@@ -0,0 +1,121 @@
...
+            Task { [weak self] in
+                try await self?.getCachedURL(for: url)
+            }
+        }
+    }
+
+    /// Create and/or get the cached file url
+    /// - Parameters:
+    ///   - url: The url for the remote data to cache into a file
+    @_spi(Internal) public func getCachedURL(for url: InputURL) async throws -> OutputURL {
```
joshdholtz: I think I would expect a method with the name of `getCachedURL` to get not "create or get" 🤔  I think based on that name, I would expect it to be a non-async method that doesn't have a `downloadFile` in it.

But that being said... I think this implementation makes sense! So I think if we could get a name that is matches doc of "Create and/or get the cached file url" would be 💯

joshdholtz (commented):
Left a few comments! Overall this is 🔥 I this is going to be super useful for videos, fonts, and other files for paywalls 💪 

Comments are mainly around `@_spi` usage and naming

Sources/Caching/KeyedDeferredValueStore.swift
JZDesign: This comment chain is why I did this. Open to changes, but I've got feedback in both directions.

https://github.com/RevenueCat/purchases-ios/pull/5477#discussion_r2276977004

Sources/Networking/HTTPClient/SimpleNetworkServiceType.swift
JZDesign: See other reply
https://github.com/RevenueCat/purchases-ios/pull/5477#discussion_r2284899741

Sources/Caching/KeyedDeferredValueStore.swift
joshdholtz: Ah okay! I totally agree with this not being `public` like in that thread 👍 

I think my comment was more along the lines of... can this be `internal` to `RevenueCat`? We don't have any initial need for this to be used by `RevenueCatUi` directly, correct?

From how I'm understanding things here, the `KeyedDeferredValueStore` will be used internally by `FileRepository`. And then `FileRepository` will be the one that needs to be `@_spi`-ed for `RevenueCatUI`?

Happy to be wrong here though! 😇

Sources/Caching/LargeItemCacheType.swift
JZDesign: Yes this is logged on line 61 of the file repository

joshdholtz (approved):
This looks good to me! LGTM :shipit:

PR #5485, opened 2025-08-18
Don't show subscriptions title if there are no subscriptions

facumenzella:
### Motivation
We are showing the `SUBSCRIPTIONS` section title even if there are no subscriptions.

### Description
- Show each title based on the content
- Show empty state if there are no purchases at all


RevenueCatUI/CustomerCenter/Views/RelevantPurchasesListView.swift
```diff
@@ -131,48 +130,44 @@ struct RelevantPurchasesListView: View {
...
-                        title: localization[.subscriptionsSectionTitle],
-                        items: customerInfoViewModel.subscriptionsSection,
-                        localization: localization
-                    ) {
-                        viewModel.purchaseInformation = $0
-                    }
-                    .tint(colorScheme == .dark ? .white : .black)
-                } else {
-                    ScrollViewSection(title: localization[.subscriptionsSectionTitle]) {
-                        NoSubscriptionsCardView(
```
vegaro, line 145: It looks like we've lost this `ScrollViewSection`. Is it needed?

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -137,7 +137,7 @@ final class CustomerCenterViewModelTests: TestCase {
 
         await viewModel.loadScreen()
 
-        expect(viewModel.hasPurchases).to(beTrue())
+        expect(viewModel.hasAnyPurchases).to(beTrue())
```
vegaro, line 140: It looks like we don't have a test that confirms this is true with only virtual currencies

RevenueCatUI/CustomerCenter/Views/RelevantPurchasesListView.swift
facumenzella: That view adds the title with the style that's needed, but the empty view does not need a title. That's part of the bug

vegaro (approved):

PR #5490, opened 2025-08-20
MON-1193 Optional transitions (delayed back button)

JZDesign:
### Checklist

### Motivation

Users want to be able to force the user to remain on the paywall for a short period of time so they don't reactively close the paywall immediately. 

### Description

Adds a transition type to the button component and renders that component with a transition if it's present

### DIFF
--- a/RevenueCat.xcodeproj/project.pbxproj
+++ b/RevenueCat.xcodeproj/project.pbxproj
@@ -58,7 +58,7 @@
 		164933242E5808BE00CE43A9 /* PaywallTransitionTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = 164933232E5808BE00CE43A9 /* PaywallTransitionTest.swift */; };
 		16A7F4C12E563B89001F9FC8 /* PaywallTransition.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16A7F4BF2E563B89001F9FC8 /* PaywallTransition.swift */; };
 		16A7F4C32E563B91001F9FC8 /* PaywallAnimation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16A7F4C22E563B91001F9FC8 /* PaywallAnimation.swift */; };
-		16A7F4C62E564B97001F9FC8 /* TransitionView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16A7F4C52E564B97001F9FC8 /* TransitionView.swift */; };
+		16A7F4C62E564B97001F9FC8 /* TransitionModifier.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16A7F4C52E564B97001F9FC8 /* TransitionModifier.swift */; };
 		16BCA3662E4D9AE400B39E7F /* LargeItemCacheType.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16BCA3652E4D9AE400B39E7F /* LargeItemCacheType.swift */; };
 		16BCA3682E4D9B0C00B39E7F /* FileRepository.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16BCA3672E4D9B0C00B39E7F /* FileRepository.swift */; };
 		16BCA36A2E4D9B2C00B39E7F /* KeyedDeferredValueStore.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16BCA3692E4D9B2C00B39E7F /* KeyedDeferredValueStore.swift */; };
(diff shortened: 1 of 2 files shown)


JZDesign:
@RCGitBot please test

JZDesign:
@RCGitBot please test

Sources/Paywalls/Components/Transitions/PaywallAnimation.swift
```diff
@@ -0,0 +1,94 @@
...
+//
+//  PaywallAnimation.swift
+//
+//  Created by Jacob Zivan Rakidzich on 8/20/25.
+// swiftlint:disable missing_docs
+
+import Foundation
+
+public extension PaywallComponent {
+    struct Animation: PaywallComponentBase {
```
vegaro, line 19: [nitpick] We normally have one empty line after type declarations, and one before the end of it https://github.com/RevenueCat/purchases-ios/blob/d98a15aa82cf6556eec4a1f0e45934edc868ea8c/Contributing/SwiftStyleGuide.swift#L18

Sources/Paywalls/Components/Transitions/PaywallTransition.swift
```diff
@@ -0,0 +1,115 @@
...
+    /// as the new view inserts itself.
+    ///
+    enum DisplacementStrategy: String, PaywallComponentBase {
+        case greedy, lazy
+    }
+
+    /// Defines the type of transition to use for paywall transitions.
+    ///
+    /// [NOTE] This is not a plain string enum because we see a future where we may want
+    /// to pass back more verbose instructions to the view layer than a simple enum case
```
vegaro: Does this mean the json is `"type": { "type": "fade" }`? Is that why you need the `TransitionTypeContainer`?

Sources/Paywalls/Components/Transitions/PaywallTransition.swift
```diff
@@ -0,0 +1,115 @@
...
+            )
+            .encode(to: encoder)
+        }
+    }
+
+    internal struct TransitionCodingContainer: Codable {
+        let type: TransitionTypeContainer
+        let displacementStrategy: DisplacementStrategy
+        let animation: PaywallComponent.Animation?
+    }
```
vegaro: I believe in other `PaywallComponent`s we have `CodingKeys` instead of a `CodingContainer` 

```swift
private enum CodingKeys: String, CodingKey {
    case type
    case displacementStrategy
    case animation
}
```

I think it's equivalent but double checking if it's worth it to use `CodingKeys` for consistency

Package.resolved
vegaro, line 1: I wonder why this file changed 🤔

Sources/Paywalls/Components/Transitions/PaywallAnimation.swift
```diff
@@ -0,0 +1,94 @@
...
+            self.msDuration = passthrough.msDuration
+        }
+
+        public func encode(to encoder: any Encoder) throws {
+            try AnimationCodingContainer(type: type.codingContainer, msDelay: msDelay, msDuration: msDuration)
+                .encode(to: encoder)
+        }
+    }
+
+    internal struct AnimationCodingContainer: Codable {
```
vegaro: Same question about `CodingContainer` vs using `CodingKeys` like in the other components

Package.resolved
facumenzella: We should discard this 🙏 
The project has nested Package.swift in some test projects (see PaywallTester => SnapshotTestingStub), and that usually causes this changes.

RevenueCatUI/Templates/V2/Components/Transition/TransitionView.swift
```diff
@@ -0,0 +1,114 @@
...
+
+    @State var isPresented: Bool = false
+
+    var body: some View {
+        ZStack {
+            if isPresented {
+                content()
+                    .transition(transition.toTransition)
+            } else {
+                if transition.displacementStrategy == .greedy {
```
rickvdl: Using a `switch` here should help catch newly added cases in the future. Or if we're 100% (can we ever be? haha) sure that there will never be another case it could be a bool instead maybe?

rickvdl (commented):
Small comment :)

Sources/Paywalls/Components/Transitions/PaywallAnimation.swift
JZDesign: No, the coding key is just the json key name, not the value type/structure. It's essentially the same as `@SerialName`

Sources/Paywalls/Components/Transitions/PaywallTransition.swift
JZDesign: Yes. This is to make it easier for us to support custom declarations in the future

Package.resolved
JZDesign: If this shouldn't get updated, should we have it gitignored?

facumenzella: It's not that it shouldn't be updated, but rather than Xcode somehow mixes everything in the same resolved file. Ultimately we should remove the sample projects from the legacy workspace, and this will stop happening

Sources/Paywalls/Components/Transitions/PaywallTransition.swift
JZDesign: After chatting with @joshdholtz I've flattened this, and we'll just modify the containing structure as needed.

PR #5491, opened 2025-08-21
Added APIs for hybrid SDKs to set presentedOfferingContext

rickvdl:
### Checklist

### Motivation

Based on PR [#2610](https://github.com/RevenueCat/purchases-android/pull/2610) for Android, adds similar internal APIs for the hybrid SDKs for iOS. Allows the hybrid SDKs to pass `PresentedOfferingContext` when launching paywalls by passing a specific offering

### Description

This PR implements the internal APIs, and uses the given `PresentedOfferingContext` when passing an `offeringIdentifier` from the hybrid SDKs to enrich the freshly fetched `Offering` with the given presentation context.

### DIFF
--- a/RevenueCat.xcodeproj/project.pbxproj
+++ b/RevenueCat.xcodeproj/project.pbxproj
@@ -68,6 +68,7 @@
 		16BCA37A2E4D9C0400B39E7F /* MockSimpleNetworkService.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16BCA3782E4D9C0400B39E7F /* MockSimpleNetworkService.swift */; };
 		16BCA37C2E4D9DD700B39E7F /* XCTestCase+Yield.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16BCA37B2E4D9DD700B39E7F /* XCTestCase+Yield.swift */; };
 		16BCA38D2E4E7B5C00B39E7F /* FileRepositoryStrings.swift in Sources */ = {isa = PBXBuildFile; fileRef = 16BCA38C2E4E7B5C00B39E7F /* FileRepositoryStrings.swift */; };
+		1DB9B2A72E57373900252D58 /* OfferingTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1DB9B2A62E57373600252D58 /* OfferingTests.swift */; };
 		1E2F911B2CC918ED00BDB016 /* ContactSupportUtilitiesTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1E2F911A2CC918ED00BDB016 /* ContactSupportUtilitiesTests.swift */; };
 		1E2F91722CCFA98C00BDB016 /* WebRedemptionStrings.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1E2F91712CCFA98C00BDB016 /* WebRedemptionStrings.swift */; };
 		1E42CC6C2D7F1A0500E0EE8D /* CacheStatus.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1E42CC6B2D7F1A0500E0EE8D /* CacheStatus.swift */; };
@@ -1528,6 +1529,7 @@
 		16BCA3782E4D9C0400B39E7F /* MockSimpleNetworkService.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = MockSimpleNetworkService.swift; sourceTree = "<group>"; };
 		16BCA37B2E4D9DD700B39E7F /* XCTestCase+Yield.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = "XCTestCase+Yield.swift"; sourceTree = "<group>"; };
 		16BCA38C2E4E7B5C00B39E7F /* FileRepositoryStrings.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = FileRepositoryStrings.swift; sourceTree = "<group>"; };
+		1DB9B2A62E57373600252D58 /* OfferingTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = OfferingTests.swift; sourceTree = "<group>"; };
 		1E2F911A2CC918ED00BDB016 /* ContactSupportUtilitiesTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ContactSupportUtilitiesTests.swift; sourceTree = "<group>"; };
 		1E2F91712CCFA98C00BDB016 /* WebRedemptionStrings.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = WebRedemptionStrings.swift; sourceTree = "<group>"; };
 		1E42CC6B2D7F1A0500E0EE8D /* CacheStatus.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = CacheStatus.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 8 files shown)


RevenueCatUI/PaywallView.swift
```diff
@@ -423,9 +424,14 @@ private extension PaywallViewConfiguration.Content {
...
-        case .defaultOffering: return Self.loadCachedCurrentOfferingIfPossible()
-        case let .offeringIdentifier(identifier): return Self.loadCachedOfferingIfPossible(identifier: identifier)
+        case let .offering(offering):
+            return offering
+        case .defaultOffering:
+            return Self.loadCachedCurrentOfferingIfPossible()
+        case let .offeringIdentifier(identifier, presentedOfferingContextOverride):
+            return Self.loadCachedOfferingIfPossible(
+                identifier: identifier
+            )?.withPresentedOfferingContext(presentedOfferingContextOverride)
```
JayShortway: I'd say it's not necessarily an "override". It's the original `PresentedOfferingContext` that would otherwise be lost if we just pass an id. 
```suggestion
        case let .offeringIdentifier(identifier, presentedOfferingContext):
            return Self.loadCachedOfferingIfPossible(
                identifier: identifier
            )?.withPresentedOfferingContext(presentedOfferingContext)
```

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -104,7 +105,33 @@ public class PaywallViewController: UIViewController {
...
+            fonts: fonts,
+            displayCloseButton: displayCloseButton,
+            shouldBlockTouchEvents: shouldBlockTouchEvents,
+            dismissRequestedHandler: dismissRequestedHandler
+        )
+    }
+    
+    /// Initialize a `PaywallViewController` with an offering identifier.
+    /// - Parameter offeringIdentifier: The identifier for the offering with paywall to display.
+    /// - Parameter presentedOfferingContext: Information about how the offering is presented..
```
JayShortway:
```suggestion
    /// - Parameter presentedOfferingContext: Information about how the offering is presented.
```

RevenueCatUI/UIKit/PaywallViewController.swift
```diff
@@ -161,9 +188,17 @@ public class PaywallViewController: UIViewController {
...
     /// - Warning: For internal use only
+    @available(*, deprecated, message: "use init with Offering instead")
     @objc(updateWithOfferingIdentifier:)
     public func update(with offeringIdentifier: String) {
-        self.configuration.content = .offeringIdentifier(offeringIdentifier)
+        self.configuration.content = .offeringIdentifier(offeringIdentifier, presentedOfferingContext: nil)
+    }
+    
+    /// - Warning: For internal use only
+    @_spi(Internal)
```
JayShortway, line 198: We don't strictly need the comment if we already have `@_spi(Internal)`. I think the comment is more of a leftover from before we used  `@_spi(Internal)`.

Sources/Purchasing/Offering.swift
```diff
@@ -291,6 +291,40 @@ import Foundation
...
+            draftPaywallComponents: draftPaywallComponents,
+            availablePackages: availablePackages.map { $0.withPresentedOfferingContext(presentedOfferingContext) },
+            webCheckoutUrl: webCheckoutUrl
+        )
+    }
+    
+    func withPresentedOfferingContext(_ presentedOfferingContext: PresentedOfferingContext?) -> Self {
+        guard let presentedOfferingContext else { return self }
+        return withPresentedOfferingContext(presentedOfferingContext)
+    }
```
JayShortway: Interesting! I guess this means we cannot set the `PresentedOfferingContext` to `nil` using this function. Reading the function signature I would expect that it allows me to set it to `nil`. Maybe we can do without this optimization?

Tests/APITesters/AllAPITests/RevenueCatUISwiftAPITester/PaywallViewControllerAPI.swift
```diff
@@ -6,7 +6,7 @@
 //
 
 import RevenueCat
-import RevenueCatUI
+@_spi(Internal) import RevenueCatUI
```
JayShortway: It's not strictly needed to test internal APIs, and it might be confusing if we do? If something breaks in this file, we need to figure out if we broke a public or internal API before we know how "bad" it is. Open to having my mind changed of course!

Tests/StoreKitUnitTests/OfferingTests.swift
```diff
@@ -0,0 +1,71 @@
...
+//  OfferingTests.swift
+//
+//  Created by Rick van der Linden on 21/08/2025.
+
+import Nimble
+@_spi(Internal) @testable import RevenueCat
+import XCTest
+
+@available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
+class OfferingTests: StoreKitConfigTestCase {
```
JayShortway, line 19: ❤️

RevenueCat.xcodeproj/project.pbxproj
JayShortway, line 1: (Not related to this file)

TIL that the `presentPaywall()` functions don't take an `offeringIdentifier`. Interesting discrepancy with Android's `launch()` functions.

RevenueCatUI/UIKit/PaywallFooterViewController.swift
```diff
@@ -43,12 +43,28 @@ public final class PaywallFooterViewController: PaywallViewController {
...
     ) {
-        super.init(content: .offeringIdentifier(offeringIdentifier),
+        super.init(content: .offeringIdentifier(offeringIdentifier, presentedOfferingContext: nil),
+                   fonts: DefaultPaywallFontProvider(),
+                   displayCloseButton: false,
+                   shouldBlockTouchEvents: false,
+                   dismissRequestedHandler: dismissRequestedHandler)
+    }
+    
+    /// Initialize a `PaywallFooterViewController` with an `Offering` identifier and a presentedOfferingContext
```
tonidero: NABD but not really needed to add docs for these `_spi` methods.

tonidero (approved):
Nothing to add to @JayShortway's comment. Looks great!! Thanks for doing this 🙌

Sources/Purchasing/Offering.swift
rickvdl: Fair point! I've removed it since it was just syntax sugar anyways :)

RevenueCatUI/PaywallView.swift
rickvdl: Agree :)

Tests/APITesters/AllAPITests/RevenueCatUISwiftAPITester/PaywallViewControllerAPI.swift
rickvdl: Ah good catch, we could have a separate file with internal API tests, but I'm not sure if that's worth the overhead. Indeed probably better to remove it for now :)

RevenueCat.xcodeproj/project.pbxproj
rickvdl: Same! Was surprised I couldn't find it

tonidero: I think PHC uses the PaywallViewController directly, instead of the present paywall functions, which are SwiftUI functions if I remember correctly. We don't use SwiftUI for the hybrids implementation, but the VC instead.

RevenueCatUI/UIKit/PaywallFooterViewController.swift
rickvdl: I removed it but it appears that SwiftLint disagrees, so I left it in for now

PR #5494, opened 2025-08-22
Fix/test flutter compilation

facumenzella:
### Motivation
We've got a report from flutter, that the codebase is not compiling with Xcode 14.2

### Description
We introduced a swift check [here](url) (a year ago). However, that check is not present through the customer center codebase, so it does not make much sense to keep it.

I'm also adding a few other things that were not compiling with Xcode 14.2


facumenzella:
@RCGitBot please test

facumenzella:
@RCGitBot please test

tonidero (approved):

PR #5500, opened 2025-08-26
Address virtual currencies with zero units in customer center

facumenzella:
### Motivation
We've noticed that there's an issue with virtual currencies and the empty state of customer center

### Description
- Consider the balance of VCs instead of the presence of a currency
- Fix ui glitch for empty state


RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -238,6 +242,15 @@ import Foundation
     }
 }
 
+private extension VirtualCurrencies {
+
+    var balanceIsZero: Bool {
+        all.reduce(into: 0) {
```
ekurutepe: nit: `all.map(\.value.balance).reduce(0,+)` ?

PR #5501, opened 2025-08-26
Add button_text to ScreenOffering

facumenzella:
### Motivation
We want to be more explicit on how to edit the button text for the ScreenOffering.

### Description
- Add button text to ScreenOffering

> This comes already translated. For backwards compatibility, we default to the key used right now.

### DIFF
--- a/Sources/CustomerCenter/CustomerCenterConfigData.swift
+++ b/Sources/CustomerCenter/CustomerCenterConfigData.swift
@@ -734,13 +734,16 @@ import Foundation
     @_spi(Internal) public struct ScreenOffering: Equatable {
         @_spi(Internal) public let type: OfferingType
         @_spi(Internal) public let offeringId: String?
+        @_spi(Internal) public let buttonText: String?
 
         @_spi(Internal) public init(
             type: OfferingType,
-            offeringId: String?
+            offeringId: String?,
+            buttonText: String?
         ) {
             self.type = type
             self.offeringId = offeringId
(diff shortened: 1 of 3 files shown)


Sources/CustomerCenter/CustomerCenterConfigData.swift
```diff
@@ -786,13 +789,15 @@ extension CustomerCenterConfigData.Screen {
...
-                    offeringId: offering.offeringId
+                    offeringId: offering.offeringId,
+                    buttonText: offering.buttonText
                 )
 
             default:
                 return CustomerCenterConfigData.ScreenOffering(
                     type: .current,
-                    offeringId: nil
+                    offeringId: offering.offeringId,
```
vegaro: should this still be nil? Or is `offering.offeringId` already `nil`?

facumenzella: If the type is current, we can ignore the id because we're using the current one from Offerings

vegaro: then we can pass `nil` here to be more explicit yeah

facumenzella: It will be null from the backend, but agree

facumenzella:
```suggestion
                    offeringId: nil,
```

vegaro (approved):

PR #5502, opened 2025-08-27
Fix dark mode button for featured offering + localized header

facumenzella:
### Motivation
1. We've noticed the dark mode featured button is not legible
2. The localized header of the empty state is not working
3. Follow-up of https://github.com/RevenueCat/purchases-ios/pull/5500 for `SubscriptionDetailView`

| Header | Header |
|--------|--------|
| Simulator Screenshot - iPhone 16 Pro Max - 2025-08-27 at 10 25 26 | Simulator Screenshot - iPhone 16 Pro Max - 2025-08-27 at 10 25 30 |

### Description
1. Set specific color for dark mode
2. Use screen title + subtitle if present

### On my mind
There's a weird state in `RelevantPurchasesListView`. We're showing the empty state header if there are no purchases, but that state should not happen. 
What do you think?

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/SubscriptionDetailView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/SubscriptionDetailView.swift
@@ -194,18 +194,7 @@ struct SubscriptionDetailView: View {
                         .animation(.easeInOut(duration: 0.3), value: viewModel.isRefreshing)
                 }
 
-                if let purchaseInformation = self.viewModel.purchaseInformation {
-                    PurchaseInformationCardView(
-                        purchaseInformation: purchaseInformation,
-                        localization: localization,
-                        accessibilityIdentifier: "0",
-                        refundStatus: viewModel.refundRequestStatus,
-                        showChevron: false
-                    )
-                    .cornerRadius(10)
-                    .padding(.horizontal)
-                    .padding(.vertical, 32)
-                } else {
+                if !customerInfoViewModel.hasAnyPurchases {
                     NoSubscriptionsCardView(
                         screenOffering: viewModel.screen.offering,
                         screen: viewModel.screen,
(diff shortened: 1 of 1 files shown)


vegaro:
> There's a weird state in RelevantPurchasesListView. We're showing the empty state header if there are no purchases, but that state should not happen.

What state @facumenzella ?

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsCardView.swift
```diff
@@ -50,12 +50,13 @@ struct NoSubscriptionsCardView: View {
 
     init(
         screenOffering: CustomerCenterConfigData.ScreenOffering?,
+        screen: CustomerCenterConfigData.Screen?,
```
vegaro, line 53: How can this be null? Maybe we should force a non-null here to simplify the code

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsCardView.swift
```diff
@@ -127,6 +128,11 @@ private struct BuySubscriptionButtonStyle: ButtonStyle {
 
     func makeBody(configuration: ButtonStyleConfiguration) -> some View {
         configuration.label
+            .foregroundStyle(
+                Color(colorScheme == .light
+                                  ? UIColor.systemBackground
+                                  : UIColor.secondarySystemBackground)
```
vegaro, line 134: Should this depend on the configured color?

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsCardView.swift
facumenzella: It can be null in the fallback screen, when there are no screen available 🤔

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsCardView.swift
facumenzella: Actually, that might make more sense

facumenzella: I've re-tested this, and I think the current solution makes more sense.
The text color is not editable in the dashboard, so makes sense to match the background of the container (having a clear color effect)

vegaro (approved):

PR #5509, opened 2025-08-29
Added the 'deepLinkID' property to PaywallComponentsData

rickvdl:
### Checklist

### Motivation

The  ~`id`~ `deepLinkID` of the paywall is needed in order to use universal links to link to the right paywall preview in the RC mobile app.

### Description

Implements the ~`id`~ `deepLinkID` property on the `PaywallComponentsData` type

### DIFF
--- a/Sources/Networking/Responses/RevenueCatUI/PaywallComponentsData.swift
+++ b/Sources/Networking/Responses/RevenueCatUI/PaywallComponentsData.swift
@@ -104,7 +104,7 @@ public struct PaywallComponentsData: Codable, Equatable, Sendable {
         case defaultLocale
         case assetBaseURL = "assetBaseUrl"
         case _revision = "revision"
-        case deepLinkID = "deep_link_id"
+        case deepLinkID = "deepLinkId"
     }
 
     public init(templateName: String,
--- a/Sources/Networking/Responses/RevenueCatUI/PaywallComponentsData.swift
+++ b/Sources/Networking/Responses/RevenueCatUI/PaywallComponentsData.swift
@@ -88,6 +88,10 @@ public struct PaywallComponentsData: Codable, Equatable, Sendable {
     public var componentsLocalizations: [PaywallComponent.LocaleID: PaywallComponent.LocalizationDictionary]
     public var defaultLocale: String
 
+    // The ID used for deeplinking
+    @_spi(Internal)
+    public let deepLinkID: String
+
     @DefaultDecodable.Zero
     internal private(set) var _revision: Int = 0
 
@@ -100,6 +104,7 @@ public struct PaywallComponentsData: Codable, Equatable, Sendable {
         case defaultLocale
         case assetBaseURL = "assetBaseUrl"
         case _revision = "revision"
+        case deepLinkID = "deep_link_id"
     }
 
     public init(templateName: String,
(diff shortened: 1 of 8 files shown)


Sources/Networking/Responses/RevenueCatUI/PaywallComponentsData.swift
```diff
@@ -73,6 +73,9 @@ public struct PaywallComponentsData: Codable, Equatable, Sendable {
         }
     }
 
+    // The unique id for this paywall
+    public let id: String
```
aboedo: for most of our entities we go with the longer `identifier` - see [Offering](https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/Offering.swift#L55) as an example. I'd recommend renaming for consistency

aboedo: we don't need to rename the internal stuff, but at least the public-facing property is good to keep consistent

aboedo (changes requested):
looking good but I left a comment about naming

Sources/Networking/Responses/RevenueCatUI/PaywallComponentsData.swift
rickvdl: Good one, agree!

joshdholtz: @aboedo This one I actually did have in mind that it is going to be an `id` and not an `identifier` 😅 

This is the `rc_public_id` which is pretty different than an offerings 'identifier` (user defined and mutable). Since its a completely different value and usage, I would still suggest this be `id` and not `identifier` but... happy to be overwritten 😇

joshdholtz (changes requested):
Not _really_ requesting changes but I want to have a full approval of using `id` vs `identifier` here 😇 

Offering's actually have an `identifier` where this is a paywall's `rc_public_id` which, IMO, is more of an `id` 🤷‍♂️ 

cc: @aboedo

Sources/Networking/Responses/RevenueCatUI/PaywallComponentsData.swift
```diff
@@ -73,6 +73,9 @@ public struct PaywallComponentsData: Codable, Equatable, Sendable {
         }
     }
 
+    // The unique identifier for this paywall
+    public let identifier: String
```
joshdholtz: @rickvdl Chatted with @aboedo during our 1:1 and we decided to call this... `deep_link_id` and we will `@_spi(Internal)` it 😊

Sources/Networking/Responses/RevenueCatUI/PaywallComponentsData.swift
```diff
@@ -88,6 +88,10 @@ public struct PaywallComponentsData: Codable, Equatable, Sendable {
     public var componentsLocalizations: [PaywallComponent.LocaleID: PaywallComponent.LocalizationDictionary]
     public var defaultLocale: String
 
+    // The ID used for deeplinking
+    @_spi(Internal)
```
tonidero, line 92: Hmm in android, the whole class is "@_spi", though I guess changing that now is a breaking change, so this might be the best compromise.

Sources/Networking/Responses/RevenueCatUI/PaywallComponentsData.swift
```diff
@@ -174,6 +182,13 @@ extension PaywallComponentsData {
             _revision = 0
         }
 
+        do {
+            deepLinkID = try container.decode(String.self, forKey: .deepLinkID)
+        } catch {
+            errors["deep_link_id"] = .init(error)
```
tonidero, line 188: [Not related to this PR!]
It's interesting how these errors are structured as a map. In android, I think these are a list. But I think this works as well 🤷

tonidero (approved):
Looking good to me!

rickvdl:
This PR is no longer necessary since the RC mobile app has been refactored to use a different API

PR #5511, opened 2025-08-29
Refetch offerings when preferred locale is set

joshdholtz:
### Motivation

Refetch current offerings (no cache) when preferred locale is set

### Description

Use a rate limiter when attempting to fetch current offerings when setting preferred locale (2 times per minute)

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -1931,6 +1931,10 @@ extension Purchases {
 
     // swiftlint:disable missing_docs
     @_spi(Internal) public func overridePreferredLocale(_ locale: String?) {
+        guard locale != self.systemInfo.preferredLocaleOverride else {
+            return
+        }
+
         self.systemInfo.overridePreferredLocale(locale)
 
         if self.overridePreferredUILocaleRateLimiter.shouldProceed() {


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1931,6 +1932,13 @@ extension Purchases {
     // swiftlint:disable missing_docs
     @_spi(Internal) public func overridePreferredLocale(_ locale: String?) {
         self.systemInfo.overridePreferredLocale(locale)
+
+        if self.overridePreferredUILocaleRateLimiter.shouldProceed() {
+            // Refetches new offerings with preferred locale
+            self.getOfferings(fetchPolicy: .default, fetchCurrent: true) { _, _ in
+                // No-op
+            }
+        }
```
ajpallares, line 1945: Perhaps we could trigger the `getOfferings` only when 
```swift
self.systemInfo.preferredLocaleOverride != locale
```

It's the most basic comparison, but at least it guards against developers setting to the already-set preferred locale.

JZDesign: I'd rather bump that responsibility to the networking layer, but I see that we are using an ephemeral session with no cache. And our e-tag cache isn't well suited for this… is there a reason we can't inject a flag to a request to use a non ephemeral session and set the cache policy to something really short like a minute?

JZDesign (approved):

Sources/Purchasing/Purchases/Purchases.swift
joshdholtz: > Is there a reason we can't inject a flag to a request to use a non ephemeral session and set the cache policy to something really short like a minute?

This is a question 🤔 I'm do not remember off the top of my head why we use the session and caching that we use. I used to know at one point but probably should understand again 🤷‍♂️ 

> I'd rather bump that responsibility to the networking layer

I'll leave it here for now since it short circuits it right at the start without touching any other layers and so its explicit that we aren't doing any extra networking calls to our API

PR #5512, opened 2025-09-01
Fix Paywall from Customer Center when `purchasesAreCompletedBy == .myApp`

ajpallares:
### Motivation
When the SDK is configured with `purchasesAreCompletedBy: .myApp`, `PaywallView` requires non-nil `performPurchase` and `performRestore` handlers at initialization. See the check: https://github.com/RevenueCat/purchases-ios/blob/c8f11707486a3b22b79ab87ac9742ff96e72cbb3/RevenueCatUI/PaywallView.swift#L174-L201

As a consequence, the paywall shown from Customer Center would raise an error when the SDK is configured with `purchasesAreCompletedBy: .myApp`
image

### Description
This PR passes `performPurchase` and `performRestore` handlers to Customer Center-initialized `PaywallView` to cover the `purchasesAreCompletedBy: .myApp` case. 

Note that these handlers are ignored when `purchasesAreCompletedBy: .revenueCat`

### DIFF
--- a/RevenueCatUI/CustomerCenter/Abstractions/CustomerCenterPurchasesType.swift
+++ b/RevenueCatUI/CustomerCenter/Abstractions/CustomerCenterPurchasesType.swift
@@ -43,6 +43,21 @@ import SwiftUI
     func promotionalOffer(forProductDiscount discount: StoreProductDiscount,
                           product: StoreProduct) async throws -> PromotionalOffer
 
+/// Initiates a purchase.
+///
+/// This method is used in two main contexts:
+/// - **PromotionalOfferViewModel**: To handle promotional offers that a customer is
+///   eligible for.
+/// - **NoSubscriptionCardView**: To serve purchases directly through your paywall UI.
+///
+/// If a `PromotionalOffer` is provided, the system will attempt to apply it.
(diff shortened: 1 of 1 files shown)


RevenueCatUI/CustomerCenter/Abstractions/CustomerCenterPurchasesType.swift
```diff
@@ -45,7 +45,7 @@ import SwiftUI
 
     func purchase(
```
facumenzella, line 61:
```suggestion
/// Initiates a purchase.
///
/// This method is used in two main contexts:
/// - **PromotionalOfferViewModel**: To handle promotional offers that a customer is
///   eligible for.
/// - **NoSubscriptionCardView**: To serve purchases directly through your paywall UI.
///
/// If a `PromotionalOffer` is provided, the system will attempt to apply it.
/// Otherwise, a standard purchase flow is executed.
///
/// - Parameters:
///   - product: The `StoreProduct` the customer intends to purchase.
///   - promotionalOffer: An optional `PromotionalOffer` to apply
/// - Returns: A `PurchaseResultData` object containing the result of the purchase.
/// - Throws: An error if the purchase flow fails or is cancelled.
    func purchase(
```

PR #5513, opened 2025-09-01
Fix compilation error in Xcode 14

ajpallares:
### Motivation
After #5512, RevenueCatUI was not building in Xcode 14 due to a missing `return` statement.

### Description
This PR adds the missing `return` to fix compilation in Xcode 14


ajpallares:
@RCGitBot please test

facumenzella (approved):

PR #5514, opened 2025-09-01
Untrack testEntitlementsComputation xcodeproj

facumenzella:
### Motivation
@ajpallares [pointed out](https://github.com/RevenueCat/purchases-ios/pull/5470) that we're tracking this, when we shouldn't.

### Description
Untrack xcodeproj files


facumenzella:
@RCGitBot please test

PR #5519, opened 2025-09-03
Updating Customer Center UI to be ready for iOS 26

hiddevdploeg:
### Checklist

### Motivation
As the release of iOS 26 approaches (with the iPhone event announced for [Sept 9](https://www.apple.com/apple-events)), the current CustomerCenter design may start to feel outdated once the app is updated. To ensure we align with the new style from day one, this PR introduces the necessary updates.

### Description
- This update introduces changes that apply starting from iOS 26.0:
- Cells are now styled with increased roundness on iOS 26 to better align with the native look and feel.
- The dismiss button now uses the new `Button(role: .close)` API.
- Labels (e.g., _Active_) are presented inside capsules to reflect the updated rounded design language.
- The prominent button now uses a default shape

#### Additional improvements (affecting other iOS versions as well):

- The [correct system icon](https://developer.apple.com/design/human-interface-guidelines/icons) is now used for Copy in the UserID cell.
- The UX and design of the UserID cell have been refined: a dedicated button now allows users to copy the ID with a single tap _(available from iOS 17+)_.
- The user ID is now selectable, making it easier to copy manually.

**NOTE: I do not think this branch should be merged / released yet as availability checking for iOS 26 is only available in Xcode 26 and up.**

https://github.com/user-attachments/assets/2c98f9a4-690c-429a-a491-f00682700157
Simulator Screenshot - iPhone 16 - 2025-09-03 at 12 36 40
hments/assets/70ef6a70-3f77-465b-8991-bd0432cc6d28" />
Simulator Screenshot - iPhone 16 - 2025-09-03 at 13 58 33

### DIFF
--- a/RevenueCatUI/CustomerCenter/ButtonStyles.swift
+++ b/RevenueCatUI/CustomerCenter/ButtonStyles.swift
@@ -32,15 +32,27 @@ struct ProminentButtonStyle: PrimitiveButtonStyle {
         let background = Color.from(colorInformation: appearance.buttonBackgroundColor, for: colorScheme)
         let textColor = Color.from(colorInformation: appearance.buttonTextColor, for: colorScheme)
 
-        Button(action: { configuration.trigger() }, label: {
-            configuration.label.frame(maxWidth: .infinity)
-        })
-        .font(.body.weight(.medium))
-        .buttonStyle(.borderedProminent)
-        .controlSize(.large)
-        .buttonBorderShape(.roundedRectangle(radius: 16))
-        .applyIf(background != nil, apply: { $0.tint(background) })
-        .applyIf(textColor != nil, apply: { $0.foregroundColor(textColor) })
+        if #available(iOS 26.0, *) {
+            Button(action: { configuration.trigger() }, label: {
+                configuration.label.frame(maxWidth: .infinity)
+            })
+            .font(.body.weight(.medium))
+            .buttonStyle(.borderedProminent)
+            .controlSize(.large)
+            .applyIf(background != nil, apply: { $0.tint(background) })
+            .applyIf(textColor != nil, apply: { $0.foregroundColor(textColor) })
+        } else {
+            Button(action: { configuration.trigger() }, label: {
+                configuration.label.frame(maxWidth: .infinity)
+            })
+            .font(.body.weight(.medium))
+            .buttonStyle(.borderedProminent)
+            .controlSize(.large)
+            .buttonBorderShape(.roundedRectangle(radius: 16))
+            .applyIf(background != nil, apply: { $0.tint(background) })
+            .applyIf(textColor != nil, apply: { $0.foregroundColor(textColor) })
+        }
+        
     }
 }
 
@@ -93,26 +105,39 @@ struct DismissCircleButton: View {
     var customDismiss: (() -> Void)?
 
     var body: some View {
-        Button {
-            if let customDismiss {
-                customDismiss()
-            } else {
-                self.dismiss()
+        
+        if #available(iOS 26.0, *) {
+            Button(role: .close) {
+                if let customDismiss {
+                    customDismiss()
+                } else {
+                    self.dismiss()
+                }
             }
-        } label: {
-            Circle()
-                .fill(Color(uiColor: .secondarySystemFill))
-                .frame(width: 28, height: 28)
-                .overlay(
-                    Image(systemName: "xmark")
-                        .font(.system(size: 14, weight: .bold))
-                        .foregroundStyle(.secondary)
-                        .imageScale(.medium)
-                )
-            }
-        .buttonStyle(.plain)
-        .accessibilityIdentifier("circled_close_button")
-        .accessibilityLabel(Text(localization[.dismiss]))
+            .accessibilityIdentifier("circled_close_button")
+            .accessibilityLabel(Text(localization[.dismiss]))
+        } else {
+            Button {
+                if let customDismiss {
+                    customDismiss()
+                } else {
+                    self.dismiss()
+                }
+            } label: {
+                Circle()
+                    .fill(Color(uiColor: .secondarySystemFill))
+                    .frame(width: 28, height: 28)
+                    .overlay(
+                        Image(systemName: "xmark")
+                            .font(.system(size: 14, weight: .bold))
+                            .foregroundStyle(.secondary)
+                            .imageScale(.medium)
+                    )
+                }
+            .buttonStyle(.plain)
+            .accessibilityIdentifier("circled_close_button")
+            .accessibilityLabel(Text(localization[.dismiss]))
+        }
     }
 
 }
@@ -130,7 +155,7 @@ struct DismissCircleButtonToolbarModifier: ViewModifier {
         if navigationOptions.shouldShowCloseButton {
             content
                 .toolbar {
-                    ToolbarItem(placement: .navigationBarTrailing) {
+                    ToolbarItem(placement: .topBarTrailing) {
                         DismissCircleButton(customDismiss: navigationOptions.onCloseHandler)
                     }
                 }
--- a/RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
@@ -95,6 +95,16 @@ struct PurchaseInformationCardView: View {
     }
 
     var body: some View {
+        if #available(iOS 26.0, *) {
+            content
+                .cornerRadius(26)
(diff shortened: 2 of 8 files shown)


facumenzella:
@RCGitBot please test

kukukaka:
Design-wise looks great :)
Thanks @hiddevdploeg

aboedo:
> ‼️NOTE: I do not think this branch should be merged / released yet as availability checking for iOS 26 is only available in Xcode 26 and up.

What we've done historically for this is to check the available compiler version and _then_ the OS version, i.e.: 
```swift
#if swift(>=6.2)
    if #available(iOS 26, *) {
```

The first one is at compile-time so the logic is entirely skipped in older Xcode. 
Here's one example:
https://github.com/RevenueCat/purchases-ios/blob/4bec38569da4afd9463976d8b55425a0be704b4b/Sources/Support/StoreMessageType.swift#L59

RevenueCatUI/CustomerCenter/Views/ScrollViewSection.swift
```diff
@@ -82,13 +81,18 @@ private struct CardStyleModifier: ViewModifier {
...
+            content
+                .padding()
+                .background(Color(colorScheme == .light ? UIColor.systemBackground : UIColor.secondarySystemBackground), in: .rect(cornerRadius: 26))
+                .padding(.horizontal)
+        } else {
+            content
+                .padding()
+                .background(Color(colorScheme == .light ? UIColor.systemBackground : UIColor.secondarySystemBackground), in: .rect(cornerRadius: 10))
+                .padding(.horizontal)
+        }
```
aboedo: if the only difference is the radius, maybe we can make that a variable and then use it? it's a little tricky because of the `#available` but one thought is doing
```swift
private var cornerRadius: CGFloat {
    if #available(iOS 26.0, *) { 26 } else { 10 }
}

var body: some View {
    content
        .padding()
        .background(
            Color(uiColor: colorScheme == .light ? .systemBackground : .secondarySystemBackground),
            in: .rect(cornerRadius: cornerRadius)
        )
        .padding(.horizontal)
}
```

where if this comes up enough, we can put the `cornerRadius` into a reusable place

RevenueCatUI/CustomerCenter/Views/ScrollViewSection.swift
```diff
@@ -151,19 +158,50 @@ struct AccountDetailsSection: View {
...
+                    Spacer()
+                    Button(localization[.copy], systemImage: didCopyID ? "checkmark" : "document.on.document") {
+                        UIPasteboard.general.string = originalAppUserId
+                        withAnimation {
+                            didCopyID = true
+                        }
+                        
+                        DispatchQueue.main.asyncAfter(deadline: .now() + 3) {
+                            withAnimation {
+                                self.didCopyID = false
```
aboedo: :chefskiss:

aboedo: really nice touch

RevenueCatUI/CustomerCenter/ButtonStyles.swift
```diff
@@ -32,15 +32,27 @@ struct ProminentButtonStyle: PrimitiveButtonStyle {
...
+                configuration.label.frame(maxWidth: .infinity)
+            })
+            .font(.body.weight(.medium))
+            .buttonStyle(.borderedProminent)
+            .controlSize(.large)
+            .buttonBorderShape(.roundedRectangle(radius: 16))
+            .applyIf(background != nil, apply: { $0.tint(background) })
+            .applyIf(textColor != nil, apply: { $0.foregroundColor(textColor) })
+        }
+
```
aboedo: similar thinking here, since we already have the `applyIf` i.e.: 

```swift
private var needsLegacyBorderShape: Bool {
    if #available(iOS 26.0, *) { false } else { true }
}

Button(action: { configuration.trigger() }) {
    configuration.label.frame(maxWidth: .infinity)
}
.font(.body.weight(.medium))
.buttonStyle(.borderedProminent)
.controlSize(.large)
.applyIf(needsLegacyBorderShape) { $0.buttonBorderShape(.roundedRectangle(radius: 16)) }
.applyIf(background != nil) { $0.tint(background) }
.applyIf(textColor != nil) { $0.foregroundColor(textColor) }
```

hiddevdploeg:
> @aboedo 
> I think we can make this mergeable today (as opposed to after xcode 26 release) if we make it get along with the current xcode build, see the `#if swift` suggestion.
> 
> other than that maybe we can have less duplicated code with some computed vars, I left comments about it.

I was thinking this! Just wasn't sure if this is something we normally do 😊 thanks for the feedback! Will make some changes today 👍

RevenueCatUI/CustomerCenter/Utilities/CustomerCenterStylingUtilities.swift
```diff
@@ -0,0 +1,36 @@
...
+//
+//  Created by Hidde van der Ploeg on 04/09/2025.
+
+import SwiftUI
+
+#if os(iOS)
+
+/// A namespace for Customer Center UI styling utilities.
+/// Provides shared constants and styling helpers across the module.
+public enum CustomerCenterStylingUtilities {
```
facumenzella: This shouldn't be public, right? It's for internal use only 🤔

RevenueCatUI/CustomerCenter/Views/PurchaseCardView.swift
```diff
@@ -95,6 +95,15 @@ struct PurchaseInformationCardView: View {
     }
 
     var body: some View {
+        if #available(iOS 26.0, *) {
+            content
+                .cornerRadius(26)
+        } else {
+            content
+                .cornerRadius(10)
+        }
```
facumenzella:
```suggestion
            content
                .cornerRadius(CustomerCenterStylingUtilities.cornerRadius)
````

File: RevenueCatUI/CustomerCenter/Utilities/CustomerCenterStylingUtilities.swift (hunk 4 above)
Comment (hiddevdploeg) [reply RevenueCatUI/CustomerCenter/Utilities/CustomerCenterStylingUtilities.swift]: Oh good catch

File: RevenueCatUI/CustomerCenter/Views/ScrollViewSection.swift [hunk 6]
```diff
@@ -151,21 +152,49 @@ struct AccountDetailsSection: View {
...
+            HStack {
+                Text(localization[.userId])
+                Spacer()
+                Text(originalAppUserId)
+                    .textSelection(.enabled)
+            }
+            .contentShape(.rect(cornerRadius: 26))
+            .contextMenu {
+                Button(localization[.copy], systemImage: "document.on.document") {
+                    UIPasteboard.general.string = originalAppUserId
```
facumenzella, line 193: Now I hate this 😆

RevenueCatUI/CustomerCenter/Views/RelevantPurchasesListView.swift
```diff
@@ -210,22 +211,36 @@ struct RelevantPurchasesListView: View {
             .prefix(RelevantPurchasesListViewModel.maxNonSubscriptionsToShow))
     }
 
+    @ViewBuilder
     private var seeAllSubscriptionsButton: some View {
```
facumenzella, line 215: This is also in `SubscriptionDetailView` 👀 
We should apply the same changes. Maybe we can extract this into its own component?

facumenzella: This is visible if there's only one purchase. Otherwise, you first go through the list and `seeAllSubscriptionsButton` won't be shown in the detail

hiddevdploeg: I thought about this but they're quite different: (Different viewModel, different background (buttonStyle vs background/cell) and different paddings)

facumenzella (approved):
sweet 🍬

RevenueCatUI/CustomerCenter/ButtonStyles.swift
```diff
@@ -27,6 +27,13 @@ struct ProminentButtonStyle: PrimitiveButtonStyle {
 
     @Environment(\.appearance) private var appearance: CustomerCenterConfigData.Appearance
     @Environment(\.colorScheme) private var colorScheme
+    private var needsLegacyBorderShape: Bool {
+        #if swift(>=6.2)
+        if #available(iOS 26.0, *) { false } else { true }
```
aboedo, line 32: why does this one need the `#if swift(>=6.2)` but the other ones don't?

PR #5523, opened 2025-09-05
Move preferred locale APIs to RevenueCat from RevenueCatUI

ajpallares:
This is a follow-up on #5292

After some discussions on https://github.com/RevenueCat/purchases-hybrid-common/pull/1267, we have decided to move the preferred locale APIs to RevenueCat (they were part of RevenueCatUI before).

This means the APIs to set the preferred locale for the UI now only require `import RevenueCat`:

```swift
import RevenueCat

let builder = Configuration
        .builder(withAPIKey: "")
        .with(preferredUILocaleOverride: "de_DE")
```
```swift
import RevenueCat

Purchases.shared.overridePreferredUILocale("de_DE")
```


ajpallares:
@RCGitBot please test

tonidero (approved):
Looks good to me! Thanks for doing this! 🙌

PR #5524, opened 2025-09-05
Fix compilation of RevenueCatUI in Xcode 14

ajpallares:
After the changes in #5519, there are some compilation errors in RevenueCatUI when using Xcode 14.

This PR fixes compilation of RevenueCatUI in Xcode 14.

## Notes
This PR adds some `#if compiler(>=5.9)` checks, disabling code not available in Xcode 14. It only affects code in RevenueCatUI and functionally this should be seamless (as Xcode 14 is not supported for App Store distribution anymore).

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionButtonsView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionButtonsView.swift
@@ -61,10 +61,12 @@ struct ActiveSubscriptionButtonsView: View {
             }
         }
         .applyIf(tintColor != nil, apply: { $0.tint(tintColor) })
+        #if compiler(>=5.9)
         .background(Color(colorScheme == .light
                           ? UIColor.systemBackground
                           : UIColor.secondarySystemBackground),
                     in: .rect(cornerRadius: CustomerCenterStylingUtilities.cornerRadius))
+        #endif
     }
 
     private var tintColor: Color? {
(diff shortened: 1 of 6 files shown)


ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

tonidero (approved):
I do wonder if we want to drop support for older Xcode versions... But for now, I think this looks good!

RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionButtonsView.swift
```diff
@@ -61,10 +61,12 @@ struct ActiveSubscriptionButtonsView: View {
             }
         }
         .applyIf(tintColor != nil, apply: { $0.tint(tintColor) })
+        #if compiler(>=5.9)
         .background(Color(colorScheme == .light
```
hiddevdploeg, line 65: We might want to change this by removing the 'in: .rect())' part then and use just background + .cornerRadius() again 😅

ajpallares:
> I do wonder if we want to drop support for older Xcode versions... But for now, I think this looks good!

We do want, yes. **But** we still need to support Xcode 14 if we want to run tests on CI in older versions of iOS (e.g. iOS 14). As far as I know (and I tried), there's no way to install a iOS 14 simulator on Xcode 16.

RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionButtonsView.swift
ajpallares: IMO since this is never going to be executed in Xcode 14 (other than in CI tests), it's fine to have this compiler check and, in general, use more modern APIs

hiddevdploeg:
Fair point! LGTM then 😊

PR #5527, opened 2025-09-05
Video Component Views

JZDesign:
### Checklist

### Motivation

We have requests for video support.

### Description

This adds a video component for apple platforms: iOS, macOS, and watchOS. There is are differences in them. But the primary difference that is noticeable in the UX is that iOS is the only one that does not have a black background if the video aspect ratio is not the same as the view's and does not fill its size. 

Also, this PR adds support for shading convex and concave masked views

For the commit history, see #5486. Moved to this PR to be a clean merge to main since that one was branched off of another one that merged recently.

closes #5486

These GIFs show some examples of this working:

Simulator Screen Recording - iPhone 16e - 2025-09-26 at 16 17 43

Simulator Screen Recording - iPhone 16e - 2025-09-05 at 12 56 50

> [!TIP]
> To test this, run PR 6173 in the revenue cat app

> [!NOTE]
> macOS and watchOS will ship with the standard black background on the video component. I don't have a good solution for handling that right now. 

> [!NOTE]
> watchOS doesn't have a looper or the ability to hide the controls, though I did hack that in there. It's not going to be as stable


Sources/Paywalls/Components/PaywallV2CacheWarming.swift
```diff
@@ -87,9 +87,12 @@ extension PaywallComponentsData.PaywallComponentsConfig {
                 urls += carousel.pages.flatMap({ stack in
                     self.collectAllImageURLs(in: stack)
                 })
-            case .video:
-                // WIP: - prewarm cache
-                break
+            case .video(let video):
+                Task {
+                    // WIP: Need to inject the file repository so we can potentially share the work in the view layer
```
JZDesign: This is an optimization that we can make to share the download task if that download has not completed before the paywall displays.

Sources/Caching/LargeItemCacheType.swift
```diff
@@ -47,7 +47,10 @@ extension FileManager: LargeItemCacheType {
 
     /// Generate a url for a location on disk based in the input URL
     func generateLocalFilesystemURL(forRemoteURL url: URL) -> URL? {
-        return cacheDirectory?.appendingPathComponent(url.pathComponents.joined(separator: "/"))
+        return cacheDirectory?
+            .appendingPathComponent(url.absoluteString.asData.md5String)
+            .appendingPathExtension(url.pathExtension)
```
JZDesign: Optimization. Also the thing I did before was unstable because `//` is valid in web URLs but is invalid on iOS's file system. 🤦

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
```diff
@@ -0,0 +1,200 @@
...
+                    .task {
+                        self.imageSource = viewModel.imageSource
+                        let fileRepository = FileRepository()
+                        if let lowResUrl = style.lowResUrl {
+                            let lowResCachedURL = try? await fileRepository.generateOrGetCachedFileURL(for: lowResUrl)
+                            self.cachedURL = lowResCachedURL
+                        }
+                        let cachedURL = try? await fileRepository.generateOrGetCachedFileURL(for: style.url)
+                        self.cachedURL = cachedURL ?? style.url
+                        self.imageSource = nil
```
joshdholtz: I think we can try to load this right away without async now with the change I made to the image component? (It's merged into `main` now)

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
```diff
@@ -0,0 +1,200 @@
...
+                                size: size,
+                                with: style
+                            )
+                        } else if let imageSource, let imageViewModel = try? ImageComponentViewModel(
+                            localizationProvider: viewModel.localizationProvider,
+                            uiConfigProvider: viewModel.uiConfigProvider,
+                            component: .init(source: imageSource)
+                        ) {
+                            ImageComponentView(viewModel: imageViewModel)
+                        }
```
joshdholtz: One change I did in the `ImageComponentView` is use our known dimensions of the image to create a "clear image" that essentially acts as a placeholder while the image is loading. 

I think this will kind of inherit that but maybe we might and to abstract something like that out at some point 🤷‍♂️

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
```diff
@@ -0,0 +1,200 @@
...
+
+                    }
+                    .applyVideoWidth(size: style.size)
+                    .applyVideoHeight(size: style.size, aspectRatio: self.aspectRatio(style: style))
```
joshdholtz: This `applyVideoWidth` and `applyVideoHeight` might end up being somewhat duplicated on top of `ImageComponentView` (since it also has something of its own)

I don't think this will be an issue but just wanted to call it out 😇

RevenueCatUI/Templates/V2/Components/Video/VideoComponentViewModel.swift
```diff
@@ -0,0 +1,234 @@
...
+        case .circle:
+            return .circle
+        case .concave:
+            return .concave
+        case .convex:
+            return .convex
+        }
+    }
+
+}
```
joshdholtz: I think we have this somewhere else 🤔 This would maybe be brought out and shared 🤷‍♂️

RevenueCatUI/Templates/V2/Components/Video/VideoPlayerView.swift
```diff
@@ -0,0 +1,155 @@
...
+        }
+
+        var body: some View {
+            VideoPlayer(player: player)
+        }
+    }
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+#Preview("No controls, no loop") {
```
joshdholtz: For some reason, I got a feeling that `#Preview` doesn't work with our infra for some reason (maybe when this runs on older Xcodes) but I could be wrong 🙃

Sources/Caching/LargeItemCacheType.swift
joshdholtz: I merged a chain for this into `main` so... this can be removed and rebased 😇

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
JZDesign: IIRC, they are different enough to keep them separate, but I'll check to be sure.

JZDesign: Nope! You were right. Copy Pasta

RevenueCatUI/Templates/V2/Components/Video/VideoPlayerView.swift
JZDesign: Seems fine?

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
```diff
@@ -0,0 +1,157 @@
...
+        with style: VideoComponentStyle
+    ) -> some View {
+        video
+            .frame(maxWidth: calculateMaxWidth(parentWidth: size.width, style: style))
+            .fitToAspectRatio(
+                aspectRatio: aspectRatio(style: style),
+                contentMode: .fill, // This must be set to fill for the modifier to work correctly
+                containerContentMode: style.contentMode // the container is what truly controls this
+            )
+
```
joshdholtz: Nit: extra line right here 😇

JZDesign:
@RCGitBot please test

RevenueCatUI/Templates/V2/Components/Video/VideoPlayerView.swift
JZDesign: Dang, I forgot I needed to run the gitbot please test thing. Got stuff to fix!!

JZDesign: Codified that https://github.com/RevenueCat/purchases-ios/pull/5527/commits/74dc534537205f805da33476f31133cac2262a9f

JZDesign:
@RCGitBot please test

joshdholtz (approved):
LGTM!!! :shipit: :shipit: :shipit:

JZDesign:
@RCGitBot please test

RevenueCatUI/Modifiers/FitToAspectRatio.swift
```diff
@@ -22,15 +22,33 @@ extension Image {
...
+                contentMode: contentMode,
+                containerContentMode: containerContentMode
+            )
+        )
+    }
 }
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, *)
-private struct FitToAspectRatio: ViewModifier {
+internal struct FitToAspectRatio: ViewModifier {
```
ajpallares: I believe this change is not needed

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
```diff
@@ -0,0 +1,156 @@
...
+                        if let cachedURL = fileRepository.getCachedFileURL(for: style.url) {
+                            self.cachedURL = cachedURL
+                            self.imageSource = nil
+                        } else if let lowResUrl = style.lowResUrl {
+                            let lowResCachedURL = fileRepository.getCachedFileURL(for: lowResUrl)
+                            self.cachedURL = lowResCachedURL
+                            self.imageSource = nil
+                        } else {
+                            self.cachedURL = style.url
+                        }
```
ajpallares, line 120: Correct me if I'm wrong, but following this code path, I see that `imageSource` is never going to be `nonnil` with `cachedURL` being `nil`, which means that in the `ZStack` below, the code will never go in the `else if let imageSource` code path.

1. In `if let cachedURL = fileRepository.getCachedFileURL(for: style.url)`, `cachedURL != nil` --> The code does not enter the `else if let imageSource` code path.
2. In `else if let lowResUrl = style.lowResUrl`, `self.imageSource` is set to `nil` --> The code does not enter the `else if let imageSource` code path.
3. In `else`. `self.cachedURL = style.url`, where `style.url` is not `nil` --> The code does not enter the `else if let imageSource` code path.

Maybe I'm missing something?

RevenueCatUI/Templates/V2/Components/Video/VideoPlayerView.swift
```diff
@@ -0,0 +1,203 @@
...
+                shouldAutoPlay: shouldAutoPlay && !reduceMotion,
+                contentMode: contentMode,
+                loopVideo: loopVideo,
+                muteAudio: muteAudio
+            )
+        }
+#elseif os(watchOS)
+        // Hiding controls on watchOS is not officially supported by Apple,
+        // disabling hit testing will prevent the user from being able to get them to show.
+        // Additionally, setting the action at end to .none will prevent them from displaying before looping.
```
ajpallares, line 60: Nice workaround!

RevenueCatUI/Templates/V2/Components/Video/VideoPlayerViewNSView.swift
```diff
@@ -0,0 +1,104 @@
...
+        case .fill:
+            playerLayer?.videoGravity = .resizeAspectFill
+        }
+
+        if shouldAutoPlay {
+            avPlayer.play()
+        }
+
+        if !loopVideo {
+        }
```
ajpallares: Something missing or should this be removed?

ajpallares (commented):
Adding some more comments! But it looks great overall. Amazing job!

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
JZDesign: That's a fair callout, the code in the view layer below only shows the image view if there is not a video url cached, so that is working. But I can clear this up. It's not very clear

RevenueCatUI/Templates/V2/Components/Video/VideoPlayerViewNSView.swift
JZDesign: Yep… I missed cleaning that up

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
ajpallares: 🤔 but if my reasoning above is correct, it will never happen that `self.cachedURL == nil && self.imageSource != nil`, so effectively, the view code in `else if let imageSource` will never execute

JZDesign: AHHHH, I updated that later and I missed that. You're right. After setting the cached url in the final block we would cancel rendering the fallback image. I am thinking on how I want to handle this at the moment.

The idea I had for the file repository was to share the download of the video in multiple spots, so we can start the prefetch and pick up that task in the middle of the download later. That's only possible if we share the instance of that cache.

If we did that, I know how I want to handle it, if we don't do that… there are things I need to consider still.

JZDesign: Okay, I've made changes. The image view will show until the image view model is made nil. 



But we also did some things… @joshdholtz and I talked about this ahead of time. We made a singleton of the file repository—not usually my go to approach, but we agreed that it was appropriate here. The point of the shared instance is that the FileRepository is equipped to basically broadcast the downloads of the files to any listener that invokes the `generateOrGetCachedFileURL` method. Meaning—if we share the instance—100 different callers could all invoke that method and we would only download the file 1 time. If some of the callers come later, they can piggy back on the original download and either get the result immediately or not have to wait for the whole download. 👈 This is precisely why we chose to do that here.

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
```diff
@@ -0,0 +1,179 @@
...
+                                }
+                            }
+                        }
+
+                        if let cachedURL = fileRepository.getCachedFileURL(for: style.url) {
+                            self.cachedURL = cachedURL
+                            self.imageSource = nil
+                        } else if let lowResUrl = style.lowResUrl {
+                            let lowResCachedURL = fileRepository.getCachedFileURL(for: lowResUrl)
+                            self.cachedURL = lowResCachedURL ?? lowResUrl
```
ajpallares, line 115: Question: aren't we missing a `self.imageSource = nil` to prevent having the two properties be nonnil at the same time?

```suggestion
                            self.cachedURL = lowResCachedURL ?? lowResUrl
                            self.imageSource = nil
```

Now, in the ZStack, we're showing the both of them if they are not nil. I guess the video would cover the image, but I want to make sure.

RevenueCatUI/Templates/V2/ViewHelpers/SizeReader.swift
```diff
@@ -0,0 +1,52 @@
...
+//
+//  SizeReader.swift
+//
+//  Created by Jacob Zivan Rakidzich on 8/18/25.
+
+import Foundation
+import SwiftUI
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+private struct SizeReader: View {
```
ajpallares: I believe that we can achieve the same with the existing `onSizeChange`. It's not using a binding but I think. we rather use the ecisting one to avoid duplicated functionality. https://github.com/RevenueCat/purchases-ios/blob/b09aaef263626c706a121719ae3f44af33ae698f/RevenueCatUI/Modifiers/ViewExtensions.swift#L268-L279

Sources/Paywalls/PaywallCacheWarming.swift
```diff
@@ -231,6 +254,17 @@ private extension Offerings {
         )
     }
 
+    var allVideosInPaywalls: Set<URL> {
+        return .init(
+            self
+                .all
+                .values
+                .lazy
```
ajpallares, line 262: Correct me if I'm wrong, but if we're returning all of them, I think `lazy` is not needed here

.swiftlint.yml
```diff
@@ -45,6 +45,18 @@ custom_rules:
...
+    regex: "#Preview"
+    message: "The '#Preview' macro will not compile in the CI pipeline during deployment, use a PreviewProvider instead."
+    severity: error
+
+  no_assignments_from_switch:
+    name: "No Assignments from Switch Statements"
+    regex: " = switch "
+    message: "In older versions of xcode build tools, this is not supported and will fail in CI"
+    severity: error
+
```
ajpallares, line 59: Nice!

Sources/Paywalls/PaywallCacheWarming.swift
JZDesign: ~~Yea, you're right, I just kept that pattern. I'll delete it.~~

Actually, I learned a thing. According to Gemini, assigning the result of a lazy sequence to a Set there are memory optimizations

## Impact of .lazy in this Scenario:

Memory Efficiency: lazy still helps with memory efficiency. Even though all elements eventually need to be processed to build the Set, the intermediate collections that would have been created by compactMap and flatMap in an eager evaluation are avoided. Instead, elements are processed one-by-one and fed directly into the Set builder. This can be a significant memory saving if offeringsToPreWarm is large.

### Without .lazy:

offeringsToPreWarm.compactMap(\.paywall): Creates a new array of all Paywall objects.
...flatMap(\.allImageURLs): Creates another new array containing all URLs.
Set(...): Consumes this final array to build the Set.
At peak, you'd have three collections in memory: the original, the Paywall array, and the URL array.

### With .lazy:

offeringsToPreWarm.lazy.compactMap(\.paywall).flatMap(\.allImageURLs): This defines the recipe.
Set(...): As the Set asks for elements, compactMap and flatMap are applied to one offering at a time. The resulting URLs are immediately added to the Set. No large intermediate arrays are ever fully materialized.
At peak, you'd primarily have the original collection and the Set being built in memory.
Execution Time: The total execution time to process all elements and build the Set might be very similar whether you use lazy or not, especially for CPU-bound transformations. The lazy overhead itself is usually minimal. However, if the memory pressure from creating intermediate arrays (without lazy) would cause significant swapping or other memory-related performance issues, then lazy could indirectly lead to faster execution by preventing those issues.

## Summary for Set Assignment:

When assigning the result of a lazy sequence to a Set (or Array), the operations in the lazy chain will still be executed for all relevant elements. However, the primary benefit of lazy in this specific case is memory optimization. It avoids the creation of large, temporary intermediate arrays, which can be crucial for performance and stability when dealing with very large initial collections.

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
JZDesign: Technically… no, that's in the `resumeDownloadOfFullResolutionVideo` function that is called on the next line. But it is delayed a bit. And I didn't think about that before 😆 so good call out. I'll add that.

Sources/Paywalls/PaywallCacheWarming.swift
ajpallares: Wow... thanks for digging into this! I assumed it didn't do any harm but I never imagined that it could even be more efficient 🤯

PR #5528, opened 2025-09-06
Paywalls image loading from cache is now synchronous

joshdholtz:
## Motivation

Make image loading from cache synchronous. This was an issue that was more obvious on bottom sheet but will actually help performance on every paywall

## Description

`RemoteImage` is now using the new `FileRepository` as its initial synchronous method for finding cached images.

If not found (even though it should be there), `RemoteImage` will fallback to existing `URLSession` cache.

### Bonus Fixes

- Only add a transition in `RemoteImageView` if loading async
- Fix issue where `ImageComponentView` where showing a `GeometryReader` (to get parent width) prevented image from loading "synchronously" 
  - Fixed this by wrapping in `ZStack`

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
@@ -60,6 +60,10 @@ struct ImageComponentView: View {
                 )
 
                 ZStack {
+                    // We need the max width of the parent view an image of a fill or
+                    // fixed width doesn't push passed the bounds.
+                    //
+                    // Once we have the width once, we can remove the GeometryReader
                     if self.maxWidth == nil {
                         GeometryReader { proxy in
                             Color.clear
@@ -74,13 +78,15 @@ struct ImageComponentView: View {
                         lowResUrl: style.lowResUrl,
                         darkUrl: style.darkUrl,
                         darkLowResUrl: style.darkLowResUrl,
+                        // The expectedSize is important
+                        // It renders a clear image if actual image is being fetched
                         expectedSize: expectedSize
                     ) { (image, size) in
                         self.renderImage(
                             image,
                             size,
                             maxWidth: self.calculateMaxWidth(
-                                parentWidth: maxWidth ?? 0,
+                                parentWidth: self.maxWidth ?? 0,
                                 style: style
                             ),
                             with: style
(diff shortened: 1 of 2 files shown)


RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
```diff
@@ -54,7 +54,19 @@ struct ImageComponentView: View {
...
+                    height: self.imageSize(style: style).height
+                )
+
+                ZStack {
+                    GeometryReader { proxy in
+                        Color.clear
+                            .onAppear {
+                                self.maxWidth = proxy.size.width
+                            }
+                    }
```
JZDesign: Since we need to do this pretty often, I have a helper coming soon. Though I need to make it a modifier. May be good to use it in this PR too and we can just clean up the merge conflicts in my PR later:


https://github.com/RevenueCat/purchases-ios/blob/9e871323882527c55f491bc26851efb15039b31d/RevenueCatUI/Templates/V2/ViewHelpers/SizeReader.swift#L18-L52

joshdholtz: I think we have an existing view modifier for this that I tried to use but it broke things even more 😅 But I think this might have broken the Emerge snapshots so looking into now 😬

Sources/Paywalls/PaywallCacheWarming.swift
```diff
@@ -90,6 +90,15 @@ actor PaywallCacheWarming: PaywallCacheWarmingType {
 
         Logger.verbose(Strings.paywalls.warming_up_images(imageURLs: imageURLs))
 
+        // Preferred method - load with FileRepository
+        let fileRepository = FileRepository()
+        for url in imageURLs {
+            Task {
+                _ = try await fileRepository.generateOrGetCachedFileURL(for: url)
+            }
```
JZDesign: With images, we may want to consider using a TaskGroup because there could be so many image URLs to process. Though we would need to make sure that we configure the task group not to bail if one fails.

This can be an improvement in a future PR though.

JZDesign (dismissed):
# Ship It

Sources/Paywalls/PaywallCacheWarming.swift
JZDesign: Actually, let's put this in the other for loop (on line 102) so you don't need to loop over all the images twice, and since the url session cache that is in the image fetcher is the backup, you can do a `try? await` IMO

PR #5529, opened 2025-09-08
Feat/test fix butotnstyles

facumenzella:
### Motivation
This PR addresses https://github.com/RevenueCat/react-native-purchases/issues/1381

It was pointed out that the dismiss is not working for iOS 15. After debugging, I noticed the issue is related to using `Environment dismiss` the way we are using it.

I've tested this on iOS15, iOS16, iOS17 and iOS18

### Description
- DismissCircleButton now requires a non-optional onDismiss: () -> Void and does not read @Environment(.dismiss).
- CustomerCenterView still captures @Environment(.dismiss) and injects it into navigationOptions.onCloseHandler, guaranteeing a working handler for all Customer Center views.

### DIFF
--- a/RevenueCatUI/CustomerCenter/ButtonStyles.swift
+++ b/RevenueCatUI/CustomerCenter/ButtonStyles.swift
@@ -89,35 +89,31 @@ extension ButtonStyle where Self == CustomerCenterButtonStyle {
 @available(macOS, unavailable)
 @available(tvOS, unavailable)
 @available(watchOS, unavailable)
+/// A circular close button used in the Customer Center toolbar.
+///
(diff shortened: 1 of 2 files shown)


facumenzella:
@ajpallares I think there's something related to where the `Environment` was placed? And maybe it wasn't propagated properly 🤔

ajpallares:
> @ajpallares I think there's something related to where the `Environment` was placed? And maybe it wasn't propagated properly 🤔

Yep, it must be due to an implementation detail in iOS 15 that got fixed in iOS 16. Probably the environment in the button was being overridden, making the `dismiss` of the button not be the same as the main view's.

PR #5535, opened 2025-09-09
Add purchase cancelled alert in Purchase Tester app

ajpallares:
### Description
This PR adds a "User cancelled" alert when cancelling a purchase. 

image

### DIFF
--- a/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Helpers/ObserverModeManager.swift
+++ b/Tests/TestingApps/PurchaseTesterSwiftUI/Shared/Helpers/ObserverModeManager.swift
@@ -6,7 +6,7 @@
 //
 
 import RevenueCat
-
+import StoreKit
 import SwiftUI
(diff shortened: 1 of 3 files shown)


ajpallares:
> Hmm so code looks good... but I wonder, do we want to display this alert? It feels a cancellation of a purchase should just return to the previous screen and not return anything. I think it's enough if we just log this happening. Wdyt? (Don't have much context on the reason for the change though, so lmk!)

Yeah, that's a good point. And I like it. Since we're showing an alert in the other two cases (success and failure), I think cancellation can totally be only a log. 
Thanks for the suggestion @tonidero! Done in https://github.com/RevenueCat/purchases-ios/pull/5535/commits/00d40dac2cfebbcb84b2ad667b061fcaca999559

Btw, I did these changes for demo purposes. I realized that the app would still show the "Purchased!" alert when cancelling a purchase. So I wanted to make it more consistent

PR #5536, opened 2025-09-09
Fix customer center iOS 26

aboedo:
We added checks for swift 6.2 for iOS 26-specific designs, but we didn't update the package.swift so the compiler was always using 5.9. 
This updates so that it still uses 5.9 when it's the latest available, otherwise 6.2 if it's available.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,15 +1,6 @@
 {
   "pins" : [
     {
-      "identity" : "accessibilitysnapshot",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/EmergeTools/AccessibilitySnapshot.git",
-      "state" : {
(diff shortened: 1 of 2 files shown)


facumenzella (approved):

Package.resolved
vegaro, line 1: I think the changes to this file can be discarded? I believe it depends on how you open the project, but I still don't understand how they happen

vegaro (approved):

Package.resolved
aboedo: I didn't even notice I'd added that file, reverting. thanks!

facumenzella: Yup we can discard them, it depends how you open your xcode

hiddevdploeg (approved):

facumenzella:
Closing this in favor of https://github.com/RevenueCat/purchases-ios/pull/5537

PR #5537, opened 2025-09-09
Fix customer center for iOS 16: alternative approach

aboedo:
Alternative to #5536, only changes the `#if swift` with `#if compiler`


facumenzella:
@RCGitBot please test

ajpallares (approved):
Yep, that does it! Tested with the RC app and the project builds and Customer Center uses Liquid Glass in iOS 26

Image

PR #5546, opened 2025-09-10
[DENG-1362] Reduce repeated logs

ajpallares:
There were certain circumstances where the SDK was sending repeated logs multiple times. This could happen, for example, when showing a paywall that used a font that the SDK couldn't find.

> WARN 2025-08-08T07:44:03.033Z,RevenueCat Log: Mapping for 'my-custom-font' could not be found. Falling back to system font.

This PR fixes it so that each of these log messages are only sent once per paywall presentation.


RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
```diff
@@ -18,11 +18,12 @@ import SwiftUI
 #if !os(macOS) && !os(tvOS) // For Paywalls V2
 
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
-struct UIConfigProvider {
+class UIConfigProvider {
```
facumenzella:
```suggestion
final class UIConfigProvider {
```

RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
```diff
@@ -60,7 +61,7 @@ struct UIConfigProvider {
             fontName = fontsConfig.ios.value
         case .googleFonts:
             // Not supported on this platform (yet)
```
facumenzella: I think we can remove this comment, code is strongly typed clear now 😅

RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
```diff
@@ -83,4 +84,44 @@ struct UIConfigProvider {
...
+        case .localizationNotFound(let identifier):
+            Logger.error("Could not find localizations for '\(identifier)'")
+        case .fontMappingNotFound(let name):
+            Logger.warning("Mapping for '\(name)' could not be found. Falling back to system font.")
+        case .customFontFailedToLoad(let fontName):
+            Logger.warning("Custom font '\(fontName)' could not be loaded. Falling back to system font.")
+        case .googleFontsNotSupported:
+            Logger.warning("Google Fonts are not supported on this platform")
+        }
+    }
```
facumenzella, line 128: We need to move this with the other logged message strings.

facumenzella: Also, I guess there's no test? It would be great to add one

RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
ajpallares: Totally. Done in https://github.com/RevenueCat/purchases-ios/pull/5546/commits/91c7144e1f8819b1e8110ed9b13763cbb0f4b1fd

RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
```diff
@@ -89,4 +89,44 @@ struct UIConfigProvider {
...
+
+    enum LogMessage: Hashable {
+        case localizationNotFound(identifier: String)
+        case fontMappingNotFound(name: String)
+        case customFontFailedToLoad(fontName: String)
+        case googleFontsNotSupported
+    }
+
+    func logMessageIfNeeded(_ message: LogMessage) {
+        guard !self.loggedMessages.contains(message) else { return }
```
vegaro, line 107: is it a concern that we never clear this? I asume it's not because we are only fetching new offerings on a new SDK configuration

ajpallares: The lifespan of the `UIConfigProvider` is that of the Paywall. It's initialized in the init of `PaywallV2View`: https://github.com/RevenueCat/purchases-ios/blob/4c88db89e14047c070aea497054805d10557f865/RevenueCatUI/Templates/V2/PaywallsV2View.swift#L124-L127

This means that there's a new instance of `UIConfigProvider` for each presentation of a paywall. Therefore, the `loggedMessages` set is effectively cleared when the paywall gets dismissed.

Does it make sense?

vegaro: It makes a lot of sense. Perfect!

RevenueCatUI/Templates/V2/ViewModelHelpers/UIConfigProvider.swift
ajpallares: Great points! 
* Strings moved in https://github.com/RevenueCat/purchases-ios/pull/5546/commits/33884cd6cb91b39b9abb28a723c886698224f950
* Tests added in https://github.com/RevenueCat/purchases-ios/pull/5546/commits/a136ae2b9ae249ab5f067de4850ef41b5fbab984

Thank you!

Tests/RevenueCatUITests/Helpers/TestLogHandler.swift
```diff
@@ -0,0 +1 @@
+../../UnitTests/TestHelpers/TestLogHandler.swift
```
facumenzella, line 1: what's this file 🤔

ajpallares: It's a symlink to a file in RevenueCat unit tests. The `TestLogHandler` class is now needed in RevenueCatUI as well in order to test logs (needed to unit test the new behavior of the `UIConfigProvider` class). Instead or rewriting the implementation of `TestLogHandler`, using a symlink we can share the same source.

Note that most of the files in [`Tests/RevenueCatUITests/Helpers`](https://github.com/RevenueCat/purchases-ios/tree/main/Tests/RevenueCatUITests/Helpers) are also symlinks to files in RevenueCat unit tests (in fact, I copied the idea for `TestLogHandler` 😅)

JZDesign: I used to use symlinks to share code like this pretty regularly before SPM made multi module projects really easy 😆 But sometimes there's still a good reason to do it

facumenzella: Got it. I think that we should at some point remove tests from the main project, and run it with tuist to avoid all these. Hopefully soon 🤞

ajpallares:
Thanks for reviewing! I need to get back to this one, as the tests in RCUI are failing on CI. And I suspect it has to do with the symlinks... will check ASAP

ajpallares:
I'm not happy about it but in the end I had to duplicate the `TestLogHandler.swift` file. Originally, I added it as a symlink, but then all RevenueCatUI tests were failing. Replacing the symlink with an actual copy of the file does work, so, for now, I think we can move on with it.

PR #5552, opened 2025-09-11
Add iOS 26 tests to CI

ajpallares:
* Adds a `run-test-ios-26` CI job
* Includes the new `run-test-ios-26` job in the `build-test` and `generate-snapshot` workflows

> [!NOTE]
> The huge diff is due to the fact that some snapshots in unit tests had to be updated because iOS 26's `Color.accentColor` changed from `#007AFF` to `#0088FF`.
>  * As a result, I had to differentiate these snapshots per version of iOS.
>  * To avoid having exact duplicated files for iOS 15, 16, 17 and 18, I made the iOS 15, 16 and 17 files to be symlinks to the iOS 18 ones.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -526,7 +526,7 @@ jobs:
   spm-release-build-xcode-15:
     executor:
       name: macos-executor
-      xcode_version: "15.2.0"
+      xcode_version: "15.4.0"
 
     steps:
       - checkout
@@ -667,7 +667,7 @@ jobs:
   run-revenuecat-ui-ios-17:
     executor:
       name: macos-executor
-      xcode_version: "15.2.0"
+      xcode_version: "15.4.0"
 
     steps:
       - checkout
@@ -683,7 +683,7 @@ jobs:
           command: bundle exec fastlane test_revenuecatui
           no_output_timeout: 15m
           environment:
-            DEVICE: iPhone 15 Pro,OS=17.2
+            DEVICE: iPhone 15 Pro,OS=17.5
       - create-snapshot-pr-if-needed:
           condition: << pipeline.parameters.generate_revenuecatui_snapshots >>
           job: "create_snapshots_repo_pr"
@@ -870,7 +870,7 @@ jobs:
   run-test-ios-17:
     executor:
       name: macos-executor
-      xcode_version: "15.2.0"
+      xcode_version: "15.4.0"
 
     steps:
       - checkout
@@ -881,7 +881,7 @@ jobs:
           command: bundle exec fastlane test_ios
           no_output_timeout: 15m
           environment:
-            SCAN_DEVICE: iPhone 15 (17.2.0)
+            SCAN_DEVICE: iPhone 15 (17.5.0)
       - compress_result_bundle:
           directory: fastlane/test_output/xctest/ios
           bundle_name: RevenueCat
@@ -931,8 +931,7 @@ jobs:
   run-test-ios-15:
     executor:
       name: macos-executor
-      # Needs to be 15.2.0 because if not running on Swift 5.9, the tests fail
-      xcode_version: "15.2.0"
+      xcode_version: "15.4.0"
 
     steps:
       - checkout
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -1069,6 +1069,11 @@ platform :ios do
     # sh runs from the Fastfile's location, but other commands run from the project root.
     output_directory_for_sh = "../#{output_directory}"
 
+    # Post-process headers to suppress StoreKit 1 deprecation warnings
+    suppress_deprecation_warnings_in_headers(
+      xcframework_path: "#{output_directory_for_sh}/#{product_name}.xcframework"
+    )
+
     # Sign XCFramework for Xcode 15
     sign_xcframework(
       file: "#{output_directory_for_sh}/#{product_name}.xcframework"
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -779,6 +779,33 @@ jobs:
...
+      - run:
+          name: Run tests
+          command: bundle exec fastlane test_ios
+          no_output_timeout: 15m
+          environment:
+            SCAN_DEVICE: iPhone 16e (26.0)
+      - compress_result_bundle:
+          directory: fastlane/test_output/xctest/ios
+          bundle_name: RevenueCat
+      # - create-snapshot-pr-if-needed:
```
fire-at-will: Nit: remove commented code before merging

.circleci/config.yml
```diff
@@ -779,6 +779,33 @@ jobs:
...
+    steps:
+      - checkout
+      - install-dependencies:
+          install_swiftlint: false
+      - run:
+          name: Run tests
+          command: bundle exec fastlane test_ios
+          no_output_timeout: 15m
+          environment:
+            SCAN_DEVICE: iPhone 16e (26.0)
```
fire-at-will, line 828: If it's available, it might be worth using an iPhone 17 just so we're using the latest of everything

```suggestion
            SCAN_DEVICE: iPhone 17 (26.0)
```

.circleci/config.yml
```diff
@@ -1537,6 +1565,7 @@ workflows:
           mac_os_executor_name: macos-executor
       - run-test-ios-17
       - run-test-ios-18
+      - run-test-ios-26
```
fire-at-will, line 1600: Might be worth removing `run-test-ios-17` here so that we're only running tests the two most recent OSs in each PR, and let `run-test-ios-17` run just in the `all-test` job

Tests/StoreKitUnitTests/TestHelpers/AvailabilityChecks.swift
```diff
@@ -55,6 +55,12 @@ enum AvailabilityChecks {
         }
     }
 
+    static func iOS26APIAvailableOrSkipTest() throws {
+        guard #available(iOS 26.0, tvOS 26.0, macOS 26.0, watchOS 26.0, visionOS 26.0, *) else {
+            throw XCTSkip("Required API is not available for this test.")
+        }
+    }
+
```
fire-at-will, line 63: This can be done in the future when/if it's needed, but it might be helpful to add `iOS26APINotAvailableOrSkipTest()` since we need those sometimes

.circleci/config.yml
ajpallares: Right, I thought the same. But it's still not available https://circleci.com/docs/guides/execution-managed/using-macos/
At the time of writing this, the latest image is
image
which does not have any of the new iPhone 17 devices (see https://circle-macos-docs.s3.amazonaws.com/image-manifest/v15636/manifest.txt)
image
We can change it once CircleCI adds the release version of Xcode 26

.circleci/config.yml
ajpallares: Thanks for flagging! In fact, this should be uncommented so that the snapshot PR is created when appropriate. Done in https://github.com/RevenueCat/purchases-ios/pull/5552/commits/a24e6026adb6738b5ac51e5d4c2bffcb4c66bc64

.circleci/config.yml
ajpallares: Great point. Done in https://github.com/RevenueCat/purchases-ios/pull/5552/commits/34c775cd378dadefeba4ed9f6c1d0e937aa6597f
Thanks!

Tests/StoreKitUnitTests/TestHelpers/AvailabilityChecks.swift
ajpallares: Yeah, why not? Done in https://github.com/RevenueCat/purchases-ios/pull/5552/commits/e8fc107f6b09290d8202797ada53b0bc832c445a

.circleci/config.yml
```diff
@@ -1535,8 +1563,8 @@ workflows:
           name: validate-package-swift-5.8
           xcode_version: "14.3.1"
           mac_os_executor_name: macos-executor
-      - run-test-ios-17
       - run-test-ios-18
+      - run-test-ios-26
       - pod-lib-lint
       - run-revenuecat-ui-ios-17
       - run-revenuecat-ui-ios-18
```
vegaro, line 1602: do we need a `run-revenuecat-ui-ios-26`?

ajpallares: Yep! Totally missed it 👍

ajpallares: Done in https://github.com/RevenueCat/purchases-ios/pull/5552/commits/ff0174b65c41bb08a6de89b9bbe3e2a70b3890cf
Thanks for the suggestion!

Tests/RevenueCatUITests/BaseSnapshotTest.swift
```diff
@@ -166,6 +166,9 @@ extension View {
 
         window.makeKeyAndVisible()
 
+        // Wait for views to render
+        RunLoop.main.run(until: Date().addingTimeInterval(0.1))
```
ajpallares, line 170: This line here is ~magical~ necessary for tests to pass in iOS 26.

For some reason I don't fully understand, without it, only when running the tests in iOS 26, the test cases in `PresentIfNeededTests.swift` would finish before receiving the paywall callbacks (test expectations or `toEventually` didn't solve this; I tried many combinations)

Tests/StoreKitUnitTests/TrialOrIntroPriceEligibilityCheckerSK2Tests.swift
```diff
@@ -210,9 +210,11 @@ class TrialOrIntroPriceEligibilityCheckerSK2Tests: StoreKitConfigTestCase {
 
         _ = try await sk2Product.purchase()
 
+        try await Task.sleep(nanoseconds: 100_000_000) // 0.1 second
```
ajpallares: I'm not a huge fan of delays, but this test `TrialOrIntroPriceEligibilityCheckerSK2Tests` has been flaky and has failed more than usual when run on iOS 26. We are probably checking intro eligibility too fast right after `purchase()`. This delay seems to reduce the flakiness greatly as it "gives time" for StoreKit testing to notice that the `sk2Product` was indeed purchased.

ajpallares:
@RCGitBot please test

Tests/RevenueCatUITests/Data/PaywallDataValidationTests.swift
```diff
@@ -283,7 +283,7 @@ private extension PaywallDataValidationTests {
     static func snapshot(
         _ paywall: PaywallData,
         file: StaticString = #file,
-        testName: String = #function,
+        testName: String = CurrentTestCaseTracker.osVersionAndTestName,
```
vegaro, line 286: Can you explain this change?

Tests/StoreKitUnitTests/TrialOrIntroPriceEligibilityCheckerSK2Tests.swift
vegaro: let's add a comment

vegaro (approved):
Ready to go after those couple small things

Tests/RevenueCatUITests/Data/PaywallDataValidationTests.swift
ajpallares: Sure! Before adding iOS 26, all the JSON snapshots for the `PaywallDataValidationTests` were exactly the same. So the snapshot file was named after only the test method name (`#function`). Now, since the JSON snapshots depend on the iOS version (because the system's `Color.accentColor` changed in iOS 26), we need the snapshot files to also include the iOS version in their name. https://github.com/RevenueCat/purchases-ios/blob/8a0787f7fba8a94b1efd16e5ccd0cffb8b22cb64/Tests/UnitTests/TestHelpers/CurrentTestCaseTracker.swift#L51-L60

Tests/StoreKitUnitTests/TrialOrIntroPriceEligibilityCheckerSK2Tests.swift
ajpallares: Good point. Added in https://github.com/RevenueCat/purchases-ios/pull/5552/commits/43ed3b2b1ce872570b163d6cc9826d171c84a28e

PR #5553, opened 2025-09-11
Use New iOS 26 Corner Radius in VC List Section

fire-at-will:
### Description
- Updates the virtual currencies list section to use the new corner radiuses when run on iOS 26
- Updates the view modifier used to render the background to match what other places in the Customer Center are doing now (using `.background(_,in:)` instead of `.background()` + `.clipShape()`

#### Before
before

#### After
Simulator Screenshot - iPhone 17 Pro - 2025-09-11 at 08 59 56

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrenciesScrollViewWithOSBackgroundSection.swift
+++ b/RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrenciesScrollViewWithOSBackgroundSection.swift
@@ -140,10 +140,12 @@ struct VirtualCurrenciesScrollViewWithOSBackgroundSection: View {
                             .padding(.vertical, 12)
                         }
                     }
+                    #if compiler(>=5.9)
                     .background(Color(colorScheme == .light
                                       ? UIColor.systemBackground
-                                      : UIColor.secondarySystemBackground))
-                    .clipShape(RoundedRectangle(cornerRadius: 10))
+                                      : UIColor.secondarySystemBackground),
+                                in: .rect(cornerRadius: CustomerCenterStylingUtilities.cornerRadius))
+                    #endif
                     .padding(.horizontal)
                 }
             }


fire-at-will:
@RCGitBot please test

RevenueCatUI/CustomerCenter/Views/Virtual Currencies/VirtualCurrenciesScrollViewWithOSBackgroundSection.swift
```diff
@@ -140,10 +140,12 @@ struct VirtualCurrenciesScrollViewWithOSBackgroundSection: View {
...
                         }
                     }
+                    #if compiler(>=5.9)
                     .background(Color(colorScheme == .light
                                       ? UIColor.systemBackground
-                                      : UIColor.secondarySystemBackground))
-                    .clipShape(RoundedRectangle(cornerRadius: 10))
+                                      : UIColor.secondarySystemBackground),
+                                in: .rect(cornerRadius: CustomerCenterStylingUtilities.cornerRadius))
+                    #endif
```
aboedo, line 148: @fire-at-will 
NABD but for stuff like this where all that's diverging is the corner radius, it'd be great to keep it centralized such that we make it a variable, like we did here: 

[CustomerCenterStylingUtilities](https://github.com/RevenueCat/purchases-ios/blob/main/RevenueCatUI/CustomerCenter/Utilities/CustomerCenterStylingUtilities.swift#L20-L33)

Usage: 
https://github.com/RevenueCat/purchases-ios/blob/a34be303b6b046249b9649c39a7a08a45d14c1d1/RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionButtonsView.swift#L68

maybe as a follow-up

fire-at-will: Totally agree! FWIW, I don't think `CustomerCenterStylingUtilities` was around when we initially built this view - maybe that would have been a good time to introduce it.

I also see five different places where we have this duplicated across the Customer Center codebase:

```swift
#if compiler(>=5.9)
               ? UIColor.systemBackground
               : UIColor.secondarySystemBackground),
         in: .rect(cornerRadius: CustomerCenterStylingUtilities.cornerRadius))
#endif
```

This could definitely be extracted out to a shared view modifier to help us avoid forgetting to update it in one place like we did with this one.

aboedo: yeah the styling utilities is a new thing that we added with the Liquid Glass update specifically for this purpose, we just didn't update the old code in all the places where it would have made sense

aboedo: why is the check against compiler 5.9? shouldn't it be 6.2?

aboedo: on a second look, what your code did was exactly what I was suggesting lol I misread the PR 🤦 apparently I don't know how to read a diff anymore. Sorry for the noise! 

anyway, I am still curious about the compiler 5.9 check

fire-at-will: > why is the check against compiler 5.9? shouldn't it be 6.2?

The `.background(:in:)` view modifier was introduced in iOS 15 ([source](https://developer.apple.com/documentation/swiftui/view/background(_:in:fillstyle:)-20tq5)), which was shipped with Xcode 15 / Swift 5.9, so we need the checks to ensure that the RevenueCatUI module compiles on Xcode 14. 

The 5.9 checks for the other `.background(:in:)` calls in the Customer Center were introduced pretty recently by @ajpallares in https://github.com/RevenueCat/purchases-ios/pull/5524 to resolve compilation issues with Xcode 14. 🙌

fire-at-will: The repo's README says that the SDK requires Xcode 15+, but I suppose it's good practice to try supporting older versions as much as possible 😎

facumenzella: I like the view modifier idea.

ajpallares: > The repo's README says that the SDK requires Xcode 15+, but I suppose it's good practice to try supporting older versions as much as possible

@fire-at-will Well, if I'm not missing anything, supporting older versions of Xcode is not something we do for good practice, as it's a pain for us in multiple ways (compiler checks like the 5.9 one mentioned in this thread, new Swift features that cannot be used, etc.). **But** AFAIK, it's the only way to run the SDK tests in older versions of iOS. For example, the SDK still supports iOS 13, but with Xcode 16 there's no way to install a simulator for iOS 13. We must use Xcode 14 for that. So, we need to keep supporting Xcode 14 only for that reason.
Otherwise, we would only need to support the oldest supported version of Xcode by Apple (Xcode 16 currently).

Issue #5290, opened 2025-06-13
SDK Not compiling on Xcode 26 due to deprecation warnings


BarMalka:
- [X] I have updated Purchases SDK to the latest version
- [X] I have read the [Contribution Guidelines](https://github.com/RevenueCat/purchases-ios/blob/main/Contributing/CONTRIBUTING.md)
- [X] I have searched the [Community](https://community.revenuecat.com)
- [X] I have read [docs.revenuecat.com](https://docs.revenuecat.com/)
- [X] I have searched for [existing Github issues](https://github.com/RevenueCat/purchases-ios/issues)

**Describe the bug**
Our project is configured with `SWIFT_TREAT_WARNINGS_AS_ERRORS = NO`.
When I was compiling our project in Xcode 26, under the current latest release (`5.28.1`) I am getting multiple deprecation errors (warnings). Here's a few:
`'SKPaymentTransaction' is deprecated: first deprecated in iOS 18.0 - Use PurchaseResult from Product.purchase(confirmIn:options:)`
`'SKPaymentQueueDelegate' is deprecated: first deprecated in iOS 18.0 - No longer supported`
`'SKPaymentTransactionObserver' is deprecated: first deprecated in iOS 18.0 - Use StoreKit 2 Transaction APIs`
See screenshot below

1. Environment
   1. Platform:
   2. SDK version:
   3. StoreKit version:
      - [ ] StoreKit 1 (default on versions <5.0.0. Can be enabled in versions >=5.0.0 with `.with(storeKitVersion: .storeKit1)`)
      - [x] StoreKit 2 (default on versions >=5.0.0)
   4. OS version: `15.4 (24E247)`
   5. Xcode version: `26.0 beta 1`
   6. Device and/or simulator:
      - [X] Device
      - [X] Simulator
   7. Environment:
      - [ ] Sandbox
      - [ ] TestFlight
      - [ ] Production
   8. How widespread is the issue. Percentage of devices affected.
2. [Debug logs](https://docs.revenuecat.com/docs/debugging) that reproduce the issue. **Complete logs with `Purchases.logLevel = .verbose` will help us debug this issue.**

<img width="2555" alt="Image" src="https://github.com/user-attachments/assets/65a89cfd-5f5a-4165-9c43-bc8c56cb4ed9" />

3. Steps to reproduce, with a description of expected vs. actual behavior

_Please fill in_
 
[...]

PR #5554, opened 2025-09-11
Silence deprecation warnings in XCFramework

ajpallares:
This PR solves #5290 

### Motivation
The generated header RevenueCat-Swift.h in the SDK's XCFramework contains interfaces that depend on StoreKit APIs that are deprecated. As a result, projects that integrate the SDK using the XCFramework would receive multiple deprecation warnings in the issue navigator
image

In particular, projects with the `SWIFT_TREAT_WARNINGS_AS_ERRORS` flag set to Yes would fail to build for this reason (see #5290)

### Description
This PR removes the deprecation warning by inserting an appropriate directive in all the generated RevenueCat-Swift.h files of the XCFramework before signing it
```
#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -1173,7 +1173,6 @@ platform :ios do
     swift_headers.each do |header_path|
       UI.message("Processing header: #{header_path}")
       
-      # Read the current header content
       content = File.read(header_path)
       
       # Check if pragma directives are already present
@@ -1187,10 +1186,7 @@ platform :ios do
 #pragma clang diagnostic ignored "-Wdeprecated-declarations"
           PRAGMA
           
-          # Insert the pragma directives
           content.insert(insertion_index, pragma_directives)
-          
-          # Write back to file
           File.write(header_path, content)
           UI.success("Added deprecation warning suppression to #{File.basename(header_path)}")
         else


fastlane/Fastfile
```diff
@@ -1159,6 +1164,45 @@ platform :ios do
...
+    swift_headers.each do |header_path|
+      UI.message("Processing header: #{header_path}")
+      
+      # Read the current header content
+      content = File.read(header_path)
+      
+      # Check if pragma directives are already present
+      unless content.include?('#pragma clang diagnostic ignored "-Wdeprecated-declarations"')
+        # Add pragma directives to suppress StoreKit 1 deprecation warnings
+        # Insert after the existing pragma clang diagnostic push
```
facumenzella: I think this comment is misleading. It's inserting after `-Wgcc-compat` right? Not a big deal though

fastlane/Fastfile
```diff
@@ -1159,6 +1164,45 @@ platform :ios do
...
+          PRAGMA
+          
+          # Insert the pragma directives
+          content.insert(insertion_index, pragma_directives)
+          
+          # Write back to file
+          File.write(header_path, content)
+          UI.success("Added deprecation warning suppression to #{File.basename(header_path)}")
+        else
+          UI.warning("Could not find insertion point in #{File.basename(header_path)}")
```
facumenzella: I think this should fail, otherwise we will most likely miss this

fastlane/Fastfile
ajpallares: I agree it's confusing. Since the code itself is clear enough, I just removed that line of the comment in https://github.com/RevenueCat/purchases-ios/pull/5554/commits/2d3f6c2c36ac9d462344f6e75467166640792f88. Lmk if you think I should do otherwise!

fastlane/Fastfile
ajpallares: Great point. I originally didn't consider it an error because not being able to insert the suppression of deprecation warnings is not a huge problem. But I agree that if we are doing it, we should be aware if it breaks in the future. And a warning would be very easy (and likely) to miss. Thanks for the improvement!

fastlane/Fastfile
facumenzella: Agreed 🚀

fastlane/Fastfile
```diff
@@ -1159,6 +1164,44 @@ platform :ios do
...
+  private_lane :suppress_deprecation_warnings_in_headers do |options|
+    xcframework_path = options[:xcframework_path]
+    
+    # Find all RevenueCat-Swift.h files in the XCFramework
+    swift_headers = Dir.glob("#{xcframework_path}/**/RevenueCat-Swift.h")
+    
+    swift_headers.each do |header_path|
+      UI.message("Processing header: #{header_path}")
+      
+      # Read the current header content
```
facumenzella:
```suggestion
```

this one is too much hehe

fastlane/Fastfile
```diff
@@ -1159,6 +1164,44 @@ platform :ios do
...
+        if pragma_insertion_point
+          insertion_index = content.index("\n", pragma_insertion_point) + 1
+          pragma_directives = <<~PRAGMA
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+          PRAGMA
+          
+          # Insert the pragma directives
+          content.insert(insertion_index, pragma_directives)
+          
+          # Write back to file
```
facumenzella:
```suggestion
```

same for this one

fastlane/Fastfile
```diff
@@ -1159,6 +1164,44 @@ platform :ios do
...
+        # Add pragma directives to suppress StoreKit 1 deprecation warnings
+        pragma_insertion_point = content.index('#pragma clang diagnostic ignored "-Wgcc-compat"')
+        
+        if pragma_insertion_point
+          insertion_index = content.index("\n", pragma_insertion_point) + 1
+          pragma_directives = <<~PRAGMA
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
+          PRAGMA
+          
+          # Insert the pragma directives
```
facumenzella:
```suggestion
```

fastlane/Fastfile
ajpallares: hahah yeah, the AI is too verbose 🙈

ajpallares: Comments removed in https://github.com/RevenueCat/purchases-ios/pull/5554/commits/a8cc223f9e60f3e0c23d07600beaffd228f6e64a. Thanks for the suggestion! 💅

PR #5556, opened 2025-09-11
clean up a couple more places that were using the old corner radius

aboedo:
Noticed a couple more places where we were using the old corner radius. 

Eyeballing it, those seem like the only places left where we had `cornerRadius(10)`

| Before | After |
| :-: | :-: |
| Screenshot 2025-09-11 at 5 22 20 PM | Screenshot 2025-09-11 at 5 22 37 PM |
| Screenshot 2025-09-11 at 5 18 59 PM  |  Screenshot 2025-09-11 at 5 19 14 PM |


facumenzella:
:shipit:

PR #5565, opened 2025-09-15
Pass navigation options explicitly to dismiss button

facumenzella:
### Motivation
We've received a report that the X button is visible whenever the CustomerCenter errors out. The environment is not being propagated properly.

### Description
- Explicitly pass the options to the view

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
@@ -268,25 +268,15 @@ private extension CustomerCenterView {
     ///
     /// - Note: Using `@Environment(.dismiss)` directly inside child views (e.g., toolbar buttons) can fail to dismiss
     /// the view when presented. To ensure reliable behavior on iOS 15, we capture `dismiss` at the
-    /// container level (CustomerCenterView) and propagate it via `navigationOptions.onCloseHandler`.
+    /// container level and propagate it via `navigationOptions.onCloseHandler`.
     var navigationOptionsWithDismiss: CustomerCenterNavigationOptions {
-        // Preserve shouldShowCloseButton exactly as provided by the caller.
         // Only inject a dismissal handler if missing, to ensure reliability on iOS 15.
-        if let onClose = self.navigationOptions.onCloseHandler {
-            return CustomerCenterNavigationOptions(
-                usesNavigationStack: self.navigationOptions.usesNavigationStack,
-                usesExistingNavigation: self.navigationOptions.usesExistingNavigation,
-                shouldShowCloseButton: self.navigationOptions.shouldShowCloseButton,
-                onCloseHandler: onClose
-            )
-        } else {
-            return CustomerCenterNavigationOptions(
-                usesNavigationStack: self.navigationOptions.usesNavigationStack,
-                usesExistingNavigation: self.navigationOptions.usesExistingNavigation,
-                shouldShowCloseButton: self.navigationOptions.shouldShowCloseButton,
-                onCloseHandler: { self.dismiss() }
-            )
-        }
+        return CustomerCenterNavigationOptions(
+            usesNavigationStack: self.navigationOptions.usesNavigationStack,
+            usesExistingNavigation: self.navigationOptions.usesExistingNavigation,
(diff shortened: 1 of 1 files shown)


RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
```diff
@@ -269,8 +270,15 @@ private extension CustomerCenterView {
...
-            return self.navigationOptions
+        // Preserve shouldShowCloseButton exactly as provided by the caller.
+        // Only inject a dismissal handler if missing, to ensure reliability on iOS 15.
+        if let onClose = self.navigationOptions.onCloseHandler {
+            return CustomerCenterNavigationOptions(
+                usesNavigationStack: self.navigationOptions.usesNavigationStack,
+                usesExistingNavigation: self.navigationOptions.usesExistingNavigation,
+                shouldShowCloseButton: self.navigationOptions.shouldShowCloseButton,
+                onCloseHandler: onClose
+            )
```
ajpallares: How is this different from the previous implementation that returned `self.navigationOptions` if `self.navigationOptions.onCloseHandler != nil`? I think it's exactly the same (only a new instance of `CustomerCenterNavigationOptions` is created now).

Also, probably this if-else can be simplified to just

```
        return CustomerCenterNavigationOptions(
            usesNavigationStack: self.navigationOptions.usesNavigationStack,
            usesExistingNavigation: self.navigationOptions.usesExistingNavigation,
            shouldShowCloseButton: self.navigationOptions.shouldShowCloseButton,
            onCloseHandler: self.navigationOptions.onCloseHandler ?? { self.dismiss() }
        )
```

RevenueCatUI/CustomerCenter/ButtonStyles.swift
```diff
@@ -155,23 +155,27 @@ struct DismissCircleButton: View {
 @available(macOS, unavailable)
 @available(tvOS, unavailable)
 @available(watchOS, unavailable)
-struct DismissCircleButtonToolbarModifier: ViewModifier {
+@available(iOS 15.0, *)
+@available(macOS, unavailable)
+@available(tvOS, unavailable)
+@available(watchOS, unavailable)
```
ajpallares: These seem to be repeated

RevenueCatUI/CustomerCenter/ButtonStyles.swift
```diff
@@ -192,14 +196,22 @@ struct DismissCircleButtonToolbarModifier: ViewModifier {
 @available(tvOS, unavailable)
 @available(watchOS, unavailable)
 extension View {
-    /// Adds a toolbar with a dismiss button if `navigationOptions.shouldShowCloseButton` is true.
-    func dismissCircleButtonToolbarIfNeeded() -> some View {
-        modifier(DismissCircleButtonToolbarModifier(customDismiss: nil))
+    /// Adds a toolbar with a dismiss button if `navigationOptions.shouldShowCloseButton` is true,
```
ajpallares: You changed the `navigationOptions` property to `options`. I'd argue to leave it as `navigationOptions` to have a smaller diff, but if changed, then this should be updated to
```suggestion
    /// Adds a toolbar with a dismiss button if `options.shouldShowCloseButton` is true,
```

RevenueCatUI/CustomerCenter/Views/CustomerCenterView.swift
facumenzella: I've been moving back and forth with this. I want this to be as clear as possible 😅 
When I realized the environment wasn't propagating properly, I tried creating a new one but of course it didn't work.

facumenzella: The previous code is cleaner IMO, what do you think?

ajpallares: I'm fine if you want to leave it like this. To me, removing the if-else is easier to read since it's clearer than the idea is to fall back to `{ self.dismiss() }` in case `self.navigationOptions.onCloseHandler == nil`. But I don't have a strong opinion about it, so feel free to leave it as you think it's better 👍

ajpallares (approved):

PR #5572, opened 2025-09-18
Fix a flaky test

ajpallares:
### Motivation
The test `test_prefetch_invokesNetwork` was failing in macOS very often. There was a race condition where the test assertion was sometimes happening before the tested code finished execution.

### Description
This PR fixes it by relaxing the assertion from `XCTAssertEqual` to `expect(...).toEventually(equal(...))` to allow the code to finish.


ajpallares:
@RCGitBot please test

PR #5573, opened 2025-09-18
Fix some more flaky tests

ajpallares:
### Motivation
Tests in `OtherIntegrationTests.swift` were sometimes failing (e.g. [here](https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/30246/workflows/1719e25f-3716-4ab2-9db5-00aee17cb2ed/jobs/376605)) with the error
> There is an issue with your configuration. Check the underlying error for more details. There's a problem with your configuration. None of the products registered in the RevenueCat dashboard could be fetched from App Store Connect (or the StoreKit Configuration file if one is being used).

I was able to consistently reproduce it in my machine with Xcode 26 (having installed the latest release of Xcode 26 recently probably helped to repro it).

### Description
I realized that `OtherIntegrationTests` were not configuring the `SKTestSession` with the StoreKit config file using the [`init(configurationFileNamed:)`](https://developer.apple.com/documentation/storekittest/sktestsession/init(configurationfilenamed:)) initializer. These tests were only relying on the StoreKit configuration file being set in the "BackendIntegrationTests" Xcode scheme.
image

However, other integration tests like `StoreKit1IntegrationTests` and `StoreKit2IntegrationTests` are indeed initializing the `SKTestSession`
https://github.com/RevenueCat/purchases-ios/blob/ce89da5e00145eace57a4f5377f0d3c37365230e/Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift#L36-L38

https://github.com/RevenueCat/purchases-ios/blob/ce89da5e00145eace57a4f5377f0d3c37365230e/Tests/BackendIntegrationTests/BaseStoreKitIntegrationTests.swift#L58-L62

This PR adds the same configuration step to `OtherIntegrationTests`, with the hope that this will reduce the flakiness due to the error "None of the products registered in the RevenueCat dashboard could be fetched from App Store Connect (or the StoreKit Configuration file if one is being used)."

### DIFF
--- a/Tests/BackendIntegrationTests/OtherIntegrationTests.swift
+++ b/Tests/BackendIntegrationTests/OtherIntegrationTests.swift
@@ -20,9 +20,15 @@ import XCTest
 
 class OtherIntegrationTests: BaseBackendIntegrationTests {
 
+    private var testSession: SKTestSession!
+
     override func setUp() async throws {
         // Some tests need to introspect logs during initialization.
         super.initializeLogger()
+
+        if self.testSession == nil {
+            try await self.configureTestSession()
+        }
 
         try await super.setUp()
     }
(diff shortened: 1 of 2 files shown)


Tests/RevenueCatUITests/PaywallViewEventsTests.swift
```diff
@@ -139,7 +139,7 @@ class BasePaywallViewEventsTests: TestCase {
 
         await self.waitForCloseEvent()
 
-        expect(self.events).to(haveCount(4))
```
ajpallares, line 142: This is a separate fix. Sometimes we were getting 2 events instead of 4 (e.g. [here](https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/30248/workflows/52766146-6bab-4fd6-a11d-5cba00bdaae9/jobs/376676/tests))

Tests/BackendIntegrationTests/OtherIntegrationTests.swift
```diff
@@ -33,6 +39,27 @@ class OtherIntegrationTests: BaseBackendIntegrationTests {
         try await super.tearDown()
     }
 
+    func configureTestSession() async throws {
+        assert(self.testSession == nil, "Attempted to configure session multiple times")
```
facumenzella, line 43:
```suggestion
        assert(self.testSession == nil, "SKTestSession already configured")
```

💅

PR #5580, opened 2025-09-19
Add Remote variants for Example tuist projects

facumenzella:
### Motivation
We want to have a way to test RC integration using SPM, our main installation method. 
Try to keep it clean for ease of maintenance.

### Description
- Add Remote variants for Example projects
- Rename maestro folders to be Maestro and MaestroRemote instead of `rc-maestro`.

Now if you run `TUIST_RC_LOCAL=false tuist generate` this is the output.
Screenshot 2025-09-19 at 13 18 15

The goal is to keep the same command by **duplicating the project, but having the same target name.**
When running local = false, example projects will use Xcode natives SPM dependency resolver. The other projects will still use tuist resolver.


Examples/MaestroRemote/Project.swift
```diff
@@ -0,0 +1,60 @@
...
+                    "REVENUECAT_PROXY_URL_HOST": "$(REVENUECAT_PROXY_URL_HOST)"
+                ]
+            ),
+            sources: ["../Maestro/rc-maestro/Sources/**/*.swift"],
+            resources: [
+                "../Maestro/rc-maestro/Resources/**/*.xcassets",
+            ],
+            dependencies: [
+                .package(product: "RevenueCat", type: .runtime),
+                .package(product: "RevenueCatUI", type: .runtime),
```
ajpallares, line 39: Maybe we should declare this in `TargetDependencies.swift` as we have others? (e.g. `.revenueCat`, `.revenueCatUI`)

ajpallares: Also, is this `.runtime` making it point to `.main` because we declare it like this in `Tuist/Package.swift`?
```
.package(
    url: "https://github.com/RevenueCat/purchases-ios",
    branch: "main"
),
```

If so, what does it use the purchases-ios from main and no the `purchases-ios-spm at 3.14.4 also declared in `Tuist/Package.swift`
```
.package(
    url: "https://github.com/RevenueCat/purchases-ios",
    branch: "main"
),
.package(
    url: "https://github.com/RevenueCat/purchases-ios-spm",
    revision: "3.14.4"
),
```

Examples/PurchaseTesterRemote/Project.swift
```diff
@@ -0,0 +1,124 @@
...
+            product: .framework,
+            bundleId: "com.revenuecat.Core",
+            deploymentTargets: allDeploymentTargets,
+            sources: [
+                "../../Tests/TestingApps/PurchaseTesterSwiftUI/Core/**/*.swift",
+            ],
+            dependencies: [
+                .package(product: "RevenueCat", type: .runtime),
+                .package(product: "RevenueCatUI", type: .runtime),
+                .package(product: "ReceiptParser", type: .runtime)
```
ajpallares, line 96: I only see receipt parser declared in `Tuist/ProjectDescriptionHelpers/TargetDependencies.swift`.

```
/// Returns a ReceiptParser dependency
/// - Returns: A TargetDependency for ReceiptParser
public static var receiptparser: TargetDependency {
    .project(
        target: "ReceiptParser",
        path: .relativeToRoot("Projects/RevenueCat")
    )
}
```

How will it find it from the remote? (very likely I'm missing something 🙏)

ajpallares (commented):
This is looking great! Especially for CI, so that we make sure that we're using SPM for tests to closer replicate the most used real-life scenarios.

However, I wonder, I see that the RevenueCat dependency added depends on the `main` branch.

Image

While that might make sense for some projects like Magic Weather (although it could  be argued that it should depend on the latest release), trying to think on our CI, ideally it should depend on the local code as a Swift Package (and not on the remote `main` branch).

facumenzella:
Closing this one

Issue #5568, opened 2025-09-17
🐛 Promotional offers not working since 5.25.0


prodigel:
### Describe the bug

We're attempting to update to the latest release (5.39.0), unfortunately we noticed that promotional offers don't show up anymore when users try to cancel their plan.
I checked through previous versions, it seems that it stopped working with 5.25.0, prior to that we see the promotional offers correctly.

We're using CustomerCenterView, no new configuration seems to exist to change its behavior, it looks like a bug.

Let me know if you need more info,
Thanks!

### Platform

iOS

### SDK version

5.25.0

### SDK integration method

Swift Package Manager

### StoreKit version

{"StoreKit 1 (default on versions <5.0.0. Can be enabled in versions >=5.0.0 with `.with(storeKitVersion" => ".storeKit1)`)"}

### OS version

iOS 18.6.2

### Xcode version

16.2

### Device and/or simulator

Device

### Environment

Sandbox

### How widespread is the issue

100%

### Debug logs

```shell
No logs
```

### Steps to reproduce

1. Configure promotional offers in Customer service dashboard
2. Attempt to cancel existing plan using the CustomerServiceView

Expected: Promotional offer should pop up
Actual: Keep/Cancel plan is shown directly

### Other information

```markdown

```

### Additional context

```markdown

```

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/64500) and notified the team. Thank you for reporting, we're checking this out!

prodigel:
A small update on the issue:
I realized the problem comes from not having a defined feedback survey. Since it was configured prior to the latest changes, it's not obvious why it was failing, I noticed that by debugging. 
I believe a safe way to handle this would be to bypass the feedback survey check if the customer center configuration was made prior to the released changes.
Furthermore it's not obvious for existing setups that a feedback survey option exists, I had to delete the existing 'Cancel' configuration and create it from scratch to see the option.

Hope that you'll find my feedback useful :)

GuilhermeMota93:
Hey @prodigel, thanks for raising!

The team is already on it, and the [#5581](https://github.com/RevenueCat/purchases-ios/pull/5581) PR aims to fix this 🙏

PR #5581, opened 2025-09-19
Open promotional offers from CustomerCenter Detail screen

facumenzella:
### Motivation
We've noticed this has not been working after a [report](https://github.com/RevenueCat/purchases-ios/issues/5568).
Promotionals can be attached to any path, not only to feedback survey

### Description
- Open promo offers from detail screen

### DIFF
--- a/RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
@@ -110,7 +110,7 @@ class BaseManageSubscriptionViewModel: ObservableObject {
                 }
 
         case let .promotionalOffer(promotionalOffer) where purchaseInformation?.store == .appStore:
-            if promotionalOffer.eligible, let productIdentifier = feedbackSurveyData?.productIdentifier {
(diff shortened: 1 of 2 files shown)


RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
```diff
@@ -110,7 +110,7 @@ class BaseManageSubscriptionViewModel: ObservableObject {
                 }
 
         case let .promotionalOffer(promotionalOffer) where purchaseInformation?.store == .appStore:
-            if promotionalOffer.eligible, let productIdentifier = feedbackSurveyData?.productIdentifier {
+            if promotionalOffer.eligible, let productIdentifier = purchaseInformation?.productIdentifier {
```
vegaro, line 113: did you test opening a promo offer from a survey after this change? I think it shouldn't matter since they are technically the same productIdentifier, but double checking

vegaro (approved):

RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
facumenzella: Feedback Survey has its own VM and handler 😞 

So good news is that it works
Bad news is this one was broken

Issue #5464, opened 2025-08-12
🐛 Gradient renders incorrectly on iOS 26 Beta 6 vs iOS 18


MattKiazyk:
### Describe the bug

Gradients are not showing up in the same place in iOS 26 Beta 5+6 (just started testing these) as they are on iOS 18. See screenshots below. 

| Paywall Editor | iOS 18 | iOS 26 |
|--------|--------|--------|
| <img width="500" height="1028" alt="Image" src="https://github.com/user-attachments/assets/f67e83a0-8e4d-449e-bcfe-52e879a5e0ec" /> | ![Image](https://github.com/user-attachments/assets/83c061a3-206c-49f8-9b92-1ef3d42689f1) | <img width="603" height="1311" alt="Image" src="https://github.com/user-attachments/assets/c648048e-d40b-41bc-84a6-1371b7badb8e" /> | 

### Platform

iOS

### SDK version

5.34.0

### StoreKit version

StoreKit 2 (default on versions >=5.0.0)

### OS version

iOS 26

### Xcode version

16.4

### Device and/or simulator

Device

### Environment

Sandbox

### How widespread is the issue

iOS 26 Beta devices

### Debug logs

```shell
Don't think logs are needed for rendering issue, but will post if requested.
```

### Steps to reproduce

In Paywall editor:
- I've added an overlay onto an image that's at the bottom. 
- The image has a fixed height, width fill, but is larger then 300 px height. 

<img width="3058" height="1546" alt="Image" src="https://github.com/user-attachments/assets/b1d0c711-b0b1-4913-81f8-b116a98c4ce4" />

Gradient shows up in paywall editor and iOS 18, but not in iOS 26.

Note for debugging: If I set the image height to Fit - the gradient renders properly at the bottom of the image on iOS 26. My guess is that the gradient renders there currently on iOS 26, just can't see it. 

<img width="628" height="1326" alt="Image" src="https://github.com/user-attachments/assets/4771df61-89b1-4b82-8c9f-f02a4f6222e6" />

### Other information

```markdown

```

### Additional context

```markdown

```

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/62861) and notified the team. Thank you for reporting, we're checking this out!

mawr92:
Hey @MattKiazyk, 

[...]

PR #5584, opened 2025-09-20
MON-1206 Fix background image—gradient overlay

JZDesign:

### Motivation

Reported in #5464 but is unrelated to the original reported issue. It's a secondary issue found in [this comment](https://github.com/RevenueCat/purchases-ios/issues/5464#issuecomment-3217026374)

The iOS sdk did not support color or gradient overlays on the image used as a paywall background. This adds that functionality.

### Description

There were 3 fixes:
1. Include the color overlay details in the model
1. Correct the radial gradient end point to be based on the view size and not fixed to 100 points
1. Clamp the size of the overlay to the view size so that the swiftUI renderer doesn't make mistakes

|Linear|Radial|
|:--:|:--:|
|one|two|


joshdholtz (commented):
Looks good to me! 

Two small asks:
1) Can we add a screenshot to the description of this PR so we can see its working 😊 
2) Can we also some snapshots of this to Emerge?

And thanks for fixing! 🙌

JZDesign:
@joshdholtz, thanks for asking for screenshots because I missed some things (mostly around radial gradients)

RevenueCatUI/Templates/V2/ViewHelpers/SizeReader.swift
JZDesign, line 1: This is also included in the video component PR so this will cause merge conflicts

RevenueCatUI/Views/GradientView.swift
```diff
@@ -80,23 +80,23 @@ struct GradientView: View {
...
                     gradient: gradient,
                     startPoint: points.start,
                     endPoint: points.end
                 )
+            case .radial:
+                RadialGradient(
+                    gradient: gradient,
+                    center: .center,
+                    startRadius: 0,
+                    endRadius: min(geometry.size.width, geometry.size.height)
```
JZDesign, line 97: Unlike with color stops, the start and end radii are point values and not percentages. So 100 didn't push to the bounds of the view, it pushed  it to 100 points then it was whatever the last color of the gradient was

RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
```diff
@@ -95,6 +80,14 @@ fileprivate extension View {
...
                         .aspectRatio(contentMode: fitMode.contentMode)
                         .ignoresSafeArea()
+                }.overlay {
+                    ZStack {
+                        if let colorOverlay {
+                            colorOverlay
+                                .toView(colorScheme: colorScheme)
+                        }
+                    }
+                    .frame(maxWidth: size?.width, maxHeight: size?.height)
```
JZDesign, line 90: There are issues with rendering overlays directly above images. It's best to enforce the exact bounds we expect the overlay to render in. If we don't, SwiftUI often pushes the gradient outside of the bounds they should be in.

RevenueCatUI/Views/GradientView.swift
JZDesign: Screenshot 2025-09-22 at 8 30 25 AM

This is what is used to look like

RevenueCatUI/Templates/V2/ViewHelpers/SizeReader.swift
ajpallares: We can probably use this existing code instead https://github.com/RevenueCat/purchases-ios/blob/b09aaef263626c706a121719ae3f44af33ae698f/RevenueCatUI/Modifiers/ViewExtensions.swift#L268-L279

RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
```diff
@@ -29,16 +29,20 @@ struct BackgroundStyleModifier: ViewModifier {
...
     func body(content: Content) -> some View {
         if let backgroundStyle {
             content
+                .sizeReader($size)
                 .apply(
                     backgroundStyle: backgroundStyle,
                     colorScheme: colorScheme,
-                    alignment: alignment
+                    alignment: alignment,
+                    size: size == .zero ? nil : size
```
ajpallares: If we use https://github.com/RevenueCat/purchases-ios/blob/b09aaef263626c706a121719ae3f44af33ae698f/RevenueCatUI/Modifiers/ViewExtensions.swift#L268-L279

then we can have `size` be nullable
```swift
@State var size: CGSize?
```

and simplify this line
```swift
size: size
```

JZDesign: Thanks! Not sure how I missed that helper.

ajpallares: Well, I only discovered it recently (yesterday) and had no idea of it either 😄

PR #5586, opened 2025-09-22
MON-1296 Fix gradient rendering issue in iOS 26

### LINKED ISSUE #5464: 🐛 Gradient renders incorrectly on iOS 26 Beta 6 vs iOS 18

JZDesign:
Resolves #5464

The iOS 26 rendering of this component did not redraw the way that is did in previous versions, but it's actually expected behavior. With images that have aspect ratios and content modes and all that, resizing them behaves differently. If the overlay is supposed to be on the final rendered image and follow the same rules as the resizing, we should apply the overlay after the resizing.

> [!NOTE]
> I made 2 commits to show the 2 options I had. I chose this option because it's not redundant.

### Checklist

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
@@ -94,6 +94,12 @@ struct ImageComponentView: View {
                     }
                     .applyImageWidth(size: style.size)
                     .applyImageHeight(size: style.size, aspectRatio: self.aspectRatio(style: style))
+                    .applyIfLet(style.colorOverlay, apply: { view, colorOverlay in
+                        view.overlay(
+                            Color.clear
+                                .backgroundStyle(.color(colorOverlay))
+                        )
+                    })
                     .clipped()
                     .padding(style.padding.extend(by: style.border?.width ?? 0))
                     .shape(border: style.border,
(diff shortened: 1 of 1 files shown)


joshdholtz:
@JZDesign  This makes a lot of sense! I'd love to run this change again the cross platform validation screenshots to make sure this still works with a lot of other existing paywall designs. Hit me up today in Slack and I can show you how I do that (and then we can document it 😊)

JZDesign:
@joshdholtz — Looks good to me

https://github.com/RevenueCat/paywall-rendering-validation/pull/363

JZDesign:
@vegaro @ajpallares Yes. The layout issue rarely appeared in 16.4 You'll note that the snapshot tests didn't have modifications (aside from Mac which seem unreliable)

JZDesign:
@RCGitBot please test

Issue #5310, opened 2025-06-20
"Switch must be exhaustive" warning, missing .unsupported cases


neave:
**Describe the bug**
Xcode shows “Switch must be exhaustive” in PurchasesError.swift `getStoreKitErrorInfoIfAny` and StoreKitError+Extensions.swift, even though `@unknown` default is present. They need the '.unsupported' case.

1. Environment
   1. Platform:
   2. SDK version:
   3. StoreKit version:
      - [ ] StoreKit 1 (default on versions <5.0.0. Can be enabled in versions >=5.0.0 with `.with(storeKitVersion: .storeKit1)`)
      - [x] StoreKit 2 (default on versions >=5.0.0)
   4. OS version:
   5. Xcode version: 16.4
   6. Device and/or simulator:
      - [x] Device
      - [x] Simulator
   7. Environment:
      - [x] Sandbox
      - [x] TestFlight
      - [x] Production
   8. How widespread is the issue. Percentage of devices affected.
2. Debug logs
```
Switch must be exhaustive
```

3. Steps to reproduce
The compiler expects all enum cases to be explicitly handled, even with `@unknown` default.
 
4. **Other information** 
May require replacing `@unknown` default

5. **Additional context**
Likely surfaced by stricter checks in recent Xcode versions or SDK updates.

RCGitBot:
👀 We've just linked this [issue to our internal tracker](https://revenuecat.zendesk.com/agent/tickets/60334) and notified the team. Thank you for reporting, we're checking this out!

HaleyRevcat:
Hi, thanks for reporting this, we will take a look. Can you make sure that you are on our most recent SDK version 5.29.0 if you aren't already? Note that as this is a warning you should be able to still run your app and make purchases but let me know if this isn't the case.

neave:
Yes, this is using the most recent SDK version. And yes, it does compile as these are only warnings in Xcode (currently).

marcgg:
Hello, just checked with the latest XCode (26 RC1) and the latest Revenuecat 5.38.1, and this is still happening. It does compile, but warnings are adding a bit of noise, so a fix would be appreciated.

<img width="550" height="334" alt="Image" src="https://github.com/user-attachments/assets/4de71076-2867-4029-b11d-6a3abe2651aa" />

fire-at-will:
[...]

PR #5589, opened 2025-09-22
Support StoreKitError.unsupported

fire-at-will:
### Description
This PR addresses https://github.com/RevenueCat/purchases-ios/issues/5310 by handling `StoreKitError.unsupported`, which was introduced in iOS 18.4.


fire-at-will:
@RCGitBot please test

Sources/Error Handling/StoreKitError+Extensions.swift
```diff
@@ -36,6 +36,12 @@ extension StoreKitError: PurchasesErrorConvertible {
             return ErrorUtils.storeProblemError(error: self)
 #endif
 
+        #if compiler(>=6.1)
+        // StoreKitError.unsupported was introduced in iOS 18.4, which shipped with Xcode 16.3 beta 1 / Swift 6.1
+        case .unsupported:
+            return ErrorUtils.purchaseInvalidError(error: self)
```
fire-at-will, line 42: I'm open to using a different mapping function to map this to a `PurchasesError` - let me know if y'all think there's a more appropriate mapping available 👍

fire-at-will:
@RCGitBot please test

Sources/Error Handling/StoreKitError+Extensions.swift
ajpallares: I've checked and I think `purchaseInvalidError` makes sense

PR #5591, opened 2025-09-23
Fix paywalls button bottom sheet

joshdholtz:
### Motivation

There is currently a defect when loading an image to the sheet. Since the sheet is animating in when the async load completes the end image location is assigned before the sheet is done moving, and the result is the loaded image appears where it would be after the sheet presents, it does not slide in with the sheet. 

### Description

This PR does a couple of things.

1. Uses the FileRepository to prefetch the image instead of relying on the URLSessionCache
1. Fixes the animation bug—_in most cases_—by using a synchronous api when possible
1. Removes an erroneous background style that is not supported by the backend


Sources/Paywalls/Components/PaywallV2CacheWarming.swift
```diff
@@ -37,14 +37,19 @@ extension PaywallComponentsData {
 extension PaywallComponentsData.PaywallComponentsConfig {
 
     var allImageURLs: [URL] {
```
JZDesign, line 43: I'm taking over this PR for Josh while he's out of the country. I wanted to get your feedback on this…

it says ALL image urls. But currently we're only returning the non-heic images.

Josh put in some logic to only override that for the sheet component so that the HEIC images are prefetched too. 

Any reason why we don't just return ALL the image urls? Including the HEIC?

@facumenzella @MarkVillacampa @fire-at-will

RevenueCatUI/Templates/V2/Components/Button/BottomSheetView.swift
```diff
@@ -65,18 +65,6 @@ struct BottomSheetOverlayModifier: ViewModifier {
         }
     }
 
-    var backgroundStyle: BackgroundStyle? {
```
MarkVillacampa, line 68: the background is now provided by the base bottom sheet stack right?

Sources/Paywalls/Components/PaywallV2CacheWarming.swift
MarkVillacampa: I guess it makes sense to only fetch hi-res images for bottom sheet because:
- Only the bottom sheet can cause animation issues when replacing the image while loading.
- For the other components, we can load the hi-res image on-demand, causing less network and disk load.

RevenueCatUI/Templates/V2/Components/Button/BottomSheetView.swift
JZDesign: Yes

Sources/Paywalls/Components/PaywallV2CacheWarming.swift
JZDesign: Okay. Then I will continue down that path and note the reason

Sources/Paywalls/Components/PaywallV2CacheWarming.swift
```diff
@@ -55,10 +57,25 @@ extension PaywallComponentsData.PaywallComponentsConfig {
...
-    private func collectAllImageURLs(in stack: PaywallComponent.StackComponent) -> [URL] {
+    private func collectAllImageURLs(
+    in stack: PaywallComponent.StackComponent,
+    includeHighResInComponentHeirarchy: (PaywallComponent) -> Bool = { component in
+        // collecting high res images from the sheet is important because async functions
+        // prevent the proper animation from playing during sheet presentation and by collecting
+        // the images ahead of time we can synchronously render the image instead.
+        return component.isSheetButton
+    }
+    ) -> [URL] {
```
ajpallares, line 68: This feels to me a bit overcomplicated due to the use of the `includeHighResInComponentHeirarchy` closure. I think this can be simplified by using a boolean instead. Something like this:

```swift
private func collectAllImageURLs(in stack: PaywallComponent.StackComponent,
                                 includeHighResImages: Bool = false) -> [URL] {
    var urls: [URL] = []
    for component in stack.components {
        // Collecting high res images from the sheet is important because async functions
        // prevent the proper animation from playing during sheet presentation and by collecting
        // the images ahead of time we can synchronously render the image instead.
        //
        // Therefore, when the sheet button component is found, we override to true regardless
        // of children to ensure high res image collection
        let includeHighResImages = includeHighResImages || component.isSheetButton

        switch component {
        ...
        }

        return urls
    }

```

Sources/Paywalls/Components/PaywallV2CacheWarming.swift
```diff
@@ -154,6 +217,26 @@ extension PaywallComponentsData.PaywallComponentsConfig {
...
+        switch self {
+        case .button(let component):
+            switch component.action {
+            case .navigateTo(let destination):
+                switch destination {
+                case .sheet:
+                    return true
+                default:
+                    return false
+                }
```
ajpallares: Nit:
```suggestion
            case .navigateTo(.sheet):
                    return true
```

ajpallares (approved):
I think it makes sense! Just a couple of comments to simplify things a bit

Sources/Paywalls/Components/PaywallV2CacheWarming.swift
JZDesign: Yea, it used to be a boolean, but what I didn't like about that is that we set it to false when it's called—but when the button-sheet component branch is traversed it overrides it to true. 

So the false at the original call site was effectively a lie and that bothered me.

Now, the closure is explicit, we will check the hierarchy when the function returns true, and we explicitly override the function in the beginning of the scope. 

Both options are a bit confusing, but the boolean option is deceptive in the beginning and that's my issue with it. Where the function version is explicit and the complexity is front loaded. So it's going to be easier to find bugs if they should arise.

JZDesign: With your option it is a nice balance, front load the override, but it also makes it impossible to pass in a false and have the function actually respect it. So maybe I'm a purist and that is something I can be more pragmatic about, but that's why I did what I did lol

JZDesign: I'm going to merge as is, but if this becomes something that bothers people I will change it.

PR #5592, opened 2025-09-23
Fix: 10 Result builder limit

JZDesign:
I added a preview which fails in CI on main. This should fix that.


JZDesign:
@RCGitBot please test

JZDesign:
@RCGitBot please test

PR #5596, opened 2025-09-25
Enable Test Store

ajpallares:
This PR removes the `SIMULATED_STORE` compiler flag to enable the Test Store functionality in the SDK

### DIFF
--- a/Sources/CodableExtensions/Store+Extensions.swift
+++ b/Sources/CodableExtensions/Store+Extensions.swift
@@ -66,9 +66,7 @@ private extension Store {
         case .rcBilling: return "rc_billing"
         case .external: return "external"
         case .paddle: return "paddle"
-        #if SIMULATED_STORE
         case .testStore: return "test"
-        #endif
         }
     }
 
--- a/RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
+++ b/RevenueCatUI/CustomerCenter/Data/PurchaseInformation.swift
@@ -569,10 +569,7 @@ extension PurchaseInformation {
         case .amazon: return .storeAmazon
         case .unknownStore: return .storeUnknownStore
         case .paddle, .stripe, .rcBilling, .external: return .storeWeb
-
-        #if SIMULATED_STORE
         case .testStore: return .simulatedStore
-        #endif
         }
     }
 }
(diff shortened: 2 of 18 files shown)


Sources/CodableExtensions/Store+Extensions.swift
```diff
@@ -66,9 +66,7 @@ private extension Store {
         case .rcBilling: return "rc_billing"
         case .external: return "external"
         case .paddle: return "paddle"
-        #if SIMULATED_STORE
         case .testStore: return "test"
-        #endif
```
JayShortway, line 71: Really torn on this one 😕 People will be forced to handle this in their `switch` statements, but they can't even use it (yet). It's a chicken and egg problem, but it would be nice if the chicken and the egg are released a bit closer to each other hahah. An alternative is using `.unknownStore` publicly for now. Not sure I love it either.

JayShortway (commented):
Let's go! Just the question of the public enum case. Maybe we can gauge the team's opinion in the sync today.

Sources/CodableExtensions/Store+Extensions.swift
ajpallares: Fair point! We had a similar internal conversation when adding the Paddle store enum case ([here](https://github.com/RevenueCat/purchases-ios/pull/4981)) some months ago and we decided to include it.

That said, it's true that we'd be adding a case to the enum before devs can even use it. I don't remember if we had the same circumstances in the Paddle case.

ajpallares: It's true that the Test Store case is not even implemented in Android (see https://github.com/RevenueCat/purchases-android/pull/2685/files). However, if we are going to eventually add the test store enum case, I'd rather be consistent from the very beginning and do this single API breaking change than initially use `.unknownStore` for this and later add the breaking change + behavior change. 

In any case, I'd argue that we should do the same in both platforms, for sure.

ajpallares: As discussed internally, Test Store products will be associated to the `Store.unknown` in the SDK. 
Done in https://github.com/RevenueCat/purchases-ios/pull/5596/commits/6f2975397ce8028d3d31c7727c9d5c0dd7da0cf2

ajpallares:
@RCGitBot please test

Sources/CodableExtensions/Store+Extensions.swift
JayShortway: Nice, yea it'll hopefully only be a short while before we can introduce `.testStore`! 🤞

PR #5597, opened 2025-09-25
Remove `github_rate_limit` in release trains

vegaro:
We noticed the lanes uses `RC_INTERNAL_GITHUB_TOKEN` during the release train for determining the changelog and next version.

That env variable wasn't set in CI, we had two other GitHub tokens, but not that particular one. And that one is used in this line https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/blob/24d8edac49bf1f7e09ffe4c184038536938077e7/lib/fastlane/plugin/revenuecat_internal/actions/bump_version_update_changelog_create_pr_action.rb#L78. That means we were making unauthenticated GitHub requests:

> You can make unauthenticated requests if you are only fetching public data. Unauthenticated requests are associated with the originating IP address, not with the user or application that made the request. The primary rate limit for unauthenticated requests is 60 requests per hour.

https://docs.github.com/en/rest/using-the-rest-api/rate-limits-for-the-rest-api?apiVersion=2022-11-28

We updated the token to use in https://github.com/RevenueCat/fastlane-plugin-revenuecat_internal/pull/86 to `GITHUB_TOKEN`, which is set in all repositories. With this, we can remove the rate limits, since we will be using an authenticated request with a higher rate limit.


ajpallares (approved):

vegaro:
I am going to merge since the failures are unrelated

PR #5602, opened 2025-09-25
Fix some tests

ajpallares:
This PR fixes API tests and iOS 14 CI execution that I broke in #5596

Also, this PR changes the execution in iOS 26 to use iPhone 17 simulators


Tests/APITesters/AllAPITests/ObjcAPITester/RCEntitlementInfoAPI.m
```diff
@@ -49,7 +49,6 @@ + (void)checkEnums {
         case RCUnknownStore:
         case RCPaddle:
         case RCExternal:
-        case RCTestStore:
```
ajpallares, line 52: I missed this one when removing the testStore enum case

JZDesign (approved):

PR #5604, opened 2025-09-26
Add displayName to SubscriptionInfo

facumenzella:
### Motivation
We've got a request to add `displayName` to `SubscriptionInfo`. It's exposed in [Android](https://github.com/RevenueCat/purchases-android/blob/85394eb53e51a1719f80e149e06a6d79a27db065/purchases/src/main/kotlin/com/revenuecat/purchases/SubscriptionInfo.kt#L88), but it's missing in iOS

### Description
- Add `displayName`
- Update `APITests`

### DIFF
--- a/Tests/APITesters/AllAPITests/SwiftAPITester/SubscriptionInfoAPI.swift
+++ b/Tests/APITesters/AllAPITests/SwiftAPITester/SubscriptionInfoAPI.swift
@@ -32,4 +32,5 @@ func checkSubscriptionInfoAPI() {
     let txId: String? = subscription.storeTransactionId
     let isActive: Bool = subscription.isActive
     let willRenew: Bool = subscription.willRenew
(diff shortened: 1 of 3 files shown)


Tests/APITesters/AllAPITests/SwiftAPITester/SubscriptionInfoAPI.swift
```diff
@@ -32,4 +32,5 @@ func checkSubscriptionInfoAPI() {
     let txId: String? = subscription.storeTransactionId
     let isActive: Bool = subscription.isActive
     let willRenew: Bool = subscription.willRenew
+    let displayName = subscription.displayName
```
tonidero: We should specify the type to make the API tests correctly
```suggestion
    let displayName: String? = subscription.displayName
```

tonidero (approved):
Looks good!

PR #5607, opened 2025-09-26
Disable a flaky PaywallsTester snapshot from Emerge

ajpallares:
This preview snapshot from the PaywallsTester app is constantly flaking in Emerge
image

E.g. here --> https://www.emergetools.com/snapshot/5075db43-27c7-4d86-aed7-01558d6dc825?tab=changed

Since it's not a view from the SDK, we thought it was best to just remove the preview from the Emerge snapshots


facumenzella (approved):
I think this makes sense, but maybe it's worth commenting in the code this is excluded? In case someone actually tries to use it in the future

ajpallares:
> I think this makes sense, but maybe it's worth commenting in the code this is excluded? In case someone actually tries to use it in the future

That's a good point! I can do that 👍 

I still don't know if we'll need to merge this, since I've been seeing lately that this snapshot is not appearing anymore among those modified. I'll keep checking

ajpallares:
Closing as #5782 already addresses this

PR #5609, opened 2025-09-30
Add flag to show / hide user section in CustomerCenter

facumenzella:
### Motivation
We've been getting a few reports about how to hide this section, or other requesting to show all of it 😅 
Instead of overcomplicating things, we're adding a flag in the dashboard to remove the section entirely.

### Description
- Add flag to hide / show section
- Hide / show section based on flag

### DIFF
--- a/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
+++ b/RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
@@ -121,6 +121,10 @@ import Foundation
         configuration?.support.displayVirtualCurrencies == true
     }
 
+    var shouldShowUserDetailsSection: Bool {
+        configuration?.support.displayUserDetailsSection == true
+    }
+
     private let currentVersionFetcher: CurrentVersionFetcher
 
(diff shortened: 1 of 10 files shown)


RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
```diff
@@ -121,6 +121,10 @@ import Foundation
         configuration?.support.displayVirtualCurrencies == true
     }
 
+    var shouldShowUserDetailsSection: Bool {
+        configuration?.support.displayUserDetailsSection == true
```
tonidero: I'm just wondering, this would default to false in case the `configuration` or `displayUserDetailsSection` are null... probably shouldn't happen but wondering if we should default to true in those cases?

tonidero (approved):
Just a question, but looks good to me!

RevenueCatUI/CustomerCenter/ViewModels/CustomerCenterViewModel.swift
facumenzella: You're right, I'll coalesce to true

PR #5614, opened 2025-10-01
Show latest expired if no active subscriptions

facumenzella:
### Motivation
We've noticed that we were not showing the latest expired subscription if there are no subscriptions available.

### Description
- Remove early exit
- Show latest expired if no subs available
- Update tests

### DIFF
--- a/Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
+++ b/Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
@@ -257,7 +257,7 @@ final class CustomerCenterViewModelTests: TestCase {
 
     func testHasAnyPurchasesIsTrueWithOnlyVirtualCurrencies() async throws {
         let mockPurchases = MockCustomerCenterPurchases(
-            customerInfo: CustomerCenterViewModelTests.customerInfoWithExpiredSubscriptions,
+            customerInfo: CustomerCenterViewModelTests.customerInfoWithoutSubscriptions,
             customerCenterConfigData: CustomerCenterConfigData.mock(displayVirtualCurrencies: true)
         )
         mockPurchases.virtualCurrenciesResult = .success(VirtualCurrenciesFixtures.fourVirtualCurrencies)
@@ -269,8 +269,7 @@ final class CustomerCenterViewModelTests: TestCase {
 
         await viewModel.loadScreen()
 
-        expect(viewModel.subscriptionsSection.count) == 1
-        expect(viewModel.subscriptionsSection.first?.isExpired).to(beTrue())
+        expect(viewModel.subscriptionsSection).to(beEmpty())
         expect(viewModel.nonSubscriptionsSection).to(beEmpty())
         expect(viewModel.virtualCurrencies).toNot(beNil())
         expect(viewModel.hasAnyPurchases).to(beTrue())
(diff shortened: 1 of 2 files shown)


Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -276,7 +257,7 @@ final class CustomerCenterViewModelTests: TestCase {
 
     func testHasAnyPurchasesIsTrueWithOnlyVirtualCurrencies() async throws {
         let mockPurchases = MockCustomerCenterPurchases(
-            customerInfo: CustomerCenterViewModelTests.customerInfoWithoutSubscriptions,
+            customerInfo: CustomerCenterViewModelTests.customerInfoWithExpiredSubscriptions,
```
vegaro: I am unsure this needs to be updated since the test is testing a case when there's only VCs

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -288,15 +269,16 @@ final class CustomerCenterViewModelTests: TestCase {
...
+        expect(viewModel.subscriptionsSection.first?.isExpired).to(beTrue())
         expect(viewModel.nonSubscriptionsSection).to(beEmpty())
         expect(viewModel.virtualCurrencies).toNot(beNil())
         expect(viewModel.hasAnyPurchases).to(beTrue())
     }
 
     func testHasAnyPurchasesIsFalseWithVirtualCurrenciesHavingZeroBalance() async throws {
         let mockPurchases = MockCustomerCenterPurchases(
-            customerInfo: CustomerCenterViewModelTests.customerInfoWithoutSubscriptions,
+            customerInfo: CustomerCenterViewModelTests.customerInfoWithExpiredSubscriptions,
```
vegaro: same here

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -1336,7 +1385,7 @@ private extension CustomerCenterViewModelTests {
         """)
     }
 
-    static let customerInfoWithoutSubscriptions: CustomerInfo = {
+    static let customerInfoWithExpiredSubscriptions: CustomerInfo = {
```
vegaro, line 1386: I feel we should have both, `customerInfoWithoutSubscriptions` and `customerInfoWithExpiredSubscriptions` to test all cases correctly

vegaro (commented):
the logic makes sense. Just a question around tests

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
facumenzella: The thing is that `CustomerCenterViewModelTests.customerInfoWithoutSubscriptions` was actually returning an expired subscription. I've just renamed it to `CustomerCenterViewModelTests.customerInfoWithExpiredSubscriptions`

vegaro: Then maybe we should create an actual `customerInfoWithoutSubscriptions` , wdyt?

facumenzella: got it https://github.com/RevenueCat/purchases-ios/pull/5614/commits/4b281b6c2623058be46c05a81275945c234981a9

Tests/RevenueCatUITests/CustomerCenter/CustomerCenterViewModelTests.swift
```diff
@@ -1380,6 +1427,12 @@ private extension CustomerCenterViewModelTests {
         )
     }()
 
+    static let customerInfoWithoutSubscriptions: CustomerInfo = {
```
vegaro, line 1430: this is not used. Did you intend to use it in https://github.com/RevenueCat/purchases-ios/pull/5614/files#diff-a2c613be0d52736ad1a27bda275d71df7e8d16c26bebb96611ff09724d3ad8adR1234?

PR #5615, opened 2025-10-01
fix broken RevenueCat.xcworkspace

facumenzella:
### Motivation
Just noticed this file does not exist anymore


facumenzella:
I need to revisit the danger rule I added. I think that it might not be working as expected

[sadyoda](https://app.graphite.dev/settings/meme-library?org=RevenueCat)

PR #5616, opened 2025-10-01
Fix failing Test Store unit test

ajpallares:
A test was failing (e.g. https://app.circleci.com/pipelines/github/RevenueCat/purchases-ios/30434/workflows/31afc45c-7e88-4f12-963c-05cdb11e6e94/jobs/380961) due to an error in the implementation of the creation of `PurchasedTransactionData` for Test Store purchases. Instead of using the storefront of the `Transaction`, the implementation was fetching `StoreKit`'s `currentStorefront`. As a result, the mocked `Storefront` from the unit tests was not actually being returned, making the test fail.


ajpallares:
@RCGitBot please test

ajpallares:
Merging even with failures in backend-integration tests as they have been addressed in #5617

PR #5623, opened 2025-10-02
Add formatted price to ProductPaidPrice

facumenzella:
### Motivation
We've got a request to match [Android](https://github.com/RevenueCat/purchases-android/blob/85394eb53e51a1719f80e149e06a6d79a27db065/purchases/src/main/kotlin/com/revenuecat/purchases/models/Price.kt#L14) implementation 

### Description
- Add formatted price to `ProductPaidPrice`
- Add tests using SK1 and SK2

### DIFF
--- a/Tests/UnitTests/Identity/ProductPaidPriceTests.swift
+++ b/Tests/UnitTests/Identity/ProductPaidPriceTests.swift
@@ -202,9 +202,9 @@ final class ProductPaidPriceTests: TestCase {
         let amount = 4.99
         let locale = Locale(identifier: "en_US")
         let expectedFormat = "$4.99"
-        
+
         var tasks: [Task<String, Never>] = []
-        
+
         // Create multiple concurrent formatting tasks
         for _ in 0..<50 {
             tasks.append(
@@ -213,13 +213,13 @@ final class ProductPaidPriceTests: TestCase {
                 }
             )
         }
-        
+
         // Collect all results
         var results: [String] = []
         for task in tasks {
             results.append(await task.value)
         }
-        
+
         // All results should be identical
         for result in results {
             expect(result) == expectedFormat
@@ -230,21 +230,21 @@ final class ProductPaidPriceTests: TestCase {
         let currency = "EUR"
         let amount = 7.99
         let locale = Locale(identifier: "fr_FR")
-        
+
         async let price1 = Task { ProductPaidPrice(currency: currency, amount: amount, locale: locale) }
         async let price2 = Task { ProductPaidPrice(currency: currency, amount: amount, locale: locale) }
         async let price3 = Task { ProductPaidPrice(currency: currency, amount: amount, locale: locale) }
         async let price4 = Task { ProductPaidPrice(currency: currency, amount: amount, locale: locale) }
         async let price5 = Task { ProductPaidPrice(currency: currency, amount: amount, locale: locale) }
-        
+
         let (result1, result2, result3, result4, result5) = await (
             price1.value, price2.value, price3.value, price4.value, price5.value
         )
-        
+
         // All should have identical formatted values
         let prices = [result1, result2, result3, result4, result5]
         let firstFormatted = prices.first!.formatted
-        
+
         for price in prices {
             expect(price.formatted) == firstFormatted
             expect(price.currency) == currency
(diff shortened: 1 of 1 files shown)


Tests/UnitTests/Identity/ProductPaidPriceTests.swift
```diff
@@ -0,0 +1,176 @@
...
+        }
+    }
+
+    // MARK: - StoreProduct Consistency Tests
+
+    func testConsistencyWithSK1StoreProductPricing() {
+        // Test that our formatting matches SK1 StoreProduct formatting
+        let usLocale = Locale(identifier: "en_US")
+
+        // Create a mock SK1Product with known price and localexx
```
facumenzella:
```suggestion
        // Create a mock SK1Product with known price and locale
```

Sources/Identity/ProductPaidPrice.swift
```diff
@@ -22,12 +22,46 @@ import Foundation
...
+        self.init(currency: currency, amount: amount, formatted: formatted)
+    }
+
+    /// Formats a price with currency using NumberFormatter
+    /// - Parameters:
+    ///   - amount: The price amount
+    ///   - currency: The currency code
+    ///   - locale: The locale for formatting
+    /// - Returns: Formatted price string (e.g., "$3.00", "€7.99")
+    static func formatPrice(amount: Double, currency: String, locale: Locale = .current) -> String {
```
tonidero: Very nitpicky but should we order the arguments the same way as in the init? 
```suggestion
    static func formatPrice(currency: String, amount: Double, locale: Locale = .current) -> String {
```

Sources/Identity/ProductPaidPrice.swift
```diff
@@ -22,12 +22,46 @@ import Foundation
...
+        let formatter = NumberFormatter()
+        formatter.numberStyle = .currency
+        return formatter
+    }()
+
     /// ProductPaidPrice initialiser
     /// - Parameters:
     ///   - currency: Currency paid
     ///   - amount: Amount paid
-    public init(currency: String, amount: Double) {
```
tonidero, line 29: Hmm interesting, I guess this is not breaking in ObjC because it was never exposed in ObjC 😅. I think it's fine keeping it like this though, we can expose it later if needed.

Sources/Identity/ProductPaidPrice.swift
```diff
@@ -22,12 +22,46 @@ import Foundation
...
+
+    /// Formats a price with currency using NumberFormatter
+    /// - Parameters:
+    ///   - amount: The price amount
+    ///   - currency: The currency code
+    ///   - locale: The locale for formatting
+    /// - Returns: Formatted price string (e.g., "$3.00", "€7.99")
+    static func formatPrice(amount: Double, currency: String, locale: Locale = .current) -> String {
+        numberFormatter.locale = locale
+        numberFormatter.currencyCode = currency
```
tonidero: Nitpick and FFTI, but I've been thinking that this could in theory cause issues if we create ProductPaidPrice in multiple threads, since we're sharing a static numberFormatter... In theory NumberFormatter is thread safe according to docs, but I guess that's within each call, but it does not ensure changes if it would happen in between these 2 calls? Maybe I'm getting it wrong though 😅

Tests/UnitTests/Identity/ProductPaidPriceTests.swift
```diff
@@ -0,0 +1,176 @@
...
+    // MARK: - Static formatPrice Method Tests
+
+    func testFormatPriceWithUSD() {
+        let usLocale = Locale(identifier: "en_US")
+        let formatted = ProductPaidPrice.formatPrice(amount: 9.99, currency: "USD", locale: usLocale)
+
+        expect(formatted) == "$9.99"
+    }
+
+    func testFormatPriceWithEUR() {
```
tonidero, line 73: Nitpick but should we also add a test where the locale does not match the currency? For example, Spain locale for USD?

tonidero (approved):
Just a small concern over the thread safety of the current approach... But I think it's looking good!

Sources/Identity/ProductPaidPrice.swift
facumenzella: I think you're right 🤔 
Let me re-think this

facumenzella: I've added a NSCache (thread safe) based on locales. I think that should be enough

Sources/Identity/ProductPaidPrice.swift
facumenzella: I've reverted this, and made it a convenience initializer (not breaking)

Sources/Identity/ProductPaidPrice.swift
facumenzella: I've also added some concurrency tests

PR #5629, opened 2025-10-03
Make managementURL public in SubscriptionInfo

facumenzella:
### Motivation
We've been asked to make this one public, and figured that there's no reason to not have it public

### Description
- Make it public
- Update API tests


facumenzella:
@RCGitBot please test

ajpallares (approved):
Should we also add ObjC API tests?

PR #5631, opened 2025-10-03
Jzdesign/video component/checksum

JZDesign:
### Checklist

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
@@ -56,6 +56,8 @@ struct VideoComponentView: View {
                 )
             ) { style in
                 if style.visible {
+                    let viewData = style.viewData(forDarkMode: colorScheme == .dark)
+
                     ZStack {
                         if let imageSource, let imageViewModel = try? ImageComponentViewModel(
                             localizationProvider: viewModel.localizationProvider,
@@ -92,7 +94,11 @@ struct VideoComponentView: View {
                                     // this method will share the async task that the cacheprewarming started
                                     // if it didn't error out, expediting the download time and reducing the memory
                                     // footprint of paywalls
-                                    let url = try await fileRepository.generateOrGetCachedFileURL(for: style.url)
+                                    let url = try await fileRepository
+                                        .generateOrGetCachedFileURL(
+                                            for: viewData.url,
+                                            withChecksum: viewData.checksum
+                                        )
                                     guard url != cachedURL else { return }
                                     await MainActor.run {
                                         self.cachedURL = url
@@ -100,17 +106,17 @@ struct VideoComponentView: View {
                                     }
                                 } catch {
                                     await MainActor.run {
-                                        self.cachedURL = style.url
+                                        self.cachedURL = viewData.url
                                         self.imageSource = nil
                                     }
                                 }
                             }
                         }
 
-                        if let cachedURL = fileRepository.getCachedFileURL(for: style.url) {
+                        if let cachedURL = fileRepository.getCachedFileURL(for: viewData.url) {
                             self.cachedURL = cachedURL
                             self.imageSource = nil
-                        } else if let lowResUrl = style.lowResUrl, lowResUrl != style.url {
+                        } else if let lowResUrl = viewData.lowResUrl, lowResUrl != viewData.url {
                             let lowResCachedURL = fileRepository.getCachedFileURL(for: lowResUrl)
                             self.cachedURL = lowResCachedURL ?? lowResUrl
                             self.imageSource = nil
--- a/RevenueCatUI/Templates/V2/Components/Video/VideoComponentViewModel.swift
+++ b/RevenueCatUI/Templates/V2/Components/Video/VideoComponentViewModel.swift
@@ -111,7 +111,9 @@ struct LocalizedVideoPartial: PresentedPartial {
                 padding: otherPartial?.padding ?? basePartial?.padding,
                 margin: otherPartial?.margin ?? basePartial?.margin,
                 border: otherPartial?.border ?? basePartial?.border,
-                shadow: otherPartial?.shadow ?? basePartial?.shadow
+                shadow: otherPartial?.shadow ?? basePartial?.shadow,
+                checksum: otherPartial?.checksum ?? basePartial?.checksum,
+                checksumLowRes: otherPartial?.checksumLowRes ?? basePartial?.checksumLowRes
             )
         )
     }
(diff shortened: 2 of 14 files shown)


RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
```diff
@@ -56,6 +56,8 @@ struct VideoComponentView: View {
                 )
             ) { style in
                 if style.visible {
+                    let viewData = style.viewData(forDarkMode: colorScheme == .dark)
```
JZDesign, line 59: I forgot to handle dark mode in the last pr 🤦

RevenueCatUI/Templates/V2/Components/Video/VideoComponentViewModel.swift
```diff
@@ -154,6 +158,10 @@ struct VideoComponentStyle {
     let margin: EdgeInsets
     let border: ShapeModifier.BorderInfo?
     let shadow: ShadowModifier.ShadowInfo?
+    let checksum: Checksum?
```
rickvdl, line 161: It could make sense to combine a few properties into a sub structure here, maybe something like
```
Source {
  let url: URL
  let checksum: Checksum
 }
```

The same could be done for the high and low res versions and light and dark. Saving some duplicate code down the road. Wdyt?

Sources/Caching/LargeItemCacheType.swift
```diff
@@ -29,20 +30,76 @@ protocol LargeItemCacheType {
...
+
+        // Write any remaining bytes missed during the while loop
+        if !buffer.isEmpty {
+            try fileHandle.write(contentsOf: buffer)
+        }
+
+        // Validate the stored data matches what the server has
+        if let checksum {
+            // If this fails… should we retry?
+            try Checksum
```
rickvdl: If this fails, should the saved data at least be discarded? That way it can at least recover on next download?

Sources/Paywalls/PaywallCacheWarming.swift
```diff
@@ -117,11 +118,14 @@ actor PaywallCacheWarming: PaywallCacheWarmingType {
         let videoURLs = offerings.allLowResVideosInPaywalls
         guard !videoURLs.isEmpty else { return }
 
-        Logger.verbose(Strings.paywalls.warming_up_videos(videoURLs: videoURLs))
+        Logger.verbose(Strings.paywalls.warming_up_videos(videoURLs: Set(videoURLs.map(\.url))))
```
rickvdl: Nitpick: any specific reason why we'd need a Set here? I assume this would cause the logs to print the URLs in an unstable order, making log filtering a bit harder

Sources/Caching/LargeItemCacheType.swift
JZDesign: Good call, I forgot to delete the temporary file.

Sources/Paywalls/PaywallCacheWarming.swift
JZDesign: That's a great question. The primary reason was to ensure deduplication, the secondary reason was that there was no need to convert anything before when I initially wrote this part of the code. Now that we're adding checksums this sticks out like a sore thumb

RevenueCatUI/Templates/V2/Components/Video/VideoComponentViewModel.swift
JZDesign: I actually agree, and I would prefer it. I'll float that idea by the team again, but I believe the reason they want to do this is because the checksum is new and the pattern with images will have to be like this for backward compatibility, so to keep the structures similar we are doing this.

Now, I could do a custom decode and make it behave like this…

JZDesign: If I do this for video, I'll want to do it for images to reduce the pattern sprawl and prep them for having a checksum (which is on our list to do). That is too much to add to this PR. 

I will perform that cleanup in a separate PR

Sources/Paywalls/PaywallCacheWarming.swift
JZDesign: I left it as a set to keep the pattern, but I updated it to log the checksums too.

JZDesign:
@RCGitBot please test

JZDesign:
RC @joshdholtz Bot - Plz review 😆

RevenueCatUI/Templates/V2/Components/Video/VideoComponentViewModel.swift
```diff
@@ -200,9 +212,43 @@ struct VideoComponentStyle {
...
+        self.checksum = checksum
+        self.checksumLowRes = checksumLowRes
+        self.darkChecksum = darkChecksum
+        self.darkChecksumLowRes = darkChecksumLowRes
         self.contentMode = fitMode.contentMode
     }
 
+    func viewData(forDarkMode: Bool) -> ViewData {
+        if forDarkMode {
+            let url = darkUrl ?? url
```
joshdholtz: Can we do something like this here to make it a little more clear on waht's happening with the darkUrl and checksums? 🤷‍♂️ 

```swift
let (resolvedUrl, resolvedChecksum): (URL, Checksum?) = {
    if let darkUrl {
        return (darkUrl, darkChecksum)
    } else {
        return (url, self.checksum)
    }
}()

return .init(
    url: resolvedUrl,
    checksum: resolvedChecksum,
    lowResUrl: darkLowResUrl,
    lowResChecksum: darkChecksumLowRes
)
```

Sources/Caching/FileRepository.swift
```diff
@@ -95,30 +105,34 @@ import Foundation
...
             throw Error.failedToSaveCachedFile(message)
         }
     }
 
-    func generateLocalFilesystemURL(forRemoteURL url: URL) -> URL? {
+    func generateLocalFilesystemURL(forRemoteURL url: URL, withChecksum checksum: Checksum?) -> URL? {
         return cacheURL?
             .appendingPathComponent(url.absoluteString.asData.md5String)
-            .appendingPathExtension(url.pathExtension)
+            .appendingPathExtension(checksum?.value ?? url.pathExtension)
```
joshdholtz, line 135: Why is the `checksum.value` being used as a path extension here? 🤔

JZDesign: To enforce a cache miss on checksum value change

JZDesign: If we only use the URL we will not be able to detect if the underlying video data has updated.

JZDesign: This kind of thing is pretty common. Keep the video file named the same (what it's for, not what it contains) but change its contents when designs change.

If the file name doesn't change when they upload it, the url won't change

PR #5636, opened 2025-10-07
[DX-520] Do not log unknown errors from health report endpoint

polpielladev:
We recently got some reports of 500 responses from the health_report endpoint from some of our users with a large number of products.

We currently show that as an unknown error, which causes confusion. To mitigate this, let's only really show the error if it is an actual valid response. 

### Checklist

### DIFF
--- a/Tests/UnitTests/Misc/SDKHealthManagerTests.swift
+++ b/Tests/UnitTests/Misc/SDKHealthManagerTests.swift
@@ -56,7 +56,7 @@ class SDKHealthManagerTests: TestCase {
         expect(report.status).to(beUnhealthyWithUnknownError())
     }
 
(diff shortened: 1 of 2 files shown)


Tests/UnitTests/Misc/SDKHealthManagerTests.swift
```diff
@@ -56,6 +56,19 @@ class SDKHealthManagerTests: TestCase {
         expect(report.status).to(beUnhealthyWithUnknownError())
     }
 
+    func testHealthReportIsNotLoggedForUnhealthyErrors() async {
```
ajpallares: Did you mean this?
```suggestion
    func testHealthReportIsNotLoggedForUnknownErrors() async {
```

ajpallares (approved):
I think it makes sense! Just one small thing about the test name

PR #5638, opened 2025-10-07
Make an API `public`

ajpallares:
API tests were failing because tests expected certain initializer of `ProductPaidPrice` to be `public`, but it wasn't.

This PR makes that initializer public.

### DIFF
--- a/Sources/Identity/ProductPaidPrice.swift
+++ b/Sources/Identity/ProductPaidPrice.swift
@@ -40,7 +40,7 @@ import Foundation
     ///   - currency: Currency paid
     ///   - amount: Amount paid
     ///   - formatted: Formatted price string with currency symbol
-    init(currency: String, amount: Double, formatted: String) {
+    public init(currency: String, amount: Double, formatted: String) {
         self.currency = currency
         self.amount = amount
         self.formatted = formatted


Sources/Identity/ProductPaidPrice.swift
```diff
@@ -40,7 +40,7 @@ import Foundation
     ///   - currency: Currency paid
     ///   - amount: Amount paid
     ///   - formatted: Formatted price string with currency symbol
-    init(currency: String, amount: Double, formatted: String) {
+    public init(currency: String, amount: Double, formatted: String) {
```
tonidero: Hmm I wonder if we should remove the API test instead... Depends on when it was added I guess... I think it was before the last release so it might make more sense? Mostly so people can't create a formatted text that doesn't have anything to do with the currency/amount

ajpallares: @facumenzella I deduced from the API tests and from the init's documentation that this was supposed to be public. Please, lmk if that's not right

tonidero (approved):
Possible change... But not a strong opinion. Happy to merge this as well.

Sources/Identity/ProductPaidPrice.swift
ajpallares: That's a good point! I assumed it was supposed to be public but I think you are right...

facumenzella: Oh damn, what @tonidero suggests makes sense. `formatted` should be public, but not the initializer

facumenzella: Thanks for catching this

ajpallares: Ok! Changing it! Thanks!

ajpallares: Done in https://github.com/RevenueCat/purchases-ios/pull/5638/commits/abdcf2eeade185f0b692d4a6e718630daf3bb7b3

ajpallares: Done in https://github.com/RevenueCat/purchases-ios/pull/5638/commits/abdcf2eeade185f0b692d4a6e718630daf3bb7b3

ajpallares:
@RCGitBot please test

PR #5640, opened 2025-10-08
Add non paid revenue reporting infra

polmiro:
### Checklist

### Motivation
Implements ad event tracking infrastructure for RevenueCat. This provides the client-side foundation needed to report ad impressions, clicks, and revenue alongside subscription data for comprehensive LTV tracking.

### Description
Adds experimental ad event tracking API with networking infrastructure. Events are **not yet sent to the backend** - the tracking methods instantiate events but don't persist or transmit them (future PR).

**Public API (`@_spi(Experimental)`):**
```swift
// Access via Purchases singleton
let adTracker = Purchases.shared.adTracker

// Track ad impression
await adTracker.trackAdDisplayed(.init(
    networkName: "AdMob",
    mediatorName: .appLovin,
    placement: "home_screen",
    adUnitId: "ca-app-pub-123",
    adInstanceId: "instance-456"
))

// Track ad click/open
await adTracker.trackAdOpened(.init(
    networkName: "AdMob",
    mediatorName: .appLovin,
    placement: "home_screen",
    adUnitId: "ca-app-pub-123",
    adInstanceId: "instance-456"
))

// Track ad revenue
await adTracker.trackAdRevenue(.init(
    networkName: "AdMob",
    mediatorName: .appLovin,
    placement: "home_screen",
    adUnitId: "ca-app-pub-123",
    adInstanceId: "instance-456",
    revenueMicros: 1500000,  // $1.50 in micros
    currency: "USD",
    precision: .exact
))
```

**Key Components:**

1. **AdTracker** - Main API entry point
   - Exposed via `Purchases.shared.adTracker`
   - Requires `@_spi(Experimental)` import
   - Available on iOS 15.0+, tvOS 15.0+, macOS 12.0+, watchOS 8.0+
   - Methods currently instantiate events but don't send them (pending future PR)

2. **Event Types** - Public data structures
   - `AdDisplayed` - Ad impression displayed
   - `AdOpened` - Ad clicked/opened  
   - `AdRevenue` - Ad generated revenue with amount, currency, precision
   - `MediatorName` - Struct for mediation network names (.appLovin, .adMob, or custom)
   - `AdRevenue.Precision` - Revenue accuracy level (.exact, .publisherDefined, .estimated, .unknown)

3. **Internal Event Model** - Type-safe routing
   - `AdEvent` enum: `.displayed`, `.opened`, `.revenue`
   - Conforms to `FeatureEvent` protocol for routing
   - Uses `Feature.ads` discriminator
   - Includes creation metadata (ID, timestamp)

4. **Networking Infrastructure** (ready for future use)
   - `EventsRequest.AdEventRequest` - Request serialization with snake_case conversion
   - `AdHTTPRequestPath` - Configures `https://a.revenue.cat/v1/events` endpoint
   - `PostEventsOperation` - **Generic** operation for posting events (eliminates duplication)
   - HTTP path routing via `HTTPRequest.AdPath.postEvents`

**Architecture Improvements:**

- **Consolidated PostEventsOperation**: Replaced duplicate `PostPaywallEventsOperation` and `PostAdEventsOperation` with a single generic `PostEventsOperation` that accepts a path parameter
  - Eliminates ~100 lines of duplicated code
  - Supports both `HTTPRequest.PaywallPath` and `HTTPRequest.AdPath` via runtime type checking
  - Single source of truth for event posting logic

- **Removed Premature API**: Deleted `InternalAPI.postAdEvents()` methods that incorrectly assumed `StoredEvent` storage would be reused
  - Storage strategy still to be determined in future PR
  - Keeps all networking infrastructure intact

**Event Schema:**
Follows standard events format with:
- `app_user_id`, `app_session_id`, `timestamp_ms`
- `network_name`, `mediator_name`, `placement`  
- `ad_unit_id`, `ad_instance_id`
- For revenue: `revenue_micros`, `currency`, `precision`

**Compiler Flag:**
Entire feature wrapped in `#if ENABLE_AD_EVENTS_TRACKING` to allow conditional compilation.

**Testing:**
- 7 unit tests covering all three event types
- Snapshot tests for JSON serialization (snake_case conversion)
- Tests verify event instantiation (not persistence/sending)
- All tests passing ✅

**Platform Support:**
- AdTracker: iOS 15.0+, tvOS 15.0+, macOS 12.0+, watchOS 8.0+ (requires `async`)
- Event data structures: No version restrictions (can be used on any supported iOS)

**Future Work (separate PRs):**
- Event persistence/storage strategy
- Event flushing and backend transmission
- Integration with ad mediation SDKs

### DIFF
--- a/Sources/Ads/AdTracker.swift
+++ b/Sources/Ads/AdTracker.swift
@@ -86,6 +86,7 @@ public final class AdTracker: NSObject {
      */
     @_spi(Experimental) public func trackAdDisplayed(_ data: AdDisplayed) async {
         let event = AdEvent.displayed(.init(id: UUID(), date: Date()), data)
+        await self.eventsManager?.track(featureEvent: event)
     }
 
     /**
@@ -108,6 +109,7 @@ public final class AdTracker: NSObject {
      */
     @_spi(Experimental) public func trackAdOpened(_ data: AdOpened) async {
         let event = AdEvent.opened(.init(id: UUID(), date: Date()), data)
+        await self.eventsManager?.track(featureEvent: event)
     }
 
     /**
@@ -134,6 +136,7 @@ public final class AdTracker: NSObject {
      */
     @_spi(Experimental) public func trackAdRevenue(_ data: AdRevenue) async {
         let event = AdEvent.revenue(.init(id: UUID(), date: Date()), data)
+        await self.eventsManager?.track(featureEvent: event)
     }
 
 }


Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1393,6 +1413,11 @@ public extension Purchases {
         await self.paywallEventsManager?.track(featureEvent: paywallEvent)
     }
 
+    /// Used by ad SDKs to keep track of ``AdEvent``s.
+    @_spi(Internal) func track(adEvent: AdEvent) async {
+        await self.adEventsManager?.track(featureEvent: adEvent)
+    }
```
JayShortway: Are app developers going to call this function in the callbacks from their ad sdk? If so, we should remove the `@_spi(Internal)` annotation, as that prevents them from doing that.

We use `@_spi(Internal)` to mark otherwise public APIs as internal. This is needed when we want to expose an internal API from `RevenueCat` to `RevenueCatUI`. In `RevenueCatUI` we basically "import RevenueCat with internal APIs available". This is used for Paywall events, as `RevenueCatUI` should be able to tell `RevenueCat` that a Paywall event has occurred. But this is not something that app developers should call on their own.

Ad events are different afaik, but maybe I'm misunderstanding!

Sources/Ads/Events/Networking/EventsRequest+Ad.swift
```diff
@@ -0,0 +1,127 @@
...
+extension EventsRequest.AdEvent: Encodable {
+
+    /// When sending this to the backend `JSONEncoder.KeyEncodingStrategy.convertToSnakeCase` is used
+    private enum CodingKeys: String, CodingKey {
+
+        case id
+        case version
+        case type
+        case appUserID = "appUserId"
+        case sessionID = "appSessionId"
```
JayShortway: I don't feel very strongly about this, but I see we're using `Id` below (`adUnitId`, `adInstanceId`). We might as well do that here too and avoid the string mapping for these? 

```suggestion
        case appUserId
        case sessionId
```

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,207 @@
...
+    var feature: Feature {
+        return .ads
+    }
+
+    var eventDiscriminator: String? {
+        return nil
+    }
+
+    /// An ad impression was displayed.
+    case displayed(CreationData, Data)
```
JayShortway: This makes me think: is this only the initial time the ad was displayed, or every impression? E.g. what if it's scrolled out and into view a few times? If that's supposed to be counted multiple times, it might be clearer to call this `impression`?

Sources/Ads/Events/AdEventsManager.swift
```diff
@@ -0,0 +1,100 @@
...
+
+    /// - Throws: if posting events fails
+    /// - Returns: the number of events posted
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+    func flushEvents(count: Int) async throws -> Int
+
+}
+
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+actor AdEventsManager: AdEventsManagerType {
```
JayShortway: I know it's confusing, but Customer Center events also use the `PaywallEventsManager`. Could we make the `PaywallEventsManager` support ad events too? (And then maybe think about renaming that in a separate PR haha.) The reason why I'm asking is because it looks like `AdEventsManager` is doing almost the same thing as `PaywallEventsManager`. 

Same with `AdEventStore` and `PaywallEventStore`. (If we want to keep the stores different for some reason, we could have 2 instances of `PaywallEventsManager`, each with their own instance of `PaywallEventsStore`.)

Sources/Ads/Events/AdEventsManager.swift
```diff
@@ -0,0 +1,100 @@
...
+                                             feature: featureEvent.feature,
+                                             appSessionID: self.appSessionID,
+                                             eventDiscriminator: featureEvent.eventDiscriminator) else {
+            Logger.error(Strings.ads.event_cannot_serialize)
+            return
+        }
+        await self.store.store(event)
+    }
+
+    func flushEvents(count: Int) async throws -> Int {
```
JayShortway: I don't think this is being called anywhere currently?

Regarding my previous comment, if we are able to use `PaywallEventsManager` (and keep it at a single instance), it would get flushed whenever we already flush paywall (and customer center) events now. If not, we should make sure the instance we end up using also gets flushed.

Sources/Ads/Events/AdEvent.swift
polmiro: It's the initial time the ad was displayed. But the wording here is a bit tricky for two reasons:
- The two major mediators do not use the same language:
    - AdMob has this callback named [onAdImpression](https://developers.google.com/admob/android/banner#ad_events), slightly different names per ad format but all around the word `ad impression`.
    - AppLovin has this callback named [onAdDisplayed](https://support.axon.ai/en/max/android/ad-formats/app-open-ads/). Interestingly they have it deprecated for banner ad formats. I assume this is because a successful load can be accounted as the impression given it's permanent on the screen?
- What the `display` or `impression` exactly means is somewhat dependent on the ad format. For example, a banner ad:
    - AdMob fires the onAdImpression every time the banner has it's content refreshed (let's say every 30 seconds).
    - AppLovin does not fire a new onAdImpression. They recommend on the ad revenue events. It's a tricky topic and I am not 100% understanding yet why the follow a different approach.

In our case I had picked `displayed` since `impression` seems more archaic and less fitting for the newer ad formats.

Sources/Ads/Events/Networking/EventsRequest+Ad.swift
polmiro: Sounds good, I followed the same pattern as `EventRequest+Paywall` just in case. But if there's no good apparent reason I agree it's better to just avoid the mapping.

Sources/Ads/Events/AdEventStore.swift
```diff
@@ -0,0 +1,231 @@
...
+//  AdEventStore.swift
+//
+//  Created by RevenueCat on 1/8/25.
+
+import Foundation
+
+protocol AdEventStoreType: Sendable {
+
+    /// Stores `event` into the store.
+    @available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
```
rickvdl: Can the @available attribute be applied to the full protocol instead of all methods as well?

Sources/Ads/Events/AdEventStore.swift
```diff
@@ -0,0 +1,231 @@
...
+
+    private static let fileManager: FileManager = .default
+
+}
+
+// MARK: - Messages
+
+// swiftlint:disable identifier_name
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+private enum AdEventStoreStrings {
```
rickvdl: Maybe strings is a bit generic, if it's only used for logging maybe the name could mention that as well

Sources/Ads/Events/AdEventsManager.swift
polmiro: It is indeed! I was feeling awkward with so much duplication but I was a bit afraid of touching too much. Thanks for the callout about customer center, I had not realized they go through the same manager. I'll try to take a stab at it.

Sources/Ads/Events/AdEventsManager.swift
polmiro: Indeed. That made me realize I'm missing some sort of integration test too. I'll be removing this class since I will consolidate it with `PaywallEventsManager`.

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,207 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  AdEvent.swift
+//
+//  Created by RevenueCat on 1/8/25.
+
+import Foundation
+
+/// An event to be sent for RC Ads tracking.
+public enum AdEvent: FeatureEvent {
```
ajpallares: We've got a thing with public enums...

Currently, the SDK does not allow adding new cases to existing enums without breaking the API (the reason is that currently integrators are not forced to deal with the `default` case in a `switch` statement).

Is this enum `frozen` in the sense that we're completely sure that no extra cases will be added in the future. It's totally OK if we can't be sure, that's the most common scenario, but, in that case we may need to either look for alternatives or mark this API as beta for now (making it explicit that we might be changing it in the future).

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,207 @@
...
+
+    /// An ad impression generated revenue.
+    case revenue(CreationData, RevenueData)
+
+}
+
+extension AdEvent {
+
+    /// The creation data of an ``AdEvent``.
+    public struct CreationData {
```
ajpallares: Is there anything missing here? Or perhaps it's only the naming. The type is called `CreationData` but it does not really contain any data. I don't know, maybe `CreationMetadata` or just `Metadata` would be more appropriate?

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,207 @@
...
+
+        // swiftlint:disable missing_docs
+        public var id: ID
+        public var date: Date
+
+        public init(
+            id: ID = .init(),
+            date: Date = .init()
+        ) {
```
ajpallares: What's the use case for this? Is this something that the SDK always generates or are developers expected to pass their `id` and `date` for this object?
Depending on the intended way of usage, we should either add documentation to these or maybe not make them all public.

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,207 @@
...
+    /// The content of a revenue ``AdEvent``.
+    public struct RevenueData {
+
+        // swiftlint:disable missing_docs
+        public var networkName: String
+        public var mediatorName: String
+        public var placement: String?
+        public var adUnitId: String
+        public var adInstanceId: String
+        public var sessionIdentifier: SessionID
```
ajpallares: I see that all thee properties are exactly the same as in `AdEvent.Data`. To avoid this duplication, I see two potential solutions:
1. Replace these with 
```suggestion
        public var data: Data
```
2. Remove these properties from `RevenueData` and change the `case revenue(CreationData, RevenueData)` to contain both `Data` and `RevenueData`
```swift
case revenue(CreationData, Data, RevenueData)
```

Which one of the two to take, I'm not sure and should probably depend on how we see these APIs being used. 

I think I would go with solution 2 since we would be making a clear differentiation between the Event data and the Revenue data, as it is not clear that `RevenueData` should contain the event `Data`

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,207 @@
...
+        // swiftlint:enable missing_docs
+
+    }
+
+}
+
+extension AdEvent {
+
+    /// The content of an ``AdEvent``.
+    public struct Data {
```
ajpallares: Perhaps it'd be a bit redundant, but since there's Foundation's `Data`, I would rename this to `EventData`
```suggestion
    public struct EventData {
```

Which makes it more explicit and easier to differentiate with `RevenueData` and `CreationData` (or whatever we end up naming it)

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,207 @@
...
+            self.precision = precision
+        }
+        // swiftlint:enable missing_docs
+
+    }
+
+}
+
+/// Enum representing the level of accuracy for reported revenue values.
+public enum Precision: String, Codable, Sendable {
```
ajpallares: This is a total nit, but sine we're declaring conformance to `Codable` and `Sendable` at the end of the file, I would move these to the end of the file as well

```
extension Precision: Codable, Sendable {}

Comment (ajpallares) [inline Sources/Ads/Events/AdEvent.swift]: Also, I think this `Precision` should be namespaced within `AdEvent.RevenueData`
```swift
extension AdEvent.RevenueData {
  public enum Precision: String, Codable, Sendable { 
    ...
  }
}
```

Otherwise, developers will constantly see `Precision` as autocompletion suggestion when they type `P` (e.g. for `Purchases`), which would be annoying.

Sources/Ads/Events/AdEventsManager.swift
polmiro: I just pushed a new commit to address this. I had to group events by ad related vs paywall/customer_center to make sure they are sent to the right domain. Context on that [here](https://docs.google.com/document/d/1sAAC1O959_o476x65wqmL6dlfJWB7IO475Pfs228W3w/edit?disco=AAABrkdhqRo). Let me know what do you think.

Sources/Ads/Events/AdEventsManager.swift
polmiro: Closing since I consolidated into `PaywallEventsManager`.

Sources/Ads/Events/AdEventStore.swift
polmiro: In the end since I am consolidating the events manager the file has been removed.

Sources/Ads/Events/AdEventStore.swift
polmiro: This file is gone after the consolidation with the `PaywallEventsManager`.

Sources/Ads/Events/AdEvent.swift
polmiro: I took the pattern from [PaywallEvent.CreationData](https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Paywalls/Events/PaywallEvent.swift#L51). I'm in the process of redoing the public API though so I may change this.

Sources/Ads/Events/AdEvent.swift
polmiro: I took the pattern from [PaywallEvent.CreationData](https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Paywalls/Events/PaywallEvent.swift#L51). I'm in the process of redoing the public API though so I may change this.

Sources/Purchasing/Purchases/Purchases.swift
polmiro: Method is now public with example docs. Let me know what do you think about the exposed API, there's a couple of options on how to approach it and I just went for one that made sense to me, I don't have a strong opinion and would be happy to change it.

Marking this as resolved though.

Sources/Ads/Events/AdEvent.swift
polmiro: There is now `ImpressionData` and the API also changed considerably. Marking this resolved. Let me know what you think about the new API approach, happy to keep iterating.

Sources/Ads/Events/AdEvent.swift
polmiro: Now it's `ImpressionData`.

Sources/Ads/Events/AdEvent.swift
polmiro: I kept the pattern for the internal SDK initialization part which is now not public. Marking as resolved. If you have a further improvement suggestion let me know!

Sources/Ads/Events/AdEvent.swift
polmiro: Since it already exists in `PaywallEvent` I will leave it as it is. But I may do a second PR to rename and consolidate things away from `PaywallEventsManager` to something broader that includes the customer events and ad events in the name. Maybe I can tackle these naming improvements as part of that effort.

Sources/Ads/Events/AdEvent.swift
polmiro: The enum is now only internal and the API changed considerably. Let me know if you have any feedback in the new API. I will close this as it's not an issue anymore. Thanks!

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1393,6 +1394,91 @@ public extension Purchases {
...
+         precision: .exact
+     ))
+     ```
+     */
+    func trackAdRevenue(_ data: AdRevenue) async {
+        await self.paywallEventsManager?.track(featureEvent: AdEvent.revenue(
+            .init(id: UUID(), date: Date()),
+            data
+        ))
+    }
```
polmiro: This is what using the new API method would look like. Looking for feedback here on the ergonomics of the API:
1. AdImpressionData can be shared for all events of the same ad instance and could potentially be reusable across events. But depending on how these callbacks are hooked into the lifecycle of the ad it may be over-complicated API and duplicating the fields in each event could be a better option.
2. Is there a specific format monetary amounts are usually exposed in the API? It could be a good idea to use it for consistency's sake.
3. Is it okay to use an enum for the precision? I don't foresee this enum value list changing but ... it's also very early project too.

Sources/Ads/Events/Networking/AdHTTPRequestPath.swift
```diff
@@ -0,0 +1,63 @@
...
+//  AdHTTPRequestPath.swift
+//
+//  Created by RevenueCat on 1/8/25.
+
+import Foundation
+
+extension HTTPRequest.AdPath: HTTPRequestPath {
+
+    // swiftlint:disable:next force_unwrapping
+    static let serverHostURL = URL(string: "https://e.revenue.cat")!
```
polmiro: @aboedo per our discussion on the doc I am using `e.revenue.cat` instead of `revenuecat.com`. I just want to make sure it does not go unnoticed.

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,210 @@
...
+import Foundation
+
+// MARK: - Public Types
+
+/// Shared impression data for all ad events.
+public struct AdImpressionData {
+
+    // swiftlint:disable missing_docs
+    public var networkName: String
+    public var mediatorName: String
```
JayShortway: Are we at all worried about differences due to capitalization/hyphens/underscores? Or are we normalizing on the backend?

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1393,6 +1394,91 @@ public extension Purchases {
...
+         networkName: "AdMob",
+         mediatorName: "MAX",
+         placement: "home_screen",
+         adUnitId: "ca-app-pub-123",
+         adInstanceId: "instance-456"
+     )
+     await Purchases.shared.trackAdDisplayed(.init(impression: impression))
+     ```
+     */
+    func trackAdDisplayed(_ data: AdDisplayed) async {
```
JayShortway: Can app developers use this already? If not, we should probably merge all public types behind a compiler flag, and then open it up once the feature is live. Otherwise people might be confused what this is for, and try to use it, and perhaps even open support tickets.

Let me know if you need help with the compiler flag!

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,210 @@
...
+//  AdEvent.swift
+//
+//  Created by RevenueCat on 1/8/25.
+
+import Foundation
+
+// MARK: - Public Types
+
+/// Shared impression data for all ad events.
+public struct AdImpressionData {
```
JayShortway: Not entirely sure whether the name `AdImpressionData` is generic enough for the different kinds of ad events we're tracking with this. What about `AdEventData` or even just `AdData`?

Sources/Paywalls/Events/PaywallEventsManager.swift
```diff
@@ -76,22 +76,60 @@ actor PaywallEventsManager: PaywallEventsManagerType {
...
+    }
 
-            throw error
+    private func postEventsByFeature(_ events: [StoredEvent]) async -> (
+        allSucceeded: Bool,
+        shouldClearEvents: Bool,
+        lastError: Error?
+    ) {
+        let adEvents = events.filter { $0.feature == .ads }
+        let nonAdEvents = events.filter { $0.feature != .ads }
```
JayShortway: How many events are we usually dealing with here, before posting? Might be nice to avoid looping twice. I'd suggest [`partition()`](https://developer.apple.com/documentation/swift/array/partition(by:)-90po8), but it modifies the array which is not great. We can always create 2 mutable arrays and fill them with a single loop. Something along these lines:

```swift
var adEvents: [Event] = []
var nonAdEvents: [Event] = []

for event in events {
    if event.feature == .ads {
        adEvents.append(event)
    } else {
        nonAdEvents.append(event)
    }
}
```

Sources/Purchasing/Purchases/Purchases.swift
JayShortway: 1. I don't see a ton of value in the `AdImpressionData` object, but I'm not against it either.
2. We (and StoreKit) expose product prices as [`Decimal`](https://developer.apple.com/documentation/foundation/decimal), but asking app developers to construct that might be error prone? Not sure. Might be simpler to ask for micros. How do AdMob and AppLovin expose this in their callbacks?
3. I would not use an enum for the sake of API evolution. We can use constants (e.g. integers) or use type  properties. For an example of the latter see [`Message.Reason`](https://developer.apple.com/documentation/storekit/message/reason-swift.struct).

cc @RevenueCat/coresdk for thoughts and suggestions.

Sources/Ads/Events/AdEvent.swift
JayShortway: Thanks for explaining it extensively! That's very annoying that they don't use the same approach. I think `displayed` is a good choice indeed!

Sources/Ads/Events/AdEventsManager.swift
JayShortway: Thanks for making this change, I think this makes sense!

Sources/Purchasing/Purchases/Purchases.swift
polmiro: 1. ~Same here, I am fairly indifferent. Happy to rework this if there's a preference.~ Update: I decided to remove `AdImpressionData`. I do think it makes a simpler API. Most SDKs follow the same pattern here.
2. Makes sense. AdMob uses micros and AppLovin uses on USD's in dollars. Same here, happy to rework this if there's a preference.
3. ~Thanks! I'll take care of that soon.~ Update: used a similar pattern to `Message.Reason`.

Sources/Ads/Events/AdEvent.swift
polmiro: Good point. I'll rename.

Sources/Paywalls/Events/PaywallEventsManager.swift
polmiro: It's difficult to say, given this happens on sync events and a bit difficult to estimate how many ads and how many events will be fired until the next sync:
- app initialization
- app foregrounded
- app backgrounded
- after successful purchase

But there are also further concerns I am uncovering:
- what happens if the events accumulate.
- what happens if the app stays foregrounded for long.

I'll look into the suggestions, alternatively we could also store them in separate arrays altogether.

I am also wondering if we should also just send all events via `e.revenue.cat` and simplify this 🤔 . The only risk is the domain blocking of `e.revenue.cat` would also block paywall, customer center, and diagnostic events. Doesn't it of make sense to keep all these separated from the main requests and not only ad revenue data?

Sources/Ads/Events/AdEvent.swift
polmiro: The thought did cross my mind and I did consider we could solve it in the backend needed be. But honestly I did not think through all that much. Do you have any suggestions to mitigate this at the API level without limiting the possible networks and mediators?

JayShortway: @ajpallares What do you think is the best way to do this in Swift without using enums? For context, we want to see if we can tighten up these fields for known networks and mediators, but still allow for custom values. So for instance, the `network` options we should support are `AdMob`, `AppLovin` and "custom string".

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,216 @@
...
+//
+//  AdEvent.swift
+//
+//  Created by RevenueCat on 1/8/25.
+
+import Foundation
+
+// MARK: - Public Types
+
+#if ENABLE_AD_TRACKING
```
JayShortway: Maybe `NON_SUBSCRIPTION_REVENUE_REPORTING`, just in case?

Sources/Ads/Events/AdEvent.swift
polmiro: In the end, I am removing `AdImpressionData` entirely in favor of a protocol shared by the three types and a simpler public API. Closing this one.

polmiro: Resolving since I am removing `AdImpressionData` from the public API and renaming internally

Sources/Ads/Events/AdEvent.swift
tonidero: Hmm in Kotlin, I guess I would create a class/interface and add the types as static constants, but the type would still need to be String to accept arbitrary values. So something like (in Kotlin):
```
object AdNetwork {
   static const AD_MOB = "AdMob"
   static const APP_LOVIN = "AppLovin"
   ...
}
// And then when creating the impression data
AdImpressionData(
   networkName: AdNetwork.AD_MOB,
   ...
)
or 
AdImpressionData(
   networkName: "unknown_network",
   ...
)
```

Not sure if something like that works for Swift?

Sources/Paywalls/Events/PaywallEventsManager.swift
```diff
@@ -76,22 +76,60 @@ actor PaywallEventsManager: PaywallEventsManagerType {
...
+    }
+
+    private func postEvents(
+        _ events: [StoredEvent],
+        using poster: ([StoredEvent]) async throws -> Void
+    ) async -> (succeeded: Bool, shouldClear: Bool, error: Error?) {
+        do {
+            try await poster(events)
+            Logger.debug(Strings.analytics.flush_events_success)
+            return (true, false, nil)
```
tonidero: Hmm shouldn't `shouldClear` be true on a success?

Edit: Hmm right, I guess the way we have it, it needs to be false, so we only clear if both requests succeed... We might want to add some comments explaining this logic, or better, apply my other suggestion to split the storage for the new type of event.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1393,6 +1394,95 @@ public extension Purchases {
...
+     await Purchases.shared.trackAdRevenue(.init(
+         impression: impression,
+         revenueMicros: 1500000,  // $1.50
+         currency: "USD",
+         precision: .exact
+     ))
+     ```
+     */
+    func trackAdRevenue(_ data: AdRevenue) async {
+        await self.paywallEventsManager?.track(featureEvent: AdEvent.revenue(
```
ajpallares: I'm OK with reusing `paywallEventsManager` for this as well, but I think it may be worth renaming it into something more generic to avoid any potential confusion

Sources/Paywalls/Events/PaywallEventsManager.swift
```diff
@@ -76,22 +76,60 @@ actor PaywallEventsManager: PaywallEventsManagerType {
 
         Logger.verbose(Strings.paywalls.event_flush_starting(count: events.count))
 
-        do {
-            try await self.internalAPI.postPaywallEvents(events: events)
-            Logger.debug(Strings.analytics.flush_events_success)
+        let (allSucceeded, shouldClearEvents, lastError) = await self.postEventsByFeature(events)
 
+        if allSucceeded || shouldClearEvents {
             await self.store.clear(count)
```
tonidero, line 83: Hmm so I believe the current code will make sure we won't clear unless both types of events have been synced (if both present) but does feel like quite complex logic and has a lot of potential edge cases and can easily cause us to post the same events multiple times (which I guess the backend should be prepared for but still)... It also would get very complex if we ever wanted to add another type of event.

I believe this would be much cleaner if we separate the entire storage for the different types of events. So CustomerCenter + Paywall events go to one storage, while these new events go to a different storage. I think this might force us to create a different event manager to be used for this new types of events so there will be some code duplication, but I feel that would be preferrable. Wdyt @polmiro @JayShortway ?

JayShortway: I think those are valid concerns! Can we keep the same manager, but use different instances, each with their own storage?

polmiro: That's what was before in the PR before I refactored to reuse `PaywallsEventsManager`.

One thing I noted in a different comment is the following:
> I am also wondering if we should also just send all events via e.revenue.cat and simplify this 🤔 . The only risk is the domain blocking of e.revenue.cat would also block paywall, customer center, and diagnostic events. Doesn't it of make sense to keep all these separated from the main requests and not only ad revenue data?

What are your thoughts on this one @tonidero ?

tonidero: Hmm we should notify the relevant teams of the risk of some adblockers mistakenly blocking all these events (FWIW, the diagnostics endpoint is already blocked in some adblockers)... I don't think there is any critical piece of information that we need from these events, but would be good to confirm with them. 

It would certainly simplify this code a lot, and we already use a separate endpoint for those events, so feels like a known risk. While this may increase the chances of it being blocked a bit, I think it might be worth it IMO.

JayShortway: >Can we keep the same manager, but use different instances, each with their own storage?

Continuing on this train of thought, could the `PaywallEventsManager` receive the endpoint URL in its constructor parameters? That way each instance can send the events where they need to go.

tonidero: That could work. One alternative could be to keep it to a single manager, but add a separate track method that only accepts the new AdEvent types, then add some check in the existing track event to make sure we don't accidentally track AdEvent's with that one. That way, we can contain it to a single manager and have the logic inside.

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,269 @@
...
+
+}
+
+extension AdEvent {
+
+    /// Internal creation metadata that is automatically generated by the SDK.
+    internal struct CreationData {
+
+        internal var id: ID
+        internal var date: Date
```
tonidero, line 259: For paywall/customer center we call this `timestamp`. Should we try to keep consistency here?

Sources/Paywalls/Events/PaywallEventsManager.swift
JayShortway: Hmm, multiple `track()` functions sounds error prone. I'd prefer to keep the complexity inside the `PaywallEventsManager` in that case. We can test it thoroughly to make sure it works as expected, add some comments as you suggested, and maybe have the private methods return private classes (e.g. `PostEventsResult`) instead of tuples.

Sources/Ads/Events/AdEvent.swift
polmiro: I took a stab ad adding it as a `MediatorName` wrapper type, see [this commit](https://github.com/RevenueCat/purchases-ios/pull/5640/commits). If that looks good feel free to resolve the comment, otherwise I can undo it or iterate on a different approach.

Sources/Ads/Events/AdEvent.swift
polmiro: I renamed the internal model for consistency but kept the serialized value sent to the backend as `timestamp_ms`. It is more explicit given UNIX timestamp are sometimes modeled as seconds and not miliseconds.

Sources/Purchasing/Purchases/Purchases.swift
polmiro: @tonidero do you have any preference/thoguhts on this one?

tonidero: Oh sorry I missed this one 🤦 
1. Right, I think it's better to try to keep a flatter API. I think the current one makes sense to me.
2. Micros helps to avoid floating point errors mostly so I'm biased toward that, but not a strong opinion if we want to use Decimal. Probably I would transform to micros for the event storage though, so when we post it, we don't post decimal numbers
3. As @JayShortway said, I think we should avoid using enums, so using constants like the pattern for Message.Reason makes the most sense yeah 👍

polmiro: Cool, seems we are all aligned then. Thanks!

Sources/Paywalls/Events/PaywallEventsManager.swift
tonidero: > multiple track() functions sounds error prone

Well, I guess it depends... If we split the new events to not be of type `FeatureEvent`, the compiler would guarantee we can only track the correct types on each track method. Kinda like I'm doing in the Android PR [here](https://github.com/RevenueCat/purchases-android/pull/2728/files#diff-707ff225a6677ba1592bb698e2775162b83cd047dd2dff1857c6f0726c336e66R162-R170)

Sources/Purchasing/Purchases/Purchases.swift
tonidero: Hmm actually rethinking about 3. Are we planning to keep the APIs internal? If so, it would be ok to keep this as an enum

polmiro: The latest is to keep them public given there is a use case to consider for people who will want to track this revenue via RC but not necessarily integrate any further.

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1393,6 +1394,90 @@ public extension Purchases {
...
+     await Purchases.shared.trackAdDisplayed(.init(
+         networkName: "AdMob",
+         mediatorName: .appLovin,
+         placement: "home_screen",
+         adUnitId: "ca-app-pub-123",
+         adInstanceId: "instance-456"
+     ))
+     ```
+     */
+    func trackAdDisplayed(_ data: AdDisplayed) async {
```
tonidero: Would it make sense to create some `AdTracker` class that has all these methods? Mostly to avoid polluting the Purchases API with too many methods.

polmiro: I like the idea, specially knowing we may be adding a bunch more down the line: loaded, failed to load, other format specific events.

Sources/Paywalls/Events/PaywallEventsManager.swift
polmiro: I implemented your suggestion. I still don't think it's the ideal result per the other conversation that's opened. Right now if one of the requests fails, we will end up re-attempting both. Going to mark this thread resolved though and focus on the other one.

Sources/Paywalls/Events/PaywallEventsManager.swift
polmiro: It should be much clearer now that I implemented @JayShortway suggestion. But as you point out, the fact that if one request fails both will be retried is not ideal. Going to to mark this resolved and we can keep the discussion about multiple storages, event managers and so on in the other thread.

Sources/Paywalls/Events/PaywallEventsManager.swift
polmiro: I followed the suggestion from @JayShortway and the implementation simplified considerably. There's still the caveat that if one request fail both will be have to retried which is far from ideal.

Given we still have the compiler flag this should have no effect though. My proposal would be to implement the refactor in a separate PR. Storages will be kept separate, and for event managers there's a couple of options I would like to discuss but rather make a proposal in the new PR itself.

How's this sound @JayShortway / @tonidero ?

Sources/Purchasing/Purchases/Purchases.swift
polmiro: What would the API look like though?
```
Purchases.shared.adTracker.trackAdRevenue(...)
```
Or with its own shared singleton?
```
Purchases.AdTracker.shared.trackAdRevenue(...)
```

tonidero: I would think the first. Better to keep it as part of the purchases instance.

polmiro: Makes sense. I will implement this tomorrow and mark it resolved then.

Sources/Paywalls/Events/PaywallEventsManager.swift
JayShortway: Sounds good to me! Makes sense to do it in a follow-up PR.

Sources/Ads/Events/Networking/EventsRequest+Ad.swift
```diff
@@ -0,0 +1,136 @@
...
+//
+//  Created by RevenueCat on 1/8/25.
+
+import Foundation
+
+#if ENABLE_AD_EVENTS_TRACKING
+
+extension EventsRequest {
+
+    struct AdEvent {
```
ajpallares: I know this is similar to how it's done for `PaywallEvent`s, but I still think it's a bit confusing to have two different objects called `AdEvent`. I'm checking for Customer Center and there's `CustomerCenterEvent` and `CustomerCenterEventBaseRequest`, which it's good to more easily differentiate between the two objects. 

Also, making the names different could prevent some potential future issues (e.g. if `EventsRequest.AdEvent` eventually becomes `Decodable`, then the line `let adEvent = try JSONDecoder.default.decode(AdEvent.self, from: jsonData)` would be ambiguous and the compiler would complain).

In this case, I would call this object `AddEventRequest` (similar to the Customer Center one, but no need of the `Base` word since it's a struct that cannot be subclassed).

Sources/Ads/Events/Networking/EventsRequest+Ad.swift
```diff
@@ -0,0 +1,136 @@
...
+            return nil
+        }
+    }
+
+    private static let version: Int = 1
+
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+private extension AdEvent {
```
ajpallares, line 96: To add to my previous point. This `AdEvent` is the other `AdEvent` object, which might not be obvious at first glance. So I think they should be named differently (and not only based on namespace)

Sources/Ads/Events/Networking/EventsRequest+Ad.swift
```diff
@@ -0,0 +1,136 @@
...
+
+    /// When sending this to the backend `JSONEncoder.KeyEncodingStrategy.convertToSnakeCase` is used
+    private enum CodingKeys: String, CodingKey {
+
+        case id
+        case version
+        case type
+        case appUserId
+        case appSessionId
+        case timestamp = "timestampMs"
```
ajpallares, line 122: Just to double check. Are we sure the field should be called "timestamp_ms"? I ask because `EventRequest.PaywallEvent` and `EventRequest.CustomerCenterEventBaseRequest` both use just "timestamp"
https://github.com/RevenueCat/purchases-ios/blob/87c47b3416c6ea9deed2ec4db693c7becf774218/Sources/Paywalls/Events/Networking/EventsRequest%2BPaywall.swift#L110
https://github.com/RevenueCat/purchases-ios/blob/87c47b3416c6ea9deed2ec4db693c7becf774218/Sources/CustomerCenter/Events/EventsRequest%2BCustomerCenter.swift#L223

Sources/Ads/Events/Networking/PostAdEventsOperation.swift
ajpallares, line 1: Not for this PR, but perhaps we should try to remove this duplication, as this is basically the same as `PostPaywallEventsOperation`, being the request's path the only change.

Sources/Ads/AdTracker.swift
```diff
@@ -0,0 +1,150 @@
...
+     adUnitId: "ca-app-pub-123",
+     adInstanceId: "instance-456",
+     revenueMicros: 1500000,  // $1.50
+     currency: "USD",
+     precision: .exact
+ ))
+ ```
+ */
+@available(iOS 15.0, tvOS 15.0, macOS 12.0, watchOS 8.0, *)
+@objc(RCAdTracker)
```
ajpallares, line 58: I see we have the `@objc` marker here, but it's not in the `track` methods or in the `AdEventData` types: `AdDisplayed`, `AdOpened` or `AdRevenue`. These are structs and cannot be ported to Objective-C as is.

So, do we want to have these APIs be compatible with Objective-C? 
* If not, we can (and should) remove the `@objc` marker
* If yes, then we might need to rethink these objects to make them `NSObject` subclasses instead of `struct`s (it pains me to propose this 😿)

Sources/Events/StoredEvent.swift
```diff
@@ -40,6 +40,7 @@ enum Feature: String, Codable {
 
     case paywalls
     case customerCenter
+    case ads
```
ajpallares, line 44: Not important, but this one could also go under `#if ENABLE_AD_EVENTS_TRACKING`, to not to have to deal with it outside of `#if ENABLE_AD_EVENTS_TRACKING`. 
Feel free to ignore, as this is an internal enum

Sources/Networking/InternalAPI.swift
```diff
@@ -113,6 +113,21 @@ class InternalAPI {
         self.backendConfig.addDiagnosticsOperation(operation, delay: .long)
     }
 
+    @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+    func postAdEvents(events: [StoredEvent], completion: @escaping ResponseHandler) {
```
ajpallares: The fact that this method already assumes the `StoredEvents` array contain `AdEvents` suggests me that we should probably refactor this to make it type safe (so that we can't possibly pass a `StoreEvent` that contains a `PaywallsEvent`, for example). Again, something to do on a separate PR

Sources/Ads/Events/AdEvent.swift
ajpallares: Sorry I'm late to this.

What @polmiro did making `MediatorName` a `struct` with `rawValue` is exactly the way to proceed in these scenarios in Swift.

Btw, if I understand correctly from the conversation, we want this for both `mediatorName` and `networkName`? I ask because I only see it implemented for `MediatorName`.

polmiro: Great, thanks for reviewing.

I don't think it makes sense for network name. That vale will be relayed from the mediator and it would be strange to have to hard-code the value. Plus the list of networks is much larger.

Sources/Ads/Events/Networking/EventsRequest+Ad.swift
polmiro: It is an inconsistent but I was doing it on purpose since it doubts for me in the backend not having explicit units, this is from another comment:
> I renamed the internal model for consistency but kept the serialized value sent to the backend as timestamp_ms. It is more explicit given UNIX timestamp are sometimes modeled as seconds and not miliseconds.

I'm not sure if I should favor consistency or this is enough justification. I'm a bit in the middle on this one.

Sources/Purchasing/Purchases/Purchases.swift
polmiro: Agreed, this one will be tackled in a different PR per our discussion today. I will mark resolved.

Sources/Ads/Events/Networking/EventsRequest+Ad.swift
polmiro: Thanks! This is accidental and I can see how it can be error-prone/confusing. I will update.

Sources/Ads/Events/Networking/EventsRequest+Ad.swift
ajpallares: I'm fine with this changing and being inconsistent if it's an improvement. I'm Ok with `timestampMs` tbh

Sources/Events/StoredEvent.swift
polmiro: Fixed.

Sources/Ads/Events/Networking/EventsRequest+Ad.swift
polmiro: Fixed!

Sources/Ads/Events/Networking/EventsRequest+Ad.swift
polmiro: Fixed!

Sources/Ads/AdTracker.swift
polmiro: Per our convo the other day, I'll tackle this in a separate PR dedicated to this goal. Marking resolved.

Sources/Paywalls/Events/PaywallEventsManager.swift
polmiro: I have removed the implementation tweaks of PaywallEventsManager and will tackle the delivery of events in a separate PR per our discussion the other day. Marking this resolved.

polmiro:
@JayShortway I have marked the APIs with `@spi_(Experimental)`

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,265 @@
...
+
+}
+
+extension AdRevenue {
+
+    /// Type representing the level of accuracy for reported revenue values.
+    @_spi(Experimental) public struct Precision: Equatable, Hashable, Codable, Sendable {
+
+        /// The raw string value of the precision type
+        public let rawValue: String
```
ajpallares: Huge nit and not really that important (as this whole API is marked as Experimental):
Since the `init` is internal, perhaps we could make the `rawValue` internal as well (I can't see why it's needed to be read by devs). 
We can always open it up later if we need

Sources/Ads/Events/Networking/PostAdEventsOperation.swift
polmiro: Addressed by creating a `PostEventsOperation` that accepts `path` initialization parameter.

Sources/Ads/AdTracker.swift
```diff
@@ -0,0 +1,141 @@
...
+         networkName: "AdMob",
+         mediatorName: .appLovin,
+         placement: "home_screen",
+         adUnitId: "ca-app-pub-123",
+         adInstanceId: "instance-456"
+     ))
+     ```
+     */
+    @_spi(Experimental) public func trackAdOpened(_ data: AdOpened) async {
+        let event = AdEvent.opened(.init(id: UUID(), date: Date()), data)
```
ajpallares, line 111: In case it was an oversight: this implementation seems to be incomplete

Sources/Ads/AdTracker.swift
```diff
@@ -0,0 +1,141 @@
...
+         adUnitId: "ca-app-pub-123",
+         adInstanceId: "instance-456",
+         revenueMicros: 1500000,  // $1.50
+         currency: "USD",
+         precision: .exact
+     ))
+     ```
+     */
+    @_spi(Experimental) public func trackAdRevenue(_ data: AdRevenue) async {
+        let event = AdEvent.revenue(.init(id: UUID(), date: Date()), data)
```
ajpallares, line 138: Same here (just in case): this implementation seems to be incomplete

RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -1122,6 +1120,17 @@
 		88B1BB132C81479F001B7EE5 /* PaywallComponentTypeTransformers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 88B1BAE72C813A3C001B7EE5 /* PaywallComponentTypeTransformers.swift */; };
 		88DE93E12C211CBC0086B6D8 /* RevenueCat.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 2DC5621624EC63420031F69B /* RevenueCat.framework */; };
 		88E679472C7503C1007E69D5 /* PaywallStackComponent.swift in Sources */ = {isa = PBXBuildFile; fileRef = 88E679462C7503C1007E69D5 /* PaywallStackComponent.swift */; };
+		901DE80E2EA0E097007EAA86 /* AdTracker.swift in Sources */ = {isa = PBXBuildFile; fileRef = 901DE80D2EA0E096007EAA86 /* AdTracker.swift */; };
+		9025C5202EA63EB700845BCE /* PostEventsOperation.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9025C51F2EA63EB700845BCE /* PostEventsOperation.swift */; };
```
ajpallares: Note that the PostEventsOperation.swift file seems to be misplaced in the repo and Xcode workspace. 

Image

It should probably be located next to the other operation files

Sources/Networking/InternalAPI.swift
polmiro: I'm removing this method from this PR given their spec is not finally. We may end up not using the same storage for the `AdEventsManager` implementation after all. Marking this resolved.  Thank you

Sources/Events/Networking/PostEventsOperation.swift
```diff
@@ -0,0 +1,69 @@
...
+        } else if let adPath = self.path as? HTTPRequest.AdPath {
+            httpRequest = HTTPRequest(method: .post(self.request), path: adPath)
+        } else {
+            assertionFailure("Unsupported path type: \(type(of: self.path))")
+            defer {
+                completion()
+            }
+            self.responseHandler?(nil)
+            return
+        }
```
ajpallares: This should not be necessary. Instead of the runtime check, we should have compile-time safety here. I can do it separately as it would require some small changes to `HTTPRequest`

PostEventsOperation.swift
ajpallares, line 1: This file seems to be duplicated. This one should be removed from the Xcode workspace and the one in `Sources/Events/Networking/PostEventsOperation.swift` should be added to the Xcode workspace

ajpallares (commented):
Looking even nicer! Thank you! I added some new comments

Sources/Ads/AdTracker.swift
polmiro: Yes, that's intended, we are missing the event manager which will go in a different PR.

RevenueCat.xcodeproj/project.pbxproj
polmiro: Ugh, my bad. I'm still struggling adding the files manually. I did not realize it would not co-locate it in the same path the file I added was in. Thanks for noticing.

Sources/Events/Networking/PostEventsOperation.swift
polmiro: Good point. Honestly I do not see a lot of value in the runtime check. I am going to remove it altogether.

Do you think its worth to refactor for compile-time safety? If so, how would you do it? With a protocol? With generic types?

ajpallares: Sorry I missed this. I think it's a bit more intricate to do. It can be done in a separate PR, no rush 👍

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,265 @@
...
+//
+//      https://opensource.org/licenses/MIT
+//
+//  AdEvent.swift
+//
+//  Created by RevenueCat on 1/8/25.
+
+import Foundation
+
+// MARK: - Public Types
```
ajpallares: Nit: probably this comment should go after `AdEventData`?

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,265 @@
...
+
+    /// AppLovin MAX mediation network
+    public static let appLovin = MediatorName(rawValue: "AppLovin")
+
+}
+
+/// Data for ad displayed events.
+@_spi(Experimental) public struct AdDisplayed: AdEventData {
+
+    // swiftlint:disable missing_docs
```
ajpallares: Just to be sure: shouldn't we have documentation on these public interfaces?

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,265 @@
...
+        self.adInstanceId = adInstanceId
+    }
+    // swiftlint:enable missing_docs
+
+}
+
+/// Data for ad opened/clicked events.
+@_spi(Experimental) public struct AdOpened: AdEventData {
+
+    // swiftlint:disable missing_docs
```
ajpallares: Same here about documentation?

Sources/Ads/Events/AdEvent.swift
```diff
@@ -0,0 +1,265 @@
...
+        self.adInstanceId = adInstanceId
+    }
+    // swiftlint:enable missing_docs
+
+}
+
+/// Data for ad revenue events.
+@_spi(Experimental) public struct AdRevenue: AdEventData {
+
+    // swiftlint:disable missing_docs
```
ajpallares: Same?

Tests/UnitTests/Ads/Events/AdEventsRequestTests.swift
```diff
@@ -0,0 +1,138 @@
...
+
+        try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
+    }
+
+    func testDisplayedEvent() throws {
+        let event = AdEvent.displayed(Self.eventCreationData, Self.eventData)
+        let storedEvent = try Self.createStoredEvent(from: event)
+        let requestEvent: EventsRequest.AdEventRequest = try XCTUnwrap(.init(storedEvent: storedEvent))
+
+        assertSnapshot(matching: requestEvent, as: .formattedJson)
```
ajpallares, line 36: Just to make sure: the purpose of these tests is to make sure that we always send the same JSON to the server?

Tests/UnitTests/Ads/Events/PurchasesAdEventsTests.swift
```diff
@@ -0,0 +1,68 @@
...
+
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+
+        try AvailabilityChecks.iOS15APIAvailableOrSkipTest()
+
+        self.setupPurchases()
+    }
+
+    func testTrackAdDisplayedDoesNotCrash() async throws {
```
ajpallares: What's the purpose of these tests?

ajpallares (approved):
Looks good! Again, great job on this!

I have just some small questions

Tests/UnitTests/Ads/Events/AdEventsRequestTests.swift
polmiro: Yes that's right. This way we can clearly seen in the snapshots what is being sent to the server so that we can be aware of any changes in the code that may modify it.

Tests/UnitTests/Ads/Events/PurchasesAdEventsTests.swift
polmiro: These public API methods are not called elsewhere so I thought it made sense to add a basic smoke test that verifies the public-facing API and the plumbing are hooked as expected.

But they are definitely too barebones. I just spotted a big bug thanks to you. The AdTracker initializes the events but does not store them via the events manager! Probably something I would have noticed when testing this end to end but still points out missing some testing coverage.

I expanded the tests to assert on stored events and fixed the bug. Thank you!

JayShortway:
Great job on this! 🙌

PR #5641, opened 2025-10-08
sort win-back offers by StoreKit order

antiraum:
Win-back offers should be returned in the best offer first order from StoreKit, so clients can present the best offer to the user.

Issue #5644


antiraum:
obsolete because addressed by https://github.com/RevenueCat/purchases-ios/pull/5646

PR #5648, opened 2025-10-09
Migrate APITests to Tuist workspace

facumenzella:
### Motivation
We want to finally clean-up the main workspace, in favor of using the Tuist generated one for internal development. In order to do that, we need to slowly move nested projects to Tuist, deleting them from the main one

### Description
- Delete APITests reference from RevenueCat.xcworkspace
- Add scheme to APITests project to build all the targets
- Edit CircleCI config file to run with Tuist.


facumenzella:
@RCGitBot please test

facumenzella:
@RCGitBot please test

facumenzella:
@RCGitBot please test

RevenueCat.xcodeproj/project.pbxproj
```diff
@@ -1294,55 +1294,6 @@
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
-		2D4C60F52C5AD1C300A29FD2 /* PBXContainerItemProxy */ = {
```
rickvdl, line 1297: Assuming these tests are now only available through Tuist, do we need to update any (internal) documentation on how to run these?

ajpallares:
@RCGitBot please test

fastlane/Fastfile
```diff
@@ -788,8 +788,12 @@ platform :ios do
   desc "run API Tests"
   lane :run_api_tests do
     # note: it's called "run" tests but these tests only need to build to run
+    workspace_path = 'RevenueCat-Tuist.xcworkspace'
```
ajpallares, line 791: Very small thing (that we can even do in a future PR), but we could perhaps create a constant outside o the lane scope for this that gets invoked by jobs that use tuist. E.g.
```
TUIST_WORKSPACE_PATH = 'RevenueCat-Tuist.xcworkspace'
```

RevenueCat.xcodeproj/project.pbxproj
facumenzella: I'll update the notion page, and I'll be adding a new section to DEVELOPMENT.md

fastlane/Fastfile
facumenzella: Yeah, I was thinking to move it to the `fastlane/.env` which we don't use yet, but's perfect for this

ajpallares: Ohh that's even better! 👌

PR #5652, opened 2025-10-10
Fix Crashes: Move large object cacheing off of user defaults to file storage

### LINKED ISSUE #5793: 🐛 Crash: Attempting to store >= '...' bytes of data in CFPreferences/NSUserDefaults on this platform is invalid
### Describe the bug

Hello, RevenueCat SDK has been crashing during configuration on watchOS since version 26.0.1. The crash looks like to occur when user updates their OS: we've received reports from 26.0.1 and 26.1, only for old installations.

Reinstalling the app seems to fix the crash. Getting the crash full stack trace is a bit tricky given the the conditions required to reproduce it.

### Platform

watchOS

### SDK version

5.47.0

### SDK integration method

Swift Package Manager

### StoreKit version

StoreKit 2 (default on versions >=5.0.0)

### OS version

watchOS 26.1

### Xcode version

26.1

### Device and/or simulator

Device

### Environment

Production

### How widespread is the issue

100% of those updating their OS

### Debug logs

```shell
Attempting to store >= 1048576 bytes of data in CFPreferences/NSUserDefaults on this platform is invalid. This is a bug in Sunlitt Watch or a library it uses.
<decode: bad range for [%@] got [offs:516 len:961 within:0]>
```

### Steps to reproduce

This is kinda tricky to replicate. A fresh installation won't trigger the issue, while old installations will after an OS update.

### Other information

```markdown

```

### Additional context

```markdown

```

JZDesign:
### Motivation

We have crashes in tvOS and watchOS due to overloading the user defaults. 

### Description

Converts the storage for all Codable objects from UserDefaults to the cache directory in the file system

resolves #5793

### Checklist

### DIFF
--- a/Tests/UnitTests/Caching/DeviceCacheTests.swift
+++ b/Tests/UnitTests/Caching/DeviceCacheTests.swift
@@ -424,7 +424,7 @@ class DeviceCacheTests: TestCase {
         // Verify that removeObject was called on the file cache with the correct key
         expect(self.mockFileCache.removeInvocations.count) == 1
         let expectedURL = self.mockFileCache.cacheDirectory?
-            .appendingPathComponent("docs–com.revenuecat.data")
+            .appendingPathComponent("docs–RevenueCat")
             .appendingPathComponent("com.revenuecat.userdefaults.offerings.\(appUserID)")
         expect(self.mockFileCache.removeInvocations.first) == expectedURL
     }


Sources/Caching/DeviceCache.swift
```diff
@@ -31,15 +33,19 @@ class DeviceCache {
...
          userDefaults: UserDefaults,
+         fileManager: LargeItemCacheType = FileManager.default,
          offeringsCachedObject: InMemoryCachedObject<Offerings> = .init()) {
         self.offeringsCachedObject = offeringsCachedObject
         self.systemInfo = systemInfo
         self.userDefaults = .init(userDefaults: userDefaults)
         self._cachedAppUserID = .init(userDefaults.string(forKey: CacheKeys.appUserDefaults))
         self._cachedLegacyAppUserID = .init(userDefaults.string(forKey: CacheKeys.legacyGeneratedAppUserDefaults))
+        self.cacheURL = fileManager.createCacheDirectoryIfNeeded(basePath: Self.defaultBasePath)
+        self.largeItemCache = .init(cache: fileManager, basePath: "RevenueCat")
```
facumenzella: this should be `defaultBasePath`

JZDesign: Thank you.

Sources/Misc/Concurrency/SynchronizedLargeItemCache.swift
```diff
@@ -0,0 +1,103 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  SynchronizedLargeItemCache.swift
+//
+//  Created by Jacob Zivan Rakidzich on 10/9/25.
+
+import Foundation
+
+/// A thread-safe wrapper around `LargeItemCacheType` for synchronized file-based caching operations.
+internal final class SynchronizedLargeItemCache {
```
ajpallares, line 17: Why not make this an `actor`? I understand that it would mean a deeper change, since the `await` calls would cascade up the call chain, but I wonder if we could take the opportunity to use more modern APIs. I mostly worry we could be adding more technical debt that we might need to face in the future (Swift 6)

Sources/Caching/DeviceCache.swift
```diff
@@ -161,8 +179,14 @@ class DeviceCache {
     // MARK: - Offerings
 
     func cachedOfferingsResponseData(appUserID: String) -> Data? {
-        return self.userDefaults.read {
-            $0.data(forKey: CacheKey.offerings(appUserID))
+        guard let cacheURL = self.cacheURL else {
```
ajpallares: I also wonder, for people updating from a prev. SDK version, should we perform some kind of cleanup/migration? 

At a minimum, if we're going to disregard the previous UserDefaults-based cache, should we perhaps cleanup and remove the values for these keys from UserDefaults?

ajpallares (commented):
Indeed, this was long overdue. Thank you for taking care of this!
I have two initial questions

Sources/Misc/Concurrency/SynchronizedLargeItemCache.swift
JZDesign: My concern was primarily cascading await's everywhere and how that may accidentally impact things elsewhere. 

I think it best to isolate the changes so if something goes wrong with either change set, it will be easier to track down the bug. If we did both at the same time, we may think we have a concurrency bug when we have an issue with the cache or visa versa.

Sources/Caching/DeviceCache.swift
JZDesign: Yea, I was thinking about that too… I could throw that in there.

ajpallares: That'd be nice. That said, I'm not sure on what the best approach would be. Probably something worth raising internally to let the rest of the team know. I honestly don't know if just being "aggressive" (e.g. removing the existing cache) is good enough

Sources/Misc/Concurrency/SynchronizedLargeItemCache.swift
ajpallares: > My concern was primarily cascading await's everywhere and how that may accidentally impact things elsewhere

Yeah, this would happen for sure. It is no small change.

> I think it best to isolate the changes so if something goes wrong with either change set, it will be easier to track down the bug. If we did both at the same time, we may think we have a concurrency bug when we have an issue with the cache or visa versa.

This I understand, but I don't think I agree. I don't think migrating `SynchronizedLargeItemCache` would be a cumulative change. It would be rather a rewrite of the class/actor, so it could be argued that if we're going to end up doing it eventually, we probably should do it that way from the beginning, since 2 rewrites is more likely to introduce bugs than only one.

With these things I'm always afraid that we're making ir harder for future us by postponing something we could tackle in the present.

Not a blocker though, as I understand that it would not be a trivial change.

JZDesign: >  if we're going to end up doing it eventually

Let's get others in the conversation and we can determine how soon this is important. I would want to use swift 6.2's approachable concurrency  if we migrate to an actor, and I am not sure if we can do that and have package.swift files for lesser versions of Swift

JZDesign: @facumenzella, @fire-at-will, @MarkVillacampa, & @joshdholtz 

What are your thoughts?

ajpallares: > I would want to use swift 6.2's approachable concurrency if we migrate to an actor, and I am not sure if we can do that and have package.swift files for lesser versions of Swift

I've been thinking a bit lately about this. It seems we're going to need to stuck with Swift 5 for a while. And I agree that it would be ideal to jump directly to Swift 6.2. However, as you say, we can't realistically have implementations work in two different versions of Swift (because that may lead to many potential bugs and behavioral differences), which means that we'll be able to jump to Swift 6 only when we're able to drop Swift 5 support, and I don't see that happening any time soon because we still need to run tests in older versions of Xcode...

JZDesign: Right, which is part of the reason I'm using locks and keeping things out of Swift Concurrency. 

And, in the spirit of Always Be Shipping, I think we should defer moving to Swift Concurrency. Getting this fix out to prevent crashes is good, especially considering what's coming with funnels. The Swift Concurrency change is a bigger risk than moving from UserDefaults to the File System because the async code changes won't be isolated to this one area, it will have to spread. And where it spreads to has to then handle the cases appropriately. 

But the file system change is essentially transparent to consumers of the Device Cache

JZDesign: It is my opinion that refactoring this to Swift Concurrency should be ticketed and worked on independently.

joshdholtz: Oh I am so not qualified for this conversation. I would retreat and say let's just write this in Objective-C

joshdholtz: Okay, I thought about this some more and I think... I agree that not making this an actor makes sense here

But like, I agree because I've had issues with using things at this type of level that are actors in areas that weren't concurrency-ed yet and it felt very backwards. 

Until we have things at a higher level concurrency-ed, I think implementing like this makes sense 🤷‍♂️ 

But also please tell me if I didn't understand this correctly 😅

Sources/Caching/DeviceCache.swift
```diff
@@ -31,15 +33,19 @@ class DeviceCache {
...
     init(systemInfo: SystemInfo,
          userDefaults: UserDefaults,
+         fileManager: LargeItemCacheType = FileManager.default,
          offeringsCachedObject: InMemoryCachedObject<Offerings> = .init()) {
         self.offeringsCachedObject = offeringsCachedObject
         self.systemInfo = systemInfo
         self.userDefaults = .init(userDefaults: userDefaults)
         self._cachedAppUserID = .init(userDefaults.string(forKey: CacheKeys.appUserDefaults))
         self._cachedLegacyAppUserID = .init(userDefaults.string(forKey: CacheKeys.legacyGeneratedAppUserDefaults))
+        self.cacheURL = fileManager.createCacheDirectoryIfNeeded(basePath: Self.defaultBasePath)
```
joshdholtz: So changing this to "cache" is changing the lifespan of this data, right? `UserDeafults` is a forever thing where `createCacheDirectoryIfNeeded` will create this in the cache directory which can be clared by the system.

To make this equivalent, does this need to be in the documents directory? 🤔 

I mean, I know why cache is good because large things but also the offerings cache _is_ kind of important (especially when infra is offline) so it might be good to put it in documents directory? 🤷‍♂️

JZDesign: Yea, that's a fair point. I can update this to use a docs directory so it's truly persistent.

Sources/Misc/Concurrency/SynchronizedLargeItemCache.swift
JZDesign: These are my 2 biggest reasons for being against actors here:

1. Contagion - Needing to update code in various places and guarantee that it works as expected
1. Reentrancy - Actors are not mutexes They are reentrant. So they are prone to edge cases. I would still need to ensure mutual exclusive executions and that is solved by NSLock fairly well.

Those are my reasons, with their powers combined, we have Captain GCD, he's our hero.

joshdholtz: I'm good with the GCD approach 😊 :shipit:

Sources/Caching/DeviceCache.swift
joshdholtz: I think that would make sense! 

cc: @ajpallares because just good to get a CoreSDK review on that too 😊

Sources/Caching/DeviceCache.swift
MarkVillacampa: Just to reiterate on this, are we fine with clearing the offerings and product entitlement mapping cache (as those are the ones I think this PR changes the storage for) for all apps updating to this version of the SDK? do we not want to provide a one-time migration from user defaults to filesystem cache?

Sources/Caching/DeviceCache.swift
ajpallares: Yeah, that's a great point. Ideally the cache should have the same lifespan as before. We just want to change where/how we store it, but not its lifespan I'd say 👍 
Great catch @joshdholtz!

Sources/Caching/DeviceCache.swift
```diff
@@ -55,17 +61,31 @@ class DeviceCache {
...
+               let decoded: Value = try? JSONDecoder.default.decode(jsonData: data, logErrors: true) {
+                value = decoded
+            }
+
             updater(&value)
-            $0.set(codable: value, forKey: key)
+
+            if let data = try? JSONEncoder.default.encode(value: value, logErrors: true) {
+                try? cache.saveData(data, to: fileURL)
+            }
```
ajpallares: This seems like it's code repeated with `SynchronizedLargeItemCache`. Why don't we just use the existing `func set<T: Encodable>(codable value: T, forKey key: DeviceCacheKeyType) -> Bool` in `SynchronizedLargeItemCache`.

Something like this:
```suggestion
        var value = self.largeItemCache.value(forKey: key) ?? defaultValue
        updater(&value)
        self.largeItemCache.set(codable: value, forKey: key)
```

I _think_ it's the same and much simpler. In addition, it allows making the `write` method in `SynchronizedLargeItemCache` just `internal`. 
That said, we'd lose the synchronized write operation that we get with the `write` method, which can be easily solved in other ways.

ajpallares: In addition, I think this 
```swift
func update<Key: DeviceCacheKeyType, Value: Codable>(
        key: Key,
        default defaultValue: Value,
        updater: @Sendable (inout Value) -> Void
    )
```

is not used anywhere. So we can probably just remove it (and then make `write` in `SynchronizedLargeItemCache` `internal`)

Sources/Caching/DeviceCache.swift
```diff
@@ -161,8 +184,20 @@ class DeviceCache {
...
+        guard let cacheURL = self.cacheURL else {
+            return nil
+        }
+
+        let fileURL = cacheURL.appendingPathComponent(CacheKey.offerings(appUserID).rawValue)
+
+        return self.largeItemCache.read { cache, _ in
+            try? cache.loadFile(at: fileURL)
         }
     }
```
ajpallares: I'm thinking we can make this method return a `Offerings.Contents` directly. That way, we can use the `func value<T: Decodable>(forKey key: DeviceCacheKeyType) -> T?` of `SynchronizedLargeItemCache` directly, so we can make the `read` of `SynchronizedLargeItemCache` also internal. It's a nicer API for `DeviceCache`, as it is making sure that the data is always being decoded as `Offerings.Contents`, as `DeviceCache` is responsible already of encoding the `Offerings.Contents` for that same key.

ajpallares (commented):
Some more comments here! 🙏 
Mostly about making the `read` and `write` methods of `SynchronizedLargeItemCache` `internal` implementation details, cleaning up a bit the `DeviceCache` API and removing an unused method.

Sources/Caching/DeviceCache.swift
```diff
@@ -160,10 +160,14 @@ class DeviceCache {
...
-        return self.userDefaults.read {
-            $0.data(forKey: CacheKey.offerings(appUserID))
+    func cachedOfferingsContents(appUserID: String) -> Offerings.Contents? {
+        // Large data used to be stored in the user defaults and resulted in crashes, we need to ensure that
+        // we are cleaning out that data
+        userDefaults.write { defaults in
+            defaults.removeObject(forKey: CacheKey.offerings(appUserID))
         }
+
+        return largeItemCache.value(forKey: CacheKey.offerings(appUserID))
```
ajpallares: Wouldn't this be the same as this?

```suggestion
        return self.value(for: CacheKey.offerings(appUserID))
```
To avoid repeating the exact same code

Sources/Caching/DeviceCache.swift
```diff
@@ -353,13 +353,23 @@ class DeviceCache {
...
     }
 
     var cachedProductEntitlementMapping: ProductEntitlementMapping? {
-        return self.userDefaults.read(Self.productEntitlementMapping)
+        // Large data used to be stored in the user defaults and resulted in crashes, we need to ensure that
+        // we are cleaning out that data
+        userDefaults.write { defaults in
+            defaults.removeObject(forKey: CacheKeys.productEntitlementMapping)
+        }
+        return self.largeItemCache.value(forKey: CacheKeys.productEntitlementMapping)
```
ajpallares: Same thing here
```suggestion
        return self.value(for: CacheKeys.productEntitlementMapping)
```

Sources/Misc/Concurrency/SynchronizedLargeItemCache.swift
```diff
@@ -0,0 +1,103 @@
...
+    private let documentURL: URL?
+
+    init(cache: LargeItemCacheType, basePath: String) {
+        self.cache = cache
+        self.lock = Lock(.nonRecursive)
+        self.documentURL = cache.createDocumentDirectoryIfNeeded(basePath: basePath)
+    }
+
+    /// Performs a synchronized read operation
+    internal func read<T>(_ action: (LargeItemCacheType, URL?) throws -> T) rethrows -> T {
```
ajpallares: Sorry, I said `internal` in my previous comment when I meant `private` (I meant "internal for the `SynchronizedLargeItemCache` 😅 )
```suggestion
    private func read<T>(_ action: (LargeItemCacheType, URL?) throws -> T) rethrows -> T {
```

Sources/Misc/Concurrency/SynchronizedLargeItemCache.swift
```diff
@@ -0,0 +1,103 @@
...
+        return try self.lock.perform {
+            return try action(self.cache, self.documentURL)
+        }
+    }
+
+    /// Performs a synchronized write operation
+    internal func write(_ action: (LargeItemCacheType, URL?) throws -> Void) rethrows {
```
ajpallares: Same here
```suggestion
    private func write(_ action: (LargeItemCacheType, URL?) throws -> Void) rethrows {
```

Sources/Caching/DeviceCache.swift
```diff
@@ -50,22 +56,13 @@ class DeviceCache {
...
-
     func value<Key: DeviceCacheKeyType, Value: Codable>(for key: Key) -> Value? {
-        self.userDefaults.read {
-            $0.value(forKey: key)
+        // Large data used to be stored in the user defaults and resulted in crashes, we need to ensure that
+        // we are cleaning out that data
+        userDefaults.write { defaults in
+            defaults.removeObject(forKey: key)
         }
+        return self.largeItemCache.value(forKey: key)
```
ajpallares, line 65: This means that all cacheable `Codable` objects are always cached in the `largeItemCache`. That is intentional, right? It's an implementation detail of the `DeviceCache`, so it should not affect callers in any way, but I'm afraid this could open up the door for inconsistencies. 
For example, in the `clearCaches` method below, we have

```swift
            // Clear offerings cache.
            self.offeringsCachedObject.clearCache()
            userDefaults.removeObject(forKey: CacheKey.offerings(oldAppUserID))

            // Clear virtual currencies cache
            userDefaults.removeObject(forKey: CacheKey.virtualCurrencies(oldAppUserID))
```

With the new changes, Offerings and VirtualCurrencies, as `Codable` objects, will be cached in the `largeItemCache`, but there we're cleaning them up from `userDefaults` only. Which means that the cache for them is not cleared from the `largeItemCache`.

Am I missing something? If I'm right, then we should ideally have a more robust implementation on this to make sure that the cache used for read, write and clear is always consistent. In addition, we should probably add specific unit tests to grant that we're always calling the right cache in all cases (read, write and clear)

JZDesign: >  but there we're cleaning them up from userDefaults only

The final line of clearCaches is cleaning up the `largeItemCache`

ajpallares: > The final line of clearCaches is cleaning up the largeItemCache

Oh, yes. You are totally right. I missed that line 🤦‍♂️ 

In addition, I assumed that all `Codable` objects were now being stored in the `largeItemCache`. That includes VirtualCurrencies and CustomerInfo, which could potentially be "large" as well (we could address that in a separate PR, because in general, I agree that Offerings is usually the largest of all).

Sorry about the confusions 🙏

Sources/Caching/DeviceCache.swift
```diff
@@ -50,22 +56,13 @@ class DeviceCache {
...
-        updater: @Sendable (inout Value) -> Void
-    ) {
-        self.userDefaults.write {
-            var value: Value = $0.value(forKey: key) ?? defaultValue
-            updater(&value)
-            $0.set(codable: value, forKey: key)
-        }
-    }
-
     func value<Key: DeviceCacheKeyType, Value: Codable>(for key: Key) -> Value? {
```
ajpallares: This one can now be `private` I think. This way we remove the temptation of calling it directly from outside the `DeviceCache`

Sources/Caching/DeviceCache.swift
ajpallares, line 91: To prevent future us from having the same confusion I had in my previous comment, I'd add an explanatory comment here saying that although Offerings cache now lives in the `largeItemCache`, we remove it from `userDefaults` to clear any pre-existing data before the migration to `largeItemCache` (something in that line, ideally better put 😅)

Sources/Misc/Concurrency/SynchronizedLargeItemCache.swift
```diff
@@ -0,0 +1,101 @@
...
+        }
+    }
+
+    /// Remove a cached item
+    func removeObject(forKey key: DeviceCacheKeyType) {
+        guard let documentURL = self.documentURL else {
+            return
+        }
+
+        let fileURL = documentURL.appendingPathComponent(key.rawValue)
```
ajpallares: Small nit: I feel like we're repeating 
```swift
let fileURL = documentURL.appendingPathComponent(key.rawValue)
```

too much. Maybe we can centralize in a method to make sure we're always using the same strategy to obtain the `fileURL` for a specific `DeviceCacheKeyType`. Something like

```swift
func getFileUrl(for key: DeviceCacheKeyType) -> URL? {
  return self.documentURL?.appendingPathComponent(key.rawValue)
}
```

We can decide when to log the `Logger.error("Cache URL is not available")` depending on the caller's site

Tests/UnitTests/Caching/DeviceCacheTests.swift
ajpallares, line 1: I think we're missing some tests:
* For `clearCaches`, to sure that the Offerings cache is removed from the file cache with the right key
* For `cachedOfferingsContents`:
  * To make sure we're removing the old user defaults cache with the right key
  * To make sure that we're returning the same OfferingsContents as the ones saved (similar to the existing `testCacheProductEntitlementMapping()`)
  * Even more importantly: to make sure that we're never saving `Offerings` into User Defaults anymore (I know that doesn't happen now, but I want to cover us from making that mistake in the future)

ajpallares (commented):
Some (I think final) comments. Sorry about the confusion on my previous review, my mistake.
Also, sorry I'm asking you to add more tests. But this feels to me like a crucial part of the SDK and we want to close the doors of eventually messing this up 🙏

ajpallares (approved):
Amazing! Thank you for all the fixes and all the iterations! 🙌 
Great job on this!

PR #5653, opened 2025-10-10
Add DEVELOPMENT file

facumenzella:
### Motivation
Have a document in the repository that describes the basics for the Tuist workspace

### Description
- Introduce file with examples


Contributing/DEVELOPMENT.md
```diff
@@ -0,0 +1,78 @@
+# Working with Tuist
```
rickvdl, line 1: Should we mention the fact that the xcworkspace also still works and that we're migrating to Tuist? Or do we want to avoid mentioning the xcworkspace explicitly?

Also if the doc starts with Tuist, maybe we should include installation instructions (or just a link to Tuist itself) for those who aren't familiar with it.

facumenzella: I like that. On it

PR #5655, opened 2025-10-10
Naive approach to handle the onPurchaseCompleted event for an offer code

JZDesign:
### Checklist

### Motivation

We have 2 issues with how the purchase handler callbacks work that can be used in swiftUI or ReactNative. 

1. The offer code flow is executed outside of the bounds of our paywall, and therefore the purchase handler does not directly handle the transaction and no event is fired back through our callbacks
2. If state mis-management causes SwiftUI to re-render the views, a disconnection can occur and the PurchaseHandler that is processing a transaction may not be the same instance that would trigger the callback through the view hierarchy.

If either of those things are true, applications that are depending on the information that comes through our `onPurchaseCompleted` callback get let down and they miss the event entirely.

### Description

Using a naive approach (the broadcast receiver pattern via the NotificationCenter), we enforce that all instances of the purchase handler will get all purchase events. They will be de-duplicated to ensure that we do not invoke the callback twice.

### DIFF
--- a/RevenueCatUI/Purchasing/PurchaseHandler.swift
+++ b/RevenueCatUI/Purchasing/PurchaseHandler.swift
@@ -134,7 +134,9 @@ final class PurchaseHandler: ObservableObject {
         purchaseResultPublisher
             .removeDuplicates(by: PurchaseResultComparator.compare)
             .receive(on: RunLoop.main)
-            .sink(receiveValue: setResult)
+            .sink { [weak self] result in
+                self?.setResult(result)
+            }
             .store(in: &cancellables)
     }
 
@@ -176,6 +178,10 @@ final class PurchaseHandler: ObservableObject {
             return
         }
         self.purchaseResult = result
+    }
+
+    deinit {
+        cancellables.removeAll()
     }
 
 }


RevenueCatUI/Templates/V2/Components/Button/ButtonComponentView.swift
```diff
@@ -74,6 +74,16 @@ struct ButtonComponentView: View {
...
+            .onReceive(NotificationCenter.default.publisher(for: .purchaseCompleted), perform: { object in
+                switch viewModel.action {
+                case .navigateTo(destination: .offerCodeRedemptionSheet):
+                    if let data = object.object as? PurchaseResultData {
+                        purchaseHandler.setResult(data)
+                    }
+                default:
+                    break
+                }
+            })
```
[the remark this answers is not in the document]
JZDesign: for example the offer code button is there, but unused, the purchase is made manually. This should not send that event.

Tests/RevenueCatUITests/Purchasing/MockStoreTransaction.swift
JZDesign, line 1: Perhaps this is something that Tuist will help with, but I copied a file and pasted it so I could get these tests to run

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -101,23 +104,38 @@ final class PurchaseHandler: ObservableObject {
...
+        purchaseResultPublisher: AnyPublisher<PurchaseResultData, Never> = NotificationCenter
+            .default
+            .purchaseCompletedPublisher()
     ) {
         self.isConfigured = isConfigured
         self.purchases = purchases
         self.performPurchase = performPurchase
         self.performRestore = performRestore
+
+        purchaseResultPublisher
```
rickvdl, line 134: Should we also add some sort of 'request identifier' or similar such that we can verify that this is the same purchase that was completed? Thinking about the theoretical case here where we still have one pending and the notification is posted later when the screen is still open / opened again and suddenly the screen closes because the result came in? Not 100% sure if that could happen

Sources/Purchasing/Purchases/TransactionNotifications.swift
```diff
@@ -0,0 +1,24 @@
...
+//
+//  Created by Jacob Zivan Rakidzich on 10/10/25.
+//
+
+import Combine
+import Foundation
+
+extension NSNotification.Name {
+    /// A notification that states a purchase has completed
+    static let purchaseCompleted = Notification.Name("RevenueCat.PurchaseCompleted")
```
rickvdl, line 13: Should we add a doc here (or spi(Internal) or both) to inform developers that this should not be relied on?

JZDesign: The spi(Internal) annotation is not needed because it's internally scoped already and not accessible outside of the target scope. But I could add some docs

rickvdl: Ah 🤦 my bad, please ignore..

RevenueCatUI/Purchasing/PurchaseHandler.swift
JZDesign: When we call `setResult` the resulting published event is not one where we are dismissing the view. We call dismiss when `purchased` is set to true from within the `performPurchase` function.

JZDesign: So, I don't think that we need to worry about that case.

Sources/Purchasing/Purchases/TransactionNotifications.swift
```diff
@@ -0,0 +1,24 @@
...
+}
+
+extension NotificationCenter {
+    // swiftlint:disable:next missing_docs
```
JZDesign:
```suggestion
    /// A publisher that wraps the `purchaseCompleted` notification that will allow us to propagate 
    /// those events for transactions that were not initiated directly by the Purchases SDK 
    /// (like promotional offers)
    ///
    /// - Important: This is not intended for public consumption and should be used with care
```

RevenueCatUI/Helpers/PurchaseResultComparator.swift
```diff
@@ -0,0 +1,20 @@
...
+enum PurchaseResultComparator {
+    static func compare(
+        _ lhs: PurchaseResultData?,
+        _ rhs: PurchaseResultData
+    ) -> Bool {
+        return lhs?.transaction == rhs.transaction &&
+            lhs?.userCancelled == rhs.userCancelled &&
+            lhs?.customerInfo == rhs.customerInfo
+    }
+}
```
joshdholtz, line 20: Is there any reason why we can't use an `Equatable` on `PurchaseResultData`?

I was going to suggest renaming `compare` to `equals` but I think `Equatable` actually makes more sense maybe???? 🤷 

But if not, I would still suggest making this `equals` since that matches a `Bool` a bit more

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -27,6 +28,8 @@ final class PurchaseHandler: ObservableObject {
 
     }
 
+    private var cancellables: Set<AnyCancellable> = Set()
```
joshdholtz, line 31: Do we need to cancel this on `deinit` at all? Is that a thing? 😅

RevenueCatUI/Helpers/PurchaseResultComparator.swift
JZDesign: Its a tuple, cant make it equatable

joshdholtz: OHHHHH.... didn't realize it was a tuple 😅  Thoughts on changing this to `equals` then? 🤷‍♂️

JZDesign: No issue with it. Comparator + compare is where my brain was at.

And I felt weird about calling it an Equator 🤣

RevenueCatUI/Purchasing/PurchaseHandler.swift
JZDesign: Looking into if this is needed. It is a thing though. Good eye.

RevenueCatUI/Helpers/PurchaseResultComparator.swift
joshdholtz: Ha! `Equator` is definitely weird 😛  I'm totally fine with this then 😊

RevenueCatUI/Purchasing/PurchaseHandler.swift
JZDesign: Sure, I'll throw it in there. Not a major vector but it is one.

joshdholtz (approved):
:shipit:

PR #5658, opened 2025-10-14
[Experimental] Add Locale to Storefront

tonidero:
### Description
This adds a couple of new APIs as experimental APIs to obtain a `Locale` object containing the storefront country code. This can make it easier to obtain a 2-letter ISO 3166-1 alpha-2 by doing `locale.regionCode` in case it's needed and to match the code returned by Android SDKs.

### DIFF
--- a/Sources/Purchasing/Purchases/Purchases.swift
+++ b/Sources/Purchasing/Purchases/Purchases.swift
@@ -2286,9 +2286,6 @@ private extension Purchases {
                         group.addTask {
                             await cache.warmUpPaywallImagesCache(offerings: offerings)
                         }
-                        group.addTask {
-                            await cache.warmUpPaywallVideosCache(offerings: offerings)
-                        }
 						group.addTask {
                             await cache.warmUpPaywallFontsCache(offerings: offerings)
                         }
--- a/RevenueCat.xcodeproj/project.pbxproj
+++ b/RevenueCat.xcodeproj/project.pbxproj
@@ -55,6 +55,9 @@
 		03F446262D2FE1510046129A /* MaskShapePropertyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03F446252D2FE1510046129A /* MaskShapePropertyTests.swift */; };
 		03F446552D303E350046129A /* PaddingPropertyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03F446542D303E350046129A /* PaddingPropertyTests.swift */; };
 		03F446572D303FA10046129A /* CornerRadiusesPropertyTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 03F446562D303FA10046129A /* CornerRadiusesPropertyTests.swift */; };
+		1622D3FB2E900DE000C20E3C /* ChecksumTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1622D3FA2E900DE000C20E3C /* ChecksumTests.swift */; };
+		1622D3FD2E900F8200C20E3C /* Checksum.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1622D3FC2E900F8200C20E3C /* Checksum.swift */; };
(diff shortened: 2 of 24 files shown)


tonidero:
@RCGitBot please test

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -251,6 +251,18 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
         systemInfo.storefront?.countryCode
     }
 
+    @available(iOS 16.0, macOS 13.0, tvOS 16.0, watchOS 9.0, *)
+    @_spi(Experimental) @objc public var storeFrontLocale: Locale? {
```
rickvdl, line 255: I think we use lowercase f in `storefront` everywhere including the Android equivalent [PR](https://github.com/RevenueCat/purchases-android/pull/2726)

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -251,6 +251,18 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
...
+        if let countryCode = systemInfo.storefront?.countryCode {
```
rickvdl: Nit: We can also do the same like `.let {}` in Swift like 
```
systemInfo.storefront?.countryCode.map { countryCode in 
    Locale(components: .init(
        languageCode: nil,
        script: nil,
        languageRegion: .init(countryCode)
    ))
}
```

Tests/StoreKitUnitTests/StorefrontTests.swift
```diff
@@ -45,6 +45,23 @@ class StorefrontTests: StoreKitConfigTestCase {
         }
     }
 
+    @available(iOS 16.0, tvOS 16.0, macOS 13.0, watchOS 9.0, *)
```
rickvdl, line 48: I think we should also add API tests for these, or do we (or should we) not do this for @_spi(Experimental)?

Sources/Purchasing/Purchases/Purchases.swift
tonidero: I actually copied this from the property above `storeFrontCountryCode`. I agree this should be lowercase, but I wonder if it's better to keep it for consistency... Wdyt?

rickvdl: Ah you're right, that's unfortunate but I agree, let's keep it consistent in that case

Tests/StoreKitUnitTests/StorefrontTests.swift
tonidero: Well, with Experimental APIs, it's "okish" to change the API, since it's experimental so we don't guarantee stability. We could add API tests as a way for us to be aware of any breaking changes, but I think it shouldn't block us from those changes at the end of the day. FWIW, in Android we haven't added these but if we agree, happy to add those API tests

rickvdl: True that also makes sense. Maybe it is nice to remind ourselves that we're changing something public facing. I agree we should be allowed to change it if necessary since it's experimental. But no strong opinion on this, it's a bit of extra work so I'm fine with both :)

Sources/Purchasing/Purchases/Purchases.swift
tonidero: Hmm getting a `Type of expression is ambiguous without a type annotation` with this, not sure why 🤔

rickvdl: Sorry it should've been
```
systemInfo.storefront.map { storefront in
    Locale(components: .init(
        languageCode: nil,
        script: nil,
        languageRegion: .init(storefront.countryCode)
    ))
}
```
since `storefront` is the optional property here

tonidero: That works! https://github.com/RevenueCat/purchases-ios/pull/5658/commits/6f4fba981d4b1966b679f4482ed09a639cc04018

JayShortway (approved):
No additional comments! 🙌

PR #5660, opened 2025-10-14
Fix: Promo/Intro Offer Text not always displaying

JZDesign:
### Checklist

### Motivation

We have a timing issue with intro offers and promo offers. There is the possibility of the task completing after the screen renders. If that is the case, the cache will be empty and we will not be able to render The appropriate text.

### Description

This makes the view update its ID after those checks complete, forcing a redraw of the paywall screen.

### DIFF
--- a/RevenueCatUI/Templates/V2/PaywallsV2View.swift
+++ b/RevenueCatUI/Templates/V2/PaywallsV2View.swift
@@ -201,17 +201,18 @@ struct PaywallsV2View: View {
                         }
                     }
                     .task {
-                        guard didFinishEligibilityCheck else {
-                            async let introCheck: Void = introOfferEligibilityContext.computeEligibility(
-                                for: paywallState.packages
-                            )
-                            async let promoCheck: Void = paywallPromoOfferCache.computeEligibility(
-                                for: paywallState.packageInfos.map { ($0.package, $0.promotionalOfferProductCode) }
-                            )
-                            _ = await (introCheck, promoCheck)
-                            didFinishEligibilityCheck = true
+                        guard !didFinishEligibilityCheck else {
                             return
                         }
+
+                        async let introCheck: Void = introOfferEligibilityContext.computeEligibility(
+                            for: paywallState.packages
+                        )
+                        async let promoCheck: Void = paywallPromoOfferCache.computeEligibility(
+                            for: paywallState.packageInfos.map { ($0.package, $0.promotionalOfferProductCode) }
+                        )
+                        _ = await (introCheck, promoCheck)
+                        didFinishEligibilityCheck = true
                     }
                     // Note: preferences need to be applied after `.toolbar` call
                     .preference(key: PurchaseInProgressPreferenceKey.self,


RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -107,6 +107,13 @@ struct PaywallsV2View: View {
     private let purchaseHandler: PurchaseHandler
     private let onDismiss: () -> Void
     private let fallbackContent: FallbackContent
+    @State private var didFinishEligibilityCheck: Bool = false
```
facumenzella, line 110: Is there a reason to put this here and not in the `StateObject`?

JZDesign: The contents of the PaywallInfo state seem to do more with the content of the paywall, and not the timing of things. While the timing does impact the state, this made more sense to me here and not in that PaywallInfo state result. Open to it though.

facumenzella: I don't have a strong opinion tbh. We can always refactor it if it becomes messy enough

RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -193,12 +201,17 @@ struct PaywallsV2View: View {
                         }
                     }
                     .task {
```
facumenzella, line 203: Maybe we can use `task(id:)` instead? That will only run the task when the id changes right?

RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -107,6 +107,13 @@ struct PaywallsV2View: View {
     private let purchaseHandler: PurchaseHandler
     private let onDismiss: () -> Void
     private let fallbackContent: FallbackContent
+    @State private var didFinishEligibilityCheck: Bool = false
+
+    // There is a timing issue where the screen will completely render before the offers cache is primed
+    // This is necessary to ensure that the view displays the offer text when it is available
+    private var redrawTrigger: String {
+        "PaywallsV2View-\(didFinishEligibilityCheck ? "Checked" : "pending")"
```
ajpallares: We probably need a `return` here (for Xcode 14)

RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -176,6 +183,7 @@ struct PaywallsV2View: View {
                         uiConfigProvider: self.uiConfigProvider,
                         onDismiss: self.onDismiss
                     )
+                    .id(redrawTrigger)
```
ajpallares, line 186: Why not this instead? This way, we don't need to have the `redrawTrigger` String that could only add noise, as it's nothing but a mapping on `didFinishEligibilityCheck`
```suggestion
                    .id(didFinishEligibilityCheck)
```

RevenueCatUI/Templates/V2/PaywallsV2View.swift
ajpallares: I was confused by this as well. But, I _think_, since the `task` is declared **after** the `id`, I don't think it will be called twice even after the `id` changes, since the `id` is only affecting the 
```
LoadedPaywallsV2View(
...
)
``` 

but not the task itself.

RevenueCatUI/Templates/V2/PaywallsV2View.swift
JZDesign: Because that is not very unique, if the SwiftUI renderer has another view in the hierarchy with a boolean as the id we can have issues… right?

RevenueCatUI/Templates/V2/PaywallsV2View.swift
JZDesign: I want to ensure that the view redraws when the id changes. But I also want to prevent running the task when it's not needed so that's why I put the guard statement in there.

RevenueCatUI/Templates/V2/PaywallsV2View.swift
JZDesign:
```suggestion
        return "PaywallsV2View-\(didFinishEligibilityCheck ? "Checked" : "pending")"
```

RevenueCatUI/Templates/V2/PaywallsV2View.swift
ajpallares: Oh, right! That makes sense. I'm not a fan of having strings hardcoded like this, but I get the benefit. Idk, a UUID could suffice and perhaps be cleaner? Not a strong opinion though

JZDesign: I would tend to agree for a component where there could be many of them that should have unique ID's, but for the entire container that should be absolutely unique, I lean toward the hardcoded. Since this is the container that should only ever have one instance displayed at a time, I think hard coding is okay.

RevenueCatUI/Templates/V2/PaywallsV2View.swift
```diff
@@ -193,12 +201,17 @@ struct PaywallsV2View: View {
...
+                            async let introCheck: Void = introOfferEligibilityContext.computeEligibility(
+                                for: paywallState.packages
+                            )
+                            async let promoCheck: Void = paywallPromoOfferCache.computeEligibility(
+                                for: paywallState.packageInfos.map { ($0.package, $0.promotionalOfferProductCode) }
+                            )
+                            _ = await (introCheck, promoCheck)
+                            didFinishEligibilityCheck = true
+                            return
+                        }
```
joshdholtz: Just a nit to move the code out of the guard for a little easier reading 🤷‍♂️ Or... easier to me at least 😛 

```suggestion
                        guard !didFinishEligibilityCheck else {
                            return
                        }
                        
                        async let introCheck: Void = introOfferEligibilityContext.computeEligibility(
                            for: paywallState.packages
                        )
                        async let promoCheck: Void = paywallPromoOfferCache.computeEligibility(
                            for: paywallState.packageInfos.map { ($0.package, $0.promotionalOfferProductCode) }
                        )
                        _ = await (introCheck, promoCheck)
                        didFinishEligibilityCheck = true
```

joshdholtz (approved):
One nit but otherwise this makes sense to me! 👍

PR #5663, opened 2025-10-15
PoC: allow overriding currencySymbol from a backend-provided ruleset

rickvdl:
We're having an issue related to currency formatting behavior in Romania (and possibly a few other currencies), internal discussion can be found [here](https://revenuecat.slack.com/archives/C02CD1TLS76/p1759416234192899).

But some context: when using the Romanian Storefront, and a Romanian locale (ro_RO) prices are formatted as `1,23 RON`, which is correct, but the more conventional way to write this in Romanian (according to reports) is `1,23 lei`.

`underlyingSK2Product.displayPrice` uses the more common `lei` rather than `RON`, however the priceFormatter we create in  `SK2StoreProduct.swift` through `priceFormatterProvider.priceFormatterForSK2` based on the currency code and locale given to us by StoreKit creates strings using `RON` instead. This is causing discrepancies between `localizedPriceString` and the price strings we generate using the formatter (for example pricePerMonth) which looks confusing on the paywall especially.

After some research and discussions it looks like we can't get the NumberFormatter API to mimic the behavior of StoreKit, and thus we've been looking into an alternative approach.

We came up with a backend driven approach where we (given a ruleset) provide the option to override the currencySymbol for specific currency codes, given a storekit storefront (similar to what we already do for paywall localizations).

This PR is just a PoC outlining the rough workings of this approach. Happy to hear any feedback

### DIFF
--- a/Sources/Misc/PriceFormatterProvider.swift
+++ b/Sources/Misc/PriceFormatterProvider.swift
@@ -74,11 +74,12 @@ public final class PriceFormatterProvider: Sendable {
                 currencyCode: currencyCode
             )
             
-            if let formatter = formatter as? CurrencySymbolOverridingPriceFormatter, formatter.currencyCode == currencyCode, formatter.locale == locale, formatter.currencySymbolOverride == currencySymbolOverride {
-                return formatter
+            if let formatter = formatter as? CurrencySymbolOverridingPriceFormatter {
+                if formatter.currencyCode == currencyCode, formatter.locale == locale, formatter.currencySymbolOverride == currencySymbolOverride {
+                    return formatter
+                }
             }
-            
-            if let formatter = formatter, formatter.currencyCode == currencyCode, formatter.locale == locale {
+            else if let formatter = formatter, formatter.currencyCode == currencyCode, formatter.locale == locale {
                 return formatter
             }
             


Sources/Networking/Responses/OfferingsResponse.swift
```diff
@@ -54,7 +54,62 @@ struct OfferingsResponse {
...
+    ]
+    
+    public func currencySymbolOverride(
+        for storefrontCountryCode: String,
+        currencyCode: String
+    ) -> CurrencySymbolOverride? {
+        return self.currencySymbolOverrides[storefrontCountryCode]?[currencyCode]
+    }
+    
+    public struct CurrencySymbolOverride: Sendable {
```
fire-at-will, line 92: Nit: I might consider adding some DocC to explain the properties here - it took me a while to realize what they were meant for

Sources/Misc/PriceFormatterProvider.swift
```diff
@@ -15,7 +15,13 @@ import Foundation
 
 /// A `NumberFormatter` provider class for prices.
 /// This provider caches the formatter to improve the performance.
-final class PriceFormatterProvider: Sendable {
+public final class PriceFormatterProvider: Sendable {
```
fire-at-will, line 18: This probably doesn't need to be public public if it's just RevenueCatUI using this, I'd probably add `@_spi(Internal)` so it's only used by RevenueCatUI

```suggestion
@_spi(Internal) public final class PriceFormatterProvider: Sendable {
```

Sources/Misc/PriceFormatterProvider.swift
```diff
@@ -43,25 +49,50 @@ final class PriceFormatterProvider: Sendable {
...
+               )
             }
-
-            return formatter
+            
+            if let formatter = formatter as? CurrencySymbolOverridingPriceFormatter, formatter.currencyCode == currencyCode, formatter.locale == locale, formatter.currencySymbolOverride == currencySymbolOverride {
+                return formatter
+            }
+            
+            if let formatter = formatter, formatter.currencyCode == currencyCode, formatter.locale == locale {
```
fire-at-will: In the logic here:

```swift
if let formatter = formatter as? CurrencySymbolOverridingPriceFormatter, formatter.currencyCode == currencyCode, formatter.locale == locale, formatter.currencySymbolOverride == currencySymbolOverride {
                return formatter
            }
            
            if let formatter = formatter, formatter.currencyCode == currencyCode, formatter.locale == locale {
                return formatter
            }
```

The first `if` checks if the `formatter` is a `CurrencySymbolOverridingPriceFormatter` and that all of the parameters match. If it fails, the second `if` still returns the `formatter` as long as the formatter's currency code & locale match.

Could this lead to a bug where if the formatter is a `CurrencySymbolOverridingPriceFormatter`, the overrides don't match, but the currency code & locale do, we return a `CurrencySymbolOverridingPriceFormatter` that doesn't match the current context?

Sources/Misc/PriceFormatterProvider.swift
```diff
@@ -90,5 +121,64 @@ final class PriceFormatterProvider: Sendable {
...
+        switch intValue {
+        case 0: return .zero
+        case 1: return .one
+        case 2: return .two
+        default: return .other
+        }
+    }
+    
+    private func formatter(for rule: PriceFormattingRuleSet.CurrencySymbolOverride.PluralRule) -> NumberFormatter {
+        if let formatter = numberFormatterCache[rule] {
```
fire-at-will, line 172: Love this 👏👏

Sources/Networking/Responses/OfferingsResponse.swift
```diff
@@ -54,7 +54,62 @@ struct OfferingsResponse {
     let placements: Placements?
     let targeting: Targeting?
     let uiConfig: UIConfig?
+    let config: Config?
```
fire-at-will, line 57: Nit: we should probably name this something more specific, like `priceFormattingConfig`, in case we decide to add other configs in the future

Sources/Misc/PriceFormatterProvider.swift
rickvdl: Unfortunately I don't think this will work, the SK2 product initializer in `StoreProduct` is public and has this type in the function signature. Will make sure to bring this up in the actual PR

Sources/Networking/Responses/OfferingsResponse.swift
rickvdl: Good one, I figured we might want to use this as a 'config envelope' to also hold future configuration values, in order to avoid having to update all initializers and mocks again. But I will make sure to bring this up in my PR and ask for the team's opinion on this as well

Sources/Misc/PriceFormatterProvider.swift
rickvdl: Good catch, I think you're right. Addressed this :)

PR #5664, opened 2025-10-15
Migrate ReceiptParser to Tuist

facumenzella:
### Motivation
We want to finally clean-up the main workspace, in favor of using the Tuist generated one for internal development. In order to do that, we need to slowly move nested projects to Tuist, deleting them from the main one.

ReceiptParser is already in the Tuist set up, so we just need to remove references from the main workspace.

Follow-up of https://github.com/RevenueCat/purchases-ios/pull/5648


facumenzella:
@RCGitBot please test

facumenzella:
@RCGitBot please test

Tests/MockBundle.swift
```diff
@@ -0,0 +1,53 @@
+//
+//  Copyright RevenueCat Inc. All Rights Reserved.
+//
+//  Licensed under the MIT License (the "License");
```
facumenzella, line 4: This is not correct 🤔

PR #1341, opened 2022-02-25
Make StoreProduct initializers public

delba:
Resolves #1340

PR #5676, opened 2025-10-17
Implement Price formatting ruleset for overriding currency symbol's per storefront and currency

rickvdl:
Based on this [PoC PR](https://github.com/RevenueCat/purchases-ios/pull/5663). 

> We're having an issue related to currency formatting behavior in Romania (and possibly a few other currencies), internal discussion can be found [here](https://revenuecat.slack.com/archives/C02CD1TLS76/p1759416234192899).

> But some context: when using the Romanian Storefront, and a Romanian locale (ro_RO) prices are formatted as 1,23 RON, which is correct, but the more conventional way to write this in Romanian (according to reports) is 1,23 lei.

> underlyingSK2Product.displayPrice uses the more common lei rather than RON, however the priceFormatter we create in  SK2StoreProduct.swift through priceFormatterProvider.priceFormatterForSK2 based on the currency code and locale given to us by StoreKit creates strings using RON instead. This is causing discrepancies between localizedPriceString and the price strings we generate using the formatter (for example pricePerMonth) which looks confusing on the paywall especially.

> After some research and discussions it looks like we can't get the NumberFormatter API to mimic the behavior of StoreKit, and thus we've been looking into an alternative approach.

> We came up with a backend driven approach where we (given a ruleset) provide the option to override the currencySymbol for specific currency codes, given a storekit storefront (similar to what we already do for paywall localizations).

### Checklist
- [ ] Check if we need any web-billing specific implementations

### DIFF
--- a/Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
+++ b/Tests/UnitTests/Purchasing/OfferingsManagerTests.swift
@@ -955,7 +955,8 @@ private extension OfferingsManagerTests {
                 ],
                 placements: nil,
                 targeting: nil,
-                uiConfig: nil
+                uiConfig: nil,
+                config: nil
             ),
             httpResponseOriginalSource: .mainServer
         )


Sources/Networking/Responses/OfferingsResponse.swift
```diff
@@ -54,7 +54,70 @@ struct OfferingsResponse {
...
+    }
+    
+    /*
+     Contains a set of currencySymbol overrides for different pluralization rules
+     */
+    public struct CurrencySymbolOverride: Sendable {
+        let zero: String
+        let one: String
+        let two: String
+        let few: String
```
rickvdl, line 127: `few` / `many` aren't really used right now, should we remove them?

Sources/Networking/Responses/OfferingsResponse.swift
```diff
@@ -54,7 +54,70 @@ struct OfferingsResponse {
     let placements: Placements?
     let targeting: Targeting?
     let uiConfig: UIConfig?
+    let config: Config?
 
+    public struct Config {
```
rickvdl: I added a `Config` 'envelope' here in order to future-proof ourselves for more configuration values. But we can also just call it `priceFormattingRuleSets` and only have this single purpose field here, happy to go either way

Sources/Purchasing/StoreKit2/ProductsFetcherSK2.swift
```diff
@@ -34,7 +40,10 @@ actor ProductsFetcherSK2 {
             }
 
             Logger.rcSuccess(Strings.storeKit.store_product_request_received_response)
-            return Set(storeKitProducts.map { SK2StoreProduct(sk2Product: $0) })
+            return Set(storeKitProducts.map { SK2StoreProduct(
+                sk2Product: $0,
+                priceFormatterProvider: .init(priceFormattingRuleSet: priceFormattingRuleSetProvider.priceFormattingRuleSet())
```
rickvdl: This will hand the SKProduct the value type (`PriceFormattingRuleSet`) for the current storefront. However when the storefront changes the storekit cache is cleared and the storekit products are refetched, so I think this is fine

Sources/Purchasing/StoreKitAbstractions/SK1StoreProductDiscount.swift
```diff
@@ -47,7 +47,7 @@ internal struct SK1StoreProductDiscount: StoreProductDiscountType {
         return self.priceFormatter.string(from: self.underlyingSK1Discount.price) ?? ""
     }
 
-    private let priceFormatterProvider: PriceFormatterProvider = .init()
+    private let priceFormatterProvider: PriceFormatterProvider = .init(priceFormattingRuleSet: nil) // todo rick
```
rickvdl: tracking todo

Tests/StoreKitUnitTests/PriceFormatterProviderTests.swift
```diff
@@ -94,5 +156,151 @@ class PriceFormatterProviderTests: StoreKitConfigTestCase {
...
+        let priceFormatter = priceFormatterProvider.priceFormatterForSK2(
+            withCurrencyCode: "RON",
+            locale: .init(identifier: "ro_RO")
+        )
+        
+        expect(priceFormatter.currencyCode) == "RON"
+        expect(priceFormatter.currencySymbol) == "RON"
+        XCTAssert(type(of: priceFormatter) == CurrencySymbolOverridingPriceFormatter.self)
+        
+        XCTAssertEqual(priceFormatter.string(from: NSNumber(integerLiteral: 0)), "0,00 lei")
```
rickvdl: real world example

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -393,14 +393,26 @@ public typealias StartPurchaseBlock = (@escaping PurchaseCompletedBlock) -> Void
         let offeringsFactory = OfferingsFactory()
         let receiptParser = PurchasesReceiptParser.default
         let transactionsManager = TransactionsManager(receiptParser: receiptParser)
+        
+        let priceFormattingRuleSetProvider: PriceFormattingRuleSetProvider = .init {
+            var priceFormattingRuleSet: PriceFormattingRuleSet?
+            if Self.isConfigured, let offeringsResponse = Self.shared.cachedOfferings?.response,
```
rickvdl: I'm wondering whether this is the way to go to get the latest offeringsResponse, happy to hear any alternatives if there are better ways of doing this

rickvdl:
@RCGitBot please test

Sources/Misc/PriceFormatterProvider.swift
```diff
@@ -71,24 +98,123 @@ final class PriceFormatterProvider: Sendable {
...
+        guard isInteger else { return .other }
+
+        // Integer-only mapping; locale-specific categories like "few"/"many" are handled elsewhere.
+        switch intValue {
+        case 0: return .zero
+        case 1: return .one
+        case 2: return .two
+        default: return .other
+        }
+    }
```
JayShortway, line 213: I think this should follow the same rules as we use for [`VariableLocalizationKey`s](https://github.com/RevenueCat/purchases-ios/blob/9bcbc56da361381a0e1569dedde2325819c710f1/RevenueCatUI/Templates/V2/Variables/VariableHandlerV2.swift#L808-L865).

Sources/Networking/Responses/OfferingsResponse.swift
JayShortway: I'm fine either way too, but I can see us having this `Config` wrapper with a single key forever. I also wouldn't mind putting it in `UIConfig`.

Sources/Networking/Responses/OfferingsResponse.swift
JayShortway: No I wouldn't. We don't know which other rules and conventions we encounter in the future. This way we automatically support those in older SDKs as well without having to release updates.

JayShortway (commented):
Some initial comments. Great job investigating this and adding support for this! 💪

Sources/Misc/PriceFormatterProvider.swift
```diff
@@ -15,27 +15,61 @@ import Foundation
 
 /// A `NumberFormatter` provider class for prices.
 /// This provider caches the formatter to improve the performance.
-final class PriceFormatterProvider: Sendable {
+public final class PriceFormatterProvider: Sendable {
```
ajpallares: I wonder, why should this API be public? Its `init` is not public.
It seems to me like it's only public to allow `StoreProduct`'s new `init` be public
```swift
convenience init(
    sk2Product: SK2Product,
    priceFormatterProvider: PriceFormatterProvider
)
```
But I think that one shouldn't be public either?

Sources/Networking/Responses/OfferingsResponse.swift
ajpallares: Yep. I think we may as well put it in `UIConfig`, as one could argue that this affects UI (how prices are presented in the UI).

ajpallares: This can be `@_spi(Internal) public` I think, as this would only be needed in RevenueCatUI and would not need to be visible by integrators

rickvdl: Good points, will move it to UIConfig

Sources/Misc/PriceFormatterProvider.swift
rickvdl: @ajpallares indeed that's the reason, but [this](https://github.com/RevenueCat/purchases-ios/blob/main/Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift#L311) was already public. There also seems to be a public API tester for so I guess it is used by developers

ajpallares: I understand. It seems like that init was made public in #1341. But I don't see why the new init that receives a `priceFormatterProvider` needs to be public. Initially, I think we should not make it public unless we find a good reason. I would keep it internal so that `PriceFormatterProvider` can also be internal

Sources/Networking/Responses/OfferingsResponse.swift
rickvdl: Moved them to UIConfig, however on tvOS it's an empty structure. I think this should still be supported on tvOS so added it to that struct as well. LMK if you (dis)agree

Sources/Misc/PriceFormatterProvider.swift
rickvdl: Ah you're right, my bad! I've made it internal again :) Thanks for clarifying

ajpallares: Also, I would make this `PriceFormatterProvider` internal (probably we need it to be `@_spi(Internal) public`). I see that it's needed in the `UIConfig`'s `init` that is already public (but that init should actually be `@_spi(Internal) public` instead). 

I would add the `@_spi(Internal)` marker to that init in `UIConfig` to allow making this `@_spi` as well.
There's precedent: in #5208 we added the `@_spi(Internal)` marker to some existing public subtypes of `UIConfig`, with no complain from anyone as these APIs are not for public use.

Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift
```diff
@@ -306,10 +306,20 @@ extension StoreProduct {
...
+    public convenience init(
+        sk2Product: SK2Product
+    ) {
+        self.init(SK2StoreProduct(sk2Product: sk2Product, priceFormatterProvider: .init()))
+    }
 
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
-    public convenience init(sk2Product: SK2Product) {
-        self.init(SK2StoreProduct(sk2Product: sk2Product))
+    public convenience init(
```
ajpallares: Just to be sure: this one should be `internal`

Sources/Misc/PriceFormatterProvider.swift
ajpallares: (unresolving to make sure this is addressed)

rickvdl: Good point and agree, will make that change :)

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift
rickvdl: Should be done :)

Sources/Misc/PriceFormatterProvider.swift
rickvdl: Should be matching now :)

rickvdl:
@RCGitBot please test

Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
```diff
@@ -116,20 +116,95 @@ public struct UIConfig: Codable, Equatable, Sendable {
...
+        // storefront country code -> ruleset
+        String: PriceFormattingRuleSet
+    ]
+
+    @_spi(Internal)
     public init(app: AppConfig,
                 localizations: [String: [String: String]],
-                variableConfig: VariableConfig) {
+                variableConfig: VariableConfig,
+                priceFormattingRuleSets: [String: PriceFormattingRuleSet]) {
```
JayShortway: Hmm, I'm having second thoughts about putting this in `UIConfig`, since this is needed even without paywalls.

ajpallares: Oh that's a very good point. In fact, we've recently seen how `UIConfig` is not received in certain situations (e.g. when there are no published v2 paywalls), which would break this.
So... operation make Get Offerings even bigger (i.e. new field)? 😬

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

RevenueCatUI/Templates/V2/Previews/TemplateComponentsViewPreviews/ButtonWithFooterPreview.swift
```diff
@@ -12,7 +12,7 @@
 //  Created by Josh Holtz on 9/26/24.
 
 import Foundation
-import RevenueCat
+@_spi(Internal) import RevenueCat
```
ajpallares: I'm curious. Why is this needed here now?

Sources/Purchasing/SimulatedStore/SimulatedStoreProductsManager.swift
```diff
@@ -64,4 +64,7 @@ final class SimulatedStoreProductsManager: ProductsManagerType {
     // This class does not implement caching.
     // See `CachingProductsManager`.
     func clearCache() { }
+
+    // todo rick: implement?
+    func updatePriceFormattingRuleSet(_ ruleSet: PriceFormattingRuleSet?) { }
```
ajpallares, line 69: TODO alert 😅

Sources/Misc/PriceFormattingRuleSetProvider.swift
```diff
@@ -0,0 +1,36 @@
...
+//
+//  PriceFormattingRuleSetProvider.swift
+//
+//  Created by Rick van der Linden on 26/11/2025.
+
+// MARK: -
+final class PriceFormattingRuleSetProvider: @unchecked Sendable {
+    private let priceFormattingRuleSet: Atomic<PriceFormattingRuleSet?>
+
+    init(priceFormattingRuleSet: PriceFormattingRuleSet? = nil) {
```
ajpallares: Do we really need a default value? I always try to avoid them as much as possible as these usually bite back in the future

Sources/Networking/Responses/OfferingsResponse.swift
```diff
@@ -49,12 +49,24 @@ struct OfferingsResponse {
         let ruleId: String
     }
 
+    struct Config {
+        @DefaultDecodable.EmptyDictionary
+        var priceFormattingRuleSets: [
+            // storefront country code -> ruleset
+            String: PriceFormattingRuleSet
+        ]
```
ajpallares: Small suggestion. I think this is clearer to read

```suggestion
        // storefront country code -> ruleset
        var priceFormattingRuleSets: [String: PriceFormattingRuleSet]
```

Sources/Networking/Responses/OfferingsResponse.swift
```diff
@@ -83,6 +95,64 @@ extension OfferingsResponse.Offering.Package: Codable, Equatable {}
...
 extension OfferingsResponse.Placements: Codable, Equatable {}
 extension OfferingsResponse.Targeting: Codable, Equatable {}
+extension OfferingsResponse.Config: Codable, Equatable {}
 extension OfferingsResponse: Codable, Equatable {}
 
 extension OfferingsResponse: HTTPResponseBody {}
+
+/*
+ Contains a set of rules that will be used when formatting a price
+ Currrently only supports overriding the currencySymbol per currency
```
ajpallares:
```suggestion
 Currently only supports overriding the currencySymbol per currency
```

Sources/Networking/Responses/OfferingsResponse.swift
```diff
@@ -83,6 +95,64 @@ extension OfferingsResponse.Offering.Package: Codable, Equatable {}
...
+ */
+struct PriceFormattingRuleSet: Sendable {
+
+    // currencyCode: CurrencySymbolOverride
```
ajpallares: To follow the format of the similar comment above

```suggestion
    // currencyCode -> CurrencySymbolOverride
```

Sources/Purchasing/StoreKitAbstractions/SK1StoreProductDiscount.swift
```diff
@@ -16,7 +16,7 @@ import StoreKit
 @available(iOS 11.2, macOS 10.13.2, tvOS 11.2, watchOS 6.2, *)
 internal struct SK1StoreProductDiscount: StoreProductDiscountType {
 
-    init?(sk1Discount: SK1ProductDiscount) {
+    init?(sk1Discount: SK1ProductDiscount, priceFormatterProvider: PriceFormatterProvider? = nil) {
```
ajpallares, line 19: Same question. Is the default value really needed?

Sources/Purchasing/StoreKitAbstractions/SK2StoreProduct.swift
```diff
@@ -16,8 +16,12 @@ import StoreKit
 @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
 internal struct SK2StoreProduct: StoreProductType {
 
-    init(sk2Product: SK2Product) {
+    init(
+        sk2Product: SK2Product,
+        priceFormatterProvider: PriceFormatterProvider = .init(priceFormattingRuleSetProvider: nil)
```
ajpallares, line 21: Same question

Sources/Purchasing/StoreKitAbstractions/StoreProduct.swift
```diff
@@ -308,8 +308,21 @@ extension StoreProduct {
     }
 
     @available(iOS 15.0, tvOS 15.0, watchOS 8.0, macOS 12.0, *)
-    public convenience init(sk2Product: SK2Product) {
-        self.init(SK2StoreProduct(sk2Product: sk2Product))
+    public convenience init(
+        sk2Product: SK2Product
+    ) {
```
ajpallares, line 313: Nit: why change the original format if there's only 1 parameter?
```suggestion
    public convenience init(sk2Product: SK2Product) {
```
In this case, I think it reads easier in one line

Sources/Purchasing/StoreKitAbstractions/StoreProductDiscount.swift
```diff
@@ -219,8 +219,11 @@ extension StoreProductDiscount {
 
 extension StoreProductDiscount {
 
-    internal convenience init?(sk1Discount: SK1ProductDiscount) {
-        guard let discount = SK1StoreProductDiscount(sk1Discount: sk1Discount) else { return nil }
+    internal convenience init?(sk1Discount: SK1ProductDiscount, priceFormatterProvider: PriceFormatterProvider? = nil) {
```
ajpallares, line 222: Same comment about default value

Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
```diff
@@ -116,20 +116,20 @@ public struct UIConfig: Codable, Equatable, Sendable {
     public var localizations: [String: [String: String]]
     public var variableConfig: VariableConfig
 
+    @_spi(Internal)
```
ajpallares: Note that this is, strictly speaking, a breaking API change. BUT, I think it's fine in this because this is not really meant to be a public API. In fact, we've done similar things in the past in these objects (e.g. in #5208 for some types in `UIConfig`)

Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
```diff
@@ -116,20 +116,20 @@ public struct UIConfig: Codable, Equatable, Sendable {
...
         self.variableConfig = variableConfig
     }
-
 }
 
 #else
 
 public struct UIConfig: Codable, Equatable, Sendable {
-
+    public init() {}
```
ajpallares: I think this is not needed?

Sources/Purchasing/ProductsManager.swift
```diff
@@ -24,6 +24,7 @@ class ProductsManager: NSObject, ProductsManagerType {
     private let diagnosticsTracker: DiagnosticsTrackerType?
     private let systemInfo: SystemInfo
     private let dateProvider: DateProvider
+    let priceFormattingRuleSetProvider: PriceFormattingRuleSetProvider
```
ajpallares: I think this can be private?
```suggestion
    private let priceFormattingRuleSetProvider: PriceFormattingRuleSetProvider
```

Tests/StoreKitUnitTests/PriceFormattingRuleSetIntegrationTests.swift
```diff
@@ -0,0 +1,308 @@
...
+
+        receivedProducts = try await manager.products(withIdentifiers: Set([identifier]))
+
+        expect(receivedProducts).notTo(beNil())
+        let unwrappedFirstProduct = try XCTUnwrap(receivedProducts?.first)
+        expect(unwrappedFirstProduct.currencyCode) == "RON"
+
+        let unwrappedPriceFormatter = try XCTUnwrap(unwrappedFirstProduct.priceFormatter)
+        // Should be a regular NumberFormatter
+        XCTAssert(type(of: unwrappedPriceFormatter) == NumberFormatter.self)
```
ajpallares, line 172: I don't know if check this is really needed or it is going too deep into the specific implementation 🤔 
Wouldn't it be enough just to check the formatted price?

Tests/StoreKitUnitTests/PriceFormattingRuleSetIntegrationTests.swift
ajpallares, line 1: Very nice set of tests

Tests/StoreKitUnitTests/ProductsManagerTests.swift
```diff
@@ -113,9 +114,86 @@ class ProductsManagerTests: StoreKitConfigTestCase {
...
+        expect(unwrappedFirstProduct.currencyCode) == "USD"
+        expect(unwrappedFirstProduct.localizedPricePerMonth) == "bucks 4.99"
+
+        testSession.locale = Locale(identifier: "nl_NL")
+        try await changeStorefront("NLD")
+
+        // Note: this test passes only because the method `clearCache`
+        // is manually executed. `ProductsManager` does not detect Storefront changes to invalidate the
+        // cache. The changes are now managed by `StoreKit2StorefrontListenerDelegate`.
+        manager.clearCache()
```
ajpallares, line 143: Nice clarification

ajpallares: Although I think it could be clearer if said 
```suggestion
        // Note: this test passes only because the method `clearCache`
        // is manually executed. `ProductsManager` does not detect Storefront changes to invalidate the
        // cache. The changes are now managed by `PurchasesOrchestrator`.
        manager.clearCache()
```
Since `PurchasesOrchestrator` is the actual `StoreKit2StorefrontListenerDelegate`. And also it's perhaps a bit confusing that this is a test for SK1 but we say that the SK2 listener delegate is the one taking care of it.

Not very important though

Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
rickvdl: Good catch, it's a leftover from previous changes, removed!

Sources/Networking/Responses/RevenueCatUI/UIConfig.swift
rickvdl: Good one, this is also a leftover. Since we now have the new Config object this can go back to being internal again

Tests/StoreKitUnitTests/PriceFormattingRuleSetIntegrationTests.swift
rickvdl: I agree actually, this is an implementation detail. Removed these lines from the tests

Tests/StoreKitUnitTests/ProductsManagerTests.swift
rickvdl: I did take this from `OfferingsManagerStoreKitTests` hehe. But updated :)

rickvdl:
@RCGitBot please test

Sources/Misc/PriceFormattingRuleSetProvider.swift
rickvdl: Agree! I just changed it and it looks like there are just a couple spots, in that case let's indeed remove it. Let's see if the tests agree 🤞

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

PR #5677, opened 2025-10-17
Update SampleCat SDK dependency version on SDK releases

ajpallares:
### Description

This PR changes the SampleCat example project to always depend on the latest version of purchases-ios

* Changes the Dependency Rule from commit to Up to Next Major Version
* Makes it so that `bump` fastlane lane also updates the minimumVersion of purchases-ios in SampleCat

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -32,8 +32,11 @@ files_with_version_number_without_prerelease_modifiers = {
   './Tests/UnitTests/Info.plist' => PLIST_VERSION_PATTERNS,
   './Tests/UnitTestsHostApp/Info.plist' => PLIST_VERSION_PATTERNS
 }
+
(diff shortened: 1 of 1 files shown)


fastlane/Fastfile
```diff
@@ -32,8 +32,11 @@ files_with_version_number_without_prerelease_modifiers = {
   './Tests/UnitTests/Info.plist' => PLIST_VERSION_PATTERNS,
   './Tests/UnitTestsHostApp/Info.plist' => PLIST_VERSION_PATTERNS
 }
+
+PACKAGE_VERSION_PATTERNS = ["repositoryURL = \"https://github.com/RevenueCat/purchases-ios\";\n\t\t\trequirement = {\n\t\t\t\tkind = upToNextMajorVersion;\n\t\t\t\tminimumVersion = {x};\n"]
```
ajpallares, line 36: I wanted to be extra safe here in case we eventually add another dependency to SampleCat. That's why I included the purchases-ios repo URL in the pattern to consider. LMK if you think I should tackle this differently

PR #5678, opened 2025-10-17
HOTFIX Media not loading

JZDesign:
### Checklist

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
@@ -146,6 +146,9 @@ struct VideoComponentView: View {
                             .removeDuplicates()
                             .debounce(for: 0.300, scheduler: RunLoop.main)
                     ) { output in
+                        // in the event that the download of the high res video is so fast that it tries to set the
+                        // url moments after the low_res was set, we need to delay a tiny bit to ensure the rerender
+                        // actually occurs. This happens consistently with small file sizes and great connection
                         cachedURL = output
                     }
                 }


joshdholtz (approved):

Sources/Caching/FileRepository.swift
```diff
@@ -130,9 +130,9 @@ import Foundation
     }
 
     func generateLocalFilesystemURL(forRemoteURL url: URL, withChecksum checksum: Checksum?) -> URL? {
+        let path = checksum?.value ?? url.absoluteString.asData.md5String
```
joshdholtz, line 133: Since the checksum should be unique, we don't need the md5 string of the url, correct?

Sources/Caching/LargeItemCacheType.swift
```diff
@@ -46,7 +46,7 @@ extension FileManager: LargeItemCacheType {
 
         // Set up file handling
 
-        let tempFileURL = temporaryDirectory.appendingPathExtension(checksum?.value ?? url.pathExtension)
+        let tempFileURL = temporaryDirectory.appendingPathComponent((checksum?.value ?? "") + url.lastPathComponent)
```
joshdholtz, line 49: Instead of empty string, should this be an md5 of the url? 🤔

Sources/Caching/FileRepository.swift
```diff
@@ -130,9 +130,9 @@ import Foundation
...
         return cacheURL?
-            .appendingPathComponent(url.absoluteString.asData.md5String)
-            .appendingPathExtension(checksum?.value ?? url.pathExtension)
+            .appendingPathComponent(path + url.lastPathComponent)
```
JZDesign, line 135: This is the fix. 2 things are different

1. PathExtension -- putting the checksum in there was a mistake. It should have been a path component that ended with the path extension.
1. I'm no longer chaining the calls. This seems to be a breaking change with iOS 26 (or perhaps the xcode26 build tools), the second call is ignored. Or something is severely wrong with my machine.

This results in the correct output in both Xcode 16.4 and Xcode 26

Sources/Caching/LargeItemCacheType.swift
JZDesign, line 49: This ended up being an issue too after changing the other file, moving didn't work. So I had to update this too.

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
```diff
@@ -139,6 +140,17 @@ struct VideoComponentView: View {
...
+                        stagedURL.publisher
+                            .eraseToAnyPublisher()
+                            .removeDuplicates()
+                            .debounce(for: 0.300, scheduler: RunLoop.main)
+                    ) { output in
+                        // in the event that the download of the high res video is so fast that it tries to set the
+                        // url moments after the low_res was set, we need to delay a tiny bit to ensure the rerender
+                        // actually occurs. This happens consistently with small file sizes and great connection
+                        cachedURL = output
+                    }
```
JZDesign, line 153: This was also not noticed until the editor changes merged because the high res and low res were the same file and therefore we just ignored the second one.

Sources/Caching/FileRepository.swift
JZDesign: That is if we have a checksum. Images don't have one. So I'm falling back to the md5

Sources/Caching/LargeItemCacheType.swift
JZDesign: This file is deleted after we confirm that the download is valid and we move the contents to the permanent file location. So the md5 here matters little to me. I could even get rid of the checksum here I think. I really just need some text + the file extension. So the `url.lastPathComponent` would really be all I need.

I specifically need to include the .mp4 or whatever it is there because the newer build tools (pretty sure it's due to Xcode build tools now, not iOS 26) for one reason or another do not consistently render the video without the file extension.

joshdholtz (approved):

PR #5679, opened 2025-10-17
Fix for paywall image sometimes not showing (in carousel)

joshdholtz:
### Motivation

Fixing issue that caused image to have a small size

### Description

Putting back the Color.clear as the base component in the ZStack


RevenueCatUI/Templates/V2/Components/Image/ImageComponentView.swift
```diff
@@ -60,6 +60,16 @@ struct ImageComponentView: View {
                 )
 
                 ZStack {
+                    // IMPORTANT: Please keep this... needed to force size
+                    //
+                    // We need the max width of the parent view an image of a fill or
+                    // fixed width doesn't push passed the bounds.
+                    //
+                    // Once we have the width once, we can remove the GeometryReader
```
JZDesign:
```suggestion
```

PR #5683, opened 2025-10-20
[Experimental] Add Locale to Storefront (#5658)

JayShortway:
Cherry-pick of https://github.com/RevenueCat/purchases-ios/pull/5658. Also adds some release train goodies from @ajpallares.

### DIFF
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -68,7 +68,8 @@ platform :ios do
   lane :automatic_bump do |options|
     next_version, type_of_bump = determine_next_version_using_labels(
       repo_name: repo_name,
-      github_rate_limit: options[:github_rate_limit]
+      github_rate_limit: options[:github_rate_limit],
+      current_version: current_version_number
     )
     options[:next_version] = next_version
     options[:automatic_release] = true


Tests/APITesters/SwiftAPITester/SwiftAPITester/StorefrontAPI.swift
```diff
@@ -5,14 +5,17 @@
 //  Created by Nacho Soto on 4/13/22.
 //
 
-import RevenueCat
+@_spi(Experimental) import RevenueCat
```
ajpallares, line 8: Not very important (even less so for this PR of major v4), but I'm thinking: would it make sense to keep API tests for experimental APIs in a separate file? Like, they are probably expected to change more often. This way, we can more easily tell them apart from released, non-experimental API tests.

(again, maybe this makes more sense in `main` than here, but I missed it from the original PR). FFTI

JayShortway: Yes, that's a good point, but I'd rather have it in `main` indeed. (I also wouldn't mind not having these tests for experimental APIs.)

PR #5685, opened 2025-10-21
Obj-c compatible APIs for non subscription revenue

polmiro:
### Checklist

### Motivation
Support objc in all new tracking experimental APIs.

### Description
- Converted structs to @objc classes.
- Added completion handler methods as alternative to async for objc.
- Added tests for objc.

### DIFF
--- a/Sources/Ads/Events/AdEvent.swift
+++ b/Sources/Ads/Events/AdEvent.swift
@@ -68,11 +68,11 @@ internal protocol AdEventData {
 @_spi(Experimental) @objc(RCAdDisplayed) public final class AdDisplayed: NSObject, AdEventData {
 
     // swiftlint:disable missing_docs
-    @objc public var networkName: String
-    @objc public var mediatorName: MediatorName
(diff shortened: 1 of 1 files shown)


Sources/Ads/Events/AdEvent.swift
```diff
@@ -26,59 +30,54 @@ internal protocol AdEventData {
...
+@_spi(Experimental) @objc(RCAdDisplayed) public final class AdDisplayed: NSObject, AdEventData {
 
-    /// The ad instance identifier.
-    public var adInstanceId: String
+    // swiftlint:disable missing_docs
+    @objc public var networkName: String
+    @objc public var mediatorName: MediatorName
+    @objc public var placement: String?
+    @objc public var adUnitId: String
+    @objc public var adInstanceId: String
```
JZDesign: should we make these `private(set)`?

polmiro: Good call, on it!

PR #5687, opened 2025-10-21
Flush many batches paywall events

polmiro:
### Checklist

### Motivation
Noticed that only a single batch was being sent when sync occurs. This can lead to some data not being synced if enough events are being generated but the rate of sync events cannot catch up to the pace. This is unlikely given the amount of events generated is small but more future-proof in case we add more events.

### Description
With this PR up to 10 batches of events will be attempted to be sent. If an unsuccessful sync occurs in a batch the delivery it will stop.

### DIFF
--- a/Sources/Paywalls/Events/PaywallEventsManager.swift
+++ b/Sources/Paywalls/Events/PaywallEventsManager.swift
@@ -68,8 +68,9 @@ actor PaywallEventsManager: PaywallEventsManagerType {
         defer { self.flushInProgress = false }
 
         var totalFlushed = 0
+        var batchesSent = 0
 
-        while true {
+        while batchesSent < Self.maxBatchesPerFlush {
             let events = await self.store.fetch(batchSize)
 
             guard !events.isEmpty else {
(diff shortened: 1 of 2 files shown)


polmiro:
Going to add a hard limit on the maximum number of batches sent at once. Very much and edge and likely not to occur, but in case someone had a very large list of stored events already, it would prevent lots of serial roundtrips given the http requests are executed serially (please correct me if I understood wrong here).

Sources/Paywalls/Events/PaywallEventsManager.swift
```diff
@@ -59,42 +59,54 @@ actor PaywallEventsManager: PaywallEventsManagerType {
...
+                }
+            }
         }
+
+        return totalFlushed
     }
 
-    static let defaultEventFlushCount = 50
+    static let defaultEventBatchSize = 50
+    static let maxBatchesPerFlush = 10
```
facumenzella: Is this 10 just an arbitrary number?

PR #5688, opened 2025-10-21
Add size limit to PaywallEventStore to prevent unbounded growth

polmiro:
### Checklist

### Description

- Add 2048 KB max file size limit
- Check size before storing each event
- Clear oldest 50 events when limit exceeded
- Add proper logging for size limit warnings and errors

Testing:
- Add tests for size limit enforcement and cleanup behavior.
- Fix MockFileHandler to return KB instead of bytes for file size.

### DIFF
--- a/Tests/UnitTests/Paywalls/Events/PaywallEventStoreTests.swift
+++ b/Tests/UnitTests/Paywalls/Events/PaywallEventStoreTests.swift
@@ -274,6 +274,34 @@ class PaywallEventStoreTests: TestCase {
         })
     }
 
+    func testSizeLimitKeepsLatest11Events() async {
+        // Store 60 events and keep track of them
+        var allEvents: [StoredEvent] = []
+        for _ in 0..<60 {
+            let event = StoredEvent.randomImpressionEvent()
+            allEvents.append(event)
+            await self.store.store(event)
+        }
+
+        // Mock file size to be over the limit (2048 KB)
+        await self.handler.setMockedFileSizeInKB(2100)
+
+        // Store one more event, which should trigger cleanup
+        let finalEvent = StoredEvent.randomImpressionEvent()
+        allEvents.append(finalEvent)
+        await self.store.store(finalEvent)
+
+        // Reset mocked size to reflect actual file size after cleanup
+        await self.handler.setMockedFileSizeInKB(nil)
+
+        // Fetch remaining events
+        let remainingEvents = await self.store.fetch(100)
+
+        // Should have the latest 11 events (last 10 from original 60 + the final event)
+        let expectedEvents = Array(allEvents.suffix(11))
+        expect(remainingEvents) == expectedEvents
+    }
+
 }
 
 // MARK: - Extensions


polmiro:
I am actually not super versed in the use cases for Paywall events and CustomerCenter events. But my understanding is these are for user analytics so it's generally okay if we were to lose some events in a compromised situation.

The goal here was to prevent growing the file without boundary which could result in worse issues than losing the events that affect the UX e.g.: crash or an iOS space issue warning being raised. In other words, this should never happen but in case it were we fail more gracefully.

polmiro:
Tagging @joshdholtz and @vegaro out of caution since this affects CustomerCenter events and Paywall events.

dpannasch:
@polmiro do we have a way to roughly estimate how many events we expect fit in the 2048 KB limit?

I'm guessing its a large # if we're going to clear 50 to get under the limit. But to explain why it matters IMO:

If ~50 events fit in the limit, and we clear 50, we lose all signal about what this user is doing on paywalls, including their latest interactions.

If ~200 events fit in the limit, and we clear 50, we still have a lot of signal about what this user is doing on paywalls. Not a complete history, but that history is relatively less important than their latest interactions.

A different way to address my concern would be to have some kind of date based rule that never clears anything <1 week old, since the vast majority of users will not have hundreds of paywall events within 1 week, but if the 2048 KB limit we expect can hold hundreds of events than this doesn't matter at all.

polmiro:
@dpannasch thanks for checking, the rough approximate is 500 bytes per event which should give you a rough maximum of 4000 events. It's fairly arbitrary of what seemed reasonable to me without much knowledge of the amount of events a user may generate.

Unfortunately, a date based rule would not necessarily address the risk here. If a bug or miss-behavior started to generate tons of events and writing them into the file with unbounded size.

If you think we should bump we definitely can.

dpannasch:
Yeah from my perspective that's totally fine. Honestly if an individual app install has generated >4k _useful_ undelivered events I'd be stunned lol. Thx for the detail!

polmiro:
By the way, this one goes in hand with https://github.com/RevenueCat/purchases-ios/pull/5687 since if we actually had lots of events piling up at such a fast pace it's unlikely that catch up sending all of them today given that a single batch of 50 events in very specific circumstances: app initialization, app foregrounded, app backgrounded and at a specific paywall interactions. This tells me that events are very unlikely to pile up in normal circumstances today or we would already be missing data.

Tests/UnitTests/Paywalls/Events/PaywallEventStoreTests.swift
```diff
@@ -221,6 +221,59 @@ class PaywallEventStoreTests: TestCase {
...
+
+        // Store one more event, which should trigger cleanup
+        await self.store.store(.randomImpressionEvent())
+
+        // Reset mocked size to reflect actual file size after cleanup
+        await self.handler.setMockedFileSizeInKB(nil)
+
+        // Should have cleared 50 events, leaving 11 (60 - 50 + 1)
+        let events = await self.store.fetch(100)
+        expect(events).to(haveCount(11))
```
JayShortway, line 243: Might be good to test that we have the expected latest 11 events?

JayShortway (approved):
Looks good! Just a suggestion for 1 more test / assertion.

Tests/UnitTests/Paywalls/Events/PaywallEventStoreTests.swift
polmiro: Just added and rebased.

ajpallares (approved):

PR #5690, opened 2025-10-21
Add a delay hook for a purchase flow

JZDesign:
### Checklist

### Motivation

We have several customers wanting to intercept the purchase flow after the purchase button is clicked, but before the purchase sheet is presented. The desire is to complete necessary steps before the purchase is handled. Like authentication or age gating.

### Description

Creates the ability to gate and intercept when a purchase / redeem button is pressed. This should allow the consumer to log or intercept the flow and present an auth flow for example.

### DIFF
--- a/Tests/APITesters/AllAPITests/RevenueCatUISwiftAPITester/PaywallViewAPI.swift
+++ b/Tests/APITesters/AllAPITests/RevenueCatUISwiftAPITester/PaywallViewAPI.swift
@@ -24,7 +24,7 @@ struct App: View {
     private var restoreStarted: RestoreStartedHandler = { }
     private var failureHandler: PurchaseFailureHandler = { (_: NSError) in }
 
-    private var purchaseInitiated = { (_: String, _: ResumeAction) in }
+    private var purchaseInitiated = { (_: Package, _: ResumeAction) in }
 
     #if !os(watchOS)
     private var paywallTierChange: PaywallTierChangeHandler = { (_: PaywallData.Tier, _: String) in }


RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
```diff
@@ -107,6 +110,17 @@ struct PurchaseButtonComponentView: View {
...
 
+        // Check if there's a purchase interceptor
+        if let interceptor = self.purchaseInitiatedAction {
+            // Wait for the interceptor to call resume before proceeding
+            await withCheckedContinuation { continuation in
+                let productIdentifier = selectedPackage.storeProduct.productIdentifier
+                interceptor(productIdentifier) {
+                    continuation.resume()
+                }
+            }
```
ajpallares: While we're waiting for the interceptor to resume, should we probably show some loading state on the button?
This could be achieved using `PurchaseHandler`'s `actionTypeInProgress`, which is only set from the `PurchaseHandler` itself. 
Since we don't really control what an integrator would do, we should block the paywall's UI as if the purchase was happening.
In terms of implementation, I'd say we could create a new `PurchaseHandler.ActionType` called `purchaseIntent` or `pendingPurchaseContinuation`, something like that, that still makes the `PurchaseHandler`'s `actionInProgress` return `true` so that the paywall UI is "blocked" accordingly

RevenueCatUI/Templates/V2/Components/Button/ButtonComponentView.swift
```diff
@@ -142,7 +146,17 @@ struct ButtonComponentView: View {
...
-    private func openCodeRedemptionSheet() {
+    private func openCodeRedemptionSheet() async {
+        // Check if there's an offer code redemption interceptor
+        if let interceptor = self.offerCodeRedemptionInitiatedAction {
+            // Wait for the interceptor to call resume before proceeding
+            await withCheckedContinuation { continuation in
+                interceptor(resume: {
+                    continuation.resume()
+                })
+            }
```
ajpallares, line 159: Same comment here about having `actionTypeInProgress` return true to block the UI while the continuation happens

RevenueCatUI/Templates/V2/Components/Packages/PurchaseButton/PurchaseButtonComponentView.swift
JZDesign: This is an interesting topic… I agree that updating the UI is appropriate, but the purchase handler isn't handling anything yet so I'm not sure I agree with it owning that state. For now I updated the View to own the data so we are handling the case correctly while I stew on the purchase handler owning this information.

It feels odd to put a new enum case in there and a function to update the `actionInProgress` to a "waiting for something unrelated" state

JZDesign: Because, it's really not a purchase action in progress right? It's a consuming app action in progress

joshdholtz: Ahhh, so I think I'm following here!

1) We all agree the button should have an activity indicator?
2) Right now its done through `PurchaseHandler` enum of `actionInProgress` which... is a weird since that isn't what this is? This is an action that is happening on the purchase button so there is purchase intent so it kind of _is_ a purchase? 🤔 

I would vote to use `actionInProgress` since its the behavior 🤷‍♂️

PR #5691, opened 2025-10-21
Track paywall source for Customer Center purchases

facumenzella:
### Checklist

### Motivation

We need to track revenue generated from Customer Center. When a paywall is presented from Customer Center, the backend should know the origin so analytics can classify the event correctly.

### Description

* Introduced `PaywallSource`, a `@_spi(Internal)` String-backed enum in the core `RevenueCat` module (currently with a single case: `.customerCenter`).
* Added a SwiftUI environment key/modifier (`.paywallSource(...)`) in `RevenueCatUI` so the source propagates through the view hierarchy to paywall event creation (V1 + V2).
* `NoSubscriptionsCardView` in Customer Center now sets `.paywallSource(.customerCenter)` on its paywall presentation.
* Updated `PaywallEvent.Data` with an `@_spi(Internal)` `source: PaywallSource?` property. Public initializers without `source` are deprecated; new `@_spi(Internal)` initializers accept the typed parameter.
* `EventsRequest.PaywallEvent` and `PostReceiptDataOperation.Paywall` include the optional `source` field, encoded as `"customer_center"` when present and omitted when `nil`.

> **Note:** Tracking of non-paywall purchases originating from Customer Center (e.g. promotional offer redemptions) is handled separately in https://github.com/RevenueCat/purchases-ios/pull/6376.

---

> [!NOTE]
> **Medium Risk**
> Touches paywall analytics and `POST /receipts` payload encoding by adding a new optional field, so regressions would affect tracking/receipt submissions even though the change is additive and gated behind `nil` defaults.
> 
> **Overview**
> **Adds paywall origin tracking.** Introduces internal `PaywallSource` plus a SwiftUI environment key/modifier (`.paywallSource(...)`) so paywall presentations can tag events with where they came from.
> 
> **Plumbs source through analytics + receipts.** `PaywallEvent.Data` gains an internal optional `source` (with new SPI initializers and deprecated public ones), and both `FeatureEventsRequest.PaywallEvent` and `PostReceiptDataOperation.Paywall` now encode `source` when present.
> 
> **Customer Center now sets the source.** `NoSubscriptionsCardView` tags its presented `PaywallView` as `.customerCenter`, and tests/snapshots are updated to cover the new field and ensure propagation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 2d5b8fc00c381e9316f2ee46e0eaba1ca965474b. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Tests/UnitTests/Events/FeatureEventStoreTests.swift
+++ b/Tests/UnitTests/Events/FeatureEventStoreTests.swift
@@ -356,7 +356,8 @@ extension PaywallEvent.Data {
             sessionID: .init(),
             displayMode: PaywallViewMode.allCases.randomElement()!,
             localeIdentifier: "es_ES",
-            darkMode: Bool.random()
+            darkMode: Bool.random(),
+            source: nil
         )
     }
 


RevenueCatUI/Data/PaywallSourceEnvironment.swift
```diff
@@ -0,0 +1,71 @@
...
+    }
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+extension View {
+
+    /// Associates the provided paywall source with this view hierarchy so downstream paywall presentations can read it.
+    @_spi(Internal) public func paywallSource(_ source: PaywallSource?) -> some View {
+        self.environment(\.paywallSource, source)
+    }
```
facumenzella: This felt cleaner than adding a parameter to the view

RevenueCatUI/Purchasing/PurchaseHandler.swift
```diff
@@ -380,6 +386,15 @@ extension PurchaseHandler {
...
     }
 
+    @MainActor
+    func updatePaywallSource(_ source: PaywallSource?) {
+        self.paywallSource = source
+
+        if let source {
+            self.eventData?.source = source.rawValue
+        }
+    }
```
facumenzella: I did this to avoid modifying `PaywallEvent`. Now I am thinking that maybe we should call this `offeringSource`, rather than `paywallSource`

Sources/Purchasing/Purchases/Purchases.swift
```diff
@@ -1399,9 +1388,11 @@ public extension Purchases {
...
     /// Used by `RevenueCatUI` to keep track of ``PaywallEvent``s.
-    func track(paywallEvent: PaywallEvent) async {
-        self.purchasesOrchestrator.track(paywallEvent: paywallEvent)
-        await self.paywallEventsManager?.track(featureEvent: paywallEvent)
+    /// - Parameter source: Optional identifier describing where the paywall was presented from.
+    func track(paywallEvent: PaywallEvent, source: String? = nil) async {
+        let eventWithSource = paywallEvent.withSource(source)
+        self.purchasesOrchestrator.track(paywallEvent: eventWithSource)
+        await self.paywallEventsManager?.track(featureEvent: eventWithSource)
     }
```
facumenzella: This makes me think if we should just add source to any `PaywallEvent` instead of mutating the event here

Examples/rc-maestro/rc-maestro/Resources/StoreKit/StoreKitConfiguration.storekit
```diff
@@ -46,56 +46,53 @@
     }
   ],
   "settings" : {
+    "_askToBuyEnabled" : false,
```
facumenzella: I need to rever this file

Tests/UnitTests/Networking/Backend/__Snapshots__/BackendPostReceiptDataTests/iOS12-testPostsReceiptDataWithPresentedOfferingContext.1.json
JZDesign, line 1: I would revert the last line change of the json files. When they are opened with Xcode, it leaves a blank line at the end. This can cause issues with tests (if these are ever compared upon load)

JZDesign: I always use VSCode for this

RevenueCatUI/Data/PaywallSource.swift
```diff
@@ -0,0 +1,35 @@
...
+    public init(stringLiteral value: StringLiteralType) {
+        self.init(rawValue: value)
+    }
+
+}
+
+@_spi(Internal) public extension PaywallSource {
+
+    /// Source identifier used when the paywall originated from Customer Center.
+    static let customerCenter: PaywallSource = "presented_from_customer_center"
```
tonidero: I think this could be simplified?
```suggestion
    static let customerCenter: PaywallSource = "customer_center"
```

RevenueCatUI/Data/PaywallSourceEnvironment.swift
```diff
@@ -0,0 +1,43 @@
...
+        get { self[PaywallSourceKey.self] }
+        set { self[PaywallSourceKey.self] = newValue }
+    }
+}
+
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+extension View {
+
+    /// Associates the provided paywall source with this view hierarchy so downstream paywall presentations can read it.
+    @_spi(Internal) public func paywallSource(_ source: PaywallSource?) -> some View {
```
tonidero: Hmm do we need to make this public even with `@_spi(Internal)`? I think since Customer center is part of the same library, it could potentially set this internally, unless I'm missing something. If we want to add other sources that need to be called from the outside, then we can think about making it public

RevenueCatUI/Purchasing/PaywallPurchasesType.swift
```diff
@@ -42,7 +42,7 @@ protocol PaywallPurchasesType: Sendable {
     func customerInfo() async throws -> CustomerInfo
 
     @Sendable
-    func track(paywallEvent: PaywallEvent) async
+    func track(paywallEvent: PaywallEvent, source: PaywallSource?) async
```
tonidero: Should we instead add the `PaywallSource` to the `PaywallEvent`? Seems like it would be nicer... But not sure if I'm missing a reason to keep it separate.

RevenueCatUI/Data/PaywallSourceEnvironment.swift
tonidero: Hmm not sure... I feel these environment properties are harder to discover and test.... But I guess not a strong opinion since this should be kept internal

Sources/Purchasing/Purchases/Purchases.swift
tonidero: Right mentioned it in another comment above, but I think that would be cleaner indeed.

Sources/Networking/Operations/PostReceiptDataOperation.swift
```diff
@@ -180,6 +184,7 @@ extension PostReceiptDataOperation.PostData {
             productData: productData,
             presentedOfferingIdentifier: data.presentedOfferingContext?.offeringIdentifier,
             presentedPlacementIdentifier: data.presentedOfferingContext?.placementIdentifier,
+            presentedOfferingSource: data.presentedPaywall?.data.source,
```
tonidero: I'm confused... why do we need `presentedOfferingSource` considering we're already sending a `source` inside the `paywall` parameter?

tonidero: Not sure if this is meant to reflect purchases performed through customer center but not the paywall? If so, I think `data.presentedPaywall?.data.source` would be wrong, since there wouldn't be a paywall?

tonidero (commented):
Some comments and questions 👍

Sources/Networking/Operations/PostReceiptDataOperation.swift
tonidero: And also, in this case, we might want to think of a better name... `presentedOfferingSource` looks like it should be part of `PresentedOfferingContext` but it isn't right now.

RevenueCatUI/Data/PaywallSourceEnvironment.swift
ajpallares: Agreed. Removed in https://github.com/RevenueCat/purchases-ios/pull/5691/commits/61a49046bade789310df7c68509c78fc3718be78

RevenueCatUI/Data/PaywallSource.swift
ajpallares: Agreed. Changed in https://github.com/RevenueCat/purchases-ios/pull/5691/commits/eaf2a30b4d74e004c2ead27483e91c2205b02d42 to `customer_center`, now matching https://github.com/RevenueCat/khepri/pull/15363

Examples/rc-maestro/rc-maestro/Resources/StoreKit/StoreKitConfiguration.storekit
ajpallares: Done in https://github.com/RevenueCat/purchases-ios/pull/5691/commits/7bb804f627fea0ef8e923cbe6e437a362b1bb02e

Sources/Paywalls/PaywallSource.swift
```diff
@@ -0,0 +1,22 @@
...
+//      https://opensource.org/licenses/MIT
+//
+//  PaywallSource.swift
+//
+//  Created by RevenueCat on 2/26/25.
+
+import Foundation
+
+/// Identifies where a paywall was presented from so that backend analytics can classify the event.
+@_spi(Internal) public enum PaywallSource: String, Sendable, Codable {
```
rickvdl, line 17: Minor nit: I think this can be excluded on platforms that don't support paywalls

ajpallares: Nice suggestion! However, since `PaywallEvent` depends on the new `PaywallSource` and that's not excluded, I feel like it would mean overcomplicating things unnecessarily... WDYT?

rickvdl: Also a valid points, fine to leave as is :)

PR #5695, opened 2025-10-22
fix: Resume request refund on Promo decline or fail

facumenzella:
### Motivation
We've got a report of refunds not working when passing through a promo offer.

### Description
- Resume path when promo offer fails or is declined
- Fix padding for ProgressView


RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
```diff
@@ -135,8 +135,25 @@ class BaseManageSubscriptionViewModel: ObservableObject {
...
+        defer {
+            self.loadingPath = nil
+        }
+
+        if let path = self.loadingPath {
+            // if its decline, execute action
+            switch action {
+            case .successfullyRedeemedPromotionalOffer:
+                return
+            case .declinePromotionalOffer, .promotionalCodeRedemptionFailed:
```
vegaro: I am not super convinced we should execute the path action if `promotionalCodeRedemptionFailed`. What happens if the user cancels the purchase in the Apple dialog? So if they accept the offer but then cancel it when the Apple UI to purchase appears. Is that an error?

In Android we are only executing the path action (the refund) if they explicitly dismiss. If they cancel the purchase in the Apple dialog, or if there's an error, we don't continue and we go back to main.

facumenzella: You're actually right.

> What happens if the user cancels the purchase in the Apple dialog?

With the current code nothing, but we should open the cancel dialog anyway. We're actually not handling it correctly.

> In Android we are only executing the path action (the refund) if they explicitly dismiss. If they cancel the purchase in the Apple dialog, or if there's an error, we don't continue and we go back to main.

I've tweaked the implementation https://github.com/RevenueCat/purchases-ios/pull/5695/commits/c45d6b90eb57724a585fb30c812d5607e028d3b7 using an existing helper 👍

RevenueCatUI/CustomerCenter/ViewModels/PromotionalOfferViewModel.swift
```diff
@@ -60,18 +60,28 @@ final class PromotionalOfferViewModel: ObservableObject {
...
+            } else {
+                let transactionId = result.transaction?.transactionIdentifier ?? "-"
+                Logger.debug(Strings.promo_offer_purchase_succeeded(productId, offerId, transactionId))
+                self.actionWrapper.handleAction(.promotionalOfferSuccess)
+                self.onPromotionalOfferPurchaseFlowComplete?(.successfullyRedeemedPromotionalOffer(result))
+            }
         } catch {
-            // swiftlint:disable:next todo
-            // TODO: Log error message
+            Logger.error(Strings.promo_offer_purchase_failed(productId, offerId, error))
```
vegaro, line 84: If the iOS payment sheet is dismissed, it goes here with a cancelation?

RevenueCatUI/CustomerCenter/ViewModels/BaseManageSubscriptionViewModel.swift
vegaro: I was just retesting this, and I think if they:
- Dismiss the promo offer, it should go to refund
- If they accept, but then cancel the payment dialog, it should stay in the promo offer
- If they say "No thanks", it should go to the refund.

I will make the changes. I believe in Android we need to fix the cancelled dialog case because it's going back to the main screen I believe, instead of staying in the promo offer screen.

RevenueCatUI/CustomerCenter/ViewModels/PromotionalOfferViewModel.swift
vegaro: tested it and it goes to the else, not the catch

vegaro:
I am going to merge this and update the behavior on cancellation of the payment dialog in another PR

PR #5697, opened 2025-10-22
fix: Use tint color for See All Purchases

facumenzella:
### Motivation
We've noticed the tint color wasn't properly applied

### Description
- Use tint color in see all button
- Extract tint logic to extension

### DIFF
--- a/RevenueCatUI/CustomerCenter/Views/FeedbackSurveyView.swift
+++ b/RevenueCatUI/CustomerCenter/Views/FeedbackSurveyView.swift
@@ -158,7 +158,7 @@ struct FeedbackSurveyButtonsView: View {
             }
             .disabled(self.loadingOption != nil)
         }
-        applyIfLet(appearance.tintColor(colorScheme: colorScheme), apply: { $0.tint($1)})
+        .applyIfLet(appearance.tintColor(colorScheme: colorScheme), apply: { $0.tint($1)})
     }
 }
 


RevenueCatUI/CustomerCenter/Views/ActiveSubscriptionButtonsView.swift
```diff
@@ -60,18 +60,14 @@ struct ActiveSubscriptionButtonsView: View {
                 }
             }
         }
-        .applyIf(tintColor != nil, apply: { $0.tint(tintColor) })
+        .applyIfLet(Color.from(colorInformation: appearance.accentColor, for: self.colorScheme), apply: { $0.tint($1)})
```
vegaro: should this be using the `tintColor` extension?

facumenzella: Yes

RevenueCatUI/CustomerCenter/Views/NoSubscriptionsCardView.swift
```diff
@@ -143,19 +143,15 @@ private struct BuySubscriptionButtonStyle: ButtonStyle {
             .padding(.vertical, 14)
             .background(
                 configuration.isPressed
-                ? tintColor?.opacity(0.8)
-                : tintColor
+                ? Color.from(colorInformation: appearance.accentColor, for: self.colorScheme)?.opacity(0.8)
+                : Color.from(colorInformation: appearance.accentColor, for: self.colorScheme)
```
vegaro: I think this can use the extension as well

PR #5700, opened 2025-10-23
Add infrastructure to enable richer server-down integration tests

ajpallares:
Before this PR, our integration tests that required a scenario where the backend failed would mock a 500 response locally (no real request happened). In addition, tests could only toggle this behavior On (all requests would be mocked with a 500 response) or Off (no requests mocked; all would go through to the real server).

This PR adds some improvements and flexibility to more closely replicate real server-down scenarios:
* Uses a specific endpoint that always responds with a 500 (simulating a real-life server down situation). 
* Allows tests to fine-tune what requests should be made to fail (instead of all or none).

### DIFF
--- a/Tests/UnitTests/Misc/TestCase.swift
+++ b/Tests/UnitTests/Misc/TestCase.swift
@@ -81,9 +81,15 @@ private enum SnapshotTests {
 
 extension ForceServerErrorStrategy {
 
-    /// Fakes server errors in all requests, including requests made to the fallback API hosts.
+    /// Forces server error in all requests, including requests made to the fallback API hosts.
     static let allServersDown: ForceServerErrorStrategy = .init { _ in
         return true // All requests fail
+    }
+
+    /// Forces server error in all requests except those made to the fallback API hosts.
+    static let failExceptFallbackUrls: ForceServerErrorStrategy = .init { (request: HTTPClient.Request) in
+        let isRequestToFallbackUrl = request.fallbackHostIndex != nil
+        return !isRequestToFallbackUrl
     }
 
 }


Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -544,10 +527,21 @@ private extension HTTPClient {
...
+        #endif
+
         let requestStartTime = self.dateProvider.now()
 
         // swiftlint:disable:next redundant_void_return
-        let task = self.session.dataTask(with: urlRequest) { (data, urlResponse, error) -> Void in
+        let task = self.session.dataTask(with: finalURLRequest) { (data, urlResponse, error) -> Void in
             self.handle(urlResponse: urlResponse,
                         request: request,
                         urlRequest: urlRequest,
```
ajpallares, line 547: Note that I'm intentionally passing here the `urlRequest` to `self.handle(urlResponse:,request:,urlRequest,...)` and not passing `finalURLRequest`.

My reasoning is: we want to make calling the `force-server-failure` endpoint as transparent as possible in terms of SDK implementation. This way, even though we would be calling the `force-server-failure` endpoint, the SDK would still behave exactly the same way as if a call to `urlRequest` happened. So that the response and diagnostics are treated exactly as if it was the call to `urlRequest` that failed.

Please, lmk what you think!

tonidero: I think that makes sense yeah! And for endpoints that support fallbacks, I was currently planning to let it actually succeed, so we should handle it the same way I think.

ajpallares: > And for endpoints that support fallbacks, I was currently planning to let it actually succeed, so we should handle it the same way I think

What do you mean exactly? 

My current implementation allows deciding what to do per each request. So we could also make fallback endpoints fail or succeed depending on the needs of the test.

tonidero: Good point, yeah this would indeed provide more flexibility to the caller... I was trying to reuse the existing logic in Android, but I will change to try to use the same logic instead 👍

ajpallares:
@RCGitBot please test

Tests/UnitTests/Misc/TestCase.swift
```diff
@@ -73,3 +78,18 @@ private enum SnapshotTests {
...
+    /// Forces server error in all requests, including requests made to the fallback API hosts.
+    static let allServersDown: ForceServerErrorStrategy = .init { _ in
+        return true // All requests fail
+    }
+
+    /// Forces server error in all requests except those made to the fallback API hosts.
+    static let failExceptFallbackUrls: ForceServerErrorStrategy = .init { (request: HTTPClient.Request) in
+        let isRequestToFallbackUrl = request.fallbackHostIndex != nil
+        return !isRequestToFallbackUrl
+    }
```
ajpallares, line 93: This is unused for now, but I plan to use it in a follow-up PR

tonidero (approved):
Amazing! Really well done!

PR #5701, opened 2025-10-23
Add fallback API integration tests

ajpallares:
### Description

Adds integration tests to verify the behavior when the SDK needs to fall back to the fallback server (that is when both the main url and load shedder are down).


Sources/OfflineEntitlements/ProductEntitlementMapping.swift
```diff
@@ -16,14 +16,14 @@ import Foundation
...
-    var entitlementsByProduct: [String: Set<String>]
+    var entitlementsByProduct: [String: [String]]
 
 }
 
 extension ProductEntitlementMapping {
 
     /// - Returns: entitlement identifiers associated to the given product identifier
-    func entitlements(for productIdentifier: String) -> Set<String> {
+    func entitlements(for productIdentifier: String) -> [String] {
```
ajpallares, line 26: This was a `Set` but didn't really need to. I changed it because:
1. It's an internal API and can be changed without braking the public API
2. An array is deterministic for unit tests
3. Outside tests, it's only used [here](https://github.com/RevenueCat/purchases-ios/blob/46bfe3cf8640f6e0e22669ce76a16bfb3d683560/Sources/OfflineEntitlements/CustomerInfo%2BOfflineEntitlements.swift#L82-L89) and is still looped over the whole Set, so there's no changed effect when looping through the array  https://github.com/RevenueCat/purchases-ios/blob/46bfe3cf8640f6e0e22669ce76a16bfb3d683560/Sources/OfflineEntitlements/CustomerInfo%2BOfflineEntitlements.swift#L82-L89
4. With an array we can deterministically add Snapshot tests for product entitlement mapping responses

Tests/BackendIntegrationTests/FallbackURLBackendIntegrationTests.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    func testCanGetOfferingsFromFallbackURL() async throws {
+        let receivedOfferings = try await self.purchases.offerings()
+
+        expect(receivedOfferings.all).toNot(beEmpty())
+        assertSnapshot(matching: receivedOfferings.response, as: .formattedJson)
+    }
+
+    func testCanGetProductEntitlementMappingFromFallbackURL() async throws {
+        let productEntitlementMapping = try await self.purchases.productEntitlementMapping()
```
tonidero, line 61: Would it be worth checking they come from the expected location? (fallback in this case)

ajpallares: That's right. But, for now, we don't have a way to check that in ProductEntitlementMapping. That's something on the roadmap, for sure! Thanks for the suggestion!

ajpallares:
@RCGitBot please test

PR #5703, opened 2025-10-24
Allowing HTTP requests to be retried with fallback host after netwok / DNS errors as well as 5xx errors

rickvdl:
### Checklist

### Motivation
Previously we required a `HTTPURLResponse` with a valid HTTP status code in the 5xx range in order to switch to the fallback URL. This however will not help us in scenarios where the server will timeout, drop the connection or fail to respond in another way.

### Description
I've updated the logic to look at the `NetworkError` instead of requiring a valid response. If the `NetworkError` is a `errorResponse` we use the logic as before on the received response. If the `NetworkError` is a dns issue, any other network error or the response is completely invalid (either no response or the response type does not match `HTTURLResponse` we now also use the fallback host for the retry operation.

### DIFF
--- a/Tests/UnitTests/Networking/HTTPClientTests.swift
+++ b/Tests/UnitTests/Networking/HTTPClientTests.swift
@@ -2764,7 +2764,7 @@ extension HTTPClientTests {
         let client = self.createClient(self.systemInfo, operationDispatcher: mockOperationDispatcher)
 
         let request = buildEmptyRequest(isRetryable: true, hasFallbackHosts: true)
-        
+
         let didRetry = client.retryRequestWithNextFallbackHostIfNeeded(
             request: request,
             error: .serverDown()
@@ -2773,13 +2773,13 @@ extension HTTPClientTests {
         expect(didRetry).to(beTrue())
         expect(mockOperationDispatcher.invokedDispatchOnWorkerThreadWithTimeInterval).to(beFalse())
     }
-    
+
     func testRetriesWithNextFallbackHostImmediatelyForUnexpectedResponse() throws {
         let mockOperationDispatcher = MockOperationDispatcher()
         let client = self.createClient(self.systemInfo, operationDispatcher: mockOperationDispatcher)
 
         let request = buildEmptyRequest(isRetryable: true, hasFallbackHosts: true)
-        
+
         let didRetry = client.retryRequestWithNextFallbackHostIfNeeded(
             request: request,
             error: .unexpectedResponse(nil)
@@ -2791,7 +2791,7 @@ extension HTTPClientTests {
 
     func testIncrementsHostIndexOnRetry() throws {
         let request = buildEmptyRequest(isRetryable: true, hasFallbackHosts: true)
-        
+
         let didRetry = self.client.retryRequestWithNextFallbackHostIfNeeded(
             request: request,
             error: .serverDown()
@@ -2803,7 +2803,7 @@ extension HTTPClientTests {
 
     func testDoesNotIncrementRetryCountOnHostRetry() throws {
         let request = buildEmptyRequest(isRetryable: true, hasFallbackHosts: true)
-        
+
         let didRetry = self.client.retryRequestWithNextFallbackHostIfNeeded(
             request: request,
             error: NetworkError.serverDown()
@@ -2819,14 +2819,14 @@ extension HTTPClientTests {
         let didRetry = self.client.retryRequestWithNextFallbackHostIfNeeded(
             request: request,
             error: NetworkError.errorResponse(
-                ErrorResponse.init(
+                ErrorResponse(
                     code: .unknownBackendError,
                     originalCode: BackendErrorCode.unknownBackendError.rawValue
                 ),
                 HTTPStatusCode.tooManyRequests
             )
         )
-        
+
         expect(didRetry).to(beFalse())
     }
 


rickvdl:
@RCGitBot please test

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -765,6 +765,19 @@ extension HTTPClient {
 
 // MARK: - Extensions
 
+fileprivate extension NetworkError {
+    var isAllowedToRetryWithFallbackHost: Bool {
+        switch self {
+        case .decoding, .unableToCreateRequest, .signatureVerificationFailed:
+            return false
+        case .dnsError, .networkError, .unexpectedResponse:
```
rickvdl, line 774: imo the `unexpectedResponse` is a bit vague, but looking at it's uses it's only really used when either a `nil` response is received or the response is not of type `HTTPURLResponse` (which should never happen AFAIK)

rickvdl, line 774: The `.networkError` case here AFAIK covers everything which prevented us from getting _any_ response. This might be too broad, I will check to see if we should be more granular

ajpallares: We should make sure we do the same here as in Android (see https://github.com/RevenueCat/purchases-android/pull/2754). 
We can discuss on opening this up more, but I think we should have the same logic as much as possible.

rickvdl: @ajpallares Agree! 

@JayShortway I believe it is related to this code in your PR https://github.com/RevenueCat/purchases-android/pull/2754/files#diff-1eb7acca809981914bd4da4d178442b7a65747a3c3f9aa0f0c4478c7251f78e0R150

It looks like any network error is caught and retried using the fallback URL, is that correct? If so I think that matches the behavior above. This would for example also mean that this logic is triggered for things like having no network connection. I got claude to create an overview of possible error codes we can expect here that we should not retry (according to claude). Wdyt?

```
        // Errors that should NOT retry with fallback
        case NSURLErrorNotConnectedToInternet,        // No internet = no host will work
             NSURLErrorCancelled,                      // User/app cancelled
             NSURLErrorBadURL,                         // URL issue, not host issue
             NSURLErrorUnsupportedURL,                 // URL scheme issue
             NSURLErrorUserCancelledAuthentication,    // User action
             NSURLErrorInternationalRoamingOff,        // Device restrictions
             NSURLErrorCallIsActive,                   // Device restrictions
             NSURLErrorDataNotAllowed,                 // Network restrictions
             NSURLErrorAppTransportSecurityRequiresSecureConnection, // ATS policy
             NSURLErrorClientCertificateRequired:      // Client config issue
            return false

        // File/download errors - should NOT retry with fallback
        case NSURLErrorFileDoesNotExist,
             NSURLErrorFileIsDirectory,
             NSURLErrorNoPermissionsToReadFile,
             NSURLErrorDataLengthExceedsMaximum:
            return false

        // Cache/loading policy errors - should NOT retry
        case NSURLErrorCannotLoadFromNetwork:
            return false

        // Download-specific errors - should NOT retry with fallback
        case NSURLErrorCannotCreateFile,
             NSURLErrorCannotOpenFile,
             NSURLErrorCannotCloseFile,
             NSURLErrorCannotWriteToFile,
             NSURLErrorCannotRemoveFile,
             NSURLErrorCannotMoveFile,
             NSURLErrorDownloadDecodingFailedMidStream,
             NSURLErrorDownloadDecodingFailedToComplete:
            return false

        // Background session errors - should NOT retry with fallback
        case NSURLErrorBackgroundSessionRequiresSharedContainer,
             NSURLErrorBackgroundSessionInUseByAnotherProcess,
             NSURLErrorBackgroundSessionWasDisconnected:
            return false
```

JayShortway: These are all good points. I think that retrying for _any_ error is already better than retrying for _none_, so I'd be fine with merging that first and then looking at more granular retry logic.

Couple rambling thoughts on being granular:
- Regarding consistency: in our current Android implementation we can (easily) get a little more granular on the network errors, e.g. DNS, SSL etc. But more device level errors such as `NSURLErrorInternationalRoamingOff` need to be checked using Android framework APIs _before_ making the request. It might make sense to not even make the main request in that case though.
- Testing: it might be challenging to explicitly test this granular behavior. For instance, we'd need to fake a DNS issue, SSL handshake issue, "Roaming off" etc.

All that to say: might be good to keep it simple first, unless we have a very good reason to believe "simple" is worse?

rickvdl: Agree, I think that makes sense as a starter, I don't see any immediate downsides to this at this moment.

rickvdl:
@RCGitBot please test

Sources/Networking/HTTPClient/HTTPClient.swift
```diff
@@ -765,6 +766,19 @@ extension HTTPClient {
...
+            return true
+        case let .errorResponse(_, statusCode, _):
+            return HTTPStatusCode(rawValue: statusCode.rawValue).isServerError
+        }
+    }
+}
+
```
JZDesign, line 781: # 🤌

JZDesign (approved):
Great addition. :shipit:

PR #5704, opened 2025-10-24
Video Background 

JZDesign:
Adds Video Background to iOS stacks. Also adds prewarming of low res background images.

see git commit history in this branch: jzdesign/MON-1374-invert I accidentally forgot to switch branches.

Lottza vids :

Simulator Screen Recording - iPhone 16e - 2025-10-24 at 10 15 24

### DIFF
--- a/RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentView.swift
@@ -821,7 +821,7 @@ extension CarouselComponentViewModel {
             )
         }
 
-        try self.init(
+        self.init(
             localizationProvider: localizationProvider,
             uiConfigProvider: .init(uiConfig: PreviewUIConfig.make()),
             component: component,
--- a/RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentViewModel.swift
+++ b/RevenueCatUI/Templates/V2/Components/Carousel/CarouselComponentViewModel.swift
@@ -34,13 +34,13 @@ class CarouselComponentViewModel {
         uiConfigProvider: UIConfigProvider,
         component: PaywallComponent.CarouselComponent,
         pageStackViewModels: [StackComponentViewModel]
-    ) throws {
+    ) {
         self.localizationProvider = localizationProvider
         self.uiConfigProvider = uiConfigProvider
         self.component = component
         self.pageStackViewModels = pageStackViewModels
 
-        self.presentedOverrides = try self.component.overrides?.toPresentedOverrides { $0 }
+        self.presentedOverrides = self.component.overrides?.toPresentedOverrides { $0 }
     }
 
     @ViewBuilder
--- a/RevenueCatUI/Templates/V2/Components/Icon/IconComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Icon/IconComponentView.swift
@@ -112,8 +112,7 @@ struct IconComponentView_Previews: PreviewProvider {
         // Default
         VStack {
             IconComponentView(
-                // swiftlint:disable:next force_try
-                viewModel: try! .init(
+                viewModel: .init(
                     localizationProvider: .init(
                         locale: Locale.current,
                         localizedStrings: [:]
@@ -144,8 +143,7 @@ struct IconComponentView_Previews: PreviewProvider {
         // Default - Background
         VStack {
             IconComponentView(
-                // swiftlint:disable:next force_try
-                viewModel: try! .init(
+                viewModel: .init(
                     localizationProvider: .init(
                         locale: Locale.current,
                         localizedStrings: [:]
--- a/RevenueCatUI/Templates/V2/Components/Icon/IconComponentViewModel.swift
+++ b/RevenueCatUI/Templates/V2/Components/Icon/IconComponentViewModel.swift
@@ -32,12 +32,12 @@ class IconComponentViewModel {
         localizationProvider: LocalizationProvider,
         uiConfigProvider: UIConfigProvider,
         component: PaywallComponent.IconComponent
-    ) throws {
+    ) {
         self.localizationProvider = localizationProvider
         self.uiConfigProvider = uiConfigProvider
         self.component = component
 
-        self.presentedOverrides = try self.component.overrides?.toPresentedOverrides { $0 }
+        self.presentedOverrides = self.component.overrides?.toPresentedOverrides { $0 }
     }
 
     @ViewBuilder
--- a/RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
+++ b/RevenueCatUI/Templates/V2/Components/Stack/StackComponentView.swift
@@ -791,7 +791,7 @@ extension StackComponentViewModel {
             )
         }
 
-        try self.init(
+        self.init(
             component: component,
             viewModels: viewModels,
             badgeViewModels: badgeViewModels ?? [],
--- a/RevenueCatUI/Templates/V2/Components/Stack/StackComponentViewModel.swift
+++ b/RevenueCatUI/Templates/V2/Components/Stack/StackComponentViewModel.swift
@@ -35,17 +35,17 @@ class StackComponentViewModel {
         badgeViewModels: [PaywallComponentViewModel],
         shouldApplySafeAreaInset: Bool = false,
         uiConfigProvider: UIConfigProvider
-    ) throws {
+    ) {
         self.component = component
         self.viewModels = viewModels
         self.uiConfigProvider = uiConfigProvider
         self.badgeViewModels = badgeViewModels
         self.shouldApplySafeAreaInset = shouldApplySafeAreaInset
-        self.presentedOverrides = try self.component.overrides?.toPresentedOverrides { $0 }
+        self.presentedOverrides = self.component.overrides?.toPresentedOverrides { $0 }
     }
 
-    func copy(withViewModels newViewModels: [PaywallComponentViewModel]) throws -> StackComponentViewModel {
-        return try StackComponentViewModel(
+    func copy(withViewModels newViewModels: [PaywallComponentViewModel]) -> StackComponentViewModel {
+        return StackComponentViewModel(
             component: self.component,
             viewModels: newViewModels,
             badgeViewModels: self.badgeViewModels,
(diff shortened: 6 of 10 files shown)


RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
```diff
@@ -122,25 +144,17 @@ extension DisplayableColorScheme {
...
-        switch self {
-        case .color(let value):
-            return .color(value)
-        case let .image(value, fitMode, color):
-            return .image(value, fitMode, color)
-        }
-    }
-
-}
-
```
JZDesign, line 143: This is pointless

JZDesign:
@RCGitBot please test

Sources/Paywalls/Components/Common/Background.swift
```diff
@@ -69,9 +91,13 @@ public extension PaywallComponent {
...
+            case video
 
         }
 
     }
 
 }
+
+public typealias Loop = Bool
+public typealias MuteAudio = Bool
```
joshdholtz: We probably don't want these to be public? 🤔  Can these be spi-ed if we need these in `RevenueCatUI`?

JZDesign: I'll nest it in PaywallComponent.Background. Otherwise I need to make those SPI Internal too.

Screenshot 2025-10-24 at 1 42 38 PM

joshdholtz:
@RCGitBot please test

JZDesign:
@RCGitBot please test

JZDesign:
@RCGitBot please test

Sources/Paywalls/Components/Common/Background.swift
```diff
@@ -35,6 +36,16 @@ public extension PaywallComponent {
...
+            case let .video(videoInfo, imageInfo, loop, mute, fitMode, colorScheme):
+                try container.encode(BackgroundType.video.rawValue, forKey: .type)
+                try container.encode(videoInfo, forKey: .value)
+                try container.encode(imageInfo, forKey: .fallbackImage)
+                try container.encode(loop, forKey: .loop)
+                try container.encode(mute, forKey: .muteAudio)
+                try container.encode(fitMode, forKey: .fitMode)
+                if let colorScheme {
+                    try container.encode(colorScheme, forKey: .colorOverlay)
+                }
```
ajpallares: NABD but isn't this the same as

```suggestion
                try container.encodeIfPresent(colorScheme, forKey: .colorOverlay)
```

RevenueCatUI/Templates/V2/ViewHelpers/DisplayableColor.swift
```diff
@@ -20,12 +20,31 @@ import SwiftUI
...
+        uiConfigProvider: UIConfigProvider,
+        localizationProvider: LocalizationProvider? = nil
+    ) -> BackgroundStyle {
         switch self {
         case .color(let color):
             return .color(color.asDisplayable(uiConfigProvider: uiConfigProvider))
         case .image(let image, let fitMode, let colorScheme):
             return .image(image, fitMode, colorScheme?.asDisplayable(uiConfigProvider: uiConfigProvider))
+        case let .video(video, image, loop, mute, fitMode, colorScheme):
+            if let viewModel = try? VideoComponentViewModel(
```
ajpallares: I'm just checked why this `init` could throw, and I think we can get rid of this

In the end, it comes down to this method marked as `throws`, but it never throws any error:
https://github.com/RevenueCat/purchases-ios/blob/72cf38a0fec0b0d47ef04bedff2deed17705a316/RevenueCatUI/Templates/V2/Components/Video/VideoComponentViewModel.swift#L130-L137

Removing that, we could get rid of the `try?` and the `else`.

Note that changing this would probably require making this method a `rethrows` one:
https://github.com/RevenueCat/purchases-ios/blob/72cf38a0fec0b0d47ef04bedff2deed17705a316/RevenueCatUI/Templates/V2/ViewModelHelpers/PresentedPartials.swift#L131-L136

ajpallares (commented):
Some initial comments. Will keep taking a look later

RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
ajpallares: Right! Nice cleanup 💅 
It was even being mapped into a nullable, so even losing its non-nullable aspect 😅

RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
```diff
@@ -122,25 +144,17 @@ extension DisplayableColorScheme {
 @available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
 extension View {
 
-    func backgroundStyle(_ backgroundStyle: BackgroundStyle?, alignment: Alignment = .center) -> some View {
+    func backgroundStyle(
+        _ backgroundStyle: BackgroundStyle?,
+        alignment: Alignment = .center,
+        uiConfigProvider: UIConfigProvider? = nil
```
ajpallares: The `uiConfigProvider` is not being used in the method. Should it be passed along somehow or just removed as a parameter?

I think it does not need to be passed along as it is already part of the `VideoComponentViewModel`... so perhaps it should just be removed

RevenueCatUI/Templates/V2/Components/Video/VideoComponentViewModel.swift
```diff
@@ -93,7 +91,18 @@ class VideoComponentViewModel {
 
         apply(style)
     }
+}
 
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+extension VideoComponentViewModel: Hashable {
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(component)
+        hasher.combine(imageSource)
```
ajpallares: Nit: I think `imageSource` is not needed here since it's actually the `component.fallbackSource` (so it's already considered in the `component`'s hash. 
That said, we might want to leave it here counting for the hash in case we change its implementation in the future? (i.e. imageSource coming from somewhere else?)

RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
```diff
@@ -82,15 +84,39 @@ fileprivate extension View {
...
+                        HStack { Spacer() }
+                        VStack { Spacer() }
                         if let colorOverlay {
                             colorOverlay
                                 .toView(colorScheme: colorScheme)
                         }
                     }
-                    .frame(maxWidth: size?.width, maxHeight: size?.height)
+                    .edgesIgnoringSafeArea(.all)
                 }
```
JZDesign, line 95: Turns out this did not always render correctly before. It was inset.

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
```diff
@@ -161,12 +148,13 @@ struct VideoComponentView: View {
                         stagedURL.publisher
                             .eraseToAnyPublisher()
                             .removeDuplicates()
-                            .debounce(for: 0.300, scheduler: RunLoop.main)
```
JZDesign, line 164: This worked well in the original context, but not so well in backgrounds 🤷‍♂️  So I changed it

RevenueCatUI/Templates/V2/Components/Video/VideoComponentView.swift
```diff
@@ -69,7 +56,7 @@ struct VideoComponentView: View {
                         if let imageSource, let imageViewModel = try? ImageComponentViewModel(
                             localizationProvider: viewModel.localizationProvider,
                             uiConfigProvider: viewModel.uiConfigProvider,
-                            component: .init(source: imageSource)
+                            component: .init(source: imageSource, fitMode: style.contentMode == .fill ? .fill : .fit)
```
JZDesign, line 72: Important detail I missed before … 😬

RevenueCatUI/Templates/V2/Components/Video/VideoComponentViewModel.swift
```diff
@@ -93,7 +91,17 @@ class VideoComponentViewModel {
...
 
+@available(iOS 15.0, macOS 12.0, tvOS 15.0, watchOS 8.0, *)
+extension VideoComponentViewModel: Hashable {
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(component)
+    }
+
+    static func == (lhs: VideoComponentViewModel, rhs: VideoComponentViewModel) -> Bool {
+        lhs.component == rhs.component && lhs.imageSource == rhs.imageSource
+    }
```
JZDesign, line 106:
```suggestion
    static func == (lhs: VideoComponentViewModel, rhs: VideoComponentViewModel) -> Bool {
        lhs.component == rhs.component
    }
```

RevenueCatUI/Templates/V2/Components/Video/VideoComponentViewModel.swift
```diff
@@ -32,31 +32,29 @@ class VideoComponentViewModel {
...
         @ViewBuilder apply: @escaping (VideoComponentStyle) -> some View
     ) -> some View {
         let localizedPartial = LocalizedVideoPartial.buildPartial(
             state: state,
             condition: condition,
-            isEligibleForIntroOffer: isEligibleForIntroOffer,
-            isEligibleForPromoOffer: isEligibleForPromoOffer,
+            // Intro and Promo offers do not impact this component
+            isEligibleForIntroOffer: false,
+            isEligibleForPromoOffer: false,
```
joshdholtz: All of the other components support this so I think I'd like for this one too as well 😇 We will be opening up "property overrides" soon which will allow overriding any property on any component with any condition and some of the conditions could be these ☝️ 

Like... somebody could have a different video for intro offer maybe 🤷‍♂️

RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
joshdholtz: I think I saw some recent issues about this? 🤔 So maybe this will fix that!

RevenueCatUI/Templates/V2/ViewHelpers/BackgroundStyle.swift
joshdholtz: This totally feels like a thing I did during a refactor 😅

RevenueCatUI/Templates/V2/ViewModelHelpers/PresentedPartials.swift
```diff
@@ -133,7 +133,7 @@ extension Array {
         P: PresentedPartial
     >(
         convert: (T) throws -> P
-    ) throws -> PresentedOverrides<P>
+    ) rethrows -> PresentedOverrides<P>
```
joshdholtz, line 136: TIL

PR #5707, opened 2025-10-24
Use cached offerings on network errors

ajpallares:
So far, if get offerings failed, the disk cache would only be returned if the failure was due a server down error (status code 5XX). 

With this PR, by default any network error will fall back to using the local disk cache, except non-500 status codes.

Some examples:
* JSON decoding errors, DNS errors, etc. now fall back to the local disc cache (new for this PR)
* 5XX status code responses still falling back to the local disk cache
* 4XX status code responses still throw the error and do not fall back to the local disk cache

Android equivalent: https://github.com/RevenueCat/purchases-android/pull/2772

### DIFF
--- a/Sources/Misc/DangerousSettings.swift
+++ b/Sources/Misc/DangerousSettings.swift
@@ -110,8 +110,8 @@ import Foundation
     /**
      * Used to initialize the SDK in UI preview mode.
      *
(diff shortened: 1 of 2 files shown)


ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

ajpallares:
@RCGitBot please test

PR #5708, opened 2025-10-24
Fix image shadow issue

JZDesign:
We have an image shadow issue on transparent images. The custom shadow modifier will shade the bounding box, including the transparent pixels. The SwiftUI modifier will only shade the non-transparent pixels.


JZDesign:
The snapshot tests for images changed intentionally here. The old way did not show under transparent pixels but this version does on purpose.

PR #5709, opened 2025-10-27
[SDK-4153] Use static fallback urls

ajpallares:
### Motivation
The signature for requests on the fallback server is done only for static endpoints, not supporting dynamic paths like the GET offerings one (`/v1/subscribers/APP_USER_ID/offerings`).

### Description
Changes the fallback urls into static ones to support signature verification.

In particular:
* The fallback url for GET offerings passes to use the `v1/offerings` path (previously it was the same relative path as for the main server: `/v1/subscribers/APP_USER_ID/offerings`)
* The fallback url for GET product entitlement mapping is maintained to use the `v1/product_entitlement_mapping` relative path.


ajpallares:
@RCGitBot please test

PR #5711, opened 2025-10-27
Adapt integration tests to changes in project

ajpallares:
### Motivation
We've made some changes in the project used for the integration tests. Mainly, setting the default paywall to be a v2 paywalls and leave a v1 paywall in an `alternate_offering`.

### Description
This PR adapts the tests to the changes in this project

### Note
There will be a follow-up PR making similar changes for the Load Shedder integration tests, as we did some similar changes to the project used for those tests. 
The changes are not immediately reflected in the Load Shedder responses, so I'm separating the changes in two different PRs not to block other PRs with failing tests


ajpallares:
@RCGitBot please test

ajpallares:
> Nice! We'll be adding new v2 integration tests in a separate PR right?

Yes @rickvdl. The idea for this PR is to fix whatever tests that broke after we changed the Integration tests project. We can add more integration tests for v2 paywalls in a follow-up PR 👍

PR #5712, opened 2025-10-28
Automated E2E Tests on all backend environments

rickvdl:
### Checklist

### Description

Implemented automated E2E testing using Maestro on all backend environments 
- production
- load shedder primary
- load shedder secondary
- fallback domain

They run two simple flows (for now) on a daily schedule (4 AM UTC) against all backend environments and ensure the following;
- offerings can be retrieved
- a paywall can be displayed
- a purchase can be made successfully from that paywall (through StoreKit testing)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -519,7 +519,7 @@ commands:
       proxy_url_host:
         type: string
         default: ""
-        description: "Optional proxy URL host (e.g., 'fortress-us-east-2.revenuecat.com')"
+        description: "Optional proxy URL host (e.g., 'proxy.revenuecat.com')"
     steps:
       - run:
           name: Create XCConfig file
--- a/fastlane/Fastfile
+++ b/fastlane/Fastfile
@@ -818,13 +818,13 @@ platform :ios do
     end
 
     # note: it's called "run" tests but these tests only need to build to run
-    workspace_path = 'RevenueCat-Tuist.xcworkspace'
-    unless File.directory?(workspace_path)
-      UI.important("#{workspace_path} not found. Run `tuist generate --no-open` before invoking this lane.")
+    xcworkspace = 'RevenueCat-Tuist.xcworkspace'
+    unless File.directory?(xcworkspace)
+      UI.important("#{xcworkspace} not found. Run `tuist generate --no-open` before invoking this lane.")
     end
 
     xcodebuild(
-      workspace: "RevenueCat-Tuist.xcworkspace",
+      workspace: xcworkspace,
       scheme: "Maestro",
       sdk: "iphonesimulator",
       configuration: "Debug",
--- a/Local.xcconfig.SAMPLE
+++ b/Local.xcconfig.SAMPLE
@@ -38,3 +38,7 @@
 // REVENUECAT_PROXY_URL_HOST = localhost:9000
 
 // Because two slashes are interpreted as a comment in this file, the scheme and host need to be set separately
+
+// The following configuration options can be used when running the E2E tests through the Tuist generated Maestro project
+// REVENUECAT_FORCE_SERVER_ERROR_STRATEGY = primary_backend_down / never
+// REVENUECAT_API_HOST = custom-api-host.revenuecat.com


rickvdl:
@RCGitBot please test

rickvdl:
Once this is merged tests will start to fail, because the load shedder and fallback backend environments are not returning the right data yet.

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

Examples/rc-maestro/rc-maestro/Sources/rc-maestro/RcMaestroApp.swift
```diff
@@ -1,18 +1,74 @@
...
+    
+    /*
+     Parses the launch argument with the e2e test flow to run
+     */
+    fileprivate var e2eTestFlow: E2ETestFlow? {
+        guard let string = UserDefaults.standard.dictionaryRepresentation()["e2e_test_flow"] as? String else {
+            return nil
         }
+        
+        return E2ETestFlow(rawValue: string) ?? nil
```
JZDesign: This returns an optional already right?

```suggestion
        return E2ETestFlow(rawValue: string)
```

Examples/rc-maestro/rc-maestro/Sources/rc-maestro/e2e-test-flows/subscribe-from-v2-paywall.swift
```diff
@@ -0,0 +1,33 @@
...
+//  Created by Rick van der Linden on 30/10/2025.
+//  Copyright © 2025 RevenueCat, Inc. All rights reserved.
+//
+
+import SwiftUI
+import RevenueCat
+import RevenueCatUI
+
+extension E2ETestFlowView {
+    struct SubscriberFromV2Paywall: View {
```
JayShortway:
```suggestion
    struct SubscribeFromV2Paywall: View {
```

Examples/rc-maestro/rc-maestro/Sources/rc-maestro/e2e-test-flows/subscribe-from-v1-paywall.swift
```diff
@@ -0,0 +1,46 @@
...
+//  Created by Rick van der Linden on 30/10/2025.
+//  Copyright © 2025 RevenueCat, Inc. All rights reserved.
+//
+
+import SwiftUI
+import RevenueCat
+import RevenueCatUI
+
+extension E2ETestFlowView {
+    struct SubscriberFromV1Paywall: View {
```
JayShortway:
```suggestion
    struct SubscribeFromV1Paywall: View {
```

rickvdl:
@RCGitBot please test

.circleci/config.yml
```diff
@@ -413,13 +422,22 @@ commands:
 
   tuist-generate-workspace:
     description: "Install Tuist and generate workspace"
+    parameters:
+      target:
```
ajpallares: small nit: instead of `target`, we should probably use the same terminology as tuist. They call this `query` (see https://docs.tuist.dev/en/cli/generate#query)

ajpallares: In particular, because we can specify more than one target, or a specific tag. For example:

```
tuist generate tag:RevenueCatTests
tuist generate UnitTests BackendIntegrationTests`
```

.circleci/config.yml
```diff
@@ -480,6 +498,50 @@ commands:
...
+        default: ""
+        description: "Optional API host for testing (e.g., 'api-staging.revenuecat.com')"
+      proxy_url_scheme:
+        type: string
+        default: ""
+        description: "Optional proxy URL scheme (e.g., 'https')"
+      proxy_url_host:
+        type: string
+        default: ""
+        description: "Optional proxy URL host (e.g., 'fortress-us-east-2.revenuecat.com')"
```
ajpallares: It's great to have this parameter, but the example url given here for fortress 2 could be misleading I think, as in reality it's being passed as `api_host` and not as `proxy_url_host`. We could make up some proxy url for the description

.circleci/config.yml
```diff
@@ -1135,6 +1197,64 @@ jobs:
...
+  run-maestro-e2e-tests:
+    parameters:
+      backend_environment:
+        type: enum
+        enum: ["production", "load_shedder_us_east_1", "load_shedder_us_east_2", "fallback"]
+    executor:
+      name: macos-executor
+      xcode_version: "26.0.1"
+    steps:
+      - macos/preboot-simulator:
```
ajpallares, line 1209: Nice touch here! Let's see if this improves the timeouts we're getting sometimes!

.circleci/config.yml
```diff
@@ -1135,6 +1197,64 @@ jobs:
...
+          version: "18.6"
```
ajpallares, line 1210: Why not iOS 26 to run the tests?

.circleci/config.yml
```diff
@@ -1135,6 +1197,64 @@ jobs:
...
+                api_key: ${RC_E2E_TEST_API_KEY_PRODUCTION_APP_STORE}
+      - revenuecat/install-maestro
+      - tuist-generate-workspace:
+          query: "Maestro"
+      - run:
+          name: Run e2e tests (<< parameters.backend_environment >>)
+          command: bundle exec fastlane run_maestro_e2e_tests
+          no_output_timeout: 15m
+        
+        # todo: implement error / success notifications through Slack
```
ajpallares, line 1251: I think, initially, I would post the results to #feed-circleci-ios-failures for now. It's good to have some initial visibility from outside the CI, at least of the failures. WDYT?

.circleci/config.yml
```diff
@@ -1663,6 +1783,14 @@ workflows:
             - slack-secrets
       - integration-tests-all
 
+  daily-maestro-e2e-tests:
+    when:
+      and:
+        - equal: [scheduled_pipeline, << pipeline.trigger_source >>]
+        - equal: ["maestro_e2e_tests", << pipeline.schedule.name >>]
+    jobs:
+      - <<: *maestro-e2e-test-job
```
ajpallares, line 1792: I'm curious. What's this syntax to run the job? Does it mean running the job with all the matrix combinations?

Examples/rc-maestro/Project.swift
```diff
@@ -21,7 +21,9 @@ let project = Project(
                     ],
                     "REVENUECAT_API_KEY": "$(REVENUECAT_API_KEY)",
                     "REVENUECAT_PROXY_URL_SCHEME": "$(REVENUECAT_PROXY_URL_SCHEME)",
-                    "REVENUECAT_PROXY_URL_HOST": "$(REVENUECAT_PROXY_URL_HOST)"
+                    "REVENUECAT_PROXY_URL_HOST": "$(REVENUECAT_PROXY_URL_HOST)",
+                    "REVENUECAT_FORCE_SERVER_ERROR_STRATEGY": "$(REVENUECAT_FORCE_SERVER_ERROR_STRATEGY)",
+                    "REVENUECAT_API_HOST": "$(REVENUECAT_API_HOST)"
```
ajpallares, line 26: Since we're adding these new xcconfig keys, perhaps it would be good to update https://github.com/RevenueCat/purchases-ios/blob/main/Local.xcconfig.SAMPLE with the new keys and their documentation (specify their possible values and mention that they are only meant for Maestro E2E test through tuist). I think it could make things easier for people trying to change these keys for running Maestro tests locally

fastlane/actions/setup_simulator_storekit_config.rb
```diff
@@ -0,0 +1,150 @@
...
+        "Installs a StoreKit configuration file in the correct location for a simulator app to use during testing"
+      end
+
+      def self.details
+        "This action copies a StoreKit configuration file to the Octane directory structure " \
+        "within the iOS simulator, allowing the app to use local StoreKit testing without " \
+        "connecting to the App Store. The action finds the correct app group container by " \
+        "reading .com.apple.mobile_container_manager.metadata.plist and checking for " \
+        "MCMMetadataIdentifier == 'group.com.apple.storekit', creates the Documents/Persistence/Octane " \
+        "folder if needed, and places the configuration file at: " \
```
ajpallares, line 99: This is amazingly useful! Great job on this!
Maybe we can eventually move this to the fastlane plugin so that it can be shared with the hybrids

fastlane/Fastfile
```diff
@@ -802,6 +802,56 @@ platform :ios do
...
+    end
+
+    # note: it's called "run" tests but these tests only need to build to run
+    workspace_path = 'RevenueCat-Tuist.xcworkspace'
+    unless File.directory?(workspace_path)
+      UI.important("#{workspace_path} not found. Run `tuist generate --no-open` before invoking this lane.")
+    end
+
+    xcodebuild(
+      workspace: "RevenueCat-Tuist.xcworkspace",
```
tonidero: Should this use the variable above? Not sure since it's named "path" but it's only the name of the workspace though...
```suggestion
      workspace: workspace_path
```

Examples/rc-maestro/rc-maestro/Sources/rc-maestro/RcMaestroApp.swift
```diff
@@ -1,18 +1,74 @@
...
+
+enum E2ETestFlow: String {
+    case subscribeFromV1Paywall = "subscribe_from_v1_paywall"
+    case subscribeFromV2Paywall = "subscribe_from_v2_paywall"
+    
+    @ViewBuilder
+    var view: some View {
+        switch self {
+        case .subscribeFromV1Paywall:
+            E2ETestFlowView.SubscribeFromV1Paywall()
```
tonidero, line 67: I'm wondering, should we just have a single view with 2 different buttons? While this makes sure the UI is entirely different to avoid accidents of pressing where we shouldn't, I think it's more complex... Not a blocker though.

Examples/rc-maestro/maestro/e2e_tests/subscribe_from_v2_paywall.yaml
```diff
@@ -0,0 +1,21 @@
...
+- tapOn:
+    text: "Yearly"
+- tapOn:
+    text: "Continue"
+- takeScreenshot: "subscribe_from_v2_paywall - IAP Sheet"
+- tapOn:
+    text: "Subscribe"
+- takeScreenshot: "subscribe_from_v2_paywall - IAP Confirmation"
+- tapOn:
+    text: "OK"
```
tonidero, line 21: Might be great to have a way to visualize "active_entitlements" in the UI, so we can verify the entitlements/purchase went through after purchasing. But this can be added later as well :)

tonidero (approved):
Amazing work!! Really love this 🫶 Left a few non-blocker comments.

Examples/rc-maestro/maestro/e2e_tests/subscribe_from_v2_paywall.yaml
ajpallares: Right! I forgot to mention this. But 100% agree. In addition to active entitlements, we should check the source of the Customer Info (see https://github.com/RevenueCat/purchases-ios/pull/5737). But I also think we can add this in a follow-up PR 💪

.circleci/config.yml
rickvdl: Good one, updated!

.circleci/config.yml
rickvdl: No strong opinion, I figured 18.6 is probably a tiny bit more stable, but happy to updated it to 26 :)

.circleci/config.yml
rickvdl: Yes agree, but I figured since I'll probably update this in a day or two with the proper reporting let's leave it for now?

.circleci/config.yml
rickvdl: Yes so it should run daily, as well as when all tests are triggered. And indeed it will run all matrix combinations (just the 4 backend environments)

.circleci/config.yml
ajpallares: Oh right. Yeah, I understand. Yeah, ok. Let's leave it in iOS 18. Once we implement the workaround (for the known issues [here](https://circleci.com/changelog/xcode-26-1-available/)) to boot up the iOS 26 simulator without errors, we can think of moving these to iOS 26 👍

.circleci/config.yml
ajpallares: I'm ok with it 👍

.circleci/config.yml
ajpallares: Nice! I was mostly asking about the `- <<: *` syntax (which I didn't know)

fastlane/actions/setup_simulator_storekit_config.rb
rickvdl: 🙌  Yes for sure! As soon as we need it in one of the Hybrids we can take it out of the repo and move it to the Fastlane repo :)

Examples/rc-maestro/Project.swift
rickvdl: Good one, done!

.circleci/config.yml
rickvdl: Aha! Same, also recently discovered it. I believe they are called [YAML anchors](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/) and more or less an aliasing feature of YAML, which saves you from repeating the same code :)

fastlane/Fastfile
rickvdl: Good catch, updated & renamed the variable as well!

Examples/rc-maestro/rc-maestro/Sources/rc-maestro/RcMaestroApp.swift
rickvdl: Good question, I think in the (near) future we'll add more to this screen, for example this week I'll add the entitlement and source assertions. Although it would probably fit on one screen in this case I'd like to keep them fully separated to avoid breaking other E2E flows when adding / moving things around. In this case they're super similar so I get your point, but wanted to define the structure for the future :)

Examples/rc-maestro/maestro/e2e_tests/subscribe_from_v2_paywall.yaml
rickvdl: Yep, fully agree! Will be working on that right after

rickvdl:
@RCGitBot please test

rickvdl:
@RCGitBot please test

PR #5713, opened 2025-10-28
Add Load shedder integration tests for Get Customer Info

ajpallares:
Adds a couple of Load Shedder integration tests:
* A test that checks that the Get Customer Info response is computed offline before there's any purchase.
* A test that checks that Get Customer Info after a successful purchase comes from the load shedder and contains the expected entitlement


ajpallares:
@RCGitBot please test

Tests/BackendIntegrationTests/LoadShedderIntegrationTests.swift
```diff
@@ -110,6 +110,12 @@ class BaseLoadShedderStoreKitIntegrationTests: BaseStoreKitIntegrationTests {
         expect(receivedOfferings.contents.originalSource) == .loadShedder
     }
 
+    func testGetCustomerInfoIsOfflineComputedBeforeAnyPurchase() async throws {
+        // Get CustomerInfo from Load Shedder fails to find the subscriber info before any purchase is made
+        let offlineComputedCustomerInfo = try await self.purchases.customerInfo()
```
tonidero, line 115: I'm guessing this also waits for the product_entitlement_mapping response to be received right? (otherwise it would not be able to enter offline entitlements)

tonidero (approved):
Makes sense! :shipit:

Tests/BackendIntegrationTests/LoadShedderIntegrationTests.swift
ajpallares: The product_entitlement_mapping response gets cached from the SDK `configure` that happens during the test `setUp`. Then for the offline customer info, the product entitlement mapping is gotten from the `DeviceCache` here https://github.com/RevenueCat/purchases-ios/blob/926dabd1178ce690ebd0b816327f98dcfd9caa3e/Sources/OfflineEntitlements/OfflineCustomerInfoCreator.swift#L86

where `self.productEntitlementMappingFetcher` is the `DeviceCache`.
image

So, in a way, this test implicitly assumes that Get product entitlement mapping succeeded in `configure`. 

But I think this is fine as we're talking about integration tests.

